Issue #1017, opened 2016-04-17
PMP::isotropic_remeshing silently failing


kekkooo:
Hi,
I have compiled CGAL 4.8 from source ( release download , not git pull )
I'm using the new isotropic_remeshing function on a Polyhedron_3 defined as 
typedef CGAL::Simple_cartesian<double> K;
typedef CGAL::Polyhedron_3<K, CGAL::Polyhedron_items_with_id_3>     Polyhedron_3;
Polyhedron_3 Poly; 
the polyhedron stores the convex hull of a point set, obtained using convex_hull_3.
when calling  PMP::isotropic_remeshing( faces(Poly), target, Poly );
the program stops without error messages or assert failures, I tried also to compile CGAL in debug mode and still there are no information about what is happening.

Thanks

afabri:
please put a program and data on a gist.

andeas

On 17/04/2016 16:24, Francesco Usai wrote:
> Hi,
> I have compiled CGAL 4.8 from source ( release download , not git pull )
> I'm using the new isotropic_remeshing function on a Polyhedron_3 defined as
> typedef CGAL::Simple_cartesian K;
> typedef CGAL::Polyhedron_3 Polyhedron_3;
> Polyhedron_3 Poly;
> the polyhedron stores the convex hull of a point set, obtained using
> convex_hull_3.
> when calling PMP::isotropic_remeshing( faces(Poly), target, Poly );
> the program stops without error messages or assert failures, I tried
> also to compile CGAL in debug mode and still there are no information
> about what is happening.
>
> Thanks
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/CGAL/cgal/issues/1017>
>

-- 
Andreas Fabri, PhD
Chief Officer, GeometryFactory
Editor, The CGAL Project

phone: +33.492.954.912    skype: andreas.fabri

kekkooo:
https://gist.github.com/kekkooo/c61c97b89b348c22a63d8c2f98cee5f8

(first time I use gist, I hope it is done properly )

janetournois:
I proposed the above pull request to try avoiding this to happen again.
Thanks @kekkooo for the report

PR #1038, opened 2016-04-22
PMP : deal with uninitialized face_index map in border_halfedges

janetournois:
In `PMP::border_halfedges()` (and as a matter of facts `PMP::isotropic_remeshing()`,
one can provide a `face_index_map` as named parameter.

If it's not provided by the user, the internal `face_index` map (if it exists) is used.
It can happen that this internal property map exists, but is not initialized. It is for example the case for a `Polyhedron_with_items_3` for which the user does not set the face ids.

This PR
- documents the fact that the internal face index map, if it exists, should be initialized
- adds a warning (via `std::cerr`) before the face index map is used, and so before the crash happens, to help the user find the problem

See e.g. the issue #1017

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -24,6 +24,7 @@
 #include <boost/graph/graph_traits.hpp>
 #include <boost/foreach.hpp>
 #include <boost/type_traits/is_same.hpp>
+#include <boost/range/size.hpp>
 
 #include <CGAL/Polygon_mesh_processing/internal/named_function_params.h>
 #include <CGAL/Polygon_mesh_processing/internal/named_params_helper.h>
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -24,7 +24,6 @@
 #include <boost/graph/graph_traits.hpp>
 #include <boost/foreach.hpp>
 #include <boost/type_traits/is_same.hpp>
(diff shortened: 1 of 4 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -157,14 +159,26 @@ namespace Polygon_mesh_processing {
...
     //face index map given as a named parameter, or as an internal property map
     FIMap fim = choose_param(get_param(np, face_index),
                              get(CGAL::face_index, pmesh));
+
+    //make a minimal check that it's properly initialized :
+    //if the 2 first faces have the same id, we know the property map is not initialized
+    typename boost::range_iterator<const FaceRange>::type it = boost::const_begin(faces);
+    if (get(fim, *it++) == get(fim, *it))
+    {
+      std::cerr << "WARNING : the internal property map for CGAL::face_index_t" << std::endl
```
sloriot: is the property map always the internal one?

janetournois: i thought yes, but no
i'll make a new commit
thanks

janetournois:
@sloriot now it is!

janetournois:
from my side, it's ready for testing

lrineau:
Will be tested tonight in CGAL-4.9-Ic-29, with 12 other PRs. I have verified that at least the polyhedron demo is not broken.

lrineau:
Well, I had to push the commit https://github.com/CGAL/cgal/pull/1038/commits/64ea03f9ec26ffcb0d651ea10bacfe4abb89f622 for it to compile on my machine. It is pushed here, and in `integration`.

janetournois:
this PR had errors in the testsuite for "old" boost versions. 
It should be fixed now

PR #1046, opened 2016-04-27
shorten include directive and use compiler include path instead

sloriot:
This is a nice to have feature for some users

### DIFF
--- a/Installation/src/CMakeLists.txt
+++ b/Installation/src/CMakeLists.txt
@@ -20,9 +20,11 @@ function (collect_cgal_library LIBRARY_NAME ADDITIONAL_FILES)
     # file in the new all_files.cpp because .cpp files would be included
     # twice, and that breaks the one-definition rule (ODR).
     if(NOT source MATCHES ".*/all_files.cpp")
-      file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/all_files.cpp "\#include \"${source}\"\n")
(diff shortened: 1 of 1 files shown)


sloriot:
@lrineau do you see any possible issue?

lrineau:
No. With `include_directories(${CMAKE_SOURCE_DIR}/src/${LIBRARY_NAME})`, it should work everywhere.

sloriot:
This is not working in branch build: the command should be `include_directories(${CMAKE_SOURCE_DIR}/${LIBRARY_NAME}/src/${LIBRARY_NAME})` which would break the flat layout behavior and would not be sufficient as the `.cpp` files for the lib CGAL are spread in different packages. The header only feature will anyway make this useless so I close this PR.

lrineau:
(For the moment the header-only feature is not ready.)

To allow its use in the flat and Git layouts, that could be:
```cmake
if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/${LIBRARY_NAME}/src/${LIBRARY_NAME})
  include_directories(${CMAKE_SOURCE_DIR}/${LIBRARY_NAME}/src/${LIBRARY_NAME})
endif()
if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/src/${LIBRARY_NAME})
  include_directories(${CMAKE_SOURCE_DIR}/src/${LIBRARY_NAME}),
endif()
```

With a `file(GLOB)` it can be adapted to multiple packages.

PR #1050, opened 2016-04-29
Fix get problems

maxGimeno:
This PR puts all get and put functions of properties_Surface_mesh.h in namespace CGAL.


maxGimeno:
replace by #1063

PR #1051, opened 2016-04-29
Polyhedron_demo : Warning when save fails

maxGimeno:
Add a warning box when the user forgets to put an extension when saving a file.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1484,15 +1484,23 @@ void MainWindow::on_actionSaveAs_triggered()
 
 void MainWindow::save(QString filename, CGAL::Three::Scene_item* item) {
   QFileInfo fileinfo(filename);
-
+  bool saved = false;
   Q_FOREACH(CGAL::Three::Polyhedron_demo_io_plugin_interface* plugin, io_plugins) {
     if(  plugin->canSave(item) &&
         file_matches_filter(plugin->saveNameFilters(),filename) )
     {
       if(plugin->save(item, fileinfo))
+      {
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1484,14 +1484,20 @@ void MainWindow::on_actionSaveAs_triggered()
 
 void MainWindow::save(QString filename, CGAL::Three::Scene_item* item) {
   QFileInfo fileinfo(filename);
-
+  bool saved = false;
   Q_FOREACH(CGAL::Three::Polyhedron_demo_io_plugin_interface* plugin, io_plugins) {
     if(  plugin->canSave(item) &&
         file_matches_filter(plugin->saveNameFilters(),filename) )
     {
       if(plugin->save(item, fileinfo))
```
sloriot: Looking at the diff I guess here you are missing some `{}`

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1484,14 +1484,20 @@ void MainWindow::on_actionSaveAs_triggered()
...
+        saved = true;
         break;
     }
+    if(!saved)
+      QMessageBox::warning(this,
+                           tr("Cannot save"),
+                           tr("The selected object %1 cannot be saved.")
```
sloriot: I would replace `cannot be saved` by `was not saved`

sloriot:
Tested locally, works great

sloriot:
@maxGimeno I have amended the last commit (wrong scope for the warning message). You need to remove your local branch or hard-reset it

PR #1055, opened 2016-05-02
Store the indices of source and target of a selected edge

afabri:
After isotropic remeshing of a mesh with selected edges, saving and loading of the mesh and the selection was not correct, as after the loading the edges are not in the same order as in the remeshed mesh.


sloriot:
Locally tested

PR #1056, opened 2016-05-02
At least allow to write polyhedral surfaces into .obj files

afabri:


sloriot:
This is a bug fix since the obj reading possibility was added into the same plugin class as the OFF read/write and since `canSave()` is global to the plugin the saving to OBJ was proposed but the output was actually an off.

sloriot:
Tested locally

PR #1060, opened 2016-05-03
Polyhedron demo : fix the display of the fixed plane for the deformation plugin

janetournois:
This PR fixes the display of the fixed plane for the `edit_polyhedron_plugin`

This problem was happening only on windows 8 (at least msvc2013), but it spotted an actual bug.

cc @maxGimeno , thanks!


sloriot:
Also merged in master

PR #1066, opened 2016-05-05
Cone spanners 2 new package uws

ws4u:
See https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Cone_based_spanners


sloriot:
Will be tested in CGAL-4.9-Ic-41

ws4u:
@afabri and @sloriot : in the testsuite report of CGAL-4.9-Ic-41, there is compilation error of theta_exact.cpp. The error seems a problem from BGL. But I'm not sure why it happens to theta_exact.cpp, but not yao_exact.cpp, as these two cpps have the same structure. I tried to remove "#include <boost/graph/adjacency_list.hpp>" from both cpps. Let's see whether this can fix the problem. Or you have other suggestions?

sloriot:
I could reproduce the compilation error with boost 1.60 and std=c++11. It might be an issue with boost. I'll have a quick deeper look and let you know.

sloriot:
I also have the problem with boost 1.58 and actually the header `boost/graph/detail/adjacency_list.hpp` has not been updated since [2 years](https://github.com/boostorg/graph/blob/develop/include/boost/graph/detail/adjacency_list.hpp).

sloriot:
The difference between `theta_exact.cpp` and `yao_exact.cpp` is the graph type (`boost::directedS` vs. `boost::undirectedS`)

sloriot:
I've produce a minimal example that is not using CGAL and does not compile. I've posted it on the boost mailing list.

sloriot:
Link in the boost [archive](http://lists.boost.org/Archives/boost/2016/05/229458.php)

afabri:
So we might put a comment that it does not work for directed graph, with a link to the bug report.
And maybe the fix makes it into the next release of boost, which is probably out before August, so that we then can write that with a directed graph one needs boost version X.Y

ws4u:
@afabri and @sloriot : thanks. Please let me know if I can do something here.

afabri:
As it concerns only C++11  you should put the directed or undirected in an `#ifdef`   in the example
and put a comment in the example with a link to the bug report.

sloriot:
you can use `#ifdef CGAL_CXX11` (which is a bit too much but shorter than `CGAL_CFG_NO_CPP0X_RVALUE_REFERENCE`)

ws4u:
I have put a comment in the example with a link to the bug report. Please check whether it is appropriate. I'm not sure what you want to achieve with  `#ifdef CGAL_CXX11`, so I haven't done it. 
I've looked at the boost pull request to fix the bug, which was dated 17 Nov 2014. I'm not sure why it hasn't been merged yet.

ws4u:
In the user manual, I've added a note about this bug from BGL, and explained a little about this workaround. Please check whether it is appropriate. Welcome to improve.

sloriot:
- There is one [warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-52/Cone_spanners_2_Examples/TestReport_jtournois_x64_Cygwin-Windows8_MSVC2015-Release-64bits.gz) remaining on windows.
- `test_merge_of_branch` does not highlight any problem with the branch
- the branch size is a bit large: 842K

ws4u:
@sloriot : thanks for the updates to the code and the doc. For the warning remaining on windows, it seems again from boost, and I'm not sure how to solve it right now. For the branch size, I checked the file sizes under the Cone_spanners_2 directory. There are only three files larger than 20K: Example-Y5.jpg: 57K, t4n20.jpg: 26K, _Plane_Scan_Tree.h: 21K. If needed, I can reduce the sizes of those two jpg files. Pls let me know.

sloriot:
@ws4u I was talking about the size of the branch (including the history of the changes)
I applied the following to the branch (which replaced in each commit the pictures by the last versions):

    git filter-branch --tree-filter \
    'for i in BisectorInThetaGraph.jpg  Example-Y5.jpg  Logo-ConeSpanners.png  pkg-small.png  t4n20.jpg  t4n9exact.jpg  t4n9inexact.jpg; do if [ -e Cone_spanners_2/doc/Cone_spanners_2/fig/$i ]; then cp /tmp/fig/$i Cone_spanners_2/doc/Cone_spanners_2/fig/$i; fi; done'\
    3e490bb5..HEAD  --not cgal/master

Now the branch size is down to 279K while the package is 142K which seems OK to me.
What you need to do right now is to reset your branch as follow:

    git fetch mine
    git checkout Cone_spanners_2-new_package-UWS
    git reset --hard mine/Cone_spanners_2-new_package-UWS

@afabri: about the warning on windows, could you have a quick look to see  if you can find a workaround?

ws4u:
@sloriot : I've reset my branch as you mentioned.

afabri:
I suppressed the warning that we get for VC15, in the CMakeList.txt, as we also get it for the [example](http://www.boost.org/doc/libs/1_61_0/libs/graph/example/dijkstra-example.cpp) coming with the distribution of boost.

sloriot:
I've pushed a few cleaning commits (renamed files and classes to match CGAL conventions, fixed trailing whitespaces and indentation issues).

There is this macro `REALLY_VERBOSE_TREE_STATE_AFTER_EVERY_TREE_UPDATE__SAFE_TO_REMOVE_FOR_PRODUCTION` I'm not sure you want to have in the release.

I also saw some comments in the code that seems to be development comments that should be gone. If you want you can do a global pass on your files if you want to clean them.

Minor point: your first commits were done with the address `w.si@uws.edu.au`. If you want github to recognize you as the author, you must add this address in https://github.com/settings/emails. If you prefer, we can also easily change the email address in the commits (only possible before the merge in master).

ws4u:
I've done a global pass to remove unwanted comments and also improve existing comments.

For my initial commits with the address w.si@uws.edu.au, pls change this email address in the commits, since the address w.si@uws.edu.au should no longer be used due to the domain name change of our university. Thanks.

sloriot:
@ws4u Done. You need to reset --hard your local branch as described above.

ws4u:
@sloriot I've reset --hard my local branch.

sloriot:
@ws4u There are some errors with ICC 2016 but they are probably not specific to you package. I'll merge your branch. Congratulation!

Issue #985, opened 2016-04-05
Polyhedron demo : improvements for point_inside_polyhedron_plugin


janetournois:
There are two improvements that should be made to the `point_inside_polyhedron_plugin`

1. the user needs to select the `polyhedron_item` AND the `point_set_item` that should be considered to select inside/outside/boundary points. When the collection of points to be tested is created from bbox (first option of the widget), it would be better that this point set is automatically considered as the one to be treated.

2. once the points are "selected" (actually only colored) it would be nice to be able to do something with them. I think the simplest/best option is to put the so-called selected vertices into an actual `scene_polyhedron_selection_item` so that we can use all the operations already provided by the `selection_plugin` (create `point_set_item` from selection, invert selection, delete selection, etc)

cc @maxGimeno (for the demo)
cc @sgiraudot (do you have extra suggestions that could be helpful for point set processing?)

sgiraudot:
I am not sure if this is how it works, but if the output points are selected (which should be the case if they are colored in red), then the plugin `Point_set_selection` can be used on the point set item to process the selection (it is for example possible to create a new point set from the selection, invert it, apply point set algorithms only on the selection, etc.).

sloriot:
For the 2nd point, I think everything is already available in the context menu.
For the 1st point, making the plugin selecting the point set generated is the only reasonable solution I see.

janetournois:
for the 2nd point, why not migrating to

janetournois:
we could for example benefit of the "invert selection" button of PR #936

sloriot:
There is already a selection plugin for point set written by @sgiraudot 
You should discuss this together because it seems to me that there is some confusion.

lrineau:
The PR #936 is merged.

janetournois:
The point set selection plugin is exactly what I wanted for 2nd point, I did not know about it. 
Thanks :-)

janetournois:
sorry, i closed too fast and forgot 1st point

PR #1068, opened 2016-05-06
If the color vector was manually set, do not translate color ids

sloriot:
Fixes a bug introduced by #985 when computing sdf values
cc @janetournois


sloriot:
Manually tested

Issue #1024, opened 2016-04-20
CGAL 4.8 compilation failure "error: 'CGAL_assertion' was not declared in this scope"


etrinh:
Hi,

I have a compilation error while compiling CGAL 4.8 on Ubuntu 16.04,.
I figured out, that in my situation, cmake built the all_files.cpp with "CGAL/io.cpp" on the top of the list.
[all_files.zip](https://github.com/CGAL/cgal/files/227635/all_files.zip)

But in CGAL/IO/io.h:378, there is a usage of CGAL_assertion, but the "CGAL/assertions.h" seems never included, is it missing?

Thanks,
Best Regards,
Eric

lrineau:
I can confirm the bug. The fix is indeed the addition of:
```C++
#include <CGAL/assertions.h>
```
in `CGAL-4.8/include/CGAL/IO/io.h`.

lrineau:
The bug is tricky, and had not triggered any compilation error in the CGAL test suite, because that depends on the order of files on the filesystem, that impacts the result of the CMake command `file(GLOB..)`.

lrineau:
Fixed by #1025. Thanks.

PR #1069, opened 2016-05-06
Sort results of file(GLOB ...) to make build deterministic

### LINKED ISSUE #1048: Non-deterministic build process causes random build failures due to missing includes
1. include/CGAL/IO/io.h needs to include \<CGAL/assertions.h\>

2. The content of the all_files.cpp files is not deterministic. For example, the missing include shows up if io.cpp ends up as first include in src/CGAL/all_files.cpp. This makes such problems extremely difficult to debug. In my case it happened consistently on 2 of ~15 architectures in an automated build environment, but never when built manually. Independent of the missing include build results should be deterministic and reproducible, see https://reproducible-builds.org/ .

lrineau:
Patch from Joachim Reichel <reichel@debian.org>, adapted from the flat
layout to the Git layout.

Fix #1048.


sloriot:
I though you'd like to include it in 4.8

lrineau:
I do not like to modify the CMake scripts without a lot of manual testing: the testsuite only tests one use-case, and even one that no user will ever use!

As that is not really a bug (unlike #1025), I prefer to postpone the reproducibility of builds to 4.9.

@joachim-reichel Do you have a way to test CGAL-4.9-dev in advance in the Debian build and test systems?

joachim-reichel:
On 05/06/2016 05:05 PM, Laurent Rineau wrote:
> I do not like to modify the CMake scripts without a lot of manual testing: the testsuite only tests one use-case, and even one that no user will ever use!
> 
> As that is not really a bug (unlike #1025), I prefer to postpone the reproducibility of builds to 4.9.

I see. I'll probably use for the Debian packages at least the patch snippet
regarding src/*/*.cpp since that is definitely needed for reproducible builds.

> @joachim-reichel Do you have a way to test CGAL-4.9-dev in advance in the Debian build and test systems?

I could test beta releases by uploading them to the "experimental" suite. You
probably don't want to make nightly and alpha releases public, right? (And
updating the packaging is a manual process.)

lrineau:
@joachim-reichel: the real compilation issue (#1024) is fixed by #1025, that is merged in `releases/CGAL-4.8-branch`.

If Debian needs deterministic builds (:+1:), then the Debian CGAL package will have to carry the patch of this PR #1069 until CGAL-4.9 is published.

joachim-reichel:
Yes, that's what I'm planning to do.

sloriot:
@lrineau anything you want to test before merging this PR?

lrineau:
I think this PR can be merged in `master`. If there is a surprising bug, we will have time to find it before CGAL-4.9, with all the manual testing that will be done until then.

It is for CGAL-4.8 that I do not want to take that risk, because there cannot be the same amount of testing (in particular there cannot be any beta-testing for bug-fix releases).

PR #1070, opened 2016-05-09
Bilateral smoothing: use put() operator instead of using ref of get()

sgiraudot:
In the _bilateral smoothing_ algorithm, the points and normals are modified by getting a non-const reference through the `get()` operator. Although this is not strictly forbidden by the [Boost API](http://www.boost.org/doc/libs/1_60_0/libs/property_map/doc/ReadablePropertyMap.html), this operation is usually done using the `put()` operator (this is the case in all the other algorithms of _Point set processing_).

More specifically, some property maps might not be able to provide a `get()` operator that returns a non-const reference but still provide a `put()` operator: this PR makes _bilateral smoothing_ compatible with such property maps.

### DIFF
--- a/Point_set_processing_3/include/CGAL/bilateral_smooth_point_set.h
+++ b/Point_set_processing_3/include/CGAL/bilateral_smooth_point_set.h
@@ -570,15 +570,18 @@ bilateral_smooth_point_set(
    for(unsigned int i = 0 ; it != beyond; ++it, ++i)
    {
 #ifdef CGAL_USE_PROPERTY_MAPS_API_V1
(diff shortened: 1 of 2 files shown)


sgiraudot:
I also added a `const` for the return value of `get()` in the _WLOP_ algorithm (in that case, it did not change anything else as the reference was not changed anyway, the `const` was already implicitly true).

sgiraudot:
I do not see any error related to this PR in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-43.shtml).

PR #1074, opened 2016-05-10
Adding less-than operator (<) to Point_d

freud14:
My editor removed useless space a the end of lines. I hope that it is fine. I did not know very well where to put the implementation of the operators. In Kernel_23, all the implementation of the operators is in two header files but it seems that in Kernel_d, it is in the specific object classes so I put the implementation there.

### DIFF
--- a/Kernel_d/doc/Kernel_d/CGAL/Kernel_d/Point_d.h
+++ b/Kernel_d/doc/Kernel_d/CGAL/Kernel_d/Point_d.h
@@ -3,243 +3,252 @@ namespace CGAL {
 /*!
 \ingroup PkgKernelDKernelObjs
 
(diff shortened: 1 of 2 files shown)


sloriot:
could you undo the removal of useless space at of the lines? we usually do not mix clean-up and addition of features (easier to review).
Also the doc and the code modification are not in sync (4 operators added, one documented)

freud14:
For the doc, I was wondering if there is some policy I am not aware of because, for instance, the operators == and != exists (In PointCd and PointHd however) but are not documented. So I deduced that the obvious operator are not documented but the less obvious are, but I was probably wrong.

I'll make a new commit later with the corrections.

Frédérik

sloriot:
Replaced by #1083

PR #483, opened 2015-11-17
Mesh 3 : add gray level image oracle

janetournois:
This PR introduces a new oracle for Mesh_3, able to deal with gray level images.

The corresponding [small feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Mesh_3_gray_images) was pre-approved on October 17 by @lrineau

PR #413, opened 2015-10-09
Surface mesher - Improve support of vtk files as input

janetournois:
This PR updates the CMakeLists to handle recent (>= 6.0) versions of vtk, and update Surface_mesher/demo/Surface_mesher/VTK examples to Qt5 and VTK6.

Also add a menu to the Surface_mesher demo to open a directory (previously one could only open a directory with drag-and-drop)

PR #1075, opened 2016-05-11
Polyhedron demo : Handle gray images

### LINKED ISSUE #774: Polyhedron demo : handle Gray-level images
We would like to be able to display gray-level images in gray level.

This task consists in
- reusing Scene_segmented_image_item and transforming it into Scene_image_item
- adding an option in the popup that opens when loading an image to decide whether the image is segmented or gray-level
- using the volume_planes_plugin to display 3 cut planes of the image

Also, the Surface_mesher demo contains load functions for image file formats that have not been ported to the Polyhedron demo yet

cc @maxGimeno @lrineau @afabri

maxGimeno:
fixes #774

This PR allows the Io_image_plugin to open images as "Gray-level images", which are images that have only volume planes.


sloriot:
@lrineau @janetournois do you approve the modifications?

janetournois:
I will test this PR asap

janetournois:
The image_io_plugin should be able to load file from *.dcm file format. See PR #413 to see how it's done in the Surface_mesher demo.

Also, when loading a labeled image, the volume planes are completely black and we can't see through. Is it what we want?

janetournois:
* when a labeled image is loaded, Mesh_3 does not appear as an option in the available operations
* when we load an image, a dialog asks for the resolution. One of the available resolutions is 1:125. It should probably be 1:128

janetournois:
Update of my above comments
* consider PR #483 to know how to mesh a gray level image
* consider PR #413 to see how to open a dicom group of files in the demo (also done in `Mesh_3/examples/Mesh_3/mesh_3D_gray_vtk_image.cpp` since then)

janetournois:
@maxGimeno is this PR ready for review?

maxGimeno:
@janetournois  Yes it is.

janetournois:
Thanks.

Is it possible to choose the iso-level we want to be meshed? For the moment I can't mesh the skull (from Surface_mesher/examples/Surface_mesher/data) though the image opens properly.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Mesh_3_plugin_cgal_code.cpp
```diff
@@ -123,41 +127,77 @@ Meshing_thread* cgal_code_mesh_3(const Image* pImage,
...
+    return new Meshing_thread(p_mesh_function, p_new_item);
   }
-  if(! polylines.empty()){
-    // Insert edge in domain
-    p_domain->add_features(polylines.begin(), polylines.end());
-    protect_features = true; // so that it will be passed in make_mesh_3
+  else
+  {
+    Gray_Image_mesh_domain* p_domain = new Gray_Image_mesh_domain(*pImage, std::bind1st(std::less<float>(), 3), 0);
+
```
lrineau: The function-object `std::bind1st(std::less<float>(), 3)` hardcodes the iso-value to `3`.

janetournois: this should be turned into parameters available for the user in the `Meshing_dialog.ui`
Those parameters are
- the iso-value (here 3)
- whether the "inside" of the object corresponds to values lower or higher than the iso-value (`std::bind1st(std::less<float>(), .., ...) ` here
- `value_outside`, which is the value (constant, and 0 here) outside the gray-level image.

Those 3 options should be added to the ui inside a `GroupBox`, and made unavailable (gray) for domains other than Gray-level images

lrineau: Instead of graying the box, I would even hide it completely, `box->setVisible(false)`, when the image is not a gray-level image.

janetournois: yes, that's even better

maxGimeno:
If you don't see anything else, I think this PR is ready to be merged

sloriot:
@maxGimeno please check that ed5e06f is correct.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Mesh_3_plugin_cgal_code.cpp
```diff
@@ -123,41 +130,114 @@ Meshing_thread* cgal_code_mesh_3(const Image* pImage,
...
+      param.facet_approx = facet_approx;
+      param.tet_sizing = tet_sizing;
+      param.tet_shape = tet_shape;
+      param.manifold = manifold;
+
+      typedef ::Mesh_function<Gray_Image_mesh_more_domain> Mesh_function;
+      Mesh_function* p_mesh_function = new Mesh_function(p_new_item->c3t3(),
+                                                         p_domain, param);
+      return new Meshing_thread(p_mesh_function, p_new_item);
+    }
```
lrineau: This block of code is more of less copy-pasted three time:
 - one after `if(!is_gray)`,
 - one after `else if(inside_is_less)`,
 - and the last one after the last `else`.

Let's work together tomorrow on improving that code.

lrineau:
https://github.com/CGAL/cgal/commit/ed5e06f3416823e76eb6040e6a93a6590926a514 is correct. It was an error not to use that parameter. Bless the warnings!

sloriot:
There are some errors in the testsuite. Not sure it is related to this branch.
However this highlight the fact that the new plugin is not in `cgal_test_with_cmake` file in the polyhedron demo directory

lrineau:
https://github.com/CGAL/cgal/commit/ed5e06f3416823e76eb6040e6a93a6590926a514 is correct. It was an error not to use that parameter. Bless the warnings!

lrineau:
@maxGimeno The compilation errors are here because the absence of VTK libraries is not handled correctly. Maxime, to reproduce that on your machine, add the CMake option `CMAKE_DISABLE_FIND_PACKAGE_VTK`, type Boolean, with the value `TRUE`, and remove from the existing CMake cache all the variables starting with `VTK`. That will instruct CMake to simulate the absence of VTK: the call to `find_package(VTK)` will do nothing.

janetournois:
The `io_image_plugin` generates a seg fault in debug mode. The error happens when `Scene_image_item::initialize_buffers()` tries to access the address of `quads` which is empty.
I don't know if the error is that `quads` is empty, or if there should be something like `if(quads.empty())`, but anyway the demo crashes.

The corresponding code was not written in this PR, but having it working is mandatory to be able to use the `mesh_3_plugin` for images

@maxGimeno

janetournois:
Tested locally, it works for me.
This PR is ready for testing

maxGimeno:
There were errors on machines without VTK. They should be fixed by  d34d7d5.

maxGimeno:
I fixed a warning when VTK is not found.

sloriot:
@lrineau are all functionality of the Mesh_3 demo covered by the polyhedron demo now?

sloriot:
What about the surface_mesher demo?

lrineau:
I do not know everything about the Mesh_3 demo. Jane (@janetournois), do you think every features from the Mesh_3 demo is now converted?

As for the Surface_mesher demo, there are still non-documented features in it, that are not converted to the Polyhedron demo.

And also the possibility to specify multiple iso-values for the meshing. That is not really useful, but I used it to stress the algorithms.

janetournois:
I think it every feature of the Mesh_3 demo is now covered, but let me check.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Image_res_dialog.ui
```diff
@@ -43,27 +31,76 @@
...
-    <widget class="QLabel" name="label_2">
-     <property name="font">
-      <font>
-       <pointsize>11</pointsize>
-       <italic>true</italic>
-      </font>
+   <item row="2" column="0" colspan="2">
+    <widget class="QGroupBox" name="groupBox">
+     <property name="title">
+      <string>GroupBox</string>
```
lrineau: I want to comment on the modification of this dialog in this PR.

I wonder why there is a group box, in this dialog. Group boxes are used to group widgets, and separate them from other widgets in UI. But in this dialog there is almost nothing to separate from! That widget can use just a `QWidget`, that is shown or not depending on the content of the `QComboBox imageType`.

And the group box title should not have been let to the default "GroupBox".

... Anyway, do not modify that. I am working on another pull-request that modifies the Mesh_3 plugin, and I will improve that dialog in the same PR.

PR #1082, opened 2016-05-13
Polyhedron demo : Path selection

### LINKED ISSUE #930: Path Selection
In the selection tool I would like to mark two vertices p and q for the mesh below

selectionpq

in order to add the feature line (which is not sharp enough).  
Note that the shortest path would be the dotted black line.
- We might have the dihedral angle at edges as a weight for the distance.
- The user might select an intermediate vertex where the path from p to q has to pass through.

maxGimeno:
(fixes #930)

This PR adds a feature to the selection tool : the path between vertices selection. It allows the user to select the shortest path between two vertices, and then to edit it by forcing the path to go through selected vertices.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
@@ -83,6 +83,7 @@ public:
     addDockWidget(dock_widget);
 
     connect(ui_widget.Select_all_button,  SIGNAL(clicked()), this, SLOT(on_Select_all_button_clicked()));
+    connect(ui_widget.Add_to_selection_button,  SIGNAL(clicked()), this, SLOT(on_Add_to_selection_button_clicked()));
     connect(ui_widget.Clear_button,  SIGNAL(clicked()), this, SLOT(on_Clear_button_clicked()));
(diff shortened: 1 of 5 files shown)


sloriot:
I compiled the branch to test it:
-  I started picking two points and got a nice polyline. Then I pressed Ctrl+Z and everything disappear. Click another point brings me back the previous points. I add to click `Add to selection` and Ctrl+Z to be back to selection one point only.
- In the .ui, all buttons and list items must be Camel Case to be consistent ( Path Between Vertices, Add to Selection, Select Components, ...)
- I guess there is something to add more than two points but I could not figure out how. Maybe we can add a tooltip?
Otherwise, this is pretty cool!

maxGimeno:
Some elements were forgotten during the merge, like clearing the path when performing a Ctrl+Z or the fact that the instructions label must always be visible. I fixed it, it should be OK now.

sloriot:
Indeed, things are better now.

sloriot:
@sloriot TODO: before merging this PR, the branch needs some rebasing to remove some commits from another branch integrated into master that was rebased before the merge.

sloriot:
Rebase was too complicated. All former modifications were merged into 3ae736a

sloriot:
The testsuite shows error not related to this branch. I locally tested the branch and it is fixes the issues.

PR #1087, opened 2016-05-18
Cast constness away as the function Surface_mesh::properties() is const

### LINKED ISSUE #1080: Surface_mesh: invalid conversion in properties and Property_selector
Dear CGAL,

I cannot seem to use the ``CGAL::Surface_mesh::properties()`` function.

Here is a simple test case:

```cpp

int main() {
  using Point = CGAL::Simple_cartesian<double>::Point_3;
  using Mesh = CGAL::Surface_mesh<Point>;

  Mesh mesh;
  auto props = mesh.properties<Mesh::Face_index>();
  
  return 0;
}
```

That’s what the compiler tells me:

```
In file included from .../cgal-4.8/include/CGAL/Surface_mesh.h:23:0,
                 from tc-sm-properties.cc:2:
.../cgal-4.8/include/CGAL/Surface_mesh/Surface_mesh.h: In instantiation of ‘std::vector<std::basic_string<char> > CGAL::Surface_mesh<K>::properties() const [with I = CGAL::SM_Face_index; P = CGAL::Point_3<CGAL::Simple_cartesian<double> >]’:
tc-sm-properties.cc:9:50:   required from here
.../cgal-4.8/include/CGAL/Surface_mesh/Surface_mesh.h:2366:14: error: invalid conversion from ‘const CGAL::Surface_mesh<CGAL::Point_3<CGAL::Simple_cartesian<double> > >*’ to ‘CGAL::Surface_mesh<CGAL::Point_3<CGAL::Simple_cartesian<double> > >*’ [-fpermissive]
       return Property_selector<I>(this)().properties();
              ^~~~~~~~~~~~~~~~~~~~~~~~~~
.../cgal-4.8/include/CGAL/Surface_mesh/Surface_mesh.h:2287:5: note:   initializing argument 1 of ‘CGAL::Surface_mesh<K>::Property_selector<CGAL::SM_Face_index, dummy>::Property_selector(CGAL::Surface_mesh<K>*) [with bool dummy = true; P = CGAL::Point_3<CGAL::Simple_cartesian<double> >]’
     Property_selector(CGAL::Surface_mesh<P>* m) : m_(m) {}
     ^~~~~~~~~~~~~~~~~
```

Looking at the ``properties()`` definition, we can see that it is defined as a ``const`` function, which makes sense:

```cpp
template<class I>
std::vector<std::string> properties() const
{
  return Property_selector<I>(this)().properties();
}
```

Now declaring the function as ``const`` makes ``this`` a ``const CGAL::Surface_mesh``. But ``Property_selector`` only defines constructors for non-const instances:

```cpp
template<bool dummy>
struct Property_selector<typename CGAL::Surface_mesh<P>::Face_index, dummy> {
  Property_selector(CGAL::Surface_mesh<P>* m) : m_(m) {}
};
```

An ugly workaround would be to ``const_cast`` inside ``properties()`` but it won't solve the problem elsewhere, in addition to being ugly.

Best,

afabri:
This fixes #1080

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
@@ -2363,7 +2363,7 @@ private: //--------------------------------------------------- property handling
     template<class I>
     std::vector<std::string> properties() const
     {
-      return Property_selector<I>(this)().properties();
+      return Property_selector<I>(const_cast<Self*>(this))().properties();
     }
 
     /// returns the property for the string "v:point".
--- a/Surface_mesh/examples/Surface_mesh/sm_properties.cpp
+++ b/Surface_mesh/examples/Surface_mesh/sm_properties.cpp
@@ -53,6 +53,11 @@ int main()
   BOOST_FOREACH( vertex_descriptor vd, m.vertices()) { 
     std::cout << name[vd] << " @ " << location[vd] << std::endl;
   }
+
+  std::vector<std::string> props = m.properties<vertex_descriptor>();
+  BOOST_FOREACH(std::string p, props){
+    std::cout << p << std::endl;
+  }
   
   // delete the string property again
   m.remove_property_map(name);


lrineau:
This is a bug fix. Can you please rebase the branch on CGAL-4.8-branch?

lrineau:
The commands to rebase are:
```shell
git fetch cgal
git rebase cgal/master --onto cgal/releases/CGAL-4.8-branch
```

Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
```diff
@@ -2363,7 +2363,7 @@ class Property_map
     template<class I>
     std::vector<std::string> properties() const
     {
-      return Property_selector<I>(this)().properties();
+      return Property_selector<I>(const_cast<Self*>(this))().properties();
     }
```
sloriot: The vector returned is not a `const&` is it on purpose?

afabri: a bug

lrineau: The code is:
```C++
    // returns a vector of all property names
    std::vector<std::string> properties() const
    {
        std::vector<std::string> names;
        for (unsigned int i=0; i<parrays_.size(); ++i)
            names.push_back(parrays_[i]->name());
        return names;
    }
```
With this code it cannot return a reference.

afabri: you are right. Additionally it is not critical unless you have tiny meshes with many properties and many calls of the function.

sloriot:
Merged in 4.8

PR #1089, opened 2016-05-18
Fixes for Polyhedron_demo

### LINKED ISSUE #1084: Polyhedron demo: polylines loader broken
When loading a `.polylines.txt` file, the polyline item created is bugged:

- Nothing is displayed (it seems the camera goes far away from the scene)
- Bounding box of the object indicates `min(0,0,0) max(0,0,0)`
- The item behaves like a group item: if another object exists when loading the polyline item, it becomes a "member" of the polyline item as if this item were a group. Alternatively, load another item when the selected item is a polyline makes it "member" of the polyline too.

maxGimeno:
Fixes  #1084 

This PR fixes :
- The multi selection that was broken for some time
- The polyline_items that had problems with their BBox
- Re-connects some actions that were broken in plugins not inheriting from the plugin helper.
- The color of items in a group when they are selected.

Plus, after this PR, when a group is added to the scene, no matter the selection, no item will be added as a child, which will avoid problems with several plugins and all the scene_items that recently inherited from Scene_group_item, such as the polylines_item or the c3t3_item.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1069,7 +1069,7 @@ void MainWindow::selectSceneItem(int i)
       proxyModel->mapSelectionFromSource(scene->createSelection(i));
 
     sceneView->selectionModel()->select(s,
-                                        QItemSelectionModel::Select);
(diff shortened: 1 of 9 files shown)


sloriot:
The testsuite shows error not related to this branch. I locally tested the branch and it is fixes the issues.

PR #1086, opened 2016-05-17
PMP isotropic remeshing : avoid creating self intersections near patch boundaries

janetournois:
the collapse step was still creating self-intersections
when it happened in a particular configuration close to patch boundaries

PR #1091, opened 2016-05-19
PMP isotropic remeshing : avoid creating self intersections near patch boundaries

janetournois:
This PR replaces #1086 

the collapse step was still creating self-intersections
when it happened in a particular configuration close to boundary


janetournois:
should be fixed

sloriot:
Tests failing at runtime on various platforms

janetournois:
I'm working on it

janetournois:
should be ok!

PR #1097, opened 2016-05-20
Mesh 3: add an example that protects features of a 3D image

lrineau:
CC @afabri @janetournois (To track my progress.)

For the moment:
  - I have modified the API of `polylines_to_protect` so that it can work with other word types.
  - I have fixed the bug that was inserting protecting balls along the edges of the bounding box of the cube even outside the domain.

*not yet ready: can be tested in integration, but not merged in `master`*


lrineau:
TODO: in `polylines_to_protect`, the arcs of the two cases with 3 colors are approximated with segments.

lrineau:
**The compiled doc [can be seen here](https://cgal.geometryfactory.com/~lrineau/Mesh_3-example-3D-images-with-features-V1/Mesh_3/#title28).**

janetournois:
I approve the doc (if needed :-))

Mesh_3/doc/Mesh_3/Mesh_3.txt
```diff
@@ -759,9 +759,9 @@ protection of the 1D-features.
 <center>
 <table border=0>
 <tr>
-  <td><img border="0" height="435px" src="./no-protection.png"></td>
-  <td><img border="0" height="435px" src="./protection-box.png"></td>
-  <td><img border="0" height="435px" src="./protection-all.png"></td>
+  <td><img border="0" src="./no-protection.png"></td>
+  <td><img border="0" src="./protection-box.png"></td>
+  <td><img border="0" src="./protection-all.png"></td>
```
lrineau: Why? I have added that parameter on purpose. I think it was Andreas @afabri that suggested to include sizes in the HTML, to avoid that the layout of the documentation page changes during the loading of images.

Andreas, can you confirm?

lrineau: ... I may have mixed up the width and the height of my images. The `height` should be `472`.

sloriot: The output was looking very bad and the images were already at the same size.

PR #1099, opened 2016-05-23
Polyhedron demo : On the fly Selection Tool

maxGimeno:
This PR acts on the Selection_plugin in the Polyhedron Demo. Currently, if the user has only one item with an associated selection_item, but several scene_items, if he selects on any item, the selection is only applied in the selection_item. With this PR, a selection_item is automatically created for the item being selected if none exists.
This PR also fixes some bugs with the Selection Tool


maxGimeno:
I realized that either we have the on the fly selection_item creation, but they will all take any selection signal, or we disable the picking while selecting, but there is no more on-the-fly creation. @sloriot  and I decided that the second possibility was better, so I created #1117 to only keep the useful fixes of this branch.

PR #1100, opened 2016-05-23
Polyhedron demo - fix a bug in selection item for `erase_selected_facets`

janetournois:
This PR fixes a bug in the selection plugin

when `erase_selected_facets` is called, the selected edges that are in the
middle of the selected area should be un-selected before they get invalidated
by the face removal


sloriot:
This PR seems already included in #1091. If you want to keep this PR, you need to remove the commits from the other branch (or just close this PR).

janetournois:
this PR was cherry-picked from #1091 because it's only a side effect of it. But they can be merged together. I close this PR

PR #1101, opened 2016-05-23
Robustify triangulate facets

### LINKED ISSUE #1093: Robustify triangulate_face(s)
`triangulate_face()`  should return a Boolean indicating whether the face was successfully triangulated.
There are several cases to handle:
- the normal computed for the projection plane is 0,0,0-> `return false;`
- change the intersection tag of the CDT to allow intersections. After the triangulation is done, if there are more vertices than the number of input points -> `return false;`
- if the dimension of the CDT is not 2 -> `return false;`
- bonus: if there are only 4 vertices, use a geometric predicate to choose the diagonal (the sign of `CGAL::cross_product(p2-p1,p0-p1) * CGAL::cross_product(p0-p3,p2-p3)` should be working IIRC).
- do not use the CDT+ but CDT (nothing to do with this issue but useless anyway)

Once this is done,  `triangulate_faces()` should return also a bool

maxGimeno:
(fixes #1093 )
This PR makes triangulate_face() and triangulate_faces() boolean values, and optimizes triangulate_face() for the quads by using split_face instead of a CDT.
Small feature : https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Robustify_triangulate_facets

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
@@ -93,123 +93,154 @@ public:
 
   bool triangulate_face(face_descriptor f, PM& pmesh)
   {
+
+
     typename Traits::Vector_3 normal =
       Polygon_mesh_processing::compute_face_normal(f, pmesh);
     if(normal == typename Traits::Vector_3(0,0,0))
       return false;
-    P_traits cdt_traits(normal);
-    CDT cdt(cdt_traits);
-
-    // Halfedge_around_facet_circulator
-    typedef typename CDT::Vertex_handle Tr_Vertex_handle;
-    halfedge_descriptor start = halfedge(f, pmesh);
-    halfedge_descriptor h = start;
-    Tr_Vertex_handle previous, first;
-    do
+    int original_size = CGAL::halfedges_around_face(halfedge(f, pmesh), pmesh).size();
+    if(original_size == 4)
     {
-      Tr_Vertex_handle vh = cdt.insert(_vpmap[target(h, pmesh)]);
-      if (first == Tr_Vertex_handle()) {
-        first = vh;
-      }
-      vh->info() = h;
-      if(previous != Tr_Vertex_handle() && previous != vh) {
-        cdt.insert_constraint(previous, vh);
-      }
-      previous = vh;
-      h = next(h, pmesh);
+      typename Kernel::Point_3 p0, p1, p2, p3;
+      halfedge_descriptor v0, v1, v2, v3;
+      v0 = halfedge(f, pmesh);
+      p0 = _vpmap[target(v0, pmesh)];
+      v1 = next(v0, pmesh);
+      p1 = _vpmap[target(v1, pmesh)];
+      v2 = next(v1, pmesh);
+      p2 = _vpmap[target(v2, pmesh)];
+      v3 = next(v2, pmesh);
+      p3 = _vpmap[target(v3, pmesh)];
 
-    } while( h != start );
-    cdt.insert_constraint(previous, first);
-
-    // sets mark is_external
-    for(typename CDT::All_faces_iterator fit = cdt.all_faces_begin(),
-          end = cdt.all_faces_end();
-        fit != end; ++fit)
-    {
-      fit->info().is_external = false;
-    }
-    std::queue<typename CDT::Face_handle> face_queue;
-    face_queue.push(cdt.infinite_vertex()->face());
-    while(! face_queue.empty() )
-    {
-      typename CDT::Face_handle fh = face_queue.front();
-      face_queue.pop();
-
-      if(fh->info().is_external)
-        continue;
-
-      fh->info().is_external = true;
-      for(int i = 0; i <3; ++i)
+      bool predicate = CGAL::cross_product(p2-p1,p0-p1) * CGAL::cross_product(p0-p3,p2-p3) > 0;
+      if(predicate)
       {
-        if(!cdt.is_constrained(typename CDT::Edge(fh, i)))
+        CGAL::Euler::split_face(v0, v2, pmesh);
+      }
+      else
+      {
+        CGAL::Euler::split_face(v1, v3, pmesh);
+      }
+
+
+    }
+    else
+    {
+      P_traits cdt_traits(normal);
+      CDT cdt(cdt_traits);
+
+      // Halfedge_around_facet_circulator
+      typedef typename CDT::Vertex_handle Tr_Vertex_handle;
+      halfedge_descriptor start = halfedge(f, pmesh);
+      halfedge_descriptor h = start;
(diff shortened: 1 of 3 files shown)


sloriot:
The doc has not been updated. I think it requires a small feature.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -93,126 +91,165 @@ class Triangulate_modifier
...
-    typedef typename CDT::Vertex_handle Tr_Vertex_handle;
-    halfedge_descriptor start = halfedge(f, pmesh);
-    halfedge_descriptor h = start;
-    Tr_Vertex_handle previous, first;
-    do
+    typename Traits::Vector_3 normal =
+      Polygon_mesh_processing::compute_face_normal(f, pmesh);
+    if(normal == typename Traits::Vector_3(0,0,0))
+      return false;
+    int original_size = CGAL::halfedges_around_face(halfedge(f, pmesh), pmesh).size();
```
sloriot: int -> std::size_t

maxGimeno:
I updated the doc

sloriot:
When the small feature is ready, don't forget to put a link by editing the top comment.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -93,126 +91,165 @@ class Triangulate_modifier
...
+      Polygon_mesh_processing::compute_face_normal(f, pmesh);
+    if(normal == typename Traits::Vector_3(0,0,0))
+      return false;
+    std::size_t original_size = CGAL::halfedges_around_face(halfedge(f, pmesh), pmesh).size();
+    if(original_size == 4)
     {
-      Tr_Vertex_handle vh = cdt.insert(_vpmap[target(h, pmesh)]);
-      if (first == Tr_Vertex_handle()) {
-        first = vh;
+      typename Kernel::Point_3 p0, p1, p2, p3;
```
sloriot: Same here, you can use `reference` from `property_traits`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -93,126 +91,165 @@ class Triangulate_modifier
...
+      v0 = halfedge(f, pmesh);
+      p0 = _vpmap[target(v0, pmesh)];
+      v1 = next(v0, pmesh);
+      p1 = _vpmap[target(v1, pmesh)];
+      v2 = next(v1, pmesh);
+      p2 = _vpmap[target(v2, pmesh)];
+      v3 = next(v2, pmesh);
+      p3 = _vpmap[target(v3, pmesh)];
+
+      bool predicate = CGAL::cross_product(p2-p1,p0-p1) * CGAL::cross_product(p0-p3,p2-p3) > 0;
```
sloriot: Did you check this is the good predicate to use?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -362,16 +407,16 @@ void triangulate_faces(FaceRange face_range, PolygonMesh& pmesh)
 *
 */
 template <typename PolygonMesh, typename NamedParameters>
-void triangulate_faces(PolygonMesh& pmesh,
+bool  triangulate_faces(PolygonMesh& pmesh,
```
sloriot: The documentation was not updated

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
lrineau: That is not a predicate in the sense of CGAL.

I have the impression that the functor [`Kernel::Compare_dihedral_angle_3`][doc] from the kernel could be used, with `0` for the cosine.

[doc]: http://doc.cgal.org/latest/Kernel_23/classKernel.html#a00792fd1a5d950d54a750f4aa9d2ba3f

But actually, to be efficient, a dedicated predicate would be better. `Compare_dihedral_angle_3` computes cross products `u` and `v`, and their scalar product `u*v`. But it also computes the squared length of `u` and `v`: it checks `(u*v)^2 > cosine * ((u*u)*(v*v))`. The computations of lengths are useless when one want to compare the sign of `u*v` with 0.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -93,126 +92,167 @@ class Triangulate_modifier
...
+      v1 = next(v0, pmesh);
+      Point_ref p1 = _vpmap[target(v1, pmesh)];
+      v2 = next(v1, pmesh);
+      Point_ref p2 = _vpmap[target(v2, pmesh)];
+      v3 = next(v2, pmesh);
+      Point_ref p3 = _vpmap[target(v3, pmesh)];
 
-    P_traits cdt_traits(normal);
-    CDT cdt(cdt_traits);
+      double p1p3= CGAL::cross_product(p2-p1,p3-p2) * CGAL::cross_product(p0-p3,p1-p0);
```
sloriot: A small comment here explaining what is the logic would be great (to avoid reverse engineering)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -93,126 +92,169 @@ class Triangulate_modifier
...
+      v2 = next(v1, pmesh);
+      Point_ref p2 = _vpmap[target(v2, pmesh)];
+      v3 = next(v2, pmesh);
+      Point_ref p3 = _vpmap[target(v3, pmesh)];
+
+      //Chooses the diagonal that will maximize the cross product. The dot product allows to
+      // choose the right diagonal if the quad is not convex.
+      double p1p3= CGAL::cross_product(p2-p1,p3-p2) * CGAL::cross_product(p0-p3,p1-p0);
+      double p0p2= CGAL::cross_product(p1-p0,p1-p2) * CGAL::cross_product(p3-p2,p3-p0);
+
```
lrineau: The comment is wrong. The expressions `p1p3` and `p0p2` are not cross products, but scalar products of cross products.

I suggest something like the following lines:
> Choose the diagonal that will split the quad in two triangles that maximize the scalar product of of the un-normalized normals of the two triangles. The lengths of the un-normalized normals (computed using cross-products of two vectors) are proportional to the area of the triangles.
> Maximize the scalar product of the two normals will avoid skinny triangles, and will also taken into account the cosine of the angle between the two normals. In particular, if the two triangles are oriented in different directions, the scalar product will be negative.

PR #1103, opened 2016-05-24
Polyhedron demo : fix a warning in polygon_soup_item

janetournois:
fix int <-> size_t conversion warning


sloriot:
Already done in #1098

PR #1105, opened 2016-05-25
Make test correctly report errors

### LINKED ISSUE #953: Orthogonal_k_neighbor_search.cpp prints errors
Looking at a random run of the testsuite: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-6/Spatial_searching/TestReport_lrineau_CentOS6-CXX11-Boost157.gz the test Orthogonal_k_neighbor_search prints many lines of the form

> Error: Point -0.999521 -0.662093 at distance 1.43741  >  0.442363

The test still returns 0 and the testsuite shows up green. Is this output expected? If so, it would be nice to reformulate it so it doesn't look so much like there is a bug... And if it isn't expected, then you know...

sloriot:
Fixes #953 

Highlight a regression introduced in CGAL 4.6 about  further neighbor search

### DIFF
--- a/Spatial_searching/include/CGAL/Kd_tree.h
+++ b/Spatial_searching/include/CGAL/Kd_tree.h
@@ -163,17 +163,25 @@ private:
     nh->set_separator(sep);
 
     int cd  = nh->cutting_dimension();
-    if(!c_low.empty())
-      nh->low_val = c_low.tight_bounding_box().max_coord(cd);
-    else
-      nh->low_val = c_low.bounding_box().min_coord(cd);
-    if(!c.empty())
-      nh->high_val = c.tight_bounding_box().min_coord(cd);
-    else
-      nh->high_val = c.bounding_box().max_coord(cd);
+    if(!c_low.empty()){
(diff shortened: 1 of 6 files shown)


lrineau:
If I understand well, this PR does not fix #953 fully: it only fixes the fact that test results reports are wrongly reported. But even with this PR, the test results are wrong.

sloriot:
Indeed, testsuite will be fully red

sloriot:
The last commit I pushed fixes the orthogonal furthest neighbor search but breaks the improvement done by @MiesA. If we keep tight bounding box we need 3 more doubles per node (we can save one by removing the cutting value in the seperator). Otherwise we need to go back to using the cutting value and kd-tree nodes bounds (+canceling the concept changes introduced that are no longer needed).
@afabri Let's have a skype meeting next week to decide what to do.

sloriot:
After doing some benchmarks, it turns out that the extra 2 double per node does not have a noticeable impact on the performance.

PR #1109, opened 2016-05-26
Mesh_criteria_3: change the default for `edge_size` from 0 to +∞

lrineau:
The previous default value for `CGAL::parameters::edge_size`, in the constructor of `CGAL::Mesh_criteria_3` was `FT(0)`.

That was in contradiction with the documentation of [`CGAL::Mesh_criteria_3`][*]:

> Each parameter has a special default value `ignored` which means that
> the corresponding criteria will be ignored.

With `edge_size=0`, the protection of 1D-features is certain to trigger an assertion or a segmentation fault.

The new default value is now `FT(DBL_MAX)`.

[*]: http://doc.cgal.org/latest/Mesh_3/classCGAL_1_1Mesh__criteria__3.html#ada26edb7ee77aa7279b2ae7e96c56eb0

*Note that the tests have been tweaked to test also the meshing of a polyhedron with sharp features (the unit tetrahedron), using the default value of `edge_size`.*


janetournois:
IIRC there is no `edge_size` in the demo `Meshing_dialog.ui`. Don't you think it should be added?

lrineau:
Yes, I think so. And we should check that, if there are 1D-features, then that parameter cannot be 0.

lrineau:
... and the default value could be set to the same as the facet size.

janetournois:
I will write an issue about that and make the changes when PR #1075 will be merged. Otherwise we'll have conflicts

PR #1110, opened 2016-05-26
Remove duplicated implementation in <CGAL/Mesh_3/polylines_to_protect.h>

### LINKED ISSUE #1107: Remove duplicated implementation
Try removing the file [`Mesh_3/include/CGAL/internal/Mesh_3/split_in_polylines.h`](https://github.com/CGAL/cgal/blob/master/Mesh_3/include/CGAL/internal/Mesh_3/split_in_polylines.h) by replacing its usage by the function in [`BGL/include/CGAL/boost/graph/split_graph_into_polylines.h`](https://github.com/CGAL/cgal/blob/master/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h).

A usage example is provided in [`simple_mcfskel_example.cpp`](https://github.com/CGAL/cgal/blob/master/Surface_mesh_skeletonization/examples/Surface_mesh_skeletonization/simple_mcfskel_example.cpp)

maxGimeno:
(fixes #1107)

This PR removes the file `Mesh_3/include/CGAL/internal/Mesh_3/split_in_polylines.h` and replaces its usage by `CGAL::split_graph_into_polylines`.


Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -25,10 +25,37 @@
 #include <map>
 #include <CGAL/tuple.h>
 #include <CGAL/Image_3.h>
-#include <CGAL/internal/Mesh_3/split_in_polylines.h>
+#include <CGAL/boost/graph/split_graph_into_polylines.h>
 #include <CGAL/internal/Mesh_3/Graph_manipulations.h>
 #include <boost/graph/adjacency_list.hpp>
```
sloriot: The class should not be defined in the global namespace. I suggest `CGAL::Mesh_3`.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -25,10 +25,37 @@
...
 
+template<typename P, typename G>
+struct Polyline_visitor
+{
+  std::vector<std::vector<P> >& polylines;
+  G graph;
+  Polyline_visitor(typename std::vector<std::vector<P> >& lines, G p_graph)
+    : polylines(lines), graph(p_graph)
```
sloriot: is it expected that the graph is copied?

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
lrineau: @sloriot Please write comments *below* the code and not *above* it. The mails that are send by Github are strange if you write comments above the code.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
lrineau: `graph` should be a reference or a pointer.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -25,13 +25,41 @@
...
+
+
 namespace CGAL {
+namespace Mesh_3{
+template<typename P, typename G>
+struct Polyline_visitor
+{
+  std::vector<std::vector<P> >& polylines;
+  G graph;
+  Polyline_visitor(typename std::vector<std::vector<P> >& lines, G& p_graph)
```
sloriot: The graph is still copied...

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -335,8 +363,8 @@ namespace CGAL {
   CGAL_assertion(nb_facets == expected_nb_facets);
   CGAL_USE(nb_facets); CGAL_USE(expected_nb_facets);
 
-
-  internal::Mesh_3::split_in_polylines(graph, polylines, K());
+  CGAL::Mesh_3::Polyline_visitor<Point_3, Graph> visitor(polylines, graph);
+  CGAL::split_graph_into_polylines(graph, visitor);
```
sloriot: @lrineau Actually the graph is also copied internally in `split_graph_into_polylines` while the former implementation was doing the modification in-place. If that matters, another version of `split_graph_into_polylines` working in-place should be added.

lrineau: The best solution to that the would be C++11 and `std::move`. If Boost Graph support the move constructors.



Le 26 mai 2016 16:25, à 16:25, Sebastien Loriot <notifications@github.com> a écrit:

sloriot: No I mean the graph is really copied (by hand) the purpose being that the input might not be a `boost::adjacency_list`.

lrineau: I understand know: the graph is copied, but not the properties of the vertices (because the properties of the new graph is just the `vertex_descriptor`s of the corresponding vertices in the original graph.

For Mesh_3 that should not be noticeable, compared to all the memory and CPU used by the algorithm that samples the 1D-features.

lrineau: But the original graph should not be copied as-is, copying also the properties. That would be a waste of CPU time and space.

sloriot: Only the connectivity is copied and the vertices of the created graph have an internal property which is the original vertex. Thus properties are not copied.

lrineau:
I have added two commits. One to remove a warning and the other one to fix a runtime error in Mesh_3, seen in the testsuite if #1097 is merged.

PR #1121, opened 2016-05-30
Polyhedron_demo: Display distance between points

maxGimeno:
This PR allows the user to display the distance between two points in the scene by pressing D and clicking on the scene. It will display a line between the two points, print their coordinates on their position and the distance in the middle of the segment. Press D alone to erase it.

screen

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.cpp
+++ b/Polyhedron/demo/Polyhedron/Viewer.cpp
@@ -1,5 +1,4 @@
 #include "Viewer.h"
-#include <CGAL/Three/TextRenderer.h>
 #include <CGAL/gl.h>
 #include <CGAL/Three/Scene_draw_interface.h>
 #include <QMouseEvent>
@@ -10,6 +9,7 @@
 #include <QOpenGLShaderProgram>
 #include <cmath>
 #include <QApplication>
+
 
 class Viewer_impl {
 public:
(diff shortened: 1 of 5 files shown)


Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -192,59 +199,111 @@ void Viewer::initializeGL()
...
-      "gl_FragColor = color*light_amb + diffuse + specular; \n"
-      "} \n"
-      "\n"
-  };
-  QOpenGLShader *vertex_shader = new QOpenGLShader(QOpenGLShader::Vertex);
-  if(!vertex_shader->compileSourceCode(vertex_source))
-  {
-      std::cerr<<"Compiling vertex source FAILED"<<std::endl;
-  }
+          "#version 120 \n"
```
sloriot: any reason to change the indentation here?

sloriot:
If you're really fast to select the points, the previous text is not erased
screenshot from 2016-05-30 18-02-38

sloriot:
When changing the rotation center (hold right and double-click), the distance segment is displayed in bold.
screenshot from 2016-05-30 18-11-34
Actually I'd prefer it that way all the time as it is not easy to distinguish it from mesh edges

maxGimeno:
The bugs have been fixed.

PR #1128, opened 2016-06-02
Add a macro to customize deprecation warnings

### LINKED ISSUE #1127: Add a new macro CGAL_DEPRECATED_MSG
I have discovered in OpenMesh that they have nice macro to deprecate things. It can display custom messages. I suggest we add in CGAL a similar macro, named `CGAL_DEPRECATED_MSG`.

Here is current code in OpenMesh version 3.2, in file `OpenMesh/Core/System/config.h`:
```C++
#if defined(_MSC_VER)
#  define DEPRECATED(msg) __declspec(deprecated(msg))
#elif defined(__GNUC__)
#  if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) >= 40500 /* Test for GCC >= 4.5.0 */
#    define DEPRECATED(msg) __attribute__ ((deprecated(msg)))
#  else
#    define DEPRECATED(msg) __attribute__ ((deprecated))
#  endif
#elif defined(__clang__)
#  define DEPRECATED(msg) __attribute__ ((deprecated(msg)))
#else
#  define DEPRECATED(msg)
#endif
```
We can add something similar in `CGAL/config.h`, but using the Boost.Config macros to check/detect compilers versions.

Example of usage, in `OpenMesh/Core/Mesh/CirculatorsT.hh`:
```C++
DEPRECATED("Do not use this error prone implicit cast. Compare to end-iterator or use is_valid(), instead.")
/**
 * \deprecated
 * Do not use this error prone implicit cast. Compare to the
 * end-iterator or use is_valid() instead.
 */
operator bool() const {
    return is_valid();
}
```

maxGimeno:
Fixes #1127 
This PR adds a macro : CGAL_DEPRECATED_MSG that allows to display a custom message in the deprecation warnings.

### DIFF
--- a/Installation/include/CGAL/config.h
+++ b/Installation/include/CGAL/config.h
@@ -424,7 +424,7 @@ using std::max;
 
 // Macro to trigger deprecation warnings with a custom message
 #ifdef CGAL_NO_DEPRECATION_WARNINGS
-#  define CGAL_DEPRECATED_MSG
+#  define CGAL_DEPRECATED_MSG(msg)
 #elif defined(__GNUC__) || __has_attribute(__deprecated__)
 #  if BOOST_GCC >= 40500 || __has_attribute(__deprecated__)
 #    define CGAL_DEPRECATED_MSG(msg) __attribute__ ((deprecated(msg)))
--- a/Installation/include/CGAL/config.h
+++ b/Installation/include/CGAL/config.h
@@ -422,6 +422,22 @@ using std::max;
 #  define CGAL_DEPRECATED
 #endif
 
+// Macro to trigger deprecation warnings with a custom message
+#ifdef CGAL_NO_DEPRECATION_WARNINGS
+#  define CGAL_DEPRECATED_MSG
+#elif defined(__GNUC__) || __has_attribute(__deprecated__)
+#  if BOOST_GCC >= 40500 || __has_attribute(__deprecated__)
+#    define CGAL_DEPRECATED_MSG(msg) __attribute__ ((deprecated(msg)))
+#  else
+#    define CGAL_DEPRECATED_MSG(msg) __attribute__ ((deprecated))
+#  endif
+#elif defined (_MSC_VER) && (_MSC_VER > 1300)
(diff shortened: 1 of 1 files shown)


Installation/include/CGAL/config.h
```diff
@@ -422,6 +422,22 @@ using std::max;
...
+// Macro to trigger deprecation warnings with a custom message
+#ifdef CGAL_NO_DEPRECATION_WARNINGS
+#  define CGAL_DEPRECATED_MSG
+#elif defined(__GNUC__) || __has_attribute(__deprecated__)
+#  if BOOST_GCC >= 40500
+#    define CGAL_DEPRECATED_MSG(msg) __attribute__ ((deprecated(msg)))
+#  else
+#    define CGAL_DEPRECATED_MSG(msg) __attribute__ ((deprecated))
+#  endif
+#elif defined (_MSC_VER) && (_MSC_VER > 1300)
```
lrineau: Recent Clang also has the syntax with the message, and even more.
http://clang.llvm.org/docs/AttributeReference.html#deprecated-gnu-deprecated

So line 429 it should be:
```C++
#  if BOOST_GCC >= 40500  || __has_attribute(__deprecated__)
```

maxGimeno:
I think it is ready to merge.

maxGimeno:
There is still another bug I need to fix before it can be merged though.

maxGimeno:
Bugs fixed, if @janetournois  says it's ok, then it is ready to merge.

Installation/include/CGAL/config.h
```diff
@@ -422,6 +422,22 @@ using std::max;
 #  define CGAL_DEPRECATED
 #endif
 
+// Macro to trigger deprecation warnings with a custom message
+#ifdef CGAL_NO_DEPRECATION_WARNINGS
+#  define CGAL_DEPRECATED_MSG
```
lrineau: There is an error, here. That lines defines a macro without argument. That will lead to a compilation error for any use of that macro with an argument.

PR #1129, opened 2016-06-02
Polyhedron_demo : fix the context menu of the children of a group

maxGimeno:
This PR fixes a bug in the hierarchy.


sloriot:
Locally tested

PR #1137, opened 2016-06-06
Polyhedron_demo : Fix the update of the viewer

maxGimeno:
- Replaces updateGL() calls by update() calls. Before the viewer used a painter, updateGL() was ok, because it was the QGLViewer that painted the scene. But a QPainter can only update its painting through a paintEvent() function, which is called by update() but not by updateGL().
- Fixes the bug that made the selection empty when an item was erased.
- Fixes the background color.


afabri:
Locally tested on Windows/VC++, where the problem had appeared.

PR #1146, opened 2016-06-09
Fix for the OBJ reading.

maxGimeno:
This PR fixes the Scene_polyhedron_item::Load_obj() function, that was only able to read the files with only vertex data, and no normal, texture, etc. 
Now it discards this unwanted information to avoid the crashing.


sloriot:
@maxGimeno : We have the files`Polyhedron_IO/include/CGAL/IO/OFF_reader.h` and  `Polyhedron_IO/include/CGAL/IO/STL_reader.h`. Would you mind create the file `OBJ_reader.h` and put the code in this file? I though @afabri already did it but no.

maxGimeno:
I did it and changed it so it can read not-triangle items.

PR #1147, opened 2016-06-09
Do not cache CGAL_THREE_SCENE_ITEM_H_PATH ; Simplify CMakeLists.txt

afabri:
tested on windows

### DIFF
--- a/Polyhedron/demo/Polyhedron/CMakeLists.txt
+++ b/Polyhedron/demo/Polyhedron/CMakeLists.txt
@@ -146,8 +146,7 @@ if(CGAL_Qt5_FOUND AND Qt5_FOUND AND OPENGL_FOUND AND QGLVIEWER_FOUND)
     HINTS ${CGAL_INCLUDE_DIRS}
     NO_DEFAULT_PATH
     DOC "Path to CGAL/Three/Scene_item.h")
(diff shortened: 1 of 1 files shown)


lrineau:
I did the patch. Two modifications:

- Instead of three `find_path` to get three different variables that are supposed to have equal values, there is only variable and one `find_path`.

- What is more, that variable is not taken from the cache, to avoid that the value stays the same if `CGAL_DIR` is changed.

lrineau:
With the commit https://github.com/CGAL/cgal/pull/1147/commits/964fdabb49ab4e0c72cd525701300f5e11070a3f, I have changed my mind: the variable is never stored in the cache, to avoid that the user sees it and assumes that the value can be changed that way.

sloriot:
When building CGAL with `WITH_demos=ON` I got the following error:
`CMake Error at Polyhedron/demo/Polyhedron/CMakeLists.txt:161 (message):
  Cannot find <CGAL/Three/Viewer_interface.h>`

lrineau:
I fixed the error.

sloriot:
I suspect the red line is due to this PR

lrineau:
Indeed. There was an error, fixed by https://github.com/CGAL/cgal/pull/1147/commits/bb5957791cc1e599792e19ce2733b9ba510c7152.

PR #1164, opened 2016-06-13
Make constructor SM_Edge_index(SM_Halfedge_index) explicit

afabri:


sloriot:
BGL_surface_mesh_Examples is fully red

sloriot:
The error is actually interesting, some overloads are missing for `filtered_graph`.
I fixed some error in the SM item in the demo.

sloriot:
@afabri I did not have fix the missing overloads.

PR #1165, opened 2016-06-13
Polyhedron_demo: Fix triangulate primitive

maxGimeno:
This PR fixes some bug in the triangulate_primitive.h file, in the workaround for the meshes with very close points.
 @janetournois  can you please check my algorithm is right?

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_item.cpp
@@ -229,12 +229,11 @@ bool CGAL::Three::Scene_item::testDisplayId(double, double, double, CGAL::Three:
 
 void CGAL::Three::Scene_item::compute_diag_bbox()const
 {
-  if(!is_bbox_computed)
-    compute_bbox();
+  Bbox b_box = bbox();
   _diag_bbox = CGAL::sqrt(
-        CGAL::square(_bbox.xmax() - _bbox.xmin())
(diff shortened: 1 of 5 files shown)


Polyhedron/demo/Polyhedron/Scene_item.cpp
```diff
@@ -19,6 +19,7 @@ CGAL::Three::Scene_item::Scene_item(int buffers_size, int vaos_size)
     vaos(vaos_size)
 {
   is_bbox_computed = false;
+  is_diag_bbox_computed = false;
```
sloriot: This does not seem to be used

Polyhedron/demo/Polyhedron/Scene_item.cpp
```diff
@@ -228,12 +229,11 @@ bool CGAL::Three::Scene_item::testDisplayId(double, double, double, CGAL::Three:
 
 void CGAL::Three::Scene_item::compute_diag_bbox()const
 {
-  if(!is_bbox_computed)
-    compute_bbox();
+  Bbox b_box = bbox();
```
sloriot: you should take a const reference

Polyhedron/demo/Polyhedron/Scene_item.cpp
maxGimeno: It was forgotten, causing a bug.  It is used in Scene_item::diagonalItem()

PR #1169, opened 2016-06-14
Avoid infinite call loop in search traits adapter

sloriot:


lrineau:
Is that a bug introduced in CGAL-4.8.1, or was it already in CGAL-4.8?

sloriot:
already in 4.8

PR #1170, opened 2016-06-14
Polyhedron_demo : Fix gray images UI

maxGimeno:
This PR fixes some problems in the UI. It also fixes the picking for the whole demo.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
@@ -222,7 +222,7 @@ public:
       connect(planeSwitch, SIGNAL(triggered()),
               this, SLOT(selectPlanes()));
       Scene* true_scene = dynamic_cast<Scene*>(scene);
-      connect(true_scene,SIGNAL(selectionChanged(int)),
+      connect(true_scene,SIGNAL(itemIndexSelected(int)),
               this, SLOT(connect_controls(int)));
     }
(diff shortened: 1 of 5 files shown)


Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
```diff
@@ -222,7 +222,7 @@ class Io_image_plugin :
       connect(planeSwitch, SIGNAL(triggered()),
               this, SLOT(selectPlanes()));
       Scene* true_scene = dynamic_cast<Scene*>(scene);
-      connect(true_scene,SIGNAL(selectionChanged(int)),
+      connect(true_scene,SIGNAL(itemIndexSelected(int)),
```
lrineau: `Scene::itemIndexSelected(int)` is a new signal added by this PR. Why is it useful, compared to `selectionChanged(int)`?

Add, actually, in `Scene`, why are there two functions that seem to do the same?
  - `void setSelectedItemIndex(int i)`
  - `void setSelectedItem(int i )`

maxGimeno: The two functions emit a different signal. One is used in case a plugin or an item must do something when the selection has changed, the other is only used to update the TreeView. They are not always called in the same conditions.

lrineau: I hope that is documented correctly, because that is really confusing, and the names of the two functions, and the names of the two signals are two similar.

PR #1172, opened 2016-06-14
Polyhedron_demo : Fix the Create_point_set_from_selection operation

maxGimeno:
This PR fixes the Create_point_set_from_selection operation in the Selection tool.


Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
```diff
@@ -316,8 +316,10 @@ public Q_SLOTS:
       for(Scene_polyhedron_selection_item::Selection_set_vertex::iterator begin = selection_item->selected_vertices.begin();
          begin != selection_item->selected_vertices.end(); ++begin) {
          point_item->point_set()->push_back((*begin)->point());
-      break;
       }
+      scene->setSelectedItem( scene->addItem(point_item) );
+      scene->itemChanged(point_item);
+      break;
```
lrineau: I really would like to know why one must call `scene->itemChanged` on an item that has just been added to the scene!

That gives me the impression that the signaling system of the scene is now broken.

PR #1177, opened 2016-06-15
Polyhedron_Demo: Enhance Shortest Path Plugin

maxGimeno:
This Pr :

- Changes the UI to make it more explicit.
- Put the items created by the plugin in a group_item
- Fixes the color of the points


sloriot:
Tested locally and it looks good

PR #1180, opened 2016-06-15
fix edge hash

sloriot:


sloriot:
@afabri The Hash_map test crashes because the test is trying to hash a default constructed edge_descriptor. Should it be hashable or should I update the  test?

afabri:
Good question. Note that the boost::graph_traits<..>::null_vertex is in our implementations also the default constructed vertex_descriptor.   Such an object could be used as a sentinel with a special value in the hash map, so I am in favor of hashing it too.

sloriot:
Do you mean that I should modified also all hash functions to handle the default constructed handle?

sloriot:
There are errors on one platforms see [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-80/Hash_map/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Debug-64bits.gz)

PR #1189, opened 2016-06-20
Fix for source on edge

### LINKED ISSUE #1179: Bugs in Shortest_path
There are several bugs in the Shortest_path computation :

- When a source point is put on an edge, depending on where the request has been made, the shortest path might be wrong (see picture below).
screen_path

- When the request and the source point are on the same edge, there is an assertion (see picture below).
screen_path1

I think the problem is located in Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h, in the function visit_shortest_path(Cone_tree_node* startNode, const Point_2& startLocation, Visitor& visitor) (line 1545)

sloriot:
Fixes #1179

@Hippocrates comments welcome!

### DIFF
--- a/Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
+++ b/Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
@@ -773,7 +773,7 @@ private:
 
     Point_2 sourcePoints[2];
     sourcePoints[0] = cb2(cv2(layoutFaces[0], 0), t0, cv2(layoutFaces[0], 1), t1);
-    sourcePoints[1] = cb2(cv2(layoutFaces[1], 0), t0, cv2(layoutFaces[1], 1), t1);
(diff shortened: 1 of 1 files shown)


Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
```diff
@@ -1570,9 +1570,9 @@ class Surface_mesh_shortest_path
 
           CGAL_assertion(bool(cgalIntersection));
 
-          Point_2* result = boost::get<Point_2>(&*cgalIntersection);
+          const Point_2* result = boost::get<Point_2>(&*cgalIntersection);
 
-          CGAL_assertion(result && "Error, did not get point intersection on path walk to source");
+          if (!result) result=&current->source_image();
```
Hippocrates: My one hesitation here would be that I'm typically wary of any code that relies specifically on the return type of a method being a const reference rather than by value (its not terribly risky here, its just a convention I typically avoid). Also, spacing around the `=`

Hippocrates:
Otherwise looks good to me.

sloriot:
@Hippocrates I rewrote my last commit to match your coding conventions. Thanks for the review.

PR #1191, opened 2016-06-20
ImageIO no longer needs openGL

sloriot:


sloriot:
@maxGimeno `Surface_mesher_demo` is using a hidden OpenGL direct draw method (see the two methods removed in this PR). Could you please have a look to see if updating the demo is easy?

maxGimeno:
From what I see I think the demo does not use gl_draw() anymore. I just left the definition because at the time I didn't dare to remove it. Even if I am wrong, I don't think it would be difficult to update the demo. I can do it if you want.

PR #1201, opened 2016-06-24
Polyhedron_demo : Fix deformation

### LINKED ISSUE #1193: Loading an off while deforming another...
.. ends up creating another deformation object. This is inconsistent with some changes done by @maxGimeno to avoid deforming more than one item at once.

### LINKED ISSUE #1194: Show ROI checkbox is broken
in deformation plugin.

I had "show as sphere checked" when I tried to hide the ROI. When unchecking "show as spheres" ROI and control vertices are not shown and they won't show up even if I check "show ROI"

@maxGimeno

### LINKED ISSUE #1197: Extending ROI/Control vertices partially working
screenshot from 2016-06-23 13-26-00

I selected the ROI and then only one control point. I used the mouse wheel to extend the region.

In the package, note that the ROI is the region that is allowed to move, and they contains the control vertices.

cc @maxGimeno

maxGimeno:
This PR fixes some bugs in the Edit_polyhedron_plugin (fixes #1193, fixes #1194 and fixes #1197 ).

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
@@ -60,7 +60,6 @@ public Q_SLOTS:
   void on_BrushSpinBoxCtrlVert_changed(int);
   void on_BrushSpinBoxRoi_changed(int);
   void on_ROIRadioButton_toggled(bool);
-  void new_item_created(int item_id);
   void on_importSelectionPushButton_clicked();
   void importSelection(Scene_polyhedron_selection_item*, Scene_edit_polyhedron_item*);
 private:
@@ -102,13 +101,7 @@ void Polyhedron_demo_edit_polyhedron_plugin::init(QMainWindow* mainWindow, CGAL:
 
   // Connect Scene::newItem so that, if dock_widget is visible, convert
   // automatically polyhedron items to "edit polyhedron" items.
-  QObject* scene = dynamic_cast<QObject*>(scene_interface);
-  if(scene) {
-    connect(scene, SIGNAL(newItem(int)), this, SLOT(new_item_created(int)));
-  } else {
-    std::cerr << "ERROR " << __FILE__ << ":" << __LINE__ << " :"
-              << " cannot convert scene_interface to scene!\n"; 
-  }
+
   ////////////////// Construct widget /////////////////////////////
   // First time, construct docking window
   dock_widget = new QDockWidget("Mesh Deformation", mw);
@@ -419,26 +412,6 @@ void Polyhedron_demo_edit_polyhedron_plugin::on_BrushSpinBoxRoi_changed(int valu
   }
 }
 
-void Polyhedron_demo_edit_polyhedron_plugin::new_item_created(int item_id)
-{
-  Scene_polyhedron_selection_item* selection_item = NULL;
-  for(int i = 0; i<scene->numberOfEntries(); i++)
-  {
-    selection_item = qobject_cast<Scene_polyhedron_selection_item*>(scene->item(i));
-    if (selection_item)
-      break;
-    else
-      selection_item = NULL;
-  }
-  if(selection_item
-     && dock_widget->isVisible()) {
-    Scene_polyhedron_item* poly_item = 
-      qobject_cast<Scene_polyhedron_item*>(scene->item(item_id));
-    if(poly_item) {
-      convert_to_edit_polyhedron(item_id, poly_item);
-    }
-  }
-}
 
 Scene_edit_polyhedron_item* 
 Polyhedron_demo_edit_polyhedron_plugin::convert_to_edit_polyhedron(Item_id i,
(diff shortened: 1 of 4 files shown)


Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
```diff
@@ -421,7 +421,17 @@ void Polyhedron_demo_edit_polyhedron_plugin::on_BrushSpinBoxRoi_changed(int valu
...
 void Polyhedron_demo_edit_polyhedron_plugin::new_item_created(int item_id)
 {
-  if(dock_widget->isVisible()) {
+  Scene_polyhedron_selection_item* selection_item = NULL;
+  for(int i = 0; i<scene->numberOfEntries(); i++)
+  {
+    selection_item = qobject_cast<Scene_polyhedron_selection_item*>(scene->item(i));
+    if (selection_item)
+      break;
+    else
```
sloriot: I guess setting it to NULL is useless since it is already

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
```diff
@@ -421,7 +421,17 @@ void Polyhedron_demo_edit_polyhedron_plugin::on_BrushSpinBoxRoi_changed(int valu
...
+      selection_item = NULL;
```
sloriot: BTW, I don't see why we need this function at all now.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -751,53 +765,74 @@ struct Is_selected_property_map{
...
+    if(ctrl_active)
+      expand_vertex_selection(active_group->ctrl_vertices_group, *poly_item->polyhedron(), steps, Is_selected_property_map(mark),
+                            CGAL::Emptyset_iterator());
+    else
+      expand_vertex_selection(deform_mesh->roi_vertices(), *poly_item->polyhedron(), steps, Is_selected_property_map(mark),
                             CGAL::Emptyset_iterator());
+  }
   else
-    reduce_vertex_selection(deform_mesh->roi_vertices(), *poly_item->polyhedron(), -steps, Is_selected_property_map(mark),
+  { if(ctrl_active)
```
sloriot: identation issue

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -751,53 +765,74 @@ struct Is_selected_property_map{
...
+    }
+    else
+    {
+      if(!deform_mesh->is_control_vertex(v))
+        mark[v->id()]=true;
+    }
   }
-
+  std::vector<bool> patron = mark;
   if(steps > 0)
```
sloriot: `patron` should be translated to `mask`

sloriot:
I tested the branch and I confirm it fixes the 3 aforementioned issues

sloriot:
In 4.7, the control vertices not in the current group were displayed in blue. Red was for the currently selected group

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -1220,6 +1309,10 @@ void Scene_edit_polyhedron_item::reset_spheres()
 {
   d->spheres = NULL;
 }
+void Scene_edit_polyhedron_item::reset_spheres_ctrl()
+{
+  d->spheres_ctrl = NULL;
+}
```
lrineau: (A general note for everybody modifying the Polyhedron demo. cc @sloriot)
The new API of [`QObject::connect`][c] allows to connect a signal to a functor. There is no longer the needs to add slots, and this modify the the `.h`.

If C++11 was allowed, the connection would be really simple, with a lambda expression:
```C++
connect(d->spheres_ctrl, &QObject::destroyed, this, [this] () { this->d->spheres_ctrl = NULL} );
```
The expression `[this] () { this->d->spheres_ctrl = NULL}` is a lambda-expression, that captures the value of `this`.

Without C++11, it can be done almost as easily:
```C++
struct Reset_spheres_ctrl {
  Scene_edit_polyhedron_item_priv* d;
  Reset_spheres_ctrl(Scene_edit_polyhedron_item_priv* d) : d(d) {}
  void operator()() const { d->spheres_ctrl = NULL; }
};
// [...]
connect(d->spheres_ctrl, &QObject::destroyed, this, Reset_spheres_ctrl(d) );
```
[c]: http://doc.qt.io/qt-5/qobject.html#connect-5

See also [New Signal Slot Syntax in Qt 5, in the wiki of Qt][*]. @maxGimeno: please read that document. You will see that it is really interesting.
[*]: https://wiki.qt.io/New_Signal_Slot_Syntax

I really would like that we allow to use C++11 in our demos. With CMake>=3.1, that is as simple as:
```CMake
target_compile_features(my_plugin PRIVATE cxx_lambdas)
```

maxGimeno:
Color fixed.

PR #1209, opened 2016-06-27
Polyhedron_demo : Keeping colors after deformation

### LINKED ISSUE #1203: Polyhedron demo : edit_polyhedron_plugin discards patch ids
When a `polyhedron_item` has patch ids (e.g. computed by "color connected components"), and the `edit_polyhedron_plugin` is started (with `Ctrl+E` e.g.), the patch ids are discarded - or maybe just not displayed anymore.

cc @maxGimeno

maxGimeno:
This fixes #1203 .
This PR saves the Colors_ vector of the polyhedron_item if it exists before the conversion to edit_item and restores it after the deformation is saved.

 It also fixes a color bug in isotropic_remeshing_plugin that occurs when it is used with a polyhedron_selection_item;

### DIFF
--- a/Polyhedron/demo/Polyhedron/Color_map.h
+++ b/Polyhedron/demo/Polyhedron/Color_map.h
@@ -9,7 +9,6 @@ compute_color_map(QColor base_color,
                   unsigned nb_of_colors,
                   Output_color_iterator out)
 {
- std::cerr<<"plouf \n";
   qreal hue = base_color.hueF();
   const qreal step = ((qreal)1) / nb_of_colors;
 
--- a/Polyhedron/demo/Polyhedron/Color_map.h
+++ b/Polyhedron/demo/Polyhedron/Color_map.h
@@ -9,6 +9,7 @@ compute_color_map(QColor base_color,
                   unsigned nb_of_colors,
                   Output_color_iterator out)
 {
+ std::cerr<<"plouf \n";
   qreal hue = base_color.hueF();
   const qreal step = ((qreal)1) / nb_of_colors;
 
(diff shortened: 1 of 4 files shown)


maxGimeno:
@janetournois said she must add something to isotropic_remeshing to complete the fix.

Polyhedron/demo/Polyhedron/Color_map.h
```diff
@@ -9,6 +9,7 @@ compute_color_map(QColor base_color,
                   unsigned nb_of_colors,
                   Output_color_iterator out)
 {
+ std::cerr<<"plouf \n";
```
sloriot: debug info to remove

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
```diff
@@ -457,6 +459,13 @@ Polyhedron_demo_edit_polyhedron_plugin::convert_to_plain_polyhedron(Item_id i,
   Scene_polyhedron_item* poly_item = edit_item->to_polyhedron_item();
   scene->replaceItem(i, poly_item);
   delete edit_item;
+  if(saved_color.size() >0)
```
sloriot: Quickly reading the code of the draw, it seems useless to save the color vector since calling `invalidateOpenGLBuffers()` will trigger a call to `compute_normals_and_vertices()` that will fill (overwrite) that vector. Simply drop the local color vector and the patch should be sufficient.

maxGimeno: You did the same mistake I did. color_vector() returns the vector Colors_
which is not the vector that is cleared with invalidateOpenGLBuffers but
the color_map that gives the color of the facets according to its patch_id.

2016-06-27 15:31 GMT+02:00 Sebastien Loriot <notifications@github.com>:

PR #1211, opened 2016-06-27
Create an example with hybrid mesh domain

lrineau:
This PR will had an example for Mesh_3 that shows how to create an hybrid mesh domain class, that combines a polyhedral surface (here a cube), with an implicit surface (here a sphere).

with-custom-init1

For the moment, the example does not protect:
- the sharp edges of the cube,
- the intersection curves between the sphere and the cube.

For that reason, the resulting mesh is not fully correct.
with-custom-init1

TODO:
- [x] protect sharp edges
- [x] add the documentation

@sloriot @janetournois @afabri I would need that PR to be tried and compiled on Windows as soon as possible.


afabri:
I just tested it with VC++.  There are no warnings, and it produces the right result.

lrineau:
After commit 997b15bf1abdad8d2ced135930f4f0de769354ea, the example now has 1D features.
with-custom-init1

lrineau:
I would appreciate feedback. The commit https://github.com/CGAL/cgal/commit/997b15bf1abdad8d2ced135930f4f0de769354ea is really not pretty. How should I document that?

- I can extract the code to a dedicated function, in a dedicated header file, and hide the details in the documentation.
- I can actually do the same for the domain class definition (see the long definition of `Hybrid_domain`, in the file [`mesh_hybrid_mesh_domain.cpp`][f]).
[f]: https://github.com/lrineau/cgal/blob/997b15bf1abdad8d2ced135930f4f0de769354ea/Mesh_3/examples/Mesh_3/mesh_hybrid_mesh_domain.cpp
- If I extract some code to dedicated functions and header files, should I make them template code?
- More generally, I wonder what I should enlighten in the documentation. Please comment as much as you want (@sloriot @janetournois @afabri).

janetournois:
I think the code of the `Hybrid_domain` IS the example, so it should not be hidden. In particular, it is a synthetic and "simple" example, so I don't see why it should be moved to an internal header file. It's not generic enough.

Ideally, it would be nice to have a complete `Hybrid_domain<MD1, MD2, BooleanOperation>` class that takes as template parameter whether the domain should be the union, intersection, or difference, of the two input domains. The implementation in this example would then be the special case `Hybrid_domain<Polyhedral_domain, Implicit_domain, Union>`.
Let's drop the `BooleanOperation` for now, but a dedicated CGAL header file should contain `Hybrid_domain<MD1, MD2>`, imho.

sloriot:
I postpone this PR to 4.10 as it seems a new class in `Mesh_3` should be added as a small feature.

lrineau:
That class is an "example" of user-made domain class. It is in Mesh_3 *examples* on purpose.

lrineau:
I eventually with @janetournois, from here comment https://github.com/CGAL/cgal/pull/1211#issuecomment-229086483: the hybrid class header belongs to examples.

sloriot:
OK, will one of you have time to write a small paragraph in the doc? Considering its length, I don't see this example fully inlined. A link to the file seems enough to me.

janetournois:
I will take care of it by the end of the week

janetournois:
I started adding a little doc, and rebased the branch.
Images coming soon...

sloriot:
@lrineau Branch size is 855K. Converting the new png figures to jpg divides the size by 4. Shall I replace them?

sloriot:
See https://github.com/sloriot/CGAL/tree/Mesh_3-hybrid_mesh_domain-GF

janetournois:
sure, go ahead! Thanks @sloriot

lrineau:
@sloriot I agree, the quality of the compressed JPEG is equivalent to the original files.

sloriot:
Done. You need to hard reset your local branch.

janetournois:
this PR should be ready for testing

PR #1219, opened 2016-06-29
Polyhedron_demo : Trivial UI change in Point_inside_polyhedron_plugin

maxGimeno:
This PR adds a spacer to the UI of Point_inside_polyhedron_plugin, so it stays compact when the widget is resized.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Point_inside_polyhedron_widget.ui
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Point_inside_polyhedron_widget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>542</width>
-    <height>258</height>
+    <width>340</width>
+    <height>225</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Plugins/PMP/Point_inside_polyhedron_widget.ui
```diff
@@ -87,6 +87,19 @@
...
     </item>
+    <item>
+     <spacer name="verticalSpacer">
+      <property name="orientation">
+       <enum>Qt::Vertical</enum>
+      </property>
+      <property name="sizeHint" stdset="0">
+       <size>
+        <width>20</width>
+        <height>40</height>
```
lrineau: Why should that spacer have a height? That wastes vertical space.

afabri: Because the Note with the floating text is really ugly.
capture

lrineau: I understand the need for a space, when there is a lot of space, so that the note and the text are at the top of the Window. But, when the dockwidget is reduced as much as possible, why let the space take a minimal height of 20 ?

lrineau: It should be `<height>0</height>`.

maxGimeno: It's just the automatic generation, I'll check if I can remove it.

sloriot:
locally tested

PR #1221, opened 2016-06-29
Polyhedron_demo: Fix Point_set_smoothing_plugin

maxGimeno:
Fixes the SEGFAULT by defining mw in the plugin's `init()` function, as it was missing.


sloriot:
Locally tested

PR #1224, opened 2016-06-30
Polyhedron_demo: Point_set_interference_plugin

maxGimeno:
This PR proposes a new plugin that allows to create noise in a point set.

 @afabri if this is what you wanted, please confirm this is OK for merging.


sloriot:
@afabri ping

afabri:
That's exactly what I want. Thanks a lot.

PR #1229, opened 2016-07-01
Make it work with Epeck

sloriot:
The branch contains several fixes, all of them are included in dedicated commit to ease the review.
@sgiraudot could you make a review? In particular 29aacb5 that remove the use of `numeric_limits`.

@afabri what do you think of 3009712?


Optimal_transportation_reconstruction_2/include/CGAL/OTR_2/Reconstruction_triangulation_2.h
```diff
@@ -983,6 +983,26 @@ class Reconstruction_triangulation_2
...
+  }
+
+  bool is_infinity(const std::pair<bool,boost::optional<FT> >& p) const
+  {
+    return p.second==boost::none;
+  }
+
+  std::pair<bool,boost::optional<FT> > m_infinity() const
+  {
+    return std::pair<bool,boost::optional<FT> >(0,boost::optional<FT>());
```
sgiraudot: Why not use directly `false` instead of 0 if it's a boolean?

PR #1239, opened 2016-07-04
Polyhedron_demo: Fix the Normal Drawing

### LINKED ISSUE #1232: drawing the normals
They should have length  min(average_spacing, the_now_fixed_length)   @maxGimeno

maxGimeno:
This PR fixes #1232 .
It sets a maximum value for the normal lengths that is the average spacing of the points set.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
@@ -10,6 +10,7 @@
 #include <CGAL/IO/write_xyz_points.h>
 #include <CGAL/Timer.h>
 #include <CGAL/Memory_sizer.h>
+#include <CGAL/compute_average_spacing.h>
 
 #include <CGAL/Three/Viewer_interface.h>
 #include <CGAL/Orthogonal_k_neighbor_search.h>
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
```diff
@@ -246,7 +247,18 @@ void Scene_points_with_normal_item_priv::compute_normals_and_vertices() const
...
+       typedef CGAL::Parallel_tag Concurrency_tag;
+#else
+        typedef CGAL::Sequential_tag Concurrency_tag;
+#endif
+
+        double average_spacing = CGAL::compute_average_spacing<Concurrency_tag>(
+              points.begin(), points.end(),
+              6);
+
+        float normal_length = (float)std::min(average_spacing, std::sqrt(region_of_interest.squared_radius() / 1000.));
```
sloriot: this will not compile on windows. Use:
`(float)(std::min)(average_spacing, std::sqrt(region_of_interest.squared_radius() / 1000.));`

sloriot: is there any reason to use `float` and not `double` for `normal_length`?

PR #1240, opened 2016-07-04
Polyhedron_demo : Submenu for Point Set Processing Operations

### LINKED ISSUE #1231: Operations -> Point Set Processing
@maxGimeno   The operations marked red should be in a sub menu entitled "Point Set Processing". 
Point Set Simplification does NOT go there because it is a proper package in the manual.

In the submenu shall be no prefix "Point Set"
point_set_processing

maxGimeno:
This PR fixes #1231.


sloriot:
`Point Set Normal Estimation` and  `Point Set Simplification Selection` should also be in the submenu

PR #1244, opened 2016-07-06
Polyhedron_demo: Restore snapshots 

### LINKED ISSUE #1208: Polyhedron demo : save snapshot broken
- open anything you like in polyhedron demo
- do File -> save snapshot
- open the image in your favorite viewer : it's all gray

@maxGimeno can you please have a look at this? we really need it to prepare slides

maxGimeno:
This PR fixes #1208 .
It restores the non-vectorial snapshots of the demo the way they were in 4.6.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.h
+++ b/Polyhedron/demo/Polyhedron/Viewer.h
@@ -99,7 +99,6 @@ protected:
   void postDraw();
   void paintEvent(QPaintEvent *);
   void paintGL();
-  void setFrustum(double l, double r, double t, double b, double n, double f);
   //! Holds useful data to draw the axis system
   struct AxisData
   {
@@ -175,7 +174,6 @@ private:
   QString message;
   bool _displayMessage;
   QTimer messageTimer;
-  QMatrix4x4 projectionMatrix;
 }; // end class Viewer
 
 
--- a/Polyhedron/demo/Polyhedron/Viewer.h
+++ b/Polyhedron/demo/Polyhedron/Viewer.h
@@ -99,6 +99,7 @@ protected:
   void postDraw();
   void paintEvent(QPaintEvent *);
   void paintGL();
+  void setFrustum(double l, double r, double t, double b, double n, double f);
   //! Holds useful data to draw the axis system
   struct AxisData
   {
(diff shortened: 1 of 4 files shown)


sloriot:
There should be an option to preserve the aspect ratio.

Polyhedron/demo/Polyhedron/Viewer.h
```diff
@@ -174,6 +175,7 @@ public Q_SLOTS:
   QString message;
   bool _displayMessage;
   QTimer messageTimer;
+  QMatrix4x4 projectionMatrix;
```
lrineau: I think the `Viewer` class should use a d-pointer, like the other ones in our framework, to avoid that modifications for internal reasons, like this one, have an impact on the `Viewer.h`.

sloriot: @lrineau I suggest that this is done in another PR. Shall you open an issue for that?

lrineau: I agree. Issue #1248.

sloriot:
Use <i>Camel Case</i> for the text in the dialog

sloriot:
I don't know what the difference when "use white background is  checked". I don't know what the option on the frustum is useful for.

Polyhedron/demo/Polyhedron/Viewer.h
lrineau: See commit https://github.com/CGAL/cgal/pull/1244/commits/5d1da468bf758415d7b6f6124c605a040683dd48. There was already a d-pointer for `Viewer`.

maxGimeno:
@sloriot  The background option makes the snapshot's background  white  even if it is not in the scene (View->Set Background Color) and the option for the Frustum is used to avoid deforming the scene when the ratio of the snapshot is not the same as the one of the viewer.

lrineau:
A side note: do not mix "[CamelCase][cc]", which is a naming convention for identifiers, with [rules for capitalization of words in English titles][e].

[cc]: https://en.wikipedia.org/wiki/CamelCase
[e]: https://www.google.com/search?q=capitalize+titles+in+english

afabri:
settings -> Settings in the dialog. I've sent a mail to Maxime.

The snapshop printing works with VC++.

janetournois:
Than you!

PR #1252, opened 2016-07-07
Add free function scalar_product(Vector_{2,3}, Vector_{2,3})

afabri:
We have  operator* and a kernel functor, but not the free function.

### DIFF
--- a/Kernel_23/doc/Kernel_23/CGAL/Kernel/global_functions.h
+++ b/Kernel_23/doc/Kernel_23/CGAL/Kernel/global_functions.h
@@ -2366,9 +2366,15 @@ const CGAL::Point_2<Kernel> &r);
 returns the scalar product of `u` and `v`.
 */
 template <typename Kernel>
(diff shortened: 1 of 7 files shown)


Kernel_23/doc/Kernel_23/CGAL/Kernel/global_functions.h
```diff
@@ -2357,6 +2357,20 @@ const CGAL::Point_2<Kernel> &r);
...
+/// \ingroup kernel_global_function
+/// @{
+
+/*!
+returns the scalar product of `u` and `v`.
+*/
+template <typename Kernel>
+Kernel::FT scalar_product( const CGAL::Vector_3<Kernel>& u,
+                           const CGAL::Vector_3<Kernel>& v );
+
```
sloriot: The 2D version is not documented

sloriot:
I labeled it as a bug since we have `cross_product`

PR #1255, opened 2016-07-08
Add a license check mechanism

afabri:
Small feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/License_check

User who have a commercial license for specific packages can check that  they do not accidentally use packages for which they do not have a commercial license. The same holds for users who want to be sure that they only use packages of CGAL released under the LGPL

To enable checking users have to define one of the following macros:

| Macro Name | Effect |
| :-- | :-- |
| `CGAL_LICENSE_WARNING` | get a warning during the compilation |
| `CGAL_LICENSE_ERROR` | get an error during the compilation |

The license checking is not a means to control users, and no information
is transmitted or collected to others.
### TODO
- [x] changes.html
- [x] Update `test_merge_of_branch` to test for the presence of the include license in GPL files


Installation/include/CGAL/config.h
```diff
@@ -439,4 +439,7 @@ typedef const void * Nullptr_t;   // Anticipate C++0x's std::nullptr_t
 
 } //namespace CGAL
 
+#include <CGAL/license_check/lgpl.h>
+xb
```
sloriot: xb?

Scripts/developer_scripts/add_license_in_pkg_header.py
```diff
@@ -0,0 +1,50 @@
...
+from sys import argv
+import os
+from os.path import join
+import codecs
+import re
+
+def add_license_include_in_file(package_name, fname):
+  # first, see if the include directive is already there
+  with codecs.open(fname, encoding='utf-8') as f:
+    if any(re.search("CGAL/license/"+package_name+".h", line) for line in f):
```
lrineau: I would search for `#include\s*<CGAL/license/` instead. If `package_name` is, for example, `Mesh_2`, but a file already includes another license checking header, like:

    #include <CGAL/license/Delaunay_conforming_2.h>

then that would be preferable that the file is skipped.

Scripts/developer_scripts/add_license_in_pkg_header.py
```diff
@@ -0,0 +1,50 @@
...
+
+from sys import argv
+import os
+from os.path import join
+import codecs
+import re
+
+def add_license_include_in_file(package_name, fname):
+  # first, see if the include directive is already there
+  with codecs.open(fname, encoding='utf-8') as f:
```
lrineau: What if the file is not encoded in utf-8? I would open the file without any encoding/decoding, to interpret the file as a sequence of bytes.

I also wonder what will happen with the end-of-lines style of the file.

sloriot: Why would the file not encoded in utf-8?

About the end-of-line, they will be transformed by git if they are not correct.

Scripts/developer_scripts/add_license_in_pkg_header.py
```diff
@@ -0,0 +1,50 @@
...
+    if any(re.search("CGAL/license/"+package_name+".h", line) for line in f):
+      return # instruction already there
+
+  # pattern is in the file, so perform replace operation.
+  inserted = False
+  with codecs.open(fname, encoding='utf-8') as f:
+    out_fname = fname + ".tmp"
+    out = codecs.open(out_fname, "w", encoding='utf-8')
+    for line in f:
+      if not inserted and re.search("#\s*define", line):
```
lrineau: Here I would search for a more explicit `#\s*define\s*CGAL.*_H`. Otherwise special files like `Kernel_23/include/CGAL/Kernel/interface_macros.h` will get corrupted.

Scripts/developer_scripts/add_license_in_pkg_header.py
lrineau: You are actually quite right. All our files are correctly encoded in ascii or utf-8, but one file:

- `Mesh_2/include/CGAL/Mesh_2/Mesh_global_optimizer_2.h` (coded with ISO-8859)

I have verified with:

    file ^build*/(doc|examples|test|demo|include)/**/*.(h*|cpp) | grep -vE '(UTF-8 Unicode|ASCII) text(, with very long lines)?$'

And you are also right that Git seems to translate the files end-of-lines, because all of them have Unix end-of-lines on my machine.

Scripts/developer_scripts/add_license_in_pkg_header.py
sloriot: What if the define macro is not correctly written? (missing CGAL_ of _H)

sloriot:
- [ ] Update `test_merge_of_branch` to test for the presence of the include license in GPL files
- [ ] Handle particular cases when a package can be split into subpackages

lrineau:
I have copied the TODO list in the first message https://github.com/CGAL/cgal/pull/1255#issue-164516031.

Scripts/developer_scripts/test_merge_of_branch
```diff
@@ -115,6 +117,15 @@ if [ -n "${project_name_demo}" ]; then
   exit 1
 fi
 
+#check no file contains non-utf8 characters
+echo '.. Checking if non utf-8 characters are used...'
+txt_not_utf8=$(for i in `find . -type f`; do out=`file $i`; if echo $out | grep -q "text"; then if ! (echo $out | egrep -q "UTF-8|ASCII|XML|EPS"); then echo $i; fi; fi; done)
```
lrineau: This loop is horrible! In my Git repository, including the build directories, I have 280440 files! For each file, at least one `file` and one `grep` is forked. That runs forever!

Scripts/developer_scripts/test_merge_of_branch
```diff
@@ -129,7 +129,7 @@ fi
 
 #check no file contains non-utf8 characters
 echo '.. Checking if non utf-8 characters are used...'
-txt_not_utf8=$(for i in `find . -type f`; do out=`file $i`; if echo $out | grep -q "text"; then if ! (echo $out | egrep -q "UTF-8|ASCII|XML|EPS"); then echo $i; fi; fi; done)
+txt_not_utf8=$(file -N `git ls-files --stage` | grep "text" | egrep -v "UTF-8|ASCII|XML|EPS")
```
lrineau: This patch replaces:
- the `find . -type f` by `git ls-files --stage`: in my working directory, I have more than 200k files, and it falls down to 15k files,
- avoid to fork a `file` and a `grep` for each of them.

This patch probably only works on Linux, where the command line is not limited. A better patch should use `git ls-files --stage | xargs file | grep ...`.

lrineau:
This PR leads to red cells in the `Installation` line on Windows:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-106.shtml#Installation
```
Compiling license_check_gpl ... 

Microsoft (R) Program Maintenance Utility Version 12.00.21005.1
Copyright (C) Microsoft Corporation.  All rights reserved.
[...]
LINK : fatal error LNK1104: cannot open file 'tbb.lib'
```

lrineau:
Nef_2 is red too:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-106.shtml#Nef_2

sloriot:
Nef2 is fixed.

lrineau:
For TBB, I suggest to define `__TBB_NO_IMPLICIT_LINKAGE` to `1`, in the cpp files, to avoid the auto-linking of TBB.

lrineau:
... well, maybe that is not related.

lrineau:
Maybe the following:
```C++
#include <CGAL/config.h>
#undef CGAL_LINKED_WITH_TBB
#include <CGAL/AABB_tree.h>
```

lrineau:
Both the `#undef` `CGAL_LINKED_WITH_TBB` or the definition of `__TBB_NO_IMPLICIT_LINKAGE` to `1` should work. The idea is to avoid that `<tbb.h>` is included (included by `Profiling_tools`), or to disable the auto-linking.

sloriot:
I simply removed these useless files

Installation/test/Installation/license_check_gpl.cpp
```diff
@@ -1,6 +0,0 @@
-// #define CGAL_AABB_TREE_COMMERCIAL_LICENSE 0
-// #define CGAL_LICENSE_WARNING
-// #define CGAL_LICENSE_ERROR
-#include <CGAL/AABB_tree.h>
-
-int main() {}
```
lrineau: And the CMakeListst.txt?
```cmake
  create_single_source_cgal_program( "license_check.cpp" )
  create_single_source_cgal_program( "license_check_gpl.cpp" )
```

PR #1257, opened 2016-07-08
Use mpq_class rather than Gmpq if gmpxx is available.

mglisse:
This needs testing / benchmarking before merging. Last time I tried, all examples I ran were slightly faster with `mpq_class` than `Gmpq` or the same speed. Another advantage is that with `CGAL_DONT_USE_LAZY_KERNEL` and `CGAL_USE_GMPXX`, this makes Epeck thread-safe.

TODO
--------
 - [x] benchmarks

### DIFF
--- a/Number_types/include/CGAL/Exact_integer.h
+++ b/Number_types/include/CGAL/Exact_integer.h
@@ -23,7 +23,9 @@
 // Author(s)     : Laurent Rineau
 
 #include <CGAL/config.h>
-#if CGAL_USE_GMP
+#if CGAL_USE_GMPXX
+#  include <CGAL/gmpxx.h>
+#elif CGAL_USE_GMP
 #  include <CGAL/Gmpz.h>
 #elif CGAL_USE_LEDA
 #  include <CGAL/leda_integer.h>
@@ -54,7 +56,11 @@ typedef unspecified_type Exact_integer;
 
 #else // not DOXYGEN_RUNNING
 
-#if CGAL_USE_GMP
+#if CGAL_USE_GMPXX
+
+typedef mpz_class Exact_integer;
+
+#elif CGAL_USE_GMP
 
 typedef Gmpz Exact_integer;
 
--- a/Number_types/include/CGAL/internal/Exact_type_selector.h
+++ b/Number_types/include/CGAL/internal/Exact_type_selector.h
@@ -57,11 +57,13 @@ namespace CGAL { namespace internal {
 
 // Two classes which tell the prefered "exact number types" corresponding to a type.
 
-// The default template chooses Gmpq, or leda_rational, or Quotient<MP_Float>.
+// The default template chooses mpq_class, Gmpq, leda_rational, or Quotient<MP_Float>.
 // It should support the built-in types.
 template < typename >
 struct Exact_field_selector
-#ifdef CGAL_USE_GMP
+#ifdef CGAL_USE_GMPXX
+{ typedef mpq_class Type; };
+#elif defined(CGAL_USE_GMP)
 { typedef Gmpq Type; };
 #elif defined(CGAL_USE_LEDA)
 { typedef leda_rational Type; };


lrineau:
It seems that clang-3.5 does not like gmpxx. Was that expected?
See https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-97/Kernel_23/TestReport_lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8.gz

mglisse:
No, that wasn't expected, but it is useful to learn. I am getting the same error with more recent versions of clang, I am surprised you don't see it more in the testsuite, but few platforms seem to have gmpxx. I'll investigate (it is likely trivial to fix) when I get the time, but possibly not before the holidays. This PR can go to sleep in the mean time.

lrineau:
Indeed. I wonder who I could have forgot to fix the same issue in `Vector_2`. Sorry. I suggest you include the fix in this PR. That will be easier to test and integrate in `master`.

mglisse:
It might be better to push the Vector_2 fix separately: it is a trivial bugfix that should go in ASAP, while this PR requires a bit of benchmarking to determine the performance impact.

lrineau:
That branch seems to cause compilation errors on one platform:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-15/AABB_tree/TestReport_lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8.gz

mglisse:
That "one platform" actually means all the platforms that define CGAL_USE_GMPXX... Yes, the testsuite provides limited testing for it.
Coercion_traits are a bit painful. I need to check what our policy is there. Most types seem to have specializations for short, int, long, float, double, so I guess I'll use that list for mpq_class as well.

lrineau:
What's the problem with `CGAL_USE_GMPXX`? Could it be set by default (assuming CMake has detected `GMPXX`)?

mglisse:
IIRC it is already set by default, I think it's just that most testsuite machines don't have it installed...

mglisse:
See #1551 (again done in a separate branch, since it is a bugfix).

lrineau:
As for #1551, did not you want to submit it against 4.9.1?

mglisse:
I hesitated a bit, but it didn't seem worth it. Feel free to retarget it if you want.

lrineau:
The column that uses GMPXX has compilation errors, like this one:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-28/Circular_kernel_2/TestReport_lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8.gz
```
In file included from /no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/cmake/platforms/x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8/test/Circular_kernel_2/test_Lazy_circular_kernel.cpp:29:
In file included from /no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Exact_circular_kernel_2.h:30:
In file included from /no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Algebraic_kernel_for_circles_2_2.h:30:
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Root_of_traits.h:188:20: error: no viable conversion from 'Sqrt_extension<class CGAL::Quotient<class __gmp_expr<mpz_t, mpz_t> >, class CGAL::Quotient<class __gmp_expr<mpz_t, mpz_t> >, [2 * ...]>' to 'Sqrt_extension<class __gmp_expr<mpq_t, mpq_t>, class __gmp_expr<mpq_t, mpq_t>, [2 * ...]>'
            return make_root_of_2<RT>(a_,b_,c_,smaller);
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Root_of_traits.h:50:12: note: in instantiation of member function 'CGAL::internal::Root_of_traits_helper<__gmp_expr<mpq_t, mpq_t>, CGAL::Field_tag>::Make_root_of_2::operator()' requested here
    return make_root_of_2(a,b,c,smaller);
           ^
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Root_of_traits_specializations.h:64:30: note: in instantiation of function template specialization 'CGAL::make_root_of_2<__gmp_expr<mpq_t, mpq_t> >' requested here
          this->et = new RO2(make_root_of_2(op1.exact(), op2.exact(),
                             ^
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Root_of_traits_specializations.h:97:24: note: in instantiation of member function 'CGAL::Lazy_exact_ro2<__gmp_expr<mpq_t, mpq_t> >::update_exact' requested here
            return new Lazy_exact_ro2<NT>(a, b, c);
                       ^
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Root_of_traits.h:41:12: note: in instantiation of member function 'CGAL::Root_of_traits<CGAL::Lazy_exact_nt<__gmp_expr<mpq_t, mpq_t> > >::Make_root_of_2::operator()' requested here
    return make_root_of_2(a,b,c);
           ^
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Algebraic_kernel_for_circles/internal_functions_on_roots_and_polynomials_2_2.h:82:32: note: in instantiation of function template specialization 'CGAL::make_root_of_2<CGAL::Lazy_exact_nt<__gmp_expr<mpq_t, mpq_t> > >' requested here
                               make_root_of_2(y_base, y_root_coeff, disc)),
                               ^
/no-raid/home/lrineau/CGAL/CGAL-4.10-Ic-28/include/CGAL/Algebraic_kernel_for_circles/function_objects_on_roots_and_polynomials_2_2.h:51:33: note: (skipping 2 contexts in backtrace; use -ftemplate-backtrace-limit=0 to see all)
    { return AlgebraicFunctors::solve<AK> ( e1, e2, res); }
                                ^
```
There are others. See https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-28.shtml

mglisse:
Indeed, I thought I had written about it somewhere but can't find it anymore. I can either specialize

    template <>
    struct Get_rational_type<mpz_class,false>{
      typedef mpq_class type;
    };

(protected by CGAL_USE_GMPXX) in Root_of_traits.h or define an arithmetic kernel for mpz_class/mpq_class. Any preference?
I had also missed one testcase, issue #1604.
The rest seem to be runtime errors, most of which are likely from the already discussed I/O issues.

lrineau:
I would prefer you provide an arithmetic kernel.

(Cc @sloriot)
Actually, I do not understand `internal::Get_rational_type` in `Root_of_traits`:
- either `Get_arithmetic_kernel` should directly provide a nested boolean constant `is_specialized` (similar to std::numeric_limits<T>`,
- and/or there should always be a minimal arithmetic kernel that provides defaults depending on `Algebraic_structure_traits<NT>::Algebraic_category` (and the default for an `Euclidean_ring_tag` would be to provide `Quotient<NT>` as `Rational`).

lrineau:
This PR as well as #1609 were tested last night. There are compilation errors and runtime errors.

In particular in [a compilation error in `test/Number_types`][1].

[1]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-31/Number_types/TestReport_lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8.gz

lrineau:
I cannot reproduce the compilation error locally with clang++-3.8.

mglisse:
I can reproduce the root_of_2 error (even with just the arithmetic kernel branch), it is a trivial adaptation. As the comment shows, it is explicitly working around the lack of arithmetic kernel, so I can easily remove the workaround. However, this gives a runtime error that I need to investigate.
As explained, most of the runtime errors are expected, they are cause by I/O (operator>> for mpq_class does not read "12.34", but that's what Lazy_exact_nt<mpq_class> prints, so we cannot round trip).

lrineau:
The `operator>>` for `CGAL::Point_3` with `Cartesian_tag` uses:
```C++
is >> iformat(x) >> iformat(y) >> iformat(z);
```

Could we use a specialization of  `CGAL::Input_rep<mpq_class>` (defined in `<CGAL/IO/io.h>`) to implement a special input method for `mpq_class` that would allow the reading of `12.34`?

lrineau:
We could use `CGAL::internal::read_float_or_quotient` (from `<CGAL/IO/io.h>`) introduced by Andreas in 1baa32971131b43bb5256cdd76ffe0963cf96dc8 for this same reason.
```c++
template <typename Int, typename Rat>
inline void read_float_or_quotient(std::istream& is, Rat &z)
```

One problem: that function assumes a constructor `Rat(Int, Int)` but that can easily be worked around, with a division, assuming that `Rat(int)` exists: `Rat(n) / d`.

(Actually, I do not understand how that our current testsuite can even work, knowing that `Quotient<MP_Float>` cannot ready `12.34` either.)

mglisse:
internal::read_float_or_quotient: that's what I suggested in https://github.com/CGAL/cgal/pull/1551#issuecomment-254011150 :smile:
Input_rep: from the same comment: "We could also make sure operator>>(*,Point_2&) does the right thing. But at some point, users may read numbers directly and it becomes impossible to avoid the fact that mpq_class will not accept 12.34". But yes, such a specialization would be useful.
mpq_class has a constructor from 2 mpz_class, but if the 2 integers are not canonical, you are supposed to use Fraction_traits::Compose instead.
For Quotient<MP_Float>, we first parse a numerator (MP_Float uses operator>> on a double, so it can read 12.34), and since the next character is not '/' we stop there, so it works, although since we go through double, we don't get a super precision.

lrineau:
Then I would go for the use of `internal::read_float_or_quotient`. We want the code to work with the existing testsuite. I users decide to read numbers by themselves, then they are by themselves: that is out of the scope of CGAL.

Thanks for the explanation about `Quotient<MP_Float>`.

lrineau:
I just merged #1618 because its code is fine, and the [`Number_types` testsuite did well in CGAL-4.10-Ic-33][*]. But there are still other runtime or compilation errors.

[*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-33.shtml#Number_types

lrineau:
Hi Marc. Our guidelines are quite wrong. Please rebase your branch onto `master`, so that the changes are easier to track in this PR.

mglisse:
" Files changed 3,256" indeed...
git doesn't like me trying to push the rebased branch. I guess I would have to force it, but would you mind doing the rebase so I don't break everything?

lrineau:
Done. You will have to `git pull --rebase` yourself, or use `git reset --hard mine/Number_types-gmpxx-glisse`.

mglisse:
Thanks.

mglisse:
> But there are still other runtime or compilation errors.

Ah, I didn't test operator>> for Lazy_exact_nt<mpq_class>. It does call read_float_or_quotient, but in a way that doesn't help for mpq_class. I wonder if I should make operator>> for Lazy_exact_nt call the right version of read_float_or_quotient, or make read_float_or_quotient auto-detect fraction types, or specialize it for mpq_class as it is done for leda_rational, or the simplest make operator>> for Lazy_exact_nt call <code>is>>iformat(e)</code> instead of calling directly read_float_or_quotient, which would remove the only call to that overload of read_float_or_quotient. I'll probably imitate leda_rational for now, although that doesn't look super nice.

mglisse:
With the patch in branch Number_types-gmpxx_IO2-glisse (I'll PR once the last issue is fixed), and ignoring the QP_Solver error that also exists without gmpxx, that only leaves the compilation failure of test_nef_3_io_EPEC.cpp which contains:

    typename Kernel::RT n( std::string("6369051672525773"));

Where RT is Lazy_exact_nt<mpq_class>. With Lazy_exact_nt<Gmpq>, we use the constructor <code>Lazy_exact_nt (const ET & e)</code>. However, mpq_class's constructor from std::string is <code>explicit</code>, so this doesn't work. We could try to add an explicit template constructor to Lazy_exact_nt that uses sfinae to accept only types for which ET has explicit (but not implicit) construction (and possibly check that they are not enum/arithmetic to avoid ambiguities). Or we could change the testcase. Now that I/O works, I guess the easiest would be:

    typename Kernel::RT n;
    std::istringstream("6369051672525773") >> n;

I vote for this last solution, it is easier. Do you agree?

lrineau:
Yes, go on,

mglisse:
Testsuite is expected to be ok once #1627 gets merged. And we can finally benchmark to find out if there was a point to doing all this work...

lrineau:
The tests now seem good. Do you need extra work before this PR is merged?

mglisse:
#1627 can be merged to master. For #1257, we first need some benchmark to justify the switch, unless you consider that making CGAL_DONT_USE_LAZY_KERNEL Epeck thread-safe is sufficient (but even in that case, we would at least want to check that performance doesn't regress). Can you think of some good examples to check for performance?

lrineau:
I have copied that TODO list to the top message as well.

lrineau:
The last edit of this PR was 2 weeks ago. Still todo: the benchmarks!

What is the plan? How is in charge of benchmarking?

mglisse:
Anyone with an application where Gmpq/Gmpz is a large part of the running time is welcome to try the branch and post their results. I'll do some minimal tests at some point, but I don't have time these days.

sloriot:
@mglisse What about doing a 3D Delaunay triangulation with `Simple_cartesian<Gmpq/Gmpz>`?

mglisse:
@sloriot benchmark/Triangulation_3/simple_2 1000 takes about 1.4s with Simple_cartesian<Gmpq> and 1.17s with Simple_cartesian<mpq_class>. That may not be representative of what everyone is doing with CGAL::Rational and CGAL::Integer though... In some packages, I fear that people took advantage of the fact that Gmpq was reference-counted to copy numbers often and all over the place.

sloriot:
@lrineau  this PR seems ready.

mglisse:
I was thinking maybe merge it at the beginning of the 4.11 cycle instead of late in 4.10, because it can potentially break more code (not covered by the testsuite) and regress performance for some algebraic stuff that copies numbers a lot, so this would give people more time to notice issues. But if you feel confident about putting it in 4.10, that's ok with me.

lrineau:
Hi Marc. Would you agree that this PR can now be merged in `master` (targetting CGAL-4.11-beta1, planned by the end of July)?

mglisse:
Yes.

lrineau:
It is now merged in `master`.

PR #1259, opened 2016-07-11
Restrict the HalfedgeGraph with properties often assumed in our functions

sloriot:


BGL/doc/BGL/Concepts/HalfedgeGraph.h
```diff
@@ -55,5 +55,11 @@ Expression                              | Returns
 `next(h, g)`                            | `halfedge_descriptor`                                                        | The next halfedge around its face.
 `prev(h, g)`                            | `halfedge_descriptor`                                                        | The previous halfedge around its face.
 `boost::graph_traits<G>::%null_halfedge()` | `halfedge_descriptor`                                                     | Returns a special halfedge that is not equal to any other halfedge.
+
+\cgalHeading{Invariants}
+
+`halfedge(edge(h,g))==h`
```
lrineau: How can that invariant be true for both `h` and `opposite(h, g)`?! They share the same edge.

sloriot: Similarly as the BGL concept IncidentEdge. An out-edge of a vertex `v` must guarantee that the source vertex is the vertex `v`. However if you consider the two out-edges from the vertex of any edge, the edges are equal while they obviously won't provide the same source vertex.

sloriot: In practice, the `edge_descriptor` store the `halfedge_descriptor`

gdamiand: I think like @lrineau , this invariant cannot be true for all halfedges because each edge is associated with only one of its halfedge.

sloriot: It is for Surface_mesh. Polyhedron and OpenMesh models. Like it is in boost for all `IncidentEdge` models

lrineau: Le Monday 11 July 2016 09:22:28 Guillaume Damiand a écrit :
> > +`halfedge(edge(h,g))==h`
> 
> I think like @lrineau , this invariant cannot be true for all halfedges
> because each edge is associated with only one of its halfedge.

As Sébastien pointed out, that means that any "edge" has at least two 
different "oriented" representations, that are considered as equal by 
`operator==`.

-- 
Laurent Rineau, PhD
R&D Engineer at GeometryFactory           http://www.geometryfactory.com/
Release Manager of the CGAL Project       http://www.cgal.org/

gdamiand: In this case, it is better to write edge(h,g)==edge(opposite(h,g),g).

sloriot: No, the invariant does not impose the same thing. It imposes that the source and target of an edge won't be swapt.

PR #1263, opened 2016-07-12
Polyhedron_demo: Fix reload item

### LINKED ISSUE #1262: Polyhedron_demo : Problem with `reload_item()`
Each time an item is reloaded, it takes more time to create it.

maxGimeno:
This PR fixes #1262 

Replacing the flat_maps used for the vertex normal map and the face normal map by another type of map fixes the problem.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
@@ -674,11 +674,11 @@ Scene_polyhedron_item_priv::compute_normals_and_vertices(const bool colors_only)
     typedef boost::graph_traits<Polyhedron>::face_descriptor   face_descriptor;
     typedef boost::graph_traits<Polyhedron>::vertex_descriptor vertex_descriptor;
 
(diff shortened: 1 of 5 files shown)


sloriot:
So the load itself was slow, right?

maxGimeno:
Not really, but when you had a big item (100K vertices), if you called
reload item twice, the second time took like 10 times more to compute the
normals.

2016-07-12 15:51 GMT+02:00 Sebastien Loriot <notifications@github.com>:

afabri:
Can you please check if we have more occurrences of flat_map in other CGAL packages, and if they have an impact.

lrineau:
No. Should we? There are use cases where the flat map can be actually a lot faster: when build at once from a range, and then queried multiple times. The use case in the scene polyhedron item was wrong, because the map was modified, and even reallocate, a lot of time.

Envoyé par BlueMail



Le 12 juil. 2016 18:56, à 18:56, Andreas Fabri <notifications@github.com> a écrit:

afabri:
They got introduced as the fastest map and  nobody realized that there is a potential problem. As far I can see it, only item classes in the demo are concerned, so please check.

maxGimeno:
It is done.

PR #1264, opened 2016-07-13
Polyhedron_demo: No std::cerr in `MainWindow::message()`

maxGimeno:
This PR removes the std::cerr in the function message() of the mainwindow. This way, the console will not print html messages anymore, but all the messages printed using this function will only be readable from the Console widget of the application.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -807,7 +807,6 @@ void MainWindow::message(QString message, QString colorName, QString font) {
   if (message.endsWith('\n')) {
     message.remove(message.length()-1, 1);
   }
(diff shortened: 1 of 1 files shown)


lrineau:
The PR has three commits, and does far more than removing a display to `std::cerr`.

What is more, I have seen that your branch namings are no longer right. Please keep following our [naming conventions for Git branches][nc]. That really help us to deal more easily with all the feature branches. This PR branch should have been named something like `Polyhedron-demo_no_cerr_in_mw_message-GF`.

[nc]: https://github.com/CGAL/cgal-dev/wiki/Guidelines-Source_Code_Management_GIT-Naming_recommendations

lrineau:
You fixed the commits while I was writing my last comment! :-)

About the branch naming convention, that is a reminder for next PR you make. Not for this PR.

PR #1265, opened 2016-07-13
Otr2 fix demo compilation

sloriot:

PR #1266, opened 2016-07-13
Property map fixes

sloriot:


Property_map/include/CGAL/property_map.h
```diff
@@ -274,26 +274,6 @@ Nth_of_tuple_property_map<N, Tuple>
   return Nth_of_tuple_property_map<N, Tuple>();
 }
 
-/// \ingroup PkgProperty_map
-/// Struct that turns a property map into a unary functor with
-/// `operator()(key k)` calling the get function with `k`
-template <class PropertyMap>
-struct Property_map_to_unary_function{
```
lrineau: 1. `Property_map_to_unary_function` is in CGAL-4.8 and documented. You cannot remove it that way.
2. `boost::property_map_function` does not seem documented.

I tend to object to that change, unless you have very good reasons.

Spatial_searching/examples/Spatial_searching/searching_with_point_with_info_inplace.cpp
```diff
@@ -11,29 +11,14 @@ typedef Kernel::Point_3 Point_3;
...
-  typedef Point key_type;
-  typedef boost::lvalue_property_map_tag category;  
-
-  My_point_property_map(const std::vector<Point_3>& pts):points(pts){}
-
-  reference operator[](key_type k) const {return points[k];}
-
-  friend reference get(const My_point_property_map& ppmap,key_type i) 
-  {return ppmap[i];}
-};
```
lrineau: That example was a good introduction to the definition of a property map that stores a pointer to a container. I would like we keep it as it was, for courses.

Point_set_processing_3/include/CGAL/Point_with_normal_3.h
```diff
@@ -138,17 +138,16 @@ struct Normal_of_point_with_normal_pmap
 
   typedef Point_with_normal key_type;
   typedef Vector value_type;
-  typedef value_type& reference;
+  typedef const value_type& reference;
   typedef boost::lvalue_property_map_tag category;
 
   /// Access a property map element
-  reference operator[](key_type& pwn) const { return pwn.normal(); }
+  value_type& operator[](key_type& pwn) const { return pwn.normal(); }
```
lrineau: If `reference` is `const value_type&`, then the lvalue property map is non-mutable. In that case, `pmap[key]` should return a `const value_type&`.

http://www.boost.org/doc/libs/1_61_0/libs/property_map/doc/LvaluePropertyMap.html

lrineau:
This PR breaks things. See https://github.com/CGAL/cgal/pull/1265#issuecomment-232382973.

Property_map/include/CGAL/property_map.h
```diff
@@ -173,7 +167,7 @@ struct First_of_pair_property_map
 {
   typedef Pair key_type; ///< typedef to `Pair`
   typedef typename Pair::first_type value_type; ///< typedef to `Pair::first_type`
-  typedef const value_type& reference; ///< typedef to `value_type&`
+  typedef value_type& reference; ///< typedef to `value_type&`
```
lrineau: `key_type` and `reference` should be related.

Using `boost::copy_cv`, they could copy the constness of `Pair`:
```
typedef Pair& key_type;
typedef typename Pair::first_type value_type;
typedef typename boost::copy_cv<typename Pair::first_type, Pair>::type& reference;
```

Property_map/include/CGAL/property_map.h
```diff
@@ -107,7 +101,7 @@ struct Dereference_property_map
 {
   typedef T* key_type; ///< typedef to 'T*'
   typedef T value_type; ///< typedef to 'T'
-  typedef const value_type& reference; ///< typedef to 'T&'
+  typedef value_type& reference; ///< typedef to 'T&'
```
lrineau: This works because it preserves the cv-qualifiers of `T`.

Point_set_processing_3/include/CGAL/Point_with_normal_3.h
```diff
@@ -147,7 +147,8 @@ struct Normal_of_point_with_normal_pmap
   typedef Normal_of_point_with_normal_pmap<Gt> Self;
   /// \name Put/get free functions
   /// @{
-  friend reference get(const Self&,const key_type& k) {return k.normal();}
+  friend const reference get(const Self&,const key_type& k) {return k.normal();}
```
lrineau: That code will break user-code that does:
```C++
  typedef typename boost::property_traits<Pmap>::reference reference;
  reference v = get(pmap, key);
```
That will break if `key` is `const`, because one cannot bind a const-ref to a ref.

sloriot: see the second overload

lrineau: I have seen the second overload, but anyway the result of the call to `get` will not bind to a non-const ref!

sloriot: I must miss something because I don't see the problem. Are you talking of `const reference` vs `const value_type&`?

lrineau:
```C++
const key_type key;
typedef typename boost::property_traits<Pmap>::reference reference;
reference v = get(pmap, key);
```

Unless I am completely wrong, that code will not compile because `reference` is a non-const `value_type&`, but the call to `get(pmap, key)` with the `const` key will return a const reference `const value_type&`...

**actually no!** `const reference` is not `const value_type&`! A reference is always const because its value (as a pointer) cannot be changed. `const reference` is the same type as `reference`. That code should not compile, because, with `const key_type& k`, `k.normal()` should not bind to a reference to a non-const value.

Try this code:
```C++
int main() {
  typedef int& reference;
  int i = 0;
  const reference r = i;
  r = 1;
  return r;
}
```

Point_set_processing_3/include/CGAL/Point_with_normal_3.h
```diff
@@ -147,7 +147,7 @@ struct Normal_of_point_with_normal_pmap
   typedef Normal_of_point_with_normal_pmap<Gt> Self;
   /// \name Put/get free functions
   /// @{
-  friend const reference get(const Self&,const key_type& k) {return k.normal();}
+  friend const value_type& get(const Self&,const key_type& k) {return k.normal();}
   friend reference get(const Self&, key_type& k) {return k.normal();}
   friend void put(const Self&,key_type& k, const value_type& v) {k.normal()=v;}
```
lrineau: `reference` should be `typedef value_type& reference;`, and both overloads should return the `reference` type.

sloriot:
After a discussion with @lrineau, he agreed that the concept `LvaluePropertyMap` should not override  the returned type of the function get(), refined from `ReadablePropertyMap`. I rebased the branch to restore that earlier change.

sloriot:
The branch was responsible for the runtime errors in OTR2. 5638160 fixes them.

lrineau:
This PR also triggered new warnings about unused typedefs:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-101/Polyhedron_Demo/TestReport_lrineau_ArchLinux-clang.gz
```
In file included from /home/cgal_tester/build/src/cmake/platforms/ArchLinux-clang/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:6:
/mnt/testsuite/include/CGAL/IO/write_ply_points.h:64:36: warning: unused typedef 'Point' [-Wunused-local-typedef]
  typedef typename Kernel::Point_3 Point;
                                   ^
/mnt/testsuite/include/CGAL/IO/write_ply_points.h:65:37: warning: unused typedef 'Vector' [-Wunused-local-typedef]
  typedef typename Kernel::Vector_3 Vector;
                                    ^
In file included from /home/cgal_tester/build/src/cmake/platforms/ArchLinux-clang/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:8:
/mnt/testsuite/include/CGAL/IO/write_off_points.h:65:36: warning: unused typedef 'Point' [-Wunused-local-typedef]
  typedef typename Kernel::Point_3 Point;
                                   ^
/mnt/testsuite/include/CGAL/IO/write_off_points.h:66:37: warning: unused typedef 'Vector' [-Wunused-local-typedef]
  typedef typename Kernel::Vector_3 Vector;
                                    ^
/mnt/testsuite/include/CGAL/IO/write_off_points.h:165:36: warning: unused typedef 'Point' [-Wunused-local-typedef]
  typedef typename Kernel::Point_3 Point;
                                   ^
5 warnings generated.
```

PR #1270, opened 2016-07-15
Polyhedron_demo : Fix the image_item's Bbox

### LINKED ISSUE #950: Polyhedron demo: buggy draw of `Scene_segmented_image_item.cpp`
From the file [`2×2×2.inr`](https://cgal.geometryfactory.com/~lrineau/2%c3%972%c3%972.inr), the loaded `Scene_segmented_image_item` is cleary wrong.
222-wrong
The real object encoded by the 3D image is more like that 3D mesh, produced by Mesh_3:
222-object

If both the mesh and the 3D  image are drawn at the same time, with the image in wireframe, that gives:
222-both
222-both-from-z

maxGimeno:
This PR fixes #950 :
 It fixes the Bbox that was too big for the image.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_image_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_image_item.cpp
@@ -640,9 +640,9 @@ Scene_image_item::compute_bbox() const
    _bbox = Bbox(-0.5*m_image->vx(),
                 -0.5*m_image->vy(),
                 -0.5*m_image->vz(),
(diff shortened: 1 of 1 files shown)


lrineau:
The bbox is now coherent with the drawing, but the whole drawing is still completely wrong, compared to 
what is meshed. #950 is not fixed. Anyway, #950 is far from being urgent. Let's merge this PR #1270, that improves the drawing a bit,  and let's fix #950 later.

PR #1273, opened 2016-07-18
OT reconstruction: return a boolean when running

sgiraudot:
This PR adds 2 changes:

* Methods `run()` and `run_until()` now return a boolean instead of `void` that indicates if the required number of steps were performed or if the algorithm was prematurely ended;
* A new method `number_of_isolated_vertices()` counts vertices with mass that are not connected to any edge with mass.


lrineau:
Is that PR needed in CGAL-4.9 or can it be postponed to CGAL-4.10?

sloriot:
Let's target 4.9 for this trivial change

PR #1279, opened 2016-07-19
Polyhedron_demo: Editable normals length

### LINKED ISSUE #1275: variable length of normals
It would be nice if we could in the GUI change the length of the normals of a point set.  @maxGimeno

maxGimeno:
Fixes #1275 
This PR adds a slider to the Scene_points_with_normals_item's context menu that changes the normals length, if the item has normals.


lrineau:
Great patch! I like that slider!

afabri:
So do I !!!

PR #1284, opened 2016-07-19
Polyhedron_demo: Fix the Cut_plugin

### LINKED ISSUE #1278: AABB tree plugin crashes
probably because my model has zero length edges, that is zero area faces.  We should only put non-zero faces in the AABB tree.  @sloriot  Do you agree?

maxGimeno:
fixes #1278 
This PR filters facets of polyhedrons to ignore the ones that are degenerated. It also makes the plugin print a warning and ignore the not pure triangle polyhedrons.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
@@ -1206,30 +1206,25 @@ void Polyhedron_demo_cut_plugin::createCutPlane() {
     if(!poly_item) continue;
     if(!poly_item->polyhedron()->is_pure_triangle())
     {
-      messages->warning(QString("%1 ignored (not pure triangle)").arg(poly_item->name()));
+      messages->warning(QString("%1 ignored (not a triangulated mesh)").arg(poly_item->name()));
       continue;
     }
     if(facet_trees.find(poly_item) == facet_trees.end()) {
       facet_trees[poly_item] = new Facet_tree();
       PPMAP pmap;
       //filter facets to ignore degenerated ones
-      for(Polyhedron::Facet_iterator fit = poly_item->polyhedron()->facets_begin(), end = poly_item->polyhedron()->facets_end(); fit!=end; ++fit)
+      for(Polyhedron::Facet_iterator
+          fit = poly_item->polyhedron()->facets_begin(),
+          end = poly_item->polyhedron()->facets_end();
+          fit!=end; ++fit)
       {
         Polyhedron::Point a(fit->halfedge()->vertex()->point()),
             b(fit->halfedge()->next()->vertex()->point()),
             c(fit->halfedge()->prev()->vertex()->point());
 
-        Polyhedron::Traits::Collinear_3  collinear = poly_item->polyhedron()->traits().collinear_3_object();
-        bool test = collinear(a,b,c);
-        if(!test)
+        if(!CGAL::collinear(a,b,c))
           facet_trees[poly_item]->insert(Facet_primitive(fit, *poly_item->polyhedron(), pmap));
       }
-
-
-    /*  insert(faces(*(poly_item->polyhedron())).first,
-                                     faces(*(poly_item->polyhedron())).second,
-                                     *poly_item->polyhedron(),
-                                     pmap );*/
 
       Scene_aabb_item* aabb_item = new Scene_aabb_item(*facet_trees[poly_item]);
       aabb_item->setName(tr("AABB tree of %1").arg(poly_item->name()));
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
```diff
@@ -1204,14 +1204,33 @@ void Polyhedron_demo_cut_plugin::createCutPlane() {
...
+          facet_trees[poly_item]->insert(Facet_primitive(fit, *poly_item->polyhedron(), pmap));
+      }
+
+
+    /*  insert(faces(*(poly_item->polyhedron())).first,
                                      faces(*(poly_item->polyhedron())).second,
                                      *poly_item->polyhedron(),
-                                     pmap );
+                                     pmap );*/
+
```
sloriot: No need to keep commented code

Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
```diff
@@ -1204,14 +1204,33 @@ void Polyhedron_demo_cut_plugin::createCutPlane() {
...
+      //filter facets to ignore degenerated ones
+      for(Polyhedron::Facet_iterator fit = poly_item->polyhedron()->facets_begin(), end = poly_item->polyhedron()->facets_end(); fit!=end; ++fit)
+      {
+        Polyhedron::Point a(fit->halfedge()->vertex()->point()),
+            b(fit->halfedge()->next()->vertex()->point()),
+            c(fit->halfedge()->prev()->vertex()->point());
+
+        Polyhedron::Traits::Collinear_3  collinear = poly_item->polyhedron()->traits().collinear_3_object();
+        bool test = collinear(a,b,c);
+        if(!test)
```
sloriot: note that there is a global function `CGAL::collinear()`

Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
```diff
@@ -1204,14 +1204,33 @@ void Polyhedron_demo_cut_plugin::createCutPlane() {
...
+    {
+      messages->warning(QString("%1 ignored (not pure triangle)").arg(poly_item->name()));
+      continue;
+    }
     if(facet_trees.find(poly_item) == facet_trees.end()) {
       facet_trees[poly_item] = new Facet_tree();
       PPMAP pmap;
-      facet_trees[poly_item]->insert(faces(*(poly_item->polyhedron())).first,
+      //filter facets to ignore degenerated ones
+      for(Polyhedron::Facet_iterator fit = poly_item->polyhedron()->facets_begin(), end = poly_item->polyhedron()->facets_end(); fit!=end; ++fit)
```
lrineau: This line of code is far too long. According to our coding guidelines, the line width should be limited to 80 columns. There is a tolerance, but you should try to reduce the width of your lines. For this, `for` loop, it can be reduced easily:
```C++
for(Polyhedron::Facet_iterator fit = poly_item->polyhedron()->facets_begin(),
                               end = poly_item->polyhedron()->facets_end();
      fit!=end; ++fit)
```

Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
```diff
@@ -1204,14 +1204,33 @@ void Polyhedron_demo_cut_plugin::createCutPlane() {
     CGAL::Three::Scene_item* item = scene->item(i);
     Scene_polyhedron_item* poly_item = qobject_cast<Scene_polyhedron_item*>(item);
     if(!poly_item) continue;
-
+    if(!poly_item->polyhedron()->is_pure_triangle())
+    {
+      messages->warning(QString("%1 ignored (not pure triangle)").arg(poly_item->name()));
```
lrineau: I am pretty sure most of CGAL users will be surprised by that wording. I think the working "not a triangulated mesh" would be better.

Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
lrineau: Well, Maxime has learned so far to write CGAL generic code, and in a generic code, all access to the geometry should be through geomtraits functors.

Here the code is wrote only for a given type, a `CGAL::Polyhedron_3` with a standard CGAL kernel, and that is the reason why one can use the global function `CGAL::collinear`.

Polyhedron/demo/Polyhedron/Plugins/AABB_tree/Cut_plugin.cpp
afabri: Or even

BOOST_FOREACH(boost::graph_traits<Polyhedron>::face_descriptor fd, faces(*poly_item))

PR #1292, opened 2016-07-21
use the latest CMap API

sloriot:
this was responsible for bugs in the demo plugin where the cmap
is created in a library and used in another one

cc @gdamiand I think it is time to remove the old API code


gdamiand:
I agree.

PR #1295, opened 2016-07-21
Add minimal support for CTest

lrineau:
With that patch, that is so simple to build and run tests (in parallel) for a package.

From the build directory configured in CMake with `WITH_tests=ON` and/or `WITH_examples=ON`, that is as simple as:
```shell
  cd test/Kernel_23
  make -j6
  ctest -j6
```

### DIFF
--- a/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
+++ b/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
@@ -28,6 +28,8 @@ function(create_single_source_cgal_program firstfile )
       add_test(NAME ${exe_name}
         COMMAND ${exe_name} ${ARGS}
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+      set_property(TEST "${exe_name}"
+        APPEND PROPERTY LABELS "${PROJECT_NAME}")
     endif(BUILD_TESTING)
 
     add_to_cached_list( CGAL_EXECUTABLE_TARGETS ${exe_name} )
--- a/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
+++ b/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
@@ -18,6 +18,18 @@ function(create_single_source_cgal_program firstfile )
 
     add_executable(${exe_name} ${all})
 
+    if(BUILD_TESTING)
+      if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${exe_name}.cin")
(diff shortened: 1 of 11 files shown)


lrineau:
I have also tested and used the feature with `test/Mesh_3` and `examples/Mesh_3`.

lrineau:
@sloriot The important commit is https://github.com/CGAL/cgal/pull/1295/commits/c65d3abe17bb3e677b8077996cdaf8672f9c4c6f.

The other commit modifies the names of temporary files used by I/O tests of different kernels, in `test/Kernel_23`, so that `ctest -j` can work (run tests in parallel).

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
```diff
@@ -18,6 +18,18 @@ function(create_single_source_cgal_program firstfile )
 
     add_executable(${exe_name} ${all})
 
+    if(BUILD_TESTING)
+      if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${exe_name}.cin")
+        set(ARGS "${CMAKE_CURRENT_SOURCE_DIR}/${exe_name}.cin")
+      elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${exe_name}.cmd")
+        file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/${exe_name}.cmd" ARGS LIMIT_COUNT 1)
```
sloriot: I don't think this will be sufficient. In some cmd, you have several lines and each line is actually a run

lrineau: I have not tried to be "sufficient" to cover all CGAL testsuite.

With more time, that can be improved.

lrineau: The current behavior of multi-line `.cmd` files is to pass all lines are argument to a single run of the test program.
With CMake/CTest, we could do the same:
  - join all lines of the `.cmd` file in the `ARGS` variable,
or:
  - create as many tests are there are non-empty lines in the file, to allow parallel runs.

lrineau: Anyway, I would like to postpone those improvements for later.
Currently, the small patch I wrote has already saved my some time, for several weeks, while working for Mesh_3, or Kernel_23. I would like to integrate the current patch in CGAL-4.8 and CGAL-4.9 as fast as possible.

sloriot: you at least need a put a TODO in the cmake.

PR #1296, opened 2016-07-22
Polyhedron_demo: Fix for some plugins

maxGimeno:
This PR fixes minor bugs detected in several plugins and segfaults that happened when using some specific operations from the context menu of an item.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Convex_decomposition/Nef_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Convex_decomposition/Nef_plugin.cpp
@@ -56,6 +56,7 @@ void init(QMainWindow* mw,
       action->setProperty("subMenuName",
                           "Boolean Operations");
     _actions<< actionConvexDecomposition;
+    connect(actionConvexDecomposition, SIGNAL(triggered()), this, SLOT(on_actionConvexDecomposition_triggered()));
 
 }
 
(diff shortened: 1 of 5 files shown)


Polyhedron/demo/Polyhedron/Plugins/Convex_decomposition/Nef_plugin.cpp
```diff
@@ -56,6 +56,7 @@ void init(QMainWindow* mw,
       action->setProperty("subMenuName",
                           "Boolean Operations");
     _actions<< actionConvexDecomposition;
+    connect(actionConvexDecomposition, SIGNAL(triggered()), this, SLOT(on_actionConvexDecomposition_triggered()));
```
lrineau: Given the names of the action and of the slot, was not that connection made with the auto-connect feature?

maxGimeno: The nef_plugin does not inherit from the Polyhedron_demo_plugin_helper, so it does not have access to the auto-connect function. I admit I did not look for the reason it is not, as it was far easier and quicker to just add the missing connect than to make it inherit form the helper

lrineau: Anyway, I have never liked those auto-connection features! :confused:

PR #1300, opened 2016-07-22
enable concept checking for Surface_mesh

sloriot:
During integration of the package the test has been disabled.


lrineau:
Even the PR targets 4.9, I will merge it in tonight's `integration` branch, so that the test is run. Probably the test will show bugs that will take times to fix.

PR #1301, opened 2016-07-22
Polyhedron_demo: Fix the image item

### LINKED ISSUE #950: Polyhedron demo: buggy draw of `Scene_segmented_image_item.cpp`

maxGimeno:
This partly fixes #950 
This PR fixes the geometry of the Scene_image_items by shrinking all the border voxels. (All voxels with a coordinate equals to 0 or dimension has its corresponding dimension divided by 2). 
With this fix, a scene_image_item and its mesh are overlapsed. (Before the fix, the Scene_image_item was twice the size of the mesh).

 The color bug is still there though.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_image_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_image_item.cpp
@@ -323,10 +323,24 @@ Vertex_buffer_helper::push_vertex(std::size_t i, std::size_t j, std::size_t k)
 {
   indices_.insert(std::make_pair(compute_position(i,j,k),
                                  vertices_.size()/3)); 
-  
-  vertices_.push_back( (i - 0.5) * data_.vx());
-  vertices_.push_back( (j - 0.5) * data_.vy());
-  vertices_.push_back( (k - 0.5) * data_.vz());
+  //resize the "border vertices"
+  double di(i),dj(j),dk(k);
+  if (di == 0)
+    di = 0.5;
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_image_item.cpp
```diff
@@ -323,18 +323,32 @@ Vertex_buffer_helper::push_vertex(std::size_t i, std::size_t j, std::size_t k)
...
+  vertices_.push_back( (dk - 0.5) * data_.vz());
 }
 
 void
 Vertex_buffer_helper::push_quads(std::size_t i, std::size_t j, std::size_t k)
 {
   int pos1 = vertex_index(i-dx(), j     , k);
   int pos2 = vertex_index(i-dx(), j-dy(), k);
-  int pos3 = vertex_index(i     , j-dy(), k);
+  int pos3 = vertex_index(i     , j-dy(), k) ;
```
lrineau: Remove the added space.

PR #1323, opened 2016-08-02
Polyhedron_demo : Add a new RenderingMode for ShadedPoints

### LINKED ISSUE #1311: Rendering mode for point sets with normals
For point sets with normals the "points" mode depicts shaded points. 
It may be relevant to add another "raw points" to depict the unshaded points.

maxGimeno:
This PR fixes #1311.
It adds a renderingMode called ShadedPoints to distinguish between the points with lighting effects and the others.

### DIFF
--- a/Three/include/CGAL/Three/Scene_interface.h
+++ b/Three/include/CGAL/Three/Scene_interface.h
@@ -45,12 +45,12 @@ class Scene_group_item;
  */
 enum RenderingMode { Points = 0,
                      PointsPlusNormals,
-                     ShadedPoints,
                      Splatting,
                      Wireframe, 
                      Flat,
                      FlatPlusEdges,
                      Gouraud,
+                     ShadedPoints,
                      NumberOfRenderingMode};
 
 
--- a/Three/include/CGAL/Three/Scene_interface.h
+++ b/Three/include/CGAL/Three/Scene_interface.h
@@ -45,13 +45,13 @@ class Scene_group_item;
  */
 enum RenderingMode { Points = 0,
                      PointsPlusNormals,
+                     ShadedPoints,
                      Splatting,
                      Wireframe, 
                      Flat,
                      FlatPlusEdges,
                      Gouraud,
-                     LastRenderingMode = Gouraud,
-                     NumberOfRenderingMode = LastRenderingMode+1 };
+                     NumberOfRenderingMode};
 
 
 namespace CGAL {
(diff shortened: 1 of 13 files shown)


Three/include/CGAL/Three/Scene_interface.h
```diff
@@ -45,13 +45,13 @@ class Scene_group_item;
...
                      PointsPlusNormals,
+                     ShadedPoints,
                      Splatting,
                      Wireframe, 
                      Flat,
                      FlatPlusEdges,
                      Gouraud,
-                     LastRenderingMode = Gouraud,
-                     NumberOfRenderingMode = LastRenderingMode+1 };
+                     NumberOfRenderingMode};
```
lrineau: That will break the binary compatibility: if you add `ShadedPoints` at the end of the enum, before `NumberOfRenderingMode`, then only the binaries using `NumberOfRenderingMode` will have to be rebuild. Otherwise, any item of plugin that uses render modes different from `Points` or `PointPlusNormals`, will have to be recompiled.

Three/include/CGAL/Three/Scene_item.h
```diff
@@ -285,6 +285,10 @@ public Q_SLOTS:
   void setPointsMode() {
     setRenderingMode(Points);
   }
+  //!Sets the RenderingMode to Points.
+  void setShadedPointsMode() {
+    setRenderingMode(ShadedPoints);
+  }
```
lrineau: Should not be switch to an API like `void setRenderingMode(int)`? That would avoid the need to change the `Scene_item.h` for a new rendering mode.

maxGimeno: I don't think it is possible. I tried but it broke the context menu, because this is needed by the slotName() function. I think replacing the current system of rendering mode to use setRenderingMode() instead of the specific functions would need a bigger change than this one.

Three/include/CGAL/Three/Scene_interface.h
maxGimeno: NumberOfRenderingMode is used by Scene_item, doesn't it mean that every item will have to be rebuilt anyway ?

lrineau: I look at the code. `NumberOfRenderingMode` is used by `Scene.cpp` and `Scene_item.cpp`. Those two must be recompiled (they are part of the `demo_framework` library). The headers or code of other scene items does not depend on the value of `NumberOfRenderingMode` as far as I have seen.

sloriot:
The new mode has no effect on a Polyhedron-item. Is this expected? If yes, I suggest to indicate that this rendering mode is not available.

PR #1324, opened 2016-08-02
Polyhedron_demo : Set the points normals after a planar shape detection

### LINKED ISSUE #1310: Normals after (planar) shape detection
For better visual depiction it may be relevant to assign to each detected (planar) point set the normal of the corresponding plane, after shape detection.

maxGimeno:
This PR fixes #1310.

After a shape detection, all points belonging to a detected plane will have their normal set to this plane's normal.


afabri:
@sgiraudot   Can you  check that this should be this way?   I say that as the normal is used by shape detection, so what happens when we want to run it with another epsilon but the  original normals.

maxGimeno:
I did not test but the sub items point sets are individual sets, setting their normal as I did does not change the normals of the original item.

sgiraudot:
Yes, what @maxGimeno explained is pretty much what I was going to say.

The code looks fine from my point of view.

PR #1328, opened 2016-08-03
Mesh_3: add a function that allows to specify and process a given set of polylines to protect

lrineau:
This PR adds an undocumented function, that allows to call `split_graph_into_polylines` on a given set of polylines to protect. This PR is used to reduce the diff with a development branch.


sloriot:
@lrineau  please update the description.

PR #1332, opened 2016-08-08
BGL : detect internal property maps, and use it in PMP isotropic remeshing

### LINKED ISSUE #1256: Isotropric remeshing: Why must face_index be an internal property?
I wonder why you request it to be an internal property and not an optional named parameter?

janetournois:
This PR is about internal property maps, in particular in PMP.

It implements a new tool that is able to detect whether a given BGL graph has an internal property map corresponding to a given property tag or not
 `struct graph_has_property<Graph, PropertyTag>`

We use it in `PMP::isotropic_remeshing()`, and at the same time add `face_index_map` as an optional named parameter

fixes #1256 

@sloriot we should discuss if we want to generate a compilation error or something special when there is no internal pmap, nor a pmap given as a named parameter. It should not compile already.

### DIFF
--- a/BGL/include/CGAL/boost/graph/internal/Has_member_id.h
+++ b/BGL/include/CGAL/boost/graph/internal/Has_member_id.h
@@ -1,0 +1,53 @@
+// Copyright (c) 2016 GeometryFactory (France).  All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org); you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public License as
+// published by the Free Software Foundation; either version 3 of the License,
+// or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+//
+// Author(s) : Jane Tournois
+
+
+#ifndef CGAL_HAS_MEMBER_ID_H
+#define CGAL_HAS_MEMBER_ID_H
+
+namespace CGAL {
+namespace internal {
+
+  template <typename Type>
+  class Has_member_id
+  {
+    typedef char yes[1];
+    typedef char no[2];
+
+    struct BaseWithId
+    {
+      void id(){}
+    };
+    struct Base : public Type, public BaseWithId {};
+
+    template <typename T, T t>
+    class Helper{};
+
+    template <typename U>
+    static no &check(U*, Helper<void (BaseWithId::*)(), &U::id>* = 0);
+
+    static yes &check(...);
+
+  public:
+    static const bool value = (sizeof(yes) == sizeof(check((Base*)(0))));
+  };
+
+}  // internal
+}  // cgal
+
+#endif /* CGAL_HAS_MEMBER_ID_H */
--- a/BGL/include/CGAL/boost/graph/properties.h
+++ b/BGL/include/CGAL/boost/graph/properties.h
@@ -60,8 +60,16 @@ enum face_external_index_t   { face_external_index   } ;
     {}
     std::string s;
     T t;
-        
   };
+
+template<typename Graph, typename PropertyTag>
+struct graph_has_property : CGAL::Tag_false {};
+  
+struct cgal_no_property
+{
+  typedef bool type;
+  typedef const bool const_type;
+};
 
 /// @}
 
--- a/BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
+++ b/BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
@@ -26,6 +26,7 @@
 #include <CGAL/squared_distance_2_1.h>
 #include <CGAL/number_utils.h>
 #include <boost/shared_ptr.hpp>
+#include <CGAL/boost/graph/internal/Has_member_id.h>
 
 #define CGAL_HDS_PARAM_ template < class Traits, class Items, class Alloc> class HDS
 
@@ -435,6 +436,29 @@ struct vertex_property_type<const CGAL::Polyhedron_3<Gt,I,HDS,A> >
 {
   typedef CGAL::vertex_point_t type;
 };
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, vertex_point_t>
+  : CGAL::Tag_true {};
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, edge_weight_t>
+  : CGAL::Tag_true {};
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, edge_index_t>
+  : CGAL::Tag_true {};
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, face_index_t>
+  : CGAL::Boolean_tag<
+      CGAL::internal::Has_member_id<
+        typename CGAL::Polyhedron_3<Gt, I, HDS, A>::Facet
+      >::value
+    >
+{};
+
+
 } // namespace boost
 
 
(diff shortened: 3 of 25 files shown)


janetournois:
will be tested in [CGAL-4.9-Ic-121](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-121.shtml)

janetournois:
will be tested in [CGAL-4.9-Ic-123](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-123.shtml)

janetournois:
merged in integration

janetournois:
the testsuite looks good

BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
```diff
@@ -435,6 +436,29 @@ struct vertex_property_type<const CGAL::Polyhedron_3<Gt,I,HDS,A> >
...
+  : CGAL::Tag_true {};
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, edge_weight_t>
+  : CGAL::Tag_true {};
+
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, edge_index_t>
+  : CGAL::Tag_true {};
+
```
sloriot: Why aren't you using `Has_member_id` also here?

BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
```diff
@@ -435,6 +436,29 @@ struct vertex_property_type<const CGAL::Polyhedron_3<Gt,I,HDS,A> >
...
+template<class Gt, class I, CGAL_HDS_PARAM_, class A>
+struct graph_has_property<CGAL::Polyhedron_3<Gt, I, HDS, A>, face_index_t>
+  : CGAL::Boolean_tag<
+      CGAL::internal::Has_member_id<
+        typename CGAL::Polyhedron_3<Gt, I, HDS, A>::Facet
+      >::value
+    >
+{};
+
```
sloriot: What about `vertex_index_t`?

BGL/include/CGAL/boost/graph/properties_Surface_mesh.h
```diff
@@ -295,6 +294,26 @@ put(CGAL::vertex_point_t p, const CGAL::Surface_mesh<Point>& g,
...
+  template<typename Point>
+  struct graph_has_property<CGAL::Surface_mesh<Point>, halfedge_index_t>
+    : CGAL::Tag_true {};
+  template<typename Point>
+  struct graph_has_property<CGAL::Surface_mesh<Point>, face_index_t>
+    : CGAL::Tag_true {};
+  template<typename Point>
+  struct graph_has_property<CGAL::Surface_mesh<Point>, CGAL::vertex_point_t>
+    : CGAL::Tag_true {};
+
```
sloriot: What about `edge_weight_t`?

BGL/include/CGAL/boost/graph/internal/Has_member_id.h
```diff
@@ -0,0 +1,53 @@
...
+
+    static yes &check(...);
+
+  public:
+    static const bool value = (sizeof(yes) == sizeof(check((Base*)(0))));
+  };
+
+}  // internal
+}  // cgal
+
```
sloriot: Could you explain why you needed more code than what is used in `Has_member_clear`?

BGL/include/CGAL/boost/graph/properties.h
```diff
@@ -60,9 +60,17 @@ enum face_external_index_t   { face_external_index   } ;
...
 
+template<typename Graph, typename PropertyTag>
+struct graph_has_property : CGAL::Tag_false {};
+  
+struct cgal_no_property
+{
+  typedef bool type;
+  typedef const bool const_type;
+};
+
```
sloriot: You're missing a specialization for HDS

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/named_params_helper.h
```diff
@@ -56,6 +58,40 @@ class GetGeomTraits
...
+};
+
+template<typename PolygonMesh, typename PropertyTag>
+typename property_map_selector<PolygonMesh, PropertyTag>::type
+get_property_map(const PropertyTag& p, const PolygonMesh& pmesh)
+{
+  property_map_selector<PolygonMesh, PropertyTag> pms;
+  return pms.get_pmap(p, pmesh);
+}
+
```
sloriot: From what I understand, this removes the need of internal point property maps ? Shall we update the documentation of all these functions?

BGL/include/CGAL/boost/graph/internal/Has_member_id.h
lrineau: I worked with Jane on that topic. The difficulty is that `::id()` can be overloaded in a given class (for example: const and non-const versions).

I think that is worth a comment that points to the link on StackOverflow. @janetournois, do your remember how to retrieve the right SO answer?

BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
janetournois: right. Fixed in commit 111e7a1

BGL/include/CGAL/boost/graph/properties_Polyhedron_3.h
janetournois: see commit 111e7a1

BGL/include/CGAL/boost/graph/properties_Surface_mesh.h
janetournois: missing. Fixed by commit 090dc89fef0085f8a38cc279436e1fe9e2b787f4

BGL/include/CGAL/boost/graph/internal/Has_member_id.h
janetournois: The difficulty was about overloaded functions, but also inherited functions. If you have look at the test I added to the BGL package, you will see what type of problems I was facing before we found [this post on stackoverflow](http://stackoverflow.com/questions/1966362/sfinae-to-check-for-inherited-member-functions).

`Has_member_id<Polyhedron_with_ids::Facet>::value` was returning `Tag_false`, though `Has_member_id<Polyhedron_with_ids::Items::Face_wrapper<Polyhedron_with_ids::HDS, K>::Face>::value)` was returning `Tag_true` -- which is the correct return value

BGL/include/CGAL/boost/graph/properties.h
janetournois: should I add a new file `properties_HalfedgeDS.h`?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/named_params_helper.h
janetournois: It does not remove the need of an internal property map. Here is what it does
- if a pmap is given as a named parameter, it is used
- then, if the polygon mesh has an internal property map, it is used
- if none of them is provided, there is a compilation error which is rather explicit, with `cgal_no_property`

BGL/include/CGAL/boost/graph/properties.h
sloriot: If the file does not exist you can ignore my remark.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/named_params_helper.h
sloriot: Let me clarify what I said. Take [this function](http://doc.cgal.org/4.7/Polygon_mesh_processing/group__PkgPolygonMeshProcessing.html#ga159ff517d211ec30f3fdb799a24834a8). We indicate that the graph must have an internal vertex point map, even if we provide it explicitly. I think your fix now remove that need. Do you agree?

janetournois: Right. It should remove that need indeed (I can't access your link though. Which function do you mention?)

janetournois:
With the latest commits, `PolygonMesh` should not need to have internal property maps anymore (unless it's not provided as a named parameter)

However I think we still have a problem with the `GetK` functor, which determines the Kernel in which things happen based on the `Point` type inside the internal `vertex_point` property map of `PolygonMesh`. I will try to figure out how to get rid of it

janetournois:
commit 2a4030f should solve the problem about `GetK` and `GetGeomTraits`

@sloriot do you have an idea about which data structure I could check that on?

sloriot:
I guess you could try with HDS fed with a `Point_3` type (since we noted that no internal property map was defined).

lrineau:
@janetournois @sloriot 
Is this PR a bug-fix, that should be selected for CGAL-4.9, or should it postponed to CGAL-4.10, once the branch for 4.9 is created?

janetournois:
This PR is an improvement, not really a bug fix. I believe we can postpone it to CGAL-4.10.

lrineau:
The PMP line in the testsuite is fully red. That is probably due to this PR.
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-2.shtml

janetournois:
All these errors have been created by a merge conflict. I just rebased the branch onto `master` to avoid them

PR #1334, opened 2016-08-09
Polyhedron_demo : make the point size of a point set editable.

### LINKED ISSUE #1320: Demo: Allow to change the points size, in points sets items

maxGimeno:
fixes #1320.
This PR adds a slider controlling the points size to the Scene_points_with_normal_item's context menu (near the normals length slider). It allows the user to set this size between 1 and 25. Of course this value is depending of the item, so you can have one point set displayed in size 1 and another one in size 20 at the time.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
@@ -40,6 +40,10 @@ struct Scene_points_with_normal_item_priv
     nb_lines = 0;
     normal_Slider = new QSlider(Qt::Horizontal);
     normal_Slider->setValue(20);
+    point_Slider = new QSlider(Qt::Horizontal);
+    point_Slider->setValue(5);
+    point_Slider->setMinimum(1);
+    point_Slider->setMaximum(25);
   }
   Scene_points_with_normal_item_priv(const Scene_points_with_normal_item& toCopy, Scene_points_with_normal_item* parent)
     : m_points(new Point_set(*toCopy.d->m_points)),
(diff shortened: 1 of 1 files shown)


janetournois:
tested locally

janetournois:
tested in [CGAL-4.9-Ic-123](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-123.shtml)

Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
```diff
@@ -68,12 +76,17 @@ struct Scene_points_with_normal_item_priv
     }
     normal_Slider = new QSlider(Qt::Horizontal);
     normal_Slider->setValue(20);
+    point_Slider = new QSlider(Qt::Horizontal);
+    point_Slider->setValue(5);
+    point_Slider->setMinimum(1);
+    point_Slider->setMaximum(25);
```
lrineau: Once C++11 will be accepted for the code of CGAL demos, the use of [delegating constructors] will avoid the code duplication when there are several constructors.

http://stackoverflow.com/q/26199431

sgiraudot:

Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
```diff
@@ -585,7 +597,7 @@ void Scene_points_with_normal_item::drawPoints(CGAL::Three::Viewer_interface* vi
 
     GLfloat point_size;
     viewer->glGetFloatv(GL_POINT_SIZE, &point_size);
-    viewer->glPointSize(4.f);
+    viewer->glPointSize(d->point_Slider->value());
```
lrineau: A note for later: I am pretty sure that the point size should be scaled according to the size of the bbox of the point set: let's say you have a point set contained in the bbox [0, 1]³, then the default point size (`5.`) will be huge for that point set.

maxGimeno: The size in glPointSize is in pixels, not in World coordinates, so it is not related to the Bbox.

PR #1336, opened 2016-08-10
Polyhedron_demo: Affine transformation for point_sets.

### LINKED ISSUE #1315: Affine transform applicable to point sets
It would be good to be able to apply the affine transform to point sets. 
Note also that I was not able to see how to specify the said transform.

maxGimeno:
fixes #1315.
This PR makes the transform_polyhedron_plugin able to transform point_sets too.
It also renames this plugin in Affine_transform_plugin, as it no longer is specific for polyhedron_items.


sgiraudot:
I tested the plugin on point sets and noticed 2 weird behaviors:
- The green copy first appears in a different location than the point set (in one case, I could not even see it). Maybe it is centered on the origin?
- When moving the camera using a large point set with the quick display active (it is by default), only a small percentage of the point set is displayed. The same happens with the green copy, but I think the points are not randomized as the small percentage is not random at all (so parts of the green point set completely disappear when moving).

maxGimeno:
It should be fixed.

sgiraudot:
Okay, I tested it locally, it looks good.

Merged in [4.9-lc-123](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-123.shtml).

sgiraudot:

PR #1340, opened 2016-08-10
AABB_tree : fix doc vs code inconsistency

### LINKED ISSUE #260: Wrong functor required/used
The AABBGeomTraits require the functor [Compute_closest_point_3](http://doc.cgal.org/latest/AABB_tree/classAABBGeomTraits.html#a97e587bad84dcd1e90d00fb18a134762), but this signature is never used in `AABB_tree` code.

The class `AABB_traits`  provides a functor `Closest_point` which calls the free function `internal::nearest_point_3` which seems to be used instead. Note that even if this functor is renamed it is still in the wrong class (should be in the geomtraits).

janetournois:
partially fixes #260 

The doc mentions `Compute_closest_point_3`, and code implements `Closest_point`

<< edit
First, `Compute_closest_point_3` should be `Construct_closest_point_3`.
<< end edit

since ~~`Compute_closest_point_3`~~ `Construct_closest_point_3` should move to the Kernel at some point, I prefer to keep this option
However, for backward compatibility with possibly existing user code, I keep Closest_point
in the non-documented code.

### DIFF
--- a/AABB_tree/doc/AABB_tree/Concepts/AABBTraits.h
+++ b/AABB_tree/doc/AABB_tree/Concepts/AABBTraits.h
@@ -145,10 +145,10 @@ A functor object to compute distance comparisons between the query and the nodes
 typedef unspecified_type Compare_distance; 
 
 /*!
-A functor object to compute closest point from the query on a primitive. Provides the operator: 
+A functor object to construct closest point from the query on a primitive. Provides the operator: 
 `Point_3 operator()(const Query& query, const Primitive& primitive, const Point_3 & closest);` which returns the closest point to `query`, among `closest` and all points of the primitive. 
 */ 
-typedef unspecified_type Closest_point; 
+typedef unspecified_type Construct_closest_point; 
 
 /*!
 A functor object to compute the squared distance between two points. Provides the operator: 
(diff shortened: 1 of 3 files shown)


janetournois:
will be tested in [CGAL-4.9-Ic-125](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-125.shtml)

janetournois:
the testsuite looks good

AABB_tree/doc/AABB_tree/Concepts/AABBTraits.h
```diff
@@ -145,10 +145,10 @@ A functor object to compute distance comparisons between the query and the nodes
 typedef unspecified_type Compare_distance; 
 
 /*!
-A functor object to compute closest point from the query on a primitive. Provides the operator: 
+A functor object to construct closest point from the query on a primitive. Provides the operator: 
 `Point_3 operator()(const Query& query, const Primitive& primitive, const Point_3 & closest);` which returns the closest point to `query`, among `closest` and all points of the primitive. 
 */ 
-typedef unspecified_type Closest_point; 
+typedef unspecified_type Construct_closest_point;
```
sloriot: I guess your missing _3 here

sloriot:
I guess you're missing some changes in the code using the traits class, otherwise the changes done in the concept are not taken into account. (Or I'm missing something)

janetournois:
this should be taken into account in #1341 and #1350

sloriot:
Changes were imported in #1350

PR #1338, opened 2016-08-10
Polyhedron_demo: Transparent background for snapshots

maxGimeno:
Fixes #1313.
This PR gives the possibility to choose if the background of the snapshot  will be of the same color as the scene, transparent, or of the chosen color. In the latter case, when the user confirms his/her choices and closes the dialog, a color dialog opens and let him/her choose the wanted color.

PR #1343, opened 2016-08-10
Polyhedron_demo: Snapshot to clipboard

### LINKED ISSUE #1314: Snapshot to clipboard
Very nice-to-have feature: copy current client to clipboard, with transparent background. So that we do not need to go through files and quickly prepare image layouts in ppt or keynote. 

Note: the current shortcut ctrl+c overrides this function from qglviewer. this means one should verify whether qglviewer does this already with transparent background or not.

maxGimeno:
Fixes #1314.
This PR implements a way to take a snapshot and send it to the clipboard. 
It uses arbitrary values for every quality parameters that are supposed to be given through the Dialog when a snapshot is taken in the regular way. 

**Question :** Would it  be better to keep the dialog for those parameters ? IMHO this shortcut is supposed to be quick, so keeping a dialog makes no sense and if the user needs some specifics parameters, then he/she can use the regular snapshot feature.

The function is bound to "Ctrl+Print Screen".
Once coupled with #1338, it will be able to use a transparent background.

@lrineau This PR makes the Viewer_impl a friend class of the viewer, as I needed access to some protected functions and I found it curious that a D pointer was not a "friend" of its original class. Is this OK for you ?

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.cpp
+++ b/Polyhedron/demo/Polyhedron/Viewer.cpp
@@ -30,6 +30,8 @@ public:
   mutable std::vector<QOpenGLShaderProgram*> shader_programs;
   QMatrix4x4 projectionMatrix;
   void setFrustum(double l, double r, double t, double b, double n, double f);
+  QImage* takeSnapshot(Viewer* viewer, int quality, int background_color, QSize size, double oversampling, bool expand);
+  void sendSnapshotToClibboard(Viewer*);
 };
 Viewer::Viewer(QWidget* parent, bool antialiasing)
   : CGAL::Three::Viewer_interface(parent)
@@ -402,7 +404,19 @@ void Viewer::keyPressEvent(QKeyEvent* e)
   else if(e->key() == Qt::Key_I && e->modifiers() & Qt::ControlModifier){
     d->scene->printPrimitiveIds(this);
     update();
+    return;
       }
+
+  else if(e->key() == Qt::Key_C && e->modifiers() & Qt::ControlModifier){
+    d->sendSnapshotToClibboard(this);
+    return;
+  }
+
+  else if(e->key() == Qt::Key_S && e->modifiers() & Qt::ControlModifier){
+    saveSnapshot(true);
+    return;
+  }
+
   //forward the event to the scene (item handling of the event)
   if (! d->scene->keyPressEvent(e) )
     QGLViewer::keyPressEvent(e);
@@ -1522,7 +1536,6 @@ private Q_SLOTS:
 
 void Viewer::saveSnapshot(bool, bool)
 {
-  GLfloat alpha = 1.0;
   qreal aspectRatio = width() / static_cast<qreal>(height());
   static ImageInterface* imageInterface = NULL;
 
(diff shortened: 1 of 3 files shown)


janetournois:
You are right, the parameters dialog should not appear while using this functionality.
Are you using the same default parameters as the ones of the dialog?

maxGimeno:
Not all.  The size is doubled and the frustum option is set to true.

janetournois:
ok. So parameters are better than "arbitrary"

lrineau:
https://github.com/CGAL/cgal/pull/1343#issue-170427777:
> @lrineau This PR makes the Viewer_impl a friend class of the viewer, as I needed access to some protected functions and I found it curious that a D pointer was not a "friend" of its original class. Is this OK for you ?

I think that is OK: `Viewer_impl` hides the implementation details of `Viewer` and the fact that `Viewer_impl` is friend of `Viewer` is just an implementation detail as well. You should have a look why `Viewer` even has anything "protected" that is not hidden in `Viewer_impl`: that might reveal a design issue.

janetournois:
it does not work on my machine. Can someone else test it?

sgiraudot:
It doesn't seem to work for me either. `Ctrl+Print Screen` just does a regular screenshot of the whole desktop for me.

maxGimeno:
Apparently some OS override the ctrl+Print_scr to do a classic Print_scr. 
@lrineau Can I move 'recenter' to Ctrl+R to use Ctrl+C, that seems the most logical for this feature ?

sgiraudot:
I agree with Maxime, `Ctrl+C` is the QGLViewer way of copying snapshots to the clipboard, we should do it that way to.

Actually, most of the time, when I want to make a snapshot in the demo, I accidentally recenter the scene…

lrineau:
I agree.

Is there a common shortcut in QGLViewer to center the viewport on the scene?

maxGimeno:
A double click with the middle button of the mouse does the same thing that Ctrl+C, as documented in the default QGLViewer help tips.

lrineau:
That is too complicated. If I remember well, the keyboard shortcut was asked by Andreas, long ago.



Le 11 août 2016 14:54, à 14:54, Maxime GIMENO <notifications@github.com> a écrit:

maxGimeno:
I don't see how a double click is more complicated than a keyboard shortcut but ok.
Is Ctrl+R ok ?

sgiraudot:
Okay, now the feature works well with `Ctrl-C`.

lrineau:
A double click with the middle button... What if the mouse does not have a middle button. Think about Mac mouses.  The keyboard shortcut is needed too.

maxGimeno:
I set the Recenter Scene to Shift+Ctrl+C.

lrineau:
<kbd>Ctrl+R</kbd> was used by something else?

maxGimeno:
No, but this way it is still C so it is closer to what it was.

janetournois:
It works for me, and it's a nice feature.
I agree with Laurent that <kbd>Ctrl+R</kbd> would be better to **R**ecenter scene

maxGimeno:
Ctrl+R it is, then.

sgiraudot:
Merged in [4.9-Ic-128](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-128.shtml).

sgiraudot:
Testsuite looks good. I'm merging it.

Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -402,7 +404,19 @@ void Viewer::keyPressEvent(QKeyEvent* e)
   else if(e->key() == Qt::Key_I && e->modifiers() & Qt::ControlModifier){
     d->scene->printPrimitiveIds(this);
     update();
-      }
+    return;
+  }
+
+  else if(e->key() == Qt::Key_C && e->modifiers() & Qt::ControlModifier){
+    d->sendSnapshotToClibboard(this);
```
sloriot: Clibboard?

Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -30,6 +30,8 @@ class Viewer_impl {
   mutable std::vector<QOpenGLShaderProgram*> shader_programs;
   QMatrix4x4 projectionMatrix;
   void setFrustum(double l, double r, double t, double b, double n, double f);
+  QImage* takeSnapshot(Viewer* viewer, int quality, int background_color, QSize size, double oversampling, bool expand);
+  void sendSnapshotToClibboard(Viewer*);
```
lrineau: @maxGimeno: I agree with Sébastien. It is spelled "clipboard" with a 'p'. A clipboard is [a board with a clip](https://www.google.fr/search?q=clipboard): ![clipboard image].

[clipboard image]: https://upload.wikimedia.org/wikipedia/commons/thumb/c/c0/Wood-clipboard.jpg/480px-Wood-clipboard.jpg

maxGimeno:
I know it was just a typo...

PR #1344, opened 2016-08-12
Polyhedron_demo: Import the c3t3_rib_export_plugin 

maxGimeno:
This PR imports the c3t3_rib_export_plugin from the Mesh_3 demo.
It seems to work, but I cannot find a way to visualize a rib file, so I am not sure.


janetournois:
- do you know what is the input for the `materialTree` in the ui?
- can you please add the property so that the action appears in the menu "Operations > Tetrahedral Mesh Generation"? (because it "belongs" to the `Mesh_3` package)

maxGimeno:
materialTree is not linked to anothing, and it was the same in 4.7. I think we can remove it.

PR #1351, opened 2016-08-18
OT reconstruction: fix: loose bbox length computation

sgiraudot:
The half of the bounding box was computed as `(xmax - xmin / 2)`, that is to say with a missing parenthesis…

This fixes that. Notice that I also changed the computation so that the maximum value between `xl` and `yl` is used: this avoid having a badly shaped domain when the point cloud is flat on an axis direction.


sgiraudot:
Merged in [4.9-Ic-130](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-130.shtml).

sgiraudot:

PR #1357, opened 2016-08-23
CTest support (cont.)

lrineau:
This PR is a follow-up of #1295.

- The code now handle the `.cin` files correctly.
- The `.cmd` files are also correctly handled:
  - Multiline `.cmd` files are correctly handled (only one test is created, for the moment).
  - Globbing patterns in `.cmd` are also handled.

I can run almost all of CGAL tests using CTest, but ones from TAU (Arrangement and Sweep).

I wonder if that PR could be integrated in CGAL-4.9.


CMakeLists.txt
```diff
@@ -16,6 +16,11 @@ include(${CMAKE_SOURCE_DIR}/Installation/cmake/modules/CGAL_SCM.cmake)
 option( CGAL_REPORT_DUPLICATE_FILES "Switch on to start (naive) detection of duplicate source- and headerfiles in packages" OFF)
 
 message( "== CMake setup (DONE) ==\n" )
+
+# Enable testing with BUILD_TESTING
+option(BUILD_TESTING "Build the testing tree." OFF)
+include(CTest)
+
```
sloriot: Is `CTest` always here if `CMake` is?

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
```diff
@@ -1,7 +1,28 @@
...
+      file(GLOB files RELATIVE ${CGAL_CURRENT_SOURCE_DIR} ${item_n})
+      list(APPEND output_list ${files})
+    else()
+      list(APPEND output_list ${item_n})
+    endif()
+#    message(STATUS "  new value of the output list: ${output_list}")
+  endforeach()
+  set(${list_name} ${output_list} PARENT_SCOPE)
+endfunction()
+
```
sloriot: Some comments at the top of this function would be nice to help understanding the code.

CMakeLists.txt
lrineau: According to the [CMake Version Compatibility Matrix], the module is shipped since CMake-2.6, and has a compatible interface since then.

[CMake Version Compatibility Matrix]: https://cmake.org/Wiki/CMake_Version_Compatibility_Matrix/StandardCMakeModulesA

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
```diff
@@ -12,26 +33,45 @@ function(create_single_source_cgal_program firstfile )
...
+	  foreach(CMD_LINE ${CMD_LINES})
+#	    message(STATUS "  command line: ${CMD_LINE}")
+	    separate_arguments(CMD_LINE_ARGS UNIX_COMMAND ${CMD_LINE})
+#	    message(STATUS "  args: ${CMD_LINE_ARGS}")
+	    list(APPEND ARGS ${CMD_LINE_ARGS})
+	  endforeach()
+	  expand_list_with_globbing(ARGS)
+	endif()
+#	message(STATUS "add test: ${exe_name} ${ARGS}")
+	add_test(NAME ${exe_name} COMMAND ${exe_name} ${ARGS})
```
sloriot: I'm surprised `add_test` is not in a loop. What's the trick to handle multiline cmd files?

Scripts/scripts/cgal_create_cmake_script
```diff
@@ -58,6 +58,10 @@ if ( CGAL_FOUND )
   include( CGAL_CreateSingleSourceCGALProgram )
 
 EOF
+    if [ -d "${SOURCE_DIR}" ] ; then
+      echo "  set(CGAL_CURRENT_SOURCE_DIR \"${SOURCE_DIR}\")"
+      echo
+    fi
```
sloriot: What about test with already commit `CMakeLists.txt` ?

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
lrineau: Done in https://github.com/CGAL/cgal/pull/1357/commits/bd156440a090700171826c9b0787a16f91506a79.

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
lrineau: For the moment, `.cmd` files are handled the same way they are by our current `cgal_test*` shell scripts: multi-line files are treated as if the lines were concatenated.

The line
```CMake 
file(STRINGS "${cmd_file}" CMD_LINES)
```
reads the files in a list `CMD_LINES`., then there is a loop `foreach(CMD_LINE ${CMD_LINES})` on the content of it. The call
```CMake
separate_arguments(CMD_LINE_ARGS UNIX_COMMAND ${CMD_LINE})
```
splits one line `CMD_LINE` of the list into shell arguments, to a list `CMD_LINE_ARGS`. And that list is appended to the list `ARGS`.

Eventually, the call `expand_list_with_globbing(ARGS)` replaces all items of the list with a `*` to a sublist corresponding to the result of the globbing.

lrineau: The current documentation of `.cmd` files says (see `Guidelines/Testing#.2A.cmd_files` in our wiki):
> For each line in program.cmd, the testsuite will then execute the program with the full line as arguments.

but that is not coherent with the current behavior.

If needed, the documented behavior is rather easy to implement.

Scripts/scripts/cgal_create_cmake_script
lrineau: If the sources already have a `CMakeLists.txt`, then the `CMAKE_CURRENT_SOURCE_DIR` is the right one, and the line
```CMake
set(CGAL_CURRENT_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
```
[in `UseCGAL.cmake`][line] will set the variable.

[line]: https://github.com/CGAL/cgal/pull/1357/files/ce782f02f8983669678f78ace85dd6a13760cece#diff-bbff21e986dce690cf536e0de94bc0f8R12

lrineau:
@sloriot Now that I have answered to your question, what do you think of this PR? Can it be selected for CGAL-4.9?

It allows to run the full testsuite locally (but Arr_2 and Sweep_2) in parallel. And with the `-R` or `-L` option, one can choose to select to run only a few test, or all tests from a selection of packages. In my opinion that is a major tool to help verifying branches before a push.

sloriot:
Sure

PR #1359, opened 2016-08-23
Deprecate the macro CGAL_VERSION

lrineau:
The macro `CGAL_VERSION` is deprecated. `CGAL_VERSION_STR` must be used instead.


lrineau:
I have scheduled it for 4.10, but it could be applied to 4.9. `CGAL_VERSION` is useless anyway.

PR #1361, opened 2016-08-24
Fix Compare_y_at_x_2 for polycurve traits

sloriot:
The code of `Compare_y_at_x_2` was assuming that xmonotone curves was oriented from <i>left-to-right</i> (meaning endpoints are lexicographically sorted), with is not the case for example for polylines.

@efifogel Do you agree with the fix?


efifogel:
Yes, I do agree with the fix.
Thanks.

                         _/



On Wed, Aug 24, 2016 at 12:39 PM, Sebastien Loriot <notifications@github.com

lrineau:
This PR has been merged in the bug-fix branch for CGAL-4.8.2, and I will merged it in `master` as well (targeting CGAL-4.9).

PR #1373, opened 2016-08-29
PMP isotropic remeshing does not support exact constructions

janetournois:
We were testing Epic twice instead of Epic and Epec.

**Edit** : it does not seem very useful to make isotropic remeshing work with an exact constructions kernel.
So, we have decided to remove support of those kernels for this function, at least for the moment.
Later, we could modify `isotropic_remeshing` so that it guarantees that output vertices lie exactly on the input surface. A `@todo` has been added to the doxygen comments to detail how to do that

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -56,14 +56,17 @@
 #ifdef CGAL_PMP_REMESHING_DEBUG
 #include <CGAL/Polygon_mesh_processing/self_intersections.h>
 #define CGAL_DUMP_REMESHING_STEPS
+#define CGAL_PMP_REMESHING_VERBOSE_PROGRESS
 #endif
 
 #ifdef CGAL_PMP_REMESHING_VERY_VERBOSE
 #define CGAL_PMP_REMESHING_VERBOSE
+#define CGAL_PMP_REMESHING_VERBOSE_PROGRESS
 #endif
 
-
-
+#ifdef CGAL_PMP_REMESHING_VERBOSE_PROGRESS
+#define CGAL_PMP_REMESHING_VERBOSE
+#endif
 
 
 namespace CGAL {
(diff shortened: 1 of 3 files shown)


janetournois:
I'm testing this PR locally and can't promise yet it will pass the testsuite successfully...

lrineau:
Jane is right: adding the tests with Epeck is actually not trivial, because the tests may time out, or expose new bugs.

lrineau:
At least it seems to compile, according to Travis CI. TODO: check if the tests pass.

janetournois:
it also compiles locally on my machine, but it has been running for about 40 hours now (no crash though) so we need to do something about it

lrineau:
I have merged this PR in CGAL-4.9-Ic-140. And the PMP lines is fully red. On Linux machines, it seems the tests finished in the time bound, but the iso-remeshing test failed.
See
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-140.shtml#Polygon_mesh_processing

janetournois:
what platform do you mention? I think I went through all the linux platforms, and they all say "killed" during the projection step

lrineau:
You are right. They all were terminated after the time bound set. I mixed up the per-test time bound (20 minutes), and the per-package time bound (100 minutes): for some platform, the global time bound was reached, but you are right that for all platform the test named `remeshing_test` was killed.

lrineau:
`remeshing_test` was killed by the timeout
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-143.shtml#Polygon_mesh_processing

lrineau:
Jane, any chance this bug can be fixed before the end of the week?

janetournois:
I hope it can. I am starting to work on it

janetournois:
`PMP::isotropic_remeshing` is not able to deal with exact constructions kernels as is.
One option would be to guarantee that output vertices exactly lie on the input surface, but this will be the topic of another PR.

I rename this PR

janetournois:
This PR will be tested in [CGAL-4.9-Ic-152](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-152.shtml)

lrineau:
@janetournois I suggest the PR should have a different title, now. And you may want to change/amend the first message, that is usually a summary of the PR.

janetournois:
I amended the 1st message. It should be ready for merging now

PR #1375, opened 2016-08-29
make relative precision thread local

afabri:
FIx a part of #1387


mglisse:
Shouldn't this be documented? If I call set_relative_precision_of_to_double, I might expect that it affects all threads (and that I am responsible for any synchronization with other get_/set_ calls in other threads).

Issue #1360, opened 2016-08-23
BGL tests : compilation error with recent OpenMesh


janetournois:
I am now using OpenMesh-6.2 and I get the following error when I compile the test `graph_concept_OpenMesh.cpp` inside the BGL package.

The error is the following : 
> CGAL\cgal-dev\STL_Extension\include\CGAL/hash_openmesh.h(31): error C2084: the function
> 'size_t OpenMesh::hash_value(const OpenMesh::BaseHandle &)' already has a body in
> C:\3rdPartyLibs\OpenMesh-6.2\include\OpenMesh/Core/Mesh/Handles.hh(110)

We should take into account the version number to define this function or not

lrineau:
@sloriot @afabri 
This issue is selected for the milestone `4.8.2`. Do we have the resources to work on it rather soon? Once the branch for CGAL-4.9 is created, it will be difficult to maintain the branch for CGAL-4.8.2.

lrineau:
Actually, I found out that `<CGAL/hash_openmesh.h>` is a file that was added to CGAL-4.9, and was not in CGAL-4.8.x. I have re-assigned the bug to 4.9 instead of 4.8.2.

lrineau:
Fixed by #1377.

PR #1377, opened 2016-08-29
hash_value is not defined inside an #ifdef OM_HAS_HASH

afabri:
So we have to add it for OpenMesh version < 6.2  
@janetournois can you test this please.

### DIFF
--- a/STL_Extension/include/CGAL/hash_openmesh.h
+++ b/STL_Extension/include/CGAL/hash_openmesh.h
@@ -22,7 +22,7 @@
 
 #include <OpenMesh/Core/Mesh/Handles.hh>
 
-#if OMVERSION < 0x60200
+#if OM_VERSION < 0x60200
 
 namespace OpenMesh {
 
--- a/STL_Extension/include/CGAL/hash_openmesh.h
+++ b/STL_Extension/include/CGAL/hash_openmesh.h
@@ -22,15 +22,20 @@
 
 #include <OpenMesh/Core/Mesh/Handles.hh>
 
-#ifndef OM_HAS_HASH
-
-#include <functional>
+#if OMVERSION < 0x60200
 
 namespace OpenMesh {
 
 inline std::size_t hash_value(const BaseHandle& h) { return h.idx(); }
 
 } // namespace OpenMesh
+#endif
+
+
+#ifndef OM_HAS_HASH
+
+#include <functional>
+
 
 namespace std {
 


STL_Extension/include/CGAL/hash_openmesh.h
```diff
@@ -22,15 +22,20 @@
 
 #include <OpenMesh/Core/Mesh/Handles.hh>
 
-#ifndef OM_HAS_HASH
-
-#include <functional>
+#if OMVERSION < 0x60200
```
sloriot: On my OpenMesh installation, the variable is named `OM_VERSION`.

janetournois:
it works for me, and fixes #1360

lrineau:
It could help if the `examples/BGL_OpenMesh/CMakeLists.txt` displays the version of OpenMesh that is being used. I would like to know which OpenMesh versions are actually tested in our testsuite.

It seems the following test platforms have OpenMesh (at least in [CGAL-4.9-Ic-137]):
  - `jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits`
  - `jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits`
  - `lrineau_Fedora-rawhide`
  - `lrineau_Fedora-strict-ansi`
  - `lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8`
  - `lrineau_x86-64_Linux-Fedora_IntelCompiler-17`
  - `lrineau_x86-64_Linux-Fedora_IntelCompiler-17-with-assertions`

[CGAL-4.9-Ic-137]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-137.shtml

I had a look and our current module `Installation/cmake/modules/FindOpenMesh.cmake` does not detect the OpenMesh version. It should read the file `${OPENMESH_INCLUDE_DIR}/OpenMesh/Core/System/config.h` and parse the line:
```C++
#define OM_VERSION 0x30200
```

afabri:
Should it be in cmake, or should it be so that it can be written in the  table of the testsuite?

lrineau:
Andreas in https://github.com/CGAL/cgal/pull/1377#issuecomment-243709612 said:
> Should it be in cmake, or should it be so that it can be written in the table of the testsuite?

One or the other, but I think the CMake way is simpler: modify `FindOpenMesh.cmake` to parse the `OM_VERSION` from `<OpenMesh/Core/System/config.h>`, then modify `examples/BGL_OpenMesh/CMakeLists.txt` so that the version string is displayed.

For the other way, first OpenMesh should be searched in the configuration of CGAL libraries, even if that is not at all a dependency of CGAL libraries, and then we would have to create a .cpp file to display the version string, and modify also two or three scripts from the testsuite infrastructure, so that in the end the version string is displayed in the table of the testsuite results.

PR #1378, opened 2016-08-30
Handle edges with undefined cost

sloriot:
@afabri do you agree with the fix?


lrineau:
Sébastien, if the PR is for CGAL-4.8.2, can you please rebase the branch and change the base of the PR?

sloriot:
Done, I did it on purpose to test the new edit feature :)

lrineau:
"undefined cost not handled yet" was never displayed in the test suite. That indicates that that code path was never tested. I suggest a new test case be added, to test cases when the cost is undefined.
(TODO)

sloriot:
Done (`simplify_polygon_test data/rectangle.poly`)

PR #1417, opened 2016-08-31
Fix bug in polyline simplification

afabri:
We had hardwired that we use Exact_predicates_tag which is slow for EPEC
in particular with Quotient<MP_float> or leda::real

We determine the appropriate tag using Algebraic_structure_traits<FT>::Is_exact


lrineau:
Was this bug already in CGAL-4.8 or was it only introduced in CGAL-4.9? I think it was already there in CGAL-4.8. Should not we target CGAL-4.8.2 instead of 4.9?

lrineau:
@afabri TODO:
- yesterday I have asked if the bug was introduced in 4.9 or before,
- should not we integrated a test case for this bug?

lrineau:
Without any answer after two days, I searched and found out the issue was already in CGAL-4.7.

I have rebased the branch onto `releases/CGAL-4.8-branch`.

Issue #1388, opened 2016-08-30
Static variables in Box_intersection_d package


sloriot:
- [x] `Box_intersection_d/include/CGAL/Box_intersection_d/segment_tree.h:278: static int level = -1;`
- [x] `Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h:42:    static std::size_t n;`

afabri:
@sloriot can you please check

lrineau:
Fixed by #1420.

PR #1420, opened 2016-08-31
Box_intersection_d: Make variables thread local

afabri:
Solves #1388

### DIFF
--- a/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
+++ b/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
@@ -39,7 +39,7 @@ template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
     Unique_numbers() {
-      CGAL::cpp11::atomic<std::size_t>(n); // initialized to 0
+      static CGAL::cpp11::atomic<std::size_t>(n); // initialized to 0
       i = n++;
     }
     std::size_t id() const { return i; }
--- a/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
+++ b/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
@@ -26,6 +26,8 @@
 #include <CGAL/Bbox_2.h>
 #include <CGAL/Bbox_3.h>
 #include <CGAL/Box_intersection_d/box_limits.h>
+#include <CGAL/tss.h>
+
 #include <algorithm>
 
 namespace CGAL {
@@ -36,14 +38,16 @@ namespace Box_intersection_d {
 template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
-    Unique_numbers() : i(n++) {}
+    Unique_numbers() {
+      CGAL_STATIC_THREAD_LOCAL_VARIABLE(std::size_t, n, 0);
+      i = n++;
+    }
     std::size_t id() const { return i; }
 private:
-    static std::size_t n;
+
     std::size_t i;
 };
-template <class Dummy>
-std::size_t Unique_numbers<Dummy>::n = 0;
+
 
 
 // Policies for id-number of boxes
--- a/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
+++ b/Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
@@ -26,7 +26,7 @@
 #include <CGAL/Bbox_2.h>
 #include <CGAL/Bbox_3.h>
 #include <CGAL/Box_intersection_d/box_limits.h>
-#include <CGAL/tss.h>
+#include <CGAL/atomic.h>
 
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
```diff
@@ -36,14 +38,16 @@ namespace Box_intersection_d {
 template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
-    Unique_numbers() : i(n++) {}
+    Unique_numbers() {
+      CGAL_STATIC_THREAD_LOCAL_VARIABLE(std::size_t, n, 0);
+      i = n++;
+    }
```
sloriot: Similarly here, I think it should be a `tbb::atomic<std::size_t>`

cc @afabri @lrineau

Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
```diff
@@ -36,14 +38,16 @@ namespace Box_intersection_d {
 template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
-    Unique_numbers() : i(n++) {}
+    Unique_numbers() {
+      CGAL::cpp11::atomic<std::size_t>(n); // initialized to 0
+      i = n++;
```
sloriot: I guess you're missing a static I guess

afabri: fixed

Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
```diff
@@ -36,14 +38,16 @@ namespace Box_intersection_d {
 template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
-    Unique_numbers() : i(n++) {}
+    Unique_numbers() {
+      static CGAL::cpp11::atomic<std::size_t>(n); // initialized to 0
```
lrineau: Have you tried to compile that code?! There should be at least a warning.
Use:
```C++
static CGAL::cpp11::atomic<std::size_t> n; // initialized to 0
```

Box_intersection_d/include/CGAL/Box_intersection_d/segment_tree.h
```diff
@@ -275,9 +275,8 @@ split_points( RandomAccessIter begin, RandomAccessIter end,
 
 
 #if CGAL_BOX_INTERSECTION_DEBUG
- static int level = -1;
  #define CGAL_BOX_INTERSECTION_DUMP(msg) { \
-   for( unsigned int i = level; i; --i ) \
+   for( unsigned int i = -1; i; --i ) \
```
lrineau: This patch completely modifies the loop. As such the loop will loop from `-1`, that is `UINTMAX` as a `unsigned`, decreasing to to `0`, whereas before the patch it was looping from `level` to `0.

afabri: I had not really understood the code. The variable `Counter<int> bla(level);`  increases level each time the function is called, so the for loop starts with `level == 0`.

Box_intersection_d/include/CGAL/Box_intersection_d/Box_d.h
```diff
@@ -36,14 +38,16 @@ namespace Box_intersection_d {
 template <class Dummy>
 struct Unique_numbers {
     typedef std::size_t ID;
-    Unique_numbers() : i(n++) {}
+    Unique_numbers() {
+      static CGAL::cpp11::atomic<std::size_t> n; // initialized to 0
+      i = n++;
+    }
```
lrineau: As `n` is now an `atomic<integral>`, [its post-increment `operator++ is equivalent to `n.fetch_add(1)`][*]. That synchronizes with other threads.

[*]: http://en.cppreference.com/w/cpp/atomic/atomic/operator_arith

lrineau: It means that those "unique IDs" will really be unique, even if there are multiple threads.

Box_intersection_d/include/CGAL/Box_intersection_d/segment_tree.h
```diff
@@ -339,6 +338,7 @@ void segment_tree( RandomAccessIter1 p_begin, RandomAccessIter1 p_end,
     const T sup = box_limits< T >::sup();
 
 #if CGAL_BOX_INTERSECTION_DEBUG
+  CGAL_STATIC_THREAD_LOCAL_VARIABLE(int, level, -1);
```
lrineau: `level` is used for debugging only. it being a thread-local variable is the right choice: anyway the `Box_intersection_d` is not thread-safe if the same data is concurrently used by multiple threads.

lrineau:
There are errors with older platforms, like:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-3/Box_intersection_d/TestReport_lrineau_CentOS6.gz

`CGAL::cpp11::atomic<T>` will not compile
  - if `CGAL_NO_THREADS` is defined,
  - or C++11 cannot be used and Boost<1.53.

If `CGAL_NO_THREADS`, `CGAL::cpp11::atomic<T>` could just be equivalent to `T` and that would work.

But what can we do in both the compiler and Boost are too old?

Issue #1414, opened 2016-08-30
Static variables in Nef_3 package


sloriot:
- [x] `Nef_3/include/CGAL/Nef_3/Visualizor.h:496:  static std::vector<Polyhedron>  polyhedra_;`
- [x] `Nef_3/include/CGAL/Nef_3/Visualizor.h:497:  static std::vector<std::string> titles_;`
- [x] `Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h:42:    static int unique = 0;`

afabri:
I will not fix it for the Visualizer class.

afabri:
@sloriot Why did you reopen it?

sloriot:
Because #1421 is not yet merged

lrineau:
#1421 is now merged.

PR #1421, opened 2016-08-31
Nef_3: Make variables atomic

afabri:
Solves  #1414

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
@@ -20,6 +20,9 @@
 
 #ifndef CGAL_NEF_SNC_INDEXED_ITEMS_H
 #define CGAL_NEF_SNC_INDEXED_ITEMS_H
+
+#include <CGAL/tss.h>
+
 #include <CGAL/Nef_3/Vertex.h>
 #include <CGAL/Nef_3/Halfedge.h>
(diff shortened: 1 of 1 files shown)


Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
```diff
@@ -39,7 +42,7 @@ class Index_generator {
  public:
   static int get_unique_index()
   {
-    static int unique = 0;
+    CGAL_STATIC_THREAD_LOCAL_VARIABLE(int, unique, 0);
```
mglisse: So the index will only be unique inside a thread, not globally. Can that be a problem? Should the doc warn about it? (maybe it already does, I didn't check)

sloriot: I think you're right, we should use a `tbb::atomic<int>`here.

cc @afabri

Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
```diff
@@ -20,6 +20,10 @@
 
 #ifndef CGAL_NEF_SNC_INDEXED_ITEMS_H
 #define CGAL_NEF_SNC_INDEXED_ITEMS_H
+
+#include <CGAL/tss.h>
```
lrineau: No need to `<CGAL/tss.h>`.

And the title of the PR is wrong.

Issue #1416, opened 2016-08-30
Static variables in Cone_spanners_2 package


sloriot:
- [x] `Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h:185:        static iterator res;`
- [x] `Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h:209:        static reverse_iterator res;`

afabri:
@sloriot   this seems ready to me; please check

lrineau:
Fixed by #1422.

@sloriot, can you confirm?

PR #1422, opened 2016-08-31
Cone_spanners: No need for thread local storage as readonly

afabri:
Solves #1416

### DIFF
--- a/Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h
+++ b/Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h
@@ -182,14 +182,14 @@ public:
 
     /* End Iterator */
     inline iterator end() {
-        static iterator res;
+        static const iterator res;
 
         return res;
     }
 
     /* Constant End Iterator */
     inline const_iterator end() const {
-        static const_iterator res;
+        static const const_iterator res;
 
         return res;
     }
--- a/Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h
+++ b/Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h
@@ -182,16 +182,12 @@ public:
 
     /* End Iterator */
     inline iterator end() {
-        static const iterator res;
-
-        return res;
+        return iterator();
     }
 
     /* Constant End Iterator */
     inline const_iterator end() const {
-        static const const_iterator res;
-
-        return res;
+        return const_iterator();
     }
 
     /* Reverse order Begin Iterator */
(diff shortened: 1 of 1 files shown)


Cone_spanners_2/include/CGAL/Cone_spanners_2/Plane_scan_tree.h
```diff
@@ -182,14 +182,14 @@ class Plane_scan_tree {
 
     /* End Iterator */
     inline iterator end() {
-        static iterator res;
+        static const iterator res;
```
lrineau: Note that the *initialization* of `static const ` local variables is not guaranteed to be thread-safe: two different threads could use this function for the first time at the same time. See this [SO answer]. Its only in C++11 and later that the C++ norm adds that guarantee.

[SO answer]: http://stackoverflow.com/a/3237024/1728537

It seems that, for Plain-Old-Data types, the initialization is actually done at compile time by compiler, so is should be safe.

**But:** the type `iterator` (a typedef for `_Iterator`) is actually not a POD for C++<11: it has a user-defined destructor.

In that case, the default-constructor of `_iterator` is so trivial: 
```C++
    _Iterator(leaf_type* start = NULL) : cell (0), leaf(start) {}`
```
that I think it would actually be faster not to use a static variable:
```C++
inline iterator end() const {
  return iterator();
}
inline const_iterator end() const {
  return const_iterator();
}
```

afabri: For when do you plan that C++11 is the minimal version?    I do not ask for this particular case, but there are other issues/PR where this means that we have to  switch to thread local storage.

lrineau: According to that page: http://en.cppreference.com/w/cpp/compiler_support
then the support for the C++11 local thread storage has been added in g++-4.8, clang-3.3, MSVC-14 (aka 2015). When will be ready to drop support for users stuck with previous versions?

For g++ [under Redhat Enterprise Linux, g++-4.8 means RHEL 7][*], the last version. We have users who are stuck with RHEL 6 at the moment. Recently I have even seen a user requesting support for g++ 4.2 !

[*]: https://access.redhat.com/solutions/19458

[The end-of-life of RHEL 6 should be around 2021][wp-RHEL]. Until then, I do not think we can drop the support for C++<11 and we will probably directly jump to support C++>=14. For now, that means that we will have to find workarounds: use tools from Boost when the C++11 version is not usable. That is what we have always done so far. The thread safety is just a rather new topic for CGAL, and around workarounds are not completely ready.

[wp-RHEL]: https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Version_history

Issue #1390, opened 2016-08-30
Static variables in Snap_rounding_2 package


sloriot:
- [x] `Snap_rounding_2/include/CGAL/Snap_rounding_2.h:90:  static SEG_Direction seg_dir;`

lrineau:
Fixed by #1424.

PR #1424, opened 2016-08-31
Snap_rounding: Make variables thread local

afabri:
Solves #1390


sloriot:
@afabri Travis says this PR breaks the checks

afabri:
@sloriot please check

Issue #1385, opened 2016-08-30
Static variables in Segment_Delaunay_graph_Linf_2 package


sloriot:
- [x] `Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_2/Segment_Delaunay_graph_Linf_hierarchy_2_impl.h:362:    static Segments_in_hierarchy_tag stag;`
- [x] `Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_2/Segment_Delaunay_graph_Linf_hierarchy_2_impl.h:392:  static Intersections_tag          itag;`
- [x] `Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_2/Segment_Delaunay_graph_Linf_hierarchy_2_impl.h:393:  static Segments_in_hierarchy_tag  stag;`

lrineau:
Fixed by #1425.

PR #1425, opened 2016-08-31
SDG_linf_2: No need for thread local storage as readonly

afabri:
Solves #1385


afabri:
@sloriot  please check

lrineau:
The patch seems good.

PR #1350, opened 2016-08-17
Add Construct_closest_point_3 to Kernel

afabri:
Add a functor Construct_closest_point_3  which before  was called Closest_point in the concept AABBTraits

This fixes #260 and #1341

PR #1428, opened 2016-09-01
Fix implementation of AABB_traits::Closest_point

### LINKED ISSUE #260: Wrong functor required/used

### LINKED ISSUE #1341: Move Functors to the Kernel
The `AABB_traits.h` file contains implementations of the functors
- ~~Compute_closest_point_3~~ `Construct_closest_point_3`
- `Compare_distance`

They should both be moved to the Kernel

This issue is related to #260

sloriot:
The problem reported by #260 was the requirement of having `Compute_closest_point_3` in `AABBGeomTraits` while it was not used. In this PR, we turn this requirement into having `Construct_projected_point_3` with operators for triangles and segments, and `Compare_squared_distance_3`. These two functors are now used it in the implementation of `Closest_point` in `AABB_traits`. Note that the current implementation might be slower as we no longer take advantage of the `bound` parameter to quit earlier the computation of the closest point.

The Kernel functor `Construct_projected_point_3` has been updated so that it provides two additional operators (one for segments and one for triangles).

Replaces #1350 
Fixes #260
Fixes #1341


sloriot:
@afabri, @janetournois  do you agree this PR fixes the aforementioned issues?

lrineau:
I have not had a look for a while, but if I remember well the changes are good.

afabri:
I do not agree with adding two files to `Kernel_23/include/CGAL/Kernel`.  The functions in the files should be dispatched  to global_functions_3.h and global_functions_internal_3.h, and there might be code that goes into functors.

sloriot:
The code is used in Kernel functors so I don't see it going  in `global_functions_internal_3.h` that contains only wrapping code. The code is used by both Cartesian and Homogeneous kernels so it cannot be put in these packages like it is done with other projection code. I can move the file into `internal` instead of Kernel or even merge the file if you want.

afabri:
Shouldn't we at least move it into global_functions_3.h ?

sloriot:
Moved into `Kernel/function_objects.h`

PR #1434, opened 2016-09-06
Point set shape detection: improve example

sgiraudot:
Documentation in the RANSAC package is missing an example to show how to recover specific parameters of shapes (using `dynamic_cast`).

### DIFF
--- a/Point_set_shape_detection_3/examples/Point_set_shape_detection_3/efficient_RANSAC_parameters.cpp
+++ b/Point_set_shape_detection_3/examples/Point_set_shape_detection_3/efficient_RANSAC_parameters.cpp
@@ -99,15 +99,17 @@ int main()
   while (it != shapes.end()) {
     
     // Get specific parameters depending on detected shape.
-    if (dynamic_cast<Plane*>(it->get()))
+    if (Plane* plane = dynamic_cast<Plane*>(it->get()))
       {
-        Kernel::Vector_3 normal = dynamic_cast<Plane*>(it->get())->plane_normal();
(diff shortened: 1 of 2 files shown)


Point_set_shape_detection_3/examples/Point_set_shape_detection_3/efficient_RANSAC_parameters.cpp
```diff
@@ -97,9 +97,30 @@ int main()
   Efficient_ransac::Shape_range::iterator it = shapes.begin();
 
   while (it != shapes.end()) {
-    // Prints the parameters of the detected shape.
-    std::cout << (*it)->info() << std::endl;
-
+    
+    // Get specific parameters depending on detected shape.
+    if (dynamic_cast<Plane*>(it->get()))
+      {
```
sloriot: I think I would use `if (Plane* plane = dynamic_cast<Plane*>(it->get())){ Vector_3 normal = plane->normal`

sloriot: Note that there is also a conversion to Traits::Plane_3 operator

sgiraudot: Good points, I'm taking care of this.

lrineau:
I think this PR could qualify for CGAl-4.9, instead of postponing it to CGAL-4.10. What do you think, Simon?

sgiraudot:
It's not critical (no `include` file is changed) and it can really help users using the package, so the sooner the better!

PR #1435, opened 2016-09-06
Triangulation_2: Use boost::transform_iterator instead of CGAL::Iterator_project

afabri:


Triangulation_2/include/CGAL/Triangulation_2.h
```diff
@@ -29,7 +29,7 @@
 #include <iostream>
 
 #include <CGAL/iterator.h>
-#include <CGAL/Iterator_project.h>
+//#include <CGAL/Iterator_project.h>
 #include <CGAL/function_objects.h>
```
sloriot: do not comment but remove

Triangulation_2/include/CGAL/Triangulation_2.h
```diff
@@ -194,13 +195,15 @@ class Triangulation_2
...
+  /*
   typedef Iterator_project<Finite_vertices_iterator, 
                            Proj_point,
 	                   const Point&, 
                            const Point*,
                            std::ptrdiff_t,
                            std::bidirectional_iterator_tag>  Point_iterator;
-
+  */
+  typedef boost::transform_iterator<Proj_point,Finite_vertices_iterator> Point_iterator;
```
sloriot: do not comment but remove

lrineau:
I suggest you change the title of the PR. It does not seem to be only about removing an unused `#include`. You did change from an iterator adapter to another.

PR #1436, opened 2016-09-07
Renewal of CMake scripts: header-only, ctest, CMake>=3.1

lrineau:
- [x] <s>The CTest support no longer tries to test demos (should be backported to 4.9).</s> (already done)
- [x] <s>The bash script `cgal_create_cmake_script` enables C++14</s> (done in PR #1533)
- [x] <s>New feature of `create_single_source_cgal_program` that allows to use C++11/14 features.</s> (done in PR #1534)
- [x] **The main goal:** Use CMake-3.1 new features to renew the configuration of CGAL.
- [x] **The second goal:** header-only step 2 (_pure header only_, without CMake):
  - [x] First implementation done
  - [x] How to generate `<CGAL/compiler_config.h>`?
  - [x] Deal with CGAL_Core with GMP disabled
  - [x] handle compilation flags modified in `Installation/CMakeLists.txt` (`-frounding-math` and others) ... Fixed in https://github.com/CGAL/cgal/pull/1436/commits/03f84dc94b5e6c707fd56325b4354173f7f116ae
  - [x] Changes about CORE and Boost thread: see #1888 ...Fixed in https://github.com/CGAL/cgal/pull/1436/commits/283f8138d80f605bf1aff663b8a305a0cd4a0ef4

TODO for later:
  - [x] Handle `make install` for pure-header-only (done in https://github.com/CGAL/cgal/pull/1436/commits/7cff890473d5a499b81fe4a9eb347e6a3d55ef10)
  - [x] Merge `CGAL_HEADER_ONLY` and `CGAL_HEADER_ONLY_STEP_2`
  - [ ] a kind of preconfiguration for pure-header-only?
  - [x] Commit `<CGAL/version.h>` in Git?
  - [x] Document the modification to `cgal_create_cmake_script`
  - [x] Document modifications to `create_single_source_cgal_program` (done, see https://github.com/CGAL/cgal/pull/1534#issuecomment-254225657)

For C++14 support, the required CMake versions are:
- CMake>=3.1 for gcc and clang on Linux,
- CMake>=3.2 for our usual platforms,
- CMake>=3.6 for Intel Compiler.


lrineau:
Problem: our Travis CI configuration uses CMake-2.8.11. We now requires CMake>=2.8.12.

lrineau:
I have verified that it works, on my machine, with CMake-2.8.12.

lrineau:
The first Travis issue is now fixed. But now clang uses g++ STL and fails.

Triangulation_3/examples/Triangulation_3/CMakeLists.txt
```diff
@@ -1,29 +1,11 @@
...
-  include( ${CGAL_USE_FILE} )
-  
-  find_package( TBB QUIET )
-
-  if( TBB_FOUND )
-    include(${TBB_USE_FILE})
-    list(APPEND CGAL_3RD_PARTY_LIBRARIES ${TBB_LIBRARIES})
-  endif()
-
-  include( CGAL_CreateSingleSourceCGALProgram )
```
cjamin: Are you sure we don't need that anymore? I'm getting an error: "Unknown CMake command "create_single_source_cgal_program""

lrineau: Actually you are quite right. If we want the CMakeLists.txt to be compatible with:
  -  CGAL in pure header only,
  - or header only step 1,
  - or CGAL precompiled,
then I have to modify the other `CGALConfig.cmake` (those that are generated), so that `CGAL_CreateSingleSourceCGALProgram` is included automatically.

cjamin: Actually, it is already the case (in both CGALConfig_XXX.cmake.in).

lrineau:
https://www.slideshare.net/LaurentRineau/clipboards/my-clips

Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake
```diff
@@ -120,7 +120,7 @@ function(CGAL_setup_CGAL_dependencies target)
     if(WIN32)
       target_compile_options(${target} ${keyword} "/fp:strict")
     else()
-      target_compile_options(${target} ${keyword} "-fp-model strict")
+      target_compile_options(${target} ${keyword} "-fp-model" "strict")
```
lrineau: @afabri: That was the issue causing the errors with the Intel Compiler. Because I had added the option `"-fp-model strict"` in one argument, instead of `-fp-model strict` in two arguments, then the FPU model was not set, and the CGAL filters were wrong.

I have remerged #2592 about the faster usage of MPFR.

lrineau:
@maxGimeno After the removal of the direct use of OpenGL, `demo//Principal_component_analysis` and `demo/Optimal_transportation_reconstruction_2` no longer compile in this branch: they do not use the `QOpenGLFunctions` API. Could you please fix them?

That is rather urgent because this PR blocks a lot of thing in `integration`.

lrineau:
@afabri Could you try to build this branch, commit https://github.com/CGAL/cgal/pull/1436/commits/26fe3df693dd8b586a757a2d23e563a9e0dc22fa, and compile CGAL_Core as a DLL on Windows?

afabri:
It works for VC++.

lrineau:
@maxGimeno I have tested your two commits 66df33abeab181ff91c685d7282e8114e24eaa90 and 30171e03ee3fba13f34d28b20cda7ee09d9875c0, and now the two demos build correctly. Thanks!

Now, our only usages of OpenGL in CGAL are in:
  - the polyhedron demo (for `GlSplat`), but that will disappear soon or later,
  - in the old demo in `demo/Surface_mesher`,
  - and as a dependency of libQGLViewer (for versions < 2.5.4).

lrineau:
@janetournois @afabri @cjamin:
The only remaining issue with this branch seems to be on Windows with the CMake variable `CGAL_HEADER_ONLY=ON`: in that case there are a lot of errors with CGAL_Core in various packages. See the column `x64_Cygwin-Windows10_MSVC2017-Release-64bits` in [the test results of 4.12-Ic-109].

[the test results of 4.12-Ic-109]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-109.shtml

 I would like you to try to reproduce, and help me debug that issue.

lrineau:
As a matter of fact, we also have issue with the Intel Compiler... but we decided to drop its support.

cjamin:
Unfortunately, I don't have MSVC 2017 for now, but here are a few questions/suggestions:
- Do you have a way to check if CGAL_CAN_USE_CXX11_THREAD_LOCAL is defined for this test?
- If yes, could you try to force the use of Boost.Thread like it's done for GCC?

cjamin:
Is the MSVC 2017 problem fixed?

lrineau:
No. I wanted that PR be merged, now that almost all issues are solved (at least in non-header-only mode). I have created https://github.com/CGAL/cgal/issues/2656 to gather the remaining issues.

cjamin:
Ok, I understand. I will rewrite my comment there.

PR #1437, opened 2016-09-07
Add a link to the dD alpha complex offered by GUDHI

afabri:


Alpha_shapes_2/doc/Alpha_shapes_2/Alpha_shapes_2.txt
```diff
@@ -43,6 +43,10 @@ it's way too large. So we will never spoon up ice-cream lying in the
 inside of the convex hull of \f$ S\f$, and hence the \f$ \alpha\f$-shape for
 \f$ \alpha \rightarrow \infty\f$ is the convex hull of \f$ S\f$.
 
+\cgal offers 2D and 3D alpha shapes. The GUDHI library offers a
+<a href="http://gudhi.gforge.inria.fr/doc/latest/group__alpha__complex.html"> dD Alpha complex</a>.
```
lrineau: We might wonder why the `Gudhi::alphacomplex::Alpha_complex< Kernel >` is not part of CGAL itself. It seems to use a lot from CGAL.

VincentRouvreau: As far as I know, CGAL has no persistence computation module, neither data structure adapted to such comptutations. Am I wrong ?
GUDHI's team is grateful of all the work done on CGAL, and also to contribute to CGAL development.

lrineau: Hi Vincent, I just discussed the subject in-real-life with @cjamin. I did not know that the code of the `Alpha_complex` from GUDHI was beyond the simple computation of the dD alpha shape. Clément told me that there was code inside it to compute persistence on the alpha shape. And that of course is the subject of GUDHI, not pure geometry computation. I am imagine that extracting a pure-geometrical dD alpha shape code from GUDHI is quite difficult.

mglisse:
One main reason to have alpha_complex in Gudhi is that we need a simplicial complex data structure that actually represents all the simplices of all dimensions so we can store the alpha value (we need it internally, not just to export the result). There is one in Gudhi, but not in CGAL AFAIK (linear cell complex might be usable for this, but not a great fit), and maps like Alpha_shape_3 uses are unlikely to scale so well.
Otherwise, no, there is no specific code to compute persistence on an alpha complex, we just produce a simplicial complex and use the separate persistence module on that, alpha_complex is really about the geometry. Essentially, we build a Delaunay triangulation in CGAL, then convert it to our simplicial complex data structure, then there is the main function that computes the alpha values (by decreasing dimension). Extracting the geometry part is easy, but then you would need to design an interface, and it doesn't seem worth the trouble for now. However, if someone from GF is volunteering to do the work, we are available to discuss the details ;-)

PR #1438, opened 2016-09-07
Clean up projects name in cmake scripts

sloriot:


lrineau:
@sloriot 
Could the test directories project names have a "_Tests" suffix? That would be better if you want to run CTest test for only the test/ directories.

For example, for the moment `ctest -L Mesh_3` launches the test from `test/Mesh_3` and `examples/Mesh_3`, and there is no way to launch only the tests from `test/Mesh_3`. If the `test/Mesh_3` project name was `Mesh_3_Tests`, then `ctest -L Mesh_3` would still launch both examples and tests, but in addition `ctest -L Mesh_3_Tests` would allow to launch only the tests from `test/Mesh_3`.

PR #1443, opened 2016-09-09
Adds CMake package version file generation

TManhente:
This allows clients to specify a minimum CGAL version in their CMake list files.
Eg.:

    find_package(CGAL 4.9)

More info: https://cmake.org/cmake/help/v3.6/manual/cmake-packages.7.html#package-version-file

### DIFF
--- a/Installation/cmake/modules/CGAL_Macros.cmake
+++ b/Installation/cmake/modules/CGAL_Macros.cmake
@@ -411,11 +411,20 @@ if( NOT CGAL_MACROS_FILE_INCLUDED )
 
   macro( create_CGALconfig_files )
 
+    include(CMakePackageConfigHelpers)
+
     # CGALConfig.cmake is platform specific so it is generated and stored in the binary folder.
     configure_file("${CGAL_MODULES_DIR}/CGALConfig_binary.cmake.in"  "${CMAKE_BINARY_DIR}/CGALConfig.cmake"        @ONLY)
+    write_basic_package_version_file("${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake"
+      VERSION "${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}.${CGAL_BUILD_VERSION}"
+      COMPATIBILITY AnyNewerVersion)
 
     # There is also a version of CGALConfig.cmake that is prepared in case CGAL in installed in CMAKE_INSTALL_PREFIX.
     configure_file("${CGAL_MODULES_DIR}/CGALConfig_install.cmake.in" "${CMAKE_BINARY_DIR}/config/CGALConfig.cmake" @ONLY)
+    configure_file(
+      "${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake"
+      "${CMAKE_BINARY_DIR}/config/CGALConfigVersion.cmake"
+      COPYONLY)
 
     #write prefix exceptions
     file( APPEND ${CMAKE_BINARY_DIR}/CGALConfig.cmake "${SPECIAL_PREFIXES}\n")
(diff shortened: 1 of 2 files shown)


Installation/cmake/modules/CGAL_Macros.cmake
```diff
@@ -411,11 +411,20 @@ if( NOT CGAL_MACROS_FILE_INCLUDED )
...
+    write_basic_package_version_file("${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake"
+      VERSION "${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}.${CGAL_BUILD_VERSION}"
+      COMPATIBILITY AnyNewerVersion)
 
     # There is also a version of CGALConfig.cmake that is prepared in case CGAL in installed in CMAKE_INSTALL_PREFIX.
     configure_file("${CGAL_MODULES_DIR}/CGALConfig_install.cmake.in" "${CMAKE_BINARY_DIR}/config/CGALConfig.cmake" @ONLY)
+    configure_file(
+      "${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake"
+      "${CMAKE_BINARY_DIR}/config/CGALConfigVersion.cmake"
+      COPYONLY)
```
lrineau: What does `configure_file` do on the file `CGALConfigVersion.cmake`?

Whould not the following be sufficient, in `Installation/CMakeLists.txt`:
```cmake
install(FILES
   ${CMAKE_BINARY_DIR}/config/CGALConfig.cmake
   ${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake
   DESTINATION ${CGAL_INSTALL_CMAKE_DIR} )
```

lrineau: The difference is that `CGALConfigVersion.cmake` is copied from `${CMAKE_BINARY_DIR}` instead of ` ${CMAKE_BINARY_DIR}/config`

TManhente: The `configure_file()` only copies the package version file generated in `${CMAKE_BINARY_DIR}` into the `${CMAKE_BINARY_DIR}/config` folder.

This is necessary if the `CGALConfig.cmake` in the `config` folder could be used directly from the build tree to satisfy a `find_package()` call.

If that's not the case (i.e. this file should only be used after install and from the installation directory), then this `configure_file()` call can be removed and the `INSTALL(FILES)` can be changed to use the version file from the `${CMAKE_BINARY_DIR}` just like you've said.

lrineau: As far as I know, the `.cmake` files in that `config/` sub-directory are only there for the purpose of the `INSTALL` command. If one tries to use them directly before they are install, they would refer to non-existing headers and libraries paths.

That is why I think the `configure_file` step is not wanted, for `CGALConfigVersion.cmake`.

Installation/cmake/modules/CGAL_Macros.cmake
```diff
@@ -411,11 +411,20 @@ if( NOT CGAL_MACROS_FILE_INCLUDED )
 
   macro( create_CGALconfig_files )
 
+    include(CMakePackageConfigHelpers)
+
     # CGALConfig.cmake is platform specific so it is generated and stored in the binary folder.
     configure_file("${CGAL_MODULES_DIR}/CGALConfig_binary.cmake.in"  "${CMAKE_BINARY_DIR}/CGALConfig.cmake"        @ONLY)
+    write_basic_package_version_file("${CMAKE_BINARY_DIR}/CGALConfigVersion.cmake"
+      VERSION "${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}.${CGAL_BUILD_VERSION}"
+      COMPATIBILITY AnyNewerVersion)
```
lrineau: That `COMPATIBILITY` setting is not the right one. The CGAL project does not commit to support compatibility if the major version changes. You should use `SameMajorVersion`.

lrineau:
The PR is now merged. Thanks.

PR #1449, opened 2016-09-13
switch to boost::bind

afabri:


lrineau:
The Travis CI test showed [compilation errors]:
```
[ 74%] Building CXX object examples/STL_Extension/CMakeFiles/min_element_if_example.dir/min_element_if_example.cpp.o
/home/travis/build/CGAL/cgal/STL_Extension/examples/STL_Extension/min_element_if_example.cpp:25:17: error: 
      call to 'bind2nd' is ambiguous
                          compose1_1(bind2nd(greater< int >(), 0),
                                     ^~~~~~~
/usr/lib/gcc/x86_64-linux-gnu/4.6/../../../../include/c++/4.6/backward/binders.h:162:5: note: 
      candidate function [with _Operation = std::greater<int>, _Tp = int]
    bind2nd(const _Operation& __fn, const _Tp& __x)
    ^
/usr/include/boost/functional.hpp:289:33: note: candidate function [with
      Operation = std::greater<int>]
    inline binder2nd<Operation> bind2nd(const Operation &op,
                                ^
1 error generated.
make[3]: *** [examples/STL_Extension/CMakeFiles/min_element_if_example.dir/min_element_if_example.cpp.o] Error 1
```

[compilation errors]: https://travis-ci.org/CGAL/cgal/jobs/159487515#L3558


What is more, the PR does not say what you are trying to fix with it. Why switch to Boost functional library? If that is because part of the `<functional>` STL header has been deprecated since C++11, then I think the better switch would be to opt for the C++11 equivalent: lambda expressions.

afabri:
I continue a replacement that we had made earlier for the include files, but not for the examples.
As we do not want require C++11, let's stick to boost.

lrineau:
Actually during the last developers meeting we agreed that C++11 could be used for some examples or demos. Those three examples are good candidates for C++11: why use an old API using boos::bind functions instead of using the nicer C++11 notations?

PR #1451, opened 2016-09-13
-ansi implies c++98 -> update script requirements of boost libraries

### LINKED ISSUE #1450: CGAL header-only errors
96 errors in the following testsuite, all of the form "undefined reference to <something in Boost>":

```
    354 undefined reference to `boost::system::system_category()'
    392 undefined reference to `boost::system::generic_category()'
   1443 undefined reference to `boost::detail::get_tss_data(void const*)'
   2886 undefined reference to `boost::detail::set_tss_data(void const*, boost::shared_ptr<boost::detail::tss_cleanup_function>, void*, bool)'
```

See for example https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-I-150/AABB_tree_Demo/TestReport_lrineau_Fedora-strict-ansi.gz

It is probably related to the feature of CGAL that allows not to use Boost.Thread with C++11.

This bug blocks CGAL-4.9.

sloriot:
Fixes #1450

### DIFF
--- a/Installation/cmake/modules/CGAL_SetupBoost.cmake
+++ b/Installation/cmake/modules/CGAL_SetupBoost.cmake
@@ -13,7 +13,8 @@ if ( NOT CGAL_Boost_Setup )
           AND CMAKE_CXX_FLAGS MATCHES "\\-std=(c|gnu)\\+\\+[01][14yxz]")
         #GCC 6.0+ without c++03 on
         OR ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6.0
-             AND NOT CMAKE_CXX_FLAGS MATCHES "\\-std=(c|gnu)\\+\\+[90][83]")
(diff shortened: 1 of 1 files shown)


lrineau:
Thanks! I had not realized that, even with g++>=6, `-ansi` meant C++98.

lrineau:
To detect C++11 or above, I wonder why we do not use CMake [`try_run`] command and read the value of `__cpluplus`.

[`try_run`]: https://cmake.org/cmake/help/v3.0/command/try_run.html

lrineau:
Actually, CMake already does the job to check the default C++ version. The variable `CMAKE_<LANG>_STANDARD_DEFAULT`, since CMake-3.4. See [cmake-developer(7)].

[cmake-developer(7)]: https://cmake.org/cmake/help/v3.2/manual/cmake-developer.7.html

PR #1452, opened 2016-09-13
Number_types Eigen::NumTraits::Literal

### LINKED ISSUE #1429: Epick_d no longer compile on recent Debian testing
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.9-Ic-141/NewKernel_d/TestReport_lrineau_Debian-Testing.gz
> /usr/include/eigen3/Eigen/src/Core/util/XprHelper.h:74:8: error: no type named 'Literal' in 'struct Eigen::NumTraits<CGAL::Interval_nt<false> >'

mglisse:
Fixes #1429 .

### DIFF
--- a/Number_types/include/CGAL/CORE_BigFloat.h
+++ b/Number_types/include/CGAL/CORE_BigFloat.h
@@ -531,6 +531,7 @@ namespace Eigen {
     typedef CORE::BigFloat Real;
     typedef CORE::BigFloat NonInteger;
     typedef CORE::BigFloat Nested;
+    typedef CORE::BigFloat Literal;
 
     static inline Real epsilon() { return 0; }
     static inline Real dummy_precision() { return 0; }
--- a/Number_types/include/CGAL/CORE_BigFloat.h
+++ b/Number_types/include/CGAL/CORE_BigFloat.h
@@ -526,7 +526,9 @@ template <> class Real_embeddable_traits< CORE::BigFloat >
 
 namespace Eigen {
   template<class> struct NumTraits;
+  template<class> struct GenericNumTraits;
   template<> struct NumTraits<CORE::BigFloat>
+   : GenericNumTraits<CORE::BigFloat>
   {
     typedef CORE::BigFloat Real;
     typedef CORE::BigFloat NonInteger;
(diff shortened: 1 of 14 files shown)


lrineau:
... On one hand, I find the patch rather long, to push it to CGAL-4.9 after just a few runs in the testsuite.

On the other hand, that is the sort of change that can be easily tested, and does not seem particularly dangerous, as soon as we test it with different versions of Eigen (which the testsuite does).

@sloriot , do you agree that this PR could be merged before 4.9? Assuming 4.9 is published before the end of next week?

sloriot:
The patch does not seem dangerous. 4.9 sounds good to me.

lrineau:
Well. The [testsuite of CGAL-4.9-Ic-52 is fully red][*]. It had not happened for a long time. I should have checked.

[*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9-Ic-152.shtml

lrineau:
Let me add that, even if that is rare, it does happen, and I should have checked. Sorry.

That patch looked so simple! That is the [Puss in Boots][*] effect. :smile: 

Puss in Boots image

[*]: https://www.google.com/search?safe=on&q=shrek+cute+cat

mglisse:
Epick_d.cpp compiles fine with both g++ and clang++ (removing the check that disables the test for clang...) here.

Is the Travis CI thing restricted to patches against master, as opposed to patches against branches?

To be honest, I was surprised that compilers accepted the specialization deriving from an undefined class, but since it was compiling with 2 compilers, I went with it. Which is silly, because Epick_d.cpp is probably one of the only cases where Eigen headers are included first and so the class is not undefined...

As a quick fix, the easiest would be to drop this patch and instead add `typedef ... Literal;` inside each specialization. Deriving from GenericNumTraits would be more future-proof, but we re/forward-declare NumTraits because I guess we don't want to include eigen unconditionally, and I am not sure how to manage both goals.

I'll do the typedef thing, should I open a separate branch/PR, or is it better to recycle somehow?

lrineau:
- Indeed, we can see that the line `NewKernel_d` in the testsuite has one green cell, on Fedora-Rawhide, running gcc-6.2.1. Your compiler is probably as recent.
- Travis CI only triggers jobs when there is a `.travis.yml` at the root of a checkout. For the moment, such a file exists only in `master` and branches/PR derives from `master`.
- I think you can recycle the branch, and thus this PR. The commits logs will even be a sort of documentation of what works or not.

PR #1454, opened 2016-09-14
CGAL_Core: fix -fvisibility issues on Linux

lrineau:
This PR allows to compile whole CGAL with `-fvisibility=hidden` with gcc and clang.


lrineau:
@afabri @janetournois I wonder if that PR would compile on Windows, with shared libraries. Can you please test the branch before I push it to `integration`? I think the build of the CGAL_Core DLL is sufficient to see if that triggers nasty warnings from MSVC about usages of `dcelspec(import/export)`.

afabri:
With the changes I made the tests Number_types/CORE_*.cpp  pass

PR #1461, opened 2016-09-15
Add missing const

sloriot:


lrineau:
Do we have a use-case where that `const` actually matters? if yes, should not it be tested in our testsuite? ..TODO

sloriot:
Done in an uncoming PR

sloriot:
but the PR target 4.10, so I need this one to be merged first.

PR #1470, opened 2016-09-16
Add a Nef_3 constructor from a FaceGraph

sloriot:
See the [small feature page](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Build_Nef_from_FaceGraph)
TODO:
- [x] update `changes.html`

### DIFF
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -49,7 +49,7 @@
 #include <CGAL/Nef_3/Mark_bounded_volumes.h>
 
 #include <CGAL/IO/Verbose_ostream.h>
-#include <CGAL/Nef_3/polyhedron_3_to_nef_3.h>
+#include <CGAL/Nef_3/polygon_mesh_to_nef_3.h>
 #include <CGAL/Nef_3/shell_to_nef_3.h>
 #include <CGAL/Polyhedron_incremental_builder_3.h>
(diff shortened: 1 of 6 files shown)


Nef_3/include/CGAL/Nef_polyhedron_3.h
```diff
@@ -49,7 +49,7 @@
 #include <CGAL/Nef_3/Mark_bounded_volumes.h>
 
 #include <CGAL/IO/Verbose_ostream.h>
-#include <CGAL/Nef_3/polyhedron_3_to_nef_3.h>
+#include <CGAL/Nef_3/polygon_mesh_to_nef_3.h>
```
lrineau: This include will crete a new hard dependency to CGAL BGL package. Do we want that?

As the BGL package is LGPL, that is probably not really an issue. But for non-GeometryFactory developers, that BGL package probably seems like the Algebraic Foundation of MPII: it is more and more an artificial dependency of a lot of CGAL packages.

sloriot: Nef is already a huge machinery. I don't see including some BGL headers as a problem here.

lrineau: As I said, the BGL package is now a dependency of a lot of other package. See https://cgal.geometryfactory.com/~lrineau/CGAL-packages-dependencies/deps.html

Let's keep it that way.

PR #1472, opened 2016-09-20
Polyhedron_demo : Optimize the point_set visualization

maxGimeno:
This PR 
- Makes the resizing of the points real fast
- Makes the resizing of the normals applied on slider released instead of every tick when the point set size is bigger than 300 000
- sets the initial value of the point size to 2 instead of 5

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
@@ -38,10 +38,11 @@ struct Scene_points_with_normal_item_priv
     nb_points = 0;
     nb_selected_points = 0;
     nb_lines = 0;
+    is_point_slider_moving = false;
     normal_Slider = new QSlider(Qt::Horizontal);
     normal_Slider->setValue(20);
     point_Slider = new QSlider(Qt::Horizontal);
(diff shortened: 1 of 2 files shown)


afabri:
Tested on windows. @lrineau : I would be glad if that was in master soon.

Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
```diff
@@ -711,17 +716,25 @@ QMenu* Scene_points_with_normal_item::contextMenu()
       {
         QMenu *container = new QMenu(tr("Normals Length"));
         QWidgetAction *sliderAction = new QWidgetAction(0);
-        connect(d->normal_Slider, &QSlider::valueChanged, this, &Scene_points_with_normal_item::invalidateOpenGLBuffers);
-        connect(d->normal_Slider, &QSlider::valueChanged, this, &Scene_points_with_normal_item::itemChanged);
-
+        if((d->nb_points + d->nb_selected_points)/3 <= 300000)
```
lrineau: I do not like magic constants! But know the constant is even hard-coded twice! Could it be a `const size_t` at the beginning of the file, so that it is at least easy to change?

Issue #1391, opened 2016-08-30
Static variables in Mesh_3 package


sloriot:
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:119:  static double default_value;`
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:120:  static double max_value;`
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:121:  static double min_value;`
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:184:  static double default_value;`
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:185:  static double max_value;`
- [x] `Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h:186:  static double min_value;`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Worksharing_data_structures.h:230:    static tbb::atomic<int> last_cell_index;`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Concurrent_mesher_config.h:58:    static Concurrent_mesher_config singleton;`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Profile_counter.h:31:          { static CGAL::Profile_counter tmp(Y); ++tmp; }`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Profile_counter.h:33:          { static CGAL::Profile_histogram_counter tmp(Y); tmp(Z); }`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Profile_counter.h:35:          static CGAL::Profile_branch_counter NAME(Y); ++NAME;`
- [ ] `Mesh_3/include/CGAL/Mesh_3/Profile_counter.h:39:          static CGAL::Profile_branch_counter_3 NAME(Y); ++NAME;`

afabri:
@cjamin, @lrineau  :  What shall I do with `static Concurrent_mesher_config singleton;` ?

cjamin:
Depends on the purpose:
* If you just want to avoid thread-safety issues, I think it's okay as it is: there is no reason to change the configuration while running parallel meshing processes.
* If you want the user to be able to run several meshing processes in parallel with different configuration parameters, then you need to have one `Concurrent_mesher_config` instance per instance.

And considering this class is not documented and is for internal/advanced use, I think you should go with option 1.

afabri:
@sloriot Can you please check if it is _ready for testing_

[...]

PR #1491, opened 2016-09-26
Mesh_3:  Remove static variables

afabri:
After discussion with @janetournois  we decided to remove the static variables. They are not documented.

This solves the fist six items of Issue #1391. 
The other items concern debugging and will not be made multi thread safe.

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h
+++ b/Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h
@@ -46,6 +46,7 @@ public:
   typedef Cell_vector_ Cell_vector;
 
 public:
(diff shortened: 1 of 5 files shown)


Mesh_3/include/CGAL/Mesh_3/sliver_criteria.h
```diff
@@ -116,10 +117,7 @@ class Min_dihedral_angle_criterion
 public:
   typedef typename Base::Cell_handle    Cell_handle;
 
-  static double default_value;
-  static double max_value;
-  static double min_value;
-
+  virtual double get_default_value() const { return 90.; }
```
janetournois: this function should return `12` to stay consistent with what was there before

PR #1500, opened 2016-09-27
Update .travis.yml

lrineau:


lrineau:
I think we should build CMake ourselves and use the cache to avoid building it at each run.

https://docs.travis-ci.com/user/migrating-from-legacy/#How-Do-I-Cache-Dependencies-and-Directories%3F

PR #1509, opened 2016-09-28
Add index function in Compact container

gdamiand:
This function allows to retreive the index of a given handle. This is particularly usefull to debug code based on compact container. This allows to write index (small numbers) instead of pointers.

This function is undocumented because it is mainly for cgal developers; and thus there is no modification of api, and thus no small feature.

I used this function in CMap test in order to test it.

### DIFF
--- a/Combinatorial_map/include/CGAL/Combinatorial_map.h
+++ b/Combinatorial_map/include/CGAL/Combinatorial_map.h
@@ -1239,11 +1239,11 @@ namespace CGAL {
       for ( typename Dart_range::const_iterator it=darts().begin();
            it!=darts().end(); ++it)
       {
-        os << " dart " << &(*it) << "; beta[i]=";
+        os << " dart " << darts().index(it)<<"; beta[i]=";
         for ( unsigned int i=0; i<=dimension; ++i)
         {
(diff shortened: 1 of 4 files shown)


lrineau:
Fine for me.

Let's hope we will eventually have the indexed-based `Compact_container`.

gdamiand:
Thanks. I agree for the indexed-based Compact_container...

However now thanks to index, operator[], is_used and the fact that when an element is removed we can continue to use iterator on it, the use of Compact_container is much simpler and its use more similar to a std::vector.

I used these functionalities in personal code and it simplify a lot the code...

lrineau:
There are several errors in the testsuite of CMap, today:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-15.shtml#Combinatorial_map

Could it be because of this PR?

gdamiand:
Yes it is.
I will correct this.

gdamiand:
Done in [this] (https://github.com/gdamiand/cgal/commit/f034fe2a928c980e42d842a7ab6304445a7df3ff commit) commit.

I added a test in [this second] (https://github.com/gdamiand/cgal/commit/5626eeae92892a1775e995045899d1137fe30d1a) commit

PR #1516, opened 2016-09-29
Remove the requirement table

sloriot:
I discussed it with @janetournois and the table is not that useful, in particular since it is not always updated when functions are added to the package.
We think it should be removed as the information provided is already in the documentation of each function.

@palliez (as the reviewer who asked for that feature): are you OK with this change?


palliez:
Sure! that's a good idea.

PR #1517, opened 2016-09-29
More bbox 3 overloads for face graph models

sloriot:
It add `vertex_bbox_3()`, `edge_bbox_3()`, and `face_bbox_3()`.

I would prefer having 3 more `bbox_3()` overloads for  but since these functions have 2 arguments this would cause an ambiguity with the 2 argument version of `bbox_3()` for a halfedge list graph model.

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
@@ -130,6 +130,9 @@ and provides a list of the parameters that are used in this package.
 - `CGAL::Polygon_mesh_slicer`
 - `CGAL::Side_of_triangle_mesh`
 - `CGAL::Polygon_mesh_processing::bbox_3()`
+- `CGAL::Polygon_mesh_processing::vertex_bbox_3()`
+- `CGAL::Polygon_mesh_processing::edge_bbox_3()`
+- `CGAL::Polygon_mesh_processing::face_bbox_3()`
 - `CGAL::Polygon_mesh_processing::border_halfedges()`
 
 
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
@@ -130,9 +130,6 @@ and provides a list of the parameters that are used in this package.
 - `CGAL::Polygon_mesh_slicer`
 - `CGAL::Side_of_triangle_mesh`
 - `CGAL::Polygon_mesh_processing::bbox_3()`
-- `CGAL::Polygon_mesh_processing::vertex_bbox_3()`
-- `CGAL::Polygon_mesh_processing::edge_bbox_3()`
-- `CGAL::Polygon_mesh_processing::face_bbox_3()`
 - `CGAL::Polygon_mesh_processing::border_halfedges()`
 
 
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
@@ -129,6 +129,9 @@ and provides a list of the parameters that are used in this package.
 ## Miscellaneous ##
 - `CGAL::Polygon_mesh_slicer`
 - `CGAL::Side_of_triangle_mesh`
+- `CGAL::Polygon_mesh_processing::vertex_bbox_3()`
+- `CGAL::Polygon_mesh_processing::edge_bbox_3()`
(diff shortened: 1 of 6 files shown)


sloriot:
This is a similar problem we encountered with `area(pm)` and `face_area(f,pm)`.

Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
```diff
@@ -129,6 +129,9 @@ and provides a list of the parameters that are used in this package.
 ## Miscellaneous ##
 - `CGAL::Polygon_mesh_slicer`
 - `CGAL::Side_of_triangle_mesh`
+- `CGAL::Polygon_mesh_processing::vertex_bbox_3()`
+- `CGAL::Polygon_mesh_processing::edge_bbox_3()`
+- `CGAL::Polygon_mesh_processing::face_bbox_3()`
 - `CGAL::Polygon_mesh_processing::bbox_3()`
```
janetournois: can you please move `bbox_3` to be the first item of this list?

afabri:
not really beautiful.

sloriot:
@afabri I'd be glad if you find better names.

lrineau:
Should not there be a small feature?

sloriot:
I finally found a way to make it work with overloads of `bbox_3()`.
The trick was to add an empty specialization of `graph_traits` for named parameters.

@lrineau  now that it works with a overloads, I don't see the point of a small feature.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/bbox.h
```diff
@@ -224,25 +224,22 @@ namespace CGAL {
...
     {
-      return edge_bbox_3(ed, pmesh,
-        CGAL::Polygon_mesh_processing::parameters::all_default());
+      return bbox_3(ed, pmesh, parameters::all_default());
     }
     template<typename PolygonMesh>
-    CGAL::Bbox_3 face_bbox_3(typename boost::graph_traits<PolygonMesh>::face_descriptor fd,
-                             const PolygonMesh& pmesh)
+    CGAL::Bbox_3 bbox_3(typename boost::graph_traits<PolygonMesh>::face_descriptor fd,
+                        const PolygonMesh& pmesh)
```
lrineau: I am completely against that sort of abuse of overload. That will lead to compilation errors for any `PolygonMesh` when all three types of descriptors are simple integers.

janetournois:
nice trick!! :)

sloriot:
Two solutions: I revert the last commit and we keep the initial names (prefixed with the simplex) or we close this PR.
@afabri , @lrineau  what do you think?

lrineau:
I really prefer the first solution (do not use overloads). I wonder why our testsuite does not show issues with overloads. Do not we have `FaceGraph`s that have simple integers for descriptors?!

sloriot:
We don't because for type safety, the descriptors in `Surface_mesh` as well as in `Open_mesh` are different index classes. But you're right that I know a user that has a hydrid c++/java model that is forced to use pure integer and this would be a problem. I even remember having seen this pb with at least another function in CGAL.

lrineau:
What is the "type-safety" you are talking about? If that is the ability to use overloads... well, that was a bad design choice.

sloriot:
No, Vertex_index != Halfedge_index in order not to use one instead of the other. Because at the end you have integers and you better have a compilation error rather than a runtime error.

lrineau:
@sloriot I suggest you  revert commit 357f26123133886b54bb9157b25 (to go back to `vertex_bbox_3`, `edge_bbox_3`...) and we go on. No need to cancel or stall this PR.

sloriot:
@afabri  do you agree?

afabri:
I do not particularly like the `vertex_`  prefix, but it does not shock me.

lrineau:
A warning in my local build:
```
/cgal/Polygon_mesh_processing/test/Polygon_mesh_processing/../../include/CGAL/Polygon_mesh_processing/bbox.h: In function ‘CGAL::Bbox_3 CGAL::Polygon_mesh_processing::edge_bbox_3(typename boost::graph_traits<Graph>::edge_descriptor, const PolygonMesh&, const NamedParameters&)’:
/cgal/Polygon_mesh_processing/test/Polygon_mesh_processing/../../include/CGAL/Polygon_mesh_processing/bbox.h:169:78: warning: typedef ‘halfedge_descriptor’ locally defined but not used [-Wunused-local-typedefs]
       typedef typename boost::graph_traits<PolygonMesh>::halfedge_descriptor halfedge_descriptor;
                                                                              ^~~~~~~~~~~~~~~~~~~
```

PR #1518, opened 2016-09-29
Update cgal_manual.bib

afabri:
fix an url


Documentation/doc/biblio/cgal_manual.bib
```diff
@@ -266,7 +266,7 @@ @article{cgal:bo-pgsms-05
 , pages =   "405--451"
 }
 
-@misc{ cgal:bpp-vrml-95
+@misc{ cgal:bpp--95
```
sloriot: If you change the reference name, you'll break the \cgalCite in the manual.

PR #1524, opened 2016-09-29
Make split_graph_into_polylines deterministic

lrineau:
- There is a new argument, `Less`, that allows to sort the `std::set` of
  vertex descriptors with a deterministic sort functor.

- The header `<CGAL/Mesh_3/polylines_to_protect.h>` defines and use such
  a functor.


lrineau:
This PR includes PR #1514 on purpose, to fix the conflicts between the two.

PR #1537, opened 2016-10-03
Replace 'see also' by 'has model' and 'models'

### LINKED ISSUE #1532: Kernel doc:  Replace \sa with \HasModels and \isModel
The kernel object man pages, e.g. for [`Point_3`](http://doc.cgal.org/latest/Kernel_23/classCGAL_1_1Point__3.html)  must use `\isModel` instead of `\sa`.   Similar for the concepts.

sloriot:
Fixes #1532


sloriot:
Tested locally

Issue #1258, opened 2016-07-09
2D alpha shapes with periodic Delaunay triangulation


grzjab:
Can you add  OUTSIDE_CONVEX_HULL and  OUTSIDE_AFFINE_HULL in  enum Locate_type in Periodic_2_triangulation_2.h for compatibility with Alpha_shape_2?
Periodic_3_triangulation_3.h has it:

```
 OUTSIDE_CONVEX_HULL, // unused, for compatibility with Alpha_shape_3
 OUTSIDE_AFFINE_HULL }; // unused, for compatibility with Alpha_shape_3
```

I found the bug while trying to compile the following code:

```
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/Periodic_2_triangulation_filtered_traits_2.h>
#include <CGAL/Periodic_2_Delaunay_triangulation_2.h>
#include <CGAL/Triangulation_vertex_base_with_info_2.h>
#include <CGAL/Alpha_shape_2.h>
 //**********TYPEDEFS********************
 typedef CGAL::Exact_predicates_exact_constructions_kernel K;
 typedef CGAL::Periodic_2_triangulation_traits_2<K> GT;
 typedef CGAL::Triangulation_vertex_base_with_info_2<unsigned, K> Vb_info;
 typedef CGAL::Alpha_shape_vertex_base_2<K, Vb_info>    Vb;
 typedef CGAL::Alpha_shape_face_base_2<K>                                               Fb;
 typedef CGAL::Triangulation_data_structure_2<Vb, Fb>    Tds;
 typedef CGAL::Periodic_2_Delaunay_triangulation_2<GT, Tds>  PDT;
 typedef CGAL::Alpha_shape_2<PDT> Alpha_shape_2;

 int main()
 {
        Alpha_shape_2 as;
 }
```

EDIT: Output from compiler:

```
In file included from /home/grzegorz/work/delaunay/example.cpp:5:0:
[...]

PR #1540, opened 2016-10-04
Compatibility issue with Alpha_shape_2

grzjab:
This pull request is directly connected with the issue https://github.com/CGAL/cgal/issues/1258

### DIFF
--- a/Periodic_2_triangulation_2/include/CGAL/Periodic_2_Delaunay_triangulation_2.h
+++ b/Periodic_2_triangulation_2/include/CGAL/Periodic_2_Delaunay_triangulation_2.h
@@ -427,13 +427,16 @@ public:
         return std::make_pair(fit, eit);
       case Triangulation::FACE:
       case Triangulation::EDGE:
-      case Triangulation::EMPTY:
+      case Triangulation::EMPTY: {
         *fit++ = fh; //put fh in OutputItFaces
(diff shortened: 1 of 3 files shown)


lrineau:
Hi @grzjab

This PR has caused a lot of new warnings in the 2D periodic triangulation. Could you check locally on your machine and try to fix the warnings?

I think we just need to add some cases in the switch:
```C++
case OUTSIDE_CONVEX_HULL: CGAL_error();
case OUTSIDE_AFFINE_HULL: CGAL_error();
```
`CGAL_error()` is a macro that evaluates to a function that is tagged with the attribute `noreturn`. At compile time, the optimizers of compilers will remove the branches of those cases.

CC @afabri. See also #1547.

afabri:
And please put a comment, that this cannot happen in a periodic triangulation,  but that otherwise we get a warning.
And where the enum  is add also a comment that although they make sense, they are added for periodic alpha shapes.

lrineau:
I have pushed a fix myself.

grzjab:
@lrineau I just realized that there is actually much more to do, if one wants to use it. I wasn't aware of the branch by nkruithof.

lrineau:
Do you want to cancel the pull-request?

grzjab:
I think it makes much more sense. Sorry for that.

MaelRL:
Closing this PR in favor of PR https://github.com/CGAL/cgal/pull/2369.

PR #1541, opened 2016-10-04
Polyhedron_demo: Fix spheres in polylines

### LINKED ISSUE #1531: Polyhedron demo : change radius of spheres in polylines graph
The scenario 
- Open a `file.polylines.txt`
- Right click on the item --> `Display corners with radius`
The radius I have choosen is too large (resp. too small), I want to change it
- Right click again --> `Display corners with radius`

Another set of spheres is drawn (a second one), with the new radius, but the initial ones remain.

Also, the `Corner spheres` sub-item cannot be removed from the Geometric Objects list without deleting its master item.

@maxGimeno

maxGimeno:
Fix #1531.

It also fixes a memory leak in the way the CGAL::Sphere_3 were cleared.


lrineau:
Please rebase the branch onto `releases/CGAL/CGAL-4.9-branch`.

PR #493, opened 2015-11-19
Remove boost thread dependency

sloriot:
https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/threadlocal_and_mutex

PR #1542, opened 2016-10-04
try_run to detect the C++ version

lrineau:
Before this PR, there was a complicated `if()` trying to detect the C++
version. That is simpler to use `try_run()` and read the version in
the value of `__cplusplus`.

### DIFF
--- a/Installation/cmake/modules/CGAL_SetupBoost.cmake
+++ b/Installation/cmake/modules/CGAL_SetupBoost.cmake
@@ -5,23 +5,16 @@ if ( NOT CGAL_Boost_Setup )
   set ( CGAL_requires_Boost_libs TRUE )
   if ( DEFINED  MSVC_VERSION AND "${MSVC_VERSION}" GREATER 1800)
     set ( CGAL_requires_Boost_libs FALSE )
-  endif()
-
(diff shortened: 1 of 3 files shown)


sloriot:
Your new test should be ran only if `CMAKE_COMPILER_IS_GNUCXX` is used.
See [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-14/AABB_tree/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.0_Release-cpp11.gz) for example. If you can fix the test for clang++ too, the documentation should be updated. See [changes](https://github.com/CGAL/cgal/pull/493/files#diff-e1f393a91b7709c4d3a8e73c552a98a7) introduced by the small feature.
See also http://clang.llvm.org/cxx_status.html#n2659
I remember than @mglisse commented on this but I can't find his comments.

mglisse:
I am not sure what I am supposed to have commented on... Anyway, `__cplusplus` is quite unreliable. VC hasn't changed it yet, while gcc and clang update it before implementing any of the features...

lrineau:
I have modified the .cpp file to use the exact same logic as in `<CGAL/config.h>`. That is what actually matters.

Issue #1387, opened 2016-08-30
Static variables in Number_type package


sloriot:
- [x] `Number_types/include/CGAL/Sqrt_extension/convert_to_bfi.h:58:  static Cache_type cache;`
- [x] `Number_types/include/CGAL/Lazy_exact_nt.h:488:  static double relative_precision_of_to_double;`
- [x] `Number_types/include/CGAL/Mpzf.h:139:  static std::vector<T> data;`
- [x] `Number_types/include/CGAL/Mpzf.h:159:    static CGAL_MPZF_TLS T data_ = 0;`
- [x] `Number_types/include/CGAL/Mpzf.h:183:    static thread_local cleaner obj;`
- [x] `Number_types/include/CGAL/Interval_nt.h:216:  static Test_runtime_rounding_modes tester;`
- [x] `Number_types/include/CGAL/Counted_number.h:37:    static unsigned long s_neg_count, s_add_count, s_sub_count,`
- [x] `Number_types/include/CGAL/GMP/Gmpzf_type.h:487:  static Gmpz s;`

afabri:
I don't see where Counted_number.h is used.

lrineau:
@sloriot @afabri Can you please check what has been "fixed" by PR #1375, check the corresponding checkbox at https://github.com/CGAL/cgal/issues/1387#issue-174031770 and maybe close this issue?

afabri:
PR #1375  only addresses that the relative precision of a `Lazy_exact_nt`  is now thread local and that this gets documented.

lrineau:
OK, so this issue remains. Thanks.

afabri:
Nothing to do for Interval_nt, as it is a test whether GCC's -frounding-math option has been forgotten.

afabri:
@lrineau I  think it is ready for testing

lrineau:
It seems this issue has been eventually fixed by #1734.

PR #648, opened 2016-01-18
Filtered kernel: Epeck reentrance

lrineau:
Add re-entrance to the `CGAL::Epeck` kernel.
- Use atomic types in `<CGAL/Handle_for.h>`,
- Use a mutex in `<CGAL/Lazy.h>`.

Depends on #642.

Fixes #4340.

PR #1548, opened 2016-10-06
Number_types / Gmpzf: make static variable thread local

afabri:
Addresses one item of Issue #1387

### DIFF
--- a/Number_types/include/CGAL/GMP/Gmpzf_type.h
+++ b/Number_types/include/CGAL/GMP/Gmpzf_type.h
@@ -23,6 +23,7 @@
 
 // includes
 #include <CGAL/basic.h>
+#include <CGAL/tss.h>
 #include <CGAL/Handle_for.h>
 #include <CGAL/gmp.h>
 #include <mpfr.h>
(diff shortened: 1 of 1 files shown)


mglisse:
If Gmpzf is now thread-safe, you may want to update Exact_ring_selector (in internal/Exact_type_selector.h) and remove the test on CGAL_HAS_THREADS (maybe after a quick benchmark to check that it is indeed a bit faster).

lrineau:
As far as I know `Gmpzf` is still not thread safe when threads shares numbers. It is only safe is different threads use different numbers.

The ref counting in `Handle_for` has to be thread-safe first, and it still is not. See the unfinished PR #648. Maybe the changes in `Handle_for` to use atomic types for the ref counting (https://github.com/CGAL/cgal/pull/648/files#diff-04b0c45a7dafd24ac604ae2d4639fb32R41) should be integrated in a small PR.

mglisse:
> It is only safe if different threads use different numbers.

That's what I was talking about. In Exact_field_selector, we use Gmpq even when there are threads, so I was assuming that we could do the same for rings. As far as I know, the ring selector is only used for some filtered predicates, where we do not share numbers between threads.

> Maybe the changes in Handle_for to use atomic types for the ref counting should be integrated in a small PR.

I thought the performance impact was too high. Now that I look again at the benchmark, while Epeck was horrible, there is also a column Simple_cartesian, which I assume from the running time was `Simple_cartesian<Gmpq>`. That does seem to indicate that the cost for Gmpq (one of the main users of Handle_for) is quite reasonable indeed.

Issue #135, opened 2015-06-25
 operator<< (std::ostream & os, const Lazy_exact_nt<ET> & a) should insert a.exact() in the stream


lrineau:
[The documentation of Lazy_exact_nt in CGAL-4.6](http://doc.cgal.org/4.6/Number_types/classCGAL_1_1Lazy__exact__nt.html#a2341c2358a9c1f61ff433519f927f5b6) says:

> `std::ostream & 	operator<< (std::ostream &out, const Lazy_exact_nt< NT > &m)`
> 	writes `m` to `ostream` `out` in an interval format. 

Well. That is false. [The actual code inserts `to_double(a)`](https://github.com/CGAL/cgal/blob/708c13570369a8a4193129586bbbae5f7c3870a4/Number_types/include/CGAL/Lazy_exact_nt.h#L1297).

*Anyway*, if we want that a save and a load preserve the value, then the `operator<<` for `Lazy_exact_nt<ET>` should insert the `exact()` value in the stream, so that no information is lost.

 @CGAL/developers: Remarks?

lrineau:
See my commit CGAL/cgal-dev@40f8f0d

sloriot:
If we set this behavior by default, this will probably break user code. A least I'm rather sure I'm using the fact that a double is output in debug code.

lrineau:
There could be a special manipulator, similar to `[CGAL::set_pretty_mode`](http://doc.cgal.org/latest/Stream_support/group__PkgIOstreams.html#ga60b0d5aa93fdf2b19089d7591de4b055), and the `operator<<` would test the mode. The default would be the current behavior.

As the current behavior of `CGAL::Gmpq` is to insert the exact representation, the default behavior (when no flag is set in the stream) would be different for `Gmpq` and `Lazy_exact_nt`.

sgiraudot:
Is this issue still valid?

If so, it seems to me the main problem to solve here is the inconsistency between the documentation and the code. As @sloriot pointed, changing the code might break user code, so a first easy step could be to change the documentation to make it clear that the numbers are printed in their `double` conversion.

Providing a way to write the number in an interval format seems like a different issue to me (that probably requires more discussion).

sgiraudot:
Was this issue solve by https://github.com/CGAL/cgal/pull/1551?

mglisse:
Nope, #1551 did not change anything about this.

MoniqueTeillaud:
[...]

PR #1551, opened 2016-10-06
Misc GMPXX fixes

mglisse:


lrineau:
There are still errors:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-16/Apollonius_graph_2/TestReport_lrineau_x86-64_Linux-Fedora19_llvm-clang-with-g++-4.8.gz

mglisse:
Right, we would need to specify `sign_a_plus_b_x_sqrt_c<RT>`. There is at least a similar issue with `sign_of_determinant` in internal/Triangulation_2_projection_traits_base_3.h, and probably a few more (when I click on the red 'n's, it regularly crashes my browser, whether I use firefox or chrome...). Those common issues are the reason I think we should only apply this PR if the benchmarks show a sufficient gain.
I should really find a machine and run the whole testsuite myself before polluting the testresults again...

mglisse:
Note that the linked testsuite error is not a problem with this PR, it is a separate issue. But I'll make the changes in the same branch if you don't want to multiply the PRs...

lrineau:
If might be a separate issue but it has been triggered either by this PR #1551 or the other PR #1257.

As this PR #1551 is a bug-fix pull-request, I would prefer you added other related patches in the same PR.

mglisse:
It is clearly triggered by PR #1257 , almost nothing uses gmpxx without it. Ok for adding other bug fixes to this branch.

Circular_kernel_3/test/Circular_kernel_3/include/CGAL/_test_sphere_compute.h
```diff
@@ -73,11 +73,11 @@ void _test_spherical_kernel_compute(SK sk)
...
 
   rt[0] = Root_for_spheres_2_3(0,1,0);
-  rt[1] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),-FT(FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(FT(1)/FT(2)),FT(2)),0);
+  rt[1] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(-FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(FT(1)/FT(2)),FT(2)),0);
   rt[2] = Root_for_spheres_2_3(-1,0,0);
-  rt[3] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),-FT(FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),-FT(FT(1)/FT(2)),FT(2)),0);
+  rt[3] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(-FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(-FT(1)/FT(2)),FT(2)),0);
   rt[4] = Root_for_spheres_2_3(0,-1,0);
-  rt[5] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),-FT(FT(1)/FT(2)),FT(2)),0);
+  rt[5] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(FT(1)/FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(-FT(1)/FT(2)),FT(2)),0);
```
lrineau: I am sure the patch is good. It seems harmless. But I do not understand its implications. @mglisse, can you comment it?

The initial code was `-FT(FT(1)/FT(2))`. `FT(1)/FT(2)` is supposed to be an `FT` (because that is a field). I imagine that the cast is there because of expression-templates: `FT(1)/FT(2)` has a special type that is convertible to `FT`. Then what does it change to put the negation inside? The negation returns also a proxy object?

Circular_kernel_3/test/Circular_kernel_3/include/CGAL/_test_sphere_predicates.h
```diff
@@ -579,11 +579,11 @@ void _test_do_overlap_predicate(SK sk) {
...
 
   rt[0] = Root_for_spheres_2_3(0,1,0);
-  rt[1] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),-FT(FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(FT(1) / FT(2)),FT(2)),0);
+  rt[1] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(-FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(FT(1) / FT(2)),FT(2)),0);
   rt[2] = Root_for_spheres_2_3(-1,0,0);
-  rt[3] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),-FT(FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),-FT(FT(1) / FT(2)),FT(2)),0);
+  rt[3] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(-FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(-FT(1) / FT(2)),FT(2)),0);
   rt[4] = Root_for_spheres_2_3(0,-1,0);
-  rt[5] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),-FT(FT(1) / FT(2)),FT(2)),0);
+  rt[5] = Root_for_spheres_2_3(CGAL::make_root_of_2(FT(0),FT(FT(1) / FT(2)),FT(2)), CGAL::make_root_of_2(FT(0),FT(-FT(1) / FT(2)),FT(2)),0);
```
lrineau: With so many usage of `FT(1)/FT(2)`, should not create constantes:
```C++
const FT one = FT(1);
const FT two = FT(2);
const decltype(auto) half = one / two;
```
and the equivalent for previous versions of C++. Do we have a type traits to retrieve the type of `half`?

afabri: for readability?

Circular_kernel_3/test/Circular_kernel_3/include/CGAL/_test_sphere_compute.h
mglisse: Yes, negation returns a proxy, same as addition, etc, and when the arguments of make_root_of_2 have different types, template deduction fails. If it hadn't been in test/, I might have gone with something like `make_root_of_2<FT>(0,FT(-1)/2,2)`, or even `make_root_of_2<FT>(0,.5,2)`.
Our number type stuff is rather painful to work with. One detail I don't like is having to write `x*FT(2)`, whereas we do not support any type for which `x*2` wouldn't work at least as well.

lrineau: Maybe we should say that  the constructor in [FromIntConstructible] must be implicit.

Note that you can replace `FT(-FT(1)/FT(2))` by `FT(FT(-1)/FT(2))`: that saves one operation.

Second note: there is a bug in [FromIntConstructible]: if `int` is 64 bits, even a `double` is not really constructible from an `int`. We should restrict the set of values to 32 bits, in [`INT32_MIN`, `INT32_MAX`].

[FromIntConstructible]: http://doc.cgal.org/latest/Algebraic_foundations/classFromIntConstructible.html

Circular_kernel_3/test/Circular_kernel_3/include/CGAL/_test_sphere_predicates.h
lrineau: My proposal is not about readability, but about the memory allocations, in case `FT` is a reference counted type (like `CGAL::Epeck::FT`). Here I see so many created objects that could be replaced by constant objects!

mglisse: I don't think reference counting is relevant here, the objects are not copied, just used multiple times, so the gains would be the same with mpq_class. But those are just a few numbers in one testcase, performance doesn't seem particularly important.

mglisse:
I have done one full pass over the build failures (ignoring the runtime failures), and the only remaining one (mentioned in the commit message) is where we try to build an Epeck::FT from a std::string. I may rerun the testsuite one of these days to check if new issues appear, but it takes forever on a laptop.

mglisse:
I/O of `Lazy_exact_nt<mpq_class>` is broken. operator<< goes through to_double and prints something like "12.34", while operator>> forwards directly to mpq_class, which only accepts "617/50". This is issue #135.

mglisse:
Some places expect to be able to read 12.34 in an exact kernel. For Lazy_exact_nt, we could specialize operator>> using `read_float_or_quotient<mpz_class,mpq_class>`, but that doesn't handle `Cartesian<mpq_class>` (or Simple_cartesian, or filtered versions of those). We could also make sure operator>>(*,Point_2&) does the right thing. But at some point, users may read numbers directly and it becomes impossible to avoid the fact that mpq_class will not accept 12.34.

lrineau:
Thanks, Marc. I has been tested, and then merged into `master`.

PR #1556, opened 2016-10-07
Polyhedron_demo: Enhance the raw image management

### LINKED ISSUE #1471: Polyhedron demo: improve loading of raw images
Suggested improvements in the loading of 3D RAW images.

  - [ ] check if the computed image size is the right one, compared to the actual size of the RAW file
  - [ ] the size of the image should taken into account the offset.
  - [ ] the fields of the dialog should have tooltips, in particular for the spacing, dimension, offset
  - [ ] if the workKind of the image is WK_FLOAT, then we know that the image is a gray-level image: no need to ask
  - [ ] allow to save the image to the .inr.gz format
  - [ ] if the image format is not right to use it as-is for the mesher, there should be a way to convert it.

See below the usual dialog:

password1

maxGimeno:
This fixes #1471  

This PR :
- won't let the user open a raw image if its size does not match the File's size
- takes the offset into account for the size's calculation
- adds tooltips to the fields of the dialog
- automates the Gray-image choice if the WORD_KIND is not WK_FIXED with a wsize of 1.
- makes it possible to save the selected rendered image to the format inr.gz
- converts all raw images to a meshable format when they are opened.

 @lrineau please check if the tooltips are explicit enough.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
@@ -806,6 +806,24 @@ bool Io_image_plugin::canLoad() const {
   return true;
 }
 
+template<typename Word>
+void convert(Image* image)
+{
+  float *f_data = (float*)ImageIO_alloc(image->xdim()*image->ydim()*image->zdim()*sizeof(float));
+  Word* d_data = (Word*)(image->data());
+  //convert image from double to float
+  for(std::size_t x = 0; x<image->xdim(); ++x)
+    for(std::size_t y = 0; y<image->ydim(); ++y)
+      for(std::size_t z = 0; z<image->zdim(); ++z)
+      {
+        std::size_t i =(z * image->ydim() + y) * image->xdim() + x;
+        f_data[i] =(float)d_data[i];
+      }
+  ImageIO_free(d_data);
+  image->image()->data = (void*)f_data;
+  image->image()->wdim = 4;
+  image->image()->wordKind = WK_FLOAT;
+}
 CGAL::Three::Scene_item*
 Io_image_plugin::load(QFileInfo fileinfo) {
   QApplication::restoreOverrideCursor();
@@ -846,64 +864,20 @@ Io_image_plugin::load(QFileInfo fileinfo) {
             {
             case WK_FLOAT:
               is_gray = true;
-              if(raw_dialog.image_word_size() == 8)
-              {
-                float *f_data = (float*)ImageIO_alloc(image->xdim()*image->ydim()*image->zdim()*sizeof(float));
-                double* d_data = (double*)(image->data());
-                //convert image from double to float
-                for(std::size_t x = 0; x<image->xdim(); ++x)
-                  for(std::size_t y = 0; y<image->ydim(); ++y)
-                    for(std::size_t z = 0; z<image->zdim(); ++z)
-                    {
-                      std::size_t i =(z * image->ydim() + y) * image->xdim() + x;
-                      f_data[i] =(float)d_data[i];
-                    }
-                ImageIO_free(d_data);
-                image->image()->data = (void*)f_data;
-                image->image()->wdim = 4;
-              }
+              convert<double>(image);
               break;
             case WK_FIXED:
             {
-              float *f_data = (float*)ImageIO_alloc(image->xdim()*image->ydim()*image->zdim()*sizeof(float));
               switch(raw_dialog.image_word_size())
               {
               case 2:
-              {
                 is_gray = true;
-                //convert image from short to char
-                short* s_data = (short*)(image->data());
-                for(std::size_t x = 0; x<image->xdim(); ++x)
-                  for(std::size_t y = 0; y<image->ydim(); ++y)
-                    for(std::size_t z = 0; z<image->zdim(); ++z)
-                    {
-                      std::size_t i =(z * image->ydim() + y) * image->xdim() + x;
-                      f_data[i] =(char)s_data[i];
-                    }
-                ImageIO_free(s_data);
-                image->image()->data = (void*)f_data;
-                image->image()->wdim = 4;
-                image->image()->wordKind = WK_FLOAT;
+                convert<short>(image);
                 break;
-              }
               case 4:
-              {
                 is_gray = true;
-                //convert image from int to char
-                int* i_data = (int*)(image->data());
-                for(std::size_t x = 0; x<image->xdim(); ++x)
-                  for(std::size_t y = 0; y<image->ydim(); ++y)
-                    for(std::size_t z = 0; z<image->zdim(); ++z)
-                    {
-                      std::size_t i =(z * image->ydim() + y) * image->xdim() + x;
-                      f_data[i] =(int)i_data[i];
-                    }
-                ImageIO_free(i_data);
-                image->image()->data = (void*)f_data;
-                image->image()->wdim = 4;
-                image->image()->wordKind = WK_FLOAT;
+                convert<int>(image);
                 break;
-              }
               default:
                 is_gray = false;
                 break;
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
@@ -848,7 +848,7 @@ Io_image_plugin::load(QFileInfo fileinfo) {
               is_gray = true;
               if(raw_dialog.image_word_size() == 8)
               {
-                float *f_data = new float[image->xdim()*image->ydim()*image->zdim()];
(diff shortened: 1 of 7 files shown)


Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Raw_image_dialog.cpp
```diff
@@ -1,19 +1,27 @@
...
-  label_image_size->setNum((int)image_word_size() *
-			   dim_x->value() *
-			   dim_y->value() *
-			   dim_z->value());
+  int size = (int)image_word_size() *
+      dim_x->value() *
+      dim_y->value() *
+      dim_z->value() +
+      offset->value();
+  label_image_size->setText(QString("%1 B").arg(size));
```
lrineau: `int` is a too-short type the size of an image. As soon as the image will be bigger than `2GB`, that will fail on Windows (when `int` is 32 bits). You must use `std::size_t` (the right type to code for a size of a container).

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Raw_image_dialog.cpp
```diff
@@ -1,19 +1,27 @@
...
+
+  if(label_image_size->text() == label_file_size->text())
+    buttonBox->button(QDialogButtonBox::Open)->setEnabled(true);
+  else
+    buttonBox->button(QDialogButtonBox::Open)->setEnabled(false);
```
lrineau: Maybe add a tooltip, when the button is disabled, that explains why. And remove the tooltip once is it enabled.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Raw_image_dialog.cpp
```diff
@@ -26,3 +34,18 @@ unsigned int Raw_image_dialog::image_word_size() const {
     return 1;
 }
```
lrineau: Here I can understand why the type was `int`: the possible values are only 1 or 2. But the change is right too, to avoid a conversion warning I suppose?

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -1,62 +1,81 @@
...
-        <property name="text" >
+     <layout class="QGridLayout" name="gridLayout">
+      <item row="2" column="0">
+       <widget class="QRadioButton" name="short_bt">
+        <property name="toolTip">
+         <string>Word kind of the image : short int.</string>
+        </property>
+        <property name="text">
          <string>Short (16 bits)</string>
         </property>
```
lrineau: I do not see the need for a tooltip. Just change the text of the button to "Short int (16 bits)" and that will add the same information.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -1,62 +1,81 @@
...
-       <widget class="QRadioButton" name="float_bt" >
-        <property name="text" >
+      <item row="3" column="0">
+       <widget class="QRadioButton" name="float_bt">
+        <property name="toolTip">
+         <string>Word kind of the image : float.</string>
+        </property>
+        <property name="text">
          <string>Float</string>
         </property>
```
lrineau: ... No comment.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -1,62 +1,81 @@
...
        </widget>
       </item>
-      <item row="0" column="0" >
-       <widget class="QRadioButton" name="char_bt" >
-        <property name="text" >
+      <item row="0" column="0">
+       <widget class="QRadioButton" name="char_bt">
+        <property name="toolTip">
+         <string>Word kind of the image : unsigned byte.</string>
```
lrineau: The tooltip's text is wrong: it can also be a "*signed* char". Anyway, just remove all those tooltips!

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -65,161 +84,191 @@
...
       <item>
-       <layout class="QGridLayout" >
-        <item row="0" column="0" >
-         <widget class="QLabel" name="label_dim" >
-          <property name="text" >
+       <layout class="QGridLayout">
+        <item row="0" column="0">
+         <widget class="QLabel" name="label_dim">
+          <property name="toolTip">
+           <string>Number of voxel along the axis.</string>
```
lrineau: That one is good. It does provide a better information than the text of the label itself.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -65,161 +84,191 @@
...
           </property>
          </widget>
         </item>
-        <item row="2" column="0" >
-         <widget class="QLabel" name="label_spacing" >
-          <property name="text" >
+        <item row="2" column="0">
+         <widget class="QLabel" name="label_spacing">
+          <property name="toolTip">
+           <string>Changing those values will scale the image according to the proportions between vx, vy and vz.</string>
```
lrineau: :+1:

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -65,161 +84,191 @@
...
           </property>
          </widget>
         </item>
-        <item row="3" column="0" >
-         <widget class="QLabel" name="offset_label" >
-          <property name="text" >
+        <item row="3" column="0">
+         <widget class="QLabel" name="offset_label">
+          <property name="toolTip">
+           <string>Specifies how many bytes must be skipped before reading the data. (Use it to skip unhandled headers fo example)</string>
```
lrineau: typo: "fo example"

and maybe add a `.` at the end of the sentence inside the parenthesis.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/raw_image.ui
```diff
@@ -65,161 +84,191 @@
...
           </property>
          </widget>
         </item>
-        <item row="3" column="1" >
-         <widget class="QSpinBox" name="offset" >
-          <property name="suffix" >
+        <item row="3" column="1">
+         <widget class="QSpinBox" name="offset">
+          <property name="toolTip">
+           <string>Specifies how many bytes must be skipped before reading the data. (Use it to skip unhandled headers fo example)</string>
```
lrineau: Same typo.

Is that really necessary to repeat the tooltip?

Polyhedron/demo/Polyhedron/Scene_image_item.cpp
```diff
@@ -333,7 +333,7 @@ Vertex_buffer_helper::push_vertex(std::size_t i, std::size_t j, std::size_t k)
     dk = 0.5;
   if (di == data_.xdim())
     di = data_.xdim()-0.5;
-  if (dj == data_.xdim())
+  if (dj == data_.ydim())
```
lrineau: :+1: Good catch!

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
```diff
@@ -816,17 +825,94 @@ Io_image_plugin::load(QFileInfo fileinfo) {
...
+                ImageIO_free(s_data);
+                image->image()->data = (void*)f_data;
+                image->image()->wdim = 4;
+                image->image()->wordKind = WK_FLOAT;
+                break;
+              }
+              case 4:
+              {
+                is_gray = true;
+                //convert image from int to char
```
lrineau: This comment is wrong. The conversion is done to `float`.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
```diff
@@ -816,17 +825,94 @@ Io_image_plugin::load(QFileInfo fileinfo) {
...
+                int* i_data = (int*)(image->data());
+                for(std::size_t x = 0; x<image->xdim(); ++x)
+                  for(std::size_t y = 0; y<image->ydim(); ++y)
+                    for(std::size_t z = 0; z<image->zdim(); ++z)
+                    {
+                      std::size_t i =(z * image->ydim() + y) * image->xdim() + x;
+                      f_data[i] =(int)i_data[i];
+                    }
+                ImageIO_free(i_data);
```
lrineau: Is not there a way to factorize all those similar loops in a member function templated by the input type and the output type?

lrineau:
I have added a lot of comments and requests for changes, but the modifications are globally pretty good. Thanks.

PR #1550, opened 2016-10-06
`packages_dependencies`: deal with OpenMesh

lrineau:
@sloriot: I have pushed my "fix" that deal with OpenMesh. It is now a non-required dependency for the target `packages_dependencies`.

PR #1563, opened 2016-10-10
Protect code that requires gsl

sloriot:
Related to #1550


lrineau:
I have merged the PR.

PR #1564, opened 2016-10-10
Fix issue #1554

### LINKED ISSUE #1554: Mesh_3 bug: The surface of c3t3 has holes
With the .cpp and the data shipped in https://gist.github.com/5e71e899c85b949487b0316e1f6e064b, one can see that the surfacic part of the c3t3 has holes.

The issue is that the circumcenters are not computed exactly, and results of queries to the domain oracle are not coherent.

I think (to be confirmed) that bug was not a bug before the commit fbf824c7113f6c9f3f9faf95cd8d458a596cc162 that merged 'Mesh_3-parallel-cjamin' into 'master': the circumcenters of cells were cached into cells, and thus computed only once.

CC @janetournois @cjamin

lrineau:
Fix #1554 

See details in 0f4661087bc6e97badbe0b42f28a56f73bf84a47.


cjamin:
Thanks for the fix.
Can you sum-up what was wrong and how you fixed it, please?

lrineau:
It is well explain in the log message for 0f4661087bc6e97badbe0b42f28a56f73bf84a47.

Let me know if I need to detail more.

cjamin:
Oh, perfect, I had missed it.

PR #1570, opened 2016-10-12
Qualify optional with boost::  as GCC6 sees an ambiguity

afabri:
This should fix a  'N' in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-16/Surface_mesh_simplification/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz)

### DIFF
--- a/Surface_mesh_simplification/test/Surface_mesh_simplification/test_edge_collapse_Polyhedron_3.cpp
+++ b/Surface_mesh_simplification/test/Surface_mesh_simplification/test_edge_collapse_Polyhedron_3.cpp
@@ -203,7 +203,7 @@ void write ( SurfaceSP aSurface, string aName )
 }
 
 template<class T>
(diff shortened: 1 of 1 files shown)


lrineau:
`Surface_mesh_simplification/test/Surface_mesh_simplification/basics.h` contains:
```C++
using namespace std ;
using namespace boost ;
using namespace CGAL ;
```
That explains why the call was ambiguous with [`std::optional` (C++17)][*].

[*]: http://en.cppreference.com/w/cpp/utility/optional

afabri:
I had qualified with `boost::`.  Is that good enough? You don't want me to remove the using statements?

lrineau:
That is sufficient. I just wanted to understand the initial issue (the ambiguity) and write it down somewhere for the records.

PR #1573, opened 2016-10-12
Check for the MPFR version in RS-based AK

luis4a0:
Since RS does not work with MPFR versions 3.1.4 and 3.1.5, this pull request enforces the version check using a Boost static assertion.
The bug in MPFR is solved in version 3.2 (which did not yet come to light).
Moreover, some checks are added to the AK code in order to quickly identify this kind of bugs.


lrineau:
Hi Luis, CGAL-4.8-branch is closed. I will keep only #1571, against CGAL-4.9-branch, and will merged it in `master` too, once it is tested.

Issue #17, opened 2015-03-12
Approximate Haussdorf distance between two surface meshes


lrineau:
We need a tool that computes an approximate (symmetric) Haussdorf distance between two surface meshes.

It would be useful for tests validation:
- in `Mesh_3`, we could use it to verify that the surfacic part of the output `c3t3` is a good approximation of the surface of the input domain (when the domain is composed from polyhedra).
- same with the ongoing work on `Constrained_Delaunay_triangulation_3`.

That tool could use the `AABB_tree` twice, to compute the one-way distance. Maybe we would need to compute a uniform sampling of the surfaces.

sloriot:
I do have something working and doing exactly this

lrineau:
I'll add a WG for the developers meeting. The goal will be try to adapt your code to the needs of `CDT_3` and `Mesh_3`.

pranavkantgaur:
@lrineau 
I am also implementing 3D CDT using CGAL based on paper "Meshing Piecewise Linear Complexes By Constrained Delaunay Triangulations" by Hang Si.

Can you please share the reference on which the implementation of CDT_3 is based? 
Is it possible for us to access that code right now?

lrineau:
Le Friday 13 March 2015 03:58:52 pranav a écrit :
> @lrineau 
> I am also implementing 3D CDT using CGAL based on paper "Meshing Piecewise
> Linear Complexes By Constrained Delaunay Triangulations" by Hang Si.
> 
> Can you please share the reference on which the implementation of CDT_3 is
> based?  Is it possible for us to access that code right now?

@pranav Our ongoing implementation is based on the same article, and it is not 
yet ready to be published. Please do not discuss that point in this bug. That 
is off-topic. Contact me privately if you want to discuss more about that.

-- 
Laurent Rineau, PhD
R&D Engineer at GeometryFactory           http://www.geometryfactory.com/
Release Manager of the CGAL Project       http://www.cgal.org/

pranav:
I think you've got the wrong @pranavkantgaur :)

lrineau:
@sloriot  We eventually have not talked about that at the developers meeting. But I still need the feature. Can you tell me (privately or not) in which branch and files I can see the current implementation?

sloriot:
The branch is Polygon_mesh_processing-GF and the first implementation is in Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h

[...]

PR #1578, opened 2016-10-14
Add approximated Hausdorff distance

maxGimeno:
This PR adds functions to compute the Hausdorff distance between meshes and/or point sets.
It also adds a plugin that colorizes two polyhedrons according to the local Hausdorff distances between each other.
 Small_feature [page](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_approximate_Hausdorff_distance_calculation_functions), pre-approved 30 Dec 2016.


sloriot:
Fixes #17

sloriot:
@maxGimeno  The branch has been rebased to eliminate duplicated commits.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+ *- `GRID` : points are generated in a grid, with a minimum of one point per triangle.
+ *
+ *- `MONTE_CARLO` : points are generated randomly in each triangle. Their number in each triangle is proportional to the corresponding face area with a minimum
+ *                  of 1.
+ *
+ */
+template<class TriangleMesh, class NamedParameters>
+void sample_triangle_mesh(const TriangleMesh& m,
+                          double parameter,
+                          std::vector<typename GetGeomTraits<TriangleMesh,
```
sloriot: `GetGeomTraits` cannot appear in the doc. Add `Point_3` as template parameter documenting that is should be the point of the Geom traits in the named parameters

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+  {}
+
+  void
+  operator()(const tbb::blocked_range<std::size_t>& range) const
+  {
+    Point_3 hint = initial_hint;
+    double hdist = 0;
+    for( std::size_t i = range.begin(); i != range.end(); ++i)
+    {
+      double d = CGAL::sqrt( squared_distance(hint,sample_points[i]) );
```
sloriot: replace by `sqrt` by `approximate_sqrt`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
-    approximated_Hausdorff_distance<Kernel>(triangles_1, triangles_2, targeted_precision),
-    approximated_Hausdorff_distance<Kernel>(triangles_2, triangles_1, targeted_precision)
+    approximated_Hausdorff_distance<Concurrency_tag>(tm1,tm2,precision,np1,np2),
+    approximated_Hausdorff_distance<Concurrency_tag>(tm2,tm1,precision,np2,np1)
   );
 }
 
+/**
+ * \ingroup PMP_distance_grp
+ * computes the approximated Hausdorff distance between `points` and `tm`.
```
sloriot: I don't think the distance is approximated here. @sgiraudot do you agree?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
-///\todo add approximated_Hausdorff_distance(FaceGraph, TriangleRange)
-///\todo add approximated_Hausdorff_distance(PointRange, FaceGraph)
-///\todo add approximated_Hausdorff_distance(PointRange, TriangleRange)
-///\todo add barycentric_triangle_sampling(Triangle, PointPerAreaUnit)
-///\todo add random_triangle_sampling(Triangle, PointPerAreaUnit)
-/// \todo maybe we should use a sampling using epec to avoid being too far from the sampled triangle
-  
+template< class Concurrency_tag,
+          class TriangleMesh,
+          class NamedParameters1>
```
sloriot: `NamedParameters1` ---> `NamedParameters`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+ *- `RANDOM_UNIFORM` : points are generated in a random and uniform way, depending on the area of each triangle.,
+ *
+ *- `GRID` : points are generated in a grid, with a minimum of one point per triangle.
+ *
+ *- `MONTE_CARLO` : points are generated randomly in each triangle. Their number in each triangle is proportional to the corresponding face area with a minimum
+ *                  of 1.
+ *
+ */
+template<class TriangleMesh, class NamedParameters>
+void sample_triangle_mesh(const TriangleMesh& m,
```
sloriot: You need an overload without the name parameter

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+ *
+ * \cgalNamedParamsBegin
+ *    \cgalParamBegin{vertex_point_map} the property map with the points associated to the vertices of `pmesh` \cgalParamEnd
+ *    \cgalParamBegin{geom_traits} an instance of a geometric traits class, model of `Kernel`\cgalParamEnd
+ * \cgalNamedParamsEnd
+ */
+template< class TriangleMesh,
+          class PointRange,
+          class NamedParameters>
+double max_distance_to_point_set(const TriangleMesh& tm,
```
sloriot: @sgiraudot what do you think about the name of the function. Should it also contain Hausdorff in its name?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+
+/*!
+ *\ingroup PMP_distance_grp
+ *  Computes the approximated Hausdorff distance between `tm` and `points`.
+ *
+ * @tparam PointRange a Range of `Point_3`.
+ * @tparam TriangleMesh a model of the concept `FaceListGraph` that has an internal property map
+ *         for `CGAL::vertex_point_t`
+ * @tparam NamedParameters a sequence of \ref namedparameters for `tm`
+ * @param tm the triangle mesh to compute the distance to
```
sloriot: in `tparam` and `param` sometime you put dot, sometime not. I think you can remove the dot since doxygen add them automatically (IIRC). The comments apply to the whole file.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +128,422 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+}
 
 
+template< class Concurrency_tag,
+          class TriangleMesh,
+          class NamedParameters1>
+double approximated_symmetric_Hausdorff_distance(const TriangleMesh& tm1,
+                                                 const TriangleMesh& tm2,
+                                                 double precision,
+                                                 const NamedParameters1& np1)
```
sloriot: `NamedParameters1` ---> `NamedParameters`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+    m_point = p;
+    m_hausdorff = h;
+  }
+  CPointH (const CPointH& p)
+  {
+    m_point = p ();
+    m_hausdorff = p.hausdorff ();
+  }
+
+  Point operator() () const { return m_point; }
```
sloriot: return `const Point&` ?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+  CRefTriangle (const PointH& a, const PointH& b, const PointH& c) // Triangle
+  {
+    m_point[0] = a;
+    m_point[1] = b;
+    m_point[2] = c;
+
+    m_edge = -1;
+    for (unsigned int i = 0; i < 3; ++ i)
+    {
+      if (CGAL::angle (m_point[(i+1)%3](), m_point[i](), m_point[(i+2)%3]())
```
sloriot: @sgiraudot  note that this is not a robust predicate. No idea if it is critical.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if (m_edge == -1)
+      m_bisector = CGAL::circumcenter (a(), b(), c());
+    else
+    {
+      m_bisector = PointH::mid_point (m_point[(m_edge+1)%3],
+                                      m_point[(m_edge+2)%3]);
+
+      // Point p0 = m_point[(m_edge+1)%3]();
```
sloriot: is the comment useful?

sloriot: cc @sgiraudot

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+//
+// Author(s)     : Simon Giraudot and Maxime Gimeno
+
+#ifndef CGAL_MESH_TO_POINT_SET_HAUSDORFF_DISTANCE_H
+#define CGAL_MESH_TO_POINT_SET_HAUSDORFF_DISTANCE_H
+
+#include <CGAL/Orthogonal_k_neighbor_search.h>
+#include <CGAL/Search_traits_3.h>
+#include <CGAL/squared_distance_3.h>
+
```
sloriot: the following should be in `CGAL::internal` namespace. You can even add another nested namespace specific to this file.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+  Point m_point;
+
+public:
+
+  CRefiner ()
+  {
+    m_lower_bound = 0.;
+    m_upper_bound = std::numeric_limits<FT>::max ();
+  }
+  ~CRefiner () { }
```
sloriot: default destructor should be good enough

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+private:
+
+  typedef typename Kernel::Point_3 Point;
+  typedef typename Kernel::Triangle_3 Triangle;
+  typedef typename Kernel::FT FT;
+  typedef CPointH<Kernel> PointH;
+  typedef CRefTriangle<Kernel> RefTriangle;
+  typedef CGAL::Orthogonal_k_neighbor_search<CGAL::Search_traits_3<Kernel> > Knn;
+  typedef typename Knn::Tree Tree;
+  std::priority_queue<RefTriangle> m_queue;
```
sloriot: you did not include the standard header for priority_queue

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
```diff
@@ -0,0 +1,362 @@
...
+        m_lower_bound = r.lower_bound ();
+      }
+    if (r.upper_bound () > m_lower_bound)
+      m_queue.push (r);
+    return true;
+  }
+  bool clean_up_queue ()
+  {
+    unsigned int before = m_queue.size ();
+    //    std::cerr << "Cleaned " << m_queue.size () << " elements to ";
```
sloriot: General remark: if it is not useful, a comment should be removed

Polyhedron/demo/Polyhedron/Plugins/PMP/Distance_plugin.cpp
```diff
@@ -0,0 +1,466 @@
...
+#include "triangulate_primitive.h"
+#include <CGAL/Polygon_mesh_processing/distance.h>
+#include <CGAL/Polygon_mesh_processing/compute_normal.h>
+#include <boost/iterator/counting_iterator.hpp>
+#include <CGAL/Search_traits_3.h>
+#include <CGAL/Spatial_sort_traits_adapter_3.h>
+#include <CGAL/property_map.h>
+#include <boost/container/flat_map.hpp>
+
+//This plugin crates an action in Operations that displays "Hello World" in the 'console' dockwidet.
```
sloriot: remove me

Polyhedron/demo/Polyhedron/Plugins/PMP/Distance_plugin.cpp
```diff
@@ -0,0 +1,466 @@
...
+          diagonal = this->diagonalBbox();
+        else
+          diagonal = 0.0;
+
+        //compute distance with other polyhedron
+        //sample facet
+        std::vector<Kernel::Point_3> sampled_points;
+
+//============== GRID ================
+        /*PMP::internal::triangle_grid_sampling<Kernel>(f->halfedge()->vertex()->point(), f->halfedge()->next()->vertex()->point(),
```
sloriot: comment useful?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
maxGimeno: I don't understand what you mean.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
sgiraudot: Yes it is, there is a precision to be given somewhere (at least there was in my code).

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
sloriot: Hausdorff distance is not symmetrical, but I'm fine with the names since they are rather explicit.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
sloriot: Actually this is a predicate that is filtered sorry. But the issue that remains is that the circumcenter is constructed with inexact constructions and might fall outside of its triangle. @janetournois @lrineau you probably have a solution for that when doing meshing?

janetournois: The solution we have in `Mesh_3` is to compute the circumcenter with inexact conctructions, and check whether the constructed point falls inside the circumsphere of the tetrahedron. It only prevents from using a circumcenter that is nonsense.
If the circumcenter is close to what it should be if computed exactly, we keep it anyway.

The problem here is different. You need to have consistency between the `is_inside` test and the actual position of inexactly computed circumcenter. I think you'd better
* compute the circumcenter (inexact)
* (maybe check whether it's not nonsense as we do in Mesh_3. If so, switch to exact constructions. See `Robust_circumcenter` in Mesh_3)
* check with orientation predicates whether it's inside the triangle or not

lrineau:
Le Tuesday, October 18, 2016 1:14:03 AM CEST Sebastien Loriot a écrit :
> But the issue that
> remains is that the circumcenter is constructed with inexact constructions
> and might fall outside of its triangle. @janetournois @lrineau you probably
> have a solution for that when doing meshing?

... A circumcenter of a triangle, exact or not, is outside of the triangle as 
soon as the triangle has an obtuse angle.

-- 
Laurent Rineau, PhD
R&D Engineer at GeometryFactory           http://www.geometryfactory.com/
Release Manager of the CGAL Project       http://www.cgal.org/

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
sloriot: @lrineau the obtuse angle test is already done. The pb is whether the inexact construction will fall outside of the triangle.

janetournois: what @lrineau suggests is exactly what I suggested. This way you don't rely on the pre-computed obtuse test, that could disagree with the constructed cc position

lrineau: Sorry Jane. I had not read carefully the thread. Your detailed description at 
https://github.com/CGAL/cgal/pull/1578#discussion_r83805064 contains everything.

Polygon_mesh_processing/examples/Polygon_mesh_processing/hausdorff_distance_remeshing_example.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+  Mesh m1, m2;
+  CGAL::make_tetrahedron(P_3(.0,.0,.0),
+                         P_3(2,.0,.0),
+                         P_3(1,1,1),
+                         P_3(1,.0,2),
+                         m1);
+  m2=m1;
+  PMP::isotropic_remeshing(m2.faces(),.05, m2);
+  std::cerr<< " Approximated Hausdorff distance : "<< CGAL::Polygon_mesh_processing::approximated_Hausdorff_distance<TAG>(m1,m2,4000)<<std::endl;
```
janetournois: @maxGimeno can you please try to keep lines shorter? This one does not even fit in the github screen. Thanks!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +127,457 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
-template <class Kernel, class TriangleRange1, class TriangleRange2>
+#ifdef CGAL_LINKED_WITH_TBB
+template <class AABB_tree, class Point_3>
+struct Distance_computation{
+  const AABB_tree& tree;
+  const std::vector<Point_3>& sample_points;
+  Point_3 initial_hint;
+  cpp11::atomic<double>* distance;
+
+  Distance_computation(const AABB_tree& tree, const Point_3& p, const std::vector<Point_3>& sample_points, cpp11::atomic<double>* d)
```
janetournois: same comment : this line is too long

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +127,457 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+ * @tparam NamedParameters2 a sequence of \ref namedparameters for `tm2`
+ *
+ * @param tm1 the triangle mesh that will be sampled
+ * @param tm2 the triangle mesh to compute the distance to
+ * @param precision the number of points per squared area unit for the random sampling
+ * @param np1 optional sequence of \ref namedparameters for `tm1` among the ones listed below
+ * @param np2 optional sequence of \ref namedparameters for `tm2` among the ones listed below
+ *
+ * \cgalNamedParamsBegin
+ *    \cgalParamBegin{vertex_point_map} the property map with the points associated to the vertices of `tm1` or `tm2` \cgalParamEnd
```
sloriot: Add _If this parameter is omitted, an internal property map for `CGAL::vertex_point_t` should be available in TriangleMesh_ and in all places where `vertex_point_map` is used.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +127,457 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
+
+template< class TriangleMesh,
+          class PointRange>
+double max_distance_to_point_set(const TriangleMesh& tm,
+                                 const PointRange& points,
+                                 const double precision)
+{
+  return max_distance_to_point_set(tm, points, precision, parameters::all_default());
+}
+// convenience functions with default parameters
```
sloriot: move the comment above the previous function and pu the convenience function for `max_distance_to_triangle_mesh` here too.

Polygon_mesh_processing/test/Polygon_mesh_processing/CMakeLists.txt
```diff
@@ -51,6 +51,14 @@ include_directories( BEFORE ../../include )
...
 
+find_package( TBB )
+if( TBB_FOUND )
+  include( ${TBB_USE_FILE} )
+  list( APPEND CGAL_3RD_PARTY_LIBRARIES ${TBB_LIBRARIES} )
+else()
+  message( STATUS "NOTICE: Intel TBB was not found. test_pmp_distance will use sequential code." )
+endif()
+
 if (EIGEN3_FOUND)
```
sloriot: Same remark for again here, even if not related to this PR

sloriot:
Could you comment on the modification of `triangulate_primitive.h`?

lrineau:
Is this PR ready to be tested an integrated in `master` (in case of successful tests)?

sloriot:
it is a non yet approved small feature and @sgiraudot needs to fix a robustness issue

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -0,0 +1,106 @@
+/// \ingroup PkgPolygonMeshProcessingConcepts
+/// \cgalConcept
+///
+/// The concept `PMPDistanceTraits` is a refinement of the
+/// concepts `AABBGeomTraits` and `SpatialSortingTraits_3`. In addition to the types required by
+/// those concepts,  it also requires types and functors necessary to
+/// define the functions `approximate_max_distance_to_point_set()`, `sample_triangle_mesh()`,
```
sloriot: _needed by the functions_

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -0,0 +1,106 @@
...
+/// `approximate_Hausdorff_distance()` and `max_distance_to_triangle_mesh()`
+///
+/// \cgalRefines `AABBGeomTraits` `SpatialSortingTraits_3`
```
sloriot: IIRC correctly, you can put 
```
\cgalRefines `AABBGeomTraits`
\cgalRefines `SpatialSortingTraits_3`
```
Please give it a try. If it does render nicely put _and_ between concepts refined.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -0,0 +1,106 @@
...
+    /*!
+     * is model of `Field` and `RealEmbeddable`
+     */
+    typedef unspecified_type FT;
+    /// 3D point type
+    /*! There must be a specialisation of CGAL::Kernel_traits for Point_3 and
+     * a function `Point_3 circumcenter(const Point_3&, const Point_3&, const Point_3&)`
+     */
+    typedef unspecified_type Point_3;
+    /// \name Creation
```
sloriot: _3D point type. A constructor from three `FT` must be available. There is a specialization of `CGAL::Kernel_traits` such that `CGAL::Kernel_traits<Point_3>::Kernel` is the model implementing this concept. A_

I don't think other constructors are needed.
Usage of `circumcenter()` must be replaced by `Construct_circumcenter_3`. Point_3 must be constructible from three `FT. Usage of `operator<()` should be replaced by `Less_XYZ_3`.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -0,0 +1,106 @@
...
+    */
+    Point_3(RT x, RT y, RT z);
+    /// @}
+
+    /// \name Access Functions
+    /// @{
+    /*!
+   * Returns the coordinate corresponding to `i`
+   */
+    FT operator[](int i);
```
sloriot: Really needed? IIRC only `.x()`, `.y()` and `.z()` are needed.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h
sgiraudot: I'm trying to see how I can fix this. My problem is the orientation test: as the scene is in 3D, there is no notion of left turn / right turn. I'm trying to think of a way to do this but I can't think of any that does not involve at least one inexact construction (using projections or using planes perpendicular to the triangle plane).

The test is not critical at all, It's just here so that the subdivision is well formed so I'm wondering if we could not just skip it and always use the longest edge (which is what we already do if the cirmcumcenter is outside). I'm not sure the impact on the performances would be noticeable anyway…

sgiraudot: I think it's okay to only use the longest edge, so I proposed a solution that works this way.

janetournois:
I am wondering whether the 'H' of "Hausdorff" in `approximate_hausdorff_distance()` should be capitalized or not.
Intuitively, I would use a 'h', but I found `Segment_Delaunay_graph` in CGAL which has a capital 'D' though "Delaunay" is not the first word, so my intuition may be wrong.
@sloriot, what do you think?

lrineau:
Just a remark: In Mesh_2 there are the functions `make_Delaunay_conforming` and `make_Gabriel_conforming`. At least Mariette thought that family names had to keep their capitalization in CGAL function names.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -116,67 +129,580 @@ sample_triangles(const TriangleRange& triangles, double distance, OutputIterator
...
-  typedef AABB_triangle_primitive<Kernel, Triangle_iterator> Primitive;
+    Vpm pmap = choose_param(get_param(np, vertex_point),
+                           get_const_property_map(vertex_point, tm));
+    typedef Creator_uniform_3<typename Geom_traits::FT,
+                              typename Geom_traits::Point_3> Creator;
+  switch(method)
+  {
+    case RANDOM_UNIFORM:
+    {
+      std::size_t nb_points = std::ceil( to_double(
```
afabri: @sloriot : Is there only a `static_cast` missing to  avoid a VC++ warning, or isn't there even a problem when the area is very small?

afabri:
Is distance.h the appropriate file name?  Shouldn't it be something with Hausdorff in it?

sloriot:
We might have more distance computation functions in the future, thus a generic name.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
sloriot: If the area is small, the parameter should be adapted larger. Maybe we should make it independent of the scale.
In any case, a `static_cast` is needed.

janetournois:
Let's stick to Mariette's decision about Delaunay and Gabriel, and keep the capital letter for Hausdorff too. Thanks!

lrineau:
Maxime, I would like to test the branch in `integration`. Could you rebase the branch, and fix the conflicts with `master`?

maxGimeno:
I rebased it but this is not ready to be integrated yet. It can be tested to detect early bugs but not integrated.

lrineau:
Thanks. I will remerge this PR in `integration` later this week.

lrineau:
This PR has a new conflict with `master`.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -38,6 +38,9 @@ class PMPDistanceTraits{
     /// It provides `FT operator()(const Point_3&, const Point_3&, const Point_3&) const`
     /// and `FT operator()(const Triangle_3&) const` and has `FT` as result_type.
     typedef unspecified_type Compute_squared_area_3;
+    /// Functor for computing squared length of a segment.
+    /// and `FT operator()(const Segment_3&) const` and has `FT` as result_type.
+    typedef unspecified_type Compute_squared_area_3;
```
lrineau: The two `Compute_squared_area_3` must be merged into one, with two "provides".

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPDistanceTraits.h
```diff
@@ -52,6 +55,7 @@ class PMPDistanceTraits{
     /// @name Functions
     /// @{
     Compute_squared_area_3 compute_squared_area_3_object();
+    Compute_squared_area_3 compute_squared_length_3_object();
```
lrineau: Or maybe actually that is a naming issue: should the second one be `Compute_squared_length`?

sloriot:
@maxGimeno The demo no longer compile since I remove sample_face. We should discuss how to fix it.

lrineau:
Warnings:
```
In file included from /Users/cgaltester/cgal_test/CGAL-4.10-Ic-94/cmake/platforms/x86-64_Darwin-13.0_Apple-clang-5.0_Release/test/Poisson_surface_reconstruction_3_Examples/poisson_reconstruction_example.cpp:15:
/Users/cgaltester/cgal_test/CGAL-4.10-Ic-94/include/CGAL/Polygon_mesh_processing/distance.h:482:43: warning: unused parameter 'out' [-Wunused-parameter]
                           OutputIterator out)
                                          ^
1 warning generated.
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-94/Poisson_surface_reconstruction_3_Examples/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.0_Release.gz

```
E:\CGAL\test\CGAL-4.10-Ic-94\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Debug-64bits\test\Polyhedron_Demo\Plugins\PMP\Distance_plugin.cpp(372): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data
E:\CGAL\test\CGAL-4.10-Ic-94\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Debug-64bits\test\Polyhedron_Demo\Plugins\PMP\Distance_plugin.cpp(376): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-94/Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz

lrineau:
Maxime, Sébastien, have you seen that compilation error?

https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-95/Poisson_surface_reconstruction_3_Examples/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Release-64bits.gz

```
[ 50%] [32mBuilding CXX object CMakeFiles/poisson_reconstruction_example.dir/poisson_reconstruction_example.cpp.obj[0m
poisson_reconstruction_example.cpp
C:\CGAL\test\CGAL-4.10-Ic-95\include\CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h(200) : error C2589: '(' : illegal token on right side of '::'
        C:\CGAL\test\CGAL-4.10-Ic-95\include\CGAL/Polygon_mesh_processing/internal/mesh_to_point_set_hausdorff_distance.h(197) : while compiling class template member function 'CGAL::CRefiner<Kernel>::CRefiner(void)'
        with
        [
            Kernel=Geom_traits
        ]
        C:\CGAL\test\CGAL-4.10-Ic-95\include\CGAL/Polygon_mesh_processing/distance.h(708) : see reference to function template instantiation 'CGAL::CRefiner<Kernel>::CRefiner(void)' being compiled
        with
        [
            Kernel=Geom_traits
        ]
        C:\CGAL\test\CGAL-4.10-Ic-95\include\CGAL/Polygon_mesh_processing/distance.h(708) : see reference to class template instantiation 'CGAL::CRefiner<Kernel>' being compiled
        with
        [
            Kernel=Geom_traits
        ]
        C:\CGAL\test\CGAL-4.10-Ic-95\include\CGAL/Polygon_mesh_processing/distance.h(746) : see reference to function template instantiation 'double CGAL::Polygon_mesh_processing::approximate_max_distance_to_point_set<TriangleMesh,PointRange,CGAL::pmp_bgl_named_params<T,Tag>>(const TriangleMesh &,const PointRange &,const double,const NamedParameters &)' being compiled
        with
        [
            TriangleMesh=Polyhedron,
            PointRange=PointList,
            T=bool,
            Tag=CGAL::all_default_t,
            NamedParameters=CGAL::pmp_bgl_named_params<bool,CGAL::all_default_t>
        ]
        C:\CGAL\test\CGAL-4.10-Ic-95\cmake\platforms\x64_Cygwin-Windows8_MSVC2012-Release-64bits\test\Poisson_surface_reconstruction_3_Examples\poisson_reconstruction_example.cpp(115) : see reference to function template instantiation 'double CGAL::Polygon_mesh_processing::approximate_max_distance_to_point_set<Polyhedron,PointList>(const TriangleMesh &,const PointRange &,const double)' being compiled
        with
        [
            TriangleMesh=Polyhedron,
            PointRange=PointList
        ]
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -62,121 +76,719 @@ triangle_grid_sampling(const typename Kernel::Point_3& p0,
...
+#ifndef CGAL_LINKED_WITH_TBB
+  CGAL_static_assertion_msg (!(boost::is_convertible<Concurrency_tag, Parallel_tag>::value),
+                             "Parallel_tag is enabled but TBB is unavailable.");
+#else
+  if (boost::is_convertible<Concurrency_tag,Parallel_tag>::value)
+  {
+    cpp11::atomic<double> distance(0);
+    Distance_computation<Tree, Point_3> f(tree, hint, sample_points, &distance);
+    tbb::parallel_for(tbb::blocked_range<std::size_t>(0, sample_points.size()), f);
+    return distance;
```
lrineau, line 525: @maxGimeno @sloriot This code is not the best one to compute a maximum using thread.

Currently:
  - the code uses `tbb::parallel_for`,
  - all threads compute a "local" maximum,
  - then they compare to a shared `atomic<double>` to synchronize the value.

Note that:
  - `atomic<double>` is not guaranteed to be lock-free,
  - there is a compile error with clang+C++20, in recent test-results:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-161/Polygon_mesh_processing/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz

I think the code should use [`tbb::parallel_reduce`](https://software.intel.com/en-us/node/506154): each thread returns a value, and the main thread is responsible for the reduction of all values to the final value.

PR #1579, opened 2016-10-14
Polyhedron_demo : Keep the c3t3_item's state when reloading

maxGimeno:
This PR introduces  a new interface for items that adds a virtual function, which is used to retrieve special properties from a scene_item. In this case, it is used in a c3t3_item to retrieve the spheres, grid, intersections and CNC from another c3t3_item. This allows to keep all of it in the same state when we reload a c3t3_item from file.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -40,7 +40,7 @@
 
 #include <CGAL/Three/Polyhedron_demo_plugin_interface.h>
 #include <CGAL/Three/Polyhedron_demo_io_plugin_interface.h>
-
+#include <CGAL/Three/Scene_item_with_properties.h>
 #include "ui_MainWindow.h"
 #include "ui_Preferences.h"
 #include "ui_Statistics_on_item_dialog.h"
@@ -884,6 +884,9 @@ void MainWindow::reloadItem() {
   new_item->setColor(item->color());
   new_item->setRenderingMode(item->renderingMode());
   new_item->setVisible(item->visible());
+  Scene_item_with_properties *property_item = dynamic_cast<Scene_item_with_properties*>(new_item);
+  if(property_item)
+    property_item->copyProperties(item);
   new_item->invalidateOpenGLBuffers();
   scene->replaceItem(item_index, new_item, true);
   item->deleteLater();
--- a/Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
@@ -390,6 +390,7 @@ struct Set_show_tetrahedra {
   Scene_c3t3_item_priv* priv;
   Set_show_tetrahedra(Scene_c3t3_item_priv* priv) : priv(priv) {}
   void operator()(bool b) {
+    qDebug()<<"coucou";
     priv->show_tetrahedra = b;
     priv->item->show_intersection(b);
   }
@@ -1483,11 +1484,23 @@ Scene_c3t3_item::setColor(QColor c)
 void Scene_c3t3_item::show_grid(bool b)
 {
   d->is_grid_shown = b;
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowGrid")
+    {
+      action->setChecked(b);
+      break;
+    }
   itemChanged();
 }
 void Scene_c3t3_item::show_spheres(bool b)
 {
   d->spheres_are_shown = b;
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowSpheres")
+    {
+      action->setChecked(b);
+      break;
+    }
   if(b && !d->spheres)
   {
     d->spheres = new Scene_spheres_item(this, true);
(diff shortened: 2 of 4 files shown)


Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -390,6 +390,7 @@ struct Set_show_tetrahedra {
   Scene_c3t3_item_priv* priv;
   Set_show_tetrahedra(Scene_c3t3_item_priv* priv) : priv(priv) {}
   void operator()(bool b) {
+    qDebug()<<"coucou";
```
janetournois: remove debug code

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1556,7 +1581,38 @@ void Scene_c3t3_item::setPosition(float x, float y, float z) {
...
 }
 
+bool Scene_c3t3_item::has_spheres()const { return d->spheres_are_shown;}
+
+bool Scene_c3t3_item::has_grid()const { return d->is_grid_shown;}
+
+bool Scene_c3t3_item::has_cnc()const { return d->cnc_are_shown;}
+
+bool Scene_c3t3_item::has_tets()const { if(d->intersection) return true; else return false; }
+
```
janetournois: `return d->intersection` is enough

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
lrineau: Hello ! ;-)

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -884,6 +884,9 @@ void MainWindow::reloadItem() {
   new_item->setColor(item->color());
   new_item->setRenderingMode(item->renderingMode());
   new_item->setVisible(item->visible());
+  Scene_item_with_properties *property_item = dynamic_cast<Scene_item_with_properties*>(new_item);
+  if(property_item)
+    property_item->copyProperties(item);
   new_item->invalidateOpenGLBuffers();
   scene->replaceItem(item_index, new_item, true);
```
lrineau: :+1: That's exactly what I had in mind.

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1483,11 +1484,23 @@ Scene_c3t3_item::setColor(QColor c)
 void Scene_c3t3_item::show_grid(bool b)
 {
   d->is_grid_shown = b;
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowGrid")
+    {
+      action->setChecked(b);
+      break;
+    }
   itemChanged();
```
lrineau: Use `contextMenu.findChild<QAction>("actionShowGrid")` to retrieve the action, instead of searching with a loop.

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1483,11 +1484,23 @@ Scene_c3t3_item::setColor(QColor c)
...
 }
 void Scene_c3t3_item::show_spheres(bool b)
 {
   d->spheres_are_shown = b;
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowSpheres")
+    {
+      action->setChecked(b);
+      break;
+    }
```
lrineau: Same: `findChild`

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1509,6 +1522,12 @@ void Scene_c3t3_item::show_spheres(bool b)
 }
 void Scene_c3t3_item::show_intersection(bool b)
 {
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowTets")
+    {
+      action->setChecked(b);
+      break;
+    }
```
lrineau: Use `findChild`!

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1535,6 +1554,12 @@ void Scene_c3t3_item::show_intersection(bool b)
 void Scene_c3t3_item::show_cnc(bool b)
 {
   d->cnc_are_shown = b;
+  Q_FOREACH(QAction* action, contextMenu()->actions())
+    if(action->objectName() == "actionShowCNC")
+    {
+      action->setChecked(b);
+      break;
+    }
```
lrineau: Use `findChild`.

And anyway, as soon as I use the same code pattern more than twice, I create a function. You should do the same.

Three/include/CGAL/Three/Scene_item_with_properties.h
```diff
@@ -0,0 +1,37 @@
...
+{
+namespace Three {
+  class Scene_item;
+
+//! Base class to allow an item to copy properties from another.
+class Scene_item_with_properties {
+public:
+  virtual ~Scene_item_with_properties(){}
+ //!Copies properties from another Scene_item
+ virtual void copyProperties(Scene_item*){}
```
lrineau: You can use:
```C++
  virtual void copyProperties(Scene_item*) = 0; // pure virtual method
```
That will avoid that somebody derives from `Scene_item_with_properties` and forgets to override `copyProperties`, or overrides it with the wrong signature.

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
lrineau: :+1:

Issue #1581, opened 2016-10-14
Circular_kernel_2: Intel compiler 


afabri:
We reduced the ICE of the Intel compiler to a [gist](https://gist.github.com/afabri/1985e8714ec55edc833e236bddcae0fc)

sloriot:
Fixed by #1582

PR #1582, opened 2016-10-14
Circular_kernel_2: Add a dummy field to the variant to work around an ICL bug

afabri:
This should fix Issue #1581  which shows up as error in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-22/Circular_kernel_2/TestReport_lrineau_x86-64_Linux-Fedora_IntelCompiler-17.gz)

### DIFF
--- a/Circular_kernel_2/include/CGAL/Circular_kernel_2/Intersection_traits.h
+++ b/Circular_kernel_2/include/CGAL/Circular_kernel_2/Intersection_traits.h
@@ -38,6 +38,12 @@ struct CK2_Intersection_traits
 {};
 
 // Intersection_traits for the circular kernel
+
+// The additional int in the variant has the only purpose to work around
+// a bug of the Intel compiler, which without it produces the error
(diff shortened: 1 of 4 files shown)


Circular_kernel_2/include/CGAL/Circular_kernel_2/Intersection_traits.h
```diff
@@ -44,7 +44,8 @@ struct CK2_Intersection_traits<CK, typename CK::Circle_2, typename CK::Circle_2>
   typedef typename
   boost::variant< typename CK::Circle_2,
                   typename std::pair< typename CK::Circular_arc_point_2,
-                                      unsigned int > >
+                                      unsigned int >,
+                  int >
```
lrineau: There should be a comment with an explanation, or pointing to a message on Github explaining why that fixes anything.

afabri:
The fix could also be in an ICL specific `#define`.

afabri:
It passes the testsuite now.

PR #1597, opened 2016-10-19
Polyhedron_demo : Load empty OFF files

maxGimeno:
This PR makes the demo create an empty scene_polyhedron_item if the OFF is empty, instead of crashing (in command line) or popping an error message.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
@@ -64,6 +64,7 @@ Polyhedron_demo_off_plugin::load_off(QFileInfo fileinfo) {
     {
       in.seekg(0);
       Scene_points_with_normal_item* item = new Scene_points_with_normal_item();
+      if (scanner.size_of_vertices()==0) return item;
       if(!item->read_off_point_set(in))
         {
           delete item;


Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
```diff
@@ -64,13 +64,11 @@ Polyhedron_demo_off_plugin::load_off(QFileInfo fileinfo) {
...
       Scene_points_with_normal_item* item = new Scene_points_with_normal_item();
-      if(!item->read_off_point_set(in))
-        {
-          delete item;
-          return 0;
-        }
+      if(item->read_off_point_set(in))
+      {
+        return item;
+      }
```
lrineau: And what if the call to `item->read_off_point_set(in))` return `false` ? You let the newly allocated `item` leak?

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -948,7 +948,8 @@ Scene_polyhedron_item::load(std::istream& in)
 
 
     in >> *(d->poly);
-
+    if(isEmpty)
+      return true;
```
lrineau: Compilation error in my local build!!
```
/home/lrineau/Git/other-cgal/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp: In member function ‘bool Scene_polyhedron_item::load(std::istream&)’:
/home/lrineau/Git/other-cgal/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp:951:15: error: cannot convert ‘Scene_polyhedron_item::isEmpty’ from type ‘bool (Scene_polyhedron_item::)() const’ to type ‘bool’
     if(isEmpty)
               ^
```

sloriot:
I've pushed a simpler fix

PR #1609, opened 2016-10-22
New GMPXX_arithmetic_kernel.

mglisse:
Another piece for PR #1257.


mglisse:
root_of_2.cpp now fails at runtime because some non-canonical mpq_class are produced (-1/-1 for the root), which breaks stuff (to_double returns -1 for it, and the square root of that is NaN). The CGAL way of canonicalizing an mpq_class seems to be `simplify()` (or building the number with `Compose` from `Fraction_traits`), it probably needs to be called somewhere... (maybe we can avoid setting a negative denominator in the first place)

mglisse:
Should be fixed now, although I haven't run the whole testsuite on it.

PR #1612, opened 2016-10-24
Fix the Unique_hash_map specialization of boost's associative property map

MaelRL:
This PR is a bugfix to the specialization of `boost::associative_property_map` using `CGAL::Unique_hash_map`. The latter uses four template parameters (last two having default values), but only the first two were used in the specialization, which created issues if one wished to use a `Unique_hash_map` that does not rely on the default parameters for `HashFunction` and `Allocator`.

I also included the improvements of commit f32b291, which seemed natural.

### DIFF
--- a/Hash_map/include/CGAL/Unique_hash_map.h
+++ b/Hash_map/include/CGAL/Unique_hash_map.h
@@ -129,10 +129,13 @@ namespace boost {
 
   struct lvalue_property_map_tag;
 
-  template <typename KeyType, typename ValueType>
-  class associative_property_map<CGAL::Unique_hash_map<KeyType,ValueType> >
+  template <typename KeyType, typename ValueType,
+            typename HashFunction, typename Allocator>
+  class associative_property_map<CGAL::Unique_hash_map<KeyType, ValueType,
+                                                       HashFunction, Allocator> >
   {
-    typedef CGAL::Unique_hash_map<KeyType,ValueType> C;
+    typedef CGAL::Unique_hash_map<KeyType, ValueType, HashFunction, Allocator> C;
+
   public:
     typedef KeyType key_type;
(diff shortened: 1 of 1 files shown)


lrineau:
@MaelRL @afabri Do you know where that functionality is tested?

MaelRL:
I don't think there is any existing code that uses both a unique hash map with something other than `handles` and a `boost::associative_property_map` at the same time.

There is however my code in cgal-public-dev/SM_Parameterization-BGL-GF, where it is used, for example to create UV property maps:

```
CGAL::Unique_hash_map<vertex_descriptor, Point_2, vertex_descriptor> > uvhm;
boost::associative_property_map<Unique_hash_map<vertex_descriptor, Point_2,
                                             boost::hash<vertex_descriptor> > > uvpm(uvhm);
```
If the type of the `PolygonMesh` is `Surface_mesh`, it will not compile currently.

Below are my parameterization files.
https://github.com/CGAL/cgal-public-dev/blob/5ed591184e54bec0502f61d36b35aad52d6cdc6b/Surface_mesh_parameterization/examples/Surface_mesh_parameterization/parameterization_tests.cpp

https://github.com/CGAL/cgal-public-dev/blob/5ed591184e54bec0502f61d36b35aad52d6cdc6b/Surface_mesh_parameterization/include/CGAL/ARAP_parameterizer_3.h

lrineau:
Mael, you can now remove the branch from your fork repository (If you add me as collaborator in http://github.com/MaelRL/cgal I would be able to remove the branch myself).

afabri:
But it is not really a good example. With a `Surface_mesh` it is more natural to use its property map. 
@MaelRL can you just add a small program in the test directory of `Unique_hash_map`. thanks.

PR #1530, opened 2016-09-30
Compute and fix packages dependencies

lrineau:
Fix #1483.
Fix #1484.
Fix #1485.
Fix #1492.
Fix #1494.

PR #1616, opened 2016-10-25
Do not early exit if IPE_FULL_VERSION is not there

lrineau:
If `IPE_FULL_VERSION` is missing, that leads to a CMake error in
`CMakeLists.txt` files using it.


lrineau:
That bug was introduced by 67e9da47708a084319db4dec41cab9e20c3f26cc, as a side-effect of PR #1530.

PR #1628, opened 2016-10-28
Fixed self intersection tests

MaelRL:
This PR fixes a bug in PMP's self_intersections.h: adjacent faces that were both coplanar and overlapping were not being detected.


lrineau:
Would you mind adding a new test case that failed before this PR, and succeed with this PR?

lrineau:
As that is a bug, the PR should be against `cgal/releases/CGAL-4.9-branch`. Mael, do you know how to use `git rebase ... --onto ...`?

MaelRL:
I don't know any of the typical .off examples that have this type of intersection overlapping triangles. The one I used to check was simply 4 points on xOy ( https://gist.github.com/MaelRL/f9bf50dc40aec0085f06b916cee20414 ) and verifying that running self intersection tests on other examples like cow.off and mannequin-devil.off gave the same results. Should I commit that file?

For the rebase, I should do `git rebase --onto cgal/releases/CGAL-4.9-branch HEAD~1` and push ?

lrineau:
For the rebase, that is right. Instead of `HEAD~1`, you can specify `cgal/master`:

    git rebase --onto cgal/releases/CGAL-4.9-branch cga/master

sloriot:
Your OFF is good. You can inline it in the example if you want.

lrineau:
TODO: @MaelRL unless I have not see it, you still have not added the test case to the intersection test suite.

MaelRL:
Actually, it doesn't really make sense to leave the option to pass `.off` files through `argv` and to use that `expected` value. Which do I remove?

lrineau:
If the OFF path is read from `argv`, when the Boolean `expected` should be read from `argv` as well, otherwise that cannot work.

PR #1527, opened 2016-09-30
Polyhedron demo: enhance point set

sgiraudot:
As I already discussed it with @lrineau, I would like to add the `Point_set_3` structure (currently reviewed as package for CGAL) in the Polyhedron demo as soon as possible:

* It changes _a lot_ of files in the demo, and as the demo changes very frequently, I always have tons of conflicts when refreshing the branch from master. As the feature might take weeks or months, this means that many more conflicts will probably happen later
* Although the API is still likely to change (as the feature is under review), the work needed to propagate such API changes will be much easier than the work to solve conflicts with the current API
* Even if the feature ends up not being accepted by the editorial board, it's still something that makes sense in the demo (colored point sets, better PLY reader/writer, reduced memory weight, etc.)
* Including the feature only in the Polyhedron demo is very straightforward, just 3 headers moved in the `Polyhedron/demo/Polyhedron/include` directory. These headers will just have to be removed when/if the actual package `Point_set_3` arrives in master

Note that this PR also modifies some files in `Surface_mesh` and `Point_set_processing` (for example, moving the `Properties` of Surface mesh out of the class and into its own namespace), but the public API does not change. Examples and tests of these packages were tested locally.

PR #1630, opened 2016-10-28
Polyhedron demo: fixes point set 3

sgiraudot:
The shape detection plugin now segfaults when called. This is caused by a bad merge between PR #1324 and #1527 (`add_normal_map()` must be called before setting normal values).

This PR fixes that and also fixes a problem of precision when saving PLY input (also due to PR #1527).


sgiraudot:
Errors and warnings from the testsuite do not seem related to that PR.

Issue #488, opened 2015-11-19
Deprecate Polyhedral_mesh_domain_with_features_3(std::string) constructor


gdamiand:
When using a Polyhedral_mesh_domain_with_features_3 constructor taking an off filename as parameter, there is no way to test if there was an error during the constructor (for example if the given file does not exist, or it there is an error during the reading...).

One solution could be to add exceptions in order to detect and catch the possible errors.

**Update:**

Andreas proposed that the constructor is deprecated.

afabri:
I would deprecate the constructor with the string.

sgiraudot:
What can we do to close this issue? Is deprecating the constructor with string the right solution? Who can take care of it?

lrineau:
#827 has removed the usage. We still have to deprecate the constructor, in the code and in the documentation.

afabri:
In fact, just deprecating is not really a solution. We then need a constructor with  a `std::istream&` as parameter. And in this case as well, the constructor might fail, even if we can check before that the stream is valid. Here are some options:
- A valid stream is good enough.
- The class may get a function that it is valid, which may be tested after the construction.
- The constructor might throw an exception. 
- The construction could be done by a free function.
I would go for the first option. 

_Question:_ Do we have more constructors which take an istream as parameter?

lrineau:
Fixed by #1640.

PR #1640, opened 2016-11-01
Mesh_3: Do not construct from a string

afabri:
@lrineau and @janetournois: I am not at all lucky with the `typedef` for `Polyhedron` in the example. 
With `Polyhedron_3<K>` it does not compile. Isn't that a contradiction to the sentence  "The only requirement ...." in the [ref man page](http://doc.cgal.org/latest/Mesh_3/classCGAL_1_1Polyhedral__mesh__domain__with__features__3.html#ac4fcfea3273e4a83b580386a3ea29da3) 

With `Mesh_polyhedron_3<K>` it does not compile neither, as it needs the `"::type` suffix -- looks as it was a traits class.

I thought that `Mesh_domain::Polyhedron` could be a good choice but it is not public.

### DIFF
--- a/Mesh_3/doc/Mesh_3/CGAL/Polyhedral_mesh_domain_with_features_3.h
+++ b/Mesh_3/doc/Mesh_3/CGAL/Polyhedral_mesh_domain_with_features_3.h
@@ -57,11 +57,6 @@ Its interior of `bounding_polyhedron` will be meshed.
 template <typename Polyhedron> 
 Polyhedral_mesh_domain_with_features_3(Polyhedron bounding_polyhedron); 
 
-/*!
-\deprecated Constructs a `Polyhedral_mesh_domain_with_features_3` from an off file. No feature 
-detection is done at this level. 
-*/ 
-Polyhedral_mesh_domain_with_features_3(const std::string& filename); 
 
 /*!
 Constructs a `Polyhedral_mesh_domain_with_features_3` from a polyhedral surface, and a bounding polyhedral surface.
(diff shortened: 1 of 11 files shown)


sloriot:
Fixes #488

Mesh_3/examples/Mesh_3/mesh_polyhedral_domain_with_features.cpp
```diff
@@ -32,8 +33,15 @@ using namespace CGAL::parameters;
 int main(int argc, char*argv[])
 {
   const char* fname = (argc>1)?argv[1]:"data/fandisk.off";
+  std::ifstream input(fname);
+  Polyhedron polyhedron;
+  input >> polyhedron;
+  if(input.bad()){
```
lrineau: It should be `input.fail()`. If the `failbit` is set, `fail()` returns `true`. If the `badbit` is set, `fail()` will also return `true`.

afabri: Supposing that it is the same for the other  examples I changed all `bad()` to `fail()`.

Mesh_3/doc/Mesh_3/CGAL/Polyhedral_mesh_domain_with_features_3.h
```diff
@@ -58,7 +58,7 @@ template <typename Polyhedron>
 Polyhedral_mesh_domain_with_features_3(Polyhedron bounding_polyhedron); 
 
 /*!
-Constructs a `Polyhedral_mesh_domain_with_features_3` from an off file. No feature 
+\deprecated Constructs a `Polyhedral_mesh_domain_with_features_3` from an off file. No feature
```
lrineau: I think a deprecated method:
 -  should have extra documentation saying by what it should be replaced,
 - a mention in the `changes.html`.

afabri:
@lrineau : It would be more helpful to tell me what you think the right polyhedron type is.

Issue #271, opened 2015-08-14
Use implementation-defined (or C++11) thread_local qualifiers, instead of Boost.Thread


lrineau:
Use `thread_local` (C++11), `__thread` (g++, clang, Intel C++), or `__declspec(thread)` instead of thread-local storage from Boost.Thread.

Probably more efficient, and that will reduce the dependency to pre-compiled libraries of Boost.

References:
- C++11: http://en.cppreference.com/w/cpp/language/storage_duration
- Wikipedia: https://en.wikipedia.org/wiki/Thread-local_storage#C_and_C.2B.2B

Warning: there is an issue with thread-local storage and Microsoft Windows before Vista (XP and before).

See also #273.

afabri:
I am facing a problem in Residue_type.h as VC++  does not to allow to declare a variable being thread_local and having at the same time a dll interface
https://msdn.microsoft.com/en-us/library/40a45kxx.aspx?f=255&MSPPError=-2147217396

sloriot:
I think the export is not needed in the case `CGAL_HAS_THREAD` since the variables have internal initialization and does not need to be accessed from outside.

afabri:
I am wondering if it is not as follows:
- We do not need the export when the variable is thread_local.
- But we need it, in case of single-thread   or boost::thread_specific_ptr.

lrineau:
* Exporting a variable through DLL boundaries is a design error. The use of that variable should be wrapped in a function or a class method.
* Anyway, the headers-only refactoring is probably the best solution.

sloriot:
@afabri Can we close this issue?

lrineau:
The issue about `Residue_type.h` still has no solution.
There will be discussion about that with @cjamin and @gdamiand at the meeting.

sloriot:
@cjamin @gdamiand is the remaining issue mentioned by @lrineau solved?

cjamin:
I did not really look into it, but in header-only mode, [CGAL_DLL_EXPORT is "empty"](https://github.com/CGAL/cgal/blob/master/Installation/include/CGAL/export/helpers.h) so maybe that answer the question?

gdamiand:
If I correctly understand, the problem is only on Windows ? If yes I cannot test...

sgiraudot:
Can someone using Windows test it so we know if it's solve by the header-only mode? @afabri ? @janetournois ?

janetournois:
[...]

PR #1641, opened 2016-11-01
Modular_arithmetic: Use the header only version

afabri:
This should fix Issue #271


Modular_arithmetic/include/CGAL/Modular_arithmetic/Residue_type.h
```diff
@@ -59,17 +59,13 @@ class Residue:
...
   {
     static const double CST_CUT = std::ldexp( 3., 51 );
     return CST_CUT;
   }
-#else // CGAL_HEADER_ONLY
-  CGAL_EXPORT static const double  CST_CUT; 
-  static const double& get_static_CST_CUT()
-  { return Residue::CST_CUT; }
-#endif // CGAL_HEADER_ONLY
+
```
lrineau: The initialization of `CST_CUT` is not thread-safe, but I doubt a race-condition could have any visible effect: in my opinion at worst, the variable is initialized twice with the same value.

Issue #1384, opened 2016-08-30
Static variables in Arrangement_on_surface_2 package


sloriot:
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h:953:    static unsigned int counter = 0;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_conic_traits_2.h:112:    static unsigned int index = 0;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_traits.h:1152:  static Vertex_const_handle m_empty_vtx_handle;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_traits.h:1153:  static Halfedge_const_handle m_empty_he_handle;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_dag.h:386:  static int depth;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_dag.h:416:  static int depth;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Rational_function.h:212:    static Algebraic_kernel_d_1 kernel;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Rational_function.h:213:    static Self x = Self(Polynomial_1(0), Polynomial_1(1), &kernel);`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h:348:    static Algebraic_kernel_d_1 kernel;`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h:349:    static typename Rational_function::Polynomial_1 numer(0);`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h:350:    static typename Rational_function::Polynomial_1 denom(1);`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h:353:    static Algebraic_real_1 x_coordinate =`
- [x] `Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h:356:    static Self default_instance(rational_function,x_coordinate);` 
[...]

PR #1642, opened 2016-11-01
Arrangement_2: make some variables atomic or tss

afabri:
Fix of some items of Issue #1384

- Arr_counting_traits_2.h
- Td_traits.h
- Td_dag.h


Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
```diff
@@ -950,7 +951,7 @@ class Arr_counting_traits_2 : public Base_traits {
    */
   static unsigned int increment(bool doit = true)
   {
-    static unsigned int counter = 0;
+    static CGAL::cpp11::atomic<unsigned int> counter;
```
lrineau: That will not compile if `CGAL_NO_ATOMIC` is defined (C++<11 and Boost<1.53).

Issue #1410, opened 2016-08-30
Static variables in Envelope_3 package


sloriot:
- [x] `Envelope_3/include/CGAL/Env_sphere_traits_3.h:695:        static Alg_kernel k;`

lrineau:
Fixed by #1643 by Andreas.

PR #1643, opened 2016-11-01
Envelope_3: make a static variable a data member

afabri:
@efifogel, I do not understand why the variable is static. There are other functions in the same file, where an `Alg_kernel k`  is just a local variable.

This is a fix for Issue #1410


efifogel:
Looks like a bug.
At least I cannot see any good reason for that.

For good measures, there should be a constructor that accepts an alg
kernel, and only one kernel should be constructed if one hasn't been passed
in, but this is only loosely related.

                         _/

lrineau:
Do we test and merge the PR as-is, or should I wait?

afabri:
So, I just made the variable local.

efifogel:
ok

                         _/

PR #1648, opened 2016-11-02
Added additional tests for CGAL::Unique_hash_map

MaelRL:
This PR adds new tests for the overload of `boost::associative_property_map` when the container is a `CGAL::Unique_hash_map`.

Specifically, two `CGAL::Unique_hash_map`s are used, one based on handles and the other on a type that cannot be dereferenced and thus for which the default hash function option (`Handle_hash_function`) cannot be used. Before PR https://github.com/CGAL/cgal/pull/1612, this would not have compiled due to an issue in the default parameters of the overload of `boost::associative_property_map`.

These tests are put in the file `Unique_hash_map_test.cpp` for consistency.

### DIFF
--- a/Hash_map/test/Hash_map/Unique_hash_map_test.cpp
+++ b/Hash_map/test/Hash_map/Unique_hash_map_test.cpp
@@ -6,6 +6,11 @@
 using namespace std;
 typedef list<int>::iterator Iterator;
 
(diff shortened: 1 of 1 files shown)


lrineau:
Mael, now that the PR has been merged. You can remove the corresponding branch from your repository.

Issue #1396, opened 2016-08-30
Static variables in STL_Extension package


sloriot:
- [x] `STL_Extension/include/CGAL/assertions_impl.h:47:  static Failure_behaviour _error_behaviour = THROW_EXCEPTION;`
- [x] `STL_Extension/include/CGAL/assertions_impl.h:52:  static Failure_behaviour _warning_behaviour = CONTINUE;`
- [x] `STL_Extension/include/CGAL/assertions_impl.h:125:  static Failure_function _error_handler = _standard_error_handler;`
- [x] `STL_Extension/include/CGAL/assertions_impl.h:130:  static Failure_function _warning_handler = _standard_warning_handler;`
- [x] `STL_Extension/include/CGAL/Concurrent_compact_container.h:84:    static unsigned int erase_counter(const Element &) { return 0; }`
- [x] `STL_Extension/include/CGAL/Concurrent_compact_container.h:98:    static unsigned int erase_counter(const Element &e)`
- [x] `STL_Extension/include/CGAL/Handle_with_policy.h:255:    static Char_allocator alloc;`
- [x] `STL_Extension/include/CGAL/Handle_with_policy.h:736:    static Bind bind;`
- [x] `STL_Extension/include/CGAL/Handle_with_policy.h:768:    static Rep_allocator allocator;`
- [x] `STL_Extension/include/CGAL/vector.h:180:    static`
- [x] `STL_Extension/include/CGAL/Handle_for.h:53:    static Allocator   allocator;`
- [x] `STL_Extension/include/CGAL/is_streamable.h:62:    static typename boost::remove_cv<typename boost::remove_reference<T>::type>::type const & x;`
- [x] `STL_Extension/include/CGAL/is_streamable.h:63:    static typename boost::remove_cv<typename boost::remove_reference<T>::type>::type  & y;`
- [x] `STL_Extension/include/CGAL/Spatial_lock_grid_3.h:598:    static tbb::atomic<unsigned int> last_id;`

afabri:
Nothing to do for
- STL_Extension/include/CGAL/is_streamable.h
- STL_Extension/include/CGAL/Spatial_lock_grid_3.h
- STL_Extension/include/CGAL/Concurrent_compact_container.h
so you might shorten the above list.

afabri:
@sloriot  can you please check that all items are addressed, so that it could be added to a testsuite.

sloriot:
What should I check?

afabri:
[...]

PR #1650, opened 2016-11-02
STL_extension: error/warning behavior/handler is not thread safe

afabri:
We document it, and don't try to make it thread safe.

@sloriot This should solve some, but not all items of  #1396

### DIFF
--- a/STL_Extension/doc/STL_Extension/CGAL/assertions_behaviour.h
+++ b/STL_Extension/doc/STL_Extension/CGAL/assertions_behaviour.h
@@ -29,24 +29,28 @@ typedef
 
 
 /*!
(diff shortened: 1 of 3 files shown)


mglisse:
Is CGAL::vector still needed for anything? There are only a couple uses, I didn't check if using std::vector would work without any other changes. (this remark does not block the patch of course)

afabri:
I had the same question, but when I disabled it with the appropriate `#define`,  I got a run time error somewhere, and  decided to stick for it for the time being. Shame on me.

mglisse:
Ah, thanks for doing that test. Since removing it is not easy, keeping it seems good to me :+1:.

PR #1660, opened 2016-11-04
Polyhedron_demo : Fix waitcursor for windows in some plugins

### LINKED ISSUE #1645: Polyhedron demo wait cursor
@maxGimeno : I do not see a wait cursor when I do jet smoothing on a point set, although the function `QApplication::setOverrideCursor(Qt::WaitCursor)` is called right before the call to the smoothing function. 

[Googling](http://www.informit.com/articles/article.aspx?p=1405544&seqNum=3) I found that the function `QApplication::processEvents()` might help, and it does.

This raises the following questions
- When is this needed?  Does this have to do with TBB?
- Is this only needed for VC++?  @janetournois can you reproduce this?
- Is there any drawback when we call  `QApplication::processEvents()` ?

maxGimeno:
This fixes #1645 .
 On windows, the modal dialogs mess with the application of the Waitcursor when they are called too soon after the closure of the said dialogs. This commit adds a call to processEvents() after the waitcursor where it is needed to fix that.


afabri:
Works on Windows/VC++

Issue #1408, opened 2016-08-30
Static variables in Optimal_transportaion_reconstruction_2 package


sloriot:
- [x] `Optimal_transportation_reconstruction_2/include/CGAL/OTR_2/Reconstruction_triangulation_2.h:164:    static CGAL::Random rng;`

lrineau:
Fixed by #1664.

PR #1664, opened 2016-11-07
Optimal Transport Reconstruction: make a static variable of a member function class local

afabri:
@sgiraudot  can you please check that there is no need to make the variable thread local.

This PR fixes Issue #1408


sloriot:
Any reason not to use `get_default_random()`?

afabri:
My primary goal is to make the package threadsafe

sgiraudot:
This looks good to me. The `static` variable was probably there for "easy-writting" reasons, so I think this change is fine.

PR #1665, opened 2016-11-07
Add range support for some CGAL concepts.

### LINKED ISSUE #1625: Add range support to access the ArrangementDcel
Add range support for all [_Access Functions_](http://doc.cgal.org/latest/Arrangement_on_surface_2/classArrangementDcel.html). I guess const and non-const versions should be provided.

Here is an implementation example:

```
Iterator_range<Prevent_deref<Vertex_iterator> >
vertex_handles()
{
  return make_prevent_deref_range(vertices_begin(), vertices_end());
}
```

Here is a documentation example:

```
/// returns a range over handles of the vertices in `dcel`.
unspecified_type vertex_handles();
```

Note that the return type can be documented if `Prevent_deref` is.

maxGimeno:
## Summary of Changes
Add range support for the arrangement DCEL.

 It is not ready, as we have to discuss if we want to add range support for triangulations as well.
## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #1625 
* Feature/Small Feature (if any): [here](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Range_Support_for_2D_Arrangements)

### DIFF
--- a/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
+++ b/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
@@ -209,7 +209,7 @@ returns a past-the-end iterator of the halfedges in `dcel`.
 Halfedge_const_iterator halfedges_end() const; 
 
 /*!
-returns a cosnt range over handles of the halfedges in `dcel`.
+returns a const range over handles of the halfedges in `dcel`.
 */
 unspecified_type halfedge_handles() const;
 
--- a/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
+++ b/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
@@ -179,7 +179,7 @@ returns a past-the-end iterator of the vertices in `dcel`.
 Vertex_const_iterator vertices_end() const; 
 
 /*!
-returns a const range over handles of the vertices in `dcel`.
+returns a const range (model of `ConstRange`) over handles of the vertices in `dcel`.
 */
 unspecified_type vertex_handles() const;
 
@@ -209,7 +209,7 @@ returns a past-the-end iterator of the halfedges in `dcel`.
 Halfedge_const_iterator halfedges_end() const; 
 
 /*!
-returns a const range over handles of the halfedges in `dcel`.
+returns a const range (model of `ConstRange`) over handles of the halfedges in `dcel`.
 */
 unspecified_type halfedge_handles() const;
 
@@ -239,7 +239,7 @@ returns a past-the-end iterator of the faces in `dcel`.
 Face_const_iterator faces_end() const; 
 
 /*!
-returns a const range over handles of the faces in `dcel`.
+returns a const range (model of `ConstRange`) over handles of the faces in `dcel`.
 */
 unspecified_type face_handles() const;
 
--- a/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
+++ b/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
@@ -164,6 +164,11 @@ returns a past-the-end iterator of the vertices in `dcel`.
 Vertex_iterator vertices_end(); 
 
 /*!
+returns a range over handles of the vertices in `dcel`.
+*/
+unspecified_type vertex_handles();
+
+/*!
 returns a begin-iterator of the vertices in `dcel`. 
 */ 
 Vertex_const_iterator vertices_begin() const; 
@@ -172,6 +177,11 @@ Vertex_const_iterator vertices_begin() const;
 returns a past-the-end iterator of the vertices in `dcel`. 
 */ 
 Vertex_const_iterator vertices_end() const; 
+
+/*!
+returns a const range over handles of the vertices in `dcel`.
+*/
+unspecified_type vertex_handles() const;
 
 /*!
 returns a begin-iterator of the halfedges in `dcel`. 
@@ -184,6 +194,11 @@ returns a past-the-end iterator of the halfedges in `dcel`.
 Halfedge_iterator halfedges_end(); 
 
 /*!
+returns a range over handles of the halfedges in `dcel`.
+*/
+unspecified_type halfedge_handles();
+
+/*!
 returns a begin-iterator of the halfedges in `dcel`. 
 */ 
 Halfedge_const_iterator halfedges_begin() const; 
@@ -192,6 +207,11 @@ Halfedge_const_iterator halfedges_begin() const;
 returns a past-the-end iterator of the halfedges in `dcel`. 
 */ 
 Halfedge_const_iterator halfedges_end() const; 
+
+/*!
+returns a cosnt range over handles of the halfedges in `dcel`.
+*/
+unspecified_type halfedge_handles() const;
 
 /*!
 returns a begin-iterator of the faces in `dcel`. 
@@ -204,6 +224,11 @@ returns a past-the-end iterator of the faces in `dcel`.
 Face_iterator faces_end(); 
 
 /*!
+returns a range over handles of the faces in `dcel`.
+*/
+unspecified_type faces_handles();
+
+/*!
 returns a begin-iterator of the faces in `dcel`. 
 */ 
 Face_const_iterator faces_begin() const; 
@@ -213,6 +238,10 @@ returns a past-the-end iterator of the faces in `dcel`.
 */ 
 Face_const_iterator faces_end() const; 
 
+/*!
+returns a const range over handles of the faces in `dcel`.
+*/
+unspecified_type faces_handles() const;
 
 /// @} 
 
--- a/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
+++ b/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
@@ -226,7 +226,7 @@ Face_iterator faces_end();
 /*!
 returns a range over handles of the faces in `dcel`.
 */
-unspecified_type faces_handles();
+unspecified_type face_handles();
 
 /*!
 returns a begin-iterator of the faces in `dcel`. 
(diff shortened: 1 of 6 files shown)


sloriot:
@efifogel does this look good to you?

afabri:
Wouldn't it make sense to have the ranges in a different section?

sloriot:
@afabri I don't think it's a good idea to separate it from its iterator version.

afabri:
Also, I think the functions should be called just `vertices()`, `faces()`, etc. because `begin(arr.vertices()) ==  arr.vertices_begin()`.

lrineau:
> `begin(arr.vertices()) == arr.vertices_begin()`

Actually no. Unless I am wrong, `begin(arr.vertices())` and `arr.vertices_begin()` do not have the same value types.

afabri:
Btw,  this PR is related to a [small feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Triangulation_iterators)

sloriot:
@lrineau you are wrong.

afabri:
Concerning the sections, I agree that it should stay as it is now, as there are sections such as _Accessing the Arrangement Vertices_

lrineau:
I insist: the iterators of the ranges use `prevent_deref`, their value types are `Foo_handle`, whereas the iterators of the arrangement have `Foo` as value type, and are equivalent to a `Foo_handle`.

sloriot:
@lrineau What I'm saying is that `begin(arr.vertices()) == arr.vertices_begin()`

afabri:
And @lrinaeau is right.  
But another argument for sticking to `vertices()` instead of `vertex_handles()`  is that in BGL the function `vertices(g)` returns  an iterator range.

efifogel:
What does make_prevent_deref_range(b, e) does to the b and e?

vertex_handles() gives the impression that the type of the returned
elements is Arrangement::Vertex_handle, which is not the case, so this is
another reason to use vertices() instead.

                         _/

sloriot:
@efifogel 
I'm not sure to understand what you mean. The new functions allows to do:

```
BOOST_FOREACH(Arr::Vertex_handle vh, arr.vertex_handles())
{
  std::cout << vh->point() << "\n";
}
```
or the `c++11` flavor

lrineau:
@efifogel, in https://github.com/CGAL/cgal/pull/1665#issuecomment-259363458:
> `vertex_handles()` gives the impression that the type of the returned elements is `Arrangement::Vertex_handle`, which is not the case

Yes it is! `vertex_handles()` allows to iterator on vertex handles: you can use it in a loop `for(Vertex_handle vh: arr.vertex_handles()`, or use its iterators, whose value types are `Arr::Vertex_handle`. That is the point: that range can hardly be named `vertices()`, because its value type is `Vertex_handle`.

afabri:
And the argument about consistency with moving towards  a BGL like naming scheme does not count?
To make it clear that is a change, we could have free functions too?

afabri:
And taking the confusion we just experience, this HAS to go in an extra section.

efifogel:
If you can iterate between some begin and end elements, then these elements
are not just handles, their type is not Blah_handle, and they should not be
referred to as handles. (Perhaps, they can be converted to handles.)

I still believe that vertex_handles() does not return a pair of
Arrangement::Vertex_handle.

                         _/

maxGimeno:
According to this : http://doc.cgal.org/latest/Circulator/classHandle.html
a handle is a kind of iterator, so one can perfectly iterate over handles.

lrineau:
@efifogel  https://github.com/CGAL/cgal/pull/1665#issuecomment-259381592:
> If you can iterate between some begin and end elements, then these elements are not just handles, their type is not Blah_handle, and they should not be referred to as handles. (Perhaps, they can be converted to handles.)

Now I understand you point. You have to get used to the `Range` API. You got used to `the Iterator` API, and that is similar:`vertices_begin()` does not return the first vertex, but an *iterator* to it, and for ranges that is similar.

There are already ranges in CGAL: for a 2D triangulation `tr`, `tr.tds().vertices()` is a *range* of vertices. That is an object `r` such that `begin(r)` and `end(r)` are a pair of iterators with value type `Tr::Vertex` (and not `Vertex_handle`). For a Boost graph, a for a CGAL model of `FaceGraph`, let's call it `g`, `vertices(g)` is a range that can be used to iterate on values of type `vertex_descriptor`: that means that `begin(vertices(g))` is an iterator with value type `vertex_descriptor`.

In this PR, at least the current version of the proposal and of the code, `arr.vertex_handles()` is a range of `Arr:Vertex_handle`. That means that `begin(arr.vertex_handles())` is not a `Vertex_handle`, but an iterator with `Vertex_handle` as value type.

Is that clearer now?

sloriot:
@efifogel 
I think there is a misunderstanding here.
Let's take a simpler example. `std::vector<Point_3>` is a range of points. Meaning, the range provides some iterator type that have `Point_3` as value type. Here, it is the same: `vertex_handles()` is a range of `Vertex_handle`, meaning the range type has an iterator type which value type is `Vertex_handle`.
The iterator type is `Prevent_def<Vertex_iterator>`.

efifogel:
ok, now I get it.
 (I should have looked at Sebstien's example on how it can be used more
carefully...)

In this case (despite the consistency issue Andreas has brought up) I would
go with vertex_handles().


                         _/

lrineau:
What is the status of this PR? Is its development finished?

I wonder if I should merge it in `integration`.

maxGimeno:
@sloriot Will we add support for Triangulation or other packages too in this PR ? I guess the small feature must be accepted before any merge is performed anyway.

afabri:
and it will be `Vertex_handles_around_vertex_iterator` ???   To go for `_handles`  seems a bad choice to me.

sloriot:
@afabri what are you talking about?

afabri:
There are other ranges than those over the vertices of an arrangement.  What to do with for example [`isolated_vertices_begin()`](http://doc.cgal.org/latest/Arrangement_on_surface_2/classCGAL_1_1Arrangement__2_1_1Face.html#a8f4ebf9a5868c84bf9ecea6acc2a1d15), or [`CCb_vertex_circulator`](http://doc.cgal.org/latest/Arrangement_on_surface_2/classCGAL_1_1Arrangement__2.html#ab4a594c193ca09a672dbe2fcafa090e9) ?

sloriot:
They can be wrapped too. But I suggest to first have the name scheme approved in the small feature before doing any more coding.

lrineau:
Who should create the small feature? Andreas or Maxime?
(@maxGimeno @sloriot @afabri)

maxGimeno:
Actually I already made [one](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Range_Support_for_2D_Arrangements) :

lrineau:
@maxGimeno: the status of 
https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Range_Support_for_2D_Arrangements is "not yet submitted". Have you send the feature to the CGAL develop mailing list, or is it stuck in a staged of "internal pre-review by GeometryFactory"?

maxGimeno:
I did not send anything.

lrineau:
The Travis build shows compilation errors.

Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Concepts/ArrangementDcel.h
```diff
@@ -179,7 +179,7 @@ returns a past-the-end iterator of the vertices in `dcel`.
 Vertex_const_iterator vertices_end() const; 
 
 /*!
-returns a const range over handles of the vertices in `dcel`.
+returns a const range (model of ConstRange) over handles of the vertices in `dcel`.
```
lrineau: I think that `ConstRange` should be quoted with backquotes, so that Doxygen interprets it as a type:
```
returns a const range (model of `ConstRange`) over handles of the vertices in `dcel`.
```

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1036,6 +1045,16 @@ class Arrangement_on_surface_2 {
                                   _dcel().vertices_end(),
                                   _Is_concrete_vertex(&m_topol_traits)));
   }
+
+  /*!
+  returns a const range over handles of the arrangement vertices .
```
lrineau: (model of `ConstRange`)

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1072,6 +1100,14 @@ class Arrangement_on_surface_2 {
                                     _dcel().halfedges_end(),
                                     _Is_valid_halfedge(&m_topol_traits)));
   }
+  /*!
+  returns a const range over handles of the arrangement halfedges .
```
lrineau: (model of `ConstRange`)

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1104,6 +1149,15 @@ class Arrangement_on_surface_2 {
     return (Edge_const_iterator(_dcel().edges_end(), _dcel().edges_end(),
                                 _Is_valid_halfedge(&m_topol_traits)));
   }
+
+  /*!
+  returns a const range over handles of the arrangement edges .
```
lrineau: (model of `ConstRange`)

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1137,6 +1199,14 @@ class Arrangement_on_surface_2 {
                                 _Is_valid_face(&m_topol_traits)));
   }
 
+  /*!
+  returns a const range over handles of the arrangement faces .
```
lrineau: (model of `ConstRange`)

maxGimeno:
I canceled travis as I only changed comments.

maxGimeno:
It has been two weeks that the small feature has been pre-approved.

lrineau:
Indeed. It will be tested and integrated pretty soon.

Installation/changes.html
```diff
@@ -241,7 +241,12 @@ <h3>Linear cell complex</h3>
     </li>
   </ul>
 
-  <h3> 2D Triangulation data structure </h3>
+ <h3>Range support</h3>
```
sloriot: A bit late but this should have been Arrangement...

lrineau: Commit directly in master

sloriot: It never made it to changes in 4.11. Not a big deal.

PR #1668, opened 2016-11-08
Polyhedron_demo : Fix point set selection

### LINKED ISSUE #1653: Polyhedron demo: point set selection display looks bad
This is probably tightly linked with https://github.com/CGAL/cgal/issues/1652 but I'm filling a different issue as it might be solved differently.

When loading a point set with "large" coordinates (see https://github.com/CGAL/cgal/issues/1652), the polyline used to display the selection rectangle looks bad: most of the time, it's not even a rectangle and if feels like its hitting walls and stopping at random locations. The selection, in the end, is fine and corresponds to the real rectangle the user traced (not the one displayed), as it is based on the projection on the screen, not on the polyline item. This makes the selection quite hard to use.

I guess the problem appears also if we load polylines with large coordinates.

So I can see 2 ways of correcting this bug:
- correcting https://github.com/CGAL/cgal/issues/1652 should do it (if we find a way to fix it, which I'm not sure we can)
- using 2D drawing on the window instead of a "fake" 2D rectangle drawed through a polyline object. I know this is possible as we talked about it with @maxGimeno. Of course, this would only solve the selection, not the polylines in general that can look bad with large coordinates.

maxGimeno:
This PR is a fix for #1653 
 It uses the QPainter 2D API instead of OpenGL to draw the selection lines, so there is no float-precision problems when the coordinates are large.
 It also fixes a CGAL_assertion that appears every time we use the free form selection and makes the feature impossible to use in Debug mode. 

Fix #1653.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_selection_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_selection_plugin.cpp
@@ -25,7 +25,6 @@
 #include <map>
 #include <fstream>
 
-
 // Class for visualizing selection 
 // provides mouse selection functionality
 class Q_DECL_EXPORT Scene_point_set_selection_visualizer : public CGAL::Three::Scene_item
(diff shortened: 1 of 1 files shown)


sgiraudot:
I tested it locally, no error or warning and it works well on my computer. Looks good!

PR #1674, opened 2016-11-09
Polyhedron_demo: Make it possible to switch between projection modes

maxGimeno:
This PR adds a switch in the MainWindow's View menu to make the projection matrix orthographic.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.ui
+++ b/Polyhedron/demo/Polyhedron/MainWindow.ui
@@ -95,6 +95,7 @@
     <addaction name="actionAntiAliasing"/>
     <addaction name="actionDrawTwoSides"/>
     <addaction name="actionQuickCameraMode"/>
+    <addaction name="actionSwitchProjection"/>
     <addaction name="actionMaxTextItemsDisplayed"/>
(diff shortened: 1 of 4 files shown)


sgiraudot:
For information, I was the one who asked for this feature (it can be very useful when dealing with objects were things are aligned, for example for selection).

I tested it locally, no error or warning and it works well.

afabri:
Compiles and works well for Windows/VC++.

Polyhedron/demo/Polyhedron/MainWindow.ui
```diff
@@ -676,6 +677,14 @@
     <string>Set Maximum  Text Items Displayed :</string>
    </property>
   </action>
+  <action name="actionSwitchProjection">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Orthographic/Perspective Projection</string>
```
lrineau: If the action is checkable, its name shoudl be the name of the state once check: "Use orthographic projection instead".

lrineau: Its "text", not its name.

PR #1696, opened 2016-11-16
Fixed an implementation of PMP's border_halfedges()

MaelRL:
This PR fixes a bug in one of the implementations of PMP's `border_halfedges()` and in the documentation.

There was an inconsistency between the version where a face index map is passed and the one where it is not. Although the documentation indicated that `border_halfedges()` returned interior halfedges (halfedges on the patch), this PR changes the documentation and one of the implementations to return exterior halfedges (halfedges are not on the patch). This is more consistent with the basic definition of a border halfedge.

PMP's isotropic remeshing already used the version returning exterior halfedges, so no changes are required there.

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -113,10 +113,9 @@ namespace Polygon_mesh_processing {
 
   /*!
   \ingroup PkgPolygonMeshProcessing
-  * collects the border halfedges of a surface patch
-  * defined as a face range, that is the halfedges who either do not have an
-  * incident face or whose incident face does not belong to the patch.
-  * The collected halfedges thus do not belong to the input faces.
+  * collects the border halfedges of a surface patch defined as a face range.
+  * For each returned halfedge `h`, `opposite(h, pmesh)` belongs to a face of the patch,
+  * but `face(h, pmesh)` does not belong to the patch.
   *
   * @tparam PolygonMesh model of `HalfedgeGraph`. If `PolygonMesh
   *  `has an internal property map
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -115,10 +113,10 @@ namespace Polygon_mesh_processing {
 
   /*!
   \ingroup PkgPolygonMeshProcessing
-  * collects the border of a surface patch
-  * defined as a face range. The border is "seen from inside" the patch,
-  * i.e. the collected halfedges are
-  * the ones that belong to the input faces.
+  * collects the border halfedges of a surface patch
+  * defined as a face range, that is the halfedges who either do not have an
+  * incident face or whose incident face does not belong to the patch.
```
janetournois: I don't like much this sentence. I suggest :
"... defined as a face range. For each returned halfedge `h`, `opposite(h, pmesh)` belongs to only one of the input faces."

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -132,12 +130,10 @@ namespace Polygon_mesh_processing {
...
   *
   * @param pmesh the polygon mesh to which `faces` belong
-  * @param faces the range of faces defining the patch
-  *              around which the border is collected
-  * @param out the output iterator that collects halfedges that form the border
-  *            of `faces`, seen from inside the surface patch
-  * @param np optional sequence of \ref namedparameters among
-     the ones listed below
+  * @param faces the range of faces defining the patch whose border halfedges
+  *              are collected
```
janetournois: similar to my previous remark, can you please stick to "which" instead of "whose"?
I suggest "around which border halfedges are collected" would be better

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -132,12 +130,10 @@ namespace Polygon_mesh_processing {
...
+  * @param out the output iterator that collects the border halfedges of the patch
```
janetournois: can you please add "seen from outside"? If we have been confused, I think it's not overkill to repeat...

lrineau:
Mael, you can now remove your branch.

PR #1715, opened 2016-11-23
Add an CGAL_error_msg when the sizing field if locally zero

lrineau:
Before this patch, the assertion was "sp > 0". Now the message says that
the sizing field is null at point (or at the corner), with the
coordinates of the point.


lrineau:
This PR is ready to be tested, but `integration` is already very busy at the moment, and I will postpone the testing of this PR.

PR #1718, opened 2016-11-24
Improve Lazy_alpha_nt_3

### LINKED ISSUE #1704: to_double(Lazy_alpha_nt_3)
Hello,

Lazy_alpha_nt_3 is a simplified version of Lazy_exact_nt just for alpha shapes. According to the documentation (http://doc.cgal.org/latest/Alpha_shapes_3/classCGAL_1_1Alpha__shape__3.html#a4f4f472d5f13900af9eb12d688d497e5) we can only access it through `approx()` and `exact()`. It also supports `operator<<` (undocumented). However, it does not support `to_double`, which is the usual way in CGAL to get a double from any number. It would be convenient to add this support :santa: :scroll: :gift:. There is the issue of which double to return, but that's the same issue as with Interval_nt or operator<<, so as long as we are consistent...

While I am looking at this, CGAL_LANT_COMPARE_FUNCTIONS seems suboptimal. When you compare 2 intervals, you get an `Uncertain<bool>`. You can ask `is_certain()` on it, which is much cheaper than casting to bool, possibly throwing an exception that you have to catch (ok, in case of an exception, we are going to do an expensive exact computation, but still). (On the other hand, the computation of Approx_squared_radius could be wrapped in try/catch, but apparently it does not contain any test so that's not needed.)
And Lazy_alpha_nt_3::exact_ could be an `optional<NT_exact>` so we don't allocate one when we aren't going to use it, std::vector could be boost::static_vector (or use directly `shared_ptr<const Input_point*[]>` and the size nb_pt (currently redundant), or even skip the whole shared_ptr thing, copying a static_vector of at most 4 pointers is cheaper than copying a shared_ptr), etc.

sloriot:
Fixes #1704 

@mglisse  for the version with `static_vector`, I have to define the `operator=` myself as the default one was not fine. Any idea why?

### DIFF
--- a/Alpha_shapes_3/doc/Alpha_shapes_3/Alpha_shapes_3.txt
+++ b/Alpha_shapes_3/doc/Alpha_shapes_3/Alpha_shapes_3.txt
@@ -238,7 +238,7 @@ and it must be parameterized with vertex and cell classes, which are model of th
 The package provides models `Fixed_alpha_shape_vertex_base_3<Gt>` 
 and `Fixed_alpha_shape_cell_base_3<Gt>`, respectively.
 
-\section Alpha_shapes_3AlphaShape3OrFixedAlphaShape3 Alpha_shape_3 vs.\ Fixed_alpha_shape_3
+\section Alpha_shapes_3AlphaShape3OrFixedAlphaShape3 Alpha_shape_3 vs. Fixed_alpha_shape_3
 
 The class `Alpha_shape_3<Dt,ExactAlphaComparisonTag>` represents the whole family
 of alpha shapes for a given set of points while the class `Fixed_alpha_shape_3<Dt>`
--- a/Alpha_shapes_3/doc/Alpha_shapes_3/CGAL/Alpha_shape_3.h
+++ b/Alpha_shapes_3/doc/Alpha_shapes_3/CGAL/Alpha_shape_3.h
@@ -72,7 +72,8 @@ resorting to exact arithmetic). Access to the interval containing the exact valu
 `FT::Approximate_nt approx() const` where `FT::Approximate_nt` is `Interval_nt<Protected>` 
 with `Protected=true`. Access to the exact value is provided through the function 
 `FT::Exact_nt exact() const` where `FT::Exact_nt` depends on the configuration of %CGAL 
-(it is `Gmpq` if `gmp` is available and `Quotient<CGAL::MP_Float>` otherwise). 
+(it is `Gmpq` if `gmp` is available and `Quotient<CGAL::MP_Float>` otherwise).
+An overload for the function `double to_double(FT)` is also available.
 It must be noted that an object of type `FT` is valid as long as the alpha shapes class that creates 
 it is valid and has not been modified. 
 For convenience, classical comparison operators are provided for the type `FT`. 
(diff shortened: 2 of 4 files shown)


mglisse:
What was going wrong with the default operator=? Did it not compile, or do strange things at runtime? It seems strange, since your implementation looks pretty close to what the default one should be...

sloriot:
I have many errors like this one when I remove the `operator=`:
```
/home/sloriot/CGAL/git/clean_branch/Alpha_shapes_3/test/Alpha_shapes_3/../../include/CGAL/Alpha_shape_3.h:1157:11: error: no viable overloaded '='
    alpha = squared_radius(cell_it);
    ~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~
/home/sloriot/CGAL/git/clean_branch/Alpha_shapes_3/test/Alpha_shapes_3/test_alpha_shape_3.cpp:53:22: note: in instantiation of member function 'CGAL::Alpha_shape_3<CGAL::Delaunay_triangulation_3<CGAL::Epick,
      CGAL::Triangulation_data_structure_3<CGAL::Alpha_shape_vertex_base_3<CGAL::Epick, CGAL::Default, CGAL::Boolean_tag<true>, CGAL::Boolean_tag<false> >, CGAL::Alpha_shape_cell_base_3<CGAL::Epick,
      CGAL::Default, CGAL::Boolean_tag<true>, CGAL::Boolean_tag<false> >, CGAL::Sequential_tag>, CGAL::Default, CGAL::Default>, CGAL::Boolean_tag<true> >::initialize_alpha_cell_map' requested here
template class CGAL::Alpha_shape_3<EF_Triangulation_3,CGAL::Tag_true>;
                     ^
```

I use boost 1.58, and I have the pb with both gcc and clang.

mglisse:
https://svn.boost.org/trac/boost/ticket/9932
you could also restrict the use of static_vector to versions of boost recent enough to have the fix...

By the way, the log says you replaced `shared_ptr<vector>` with `static_vector`, but you still have `shared_ptr` in there, which seems wasteful.

sloriot:
Do you mean that even with `std::vector` you would drop `shared_ptr`?

mglisse:
Ah, I misread the patch, sorry. Indeed, with std::vector you can keep shared_ptr if you like.

sloriot:
what version of boost are you using?

mglisse:
The one in Debian testing (they call it 1.62.0.1).

sloriot:
Actually the bug is still in the latest version of boost. The error goes away if I use c++11, which is what you're using I guess.

sloriot:
I replaced `static_vector` by a small struct, I think it should be OK now. @mglisse any other remark?

mglisse:
The name of the parameter `Ptr` is a bit strange since it gets instantiated with `Input_point`. Otherwise looks fine. ( https://svn.boost.org/trac/boost/ticket/12622 for the boost issue )

mglisse:
Ah, one thing I was wondering, does one of the sanitizers or valgrind complain when you copy a Lazy_alpha_nt_3 defined with less than 4 points? It shouldn't matter, but if you use those tools regularly, you might not like the noise (initializing the other pointers to 0 would likely solve any such issue).

sloriot:
I've renamed the template parameter and initialized the pointers even when not used.

PR #1720, opened 2016-11-24
Use faster Eigen implementation for diagonalization

sgiraudot:
Eigen provides [a faster implementation](https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html#afe520161701f5f585bcc4cedb8657bd1) for matrix diagonalization (if the dimension is 2x2 or 3x3 which is pretty much all that CGAL uses). This PR makes this implementation used by default.

There is indeed a noticeable gain of performance (from 4.5 to 3.0 seconds on my test computing covariance matrices on 6 nearest neighbors for each points on a sphere sampled by 10 million points).

I added a commentary and the possibility to fallback to the slower version for precision issues.

### DIFF
--- a/Solver_interface/include/CGAL/Eigen_diagonalize_traits.h
+++ b/Solver_interface/include/CGAL/Eigen_diagonalize_traits.h
@@ -24,6 +24,17 @@
 #include <Eigen/Dense>
 #include <Eigen/Eigenvalues>
 
+
+// If the matrix to diagonalize is of dimension 2x2 or 3x3, Eigen
+// provides a faster implementation using a closed-form
+// algorithm. However, it offers less precision. See:
+// https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html
+// This is usually acceptable for CGAL algorithms but one might want
+// to use the slower but more accurate version. In that case, just
+// uncomment the following line:
+
+//#define DO_NOT_USE_EIGEN_COMPUTEDIRECT_FOR_DIAGONALIZATION
+
 #include <CGAL/array.h>
 
 namespace CGAL {
@@ -67,7 +78,14 @@ private:
   diagonalize_selfadjoint_matrix (EigenMatrix& m,
 				  EigenMatrix& eigenvectors,
                                   EigenVector& eigenvalues) {
-      Eigen::SelfAdjointEigenSolver<EigenMatrix> eigensolver(m);
+      Eigen::SelfAdjointEigenSolver<EigenMatrix> eigensolver;
+
+#ifndef DO_NOT_USE_EIGEN_COMPUTEDIRECT_FOR_DIAGONALIZATION
+      if (dim == 2 || dim == 3)
+        eigensolver.computeDirect(m);
+      else
+#endif
+        eigensolver.compute(m);
 
       if (eigensolver.info() != Eigen::Success) {
           return false;


mglisse:
Eigen_diagonalize_traits doesn't document the requirements on its argument FT, is it implicitly restricted to float, double and long double? I am trying to think if the fact that computeDirect uses cos/sin/tan is an issue, but except for CORE::Expr (and maybe Interval_nt, but that would be a lot more work), I don't think we have many types where diagonalizing makes sense, and Eigen already does the wrong thing for it, so this patch seems fine from that point of view.

sgiraudot:
> Eigen_diagonalize_traits doesn't document the requirements on its argument FT, is it implicitly restricted to float, double and long double? 

I think it is. From what I understand [in the manual](https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html), Eigen supports all floating types + complex types but I don't see anything about exact types.

So this PR should not change anything about the requirement, but maybe I can use the opportunity to add a sentence about requirements in the manual?

mglisse:
https://eigen.tuxfamily.org/dox/TopicCustomizing_CustomScalar.html
Eigen supports custom types (the specialization of NumTraits is actually optional, despite what that page says), and NewKernel_d uses Eigen with Interval_nt and Gmpq everywhere.
But diagonalizing is not just any operation, it involves finding the roots of a polynomial, which makes no sense for exact types like Gmpq, and would require special syntax for CORE::Expr, so I don't expect people will be surprised that it only supports builtin types. But if you want to add a word about it, feel free.

lrineau:
There are runtime errors on Windows that might be due to this PR:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-58.shtml#Point_set_processing_3
```
cgal /srv/CGAL/www/testsuite/CGAL-4.10-Ic-58 $ zgrep -E 'ERROR|Assertion' Point_set_processing_3*/*gz
Point_set_processing_3/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Debug-64bits.gz:   ERROR:    execution   of vcm_plane_test
Point_set_processing_3/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Debug-64bits.gz:Assertion failed: points[0].second.x() < epsilon && points[0].second.x() > -epsilon, file C:\CGAL\test\CGAL-4.10-Ic-58\cmake\platforms\x64_Cygwin-Windows8_MSVC2012-Debug-64bits\test\Point_set_processing_3\vcm_plane_test.cpp, line 35
Point_set_processing_3/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Release-64bits.gz:   ERROR:    execution   of vcm_plane_test
Point_set_processing_3/TestReport_afabri_x64_Cygwin-Windows8_MSVC2012-Release-64bits.gz:Assertion failed: points[0].second.x() < epsilon && points[0].second.x() > -epsilon, file C:\CGAL\test\CGAL-4.10-Ic-58\cmake\platforms\x64_Cygwin-Windows8_MSVC2012-Release-64bits\test\Point_set_processing_3\vcm_plane_test.cpp, line 35
Point_set_processing_3/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:   ERROR:    execution   of vcm_plane_test
Point_set_processing_3/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:Assertion failed: points[0].second.x() < epsilon && points[0].second.x() > -epsilon, file E:\CGAL\test\CGAL-4.10-Ic-58\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Debug-64bits\test\Point_set_processing_3\vcm_plane_test.cpp, line 35
Point_set_processing_3/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz:   ERROR:    execution   of vcm_plane_test
Point_set_processing_3/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz:Assertion failed: points[0].second.x() < epsilon && points[0].second.x() > -epsilon, file E:\CGAL\test\CGAL-4.10-Ic-58\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Release-64bits\test\Point_set_processing_3\vcm_plane_test.cpp, line 35
```

sgiraudot:
I think I can safely confirm that this is due do this PR. The result is supposed to be `Vector_3(0,0,1)` with a precision of `epsilon=1e-5` but here we have `Vector_3(0.00219548, 0.000622892, -0.999997)` so the precision is not good enough.

The loss of precision is known as I pointed before but I did not think it would be sufficiently big to make such test fail… that might be an argument against integrating this change, or at least for making it non-default. What do you think?

sgiraudot:
From what I see, the errors are gone, probably fixed by this PR: https://github.com/CGAL/cgal/pull/1731
(Problem was in VCM.)

sgiraudot:
Testsuite looks good.

PR #1730, opened 2016-12-01
Fixed a compilation issue in PMP's border_halfedges()

MaelRL:
This PR fixes a compilation error in `border_halfedges()` of `PMP`, which appeared if a face index map was not provided in named parameters, nor as an internal property map.

Note that it still requires the Polygon mesh to have a dummy that matches `boost::property_map<PolygonMesh, boost::face_index_t>` otherwise the `property_map_selector` will not compile.

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -80,11 +80,13 @@ namespace Polygon_mesh_processing {
 
     template<typename PM
            , typename FaceRange
-           , typename HalfedgeOutputIterator>
+           , typename HalfedgeOutputIterator
+           , typename NamedParameters>
     HalfedgeOutputIterator border_halfedges_impl(const FaceRange& faces
                                                , typename boost::cgal_no_property::type
                                                , HalfedgeOutputIterator out
-                                               , const PM& pmesh)
+                                               , const PM& pmesh
+                                               , const NamedParameters& /* np */)
     {
       return border_halfedges_impl(faces, out, pmesh);
     }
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -80,27 +80,45 @@ namespace Polygon_mesh_processing {
...
       typedef typename boost::graph_traits<PM>::halfedge_descriptor halfedge_descriptor;
       typedef typename boost::graph_traits<PM>::face_descriptor     face_descriptor;
 
+      //make a minimal check that it's properly initialized :
+      //if the 2 first faces have the same id, we know the property map is not initialized
+      if (boost::is_same<typename GetFaceIndexMap<PM, NamedParameters>::Is_internal_map,
+                         boost::true_type>::value)
+      {
+        typename boost::range_iterator<const FaceRange>::type it = boost::const_begin(faces);
+        if (get(fmap, *it) == get(fmap, *cpp11::next(it)))
```
lrineau: This will segfault if the `pmesh` is empty or have only one face.

MaelRL: There are already two checks in the main function `border_halfedges()`:
` if (faces.empty()) return out;`
and
` if (boost::is_same<FIMap, Unset_FIMap>::value || faces.size() == 1)`
who should handle those cases. Do I repeat them there ?

lrineau: No, that is fine. I had not seen that it was checked in the main function.

PR #1731, opened 2016-12-01
Point_set_processing: define CGAL_VORONOI_COVARIANCE_USE_CONSTRUCTIONS

afabri:
Use halfspace intersection that produces better vertex coordinates (even if the topology of the CH3 might not be fully correct).

halfspace_intersections()  for a range of planes has a robustness problem when `intersection()` is called with three planes.  To `#define CGAL_VORONOI_COVARIANCE_USE_CONSTRUCTIONS` is a quick solution, and the problem wih the intersection must be investigated later.


lrineau:
Andreas, is there an issue for the "the problem wih the intersection must be investigated later"? As far as I know the CGAL-develop mailing-list has not received any mail on 1st December from you.

PR #1737, opened 2016-12-02
Polyhedron_demo: Fix display bugs when the scene is greatly offsetted

### LINKED ISSUE #1652: Polyhedron demo: bad display when coordinates are large
This is a known problem but I'm filling this issue so that we can keep track of it.

When loading an item that has "large" coordinates (what is the "large" limit is a gray area to me), the display looks bad: for example, with a point set, points flicker when moving the objects and look falsy "aligned" on a grid.

The solution I use for now is to recenter the object, but that's obviously not ideal: one may want to work on an item without changing its coordinates (if it needs to be saved and used with other registered items, it's not possible). Could there be a way to, for example, only "virtually" move if from the OpenGl point of view while keeping the real coordinates in memory?

(cc @maxGimeno @lrineau)

maxGimeno:
This PR fixes #1652 . 
 When the scene Bbox is centered far from (0,0,0) (if log2(coord)>13), the viewer gets an offset, and every item will be displayed by this offset, so it is re-centered closer to (0,0,0). This only affects the display and the item's data remains unchanged. 
This will not change the relative distance between items, so let's say if an item is centered in (200000, 0, 0) and another around (0, 200000,0), the scene will remain unreadable because of the distance between the items.
However, when a single item is drawn, this will fix the display problem making it "shaking".

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1,3 +1,5 @@
+#include <cmath>
+
 #include "config.h"
 #include "MainWindow.h"
 #include "Scene.h"
@@ -863,7 +865,7 @@ void MainWindow::updateViewerBBox()
   qglviewer::Vec offset(0,0,0);
   for(int i=0; i<3; ++i)
   {
-    if(log2(bbox.min(i)) > 13.0 )
+    if(std::log2(bbox.min(i)) > 13.0 )
     {
       offset[i] = -bbox.min(i);
     }
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -864,9 +864,9 @@ void MainWindow::updateViewerBBox()
   bool new_offset = false;
   for(int i=0; i<3; ++i)
   {
-    if(log2(bbox.min(i)) > 13.0 )
+    //if(log2(bbox.min(i)) > 13.0 )
     {
-      offset[i] = -bbox.min(i);
+      offset[i] = -200;//-bbox.min(i);
       new_offset = true;
     }
   }
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -864,7 +864,7 @@ void MainWindow::updateViewerBBox()
   bool new_offset = false;
   for(int i=0; i<3; ++i)
   {
-    //if(log2(bbox.min(i)) > 13.0 )
+    if(log2(bbox.min(i)) > 13.0 )
     {
       offset[i] = -bbox.min(i);
       new_offset = true;
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -866,7 +866,7 @@ void MainWindow::updateViewerBBox()
   {
     //if(log2(bbox.min(i)) > 13.0 )
     {
-      offset[i] = -200;//-bbox.min(i);
+      offset[i] = -bbox.min(i);
       new_offset = true;
     }
   }
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -860,6 +860,14 @@ void MainWindow::updateViewerBBox()
   qglviewer::Vec 
     vec_min(xmin, ymin, zmin),
     vec_max(xmax, ymax, zmax);
+  qglviewer::Vec offset(0,0,0);
+  for(int i=0; i<3; ++i)
+  {
+    if(log2(bbox.min(i)) > 13.0 )
+      offset[i] = -bbox.min(i);
+  }
+  viewer->setOffset(offset);
+  std::cerr<<offset.x<<", "<<offset.y<<", "<<offset.z<<std::endl;
   viewer->setSceneBoundingBox(vec_min,
                               vec_max);
   viewer->camera()->showEntireScene();
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -861,13 +861,20 @@ void MainWindow::updateViewerBBox()
     vec_min(xmin, ymin, zmin),
     vec_max(xmax, ymax, zmax);
   qglviewer::Vec offset(0,0,0);
+  bool new_offset = false;
   for(int i=0; i<3; ++i)
   {
-   // if(log2(bbox.min(i)) > 13.0 )
+    if(log2(bbox.min(i)) > 13.0 )
+    {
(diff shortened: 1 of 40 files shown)


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -860,6 +860,21 @@ void MainWindow::updateViewerBBox()
   qglviewer::Vec 
     vec_min(xmin, ymin, zmin),
     vec_max(xmax, ymax, zmax);
+  qglviewer::Vec offset(0,0,0);
+  for(int i=0; i<3; ++i)
+  {
+    if(log2(bbox.min(i)) > 13.0 )
```
lrineau: This line does not compile on Windows: you must use [`std::log2`] and include `<cmath>`.

[`std::log2`]: http://en.cppreference.com/w/cpp/numeric/math/log2

lrineau:
I have removed 598f5d5 from this PR with a forced push. See #1804.

lrineau:
I have tested on Linux (Fedora 24, NVIDIA graphics drivers), and it seems fine (at least for the items Polyhedron, C3t3, spheres, polylines).

lrineau:
@CGAL/geometryfactory Can you please test the polyhedron demo from this branch, with your favorite items? At least we would need to test:
  - point set related items,
  - the selection plugin,
  - ... on Windows.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -860,6 +862,21 @@ void MainWindow::updateViewerBBox()
...
+  {
+    if(std::log2(bbox.min(i)) > 13.0 )
+    {
+      offset[i] = -bbox.min(i);
+    }
+  }
+  if(offset != viewer->offset())
+  {
+    viewer->setOffset(offset);
+    recomputeItems();
```
lrineau: The strategy is strange. Instead of `std::log2(bbox.min(i))`, I would test something like the "distance" (l-infiinity distance) between the center of the bbox and the offset, something like:
```C++
max(abs(center.x() + offset.x()),
    max(abs(center.y() + offset.y()),
        abs(center.z() + oftset.z())))
```
and if that distance log2 is greater than 13 (half of the precision of `float`), then change the offset to the center of the bbox.

Using the `min` of the bbox as the offset is not the best choice, because it means that the sign-bit of the `float`s will never be used, and that wastes one bit of precision for no reason.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -869,15 +886,13 @@ void MainWindow::reloadItem() {
   QAction* sender_action = qobject_cast<QAction*>(sender());
   if(!sender_action) return;
   
-  bool ok;
-  int item_index = sender_action->data().toInt(&ok);
-  QObject* item_object = scene->item(item_index);
-  if(!ok || !item_object || sender_action->data().type() != QVariant::Int) {
+  Scene_item* item = (Scene_item*)sender_action->data().value<void*>();
+  QObject* item_object = item;
+  if(!item || !item_object) {
```
lrineau: I do not understand!

1. `item_object` seems useless: it cannot be zero if `item` is non-zero.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1933,9 +1942,18 @@ void MainWindow::resetHeader()
...
   sceneView->header()->resizeSection(Scene::ABColumn, sceneView->header()->fontMetrics().width(QString("_AB_")));
   sceneView->header()->resizeSection(Scene::VisibleColumn, sceneView->header()->fontMetrics().width(QString("_View_")));
+
+}
+void MainWindow::recomputeItems()
+{
+  for(int i=0; i<scene->numberOfEntries(); ++i)
+  {
+    scene->item(i)->invalidateOpenGLBuffers();
+    scene->item(i)->itemChanged();
```
lrineau: Why do we need to call the two functions?

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -354,6 +354,7 @@ protected Q_SLOTS:
   void updateMenus();
   void load_plugin(QString names, bool blacklisted);
   void recurseExpand(QModelIndex index);
+  void recomputeItems();
```
lrineau: `recomputeItems` is called only once. Just remove it and inline its body in the caller. That will avoid the recompilation of all plugins (only items will be recompiled).

afabri:
Where do we find offsetted data sets of the various types?

lrineau:
I used geological meshes from one of our user. Maybe we could extract point sets from such data. And Surface⁣_mesh, and polygon soups. 

Envoyé par BlueMail ​

Le 11 janv. 2017 08:01, à 08:01, Andreas Fabri <notifications@github.com> a écrit:

sgiraudot:
Tested on my computer:
- loading a greatly offsetted PLY point set
- smoothing it
- reconstructing it with advancing front

Everything looks good, the point set before and after smoothing and the Polyhedron create too. No more weird display :)

maxGimeno:
@afabri you can ;ake an offset for a polyhedron or a point_set with the affine_transformation_plugin

maxGimeno:
@lrineau 

> All functions that call
> 
> static_cast<CGAL::Three::Viewer_interface*>(QGLViewer::QGLViewerPool().first())
> should have the viewer directly as parameter.

That doesn't seem possible to me. Most (not to say all, as I didn't check every case) of those functions are called from places that have no access to a viewer. For some functions it would be possible but that would mean change a significative part of the rendering pipeline.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -857,19 +857,23 @@ void MainWindow::updateViewerBBox()
...
+    vec_max(xmax, ymax, zmax),
+    bbox_center((xmin+xmax)/2, (ymin+ymax)/2, (zmin+zmax)/2);
   qglviewer::Vec offset(0,0,0);
-  for(int i=0; i<3; ++i)
-  {
-    if(std::log2(bbox.min(i)) > 13.0 )
+  double l_dist = (std::max)((std::abs)(bbox_center.x + viewer->offset().x),
+                      (std::max)((std::abs)(bbox_center.y + viewer->offset().y),
+                          (std::abs)(bbox_center.z + viewer->offset().z)));
+  if((std::log2)(l_dist) > 13.0 )
```
lrineau: **Note for next time** (no need to change the commit): only `min` and `max` have to be protected by a pair of parenthesis (that prevents the macro substitution). The reason are those two lines in `<stdlib.h>` in the MSVC standard library:
```C++
#define max(a,b)    (((a) > (b)) ? (a) : (b))
#define min(a,b)    (((a) < (b)) ? (a) : (b))
```

Without the protection against substitution of `max`, `std::max(x, y-1)`, for example, is turned into:
```C++
std::   (((x) > (y-1)) ? (x) : (y-1))
```
which leads to a parsing error.

No need to protect `abs` or `log2`, or anything else but `min` and `max`, as far as we known.

PR #1741, opened 2016-12-02
fix Construct_projected_point_3 with degenerate triangles

lrineau:
`Construct_projected_point_3::operator()(Point_3, Triangle_3)` now
checks if the triangle is degenerated (actually it checks if its
supporting plane is degenerated). If it is, it falls back to
`operator()(Point_3, Segment_3)` with the segment that is equivalent to
the triangle. When using non-exact floating point numbers, it chooses
the edge of the triangle that has the biggest l-infinity length.

### DIFF
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -2215,9 +2215,52 @@ namespace CommonKernelFunctors {
         k.construct_supporting_plane_3_object();
       typename K::Construct_projected_point_3 projection =
         k.construct_projected_point_3_object();
+      typename K::Is_degenerate_3 is_degenerate = k.is_degenerate_3_object();
+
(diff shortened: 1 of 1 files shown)


lrineau:
@sloriot, @janetournois:  If I remember well, you did move the code from AABB tree to the kernel. Can you review this patch, please?

sloriot:
@lrineau did you handled the case of a degenerate segment in the other `operator()`?

lrineau:
@sloriot, https://github.com/CGAL/cgal/pull/1741#issuecomment-264444252:
> is the case of a degenerate segment handled in the other `operator()`?

I saw that the `operator()(Point_3 p , Segment_3 seg)` already had a test that checks if the segment `seg` is degenerate. That is why the fallback from `Triangle_3` to `Segment_3` also works if the triangle is degenerate to a point.

Issue #1723, opened 2016-11-25
Fuzzy_sphere::outer_range_contains always true for radius 0


mglisse:
I noticed accidentally that, for a Fuzzy_sphere where radius=epsilon=0, outer_range_contains does not even look at the coordinates: distance and squared_radius are 0, the main loop is

    for (int i = 0; cit != end && (distance < squared_radius)...

so it exits immediately, and we end up with

    return (distance <= squared_radius);

meaning that this point contains all rectangles... This yields strange results in Kd_tree::search. I don't know if we want to support a radius of 0, but if we don't, it would be nice to document it and add a precondition check.

afabri:
You mean, you want at least find the point which has the coordinate of the center of the sphere?

mglisse:
1) Indeed I would like to find the point which coincides with the center of the sphere. I can understand if you don't want to support it (maybe Fuzzy_sphere represents an open ball or something).
2) Most importantly, I do not want to find points which are far away, which is what CGAL currently returns...

afabri:
Thank  you for the bug report and fix.  Concerning point 1) above, it is the function `Fuzzy_sphere::contains(Point)` which at the end decides if a point is inside it or not.

lrineau:
Fixed by #1743.

PR #1743, opened 2016-12-04

afabri:
When the query is a radius 0 sphere, we never enter the loop with ` distance < squared_radius`, and hence the real ` distance` does not get computed.

This fixes Issue  #1723


afabri:
@lrineau  The spatial searching row is competely green, so  it could be merged in the 4.9 bug fix branch.

PR #1745, opened 2016-12-05
Add corefinement operations in PMP

sloriot:
Add surface intersection, corefinement and boolean operations using corefinement.

see the small feature [page](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Corefinement_in_PMP) for details


sloriot:
`test_merge_of_branch` reported no error.
Branch weight is 1.3Mo

lrineau:
I will be tested tonight, in 4.10-Ic-68.

Number_types/include/CGAL/Interval_nt.h
```diff
@@ -945,7 +945,7 @@ namespace INTERN_INTERVAL_NT {
       return approx.pair();
     }
     else
-      return std::pair<double,double>(l,l);
+      return std::pair<double,double>((double)l,(double)l);
```
lrineau: Please use `double(l)` or `static_cast<double>(l)`.

The C-style cast is an abomination without any checks, that allows to cast even carrots into space ships.

Operations_on_polyhedra/include/CGAL/internal/Polyhedron_plane_clipping_3.h
```diff
@@ -131,7 +131,7 @@ Polyhedron clip_to_bbox(const Bbox_3& bbox, const Plane_3& plane)
     if (orientations[i2]==ON_ORIENTED_BOUNDARY) continue;
     if (orientations[i1]!=orientations[i2])
       intersection_points.push_back(
-        get<Point_3>(
+        boost::get<Point_3>(
```
lrineau: Adding the qualification is a bad fix. You no longer use ADL is the call is qualified.

Read this StackOverflow question and best answer:  http://stackoverflow.com/a/2953783/1728537

The best solution is `using boost::get; boost::get<Point_3>(...)`. That way, even if we switch to `std::optional`, the ADL call will remain correct. The `using` is there so that the C++ parser finds at least one `get` function. But ADL will allow to find better matches during the lookup.

sloriot: That's a work around for a compiler bug. The argument of get is a `boost::variant` so ADL should work but it doesn't.

lrineau: Go and read the SO question and answer. The compilation error has nothing to with ADL, but with the parsing of the expression `get<Point_3>`: even before the name lookup rules are used (including ADL), the expression must be parse as a correct C++ expression. If the compiler has no "view", as regards the current namespace/scope, of a correct `get` template function, then the parsing trigger a compilation error.

You can even use `using CGAL::get` (if it is defined in the included headers) even if in the end the argument-dependent lookup will conclude that `boost::get` is the right match.

lrineau:
This PR weights around 1.5MB.

sloriot:
Except from an [ICE](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-83/Polygon_mesh_processing_Examples/TestReport_afabri_x64_Cygwin-Windows8_MSVC2013-Release-32bits.gz) tests were good.

sloriot:
I've reduce the size of the branch to 764K by removing an unused png and replacing a png by a jpg.

sloriot:
@kintel You might be interested by this new official method for doing fast boolean operations of closed manifold surfaces.

Here is the user manual documentation:
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.10-Ic-100/output/Polygon_mesh_processing/index.html#title12

PR #1747, opened 2016-12-05
Surface_mesh:  The get/put functions are not templated

afabri:
This should fix  the testcases with an ICE for Visual 2013


Polygon_mesh_processing/test/Polygon_mesh_processing/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ if (EIGEN3_FOUND)
   create_single_source_cgal_program("measures_test.cpp")
   create_single_source_cgal_program("triangulate_faces_test.cpp")
   create_single_source_cgal_program("test_pmp_remove_border_edge.cpp")
+  create_single_source_cgal_program("test.cpp")
```
sloriot: the name of the file is too generic and is actually not commited. I guess you did not want to commit that change.

lrineau:
The testsuite CGAL-4.10-Ic-67 seemed fine. It that sufficient to merge this PR? I wonder why I had put the label "not yet approved".

@sloriot Do you have any other comment on the change of this PR?

PR #1748, opened 2016-12-06
Polyhedron_demo: Enhancements for Io_image_plugin

### LINKED ISSUE #1721: VolumePlaneControl  feature request for Polyhedron demo
The planes and the intersection lines, as well as the sliders should be colored as shown. 
I would like to have handles on the sides of the frames, so that I can slide them directly.

volumeplanecontrol

maxGimeno:
This PR fixes #1721 by adding colors and handles for the Volume planes. 
It also fixes the fluidity problems when handling the planes from the ManipulatedFrame interface and destroys the planes of an image when the said image is destroyed, to avoid an unstable state of the sliders when there is several images open at once.

@afabri and @sloriot, can you test it and say if you are satisfied with the modifications before it is integrated ? 
@lrineau This PR needs #1737 about the offseted items.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Volume_plane.h
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Volume_plane.h
@@ -37,10 +37,10 @@ public:
   void constrainTranslation(qglviewer::Vec& t, qglviewer::Frame* const frame) {
     WorldConstraint::constrainTranslation(t, frame);
     qglviewer::Vec pos = frame->position();
-    double start = pos[Dim];
+    const qglviewer::Vec offset = static_cast<CGAL::Three::Viewer_interface*>(QGLViewer::QGLViewerPool().first())->offset();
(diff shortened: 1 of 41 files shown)


sloriot:
This is a great improvement for moving planes.

afabri:
Works well. For me it could get integrated into master.

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Volume_plane.h
```diff
@@ -219,7 +232,7 @@ class Volume_plane : public Volume_plane_interface, public Tag {
 
   void drawSpheres(x_tag) const
   {
-      int max_dim = (std::max)((std::max)(adim_, bdim_ ), cdim_);
+      update_maxDim();
```
lrineau: I do not understand why `max_dim` become a mutable member data, instead writing;
```C++
double max_min = compute_max_dim();
```

PR #1749, opened 2016-12-06
Barycentric coordinate -> Barycentric coordinates

sloriot:
As reported privately by @efifogel , coordinate should be plurial.

The rename comes with a fallback for deprecated code (I locally tested that example of 4.9 and previous compiles with updated headers).

CC @Hippocrates do you agree?

### DIFF
--- a/Surface_mesh_shortest_path/doc/Surface_mesh_shortest_path/Concepts/SurfaceMeshShortestPathTraits.h
+++ b/Surface_mesh_shortest_path/doc/Surface_mesh_shortest_path/Concepts/SurfaceMeshShortestPathTraits.h
@@ -43,7 +43,7 @@ public:
   typedef unspecified_type Triangle_2;
 
   /// An ordered triple to specify barycentric coordinates in triangles.
-  typedef unspecified_type Barycentric_coordinate;
+  typedef unspecified_type Barycentric_coordinates;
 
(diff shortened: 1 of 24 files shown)


Surface_mesh_shortest_path/doc/Surface_mesh_shortest_path/Concepts/SurfaceMeshShortestPathTraits.h
```diff
@@ -247,31 +247,31 @@ class SurfaceMeshShortestPathTraits
 
   /*!
   Function object type that provides
-  `Barycentric_coordinate operator()(FT a, FT b, FT c)`
-  to introduce a new triangular barycentric coordinate.
+  `Barycentric_coordinates operator()(FT a, FT b, FT c)`
+  to introduce some new triangular barycentric coordinates.
```
Hippocrates: Random note: The use of the term `some` feels a little weird to me. Its technically correct, but it feels oddly informal. You might consider `to construct new triangular barycentric coordinates` instead, but its just my gut reaction, I don't have a specific reason to prefer it (and its definitely not a big deal).

Hippocrates:
In any case, this seems fine to me. You are correct that the pluralization is a more correct way to express this (since the elements of the tuple are themselves each one coordinate).

PR #1751, opened 2016-12-06
Fix doc testsuite

sloriot:


lrineau:
Tested in CGAl-4.10-Ic-80, documentation-only.

PR #1752, opened 2016-12-07
Fix the patch from #1524

lrineau:
The main goal of this PR is to fix a bug introduced by #1524 (for CGAL-4.9.1). This PR also improves or adds minor fixes:

  - in `polylines_to_protect`, change the split angle from 135° to 90°,

  - improve debug messages,

  - fix `CGAL_MESH_3_PROTECTION_DEBUG` with a mesh domain where the
    `Surface_patch_index` is `pair<int, int>`.

### DIFF
--- a/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
+++ b/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
@@ -62,10 +62,32 @@ public:
 
   typedef typename boost::graph_traits<G_copy>::vertex_descriptor
     g_copy_vertex_descriptor;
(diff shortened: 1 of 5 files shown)


lrineau:
There a reproducible runtime error, in Debug, of `examples/Mesh_3/mesh_3D_image_with_features.cpp`.
```
Expr: degree(v, graph) == 1
File: /home/lrineau/Git/other-cgal/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
Line: 130
```

See https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-70.shtml#Mesh_3_Examples

PR #1755, opened 2016-12-09
Allow to load OBJ files containing polygon soups

lrineau:
If the content of an OBJ file is not a correct polyhedron, but a polygon soups, then the current code cannot load the file.

This PR allows to load the file anyway, producing a polyhedron with self-intersections.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
@@ -3,10 +3,13 @@
 #include "Scene_points_with_normal_item.h"
 #include "Polyhedron_type.h"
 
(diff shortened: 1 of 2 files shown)


sloriot:
The logic for OFF file is to create a polygon soup if the input is not a valid polyhedron. Any reason not to do the same here?

lrineau:
@sloriot With c94b2f3309a9f907d65fac594e1cff1b6cb323b3, I have moved the code that parse OBJ files in the OFF plugin, that will now create either a polyhedron or a polygon soup.

lrineau:
Note: I have kept the code of `Scene_polyhedron_item::load_obj` by compatibility. I could also revert by first commit 526bee6094ac2a38b54d077ea555b736a9b07c9d for a full compatibility.

PR #1759, opened 2016-12-13
Use correct Exponent type

xantares:


Number_types/include/CGAL/GMP/Gmpzf_type.h
```diff
@@ -422,7 +422,7 @@ std::pair<std::pair<double, double>, long> Gmpzf::to_interval_exp() const
   // get surrounding interval of the form [l * 2 ^ k, u * 2^ k]
   // first get mantissa in the form l*2^k, with 0.5 <= d < 1;
   // truncation is guaranteed to go towards zero
-  long k = 0;
+  Exponent k = 0;
   double l = mpz_get_d_2exp (&k, man());
```
mglisse: Well, the correct type for k is whatever `mpz_get_d_2exp` is documented to take as first argument (minus the pointer). According to the GMP documentation, `long` is right, and the reverse patch in function `to_double` would make sense.

xantares: it's for mpir, yes, it helps if it's a typedef

mglisse: https://github.com/wbhart/mpir/blob/master/gmp-h.in#L918 seems to say that `mpz_get_d_2exp` also expects a `long` in MPIR now. If we change things to help older versions of MPIR, that's going to break future ones...

mglisse:
I am ok with this patch for consistency reasons. If some day we bother to improve things, we can change the 3 instances at the same time.

xantares:
exactly

PR #1760, opened 2016-12-13
Use mp_exp_t for Exponent

xantares:


lrineau:
This PR seems related to the previous one, #1759, is not it?

xantares:
I'm sure about the previous one, but for this one I don't really know what I'm doing.

lrineau:
@mglisse Can you have a look at the proposed patch?

mglisse:
I'll look at it, but it would help to know what your goal is with this patch series. Are you trying to get support for MPIR? Nothing wrong if you are, but I'll look at the patches under a different angle depending if you are fixing some crashes, removing some warnings, getting CGAL to compile on cray, etc.

xantares:
yes it's mpir, turns out we had to replace this with mpir_si

PR #1765, opened 2016-12-15
Polyhedron_demo : update for the documentation

maxGimeno:
This PR updates the documentation of the package Three for the polyhedron_demo.


lrineau:
Tested in Ic-80, as a doc-only modification.

PR #1756, opened 2016-12-09
Fixed some typos in Arrangement_2

MaelRL:
This PR fixes some typos in comments and in the doc of the Arrangement_2 package.

PR #1769, opened 2016-12-17
Fix spelling of 'matches'

drewish:
Looks like the misspelling got copy pasted a few other places.

#1756 has some other spelling corrections but it doesn't look like they noticed this.


lrineau:
Thanks. I have merged your fixes in CGAL `master` branch.

PR #1772, opened 2016-12-19
Spatial_searching: Fix early loop abort.

mheistermann:
If distance==squared_radius, we don't yet know if we can abort or not.
This is the same kind of bug that was found and fixed in a related
method two weeks ago in this commit:

ed526b8f22ca0d075ca42c6337a39bf5f0421c68


mheistermann:
On closer inspection, the method comment suggests the check should be a strict inequality instead (test whether sphere _interior_ intersects). I'll create an updated PR.

PR #1773, opened 2016-12-19
Spatial_searching: Fix strict equality / inconsistent early loop abort.

mheistermann:
Compare https://github.com/CGAL/cgal/pull/1772 - I can't quickly find an answer in the documentation as to whether points on the sphere surface should be returned for an (exact) sphere query, so I fixed the inconsistency by relying on the function comment.

**Commit message:**

The method comment suggests the check should be a strict inequality
instead of less-or-equals to test whether the sphere _interior_ intersects.

Additionally, the early abort condition was inconsistent with the final
comparison.
The early abort part is the same kind of bug that was found and fixed in a
related method two weeks ago in this commit:

ed526b8f22ca0d075ca42c6337a39bf5f0421c68

### DIFF
--- a/Spatial_searching/include/CGAL/Fuzzy_sphere.h
+++ b/Spatial_searching/include/CGAL/Fuzzy_sphere.h
@@ -98,7 +98,7 @@ namespace CGAL {
 				((*cit)-rectangle.max_coord(i))*((*cit)-rectangle.max_coord(i));
 		}
 		
-		return (distance <= squared_radius);
+		return (distance < squared_radius);
 	}
 
 


mglisse:
@MaelRL switched from strict to large inequality in c0cbdab3d850749333f4949b6e759b9f37f387bd, he might have something to say about this...

MaelRL:
c0cbdab was just for consistency with 5ed7255, although I seem to have messed up the in-loop checks. It came from a "bug" in the anisotropic mesher where calling `search` from a `kd_tree` would miss points that were on the border of queries (in my case, it was a fuzzy iso box with epsilon set to 0). The documentation wasn't (isn't) clear about what is supposed to happen when epsilon is 0 and points are on the boundary.

I figured that it would make more sense if points contained in the queries were actually returned instead of having to pass artificially-grown queries.

The changes to the `inner` and `outer` intersection functions do go against the comments, which I must have missed when transporting changes from `Fuzzy_iso_box` to `Fuzzy sphere`... This is, I think, because if you have epsilon = 0, you might have the `inner` and `outer` intersection functions both returning "no intersection" while `contains` returns that the Fuzzy_sphere actually contains the point. (I am not so sure about this point anymore, but I think it actually happened in practice). Thus, although I messed up the in-loop checks, changing with `<=` instead of `<` was consistent with the change made in `contains`.

To sum up, this patch should either be a complete revert of both my commits c0cbdab and 5ed7255 (along with ed526b8) and an explanation in the doc that says epsilon = 0 is not handled, or fixing `<=` in the in-loops and changing the comments (and making sure that the changes in `inner/outer` intersection functions of `Fuzzy iso box` and `Fuzzy sphere` to consider the intersection with the whole item and not just its interior do actually make sense).

MaelRL:
I went with option A (reverting everything), see https://github.com/CGAL/cgal/pull/1808.

The change of this PR is included in the other one.

PR #1782, opened 2016-12-23
MPIR support

### LINKED ISSUE #439: Compilation error when using mpir instead of gmp
I'm trying to compile CGAL using Visual Studio 2010. I'm using MPIR version 2.6.0 and I get the following compilation error:

```
  f:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\Number_types\include\CGAL/GMP/Gmpzf_type.h(403): error C2664: '__gmpz_get_d_2exp' : cannot convert parameter 1 from 'CGAL::Gmpzf::Exponent *' to 'mpir_si *' [F:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\build\Installation\src\CGAL\CGAL.vcxproj]
  f:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\Number_types\include\CGAL/GMP/Gmpzf_type.h(414): error C2664: '__gmpz_get_d_2exp' : cannot convert parameter 1 from 'CGAL::Gmpzf::Exponent *' to 'mpir_si *' [F:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\build\Installation\src\CGAL\CGAL.vcxproj]
  f:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\Number_types\include\CGAL/GMP/Gmpzf_type.h(425): error C2664: '__gmpz_get_d_2exp' : cannot convert parameter 1 from 'long *' to 'mpir_si *' [F:\Miniconda64\conda-bld\work\cgal-releases-CGAL-4.7\build\Installation\src\CGAL\CGAL.vcxproj]
```

I found this e-mail on the cgal mailing list: http://cgal-discuss.949826.n4.nabble.com/compilation-error-in-CGAL-4-7-beta-1-in-VC2015-on-x64-Platform-td4661087.html.

Marc Glisse (@mglisse) in the e-mail states that:

>The problem is that MPIR broke compatibility with GMP (not necessarily for 
> bad reasons), it would work if you were using GMP. (Well, the original 
> problem is that Microsoft chose a weird LLP64 ABI instead of LP64 like 
> everyone else.) We will have to think of a workaround...

xantares:
fixes #439, needs #1759 in conjunction, also mpir must be patched to include https://github.com/wbhart/mpir/commit/b84e7e24ca24fd4a965017ff56c25728d2d10fdc (not included in latest 2.7.2 as of writing)

### DIFF
--- a/Installation/cmake/modules/FindGMP.cmake
+++ b/Installation/cmake/modules/FindGMP.cmake
@@ -40,7 +40,7 @@ else()
     cache_set( GMP_IN_CGAL_AUXILIARY TRUE )
   endif()
 
-  find_library(GMP_LIBRARIES NAMES gmp libgmp-10
+  find_library(GMP_LIBRARIES NAMES gmp libgmp-10 mpir
     HINTS ENV GMP_LIB_DIR
           ENV GMP_DIR
           ${CGAL_INSTALLATION_PACKAGE_DIR}/auxiliary/gmp/lib


mglisse:
Indeed, mpir's build.vc/postbuild.bat copies mpir(xx).h to gmp(xx).h, but it doesn't copy the libraries, so something like this (and the same for gmpxx?) seems necessary to support MPIR.
If you rely on the extra commit from mpir's github, I think #1759 is not necessary (we can still do it of course).

PR #1789, opened 2017-01-03
Kernel_23: Make two functors for weighted points CGAL_Kernel_pred_RT

afabri:
I locally compiled the packages Kernel, Triangulation_3, Mesh_3, and Skin_surface

### DIFF
--- a/Kernel_23/include/CGAL/Kernel/interface_macros.h
+++ b/Kernel_23/include/CGAL/Kernel/interface_macros.h
@@ -580,11 +580,11 @@ CGAL_Kernel_pred_RT(Compare_power_distance_2,
 		    compare_power_distance_2_object)
 CGAL_Kernel_pred_RT(Compare_power_distance_3,
 		    compare_power_distance_3_object)
-CGAL_Kernel_pred(Compare_weighted_squared_radius_3,
+CGAL_Kernel_pred_RT(Compare_weighted_squared_radius_3,
                  compare_weighted_squared_radius_3_object)
-CGAL_Kernel_pred(Power_side_of_bounded_power_sphere_3,
+CGAL_Kernel_pred_RT(Power_side_of_bounded_power_sphere_3,
                  power_side_of_bounded_power_sphere_3_object)
-CGAL_Kernel_pred(Oriented_side_2,
+CGAL_Kernel_pred_RT(Oriented_side_2,
 		 oriented_side_2_object)
 CGAL_Kernel_pred(Oriented_side_3,
 		 oriented_side_3_object)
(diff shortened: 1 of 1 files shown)


mglisse:
I see a division in squared_radius_orthogonal_sphereC3, so I am not sure how safe this is. Maybe try recompiling after removing `operator/` from Mpzf (or disable Mpzf, I think Gmpzf and MP_Float don't have `operator/`)? Did you get runtime errors? (if the division is always exact, you might not get any error)

mglisse:
One of the overloads of power_side_of_bounded_power_sphereC3 also does divisions (by 2, so those are safe with Mpzf, but will not compile with others). It could probably be changed to use only multiplications.

afabri:
Conclusion is that this PR makes no sense.

mglisse:
I wouldn't say it makes no sense. It points at prerequisites. It may be a good idea to check which predicates currently use unnecessary divisions. My previous comment seems to indicate that power_side_of_bounded_power_sphereC3 would be a good candidate to become division-free.
It also points out that having many overloads in the "same" predicate makes it more complicated to optimize them.

afabri:
The question is if the split can be done on the granularity of the operators, or only on the granularity of functors.

mglisse:
One goal of the current _RT stuff is to keep things simple (just change the macro used in interface_macros.h). Doing it per operator is of course possible, but in my opinion it is complicated enough to make it not worth the trouble. If you want a different behavior for the different overloads of operator(), my suggestion would be to create new functors, one per overload, and if absolutely necessary keep some kind of wrapper as the old functor to dispatch to the new functors (though one has to make sure that this wrapper doesn't get filtered and forwards to filtered functors, I don't remember how hard that would be in Kernel_23).

lrineau:
I wonder why all predicates do not use a RT by default. `power_side_of_bounded_power_sphereC3` seems to be the only predicate that actually uses divisions. As it can be check using `MP_Float`, why not run the testsuite with `#define CGAL_DO_NOT_USE_MPZF 1` to check if our predicates could use RT instead of FT.

Actually, I found out that none of the predicates on weighted points have tests. That complicates the testing I am talking about.

lrineau:
https://github.com/CGAL/cgal/pull/1789#issuecomment-276696234:
> The question is if the split can be done on the granularity of the operators, or only on the granularity of functors.

No. That is not the question. `power_side_of_bounded_power_sphereC3` has two overloads, and one of the two currently uses divisions. But that overload could be rewrote using only multiplications: for example, the sign of `a - (b/c)` is the same as the sign of `a*c - b`, if `c>0`. Here the predicate is:
```C++
sign (dr2 - dp2/FT4 + dpr*dpw/dp2 - drw )
```
`dp2` is known to be > 0, because it is a sum of squares. So the expression could be multiplied by `FT(4)*dp2`, and it would keep the same sign.

PR #1796, opened 2017-01-05
Add surf format reader

maxGimeno:
This PR adds a reader for the .surf files and a plugin for the Polyhedron_demo.

 The parsing function is the following :
template<class Mesh>
void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<MaterialData>& metadata);

- Mesh is a model of FaceListGraph
- MaterialData is a struct containing two std::pairs<int, std::string>

When the parsing is over, output is filled with one Mesh per patch, and metadata contains one MaterialData per patch. 
metadata[i] contains the metada for the `i`th patch, saying the ID and name of the inner region's material and the outer region's.


afabri:
The function should not have a `_file` as part of its name as it reads from a stream.

maxGimeno:
I edited it to make it read_surf()

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -27,7 +27,7 @@ void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<Mater
   std::vector<Point_3> points;
   std::string line;
   std::istringstream iss;
-  int nb_vertices(0);
+  std::size_t nb_vertices(0);
```
lrineau: Note for @maxGimeno: the size of something is always `std::size_t`.

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -101,10 +100,12 @@ void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<Mater
...
   metadata.resize(nb_patches);
-  for(int i=0; i < nb_patches; ++i)
+  output.resize(nb_patches);
+  const vertex_descriptor null_vertex = boost::graph_traits<Mesh>::null_vertex();
+  std::vector<vertex_descriptor> vertices(nb_vertices, null_vertex);
+  for(int i=0; i < nb_patches; /* i is incremented in the body */)
   {
-    std::map<std::size_t, vertex_descriptor> vertices;
-    Mesh mesh;
+    Mesh& mesh = output[i];
```
lrineau: The line 
```C++
  output.push_back(mesh);
```
is catastrophic: the vector may get reallocated, and all its content copied/moved. And a `Mesh` can be a huge surface, let's not copy it.

Instead, as soon as we get `nb_patches`, let's resize the vector (it is filled with empty `Mesh`es), and then line 108 `mesh ` is a ref to the ith mesh of the `output` vector, and is filled in-place. That's speed-up a lot!

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -101,10 +100,12 @@ void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<Mater
     }
   std::cout<<nb_patches<<" patch(es)"<<std::endl;
   metadata.resize(nb_patches);
-  for(int i=0; i < nb_patches; ++i)
+  output.resize(nb_patches);
+  const vertex_descriptor null_vertex = boost::graph_traits<Mesh>::null_vertex();
+  std::vector<vertex_descriptor> vertices(nb_vertices, null_vertex);
+  for(int i=0; i < nb_patches; /* i is incremented in the body */)
   {
-    std::map<std::size_t, vertex_descriptor> vertices;
```
lrineau: The next speedup is the removal of the `std::map` `vertices. Instead, I allocated once a big vector, and fill it with null vertex descriptors...

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -158,20 +159,23 @@ void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<Mater
...
       for(int id=0; id<3; ++id)
       {
-        if(vertices.find(index[id]) == vertices.end())
+        if(vertices[index[id]] == null_vertex)
         {
-          vertices.insert(std::make_pair(index[id], add_vertex(points[index[id]-1], mesh)));
+          vertices[index[id]] = add_vertex(points[index[id]-1], mesh);
         }
+        face[id] = vertices[index[id]];
       }
```
lrineau: ... and here, if `vertices[index[id]] == null_vertex`, that means that vertex with ID `index[id]` is a new one.

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -158,20 +159,23 @@ void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<Mater
...
-        face.push_back(vertices[index[id]]);
       CGAL::Euler::add_face(face, mesh);
     }
-    output.push_back(mesh);
-  }
+    // reset the `vertices` vector, unless that was the last iteration of
+    // the loop
+    if(++i < nb_patches) std::fill(vertices.begin(),
+                                   vertices.end(),
+                                   null_vertex);
```
lrineau: And at the end of each iterations, the vector `vertices` is filled again with the null vertex descriptor.

lrineau:
@maxGimeno Please read my comments in commit 8133cbbe77dde0d51d53f65859666c04e2555d6f, where I have optimized a lot the memory handling of the loading of the patches.

Polyhedron/demo/Polyhedron/Plugins/IO/XYZ_io_plugin.cpp
```diff
@@ -73,9 +73,6 @@ class Polyhedron_demo_xyz_plugin :
   bool canSave(const CGAL::Three::Scene_item*);
   bool save(const CGAL::Three::Scene_item*, QFileInfo fileinfo);
   bool applicable(QAction*) const { return true;}
-  QList<QAction*> actions() const {
```
janetournois: this function is missing now, and causes a compilation error : "cannot instantiate abstract class" for `Polyhedron_demo_xyz_plugin`

Polyhedron/demo/Polyhedron/Plugins/IO/Surf_io_plugin.cpp
```diff
@@ -0,0 +1,81 @@
...
+  std::vector<MaterialData> material_data;
+  read_surf(in, patches, material_data);
+  for(std::size_t i=0; i<material_data.size(); ++i)
+  {
+   std::cout<<"The patch #"<<i<<":\n  -inner region : material's id = "<<material_data[i].innerRegion.first<<" material's name = "
+           <<material_data[i].innerRegion.second<<"\n  -outer region: material's id = "<<material_data[i].outerRegion.first<<" material's name = "
+             <<material_data[i].outerRegion.second<<std::endl;
+  }
+  std::vector<QColor> colors_;
+  compute_color_map(QColor(100, 100, 255), patches.size(),
```
janetournois: `std::size_t` to `unsigned int` warning happening here

lrineau:
No. I do not understand. I have re-pushed the commit with a `push --force`.

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -0,0 +1,186 @@
...
+};
+
+/*!
+ * read_surf reads a file which extension is .surf and fills `output` with one `Mesh` per patch.
+ * Mesh is a model of FaceListGraph.
+ */
+template<class Mesh, class NamedParameters>
+void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<MaterialData>& metadata,  const NamedParameters&)
+{
+  typedef typename GetGeomTraits<Mesh,
```
janetournois: in `master`, `GetGeomTraits` now is `CGAL::GetGeomTraits`.
It should be updated before testing.
Thanks!

lrineau:
@sloriot: This PR can be tested in `integration`. It has been tested locally on Linux and Windows (by me and @janetournois).

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -0,0 +1,186 @@
...
+  material innerRegion;
+  material outerRegion;
+};
+
+/*!
+ * read_surf reads a file which extension is .surf and fills `output` with one `Mesh` per patch.
+ * Mesh is a model of FaceListGraph.
+ */
+template<class Mesh, class NamedParameters>
+void read_surf(std::istream& input, std::vector<Mesh>& output, std::vector<MaterialData>& metadata,  const NamedParameters&)
```
lrineau: @sloriot @afabri 
Do you know a documented method from CGAL that is supposed to create a sequence of polyhedra. Output iterators are a general tool to build an API that generate things, but when the objects are big, that is not efficient. Here I have chosen to hard-code a non-const reference to a `std::vector<Mesh>` but that is not pretty.

afabri: Do you know the number of polyhedra to avoid reallocation?

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
```diff
@@ -0,0 +1,186 @@
...
+  typedef typename Kernel::Point_3 Point_3;
+  typedef typename boost::graph_traits<Mesh>::vertex_descriptor vertex_descriptor;
+  std::vector<Point_3> points;
+  std::string line;
+  std::istringstream iss;
+  std::size_t nb_vertices(0);
+  std::vector<material> materials;
+  //ignore header
+  while(std::getline(input, line))
+    if(line.compare(0, 13, "    Materials") != 0)
```
janetournois: this (and all similar calls to `line.compare()` below is not robust to indentation changes.
You can use [std::string::find_first_not_of](http://www.cplusplus.com/reference/string/string/find_first_not_of/) or [std::isspace](http://en.cppreference.com/w/cpp/string/byte/isspace) and `std::string::erase` to remove spaces at the beginning of the line and make it more robust.

janetournois: this is one option `line.erase(0, line.find_first_not_of(" \t"));
it erases spaces, and tabs

Polyhedron/demo/Polyhedron/include/CGAL/IO/read_surf_trianglemesh.h
lrineau: @afabri Actually no: the number of polyhedra is only know once the beginning of the file is read.

afabri:
You can also read into a `std::string` with   `operator>>(istream&, string&)`and then read until the end of the line with ` in.ignore(std::numeric_limits<std::streamsize>::max(), '\n') `

afabri:
Isn't a list then a better choice to avoid copies of polyhedra made when a pushback leads to reallocation?

lrineau:
@afabri: The current API uses a non-const ref `std::vector<Polyhedron>&`, and the vector is reserved as soon as the number of polyhedra is known.

If the API was using an output iterator, then I agree a `std::list` or a `std::deque` would be better than a `std::vector`. But, even with an `std::insert_iterator` to a `std::list`, the expression `*out++ = polyhedron` does copy the polyhedra to the memory pointed by the output iterator.

With C++11, if our `CGAL::Polyhedron_3` was efficiently movable, that would not be an issue at all, but for a cross-platform C++98/C++11 API, I cannot find a better way than a non-const ref to an existing container.

sloriot:
At least a new warning [https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-111/Polyhedron_Demo/TestReport_lrineau_ArchLinux-clang-CXX14.gz](here)

maxGimeno:
@sloriot  I don't really know what to do about this warning. It concerns this line : 
CGAL::cpp11::array<unsigned int, 3> grid_size = {1, 1, 1};

Which apparently is correct, but I found several threads on stackoverflow ([this one](http://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis) and [this one](http://stackoverflow.com/questions/13905200/is-it-wise-to-ignore-gcc-clangs-wmissing-braces-warning)) that suggest that it is the compiler that is wrong. Should I use double braces to silent the warning even though the proper way to initialize an array is single braces ?

janetournois:
Note that without this initialization, we have the "variable may be non initialized" warning

sloriot:
@maxGimeno Use double braces.

PR #1808, opened 2017-01-09
Spatial searching: Fix fuzzy query items

MaelRL:
This PR is a more exhaustive version of https://github.com/CGAL/cgal/pull/1773.

The core of the PR is the reversion of some old commits of mine that modified the behavior of Fuzzy query items such that points on the boundary of the inner approximation would always be reported (commits c0cbdab and 5ed7255). Additionally to introducing some bugs in Fuzzy sphere items (https://github.com/CGAL/cgal/issues/1723), these changes are not logically pleasing and are thus reverted. Along the way, related examples/tests are fixed and the documentation was clarified.

To make sure that points on the boundary of the inner approximation are always found, one should for now grow the query item (through its radius, and not epsilon!). That is not completely satisfying, but it's the lesser evil.

### DIFF
--- a/Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
+++ b/Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
@@ -10,12 +10,12 @@ as outer approximation a sphere with radius \f$ r+\epsilon\f$.
 
 \attention The fuzziness of a `Fuzzy_sphere` is specified by a parameter \f$ \epsilon\f$
 denoting a maximal allowed distance to the boundary of a sphere.
-If the distance to the boundary is at least \f$ \epsilon\f$, points inside the
+If the distance to the boundary is greater than \f$ \epsilon\f$, points inside the
 object are always reported and points outside the sphere are never reported.
-Points within distance \f$ \epsilon\f$ to the boundary may be or may be not reported.
-Subsequently, points on the inner and outer spheres may or may not be reported
-by a search query. Specifically when \f$ \epsilon = 0\f$, points on the sphere
-of radius \f$ r\f$ may or may not be reported.
+Points whose distance to the boundary is less than or equal to \f$ \epsilon\f$
+may or may not be reported. Subsequently, points on the inner and outer spheres
+may or may not be reported. Specifically when \f$ \epsilon = 0\f$, points
+on the sphere of radius \f$ r\f$ may or may not be reported.
 
 \tparam Traits must be a model of the concept
 `SearchTraits`, for example `CGAL::Cartesian_d<double>`.
@@ -90,7 +90,7 @@ bool inner_range_intersects(const Kd_tree_rectangle<FT,D>& rectangle) const;
 /*!
 Test whether the outer sphere encloses the rectangle associated with a node of a tree.
 
-That is, whether the minimal distance between the center of the fuzzy sphere and
+That is, whether the maximal distance between the center of the fuzzy sphere and
 `rectangle` is less than \f$ r+\epsilon\f$.
 */
 bool outer_range_contains(const Kd_tree_rectangle<FT,D>& rectangle) const;
--- a/Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
+++ b/Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
@@ -22,14 +22,14 @@ of radius \f$ r\f$ may or may not be reported.
 
 \cgalModels `FuzzyQueryItem`
 
-\sa `FuzzyQueryItem` 
+\sa `FuzzyQueryItem`
 
 */
 template< typename Traits >
 class Fuzzy_sphere {
 public:
 
-/// \name Types 
+/// \name Types
 /// @{
 
 /*!
@@ -38,19 +38,19 @@ Dimension Tag.
 typedef unspecified_type D;
 
 /*!
-Point type. 
-*/ 
-typedef Traits::Point_d Point_d; 
+Point type.
+*/
+typedef Traits::Point_d Point_d;
 
 /*!
-Number type. 
-*/ 
-typedef Traits::FT FT; 
+Number type.
+*/
+typedef Traits::FT FT;
 
-/// @} 
+/// @}
 
-/// \name Creation 
-/// 
+/// \name Creation
+///
 /// @{
 
 /*!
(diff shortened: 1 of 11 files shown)


Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
```diff
@@ -3,25 +3,33 @@ namespace CGAL {
...
+as outer approximation a sphere with radius \f$ r+\epsilon\f$.
+
+\attention The fuzziness of a `Fuzzy_sphere` is specified by a parameter \f$ \epsilon\f$
+denoting a maximal allowed distance to the boundary of a sphere.
+If the distance to the boundary is at least \f$ \epsilon\f$, points inside the
+object are always reported and points outside the sphere are never reported.
+Points within distance \f$ \epsilon\f$ to the boundary may be or may be not reported.
+Subsequently, points on the inner and outer spheres may or may not be reported
+by a search query. Specifically when \f$ \epsilon = 0\f$, points on the sphere
+of radius \f$ r\f$ may or may not be reported.
```
mglisse: It looks like this contradicts the previous paragraph: points on the inner sphere are at distance "at least epsilon", so it says they "are always reported".

Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
```diff
@@ -30,53 +38,62 @@ Dimension Tag.
...
+*/
+bool inner_range_intersects(const Kd_tree_rectangle<FT,D>& rectangle) const;
 
 /*!
-test whether the outer sphere encloses the rectangle associated with a node of a tree. 
-*/ 
-bool outer_range_contains(const Kd_tree_rectangle<FT,D>& rectangle) const; 
+Test whether the outer sphere encloses the rectangle associated with a node of a tree.
+
+That is, whether the minimal distance between the center of the fuzzy sphere and
```
mglisse: Should this be "maximal" instead?

Spatial_searching/include/CGAL/Fuzzy_iso_box.h
```diff
@@ -117,7 +117,7 @@ namespace CGAL {
 	  Cartesian_const_iterator_d pit = construct_it(p);
 	  Cartesian_const_iterator_d minit= min_begin, maxit = max_begin; 
 		for (unsigned int i = 0; i < dim; ++i, ++pit, ++minit, ++maxit) {
-			if ( ((*pit) < (*minit)) || ((*pit) > (*maxit)) ) return false;
+			if ( ((*pit) < (*minit)) || ((*pit) >= (*maxit)) ) return false;
```
mglisse: So the box is half-open (closed on the min side, open on the max side). Is there a particular reason for that choice?

Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
MaelRL: The previous paragraph is some remnant that I had rephrased to use less/greater... but is somehow there without these changes. There were many "distance of at least.... distance of at most... distance within" which were quite confusing; should be good now.

Spatial_searching/doc/Spatial_searching/CGAL/Fuzzy_sphere.h
MaelRL: Yes.

Spatial_searching/include/CGAL/Fuzzy_iso_box.h
MaelRL: I know at the moment little of the inner workings of this package. It is simply what was before.

MaelRL:
To clarify on "not logically pleasing" in the initial message of this PR:

The historical approach assumed that the boundary of the inner and outer approximations are themselves fuzzy regions. Although it is counter intuitive to use when `epsilon` is 0 (especially with all the "exact" keyword thrown around in the doc/comments), this allows for a smooth transition when `epsilon` goes to 0: it is still a fuzzy region.

Instead, if we assume e.g. that all points on the boundary of the inner approximation should be reported, then it is not clear what should be done when `epsilon` is 0.

This PR simply aims to fix things up, but maybe some further work could assume that we always want points on the boundary of the inner approximation to be found (even when `epsilon` is 0)? But as I said in the other PR, I don't know the package well-enough right now to know if the open aspect of the approximation is used.

lrineau:
1. Marc (@mglisse), Mael (@MaelRL), Andreas (@afabri0, do you all agree on the modifications?
1. If yes, does this PR modify the documentation sufficiently to require a small feature?
2. (Whatever is the answer to the previous question) do we target CGAL-4.10-beta1 (code freeze has started) or postpone for 4.11?

afabri:
This PR is a cleanup of the documentation which explains why it is a feature and not a bug that the points **ON** a fuzzy sphere with fuzziness epsilon == 0 may not be reported. So I don't see a need for a small feature.

mglisse:
1. I would like to be able to search for points at distance <= cst from a given point (no fuzziness). I didn't check carefully, but this patch might be making it a bit more difficult (however, CGAL never guaranteed it). On the other hand, in the particular case I have in mind, I might be better off using Kd_tree::search_any_point with my own non-fuzzy region rather than using the nearest neighbor infrastructure.
2. (don't forget to append `?w=1` to the URL to get a sensible diff) It is mostly clarifying something that was unclear before, and where the behavior changed from release to release. I don't know if that requires a small feature or not.

sloriot:
@mglisse Any other remark or are you OK if you merge the branch now it's been successfully tested?

mglisse:
(Mostly from memory, no time to reread the whole thing) I am ok with merging, essentially it clarifies things we cannot rely on. But I want to mention again that some people (including me in Gudhi) need a closed square for exact queries, which duplicates a lot of code from Fuzzy_iso_box (with epsilon=0) just to be sure that points on the boundary are included (I guess that with spheres and the rounding in computing the distance, it is less relevant). So it might make sense to define the behavior on the boundary that way at some point in the future.

lrineau:
I have merged the PR. Marc is right that in the future we may want to help the user who wants an exact (non-fuzzy) query object.

lrineau:
By the way, Marc, can you point us to the "fork" of `Fuzzy_iso_box` that you have? Could not we try to factorize the code between the fuzzy iso-box, and a non-fuzzy iso-box, using a common base class, maybe templated by a Boolean?

mglisse:
I don't really have a fork, I wrote the following as an experiment. It isn't that long, not that much to share, mostly that it would have been nice not to have to write it. And actually I am not completely certain that we really need the thing to be closed... It would probably be a little bit more efficient to store c.x±size and c.y±size.

```
struct myrectangle {
  typedef CGAL::Dimension_tag<2> D;
  typedef Internal_point Point_d;
  typedef double FT;
  bool contains(Point_d p) const {
    return std::abs(p.x()-c.x())<=size && std::abs(p.y()-c.y())<=size;
  }
  bool inner_range_intersects(CGAL::Kd_tree_rectangle<FT,D> const&r) const {
    return
      r.max_coord(0) >= c.x() - size &&
      r.min_coord(0) <= c.x() + size &&
      r.max_coord(1) >= c.y() - size &&
      r.min_coord(1) <= c.y() + size;
  }
  bool outer_range_contains(CGAL::Kd_tree_rectangle<FT,D> const&r) const {
    return
      r.min_coord(0) >= c.x() - size &&
      r.max_coord(0) <= c.x() + size &&
      r.min_coord(1) >= c.y() - size &&
      r.max_coord(1) <= c.y() + size;
  }
  Point_d c;
  FT size;
};
```

PR #1811, opened 2017-01-10
make Nth_of_tuple_property_map compatible with std::tuple

### LINKED ISSUE #1809: Nth_of_tuple_property_map hard-coded for Boost
We have `<CGAL/tuple.h>` that binds `CGAL::cpp11::tuple` either to `std::tuple` or `boost::tuple`, depending on what's available. But the property map `CGAL::Nth_of_tuple_property_map` is hard-coded for _Boost_ only, which makes it incompatible with `CGAL::cpp11::tuple` if the STD version is available.

This seems rather counter-intuitive and forces users to use `boost::tuple` if they want to use the property map provided by CGAL.

(Also, related but debatable: we have `CGAL::cpp11::array` and `CGAL::cpp11::tuple`, but `CGAL::make_array` and `CGAL::cpp11::make_tuple`. The namespaces for the `make_something` should be the same.)

sloriot:
Fixes  #1809 

I consider this as a bug and should target 4.10

### DIFF
--- a/Point_set_processing_3/examples/Point_set_processing_3/property_map.cpp
+++ b/Point_set_processing_3/examples/Point_set_processing_3/property_map.cpp
@@ -22,11 +22,6 @@ typedef K::Vector_3 Vector_3;
 // Point with normal vector stored in a std::pair.
 typedef std::pair<Point_3, Vector_3> PointVectorPair; 
 
-// Data type = index, followed by the point, followed by a boolean
-// that tells us whether the normal is oriented or not, followed by the normal vector.
-typedef boost::tuple<int, Point_3, bool, Vector_3> IndexedPointWithOrientableNormalTuple;
-
-
 
 // This is an implementation detail of the process_point_set function.
 // We need this function because in process_point_set() we use std::sort.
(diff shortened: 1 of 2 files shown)


lrineau:
Why not target 4.9.1 if that is a bug?

Point_set_processing_3/examples/Point_set_processing_3/property_map.cpp
```diff
@@ -156,5 +154,27 @@ int main()
       std::cout << points[i]  << std::endl;
     }     
   }
+
+  #ifndef CGAL_CFG_NO_CPP0X_TUPLE
+  // same test with std::tuple
+  {
```
lrineau: Are you sure you want that "test" in an example?

sloriot: It is indeed not ideal but there is nothing in test and the exemple is not inlined in the doc.

sloriot:
If you prefer I can rebase the branch onto the 4.9-branch.

lrineau:
MSVC does not like that patch.
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-98.shtml#Point_set_processing_3_Examples

PR #1816, opened 2017-01-10
Break the travis tests into several jobs

### LINKED ISSUE #1779: Fix our .travis.yml file
Our `.travis.yml` triggers jobs that are too long to finish before the timeout.

The solution is to decompose our current jobs (g++ and clang++) into multiple shorter jobs.

maxGimeno:
This PR fixes #1779 
It uses one job per package, that builds the examples and the tests.


lrineau:
For demos, the trusty CI environment of Travis does have qt5 (we would have to add packages in our line `addons.apt.packages:`) but the `libqglviewer` package from Trusty
1. is not [whitelisted by Travis],
2. and is the Qt4 variant only.

So to build demos in Travis, we will have to build QGLViewer ourself, and use the cache to keep it built in the travis cache.

[whitelisted by Travis]: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty

lrineau:
The [travis build #418](https://travis-ci.org/CGAL/cgal/builds/190656063) took more than 17 hours. I suggest to remove `-O3` from `CMAKE_CXX_FLAGS_RELEASE`, with
```shell
cmake -DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG"
```
to try reduce the compilation time.

And use `make -j2` or `make -j3` too.

lrineau:
The build errors in Generator and Mesh_3 are both due to Mesh_3 compilations requiring too much memory.
See those travis-ci issues:
- https://github.com/travis-ci/travis-ci/issues/1972
- https://github.com/travis-ci/travis-ci/issues/2407
- https://github.com/travis-ci/travis-ci/issues/5926

In 6ab9e1620a537b5ee2f4816ac69e1dd9e1d7f26f, I have modified the `.travis.yml` to try non-container based workers. It is documented that it provides more memory (>7GB instead of 3GB).

lrineau:
New tasks to do:
  1. create a `.travis` directory at the root (next to the file `.travis.yml`), and push the script `build_package.sh` in it,
  2. commit the scripts that allows to generate the list of packages to test,
  3. add a line in the matrix, that, instead of launching `build_package.sh` for a package, would launch another script, that generates a matrix, compares it with the one that is hard-coded in `travis.yml`, and then fails if the hard-coded list differs from the newly generated one.

The idea with tasks 2. and 3. is that if a pull-request introduces a *new* package, and forgets to update the `.travis.yml`, then Travis CI will report an error.

(Cc: @sloriot)

sloriot:
Does the release creation script handle correctly all those new files at the root of the repository?

lrineau:
@sloriot, https://github.com/CGAL/cgal/pull/1816#issuecomment-272432112:
> Does the release creation script handle correctly all those new files at the root of the repository?

Yes, because it does not look at any file at the root of the repo. Actually, it does look at them, but they are ignored: the function `list_packages` in the perl script `create_internal_release` is
```perl
sub list_packages($) {
  my ($packages_directory) = @_;
  opendir PACKAGESDIR, $packages_directory or die;
  while (defined($package_name = readdir(PACKAGESDIR))) {
    my $package_full_path = File::Spec->catdir($packages_directory, $package_name);
    next if $package_name =~ /^\..*$/;
    next if (! -d $package_full_path );
    next if ($package_full_path =~ /Maintenance$/); # skip Maintenance package
    $LISTOFALLPACKAGES{$package_name} = $packages_directory;
  }
  closedir(PACKAGESDIR);
}
```

The first `next if` will skip any file or directory starting with a dot, the second one skips anything but directories, and the last one skips `Maintenance`.

maxGimeno:
I added some files in .travis:
- generate_travis.sh will generate the .travis.yml file with an automaticaly generated matrix containing all directories that math PACKAGE/(demo || test || example)/PACKAGE.
- template.txt is the template for the .travis.yml file. If we want to modify its behaviour, we should do it only in the template, and then call the previous script.
- packages.txt is the list containing all the packages that were last used to generate the MATRIX. It is used by build_packages.sh to check that no package is missing from the MATRIX.

lrineau:
https://travis-ci.org/CGAL/cgal/requests
> Build matrix exceeds 200 jobs.

You'll have to find a way to reduce the number of jobs.

lrineau:
If a PR adds a new package and forgets to rebuild the `.travis.yml` by calling `.travis/generate_travis.sh `, how will the Travis script detect the change?

I think the script `.travis/generate_travis.sh` should be called at each Travis run (in the `CHECK` job), and then the list of packages should be compared.

maxGimeno:
@lrineau if one forgets to call generate_travis.sh, the file packages.txt will not be updated either. So when the CHECK phase is performed, the current list of packages will be compared to the one saved in packages.txt and they won't match. 
Moreover, I fear that replacing the .travis.yml at runtime would be dangerous.

lrineau:
(Sorry for the late comment)

I do not understand... if a PR adds a new package without modifying the `.travis.yml`, will the job named `CHECK` fail automatically?

PR #1823, opened 2017-01-12
Polyhedron_demo : Enhance point set selection plugin

maxGimeno:
This PR removes the temporary item in the point_set_selection_plugin and adds a select_box as a third way of selecting.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
@@ -102,14 +102,9 @@ struct Scene_edit_box_item_priv{
     pool[1] = bb.ymin(); pool[4] = bb.ymax();
     pool[2] = bb.zmin(); pool[5] = bb.zmax();
 
-    double sq_diag =
-        (bb.xmax() - bb.xmin()) * (bb.xmax() - bb.xmin()) +
-        (bb.ymax() - bb.ymin()) * (bb.ymax() - bb.ymin()) +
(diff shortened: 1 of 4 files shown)


sgiraudot:
I tested it on my computer and it works pretty well.

Some remarks:

- when selecting `Selection Box`, a new button `Add to Selection` appears, the plugin's widget becomes taller and there's a graphic bug with Qt (the last button of the widget is superimposed with the Console/Info widgets above). I'm not sure if this is a Qt bug or if this is something we prevent from our side
- on some point sets, the red balls are very very big, which is kind of weird. I don't know how the radius are computed but maybe it should be proportional to the size of the box? Otherwise it's difficult to see anything when the box is really small (it's covered by the balls)
- I'm not a fan of the behavior "Add to Selection" that differs from the original behavior (originally, when you release the mouse, the selection appears, but here you have to click to that button and you can't really preview what will be selected). But I understand that it's difficult to do otherwise, so as I don't have a better idea to suggest, I'm okay with it.

lrineau:
Simon, let us know if the commit 1685eeb45b993897e03d116eb9cdd4ea65b9128e has fixed the issues you pointed out.

lrineau:
ping @sgiraudot: https://github.com/CGAL/cgal/pull/1823#issuecomment-272194427

sgiraudot:
Yes, looks good to me. The fix for the "Add to selection" (which is now called "Validate" and inserted in a way that does not distort the widget), is also good.

I don't have any other remark to make, this PR looks good to me.

lrineau:
Tested in Ic-103, and approved by Simon. I merge...

PR #1824, opened 2017-01-12
Polyhedron demo : various fixes and enhancements

maxGimeno:
This PR :
- fixes various bugs in the selection_plugin
- Specifies the wanted extension of the file being saved in the IO Dialog
- Fixes the "Duplicate Filter" warning
- Makes all selection_item's simplices dependent of its selected color, to help distinguish between multiple selection_items.

PR #1826, opened 2017-01-12
Polyhedron demo : extend selection

### LINKED ISSUE #1775: Feature needed: selection from edges to facets
A feature that would be very much needed: I have made a selection of edges: I would like to convert it into a selection of facets adjacent to these edges.. Or simply to create another facet selection from the current selected edges.

Motivation: to cleanup the surface mesh around holes when the geometry is very messy. 

We can also imagine more general conversion operations, ie from edges to facets, from facets to boundary edges, from edges to edge vertices, etc. 

Pierre

maxGimeno:
This PR fixes #1775 

 It adds conversion between edges selection and adjacent facet selection, and between facets selection and bounding edges selection.

It also adds a lasso selection.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
@@ -98,6 +98,8 @@ public:
     connect(ui_widget.Create_selection_item_button,  SIGNAL(clicked()), this, SLOT(on_Create_selection_item_button_clicked()));    
     connect(ui_widget.Selection_type_combo_box, SIGNAL(currentIndexChanged(int)), 
             this, SLOT(on_Selection_type_combo_box_changed(int)));
+    connect(ui_widget.lassoCheckBox, &QCheckBox::toggled,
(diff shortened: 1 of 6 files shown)


Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
```diff
@@ -627,6 +743,7 @@ public Q_SLOTS:
     if (scene_ptr)
       connect(selection_item,SIGNAL(simplicesSelected(CGAL::Three::Scene_item*)), scene_ptr, SLOT(setSelectedItem(CGAL::Three::Scene_item*)));
     connect(selection_item,SIGNAL(isCurrentlySelected(Scene_polyhedron_item_k_ring_selection*)), this, SLOT(isCurrentlySelected(Scene_polyhedron_item_k_ring_selection*)));
+    on_LassoCheckBox_changed(ui_widget.lassoCheckBox->isChecked());
```
lrineau: This has a conflict with [that modification] from #1824.

[that modification]:  https://github.com/CGAL/cgal/pull/1824/files#diff-4fa131e055014434c9fc0405938df65e

PR #1827, opened 2017-01-12
Demo Distance_plugin : fix compilation error

janetournois:
This PR removes a `typename` that causes a compilation error on msvc :
"C2899: typename cannot be used outside a template declaration"

The compilation error was introduced by PR #1578


janetournois:
@maxGimeno @lrineau 
Since the PR #1578 had been tested, I had a look at the corresponding testsuite.
It seems that on [this platform](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-97/Polyhedron_Demo/TestReport_afabri_x64_Cygwin-Windows8_MSVC2013-Debug-64bits.gz) for example, the `distance_plugin` was not compiled at all.

This is something to be checked in the upcoming tests.

PR #1830, opened 2017-01-13
Linear_cell_complex_demo:  Reorder includes to work around a bug in b…

afabri:
…oost

Reordering includes of CGAL headers changes the inclusion order
for boost headers.  The test file  flat.cpp should reveal the bug

This should fix a bug reported in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-100/Linear_cell_complex_Demo/TestReport_lrineau_ArchLinux.gz)


sloriot:
What's the point of `flat.cpp`? I'm afraid it includes non documented headers and will not compile depending on the version of boost used.

afabri:
It's the two files of boost, where we think that the inclusion order leads to a problem.
Once this is confirmed we can file a bug report at boost.org

afabri:
What is strange:  The demo is fixed in Ic-103, but the small proof  of bug flat.cpp does not fail on ArchLinux, but does on one VC++ platform.

Linear_cell_complex/test/Linear_cell_complex/flat.cpp
```diff
@@ -0,0 +1,12 @@
...
+#include <boost/property_tree/ptree.hpp>
+
+#include <boost/container/flat_set.hpp>
+
+int main()
+{
+  boost::container::flat_set<int> fs;
+
+  return 0;
+}
```
lrineau: What is the plan for that file? We thought it was a minimal test case for the bug, but it is not.

Actually, all the docker images have been rebuild two days ago, and the version of Boost has changed. Maybe the bug is already fixed.

I suggest the file to be removed.

lrineau:
Now that the file `flat.cpp` is deleted, I will merged this PR:
- the bug seems to be only in Boost-1.62,
- the patch has been tested in CGAL-4.10-Ic-103, and the column Debian-Testing shows that the patch does fix the bug.
- `flat.cpp` is not the minimal failing test file (does not fail). It does fail on Windows with Boost-1.55... but we give up and will not report any bug.

PR #1832, opened 2017-01-13
Handle boost prior to 1 51

sloriot:
This fixes warnings and errors boost such as [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-100/Polygon_mesh_processing/TestReport_lrineau_CentOS5.gz) and these [warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-100/Polygon_mesh_processing_Examples/TestReport_lrineau_CentOS5.gz).

It also includes a fix for the availability of the parallel mesh distance computation.


Polygon_mesh_processing/examples/Polygon_mesh_processing/hausdorff_distance_remeshing_example.cpp
```diff
@@ -4,7 +4,7 @@
 #include <CGAL/Polygon_mesh_processing/distance.h>
 #include <CGAL/Polygon_mesh_processing/remesh.h>
 
-#if CGAL_LINKED_WITH_TBB
+#if defined(CGAL_LINKED_WITH_TBB) && !defined(CGAL_NO_ATOMIC)
```
lrineau: If an implementation uses TBB, then it can use TBB atomic types, instead of those provided by the STL or Boost.

sloriot:
You mean that we should update the mechanism in `atomic.h` to also try to use TBB if it is here and no other solution is available?

lrineau:
... yes, that would be the simplest solution: all the hacks in `<CGAL/atomic.h>`.

sloriot:
It seems the [API](https://software.intel.com/en-us/node/506277) of `tbb::atomic` is not similar and not complete wrt `std::atomic`. In particular, it seems there is only 2 memory ordering constraints and no atomic fence. We cannot use it except if we reduce the requirements of  `CGAL::cpp11`. Since this is not trivial, it is out of the scope of this PR IMO.

sloriot:
Another solution is to hardcode `tbb::atomic` if `CGAL_NO_ATOMIC` is defined in `distance.h`. @lrineau, what do you think?

lrineau:
https://github.com/CGAL/cgal/pull/1832#issuecomment-272805597:
> Another solution is to hardcode `tbb::atomic` if `CGAL_NO_ATOMIC` is defined in distance.h.
@lrineau, what do you think?

I agree. That was actually what I had in mind with my first comment (https://github.com/CGAL/cgal/pull/1832#discussion_r95990933): `distance.h` already uses the TBB API. It can go on with the TBB API and TBB atomic types directly instead of using `<CGAL/atomic.h>`.

lrineau:
The errors and warnings are indeed fixed.

PR #1841, opened 2017-01-18
Fix a bug in Sweep-line

sloriot:
I've added the test failing in the testsuite before my two patches and that passes after. I've run the testsuites of arrangement, sweep-line and Boolean set operations.
@efifogel could  you have a look?

Here are the 3 polycurves producing the bug:
argt


efifogel:
I'll check it thoroughly next week.
Who reported this?

                         _/



On Wed, Jan 18, 2017 at 10:36 AM, Sebastien Loriot <notifications@github.com

sloriot:
That's an example I created.

sloriot:
@efifogel I also pushed 95ea546 that fixes a precondition raised in `Multiset.h` when the event point is not the left end point of the curve being inserted.

efifogel:
Removing the statement below on line 320 (before your change) is NOT the
solution!

*iter = curve; // replace the current curve with the new one.

I've introduced this line in my last fix.
I haven't updated the testsuit yet, so before I continue, try the attached
program. I believe that, after your fix, it will fail.

How would you like to proceed?
We can work on it together. In this case we need to have a branch
accessible to both of us, or
I continue to work on it a bit, and let you know.

                         _/

efifogel:
@Sebastien, do you have a case that triggers the fix
to Sweep_line_functors.h ?

                         _/

sloriot:
The example added in the testsuite fails without the fix. The reason is a curve that is on the right of an event that is no longer in the status line since it had been replaced by the line I commented. I tried you example and it indeed failed with the fix. I'm adding it in the testsuite in this branch. I also pushed the branch in `cgal-public-dev`.

efifogel:
I know that the example you added fails without the fix and I know the
reason for the failure.
As a matter of fact, I believe that I know how to fix it for good.
As you added the branch to cgal-public-dev I'm going to work on the branch
now.

BW, these tests should not go into the testsuit, at least not as are. I'll
work them out as well.

                         _/

afabri:
Hi Efi,  It would be good if you also documented the reasoning of your fix to come.

lrineau:
@efifogel Those "cleanup" commits have nothing to do in a bug-fix branch like `releases/CGAL-4.9-branch`. That just makes the bug-fix patch harder, or impossible, to review!

efifogel:
The cleanup and the bug fix are separate commits.
You cannot stop me from fixing indentation and similar things, which I
refer to as "Clean ups".
BW, I don't do it randomly; I do it for files developed at TAU; I do it
only if I need to work on a file, and only if the styling is inconsistent;
I do it before I start the actual work in separate commits (at least that's
my intention).
On the bright side, I don't think that there are many files originated at
TAU that I haven't touched yet...

                         _/

lrineau:
> The cleanup and the bug fix are separate commits. 

That is right. And the clean-up commits should go to the development branch only, where the bug-fix should go to the bug-fix branch `CGAL-4.9-branch` and the development branch. Should we split this PR into two distinct PRs (one for 4.9 and one for 4.10)?

sloriot:
I'll handle when we're done with the fixes. For the moment we are still working on the last fix.

lrineau:
This pull-request is blocking CGAL-4.9.1. What is the schedule for it?

lrineau:
I should also say that CGAL-4.9.1 is a blocker for CGAL-4.10-beta1.

sloriot:
@lrineau we target 4.11 with this branch.

sloriot:
I have removed all clean-up commits

lrineau:
@sloriot Do you agree that this PR is stalled, or re-scheduled for CGAL-4.12 ?

sloriot:
It's not but the code is totally different from the one and currently in another branch.

efifogel:
Hi Guys,

The branch mine/Aos_2-surface_sweep-efif contains a significant improvement
of the entire sweep code.
Details follow.

1. The changes introduced by this branch do not affect the exposed API
whatsoever.
2. I've change the name of the main directory from Sweep_line_2 to
Surface_sweep_2. It's a better name in general, but it also matches the
name of the directory that holds the 2D arrangement code, namely
Arrangement_on_surface_2.
3. Sometime ago Andreas and perhaps others moved files from the Aos_2
directory to the Sl_2 (Now Ss_2) also changing the included license file.
The move was done a bit hastily. I've moved one file that was left out, and
I moved back some files that shouldn't have been moved in the first place
(naturally, updating the included license file.)
4. In the new Ss_2 I've introduced a namespace, namely Surface_sweep_2.
However, I'm using it at this point only internally, so the exposed
functions are not affected.
5. The changes where made in order to get rid of the use of 'void *' and
castings (both C and C++ styles) which where used to handle
interdependencies. For example, the surface sweep uses something called
Event and Subcurve, and each template contains references to object of the
other. Now the code has better type checking and I manage to optimize it a
bit, so it is more efficient space and time wise.
6. I've also fixed the scheme of instantiation so that the surface-sweep
stuff can be exposed (once the appropriate documentation is provided as
part of a formal feature submission). For example, the main sweep template
is parameterized only by a visitor, and the visitor is parameterized by the
traits, the event, the subcurve, and the allocator. (In the past several
types where parameterized by several other and all had defaults, which
obviously didn't make sense). In particular, I've fixed the definition of
the Allocator so that a user of the sweep can provide her own.
7. I had to use curiously recurring template pattern (CRTP) in several
places and they are intermixed. Unfortunately, I cannot use C++11 features,
but I managed. Everything compiles on G++ 4.8.4 and VC 14. I have a feeling
that I'll have to work on it to get older VC versions to work, but this is
to come.

I haven't gone further with the complete submission of the feature, because
there is a more urgent task to do---we need to fix the bug reported by a
user in the sweep code first.

I would like to integrate this branch at this point, continue with the bug
fix (that is, continue to work with Sebastien on the bug fix) and then come
back and submit a feature that exposes the surface-sweep code.

If this is ok with you, I'll go ahead and create a pull request. Otherwise,
let me know what you think I should do.

Thanks,
Efi

                         _/

afabri:
It is up to Sébastien to say if he finds that a good or bad idea. With the limited insight I have I would say fix the bug first, and then apply your changes paying attention that there is no regression.

efifogel:
If Sebastien is able to fix the bug, then I could certainly wait.
When Sebastien and I talked about it, we suspected that the fix would
require massive changes, and so it makes sense to make them on top of my
big change. Besides, with tighter type checking it can help the development
and debugging in general.
Anyhow, Sebastien, we can discuss it further if needed.

                         _/

sloriot:
I've managed to workaroud the issue I mentioned to you last week. Now the testsuites+examples of Arrangement, Sweep_line and Boolean_set_operation_2 is working. I'm left with one issue relate to the bug reported by a user, but I know what it's going on. I need to take a break on this topic as I have another deadline. I'll hope to get back to it next week and as soon as I'm done we will need to discuss the changes I've made to see if you agree with them.

Code freeze is today so your modifications will go in 4.12 while I hope we will be able to include the bug-fix in 4.11. So from my point on view you can already start the PR. On the other hand, there are so many PRs open that target 4.11 that there is nothing urgent ( I guess @lrineau will wait for that number to decrease before opening the 4.11-branch and make master target 4.12).

efifogel:
ok, I'll go ahead an create a PR.
Actually, I'd like to run this brunch through the test suite once to see
how it runs on all platforms.

                         _/

sloriot:
I've created  PR #2342 which is covering all cases that were previously failing.

PR #1843, opened 2017-01-18
Accelerate copy_face_graph

sloriot:
If an internal index property map is available, we use it a vector instead of an unordered map.
The implementation was also changed to avoid using `Euler::add_face()`.

There is a questionable point. If the index property map is writable, it is always reinitialized which might be unwanted.


lrineau:
There are compilation errors in the Polyhedron_Demo line of the testsuite:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-105.shtml#Polyhedron_Demo

lrineau:
Do we select this PR for 4.10 or 4.11?

lrineau:
I also think this PR is the reason of all the read lines with segfaults. For example, `test/Polytope_distance_d/test_width_cube` on my machine sefaults:
```
Program received signal SIGSEGV, Segmentation fault.
CGAL::halfedge<CGAL::Triangulation_vertex_base_with_info_2<int, CGAL::GT3_for_CH3<CGAL::Cartesian<double> >, CGAL::Triangulation_vertex_base_2<CGAL::GT3_for_CH3<CGAL::Cartesian<double> >, CGAL::Triangulation_ds_vertex_base_2<void> > >, CGAL::Convex_hull_face_base_2<int, CGAL::Cartesian<double>, CGAL::Triangulation_face_base_2<CGAL::Cartesian<double>, CGAL::Triangulation_ds_face_base_2<void> > > > (v=..., g=...)
    at /home/lrineau/Git/cgal/Triangulation_2/include/CGAL/boost/graph/graph_traits_Triangulation_data_structure_2.h:356
356         int i = fd->index(v);
```
and in the backtrace the function just before is:
```
CGAL::internal::copy_face_graph_impl<CGAL::Triangulation_data_structure_2<[....]
```

I will stop have a look at CGAL-4.10-Ic-105 because probably all the errors comes from this PR.

lrineau:
What is the plan for this PR?

sloriot:
I'm missing time for now. I'd postpone it to 4.11

sloriot:
@afabri seems to have fix the issue.

afabri:
@sloriot Does the "questionable point" need further discussion?

sloriot:
No, it means ready to be tested.

lrineau:
So, does it mean that we can still consider this PR for 4.10?

lrineau:
Can this PR explain [those compilation errors](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-122/AABB_tree/TestReport_afabri_x64_Cygwin-Windows8_MSVC2010-Release-64bits.gz)? That is as if `CGAL_CFG_NO_CPP0X_TUPLE` should have been defined but is not.

afabri:
The same error is in the testsuite of the day before without this PR, but with BGL-reserve-GF.
As both testsuites are partial we better wait for a complete VC2010 testsuite.

lrineau:
On Picasso, both Ic-121 and Ic-122 suffered from a very partial test. Let's retest again with a full testsuite. There might be side-effects in the way a partial testsuite is launched.

PR #1857, opened 2017-01-25
Polyhedron_demo: Fix Edit_polyhedron_plugin

maxGimeno:
## Summary of Changes
This PR fixes the segfault that occurs when the Edit_polyhedron_plugin is called on an item without offset.

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
@@ -210,28 +210,29 @@ void Scene_edit_polyhedron_item_priv::initializeBuffers(CGAL::Three::Viewer_inte
 {
     //vao for the facets
     {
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -244,7 +244,8 @@ void Scene_edit_polyhedron_item_priv::initializeBuffers(CGAL::Three::Viewer_inte
         item->buffers[Facet_normals].release();
         item->vaos[Facets]->release();
         program->release();
-        delete vertices;
+        if(offset.norm() !=0)
+          delete vertices;
```
lrineau: Why was `vertices` created on the heap anyway? If that `std::vector` is meant to be temporary, then just create it on the stack!
```C++
std::vector<GLdouble> vertices;
if(offset.norm() !=0) {
  vertice.resize(positions.size()));
  // ...
}

PR #1877, opened 2017-02-01
Fix doc headers

### LINKED ISSUE #1874: Detect documentation bugs
## Issue Details

SInce the documentation is not inline for all packages, we should guarantee that documented header and real headers are the same. For example, we document `#include <CGAL/IO/output_surface_facets_to_off.h>` for `CGAL::output_surface_facets_to_off ()` while this header does not exist  (the real header is `CGAL/IO/Complex_2_in_triangulation_3_file_writer.h`)

- [ ] The simple thing we can do it to update `test_merge_of_branches` to check that any header in `doc/PKG/CGAL` exists. There might be false positive if the header is not in the package but I'm pretty sure we can workaround these.
- [ ] The second part of this issue is to fix the errors

maxGimeno:
## Summary of Changes
This PR adds a script that checks if all the headers in Pkg/doc/pkg/CGAL actually have an equivalent. It also adds the test_merge_branch to the travis tests.
## Release Management

* Issue(s) solved (if any): fix #1874


maxGimeno:
@lrineau  The script detected a license error in master

sloriot:
`test_merge_of_branches` seems to be called twice for each job. Can't we simply call it once?

maxGimeno:
I don't think I can do better than this, AFAIK, all the configuration is called for each job.

.travis.yml
```diff
@@ -45,11 +45,13 @@ env:
   - PACKAGE='CHECK' 
 install: 
 before_script: 
+- sudo apt-get install zsh -y
```
lrineau: No. Add `zsh` in the list `addons.apt.packages` instead. That avoids that apt is called twice.

.travis.yml
```diff
@@ -45,11 +45,13 @@ env:
...
 - mkdir -p build 
 - cd build 
-- cmake -DCGAL_HEADER_ONLY=ON -DCMAKE_CXX_FLAGS_RELEASE=-DCGAL_NDEBUG .. 
+- cmake -DCGAL_HEADER_ONLY=ON -DWITH_demos:BOOL=TRUE -DWITH_examples:BOOL=true -DWITH_tests:BOOL=TRUE -DCMAKE_CXX_FLAGS_RELEASE=-DCGAL_NDEBUG .. 
 - make 
 - cd .. 
+- zsh -x -e ./Scripts/developer_scripts/test_merge_of_branch HEAD
```
lrineau: That line `zsh -x -e ./Scripts/developer_scripts/test_merge_of_branch HEAD` should be in the `CHECK` section of `.travis/build_package.sh`.

lrineau: By the way, please put the `PACKAGE='CHECK'` as the first job, to allow early detections of issues.

lrineau:
For your information, the call to `test_merge_branch` in `.travis.yml` works, and has already detected one real issue in a PR of mines:
```
.. Checking include directives in header files...

Mesh_3/include/CGAL/Polyhedral_complex_mesh_domain_3.h

(aborting the merge now)
```
https://travis-ci.org/CGAL/cgal/jobs/197708820

Thanks for this PR. That will help me a lot as a release manager!

PR #1879, opened 2017-02-02
Polyhedron_demo: Fix the column resizing

maxGimeno:
## Summary of Changes
The name column gets resized instead of the AB column.
## Release Management


janetournois:
This PR should be split into two (one for the edit plugin, and one for the columns size), right?
Doing this will also allow you to remove the extra empty "merge" commits.

sloriot:
What is `CustomTreeView`? it does not seem to be used here.

maxGimeno:
No it was just a bug from a pull of master I think. I didn't touch the edit plugin.

janetournois:
than you should rebase the branch to have only the commit of interest in the PR

maxGimeno:
That is what I did

sloriot:
Tested locally, it fixes a very annoying bug

janetournois:
thanks

lrineau:
There is a conflict with the PR #1796 recently merged. Be careful with the conflict. Do not remove the lines added by #1796.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1934,11 +1927,22 @@ void MainWindow::resetHeader()
...
+    sceneView->header()->resizeSection(Scene::NameColumn, name_width);
+}
+
+bool MainWindow::eventFilter(QObject *, QEvent *event)
+{
+  if(event->type() == QEvent::Resize)
+  {
+    resetHeader();
+  }
+  return false;
```
maxGimeno: Apparently not. Even with ABColumn set to Fixed, it getd stretched and there is not enough space for the NameColumn. Plus, I don't remember exactly what Andreas specifically wanted when I started to change things in the sceneView, but I personally find it better to be able to resize the NameColumn as wanted by the user.

lrineau: Re-read carefully the documentation of `QHeaderView::setSectionResizeMode`!

maxGimeno: yes...

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -394,6 +394,7 @@ MainWindow::MainWindow(QWidget* parent)
 
   // setup menu filtering
   connect(ui->menuOperations, SIGNAL(aboutToShow()), this, SLOT(filterOperations()));
+  ui->sceneDockWidget->installEventFilter(this);
```
lrineau: You can remove this line, now.

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -212,6 +212,7 @@ public Q_SLOTS:
 
   /// This slot is used to test exception handling in Qt Scripts.
   void throw_exception();
+
```
lrineau: And not adding that empty line in the .h file will save the need to rebuild all plugin including that header file. You should remove it, so that `MainWindow.h` stays untouched by the PR.

PR #1885, opened 2017-02-06
Add a .mesh reader

### LINKED ISSUE #1851: Polyhedron demo : add a tetrahedron soup reader
Similarly to the polygon soup reader, it would be nice to have a tetrahedron soup reader.
The idea is to be able to
- [x] read a `.mesh` file (there are other formats, to be added later probably), 
- [x] display the tet soup (similarly to a `C3t3`)
- [x] get statistics on it (edge lengths, dihedral angles, nb of subdomains...)

It would be useful for our experiments on meshing, and would sometimes avoid us to resort to external tools.

I have some draft code to represent the tet soup, and read several file formats into a tet soup. Ask me when needed!

@lrineau @maxGimeno

maxGimeno:
## Summary of Changes
This PR:
- provides a reader for the .mesh files
-  allows the c3t3_io_plugin of the Polyhedron_demo to load tetrahedron soups from .mesh files into "invalid" Scene_c3t3_items. 
- Adds statistics to the scene_c3t3_item
- Adds a slider to _shrink_ the facets of a scene_c3t3_item (by using the slider in the item's context menu or using + and - ).

## Release Management

* Affected package(s): Polyhedron_demo, Mesh_3
* Issue(s) solved (if any): fix #1851

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
+++ b/Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
@@ -1,0 +1,674 @@
+#ifndef CGAL_MESH_3_TET_SOUP_TO_C3T3_H
+#define CGAL_MESH_3_TET_SOUP_TO_C3T3_H
+
+#include <CGAL/IO/File_medit.h>
+
+#include <CGAL/assertions.h>
+
+#include <boost/array.hpp>
+#include <boost/unordered_set.hpp>
+#include <boost/unordered_map.hpp>
+#include <boost/foreach.hpp>
+
+// The purpose of this file is to rebuild manually a triangulation from a c3t3
+// output file (because generating a new triangulation every time is expensive)
+
+namespace CGAL
+{
+// the output gives all the cells :
+// - the cells in complex are under 'Tetrahedra', with a reference field that is
+//   their subdomain's index
+// - the finite cells not in complex are also given under 'Tetrahedra', but their
+//   reference field is '-1'
+// - the infinite cells are given under 'InfiniteCells', which is not a medit
+//   keyword and will get ignored by medit, but it is useful information (to me)
+//   to build a triangulation from the output file
+
+template <class C3T3,
+          class Vertex_index_property_map,
+          class Facet_index_property_map,
+          class Facet_index_property_map_twice,
+          class Cell_index_property_map>
+void
+output_to_medit_all_cells(std::ostream& os,
+                          const C3T3& c3t3,
+                          const Vertex_index_property_map& vertex_pmap,
+                          const Facet_index_property_map& facet_pmap,
+                          const Cell_index_property_map& cell_pmap,
+                          const Facet_index_property_map_twice& facet_twice_pmap = Facet_index_property_map_twice(),
+                          const bool print_each_facet_twice = false)
+{
+  typedef typename C3T3::Triangulation                  Tr;
+  typedef typename C3T3::Facets_in_complex_iterator     Facet_iterator;
+
+  typedef typename Tr::Finite_vertices_iterator         Finite_vertices_iterator;
+  typedef typename Tr::All_cells_iterator               All_cells_iterator;
+  typedef typename Tr::Finite_cells_iterator            Finite_cells_iterator;
+  typedef typename Tr::Vertex_handle                    Vertex_handle;
+  typedef typename Tr::Point                            Point_3;
+
+  const Tr& tr = c3t3.triangulation();
+
+  //-------------------------------------------------------
+  // Header
+  //-------------------------------------------------------
+  os << std::setprecision(17);
+
+  os << "MeshVersionFormatted 1" << std::endl
+     << "Dimension 3" << std::endl;
+
+  //-------------------------------------------------------
+  // Vertices
+  //-------------------------------------------------------
+  os << "Vertices" << std::endl
+     << tr.number_of_vertices() << std::endl;
+
+  std::map<Vertex_handle, int> V;
+  int inum = 1;
+
+  Finite_vertices_iterator fvend = tr.finite_vertices_end();
+  for(Finite_vertices_iterator vit = tr.finite_vertices_begin();
+                               vit != fvend; ++vit)
+  {
+    V[vit] = inum++;
+    Point_3 p = vit->point();
+    os << CGAL::to_double(p.x()) << " "
+       << CGAL::to_double(p.y()) << " "
+       << CGAL::to_double(p.z()) << " "
+       << get(vertex_pmap, vit)
+       << std::endl;
+  }
+
+  //-------------------------------------------------------
+  // Facets
+  //-------------------------------------------------------
+  typename C3T3::size_type number_of_triangles = c3t3.number_of_facets_in_complex();
+
+  if(print_each_facet_twice)
+    number_of_triangles += number_of_triangles;
+
+  os << "Triangles" << std::endl
+     << number_of_triangles << std::endl;
+
+  Facet_iterator ficend = c3t3.facets_in_complex_end();
+  for(Facet_iterator fit = c3t3.facets_in_complex_begin();
+                     fit != ficend; ++fit)
+  {
+    for(int i=0; i<4; i++)
+    {
+      if(i != fit->second)
+      {
+        const Vertex_handle vh = (*fit).first->vertex(i);
+        os << V[vh] << " ";
+      }
+    }
+    os << get(facet_pmap, *fit) << std::endl;
+
+    // Print triangle again if needed
+    if(print_each_facet_twice)
+    {
+      for(int i=0; i<4; i++)
+      {
+        if(i != fit->second)
+        {
+          const Vertex_handle vh = (*fit).first->vertex(i);
+          os << V[vh] << " ";
+        }
+      }
+      os << get(facet_twice_pmap, *fit) << std::endl;
+    }
+  }
+
+  std::cout << "output: " << std::endl;
+  std::cout << tr.number_of_vertices() << " vertices" << std::endl;
+  std::cout << c3t3.number_of_cells_in_complex() << " finite cells in complex" << std::endl;
+  std::cout << tr.number_of_finite_cells() << " finite cells" << std::endl;
+  std::cout << tr.number_of_cells() << " total cells" << std::endl;
+
+  //-------------------------------------------------------
+  // Tetrahedra
+  //-------------------------------------------------------
+
+  os << "Tetrahedra" << std::endl
+     << tr.number_of_finite_cells() << std::endl;
+
+  Finite_cells_iterator fcend = tr.finite_cells_end();
+  for(Finite_cells_iterator cit = tr.finite_cells_begin();
+                            cit != fcend; ++cit )
+  {
+    for(int i=0; i<4; i++)
+      os << V[cit->vertex(i)] << " ";
+    os << get(cell_pmap, cit) << std::endl;
+  }
+
+  //-------------------------------------------------------
+  // Infinite cells
+  //-------------------------------------------------------
+
+  os << "InfiniteCells" << std::endl
+     << tr.number_of_cells() - tr.number_of_finite_cells() << std::endl;
+
+  All_cells_iterator acend = tr.all_cells_end();
+  for(All_cells_iterator cit=tr.all_cells_begin();
+                         cit!=acend; ++cit )
+  {
+    if(!tr.is_infinite(cit))
+      continue;
+
+    for (int i=0; i<4; i++)
+    {
+      Vertex_handle vh = cit->vertex(i);
+      if(tr.is_infinite(vh))
+        os << "0 ";
+      else
+        os << V[vh] << " ";
+    }
+    os << "-1" << std::endl;
+  }
+
+  //-------------------------------------------------------
+  // End
+  //-------------------------------------------------------
+  os << "End" << std::endl;
+}
+
+
+template <class C3T3, bool rebind, bool no_patch>
+void
+output_to_medit_all_cells(std::ostream& os,
+                          const C3T3& c3t3)
+{
+  typedef CGAL::Mesh_3::Medit_pmap_generator<C3T3,rebind,no_patch> Generator;
+  typedef typename Generator::Cell_pmap Cell_pmap;
+  typedef typename Generator::Facet_pmap Facet_pmap;
+  typedef typename Generator::Facet_pmap_twice Facet_pmap_twice;
+  typedef typename Generator::Vertex_pmap Vertex_pmap;
+
+  Cell_pmap cell_pmap(c3t3);
+  Facet_pmap facet_pmap(c3t3,cell_pmap);
+  Facet_pmap_twice facet_pmap_twice(c3t3,cell_pmap);
+  Vertex_pmap vertex_pmap(c3t3,cell_pmap,facet_pmap);
+
+  output_to_medit_all_cells(os,
+                            c3t3,
+                            vertex_pmap,
+                            facet_pmap,
+                            cell_pmap,
+                            facet_pmap_twice,
+                            Generator().print_twice());
+
+#ifdef CGAL_MESH_3_IO_VERBOSE
+  std::cerr << "done.\n";
+#endif
+}
+
+// The function below reads a medit file and builds a triangulation from it.
+
+// -- If we're given every cell (finite & infinite), we can easily build
+//    the full triangulation with the same connectivity than the input !
+// -- If we're only given every finite cell, we can deduce the infinite cells and
+//    build the full triangulation
+// -- If we're only given the cells in complex, well... Since the cells in complex
+//    are usually not the convex hull of the points it's problematic and the input
+//    is remeshed (or rejected)
+
+// Side note on the numbering: since the vertex n° 0 is the infinite vertex,
+// we have to shift everything by one (but in the other direction than medit...)
+
+template<class Tr>
+void build_vertices(Tr& tr,
+                    const std::vector<typename Tr::Point>& points,
+                    std::vector<typename Tr::Vertex_handle>& vertex_handle_vector)
+{
+  typedef typename Tr::Vertex_handle            Vertex_handle;
+
+  vertex_handle_vector[0] = tr.tds().create_vertex(); // creates the infinite vertex
+  tr.infinite_vertex() = vertex_handle_vector[0];
+
+  // build vertices
+  for(std::size_t i=0; i<points.size(); ++i)
+  {
+    Vertex_handle vh = tr.tds().create_vertex();
+    vertex_handle_vector[i+1] = vh;
+    vh->set_point(points[i]);
+  }
+}
+
+template<class Tr>
+void add_facet_to_incident_cells_map(const typename Tr::Cell_handle c, int i,
+                                     std::map<std::set<typename Tr::Vertex_handle>,
+                                              std::vector<std::pair<typename Tr::Cell_handle,
+                                                                    int> > >& incident_cells_map)
+{
+  typedef typename Tr::Vertex_handle                            Vertex_handle;
+  typedef typename Tr::Cell_handle                              Cell_handle;
+  typedef std::set<Vertex_handle>                               Facet;
+  typedef std::pair<Cell_handle, int>                           Incident_cell;
+  typedef std::map<Facet, std::vector<Incident_cell> >          Incident_cells_map;
+
+  // the opposite vertex of f in c is i
+  Facet f;
+  f.insert(c->vertex((i + 1) % 4));
+  f.insert(c->vertex((i + 2) % 4));
+  f.insert(c->vertex((i + 3) % 4));
+  CGAL_precondition(f.size() == 3);
+
+  Incident_cell e = std::make_pair(c, i);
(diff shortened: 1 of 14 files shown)


Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -699,6 +700,7 @@ void Viewer::attribBuffers(int program_name) const {
     case PROGRAM_INSTANCED:
     case PROGRAM_CUTPLANE_SPHERES:
     case PROGRAM_SPHERES:
+      case PROGRAM_C3T3_TETS:
```
lrineau: Is not there an indentation issue, here?

Polyhedron/demo/Polyhedron/Scene_c3t3_item.h
```diff
@@ -39,6 +39,28 @@ using namespace CGAL::Three;
...
+    NB_VERTICES,
+    NB_TETS,
+    SMALLEST_RAD_RAD,
+    SMALLEST_EDGE_RAD,
+    BIGGEST_VL3_CUBE,
+    NB_SUBDOMAINS
+  };
+  bool has_stats()const {return true;}
+  QString computeStats(int type);
+  CGAL::Three::Scene_item::Header_data header() const;
```
lrineau: Those three new methods of `Scene_c3t3_item`, are they virtual methods?

maxGimeno: Yes.

Polyhedron/demo/Polyhedron/Scene_c3t3_item.h
```diff
@@ -39,6 +39,28 @@ using namespace CGAL::Three;
...
+    MEAN_DIHEDRAL_ANGLE,
+    NB_SPHERES,
+    NB_CNC,
+    NB_VERTICES,
+    NB_TETS,
+    SMALLEST_RAD_RAD,
+    SMALLEST_EDGE_RAD,
+    BIGGEST_VL3_CUBE,
+    NB_SUBDOMAINS
+  };
```
lrineau: That enum is part of the implementation internals of the item. It should be in the `.cpp` and not in the `.h`, to avoid a recompilation of all dependent code if that list changes.

lrineau:
- There is now a conflict with `master`.
- The [Travis job "CHECK" showed an issue with licenses](https://travis-ci.org/CGAL/cgal/jobs/198812253), probably about the new header file added by this PR.

MaelRL:
Some remarks:

- The file tet_soup_to_c3t3.h contains what you need but also contains too much and should be trimmed. There is no need to keep the very specific output functions (`output_all_cells()` and others). If they are useful or interesting to put in general (which they didn't seem to me), they should be moved somewhere else (File_medit.h or something like that) and probably another PR.

- The `border_info` container does not make sense here; should be removed.

- The "main" function can be cleaned because we will never pass infinite cells to that function or will never attempt to use Mesh_3 to rebuild a complete complex from incomplete information.

- There are a few easy optimizations to do in that file too with unordered maps and sets.

- All the `.mesh` keywords should be tested and their data discarded (or maybe even used ?)

- The assumptions in my file about the reference ('r' if a tetrahedron is described by the line `n1 n2 n3 n4 r`) make sense because I had created the files that I was reading. You probably can't deduce anything from that number in a general case (esp not that they are in the complex if they are positive, for example).

janetournois:
The c3t3 builder crashes on a `.mesh` file I'm trying to read. I'm trying to figure out why

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -1,6 +1,9 @@
 #ifndef CGAL_MESH_3_TET_SOUP_TO_C3T3_H
 #define CGAL_MESH_3_TET_SOUP_TO_C3T3_H
 
+#include <CGAL/license/Mesh_3.h>
+
```
lrineau: In addition this new file should also have the usual copyright+license preamble, like any other header file include `include/CGAL/`.

janetournois:
@lrineau Do you think the file `test_soup_to_c3t3.h` should be in `Mesh_3/include/CGAL/Mesh_3`? I'd rather move it to `Mesh_3/include/CGAL/IO`, or `Mesh_3/include/CGAL/internal`, don't you think?

maxGimeno:
@MaelRL 

> All the .mesh keywords should be tested and their data discarded (or maybe even used ?)

From what I see in the code, we only use what we need without implying any order. Why should we test information we won't use ?

lrineau:
@janetournois, https://github.com/CGAL/cgal/pull/1885#issuecomment-277723238:
> @lrineau Do you think the file test_soup_to_c3t3.h should be in Mesh_3/include/CGAL/Mesh_3? I'd rather move it to Mesh_3/include/CGAL/IO, or Mesh_3/include/CGAL/internal, don't you think?

I do not know, yet. It depends if the file contains functions that users may want to use or not. If it is about internals of Mesh_3, then both `include/CGAL/Mesh_3/` or `include/CGAL/internal/Mesh_3/` are fine: Mesh_3 was started before the guidelines about `include/CGAL/internal` were formalized.

MaelRL:
Yes, sorry, the `while(is >> word && word != "End")` should do the job of reading through unmatched keywords. Maybe it's still worth checking out to see if some are interesting to display (I think there is something that corresponds to "feature edge" for example).

lrineau:
Mael, in https://github.com/CGAL/cgal/pull/1885#issuecomment-277727231:
> Maybe it's still worth checking out to see if some are interesting to display (I think there is something that corresponds to "feature edge" for example).

Let's not complicate the subject. There are already enough comments that Maxime has a lot to do. Let's improve the current version, with the same feature set, and then after we can try to add features to the code.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -304,8 +304,6 @@ void build_finite_cells(Tr& tr,
     Cell_handle c = tr.tds().create_cell(vs[0], vs[1], vs[2], vs[3]);
     c->info() = tet[4]; // the reference encodes the interior/exterior info
 
-    // the reference must either be -1 for finite interior or > 0 for subdomains
-    CGAL_precondition(tet[4] != 0);
```
lrineau: The code of `Scene_c3t3_item` will crash if there is a reference material <= 0.

janetournois: ok sorry, I'm fixing it

janetournois:
I juste made the following experiment :
- open elephant.off
- mesh it with `Mesh_3` (default parameters in the widget)
- save it as `elephant.mesh`
- and try to reopen it.
It crashes. 
Can you please have a look @maxGimeno? Can you reproduce the same behaviour on linux?
Thanks

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -362,7 +362,7 @@ bool build_triangulation_from_file(std::istream& is,
 
   // grab the vertices
   int dim;
-  int nv, nf, ntet;
+  int nv, nf, ntet, useless;
```
janetournois: Please have a look at the `.mesh` writer to find out what these `useless` values actually are.
Then
- change their names,
- amend this commit (51bff2526e48faa14d757252075392fbd19a2e42) with the updated variable names, and a more explicit commit message
thanks

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -378,8 +379,8 @@ bool build_triangulation_from_file(std::istream& is,
       for(int i=0; i<nv; ++i)
       {
         double x,y,z;
-        is >> x >> y >> z;
-        points.push_back(Point_3(x,y,z));
+        is >> x >> y >> z >> w;
```
lrineau: In Medit files, the points do not have weights!

afabri: And in order to be fast for VC++ you should write `is >> iformat(x) >> ..."

janetournois:
I still have the following assertion
> Expression : f.size() == 3
> File       : ....../include/CGAL/Mesh_3/tet_soup_to_c3t3.h
> Line       : 57

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+#include <CGAL/assertions.h>
+
+#include <boost/array.hpp>
+#include <boost/unordered_set.hpp>
+#include <boost/unordered_map.hpp>
+#include <boost/foreach.hpp>
+
+// The purpose of this file is to rebuild manually a triangulation from a c3t3
+// output file (because generating a new triangulation every time is expensive)
+
```
MaelRL: Comment to be cleaned

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
+#ifndef CGAL_MESH_3_TET_SOUP_TO_C3T3_H
```
MaelRL: Header is missing

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+      vs[j] = vertex_handle_vector[tet[j] + 1];
+      CGAL_postcondition(vs[j] != Vertex_handle());
+    }
+
+    // this assertion also tests for degeneracy
+    CGAL_assertion(CGAL::orientation(vs[0]->point().point(), vs[1]->point().point(),
+        vs[2]->point().point(), vs[3]->point().point()) == POSITIVE);
+
+    Cell_handle c = tr.tds().create_cell(vs[0], vs[1], vs[2], vs[3]);
+    c->info() = tet[4]; // the reference encodes the interior/exterior info
```
MaelRL: The comment does not make sense

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+
+    // the reference must either be -1 for finite interior or > 0 for subdomains
```
MaelRL: The comment does not make sense anymore

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+                          std::vector<std::pair<typename Tr::Cell_handle,
+                          int> > >& incident_cells_map)
+{
+  typedef boost::array<int, 4>                            Tet; // 4 ids
+
+  typedef typename Tr::Vertex_handle                      Vertex_handle;
+  typedef typename Tr::Cell_handle                        Cell_handle;
+  typedef std::set<Vertex_handle>                         Facet;
+  typedef std::pair<Cell_handle, int>                     Incident_cell;
+  typedef std::map<Facet, std::vector<Incident_cell> >    Incident_cells_map;
```
MaelRL: Should be an unordered map

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+
+  // build the infinite cells if provided
```
MaelRL: There is no reading of infinite cells anymore so the whole `for()` can be removed

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+    typename Incident_cells_map::iterator it = incident_cells_map.find(f);
+    CGAL_assertion(it != incident_cells_map.end());
+    (it->second).push_back(std::make_pair(c, inf_pos));
+
+    // the three infinite facets
+    add_infinite_facets_to_incident_cells_map<Tr>(c, inf_pos, incident_cells_map);
+  }
+
+  // manually build the infinite cells if they were not provided
+  if(infinite_cells.empty())
```
MaelRL: This `if` will always be true

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    if(word == "Triangles")
+    {
+      is >> nf;
+      for(int i=0; i<nf; ++i)
+      {
+        int n1, n2, n3, surface_patch_id;
+        is >> n1 >> n2 >> n3 >> surface_patch_id;
+        // no use for boundary facets for now...
```
MaelRL: Comment is apparently not true anymore

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+        t[3] = n3 - 1;
+        t[4] = reference;
+        finite_cells.push_back(t);
+      }
+    }
+  }
+
+
+
+  if(!finite_exterior_cells_counter)
```
MaelRL: The variable and this test are unused and should be cleaned

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+
+    // this assertion also tests for degeneracy
+    CGAL_assertion(CGAL::orientation(vs[0]->point().point(), vs[1]->point().point(),
+        vs[2]->point().point(), vs[3]->point().point()) == POSITIVE);
+
+    Cell_handle c = tr.tds().create_cell(vs[0], vs[1], vs[2], vs[3]);
+    c->info() = tet[4]; // the reference encodes the interior/exterior info
+
+    // the reference must either be -1 for finite interior or > 0 for subdomains
+    CGAL_precondition(tet[4] != 0);
```
MaelRL: This does not make sense. The reference in your input file can be anything, including 0. If the scene item cannot use a cell with info 0, then it should just be discarded.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
lrineau: Actually the comment is wrong. In a Medit file, there is no tetrahedron that is "exterior". All the tetrahedron listed in the file are in the mesh. The "reference" of cells is just another name for the "subdomain index". Different software, different wording.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
lrineau: This precondition should be replaced by `CGAL_assertion(tet[4]>0);`. If the reference is 0, then there will be something that will crash afterward in the C3t3.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
MaelRL: It's not wrong, my old code used medit references to distinguish interior and exterior finite cells. It just doesn't make sense in this new use and must be removed.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
MaelRL: Then it should be replaced by some arbitrary positive value rather than making files that decided to use negative references unreadable.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+
+
+
+  if(!finite_exterior_cells_counter)
+  {
+    // The finite interior cells MUST be the convex hulls of the point
+    std::cerr << "WARNING: no finite exterior cell provided..." << std::endl;
+  }
+
+  CGAL_precondition(!finite_cells.empty());
```
janetournois: replace this precondition by a pop-up message "your c3t3 is empty", and return `false`

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
lrineau: In the source code of medit, the sign bit of the `ref` integer is cleared, so that `ref` variable is supposed to be >=0. I wonder if the ref 0 is a valid one.

The source code also shows that the ref 0 is a special one, not associated to any material.

I think we can assume the ref of triangles are > 0.

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
lrineau: You cannot have a popup menu in a header file like this one. How are errors handled, in the loading function?

janetournois: Right. This function should return false, and the loader check if the number of cells is empty. In this case, show a popup

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -0,0 +1,447 @@
...
+  typedef std::set<Vertex_handle>               Facet;
+
+  // associate to a face the two (at most) incident tets and the id of the face in the cell
+  typedef std::pair<Cell_handle, int>                   Incident_cell;
+  typedef std::map<Facet, std::vector<Incident_cell> >  Incident_cells_map;
+
+  Incident_cells_map incident_cells_map;
+  std::vector<Vertex_handle> vertex_handle_vector(points.size() + 1); // id to vertex_handle
+
+  CGAL_precondition(!points.empty() && !finite_cells.empty());
```
janetournois: `finite_cells` can be empty, and should better trigger a warning message and load only the points

maxGimeno: That is in opposition with your previous comment and why load points that cannot be displayed ?

maxGimeno: I find it strange that the treatment for empty finite cells differ in the two  functions.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -245,6 +245,9 @@ void Scene_edit_polyhedron_item_priv::initializeBuffers(CGAL::Three::Viewer_inte
         item->buffers[Facet_normals].release();
         item->vaos[Facets]->release();
         program->release();
+
+        if(offset.norm() !=0)
+          delete vertices;
```
lrineau: 1. I do not understand why this PR about the loading of c3t3 items modifies this file in cabf3fb0dbb98f8671c620779c81994de31070d3.
2. Those lines are compilation error:
```
test/Polyhedron_Demo/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp:250:11: 
error: cannot delete expression of type 'std::vector<GLdouble>'
          delete vertices;
          ^      ~~~~~~~~
1 error generated.
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-122/Polyhedron_Demo/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.0_Release.gz

maxGimeno: How the heck did that come back here ? I am sure I removed it with a rebase !

lrineau:
@janetournois:
- Is the implementation of "shrink the tetrahedra" conform to what you expect?
- What about the other feature "be able to provide a file with predefined colors for subdomains" of issue #1851?

maxGimeno:
@lrineau I discussed  "be able to provide a file with predefined colors for subdomains" with @MaelRL and he told me he had no use for it and I shouldn't waste time to implement it until someone has a real use for it.

janetournois:
@MaelRL I forward @lrineau 's question : 
>> - Is the implementation of "shrink the tetrahedra" conform to what you expect?

MaelRL:
It seemed to work well for the surface of the mesh, yes. I think it was not completely finished with respect to shrinking while having a cut plane and shrinking tetrahedra, so I don't know about that.

janetournois:
There are two bugs left
- the bounding box of the loaded mesh is not set properly. Sometimes the mesh is very-very small, and sometimes it's not even in the window (I don't know if it's too small or too far)
- once a `.mesh` file is loaded to a `c3t3_item`, the "Reload item from file" right-click operation makes the demo crash

PR #1855, opened 2017-01-25
Kernel: Add overloads of squared_distance()

afabri:
## Summary of Changes

We add overloads for Point_3/Segment_3 and Point_3/Triangle_3 distance computation, that tells us if the point is on a vertex, an edge or in the interioir.

## Release Management

* Affected package(s): Kernel_23, Distance
* Issue(s) solved (if any):
* Feature/Small Feature (if any): [Kernel::squared_distance_with_dim_and_index](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::squared_distance_with_dim_and_index)

PR #1887, opened 2017-02-07
Aabb tree add closest point location and primitive

sloriot:
## Summary of Changes

Add a function in the AABB-tree that in addition of the closest primitives can provide more information on where on the closest primitive the closest point is. For example in a triangle mesh, we would know if the closest point is on a vertex, an edge or inside a face of the mesh.

## Release Management

* Affected package(s): `AABB_tree`, `Kernel_23`
* Feature/Small Feature (if any): TODO
* Replaces:  #1855

### DIFF
--- a/AABB_tree/doc/AABB_tree/Concepts/AABBGeomTraits.h
+++ b/AABB_tree/doc/AABB_tree/Concepts/AABBGeomTraits.h
@@ -71,6 +71,14 @@ A functor object to compute the point on a geometric primitive which is closest 
 typedef unspecified_type Construct_projected_point_3;
 
 /*!
+A functor object to compute the point and location on a geometric primitive which is closest from a query.
+It must support the `result_of` protocol and provides the operator:
+`operator()(const Type_2& type_2, const Point_3& p);` where `Type_2` is any type among `Segment_3` and `Triangle_3`.
+The operator returns an object which is explicitly convertible to `Point_3` and that represents the point on `type_2` which is closest to `p`.
+*/
+typedef unspecified_type Construct_projected_point_and_location_3;
+
+/*!
 A functor object to compare the distance of two points wrt a third one.
 Provides the operator:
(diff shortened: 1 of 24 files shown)


sloriot:
Current documentation issue:
The function `closest_point_location_and_primitive()` requires an initialization of the hint. The way `AABB_tree::Point_location` is documented does not imposed anything on the return type of  `AABBTraits::Closest_point` (for the second overload) but to be explicitly convertible to `Point_3`. For now the code set a `dimension` and a `index` members that are not documented. Also in `AABB_traversal_traits.h`,  `Projection_traits_point_and_location::intersection()` the update of the closest primitive is also using these undocumented members.

sloriot:
Code issue
`closest_point_and_primitive()` uses a construction for getting the closest primitive in the traversal traits. If we want to guarantee that the primitive returned is the closest with a kernel with exact predicates only (even if the projection is approximated) we need to use a predicate that relies on an exact projection or that uses only the current closest primitive. This issue is not specific to this PR but since this PR introduces a new similar function, a solution should be found for both. Something to take into account is also to avoid penalizing a usage where an approximated closest primitive is fine (having several versions of the function?)

sloriot:
Code issue:
The functor `Construct_projected_point_and_location_3` is a construction and require a kernel with exact constructions. It is used in the `AABB_tree` in the function `closest_point_location_and_primitive()` when using a CGAL Kernel, which in practice is always a kernel with exact predicates but inexact constructions.

lrineau:
@afabri @sloriot:

 That pull-request is related to what we discussed today, and the work seems stalled (nothing changed since mi-February).

sloriot:
The purpose of the PR was to have a function to also get the location of the closest point but there are some problems that I listed above. The only relationship with what we discussed today is the sphere construction that is a side problem in this PR.

lrineau:
@sloriot Maybe this PR could be revived...

sloriot:
There is nothing new, I wrote a pretty complete summary of the situation but got no reaction.

lrineau:
Right. The important message is https://github.com/CGAL/cgal/pull/1887#issuecomment-277954928.

PR #1791, opened 2017-01-04
New package Point_set_3

sgiraudot:
[Full feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Point_set_3) reviewed as API ready.

PR #1890, opened 2017-02-08
Fix <CGAL/Surface_mesh/Properties.h> introduced by #1791

lrineau:
The PR #1791 introduced `CGAL::Point_set_3`, and factorized the
`Property_map` implementation of `Surface_mesh` into a new class
template `CGAL::Properties::Property_map<Key, T>`.

The backward compatibility was supposed to be ensured by the fact that
`Surface_mesh<Point>::Property_map<Key,T>` had
`Properties::Property_map<Key, T>` as base class. But that compatibility
was only partial, and there could be ambiguities in the calls to `get`
and `put` with those property maps.

This commit/PR renamed `Properties::Property_map` to
`Properties::Property_map_base`, and added a third parameter, used as
the `Derived` class in the CRTP used by `boost::put_get_helper`. That
way, the `get` and `put` functions are defined directory on the real
class `Surface_mesh::Property_map<Key, T>` and not on its base class.

The same has been modified in `CGAL::Point_set_3`.

That makes the use of `Property_base_base` and `Property_container` a
little trickier, but that removes several ugly non-portable hacks, like
the `reinterpret_cast` that were used to convert a pointer to base class
of the pmap to the pointer the real pmap. Now that is a lot cleaner.

I have verified that the documentation is not modified. Actually, the
PR #1791 did modify the documentation of `Surface_mesh` (see the member
method `add_property_map` for example), and this PR fixes the situation:
the documentation of `Surface_mesh` is put back to the version in
CGAL-4.9.

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -86,15 +86,22 @@ public:
 
   class Index;
 
-  typedef typename Properties::Property_container<Index> Base;
+  typedef typename Properties::Property_container<Point_set, Index> Base;
 
   template <class Type>
   struct Property_map
-    : public Properties::Property_map<Index, Type>
+    : public Properties::Property_map_base<Index, Type, Property_map<Type> >
   {
+    typedef Properties::Property_map_base<Index, Type, Property_map<Type> > Base;
+    Property_map() : Base() {}
+    Property_map(const Base& pm): Base(pm) {}
   };
   typedef Property_map<Index> Index_map;
 
+  template <typename Key, typename T>
+  struct Get_property_map {
+    typedef Property_map<T> type;
+  };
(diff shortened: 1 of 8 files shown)


lrineau:
- Cc @MaelRL who detected the issue with the ambiguity on the calls to `get`/`put`.
- Cc @afabri, one of the main authors of Surface_mesher.

sgiraudot:
Actually, https://github.com/CGAL/cgal/pull/1527 was the PR that changed the documentation. And that was a mistake, the documentation shouldn't have been changed: the inherited `Property_map` in Surface Mesh was introduced in order not to break the API, it seems that I missed some changes that I should have reverted.

I'll check the code, but if the API is not modified and the dirty hacks were removed, I think this is PR definitely makes sense.

PR #1896, opened 2017-02-10
Polyhedron_demo: Fix the highlighting in the selection_plugin

maxGimeno:
## Summary of Changes
This PR fixes a bug that happened when the brush size was not 0 with the highlighting in the selection_plugin.
## Release Management


lrineau:
I wonder why the Travis task for this PR was canceled. Can we restart it?

maxGimeno:
I canceled it because it only touches the Plyhedorn_demo, which is not tested by travis, and I needed another PR to be tested by travis. You can restart it if you want but as I sayed, only demo files are affected by this PR, so travis won't detect anything.

lrineau:
That's right. Thanks for the explanation.

PR #1907, opened 2017-02-14
Fix TBB installation instructions in Manual

### LINKED ISSUE #1326: Intallation documentation: fix TTB-related instructions for non-Windows systems
It seems the [TBB installation instructions](http://doc.cgal.org/latest/Manual/installation.html#title56) are not correct for non-Windows systems. For example, the compiled libs are not in `bin` but in `lib`. Here are [more information coming from TBB documentation](https://www.threadingbuildingblocks.org/docs/help/tbb_userguide/Linux_OS.html). Could any Linux guy check all this and fix the documentation, please?

maxGimeno:
## Summary of Changes
Minor fixes in the documentation about TBB installation
## Release Management

* Issue(s) solved (if any): fix #1326

### DIFF
--- a/Documentation/doc/Documentation/Installation.txt
+++ b/Documentation/doc/Documentation/Installation.txt
@@ -1162,9 +1162,9 @@ environment variable. The environment variable `TBB_ARCH_PLATFORM=<arch>/<compil
 `<arch>` is `ia32` or `intel64`. `<compiler>` describes the Linux kernel, gcc version or Visual Studio version
 used. It should be set to what is used in `$TBBROOT/lib/<arch>`.
 
-The folder `TBBROOT/bin/<arch>/<compiler>` should be added to the `PATH` variable.
+For windows users, the folder `TBBROOT/bin/<arch>/<compiler>` should be added to the `PATH` variable.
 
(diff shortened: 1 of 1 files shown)


Documentation/doc/Documentation/Installation.txt
```diff
@@ -1162,9 +1162,9 @@ environment variable. The environment variable `TBB_ARCH_PLATFORM=<arch>/<compil
 `<arch>` is `ia32` or `intel64`. `<compiler>` describes the Linux kernel, gcc version or Visual Studio version
 used. It should be set to what is used in `$TBBROOT/lib/<arch>`.
 
-The folder `TBBROOT/bin/<arch>/<compiler>` should be added to the `PATH` variable.
+The folder `TBBROOT/lib/<arch>/<compiler>` should be added to the `PATH` variable.
```
lrineau: That does not make any sense! Nothing from `TBBROOT/lib` should be in the `PATH`. On Windows, the DLL (the runtime part of shared libraries) are in `TBBROOT/bin`, and DLLs are searched in the `PATH`. On non-Windows, as far as I know there is nothing to add in the `PATH`.

PR #1909, opened 2017-02-16
WLOP Bugfix: use radius2 in parallel version too

sgiraudot:
## Summary of Changes

WLOP gives different results in parallel and sequential versions. This is because the parallel version wrongly uses the radius instead of the squared radius. This PR fixes this.

## Release Management

* Affected package(s): Point Set processing


lrineau:
Is that a bug specific to `master` or was it already in CGAL-4.9? In the later case, you should rebase your patch onto `releases/CGAL-4.9-branch`, and we will modify the PR to apply it to `CGAL-4.9-branch`.

sgiraudot:
Yes it was. I rebased and changed the PR base.

PR #1916, opened 2017-02-20
Fix Expr computation when a Node in the tree has a large MSB bound

sloriot:
## Summary of Changes
When the lower bound of the Most Significant Bit (MSB) is too large or too small, the evaluation of ExprRep was not done.
From what I understand, the code is able to handling it.

This was reveled by the test added. What happen is that the lower bound of MSB, computed with:
```
  /// lower bound on MSB
  /** defined to be cel(lg(real value));
      ilogb(x) is floor(log_2(|x|)). 
      Also, ilogb(0) = -INT_MAX.  ilogb(NaN) = ilogb(+/-Inf) = INT_MAX */
  extLong lMSB() const {
    return extLong(ilogb(core_abs(fpVal)-maxAbs*ind*CORE_EPS));
  }
```
is `INT_MAX` because of the operation _4.0000000000000000 - 4.0000000000000009_ that is close to 0.
Then, without the current patch the approximation was not refined and an erroneous decision was taken when computing a determinant (`Gmpq` is used as ground truth here but `BigFloat` could have also been used).

I'm not an expert so if for example @luis4a0, @akobel, @pougetma, or @mhsaar  can comment on this patch, it would be great.

## Release Management

* Affected package(s): CGAL_Core

### DIFF
--- a/CGAL_Core/include/CGAL/CORE/ExprRep.h
+++ b/CGAL_Core/include/CGAL/CORE/ExprRep.h
@@ -1220,20 +1220,24 @@ void AddSubRep<Operator>::computeApproxValue(const extLong& relPrec,
     appValue() = first->getAppValue(relPrec, absPrec);
     return;
   }
-  if (lMSB() < EXTLONG_BIG && lMSB() > EXTLONG_SMALL) {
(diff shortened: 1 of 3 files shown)


akobel:
Late to the party, sorry. I don't feel very confident with CORE internals. But FWIW: I agree that extLong should cover the overflow handling. Hence, LGTM.

PR #1929, opened 2017-02-27
Optimizations for grid simplication

sgiraudot:
## Summary of Changes

Several optimizations for grid simplification:

* Replacing `std::floor()` with `static_cast<int>()` (with some adaptations necessary for negative numbers so that the result is the same)
* Not constructing `Point_3` objects for comparison
* Using `std::unordered_set` instead of `std::set` (I added a new header `<CGAL/unordered.h>` to choose either `std` or `boost` similarly to what is done with `array` - also including `unorderd_map`)

The total speed up is of 75% (from 12 to 3 seconds on my tests).

## Release Management

* Issue(s) solved (if any): 


STL_Extension/include/CGAL/unordered.h
```diff
@@ -0,0 +1,50 @@
...
+using std::unordered_set;
+using std::unordered_map;
+#else
+using boost::unordered_set;
+using boost::unordered_map;
+#endif
+
+} // cpp11
+
+namespace cpp0x = cpp11;
```
lrineau: I would add that line. `CGAL::cpp0x::unordered_set` has never existed in previous versions of CGAL. No need to add a "compatibility".

lrineau:
Simon, are you confident enough to push that change into CGAL-4.10 instead of CGAL-4.11?

The speedup (300%, not 75%) is rather impressive. Users of CGAL-4.10 could benefit from it. I wonder what are the risks of regressions. I think that, if we want this PR into CGAL-4.10, we should push it also in CGAL-4.10-beta1 (and that means delay that beta release *again*).

Installation/include/CGAL/config.h
```diff
@@ -152,6 +152,10 @@
     defined(BOOST_NO_CXX11_HDR_ARRAY) || BOOST_VERSION < 104000
 #define CGAL_CFG_NO_CPP0X_ARRAY 1
 #endif
+#if defined(BOOST_NO_CXX11_HDR_UNORDERED_SET) || \
+    defined(BOOST_NO_CXX11_HDR_UNORDERED_MAP)
+#define CGAL_CFG_NO_CPP0X_UNORDERED 1
+#endif
```
sloriot: Be careful that in boost 1.48 the macro seems to be named `BOOST_NO_0X_HDR_UNORDERED_SET`

sloriot:
I would not target `4.10`. The patch implies that coordinates of the point are hashable.

Point_set_processing_3/include/CGAL/grid_simplify_point_set.h
```diff
@@ -44,11 +45,26 @@ namespace CGAL {
...
 
-/// Utility class for grid_simplify_point_set():
-/// Less_epsilon_points_3 defines a 3D points order / 2 points are equal
-/// iff they belong to the same cell of a grid of cell size = epsilon.
+// Round number to multiples of epsilon
+inline int round_epsilon(double value, double epsilon)
+{
+  double x = value / epsilon;
+  if (x > 0.)
+    return static_cast<int> (x);
```
sloriot: If epsilon is small, don't you have a risk of being affected by overflow?

sgiraudot: Indeed… one first step would be to replace that by `ptrdiff_t` to have the largest range possible, but the problem remains. In practice, I doubt this will ever happen in real cases. Maybe a solution would be to add a fallback mode with a define CGAL_GRID_SIMPLIFY_USE_ROUND or something?

Installation/include/CGAL/config.h
lrineau: [Sébastien is right][*]. You should add 
```C++
                               || \
BOOST_NO_0X_HDR_UNORDERED_MAP  || \
BOOST_NO_0X_HDR_UNORDERED_SET
```

  [*]: http://www.boost.org/doc/libs/1_48_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.macros_that_describe_possible_c__0x_features

sgiraudot:
> The patch implies that coordinates of the point are hashable, which would not be OK with EPECK for example.

I'm not sure I understand this correctly, but the hash function relies on the computation of integer rounded coordinates that were _already_ here in this package, so I don't think this breaks anything (most probably grid simplification already does not work with EPECK but I haven't checked).

sgiraudot:
Ok, I added the missing boost macros and revert the change from `std::floor` to casting to integer. Turns out this optimization was minor compared to the rest. My test that takes 12s with the master version took 3s with my version and 3.2s without this casting optimization. So it's not worth risking an integer overflow, I've reverted.

sloriot:
If it did not work with EPECK, then it is indeed not breaking anything.

lrineau:
It is merged in `master`, now. Let's hope nothing will break.

PR #1936, opened 2017-02-28
Mesh_3 Refer to the example that allows to read DICOM files

afabri:
## Summary of Changes

Tell that there is also an example that allows to read DICOM files.
@sloriot Is there a way to create a link to the example file? 

## Release Management

* Affected package(s): Mesh_3  but only the doucmentation


afabri:
It works. See [doc testsuite](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.10-Ic-142/output/Mesh_3/index.html#Chapter_3D_Mesh_Generation)

PR #1939, opened 2017-02-28
Fix non-determinism!

lrineau:
Fix an old non-determinism bug in
`<CGAL/Mesh_3/Detect_features_in_polyhedra.h>`.


lrineau:
Cc @janetournois @afabri

Issue #1931, opened 2017-02-27
Polyhedron demo: port plugins to use Surface_mesh


lrineau:
We would like to port plugin of the "polyhedron demo" to use the BGL API and the `Surface_mesh` class instead of `Polyhedron_3`.

The rational (taken from #1932): 
> We should use the Scene_surface_mesh_item where it is possible, as it is more efficient in terms of memory consumption. We should integrate the plugins by package, to dispatch tests and to avoid a too important difference with the current master branch.

- [x] AABB_tree (#1953)
- [ ] Convex_decomposition
- [ ] Convex_hull
- [x] IO
- [ ] Mesh_2
- [ ] Mesh_3
- [x] Operations_on_polyhedra (#1941)
- [ ] PCA
- [ ] PMP
- [ ] Point_set
- [ ] Subdivision_methods
- [ ] Surface_mesh
- [ ] Surface_mesh_deformation

Here is the list of files that include `Scene_polyhedron_item.h` *and* do not include `Scene_surface_mesh_item.h`:

  - [ ] `Scene_c3t3_item.h`
  - [ ] `Scene_nef_polyhedron_item.cpp`
  - [ ] `Scene_polygon_soup_item.cpp`
  - [ ] `Scene_polyhedron_item.cpp`
  - [ ] `Plugins/AABB_tree/Cut_plugin.cpp`
  - [ ] `Plugins/PCA/Create_bbox_mesh_plugin.cpp`
  - [ ] `Plugins/PCA/Pca_plugin.cpp`
  - [ ] `Plugins/PCA/Scene_polyhedron_transform_item.h`
  - [ ] `Plugins/PCA/Edit_box_plugin.cpp`
  - [ ] `Plugins/Mesh_3/Detect_sharp_edges_plugin.cpp`
  - [ ] `Plugins/Mesh_3/Optimization_plugin_cgal_code.cpp`
  - [ ] `Plugins/Mesh_3/Mesh_3_plugin.cpp`
  - [ ] `Plugins/IO/OFF_io_plugin.cpp`
  - [ ] `Plugins/IO/GOCAD_io_plugin.cpp`
  - [ ] `Plugins/IO/Surf_io_plugin.cpp`
  - [ ] `Plugins/IO/STL_io_plugin.cpp`
  - [ ] `Plugins/IO/VTK_io_plugin.cpp`
  - [ ] `Plugins/Convex_decomposition/Nef_plugin.cpp`
[...]

PR #1946, opened 2017-03-02
Polyhedron_demo: Scene_facegraph_interface

maxGimeno:
This PR is the first step to use Scene_surface_mesh_item in the Polyhedron_demo. ( #1931 )
## Summary of Changes
- Adds an interface that gives a function to return a facegraph
- Make Scene_polyhedron_item and Scene_surface_mesh_item inherit from this interface

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
+++ b/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
@@ -3,6 +3,7 @@
 
 #include "Scene_polyhedron_item_config.h"
 #include  <CGAL/Three/Scene_item.h>
+#include  <CGAL/Three/Scene_facegraph_interface.h>
 #include  <CGAL/Three/TextRenderer.h>
 #include "Polyhedron_type_fwd.h"
 #include "Polyhedron_type.h"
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
```diff
@@ -82,6 +85,9 @@ class SCENE_POLYHEDRON_ITEM_EXPORT Scene_polyhedron_item
     Polyhedron*       polyhedron();
     const Polyhedron* polyhedron() const;
 
+    Polyhedron* facegraph(){return polyhedron();}
```
lrineau: Either you chose Qt naming guidelines: `faceGraph`, or you choose CGAL ones: `face_graph`.

maxGimeno:
After testing, I realize this interface is useless. The fact that it is templated keeps me from factorizing the plugin's code further. The goal was to avoid checking if an item was a surface_mesh_item or a polyhedron_item, but I still have to do so to determine if the templated type of the interface should be SM or Polyhedron. Therefore I close the PR and delete the branch.

Issue #1928, opened 2017-02-23
Selection of duplicated points is broken


sloriot:
In the polyhedron demo, the selection of duplicated points (using the context menu of the item) seems broken.

With the current [point set](https://gist.github.com/sloriot/652c2d951362de36f6ecd9a0bb0bcd3d), not all the 32 duplicated points are selected.

afabri:
@sgiraudot  can you fix that please.

PR #1951, opened 2017-03-06
Polyhedron demo: fix duplicated points selection

sgiraudot:
## Summary of Changes

Iterators of Point_set_3 are invalidated when removing (selecting in the demo) one point. A remaining of the old API of the demo's Point_set_3 was producing a bug (not selecting all duplicate points). This is a fix for this.

## Release Management

* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/1928

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
@@ -467,9 +467,28 @@ void Scene_points_with_normal_item::resetSelection()
 void Scene_points_with_normal_item::selectDuplicates()
 {
   std::set<Kernel::Point_3> unique_points;
-  for (Point_set::iterator ptit = d->m_points->begin(); ptit!= d->m_points->end();++ptit )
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_points_with_normal_item.cpp
```diff
@@ -467,9 +467,28 @@ void Scene_points_with_normal_item::resetSelection()
...
+
+  if (selected.empty ())
+    {
+      d->m_points->unselect_all();
+    }
+  else
+    {
+      d->m_points->set_first_selected
+        (d->m_points->begin() + unselected.size());
+    }
```
lrineau: It seems you use a different coding style from the initial author (and probably different from most CGAL developers), where both the `{` and its content are indented.

1. Is that on purpose? In my opinion that coding style wastes to many horizontal spaces.
2. Anyway, where you modify an existing file that was not created by you, you should try to mimic the existing coding style.

sgiraudot:
Indentation problem is fixed.

PR #1952, opened 2017-03-06
Point set shape detection: Fix incomplete clear

sgiraudot:
## Summary of Changes

The `clear()` method of `CGAL::Efficient_RANSAC` states that _all internal structures are cleaned, including formerly detected shapes_, but in practice it does not clear the shape factories. It seems rather counter-intuitive: a user reported a bug on the `cgal-discuss` list (he was looping on a RANSAC object, clearing and adding a shape factory everytime, so the number of shape factories became higher and higher). This PR makes the `clear()` function also call `clear_shape_factories()`.

## Release Management

* Affected package(s): Point set shape detection
* Issue(s) solved (if any): (see `cgal-discuss`)

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
@@ -397,6 +397,7 @@ shape. The implementation follows \cgalCite{schnabel2007efficient}.
       m_num_available_points = m_num_total_points;
 
       clear_octrees();
+      clear_shape_factories();
     }
     /// @}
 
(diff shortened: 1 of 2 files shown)


lrineau:
As a bug fix of a released function/class, that should probably go into the `releases/CGAL-4.9-branch`.

sloriot:
The line `Point_set_shape_detection` is fully red

sgiraudot:
The bug should be fixed now.

sgiraudot:
There is still one warning but it is due to this PR: https://github.com/CGAL/cgal/pull/1964

sloriot:
@sgiraudot could this [error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-153/Point_set_shape_detection_3/TestReport_lrineau_ArchLinux.gz) be related to this PR?

sgiraudot:
I doubt it, the modified function `clear()` is never called in this test.

sloriot:
Indeed, this is a time-out.

sloriot:
You did not answer @lrineau : shall it go into 4.9?

sgiraudot:
I thought it was a statement, not a question ;)

The answer is yes.

PR #1954, opened 2017-03-06
Polyhedron_demo: Update Scene_surface_mesh_item

maxGimeno:
## Summary of Changes

- makes the Scene_surface_mesh_item able to be geometrically updated. (Without it, calling invalidateOpenGLBuffers will not re-compute the geometry)
- adds a default constructor to the item.
- adds a global mode for the facegraph_item default type. If this mode is Polyhedron(default), a plugin that outputs a facegraph_item from another type of item will produce a Scene_polyhedron_item. Else it will create a Scene_surface_mesh_item.

Almost all the PR fixing #1931 depend on this PR. It should be the first one to be integrated.
## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1354,7 +1354,6 @@ void MainWindow::writeSettings()
     else settings.remove("plugin_blacklist");
   //setting polyhedorn mode
   settings.setValue("polyhedron_mode", scene->isPolyhedronMode());
-  qDebug()<<"writing "<<scene->isPolyhedronMode();
 
   }
   std::cerr << "Write setting... done.\n";
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -254,8 +254,6 @@ MainWindow::MainWindow(QWidget* parent)
           viewer, SLOT(setFastDrawing(bool)));
   connect(ui->actionSwitchProjection, SIGNAL(toggled(bool)),
           viewer, SLOT(SetOrthoProjection(bool)));
-  connect(ui->actionPolyhedron_Mode, SIGNAL(toggled(bool)),
-          scene, SLOT(setPolyhedronMode(bool)));
 
   // add the "About CGAL..." and "About demo..." entries
   this->addAboutCGAL();
@@ -1338,6 +1336,8 @@ void MainWindow::readSettings()
     // read plugin blacklist
     QStringList blacklist=settings.value("plugin_blacklist",QStringList()).toStringList();
     Q_FOREACH(QString name,blacklist){ plugin_blacklist.insert(name); }
+    scene->setPolyhedronMode(settings.value("polyhedron_mode", false).toBool());
+    qDebug()<<"reading "<<scene->isPolyhedronMode();
 }
 
 void MainWindow::writeSettings()
@@ -1352,6 +1352,10 @@ void MainWindow::writeSettings()
     Q_FOREACH(QString name,plugin_blacklist){ blacklist << name; }
     if ( !blacklist.isEmpty() ) settings.setValue("plugin_blacklist",blacklist);
     else settings.remove("plugin_blacklist");
+  //setting polyhedorn mode
+  settings.setValue("polyhedron_mode", scene->isPolyhedronMode());
+  qDebug()<<"writing "<<scene->isPolyhedronMode();
+
   }
   std::cerr << "Write setting... done.\n";
 }
@@ -1619,7 +1623,9 @@ void MainWindow::on_actionPreferences_triggered()
   QDialog dialog(this);
   Ui::PreferencesDialog prefdiag;
   prefdiag.setupUi(&dialog);
-  
+  prefdiag.polyhedronMode_checkbox->setChecked(scene->isPolyhedronMode());
+  connect(prefdiag.polyhedronMode_checkbox, &QCheckBox::clicked,
+          scene, &Scene::setPolyhedronMode);
   
   QStandardItemModel* iStandardModel = new QStandardItemModel(this);
   //add blacklisted plugins
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1620,8 +1620,11 @@ void MainWindow::on_actionPreferences_triggered()
   QDialog dialog(this);
   Ui::PreferencesDialog prefdiag;
   prefdiag.setupUi(&dialog);
-  prefdiag.polyhedronMode_checkbox->setChecked(scene->property("is_polyhedron_mode").toBool());
-  connect(prefdiag.polyhedronMode_checkbox, &QCheckBox::clicked,
+  if(scene->property("is_polyhedron_mode").toBool())
+    prefdiag.polyRadioButton->setChecked(true);
+  else
+    prefdiag.smRadioButton->setChecked(true);
+  connect(prefdiag.polyRadioButton, &QRadioButton::toggled,
           this, &MainWindow::set_face_graph_default_type);
   
   QStandardItemModel* iStandardModel = new QStandardItemModel(this);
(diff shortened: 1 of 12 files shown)


Three/include/CGAL/Three/Scene_interface.h
```diff
@@ -168,7 +113,12 @@ class Scene_interface {
   //!
   //!@returns the index of the Item_B
   virtual Item_id selectionBindex() const = 0;
-
+  /*!
+   * \brief isPolyhedronMode determines if the global state is Polyhedron mode or Surface_mesh mode.
+   * @returns true if the global state is Polyhedron mode
+   * @returns false if the global state is Surface_mesh mode
+   */
```
lrineau: That documentation is completely useless. That is similar to:

```
/// Tells is Foo is true or not
bool isFoo() const;
```

The documentation should say what is changed by that global state.

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -87,6 +87,7 @@ struct Scene_surface_mesh_item_priv{
   mutable bool is_filled;
   mutable bool isinit;
   mutable std::vector<unsigned int> idx_data_;
+  mutable std::map<unsigned int, unsigned int> current_indices; //map im values to ghosts-free values
```
lrineau: Seems to be used only by `compute_elements()`. Why does it need to be a member of the item_priv object?

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
```diff
@@ -98,6 +99,9 @@ class SCENE_POLYHEDRON_ITEM_EXPORT Scene_polyhedron_item
     std::vector<QColor>& color_vector();
     void set_color_vector_read_only(bool on_off);
     bool is_color_vector_read_only();
+
+    void set_patch_id(Polyhedron::Face_handle f,int i) const;
```
lrineau: The basic design of the item polyhedron demo, at the beginning was to avoid have the CGAL API exposed in the items API. Why do you need `set_patch_id` at the level of the item?!

lrineau: It does not seem to be used anywhere. Please remove it.

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.h
```diff
@@ -29,15 +32,26 @@ class SCENE_SURFACE_MESH_ITEM_EXPORT Scene_surface_mesh_item
...
   typedef CGAL::Surface_mesh<Point> SMesh;
+  typedef SMesh FaceGraph;
   typedef boost::graph_traits<SMesh>::face_descriptor face_descriptor;
+  typedef boost::graph_traits<SMesh>::vertex_descriptor vertex_descriptor;
+  typedef boost::graph_traits<SMesh>::halfedge_descriptor halfedge_descriptor;
+  typedef SMesh::Property_map<vertex_descriptor,int> Vertex_selection_map;
+  typedef SMesh::Property_map<face_descriptor,int> Face_selection_map;
+  typedef SMesh::Property_map<halfedge_descriptor, bool> Halfedge_is_feature_map;
+  typedef SMesh::Property_map<face_descriptor, int> Face_patch_id_map;
+  typedef SMesh::Property_map<vertex_descriptor,int> Vertex_num_feature_edges_map;
```
lrineau: In my opinion, the header of the item should not defined any property map. it should be moved to a header `Surface_mesh_type.h`, that would be the CGAL header used by plugins needing to deal with the surface mesh as a CGAL object.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
maxGimeno: It will be used by incoming PR.

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.h
maxGimeno: Are there plugins that deal with Surface_mesh but no with Surface_mesh_item ?

lrineau: The point is reducing dependencies between files, decouple the interface and the implementation, and make compilations and recompilation shorter.

maxGimeno: What I don't understand is how including 2 headers instead of one makes less dependency and why having a header including another header makes the compilation quicker. I agree that it seems cleaner to separate the code dealing with Surface_mesh from the code specific for the item, like we did in Polyhedron, but how does it make anything quicker, as everywhere Surface_mesh_type.h will be inclluded, so will Scene_surface_mesh_item ?

lrineau: No `Surface_mesh_type.h` may not be always included. For a plugin with a lot of code, like the mesh_3_plugin, the part of the plugin that actually deal with CGAL code is in a dedicated .cpp files, and the main .cpp of the plugin only includes `Scene_polyhedron_item.h`, and not `Polyhedron_type.h`. That way, the main .cpp file of the plugin, that deals with the Qt user interface of the plugin, does not include the CGAL headers, but only Qt5 headers.

maxGimeno: Ok, I see it now. Thanks !

maxGimeno:
@lrineau : I think I made the changes you requested. If I did not forget anything, this PR should be ready for integration.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
```diff
@@ -24,6 +24,7 @@ class SCENE_POLYHEDRON_ITEM_EXPORT Scene_polyhedron_item
         : public CGAL::Three::Scene_item{
     Q_OBJECT
 public:
+  typedef Polyhedron FaceGraph;
     enum STATS {
       NB_VERTICES = 0,
```
lrineau: - In my opinion the naming convention should be the one of CGAL for types: `Face_graph`.
- What is more: should not you also add a function `Face_graph* face_graph()`?

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.h
```diff
@@ -49,10 +57,22 @@ class SCENE_SURFACE_MESH_ITEM_EXPORT Scene_surface_mesh_item
...
 
+  // Only needed for Scene_polyhedron_item
+  void setItemIsMulticolor(bool){}
+  void update_vertex_indices(){}
+  void update_halfedge_indices(){}
+  void update_facet_indices(){}
+
+  std::vector<QColor>& color_vector();
+  void set_patch_id(SMesh::Face_index f,int i)const;
+  int patch_id(SMesh::Face_index f)const;
```
lrineau: I have already said in my previous review that the item type should not have any function to get/set the `patch_id` of a face. That is too low-level for the items API: a plugin using a `Surface_mesh` should call `item->face_graph()` to get the pointer to the `Surface_mesh`, and then get/set the patch ids using the CGAL object, and not the item object, that should not know anything about CGAL.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.h
maxGimeno: Ther eis already the polyhedron() function that does that. In surface_mesh, the mesh is called SMesh, not Polyhedron, hence the need of this typedef, but polyhedron() exists for both, so there shouldn't be a need for a function called face_graph().

afabri: Why no just replacing `polyhedron()` with `face_graph()` ?

maxGimeno: That would be possible.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1961,5 +1961,8 @@ void MainWindow::set_face_graph_default_type(Face_graph_mode m)
 
 void MainWindow::set_facegraph_mode_adapter(bool b)
 {
-  this->setProperty("is_polyhedron_mode", b);
+  if(b)
+   set_face_graph_default_type(SURFACE_MESH);
+  else
+    set_face_graph_default_type(POLYHEDRON);
```
lrineau: The variable `b` should be renamed `polyhedron_mode`... and then you see the bug: you reversed the logic!

maxGimeno:
It should be ok now.

maxGimeno:
@lrineau are my changes OK or should I still modify something ?

lrineau:
This PR #1954 is now merged. Other PRs based on it can be rebased onto `master`.

PR #1960, opened 2017-03-08
Fix bug in split_graph_into_polylines

sloriot:
## Summary of Changes
The order used to sort vertex descriptors was not correct and ended up in some cases to identify the wrong vertex in the set. Using the default is fine since no vertex is removed.
See [_Iterator and Descriptor Stability/Invalidation_](http://www.boost.org/doc/libs/1_63_0/libs/graph/doc/adjacency_list.html), and more particularly the table at the end.

## Release Management

* Affected package(s): BGL

### DIFF
--- a/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
+++ b/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
@@ -51,6 +51,45 @@ struct Dummy_visitor_for_split_graph_into_polylines
 
 namespace internal {
 
+template <typename G_copy, typename Less_on_orig_vertex_descriptors>
+class Less_on_G_copy_vertex_descriptors {
+  const G_copy& g_copy;
+  const Less_on_orig_vertex_descriptors& less;
+public:
+  Less_on_G_copy_vertex_descriptors ( const G_copy& g_copy,
+    const Less_on_orig_vertex_descriptors& less)
+    : g_copy(g_copy), less(less) {}
+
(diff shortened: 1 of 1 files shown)


sloriot:
@lrineau  I think this is breaking a change you did recently to have some determinism in Mesh_3. Am I correct?

lrineau:
Bug introduced by #1524... I do not know how to fix it generally.

lrineau:
I mean... generically.

lrineau:
I introduced first a new "less" functor with #1524, and then I "fixed" it in #1752. The issue I was trying is solve is the non-determinism of the polyline extraction in Mesh_3. I do not know how to fix that issue generically, for the moment.

sloriot:
Actually the order was relying on incident edges and since edges was removed while the vertex was not, the set was no longer sorted and the wrong vertex was possibly removed

lrineau:
This PR conflicts with `master`.

sloriot:
@lrineau what action do you expect? It is not in conflict with the branch 4.9 it targets. I can't do anything but handle the conflict when merging 4.9 in master.

lrineau:
That was actually tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9.1-Ic-164.shtml


@sloriot, can you please handle the merge of `CGAL-4.9-branch` into `master`?

Actually I wonder if we should try the merge into `integration` first, before the conflict seems non-trivial to handle.

sloriot:
done

PR #1964, opened 2017-03-10
Fix iterator range size type

sgiraudot:
## Summary of Changes

Currently, our `CGAL::Iterator_range` has a method `size()` which returns a signed integer (`difference_type`). This is probably due to the fact that we use `std::distance()` to compute it, which does return a signed number.

However, it seems very counter-intuitive to have a signed size (a range where `begin()` is after `end()` souldn't be valid anyway) and it leads to warnings if a user does a comparison of `size()` with an unsigned integer (this is how I discovered this problem). In addition, the [boost::iterator_range](http://www.boost.org/doc/libs/1_55_0/libs/range/doc/html/range/reference/utilities/iterator_range.html) that we cloned (as the documentation says) uses an unsigned type.

This PR makes `CGAL::Iterator_range::size()` return a `std::size_t` number (with a static cast internally – I'm not sure if this should be indicated in the doc, as we indicate we are using `std::distance`).

## Release Management

* Affected package(s): STL_Extensions

### DIFF
--- a/Point_set_shape_detection_3/test/Point_set_shape_detection_3/test_regularization.cpp
+++ b/Point_set_shape_detection_3/test/Point_set_shape_detection_3/test_regularization.cpp
@@ -96,7 +96,7 @@ bool planes_differ (const Plane& a, const Plane& b)
           (std::fabs (a.d() - b.d()) > 1e-6));
 }
 
-void check_ransac_size (const Efficient_ransac& ransac, std::ptrdiff_t nb)
+void check_ransac_size (const Efficient_ransac& ransac, std::size_t nb)
 {
   if (ransac.shapes().size() != nb)
     {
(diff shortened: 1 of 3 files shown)


sgiraudot:
One warning in shape detection that I just fixed.

mglisse:
Note that the C++ committee generally considers that using unsigned types so much in the standard was a mistake, and they would make std::vector::size() return a signed type if they were not constrained by backwards compatibility.
(that doesn't mean this patch is wrong, it makes things more consistent)

lrineau:
@mglisse What is the rational to prefer a signed size?

mglisse:
(you can find links through google, I think http://stackoverflow.com/q/18795453/1918193 contains several of them)
The first 2 reasons that come to mind:

- it is less error-prone and more intuitive, i<j is the same as j-i>0, you can do a loop for(;i>=0;--i), etc.
- overflow is undefined behavior (not wrapping like unsigned): this is good for optimizations (the compiler can assume that overflow doesn't happen), but this is also good for debugging as it makes -fsanitize=signed-integer-overflow (included in -fsanitize=undefined) a perfectly valid behavior for the compiler.

PR #1975, opened 2017-03-16
Fix polylines reading

### LINKED ISSUE #1963: Improve loading of polylines
Right now it is painful to open a polyline file containing many polylines. For example opening 100000 segments takes ages. Also the std::cerr for each polyline should be removed.

We should have a look at what is going on. @maxGimeno, could you have a look?

maxGimeno:
## Summary of Changes

Makes the reading of polylines faster in the Polyhedron_demo

## Release Management

* Issue(s) solved (if any): fix #1963


sloriot:
Locally tested and it works at the speed of light compared to the previous version

lrineau:
Houra for this PR. The previous code was really inefficient!

PR #1981, opened 2017-03-21
Polyhedron demo : fix surf reader

janetournois:
## Summary of Changes
This PR intends to fix the behavior of the `.surf` reader when the input surface is non-manifold.
Because `CGAL::Polyhedron_3` cannot handle a non-manifold surface, the built polyhedron has holes wherever the builder tries to make a non-manifold configuration.

We use `PMP::orient_polygon_soup` and `PMP::polygon_soup_to_polygon_mesh` to build a valid polygon mesh, which can be a `CGAL::Polyhedron_3` with duplicated vertices on geometrically non-manifold edges/vertices.

## Release Management


janetournois:
note more patches need to be pushed before testing

lrineau:
Is this PR now ready?

janetournois:
It is. Unless we want to try to speed it up (?)

janetournois:
let's say it's ready for now. The speedups would be in the repair functions from PMP, so they would be made in other PR(s).

PR #1983, opened 2017-03-22
Polyhedron_demo: Plugins use SM: Second Step

### LINKED ISSUE #1974: Selecting faces when Polyhedron has degenerate faces
If a polyhedron has some degenerate faces, the selection of non-degenerate faces (picking and selection item) will fail as the AABB-tree will fail on those faces. A simple solution is to filter out those faces while constructing the tree.

maxGimeno:
## Summary of Changes
The selection_plugin and edit_polyhedron_plugin are applicable to  Scene_surface_mesh_items as well as Scene_polyhedron_items. To do so, the Detect_features_in_polyhedra files have been moved in PMP and properties have been added to the Surface_mesh_type and Polyhedron_type of the demo.

The Mesh_3 plugin is also applicable to a Scene_surface_mesh_item. To do so, the Mesh_polyhedral_domain has been bgl-ized.

This PR also fixes a bug in the AABB_trees of the Scene_polyhedron_item and the Scene_surface_mesh_item that occurred when the graph has degenerated faces. 

## Release Management
* Many PR depend on this one.
* Based on PR #1954.
* Affected package(s):Polyhedron_demo, Mesh_3, Polygon_mesh_processing
* Issue(s) solved (if any): fix #1931 fix #1974


lrineau:
This pull-request modifies files from the BGL and Mesh_3 package. I would like to discussion the modifications with you, next week.

maxGimeno:
Did you test your fix for polyhedron ? I think all those property_maps and their get() should not be in Polyhedron_type.h but in properties_Polyhedron_3.h. For now, on my machine, polyhedron_item does not compile.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -964,53 +1357,120 @@ void Scene_edit_polyhedron_item::drawEdges(CGAL::Three::Viewer_interface* viewer
...
+                 bbox_program.setUniformValue("rotations", f_mat);
+                 bbox_program.setUniformValue("translation", vec);
+                 bbox_program.setUniformValue("translation_2", vec2);
+                 bbox_program.setUniformValue("mvp_matrix", mvp_mat);
+                 program->setAttributeValue("colors", QColor(255,0,0));
+                 viewer->glDrawArrays(GL_LINES, 0, static_cast<GLsizei>(nb_bbox/3));
+                 bbox_program.release();
+                 item->vaos[BBox]->release();
+            }
+       }
```
lrineau: The indentation of that whole block is not consistent with the rest of the file.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.h
```diff
@@ -31,93 +32,172 @@
...
+  Id_setter(Polyhedron* poly)
+    :poly(poly), sm(NULL){}
+
+    Id_setter(SMesh* sm)
+    :poly(NULL), sm(sm)
+  {
+    im = get(boost::vertex_index, *sm);
+  }
+
+    std::size_t get_id(p_vd vd)
```
lrineau: Indentation

Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "Scene_polygon_soup_item.h"
 #include "Scene_polyhedron_item.h"
+#include "Scene_surface_mesh_item.h"
```
lrineau: Why is that header included but never used?

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -862,6 +863,18 @@ invalidate_stats()
...
+Scene_polyhedron_item::Vertex_selection_map 
+Scene_polyhedron_item::vertex_selection_map()
+{
+  return get(boost::vertex_index,*d->poly);
+}
+
+Scene_polyhedron_item::Face_selection_map 
+Scene_polyhedron_item::face_selection_map()
+{
+  return get(boost::face_index,*d->poly);
```
lrineau: There are two property tag `vertex_selection` and `face_selection`. Is not there a bug here?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/properties.h
```diff
@@ -0,0 +1,36 @@
...
+#include <CGAL/license/Polygon_mesh_processing.h>
+
+namespace CGAL
+{
+enum face_patch_id_t            { face_patch_id };
+enum vertex_num_feature_edges_t { vertex_num_feature_edges };
+enum halfedge_is_feature_t      { halfedge_is_feature };
+enum vertex_selection_t         { vertex_selection};
+enum edge_selection_t           { edge_selection};
+enum face_selection_t           { face_selection};
```
lrineau: The name `face_selection` should be `facet_selection`. The dimension of a "face" in a polyhedron can be 0, 1, or 2. A facet is the face with dimension 2.

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -1,30 +1,68 @@
...
+  }
+public:
+  Id& id() { return m_it; }
+  const Id& id() const { return m_it; }
+  Datum& datum() { return m_datum; }
+  const Datum& datum() const { return m_datum; }
+
+  /// Returns a point on the primitive
+  Point reference_point() const { return m_datum.vertex(0); }
+};
```
lrineau: Looks like the generic primitive class `CGAL::AABB_face_graph_triangle_primitive` should have been used. @sloriot, can you comment?

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.h
```diff
@@ -29,17 +33,42 @@ class SCENE_SURFACE_MESH_ITEM_EXPORT Scene_surface_mesh_item
...
+  typedef SMesh FaceGraph;
   typedef boost::graph_traits<SMesh>::face_descriptor face_descriptor;
+  typedef boost::graph_traits<SMesh>::vertex_descriptor vertex_descriptor;
+  typedef boost::graph_traits<SMesh>::halfedge_descriptor halfedge_descriptor;
+  typedef SMesh::Property_map<vertex_descriptor,int> Vertex_selection_map;
+  typedef SMesh::Property_map<face_descriptor,int> Face_selection_map;
+  typedef SMesh::Property_map<halfedge_descriptor, bool> Halfedge_is_feature_map;
+  typedef SMesh::Property_map<face_descriptor, int> Face_patch_id_map;
+  typedef SMesh::Property_map<vertex_descriptor,int> Vertex_num_feature_edges_map;
+
```
lrineau: In the spirit of the organization of the demo, the header `Scene_surface_mesh_item.h` should be very lightweight, and should not include any CGAL header file (but forward declaration). That would avoid to recompile all plugins including it any time of those header is modified.

lrineau:
There are conflicts with `master` that prevent the testing of the PR by Travis, or by us in `integration`.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/properties.h
afabri: I don't remember why but for Surface_mesh and the BGL-API we have switched to face instead of facet.

Polyhedron/demo/Polyhedron/Plugins/PMP/Scene_facegraph_item_k_ring_selection.h
```diff
@@ -0,0 +1,667 @@
...
+  }
+
+  friend void put(FG_is_selected_edge_property_map map, edge_descriptor ed, bool b)
+  {
+    CGAL_assertion(map.is_selected_ptr!=NULL);
+    (*map.is_selected_ptr)[map.id(ed)]=b;
+  }
+};
+
+class SCENE_FACEGRAPH_ITEM_K_RING_SELECTION_EXPORT Scene_facegraph_item_k_ring_selection
```
maxGimeno: No it should not. If the class's name change, the moc files will change too, and we won't ever be able to compile both poly and SM verisons of the selection_plugin at the same time.

Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
```diff
@@ -65,7 +79,7 @@ class Polyhedron_demo_selection_plugin :
     Q_PLUGIN_METADATA(IID "com.geometryfactory.PolyhedronDemo.PluginInterface/1.0")
 public:
   bool applicable(QAction*) const { 
-    return qobject_cast<Scene_polyhedron_item*>(scene->item(scene->mainSelectionIndex()))
+    return qobject_cast<Scene_face_graph_item*>(scene->item(scene->mainSelectionIndex()))
```
maxGimeno: And again, same aswer about the moc.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Edit_polyhedron_plugin.cpp
```diff
@@ -83,9 +99,25 @@ QList<QAction*> Polyhedron_demo_edit_polyhedron_plugin::actions() const {
...
+        return true;
+      }
+    }
+    else
+    {
+      if (qobject_cast<Scene_facegraph_item*>(edit_item->sm_item()))
+      {
+        return true;
+      }
+    }
```
maxGimeno: Yes, it is because the Scene_selection_item is split into two targets, and so we also must have two targets for the edit_plugin, which was not conceived to do so. We therefore have two identical targets with two different names.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/properties.h
maxGimeno: Plus I think it would be weird to have facet_selection with face_descriptors, face_iterators, etc.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Detect_features_in_polyhedra.h
```diff
@@ -0,0 +1,305 @@
...
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+//
+//
+// Author(s)     : Laurent Rineau, Stephane Tayeb
```
maxGimeno: So we put it back in Mesh_3 ?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/properties.h
lrineau: I see that that point was raised during the review.

https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/BGL/1st_round

But the naming in existing packages of CGAL, at the time the "FaceGraph" was added, was not consistent.

Anyway:
- in a 2D planar graph (such as a 2D triangulation), that is common to call "face" the 2-faces,
- for a polyhedron, the correct term for 2-face is "facet", and "face" is for any dimension: vertices are 0-faces, edges are 1-faces, and so on...
- more generally, facets are faces of co-dimension 1: that means that their dimension is the ambient dimension minus 1,
- the naming for the BGL extension "FaceGraph" is ambiguous, because that concept does not imply any ambient dimension.

Let's forget that, and keep face, at it was decided during the review. At least that is consistent with the usual wording of a planar graph.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Detect_features_in_polyhedra.h
lrineau: Let's discuss the issue calmly before rushing into any solution. That will be a topic to be discussed on Monday with @afabri.

Polyhedron/demo/Polyhedron/Surface_mesh_type.h
```diff
@@ -1,13 +0,0 @@
...
-#include <CGAL/Polygon_mesh_processing/compute_normal.h>
-#include "triangulate_primitive.h"
-#include <CGAL/boost/graph/graph_traits_Surface_mesh.h>
-#include <CGAL/boost/graph/properties.h>
-#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
-#include <CGAL/Polygon_mesh_processing/properties.h>
-typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
-typedef Kernel::Point_3 Point;
-typedef CGAL::Surface_mesh<Point> SMesh;
-#endif // SURFACE_MESH_TYPE_H
```
lrineau: I do not understand why the commit is named "remove empty file". That file was not empty.

maxGimeno: Well it was supposed to be. I guess I didn't save when I deleted its content.

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -751,7 +751,17 @@ void clear(FaceGraph& g)
   CGAL_postcondition(num_faces(g) == 0);
 }
 
-
+/**
+ * \ingroup PkgBGLHelperFct
+ *
+ * returns the degree of the face with halfedge `hd`
+ *
+ * @tparam FaceGraph model of `MutableHalfedgeGraph` and `MutableFaceGraph`
```
lrineau: 1. Why mutable?
2. `FaceGraph` already refines `HalfedgeGraph`. No need to mention both: you can just mention `FaceGraph`.

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -214,13 +214,13 @@ public Q_SLOTS:
...
-   * If isPolyhedronMode is true, the default output type will be Polyhedron (obviously), and if not, it will
-   * be Surface_mesh.
+   * \brief set_face_graph_default_type sets the global state of the application to `Polyhedron mode` or `Surface_mesh mode`.
+   * In the application, a plugin that outputs a `Facegraph_item` will match the input type for the output.
+   * But when the input is not a `Facegraph_item`, the plugin will use this state to know what type to use.
+   * If `isPolyhedronMode()` is `true`, the default output type will be `Polyhedron` (obviously), and if not, it will
+   * be `Surface_mesh`.
    */
-  void setPolyhedronMode(bool b);
+  void set_face_graph_default_type(bool b);
```
lrineau: I did not say it because it was kind of obvious for me: one the name is changed to `set_face_graph_default_type`, then the argument type cannot not be a `bool`: it has to be an enum, with two values with names like `POLYHEDRON_3` and `SURFACE_MESH` (the names of the enumerators have to be uppercase, by CGAL coding conventions).

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.cpp
```diff
@@ -1041,7 +1049,10 @@ void Scene_edit_polyhedron_item_priv::remesh(Mesh* mesh)
   unsigned int nb_iter = ui_widget->remeshing_iterations_spinbox->value();
 
   std::cout << "Remeshing (target edge length = " << target_length <<")...";
-  Is_constrained_map<Mesh> icm(&constrained_vec, mesh);
+  SMesh::Property_map<sm_vertex_descriptor,int> smap;
+  std::map<poly_vertex_descriptor,int> polymap;
+
+  Is_constrained_map<Mesh> icm(&constrained_vec,&polymap, &smap,  mesh);
```
lrineau: That API is really strange. `smap` and `polymap` should remain internals of the property map. For the internals
 - of `Is_constrained_map<Poiyhedron>`, you can use `std::shared_ptr<std::map<poly_vertex_descriptor,int>>` and the corresponding `std::make_shared`, to let C++ deal with the handling of memory,
 - and for `Is_constrained_map<SMesh>`, you could also use `std::shared_ptr<SMesh::Property_map<sm_vertex_descriptor,int> >` and create the shared pointer with a lambda-function as the `Deleter`, that would execute `sm_item->polyhedron()->remove_property_map(smap)` once that resource is not longer used.

Polyhedron/demo/Polyhedron/Plugins/PMP/Scene_facegraph_item_k_ring_selection.h
lrineau: I thought about it and that ODR problem is a real problem: the two different classes with the same name even share the same meta-object. That means that `qobject_cast` will be messed-up.

I really think that is important. And about the MOC files, the best would be to let AUTOMOC produce the right files, instead of including a `.moc` file directly.

maxGimeno: qobject_cast is not messed-up, as every plugin that needs is only linked to one and only version of the file. This way there is no conflict. It means the sm version of a plugin will only know the sm version of the sm_item_k_ring_selection, and the poly version will only know the poly_item_k_ring_selection. I agree this is not very clean and this is not easy to decypher, but it works and rethinking the strategy of double targets will take ages and mess-up half a dozen PRs.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Detect_features_in_polyhedra.h
afabri: @palliez just confirmed me by mail that the code is small and untechnical enough to be moved to the PMP package.

afabri: @klrineau: Note that there is already a dependency on PMP for normal computations.

Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
```diff
@@ -285,15 +299,18 @@ public Q_SLOTS:
...
     if(!poly_item) {
       print_message("Error: there is no selected polyhedron item!");
       return; 
     }
     // all other arrangements (putting inside selection_item_map), setting names etc,
     // other params (e.g. k_ring) will be set inside new_item_created
+    std::cerr << "B"<< std::endl;
     Scene_polyhedron_selection_item* new_item = new Scene_polyhedron_selection_item(poly_item, mw);
     ui_widget.modeBox->setCurrentIndex(last_mode);
+    std::cerr << "C"<< std::endl;
```
afabri: oops. I just spoted the std::cerr << "A"  etc.  I think I accidentally committed this debug output.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
afabri: In the original code the selection status was stored in the index which is a writable property of the enriched polyhedron type. So I just sticked to the index as I was not sure if there are any side effects if we write in the selection property.

afabri: In fact in the enriched polyhedron is no additional storage for selection and the index IS the storage for selection.

Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
afabri: I just fixed this.

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
sloriot: Indeed.

maxGimeno: I need to triangulate the primitives, so I cannot use the default type.

sloriot: after discussing with @maxGimeno it turns out that its a special primitive to handle non-triangulated meshes so a dedicated one is needed. It can probably be shared with polyhedron.

lrineau:
@maxGimeno @afabri Can one of you rewrite the title and the first message of this PR?

And what is the relationship with #2023?

maxGimeno:
@lrineau #2023 is based on this PR.

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -751,6 +751,20 @@ void clear(FaceGraph& g)
...
+face_degree(typename boost::graph_traits<FaceGraph>::halfedge_descriptor hd, const FaceGraph& g)
+{
+  typename boost::graph_traits<FaceGraph>::degree_size_type res = 0;
+  typename boost::graph_traits<FaceGraph>::halfedge_descriptor done = hd;
+  do {
+    ++res;
+    hd = next(hd,g);
+  }while(! (hd == done));
+  return res;
+}
```
lrineau: > Should not this function be documented, similarly to CGAL::clear(..)?

@afabri I have asked that question three weeks ago. No answer, no modification, no progress!...

Mesh_3/include/CGAL/Mesh_polyhedron_3.h
```diff
@@ -215,6 +217,117 @@ struct Mesh_polyhedron_3
   typedef type Type;
 };
 
+
+template <typename Gt, typename Patch_id>
+struct Patch_id_pmap {
```
lrineau: @maxGimeno: There are two `Patch_id_pmap` in this PR, and I have already worked with you, in another PR (the one about `Surface_mesh` and `Mesh_3`) to unify all those property maps into a better reusable implementation.

BGL/include/CGAL/boost/graph/helpers.h
afabri: In fact it should be called just `degree()`. I think we did it call this way and had a problem with sfinae. Note that the PR #2101 introduces implementations of this function in the models, as it is listed in the concept.  One might (should??) argue that the degree functions should only be generic, as it is nothing but counting `next()` steps around a face, or `opposite(next())` steps around a vertex.

afabri: We cannot call it degree, as the argument is a halfedge_descriptor and not a face_descriptor. Let me read what the context is, and come up with a proposal for what to do.

afabri: And we cannot call `degree(face(hd,g), g)` as the face might be the  `null_face()`.  

We will remove the new function, and in the code write  `boost::distance(CGAL::halfedges_around_face(hd,g))`.

maxGimeno: >I have asked that question three weeks ago. No answer, no modification, no progress!...

Are not lines 754 to 764 documentation I wrote three weeks ago ?

Mesh_3/include/CGAL/Mesh_polyhedron_3.h
maxGimeno: Can't we just ignore it for now and wait for #2023, which will apply all the modifications in the right order, instead of trying to cherry pick commits that manipulate properties that have been deeply reworked in #2023 but not here, creating conflicts that I do not understand enough to fix them ?

lrineau:
Here are the list of files (outside `Polyhedron/demo/Polyhedron/`) that are modified by this PR:
```
BGL/include/CGAL/boost/graph/Graph_with_descriptor_with_graph.h
BGL/include/CGAL/boost/graph/helpers.h
Mesh_3/examples/Mesh_3/CMakeLists.txt
Mesh_3/examples/Mesh_3/data/cheese-box.off
Mesh_3/examples/Mesh_3/data/fandisk-box.off
Mesh_3/examples/Mesh_3/mesh_polyhedral_domain.cpp
Mesh_3/examples/Mesh_3/mesh_polyhedral_domain_sm.cpp
Mesh_3/examples/Mesh_3/mesh_polyhedral_domain_with_features.cpp
Mesh_3/examples/Mesh_3/mesh_polyhedral_domain_with_features_sm.cpp
Mesh_3/examples/Mesh_3/mesh_polyhedral_domain_with_two_polyhedra.cpp
Mesh_3/include/CGAL/Mesh_3/Detect_features_in_polyhedra.h
Mesh_3/include/CGAL/Mesh_3/experimental/Get_facet_patch_id.h
Mesh_3/include/CGAL/Mesh_polyhedron_3.h
Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
Mesh_3/include/CGAL/Polyhedral_mesh_domain_with_features_3.h
Mesh_3/include/CGAL/Triangle_accessor_3.h
Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Detect_features_in_polyhedra.h
Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Detect_features_in_polyhedra_fwd.h
Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/properties.h
Polyhedron/include/CGAL/boost/graph/PMP_properties_Polyhedron_3.h
Polyhedron/include/CGAL/boost/graph/properties_Polyhedron_3.h
Property_map/include/CGAL/property_map.h
Surface_mesh/include/CGAL/boost/graph/PMP_properties_Surface_mesh.h
Surface_mesh/include/CGAL/boost/graph/graph_traits_Surface_mesh.h
Surface_mesh/include/CGAL/boost/graph/properties_Surface_mesh.h
Three/include/CGAL/Three/Scene_interface.h
Triangulation_2/include/CGAL/Triangulation_line_face_circulator_2.h
```

Maxime, can you please modify the title and the first message of this PR, to describe what is really changed by the PR?

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Mesh_3_plugin_cgal_code.cpp
```diff
@@ -42,23 +62,24 @@ Meshing_thread* cgal_code_mesh_3(const Polyhedron* pMesh,
...
             << "  approximation bound: " << facet_approx << std::endl;
-  if (pMesh->is_closed())
+  if (is_closed(*pMesh))
     std::cerr << "  tetrahedra size bound: " << tet_sizing << std::endl;
 
   std::cerr << "Build AABB tree...";
   CGAL::Real_timer timer;
   timer.start();
 
+  typedef typename Polyhedral_mesh_domain_selector<Mesh>::type Pwd;
```
lrineau: What is that name `Pwd`? Should not it be just something like ` Mesh_domain`?

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.h
```diff
@@ -49,17 +54,61 @@ class SCENE_SURFACE_MESH_ITEM_EXPORT Scene_surface_mesh_item
...
+  void update_vertex_indices(){}
+  void update_halfedge_indices(){}
+  void update_facet_indices(){}
+  Vertex_selection_map vertex_selection_map();
+  Face_selection_map face_selection_map();
+
+  std::vector<QColor>& color_vector();
+
+  void set_patch_id(SMesh::Face_index f,int i)const;
+  int patch_id(SMesh::Face_index f)const;
```
lrineau: The code of `Scene_surface_mesh_item.h` is wrong: it should not expose any CGAL API.

- All the typedefs `<foo>_descriptor`,
- all the functions that access or modify properties of elements of the mesh (instead of global properties of the item) 
should be moved to another file, such as the file `Polyhedron_type.h` for polyhedron.

Anyway, I have checked and, for example, the function `Scene_surface_mesh_item::set_patch_id` is never used anywhere in the demo.

Polyhedron/demo/Polyhedron/Travel_isolated_components.h
```diff
@@ -5,12 +5,58 @@
...
+    return get(fmap, fd);
+  }
+  std::size_t id(mesh_hd hd)
+  {
+    return get(hmap, hd);
+  }
+  std::size_t id(mesh_ed ed)
+  {
+    return get(emap, ed);
+  }
```
lrineau: Remark for @afabri and @sloriot, not really related to this PR: that sort of API will fail as soon as we have a model of `FaceGraph` that use simple integers for all types of descriptor. An API assuming that, for example, `vertex_descriptor` and `face_descriptor` are different, cannot be generic.

Property_map/include/CGAL/property_map.h
```diff
@@ -39,6 +39,31 @@ namespace CGAL {
...
+  get(const static_property_map& pm, const key_type&)
+  {
+    return pm.v;
+  }
+
+  inline friend
+  void
+  put(static_property_map& pm, const key_type&, const value_type&)
+  {}
+};
```
lrineau: @afabri:

- That class `static_property_map` should be named `Static_property_map`, to follow CGAL naming conventions,
- how is that different from the better implementation `boost::static_property_map`?

Triangulation_2/include/CGAL/Triangulation_line_face_circulator_2.h
```diff
@@ -161,6 +161,10 @@ Triangulation_line_face_circulator_2(Vertex_handle v,
   // dir 
   // or null iterator
 {
+  if(!((!_tr->is_infinite(v)) &&
+       (_tr->dimension() == 2)  &&
+       (! _tr->xy_equal(v->point(),dir))))
+    std::cout<<"";
```
lrineau: Please remove that debugging code!

Polyhedron/demo/Polyhedron/Travel_isolated_components.h
afabri: They must be different as we overload functions such as `halfedge()` for vertex and face descriptor. I must admit that there are no overlaods in the boost BGL.

Property_map/include/CGAL/property_map.h
```diff
@@ -49,18 +51,18 @@ class static_property_map
...
   {
     return pm.v;
   }
 
   inline friend
   void
-  put(static_property_map& pm, const key_type&, const value_type&)
+  put(Static_property_map& pm, const key_type&, const value_type&)
   {}
 };
```
lrineau: This commit was not correctly tested. It triggers compilation errors, detected by Travis, [in Mesh_3][m3], and [in the Polyhedron demo][demo]. It seems to be the same culprit: two occurrences of the old name in `Surface_mesh/include/CGAL/boost/graph/PMP_properties_Surface_mesh.h`.

  [m3]: https://travis-ci.org/CGAL/cgal/jobs/233619359#L4138
  [demo]: https://travis-ci.org/CGAL/cgal/jobs/233619449#L5412

Mesh_3/include/CGAL/Mesh_3/experimental/Get_facet_patch_id.h
```diff
@@ -74,6 +74,53 @@ get(const Get_facet_patch_id<Primitive>, const typename Primitive::Id& primitive
...
+namespace CGAL { namespace Mesh_3 {
+
+template <typename MeshDomain>
+typename boost::property_traits< Get_facet_patch_id_sm<MeshDomain> >::value_type
+get(const Get_facet_patch_id_sm<MeshDomain>, const typename MeshDomain::AABB_primitive::Id& primitive_id) {
+  typedef typename boost::property_map<typename MeshDomain::Polyhedron::Graph, face_patch_id_t<typename MeshDomain::Patch_id> >::type Fpim;
+  Fpim fpim = get(face_patch_id_t<typename MeshDomain::Patch_id>(),*((*primitive_id).graph));
+  typename MeshDomain::Patch_id patch_index =  get(fpim, (*primitive_id).descriptor);
+  return patch_index;
+}
```
lrineau: @afabri I do not understand that patch to `Mesh_3/include/CGAL/Mesh_3/experimental/Get_facet_patch_id.h`. Why is that so complicated?! It should be simpler with a `Surface_mesh` than with a `Polyhedron_3`.

Polyhedron/demo/Polyhedron/SMesh_type.h
```diff
@@ -0,0 +1,21 @@
+#ifndef SMESH_TYPE_H
+#define SMESH_TYPE_H
+
+#include <boost/scoped_ptr.hpp>
+#include <boost/array.hpp>
+#include <set>
```
lrineau: Why include `<boost/scoped_ptr.hpp>`, `<boost/array.hpp>`, and `<set>` whereas they are not used?

Polyhedron/demo/Polyhedron/SMesh_type.h
```diff
@@ -0,0 +1,21 @@
...
+#include <CGAL/Surface_mesh/Surface_mesh_fwd.h>
```
lrineau: If the header `SMesh_type.h` is suppose to define the type `SMesh`, then the real header should be included!

Polyhedron/demo/Polyhedron/Polyhedron_type.h
```diff
@@ -193,10 +195,38 @@ class Polyhedron_demo_items : public CGAL::Polyhedron_items_3 {
...
+  typedef boost::property_map<Polyhedron,boost::vertex_index_t>::type type;
+  typedef boost::property_map<Polyhedron,boost::vertex_index_t>::const_type const_type;
+};
+template <>
+struct property_map<Polyhedron, CGAL::face_selection_t>
+{
+  typedef boost::property_map<Polyhedron,boost::face_index_t>::type type;
+  typedef boost::property_map<Polyhedron,boost::face_index_t>::const_type const_type;
+};
+}//boost
```
lrineau: What are the lines 198-266 supposed to do? Why are they in a private header of the demo, and not in the CGAL headers?

Mesh_3/examples/Mesh_3/CMakeLists.txt
```diff
@@ -73,8 +73,11 @@ if ( CGAL_FOUND )
     create_single_source_cgal_program( "mesh_cubes_intersection_with_features.cpp" )
     create_single_source_cgal_program( "mesh_implicit_domains.cpp" "implicit_functions.cpp" )
     create_single_source_cgal_program( "mesh_polyhedral_domain.cpp" )
+    create_single_source_cgal_program( "mesh_polyhedral_domain_sm.cpp" )
     create_single_source_cgal_program( "remesh_polyhedral_surface.cpp" )
     create_single_source_cgal_program( "mesh_polyhedral_domain_with_features.cpp" )
+    create_single_source_cgal_program( "mesh_polyhedral_domain_with_features_sm.cpp" )
+    create_single_source_cgal_program( "mesh_polyhedral_domain_with_two_polyhedra.cpp" )
```
lrineau: @afabri Are those new example just supposed to exist, without any explanation/documentation for users?

Polyhedron/demo/Polyhedron/Polyhedron_type.h
maxGimeno: I guess I left them here because I have absolutely no idea where else they are supposed to go.

maxGimeno: Plus they look like demo specific code that are only here so some plugins are compatible with both Polyhedorn and SM, as those property maps only forward the calls to other property_maps.

lrineau: That is right. That seems a strange trick to use the property associated to `boost::face_index`, for example, as a replacement to the one associated to `CGAL::face_selection`. I wonder how that can work.

maxGimeno: It was face_index and vertex_index that were used initially, but as those are not writable for SM, Andreas added vertex and face selection, that replace _index. For Polyhedron it is all the same, so we forward it to _index, but SM have a real implementation for _selection, IIRC.

sloriot:
@afabri @lrineau 
You mentioned in a private thread the usage of `Surface_mesh` in the class `Polyhedral_mesh_domain`.

The current solution you choose was to use `Graph_with_descriptor_with_graph`. An alternative is to put false for the `Boolean` `OneFaceGraphPerTree` of `AABB_face_graph_triangle_primitive`. However this does not change that you need to detect automatically how to set this value. Here are several proposal:
1 - Always use an AABB-tree of AABB-tree (IIRC you told me it was faster)
2 - Add a template parameter in `Polyhedral_mesh_domain` that is controlling the value of `OneFaceGraphPerTree` and mark it as advanced. The default is value is `CGAL::Default`. Internally, to set  the value of `OneFaceGraphPerTree` you always put `false` but in the following cases:
  -  Something different from `Default` was passed to the extra template parameter
  -  FaceGraph is an `Polyhedron_3` or [`is_base_of`](http://www.boost.org/doc/libs/master/libs/type_traits/doc/html/boost_typetraits/reference/is_base_of.html) `Polyhedron_3`.
3 - Split  `Polyhedral_mesh_domain` in two classes (one if you have a single polyhedra, and one if you have several)

lrineau:
@maxGimeno @afabri

Have a look at the test results https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-37.shtml were this PR was merged.

There are:
  - compilation errors on Windows (`dcelspec` stuff) and an all platforms, in the Polyhedron demo. Too bad they were not tested before,
  - warnings in the three Apple columns,
  - strange compilation errors [in `Generator`][g] and [in `Mesh_3`][m] with MSVC2012.

  [g]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-37.shtml#Generator
  [m]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-37.shtml#Mesh_3

maxGimeno:
The testsuite is finally green ! Oh happy day !

PR #1985, opened 2017-03-22
Polygon Mesh Processing: Finer grain for license check

afabri:
## Summary of Changes
Create license check files for the different subsections. 

## Release Management


afabri:
@sloriot As we know include internal files, what will replace these calls to internal functions in the long run?

sloriot:
@afabri  I don't understand the question.

lrineau:
Should not we actually define `CGAL_LICENSE_ERROR` and an appropriate set of `CGAL_<pkg>_COMMERCIAL_LICENSE` in a few .cpp test files, to actually check that it works?

lrineau:
This PR blocks CGAL-4.10-beta1, and not CGAL-4.10. We cannot import such a important change after the beta release.

PR #1993, opened 2017-03-24
Io plugins with surface mesh

maxGimeno:
## Summary of Changes

Make the Io_plugins able to manage Surface_mesh_items as well as Polyhedron_items

This PR is based on #1954 
## Release Management

* Issue(s) solved (if any): fix #1931


lrineau:
This PR is rather simple. it can be reviewed as-is.

It should be first rebased on the latest version of #1954.

Polyhedron/demo/Polyhedron/Plugins/IO/STL_io_plugin.cpp
```diff
@@ -1,25 +1,44 @@
...
+            CGAL::Three::Scene_interface* scene_interface,
+            Messages_interface*) {
+    //get the references
+    this->scene = scene_interface;
+    this->mw = mainWindow;
+  }
+  QList<QAction*> actions() const {
+    return QList<QAction*>();
+  }
+  bool applicable(QAction*) const { return false;}
```
lrineau: Why is that I/O plugin turned into a regular plugin, with no action? Just to get the access to `this->mw` and then to the property?!

maxGimeno: Yes

maxGimeno:
Ready for testing.

PR #1994, opened 2017-03-27
fix a bug when removing redundant edges after an overlay for a boolea…

sloriot:
## Summary of Changes

inner ccbs and outer ccb were reused to avoid a reallocation. However,
since the face field of inner/outer ccbs are needed will reassign them,
we now delay the setting of the new face for the ccbs to avoid
overwritting the face field

## Release Management

* Affected package(s): 2D Regularized Boolean Set Operations

cc @efifogel


lrineau:
That was actually tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.9.1-Ic-164.shtml

Issue #1996, opened 2017-03-29
Document that CGAL-4.10 requires CMake>=3.1


lrineau:
In `master`, our `Installation/changes.html` "documents" that the minimal required version of CMake is 3.1. But we forgot to write that new requirement in `INSTALL.md` and in the installation manual.

lrineau:
Cc @sloriot @cjamin

lrineau:
Fixed by #2013.

PR #2013, opened 2017-04-03
Document that CGAL-4.10 requires CMake>=3.1 -  Fix for #1996

afabri:


afabri:
Tested  in testsuite [Ic-170](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.10-Ic-170/output/Manual/installation.html#secconfigwithcmake)

PR #2017, opened 2017-04-03
CGAL_Core: make it less verbose to fix #1915

### LINKED ISSUE #1915: Core is too verbose
There are some `std::cerr` in Core and also the file `Core_Diagnotics` that is created on the disk when an error occurs. I think they should be protected by the macro `CORE_TRACE`. Do you agree?
While at it, shall we rename the debug macro `CORE_TRACE` to `CGAL_CORE_TRACE`?

afabri:

### DIFF
--- a/CGAL_Core/include/CGAL/CORE/BigFloat_impl.h
+++ b/CGAL_Core/include/CGAL/CORE/BigFloat_impl.h
@@ -137,7 +137,7 @@ BigFloatRep::BigFloatRep(double d) : m(0), err(0), exp(0) {
       exp--;
       stop++;
     }
-#ifdef CORE_DEBUG
+#ifdef CGAL_CORE_DEBUG
     CGAL_assertion (s >= 0);
 #endif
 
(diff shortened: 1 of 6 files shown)


CGAL_Core/include/CGAL/CORE/BigFloat_impl.h
```diff
@@ -137,7 +137,7 @@ BigFloatRep::BigFloatRep(double d) : m(0), err(0), exp(0) {
       exp--;
       stop++;
     }
-#ifdef CORE_DEBUG
+#ifdef CGAL_CORE_DEBUG
     CGAL_assertion (s >= 0);
 #endif
```
lrineau: (That is just a remark, and not a request for a change.)

I find strange that such an assertion is protected by macro `CORE_DEBUG`, knowing that `CGAL_assertion` itself it protected by `NDEBUG` and `CGAL_NDEBUG`.

afabri:
@lrineau Any idea why this fails?  It does not do on VC++ on my machine.

lrineau:
@afabri, in https://github.com/CGAL/cgal/pull/2017#issuecomment-291422651:
> @lrineau Any idea why this fails?  It does not do on VC++ on my machine.

What do you mean by it fails? If that is shown by the testsuite, then please copy-paste a link to the error log.

afabri:
When I had written that it failed travis was red. Now it is green.  A miracle.

lrineau:
Sometime Travis jobs fails for networks errors, and it happens that I restart jobs that failed that way.

PR #2019, opened 2017-04-03
Add genus for closed polyhedra.

### LINKED ISSUE #1770: Need extra properties in statistics...
It is possible to compute statistics and list properties on polyhedral surfaces. 
I think it would be important to also list:
- genus (trivial from Euler formula once we have components, boundaries, etc)
- stats per face 
- size of bbox

maxGimeno:
## Summary of Changes

Add the genus in the statistics of a closedpolyhedorn item.
## Release Management

* Issue(s) solved (if any): fix #1770

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
@@ -160,6 +160,7 @@ struct Scene_polyhedron_item_priv{
   mutable QOpenGLShaderProgram *program;
   unsigned int number_of_null_length_edges;
   unsigned int number_of_degenerated_faces;
+  int genus;
   bool self_intersect;
   int m_min_patch_id; // the min value of the patch ids initialized in init()
   mutable bool all_ids_displayed;
@@ -860,6 +861,7 @@ invalidate_stats()
   volume = -std::numeric_limits<double>::infinity();
   area = -std::numeric_limits<double>::infinity();
   self_intersect = false;
+  genus = -1;
 }
 
 Scene_polyhedron_item*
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -160,6 +160,7 @@ struct Scene_polyhedron_item_priv{
   mutable QOpenGLShaderProgram *program;
   unsigned int number_of_null_length_edges;
   unsigned int number_of_degenerated_faces;
+  double genus;
```
lrineau: I do not really understand why the `genus` is a `double`. It should be an integral type.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -1487,6 +1489,28 @@ QString Scene_polyhedron_item::computeStats(int type)
...
+  {
+    if(!d->poly->is_closed())
+    {
+      return QString("n/a");
+    }
+    else if(d->genus == -std::numeric_limits<double>::infinity())
+    {
+      std::size_t S(d->poly->size_of_vertices()), A(d->poly->size_of_halfedges()/2), F(d->poly->size_of_facets());
+      //cast in int to avoid intermediary state containing a negative size_t.
+      d->genus = 1.0 - (static_cast<int>(S)-static_cast<int>(A)+static_cast<int>(F))/2.0;
```
lrineau: Let's say the type of `d->genus` is `std::ptrdiff_t`, then:
```c++
std::ptrdiff_t S(d->poly->size_of_vertices()), A(d->poly->size_of_halfedges()/2), F(d->poly->size_of_facets());
d->genus =  1 - (S-A+F)/2;
```

I'd like also to remind that the naming convention of CGAL says that variable names should be lowercase.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -1497,7 +1497,9 @@ QString Scene_polyhedron_item::computeStats(int type)
     }
     else if(d->genus == -1)
     {
-      int s(d->poly->size_of_vertices()), a(d->poly->size_of_halfedges()/2), f(d->poly->size_of_facets());
+      int s(static_cast<int>(d->poly->size_of_vertices())),
+          a(static_cast<int>(d->poly->size_of_halfedges())/2),
+          f(static_cast<int>(d->poly->size_of_facets()));
       //cast in int to avoid intermediary state containing a negative size_t.
       d->genus = 1.0 - (s-a+f)/2.0;
```
lrineau: I do not understand the comment:
```C++
//cast in int to avoid intermediary state containing a negative size_t.
```

A `std::size_t` unsigned. If you want the "signed-`size_t`", then the type is `std::ptrdiff_t`, and not `int`!

I recommend to use the right types and less `static_cast` unless one has to use two different incompatible APIs.

... If you use `std::ptrdiff_t`, you will have to cast `s-a+f` to double:
```
d->genus = 1.0 - double(s-a+f)/2.0;
```
because the precision of `ptrdiff_t` might be 64 bits where a double only has 53 bits of precision in its mantissa... and compilers (like MSVC) will warn about a possible loss of precision.

maxGimeno:
done

PR #2020, opened 2017-04-03
KDS: do not open /dev/null  Fix #1988

### LINKED ISSUE #1988: Windows: Creation of dev/null file
## Issue Details

On Windows:
If "dev" folder is present, a "dev/null" file is created.
In CGAL/Tools/Log.h null_ stream opens null device file "/dev/null" which is only correct on Unix.

## Environment

* Operating system: Windows 8.1
* CGAL version: 4.9

afabri:
Fix #1988


afabri:
Tested in [4.10-Ic-170](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-170.shtml)

PR #2022, opened 2017-04-04
Polyhedron_demo : Reverse alwaysUse

### LINKED ISSUE #2008: Polyhedron demo : add `Load as...`
## Issue Details
When we open a file (e.g. a `mesh.cgal`), choose a plugin to open it (e.g. `C3t3_io_plugin`), and click "Always open with...", it is never possible to make another decision while keeping the demo open.

Let's say I want to open a `.off` file as a `Surface_mesh`, and later another as a `Point_set`. They will all be opened as `Surface_mesh`, and then I need to use the conversion tool, which is not always available, and/or not always complete.

## Source Code

Todo : Add a "Load as..." submenu in the `File` menu, that allows to change your mind on the plugin to use to open some new file (as if the extension has never been met before, or the "always" checkbox never clicked).

maxGimeno:
## Summary of Changes

Add an action to the menuFile to reset the default_loaders_map, allowing the user to choose the loader for all items again, even if alwaysUse was checked.

This action appears if alwaysUse is checked, and disappears when it is clicked.

## Release Management

* Issue(s) solved (if any): fix #2008

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -351,7 +351,7 @@ MainWindow::MainWindow(QWidget* parent)
   }
 
   QMenu* menuFile = findChild<QMenu*>("menuFile");
-  insertActionAfter(menuFile, QString("Load"), actionAddToGroup);
+  insertActionBefore(menuFile, actionAddToGroup);
   statistics_dlg = NULL;
   statistics_ui = new Ui::Statistics_on_item_dialog();
 
@@ -1010,7 +1010,7 @@ void MainWindow::open(QString filename)
     connect(actionResetDefaultLoaders, SIGNAL(triggered()),
             this, SLOT(reset_default_loaders()));
     default_plugin_selection[fileinfo.completeSuffix()]=load_pair.first;
-    insertActionAfter(ui->menuFile, QString("Load"), actionResetDefaultLoaders);
+    insertActionBefore(ui->menuFile, actionResetDefaultLoaders);
   }
   
   
@@ -1944,7 +1944,7 @@ void MainWindow::reset_default_loaders()
   menu->removeAction(actionResetDefaultLoaders);
 }
 
-void MainWindow::insertActionAfter(QMenu* menu, QString actionAfterName, QAction* actionToInsert)
+void MainWindow::insertActionBefore(QMenu* menu, QAction* actionToInsert)
 {
   const char* prop_name = "Menu modified by MainWindow.";
   if(menu)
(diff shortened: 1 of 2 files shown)


janetournois:
It works for me (msvc 2013, release mode). Thank you!

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1945,3 +1928,53 @@ void MainWindow::resetHeader()
...
+      }
+    }
+
+    // Insert "Load implicit function" action
+    if ( NULL != actionAfter )
+    {
+      menu->insertAction(actionAfter,actionToInsert);
+    }
+  }
+}
```
lrineau: That function is complicated, and there is a lot better to do using the Qt API.

The "Load" action is simply `ui->actionLoad`, or (in case you do not have a direct access to the action pointer) `this->findChild<QAction*>("actionLoad")`.

Then you can do:
```C++
if(!menu->actions().contains(actionToInsert)) {
  menu>insertAction(ui->actionLoad, actionToInsert);
}
```

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1945,3 +1928,34 @@ void MainWindow::resetHeader()
...
+
+void MainWindow::insertActionBefore(QMenu* menu, QAction* actionToInsert)
+{
+  const char* prop_name = "Menu modified by MainWindow.";
+  if(menu)
+  {
+    bool menuChanged = menu->property(prop_name).toBool();
+    if(!menuChanged) {
+      menu->setProperty(prop_name, true);
+    }
```
lrineau: I am pretty sure you can remove `prop_name`, `menuChanged` amd the `setProperty(prop_name, true)`. That property is always set to `true`, and its value does not change the behavior of the functions.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1945,3 +1928,34 @@ void MainWindow::resetHeader()
...
+    return;
+  bool menuChanged = menu->property(prop_name).toBool();
+  if(!menuChanged) {
+    menu->setProperty(prop_name, true);
+  }
+  QList<QAction*> menuActions = menu->actions();
+  menu->removeAction(actionResetDefaultLoaders);
+}
+
+void MainWindow::insertActionBefore(QMenu* menu, QAction* actionToInsert)
```
lrineau: - Either the name should be changed to `'insertActionBeforeLoadPlugin`,
- or the function should take a third parameter: `insertActionBefore(QMenu* menu, QAction* action, QAction* actionToInsert)`.

PR #2005, opened 2017-03-30
Detect_sharp_edges_plugin works with a Scene_surface_mesh_item

maxGimeno:
## Summary of Changes
The Detect_sharp_edges_plugin now works with SM_items. 

To do so, the file Detect_features_in_polyhedra.h has been BGL-ized and a new property_map has been added to Polyhedron_type, Scene_surface_mesh_item and Mesh_polyhedron_3. 

This property map contains the feature halfedges.
## Release Management

This PR is based on #1983 

* Affected package(s):Polyhedron_demo, Polygon_mesh_processing
* Issue(s) solved (if any): fix #1931

PR #2023, opened 2017-04-04
Mesh_3: Make make_mesh_3() work for Surface_mesh

afabri:
## Summary of Changes

This is WIP.

The partial specialisations for the `Triangle_accessor` for a `Graph_with_descriptor_with_graph` are enough to make the polyhedral domain class work for a `Surface_mesh`.   The next step will be to deal with domains with features, but for that we better wait for the merge of the branch that makes the Detect_sharp_features_plugin work for a `Surface_mesh`.

## Release Management

* Affected package(s): Mesh_3


maxGimeno:
The branch is now rebased on #2005 to take advantage of the Mesh_polyhedron_3 and Detect_features_in_polyhedra alterations.

Property_map/include/CGAL/property_map.h
```diff
@@ -39,6 +39,28 @@ namespace CGAL {
...
+  get(const static_property_map& pm, const key_type&)
+  {
+    return pm.v;
+  }
+
+  inline friend
+  void
+  put(static_property_map& pm, const key_type&, const value_type&)
+  {}
+};
```
lrineau: Why do you add a `CGAL::static_property_map` instead of using the boost version [`boost::static_property_map`][doc]?

  [doc]: http://www.boost.org/doc/libs/1_63_0/libs/property_map/doc/static_property_map.html

Property_map/include/CGAL/property_map.h
```diff
@@ -39,6 +39,28 @@ namespace CGAL {
...
+
 
 template <typename PM1, typename PM2>
 class OR_property_map {
```
lrineau: That property map `OR_property_map` is not used anywhere in CGAL. It should be removed.

lrineau:
The mesh_3_plugin builds again (with `Polyhedron_3<..>`) after the following three commits:
- 207bf6f :  Use the new header `<CGAL/boost/graph/properties_Polyhedron_3.h>`
- 7d51b99 : Add compatibility: support when `Patch_id` is `Tag_true`/`Tag_false`
- c746be9 : Move properties from `<CGAL/Mesh_polyhedron_3.h>` to `<CGAL/boost/graph/properties_Polyhedron_3.h>`

lrineau:
@maxGimeno I think after the last three commits you can start again working on the branch. The properties are now reorganized like we planned, and generic: they will work on a polyhedron type implementing the implicit API.

BGL/include/CGAL/boost/graph/Graph_with_descriptor_with_graph.h
```diff
@@ -135,11 +139,18 @@ struct Graph_with_descriptor_with_graph
...
+
   Graph_with_descriptor_with_graph(Graph& graph)
     : graph(&graph)
   {}
-private:
-  Graph_with_descriptor_with_graph(const Graph_with_descriptor_with_graph&){} // disable copy-constructor to avoid non-wanted copies
+  
+  // private:   
+  // disable copy-constructor to avoid non-wanted copies
+  // AF: Why was this private??
```
maxGimeno: It is supposed to be private, to keep the user from using it.

Problem is : Polyhedral_mesh_domain_with_features_3 requires a copy of the template type Polyhedron. If we disable this, what do we do for surface_only remeshing ?

afabri: I see no reason why this should be private

maxGimeno: Because it forbids its use. This way you get a compilation error instead of a segfault.

sloriot: We could have another function doing the copy. That way if you want to a copy it is done using that function and it is not an accidental use.

janetournois:
@lrineau @maxGimeno what is this PR waiting for?

lrineau:
This PR is for CGAL-4.11. It is waiting for the branch `releases/CGAL-4.10-branch`, so that `master` and `integration` can trigger CGAL-4.11.

maxGimeno:
It is also waiting for the branches it is based on.

maxGimeno:
This branch is now merged in #1983 .

lrineau:
This PR should be close now, no?

maxGimeno:
Yes it should

lrineau:
PR merged (and thus replaced) into #1983.

lrineau:
I close this PR, merged in #1983.

PR #2025, opened 2017-04-05
Point set 3: fix index type

sgiraudot:
## Summary of Changes

This PR fixes two problems:

1. The point set index violates the `Index` concept as the constructor using a `size_t` is private (this was done to avoid some mis-usage of the point set but should be removed)
2. The index map is using a `std::size_t` which can lead to an overly large memory usage when loading very big point sets. This PR makes it use `boost::uint32_t` which should be half the size and can still address more than 4 billion points. Anyway, loading more than 4 billion points requires at least 50GB of RAM with `CGAL::Simple_cartesian<float>`, which is not common at the moment. I added a `#ifdef` macro `CGAL_POINT_SET_3_USE_STD_SIZE_T_AS_SIZE_TYPE` which is undefined by default so that a user who actually needs to load more than 4 billion points can still do it.

## Release Management

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -120,12 +120,9 @@ public:
     friend class std::vector<Index>;
     std::size_t value;
     
-    // Only Point_set_3 and other friend classes are allowed to
-    // instantiate an Index with a specific value
-    Index (const std::size_t& value) : value (value) { }
-    /// \endcond
   public:
     Index (const Index& index) : value (index) { }
+    Index (const std::size_t& value) : value (value) { }
     Index () : value ((std::size_t)(-1)) { }
     Index operator= (const Index& index) { value = index.value; return *this; }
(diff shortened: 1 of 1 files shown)


lrineau:
Tested last night. Warnings on Windows: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10-Ic-173.shtml#Point_set_3

lrineau:
Should this PR considered as a bug-fix, and selected for CGAL-4.10?

sgiraudot:
The warnings should be fixed. Yes, the PR can be selected for CGAL-4.10.

lrineau:
There was also a warning threre:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-Ic-173/Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz

Is it handled too?

sgiraudot:
Yes, it was coming from the same line.

PR #2044, opened 2017-04-10
Polyhedron demo: Reorganize selection plugin

sgiraudot:
The selection plugin is overly long and usually makes the left column way to high.

## Summary of Changes

__Note: this PR is not ready to merge, I'm creating it so we can discuss it.__

Using tab widgets and moving stuff around, I managed to reduce the height by about 50%. It's a bit wider than before, but as screens are mainly 16/9 now, I think it's less of an issue.

The main problem remaining is the Euler operations: because of the images, the height becomes really important when clicking on the tab. Note that before, there was a bug when switching back to selection mode (the widget was dropping to the bottom) that is fixed, but now there's a bug when switching to edition mode (the widget goes over the bottom bar).

Is there a reason why this Euler operation is not in a different widget? It seems to me that there should even be two separated actions (one for Selection, one for Edition) that each raises its own widget. I understand that both share a code about selection, but that seems counter-intuitive to go to Selection to actually use the Euler operations.

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
@@ -105,7 +105,7 @@ public:
     connect(ui_widget.validateButton, SIGNAL(clicked()), this, SLOT(on_validateButton_clicked()));
     connect(ui_widget.Expand_reduce_button, SIGNAL(clicked()), this, SLOT(on_Expand_reduce_button_clicked()));
     connect(ui_widget.Select_sharp_edges_button, SIGNAL(clicked()), this, SLOT(on_Select_sharp_edges_button_clicked()));
-    connect(ui_widget.modeBox, SIGNAL(currentIndexChanged(int)), this, SLOT(on_ModeBox_changed(int)));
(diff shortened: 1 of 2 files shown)


sloriot:
I think it would help the discussion if you can inline a screenshot before/after the change.

sgiraudot:
The current widget on the left, the proposed one on the right:

Selection widget

The edition tab is as big as before, but as long as you don't click on edition, the widget stays small. I'm not quite satisfied with this, as I said, I think edition should be a separated widget and a separated entry in the menu.

sloriot:
Thanks, now I understand. It is indeed better.

lrineau:
@sgiraudot, in https://github.com/CGAL/cgal/pull/2044#issue-220571955:
> Note: this PR is not ready to merge

Can you comment on that? Is there a common agreement that this PR should be merged now in `master`?

sgiraudot:
I was expecting a discussion to see if we could improve my proposal, but if everyone is happy about it, I think it can be merged.

(There is still the problem that the widget becomes overly big again when we click on Edition, but I think the best way to fix it is to move the Edition to a separated widget and that would probably require way more work and another PR. Let's move one step at a time.)

PR #2054, opened 2017-04-14
Improve deprecation_warning.h and fix some missing / improper deprecated macros

MaelRL:
## Summary of Changes

[This commit](https://github.com/CGAL/cgal/commit/6d13fa2c3178dfecd30b04bcefc5861304466ce8) and other commits deprecated Regular_triangulation_euclidean_traits, but the macros `CGAL_NO_DEPRECATED_CODE` are missing in the files.

## Release Management

* Affected package(s): Triangulation_2/3

### DIFF
--- a/Installation/include/CGAL/internal/deprecation_warning.h
+++ b/Installation/include/CGAL/internal/deprecation_warning.h
@@ -12,44 +12,69 @@
 // This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
 // WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
 //
-// $URL$
-// $Id$
-// 
-//
-// Author(s)     : Philipp Möller
+// Author(s)     : Philipp Möller, Mael Rouxel-Labbé
 
-// Including this header will issue a warning during compilation or
-// cause compilation to fail if CGAL_NO_DEPRECATED_CODE is defined.
-// CGAL_DEPRECATED_HEADER and CGAL_REPLACEMENT_HEADER can be defined
+// Including this header will cause compilation to fail
+// if CGAL_NO_DEPRECATED_CODE is defined. If this is not the case, it will issue
+// a warning during compilation unless CGAL_NO_DEPRECATION_WARNINGS.
+
+// CGAL_DEPRECATED_HEADER, CGAL_REPLACEMENT_HEADER, and
+// CGAL_DEPRECATED_MESSAGE_DETAILS can be defined
 // to a string literal to customize the warning.
-// CGAL_DEPRECATED_HEADER and CGAL_REPLACEMENT_HEADER are undefined,
-// after the file is included.  
+
+// CGAL_DEPRECATED_HEADER, CGAL_REPLACEMENT_HEADER, and
+// CGAL_DEPRECATED_MESSAGE_DETAILS are undefined after the file is included.
 
 // The lack of an include guard is intentional and necessary.
 
 #include <CGAL/assertions.h>
 
-#ifndef CGAL_NO_DEPRECATION_WARNINGS
-
+// whether to print Warning or Error
 #if defined(CGAL_NO_DEPRECATED_CODE)
-// No deprecated code.
-CGAL_static_assertion_msg(false, "A deprecated header has been included and CGAL_NO_DEPRECATED_CODE is defined.");
-#endif // CGAL_NO_DEPRECATED_CODE
-
-// Build the message
-#define CGAL_INTERNAL_DEPRECATED_MESSAGE "Warning: A deprecated header has been included."
-
-#if defined(CGAL_DEPRECATED_HEADER) && defined(CGAL_REPLACEMENT_HEADER)
-#  undef CGAL_INTERNAL_DEPRECATED_MESSAGE
-#  define CGAL_INTERNAL_DEPRECATED_MESSAGE "Warning: The header " CGAL_DEPRECATED_HEADER " is deprecated. " \
-                                           "Please use " CGAL_REPLACEMENT_HEADER " instead."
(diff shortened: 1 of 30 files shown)


Triangulation_2/include/CGAL/Regular_triangulation_euclidean_traits_2.h
```diff
@@ -24,6 +24,11 @@
 
 #include <CGAL/license/Triangulation_2.h>
 
+#ifdef CGAL_NO_DEPRECATED_CODE
```
janetournois: I think this should be `#ifndef`, instead of `#ifdef`
And, should it really make an error? A warning would be enough, maybe...?

sloriot:
See also `CGAL/git/cgal/Installation/include/CGAL/internal/deprecation_warning.h`

janetournois:
thank you!

lrineau:
The macro `CGAL_NO_DEPRECATED_CODE` is supposed to trigger errors if deprecated code is used.

http://doc.cgal.org/4.9/Manual/preliminaries.html:
> In order to help users manage the changes to apply to their code, we attempt to make CGAL code emit warnings when deprecated code is used. This can be done using some compiler specific features. Those warnings can be disabled by defining the macro CGAL_NO_DEPRECATION_WARNINGS. On top of this, we also provide a macro, CGAL_NO_DEPRECATED_CODE, which, when defined, disables all deprecated features. This allows users to easily test if their code relies on deprecated features.

MaelRL:
Was successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-8.shtml

lrineau:
As pointed out by me and Sébastien in https://github.com/CGAL/cgal/pull/2054#issuecomment-294152430 and https://github.com/CGAL/cgal/pull/2054#issuecomment-294156350, deprecated headers should:
  - error out if `CGAL_NO_DEPRECATED_CODE` is defined, 
  - display a warning, unless `CGAL_NO_DEPRECATION_WARNINGS` is defined.

Both is already handled by the use of `<CGAL/internal/deprecation_warning.h>`. See for example `Git/cgal/AABB_tree/include/CGAL/AABB_polyhedron_triangle_primitive.h` for a usage.

MaelRL:
I disregarded those comments because `deprecation_warning.h` does not allow to pass custom messages and there is no replacement header for these deprecated classes. Also because there are other CGAL classes that are deprecated and do not use the `deprecation_warning.h` include.

Two possibilities:
1. I use `deprecation_warning.h` and the user must go look up in the code/doc what's wrong.
2. I change the code of `deprecation_warning.h` such that a macro like `CGAL_INTERNAL_DEPRECATED_MESSAGE_DETAILS` is appended (if it exists) to `CGAL_INTERNAL_DEPRECATED_MESSAGE`.

lrineau:
I would prefer the second choice (improvement of `deprecation_warning`). Would you mind amending also the corresponding documentation in our guidelines? Currently is says:
> To deprecate something, she should make modifications in the code and in the documentation.
>  * In the .h code-file she wants to deprecate, she should add the following :
>
>     #define CGAL_DEPRECATED_HEADER "<CGAL/old_header_file.h>"
>     #define CGAL_REPLACEMENT_HEADER "<CGAL/new_header_file.h>"
>     #include <CGAL/internal/deprecation_warning.h>


https://github.com/CGAL/cgal/wiki/Developing-Features-with-Git#coding

MaelRL:
If the name `CGAL_DEPRECATED_MESSAGE_DETAILS` for the optional custom text macro is deemed alright, I will edit the doc.

I made a few additional changes, see 6550621 commit message.

MaelRL:
d3bb2c2 adds a lot of warnings due to two deprecated headers that are often included:
- `Regular_triangulation_euclidean_traits_3.h`. All of these warnings are already fixed in the implicit conversion branch, which is almost ready to be merged, so I'd rather not do anything about the warnings in this branch.
- `Weighted_point.h`. `Surface_mesher` did not get converted to the kernel's weighted points and the class `Weighted_point` is used a lot. I don't really have time to look into it, so do I just comment the include `<CGAL/internal/depreciation_warning.h>` for now ?

MaelRL:
I silenced both deprecations so this branch is clean. I'll re-enable the depreciation macros of `Regular_triangulation_euclidean_traits_23` in the implicit conversion branch (thus this branch must be merged first). For `Weighted_point.h`... someone will need to clean `Surface_mesher` some day.

If the `new deprecation_warning.h` is satisfying, this is ready to be tested.

MaelRL:
Rebased and all warnings fixed.

MaelRL:
I changed the base to master because it's not urgent stuff and I think it's better if it's in the same release as the branch that removes implicit conversions between P/WP (which will re-enable the temporarily warnings that were temporarily silenced here).

Installation/include/CGAL/internal/deprecation_warning.h
```diff
@@ -12,44 +12,69 @@
...
+#else
+#  define CGAL_INTERNAL_DEPRECATED_MESSAGE \
+     CGAL_INTERNAL_DEPRECATED_MESSAGE_STATUS \
+     CGAL_INTERNAL_DEPRECATED_MESSAGE_HEADERS
 #endif
 
-// don't trigger on NO_DEPRECATIOON_WARNINGS and don't trigger twice on NO_DEPRECATED_CODE
-#if !defined(CGAL_NO_DEPRECATION_WARNINGS) && !defined(CGAL_NO_DEPRECATED_CODE)
+#if defined(CGAL_NO_DEPRECATED_CODE) // No deprecated code.
+CGAL_static_assertion_msg(false, CGAL_INTERNAL_DEPRECATED_MESSAGE);
```
lrineau, line 76: @MaelRL:
(This in an old PR, from about 10 years ago.)

Why not a plain `#error`, instead of a static assertion with `false` as the condition?

PR #1795, opened 2017-01-05
Work in progress: Kernel_23: Add Compare_slope_3 and Angle_3()(Point_3, Point_3, Point_3, Vector_3)

afabri:
(**update** by @lrineau: I took over this PR, and in particular the text of this first message.)

This PR adds two things in the CGAL Kernel concepts (and its models), plus the global functions:

- A functor `Compare_slope_3`,

- and an overload of `Angle_3::operator()(Point_3, Point_3, Point_3, Vector_3)`.

There are two a small features:
- [`Kernel::Compare_slope_3`][slope]
- [`Kernel::Angle_3`][angle]

  [slope]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::Compare_slope_3
  [angle]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::Angle_3

The PR also adds the free functions, and tests.

PR #1891, opened 2017-02-08
Use macros to add parameters in PMP

maxGimeno:
## Summary of Changes
- declares two macro in named_fucntion_params.h that help adding new parameters
- replaces the parameters declarations by calls to the macro if possible

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #1798 
* Feature/Small Feature (if any):

PR #2057, opened 2017-04-14
PMP : fix `PMP::is_outward_oriented()`

### LINKED ISSUE #2048: Bug in CGAL::PMP::is_outward_oriented()
## Issue Details

The is a bug in [`PMP::is_outward_oriented()`][doc]. It was [reported by a user on the `CGAL-discuss` mailing list][mail].

  [doc]: http://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__orientation__grp.html#gad71360b56f0d5340bf3006febd49286c`
  [mail]: https://sympa.inria.fr/sympa/arc/cgal-discuss/2017-04/msg00017.html

> Greetings,
> 
> I have one mesh whose normals are pointing outwards, but the function
> `CGAL::Polygon_mesh_processing::is_outward_oriented()` returns
> otherwise.
> 
> I have checked the code and the average of the normals of the faces
> incident to the min-vertex is used, which conceptually I doubt should
> work… And it doesn’t work in my case (unless I made a mistake
> somewhere), you can download my OFF file, along with the code I used,
> there:
> https://www.dropbox.com/sh/49as9yt139v6kxk/AABXI3Nh4ZfqCEDWHyv8Q-Zja?dl=0
> 
> If other ways to compute the normal were used (eg by using the area of
> the face as a weight), then the x-component of the vector could be
> negative (as is used by cgal), but I am wondering if that would always
> work. I think not.
> 
> Any comments appreciated, cheers,
> Hugo Ledoux

# The Problem

The current code is computing the vertex normal of a vertex with the lowest `x` coordinate, and then try to conclude on the coordinate of that normal.

* First problem: the computation of the vertex normal is currently wrong (issue #2047).
* Anyway, even if it is was computed correctly, `is_outward_oriented()` would make a Boolean decision based on a construction, that is inexact even with the `Epick` kernel.

# A Proposed Solution

A possible other implementation would be the following: try to find an extreme faces of the mesh, and test is the normal of that face is pointing to the right side.

More concretely:

1. Find a vertex `v` with the maximum `z` coordinate. Can be done using the predicate `Kernel::Compare_z_3`.
2. Among the incident edges to that vertex `v`, find one of the edges `e` with the <s>maximal</s> **minimal** slope, using `Kernel::Compare_slope_3` from the PR #1795. (text **updated** after https://github.com/CGAL/cgal/issues/2048#issuecomment-293601105)
3. Then, around that edge `e`, there are two faces (because the mesh is without borders) , and take the face `f` for which the other edge incident to `f` and `v` has the <s>maximal</s> **minimal** slope. (text **updated**)

My claim is that that face `f` appears in the superior envelop of the mesh (with a projection along the `z`-axis), and that a vertical ray piercing that face should *enter* the object delimited by the mesh. That means that the `z`-coordinate of the normal of `f` can tell if the mesh is correctly oriented: it is >0 if and only if the mesh is oriented outwardly.

Note: `f` cannot be vertical, because that would mean that all edges incident to `v` are vertical, and that would violate a precondition on the mesh (should not be auto-intersecting).

Cc: @afabri

janetournois:
## Summary of Changes

This PR fixes the implementation of `PMP::is_outward_oriented()`, as detailed by @lrineau in issue #2048.
The implementation is now based only on kernel predicates and does not require the construction of a normal anymore.

## Release Management

* Affected package(s): PMP
* Issue(s) solved: fix #2048
* This PR is based on PR #1795 (now closed as included in this PR) that corresponds to two small features: 
  - [`Kernel::Compare_slope_3`][slope]
  - [`Kernel::Angle_3`][angle]

  [slope]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::Compare_slope_3
  [angle]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::Angle_3

Note I've run the PMP tests locally and they work fine. The data given in the cgal-discuss thread has also been tested.


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -36,22 +35,27 @@ namespace CGAL {
...
+    typename GT::Compare_z_3 compare_z;
     VPmap vpmap;
 
-   Compare_vertex_points_xyz_3(VPmap const& vpmap)
-	: vpmap(vpmap){}
+    Compare_vertex_points_z_3(VPmap const& vpmap, const GT& gt)
+      : vpmap(vpmap)
+    {
+      compare_z = gt.compare_z_3_object();
+    }
```
lrineau: The following works also if `GT::Compare_z_3` does not have a default constructor (may happen with a geom traits with state):
```c++
Compare_vertex_points_z_3(VPmap const& vpmap, const GT& gt)
     : compare_z(gt.compare_z_3_object()), vpmap(vpmap)
{}
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -97,25 +101,55 @@ bool is_outward_oriented(const PolygonMesh& pmesh,
   VPMap vpmap = choose_param(get_param(np, vertex_point),
                              get_const_property_map(vertex_point, pmesh));
   //Kernel
-  typedef typename GetGeomTraits<PolygonMesh, NamedParameters>::type Kernel;
-
-  internal::Compare_vertex_points_xyz_3<typename Kernel::Less_xyz_3, VPMap >
-    less_xyz(vpmap);
+  typedef typename GetGeomTraits<PolygonMesh, NamedParameters>::type GT;
+  GT gt = choose_param(get_param(np, geom_traits), GT());
```
lrineau: just a remark: it seems that the API of PMP does not allow to get the geom traits by a reference. So the geom traits objects used by PMP have to be lightweight.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -97,25 +101,55 @@ bool is_outward_oriented(const PolygonMesh& pmesh,
...
-            normal_v_min[0] == 0 && (
-              normal_v_min[1] < 0  ||
-              ( normal_v_min[1]==0  && normal_v_min[2] < 0 )
-            )
-         );
+
+  internal::Compare_vertex_points_z_3<GT, VPMap> less_z(vpmap, gt);
+  typename boost::graph_traits<PolygonMesh>::vertex_iterator v_max_it
+    = std::max_element(vbegin, vend, less_z);
+  typename boost::graph_traits<PolygonMesh>::vertex_descriptor v_max = *v_max_it;
```
lrineau: I think something should be done for empty meshes: either add a predication `!empty`, or return `true` for empty meshes.

lrineau: > predication

?! I meant "precondition"...

janetournois:
thank you for the remarks, the 2 last commits should implement them

sloriot:
@lrineau Does it make sense to target 4.10 by hiding the doc changes in the Kernel? I agree that this will break if a user implements its own kernel but I have the feeling that they are more likely to be affected by the bug in the function rather than not respecting the concept.

lrineau:
This should indeed target 4.10. I had already flagged it so last week. 4.10 is in about a month. We have plenty of time to get the small features approved and the pull-requests merged.

lrineau:
There is conflict with `master` and it does not seem trivial to fix. I cannot test this PR.

janetournois:
I have rebased and resolved conflicts

lrineau:
Travis reports a compilation error: https://github.com/CGAL/cgal/pull/2057

lrineau:
Travis reports a compilation error: https://travis-ci.org/CGAL/cgal/jobs/223883331#L4665

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -40,30 +40,34 @@ namespace CGAL {
...
+
   template<typename Kernel, typename PM, typename NamedParameters>
   bool is_outward_oriented(typename boost::graph_traits<PM>::vertex_descriptor vd,
                            const PM& pmesh,
                            const NamedParameters& np)
   {
-    const typename Kernel::Vector_3&
-      normal_v_min = compute_vertex_normal(vd, pmesh, np);
+    const typename Kernel::Vector_3& normal_v_min
+      = CGAL::Polygon_mesh_processing::compute_vertex_normal(vd, pmesh, np);
```
lrineau: That function `is_outward_oriented(vertex_descriptor, Mesh)` is actually no longer used. It was used by `is_outward_oriented(Mesh)` but it is no longer, since the implementation changed.

janetournois: it is actually used in `PMP::does_bound_a_volume()` inside corefinement.h
Maybe it should not, but it is.

I will try to have a look whether this use can be replaced too

janetournois:
@sloriot Can you please confirm my last commit is correct?

lrineau:
I do not think so.

The `is_outward_oriented(mesh)` (without the vertex descriptor) assumes that there is only one connected component, and that the mesh is either fully oriented inward, or fully oriented outward.

But in the corefinement, the verification of the orientation seems to be done by connected component: there could be one component oriented outward, and another one oriented inward.

sloriot:
@lrineau is right, the patch is incorrect. I'll fix it on Monday.

sloriot:
With the last commit, we no longer need the `Angle_3` functor.

lrineau:
TODO: #1795 should be dealt with before this PR is merged.

lrineau:
@sloriot, @janetournois: this PR is for 4.10, but the PR #1795 has been prepared for 4.11.

My suggestion is the following: remove from this PR any change about `Angle_3`, and remove the documentation part about `Compare_slope_3`. That seems the easiest.

Another possibility: rebase #1795 onto `4.10-branch`, and then rebase this PR #2057 onto #1795.

sloriot:
I rebased #1795 onto 4.10-branch put the extra commit from this PR on top.

sloriot:
Looks like the implementation is still buggy. Don't merge the branch for now.

sloriot:
It's fixed.

lrineau:
This PR has conflicts with `master`. It cannot be tested easily in `integration`.

sloriot:
Working on it...

Kernel_23/include/CGAL/Kernel/global_functions_2.h
```diff
@@ -355,6 +355,8 @@ compare_slope(const Segment_2<K> &s1, const Segment_2<K> &s2)
   return internal::compare_slope(s1, s2, K());
 }
 
+
+#ifndef CGAL_NO_DEPRECATED_CODE
```
lrineau: Do not we have also a macro that use attributes to mark function as deprecated?

sloriot: None that I'm aware of.

lrineau: Its `CGAL_DEPRECATED` and `CGAL_DEPRECATED_MSG`, from `<CGAL/config.h>`.

Maybe that is not yet documented.

sloriot: Fixed

lrineau:
@janetournois @sloriot: This PR is incompatible with the PR #1891, already merged in `master`, that changes the namespace where named parameters are defined. Sébastien, you seem the one having changed the namespace (commit e9395aba6f1cc1a273e8625c168f2bdd3f1b66eb); can you fix the code of `orientation.h`, or tell Jane what should be changed?

The compilation error is:
```
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/orientation.h: In function 'bool CGAL::Polygon_mesh_processing::internal::is_outward_oriented(typename boost::graph_traits<Graph>::vertex_descriptor, const PolygonMesh&, const NamedParameters&)':
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/orientation.h:79:40: error: 'geom_traits' was not declared in this scope
     GT gt = choose_param(get_param(np, geom_traits), GT());
                                        ^~~~~~~~~~~
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/orientation.h:79:40: note: suggested alternatives:
In file included from /mnt/testsuite/include/CGAL/Polygon_mesh_processing/compute_normal.h:35:0,
                 from /mnt/testsuite/include/CGAL/Polygon_mesh_processing/orientation.h:29,
                 from /home/cgal_tester/build/src/cmake/platforms/Fedora-Release/test/Polygon_mesh_processing/orient_polygon_mesh_test.cpp:5:
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/internal/parameters_interface.h:23:41: note:   'geom_traits'
 CGAL_add_named_parameter(geom_traits_t, geom_traits, geom_traits)
                                         ^
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/internal/named_function_params.h:38:12: note: in definition of macro 'CGAL_add_named_parameter'
   enum X { Y };
            ^
In file included from /mnt/testsuite/include/CGAL/Polygon_mesh_processing/compute_normal.h:35:0,
                 from /mnt/testsuite/include/CGAL/Polygon_mesh_processing/orientation.h:29,
                 from /home/cgal_tester/build/src/cmake/platforms/Fedora-Release/test/Polygon_mesh_processing/orient_polygon_mesh_test.cpp:5:
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/internal/parameters_interface.h:23:54: note:   'CGAL::Polygon_mesh_processing::parameters::geom_traits'
 CGAL_add_named_parameter(geom_traits_t, geom_traits, geom_traits)
                                                      ^
/mnt/testsuite/include/CGAL/Polygon_mesh_processing/internal/named_function_params.h:100:3: note: in definition of macro 'CGAL_add_named_parameter'
   Z(const K& k)                                  \
   ^
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.11-Ic-14/Polygon_mesh_processing/TestReport_lrineau_Fedora-Release.gz

sloriot:
IIRC, it should be `internal_np::geom_traits`. However, we cannot do the fix in this branch since the namespace fix was done in master only. I can make the change locally in `integration`. After integration in 4.10, I can cherry-pick that commit after the merge of the branch 4.10 in master. What do you think?

lrineau:
I agree with your plan. I just pushed a big change to `integration`. Now, I let you deal with the merge of this PR in `integration`.

lrineau:
Actually, what are internal named parameters? Why is `geom_traits` one of those? What is the documented way for users to specify a custom geometric traits ?

sloriot:
the namespace change is an implementation detail and the user is not affected by it (it avoid pollution of the CGAL namespace IIRC, that is `CGAL::internal_np::geom_traits` instead of `CGAL::geom_traits`).

lrineau:
The name `internal_np` sounds like "internal stuff", as if the user should not use it. So, can the user specify the geom traits?

sloriot:
using `CGAL::Polygon_mesh_processing::parameters::geom_traits(MyTraits());`

sloriot:
Done in 55ef26b

lrineau:
Tested in CGAL-4.11-Ic-16.

PR #2065, opened 2017-04-19
WIP : fix PMP::compute_vertex_normal

### LINKED ISSUE #2047: Bug in PMP::compute_vertex_normal()
## Issue Details

There is a bug in [`PMP::compute_vertex_normal()`][doc]

  [doc]: http://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__normal__grp.html#ga240ed0a1f5efbea472f998fada07e112

For a vertex of a surface mesh, it computes the normalized normals of all incident faces, and compute the unweighted mean of those normalized normals. ... That is wrong!

Steps to easily see the bug:
- Load the OFF file `demo/Polyhedron/data/cube.off` (that is a simple triangulated unit cube: 8 vertices, 12 faces), in the polyhedron demo,
- Use "create a point set from vertices". That will display the normals.
- For each vertex that is incident to four faces, you will see that the normal is wrong : all vertex normals of the 8 vertices of the cube should be proportional to the diagonal vectors (±1, ±1, ±1).

There are several methods to compute vertex normals of a mesh, and the unweighted mean is clearly always wrong (but if the mesh is very isotropic). See details our three different methods [in this blog][blog]. The page points to articles, one comparing six different methods to compute normals. The best one seems to weight the mean with the angles of the face around the vertex.

  [blog]: http://meshlabstuff.blogspot.fr/2009/04/on-computation-of-vertex-normals.html

## Environment
* CGAL version: current `master` branch (pre-CGAL-4.10-beta1, commit 82286eb79181a54158fdddc4d78e06d56f0e1f1f)

janetournois:
This PR should be discussed in the issue before going further.

## Summary of Changes
This PR intends to improve the `PMP::compute_vertex_normal(v)` function, by changing the weights of face normals that are summed to compute a normal at vertex `v`.

The former implementation used `1` as weight, and the new weight is `sin(angle at v in f)`.
It is cheaper than computing `(angle at v in f)`, and should behave similarly.

## Release Management

* Affected package : PMP
* Issue solved : fixes #2047

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
@@ -261,7 +261,8 @@ compute_vertex_normal(typename boost::graph_traits<PolygonMesh>::vertex_descript
                            get(vpmap, target(next(he, pmesh), pmesh)));
       //v(i) and v(i+1) must me seen in ccw order, from v, so we reverse v1 and v2
       Vector ni = traits.construct_cross_product_vector_3_object()(v2, v1);
-      ni = traits.construct_scaled_vector_3_object()(ni, CGAL::sqrt(1./(sqlen(v1)*sqlen(v2))));
+      ni = traits.construct_scaled_vector_3_object()(ni,
+        CGAL::approximate_sqrt(FT(1)/(sqlen(v1)*sqlen(v2))));
 
       normal = traits.construct_sum_of_vectors_3_object()(normal, ni);
     }
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
@@ -229,25 +229,23 @@ compute_vertex_normal(typename boost::graph_traits<PolygonMesh>::vertex_descript
                       )
 {
   using boost::choose_param;
+  using boost::get_param;
 
   typedef typename GetGeomTraits<PolygonMesh, NamedParameters>::type GT;
   typedef typename GT::Vector_3 Vector;
+  typedef typename GT::FT       FT;
   GT traits = choose_param(get_param(np, CGAL::geom_traits), GT());
 
-  typedef typename GetFaceNormalMap<PolygonMesh, NamedParameters>::NoMap DefaultMap;
-  typedef typename boost::lookup_named_param_def <
-    CGAL::face_normal_t,
-    NamedParameters,
-    DefaultMap> ::type FaceNormalMap;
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
```diff
@@ -257,9 +255,15 @@ compute_vertex_normal(typename boost::graph_traits<PolygonMesh>::vertex_descript
...
-      Vector n = fnmap_valid ? get(fnmap, face(he, pmesh))
-                             : compute_face_normal(face(he, pmesh), pmesh, np);
-      normal = traits.construct_sum_of_vectors_3_object()(normal, n);
+      Vector v1 = vector_3(get(vpmap, v /*target(he, pmesh)*/),
+                           get(vpmap, source(he, pmesh)));
+      Vector v2 = vector_3(get(vpmap, v /*target(he, pmesh)*/),
+                           get(vpmap, target(next(he, pmesh), pmesh)));
+      //v(i) and v(i+1) must me seen in ccw order, from v, so we reverse v1 and v2
+      Vector ni = traits.construct_cross_product_vector_3_object()(v2, v1);
+      ni = traits.construct_scaled_vector_3_object()(ni, CGAL::sqrt(1./(sqlen(v1)*sqlen(v2))));
```
lrineau: The expression `CGAL::sqrt(1./(sqlen(v1)*sqlen(v2)))` does not compile with `Gmpq` as the `FT`. That error was detected by Travis: https://travis-ci.org/CGAL/cgal/jobs/223571577#L4670

Did you get a mail from Travis, by the way?

janetournois: right. I should use `CGAL;;approximate_sqrt`. 
Just a remark : the documentation says that, even with an exact kernel, the sqrt inside will involve non exact computations.

And No, I did not get an email from Travis.

lrineau: You should also write `FT(1)/x` instead of `1./x`, so that it compiles with number types that cannot be constructed from a `double`.

lrineau:
This PR conflicts with `master`... anyway, if I remember well, it waits for a larger discussion among us.

lrineau:
@janetournois, do you know why this pull-request was left unattended? It seems there was a discussion, and then... I do not know.

janetournois:
Yes I remember.
We started a discussion in issue #2047,  but it was not conclusive to this PR stayed as is.

Imho it's not obsolete and the implemented solution should be improved at some point.

sloriot:
Note that @MaelRL has a very good method implemented for another project and he will at some point export it properly.

janetournois:
Sounds good!

PR #2070, opened 2017-04-21
Polyhedron_demo: Use a recent OpenGL context if possible

### LINKED ISSUE #2055: Polyhedron_demo: Use a recent opengl context when possible
## Issue Details
It should be possible to implement a recent OpenGL context and check if it is valid. If not, keep the 2.1, but in the case it is possible, it would allow us to use geometry shaders and many features that would be really appreciated. 
This would permit to fix the polyhedron_item's flat mode with draw one side problem, and make it easier to make decent transparency in the future.

maxGimeno:
## Summary of Changes
Set an OpenGL 4.3 format if possible and create a QOpenGLFunctions_4_3_Compatibility in that case.
Use this recent context to link geometry shaders in order to use an efficient flat shading in polyhedron_item and to fix the visual problems remaining in #950.
## Release Management

* Issue(s) solved (if any): fix #2055

### DIFF
--- a/.travis/build_package.sh
+++ b/.travis/build_package.sh
@@ -19,7 +19,7 @@ function build_demo {
   cd build-travis
   if [ $NEED_3D = 1 ]; then
     #install libqglviewer
-    git clone --depth=4 -b v2.6.3 --single-branch https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
+    git clone --depth=1 https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
     pushd ./qglviewer/QGLViewer
     #use qt5 instead of qt4
 #    export QT_SELECT=5
--- a/.travis/build_package.sh
+++ b/.travis/build_package.sh
@@ -19,7 +19,7 @@ function build_demo {
   cd build-travis
   if [ $NEED_3D = 1 ]; then
     #install libqglviewer
-    git clone --depth=1 https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
+    git clone --depth=4 -b v2.6.3 --single-branch https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
     pushd ./qglviewer/QGLViewer
     #use qt5 instead of qt4
 #    export QT_SELECT=5
--- a/Polyhedron/demo/Polyhedron/Viewer.h
+++ b/Polyhedron/demo/Polyhedron/Viewer.h
@@ -151,8 +151,8 @@ protected:
   double prev_radius;
 
 public:
-  bool isOpenGL_4_3() const;
-  QOpenGLFunctions_4_3_Compatibility* recentFunctions();
+  bool isOpenGL_4_3() const Q_DECL_OVERRIDE;
+  QOpenGLFunctions_4_3_Compatibility* openGL_4_3_functions() Q_DECL_OVERRIDE;
 
 }; // end class Viewer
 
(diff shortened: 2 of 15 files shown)


sgiraudot:
Tested on my computer, no regression and no crash.

Polyhedron/demo/Polyhedron/Viewer.h
```diff
@@ -127,6 +127,10 @@ public Q_SLOTS:
   Viewer_impl* d;
   double prev_radius;
 
+public:
+  bool isRecent() const;
+  QOpenGLFunctions_4_3_Compatibility* recentFunctions();
```
lrineau: The name `recentFunctions` is badly chosen. Imagine you read that doc in 10 years... It should be, for example, `openGLFunctions_4_3`…

lrineau:
This PR conflicts with other PR in integration. It will not be tested in CGAL-4.11-Ic-42.

Polyhedron/demo/Polyhedron/Viewer.h
lrineau: Thanks for the renaming of `isRecent()` to `isOpenGL_4_3()`.

My remark about the name of `recentFunctions()` remains: it must be renamed to something like `openGLFunctions_4_3`.

lrineau:
... This PR has also conflicts with `master`.

.travis/build_package.sh
```diff
@@ -19,7 +19,7 @@ function build_demo {
   cd build-travis
   if [ $NEED_3D = 1 ]; then
     #install libqglviewer
-    git clone --depth=1 https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
+    git clone --depth=4 -b v2.6.3 --single-branch https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
```
sloriot: is this modification intended?

maxGimeno: Probably not...

maxGimeno:
rebased

.travis/build_package.sh
lrineau: Yes it was intended, because of the problems we had with libQGLViewer 2.7.0. Maxime did the patch in #2186 , and I cherry-picked it in all pull-requests, so that the Travis status of the pull-requests could be green again.

sloriot:
Test failing, see [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-9/Polyhedron_Demo/TestReport_lrineau_ArchLinux-clang.gz)

maxGimeno:
This is getting complicated. We now need to manage different cases according to qglviewer's version, which treat different cases according to available OpenGLVersions. Maybe we should consider dropping support for old versions of qglviewer at some point, when the QOpenGLWidget environment is considered stable...

sloriot:
What versions would you like to support?

maxGimeno:
In a better world, none. We just get rid of it and re-implement our own viewer system.
An alternative would be to restrict our support to 2.7, our none of this if my fix holds on long enough. It could be ok like that, but it would require an intensive testing session to be sure of that.

sloriot:
This is not reasonable, 2.7 was released in June. Most distributions haven't yet integrated it.

maxGimeno:
That is why I said "when the QOpenGLWidget environment is considered stable".

maxGimeno:
@lrineau Is the new name "opengl_4_3_functions" ok ?

lrineau:
Yes, `opengl_4_3_functions` follows the CGAL naming convention.

lrineau:
Is this PR ready to be tested today?

maxGimeno:
I think so

PR #2071, opened 2017-04-21
Polyhedron demo : improve .surf reader

janetournois:
## Summary of Changes

This PR introduces improvement to the `.surf` reader, and the corresponding plugins.
- deal with non-manifold surfaces (there were holes in the generated surfaces, next to non-manifold configurations)
- number materials from `0` (Exterior) to `n-1` to make sure Mesh_3 will not get into trouble
- ignore isolated vertices
- reduce memory footprint

## Release Management

* Affected packages : Polyhedron demo, PMP (internal code only)

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
@@ -29,6 +29,7 @@
 #include <CGAL/Polygon_mesh_processing/orient_polygon_soup.h>
 #include <CGAL/algorithm.h>
 #include <set>
+#include <boost/dynamic_bitset.hpp>
 
 #include <boost/range/size.hpp>
 #include <boost/range/value_type.hpp>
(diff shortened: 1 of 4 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
```diff
@@ -67,13 +68,29 @@ class Polygon_soup_to_polygon_mesh
       _polygons(polygons)
   { }
 
-  void operator()(PM& pmesh)
+  void operator()(PM& pmesh, const bool insert_isolated_vertices = true)
```
lrineau: I am not a big fan of Boolean parameters. When one see a call `polygon_soup_to_polygon_mesh(mesh, true)`, its hard to know, and to remember, if the `true` means "keep the isolated vertices" or "remove the isolated vertices". Explicit enumerators allows to see call like `polygon_soup_to_polygon_mesh(mesh, REMOVE_ISOLATED_VERTICES)` which is a lot more explicit.

That is just a remark, for a later time.

janetournois: ok. This is just internal code, it should not be used much

Issue #2069, opened 2017-04-21
Covariance matrix assembly in CGAL is very unefficient


sgiraudot:
## Issue Details

I just noticed that applying `CGAL::linear_least_squares_fitting_3` to a big set of triangles was suspiciously long. Using _Valgrind_, I saw that most of the time was spent on `new` and `delete` operators. It seems that the internal code to assemble the covariance matrix uses `CGAL::Linear_algebraCd` were all matrices are dynamically allocated (and as one matrix is computed per triangle and multiplied, the penalty is high), see in `Principal_component_analysis/include/CGAL/PCA_util.h`.

I tried to replace the few lines of linear algebra using the `Eigen` library (matrices are static) and, well, it's about __200 times faster__ (yes, 200 times, from 2 minutes to 0.6 seconds on my test). I'm obviously going to make a proper branch and submit a PR, but I'm not sure how we want to do this:

- PCA is an old package that can use Eigen (`Eigen_diagonalize_traits`) but also has a fallback mode using internal diagonalization code (`Diagonalize_traits`)
- For backward compatibility, we should keep this fallback mode
- My first idea was to specialize the functions (when `DiagonalizeTraits` is `Eigen_diagonalize_traits`) but there are already many functions and code duplicate (for handling points, triangles, cuboids, etc.). Sadly, `CGAL::Linear_algebraCd` and `Eigen` do not have the exact same API (matrix access is `m[][]` in CGAL and `m(,)` in Eigen) so we can't do a minimal `#ifdef/#else` somewhere
- Maybe add a layer of abstraction just to bind the few linear algebra operations either to `CGAL::Linear_algebraCd` and `Eigen`?

Also, I'm not sure where `CGAL::Linear_algebraCd` is used in CGAL, but it seems to me that in many cases, matrices have a known size at compilation time, so it could also be a good idea to provide matrices with template size parameters to avoid the numerous allocation/disallocation. But that's probably more complicated than that, so maybe not a good idea to change this code.

__Note:__ just for the record, this problem is not present when using PCA on a set of points (there is no linear algebra involved here), which is the most widely used case.

lrineau:
`Linear_algebraCd` is used mainly in the old kernel `Carterian_d` (and relatives), in Jet_fitting, and PCA:
```shell
renoir ~/Git/cgal $ ack Linear_algebraCd ^build*/(include|examples|demo) -l
Jet_fitting_3/include/CGAL/Monge_via_jet_fitting.h
[...]

PR #2074, opened 2017-04-24
Principal component analysis: Enhance matrix assembly

sgiraudot:
## Summary of Changes

- Specialization of `assemble_covariance_matrix_3()` when `CGAL::Eigen_diagonalize_traits` is used, so that we stop relying on `Linear_algebraCd` internally (which is very slow) and use the Eigen library instead (which is by definition available if `CGAL::Eigen_diagonalize_traits` is used)
- Addition of lines in the test files so that both the Eigen and internal CGAL versions are tested
- Addition of notes in the doc to advise the user to use Eigen (which is much faster)

## Release Management

* Affected package(s): Principal Component Analysis
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2069


lrineau:
For 4.10 or 4.11?

sgiraudot:
4.10 would be nice.

Do I rebase?

lrineau:
No need. It seems your history does not contain any commit that is only in `master`.

PR #2075, opened 2017-04-24
Do not include graph_traits_{Polyhedron | Surface_mesh}.h

afabri:
## Summary of Changes

In the examples we no longer include the files that define the graph_traits and properties for Polyhedron_3 and Surface_mesh, as their header files already include those files.


afabri:
Successfully tested in CGAL-4.10-Ic-188

PR #2077, opened 2017-04-26
BGL Add a description to doxygen groups

afabri:
## Summary of Changes

For each group there is a line in a generated table of the entry page of the reference manual.
* Adding a sentence improves this table
* On the page of the group with functions I added the name of the header file, as when the functions only have a short dfescription, the line with the `#include` is not added automatically. 

## Release Management

* Affected package(s): BGL


afabri:
@lrineau I added it for testing to CGAL-4.11-Ic-12 as  the documentation of 4.10-Ic-190  is no longer accessible.

PR #2082, opened 2017-04-26
Polyhedron_demo: Fix segfault in Scene::replaceItem().

### LINKED ISSUE #2080: Polyhedron demo segfault
## Issue Details

Load an off file
Add a selection item
Reload the off file   (I wanted to reload after having selected/isotropic remeshed)

seg fault.

@maxGimeno I guess you must first remove or clear the selection item.

maxGimeno:
## Summary of Changes

The `itemAboutToBeDestroyed()` signal is emitted in `replaceItem()`, so the d-pointer of some items gets deleted, but is called in `isEmpty()`, which is also called in `replaceItem()` when the Bbox is recomputed.

The fix is to move the call to `itemAboutToBeDestroyed()` after the item is used for the last time in `replaceItem()`.

## Release Management

* Issue(s) solved (if any): fix #2080

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene.cpp
@@ -91,10 +91,6 @@ Scene::replaceItem(Scene::Item_id index, CGAL::Three::Scene_item* item, bool emi
     if(index < 0 || index >= m_entries.size())
         return 0;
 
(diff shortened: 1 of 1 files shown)


sloriot:
The PR must be rebased onto the 4.10 branch since it target 4.10

maxGimeno:
done

PR #2091, opened 2017-04-29
kd-tree removal + misc improvements

mglisse:
## Summary of Changes

This adds a function to remove points from a kd-tree. It also includes various unrelated optimizations to the kd-tree, so it would be good if you could have a quick look to check that you don't find one of them objectionable.

## Release Management

* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Remove_from_Kd_tree

### DIFF
--- a/Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
+++ b/Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
@@ -109,7 +109,7 @@ explicit call to `build()` or implicitly by the next query or removal. The only
 reason to call this function explicitly is to rebalance the tree after some
 number of removals.
 */
-void invalidate_built();
+void invalidate_build();
 /// @}
 
 /// \name Operations
--- a/Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
+++ b/Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
@@ -96,12 +96,20 @@ template <class InputIterator> Kd_tree(InputIterator first, InputIterator beyond
 The constructor does not build the internal data structure, and it
 is also not updated after calls to `insert()`.  
 The method `build()` is called implicitly
-at the first call to a query member function. You can call
+at the first call to a query or removal member function. You can call
 `build()` explicitly to ensure that the next call to
 query functions will not trigger the reconstruction of the
 data structure.
 */
 void build();
+
+/*!
+This clears the internal data structure, which then gets rebuilt either by an
+explicit call to `build()` or implicitly by the next query or removal. The only
+reason to call this function explicitly is to rebalance the tree after some
+number of removals.
+*/
+void invalidate_built();
 /// @}
 
 /// \name Operations
@@ -109,6 +117,9 @@ void build();
 
 /*!
 Inserts the point `p` in the `k-d` tree.
+\note Insertions do not dynamically update the internal data structure. The
+next query, or a call to `build()`, automatically triggers a rebuild of the
+whole structure.
 */
 void insert(Point_d p);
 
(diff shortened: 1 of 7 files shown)


Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
```diff
@@ -96,19 +96,30 @@ template <class InputIterator> Kd_tree(InputIterator first, InputIterator beyond
...
 */
 void build();
+
+/*!
+This clears the internal data structure, which then gets rebuilt either by an
+explicit call to `build()` or implicitly by the next query or removal. The only
+reason to call this function explicitly is to rebalance the tree after some
+number of removals.
+*/
+void invalidate_built();
```
sloriot: Wouldn't `invalidate_build` be a better name?

Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -339,6 +355,98 @@ class Kd_tree {
...
+#if 0
+    // This code could have quadratic runtime.
+    if (!is_built()) {
+      std::vector<Point_d>::iterator pi = std::find(pts.begin(), pts.end(), p);
+      // Precondition: the point must be there.
+      CGAL_assertion (pi != pts.end());
+      pts.erase(pi);
+      return;
+    }
+#endif
```
sloriot: do you want to keep the commented bloc?

mglisse: I'd like to keep at least a comment to that effect. Keeping the commented code avoids having to paraphrase what it does, but I can replace that with a short paragraph of explanation if you prefer.

Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
mglisse: I don't have a strong preference, we invalidate what was built, we invalidate the build, whatever. Maybe I was inspired by is_built(). Should I rename to invalidate_build?

Spatial_searching/include/CGAL/Kd_tree.h
sloriot: I just wanted to be sure it was left on purpose.

Spatial_searching/doc/Spatial_searching/CGAL/Kd_tree.h
sloriot: `invalidate_build()` sounds better to me.

mglisse: Done. (actually, this function has been there for a long time, I just documented it)

Spatial_searching/test/Spatial_searching/remove.cpp
```diff
@@ -0,0 +1,29 @@
...
+  t.remove(Point(3,4));
+  t.insert(Point(5,5));
+  t.build();
+  assert(t.size()==2);
+  t.clear();
+  for(int i=0;i<1000;++i)
+    t.insert(Point(i,-i));
+  for(int i=0;i<1000;++i)
+    t.remove(Point(i,-i));
+  t.print();
```
lrineau: This test should use `assert`, to check something (like the number of remaining points in the tree, and the closest point of a known query) because, in the testsuite, a few columns of the testsuite compile and run the tests with `CGAL_NDEBUG`.

lrineau:
The three Windows Debug columns of the testsuite both show a runtime error:
```
- ProgramOutput.remove.x64_Cygwin-Windows8_MSVC2012-Debug-64bits
CGAL::Random()::get_seed() = 1495250632
C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\INCLUDE\vector(72) : Assertion failed: vector iterator not dereferencable
```
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-20.shtml#Spatial_searching

afabri:
@mglisse   I will have a look.

mglisse:
Hmm, indeed, I can reproduce that with -D_GLIBCXX_DEBUG, my fault for not checking. I'll look at it, thanks.

mglisse:
Hmm, build() is not supposed to be called when the list of points is empty? It assumes there is at least one point. I'll see about adding an early escape / adding an assertion / not calling build in that case.

mglisse:
Apparently people don't often try to print an empty Kd_tree... At least that was easy to fix 😄 .

afabri:
@mglisse  Do you have got the green light from the Inria technology-transfer department, that they don't want IP of your small scale contributions to the kd-tree owned by Utrecht University?

mglisse:
I only have informal talk, no paperwork.

PR #2072, opened 2017-04-21
Mesh 3 : small improvements

janetournois:
## Summary of Changes
This PR introduces small improvements that we have recently made to the `Mesh_3` package, ranging from small bug fixes to time and memory footprint improvement.

## Release Management

* Affected package : Mesh_3

PR #2097, opened 2017-05-03
WIP: Mesh_3: another bug-fix with time stamps

lrineau:
A bug-fix selected for CGAL-4.9.2 and CGAL-4.10: because the time stamp of a `Vertex` can change while the `Vertex_handle` is inserted in a sorted container (`set`/`map`), that container can be corrupted.

Now the modified vertex handles are also removed from `unchecked_vertices_`.

Work-in-progress... TODO:
  - [ ] remove the two.cpp
  - [ ] fix the bug when a c3t3 is copied.


lrineau:
This PR is replaced by #2072, that will be merged into 4.10 (and not in 4.9.2).

PR #2099, opened 2017-05-05
Polyhedron_demo: Fix segfault in distance_item 

maxGimeno:
## Summary of Changes
If a facet is too small, a precision problem will cause it's triangulation to fail. This PR fixes the resulting segfault by stopping the triangulation process if there are too few points, and not display the facet. As it only happens for facets with a diagonal smaller than 0.0001 times the item's Bbox's diagonal, the user should not even notice it unless he(she) is looking for it. 
A message is displayed in the console in that case.
## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/triangulate_primitive.h
+++ b/Polyhedron/demo/Polyhedron/triangulate_primitive.h
@@ -103,6 +103,11 @@ public:
      std::cerr<<"Facet not displayed"<<std::endl;
 
   }
+  ~FacetTriangulator()
+  {
+   if (cdt )
+     delete cdt;
+  }
 
 private:
   bool triangulate( std::vector<PointAndId > &idPoints,
@@ -128,11 +133,11 @@ private:
      typename CDT::Vertex_handle vh;
      //Always insert the first point, then only insert
      // if the distance with the previous is reasonable.
-     if(first == 0 || CGAL::squared_distance(idPoint.point, previous->point()) > min_sq_dist)
+     if(first == typename CDT::Vertex_handle() || CGAL::squared_distance(idPoint.point, previous->point()) > min_sq_dist)
      {
        vh = cdt->insert(idPoint.point);
        v2v[vh] = idPoint.id;
-       if(first == 0) {
+       if(first == typename CDT::Vertex_handle()) {
          first = vh;
        }
        if(previous != 0 && previous != vh) {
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/triangulate_primitive.h
```diff
@@ -82,25 +83,29 @@ class FacetTriangulator
...
   {
-   triangulate_with_points(idPoints, more_points, normal, item_diag);
+   if(!triangulate_with_points(idPoints, more_points, normal, item_diag))
+     std::cerr<<"Facet not displayed"<<std::endl;
+
   }
 
 private:
-  void triangulate( std::vector<PointAndId > &idPoints,
+  bool triangulate( std::vector<PointAndId > &idPoints,
```
lrineau: As far as I have seen There is a memory leak in that class:
```C++
 cdt = new CDT(cdt_traits);
```

Why is `cdt` a member pointer of the class, instead of a non-pointer variable local to the function?

Polyhedron/demo/Polyhedron/triangulate_primitive.h
```diff
@@ -145,6 +150,8 @@ class FacetTriangulator
      previous = vh;
      }
     }
+    if(last_inserted == 0)
```
lrineau: A `Vertex_handle` is not supposed to be comparable to `0`. I may compile, but the correct code is `last_inserted==Vertex_handle()`.

Polyhedron/demo/Polyhedron/triangulate_primitive.h
maxGimeno: It is a member so it can be accessed in the item's triangulated function, to get the resulting finite_faces.

lrineau: And what about the memory ownership? Which object is responsible for that memory allocation and should deallocate it?

You should use `std::shared_ptr` and `std::make_shared`. In the Polyhedron demo, C++11 is allowed, and you should no longer call `new` and `delete` anymore.

maxGimeno: The functor is responsible for it. The triangulate() function is private, it is only called by the constructors, so there is only one CDT per instance. I juste added a destructor with a delete.

PR #2100, opened 2017-05-05
PMP : fix isotropic remeshing of a selection

janetournois:
## Summary of Changes

Fix a bug when the face range to remesh is incident to the boundary of the mesh.

Vertices that are at the "corner" of the face range to be remeshed,
i.e. on the border of the face range, and also on the border of the mesh,
incident to both the selected and unselected faces, should be constrained,
to avoid making a "axe blow" in the border

## Release Management
* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -303,6 +303,8 @@ namespace internal {
     {
       tag_halfedges_status(face_range); //called first
 
+      constrain_patch_corners(face_range);
+
       BOOST_FOREACH(face_descriptor f, face_range)
       {
         input_triangles_.push_back(triangle(f));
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -1372,6 +1374,43 @@ namespace internal {
         return PMP::compute_face_normal(f, mesh_);
     }
 
+    template <typename FaceRange>
+    void constrain_patch_corners(const FaceRange& face_range)
+    {
+      std::set<vertex_descriptor> visited;
```
lrineau: After all the issues we have had with the inefficiency of `std::set`, do we really want a `std::set`, instead of an `boost::unordered_set` or a `boost::flat_set`?

janetournois: you are absolutely right

PR #2104, opened 2017-05-10
Polyhedron_demo: Fix draw two sides for Polyhedron

### LINKED ISSUE #2103: Add a "draw two side" option in the context menu of Polyhedron item
When ticked, it will fill the vector of normals and use the same shader as for polygon soups. When unticked, the vector of normals is cleared. Special care need to be taken when multicolors are ON.

maxGimeno:
## Summary of Changes
The PROGRAM_FLAT allows to save memory by not keeping the flat normals and computing them in real time in the fragment shader, but it loses the orientation of the faces. To fix it, we decided to drop this enhancement in the case the draw_two_sides option is set, and keep computing the flat normals.

## Release Management

* Issue(s) solved (if any): fix #2103 

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
@@ -117,7 +117,7 @@ struct Scene_polyhedron_item_priv{
   }
 
 
-  void compute_normals_and_vertices(const bool colors_only = false, const bool is_two_side = false) const;
+  void compute_normals_and_vertices(const bool colors_only = false, const bool draw_two_sides = false) const;
   bool isFacetConvex(Facet_iterator, const Polyhedron::Traits::Vector_3&)const;
 
   void triangulate_convex_facet(Facet_iterator,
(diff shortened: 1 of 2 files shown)


maxGimeno:
Travis failing is not due to this code.

lrineau:
> Unable to connect to ppa.launchpad.net:http:

I have restarted the failed job.

lrineau:
... and the restarted job worked. The Travis status of this PR is now green.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -117,15 +117,15 @@ struct Scene_polyhedron_item_priv{
   }
 
 
-  void compute_normals_and_vertices(const bool colors_only = false) const;
+  void compute_normals_and_vertices(const bool colors_only = false, const bool is_two_side = false) const;
```
lrineau: I do not really understand why the parameter here is named `is_two_side`. At the very least, there should be an 's': `is_two_sides`.

Anyway, there is already a good name: `draw_two_sides`.

Polyhedron/demo/Polyhedron/Scene_polyhedron_item.cpp
```diff
@@ -491,7 +500,14 @@ Scene_polyhedron_item_priv::initialize_buffers(CGAL::Three::Viewer_interface* vi
 {
     //vao containing the data for the facets
     {
-    program = item->getShaderProgram(Scene_polyhedron_item::PROGRAM_FLAT, viewer);
+    if(viewer->property("is_two_sides").toBool())
```
lrineau: Same: `draw_two_sides`

PR #2105, opened 2017-05-10
prevent the merge of two vertices that would create a non-manifold edge

sloriot:
Without this fix the stitching can create non-manifold edges by merging two vertices that are already incident to the same vertex. This PR avoid doing this.


Polygon_mesh_processing/test/Polygon_mesh_processing/test_stitching.cpp
```diff
@@ -78,6 +79,7 @@ int main()
   test_surface_mesh("data_stitching/multiple_incidence.off");
   test_surface_mesh("data_stitching/incidence_3.off");
   test_polyhedron<Epic>("data_stitching/incoherent_patch_orientation.off");
+  test_polyhedron<Epic>("data_stitching/non_stitchable.off");
```
lrineau: Why add the same call twice?

lrineau: Same for the other line, that is also twice:
```c++
test_polyhedron<Epic>("data_stitching/incoherent_patch_orientation.off");
```

sloriot:
The patch is not sufficient, we must check that the intersection of the link of the two vertices to be merged is empty (and not only considering the first incident edge)

sloriot:
I fixed the test.

PR #2126, opened 2017-05-18
Polyhedron_demo: Fix empty group_item RenderingMode

maxGimeno:
## Summary of Changes

If a group_item is empty, supportsRenderingMode() must return true, not false, or else the scene will enter an infinite loop as it does not expect any item returning false to every renderingMode.

## Release Management


lrineau:
Is that a bug serious that we should merge in into 4.10 or 4.10.1?

lrineau:
@maxGimeno, I ask it **again** (https://github.com/CGAL/cgal/pull/2126#issuecomment-302661055):
> Is that a bug serious that we should merge in into 4.10 or 4.10.1?

lrineau:
Alright. I have cherry-picked the lone commit in the branch, as the commit 7cf7158c330312ce1b0022fed6016e93b714d971.

PR #2127, opened 2017-05-19
Kernel_23: Add static filtering for Plane_3

afabri:
## Summary of Changes

Avoid the usage of `Filtered_predicate` if possible.

## Release Management

### DIFF
--- a/Filtered_kernel/include/CGAL/internal/Static_filters/Is_degenerate_3.h
+++ b/Filtered_kernel/include/CGAL/internal/Static_filters/Is_degenerate_3.h
@@ -35,6 +35,7 @@ class Is_degenerate_3
 {
   typedef typename K_base::Ray_3     Ray_3;
   typedef typename K_base::Segment_3 Segment_3;
+  typedef typename K_base::Plane_3 Plane_3;
   typedef typename K_base::Is_degenerate_3 Base;
   typedef typename K_base::Construct_source_3 Construct_source_3;
   typedef typename K_base::Construct_target_3 Construct_target_3;
@@ -70,6 +71,26 @@ public:
   {
     return Equal_3()(Construct_source_3()(r), Construct_second_point_3()(r));
   }
+
+  result_type 
+  operator()(const Plane_3& p) const
+  {
+    CGAL_BRANCH_PROFILER(std::string("semi-static attempts/calls to   : ") +
+                         std::string(CGAL_PRETTY_FUNCTION), tmp);
+
+    Get_approx<Plane_3> get_approx; // Identity functor for all planes
+                                    // but lazy planes
+    double a, b, c;
+
+    if (fit_in_double(get_approx(p).a(), a) && fit_in_double(get_approx(p).b(), b) &&
+        fit_in_double(get_approx(p).c(), c) )
+    {
+      CGAL_BRANCH_PROFILER_BRANCH(tmp);
+      return a == 0 && b == 0 && c == 0;
+    }
+    return Base::operator()(p);
+  }
+
 
 }; // end class Is_degenerate_3
 


mglisse:
Just checking: the kernel for which you are trying to speed things up is Epick? Overloaded functors are so inconvenient, we can't handle Is_degenerate_3 the same as Less_x_2 just because of triangles and tetrahedra...
With `g++ -ffinite-math-only`, I think we generate the same code for Epick before/after your patch 😉 (we could also tweak Interval_nt.h, replacing a<b with !(a>=b) in a few places, so that `-fno-trapping-math` is sufficient).

afabri:
@lrineau Can we discuss Marc's comments.

lrineau:
@mglisse, in https://github.com/CGAL/cgal/pull/2127#issuecomment-302966048:
> Just checking: the kernel for which you are trying to speed things up is Epick?

Yes, mostly in Epick, but static filters will also benefit to Epeck.

> With g++ -ffinite-math-only, I think we generate the same code for Epick before/after your patch

I do not understand. In `master`, `Epick::Is_degenerate_3` is a `Filtered_predicate`, that does:
```
Protect_FPU_rounding<Protection> p;
      try
	{
	  Ares res = ap(c2a(args)...);
	  if (is_certain(res))
	    return get_certain(res);
	}
      catch (Uncertain_conversion_exception) {}
```
where `ap(c2a(args)...)` call the `Is_degenerate_3` predicates with intervals.

How could the compiler reduce that code to a simple call to `a==0 && b==0 && c ==0`? I'd like to see the assembly code.

mglisse:
```

typedef CGAL::Epick K;
bool marc(K::Plane_3 p, K k){
  return k.is_degenerate_3_object()(p);
}
```
compiled with `g++ -DNDEBUG -O3 -ffinite-math-only`
```
_ZNK4CGAL18Filtered_predicateINS_20CommonKernelFunctors15Is_degenerate_3INS_16Simple_cartesianINS_4GmpqEEEEENS2_INS3_INS_11Interval_ntILb0EEEEEEENS_19Cartesian_converterINS_21Type_equality_wrapperINS_27Cartesian_base_no_ref_countIdNS_5EpickEEESE_EES5_NS_12NT_converterIdS4_EEEENSB_ISG_S9_NSH_IdS8_EEEELb1EEclIJNS_7Plane_3ISE_EEEEEbDpRKT_:
.LFB16032:
        .cfi_startproc
        .cfi_personality 0x3,__gxx_personality_v0
        stmxcsr -4(%rsp)
        movl    -4(%rsp), %ecx
        andl    $24576, %ecx
        stmxcsr -4(%rsp)
        movl    -4(%rsp), %eax
        andb    $-97, %ah
        orb     $64, %ah
        movl    %eax, -4(%rsp)
        ldmxcsr -4(%rsp)
        pxor    %xmm0, %xmm0
        movsd   16(%rsi), %xmm2
        movsd   8(%rsi), %xmm1
        comisd  (%rsi), %xmm0
        sete    %al
        jne     .L16
        comisd  %xmm0, %xmm1
        sete    %al
        jne     .L16
        comisd  %xmm0, %xmm2
        sete    %al
.L16:
        stmxcsr -4(%rsp)
        movl    -4(%rsp), %edx
        andb    $-97, %dh
        orl     %ecx, %edx
        movl    %edx, -4(%rsp)
        ldmxcsr -4(%rsp)
        ret
        .cfi_endproc
```
`pxor    %xmm0, %xmm0` puts 0 in xmm0, then 3 doubles are compared to it (`comisd`). Sadly, the rounding-mode manipulations (`stmxcsr`, `ldmxcsr`) are still there, and skipping those is indeed worth writing a special case of static filter.
The simplification is not very surprising, converting a double to an interval just turns x into a pair `(x,x)`, then testing if this is 0 does `if(x<0||x>0)return false;if(x==0&&x==0) return true;unreachable;` where `-ffinite-math-only helps prove that the end is unreachable, and that `x<0||x>0` is `x!=0`.

PR #2132, opened 2017-05-22
force the usage of CGAL compare function

sloriot:
Fixes [these errors](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10-I-194/Polygon_mesh_processing/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.1_Release-LEDA-without-GMP.gz).


lrineau:
That is probably a blocker for CGAL-4.10.

PR #2133, opened 2017-05-22
Silent cmake warnings about CMP0053

maxGimeno:
## Summary of Changes

Explicitly set the policy of CMP0053 to OLD to silent the warning when it is not found.

## Release Management

* Affected package(s): All demos


lrineau:
Can you please rebase onto `releases/CGAL-4.10-branch`? That would also remove the warnings from the CGAL-4.10-* testsuites.

PR #2137, opened 2017-05-24
Polygon Mesh Processing bug fixes

sloriot:
* only insert non-degenerate faces in the AABB-tree for the reprojection step in isotropic remeshing
* handling case when target edge length is 0 in isotropic remeshing
* bug fix in undocumented header for removal of degenerate faces

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -198,8 +198,8 @@ namespace Polygon_mesh_processing {
     }
 
     //face index map given as a named parameter, or as an internal property map
-    FIMap fim = choose_param(get_param(np, CGAL::face_index),
-                             get_const_property_map(CGAL::face_index, pmesh));
+    FIMap fim = boost::choose_param(get_param(np, CGAL::face_index),
+                                    get_const_property_map(CGAL::face_index, pmesh));
 
     return internal::border_halfedges_impl(faces, fim, out, pmesh, np);
   }
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
@@ -57,7 +57,8 @@ namespace Polygon_mesh_processing {
 *
 * @param pmesh a polygon mesh with triangulated surface patches to be remeshed
 * @param faces the range of triangular faces defining one or several surface patches to be remeshed
-* @param target_edge_length the edge length that is targetted in the remeshed patch
+* @param target_edge_length the edge length that is targetted in the remeshed patch.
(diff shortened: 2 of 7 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -230,10 +231,12 @@ void isotropic_remeshing(const FaceRange& faces
 #ifdef CGAL_PMP_REMESHING_VERBOSE
     std::cout << " * Iteration " << (i + 1) << " *" << std::endl;
 #endif
-
-    remesher.split_long_edges(high);
-    remesher.collapse_short_edges(low, high);
-    remesher.equalize_valences();
+    if (target_edge_length!=0)
```
janetournois: what do you think of changing that with `if(target_edge_length > 0)` ?
Or maybe leave this, and add an assertion `CGAL_assertion(target_edge_length >= 0)` instead

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -230,10 +231,12 @@ void isotropic_remeshing(const FaceRange& faces
...
+    {
+      remesher.split_long_edges(high);
+      remesher.collapse_short_edges(low, high);
+      remesher.equalize_valences();
```
janetournois: Since `equalize_valences()` only considers the connectivity and not the edge lengths, we should maybe leave it outside the `if`. Did you experiment with and without it? What do you think?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -198,8 +198,8 @@ namespace Polygon_mesh_processing {
     }
 
     //face index map given as a named parameter, or as an internal property map
-    FIMap fim = choose_param(get_param(np, CGAL::face_index),
-                             get_const_property_map(CGAL::face_index, pmesh));
+    FIMap fim = boost::choose_param(get_param(np, CGAL::face_index),
```
lrineau: As far as I understand, here we use non-documented function templates overloads from `<boost/graph/named_function_params.hpp>`.

1. Is not that dangerous? What is the plan is Boost internals are changed in an incompatible way?
2. Why do you need to qualify `choose_param`? It seems the argument-dependent lookup (ADL) should find it.
3. Why do you need to qualify `choose_param`, but not `get_param`?

sloriot: 1- yes but we should not have used it in the first place and have our own functions (it is not related to this bug-fix)
2- no element from the boost namespace is a parameter, ADL cannot work
3- choose_param is overloaded in the CGAL namespace IIRC

lrineau:
Travis shows a compilation error in the Polyhedron demo!

That is a blocker for 4.10.

PR #2138, opened 2017-05-24
Point set processing: LAS I/O

sgiraudot:
## Summary of Changes

Input/output functions for LAS format (+ updated functions for PLY).

## Release Management

* Affected package(s): `Point_set_processing_3`
* Small Feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Read_and_write_LAS_point_sets


sgiraudot:
Bugs should be fixed.

sgiraudot:
Note (cc @lrineau): we should install LASlib on some of the platforms of the testsuite so that this new feature is tested properly (we now have [a dedicated fork](https://github.com/CGAL/LAStools) that should be easy to install).

lrineau:
@sgiraudot @maxGimeno You should work together and propose a pull-request to some docker images at https://github.com/CGAL/cgal-testsuite-dockerfiles.

afabri:
And you (Simon&Maxime) should also look with @janetournois for Windows.

Installation/cmake/modules/UseLASLIB.cmake
```diff
@@ -1,4 +1,4 @@
 # This module setups the compiler for using LASlib library.
 # It assumes that find_package(LASLIB) was already called.
 
-add_definitions(-DCGAL_LINKED_WITH_LASLIB)
+add_definitions(-DCGAL_LINKED_WITH_LASLIB -DUSE_AS_DLL)
```
lrineau: Where does that name `USE_AS_DLL` come from? It should have a kind of namespacing: `LASLIB_USED_AS_DLL`, or 
`CGAL_USE_LASLIB_AS_DLL`...

sgiraudot: It's from _LASlib_.

We forked it, so I could change it, but I'd rather change as little things as possible in the library. Maybe I can add `#define USE_AS_DLL` inside the CGAL code just before including the LASlib header and `#undef` right after (there is exactly one inclusion of LASlib). What do you think?

lrineau: If the CMake command was a `target_compile_definitions`, just for a specific target, that would be fine. But here the line `add_definitions(-DCGAL_LINKED_WITH_LASLIB -DUSE_AS_DLL)` will add `-DUSE_AS_DLL` to all compilation commands of the CMake project.

If that is not too complicated to hide that macro in just one header of CGAL, I would prefer.

janetournois:
Sosno should be ready to test this PR on visual studio 2015.

lrineau:
1. The Travis build failed, for a trivial reason (the license-checking macros).
2. More important: that PR does not compile on my machine. The error is in `FindLASLIB`.

Installation/cmake/modules/FindLASLIB.cmake
```diff
@@ -0,0 +1,35 @@
...
+find_library(LASLIB_LIBRARIES
+             NAMES las
+             PATHS ENV LD_LIBRARY_PATH
+                   ENV LIBRARY_PATH
+                   /usr/local/lib
+                   ${LASLIB_INCLUDE_DIR}/../../lib
+                  ENV LASLIB_LIB_DIR
+            )
+
+if(LASLIB_LIBRARIES)
```
lrineau: If `LASLIB_LIBRARIES` evaluates to `TRUE`, but not `LASZIP_INCLUDE_DIR`, there is a CMake error.

On my machine, there is a file `/usr/lib64/liblas.so`, from the package `liblas-devel-1.8.0-13.fc25.x86_64`, that corresponds to http://www.liblas.org, and `LASLIB_LIBRARIES` is automatically set to `/usr/lib64/liblas.so`, but `LASZIP_INCLUDE_DIR` remains `NOTFOUND`.

sgiraudot: The main problem here is that `libLAS != LASlib`, although both libraries use the same filename… so anyway, there's no way the code will compile, even if we add LASzip. I'm not sure at all how we can handle that.

lrineau: I think `LASLIB_FOUND` should not be set to true if `LASLIB_INCLUDE_DIR` is not-found. That would probably solve the issue I have had.

sgiraudot:
I added the test for `LASLIB_INCLUDE_DIR`.

lrineau:
That works correctly now on my machine: instead of a CMake critical error, I just have the message

> the LAS IO plugin needs LAS libraries and will not be compiled.

as expected.

lrineau:
It seems that PR change the API that was introduced in CGAL-4.8 by [Small_Features/Point_set_processing_ply_io][1] and modified in CGAL-4.9 by [Small_Features/Extend_PLY_point_set_reader][2]. What is the upgrade path for users of that documented API? If that is a simple breakable, then the `changes.html` should mention the breakage, with a bold **Breaking change:** in the text.

  [1]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Point_set_processing_ply_io
  [2]: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Extend_PLY_point_set_reader

sgiraudot:
Good point. I've updated `changes.html`.

Installation/changes.html
```diff
@@ -338,9 +338,11 @@ <h2 id="release4.10">Release 4.10 </h2>
...
     format), with or without taking additional properties into
     account.</li>
-    <li>The API of the PLY function to read points with properties is
-    modified for unification with LAS. A new function to write PLY
-    with properties is provided.</li>
+    <li><b>Breaking change:</b> The API of the PLY function to read
+    points with properties is modified for unification with LAS
+    (see <code>CGAL::read_ply_points_with_properties()</code>). A new
+    function to write PLY with properties is provided
+    (<code>CGAL::write_las_points_with_properties()</code>).</li>
```
lrineau: I think you have swapped the two names of functions <code>CGAL::read_ply_points_with_properties()</code>/<code>CGAL::write_las_points_with_properties()</code> in the description.

lrineau:
The Travis error will remain:
```
Point_set_processing_3/include/CGAL/IO/write_las_points.h
Point_set_processing_3/include/CGAL/IO/read_las_points.h
```
https://travis-ci.org/CGAL/cgal/jobs/238376800#L4070

The two files should include:
```C++
#include <CGAL/license/Point_set_processing_3.h>
```

sgiraudot:
Alright, I fixed both problems (error in `changes.html` + license check).

lrineau:
Doxygen error:
```
.../doc/Documentation/Installation.txt:576: Warning: Unexpected html tag <span> found within <a href=...> context
.../doc/Documentation/Installation.txt:576: Warning: Unexpected html tag </span> found within <a href=...> context
```
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-35/Documentation.log

sgiraudot:
Fixed. Apparently, Doxygen did not like the `\cgal` inside an HTML link `<a href>...</a>`.

lrineau:
@sgiraudot @janetournois There is a compilation error with LASLib on Windows:
```
C:\Users\Jane\Documents\CGAL\LAStools\LASzip\src\mydefs.hpp(41): fatal error C1017: invalid integer constant expression
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.11-Ic-35/Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz

Jane, do you reproduce it on SOSNO?

sgiraudot:
Many errors were coming from the use of C++11 features. After discussing with @sloriot, it seems more reasonable to make this feature C++11 only (as making it compatible with C++03 would be quite long and painful, and probably make the code less readable).

I added some precondition with error raising in the headers, conditions in the examples/tests/demo and a Doxygen macro `cgalRequiresCPP11` to document it correctly on each function.

Point_set_processing_3/include/CGAL/IO/read_ply_points.h
```diff
@@ -22,6 +22,9 @@
 
 #include <CGAL/license/Point_set_processing_3.h>
 
+#ifndef CGAL_CXX11
```
lrineau: You must include `<CGAL/config.h>` before any test of that macro `CGAL_CXX11`.

sgiraudot: Okay, I've done it.

lrineau:
This PR has a conflict with `master`. The conflict comes from commit d005981b529b6da5535ea2bdeb4b87eae075ce62 from PR #2139 merged last week.

sgiraudot:
Conflict fixed.

sgiraudot:
There are errors in `Point_set_3` (and therefore in `Polyhedron demo`) because of bad `#ifdef/#endif` macros. Sorry about that, it is fixed.

There's another error: apparently the test I added in the `CMakeLists` to make sure that platforms without C++11 would not try to compile the examples failed. I used the same as in the `CMakeLists` of Polyhedron demo:

```cmake
  list(FIND CMAKE_CXX_COMPILE_FEATURES cxx_generalized_initializers has_cpp11)
  if(has_cpp11 LESS 0)
    message(STATUS "NOTICE: LAS/PLY IO examples require a C++11 compiler and will not be compiled.")
  else()
    create_single_source_cgal_program( "read_ply_points_with_colors_example.cpp" )
    create_single_source_cgal_program( "write_ply_points_example.cpp" )
    
    find_package(LASLIB)
    if (LASLIB_FOUND)
      include(${LASLIB_USE_FILE})
      include_directories(${LASLIB_INCLUDE_DIR})
      include_directories(${LASZIP_INCLUDE_DIR})
      create_single_source_cgal_program( "read_las_example.cpp" )
      target_link_libraries(read_las_example ${LASLIB_LIBRARIES})
    else()
      message(STATUS "NOTICE : the LAS reader test requires LASlib and will not be compiled.")
    endif()
  endif()
```

It seems that platforms without C++11 still tried to compile and, of course, ran into my `#error` that I put in the beginning of the files that require it.

Did I do something wrong here?

janetournois:
I tried to reproduce the error of [CGAL-4.11-Ic-35](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.11-Ic-35/Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz) and ran into the following 

```
1>Point_set_processing_3\include\CGAL/IO/read_las_points.h(27): fatal error C1189: #error:  CGAL LAS reader requires a C++11 compiler
2>Point_set_processing_3\include\CGAL/IO/read_ply_points.h(27): fatal error C1189: #error:  CGAL PLY reader requires a C++11 compiler
3>Point_set_processing_3\include\CGAL/IO/write_ply_points.h(27): fatal error C1189: #error:  CGAL PLY writer requires a C++11 compiler
```

there is no vs 2015 testsuite today, but that's what I get from the examples of this package.

lrineau:
At the CMake level, the script detects the feature `cxx_generalized_initializers`: as soon as the CMake version is recent enough, and the compiler supports that C++ feature, that value is detected by CMake.

The definition of the macro `CGAL_CXX11` in `<CGAL/config.h>` does not work the same way. In particular, I think [Microsoft compilers still define `__cplusplus` to its old value `199711L`][so].

  [so]: https://stackoverflow.com/questions/14131454/visual-studio-2012-cplusplus-and-c-11

Do you know which C++11 features you need? We could try to use more specific macros.

sgiraudot:
If I'm not forgetting anything, I only need variadic templates and rvalue references (&&).

lrineau:
Then I think you should test the CGAL macros `CGAL_CFG_NO_CPP0X_RVALUE_REFERENCE` and `CGAL_CFG_NO_CPP0X_VARIADIC_TEMPLATES`, and test the same features in the CMake script.

lrineau:
What is more, the content of `CMAKE_CXX_COMPILE_FEATURES` is filled with the features that the compiler is able to have... **with the right flags**.

Take g++-5.1, for example, that implements, I think, all of C++11: `CMAKE_CXX_COMPILE_FEATURES` is filled with a lot of features, but the compiler will refuse to compile any C++11 code with `std=c++11` or an equivalent flag.

You must call the CMake command `target_compile_features`, with the features you want, on the two targets using them.

sgiraudot:
Ok, I changed my tests and checks, it should work better now.

lrineau:
Hi Simon, have you seen that the Travis tests still fail? https://travis-ci.org/CGAL/cgal/jobs/241973381#L4230

There is also something that will require a change in our automatic testing:
https://travis-ci.org/CGAL/cgal/jobs/241973325#L7546
The new C++11-only headers do not pass the automatic `-fsyntax-only` compilation. Either 
  - we change the headers to remove the `#error` and replace it with something else (a `CGAL_static_assertion_msg` in the body of the main functions),
  - or I whitelist those two headers.

sgiraudot:
Yes, I've seen the error and I manage to reproduce it using `clang` (although for some reason, my tests on `clang` also run into the `#error`, which they should not). It seems that it's not handling variadic template correctly.

For the syntax only compilation, I don't know: if the syntax has `C++11` features, won't it fail anyway?

lrineau:
https://github.com/CGAL/cgal/pull/2138#issuecomment-308045646:
> For the syntax only compilation, I don't know: if the syntax has C++11 features, won't it fail anyway?

Yes, you are right. I will whitelist the two headers...

lrineau:
The Travis job still shows a compilation error:
https://travis-ci.org/CGAL/cgal/jobs/242362528#L4230

sgiraudot:
Ok, let's see if `clang` works well with `std::tuple`…

sgiraudot:
Still some errors, most probably because I still used some `cpp11` code from CGAL to handle tuples (`make_tuple`, `get`) where I should just use `std`. I'll fix this.

sgiraudot:
Another bug, it seems I have to use `cpp11::tuple` outside of the function (examples) because otherwise `Nth_of_tuple_property_map` is not coherent. This is fixed.

Let's cross our fingers…

lrineau:
The Travis build failed for a simple reason:
```
[ 15%] Building CXX object CMakeFiles/read_ply_points_with_colors_example.dir/read_ply_points_with_colors_example.cpp.o
/home/travis/build/CGAL/cgal/Point_set_processing_3/examples/Point_set_processing_3/read_ply_points_with_colors_example.cpp:53:24: error: 
      no matching function for call to 'get'
      const Point& p = std::get<0>(points[i]);
                       ^~~~~~~~~~~
```
https://travis-ci.org/CGAL/cgal/jobs/243160178#L4238

sgiraudot:
There are errors in Travis, but this time it does not seem related to this PR (Point Set Processing is good). Errors with OpenGL, pretty weird…

lrineau:
There is a lot do to:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-55.shtml#Point_set_processing_3_Examples

- On a lot of platforms, there is `ERROR:    execution   of write_ply_points_example`.
- And also those errors:
```
Point_set_processing_3_Examples/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:   ERROR:    compilation of read_las_example
Point_set_processing_3_Examples/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:   ERROR:    not executed   read_las_example
Point_set_processing_3_Examples/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz:   ERROR:    compilation of read_las_example
Point_set_processing_3_Examples/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz:   ERROR:    not executed   read_las_example
Point_set_processing_3_Examples/TestReport_lrineau_ArchLinux-clang.gz:ERROR: cgal_test_with_cmake exited with error condition 
Point_set_processing_3_Examples/TestReport_lrineau_ArchLinux-clang.gz:   ERROR:    compilation of normal_estimation
Point_set_processing_3_Examples/TestReport_lrineau_ArchLinux-clang.gz:   ERROR:    not executed   normal_estimation
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-2016.gz:   ERROR:    compilation of read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-2016.gz:   ERROR:    not executed   read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-2016-Release.gz:   ERROR:    compilation of read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-2016-Release.gz:   ERROR:    not executed   read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC.gz:   ERROR:    compilation of read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC.gz:   ERROR:    not executed   read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-Release.gz:   ERROR:    compilation of read_ply_points_with_colors_example
Point_set_processing_3_Examples/TestReport_lrineau_CentOS7-ICC-Release.gz:   ERROR:    not executed   read_ply_points_with_colors_example
```

lrineau:
And a few compilation errors in `Point_set_3_Examples`:
  https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-55.shtml#Point_set_3_Examples

lrineau:
Plus two errors on Windows in the polyhedron demo:
```
Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:C:\Users\Jane\Documents\CGAL\LAStools\LASzip\src\mydefs.hpp(41): fatal error C1017: invalid integer constant expression
Polyhedron_Demo/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz:C:\Users\Jane\Documents\CGAL\LAStools\LASzip\src\mydefs.hpp(41): fatal error C1017: invalid integer constant expression
```
(Cc @janetournois for the expertise on Windows)

janetournois:
I have pushed one commit to CGAL/LAStools [4f2410a](https://github.com/CGAL/LAStools/commit/4f2410abf60dac6df3185fa4cab3f37dcc2bc569) but I locally have new compilation errors.
I will have a look

lrineau:
@sgiraudot: This PR also has a small merge conflict with `master`, in `Polyhedron/demo/Polyhedron/CMakeLists.txt`.

janetournois:
With this commit [LAStools/80610f](https://github.com/CGAL/LAStools/commit/80610f00089b3fd4c2ce99a92391ff9fcb055b5b) the demo compiles on my machine

sgiraudot:
I fixed the bug of the `write_ply_example`. Turns out `CGAL::write()` was segfaulting. I don't really understand why, it seems that it was called many many times (to many compared to the normal behavior) with always the same values of color...

Anyway, I also found out that this function is not documented, so it's better not to use it in an example. I'm not really a fan of the `reinterpret_cast` in the example either, but there does not seem to be a cleaner way of writing binary…

sgiraudot:
I think the example of `Point_set_3` should be fixed too.

Point_set_processing_3/examples/Point_set_processing_3/write_ply_points_example.cpp
```diff
@@ -32,7 +32,7 @@ namespace CGAL {
       if (is_ascii(out))
         out << int(c[0]) << " " << int(c[1]) << " " << int(c[2]) << " " << int(c[3]);
       else
-        CGAL::write(out, c);
+        out.write(reinterpret_cast<const char*>(&c), sizeof(c));
```
lrineau: That is really strange. Unless there is an overload of `CGAL::write` that we do not know, the two codes are equivalent.

lrineau: I suspect a bug in `sizeof(c)` where `c` is a reference to an array. If the bug persists, you can hardcode the value `4` instead of `sizeof(c)`.

sgiraudot: Actually it isn't: there's a tag `Io_tag()` that's deduce. If it is `io_Read_write()`, it does the `reinterpret_cast` but apparently, in that case, it is `io_Operator`: as a result, it calls `oformat()`… which calls this method, which calls `oformat()`, etc. Hence the bug.

I'm not sure how these tags are supposed to be handled.

lrineau: You are right. The default implementation of `CGAL::write` calls `oformat`! Good catch!

sgiraudot:
There are some errors on `Point_set_processing` and `Point_set_3`. I'm investigating…

sgiraudot:
It seems that our hack for ICC did not work…

sgiraudot:
I fixed one of the two errors (`IO.h` was not needed in one example). I'm still trying to solve the second (it's going through the static assertion whereas it shouldn't).

( @lrineau I inadvertantly erased your latest patch, if you still have it locally, can you push it again? Otherwise I think I remember what is was and I can do it again.)

sgiraudot:
The error polluted several packages.

One easy way to fix it would be to simply replace the static assertion by a runtime message on `std::cerr` (and always returning `false`, which is documented as a failure return). But it's less optimal, a compile-time error would be better…

lrineau:
I think I found [a StackOverlow answer][1] that explains why the static assertion triggered:
> That's because the condition does not depend in any way on the template parameters. Therefore, the compiler can evaluate it even before instantiating that template, and produces the associated compilation error message if it the evaluation yields false.

   [1]: https://stackoverflow.com/a/14637372/1728537

[That other answer][2] provides a workaround:
```c++
template<typename T>
struct foobar : std::false_type
{ };

template <typename T>
inline T getValue(AnObject&)
{
    static_assert( foobar<T>::value , "this function has to be implemented for desired type");
}
```

  [2]: https://stackoverflow.com/a/14637534/1728537

lrineau:
Note also that there is a conflict with master in `Installation/include/CGAL/config.h` and Travis no longer tests the new commits.

lrineau:
As for the missing commit, I have rebased my local branch on your new version, and I pushed the result. The commit is https://github.com/CGAL/cgal/pull/2138/commits/81fafa3356ffb1406e8964871c3e1c43824ccea5.

sgiraudot:
The static assertion should be fixed and the branch is rebased.

sgiraudot:
Everything looks good except for the remaining warning on Windows DLL. I deactivated it in the code: it comes from LASlib and I'd rather change our fork of LASlib as little as possible (and this warning seems to require many additions of the export DLL mechanism). Anyway, we already deactivated a warning from GCC for the inclusion of LASlib…

sgiraudot:
The remaining error on Windows should be fixed.

PR #2139, opened 2017-05-24
Point set 3: Bugfix with carriage return

sgiraudot:
## Summary of Changes

The `operator>>` of the class `Point_set_3` uses the first line read (tag) to estimate the format. I just realized that using a simple string comparison could lead to bugs if the file read was not generated from the same platform (the usual differences of carriage return between `\n`, `\r\n`, etc.).

Using `std::string::find()` and checking if the line starts with the tag is enough and solves the bug.

## Release Management

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3/IO.h
+++ b/Point_set_3/include/CGAL/Point_set_3/IO.h
@@ -201,10 +201,11 @@ std::istream& operator>>(std::istream& is,
   std::string line;
   if (!getline(is, line))
     return is;
+
   is.seekg(0);
-  if (line == "OFF" || line == "NOFF")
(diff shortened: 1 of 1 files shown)


lrineau:
You say that is a bug-fix. Should not that bug be fixed in release branches, for 4.9.2 or 4.10.1 (or maybe directly 4.10 if that is not too late)?

sgiraudot:
`Point_set_3` did not exist in 4.9 and was introduced in 4.10, so this fix should target 4.10 or 4.10.1 (not sure how this works).

lrineau:
Can you please rebase the branch onto `releases/CGAL-4.10-branch`? I will merge the PR in it.

That will probably be too late for 4.10. It will be in 4.10.1.

sgiraudot:
Okay, I rebased it.

PR #2147, opened 2017-06-01
Polyhedron demo: Fixes and enhancements

### LINKED ISSUE #2128: Polyhedron demo visualisation bug on Mac OS
## Issue Details

[data.zip](https://github.com/CGAL/cgal/files/1014167/data.zip)

When running the CGAL Polyhedron demo on my Mac OS, I have some visualisation issues. I attach three test point clouds (a torus, a cylinder, and a shuttle) and a set of screenshots. With the torus, the point visualisation mode does not work, I do not see any points (screen 1). The same for points+normals mode, I see normals, but I do not see points (screen 2). The shaded point mode works (screen 3).

If I then apply shape detection algorithm, then again I do not see any result (screen 4). Actually, after shape detection, the only mode, which works, is shaded points, but only if I disable the group (screen 5), together with the group none of the modes works (screen 6 for splats).

I did a similar example with a cylinder. In this case, I can see the points (screen 1), but when I use the points+normals mode, all normals are disattached from the points (screen 2).

For a shuttle example, I can see points, but they do not make any sense (screen 1 with points and screen 2 with points+normals).

I tested all these point clouds in MeshLab and they are visualised correctly there. In addition, I  did the torus example on Linux, and again everything was working as expected.

The same happens for other point clouds. So, it looks like always a problem with visualising points, but maybe I am doing something wrong. Please let me know.

**To reproduce the error (if it is a one), it should be the following configuration:**

MacBook Pro 2016, 2.9 GHz Intel Core i7, 16 GB 2133 MHz LPDDR3, Radeon Pro 460 4096 MB Intel HD Graphics 530 1536 MB, Mac OS Sierra.

**I compiled the latest release version of the CGAL public-dev with** 
* libQGLViewer 2.6.4
* Qt 5.8.0_2
* Eigen 3.3.3
* TBB 2017
* VTK 7.1.1_1 
* MFR 3.1.5 
* Boost 1.64.0_1 

 The compiler is Apple LLVM 8.1.0 (clang-802.0.42), target x86_64-apple-darwin16.5.0

maxGimeno:
## Summary of Changes

 - Fix point set visualization on mac
 - Fix Point_set_shape_detection_plugin
 - Enhance the display of the axis system
 - Workaround for focus problem on mac
## Release Management

* Issue(s) solved (if any): fix #2128

### DIFF
--- a/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
+++ b/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
@@ -68,7 +68,7 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
   // On Apple, the first time the application is launched, the menus are unclicable, and
   // the only way you can fix it is to unfocus and re-focus the application.
   // This is a hack that makes the application lose the focus after it is started, to force the user
-  // to re-focus it.
+  // to re-focus it. (source : http://www.alecjacobson.com/weblog/?p=3910)
 #ifdef __APPLE__
     system("osascript -e 'tell application \"System Events\" "
       "to keystroke tab using {command down, shift down}'");
--- a/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
+++ b/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
@@ -64,7 +64,10 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
 
   mainWindow.setWindowTitle(main_window_title);
   mainWindow.show();
-
+#ifdef __APPLE__
+    system("osascript -e 'tell application \"System Events\" "
+      "to keystroke tab using {command down, shift down}'");
+#endif
   if(parser.isSet(use_meta)) {
     mainWindow.setAddKeyFrameKeyboardModifiers(::Qt::MetaModifier);
   }
--- a/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
+++ b/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
@@ -64,6 +64,11 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
 
   mainWindow.setWindowTitle(main_window_title);
   mainWindow.show();
+
+  // On Apple, the first time the application is launched, the menus are unclicable, and
+  // the only way you can fix it is to unfocus and re-focus the application.
+  // This is a hack that makes the application lose the focus after it is started, to force the user
+  // to re-focus it.
 #ifdef __APPLE__
     system("osascript -e 'tell application \"System Events\" "
       "to keystroke tab using {command down, shift down}'");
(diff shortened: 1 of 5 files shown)


Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
```diff
@@ -64,7 +64,10 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
 
   mainWindow.setWindowTitle(main_window_title);
   mainWindow.show();
-
+#ifdef __APPLE__
+    system("osascript -e 'tell application \"System Events\" "
+      "to keystroke tab using {command down, shift down}'");
+#endif
```
lrineau: You cannot include that ugly hack in the code without a strong comments/documentation explaining why that is needed!

Polyhedron/demo/Polyhedron/Scene_group_item.cpp
```diff
@@ -141,70 +141,25 @@ void Scene_group_item::draw(CGAL::Three::Viewer_interface* viewer) const {
...
-    }
-    switch(child->renderingMode()) {
-    case Points:
-    case PointsPlusNormals:
-      child->drawPoints(viewer); break;
-    default: break;
-    }
-  }
-  already_drawn = true;
+  draw(viewer);
```
lrineau: I strongly disagree! If `drawEdges` and `draw` call the exact same code, and in the `FlatPlusEdges` mode, the drawing will be done twice, and in the `Writeframe` mode, the drawing will be wrong.

There may be a bug in the current logic, but the solution cannot be what you proposed.

lrineau:
If the bugs are important, I think that PR should be against CGAL-4.10, so that users will have access to the patch once CGAL-4.10.1 is published, will not have to wait for CGAL-4.11.

lrineau:
@maxGimeno @danston: 

You did not answered my messgae from yesterday (https://github.com/CGAL/cgal/pull/2147#issuecomment-305447978):

Should this PR be merged in CGAL-4.10.1 as well?

danston:
Hey, sorry! I did not receive a notification about the message. I think, it should be merged as soon as possible to become public for users. Some bugs are not very important like the focus one, but some other bugs like visualisation of points are very basic functionality and without them the demo is not very usable on Mac OS, at least for point sets.

danston:
I tested the current fixed version and it works pretty good. For the moment I did not find any severe artefacts.

lrineau:
It seems Maxime already rebased on `releases/CGAL-4.10-branch` and did not tell us. Good!

danston:
Ok, great.

Polyhedron/demo/Polyhedron/Scene_group_item.cpp
```diff
@@ -141,10 +141,13 @@ void Scene_group_item::draw(CGAL::Three::Viewer_interface* viewer) const {
     }
     switch(child->renderingMode()) {
     case Points:
+    case ShadedPoints:
     case PointsPlusNormals:
       child->drawPoints(viewer); break;
     default: break;
     }
+    if(child->renderingMode() == Splatting)
+      child->drawSplats(viewer); break;
```
lrineau: I am surprised there was no warning from gcc.

PR #2148, opened 2017-06-01
Increase timeout for polyhedron_demo in testsuite

maxGimeno:
## Summary of Changes

This should allow picasso to build the Polyhedron demo entirely. This is the first step for automatically generate a pre-compiled demo for windows.

### DIFF
--- a/Testsuite/test/run_testsuite_with_cmake
+++ b/Testsuite/test/run_testsuite_with_cmake
@@ -154,11 +154,11 @@ test_directory()
       START=`date +%s`
       run_local_cgal_test &
 
-      TIMEOUT=1200
+      TIME_PERIOD=1200
       if [ "$1" = "Polyhedron_Demo" ]; then
-        TIMEOUT=5400
+        TIME_PERIOD=2400
       fi
-      if wait_for_process "$!" "$TIMEOUT" "5"
+      if wait_for_process "$!" "$TIME_PERIOD" "5"
       then
 	if [ -f test_failure  ] ; then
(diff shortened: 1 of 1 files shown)


Testsuite/test/run_testsuite_with_cmake
```diff
@@ -154,7 +154,11 @@ test_directory()
       START=`date +%s`
       run_local_cgal_test &
 
-      if wait_for_process "$!" "1200" "5"
+      TIMEOUT=1200
+      if [ "$1" = "Polyhedron_Demo" ]; then
+        TIMEOUT=5400
+      fi
+      if wait_for_process "$!" "$TIMEOUT" "5"
```
lrineau: If you have a look at the comments documenting `wait_for_process`, [above in the file][l], you can see that with `TIMEOUT=1200` the function `wait_for_process` waits 1200 times for 5 seconds (and that means 100 minutes). With  
`TIMEOUT=5400`, that would mean 450 minutes (7 hours and a half)! That seems too much.

  [l]: https://github.com/maxGimeno/cgal/blob/b7225bbf078ef2da1c79c228bba3ac855203a5c3/Testsuite/test/run_testsuite_with_cmake#L46

100 minutes is already a lot! How long need Visual Studio to compile the whole demo?! I think we should try first with just the double: `TIMEOUT=2400`

lrineau: By the way, the name `TIMEOUT` is not right, because that is the timeout divided by 5.

PR #2153, opened 2017-06-06
PMP - add random perturbation for triangle meshes

janetournois:
## Summary of Changes

This PR introduces a simple algorithm that moves all vertices of a triangle mesh in a random manner. Moves are constrained by 
- a maximum length for each move,
- a property map `vertex_is_constrained_map` of vertices that are not moving,
- vertices keep interpolating the input surface, because they are always reprojected to the input surface,
- boundary vertices do not move.

Note that combinatorics do not change, and this can introduce wrapping and self-intersections (on purpose).

A plugin and an example are part of this PR.

_edit :_ link to the [small feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_PMP::random_perturbation()#Status)

## Release Management

* Affected package(s): PMP, Polyhedron demo.
* Feature/Small Feature (if any): The documentation is ready. Shall I expose it? I'm not sure it is of big interest, but I needed this algorithm, so maybe I'm not the only one...
@sloriot @lrineau 

@kkatrio You can/should use this plugin to generate some input data for your smoothing algorithms. Depending on the `perturbation_max_size` you use, you can make valid or non-valid meshes, and it works for planar or non-planar input meshes.


lrineau:
I think that plugin should remain a debugging tool in the demo, like the plugin that shuffle the orientation of triangles in a surface mesh.

kkatrio:
very helpful, thanks!

sloriot:
You should add a range of vertices as input and use named parameters. I would make the reprojection step optional. I see no reason not to document it.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -0,0 +1,172 @@
...
+* @tparam NamedParameters a sequence of \ref namedparameters
+*
+* @param tmesh the triangulated surface mesh to be perturbed
+* @param perturbation_max_size the maximal length of moves that can be applied to
+*        vertices of `tmesh`.
+* @param np optional sequence of \ref namedparameters among the ones listed below
+*
+* \cgalNamedParamsBegin
+*  \cgalParamBegin{geom_traits} a geometric traits class instance, model of `Kernel`.
+*    Exact constructions kernels are not supported by this function.
```
sloriot: I think in "Exact constructions kernels" exact constructions is used as an adjective and it should be singular. A better formulation would be "Kernels with exact constructions ..."

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -0,0 +1,172 @@
...
+      internal::No_constraint_pmap<vertex_descriptor>//default
+    > ::type VCMap;
+  VCMap vcmap = choose_param(get_param(np, internal_np::vertex_is_constrained),
+                             internal::No_constraint_pmap<vertex_descriptor>());
+
+  internal::random_perturbation_impl<GT>(tmesh,
+                                         perturbation_max_size,
+                                         vcmap,
+                                         vpmap,
+                                         0/*seed*/);
```
sloriot: Adding a name parameter is now very easy with the new set of macros. I would definitely expose the seed using a named parameter.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: Actually, there is no such restriction.
Commit ec6eb417775f9fa8d0eca3007a7fdb2a7dbc9683 fixes that

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: with commit a874473106dba839e603a85cad6dfd8b9c9161e4, the seed becomes a named parameter

janetournois:
@sloriot why would like to add a range of vertices as input? The `vertex_is_constrained_map` is there to constrain non-moving vertices, so the user should be able to do what you want, right?

janetournois:
I think this PR is ready to be tested.
The only remaining thing is the complete documentation, if we decide to document it

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -0,0 +1,204 @@
...
+               FT(rng.get_double(-max_size, max_size)));
+  }
+
+  template<typename GT, typename PM, typename VCMap, typename VPMap, typename RNG>
+  void random_perturbation_impl(PM& tmesh,
+                                const double& max_size,
+                                VCMap vcmap,
+                                VPMap vpmap,
+                                bool do_project,
+                                RNG& rng,
```
kkatrio: may I propose RNG&& rng  ?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -0,0 +1,204 @@
...
+
+namespace CGAL {
+
+namespace Polygon_mesh_processing {
+
+namespace internal {
+
+  template<typename GT, typename RNG>
+  typename GT::Vector_3 construct_random_vector_3(const double& max_size,
+                                                  RNG& rng,
```
kkatrio: same here RNG&& rng

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -112,6 +110,8 @@ namespace internal {
 *    constrained-or-not status of each vertex of `tmesh`. A constrained vertex
 *    cannot be modified at all during remeshing
 *  \cgalParamEnd
+*  \cgalParambegin{random_seed} an non-negative integer value to seed the random
```
sloriot: a non-negative...

sloriot:
The new named parameter is not documented in `Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt`

sloriot:
About the input vertex range: if you think about the usage, you don't want to constrained all the vertices but a small patch. Providing the small patch of vertices is cheaper and easier. Is there any API issue?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: why do you think this is better?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: "non-negative" means `>=0`. What would you say othewise? "Positive" means `>0`...

janetournois: right, thanks!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
kkatrio: CGAL::Random() which is passed to random_perturbation_impl is (I think) a rvalue, so can't be referenced, unless considered as a rvalue-reference with the && from C++11. If it must be able to work without C++11, then it's not any better I guess...

lrineau:
Travis build error: https://travis-ci.org/CGAL/cgal/jobs/241962381#L5650

Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt
```diff
@@ -319,6 +319,23 @@ Monte-Carlo methods.
...
+\b Default value is `true`
+\cgalNPEnd
+
+\cgalNPBegin{random_seed} \anchor PMP_random_seed
+Parameter used in `random_perturbation()` to choose a seed for the random number generator
+`CGAL::Random()`. If this parameter is not used, perturbation is not deterministic
+(i.e. not reproducible from one run to the other).
+\n
+\b Type : `unsigned int` \n
+\b Default is that this variable is unset
```
sloriot: I think unset and non set is not the same thing. Anyway, I guess it is initialized to some seed.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
```diff
@@ -0,0 +1,233 @@
...
+*
+* @tparam VertexRange model of `Range`, holding
+*         vertices of type `boost::graph_traits<TriangleMesh>::vertex_descriptor`.
+*         Its iterator type is `ForwardIterator`. 
+* @tparam TriangleMesh model of `MutableFaceGraph`.
+* @tparam NamedParameters a sequence of \ref namedparameters
+*
+* @param vertices the range of vertices to be perturbed
+* @param tmesh the triangulated surface mesh to which the perturbed vertices belong
+* @param perturbation_max_size the maximal length of moves that can be applied to
```
sloriot: I guess it does not take into account the reprojection, right? If so we should mention it.

Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt
janetournois: do you mean that "unset" is an action, while "not set" is passive?
If the parameter is not given, the constructor `CGAL::Random()` is used, instead of `CGAL::Random(seed)`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: No, it does not take reprojection into account. But the re-projected point should be closer than the perturbed-with-no-projection point. Provided the point does not migrate to another surface patch (which should be checked in a later version, maybe), the re-projected point is closer, so this parameter remains the "maximal length".

lrineau:
Before the PR can be tested in `integration`:
1. This PR does not build in Travis: https://travis-ci.org/CGAL/cgal/jobs/243148759#L5655
2. The remark https://github.com/CGAL/cgal/pull/2153#pullrequestreview-43324621 seems a valid one, that should be addressed.
3. There is a conflict with `master`.

janetournois:
I have rebased the branch. Conflicts should be solved

janetournois:
the Travis build error is "expects an l-value for 7th argument",
so it's the same issue as remark https://github.com/CGAL/cgal/pull/2153#pullrequestreview-43324621
I'm having a look...

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/random_perturbation.h
janetournois: `RNG&&` is what we would like, indeed (note that `clang++` even made a compilation error in the Travis tests, saying that this function required an l-value), but for now CGAL must be able to work without C++11.

So I'm changing the initialization of `rng` with commit 55157013aa7ccce34402c1870e89c9e8bf5cc8ef

thank you!

lrineau:
This PR does not have a Travis job. I do not know why. I will close and reopen it, to see if Travis reacts.

lrineau:
That worked: a Travis job is now open.

lrineau:
I re-schedule the PR for 4.12, because:
* There are conflicts with `master`,
* the PR is more useful for developers than for CGAL users,
* Jane is not there.

No need to hurry and integrate it this week.

sloriot:
It was trivial. I've fixed it.

PR #2157, opened 2017-06-07
Try clang-3.6 with Travis

lrineau:
Update Travis to use clang-3.6 instead of the default clang-3.5


lrineau:
@maxGimeno The Travis job went well, but the clang version that was used it still clang-3.5. Can you have a look?

maxGimeno:
And don't forget to edit .travis/template.txt

lrineau:
Close. Replaced by #2184.

PR #2161, opened 2017-06-08
Surface_mesher IO.h: Add missing headers

lrineau:
The new compiler g++-7 has change the C++ headers dependencies. `std::algorithm` is no longer found.

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-35.shtml#Polyhedron_Demo


lrineau:
@maxGimeno: That is the reason why `std::algorithm` is not found: `<algorithm>` was not included.

PR #2164, opened 2017-06-08
Polyhedron_demo: Enhance the 'Save as' dialog

### LINKED ISSUE #2162: Polyhedron demo : add a default file name for "Save as..."
## Issue Details

When "Save as..." is called in the demo, it would be nice that the file name is automatically filled with the item's name, as a default.
It would avoid typing a file name that is most-probably the same as the item's name (save time and avoid typos!)

maxGimeno:
## Summary of Changes

Add an automatically generated name to the 'Save as...' dialog.

## Release Management

* Issue(s) solved (if any): fix #2162

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1534,10 +1534,14 @@ void MainWindow::on_actionSaveAs_triggered()
     return;
   }
   QString caption = tr("Save %1 to File...%2").arg(item->name()).arg(ext);
+  //remove `)`
+  ext.chop(1);
+  //remove `(*.`
+  ext = ext.right(ext.size()-3);
   QString filename = 
     QFileDialog::getSaveFileName(this,
                                  caption,
-                                 QString(),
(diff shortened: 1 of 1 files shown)


sloriot:
From what I understand, you are taking the name of the item as suggestion. I think it might make sense to first look whether the item has an associated input filename.

maxGimeno:
@sloriot Why ? If you load an item, and then edit it, you would rather have it named like the original than like the edited version ?

janetournois:
why do you think it's better to look for the file name?
Here is the workflow I was thinking of
- open an the file `poly.off`
- The newly created item is named `poly`
- apply your favorite algo, that creates a new item, named 'poly_after_algo'
- I'd rather save it as "poly_after_algo.off", not "poly.off"

don't you think?

janetournois:
oops sorry @maxGimeno I did not see your answer - I agree

sloriot:
well, I'm usually using the demo to fix a off file. But the name of the item being the name of the file it should work too.

janetournois:
I tested it on visual 2013, it works fine, and it's very helpful!
Thank you!

PR #2167, opened 2017-06-09
Polyhedron_demo: Point set to mesh distance plugin

### LINKED ISSUE #2163: Polyhedron demo: Analyse distance between point set and mesh
_(This is a feature request, not urgent.)_

## Issue Details

It would be nice to have the following feature on the demo:
1. Select a point set item and a mesh item
2. Compute the distance from each point to the mesh
3. Output the statistics (min distance, first decile, median, last decile, max)
4. With a ramp, color each point from the closest to the farthest from the surface (to easily identify which points are well represented by the mesh and which points are not)

That could be very useful to analyse the output of reconstruction algorithms.

@maxGimeno if/when you take care of this issue, I can give you the code for the distance computation + statistical analysis, based on the AABB tree. It's pretty similar to Hausdorff anyway (but instead of taking the max, you just save all the distance values in a vector that you sort afterwards).

## Tip

You can store colors inside the point set item (adding property maps `r`, `g` and `b` with `unsigned char` type), so the display is already handled. But this could be a problem if the point set _already_ has colors (I haven't thought it over).

maxGimeno:
## Summary of Changes

This is a plugin that allows to colorize a point set according to the distance of each point to a selected mesh (polyhedron or surface_mesh.
## Release Management

* Issue(s) solved (if any): fix #2163

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Distance_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Distance_plugin.cpp
@@ -294,6 +294,8 @@ private:
       double hausdorff = compute_distances(*poly_B,
                                            sorted_points,
                                            distances);
+      if(hausdorff == 0)
+        hausdorff++;
       //compute the colors
       colors.resize(sorted_points.size()*3);
       for(std::size_t i=0; i<sorted_points.size(); ++i)
--- a/Polyhedron/demo/Polyhedron/Plugins/Point_set/CMakeLists.txt
+++ b/Polyhedron/demo/Polyhedron/Plugins/Point_set/CMakeLists.txt
@@ -65,3 +65,7 @@ endif()
   qt5_wrap_ui( alpha_shapeUI_FILES Alpha_shape_widget.ui )
   polyhedron_demo_plugin(alpha_shape_plugin Alpha_shape_plugin ${alpha_shapeUI_FILES})
   target_link_libraries(alpha_shape_plugin scene_points_with_normal_item scene_c3t3_item)
+
+  qt5_wrap_ui( distanceUI_FILES Point_set_to_mesh_distance_widget.ui )
+  polyhedron_demo_plugin(point_set_to_mesh_distance_plugin Point_set_to_mesh_distance_plugin ${distanceUI_FILES})
+  target_link_libraries(point_set_to_mesh_distance_plugin scene_points_with_normal_item scene_surface_mesh_item scene_polyhedron_item scene_color_ramp)
--- a/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_to_mesh_distance_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_to_mesh_distance_plugin.cpp
@@ -1,0 +1,290 @@
+#include <CGAL/Three/Polyhedron_demo_plugin_interface.h>
+#include <CGAL/Three/Polyhedron_demo_plugin_helper.h>
+#include <QApplication>
+#include <QDockWidget>
+#include <QObject>
+#include <QAction>
+#include <QMainWindow>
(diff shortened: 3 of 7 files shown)


maxGimeno:
@sgiraudot don't hesitate to check what I did for the stats.

sloriot:
Just a remark on what @sgiraudot said. Maybe we can avoid duplicating the point set. Since we have the notion of groups, maybe it is not hard to have additional properties for a point set accessible in the GUI. That way we could have several color properties without duplicating the point set.

sgiraudot:
In that case, it would be nice to have something fully extensible (not sure if possible): when reading PLY, we can store several properties on the point set. Let's say we have colors + two properties with type `double` (for example, _intensity_ and _error_). Then there could be a list of available displays:
* colors
* intensity
* error

Colors would be displayed as usual, but additional properties (that we typically don't know at compile time) would simply be displayed using ramps (recomputed each time we switch the display). This is pretty much what I'm doing by hand for the classification feature (except that I store the recomputed ramps in the RGB properties for convenience).

For this particular plugin, we wouldn't store the colors of the ramp corresponding to the distance, simply the distances themselves. Which would be also nice, because if the user saves the item in PLY format, it will simply have an additional `property double distance` which makes more sense than the colors.

But then again, that might be a much more complicated feature than what I asked on my issue about this distance plugin…

lrineau:
Maybe we should plan to change completely the UI of a `Scene_item`. Currently, the sole properties of an item that can be changed easily are: its name, its "color", and the drawing mode (wireframe, flat...). Maybe there should be just a clickable button that would open a "properties dialog".

sloriot:
Sorry for messing things up here. An issue should be opened and it should reference this PR. In the mean time, the solution @sgiraudot proposed is the simplest.

sgiraudot:
Okay, it looks good now, I don't have any other remark to do :)

sgiraudot:
(I fixed the second bug, it was happening if the property maps for colors were `r` `g` and `b` instead of `red` `green` and `blue`. I replaced by something safer: remove all existing color maps and add the ones you need.)

sgiraudot:
It seems that you removed my fix with your last commit…

maxGimeno:
No, I don't think so. Wasn't it af142a3 and ca25b21 ?

maxGimeno:
Indeed, I guess I forgot to reset the branch. It is fixed now.

sloriot:
If it is not hard to do, having a way to selection points that are further from a given distance from a surface could be interesting.

afabri:
it works with VC++ for Polyhedron and Surface_mesh.

lrineau:
This PR is currently selected for CGAL-4.12 (not 4.11). Do we still agree on that?

afabri:
ok for me. As it is a plugin I just tested it interactively.

Polyhedron/demo/Polyhedron/Plugins/PMP/Distance_plugin.cpp
```diff
@@ -294,6 +294,8 @@ class Scene_distance_polyhedron_item: public Scene_item
       double hausdorff = compute_distances(*poly_B,
                                            sorted_points,
                                            distances);
+      if(hausdorff == 0)
+        hausdorff++;
```
sloriot: Why do you need this?

maxGimeno: To insure there will not ba a division by 0.

sloriot: Could you be more precise? What is surprizing is that you have no way to distinguish whether the distance is 1 or 0 after the increment. Why is it not important?

maxGimeno: This distance is only here to scale the color ramp.

PR #2175, opened 2017-06-14
Bug fix in CMap and GMap copy functions.

gdamiand:

## Summary of Changes

Remove return since copy is a procedure (does not return anything).

(why this code compile ? strange...)

## Release Management

* Affected package(s): CMap and GMap


lrineau:
That compiles because you can return a `void` object from a function returning `void`.

See that SO answer: https://stackoverflow.com/a/2249281/1728537
> Interestingly, you can also return void from a void function. For example:

    void foo()
    {
      return void();
    }
> As expected, this is the same as a plain return;. It may seem esoteric, but the reason is for template consistency:

    template<class T>
    T default_value()
    {
      return T();
    }

gdamiand:
Thanks @lrineau for the explanations.
I didn't know that.
Anyway in this case (CMap and GMap), there is no interest of keeping the return.

PR #2176, opened 2017-06-15
Polyhedron_demo: Fixes for the override cursor.

maxGimeno:
## Summary of Changes
- Removes the save/load dialog displayed after validation of the path so it doesn't mess with the cursor.
- Restores the cursor before the `No self intersection` dialog appears
## Release Management


lrineau:
For those cursor overrides, we should create a guard class: its constructor would override the cursor, and its destructor would restore it. That way, we would be sure that the cursor would be restored when the function returns.

maxGimeno:
That's a great idea ! 
But in this context it wouldn't have helped at all. One bug was that the ghost of a file dialog hid the waitcursor, and the other bug was a modal messagebox that blocked the function before the restoration happened. After clicking`OK` the restoration was performed normally.

PR #2183, opened 2017-06-16
Post processing for doc testsuite

maxGimeno:
## Summary of Changes
Fix a path and add post processing in the doc testsuite

## Release Management

### DIFF
--- a/.travis/build_package.sh
+++ b/.travis/build_package.sh
@@ -19,7 +19,7 @@ function build_demo {
   cd build-travis
   if [ $NEED_3D = 1 ]; then
     #install libqglviewer
-    git clone --depth=1 https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
+    git clone --depth=4 -b v2.6.3 --single-branch https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
     pushd ./qglviewer/QGLViewer
     #use qt5 instead of qt4
     export QT_SELECT=5
--- a/Documentation/doc/scripts/test_doxygen_versions.sh
+++ b/Documentation/doc/scripts/test_doxygen_versions.sh
@@ -38,7 +38,6 @@ mkdir ./build_doc
 cd ./build_doc
 cmake -DCGAL_GENERATE_XML=ON -DDOXYGEN_EXECUTABLE="$PATH_TO_1" ../..  &> /dev/null
 make -j$NB_CORES doc  &> /dev/null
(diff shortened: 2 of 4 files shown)


lrineau:
@sloriot: Do you think the target `doc_and_postprocessing` can work with parallel make (`-j` option), now?

sloriot:
It will but the postprocessing will not be parallel.

sloriot:
I suspect the fix is not at the right place. I guess the runs with the doxygen output is not the one used when testing the compilation per package (since the output is redirected to /dev/null).

maxGimeno:
Don't know what you mean. The fix is in test_doxygen_versions.sh, which is what replaces the original call in run_doxygen_testsuite script.

sloriot:
I'm talking about the table [here](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-42) for example. Then maybe there are 3 runs one for the table, and one for each documentation.

maxGimeno:
If I understand (or remember)  correctly, the doc is built once for master, once for 1.8.4, they are compared, then the python script that runs the testsuite is called on the 1.8.4 build dir, and generates the table you are talking about. And if this is not the case, this table is generated by something I didn't touch, so there is no reason anything changed.

sloriot:
We'll see

lrineau:
We can still read 
> Acknowledging %CGAL

in the side-nav widget.

https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-46/output/Manual/index.html

That means that the post-processing is not done for the build with Doxygen-1.8.4.

On the other hand, the build with Doxygen-master had the post-processing: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-46/master/Manual/index.html

maxGimeno:
Ok the doc was build twice with 1.8.4, I forgot to call the post-processing on the second pass.

lrineau:
Thanks. That will be tested tonight, in CGAL-4.11-Ic-48.

... And for convenience, I give the (future) URLs to the results (to check if that worked):
- https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-48/output/Manual/index.html
- https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.11-Ic-48/master/Manual/index.html

PR #2102, opened 2017-05-10
Remove implicit conversions between points and weighted points

MaelRL:
## Summary of Changes
The current implementation of weighted points in the CGAL kernel allows for implicit conversions between points and weighted points, meaning that the compiler can convert a `Weighted_point_23` into a `Point_23` (and reversely) if it allows to match a function. This is dangerous because we can accidentally drop weights or call the wrong function unintentionally.

This small feature does not remove these conversions but makes them explicit, meaning that the constructor must be called explicitly to be used (and thus the compiler is not allowed to use them on its own). 

## Release Management

* Affected package(s): Main changes in kernel packages, incident changes  a bit everywhere (mostly `Alpha_shapes_23`, `Interpolation`, `Kinetic_data_structures`, `Mesh_3`, `Skin_surface_3`, and `Triangulation_23`). 
* Issue(s) solved (if any): fix #2144, fix #1878
* Feature/Small Feature (if any): [Remove Point -- Weighted point implicit conversions](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Remove_Point_--_Weighted_point_implicit_conversions)

PR #2185, opened 2017-06-16
Triangulation: add dD regular triangulation

cjamin:
* Affected package: Triangulation
* Feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Regular_triangulation

TODO:
- [x] Pass the testsuite


MaelRL:
The new `convex_hull.cpp` of `Triangulation` will conflict with `Envelope_2`'s `convex_hull.cpp`.

MaelRL:
More like review comments, but:
- Why not document the regular triangulation's `Bare_point` ?
- The concept `TriangulationVertex` has:
`typedef unspecified_type TriangulationVertex::Point`
and it adds:
`It must be the same as the point type TriangulationTraits::Point`
but none of traits concepts define a `Point` type. So this must be rephrased.

Actually, for the second point, I'd advise to restructure a bit and take the road of https://github.com/CGAL/cgal/pull/2102: instead of defining the (basic) triangulation point type from
`typedef typename Geom_traits::Point_d           Point;`
you can define it from
`typedef typename TDS::Vertex::Point               Point;`
The fact that `TDS::Vertex::Point` is `Point_d` or `Weighted_point_d` depending on your triangulation is already imposed right now. But with this definition, you can then completely remove all the adapter traits, provided you add a few `Construct_point_d` here and there. Your `RegularTriangulationTraits` anyway already requests `Bare_point Construct_point_d()(Weighted_point)` so adding a requirement `Bare_point Construct_point_d()(Bare_point)` is epsilon cost.

lrineau:
I have changed the title of the PR to something more meaningful, at least for me. Change it again if it is wrong.

lrineau:
There are errors with MSVC-2013, in point set processing, and more generally in anything that depends on `Spatial_searching/include/CGAL/Search_traits_adapter.h`:

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-43.shtml#Point_set_3_Examples

cjamin:
Regarding `convex_hull.cpp`, shouldn't the one from `Envelope_2` be called `convex_hull_2.cpp`?

lrineau:
The testsuite is also pretty red on Linux in the following two lines:
* https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-43.shtml#Triangulation
* https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-43.shtml#Triangulation_Examples

cjamin:
@lrineau Yes, I already started to fix most of the errors & warnings in my branch.
And I'm trying to reproduce the errors from MSVC 2013 (no errors with MSVC 2015).

cjamin:
The MSVC 2013 is now fixed, I will merge it again in integration tonight.
@MaelRL I also documented Bare_point, and will have a look at your other comment soon.

cjamin:
@MaelRL This is a big change. For example, I'm not sure how to manage the call to `spatial_sort(points.begin(), points.end(), geom_traits());`where `points` contains weighted points. For now, the adapter provides the correct `Less_coordinate_d`.
I can see that in your branch, you use:
`spatial_sort(points.begin(), points.end(), Search_traits_3(boost::make_function_property_map<Weighted_point, Ret, Construct_point_3>(geom_traits().construct_point_3_object()), geom_traits()));`
but I can't see any mention of this Search_traits_3 constructor in your documentation.
Did I miss something?

MaelRL:
The `Search_traits_3` is `CGAL::Spatial_sort_traits_adapter_3` ([doc](http://doc.cgal.org/latest/Spatial_sorting/classCGAL_1_1Spatial__sort__traits__adapter__3.html)), with the second constructor. Seems like the `_d` version exists with similar API.

Adapting this spatial search was one of the biggest changes in `T2`/`T3` in the other branch. I don't know `Td` very well, but in T2 and T3, there were a lot of functions wrapping the traits already, meaning that the only thing that was needed was adding some `construct_point_3` in those wrappers.

cjamin:
Thank you. Given the name, I had assumed it was `CGAL::Search_traits_3` and did not see the typedef which was right above.
You are right, the same thing exists in dD. I will have a look.

cjamin:
Well, I'm not as lucky as you were. And more than that, the problem is that we don't know the dimension, so everything goes through ranges or iterators (e.g. parsing the points of a cell). So it's not just adding a few `Construct_point_d`: we need to call boost::make_transform_iterator.
I'm not sure this will be costless in the DT case...

MaelRL:
Yes, I see. I don't know what the cost would be.

Maybe you can at least get rid of the overload of
 `typedef typename K::Weighted_point_d Point_d;`
by using `TDS::Vertex::Point`. Should be able to also get rid of `Bare_point_d` / `Weighted_point_d` typedefs from the traits class too.

Maybe it's not that important since the adapter overloads predicates that are relatively safe from a weighted-or-not point of view. It's still not ideal because you don't have `Tr::Gt == Vertex::Gt` (except if you use the default `TDS`), but oh well.

cjamin:
The default TDS is `Triangulation_data_structure<Maximal_dimension_, Triangulation_vertex<TriangulationTraits>, Triangulation_full_cell<TriangulationTraits> >`, which takes its point type from the traits. So I need to keep `typedef typename K::Weighted_point_d Point_d`.

In the traits, the type `Weighted_point_d` is used in the functors' parameters. So OK, it's not absolutely needed as a public type, but it's much more convenient like this.

In the end, I could get rid of `Bare_point_d` in the adapter and in `Regular_triangulation.h`.

cjamin:
@lrineau @MaelRL What about the `convex_hull.cpp` that also exists in `Envelope_2`? Will it actually create a conflict? I was suggesting to rename the one from `Envelope_2`to `convex_hull_2.cpp`. But I don't mine renaming mine either...

MaelRL:
I guess you mean the `Vertex_base` needs the point type. Sure, I could only use `TDS::Vertex::Point` in `T2`/`T3` because there existed a class `Regular_triangulation_vertex_base_3` that defines
`typedef Weighted_point_3 Point;` so you'd have to define it.

In any case, even if you leave things like they are right now, the concept `TriangulationVertex` must say `Traits::Point_d`.

lrineau:
@cjamin, https://github.com/CGAL/cgal/pull/2185#issuecomment-309715302
> What about the `convex_hull.cpp` that also exists in `Envelope_2`? Will it actually create a conflict? I was suggesting to rename the one from `Envelope_2` to `convex_hull_2.cpp`. But I don't mine renaming mine either...

That is a problem, in case one wants to configure all CGAL (with examples and tests) at once in the same project. The Travis builds do that check in the `PACKAGES=CHECK` lines but, as you know, Travis reports have false errors for the moment.

So one or the other must be renamed. If you prefer renaming the one from `Envelope_2`, make sure any reference to it (in the documentation for example) is updated as well.

cjamin:
Ok, I renamed the one from `Envelope_2`. Thanks.

lrineau:
Now the test results are a lot better. There is a compilation error on `CentOS5`, that cannot be fixed because the compiler is too old for the new kernel `Epick_d`, but the compilation errors on CentOS6 could be fixed, I think.

lrineau:
I had not seen that you had already modified the code.

lrineau:
This PR does not pass the line "CHECK" of Travis builds:
```
.. Checking include directives in header files...
Triangulation_2/include/CGAL/IO/Triangulation_off_ostream_2.h
Triangulation_3/include/CGAL/IO/Triangulation_off_ostream_3.h
Triangulation/include/CGAL/IO/Triangulation_off_ostream.h
Triangulation/include/CGAL/Regular_triangulation_traits_adapter.h
Triangulation/include/CGAL/Regular_triangulation.h
```
See https://travis-ci.org/CGAL/cgal/jobs/245365600#L4433

Those five new headers need a [license checking header][doc1].

  [doc1]: http://doc.cgal.org/latest/Manual/preliminaries.html#licenseCheck

@sloriot: Do we have a developers manual page that explains which license checking headers must be included?

@cjamin What I know is that all headers from `Triangulation/include/` do have:
```C++
#include <CGAL/license/Triangulation.h>
```
at the top of the file. Those five new headers should probably have it too. Maybe Sébastien will correct that statement.

sloriot:
See https://github.com/CGAL/cgal/wiki/License

cjamin:
Fixed. Thank you for the help.

lrineau:
I have pushed the commit https://github.com/CGAL/cgal/pull/2185/commits/2c6980bb4aadfa9cb54724a597426539012b4d37 because the script, run by Travis, was chocking on this branch, on filenames containing the pattern ` - ` (space-dash-space). The script was indeed not robust, and I have strengthened it a bit.

lrineau:
Now merged in `master`. Congratulation!

cjamin:
Thank you Laurent!

PR #2186, opened 2017-06-16
Fix qglviewer compilation in travis

maxGimeno:
## Summary of Changes

This PR Uses an older version of qglviewer as travis doesn't have a qt version recent enough to work with master.
## Release Management

* Affected package(s):


maxGimeno:
It's working but I need travis as soon as possible so I didn't let it do the whole tests.

PR #2194, opened 2017-06-20
Shape detection: bugfix on connected components

sgiraudot:
## Summary of Changes

When a shape is detected, the largest connected component is sought after. In the implementation, it seems that a cache mechanism was partially implemented: there is a boolean that tells if the largest connected component was already computed. But this component is not saved (only its cardinality is saved), which produces the following bug: if a shape is discarded once but then tried again (it can happen that a shape is not discarded at the second trial because of a variation of probability), the largest connected component is not computed. Therefore, _all the points_ are considered part of the largest connected component (which is usually wrong).

I removed this boolean and the associated test: the bug is gone and there's no noticeable impact on the performances.

## Release Management

* Affected package(s): _Point Set Shape Detection 3_

### DIFF
--- a/.travis/build_package.sh
+++ b/.travis/build_package.sh
@@ -19,7 +19,7 @@ function build_demo {
   cd build-travis
   if [ $NEED_3D = 1 ]; then
     #install libqglviewer
-    git clone --depth=1 https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
+    git clone --depth=4 -b v2.6.3 --single-branch https://github.com/GillesDebunne/libQGLViewer.git ./qglviewer
     pushd ./qglviewer/QGLViewer
     #use qt5 instead of qt4
     export QT_SELECT=5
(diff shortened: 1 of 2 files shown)


lrineau:
Usual question about bug-fixes: was the bug introduced only in `master`, after the release of CGAL-4.10, or is the bug already in CGAL-4.10 or even in CGAL-4.9.1?

sgiraudot:
It's been here since the package exists (4.7 I think).

lrineau:
Then I suggest to rebase the PR onto CGAL-4.9.1 or CGAL-4.10. Use for example:
```shell
    git fetch cgal
    git checkout Shape_detection-Bugfix_connected_components-GF
    git rebase --onto cgal/releases/CGAL-4.10-branch cgal/master
```
and then use `git push -f` cautiously.

sgiraudot:
Done.

sgiraudot:
Travis failed but it doesn't seem related to the PR (`apt-get install failed`).

lrineau:
I have restarted the failed sub-job of Travis. Hopefully it will success, this time.

PR #2208, opened 2017-06-26
Aabb tree doc update

sloriot:
update a sentence that we forgot to update

### DIFF
--- a/AABB_tree/include/CGAL/AABB_face_graph_triangle_primitive.h
+++ b/AABB_tree/include/CGAL/AABB_face_graph_triangle_primitive.h
@@ -47,10 +47,9 @@ namespace CGAL {
  *                         The default is `typename boost::property_map< FaceGraph,vertex_point_t>::%type`.
  *\tparam OneFaceGraphPerTree is either `CGAL::Tag_true` or `CGAL::Tag_false`.
  * In the former case, we guarantee that all the primitives will be from a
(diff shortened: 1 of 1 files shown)


AABB_tree/include/CGAL/AABB_polyhedron_triangle_primitive.h
```diff
@@ -154,8 +155,8 @@ namespace CGAL {
     };  // end class AABB_polyhedron_triangle_primitive
 
 
-
 }  // end namespace CGAL
+/// \endcond
```
lrineau: I disagree with that choice. For our users, it is a lot more useful to get a documentation of the class, with the deprecation notice, mentioning by what it is should replaced, instead of having the documentation missing.

sloriot: Except if I'm wrong, our rules says that after two releases, we can remove deprecated classes. Theses classes were deprecated since 4.3

lrineau: The rules says we **can** **remove** the classes... implicitly: if there is a benefit (for example, easier maintenance). The rules do not say we should remove the documentation for free. And if we remove classes (not the documentation), we usually announce it in the `changes.html`.

PR #2210, opened 2017-06-26
Polyhedron_demo: 'Save as' for Surface_mesh items

### LINKED ISSUE #2169: Polyhedron Demo: Add "Save As" for Surface_mesh_item

maxGimeno:
## Summary of Changes
- Adds a Save_as in OFF and OBJ format for the Surface_mesh items
- Adds a color writing for the OFF format. As it does not seem to be officially supported by Wavefront OBJ format, I didn't think it was a good idea to add it, I don't want to break any compatibility.

## Release Management

* Affected package(s):Surface_mesh, Polyhedron_demo
* Issue(s) solved (if any): fix #2169

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/IO/Surface_mesh_io_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/IO/Surface_mesh_io_plugin.cpp
@@ -35,11 +35,11 @@ public:
      return QList<QAction*>();
    }
    QString name() const { return "surface_mesh_io_plugin"; }
-   QString nameFilters() const { return "OFF files to Surface_mesh (*.off)"; }
+   QString nameFilters() const { return "OFF files to Surface_mesh (*.off);;Wavefront Surface_mesh OBJ (*.obj)"; }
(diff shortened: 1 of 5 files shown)


lrineau:
There are compilation errors due to this PR:
  https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-55.shtml#Polyhedron_Demo

maxGimeno:
I couldn't reproduce but from what I read that should fix the problem. By the way, I didn't  touch that function (read_mesh) and don't understand why or how that branch made it compiled when it wasn't before.

lrineau:
`std::tolower` is an overloaded function. Specifying a pointer to such a function requires an explicit static cast:
http://en.cppreference.com/w/cpp/language/overloaded_address

`::tolower` is a different function, the C version. It is not overloaded (C does not know overloads).

The reason why that errors occurred suddenly depends on which headers are included: if `<locale>` is not included, then the compiler does not see the overload, and does not complain about the ambiguity.

maxGimeno:
Ok, I guess one of my #include triggered it then.

PR #2213, opened 2017-06-27
Fix demos for qglviewer 2.7.0

maxGimeno:
## Summary of Changes
Rework for everything that uses qglviewer so it is not broken by the version 2.7.0.
## Release Management

* Affected package(s): All 3D Demos


lrineau:
There is a conflict with `master`.

maxGimeno:
Resolved

lrineau:
Maxime, do you think the patch of this branch could be applied easily to CGAL-4.10?

maxGimeno:
It mainly modifies demos we almost never touch, so I guess this should be easy.

lrineau:
I ask because the issue #2223, is reported by a user against CGAL-4.10. Could you prepare a PR for the 4.10 branch?

maxGimeno:
Ok

PR #2159, opened 2017-06-08
Scale space reconstruction: Improve API

sgiraudot:
## Summary of Changes

Separation of meshing and smoothing algorithm, addition of two new (better) algorithms. This is a pre-approved small feature.

## Release Management

* Affected package(s): Scale Space Reconstruction (and Polyhedron demo for the plugin)
* [Small Features/Improve API of Scale Space Reconstruction](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Improve_API_of_Scale_Space_Reconstruction)

PR #2226, opened 2017-07-04
Scale space reconstruction: improve API cleanup

sgiraudot:
## Summary of Changes

Fixes the two remaining issues from [the original PR](https://github.com/CGAL/cgal/pull/2159):
* cleans up commented pieces of code
* adds a section to `changes.html`

## Release Management

* Affected package(s): Scale Space Reconstruction

### DIFF
--- a/Installation/changes.html
+++ b/Installation/changes.html
@@ -1,4 +1,4 @@
-﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
 <!-- $Id$
      $URL$
 -->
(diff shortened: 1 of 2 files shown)


Installation/changes.html
```diff
@@ -1,4 +1,4 @@
-﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
```
sgiraudot: I don't get what's going on here… I did not change anything on this line.

lrineau: On my machine, this commit starts with:
```patch
diff --git a/Installation/changes.html b/Installation/changes.html
index 6fa29bf..a6cd54a 100644
--- a/Installation/changes.html
+++ b/Installation/changes.html
@@ -1,4 +1,4 @@
-<U+FEFF><!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
 <!-- $Id$
      $URL$
 -->
```

Your editor has removed the [BOM]. That is fine. It is optional anyway. However, your editor is not so polite, because it should not remove it when it is in the file.

  [BOM]: https://en.wikipedia.org/wiki/Byte_order_mark "Byte order mark"

sgiraudot: Weird, that never happened before…

lrineau: Well, that BOM does appear and disappear from time to time. It has already been added by Clément, removed by Andreas, re-added by Clément, and probably your PR will remove it again.

PR #2227, opened 2017-07-04
Polyhedron_demo: Fix conversion warning in Surface_reconstruction_plugin

maxGimeno:
## Summary of Changes
Fixes the warnings about `std::size_t` to `unsigned int` conversion in Surface_reconstruction_plugin
## Release Management


janetournois:
tested locally, warnings are gone

lrineau:
Are the warnings only in `master`/`integration` (CGAL-4.11-Ic-*) or are they also in CGAL-4.10?

lrineau:
I wonder why the index type is sometimes `std::size_t` and sometimes `unsigned int`.

maxGimeno:
The warnings came with #2159, which is based on master, so it shouldn't be in 4.10. Testsuite on last Monday has no warning in 4.10.

PR #2237, opened 2017-07-05
Fix testsuite when GMP is not present

sloriot:
The fix is not perfect has it only checks for GMP presence but it will fix the testsuite.

### DIFF
--- a/Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
+++ b/Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
@@ -873,6 +873,10 @@ endfunction()
 #---------------------------------------------------------------------#
 function(test_polycurve_conic_traits)
 #  echo polycurve test starting
+  if($ENV{CGAL_DISABLE_GMP})
+    MESSAGE(STATUS "test_polycurve_conic_traits requires CORE and will not be executed")
+    return()
+  endif()
   set(nt ${CORE_EXPR_NT})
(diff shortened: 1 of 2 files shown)


Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
```diff
@@ -873,6 +873,10 @@ endfunction()
 #---------------------------------------------------------------------#
 function(test_polycurve_conic_traits)
 #  echo polycurve test starting
+  if($ENV{CGAL_DISABLE_GMP})
+    MESSAGE(STATUS "test_polycurve_conic_traits requires CORE and will be executed")
```
lrineau: > will **not** be executed

sloriot:
Amended. Thanks.

sloriot:
The compilation errors are gone. We are now left with runtime errors...

lrineau:
Anyway, the results of CGAL-4.11-Ic-59 were too incomplete, and I have repushed the `integration` branch of yesterday (plus new merges).

lrineau:
@afabri A lot of Windows results of Periodic_3 did not finish before the timeout, but those that has finished did finish in a lot less than the time bound. That is strange. Could you have a look?

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-61.shtml#Periodic_3_triangulation_3

lrineau:
Sorry, wrong PR.

PR #2250, opened 2017-07-12
Workaround error in doxygen xml parsing

sloriot:
I used Mathjax symbol instead of html symbol.

Tested locally, `documentation_parser.py` is now happy.


lrineau:
Thanks.

lrineau:
I have cancelled the Travis job, because our Travis backlog is now 19 hours late, and this PR cannot break anything that Travis tests.

lrineau:
I think this issue was the reason why the `doc/master` documentation was no longer present. It should be fixed soon.

PR #2101, opened 2017-05-09
BGL: Add generic functions read_off() and write_off()

afabri:
## Summary of Changes
Add generic functions that may be overlaoded for specific models of `FaceGraph`, in case one can provide  a better implementation.

There is more to do, namely to provide the overloads, and to provide other formats than OFF.

## Release Management

* Affected package(s): BGL
* Feature/Small Feature (if any):  [BGL_load_off](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/BGL_load_off) (pre-approved Gdamiand 26 June 2017 )

PR #2265, opened 2017-07-17
Surface_mesh: Set the failbit instead of writing a warning message to std::cout

### LINKED ISSUE #2195: Misleading warning message in Surface_mesh
## Issue Details

Using the `operator>>()` with Surface_mesh, the user can run into the following message:

```
"Warning: Facets don't seem to be oriented. Loading a Soup of polygons instead."
```

But if we look at the code, here is what happens:

```cpp
      Face_index fi = sm.add_face(vr);
      if(fi == sm.null_face())
      {
       std::cout<< "Warning: Facets don't seem to be oriented. Loading a Soup of polygons instead."<<std::endl;
       sm.clear();
       return is;
      }
```
It seems to me this message is very misleading:
* From what I understand, this is much closer to an _Error_ than to a _Warning_
* The program does not "_load a Soup of polygons instead_" as suggested, it actually does nothing and returns an empty surface mesh

Note also that this behavior is not documented.

## Environment

* CGAL version: master

afabri:
## Summary of Changes

Instead of issuing a warning to `std::cout` we set the failbit in the  `operator>>(istream&, Surface_mesh&)`

## Release Management

* Affected package(s):  Surface_mesh
* Issue(s) solved (if any): fix #2195

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
@@ -2043,8 +2043,10 @@ private: //------------------------------------------------------- private data
   /// Extracts the surface mesh from an input stream in Ascii OFF, COFF, NOFF, CNOFF 
   /// format and appends it to the surface mesh `sm`.
   /// The operator reads the point property as well as "v:normal", "v:color", and "f:color".
-  /// Vertex texture coordinates are ignored.
+  /// Vertex texture coordinates are ignored. The surface mesh `sm` is initially cleared.
   /// \pre `operator>>(std::istream&,const P&)` must be defined.
+  /// \pre The data in the stream must represent a two-manifold. If this is not the case
+  ///      the `failbit` of `is` is set and the mesh cleared.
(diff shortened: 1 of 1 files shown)


lrineau:
There is already a conflict with `master`. The code was modified by the merge of PR #2101 (Generic read-write).

Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
```diff
@@ -2043,8 +2043,10 @@ class Surface_mesh
   /// Extracts the surface mesh from an input stream in Ascii OFF, COFF, NOFF, CNOFF 
   /// format and appends it to the surface mesh `sm`.
   /// The operator reads the point property as well as "v:normal", "v:color", and "f:color".
-  /// Vertex texture coordinates are ignored.
+  /// Vertex texture coordinates are ignored. The surface mesh `sm` is initially cleared.
```
lrineau: I think the implementation does not clear `sm`, but only in case of a failure.

PR #2267, opened 2017-07-17
doc.cgal.org/master: Try to be more robust on errors

lrineau:
- Use rsync instead of a symbolic link,
- and exit earlier on errors.

If there is an error, then the rsync will not overwrite the current documentation.

### DIFF
--- a/Documentation/doc/scripts/compare_testsuites.sh
+++ b/Documentation/doc/scripts/compare_testsuites.sh
@@ -40,8 +40,8 @@ cd ..
 echo "Output generated"
 if ! [ -d "$DOC_REF" ]; then
   echo "No reference given. Script is finished."
-  exit 2
+  exit 0
 fi
 
 #diff the output and the reference output, ignoring the differences in whitespaces
-diff -u -N -w ./doc_data $DOC_REF > ./diff.txt
+diff -u -N -w ./doc_data $DOC_REF > ./diff.txt || true
--- a/Documentation/doc/scripts/test_doxygen_versions.sh
+++ b/Documentation/doc/scripts/test_doxygen_versions.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e
+
 if [ "$1" == '--help' ]; then
   echo "Usage: $0 <doxygen_1> [doxygen_2] [publish_dir]"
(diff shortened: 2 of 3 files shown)


Documentation/doc/scripts/compare_testsuites.sh
```diff
@@ -40,8 +40,8 @@ cd ..
 echo "Output generated"
 if ! [ -d "$DOC_REF" ]; then
   echo "No reference given. Script is finished."
-  exit 2
+  exit 0
```
lrineau: @maxGimeno It seems that exit is a normal exit of the script. That is why I have change the exitcode to 0.

Documentation/doc/scripts/compare_testsuites.sh
```diff
@@ -40,8 +40,8 @@ cd ..
...
 fi
 
 #diff the output and the reference output, ignoring the differences in whitespaces
-diff -u -N -w ./doc_data $DOC_REF > ./diff.txt
+diff -u -N -w ./doc_data $DOC_REF > ./diff.txt || true
```
lrineau: I have added `|| true` so that the diff does not interrupt the script if it is launched with the `-e` option.

Documentation/doc/scripts/test_doxygen_versions.sh
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e
+
```
lrineau: I have added `-e` to that the errors better reported.

Documentation/doc/scripts/test_doxygen_versions.sh
```diff
@@ -44,7 +47,7 @@ make -j$NB_CORES doc  &> /dev/null
 echo "done."
 cd ../ #scripts
 echo "Creating text files for diff...."
-bash compare_testsuites.sh $PWD/build_doc/doc_output 1> /dev/null
+bash -$- compare_testsuites.sh $PWD/build_doc/doc_output 1> /dev/null
```
lrineau: `$-` is a positional parameter of the shell that contains all the options set by `set`. `-$-` is a way to pass those options to sub-shells.

PR #2275, opened 2017-07-19
Convex_hull_3: Make it work for OpenMesh

afabri:
## Summary of Changes

Add examples for using 3D convex hull functions with open mesh.  And fix a bug in convex hull code as it does use more than in the concept `MutableFaceGraph`.

## Release Management

### DIFF
--- a/Convex_hull_3/doc/Convex_hull_3/examples.txt
+++ b/Convex_hull_3/doc/Convex_hull_3/examples.txt
@@ -1,6 +1,8 @@
 /*!
 \example Convex_hull_3/dynamic_hull_3.cpp
+\example Convex_hull_3/dynamic_hull_OM_3.cpp
 \example Convex_hull_3/quickhull_3.cpp
+\example Convex_hull_3/quickhull_OM_3.cpp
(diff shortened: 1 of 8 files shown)


lrineau:
This PR could be against the bug-fix release CGAL-4.10.1, instead of CGAL-4.11-beta1.

What is best for the user who reported the issue?

lrineau:
There is unfortunately a check, added by Sébastien (cc @sloriot), that blocks the integration into `master` today.

```
.. Checking if all CMakeLists.txt project names are correct...
CMakeLists with incorrect project name
Convex_hull_3/examples/Convex_hull_3/CMakeLists.txt
```
https://travis-ci.org/CGAL/cgal/jobs/255203948#L4304

The name `Convex_hull_3_Examples` will have to be restored.

Andreas, which tool did you use to generate the CMake file? It will have to be patched accordingly.

afabri:
cgal_create_CMakeLists

afabri:
It is now the same project name as in master. No idea if that needs another round of integration.

lrineau:
If I run the check scripts manually that will be sufficient.

PR #2277, opened 2017-07-20
FIX: crash in Surface_mesh_shortest_path::nearest_to_location

ocarre:
## Summary of Changes
Fix the crash that occurs if the point passed to the Surface_mesh_shortest_path::nearest_to_location function is located on a border edge.
It now returns directly the main face if the point is located on a border edge and it does not look anymore for a non-existent opposite face.

## Release Management
* Affected package(s): Triangulated Surface Mesh Shortest Paths

### DIFF
--- a/Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
+++ b/Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
@@ -45,6 +45,7 @@
 #include <CGAL/Surface_mesh_shortest_path/internal/Cone_tree.h>
 #include <CGAL/Surface_mesh_shortest_path/internal/misc_functions.h>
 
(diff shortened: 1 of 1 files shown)


Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h
```diff
@@ -1738,29 +1738,34 @@ class Surface_mesh_shortest_path
...
-          {
-            return otherFace;
-          }
-          else if (otherFace.first.first == NULL)
-          {
-            return mainFace;
-          }
-          else
+          halfedge_descriptor oppositeHalfedge = opposite(he, m_graph);
+          if(!oppositeHalfedge->is_border())
```
ocarre: Change request has been applied. Thank you for the hint on the new API.

lrineau:
Thanks @ocarre. I am now waiting for a review from @sloriot, who is, I think, one of the maintainers of that package.

PR #2298, opened 2017-07-25
Polyhedron_demo: Fix Qt warning about duplicate filters

### LINKED ISSUE #2294: Polyhedron demo: error about file filters in I/O plugins
From `master`, version 15d9bc94d516316343b9e2ec727eb0da35558205. Open the Polyhedron demo with all plugins, then choose the action (from menu) "File"->"Open...". The text console shows those messages while the open dialog is displayed:
```
Duplicate Filter:  "selection_io_plugin"
This filter will not be available.
Duplicate Filter:  "surf_io_plugin"
This filter will not be available.
Duplicate Filter:  "vtk_plugin"
This filter will not be available.
Duplicate Filter:  "vtk_plugin"
This filter will not be available.
Duplicate Filter:  "vtk_plugin"
This filter will not be available.
```

That seems to mean that some I/O plugins will not work. That is almost certainly due to the way the plugins have been modified to support `Surface_mesh`.

Assigned to: @maxGimeno and @afabri. To be fixed ASAP for CGAL-4.11.

maxGimeno:
## Summary of Changes

This PR modifies the nameFilters of the affected IO plugins to avoid duplication.
## Release Management

* Issue(s) solved (if any): fix #2294

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/IO/Selection_io_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/IO/Selection_io_plugin.cpp
@@ -12,11 +12,12 @@ class Polyhedron_demo_selection_io_plugin :
 public:
 #ifdef USE_SURFACE_MESH
     QString name() const { return "selection_io_sm_plugin"; }
+    QString nameFilters() const { return "Selection files (Surface_mesh) (*.selection.txt)"; }
 #else
     QString name() const { return "selection_io_plugin"; }
+    QString nameFilters() const { return "Selection files (Polyhedron) (*.selection.txt)"; }
 #endif
 
-    QString nameFilters() const { return "Selection files (*.selection.txt)"; }
 
     bool canLoad() const { return true; }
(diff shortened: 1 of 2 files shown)


maxGimeno:
I could not reproduce the issue so someone please tries and validates that it works. Don't forget to remove old library files from your build directory before you try.

lrineau:
The issue #2294 is also about the `surf_io_plugin`. Will this PR fix it as well?

maxGimeno:
I didn't have any problem with the surf_io_plugin, which is the only plugin able to read Amira files, as far as I know. So there shouldn't be any problem. Wasn't there an old version of the surf_io_plugin or the surf_io_sm_plugin remaining in you build dir when you tested ?

lrineau:
I confirm the fix. Tested locally on my Linux with https://github.com/CGAL/cgal/commit/5658adbc263aa01f0adcaecbfaf9148e2010e928 plus this PR merged in it.

PR #2301, opened 2017-07-25
Polyhedron_demo : Robustify split_polyhedron_plugin 

maxGimeno:
## Summary of Changes
This PR robustifies the split_polyhedron_plugin by checking if the extracted connected components are valid. If this is not the case, a warning popup appears to alert the user and ask to call Remove self intersections.


sloriot:
I think it will conflict with the fix in #2304. In particular, if `is_valid()` was already returning false before calling the split function. Since `Polyhedron_3` and `Surface_mesh` can _handle_ non-manifold vertices I would close this PR. @maxGimeno Do you agree?

maxGimeno:
#2304 fixes the issue.

lrineau:
The changes to `Polyhedron/demo/Polyhedron/Plugins/PMP/Repair_polyhedron_plugin.cpp` seemed good. You should push it to another PR.

PR #2304, opened 2017-07-26
Handle non-manifold vertices when copying a mesh

sloriot:
## Summary of Changes

The way the next halfedge of each border halfedge was set assumed that no vertex was non-manifold. With the following patch we use the connectivity of the input mesh for setting the next pointer.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/BGL/include/CGAL/boost/graph/copy_face_graph.h
+++ b/BGL/include/CGAL/boost/graph/copy_face_graph.h
@@ -46,6 +46,7 @@ void copy_face_graph_impl(const SourceMesh& sm, TargetMesh& tm,
 {
   typedef typename boost::graph_traits<SourceMesh>::vertex_descriptor sm_vertex_descriptor;
   typedef typename boost::graph_traits<TargetMesh>::vertex_descriptor tm_vertex_descriptor;
+  typedef typename boost::graph_traits<TargetMesh>::vertex_iterator   tm_vertex_iterator;
 
   typedef typename boost::graph_traits<SourceMesh>::face_descriptor sm_face_descriptor;
   typedef typename boost::graph_traits<TargetMesh>::face_descriptor tm_face_descriptor;
@@ -153,8 +154,10 @@ void copy_face_graph_impl(const SourceMesh& sm, TargetMesh& tm,
   }
 
   // update halfedge vertex of all but the vertex halfedge
-  BOOST_FOREACH(tm_vertex_descriptor v, vertices(tm))
+  for(tm_vertex_iterator vit = vertices(tm).first;
+      vit != vertices(tm).second; ++vit)
   {
+    tm_vertex_descriptor v = *vit;
     tm_halfedge_descriptor h = halfedge(v, tm);
     tm_halfedge_descriptor next_around_vertex=h;
     do{
--- a/BGL/include/CGAL/boost/graph/copy_face_graph.h
+++ b/BGL/include/CGAL/boost/graph/copy_face_graph.h
@@ -61,6 +61,7 @@ void copy_face_graph_impl(const SourceMesh& sm, TargetMesh& tm,
     conv;
 
   std::vector<tm_halfedge_descriptor> tm_border_halfedges;
+  std::vector<sm_halfedge_descriptor> sm_border_halfedges;
 
   tm_face_descriptor tm_null_face = boost::graph_traits<TargetMesh>::null_face();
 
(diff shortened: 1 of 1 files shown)


lrineau:
There is [a line completely red in the testsuite][1], because of a runtime error:
```
- ProgramOutput.extract_segmentation_into_mesh_example.ArchLinux
CGAL::Random()::get_seed() = 1501122203
Segment 0's area is : 0.200531
Expr: next(tm_h_prev, tm) == boost::graph_traits<TargetMesh>::null_halfedge()
File: /mnt/testsuite/include/CGAL/boost/graph/copy_face_graph.h
Line: 147
./cgal_test_with_cmake: line 70: 12281 Aborted                 (core dumped) ./extract_segmentation_into_mesh_example
```
  [1]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-77.shtml#Surface_mesh_segmentation_Examples

sloriot:
It makes me think that if the target mesh is for example a `Surface_mesh` with garbage inside, an assertion about next halfedge of a newly create halfedge might break since it can be non null. In my initial code, I've put a call to `set_next` with `null_halfedge` but it was breaking because `set_prev` was internally called with a null halfedge. @afabri shall we update all call to `set_next` in the models so that `set_prev` is not call if a null halfedge is passed?

sloriot:
The error is not related to this fix and is in master. It is fixed (tested locally) with #2311.

lrineau:
I think this PR is responsible for segfaults in [Convex_hull_3][ch] and [Point_set_processing_3][psp] testsuites, on Windows 64 bits with MSVC 2015 and 2017.

[ch]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-78.shtml#Convex_hull_3
[psp]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.11-Ic-78.shtml#Point_set_processing_3

@janetournois Would you have time to checkout yesterday's integration (CGAL/cgal-dev@fde4da14f381cc1dd6ceeac294f87fa35f5a8dec) and try to compile and run, for example, `test/Convex_hull_3/quickhull_test_3.cpp` in Release with assertions, with MSVC 2015 or 2017?

janetournois:
one option is to
> update all calls to `set_next` in the models so that `set_prev` is not call if a null halfedge is passed?
but we could also change the `add_edge` function to make sure its `next` is the `null_halfedge`. Having something recycled without knowing about it is dangerous...

janetournois:
@lrineau I was looking at these two PRs. I am going to test the things you mention

janetournois:
I can reproduce the crash.
I am having a look at it...

janetournois:
This bug is hard to understand.
It happens only in `Release` mode, not even in `RelWithDebInfo`, and it is instable. For time to time, it either
- falls in an infinite loop inside `vertices(tm)` (line 156 of the latest commit), where `v` always points to the same point `(0, 0, 0)`
- crashes after the end of the first loop on `vertices(tm)` (same place)

janetournois:
replacing `BOOST_FOREACH` by a simple `for` loop resolves the crash... for some reason (??)

sloriot:
@janetournois  I agree about `add_edge()`, it should even be the same for `add_vertex()` and `add_face()` in general. I also think DS like `Surface_mesh` should reinit properties of elements newly added in a formerly occupy location but @afabri did not like the idea IIRC.

lrineau:
This PR has been tested last night, the results seem good, and the release CGAL-4.11-beta1 should be tomorrow.

Should I merge this PR today, or wait for modifications to `add_vertex`, `add_edge`, and others?

janetournois:
@sloriot I agree with re-initializing properties when recycling happens, imho it is too risky to keep outdated properties "valid".
@lrineau I think the question about `add_vertex` etc is quite independent from this PR, so the PR should be merged, and this discussion continued, maybe in a new issue. @sloriot do you agree?

lrineau:
> @lrineau I think the question about add_vertex etc is quite independent from this PR, so the PR should be merged, and this discussion continued, maybe in a new issue. @sloriot do you agree?

Good. I plan to merge this PR in `master` today, during the afternoon, to compose the version of `master` that will be CGAL-4.11-beta1.

BGL/include/CGAL/boost/graph/copy_face_graph.h
```diff
@@ -153,8 +154,10 @@ void copy_face_graph_impl(const SourceMesh& sm, TargetMesh& tm,
   }
 
   // update halfedge vertex of all but the vertex halfedge
-  BOOST_FOREACH(tm_vertex_descriptor v, vertices(tm))
+  for(tm_vertex_iterator vit = vertices(tm).first;
+      vit != vertices(tm).second; ++vit)
   {
+    tm_vertex_descriptor v = *vit;
```
lrineau: This was the previous instance of the "`BOOST_FOREACH` problem".
- the loop was not nested in another loop,
- the triangular mesh was indeed modified, but its vector of vertices was not.

lrineau: Could there be something that cause the reallocation of the vector of vertices of the mesh during the loop?

lrineau: - Unless the second issue with `BOOST_FOREACH`: this time MSVC 2017 was impacted (but which update of it)?

lrineau: Is not the loop at https://github.com/janetournois/cgal/blob/9b00be7d7bddee92e6c97f7af089aa0b0a52cc25/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h#L283 dangerous?

sloriot: The issue with MSVC is for a user that was using a version of this file that was distributed privately with CGAL 4.9, so way before this fix.

sloriot: The helper function is only using next/opposite to turn around the target vertex

sloriot: All the vertices are created in the loop before so no reallocation after using tm vertex vector.

PR #2310, opened 2017-07-27
Polyhedron_demo: No shift for Edit_box

maxGimeno:
## Summary of Changes
Remove the need for shift to manipulate the Editable box, which avoids problems with a selection_item.
## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
@@ -1022,9 +1022,11 @@ void Scene_edit_box_item_priv::reset_selection()
 }
 
 //intercept events for picking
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
```diff
@@ -1024,12 +1024,12 @@ void Scene_edit_box_item_priv::reset_selection()
 //intercept events for picking
 bool Scene_edit_box_item::eventFilter(QObject *, QEvent *event)
 {
+  QGLViewer* viewer = *QGLViewer::QGLViewerPool().begin();
```
lrineau: `viewer` should be equal to the first argument `QObject *` of the eventFilter. The more we use `QGLViewerPool()`, the more difficult it will be to have multiple viewers.

PR #2311, opened 2017-07-27
Fix Face_filtered_graph bugs

sloriot:
## Summary of Changes

Fix next/prev for `Face_filtered_graph` as well as the change of selected faces.

## Release Management

* Affected package(s): BGL


lrineau:
There are segfault in Convex_hull_3 and PSP testsuite. I cannot tell if this PR is right or not. See https://github.com/CGAL/cgal/pull/2304#issuecomment-318595954.

sloriot:
This PR is safe in my opinion as it only affects Face_filtered_graph and fixes _obvious_ bugs.

PR #2312, opened 2017-07-27
Polyhedron_demo: Add visibility to the status messages

maxGimeno:
## Summary of Changes
Add color to the status text so it is more visible. The status is the text displayed at the bottom of the mainWindow, which is also written in the console widget.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -792,7 +792,7 @@ void MainWindow::message(QString message, QString colorName, QString font) {
     message.remove(message.length()-1, 1);
   }
   statusBar()->showMessage(message, 5000);
-  QTimer::singleShot(5000, this, SLOT(resetStatusColor()));
+  QTimer::singleShot(5000, [this]{this->statusBar()->setStyleSheet("");});
   message = "<font color=\"" + colorName + "\" style=\"font-style: " + font + ";\" >" +
       message + "</font><br>";
   message = "[" + QTime::currentTime().toString() + "] " + message;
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2029,3 +2033,8 @@ void MainWindow::set_facegraph_mode_adapter(bool is_polyhedron)
   else
     set_face_graph_default_type(SURFACE_MESH);
 }
+
+void MainWindow::resetSatusColor()
```
lrineau: typo

maxGimeno: Status is wanted this way. It is what is displayed in the QStatusBar.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2029,3 +2033,8 @@ void MainWindow::set_facegraph_mode_adapter(bool is_polyhedron)
   else
     set_face_graph_default_type(SURFACE_MESH);
 }
+
+void MainWindow::resetStatusColor()
+{
+  statusBar()->setStyleSheet("color: black");
+}
```
lrineau: The default color may not be black. If the user uses a "Dark" style, then the default color might be closer to white. Is not their a way to restore the default style sheet?

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -792,22 +792,26 @@ void MainWindow::message(QString message, QString colorName, QString font) {
     message.remove(message.length()-1, 1);
   }
   statusBar()->showMessage(message, 5000);
+  QTimer::singleShot(5000, this, SLOT(resetStatusColor()));
```
lrineau: As `resetStatusColor()` is used only once, here, that is a good opportunity to use a lambda expression, instead of a new slot: since Qt-5.4, `QWidget::singleShot` can be connected to lambda expressions.

lrineau:
There is a build error in Travis, but it is due to a too-old version of Qt (5.2.1). See the issue #2323.

PR #2314, opened 2017-07-28
Polyhedron_demo: Enhance the plugins management in Preferences

maxGimeno:
## Summary of Changes

Add a hierarchy to the list of plugins in the Preferences dialog based on their location :

preferences-screenshot

Edit : 
now, it looks like this : 
preferences-screenshot
 and items have a tooltip if they are not green, saying why they were not loaded. (Apparently, my screenshot manager does not capture the cursor neither the tooltips, sorry.)

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1669,9 +1669,9 @@ void MainWindow::on_actionPreferences_triggered()
   //QStandardItemModel* iStandardModel = new QStandardItemModel(this);
 
   std::vector<QTreeWidgetItem*> items;
-  QBrush successBrush(QColor(Qt::green)),
-      errorBrush(QColor(Qt::red)),
-      ignoredBrush(QColor(Qt::lightGray));
+  QBrush successBrush(Qt::green),
+      errorBrush(Qt::red),
+      ignoredBrush(Qt::lightGray);
 
   //add blacklisted plugins
   Q_FOREACH (QString path, PathNames_map.keys())
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -61,14 +61,14 @@
 #  include <QScriptValue>
 #include "Color_map.h"
 using namespace CGAL::Three;
-QScriptValue 
-myScene_itemToScriptValue(QScriptEngine *engine, 
+QScriptValue
+myScene_itemToScriptValue(QScriptEngine *engine,
                           CGAL::Three::Scene_item* const &in)
-{ 
-  return engine->newQObject(in); 
+{
+  return engine->newQObject(in);
 }
 
-void myScene_itemFromScriptValue(const QScriptValue &object, 
+void myScene_itemFromScriptValue(const QScriptValue &object,
                                  CGAL::Three::Scene_item* &out)
 {
   out = qobject_cast<CGAL::Three::Scene_item*>(object.toQObject());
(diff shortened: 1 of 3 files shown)


lrineau:
You should add a screenshot.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1652,50 +1663,49 @@ void MainWindow::on_actionPreferences_triggered()
...
+      name.remove(QRegExp("\\..*"));
+      QTreeWidgetItem *item = new QTreeWidgetItem(pluginItem);
+      item->setText(1, name);
+      if(plugin_blacklist.contains(name))
+        item->setCheckState(0, Qt::Checked);
+      else
+        item->setCheckState(0, Qt::Unchecked);
+      items.push_back(item);
+    }
+  }
```
lrineau: The `loadPlugins` function should store enough information, so that this one does not have to scan the directories again.

And, anyway, we do not care which are the available at the time the dialog is displayed: we want to know which ones were actually loaded.

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -388,7 +388,7 @@ protected Q_SLOTS:
   void setMenus(QString, QString, QAction *a);
   /// plugin black-list
   QSet<QString> plugin_blacklist;
-  QList<QDir> plugins_directories;
+  QMap<QString, std::vector<QString> > PathNames_map; //For each non-empty plugin directory, contains a vector of plugin names
```
lrineau: There a problem with the naming of that variable. It should follow the CGAL naming convention `path_names_map` (or Qt convention `pathNamesMap`).

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -612,8 +616,14 @@ void MainWindow::loadPlugins()
...
-        load_plugin(abs_name, true);
-        loaded.insert(abs_name);
+        if(load_plugin(abs_name, true))
+        {
+          QString name = QFileInfo(fileName).fileName();
+          name.remove(QRegExp("^lib"));
+          name.remove(QRegExp("\\..*"));
+          PathNames_map[pluginsDir.absolutePath()].push_back(name);
+          loaded.insert(abs_name);
+        }
```
lrineau: If think the lines 
```C++
         QString name = QFileInfo(fileName).fileName();
         name.remove(QRegExp("^lib"));
         name.remove(QRegExp("\\..*"));
         PathNames_map[pluginsDir.absolutePath()].push_back(name);
```
should be in `load_plugin`. That way, if the user loads a plugin with the "File, Load Plugin" menu entry, it will be registered as well.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -528,7 +530,7 @@ void MainWindow::load_plugin(QString fileName, bool blacklisted)
       {
         if ( plugin_blacklist.contains(name) ){
           qDebug("### Ignoring plugin \"%s\".", qPrintable(fileName));
-          return;
+          return true;
```
lrineau: If the plugin is blacklisted, it is not loaded, and I think the function should return `false`, otherwise the plugin will be listed in the loaded plugins.

maxGimeno: Actually the function returns true if fileName leads to a plugin, not if the plugin is loaded. I needed a way to get all plugins in load_plugins, and load_plugin() is the one that filters the files, so I used the return value. using a reference to a bool might be clearer though.

maxGimeno: I used this retur value to get all files that are plugins, not only those that are loaded, as the filtering is done in load_plugin() and not in load_plugins().

lrineau:
The Travis compilation failed with clang:
https://travis-ci.org/CGAL/cgal/jobs/310114410#L3397

lrineau:
@maxGimeno Does it fix an issue?

PR #2324, opened 2017-07-31
Make Periodic_3_regular_triangulation_3 compatible with Alpha_Shapes_3

MaelRL:
## Summary of Changes

The `Is_Gabriel` function (and some other minor details) were missing for P3RT3 to be usable in Alpha Shapes 3D. An example is added and the doc is modified.

Geomview's functions that output Alpha Shapes are fixed to work with periodic triangulations.

I have also added a test to `Periodic_3_triangulation_3` to check that a fake periodic triangulation built using `Regular_triangulation_3` gives the same thing as if it were built with `Periodic_3_regular_triangulation_3`.

## Release Management

* Affected package(s): `Alpha Shapes 3D`, `Periodic_3_triangulation_3`

### DIFF
--- a/Periodic_3_triangulation_3/test/Periodic_3_triangulation_3/test_p3rt3_versus_rt3.cpp
+++ b/Periodic_3_triangulation_3/test/Periodic_3_triangulation_3/test_p3rt3_versus_rt3.cpp
@@ -1,0 +1,158 @@
+// Copyright (c) 2017  INRIA Sophia-Antipolis (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org).
+// You can redistribute it and/or modify it under the terms of the GNU
+// General Public License as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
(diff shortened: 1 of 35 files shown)


MaelRL:
Ideally, this would make it into `4.11` since it is just a sub-part of `P3RT3`, but it is not urgent (there is a public branch to use this code if needed).

Periodic_3_triangulation_3/test/Periodic_3_triangulation_3/test_p3rt3_versus_rt3.cpp
```diff
@@ -0,0 +1,158 @@
...
+// either version 3 of the License, or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// Author(s)     : Mael Rouxel-Labbé
+
```
sloriot: no need for copyright headers in tests and examples

Issue #2325, opened 2017-07-31
CGAL-4.10, Convex_hull_3, OpenMesh: write_OFF compilation error


lrineau:
Something to fix before CGAL-4.10.1 can be published: with `releases/CGAL-4.10-branch`, the testsuite of Convex_hull_3 had compilation errors as soon as OpenMesh is detected and used:

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.10.1-I-205.shtml#Convex_hull_3_Examples

```
[ 50%] Building CXX object CMakeFiles/quickhull_OM_3.dir/quickhull_OM_3.cpp.obj
	C:\PROGRA~2\MICROS~3.0\VC\bin\amd64\cl.exe  @C:\Users\Jane\AppData\Local\Temp\nmE480.tmp
quickhull_OM_3.cpp
E:\CGAL\test\CGAL-4.10.1-I-205\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Release-64bits\test\Convex_hull_3_Examples\quickhull_OM_3.cpp(36): error C2039: 'write_off': is not a member of 'CGAL'
E:\CGAL\test\CGAL-4.10.1-I-205\include\CGAL/boost/graph/graph_traits_PolyMesh_ArrayKernelT.h(668): note: see declaration of 'CGAL'
E:\CGAL\test\CGAL-4.10.1-I-205\cmake\platforms\x64_Cygwin-Windows10_MSVC2015-Release-64bits\test\Convex_hull_3_Examples\quickhull_OM_3.cpp(36): error C3861: 'write_off': identifier not found
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.10.1-I-205/Convex_hull_3_Examples/TestReport_jtournois_x64_Cygwin-Windows10_MSVC2015-Release-64bits.gz

PR #2326, opened 2017-07-31
Convex_hull_3: Fix OpenMesh example

afabri:
Fix for #2325

### DIFF
--- a/Convex_hull_3/examples/Convex_hull_3/quickhull_OM_3.cpp
+++ b/Convex_hull_3/examples/Convex_hull_3/quickhull_OM_3.cpp
@@ -32,8 +32,7 @@ int main(int argc, char* argv[])
   Mesh sm;
   
   CGAL::convex_hull_3(points.begin(), points.end(), sm);
-
-  CGAL::write_off((argc>2)?argv[2]:"out.off", sm);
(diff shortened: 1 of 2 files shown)


Convex_hull_3/examples/Convex_hull_3/quickhull_OM_3.cpp
```diff
@@ -32,8 +32,7 @@ int main(int argc, char* argv[])
   Mesh sm;
   
   CGAL::convex_hull_3(points.begin(), points.end(), sm);
-
-  CGAL::write_off((argc>2)?argv[2]:"out.off", sm);
```
sloriot: Doesn't this spot a bug in `write_off()`?

lrineau: I think the generic `write_off` has been introduced in CGAL-4.11. And the PR #2275 has been tested in `integration` only, the branch for CGAL-4.11.

lrineau:
Note to Sébastien @sloriot: be careful when you will merge this PR in `master`! You will have to merge it with the `ours` strategy, to ignore the changes: `git merge -s ours ...`, because we do not want that patch in `master`.

PR #2328, opened 2017-08-01
Polyhedron_demo: Robustify Mean_curvature_flow_skeleton_plugin

maxGimeno:
## Summary of Changes

This PR is a rework of this plugin, using group items to store variables that were hold by the plugin before, allowing it to manage several items and avoiding segfaults. 
It also adds a label to wxplain how it works, so the user is not surprised anymore when he/she asks for a skeletonization before runing any iteration.

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Scene_mcf_item.h
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Scene_mcf_item.h
@@ -1,0 +1,50 @@
+#ifndef SCENE_MCF_ITEM_H
+#define SCENE_MCF_ITEM_H
+
+#ifdef scene_mcf_item_EXPORTS
+#  define SCENE_MCF_ITEM_EXPORT Q_DECL_EXPORT
+#else
+#  define SCENE_MCF_ITEM_EXPORT Q_DECL_IMPORT
+#endif
+
+#include <CGAL/Three/Scene_group_item.h>
+#include <CGAL/extract_mean_curvature_flow_skeleton.h>
+#ifdef USE_SURFACE_MESH
+#include "SMesh_type.h"
+typedef SMesh Face_graph;
+#else
+#include "Polyhedron_type.h"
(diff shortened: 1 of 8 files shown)


sloriot:
The parameter for medially centered trade off should only be available is medially center is checked

sloriot:
I've updated the help text. Maybe you should consider avoid a help button creating a dialog with the text rather than having it in the widget.

sloriot:
All issues mentioned are fixed. For some reason, I need to click on OK as much time as I created a mcf group to close the help dialog box.

Polyhedron/demo/Polyhedron/Plugins/PMP/Scene_mcf_item.h
```diff
@@ -0,0 +1,50 @@
+#ifndef SCENE_MCF_ITEM_H
+#define SCENE_MCF_ITEM_H
+
+#ifdef scene_mcf_item_EXPORTS
+#  define SCENE_MCF_ITEM_EXPORT Q_DECL_EXPORT
+#else
+#  define SCENE_MCF_ITEM_EXPORT Q_DECL_IMPORT
+#endif
```
lrineau: That will not work. The code is compiled into two different libraries, named `scene_mcf_poly_item` and `scene_mcf_sm_item`. The macro `scene_mcf_item_EXPORTS` will never be defined.

You should add, before line 4:
```C++
#ifdef scene_mcf_sm_item_EXPORTS
#  define scene_mcf_item_EXPORTS
#endif
#ifdef scene_mcf_poly_item_EXPORTS
#  define scene_mcf_item_EXPORTS
#endif
```
or rewrite the line 4 as:
```C++
#if defined(scene_mcf_poly_item_EXPORTS) || defined(scene_mcf_poly_item_EXPORTS)
```

sloriot:
The travis error is an apt-get error.

lrineau:
It can happen because of temporary network delays. I have restarted the failed sub-job.

PR #2333, opened 2017-08-02
Move detect_features_in_polyhedra.h in internal until it is documented

maxGimeno:
## Summary of Changes
This PR moves the files  detect_features_in_polyhedra.h and  detect_features_in_polyhedra_fwd.h in internal. The will later be replaced and documented.

## Release Management

* Affected package(s):PMP


lrineau:
I object. There are external projects using that header file. Stop moving header files without providing compatibility headers.

lrineau:
Anyway, the documentation of that header is planned soon. It was not documented in CGAL-4.10, it will not be documented in CGAL-4.11 either, but it will be in CGAL-4.12.

sloriot:
@lrineau  The change was done in 4.11 so I do not see why a compatibility header would be needed. I object against closing the PR

lrineau:
4.11 is already published. I do not want to have a special support for 4.11 beta 1 in our projects, in addition to 4.10 and 4.11. What is more, our handling of `CGAL_VERSION_NR` is rather bad, and I think that there is currently no way to discriminate between a release and its beta versions. I still recommend and want to reject this PR: it will break code, and I see no really advantage.

Issue #2225, opened 2017-07-03
MinSphereOfSpheresTraits::D


mglisse:
[The documentation of the concept][doc] says:
```C++
    typedef unspecified_type D;
```
as if `D` was a type, but this appears in a section called "Constants" and is used as `static const int D = Traits::D;`, so I assume the documentation of the concept should be fixed.

More importantly, this interface is impossible to use with a dynamic dimension, which looks like a very strong limitation.

  [doc]: http://doc.cgal.org/4.10/Bounding_volumes/classMinSphereOfSpheresTraits.html

sgiraudot:
Inside the code of each model, it is an `int` constant:

```c++
Min_sphere_of_spheres_d_traits_2.h:    static const int D = 2;                  // dimension
Min_sphere_of_spheres_d_traits_3.h:    static const int D = 3;                  // dimension
Min_sphere_of_spheres_d_traits_d.h:    static const int D = Dim_;               // dimension
```

It's only in the separated header files that they are documented as the concept:

```c++
Min_sphere_of_spheres_d_traits_2.h:typedef unspecified_type D; 
Min_sphere_of_spheres_d_traits_3.h:typedef unspecified_type D; 
Min_sphere_of_spheres_d_traits_d.h:typedef unspecified_type D;
```

It seems to me that we should simply replace the `typedef unspecified_type` in both the concept's doc and the three model's doc by `static const int`, as it seems consistent with the code. If we want to keep the implementation opened to dynamic dimension, maybe only document `static int` in the concept?

mglisse:
> It seems to me that we should simply replace the typedef unspecified_type in both the concept's doc and the three model's doc by static const int, as it seems consistent with the code.

That's the minimal fix for consistency.

> If we want to keep the implementation opened to dynamic dimension, maybe only document static int in the concept?

That wouldn't really fit with the way we do things in the rest of CGAL, I'd rather get a consistent interface. But if it is the only (or at least easiest) way to get the functionality, I'll take it.

sgiraudot:
Alright, I made a pull request to integrate the first fix. The dynamic dimension problem probably requires more discussion.

mglisse:
Thanks for the patch. I agree that handling the dynamic dimension case, although it should not be very hard, is not as straightforward as updating the doc to match the code, and it is likely to require a breaking change in the interface.

mglisse:
@lrineau Should I open a separate issue asking for dynamic dimension support if this one is closed?

[...]

PR #2352, opened 2017-08-17
Bounding volumes: Fix wrong dimension doc

sgiraudot:
## Summary of Changes

The dimension is documented as a type (unspecified), which is wrong and does not make any sense. This PR makes it consistent with the code (dimension is `static const int` in every case).

## Release Management

* Affected package(s): Bounding Volumes
* Issue(s) solved: fix https://github.com/CGAL/cgal/issues/2225

### DIFF
--- a/Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_3.h
+++ b/Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_3.h
@@ -35,7 +35,7 @@ public:
 /*!
 is the constant 3, i.e.\ the dimension of \f$ \mathbb{R}^3\f$. 
 */ 
-typedef unspecified_type D; 
+static const int D;
 
 /// @} 
 
--- a/Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_d.h
+++ b/Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_d.h
@@ -17,6 +17,8 @@ a model for concept `MinSphereOfSpheresTraits`. It uses the
 type `FT` of concept `MinSphereOfSpheresTraits`: It must be 
 either `double` or `float`, or an exact number type. 
 
+\tparam Dim is the dimension.
+
 \tparam UseSqrt fulfills the 
 requirements of type `Use_square_roots` of concept 
 `MinSphereOfSpheresTraits`: It must be either `Tag_true` or `Tag_false`
@@ -28,7 +30,7 @@ concept `MinSphereOfSpheresTraits`: It must be either
 and it defaults to `Default_algorithm`.
 
 */
-template< typename K, typename FT, typename Dim, typename UseSqrt, typename Algorithm >
+template< typename K, typename FT, int Dim, typename UseSqrt, typename Algorithm >
 class Min_sphere_of_spheres_d_traits_d {
 public:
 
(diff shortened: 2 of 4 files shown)


Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_3.h
```diff
@@ -35,7 +35,7 @@ class Min_sphere_of_spheres_d_traits_3 {
 /*!
 is the constant 3, i.e.\ the dimension of \f$ \mathbb{R}^3\f$. 
 */ 
-typedef unspecified_type D; 
+static const int D;
```
mglisse: I wonder if we should document it as `static const int D = 3;`. Doesn't matter.

sgiraudot: Yes, I hesitated to do this, but as the original doc was completed by `is the constant 3`, I thought I'd leave it that way.

Bounding_volumes/doc/Bounding_volumes/CGAL/Min_sphere_of_spheres_d_traits_d.h
```diff
@@ -38,7 +38,7 @@ class Min_sphere_of_spheres_d_traits_d {
 /*!
 is the constant `Dim`. 
 */ 
-typedef unspecified_type D; 
+static const int D;
```
mglisse: Probably "is the constant Dim::value" now? Maybe say somewhere in this file that the template parameter Dim cannot be Dynamic_dimension_tag, it has to be `Dimension_tag<D>`?

sgiraudot: Actually that seems like another doc problem: `Dim` is a template parameter documented as `typename Dim`, but it seems to me that it should be `int Dim`. Well, actually I checked and this is exactly what is done in the include file…

```c++
  template<typename K_,                      // kernel
    typename FT_,                            // number type
    int Dim_,                                // dimension
    typename UseSqrt_ = Tag_false,           // whether to use square-roots
    typename Algorithm_ = Farthest_first_heuristic> // algorithm to use
  class Min_sphere_of_spheres_d_traits_d {
  public: // types:
    typedef FT_ FT;
    typedef FT_ Radius;
    typedef typename K_::Point_d Point;
    typedef std::pair<Point,Radius> Sphere;
    typedef typename K_::Cartesian_const_iterator_d Cartesian_const_iterator;
    typedef UseSqrt_ Use_square_roots;
    typedef Algorithm_ Algorithm;

  public: // constants:
    static const int D = Dim_;               // dimension
```

Here again, I would be tempted to simply correct the documentation and replace `typename Dim` by `int Dim` which is exactly what the code is.

(At some point, we should really get rid of these separated documentation header files, I know this is something inherited from old doc generation of CGAL, but it is really error-prone.)

sgiraudot: (Note that the example is clear on that point too:

```c++
const int D = 3;                          // dimension of points
typedef CGAL::Min_sphere_of_spheres_d_traits_d<K,FT,D> Traits;
```
)

mglisse: Indeed, makes sense, let's first make the doc match the code.

PR #2358, opened 2017-08-22
Surface mesh: reinit connectivies when reusing garbage memory

sgiraudot:
## Summary of Changes

When removing simplices from a `CGAL::Surface_mesh` object, they are not deleted and their memory can be reused later on. The problem is that the connectivity structure of a newly added simplex is not reinitialized, which can lead to problems.

For example, if you remove an isolated vertex (which should not pose any structure issue) and add a new one, the new one will be considered adjacent to a random face because the connectivity structure was used to keep track of removed vertices. Although it is said in the manual that removing isolated vertices could leave the surface mesh in an invalid state, it seems counter-intuitive to get an invalid state after _adding_ a new vertex (especially if the previous removed vertex was isolated).

This PR adds a reinitialization step of the connectivity structures just after their addition, in the case garbage memory is reused. It does not break what is currently working and it decreases the chances of running into an unexpected behavior. I benched it and I saw no visible slowing down.

## Release Management

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
@@ -61,6 +61,9 @@ public:
     /// Extend the number of elements by one.
     virtual void push_back() = 0;
 
+    /// Reset element to default value
+    virtual void reset(size_t idx) = 0;
+
     virtual bool transfer(const Base_property_array& other) = 0;
 
     /// Let two elements swap their storage place.
@@ -117,6 +120,11 @@ public: // virtual interface of Base_property_array
     virtual void push_back()
     {
         data_.push_back(value_);
+    }
+
+    virtual void reset(size_t idx)
+    {
+        data_[idx] = value_;
     }
 
     bool transfer(const Base_property_array& other)
@@ -395,6 +403,13 @@ public:
         for (std::size_t i=0; i<parrays_.size(); ++i)
             parrays_[i]->push_back();
         ++size_;
+    }
+
+    // reset element to its default property values
+    void reset(size_t idx)
+    {
+        for (std::size_t i=0; i<parrays_.size(); ++i)
+            parrays_[i]->reset(idx);
     }
 
     // swap elements i0 and i1 in all arrays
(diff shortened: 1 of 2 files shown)


sloriot:
There was a discussion about those issues in #2304 (I should have created an issue) about properties in general that should be reset. I haven't looked at the code but I guess a more global fix can be done.
What do you think?

sgiraudot:
It seems complicated to handle all properties, especially because there can be user-defined ones on which the internal code as few control (if not none).

Handling the internal properties seems like something that is expected to be done, but maybe the other ones can be left under the responsibility of the user?

sloriot:
IIRC correctly a default value is associated to a property so I'm surprised we can't take advantage of it.

sgiraudot:
Actually, I think an easy way to do it would be to add a method `reset(Index idx)` in the property container class: currently, this class has a method `push_back()` (with no argument) that creates new elements for all properties with default values. This method is used when creating a new item (from scratch). If the new item is created from existing memory, we could call a method `reset()` with the existing index.

sloriot:
Sounds good and would also apply to your scenario. Are you willing to do it that way?

sgiraudot:
I did it and it seems to work well.

Still, I'm seeing something quite shady in the `add_edge()` function:
```c++
    /// adds a new edge, and resizes edge and halfedge properties if necessary.
    Halfedge_index add_edge()
    {
      Halfedge_index h0, h1;
      size_type inf = -1;
      if(edges_freelist_ != inf){
        size_type idx = edges_freelist_;
        edges_freelist_ = (size_type)hconn_[Halfedge_index(edges_freelist_)].next_halfedge_;
        --removed_edges_;
        eremoved_[Edge_index(Halfedge_index(idx))] = false;
        hprops_.reset(Halfedge_index(idx));
        return Halfedge_index(idx);
      } else {
        eprops_.push_back();
        hprops_.push_back();
        hprops_.push_back();

        return Halfedge_index(num_halfedges()-2);
      }
    }
```

It seems to me that the `eprops_` and `hprops_` (properties of edges and halfedges respectively) are not really handled well in the case of using garbage memory (first case of the `if`).

(Note that this is a general remark on this method, as my patch did not introduce any different behavior than the one that already existed – except that one of the `hprops_` is now initialized.)

lrineau:
This PR has been tested on Friday night, with success. Is it ready to be merged into `master`, for CGAL-4.12?

sgiraudot:
Yes it is, it's quite self-contained and if it passes the testsuite, I can't see what would go wrong.

PR #2360, opened 2017-08-23
Update license files

sloriot:
This makes the creation of license.h file easier by displaying the name of the macros to be defined.


afabri:
I tested it on Windows.  It works for warnings and errors. I made a suggestion in a private mail to split the warning in two lines which makes it more readable and easier to grep for producing the `#define`s for license.h

PR #2365, opened 2017-08-23
Fix compilation errors with new implement of assert() in glibc

lrineau:
## Summary of Changes

In glibc 2.25 and 2.26, the implementation of `assert()` triggers
compilation errors with values that are not comparable to an `int`, such
a iostream, or `optional`.

That is the "bug" https://sourceware.org/bugzilla/show_bug.cgi?id=21972

In my opinion that is not really a bug: the value test by `assert` is
supposed to be an integral value, and not something explicitly
convertible to `bool`.

This commit uses an explicit conversion to `bool`, in all problematic
calls to `assert()`.

The problem was seen on my platform, Linux Fedora 26.

## Release Management

* Affected package(s): a lot!

**The branch is against `master`, but it should be backported to `4.10.1` and `4.11`. There are conflicts, about missing files.**

### DIFF
--- a/AABB_tree/test/AABB_tree/aabb_test_singleton_tree.cpp
+++ b/AABB_tree/test/AABB_tree/aabb_test_singleton_tree.cpp
@@ -39,8 +39,8 @@ int main()
     tree.accelerate_distance_queries();
     tree.all_intersections(triangle_query, devnull);
     tree.all_intersected_primitives(triangle_query, devnull);
-    assert(tree.any_intersected_primitive(triangle_query));
-    assert(tree.any_intersection(triangle_query));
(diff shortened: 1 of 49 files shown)


lrineau:
1. This PR triggered [warnings] in CGAL-4.12-Ic-24 on Windows.
2. The Linux Fedora 26 distribution has been updated by [an update] rather fast.

  [warnings]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-23.shtml
  [an update]: https://bodhi.fedoraproject.org/updates/FEDORA-2017-781951efb3

That is why I prefer to withdraw the pull-request instead of completing it to remove the Windows warnings.

PR #2249, opened 2017-07-11
Read PLY polygon soup

sgiraudot:
## Summary of Changes

This is a first step for solving the issue https://github.com/CGAL/cgal/issues/2196:
* it adds some (internal) needed methods/functions in the existing PLY headers (in point set processing)
* it adds functions to read/write PLY polygon soups (and to write face graphs)
* it adds mesh polygon soup support to the PLY plugin

So far, this PR only modifies internal code and demo code. To fully solve the issue https://github.com/CGAL/cgal/issues/2196, there should also be a small feature to document `read_OFF()` and these new PLY functions.

## Release Management

* Affected package(s): Polyhedron demo, Point Set Processing, Polyhedron IO (but no API change and no new documented function)
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2196
* Dependences: https://github.com/CGAL/cgal/pull/2138 (should not be merged before this one)

PR #2370, opened 2017-08-24
PLY reader: Fix face properties bug

sgiraudot:
## Summary of Changes

Properties in PLY input were badly handled: all simple properties were assigned to points whereas face can have simple properties too (other properties than just the list of vertices). As a result, loading a PLY with colored _faces_ in the demo was leading to a crash. This fixes it.

## Release Management

* Affected package(s): Point_set_processing, Polyhedron_IO

### DIFF
--- a/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
+++ b/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
@@ -41,19 +41,19 @@
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


lrineau:
Which release should be triggered? 4.10.1, 4.11, 4.12?

sgiraudot:
It fixes something introduce in PR https://github.com/CGAL/cgal/pull/2249 (merged 2 weeks ago), so I guess 4.12?

lrineau:
So 4.12 indeed. Thanks.

PR #2394, opened 2017-09-06
Parameterization:  Remove VC++ compiler option; Remove a macro concerning Eigen

afabri:
## Summary of Changes

The compiler option `/FR` leads to an [ICE in the testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-34/Surface_mesh_parameterization/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz) for just `#include <Eigen/Sparse>`.

I at the same time removed the line `#define EIGEN_YES_I_KNOW_SPARSE_MODULE_IS_NOT_STABLE_YET`, as according to Gael Guennebaud this package is stable now. 

## Release Management

* Affected package(s): Surface_mesh_parameterization

### DIFF
--- a/Solver_interface/include/CGAL/Eigen_matrix.h
+++ b/Solver_interface/include/CGAL/Eigen_matrix.h
@@ -21,7 +21,6 @@
 
 #include <CGAL/basic.h> // include basic.h before testing #defines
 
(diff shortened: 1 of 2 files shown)


lrineau:
Why only for `master`? It could be for 4.10.x and 4.11.x as well.

PR #2409, opened 2017-09-09
Shape Detection: avoid index out of range with large bitmap sizes

### LINKED ISSUE #2408: crash in Shape Detection when bitmap size exceeds max int
## Issue Details

Using shape detection to detect planes I encountered a seg fault caused by incorrect indexing into the bitmap vector in the `Shape_base::connected_component()` method.

## Source Code

I don't have a small repro case.  I ran into this with a rather large set of points.

## Environment

CGAL 4.10 on mac and linux.

caedmon:
With large u and v, the product u * v exceeds the capacity of an int.  The result when coerced to size_t can be an out of range index into bitmap.
Avoid this by casting u and v to size_t before multiplying.

## Release Management

* Affected package(s): Point set shape detection
* Issue(s) solved (if any): fix #2408

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Shape_base.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Shape_base.h
@@ -168,7 +168,7 @@ namespace CGAL {
         u = (u < 0) ? 0 : (((std::size_t)u >= u_extent) ? (int)u_extent - 1 : u);
         v = (v < 0) ? 0 : (((std::size_t)v >= v_extent) ? (int)v_extent - 1 : v);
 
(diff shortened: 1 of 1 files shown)


sloriot:
@sgiraudot there are so many cast in the code around into `int` and `size_t`. Do you see a technical reason not to switch to `std::size_t` everywhere (in particular for `u` and `v`)?

lrineau:
TODO: apply first the PR on `master`, and then cherry-pick the commit in the release branch for 4.11.

lrineau:
(That was a "TODO" for myself.)

@caedmon If you have time to rebase your branch onto `releases/CGAL-4.11-branch` that would be easier for us.

PR #2419, opened 2017-09-12
Polyhedron demo: c3t3_item, fix the color of balls of corners

lrineau:


lrineau:
@sloriot: That is the bug you reported to me, about the color of some of the protecting balls.

PR #2422, opened 2017-09-13
Polyhedron demo: reset points order before saving

sgiraudot:
## Summary of Changes

In order to handle the quick display of point sets (only displaying a random fraction of the points when moving the camera), indices of points are shuffled. But as result, when saving a point set, even unaltered, the points have a different order than in the input file, which can be disturbing for users.

This PR resets the indices before saving.

## Release Management


lrineau:
Then, once you have saved a point set item, are the indices shuffled again? Otherwise the display may no longer be correct.

sgiraudot:
The shuffle is done everytime the OpenGL buffers are filled, so yes.

PR #2426, opened 2017-09-14
Polyhedron demo: Fix for Color Connected Components in polyhedron mode

### LINKED ISSUE #2421: Color connected components segfaults in the demo
Take the elephant and the sphere, join them and then try to color each cc.
We should check whether the bug is in the 4.11 branch also.
I tried only with the polyhedron item, no idea for the surface mesh item.

my HEAD is 6f90d5d4f0601266ed020cec5874cc2289f6a271

maxGimeno:
## Summary of Changes

Fixes multicolor rendering in Polyhedron_item.

## Release Management

* Issue(s) solved (if any): fix #2421


lrineau:
@sloriot said, in https://github.com/CGAL/cgal/issues/2421#issue-257280120:
> We should check whether the bug is in the 4.11 branch also.

@maxGimeno Can I assume you have checked and found out that the bug in only in `master`, and not in `releases/CGAL-4.11-branch`?

maxGimeno:
Yes, I checked. it is not in 4.11.

lrineau:
Good. Thanks.

Issue #2410, opened 2017-09-10
crash in Point_set_2::dfs due to stack overflow


mattanf2:
## Issue Details

Gdal crashes while calling (probably) `CGAL::Point_set_2<SurfaceTriangulationTraits>::range_search()`.
Actual crash is inside the dfs function in `Point_set_2.h` file. It is caused because of a stack overflow (The function calls itself too many times)

## Source Code

I don't know how to recreate the problem without a lot of code but fixing it seems simple enough. remove the recursion. you can use the following code:

``` c++
void dfs(Vertex_handle v, const Circle& C, std::list<Vertex_handle>& L)
{
	struct DfsStackItem {
		Vertex_circulator vc;
		Vertex_circulator start;
	}; 
	
	Vertex_handle act = v;
	std::list<DfsStackItem> stack;
	bool pushToStack = true;
	do
	{
		DfsStackItem* top = nullptr;
		bool hadPush = pushToStack;
		if (pushToStack)
		{
			stack.push_back(DfsStackItem());
			top = &(stack.back());
			mark_vertex(act);
			top->vc = incident_vertices(act);
			top->start = top->vc;
			pushToStack = false;
		}
		else
		{
			top = &(stack.back());
			++(top->vc);
		}
		if ((hadPush) || (top->vc != top->start))
		{
			act = top->vc;
			if (!is_infinite(act)) {
				if (!is_marked(act) && !(tr_circleptori(C, act->point()) == ON_UNBOUNDED_SIDE))
					pushToStack = true;
			}
		}
		else stack.pop_back();
	
	} while (!stack.empty());
}
```

## Environment

* Operating system Windows
* Compiler: Visual studio 2015
* Release or debug mode: Release
* Specific flags used (if any):
* CGAL version: produced on 4.0.2. However the bug seams relevant for the current 4.0.10 release.

sgiraudot:
Thanks for the report, recursion should indeed be avoided. I'll fix it.

lrineau:
Should be fixed by #2427 in the release branches for 4.10.2 and 4.11.

mattanf2:
I forgot the most important line in the whole function. The function doesn't do anything right now.
inside the section of if (pushToStack) add the following line

		if (pushToStack)
		{
			stack.push_back(DfsStackItem());
			top = &(stack.back());
			mark_vertex(act);
			top->vc = incident_vertices(act);
			top->start = top->vc;
			pushToStack = false;

[...]

PR #2427, opened 2017-09-14
Point set 2: make method dfs non-recursive

sgiraudot:
## Summary of Changes

One method of `Point_set_2` is recursive and may lead to stack overflow. Simply replacing it with an `std::stack` (with an additional check when destacking) solves the issue. I tested it locally.

## Release Management

* Affected package(s): Point Set 2
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2410

### DIFF
--- a/Point_set_2/include/CGAL/Point_set_2.h
+++ b/Point_set_2/include/CGAL/Point_set_2.h
@@ -326,25 +326,35 @@ public:
   
   void dfs(Vertex_handle v,const Circle& C, std::list<Vertex_handle>& L)
   {
-    L.push_back(v);
-    mark_vertex(v);
-    
-    // get incident vertices of v ...
-    Vertex_circulator vc = incident_vertices(v);
-    Vertex_circulator start =vc;
-     
-    Vertex_handle act;
-     
-    // go through the vertices ...
-    do {
-      act = vc;
- 
-       if (! is_infinite(act)) {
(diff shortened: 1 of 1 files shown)


lrineau:
Why only for `master`? #2410 was reported against CGAL-4.2 and CGAL-4.10.

lrineau:
TODO: rebase only CGAL-4.10-branch

sgiraudot:
I rebased it and also changed the names of the method. They are internal so it does not really matter, but as @sloriot pointed out, it's no longer a Depth-First Search (DFS). The fact that it was a DFS was not documented and the search functions (`nearest_neighbors()` and `range_search()`) that use it are still working properly.

lrineau:
@maxGimeno Can you please try to understand, and fix, why this PR has a problem with Qt5 in Travis?

https://travis-ci.org/CGAL/cgal/jobs/275391478

It is only two commits ahead of `cgal/releases/CGAL-4.10-branch`, and that one has [perfect Travis results][*].

[*]: https://travis-ci.org/CGAL/cgal/builds/273347976

maxGimeno:
The only file changed has nothing to do with Qt5 config, so I guess either our sources path is not valid anymore, and in this case every Travis run will fail from now on, or a network error occured, keeping travis from getting qt5 for this run, and it is nothing. I restarted the job to see if it works better.

PR #2428, opened 2017-09-14
Polyhedron demo: Fix shrinking in c3t3

### LINKED ISSUE #2418: Default shrink factor of a c3t3 in the demo
When loading a c3t3 in the demo, the facets are already shrinked a bit. Is it a bug or is it expected?

maxGimeno:
## Summary of Changes

Set the shrinking cursor value after setting its maximum.

## Release Management
* Issue(s) solved (if any): fix #2418

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
@@ -257,9 +257,9 @@ struct Scene_c3t3_item_priv {
   {
     init_default_values();
     tet_Slider = new QSlider(Qt::Horizontal);
-    tet_Slider->setValue(100);
     tet_Slider->setMinimum(0);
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -273,9 +273,9 @@ struct Scene_c3t3_item_priv {
   {
     init_default_values();
     tet_Slider = new QSlider(Qt::Horizontal);
-    tet_Slider->setValue(100);
     tet_Slider->setMinimum(0);
     tet_Slider->setMaximum(100);
+    tet_Slider->setValue(100);
```
lrineau: Whoah, that one was a nasty bug! Previously the `setValue(100)` was called before `setMaximum(100)`, and the default maximum is... 99, as far as I know!

PR #2429, opened 2017-09-14
Polyhedron demo: Fix polyhedron orientation errors

### LINKED ISSUE #2423: Two-side drawing of a tetrahedron
The color used for one of the face (seen) from the outside is actually the color when seen from the inside. The problem goes away if I add a center vertex in that face (adding a center vertex in another face does not solve the pb). The tetrahedron I used is the default one generated by basic generators.

The bug is observed in master and we should check whether it is also in 4.11

maxGimeno:
## Summary of Changes

Fix orientation of facets in modern flat

## Release Management

* Issue(s) solved (if any): fix #2423

### DIFF
--- a/Polyhedron/demo/Polyhedron/resources/shader_flat.g
+++ b/Polyhedron/demo/Polyhedron/resources/shader_flat.g
@@ -6,7 +6,6 @@ layout (triangle_strip, max_vertices = 3) out;
 in VS_OUT
 {
   vec4 fP;
-  vec3 normal;
   vec4 out_color;
 } gs_in[3];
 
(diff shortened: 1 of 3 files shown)


maxGimeno:
@lrineau the bug is not in 4.11, it is related to the modern flat, which was not in 4.11

Polyhedron/demo/Polyhedron/resources/shader_flat.g
```diff
@@ -18,9 +17,13 @@ out GS_OUT
 } gs_out;
 
 uniform mat4 mv_matrix;
+uniform mat4 mvp_matrix;
 
 void main(void)
 {
+  vec4 norm1 = inverse(mvp_matrix)*gl_in[1].gl_Position - inverse(mvp_matrix)*gl_in[0].gl_Position;
+  vec4 norm2 = inverse(mvp_matrix)*gl_in[2].gl_Position - inverse(mvp_matrix)*gl_in[1].gl_Position;
```
lrineau: You should store the result of the expression `inverse(mvp_matrix)`.

lrineau: ping @maxGimeno

maxGimeno:
done

PR #2434, opened 2017-09-15
Polyhedron demo: New widget for color computation

maxGimeno:
## Summary of Changes
Adds a widget that allows to colorize a surface_mesh according to the smallest angle of each of its faces, or according to their scaled jacobian.
## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Color_ramp.cpp
+++ b/Polyhedron/demo/Polyhedron/Color_ramp.cpp
@@ -120,12 +120,19 @@ Color_ramp::build_thermal()
   g_.rebuild(1,0);
   b_.rebuild(1,0);
 
-  r_.add(0.3,1);
-  r_.add(0.05,1);
+  r_.add(0.0,0);
+  r_.add(1,1);
   g_.add(0.05,0.8);
   g_.add(0.3,0.5);
-  b_.add(0.05,0.6);
-  b_.add(0.05,0.3);
+  b_.add(0.0,1.0);
+  b_.add(1.0,0.0);
+}
+void Color_ramp::
+build_red_to_green()
+{
+  r_.rebuild(1,0);
+  g_.rebuild(0,1);
+  b_.rebuild(0,0);
 }
 
 void
--- a/Polyhedron/demo/Polyhedron/Color_ramp.cpp
+++ b/Polyhedron/demo/Polyhedron/Color_ramp.cpp
@@ -87,6 +87,19 @@ Color_ramp::Color_ramp()
 { 
 }
 
+Color_ramp::Color_ramp(const double r0, const double r1,
+           const double g0, const double g1,
+           const double b0, const double b1)
+{
+  r_.rebuild(r0,r1);
+  g_.rebuild(g0,g1);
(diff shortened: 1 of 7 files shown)


lrineau:
A Travis check, about `override`, has failed:
https://travis-ci.org/CGAL/cgal/jobs/275923790#L4127

janetournois:
Nice feature, thank you!

It works fine, but there is something strange in the behavior :
- load a polyhedral surface
- click "Display properties" : the scaled jacobian is chosen by default (maybe having the smallest angle as the default would be more intuitive btw)
- click "Colorize" : the values for min and max of the ramp change in the spin boxes, but not on the colored ramp. That's a bit weird. But then clicking on "replace ramp" changes the min and max of the ramp so it's ok.
- change the property to "smallest angle", and click "colorize": the ramp extrema change in the spinboxes, but the ramp does not change at all (not the color, nor the extrema). And this time, it looks more like a bug.

Imho, it is acceptable that the ramp does not change on-the-fly when the spinboxes values are changed, but that the same criterion is taken into account, until one clicks "Replace ramp".
However, when the criterion iss changed, the ramp should be updated when "Colorize" is clicked.

maxGimeno:
* click "Colorize" : the values for min and max of the ramp change in the spin boxes, but not on the colored ramp. That's a bit weird. But then clicking on "replace ramp" changes the min and max of the ramp so it's ok.
-> It is wanted : having the values in the min and max boxes scaled on the extrema of the surface allows to get its values, which gives an idea for the values we want for the ramp, but we don't want the ramp to change everytime we colorize a mesh. The upside is that it makes the economy of a button, but if you think it is disturbing, I guess it is better to add a button anyway.

janetournois:
Thank you for the changes.

About the smallest angle scale :
I agree that, when loading a triangle mesh, the initial angle bounds should be `[0;60]`, but when the mesh is not triangular, the smallest angle has different bounds, and `[0;180]` would be a reasonable choice in case of non-convex polygons.
 
Note that the mesh can be hybrid (for example triangles and quads), so I would suggest that 
- if `CGAL::is_triangle_mesh()` returns true, the initial bounds are set to `[0;60]`
- if `CGAL::is_quad_mesh()` returns true, the initial bounds are set to `[0,90]` (even with non-convex polygons, it should be correct 
- otherwise, the inital bounds are set to `[0; 180]`

janetournois:
When "Replace Ramp" is clicked, is it on purpose that the faces are not re-colorized automatically?

maxGimeno:
Yes, you might just want to check if your new ramp looks like you want before re-colorizing all the faces. I guess.

lrineau:
[Warnings in the testsuite]:
```
test\Polyhedron_Demo\Plugins\Display\Display_property_plugin.cpp(172): warning C4305: 'initializing': truncation from 'double' to 'float'
test\Polyhedron_Demo\Plugins\Display\Display_property_plugin.cpp(173): warning C4305: 'initializing': truncation from 'double' to 'float'
```

[Warnings in the testsuite]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-45.shtml#Polyhedron_Demo

PR #2437, opened 2017-09-18
Polyhedron demo: Fix float conversion at buffer filling in SM item

### LINKED ISSUE #2433: Polyhedron demo: bad surface mesh display on large coordinates
## Issue Details

The Surface Mesh item in the demo has a bad display when using large coordinates (which makes it non-aligned with other objects that have the exact same coordinates).

## Source Code

Consider the following OFF file:

```
OFF
3 1 0
8000000 5000000 0.3
8000002.3 5000001.5 -0.11
8000001.1 5000003.92 0.999
3 0 1 2
```

You can view the problem with 2 simple methods:
* Open it once with the Polyhedron OFF reader and once with the Surface Mesh OFF reader: the triangle is not the same
* Or open it with the Surface Mesh OFF reader, select all edges: the selection item is not aligned with the edges of the surface mesh (same if you create a polyline)

Visually, Polylines, Point Sets and Polyhedron seem to agree on the coordinates, so my guess is that the problem comes from the Surface Mesh item.

maxGimeno:
## Summary of Changes

Add missing parenthesis at offset addition.
## Release Management

* Issue(s) solved (if any): fix #2433

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
@@ -297,9 +297,9 @@ void Scene_surface_mesh_item_priv::addFlatData(Point p, EPICK::Vector_3 n, CGAL:
 {
   const qglviewer::Vec offset = static_cast<CGAL::Three::Viewer_interface*>(QGLViewer::QGLViewerPool().first())->offset();
 
-  flat_vertices.push_back((cgal_gl_data)p.x()+offset[0]);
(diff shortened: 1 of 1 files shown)


lrineau:
Is this only for 4.12, or was the bug already in 4.11?

maxGimeno:
Oops, it must be in 4.11 too. I'll take care of this

maxGimeno:
done

lrineau:
I confirm that this trivial PR fixes #2433.

PR #2442, opened 2017-09-19
Polyhedron demo: Fix the Clipping Box Plugin

### LINKED ISSUE #2440: Clipping box item
I don't know what the clipping box should be doing but with bac04522 when I click on `Clip` after the generation of the box with `elephant` selected, edges outside the box are no longer displayed but faces are. Also the `clip` button stays pressed. 

No idea if the issues is also in 4.11

maxGimeno:
## Summary of Changes

Add clipbox management in flat_shader.

## Release Management

* Issue(s) solved (if any): fix #2440

### DIFF
--- a/Polyhedron/demo/Polyhedron/resources/shader_flat.f
+++ b/Polyhedron/demo/Polyhedron/resources/shader_flat.f
@@ -5,6 +5,7 @@ in GS_OUT
   vec4 fP;
   flat vec3 normal;
   flat vec4 color;
+  float dist[6];
 } fs_in;
 
 uniform bool is_two_side;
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/resources/shader_flat.f
```diff
@@ -13,11 +14,20 @@
...
 void main(void)
 {
+  if(is_clipbox_on)
+    if(fs_in.dist[0]>0 ||
+       fs_in.dist[1]>0 ||
+       fs_in.dist[2]>0 ||
+       fs_in.dist[3]>0 ||
+       fs_in.dist[4]>0 ||
+       fs_in.dist[5]>0)
+      discard;
```
lrineau: You should bench that. Adding a `if` in a shader might penalize the performance of the shader, even for items when the `is_clipbox_on` is always false. If the performance is impacted, you should use two different shaders.

maxGimeno: That is easier said than done. That would mean that all impacted items should test for the presence of an active clipping box  in the scene, qhich they cannot do because they do not have a reference to the scene, and choose their program accordingly. So just for that little plugin we need to add a boolean to the Scene_facegraph_items and 2 programs to the list in the viewer. 
I agree that it is probably impacting the rendering, but the alternative does not satisfy me either.

maxGimeno: I benched it, it doesn't change a thing.

PR #2443, opened 2017-09-19
Fix determinism of points inside 2D/3D triangles, and tetrahedron

sloriot:


sloriot:
Like my commit says I removed a default value so that the default copy-constructor is not hidden and replace by a constructor using the defaut random number generator.

MaelRL:
From a symptomatic point of view, this meant that the seed was usually re-initialized in between calls such as `*generator++`. So, it would produce a different output every time, even if you had passed a seed to `CGAL::Random`.

PR #2444, opened 2017-09-19
Polyhedron demo: bug fix for writing binary PLY

afabri:
## Summary of Changes

For VC++ the stream must be opened in binary mode.

## Release Management


lrineau:
Again, if that is a bug-fix, then the branch is probably not `master`, but one of the release branches, for 4.10.2, or 4.11.1.

PR #2373, opened 2017-08-25
Classification (new package)

sgiraudot:
## Summary of Changes

This is a new package. Not yet approved [full feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Classification).

## Release Management

* Affected package(s): Classification (new package), Polyhedron demo
* Feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Classification

PR #2446, opened 2017-09-19
PSP:  make read_ply_points faster for VC++ < 2017

afabri:
## Summary of Changes

VC++ < 2017 has a slow   istream >> double

## Release Management

* Affected package(s): Point set processing

### DIFF
--- a/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
+++ b/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
@@ -165,6 +165,11 @@ namespace internal {
       c = static_cast<unsigned char>(s);
     }
 
+    void read_ascii(std::istream& stream, double& d) const
+    {
+      stream >> iformat(d);
+    }
+
     // Default template when Type is not a char type
     template <typename Type>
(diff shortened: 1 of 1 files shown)


Point_set_processing_3/include/CGAL/IO/read_ply_points.h
```diff
@@ -165,6 +165,11 @@ namespace internal {
       c = static_cast<unsigned char>(s);
     }
 
+    void read_ascii(std::istream& stream, double& d) const
+    {
+      stream >> iformat(d);
+    }
+
```
lrineau: This change is already present in the PR #2373, that is about to be merged in `master`, see:
https://github.com/CGAL/cgal/pull/2373/files#diff-aa78664827be8dc329359cf1b4cb0080R174

Unless you want to apply this PR to previous versions of CGAL, such as CGAL-4.11, I suggest we cancel this PR #2446.

afabri: You can cancel it.

PR #2350, opened 2017-08-16
Fix the link computation for edges on the border

sloriot:
The condition for adding vL and vR was incorrect

## Release Management

* Affected package(s): Surface Mesh Simplification

PR #2450, opened 2017-09-21
Fix VC performance warning

afabri:
## Summary of Changes

Fix for VC  `warning C4800: 'void *' : forcing value to bool 'true' or 'false'`

## Release Management

* Affected package(s): Surface_mesh_simplification
* Issue(s) solved (if any): fix #2350


Surface_mesh_simplification/test/Surface_mesh_simplification/test_edge_profile_link.cpp
```diff
@@ -65,7 +65,7 @@ void test(const char* fname)
 {
   Mesh m;
   std::ifstream input(fname);
-  assert( bool(input) );
+  assert( ! input.bad()) );
```
lrineau: Syntax error: there is an extra `)`.

PR #2454, opened 2017-09-21
Polyhedron demo: Fix tets rendering on Windows

### LINKED ISSUE #2453: Polyhedron demo : bugs in the `mesh_3_plugin`
## Issue Details

I am trying to mesh in the demo a given `Polyhedron` (the problem is the same with `Surface_mesh` with a very small tetrahedron size, and the output mesh is always more or less the same, with very few simplices compared to what I expect.

Then, the `C3t3` is displayed with the cut plane, and when I click "Show tetrahedra", it is only possible to have the intersected tetrahedra in "wire" mode. The "flat+edges" mode displays the same as the "wire" mode.

## Environment

* Compiler: msvc2013
* Specific flags used (if any): CGAL_HEADER_ONLY
* CGAL version: master
* Boost version: 1.59

maxGimeno:
## Summary of Changes
Adds a clipbox management in the c3t3_item's intersection to have coherent shaders.
## Release Management
* Issue(s) solved (if any): partly fixes #2453


Polyhedron/demo/Polyhedron/resources/shader_c3t3_tets.v
```diff
@@ -26,5 +47,6 @@ void main(void)
     0, shrink_factor, 0, 0,
     0, 0, shrink_factor, 0,
     0, 0, 0, 1);
+    void compute_distances();
```
lrineau: The `void` keyword should problably be removed from this line. Does it compile?!

lrineau:
Bug for 4.11, or `master`/4.12?

maxGimeno:
This bug is only in 4.12.

janetournois:
My bad, this was fixed by the last commit.
Now it would be nice to have the "inside" tets in a different color than the surface of the mesh.

PR #2461, opened 2017-09-22
Use CGAL::cpp11::is_enum

sloriot:
To work around a bug observed with boost 1.62 with c++11 ON.

Fixes [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-46/Surface_mesher/TestReport_lrineau_Debian-Testing.gz)


sloriot:
The [error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-47/Surface_mesher/TestReport_lrineau_Debian-Testing.gz) is gone.

PR #2467, opened 2017-09-26
Mesh 2: fix clusters (issue #2456)

### LINKED ISSUE #2456: Mesh_2 segfault
I have an old file that is making mesh_2 crashing in ipe but also in the mesh_2 demo (IIRC).
For the record I'm putting the files here. It might be a duplicate of #980
[mesh_2_br.zip](https://github.com/CGAL/cgal/files/1323914/mesh_2_br.zip)

lrineau:
## Summary of Changes

When two clusters share a subsegment, the subsegment is split at its
middle, instead of using the length imposed by the clusters. If the
clusters were partially refined before that subsegment is split, one
must reset the clusters, because the minimal length may have changed.

This PR:
  - fixes the bug,
  - add more debug output, in conditionals `CGAL_MESH_2_DEBUG_CLUSTERS`.

## Release Management

* Affected package(s): Mesh_2
* Issue(s) solved (if any): fix #2456 

### DIFF
--- a/Mesh_2/include/CGAL/Mesh_2/Clusters.h
+++ b/Mesh_2/include/CGAL/Mesh_2/Clusters.h
@@ -325,6 +325,16 @@ update_cluster(Cluster& c, iterator it, Vertex_handle va,
   c.vertices.erase(vb);
   c.vertices[vm] = reduction;
 
+  if(false == reduction) {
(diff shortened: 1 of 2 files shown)


lrineau:
Note that the bug has been there since the very beginning of Mesh_2.

PR #2471, opened 2017-09-27
Include OpenGL ES for arm

jubalh:
## Summary of Changes

If we compile CGAL with having Qt present on arm, Qt will use OpenGL ES
headers. But CGAL/gl.h includes specifically include <GL/gl.h> which are
the desktop headers. They cannot be mixed.

However I am not sure whether there should be a differentiation between OpenGL ES 2 and 3 though. I just used 3 since that what I have here. Please give feedback in case you want me to change something.

## Release Management

* Affected package(s): cgal 4.11
* Feature/Small Feature (if any): let compile on arm(v7) with Qt present


lrineau:
Thanks for your patch. It is now integrated in the `master` branch of CGAL, targetting CGAL-4.12 (due Q1 2018).

Issue #2468, opened 2017-09-26
New warning about counting_iterator


lrineau:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.11.1-I-91/Polytope_distance_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

The warning appeared suddenly between CGAL-4.11-I-90 and CGAL-4.11.1-I-91, that are very similar. I do not understand.

PR #2475, opened 2017-09-29
Fix counting iterator warnings

sgiraudot:
## Summary of Changes

Fix for https://github.com/CGAL/cgal/issues/2468. As I was getting tired of this warning, I wrote a file `CGAL/counting_iterator.h` that just does the inclusion surrounded by the warning deactivation, but then I realized that it already existed (for the very same reason, I guess), see in `Installation/include/CGAL/boost/iterator/counting_iterator.h`. It was already used by some CGAL headers.

So this PR replaces all occurrences of the boost inclusion in CGAL files by this one, which should fix the problem everywhere and prevent it from happening again. We should always use it from now on.

## Release Management

* Affected package(s):
  - Arrangement on Surface 2
  - Convex hull 2
  - NewKernel D
  - Point Set Shape Detection 3
  - Polyhedron demo
  - Segment Delaunay Graph 2 (+ Linf 2)
  - Spatial Searching
  - Spatial Sorting
  - Triangulation 2
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2468 (and more...)

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/graph_traits_Arrangement_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/graph_traits_Arrangement_2.h
@@ -33,7 +33,7 @@
 #include <CGAL/boost/graph/named_function_params.h>
 
 #include <boost/graph/graph_concepts.hpp>
-#include <boost/iterator/counting_iterator.hpp>
+#include <CGAL/boost/iterator/counting_iterator.hpp>
 #include <CGAL/Arrangement_on_surface_2.h>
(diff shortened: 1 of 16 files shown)


sgiraudot:
(I just realized that the incriminating package Polytope Distance D is not in the list. The problem might be that another header of Boost already includes the counting iterator… that's going to require more work to fix it, probably.)

lrineau:
Please rebase for `releases/4.11-branch`.

sgiraudot:
Done.

PR #2477, opened 2017-10-02
Polyhedron demo: several fixes and enhancements for PLY I/O

sgiraudot:
## Summary of Changes

Several fixes and enhancements:

* To decide to read a PLY input whether as a point set or as a polygon soup, we check if `element face` is present in the header. In fact, we should also check that it is not `element face 0`, which some software use when they save PLY point sets (in our case, we load an empty polygon soup, we can then recover the points but it's not intuitive at all).
* Test `!stream.bad()` instead of `stream.good()` to handle cases with missing carriage return at the end of an input file
* Handle another commonly used tag in the header (vertex_index instead of vertex_indices)
* Add a way to read and write comments in the PLY header (undocumented so far but that can be documented and generalized later on if we find interest in it)
* Save/load label indices and names in classification plugin (using PLY comments)

## Release Management

* Affected package(s): Polyhedron demo, Point Set 3, Polyhedron IO


sgiraudot:
(Please don't merge this branch yet, I still have some additions to make which will have to be tested.)

sgiraudot:
Okay, the branch can be tested again. For information, I added:

* Another fix to handle another commonly used tag in the header (`vertex_index` instead of `vertex_indices`)
* A way to read and write comments in the PLY header (undocumented so far but that can be documented and generalized later on if we find interest in it)

I also added a mechanism in the classification plugin to read and write label informations using the PLY comments. It's a bit out of scope of the original PR, but as it's closely related to the comment handling in PLY, I thought it made sense to include it rather than add yet another PR after this one… Let me know if this is a problem, I can easily separate the commits if needed.

lrineau:
Simon, remove the `TODO` label from this PR once you consider it is finished.

sgiraudot:
Oh yes, I forgot. It's done.

PR #2478, opened 2017-10-02
fix bug -> 2x 2 missing "inline" to prevent duplicate symbols during …

mtola:
… link process.

---> cgal 4.11

* Affected package(s):

* BGL
* Linear_cell_complex


mtola:
NB : with last cgal 4.11 (release)

lrineau:
Thanks for your patch. It is now integrated in the branch targetting CGAL-4.11.1, and in the main `master` branch targetting CGAL-4.12.

Issue #2483, opened 2017-10-02
Region_growing can duplicate points when assigning to shapes.


caedmon:
## Issue Details

When using Region_growing to detect shapes in a point cloud, it can return the same point twice in the list of assigned points.

## Source Code

  Region_growing detector;
  detector.set_input(points);
  detector.add_shape_factory<Plane>();

  Region_growing::Parameters parameters;
  // Minimum number of points in a detected shape
  parameters.min_points = options.minPoints;
  // Sets maximum Euclidean distance between a point and a shape.
  parameters.epsilon = options.epsilon;
  // Sets maximum Euclidean distance between points to be clustered.
  parameters.cluster_epsilon = options.clusterEpsilon;
  // Sets maximum normal deviation, dot(surface_normal, point_normal)
  parameters.normal_threshold = options.normalThreshold;

  detector.detect(parameters);

## Environment

* Operating system OSX, Ubuntu
* Compiler: clang, g++
* Release or debug mode: release
* Specific flags used (if any):
* CGAL version: 4.12
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

PR #2484, opened 2017-10-02
don’t assign same point to shape twice

caedmon:
## Summary of Changes

Region growing sometimes includes a point more than once in the list of points assigned to a shape.  This change may not be complete or the best way to fix this, but it did solve the problem in my situation.

## Release Management

Possible fix for #2483


sgiraudot:
Actually, the duplicate index problem that you found only happens for the first detected shape, and it's because of a badly initialized variable. I created a dedicated fix, but thanks for the report!

caedmon:
Yes, that looks like a better fix.  Thanks!

PR #2494, opened 2017-10-04
Polyhedron_demo: Display_plugin can zoom to extrema

### LINKED ISSUE #2507: Polyhedron demo: Improvements for Display Properties
## Issue Details

The widget is cool to have!! 
But it takes too much screen space, and it is in particular too wide. 
 
capture

- No need for the grey area to the left of the ramp  (A)
- 1/3 of the width of the color  ramp is enough  (A)
- Replace (B) with a color chooser.
- Write Ramp Extrema on top of the spin box

Also when the widget stretches vertically, it is ugly that the ramp and the text are not horizontally aligned. 
All elements in the widget should be at the top and a stretcher (if that is the right name) at the bottom.

@janetournois  I think you had some additional remarks.

maxGimeno:
## Summary of Changes
Add buttons for zooming to extrema faces in the Display_properties_plugin.

fixes #2507

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Display/Display_property.ui
+++ b/Polyhedron/demo/Polyhedron/Plugins/Display/Display_property.ui
@@ -6,15 +6,28 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>504</width>
-    <height>369</height>
+    <width>292</width>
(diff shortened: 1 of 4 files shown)


afabri:
It works on Windows and is a great feature!!!

PR #2496, opened 2017-10-04
Replace std::unary_function

### LINKED ISSUE #2415: Add a test of `/std:c++latest` with  Visual C++ 2017.
Newest versions of VC++2017 have a new flags `/std:c++latest`, to enable the bleeing-edge support of C++ (C++14, C++17, and later). I think one of our testsuite configuration should test that flag.

https://docs.microsoft.com/en-us/cpp/what-s-new-for-visual-cpp-in-visual-studio

See also [the documentation of the `/std` option](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version).

afabri:
## Summary of Changes

As `std::unary_function` got deprecated and as there are compiler options that make that it no longer there for backward compatibility, we add `CGAL::unary_function` to CGAL. It is mainly an implementation detail. For those where it is documented we have  to do more.

@lrineau can you please replace all occurences of `std::unary_function` with `CGAL::unary_function` with perl, awk, or whatsoever. thank you. 
 
## Release Management

* Affected package(s): lots
* Issue(s) solved (if any): fix #2415


lrineau:
There should be also a replacement for `std::binary_function`, that is removed from C++17 as well (deprecated since C++11).

lrineau:
I have documented `CGAL::unary_function`, in https://github.com/CGAL/cgal/pull/2496/commits/23f954af447168059296f7ac71c064174e196ef6, becaus e it is mentioned by other documentation pages.

lrineau:
This PR should be based on CGAL-4.11. We will rebase it at the end.

afabri:
I just added  `binary_function`   in  `include/CGAL/unary_function.h`   We might rename it to ` functional.h`.

lrineau:
I have renamed the file, and replaced the occurrences of `std::binary_function`.

PR #2499, opened 2017-10-04
Mesh_3: the "maximal number of vertices" feature

lrineau:
## Summary of Changes

This PR imports from the old branch `Mesh_3-experimental-GF` the feature that allows to set a maximal number of vertices. That maximal number will end-up the refinement process forcibly, and an error code is set.

## Release Management

* Affected package(s): Mesh-3
* Feature/Small Feature (if any): none... that is not yet documented.

Question for @cjamin: for the moment it only works if the `Parallel_tag` is not used. Do you see a way to implement something similar in the parallel mode?

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
+++ b/Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
@@ -32,6 +32,8 @@
 
 #include <CGAL/Mesh_3/config.h>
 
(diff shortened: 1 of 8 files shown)


cjamin:
In parallel mode, it would be tricky to match the exact maximal number of vertices, but we could probably find a way to limit the number of inserted vertices to a reasonnably close value (depends on the number of threads)... We can discuss it during the CGAL meeting if you want.

PR #2509, opened 2017-10-06
Add a plugin for inset and offset of a polyhedron

sloriot:
This is a minimal plugin working with Surface_mesh and the polyhedron item that can generate naive inset and offset meshes.


lrineau:
There is a warning in the polyhedron demo:
```
Plugins/Surface_mesh/Offset_meshing_plugin.cpp:77:52: warning: unused parameter 'parent' [-Wunused-parameter]
```

lrineau:
> This is a minimal plugin working with Surface_mesh and the polyhedron item that can generate naive inset and offset meshes.

As far as I can see, only Polyhedron_3 can be used as input, and not Surface_mesh.

sloriot:
Yep, to use Surface_mesh it should need to be ported to `Mesh_3`. @maxGimeno Are you interested?

maxGimeno:
I'll look into it.

sloriot:
Actually, a `Surface_mesh` can be handled that's just a matter of handling the output.

PR #2514, opened 2017-10-09
Polyhedron demo - fix isotropic remeshing plugin

janetournois:
## Summary of Changes

In the `isotropic_remeshing_plugin`, the `Patch_id_map` that `get(..., pmesh)` returns was given as a parameter to `PMP::isotropic_remeshing()`, however it had been initialized or not.

The consequence was that `PMP::isotropic_remeshing()` "thought" it was using a valid `face_patch_map` and hence failed to make valid collapisibility/topology checks next to constrained edges.
The proper behavior when the face patch map is not provided (or invalid) is to compute the connected components from constrained edges. This is automatically done inside the remeshing function.

Now we use the face_patch_map only when it is valid, i.e. with not all its values `== 1` (because `1` is the default value both for `Polyhedron_3` and `Surface_mesh`.

## Release Management

* Affected packages : Polyhedron demo (and minor changes in PMP)

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -617,6 +617,10 @@ namespace internal {
       std::cout << "Collapse short edges (" << low << ", " << high << ")..."
                 << std::endl;
 #endif
+#ifdef CGAL_PMP_REMESHING_VERBOSE_PROGRESS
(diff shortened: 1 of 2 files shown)


lrineau:
This PR has a  conflict with #2519, and I will test it later.

lrineau:
Actually #2519 targets 4.10.2, and conflicts with `master`. I have asked Jane to merge #2519 into #2514, so solve conflicts.

janetournois:
the merge is done

lrineau:
The Travis build fails in the Polyhedron demo:
https://travis-ci.org/CGAL/cgal/jobs/288545504#L3896

PR #2518, opened 2017-10-10
Fix license checking for some lgpl headers

sloriot:
Some headers were LGPL while being in some GPL packages. The license checking mechanism would have reported wrong positives. The simplest fix was to make those headers GPL.

I used `ack -l "Lesser General" | xargs grep -l "CGAL/license"` to detect them.

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_algebraic_segment_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_algebraic_segment_traits_2.h
@@ -1,10 +1,10 @@
 // Copyright (c) 2006,2007,2008,2009,2010,2011 Max-Planck-Institute Saarbruecken (Germany).
 // All rights reserved.
 //
-// This file is part of CGAL (www.cgal.org); you can redistribute it and/or
-// modify it under the terms of the GNU Lesser General Public License as
-// published by the Free Software Foundation; either version 3 of the License,
-// or (at your option) any later version.
+// This file is part of CGAL (www.cgal.org).
+// You can redistribute it and/or modify it under the terms of the GNU
+// General Public License as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
 //
 // Licensees holding a valid commercial license may use this file in
 // accordance with the commercial license agreement provided with the software.
--- a/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
@@ -1,10 +1,10 @@
 // Copyright (c) 2007,2008,2009,2010,2011 Max-Planck-Institute Saarbruecken (Germany),
 // and Tel-Aviv University (Israel).  All rights reserved.
 //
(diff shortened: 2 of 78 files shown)


mglisse:
I don't like this automatic change. CGAL/config.h really? Besides, some basic functionality ended up in one GPL package or another almost by accident.

sloriot:
`CGAL/config.h` was a false positive. I don't see any simpler fix. Is there any particular file that you'd like to stay LGPL?

mglisse:
> I don't see any simpler fix.

Change the checker to allow files with a less restrictive license than their package?

> Is there any particular file that you'd like to stay LGPL?

The list of files changed is so long... Randomly, things like CGAL/certified_*, some IO/*, Static_filters/*, Installation/* might be candidates to remain LGPL. (I haven't read any of those files)

mglisse:
In the list of *easy* solutions, we could also admit that the LGPL part of CGAL is almost useless by itself and move everything to GPL.
Anyway, as long as you ask package maintainers first, I guess you can make those changes if you want.

afabri:
The LGPL part is NOT useless. It is the foundation layer on that the GPLed components are based.
The LGPLed parts are even a dependency for some other software projects with licenses such as BSD, Mozilla, so switching to GPL would be really serious.

If the Static_filters got relicensed from LGPL to GPL by the scriptt this is a mistake.

mglisse:
> The LGPL part is NOT useless. It is the foundation layer on that the GPLed components are based.

I didn't say it was useless, I said having it LGPL instead of GPL was useless, because you won't go far if you can't use triangulations, arrangements, etc.

> The LGPLed parts are even a dependency for some other software projects with licenses such as BSD, Mozilla, so switching to GPL would be really serious.

Really? That's interesting, I never heard about those, although the existence of such projects was always the reason to have some parts LGPL. Is there a list somewhere? I'd like to see what parts of CGAL they are using exactly.

sloriot:
I re-checked the files modified and these files are inside packages that are GPL so basically they are useless without the corresponding package (like IO for triangulation, skeleton, arrangement or `graph_traits`). The only exception are the files in `Polyhedron_IO` to read STL, OBJ, ... that should IMO stay LGPL but these files will be moved in another package soon when they'll get documented, and this package will be LGPL.

lrineau:
I do not understand which issue this PR is trying to fix. A GPL package can use LGPL/BSD/permissive files. And there is no issue either if a GPL package includes such files. The script `Scripts/developer_scripts/detect_packages_licenses` has a kind of logic to allow that.

@sloriot, what was the first issue you were trying to solve?

Anyway, we cannot change the license of a file without the permission from all the copyrights owners of the file. The only exception I know is about files under LGPLv2.1, that explicitly allows to relicense under GPL.

sloriot:
The issue is that some files are inside GPL packages but their header is LGPL but they include the license header of the GPL package which is a mistake. I though changing the license of those files was simpler than adding a special case in the checking script (as I don't see why these files are LGPL anyway).

mglisse:
> Anyway, we cannot change the license of a file without the permission from all the copyrights owners of the file. The only exception I know is about files under LGPLv2.1, that explicitly allows to relicense under GPL.

I thought it was also always fine to relicense from LGPL3 to GPL3. Didn't check though.

PR #2535, opened 2017-10-12
Surface_mesh: Added missing 'explicit' on SM_Edge_index constructor

MaelRL:
## Summary of Changes

Constructors of the index classes of `Surface_mesh` should be explicit (to avoid dangerous implicit conversions!).

## Release Management

* Affected package(s): `Surface_mesh`


MaelRL:
@sloriot @maxGimeno I'd also like to argue for a partial revert of 3bc4308abbc50ee2aa093a295158ab814fb74c96 to not remove the `explicit` keyword from the documentation.

PR #2538, opened 2017-10-13
Polyhedron demo: improve point set selection widget

sgiraudot:
## Summary of Changes

This PR reduces drastically the size of the point set selection widget and makes it easier to use (using similar icons – with tooltips – as selections in photo editors such as GIMP). Most buttons are moved to a dedicated menu.

Before:
Before

After:
After

## Release Management


sgiraudot:
(The Travis error is due to exceeding the limit of compilation time.)

PR #2550, opened 2017-10-23
Add SPDX License Identifier

sloriot:
Add license identifiers for all files having a license header
(Having kept the URL and Id identifiers would have made my life easier here...)

Question:
- Some test files and scripts have a license header.Do we want to keep them or shall we make remove the license headers to make it uniform?

TODO:
- [x] update test_merge_of_branch to check for the presence of the tag
- [x] use the SPDX tag to check for the presence of the license include directive


lrineau:
Are there tools to verify that the license identifiers actually correspond to the text of the license notice that is at the top of the files?

sloriot:
I'll check of course but note that for my last commit, I don't think it would have helped since it is written in the header that it is a copy of a boost header.

lrineau:
Travis error in license checking: https://travis-ci.org/CGAL/cgal/jobs/292517425#L2479

sloriot:
`test_merge_of_branch` has been updated to test for the presence of the identifiers. I checked that the identifier I put are correct using:
```
> find */include -name '*.*' | xargs egrep -l "SPDX-License-Identifier: LGPL-3\.0\+$" | xargs grep -L "Lesser General Public License"
Nothing returned
> find */include -name '*.*' | xargs egrep -l "SPDX-License-Identifier: GPL-3\.0\+$" | xargs grep -L "General Public License"
Nothing returned
> find */include -name '*.*' | xargs egrep -l "SPDX-License-Identifier: GPL-3\.0\+$" | xargs grep -l "Lesser"
Nothing returned
> find */include \( -name '*.h' -o -name '*.hpp' \) | xargs egrep -L "SPDX-License-Identifier: GPL-3\.0\+$" | xargs egrep -L "SPDX-License-Identifier: LGPL-3\.0\+$"
A few files are returned and I manually checked the id is correct (boost license for all of them but one that is GPL-2+)

```

Issue #2563, opened 2017-10-27
bug in Two_vertices_parameterizer_3.h


jiabinliang:
## Issue Details
runtime error
given a mesh with only one triangle 
-27.9902 8.96537 -36.5647
-27.9902 8.96613 -36.5647
-27.9901 8.96704 -36.5647
parameterize it with Square_border_arc_length_parameterizer_3 and LSCM_parameterizer
-> crash, using null pointer

bug is in:
Two_vertices_parameterizer_3.h:135-137
in 131-149 we suppose to find the bounding box of the mesh.
so  xmax should be initialize as -inf(std::numeric_limit<double>lowest),
but cgal is using numeric_limite<double>::min
which is the minimal positive value of double, epsilon. close to 0
so at the end xmax =0, (it like we extend bbox with 0,0,0)
then cgal chose Z then X as axis
as we all have the same Z, so all vertex will paramterized as u=0

then in line:230,232, again, using min instead of lowest
umax is epsilon now
we never enter if in line:255(u=0<epsilon)
vxmax is empty. crash in line 262

 

* Operating system (Windows/Mac/Linux, 32/64 bits):win64
* Compiler:MSVC
* Release or debug mode: release
* Specific flags used (if any):
* CGAL version:4.10, also in 4.11
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

lrineau:
@palliez: That is a bug in Surface_mesh_parameterization, and the reporter @jiabinliang has already found out the bug and the patch.

Note that [`std::numeric_limits<T>::lowest`] is C++11.
I suggest to use [`- std::numeric_limits<T>::infinity`] (negation of infinity) instead.

[`std::numeric_limits<T>::lowest`]: http://en.cppreference.com/w/cpp/types/numeric_limits/lowest
[`- std::numeric_limits<T>::infinity`]: http://en.cppreference.com/w/cpp/types/numeric_limits/infinity

afabri:
@MaelRL can you please fix this

MaelRL:
Can reproduce and changing the bound does fix it.

I haven't opened a PR yet because there are other places that seem to have the same issue, like:
* https://github.com/CGAL/cgal/blob/master/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h#L628
* https://github.com/CGAL/cgal/blob/master/Polygon_mesh_processing/examples/Polygon_mesh_processing/point_inside_example.cpp#L18

[...]

PR #2564, opened 2017-10-27
Fix wrong uses of std::numeric_limits::min() (4.10)

MaelRL:
## Summary of Changes

Various packages wrongly assumed that `min()` returns the lowest possible value.

## Release Management

* Affected package(s): `Point_set_shape_detection_3`, `PMP`, `SMP`
* Issue(s) solved (if any): Partial fix of #2563 

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h
@@ -622,12 +622,12 @@ namespace CGAL {
       }
         
       const Bbox_3 &buildBoundingCube() {
-        FT min[] = {(std::numeric_limits<FT>::max)(),
-                    (std::numeric_limits<FT>::max)(),
-                    (std::numeric_limits<FT>::max)()};
-        FT max[] = {(std::numeric_limits<FT>::min)(),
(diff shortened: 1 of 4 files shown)


Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h
```diff
@@ -622,12 +622,12 @@ namespace CGAL {
...
-                    (std::numeric_limits<FT>::max)()};
-        FT max[] = {(std::numeric_limits<FT>::min)(),
-                    (std::numeric_limits<FT>::min)(),
-                    (std::numeric_limits<FT>::min)()};
+        FT min[] = {std::numeric_limits<FT>::infinity(),
+                    std::numeric_limits<FT>::infinity(),
+                    std::numeric_limits<FT>::infinity()};
+        FT max[] = {-std::numeric_limits<FT>::infinity(),
+                    -std::numeric_limits<FT>::infinity(),
+                    -std::numeric_limits<FT>::infinity()};
```
lrineau: Actually `±max()` would have been a correct value too. By definition there does not exist any `FT` in the interval `]std::numeric_limits<FT>::max, std::numeric_limits<FT>::infinity()[`.

PR #2565, opened 2017-10-27
Fix wrong uses of std::numeric_limits::min() (4.11)

### LINKED ISSUE #2563: bug in Two_vertices_parameterizer_3.h

MaelRL:
## Summary of Changes

Second part of #2564, because `SMP` was restructured with CGAL-4.11.

## Release Management

* Affected package(s): `SMP`
* Issue(s) solved (if any): fixes  #2563.

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Octree.h
@@ -622,12 +622,12 @@ namespace CGAL {
       }
         
       const Bbox_3 &buildBoundingCube() {
-        FT min[] = {(std::numeric_limits<FT>::max)(),
-                    (std::numeric_limits<FT>::max)(),
-                    (std::numeric_limits<FT>::max)()};
-        FT max[] = {(std::numeric_limits<FT>::min)(),
-                    (std::numeric_limits<FT>::min)(),
-                    (std::numeric_limits<FT>::min)()};
+        FT min[] = {std::numeric_limits<FT>::infinity(),
+                    std::numeric_limits<FT>::infinity(),
+                    std::numeric_limits<FT>::infinity()};
+        FT max[] = {-std::numeric_limits<FT>::infinity(),
+                    -std::numeric_limits<FT>::infinity(),
+                    -std::numeric_limits<FT>::infinity()};
 
         for (std::size_t i = 0;i<this->size();i++) {
           Point_3 p = get(m_point_pmap, *this->at(i));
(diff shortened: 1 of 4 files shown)


lrineau:
@mglisse https://github.com/CGAL/cgal/pull/2565#issuecomment-339972470
> Out of curiosity, is it properly documented that this only works if `FT` is float, double or similar (infinity, max and min are likely to return 0 for most types)? Not that anyone is going to try, but you never know... I see a mention in Shape_detection_traits.h, but supposedly one could provide their own class satisfying the `ShapeDetectionTraits` concept.

That is one reason to use `max()` instead of `infinity()`: there is more chance that it is implemented.

lrineau:
Actually... it seems all the use cases are the computation of a `Bbox_3`. And the number type of `CGAL::Bbox_3` is always `double`, for all CGAL kernels.

mglisse:
> That is one reason to use max() instead of infinity(): there is more chance that it is implemented.

Note that I wasn't suggesting changing the code, just making sure the doc was in sync.

> Actually... it seems all the use cases are the computation of a Bbox_3. And the number type of CGAL::Bbox_3 is always double, for all CGAL kernels.

Well the code in Octree.h does use FT directly, which isn't double for Epeck.

sgiraudot:
The `Shape_detection_traits` says that FT should be `float` or `double`, so I think we're good.

mglisse:
@sgiraudot "I see a mention in Shape_detection_traits.h, but supposedly one could provide their own class satisfying the ShapeDetectionTraits concept.", I just wonder if the remark should be copied from Shape_detection_traits to the concept.

sgiraudot:
Most probably… I don't think the package would work with something else than `float` or `double` (it's one of the many).

Issue #2119, opened 2017-05-12
Shape_detection_3: Seemingly arbitrary stop criterion might abort too early


ochilan:
## Issue Details

In the `detect` function of `Efficient_RANSAC`, there is a forced exit criterion:
```
if (failed_candidates >= 10000)
  force_exit = true;
```
Is there a specific reason for this choice? Running the algorithm on more complex point clouds sometimes seems to overlook quite a lot of primitives that are otherwise detected when increasing this value. Should it be chosen proportionally to the total number of points or something the like?

## Environment

* Operating system: Arch Linux 64-bit
* Compiler: gcc
* Release or debug mode: Both
* CGAL version: 4.9.1
* Boost version: 1.63.0

sgiraudot:
A CGAL developer is currently doing a code review for the `Efficient_RANSAC` class as we noticed that it sometimes did not give satisfying result. Thank you for pointing this issue, this could be one of the reasons of these unsatisfying results.

I'll update this issue when the code review is done.

lrineau:
Simon, what is the status of this issue, now?

sgiraudot:
@danston from your experiments with the `Efficient_RANSAC` code, did it look like this hard-coded exit criterion had any effect on the output?

Would you have any intuition on how we could tune this parameter automatically? (Maybe, as OP suggested, proportionally to the number of points?)

danston:
Hi @sgiraudot I honestly did not see this parameter, so it might be a reason for some of my results to not converge. I will do a few experiments and let you know asap.

danston:
Ok, I did a few tests. Actually, with my data sets (since they are quite simple) I never hit this maximum value, but I agree, this hard-coded value is weird and for large data sets it can become a problem. The simplest idea, which comes to my mind, is to let users to change this parameter (like maximum number of iterations). If the solvers hits this maximum value, we can output the warning and let solver run for some extra time based, for example, on the number of remaining points (if it is bigger than some specified value in percents, we continue). I also like the idea of making this value proportional to the number of all points, but we have to test what is better and if Ransac still works well after this modification.

sgiraudot:
@ochilan what is the size of the data you used? Number of points and number of shapes? Are you sure you hit this arbitrary limit? RANSAC can miss shapes for many reasons, I just want to make sure this is the one.

[...]

PR #2568, opened 2017-10-31
Shape detection: robustify RANSAC

sgiraudot:
## Summary of Changes

This PR makes the RANSAC algorithm more robust, especially in the case of very large scenes with small details (that currently lead to many missed detections):
* The depth limit of the octree is now computed dynamically with respect to the ratio between the bbox diagonal and the cluster epsilon parameter (to make sure the smallest level's cells are small enough)
* The hard limit of failed candidates that could lead to mistakingly early termination is now computed dynamically too with respect to the size of the input ([this was already reported and closed](https://github.com/CGAL/cgal/issues/2119) because I couldn't reproduce, but now I had the chance to experiment on this)
* One parameter used to compute the termination criteria was badly computed compared to what was done in the original article and prototype code (`generated_candidates`). _(I consider this latest thing as a bugfix and not just an enhancement.)_

All of these changes make the algorithm work much better on large data sets. I did not notice any regression on other point sets.

All of these changes are implementation "details" and the API is unchanged.

## Release Management

* Affected package(s): Point Set Shape Detection 3
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2119

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
@@ -350,7 +350,7 @@ shape. The implementation follows \cgalCite{schnabel2007efficient}.
           0);
 
         m_available_octree_sizes[s] = subsetSize;
-        m_direct_octrees[s]->createTree();
+        m_direct_octrees[s]->createTree(m_options.cluster_epsilon);
 
         remainingPoints -= subsetSize;
       }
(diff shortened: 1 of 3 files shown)


afabri:
VC++ problem with `std::log2`.  Maybe including `<cmath>` does it.  Otherwise [`log(...)/log(2.0)`](https://stackoverflow.com/questions/22388630/is-stdlog2-an-extension-or-c-standard)

lrineau:
http://en.cppreference.com/w/cpp/numeric/math/log2

PR #2572, opened 2017-11-03
Add missing null descriptor functions in graph_traits of triangulations

sloriot:


lrineau:
Bug: the documented concept talks about a free function `null_face(Graph)`, and so on, and not a static function in the `graph_traits`. What is right?

afabri:
The static function.

lrineau:
Indeed `graph_traits<G>::null_vertex()` is part of the Boost `Graph` concept.

Then... How can it work with LCC?

sloriot:
if it is not used we won't see it (as for the functions added in this PR).

sloriot:
The doc issue is addressed in a separate PR. @lrineau can you mark the PR as ready or do you see another issue with this PR?

afabri:
Note that the way we document the free functions related to a concept are now documented the doxygen way, and no longer with a table so that it looks like the BGL documentation in boost.

PR #2571, opened 2017-11-02
Minkovski_sum_2: reduced conv collision detection

efifogel:
This fixes a bug in the Minkowski_sum_2 package.
In particular, it fixes a bug in the reduced-convolution  computation.
Specifically, given a face of the arrangement induced by the reduced convolution set, we need to test whether the face is in the sum or not. We pick a point x in the face and test whether P and (x \oplus -Q) collide. The changes in this brunch fix the code the implements the testing logic of the collision detection.

## Release Management

* Affected package(s): Minkowski_sum_2
* Issue(s) solved (if any): fix #2517
* Feature/Small Feature (if any): None

PR #2576, opened 2017-11-05
Aos_2: data traits construct opposite

### LINKED ISSUE #2063: Flipped segment with data does not contain original data
Problem
---

I use segments with data, in particular `CGAL::Arr_curve_data_traits_2<Segment_traits_2, int>::Curve_2`,the data is integer in this case. The problem is that if I call `flip()` function on a segment then the resulting segment does not have the data of the original segment.

Code 
-------

Example code reproducing this behavior:
```
#include <CGAL/Arr_segment_traits_2.h>
#include <CGAL/Arr_curve_data_traits_2.h>

typedef CGAL::Exact_predicates_exact_constructions_kernel	Kernel;
typedef Kernel::Point_2						Point_2;
typedef Kernel::Segment_2					Segment_2;
typedef CGAL::Arr_segment_traits_2<Kernel>			Segment_traits_2;
typedef CGAL::Arr_curve_data_traits_2<Segment_traits_2, int>	Traits_2;
typedef Traits_2::Curve_2					Curve_2;


int main(){
  Point_2 p1 = {0.0,0.0};
  Point_2 p2 = {1.0,0.0};

  Segment_2 s = {p1,p2};
  Curve_2 c = {s,42};
  Curve_2 fc = c.flip();

  std::cout << "Curve: " << c << std::endl;
  std::cout << "Curve id: " << c.data() << std::endl;
  std::cout << "Flipped curve: " << fc << std::endl;
  std::cout << "Flipped curve id: " << fc.data() << std::endl;
}
```

###  Output
```
Curve: 0 0 1 0
Curve id: 42
Flipped curve: 1 0 0 0
Flipped curve id: 0
```
### Expected output
```
Curve: 0 0 1 0
Curve id: 42
Flipped curve: 1 0 0 0
Flipped curve id: 42
```

Additional problems
---

I also encounter problems while doing Boolean operations on polygons whose segments have data. Doing unions and intersections works fine but when I do difference or symmetric difference then some of the edges lose its data. This is illustrated  on the picture below

screenshot from 2017-04-19 14-27-21

The shapes from left to right: 
red box(A)
green box(B)
reversed version of A
arrangement created by combining A and B
union of A and B
intersection of A and B
difference of A and B
symmetric difference of A and B

As you can see, some of the edges lost their color for some reason.

Maybe this is intended behavior, but it is quite unexpected.

efifogel:
This cleans the code of the `Arr_curve_data_traits_2<Tr, ...>` class template and fixes a bug in the code. Any instance of this template is a model of the `ArrangementTraits_2` concept and serves as a decorator class that allows the extension of curves defined by the base traits-class (the `Tr` parameter). The fix adds a wrapper to the `Construct_opposite_2` functor required by the concept `ArrangementDirectionalXMonotoneTraits_2`. The `()` operator of this functor accepts an x-monotone curve and returns its opposite curve. The wrapper attaches the data attached to the input curve to the output curve.

The implementation uses a helper class to detect the presence of a functor with the same name in the base geometry class. While at it, I've changed the code of the Are_mergeable_2 and Merge_2 wrappers to use similar technique, instead of the using the tags defined in the base traits class). In the future, I plan to use the same technique everywhere and render the tags obsolete.

## Release Management

* Issue(s) solved (if any): fix #2063
* Feature/Small Feature (if any): None

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
@@ -28,11 +28,13 @@
  * Definition of the Arr_curve_data_traits_2<> class template.
  */
 
-#include<CGAL/Object.h>
-#include<CGAL/tags.h>
-#include<CGAL/Arr_tags.h>
-#include<CGAL/Arr_geometry_traits/Curve_data_aux.h>
-#include<list>
+#include <list>
+#include <boost/utility/enable_if.hpp>
+
+#include <CGAL/Object.h>
+#include <CGAL/tags.h>
+#include <CGAL/Arr_tags.h>
+#include <CGAL/Arr_geometry_traits/Curve_data_aux.h>
 
 namespace CGAL {
 
@@ -93,10 +95,10 @@ public:
   /// \name Construction.
   //@{
 
-  /*! Default constructor. */
+  /*! Construct default. */
   Arr_curve_data_traits_2() {}
 
-  /*! Constructor from a base-traits class. */
+  /*! Construct from a base-traits class. */
   Arr_curve_data_traits_2(const Base_traits_2& traits) : Base_traits_2(traits) {}
   //@}
 
@@ -160,10 +162,10 @@ public:
     Split_2(const Base_traits_2& base) : m_base(base) {}
 
     /*! Split a given x-monotone curve at a given point into two sub-curves.
-     * \param cv The curve to split
-     * \param p The split point.
-     * \param c1 Output: The left resulting subcurve (p is its right endpoint).
-     * \param c2 Output: The right resulting subcurve (p is its left endpoint).
+     * \param cv[in] The curve to split
+     * \param p[in] The split point.
+     * \param c1[out] The left resulting subcurve (p is its right endpoint).
+     * \param c2[out] The right resulting subcurve (p is its left endpoint).
      * \pre p lies on cv but is not one of its end-points.
      */
     void operator()(const X_monotone_curve_2& cv, const Point_2& p,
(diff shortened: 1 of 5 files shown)


lrineau:
The Aos_2 test results were pretty bad last night:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-92.shtml#Arrangement_on_surface_2

It could be because of this PR, or #2571.

efifogel:
No idea why it fails. (Runs fine on my machine.)
I've enhanced the test, so let's give it another try.

                         _/

lrineau:
On my machine it says:
```shellsession
 % ../../../build/test/Arrangement_on_surface_2/test_data_traits 
Construct opposite failed: 1234 != 32577
```

efifogel:
I pushed something, but I don't think it should make a difference, and I
ran out of options...
Could you please try again. If it still fails, I'll need to get access to
some platform that the test fails on.

                         _/

Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
```diff
@@ -49,407 +46,417 @@ namespace CGAL {
...
+     * a nested type named Are_mergeable_2.
      */
-    bool operator() (const X_monotone_curve_2& cv1,
-                     const X_monotone_curve_2& cv2) const
-    {
-      return (_are_mergeable_base_imp (cv1, cv2, Base_has_merge_category()));
-    }
+    template <typename T>
+    struct has_are_mergeable_2 {
+      typedef char yes[1];
```
sloriot: You can use BOOST_MPL_HAS_XXX_TRAIT_NAMED_DEF

Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
```diff
@@ -49,407 +46,417 @@ namespace CGAL {
...
      */
-    void operator() (const X_monotone_curve_2& cv1,
-                     const X_monotone_curve_2& cv2,
-                     X_monotone_curve_2& c) const
-    {
-      // The function is implemented based on the base Has_merge category.
-      _merge_imp (cv1, cv2, c, Base_has_merge_category());
-    }
+    template <typename T>
+    struct has_merge_2 {
```
sloriot: same here

Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
```diff
@@ -49,407 +46,417 @@ namespace CGAL {
...
-      return (X_monotone_curve_2 (base_cv, X_monotone_curve_data()));
+  class Construct_opposite_2 {
+  private:
+    const Base_traits_2& m_base;
+
+    /*! Helper class template to find out whether the base geometry traits has
+     * a nested type named Construct_opposite_2.
+     */
+    template <typename T>
+    struct has_construct_opposite_2 {
```
sloriot: same here

efifogel:
Aha,
I've used it x3 (and simplified the code accordingly).
So let's try the branch again.

                         _/

lrineau:
The PR conflicts with `master`!

PR #2578, opened 2017-11-06
Polyhedron demo: Add 2D image management.

### LINKED ISSUE #2573: Polyhedron demo, 3D image viewer: support 2D images
When a "3D image" has a dimension equal to 1, then that is actually a 2D image. In that case, only one plane should be showed (the one with dim=1), and all sliders should be deactivated.

maxGimeno:
## Summary of Changes
Disable the sliders and hide the handles if the loaded/selected image has one dimension with value = 1.
## Release Management

* Issue(s) solved (if any): fix #2573


lrineau:
I have tested the code on my Linux machine, and it works as expected. Thanks.

PR #2589, opened 2017-11-09
Polyhedron demo: Fix pixel value in Io_image_plugin

### LINKED ISSUE #2588: Demo io_image_plugin: Wrong display of picked pixel values
## Issue Details

In the Polyhedron Demo, when the gray-level image `examples/Mesh_3/data/skull_2.9.inr` is open, and when with <kbd>Shift</kbd>+left-click a pixel is picked, the value of the pixel is displayed, at the top of the "Volume Plantes Control" dock-widget:

spectacle ve5661

Two issues:
1. The string "Isovalue at point:" should be "Value of that pixel:".
2. The value is displayed as an integer, even if the image word type is `float`. There is some code in the plugin to deal with that, but it must have been modified and a regression was introduced.

Please fix in 4.11 is the bug can be confirmed in that branch. It is sufficient to compile `io_image_plugin`.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux Fedora 26
* Compiler: g++
* Specific flags used (if any): `-g -Og`
* CGAL version: Tuesday `integration`, but that is probably the same with `master` (64023d2b3334eb2fed2581e397e456ad160b7871)

maxGimeno:
## Summary of Changes
This PR changes the signal emitted when a pixel is picked so it displays either an integer or a double according the the isovalue type of the image
## Release Management

* Issue(s) solved (if any): fix #2588

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
@@ -93,7 +93,7 @@ public Q_SLOTS:
     getPixel(e->pos());
   }
 Q_SIGNALS:
-  void x(int);
+  void x(QString);
 
 public:
   void setIC(const IntConverter& x) { ic = x; fc = boost::optional<DoubleConverter>(); }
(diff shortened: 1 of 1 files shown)


lrineau:
In #2588, I said:
> Please fix in 4.11 is the bug can be confirmed in that branch. 

Could you please rebase the patch and the PR?

maxGimeno:
rebased on 4.11

PR #2592, opened 2017-11-13
Faster to_interval(Gmpq)

mglisse:
## Summary of Changes

Make conversion to interval through MPFR3 faster. Keep old code for compatibility with centos with mpfr-2.2 or 2.4.

## Release Management

### DIFF
--- a/Number_types/include/CGAL/double.h
+++ b/Number_types/include/CGAL/double.h
@@ -138,8 +138,7 @@ inline double sse2fabs(double a)
   __m128d temp = _mm_set1_pd(a);
   
   temp = _mm_and_pd(temp, absMask.m);
-  _mm_store_sd(&a, temp);
-  return a;
+  return _mm_cvtsd_f64 (temp);
 }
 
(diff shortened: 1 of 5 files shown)


lrineau:
Only for `master` (4.12)?

Number_types/include/CGAL/double.h
```diff
@@ -138,8 +138,7 @@ inline double sse2fabs(double a)
   __m128d temp = _mm_set1_pd(a);
   
   temp = _mm_and_pd(temp, absMask.m);
-  _mm_store_sd(&a, temp);
-  return a;
+  return _mm_cvtsd_f64 (temp);
```
lrineau: Is that change on purpose? It does not seem related to the description of the PR?

lrineau: Marc told me that it was on purpose.

mglisse:
Yes, only master, it is a new optimization, it doesn't seem necessary to backport.

PR #2595, opened 2017-11-13
BGL/LCC: use generic load

gdamiand:
## Summary of Changes

In BGL for LCC, replace the use of my own copy of cgal/bgl/graph/io.h by the original file (which is now integrated in master since https://github.com/CGAL/cgal/commit/b1c08973d0a9b578248c14e9b762c727097e2614).

Update two tests in BGL that can now use the generic load function instead of some specialized functions.

## Release Management

* Affected package(s): LCC, BGL


lrineau:
Guilaume, would you mind rebase on `master` and resolve the conflict? That will ease the tests in `integration`.

gdamiand:
Done. By the way, we will have conflicts with all our branches following the modifications of all cgal header file (add of // SPDX-License-Identifier:)

lrineau:
The conflicts will only be for deleted files (because they were modified in-between in `master`). And there is also a kind of conflict with added files: Git will not see the conflict, but our testing scripts will report that the new files do not have the `SPDX-License-Identifier`. Cc @sloriot

sloriot:
Note that the dev manual is updated in https://github.com/CGAL/cgal/pull/2605 to mention the presence of the 3 tags.

PR #2596, opened 2017-11-13
Mesh 2 : fix the use of Lloyd from any CDT_2

janetournois:
## Summary of Changes

This PR fixes the use of `CGAL::lloyd_optimize_mesh_2()` when used on a `CDT_2` that was not constructed using `Mesh_2`.
Formerly, the convergence test `check_convergence()` was returning `true` all the time because no facet was tagged as "in the domain".
The convergence test should not suffer from that change, mainly because constrained vertices are not considered.

## Release Management

* Affected package : Mesh_2 (smoothing only)

Thanks @gdamiand for the bug report!

### DIFF
--- a/Mesh_2/doc/Mesh_2/CGAL/lloyd_optimize_mesh_2.h
+++ b/Mesh_2/doc/Mesh_2/CGAL/lloyd_optimize_mesh_2.h
@@ -101,13 +101,13 @@ lloyd_optimize_mesh_2(cdt,
 
 \endcode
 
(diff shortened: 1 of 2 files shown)


lrineau:
As you know, the testsuite reported several errors:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-106.shtml#Mesh_2

lrineau:
Ping @janetournois. This PR still blocks 4.10.2.

lrineau:

Issue #1892, opened 2017-02-09
Remnants of SVN in headers & documentation


MaelRL:
The [File header documentation](http://doc.cgal.org/latest/Manual/devman_code_format.html#secfile_header) and some headers (e.g. [this one](https://github.com/CGAL/cgal/blob/master/AABB_tree/include/CGAL/internal/AABB_tree/Primitive_helper.h)) still mention SVN.

Is there any reason not to clean it? Should it be updated to do the same with Git or just completely removed?

sloriot:
I would vote for removing them

afabri:
@lrineau any opinion on this?

lrineau:
It seems that most of the existing headers (2852 out of 3032) have `$URL$` and `$Id$`.

I would not recommend that we forcefully remove those SVN keywords from all headers.

I propose a choice, for a discussion:
- The first possibility is to deprecate the use of those headers: remove their mention from the documentation.
- The second possibility is to restore a usage of those keywords: for example, the release creation script could substitute them to give of the mapping between the Git layout and the tarball layout.

afabri:
In order to get what from the second possibility?

lrineau:
> mapping between the Git layout and the tarball layout

For example, the file `CGAL-x.y/include/CGAL/Has_timestamp.h` of the tarball would mention in `$Id: $` its full path in the Git layout: `STL_Extension/include/CGAL/Has_timestamp.h`, as well as the "Date" and the hash of the last commit that modified it. Maybe the `$URL: $` could point to the file in the Github website, so that the use can access the history, or other contextual information given by the Github user interface.

sgiraudot:
Can we carry on this discussion and take a decision?

I like the second possibility, with the drawback that it can create broken links on old releases.

sloriot:
I'm adding the dev-meeting flag

MaelRL:
Putting the path of the file with the commit hash (of the release), e.g.
[...]

PR #2604, opened 2017-11-15
Add missing URL/Id tags

sloriot:
Part of fixing #1892


MaelRL:
Was a decision taken on the (future?) purpose of these tags ?

sloriot:
When the release is generated, we'll put the canonical link on github and the release number.

Issue #1728, opened 2016-12-01
Handling the last dont_submit files


sloriot:
Follow up of #1446

Some `dont_submit` files are remaining:
- [x] `Installation` : 4 files related to the `changes.html` file
- [ ] `wininst` : some files are ignored. Can the whole directory be ignored?
- [ ] `Three` : Only the documentation,  while waiting for a review.
- [x] `Operations_on_polyhedra` :  will disappear soon with a merge of a small feature

lrineau:
@sloriot: Why have you assign this task to Maxime? For all items in the list, you need to know a lot about the internals of the CGAL project to be able to conclude.

sloriot:
I'm sure he can investigate how things work for creating the windows installer from the script that is in the docker container (my guess was that the whole directory can be ignored by the release creation script) and propose something for Installation.

lrineau:
@maxGimeno I have reassigned the issue to you. Discuss with Sébastien for the details.

sgiraudot:
@maxGimeno any update on this issue? Are you still working on it?

maxGimeno:
Still waiting for clear instructions.
Low priority.

lrineau:
- For `Installation/`: we should try to convert our changes.html to a Markdown file,
- for `wininst/`, the `dont_submit` file can be removed,
- for `Three/doc`, that is one good reason to keep `dont_submit` files.

Issue #23, opened 2015-03-24
Build a release using cmake


lrineau:
We really need that script...

**Update:**
Sébastien said, in https://github.com/CGAL/cgal/issues/23#issuecomment-241346665:

> From what I remember the branch contains two things:
- [x] building a release using CMake (see PR #2606)
- [x] ignoring the library configuration tests when only building the documentation

**Update:**
Issue renamed to _Build a release using cmake_

sgiraudot:
Is this still needed today? Who can take care of this?

I can see the [branch](https://github.com/CGAL/cgal-dev/tree/Installation-make_release-pmoeller) is still in `cgal-dev` but seems very old (but was already pretty old when this issue was opened).

lrineau:
I know the branch is very old. It was done by Philipp (@bo0ts). I filled the issue not to forget that the code of the branch should be reworked and integrated. That could also be a topic for next meeting.

@sloriot Do you have an opinion on the feature, or the code?

sloriot:
From what I remember the branch contains two things:
- building a release using CMake
- ignoring the library configuration tests when only building the documentation
The reason why the branch was not merged is that the testsuite showed some strange issue on windows that we could not reproduce.

sloriot:
See #1444 that removes the library configuration step if we're interested only by building the doc

sloriot:
@maxGimeno  @lrineau  Since the remaining files in `dont_submit` are neither headers, nor examples/demos files this is no longer a blocking point for this issue.

lrineau:
Fixed by #2606.

PR #2606, opened 2017-11-17
Add script to create release with cmake

sloriot:


sloriot:
Related issues: #1892 and #23

sloriot:
And also #1728

sloriot:
Note that the script requires that `Installation/include/CGAL/version.h` exists, according to another PR from @lrineau

lrineau:
When you mention related issues: will this PR fix those issues or not?

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```diff
@@ -61,8 +64,17 @@ foreach(pkg ${files})
...
+        #copy the file (replace $URL$ and $ID$ for *.h and *.hpp)
+        get_filename_component(fext ${fname} EXT)
+        if ("${fext}" STREQUAL ".h" OR "${fext}" STREQUAL ".hpp")
+          file(READ "${pkg_dir}/${f}" file_content)
+          string(REPLACE "$URL$" "${GITHUB_PREFIX}/${pkg}/${f}" file_content ${file_content})
+          string(REPLACE "$Id$" "This file is from the release ${CGAL_VERSION} of CGAL" file_content ${file_content})
+          file(WRITE ${release_dir}/${afile_dir}/${fname} ${file_content})
+        else()
+          file(COPY ${afile} DESTINATION ${release_dir}/${afile_dir})
+        endif()
```
lrineau: 1. Does not this slow down the script a lot?
2. The `$URL` and `$Id$` must be replaced by something starts with `$URL: ` or `$Id: ` and ends with `$`.

sloriot:
Yes it will

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
sloriot: The script is pretty fast (still way faster than the old one even on my laptop).
Do you mean I should replace by `$URL: https://www.github.com/....`?

lrineau: By `$URL: https://www.github.com/....$`

sloriot:
Any proposal for the replacement of Id of is what I put OK?

MaelRL:
dd5a9163d01577101c2a15e9c97fa3f6dcdae2fa should do the same for ID.

/Manual/devman_code_format.html#secfile_header needs to be edited too.

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```diff
@@ -0,0 +1,110 @@
...
+  message(FATAL_ERROR "Cannot find Installation/include/CGAL/version.h. Make sure you are at the root of a CGAL branch")
+endif()
+
+file(READ "${CMAKE_BINARY_DIR}/Installation/include/CGAL/version.h" version_file_content)
+string(REGEX MATCH "define CGAL_VERSION (.*)\n#define CGAL_VERSION_NR" CGAL_VERSION_FOUND "${version_file_content}")
+
+if (CGAL_VERSION_FOUND)
+  set(CGAL_VERSION "${CMAKE_MATCH_1}")
+  set (GITHUB_PREFIX "https://github.com/CGAL/cgal/blob/releases/CGAL-${CGAL_VERSION}")
+else()
```
lrineau: That URL will be almost always wrong. Instead you should use the real hash,`${CGAL_GIT_HASH}`, after having included `Installation/cmake/modules/CGAL_SCM.cmake`.

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
lrineau: `$Id$` is supposed to be turned into `$Id: <id>$` when `<id>` is a "summary of the other information":
http://svnbook.red-bean.com/en/1.7/svn.advanced.props.special.keywords.html

Probably to be used by tool to designate the file. In the context of Git, it could be the full filename, plus the result of `git describe HEAD`.

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
sloriot: The URL will indeed be wrong for internal releases but is perfectly fine for official public releases.
Using the hash is not a solution for internal releases since the hash will no longer be valid after a week (IIRC we don't have tags for Ic/I releases).

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```diff
@@ -70,7 +70,16 @@ foreach(pkg ${files})
         if ("${fext}" STREQUAL ".h" OR "${fext}" STREQUAL ".hpp")
           file(READ "${pkg_dir}/${f}" file_content)
           string(REPLACE "$URL$" "$URL: ${GITHUB_PREFIX}/${pkg}/${f} $" file_content ${file_content})
-          string(REPLACE "$Id$" "This file is from the release ${CGAL_VERSION} of CGAL" file_content ${file_content})
+          if(EXISTS ${CMAKE_BINARY_DIR}/.git)
```
lrineau: `CMAKE_SOURCE_DIR`

sloriot: cmake must be called at the root of the repository which is `CMAKE_BINARY_DIR`, so it should be correct. Note that I used only `CMAKE_BINARY_DIR` in this script.

lrineau:
What is the next step, now?

sloriot:
I think the last step is now to use the cmake script to create the releases for the testsuites. Shall @maxGimeno take care of it?

lrineau:
Hi, I am trying to find out how to use the script. Its first lines says:
```
#option :
# DESTINATION the path where the release is create, default is /tmp
# PUBLIC=[ON/OFF] indicates if a public release should be built
# INTERNAL=[ON/OFF] indicates if an internal release should be built
# VERBOSE=[ON/OFF] makes the script more verbose
```
Reading/grepping quickly the code, I found out that:
1. One cannot build both a public and an internal releases at the same time, like the previous script: the target will be a public release if and only if `PUBLIC` is set to `TRUE`.
2. The option `INTERNAL` is useless: it is the default, and conflicts with `PUBLIC`.

And there is no way to set the name of the internal release, like `CGAL-4.12-Ic-142`, for the moment.

I propose to:
 - remove the option `INTERNAL`, unless we allow the creation of two distinct releases (internal+public) at the same time,
 - add an option to choose the name of the internal release.

sloriot:
Indeed, you create either a public or an internal release. I agree with your 2 points.

lrineau:
I ran that script with a checkout of 9c6b1314ebe4781b29bd61de2e6f57b9351b5b1d, and compared it with https://cgal.geometryfactory.com/CGAL/Members/Releases/CGAL-4.12-I-131.tar.gz, and the results are really different. The CMake script seems to tripped all `;` character from any file it processes.

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
lrineau: Here is something I did not know:
> When run in `-P` script mode, CMake sets the variables `CMAKE_BINARY_DIR`, `CMAKE_SOURCE_DIR`, `CMAKE_CURRENT_BINARY_DIR` and `CMAKE_CURRENT_SOURCE_DIR` to the current working directory.

https://cmake.org/cmake/help/v3.6/variable/CMAKE_BINARY_DIR.html

So `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` are identical and set to the current working directory.

sloriot:
See #2707

lrineau:
The script `Scripts/developer_scripts/cgal_create_release_with_cmake.cmake` does not seem to be documented in our wiki. Is it documented somewhere else? I know it starts with detailed comments, but that cannot replace a documentation of usage.

PR #2607, opened 2017-11-17
Polyhedron demo: Fixes for QGLViewer 2.7 and remove GlSplat

maxGimeno:
## Summary of Changes
 This PR fixes several problems faced with recent versions of QGLViewer, mainly in the selections plugin.


Polyhedron/demo/Polyhedron/Plugins/PCA/Scene_edit_box_item.cpp
```diff
@@ -773,6 +773,7 @@ void Scene_edit_box_item::highlight()
 {
   d->ready_to_hl = true;
   Viewer_interface* viewer = dynamic_cast<Viewer_interface*>(*QGLViewer::QGLViewerPool().begin());
+  viewer->makeCurrent();
```
lrineau: How will it be implemented in our future multi-viewer world?

maxGimeno: makeCurrent() is a function from Qt, not qglViewer, so it will be the same as I intend to use QOpenGLWidget (which works for Android, unlike the old QGLWidget)

lrineau: I mean: if there are multiple viewer, `QGLViewer::QGLViewerPool()->begin()` may not be the right viewer! I think the API of that item will have to be changed so that the `Scene_edit_box_item` knows in which viewer it is used.

maxGimeno: Oh, that. I could use the lambda function version of the singleShot to send the viewer as argument for the slot `highlight()`.

maxGimeno: An object should not be usable in only one viewer, especially if it is not in ht emain one, as it can be destroyed any time. The solution I propose would be consistent with what you are saying. Adding an argument to `highlight()` would avoid having to cast from the pool, and the function that calls highlight casts from its `obj`argument, which is the affected viewer. So I think it is the best solution.

lrineau:
This PR is broken. Try to compile it locally.

lrineau:
Note that commit https://github.com/CGAL/cgal/pull/2607/commits/1ac1eb3fdb34094fc3e589f670bab3f314980900 removes GlSplat. I have modified the title of the PR.

sloriot:
@lrineau  do you know why this PR was not selected for 4.12? Last change was November 15.
Someone on [SO](https://stackoverflow.com/questions/49055638/error-while-building-polyhedron-demo) seems to be affected by the issue.

maxGimeno:
Actually I rebased it a few days ago to get rid of some conflicts.

lrineau:
I do not know why it was not selected. Maybe I thought it was unmaintained?

Let's try to get it into CGAL-4.12-beta2.

lrineau:

sloriot:
@maxGimeno this PR is breaking the change of the rotation center (right-click+double left-click).

maxGimeno:
See #2904. .

PR #2598, opened 2017-11-14
Correctly handle faces with more than one cc of the boundary

sloriot:
Since faces were not triangulated in the conversion from Nef_3 to FaceGraph, faces with more than one boundary where not correctly represented.

Fixes issues reported on [cgal-discuss](https://sympa.inria.fr/sympa/arc/cgal-discuss/2017-11/msg00022.html)

PR #2614, opened 2017-11-17
Speed Improvement/Refactoring nef polyhedron 3.

GilesBathgate:
## Summary of Changes

I would like to propose some changes which measurably improve the performance (1.3x faster in my tests) of converting a Nef_polyhedron_3 to a Polyhedron_3. This is primarily achieved by removing the unique hash maps and using `Triangulation_vertex_base_with_info_2`, and `Triangulation_face_base_with_info_2` to decorate the facets themselves with the required data, and save the lookup in the hash maps altogether.

Whilst trying to understand the implementation of `Nef_polyhedron_3::Triangulation_handler2` i noticed the code wasn't very clear or easy to understand, so I have also done some refactoring to make the code clearer. (These changes are easier to see by looking at the individual commits, rather than the PR as a whole.) 

## Release Management

* Affected package(s): The `Nef_polyhedron_3::Triangulation_handler2` class and the `convert_to_polyhedron()` function.
* Feature/Small Feature (if any): 1.3x performance improvement when calling `convert_to_polyhedron()`

### DIFF
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -801,7 +801,8 @@ protected:
 
           Halffacet_const_handle f = opposite_facet->twin();
 
-          if(needs_triangulation(f)) {
+          int s = sides(f);
+          if(s > 4 || has_holes(f)) {
               Vector_3 orth = f->plane().orthogonal_vector();
               int c = CGAL::abs(orth[0]) > CGAL::abs(orth[1]) ? 0 : 1;
               c = CGAL::abs(orth[2]) > CGAL::abs(orth[c]) ? 2 : c;
@@ -817,29 +818,48 @@ protected:
                   th.handle_triangles(B, VI);
               } else
                   CGAL_error_msg( "wrong value");
-
           } else {
-
               B.begin_facet();
               Halffacet_cycle_const_iterator fc = f->facet_cycles_begin();
               SHalfedge_const_handle se(fc);
               CGAL_assertion(se != 0);
               SHalfedge_around_facet_const_circulator hc(se),he(hc);
+              int t = 0, fv;
               CGAL_For_all(hc,he) {
                   Vertex_const_handle cv = hc->source()->center_vertex();
                   CGAL_NEF_TRACEN("   add vertex " << cv->point());
-                  B.add_vertex_to_facet(VI[cv]);
+                  int i=VI[cv];
+                  B.add_vertex_to_facet(i);
+                  if(s==4) {
+                      if(t==0) fv=i;
+                      if(++t==3) {
+                        B.end_facet();
+                        B.begin_facet();
+                        B.add_vertex_to_facet(i);
+                      }
+                  }
               }
+              if(s==4)
+                  B.add_vertex_to_facet(fv);
               B.end_facet();
           }
       }
 
-      inline bool needs_triangulation(Halffacet_const_handle f)
+      inline int sides(Halffacet_const_handle f)
       {
           SHalfedge_around_facet_const_circulator
                   sfc1(f->facet_cycles_begin()), sfc2(sfc1);
-          return ++f->facet_cycles_begin() != f->facet_cycles_end() ||
-                  ++(++(++sfc1)) != sfc2;
+          if(++(++(++sfc1)) == sfc2)
+              return 3;
+          if((++sfc1) == sfc2)
+              return 4;
+           //We only care whether there are 3,4 or greater sides.
+          return 0xFF;
+      }
+
+      inline bool has_holes(Halffacet_const_handle f)
+      {
+          return cpp11::next(f->facet_cycles_begin())!=f->facet_cycles_end();
       }
 
       void visit(SFace_const_handle) {}
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -712,6 +712,10 @@ protected:
             }
         CGAL_assertion(ct.is_valid());
 
+        Finite_face_iterator fi;
+        for(fi=ct.finite_faces_begin();fi!=ct.finite_faces_end(); ++fi)
+            fi->info() = false; // Mark faces unvisited
+
         CGAL_NEF_TRACEN("number of finite triangles " << ct.number_of_faces());
 
         Face_handle infinite = ct.infinite_face();
@@ -728,9 +732,8 @@ protected:
         CGAL_assertion(!ct.is_infinite(first));
         traverse_triangulation(first, first->index(opposite));
 
-        Finite_face_iterator fi;
         for(fi=ct.finite_faces_begin();fi!=ct.finite_faces_end(); ++fi) {
-                if(fi->info() == false) continue;
+                if(!fi->info()) continue; // Skip unvisited faces
                 CTVertex_handle p=fi->vertex(0);
                 CTVertex_handle q=fi->vertex(1);
                 CTVertex_handle r=fi->vertex(2);
@@ -759,7 +762,7 @@ protected:
     }
 
     void traverse_triangulation(Face_handle f, int parent) {
-       f->info() = true;
+       f->info() = true; // Mark face as visited
        descend_triangulation(f,ct.cw(parent));
        descend_triangulation(f,ct.ccw(parent));
     }
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -802,7 +802,7 @@ protected:
           Halffacet_const_handle f = opposite_facet->twin();
 
           int s = sides(f);
-          if(s > 4 || has_holes(f)) {
+          if(s > 3 || has_holes(f)) {
               Vector_3 orth = f->plane().orthogonal_vector();
               int c = CGAL::abs(orth[0]) > CGAL::abs(orth[1]) ? 0 : 1;
               c = CGAL::abs(orth[2]) > CGAL::abs(orth[c]) ? 2 : c;
@@ -820,27 +820,14 @@ protected:
                   CGAL_error_msg( "wrong value");
           } else {
               B.begin_facet();
-              Halffacet_cycle_const_iterator fc = f->facet_cycles_begin();
-              SHalfedge_const_handle se(fc);
+              SHalfedge_const_handle se(f->facet_cycles_begin());
               CGAL_assertion(se != 0);
               SHalfedge_around_facet_const_circulator hc(se),he(hc);
-              int t = 0, fv;
               CGAL_For_all(hc,he) {
                   Vertex_const_handle cv = hc->source()->center_vertex();
                   CGAL_NEF_TRACEN("   add vertex " << cv->point());
-                  int i=VI[cv];
-                  B.add_vertex_to_facet(i);
-                  if(s==4) {
-                      if(t==0) fv=i;
-                      if(++t==3) {
-                        B.end_facet();
-                        B.begin_facet();
-                        B.add_vertex_to_facet(i);
-                      }
-                  }
+                  B.add_vertex_to_facet(VI[cv]);
               }
-              if(s==4)
-                  B.add_vertex_to_facet(fv);
               B.end_facet();
           }
       }
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -67,6 +67,7 @@
 #include <CGAL/Projection_traits_yz_3.h>
 #include <CGAL/Projection_traits_xz_3.h>
 #include <CGAL/Triangulation_vertex_base_with_info_2.h>
+#include <CGAL/Triangulation_face_base_with_info_2.h>
 #include <CGAL/Constrained_triangulation_face_base_2.h>
 #include <list>
 
@@ -672,7 +673,8 @@ protected:
   class Triangulation_handler2 {
 
     typedef typename CGAL::Triangulation_vertex_base_with_info_2<Vertex_const_handle,Kernel> Vb;
-    typedef typename CGAL::Constrained_triangulation_face_base_2<Kernel>     Fb;
+    typedef typename CGAL::Triangulation_face_base_with_info_2<bool,Kernel> Info;
+    typedef typename CGAL::Constrained_triangulation_face_base_2<Kernel,Info>     Fb;
     typedef typename CGAL::Triangulation_data_structure_2<Vb,Fb>             TDS;
     typedef typename CGAL::Constrained_triangulation_2<Kernel,TDS>           CT;
 
@@ -683,13 +685,12 @@ protected:
     typedef typename CT::Edge                  Edge;
 
     CT ct;
-    CGAL::Unique_hash_map<Face_handle, bool> visited;
     Halffacet_const_handle f;
     Plane_3 supporting_plane;
 
   public:
     Triangulation_handler2(Halffacet_const_handle f) : 
-      visited(false), supporting_plane(f->plane()), f(f) {}
+      supporting_plane(f->plane()), f(f) {}
 
     template<typename PIB, typename Index>
     void handle_triangles(PIB& pib, Index& VI) {
@@ -728,7 +729,7 @@ protected:
       traverse_triangulation(first, first->index(opposite));
     Finite_face_iterator fi;
     for(fi=ct.finite_faces_begin();fi!=ct.finite_faces_end(); ++fi) {
-       if(visited[fi] == false) continue;
+       if(fi->info() == false) continue;
 	   CTVertex_handle p=fi->vertex(0);
 	   CTVertex_handle q=fi->vertex(1);
 	   CTVertex_handle r=fi->vertex(2);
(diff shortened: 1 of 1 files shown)


sloriot:
Did you compare with the runtime of the function [`CGAL::convert_nef_polyhedron_to_polygon_mesh()`](https://doc.cgal.org/latest/BGL/group__PkgBGL.html#ga7cf62f17d3bb060519e682ad9499c7d4)?

GilesBathgate:
@sloriot No, I wasn't able to test that function on my setup, although I have no doubt that the function will be faster. My primary objective was to improve this for legacy code and indeed I wrote the change against CGAL 4.9 and forward ported it to the current codebase. Whilst forward porting I did notice the PMP conversion functions in my diff tool, and I am interested in porting my legacy code (that currently depends on `Polyhedron_3`) to use PMP at some point in the future.

GilesBathgate:
@sloriot I thought you were referring to `polygon_mesh_to_nef_3` https://github.com/CGAL/cgal/blob/master/Nef_3/include/CGAL/Nef_polyhedron_3.h#L625 (which is what I saw in my diff tool) in retrospect I realise that that function goes in the opposite direction. Either way the function you reference doesn't appear to be available to my legacy codebase, so I think the gist of my previous comment still makes sense ;)

GilesBathgate:
@sloriot I have created the following 3 performance tests. It simply does convert_to_polyhedron et al 1000 times:

https://github.com/GilesBathgate/cgal/commit/93ff95394cfb29ccabb65604706594b2eabf05ec
https://github.com/GilesBathgate/cgal/commit/b75fd12e11323ac0be8129cc970ab367d5c2b731
https://github.com/GilesBathgate/cgal/commit/8276f7f1db78a263f493573d9816d85438ada462

## original
giles@BP0039a:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off
real	0m9.739s
user	0m9.736s
sys	0m0.004s

## with refactoring
giles@BP0039a:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off
real	0m8.012s
user	0m8.008s
sys	0m0.004s

## convert_nef_polyhedron_to_polygon_mesh
giles@BP0039a:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off
real	0m0.303s
user	0m0.288s
sys	0m0.016s

While its clear that the latter is much faster, my application needs the polyhedron faces to be triangulated, and I couldn't get the `CGAL::Polygon_mesh_processing::triangulate_faces` function to work without compilation errors. (I think that would be needed for a fair test)

GilesBathgate:
@sloriot I managed to fix the compile error by switching to a different kernel, this makes all the tests run faster ;)

## original
```
giles@home:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off 
real	0m2.881s
user	0m2.876s
sys	0m0.004s
```

## with refactoring
```
giles@home:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off 
real	0m2.243s
user	0m2.240s
sys	0m0.000s
```

## convert_nef_polyhedron_to_polygon_mesh (and triangulate)
```
giles@home:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off 
real	0m2.131s
user	0m2.112s
sys	0m0.020s
```

So `convert_nef_polyhedron_to_polygon_mesh` is still the winner, but the difference is now less substantial.

GilesBathgate:
@sloriot I seem to be getting different output from the refactored version vs the convert_to_mesh method, essentially it seems that the latter is not correctly triangulating polygons with holes. Is there some additional step needed?

The outputs can be seen in Github's online stl viewer (but you will have to switch to wireframe view)

### with refactoring
https://github.com/GilesBathgate/cgal/blob/performance-test1/Nef_3/examples/Nef_3/cube.stl

### convert_nef_polyhedron_to_polygon_mesh (and triangulate)
https://github.com/GilesBathgate/cgal/blob/performance-test3/Nef_3/examples/Nef_3/cube.stl

kintel:
@sloriot I didn't read through the code in too much detail, but at first glance it looks like `CGAL::convert_nef_polyhedron_to_polygon_mesh()` ignores holes in Nef3 faces. Could you comment on that?

sloriot:
@kintel Indeed, this is fixed in https://github.com/CGAL/cgal/pull/2598

GilesBathgate:
@sloriot Ah ok great. I've applied that patch, and now the STL output is fixed. It did significantly impact the timing though :(

### Patched convert_nef_polyhedron_to_polygon_mesh (and triangulate)
```
giles@home:~/cgal/Nef_3/examples/Nef_3$ time ./interface_polyhedron < cube.off
real	0m4.893s
user	0m4.876s
sys	0m0.016s
```

lrineau:
Now it seems your version ("with refactoring") is the fastest.

GilesBathgate:
@lrineau Yes, although this is a simple test case. For example, I don't think there is any advantage in this specific input geometry for the triangulation to require being delaunlay. `convert_to_polyhedron` uses a `Contrained_triangulation_2`, whereas `convert_nef_polyhedron_to_polygon_mesh()` uses `Contrained_Delaunlay_triangulation_2` so that could be the cause of the performance hit. I can run some tests with some different input if you would be so kind as to suggest ones to try.

GilesBathgate:
...also @kintel do you have some examples of .nef files I could try, you mentioned you had robustness issues with `convert_to_polyhedron` and I wanted to see if `convert_nef_polyhedron_to_polygon_mesh` faired any better.

lrineau:
There are `[-Wreorder]` warnings:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-110.shtml#Nef_3

GilesBathgate:
@lrineau Yes I do work for blueprism, however all this work was done during my personal time and during annual leave. I am happy to sign copyright back over to the original copyright owner. If you can give me some further detail on the -Wreorder warnings I would be happy to fix them.

lrineau:
As for the warning, that is easy. The warning is:
```
.../include/CGAL/Nef_polyhedron_3.h:690:13: warning: 'CGAL::Nef_polyhedron_3<CGAL::Epeck>::Triangulation_handler2<CGAL::Projection_traits_yz_3<CGAL::Epeck> >::supporting_plane' will be initialized after [-Wreorder]
     Plane_3 supporting_plane;
             ^~~~~~~~~~~~~~~~
.../include/CGAL/Nef_polyhedron_3.h:689:28: warning:   'CGAL::Nef_polyhedron_3<CGAL::Epeck>::Halffacet_const_handle CGAL::Nef_polyhedron_3<CGAL::Epeck>::Triangulation_handler2<CGAL::Projection_traits_yz_3<CGAL::Epeck> >::f' [-Wreorder]
     Halffacet_const_handle f;
                            ^
.../include/CGAL/Nef_polyhedron_3.h:693:5: warning:   when initialized here [-Wreorder]
     Triangulation_handler2(Halffacet_const_handle f) :
     ^~~~~~~~~~~~~~~~~~~~~~
```
You must either reorder the initialization of fields in the constructor, or reorder the declaration of fields in the class itself. The two orders must match.

kintel:
@GilesBathgate I believe all my stability issues were after snap-rounding from Gmpq to float. I don't believe I have any testcases in .nef format lying around. I'm thinking about moving to header-only CGAL, in which case I can start looking at removing workaround for older CGAL versions. That would be a good time to revisit some of my older issues.

GilesBathgate:
@kintel I did my own experiments to "snap" round https://git.io/vb8Uy . It doesn't work when the facets are made of quads and are not axis aligned because snap rounding can cause the loss of coplanarity. I think I could fix it by first triangulating. I wonder if https://doc.cgal.org/latest/Snap_rounding_2/index.html would provide a more robust solution?

kintel:
I believe planarity is a precondition for building Polyhedron_3 polygons, so that also needs to be dealt with. My technique has been to try building CGAL objects, catch precondition exceptions (e.g. coplanarity checks), and then try again using triangulated shapes.

Since "snap rounding" in my case can mean round to float/double, I don't think the `Snap_rounding_2` method would work for me. Snap-rounding can also have fun effects, like introducing self-intersections.

GilesBathgate:
@kintel I think we both have a requirement to convert nef polyhedra into an output mesh file (whether it be stl, amf, off etc) most of which use either float or double as their number type representation. In the specific application of 3D printing, it's probably also desirable to round to the nearest 0.001 units (representing a fairly standard 10 microns resolution used by many FDM printers)  So I would like to know a method to able to convert robustly from Gmpq based mesh to float based mesh without the pitfalls, such as self-intersections and dealing with coplanarity issues. Perhaps @sloriot or @lrineau have some suggestions.

sloriot:
@GilesBathgate Sorry it took me some time to get back to your PR.

I did some benchmark on my own and I have some pretty different results.
I used this [test program](https://gist.github.com/sloriot/a18597ddb33bf5d0572a72e717d3da62) with your [data set](https://gist.github.com/sloriot/74e725e1040dec95a1f6af61b0049fb0) converted to off.

Running it with N=10000000 I have:

```
convert_to_polyhedron 10.9207
convert_nef_polyhedron_to_polygon_mesh 0.721564
```

and if I modify `convert_nef_polyhedron_to_polygon_mesh()` to always retriangulate the faces I have:

```
convert_to_polyhedron 10.8969
convert_nef_polyhedron_to_polygon_mesh 0.735654
```
So except if I have done something wrong the generic version is faster.
If you want I can add an option to the generic function to always retriangulate faces even if they do not have holes.

GilesBathgate:
@sloriot Hi, thanks for taking the time to look at this. In my test I was explicitly calling `CGAL::Polygon_mesh_processing::triangulate_faces()` because I assumed it was necessary, e.g:
 
```c++
for (int i=0; i<N; ++i)
  {
    Polyhedron poly_out;
    CGAL::convert_nef_polyhedron_to_polygon_mesh(nef, poly_out);
    CGAL::Polygon_mesh_processing::triangulate_faces(poly_out);
    if(poly_out.size_of_vertices()!=poly_in.size_of_vertices())
      std::cerr <<"ERROR2\n";
  }
```
https://github.com/CGAL/cgal/commit/580486a1f9700d10d104e396312593683c605640

You didn't mention whether or not you applied https://github.com/CGAL/cgal/pull/2598 or whether the output when converted back to OFF retains holes. In my tests, I was still explicitly calling  `triangulate_faces()` even with that patch applied, which might have been a mistake.

In the original implementation by Peter Hachenberger, the facets of Nef polyhedron are only triangulated when there is more than one facet cycle (i.e a hole) and when the facet is not already a triangle. If that information can be known by the time it is converted to polygon soup, then I think it would be desirable to have this done automatically.

sloriot:
My bench is using the master branch and includes the triangulation. As I said I also modified the code to triangulate all the faces (second bench) and the time was slightly up. As proposed I can add an option to the generic function to do the retriangulation all the time.

lrineau:
I suggest you both share your bench code, to compare.

GilesBathgate:
@lrineau Yes, I had previously shared my benchmark code, (but in hindsight made it difficult to find) it's here: https://git.io/vb2wj

I think I will see if I get the expected result by removing line 74.

@sloriot I think I understand now.  I don't think its necessary to add an option to the generic function to do the re-triangulation all the time. I can't envisage a use case where that would be necessary, as it seems that when converting `CGAL::Polyhedron_3`  to STL (which must consist only of triangles) that the assertion that there are only 3 vertices in the facet was always true. See: https://git.io/vb2wg

What I would like to mention is that the documentation: (https://doc.cgal.org/latest/BGL/group__PkgBGL.html#ga7cf62f17d3bb060519e682ad9499c7d4) is a bit misleading as it says "the output is not triangulated." Perhaps it should say that the output may contain additional triangles to maintain the correct topology or words to that effect. 

I think within the Nef_3 documentation making it clearer that this method can be used to convert to any MutableFaceGraph (which includes `CGAL::Polyhedron_3`) would be useful.

GilesBathgate:
@sloriot I've merged cgal master into my performance-test3 branch, and removed the explicit re-triangulation step: https://github.com/GilesBathgate/cgal/commit/e9ab08b6a9459bfae60f93c984ee8e1d3ff9bdd0 The output isn't what I expected, https://git.io/vb2Xr half the triangles are missing, is this a bug?

GilesBathgate:
@sloriot I am now seeing the same performance as you. I am not sure why my initial results were different:

https://github.com/GilesBathgate/cgal/commit/d34e8a975aed8bb9b2279e64aee4c3c72a07ddd1
```
convert_nef_polyhedron_to_polygon_mesh 9.31643
convert_to_polyhedron 13.6901
```

What do you think about the following convenience function added to `CGAL\Nef_polyhedron_3.h`

```c++
 template<typename Polyhedron>
 void convert_to_triangulated_polyhedron(Polyhedron& P) const {
    CGAL::convert_nef_polyhedron_to_polygon_mesh(*this, P);
    CGAL::Polygon_mesh_processing::triangulate_faces(P);
 }
```

sloriot:
Note that the doc has been updated when I fixed the function. See the output [here](https://cgal.geometryfactory.com/CGAL/doc/master/BGL/group__PkgBGL.html#ga7cf62f17d3bb060519e682ad9499c7d4). Thinking about it, adding the option to triangulate face without holes is interesting as it will be faster than calling the function. I'll prepare a PR for that.

sloriot:
See https://github.com/CGAL/cgal/pull/2682

sloriot:
Final results:
This [test program](https://gist.github.com/sloriot/a18597ddb33bf5d0572a72e717d3da62) with your [data set](https://gist.github.com/sloriot/74e725e1040dec95a1f6af61b0049fb0) converted to off. With the extra boolean added to the generic function + your modified version of `Nef_polyhedron_3.h`:

```
convert_to_polyhedron 12.9375
convert_nef_polyhedron_to_polygon_mesh(true) 9.94739
convert_nef_polyhedron_to_polygon_mesh 4.78685
```

PS: I realized that my previous benchmark was not testing much as I forgot to put the file in the same dir.

GilesBathgate:
@sloriot I am now finding that the new method is slower than seperate calls to `convert_nef_polyhedron_to_polygon_mesh` and `triangulate_faces`

```
convert_nef_polyhedron_to_polygon_mesh 46.7653
convert_nef_polyhedron_to_polygon_mesh + triangulate_faces 56.957
convert_nef_polyhedron_to_polygon_mesh(true) 74.6132
convert_to_polyhedron 90.3888
```
The benchmark code is here:  https://github.com/GilesBathgate/cgal/commit/ada918f5b61162df1aceaafb7940a9a2197bb5f2

sloriot:
you are missing some `timer.reset()` calls.

GilesBathgate:
@sloriot I've removed the STL exporting bits of the code to make it clearer, but I am sorry I can't see where I need to add additional `timer.reset()` calls, there is one before every `timer.start()`, 4 in total. https://github.com/GilesBathgate/cgal/blob/2111059e688707cd31680634d38c28b3a675381e/Nef_3/examples/Nef_3/interface_polyhedron.cpp

sloriot:
Sorry you're right, for some reason the output from github was misleading the first time I looked.

GilesBathgate:
@sloriot I think it must be because of this optimisation regarding quads: https://github.com/cgal/cgal/blob/master/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h#L106

GilesBathgate:
@sloriot I added a further optimisation, which is to triangulate quads with trivial triangulation and this improves the performance as I expected.

```
convert_nef_polyhedron_to_polygon_mesh 46.9208
convert_nef_polyhedron_to_polygon_mesh + triangulate_faces 56.9069
convert_nef_polyhedron_to_polygon_mesh(true) 75.0864
convert_to_polyhedron 51.6953
```

`convert_to_polyhedron` is now the fastest of the three that produce the correct output topology.

GilesBathgate:
@sloriot I have realised that my latest changes do not handle convex quads as can be seen here: https://github.com/GilesBathgate/cgal/commit/ce2b3cc10750ca1a982311395aa0cdb46d5736cd I will address this and update you on any performance changes.

sloriot:
@GilesBathgate I've updated the branch of my PR to have the special cases for boundary of size 3 and 4

GilesBathgate:
@sloriot Your code now produces an error:
```terminate called after throwing an instance of 'CGAL::Precondition_exception'
Expr: is_polygon_soup_a_polygon_mesh(polygons)
File: /home/giles/cgal/Installation/lib/cmake/CGAL/../../../../Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
Line: 216
Explanation: Input soup needs to be a polygon mesh!
Aborted (core dumped)
```

Test code: https://github.com/GilesBathgate/cgal/blob/e2354996a67cf1c370b130ca61f92f16448a5133/Nef_3/examples/Nef_3/interface_polyhedron.cpp

sloriot:
@GilesBathgate sorry, I copy-paste code too fast. It should be fine now.

GilesBathgate:
@sloriot Cool. Here are the latest results:

```
convert_nef_polyhedron_to_polygon_mesh 47.2151
convert_nef_polyhedron_to_polygon_mesh + triangulate_faces 57.32
convert_nef_polyhedron_to_polygon_mesh(true) 57.6513
convert_to_polyhedron 57.8061
```

I don't understand why doing `triangulate_faces` separately is still a touch faster, but there is so little difference between them all now that I don't think its worth worrying. @lrineau I think this PR can now be closed in favour of #2682

lrineau:
Replaced by PR #2682.

@GilesBathgate, @sloriot: Thanks to both of you for the work you have done on benches, analysis, and improvements.

PR #2617, opened 2017-11-20
Replace CGAL::Open_hash by boost::unordered_set

afabri:
## Summary of Changes

Replace a handwritten hash map that gets rehashed during the sweep in order to have an "optimal" filling ratio. 
At least my benchmarks show that there is a performance gain, but we have to confirm that on a broader base of test cases. 
@efifogel - any comment on that? I guess the code was written when `boost::unordered_set` didn't exist. 

## Release Management

* Affected package(s): Arrangement_2


Sweep_line_2/include/CGAL/Sweep_line_2/Sweep_line_curve_pair.h
```diff
@@ -220,4 +224,20 @@ class random_access_input_iterator
...
+namespace boost {
+  
+  template <class Subcurve_>
+  struct hash<CGAL::Curve_pair<Subcurve_> >
+    :  public CGAL::unary_function<CGAL::Curve_pair<Subcurve_>, std::size_t>  {
+
+    std::size_t operator()(const CGAL::Curve_pair<Subcurve_>& cp) const
+    {
+      return boost::hash<std::pair<Subcurve_*, Subcurve_*> >()(cp.pair());
+    }
```
lrineau: The documentation of `boost::hash` does not say it is allowed to specialize it. What is documented is [the creation of an overload of  `hash_value`][*], found by ADL.

[*]: http://www.boost.org/doc/libs/1_65_1/doc/html/hash/custom.html

Sweep_line_2/include/CGAL/Sweep_line_2/Sweep_line_curve_pair.h
```diff
@@ -220,4 +224,20 @@ class random_access_input_iterator
...
+
+  };
+  
+} // namespace std
```
sloriot: the namespace ending here is boost

lrineau:
Compilation error:
https://travis-ci.org/CGAL/cgal/jobs/305214878#L2726

sloriot:
There are some compilation issues in `BSO_2` and `Minkowski_sum_2` packages reported by Travis:
`home/travis/build/CGAL/cgal/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_agg_op_sweep.h:125:29: error:  no member named 'resize'`

sgiraudot:
You should use `CGAL::unordered_set` (`#include <CGAL/unordered.h>`) that will use either the `std` or the `boost` version depending on what's available.

afabri:
`CGAL::cpp11::unordered_set` ?

sgiraudot:
Oh yes, it's in the `cpp11` namespace, I forgot.

I did not think about the `hash` part, but most probably yes (we can have a `CGAL::cpp11::hash` that does the same thing).

I think the benefit is the same as for all of our `CGAL::cpp11` wrappers (`array`, `tuple`, etc.), is that people using a C++11 compliant compiler will benefit from the eventual improvements in the `std` implementations (@lrineau what do you think?).

lrineau:
> I think the benefit is the same as for all of our `CGAL::cpp11` wrappers (`array`, `tuple`, etc.), is that people using a C++11 compliant compiler will benefit from the eventual improvements in the std implementations (@lrineau what do you think?).

I agree. For `CGAL::cpp11::unordered_set`, as Andreas said, that means we have to suppose the extension of `boost::hash` (via the `hash_value` function), *and* the extension of `std::hash` (via the specialization of it), or pass explicitly the `Hash` template parameter.

@sgiraudot You added `CGAL::cpp11::unordered_set` and -`map` in CGAL, and you are the sole user of it currently, in `Point_set_processing_3/include/CGAL/grid_simplify_point_set.h`. What do you recommend?

sgiraudot:
It would be nice to have a `CGAL::cpp11::hash` that does the `boost`/`std` switch, but if I understand correctly, that is a bit more complicated (see [this discussion on Stack Overflow](https://stackoverflow.com/questions/14863848/use-boosthash-value-to-define-stdhash-in-c11)). I'm not sure it is possible to offer one wrapper that fits both behaviors.

Pass explicitly the `Hash` template parameter would solve the issue but it's a bit annoying for users that they can't use the `hash` extension…

afabri:
Implementing `std::hash` is trivial as one can forward it to the same member function as `boost::hash`.

afabri:
I would like to stick to `boost::unordered_map` and `boost::hash`, as the latter provides overloads for what we need, namely `hash<std::pair<T*, T*> >`.   Even [cppreference.com](http://en.cppreference.com/w/cpp/utility/hash) refers to the extensions offered by boost.
@lrineau can we proceed with this PR?

lrineau:
There first thing I see in this PR is the summary at the end:
spectacle t56448

Don't you see the message about the branch having conflicts? Sure I will proceed with this PR once it is testable, like any other PR.

lrineau:
We are also waiting for the comments/review of @efifogel, that you asked for at the beginning of the pull-request:
> @efifogel - any comment on that? I guess the code was written when boost::unordered_set didn't exist.

sloriot:
@efifogel Could you have a look at the branch and let us know if you approve the change?

lrineau:
This PR now conflicts with recent changes in `master`. As it has only four commits, a rebase can probably be done with too much pain.

lrineau:
conflicts

sloriot:
Rebased

lrineau:
Tested in [CGAL-4.12-Ic-142].

  [CGAL-4.12-Ic-142]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-142.shtml

PR #2626, opened 2017-11-23
PMP - fix remove_self_intersections() by removing BOOST_FOREACH

janetournois:
## Summary of Changes

In `PMP/repair.h`, we had to replace `BOOST_FOREACH()` with a c++11 for loop to avoid a seg fault.

For some reason, the modified Polyhedron gets corrupted when
`BOOST_FOREACH()` loops are used here, even though these loops are
not executed (because `border_edges_found` is false)

Note this crash only concerns visual studio 2015 (tested with visual 2013, clang, and g++), with `-O2` and not `-O1`.
I am using boost 1.59

edit : Note also that I did not have to change all the `BOOST_FOREACH` in the file, changing those corresponding to code modifying the mesh is enough

## Release Management

* Affected package : PMP
* Platform : visual 2015

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
@@ -1430,7 +1430,6 @@ bool remove_self_intersections_one_step(TriangleMesh& tm,
       // remove small border cycle that are imposing a self-intersection
       // that could not be fixed if kept. (This also prevent small island
       // if a small hole is incident to the faces to be removed).
-      if (border_edges_found){
         std::set<halfedge_descriptor> cycles;
         std::set<halfedge_descriptor>  boundary_set;
         BOOST_FOREACH(halfedge_descriptor h, boundary_hedges)
@@ -1472,7 +1471,6 @@ bool remove_self_intersections_one_step(TriangleMesh& tm,
           }
           tmp.swap(boundary_hedges);
         }
-      }
 
       // do not remove edges on the boundary of the selection of faces,
       // nor its vertices
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
@@ -1288,7 +1288,7 @@ struct Is_selected{
   {}
 
   friend bool get(Is_selected is, Descriptor d){
-    return is.selection.count(d);
+    return (is.selection.count(d) == 1);
   }
 
   friend void put(Is_selected is, Descriptor d, bool b){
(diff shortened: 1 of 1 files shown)


lrineau:
That is only for C++11 or later. As far as I know, `PMP/repair.h` is supposed to support all our supported compilers, including those using C++<11.

janetournois:
Actually the issue #2627 suggests to add a `CGAL_FOREACH()` function that would use `BOOST_FOREACH()` in the general case, and the c++11 `for` loop for platforms that meet the issue.
Maybe we should do that first.
But for now, this code only available in the demo.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
```diff
@@ -1511,20 +1518,20 @@ bool remove_self_intersections_one_step(TriangleMesh& tm,
...
 
     /// now get one halfedge per hole
     std::set<halfedge_descriptor> visited;
-    BOOST_FOREACH(halfedge_descriptor h, boundary_hedges)
+    for(halfedge_descriptor h : boundary_hedges)
     {
       if (visited.insert(h).second)
       {
         one_halfedge_per_border.push_back(h);
-        BOOST_FOREACH(halfedge_descriptor hh, halfedges_around_face(h, tm))
```
lrineau: That `BOOST_FOREACH` is nested... Would it be sufficient to convert only that one?

sloriot:
I suggest that we wait to see if #2634 make it into master, in such a case we would simply ignored this PR.

lrineau:
@janetournois: This PR modifies more that just the for-loops. See [the diff]. Is that wanted?

If it is, we cannot just ignore this PR in favor of #2634.

[the diff]: https://github.com/CGAL/cgal/pull/2626/files/9b00be7d7bddee92e6c97f7af089aa0b0a52cc25

janetournois:
you are right, the only commit about for-loops is 0fe55d3.

This PR also fixes a warning and slighlty improves the code. I suggest that I wait for #2634 to be merged, and then make a new PR out of this one with the 3 other commits (of lower importance, btw)

lrineau:
Do we want to fix that bug only in `master` (CGAL-4.12)?

sloriot:
4.12 and as stated above #2634 might replace it

lrineau:
But that means that the bug in `<CGAL/Polygon_mesh_processing/repair.h>` will only be fixed in `master`, and not in release branches. Is that wanted?

sloriot:
It is inside a function only used in the demo (not documented).

lrineau:
Replaced by #2634.

sloriot:
#2711 contains the 3 commits that there not related to the for loops.

PR #2628, opened 2017-11-24
Surface_mesh: Derivation, boost::graph_traits and boost::property_map

afabri:
## Summary of Changes

Add an example that shows that one has to provide specializations for `boost::graph_traits` and `boost::property_map` so that BGL-style algorithms work on a class derived from `Surface_mesh`.

## Release Management

* Affected package(s): Surface_mesh   (no code change, only an example)
* Issue(s) solved (if any): Question was raised by a user in a private mail

### DIFF
--- a/Surface_mesh/examples/Surface_mesh/sm_derivation.cpp
+++ b/Surface_mesh/examples/Surface_mesh/sm_derivation.cpp
@@ -1,0 +1,56 @@
+#include <iostream>
+
+#include <CGAL/Simple_cartesian.h>
+#include <CGAL/Surface_mesh.h>
+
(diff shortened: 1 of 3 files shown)


Surface_mesh/examples/Surface_mesh/sm_derivation.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace boost {
+
+  template <>
+  struct graph_traits<My::Mesh>
+    : public boost::graph_traits<My::Mesh::Base>
+  {};
+
+  template <typename T>
+  struct property_map<My::Mesh, T>
```
sloriot: why inheritance not sufficient here?

PR #2630, opened 2017-11-25
fix sudo make install error because of incorrect file name

jasjuang:
If you look in https://github.com/CGAL/cgal/tree/master/Installation, you will realize the file "CHANGE" does not exist, hence if `sudo make install` there will be errors like

```
-- Installing: /home/ubuntu/cgal/install/share/doc/CGAL-4.12-I-900/AUTHORS
CMake Error at Installation/cmake_install.cmake:41 (file):
  file INSTALL cannot find "/home/ubuntu/cgal/Installation/CHANGES".
  cmake_install.cmake:42 (include)


Makefile:107: recipe for target 'install' failed
```

Change "CHANGE" to "changes.md" solves the problem

### DIFF
--- a/Installation/CMakeLists.txt
+++ b/Installation/CMakeLists.txt
@@ -858,7 +858,7 @@ create_CGALconfig_files()
 # DESTINATION option is mandatory; skipping it breaks CPack!
 
 if(CGAL_INSTALL_DOC_DIR)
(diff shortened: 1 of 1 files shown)


lrineau:
This PR is a follow-up to the PR #2603.

Thanks @jasjuang for pointing out the issue.

jasjuang:
@lrineau Would you consider adding a `make install` command in travis.yml so this kind of error will be caught by the CI?

lrineau:
Good idea!

I think our Travis CI test should even use an installed version of CGAL as `CGAL_DIR`, to check that usage. @maxGimeno, could you please implement that?

maxGimeno:
done in #2636

PR #2632, opened 2017-11-26
Kd_tree::search_any_point: don't check outer_range_contains

mglisse:
## Summary of Changes

This is purely a performance patch. For my use case, in 2D, using a custom point type with a custom distance (sup norm), this is measurably faster. And intuitively it makes sense that checking if a subtree is entirely contained in the query is more useful for search (report all points) than search_any (report one arbitrary point). But if someone else uses search_any, they are welcome to benchmark this on their testcase. It is certainly possible to find types and distributions for which the patch makes performance worse, but I expect them to be rather contrived.

## Release Management


lrineau:
Andreas, I request a review from you, two days ago. Please tell us if you cannot do the review.

PR #2634, opened 2017-11-27
Add CGAL_FOREACH

### LINKED ISSUE #2627: Issue with BOOST_FOREACH on Visual Studio 2015
## Issue Details

The use of `BOOST_FOREACH()` seems to corrupt the data when used while modifying the mesh.

In PR #2626, replacing the use of `BOOST_FOREACH()` by c++11 range-"for" loop in the lines corresponding to code modifying the mesh was enough to fix the seg fault.

We had already met that issue in PR #2304, with visual 2015 and 2017.

We could have a `CGAL_FOREACH()` that uses the c++11 for loop for visual >= 2015, and `BOOST_FOREACH()` otherwise.

## Source Code

This is the way our test crashes :
```
void fct(TM& tm)
{
  is_valid(tm); //crashes here
  if(some_boolean_that_evaluates_to_false_in_our_case)
  {
     // code using BOOST_FOREACH, not executed
   }
}
int main()
{
  TM tm;
  is_valid(tm);
  fct(tm);
} 
```
## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 64 bits
* Compiler: visual 2015
* Release or debug mode: Release with -O2 (Release with -O1 is fine)
* CGAL version: master
* Boost version: 1.59

cc @sloriot

sloriot:
Introduces CGAL_FOREACH that uses c++11 range loop for if available or BOOST_FOREACH.
There is a tiny variation as `std::pair<T,T>` is binded on the fly to `Iterator_range<T>`.

The files to be reviewed is `STL_Extension/include/CGAL/foreach.h`.

In passing 2bb0e40  and e34ca45 are bug-fixes not related to the change.

Fixes #2627 
Replaces #2626


STL_Extension/include/CGAL/foreach.h
```diff
@@ -0,0 +1,73 @@
...
+}
+
+template<class T>
+constexpr T&& make_range( T&& t )
+{
+  return std::forward<T>(t);
+}
+
+} }
+
```
lrineau: @mglisse Would you mind reviewing this file with the four overloads of `CGAL::for_range_loop::make_range`?

I am not sure but I think:
1. The three overloads about `std::pair` could be replaced by one taking its parameter by copy.
2. Maybe the fourth overload (generic) of `make_range` is not optimal.

STL_Extension/include/CGAL/foreach.h
```diff
@@ -0,0 +1,73 @@
...
+#define CGAL_FOREACH(A,B) for(A : CGAL::for_range_loop::make_range(B))
+#else
+#include <CGAL/foreach.h>
```
mglisse: Did you mean to include a boost header instead here?

sloriot: Right, I did not exclude the header from `sed` input files

STL_Extension/include/CGAL/foreach.h
```diff
@@ -0,0 +1,73 @@
...
+
+#define CGAL_FOREACH(A,B) for(A : CGAL::for_range_loop::make_range(B))
+#else
+#include <boost/foreach.hpp>
+#define CGAL_FOREACH(A,B) BOOST_FOREACH(A, B)
+#endif
+#else
+/// \ingroup  PkgStlExtension
+/// If the version of `boost` is at least 1.51 and the compiler used support range-based for loops
+/// introduced with \cpp11, use it; otherwise fallback onto `CGAL_FOREACH`.
```
mglisse: Overeager sed again?

mglisse:
The generic version is dangerous. If I have a function f that returns a vector (by value), CGAL_FOREACH(whatever,f()) will destruct the vector before even calling begin().

mglisse:
Special handling for std::pair is a hack to delay the inevitable: stop using std::pair as a range. I can understand not being in a hurry to fix it though.

lrineau:
> The generic version is dangerous. If I have a function f that returns a vector (by value), `CGAL_FOREACH(whatever,f())` will destruct the vector before even calling begin(). It would be safer to have `T make_range(T&&)` and `T& make_range(T&)`, if you keep a make_range.

What do you mean?

`f()` will be a temporary: a r-value of type `vector<T>`, so the generic `make_range` will be instantiated with `vector<T>&&`, and return a `vector<T>&&` as well. I do not see why you say the temporary will be destroyed before the call to `begin(..)`.

mglisse:
Try it on an example?
`for(lhs:rhs)` starts by doing `auto&&range=rhs;`. If rhs is a prvalue (a temporary object), range binds to it and extends its lifetime. If rhs is an xvalue (rvalue reference), range binds to the same object as the xvalue, but no lifetime extension occurs. The vector is destroyed. Next, `for` proceeds to initialize begin, end, and do the looping, on an object that is already dead.

lrineau:
Travis [detected an error](https://travis-ci.org/CGAL/cgal/jobs/308454571#L3074):
```
/home/travis/build/CGAL/cgal/Arrangement_on_surface_2/test/Arrangement_on_surface_2/test_spherical_removal.cpp:88:16: error: 
      unexpected type name 'Halfedge_handle': expected expression
  CGAL_FOREACH(Halfedge_handle hh, arr.halfedge_handles())
               ^
/home/travis/build/CGAL/cgal/Arrangement_on_surface_2/test/Arrangement_on_surface_2/test_spherical_removal.cpp:111:45: error: 
      expected '(' for function-style cast or type construction
    CGAL_FOREACH(Arrangement_2::Face_handle fh, arr.face_handles())
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
```

STL_Extension/include/CGAL/foreach.h
mglisse: > The three overloads about std::pair could be replaced by one taking its parameter by copy.

Forgot to comment about this part but indeed, the last 2 std::pair overloads are redundant, the first one (`const&`) should suffice.

sloriot:
@mglisse I think my last commit fixes the issue you mentioned. Do you see a better solution? (I cannot remove the extra overloads for `std::pair` as it does not pass the testsuite. I'll investigate).

mglisse:
> I think my last commit fixes the issue you mentioned.

Yes.

> Do you see a better solution?

Stop using std::pair as ranges so we don't need the make_range helper at all? Stop supporting C++03?

> (I cannot remove the extra overloads for std::pair as it does not pass the testsuite. I'll investigate).

You mean removing the `&` and `&&` versions, keeping `const&`? I am curious what might cause it to fail...

sloriot:
Here is the problematic call:
```
CGAL_FOREACH(Cell_handle& ch,
  std::make_pair(cells.begin(), cells.end()))
```

mglisse:
Ah yes, of course, once you have a fully generic make_range overload, it is more specialized than the `const pair&` overload for a pair rvalue. We could use sfinae to disable the generic version for pairs, or dispatch some other way, but I guess what you have now is simpler. Let's just hope that nobody writes a function returning a const pair, or you'll need the extra `const&&` overload...

By the way, could we start replacing those std::make_pair with boost::make_iterator_range? These are the easy cases where the use of std::pair is not baked in the API.

mglisse:
> replace std::make_pair with CGAL::make_range in foreach loops

thanks 👍

lrineau:
conflicts

sloriot:
Rebased

sloriot:
There is the following error reported by travis with gcc 4.8.4.
I don't understand why it is ambiguous since `get_chill()` returns a lvalue reference.

```
/home/travis/build/CGAL/cgal/Classification/examples/Classification/../../include/CGAL/Classification/Sum_of_weighted_features_classifier.h: In member function ‘bool CGAL::Classification::Sum_of_weighted_features_classifier::load_configuration(std::istream&, bool)’:
/home/travis/build/CGAL/cgal/Classification/examples/Classification/../../include/CGAL/Classification/Sum_of_weighted_features_classifier.h:736:5: error: call of overloaded ‘make_range(boost::property_tree::basic_ptree<std::basic_string<char>, std::basic_string<char> >&)’ is ambiguous
     CGAL_FOREACH(boost::property_tree::ptree::value_type &v, tree.get_child("classification.features"))
     ^
/home/travis/build/CGAL/cgal/Classification/examples/Classification/../../include/CGAL/Classification/Sum_of_weighted_features_classifier.h:736:5: note: candidates are:
/usr/local/include/CGAL/foreach.h:55:14: note: constexpr T& CGAL::for_range_loop::make_range(T&) [with T = boost::property_tree::basic_ptree<std::basic_string<char>, std::basic_string<char> >]
 constexpr T& make_range(T& t)
              ^
/usr/local/include/CGAL/foreach.h:61:13: note: constexpr T CGAL::for_range_loop::make_range(T&&) [with T = boost::property_tree::basic_ptree<std::basic_string<char>, std::basic_string<char> >&]
 constexpr T make_range(T&& t)
             ^
```

mglisse:
> There is the following error reported by travis with gcc 4.8.4.

Looks like a bug in gcc-4.8, it compiles fine with 4.9.

lrineau:
Travis shows a compilation issue: https://travis-ci.org/CGAL/cgal/jobs/325369172#L1550

What is more, [CGAL-4.12-Ic-134][.] seems rather red on older Windows, because of the lack of support of `constexpr`.

[.]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-134/

lrineau:
On my machine, `/usr/include/boost/config/compiler/visualc.hpp` contains:
```c++
// C++11 features supported by VC++ 14 update 3 (aka 2015)
//
#if (_MSC_FULL_VER < 190024210)
#  define BOOST_NO_CXX14_VARIABLE_TEMPLATES
#  define BOOST_NO_SFINAE_EXPR
#  define BOOST_NO_CXX11_CONSTEXPR
#endif
```
That means that `constexpr` was supported very late my MSVC: only in 2015 update 3. That explains why the code does not compile for versions of MSVC that know c++11 range-based for, but not `constexpr`.

lrineau:
Should this PR be removed?

sloriot:
well the issue is still here but I need to look again to see how to fix the remaining issues.

sloriot:
No longer needed with the move to C++14

PR #2638, opened 2017-11-28
Add experimental autorefinement code

sloriot:
Add some undocumented code that is modifying the internal implementation of the corefinement. The code is experimental as it is not yet working in all cases but is in a particular scenario.


Property_map/include/CGAL/property_map.h
```diff
@@ -456,8 +456,52 @@ struct Default_property_map{
...
 };
 
+/// \ingroup PkgProperty_map
+/// Read-write Property map turning a set (such a `std::set`,
+/// `boost::unordered_set`, `std::unordered_set`) into a property map
+/// associating a Boolean to the value type of the set. The function `get` will
+/// return `true` if the key is inside the set and `false` otherwise. The `put`
+/// function will insert an element in the set if `true` is passed and erase it
+/// otherwise.
+/// \cgalModels `ReadWritePropertyMap`
```
lrineau: Should have been documented as a small feature...

sloriot: I can hide it in the documentation if you want.

lrineau: I will merge the PR anyway. And I agree with the documentation.

PR #2643, opened 2017-11-29
Use a tight count for the number of edges

sloriot:
This is a real issue for example when using  `Surface_mesh` with some garbage edges not collected.

cc @afabri


lrineau:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-107.shtml#Surface_mesh_simplification_Examples

PR #2647, opened 2017-12-01
Polyhedron demo : export surface_mesh_selection_item for external plugins

janetournois:
## Summary of Changes

Export the selection item for `Surface_mesh`.
This is needed for external plugins using the selection plugin with `Surface_mesh_item`

## Release Management

* Affected package(s): CGAL demo


lrineau:
Only in CGAL-4.12?

janetournois:
Having it in `master` is enough for me.
If you think someone might create a new external plugin using a previous version of CGAL, then we should add it to 4.11 (when did `Surface_mesh` make it to the demo?)

The patch is light, it shouldn't hurt!

PR #2649, opened 2017-12-01
Regroup PMP named parameters with BGL named parameters

sgiraudot:
## Summary of Changes

This is a first step for organizing and generalizing Named Parameters in CGAL. There was already some duplication of code between PMP and BGL: all the code is merged in BGL (for genericity), only the specific parameters macro remain in PMP (and also some `#define` and other small tricks for backwards compatibility).

*  The few additional code that was in `CGAL/Polygon_mesh_processing/internal/named_function_params.h` but not in `CGAL/boost/graph/named_function_params.h` is integrated in it (it concerns the `default` behavior)
* `pmp_bgl_named_params` disappears, everything is now `cgal_bgl_named_params`
* the helpers functions that were either in `CGAL/boost/graph/named_function_params.h` or `CGAL/Polygon_mesh_processing/internal/named_params_helper.h` are regrouped in a new file `CGAL/boost/graph/named_params_helper.h`
* Both versions of `parameters_interface.h` from BGL and PMP are kept as is. That is debatable. In my mind, packages that want to use named parameters should create a similar file and include it in the right places.

(Again, this is only a first step to be able to used Named Parameters in other packages without adding a dependency to PMP and without duplicating code. We can still improve by increments when we need to. Reorganizing the documentation at some point might also be a good idea.)

There is no change of public API and as far as I tested, it did not break anything on PMP.

## Release Management

* Affected package(s): BGL, PMP


BGL/include/CGAL/boost/graph/named_function_params.h
```diff
@@ -91,6 +92,7 @@ using boost::visitor;
 #define CGAL_add_named_parameter(X, Y, Z)            \
   enum X { Y };
 #include <CGAL/boost/graph/parameters_interface.h>
+#include <CGAL/Polygon_mesh_processing/internal/parameters_interface.h>
```
sloriot: you don't want to introduce a dependy to PMP in BGL

afabri: When I discussed this with Simon we said that we should have a dedicated package for named parameters.

sgiraudot: So I should move the `parameters_interface.h` content to the BGL one?

sloriot: You can move them all into the BGL files now that the np classes are the same.

sloriot: @afabri I don't think there is a need for a dedicated package. NP are from BGL so they are at the correct place. For the moment, each package will have its own NP page documentation as it seems non obvious that they'll have the exact same meaning across packages.

afabri: ok, for putting it into the  BGL package.  It only makes no  sense to make things depend on PMP. We should also do this for Surface_mesh_simplification (I started doinf this in the branch SMS-parallelization-GF).

sgiraudot:
Ok, `parameters_interface.h` is now merged too.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -269,7 +269,7 @@ self_intersections(const TriangleMesh& tmesh
 #ifdef DOXYGEN_RUNNING
                  , const NamedParameters& np)
 #else
-                 , const pmp_bgl_named_params<P,T,R>& np)
+                 , const cgal_bgl_named_params<P,T,R>& np)
```
sloriot: The macro `CGAL_PMP_NP_TEMPLATE_PARAMETERS` can be used instead of `class P, class T, class R`
and `CGAL_PMP_NP_CLASS` can be used instead of `cgal_bgl_named_params<P,T,R>`.

sgiraudot:
I've reorganized the doc with BGL.

MaelRL:
- It would be nice if BGL/include/CGAL/boost/graph/parameters_interface.h was tested. There is some incomplete skeletons in https://github.com/CGAL/cgal/blob/master/BGL/test/BGL/test_cgal_bgl_named_params.cpp and the equivalent in PMP. They should probably be merged with BGL's if the named parameters are regrouped.
- Why is the documentation of the different parameters in PMP (https://github.com/sgiraudot/cgal/blob/8c996bb7887f8999ce0b0ce8a3c45ebf87934442/Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt)? For example, why is the description of `vertex_point_map` which is a BGL named parameter described in PMP? It seems strange that `Subdivision_methods`, which needs a named parameter from BGL, should go look into PMP for the description of a BGL named parameter. This seems even stranger now that the parameters are in /BGL. Shouldn't there be a NamedParameters.txt for each package that defines its named parameters (and if needed, point to other packages' named parameters)?

MaelRL:
In ad947da :
- BGL, PMP, SM, SMS get their own NamedParameters.txt which describe the respective named parameters (referring to BGL's named parameters when needed).
- Proper presentation of named parameters for `edge_collapse` (todo: do it in SMS's user manual too; probably will be done in the parallel branch rather than here).
- Cleaned named parameters (especially PMP) along the way.

Still todo: test the NPs in BGL

lrineau:
Doc issue: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.12-Ic-115/Polygon_mesh_processing.log

sgiraudot:
If I'm not mistaken, these are already fixed by Mael's latest commits.

lrineau:
Travis detected a compilation error:
https://travis-ci.org/CGAL/cgal/jobs/315065507#L4808

lrineau:

PR #2651, opened 2017-12-04
Use readable property map API

sloriot:
## Summary of Changes

The code uses rvalue pmap API rather than readable pmap API

## Release Management

* Affected package(s): Ridges_3


lrineau:

Issue #2639, opened 2017-11-28
Comparison operators of Weighted_point_23


MaelRL:
Implicit conversions have been removed between points and weighted points, but comparison operators (!=, ==, <, etc.) entirely rely on the bare point contained within the weighted point, which create the same issues:
```
WP(P(1,2,3), 10) == WP(P(1,2,3), 0)
```
returning `true` is really misleading. 

I didn't previously miss them since I added some warnings (e.g. https://doc.cgal.org/latest/Kernel_23/classCGAL_1_1Weighted__point__2.html) some time ago, but I don't see any reason why I left them. If `<` is defined by comparing bare points and weights are compared when bare points are equal, it should be fine (?).

Additionally, `equal_[23]` is missing some tests and an overload for weighted points.

MaelRL:
I was looking into fixing #1261 on all triangulations, and I ran into this issue since two weighted points are currently equal regardless of their weight.

The current weighted point comparison operators are (ab)used in quite a few places, for example in the triangulation packages for the location and move functions.

In my opinion, this should all be cleaned to use proper bare points when needed (e.g. the face line tracers should not use the triangulation point type but the bare point type). There is also a great deal of inconsistency between the move functions of T2/T3/P2T2/P3T3 that ought to be fixed along the way. Some minor bugs due to these bad comparison operators too (when moving a point is simply changing its weight, for example).

All in all, it turned into a rather large task...

PR #2658, opened 2017-12-06
Periodic_3_mesh_3, new package

### LINKED ISSUE #570: concept-model differences in Mesh_3
According to the documentation,  [Mesh_complex_3_in_triangulation_3](http://doc.cgal.org/latest/Mesh_3/classCGAL_1_1Mesh__complex__3__in__triangulation__3.html) is a model of [MeshComplexWithFeatures_3InTriangulation_3](http://doc.cgal.org/latest/Mesh_3/classMeshComplexWithFeatures__3InTriangulation__3.html). However, a few functions differ from the concept to the model: for example `number_of_corners()` is called `number_of_vertices_in_complex()` in the model; additionally, overloads with a subdomain index are also missing.

The documentation uses the terminology `corners`, `curve segments` but `vertices_in_complex` is simpler so I am not sure which one should be kept.

### LINKED ISSUE #1927: MeshEdgeCriteria_3 is wrong
http://doc.cgal.org/4.9/Mesh_3/classMeshEdgeCriteria__3.html

That documentation is wrong: the `operator()` is actually never used by Mesh_3, and it is not implemented in the "model" [`Mesh_edge_criteria_3`][*].

  [*]: http://doc.cgal.org/4.9/Mesh_3/classCGAL_1_1Mesh__edge__criteria__3.html

### LINKED ISSUE #2134: Mesh_3 requires more than RegularTriangulationTraits_3 provides
In the documentation, `Mesh_3` [requires](http://doc.cgal.org/latest/Mesh_3/classCGAL_1_1Mesh__triangulation__3.html) a traits class that is a model of `RegularTriangulationTraits_3`. `RegularTriangulationTraits_3` will include `TriangulationTraits_3` (and soon `SpatialSortingTraits_3` as it should), but still uses quite a few additional functors, like `compute_approximate_dihedral_angle_3_object`, `construct_translated_point_3_object`, `compute_area_3_object`, etc. that are not included by `RegularTriangulationTraits_3`.

Since `RegularTriangulationTraits_3` probably shouldn't be polluted with that stuff, should there be a `Mesh_3` traits concept?

### LINKED ISSUE #3011: Mesh_domain_with_features_3: add `add_corner()` and `add_corner_with_context()`
## Issue Details

Feature request: in the class template `CGAL::Mesh_domain_with_features_3`, you may need to add isolated vertices as corners, in the mesh (like the tip of a 3D cone), or register that a corner is incident to specific surfaces (imagine the intersection of a 1D polylines, like a well, with a surface).

(Cc @SLopezBRGM)

### LINKED ISSUE #3040: Delaunay_triangulation_2<Gt> does not inherit Triangulation_2<Gt>
## Issue Details

The default template parameter Tds for Delaunay_triangulation_2<Gt> is different of the default template parameter Tds for Triangulation_2<Gt>. (Specifically, the former does not specify a Triangulation_face_base_2, which then defaults to a version without a Traits class).

This is annoying if I want to build a triangulation that is Delaunay, and then treat it like a non-Delaunay triangulation, without conserving the Delaunay property. Which is the case.

It can be avoided by directly specifying the Tds, but it looks like a (very small) design flaw. And it's not what's documented, also.

## Source Code

typedef Projection_traits_xy_3<Kernel> Gt;

Delaunay_triangulation_2<Gt> delaunay_tr;
// fill the triangulation
Triangulation_2<Gt> tr(delaunay_tr); // Does not compile, obscure templaty error message where a Traits class is actually void

## Environment

* CGAL version: Installed in April 2018 with ubuntu's apt install

### LINKED ISSUE #3175: Re-inserting point in a Periodic_3_regular_triangulation
## Issue Details

Inserting and removing the same point to the triangulation produces the error
`terminate called after throwing an instance of 'CGAL::Assertion_exception'
Expr: cells_with_too_big_orthoball.empty()
/usr/local/include/CGAL/Periodic_3_regular_triangulation_3.h
Line: 329
Aborted (core dumped)
`
Edit: The problem seems to happen when I remove all points from a periodic triangulation and then add *any* point. 

## Source Code

``` c++
#include <CGAL/Periodic_3_regular_triangulation_traits_3.h>
#include <CGAL/Periodic_3_regular_triangulation_3.h>

typedef CGAL::Periodic_3_regular_triangulation_traits_3<K>       Gt;
typedef CGAL::Periodic_3_regular_triangulation_3<Gt>             P3RT3;

int main()
{
  
  P3RT3 PT;

  P3RT3::Weighted_point wp (P3RT3::Bare_point(0.5,0.5,0.5),0.01);
   
  P3RT3::Vertex_handle v = PT.insert(wp);
  PT.remove(v);
  PT.insert(wp);
}
``` 
## Environment

* Operating system: Linux Mint 18.03
* Compiler: g++
* CGAL version: 4.12

MaelRL:
## Summary of Changes

This feature introduces a new package to generate periodic meshes. This is achieved by creating an interface between the existing packages Mesh_3 and Periodic_3_triangulation_3. 

Feature developed with @MoniqueTeillaud, @aympelle .

## Release Management

* Affected package(s): Mainly `Kernel_23`, `Mesh_3`, `Periodic_3_mesh_3`, `Periodic_3_triangulation_3`, `Triangulation_3`
* Issue(s) solved (if any): fix #2134 , fix #570, fix #1927, fix #3040, fix #3011, fix #3175
* Feature/Small Feature (if any): [`Features/Periodic_3_mesh_3`](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Periodic_3_mesh_3)


MaelRL:
Going to need some help with interpreting what is the issue with Travis...

lrineau:
`master` is broken, as for Travis, and you merged `master` is this PR yesterday.

MaelRL:
- I don't see P3M3 being tested in travis, I have added it to .travis/packages.txt. Is there anything else to do ? Can't find anything in the doc.
- https://travis-ci.org/CGAL/cgal/jobs/312901991#L4864 Clang is using GCC's libraries, is it an issue from me or in the travis configuration ?

lrineau:
About the Travis issues...

## Travis detects incorrect headers file ##
From https://travis-ci.org/CGAL/cgal/jobs/312901937#L4671:
```
.. Checking include directives in GPL header files...
Kernel_23/include/CGAL/Exact_kernel_selector.h
Periodic_3_mesh_3/include/CGAL/IO/Medit_IO.h
STL_Extension/include/CGAL/internal/Has_member_visited.h
```
The documentation for that problem is here: https://github.com/CGAL/cgal/wiki/License#including-license-headers-in-package-header-files
**TL;DR:** add lines like that one in the corresponding files:
```C+++
#include <CGAL/license/Periodic_3_triangulation_3.h>
```
Of course, the license-checking header depends on the package. Copy-paste what is used in the header files of the same directory.

## Travis does not test P3Mesh_3 ##
https://github.com/CGAL/cgal/wiki/Directory-Structure-for-Packages#requirements-and-recommendations

**TL;DR:**
```shellsession
$ cd .travis
$ ./generate_travis.sh
```
and then commit/push the modified Travis files. If you have any question about that script `./generate_travis.sh`, @maxGimeno is its author.

MaelRL:
> Travis detects incorrect headers file

It wasn't a license issue (except for one), just a bad copy paste of SPDXwhatever tags, it's already fixed in 3bd4d74.

> Travis does not test P3Mesh_3

Thanks. I was expecting that to be in the testing section of the wiki...

lrineau:
>> Travis does not test P3Mesh_3
>> https://github.com/CGAL/cgal/wiki/Directory-Structure-for-Packages#requirements-and-recommendations

> Thanks. I was expecting that to be in the testing section of the wiki...


I sound reasonable too. I suggest we add at least a link. Where would it fix best in testing section?

MaelRL:
> I sound reasonable too. I suggest we add at least a link. Where would it fix best in testing section?

I have just discovered that the repo search actually only shows code results initially, and you have to click 'wiki' at the top of the results page to get the results in the wiki... That's why I couldn't find anything in the results.

From what I see, this wiki page is the only place where travis is mentioned. I think there should be a section "Automated testing" that mentions travis and the nightly test suite.

I have added some words there: https://github.com/CGAL/cgal/wiki/Testing#automated-testing.

MaelRL:
`set(CMAKE_CXX_STANDARD 14)` (left accidentally in P3T3/test) made Travis use gcc's libs while compiling with Clang: https://travis-ci.org/CGAL/cgal/jobs/312901991#L4864...

lrineau:
Indeed, the last commit was flagged as OK by Travis.

lrineau:
Doc issue: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.12-Ic-115/Periodic_3_mesh_3.log
```
.../CGAL-4.12-Ic-115/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt:272: Warning: unable to resolve link to `CGAL::Mesh_domain_with_polyline_features_3::add_features()' for \link command
.../CGAL-4.12-Ic-115/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt:273: Warning: unable to resolve link to `CGAL::Mesh_domain_with_polyline_features_3::add_corners()' for \link command
```

lrineau:
The [testsuite of CGAL_4.12-Ic-115][*] also has several new errors, in the following lines:
- Envelope_3
- Envelope_3_Examples
- Generator
- Generator_Examples
- Mesh_3
- Mesh_3_Examples
- Periodic_3_mesh_3
- Periodic_3_mesh_3_Examples

[*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-115.shtml

afabri:
Envelope_3 is related to the arrangement branch.

MaelRL:
`set(CMAKE_CXX_STANDARD 14)` (left accidentally in P3T3/test) made Travis use gcc's libs while compiling with Clang: https://travis-ci.org/CGAL/cgal/jobs/312901991#L4864...

lrineau:
In integration (commit 7ab281e0400563353b8252513863caf0a557461e), CTest told me:
```
The following tests FAILED:
        3293 - execution___of__test_implicit_shapes_with_protection (Failed)
```

MaelRL:
Can't reproduce (neither in that branch nor integration):
```
[mrouxell@kandinsky build]$ ctest -R "test_implicit_shapes_with_protection"
Test project /home/mrouxell/cgal_quinquies/build
    Start 3138: Periodic_3_mesh_3_Tests_SetupFixture
1/4 Test #3138: Periodic_3_mesh_3_Tests_SetupFixture ...................   Passed    0.09 sec
    Start 3140: compilation_of__test_implicit_shapes_with_protection
2/4 Test #3140: compilation_of__test_implicit_shapes_with_protection ...   Passed   43.10 sec
    Start 3141: execution___of__test_implicit_shapes_with_protection
3/4 Test #3141: execution___of__test_implicit_shapes_with_protection ...   Passed    5.33 sec
    Start 3139: Periodic_3_mesh_3_Tests_CleanupFixture
4/4 Test #3139: Periodic_3_mesh_3_Tests_CleanupFixture .................   Passed    0.07 sec

100% tests passed, 0 tests failed out of 4
```

Can you please give me the log, and possibly with:
```
#define CGAL_MESH_3_VERBOSE
#define CGAL_MESHES_DEBUG_REFINEMENT_POINTS
#define CGAL_MESH_3_DEBUG_FACET_CRITERIA
#define CGAL_MESH_3_DEBUG_CELL_CRITERIA
#define CGAL_MESH_3_PROFILING
#define CGAL_MESH_3_DEBUG_SLIVERS_EXUDER
#define CGAL_MESH_3_PROTECTION_DEBUG 1111
```
added at the top of the test ?

Related: I was going to add those macros to try and understand what happens on that Fedora 32 plateform that segfaults, is there a maximum log size on the test suite ?

lrineau:
Even without those macros, the test is rather verbose:
```
3293: epsilon: 2.2204460492503130808e-16
3293: snapped 0.99999999999999988898 0.89097009971699747677 0.70018011206237806832 to 1 0.89097009971699747677 0.70018011206237806832
3293: epsilon: 2.2204460492503130808e-16
3293: snapped 0.99999999999999988898 0.88952353407756690196 0.70941179604015447424 to 1 0.88952353407756690196 0.70941179604015447424
3293: epsilon: 2.2204460492503130808e-16
3293: snapped 0.15430344430434703162 0.99999999999999988898 0.13369772444327393535 to 0.15430344430434703162 1 0.13369772444327393535
3293: epsilon: 2.2204460492503130808e-16
3293: snapped 0.99999999999999988898 0.9064230731443844169 0.69819641169733337449 to 1 0.9064230731443844169 0.69819641169733337449
3293: epsilon: 2.2204460492503130808e-16
3293: terminate called after throwing an instance of 'CGAL::Precondition_exception'
3293:   what():  CGAL ERROR: precondition violation!
3293: Expr: known_point != Bare_point()
3293: File: /home/lrineau/Git/cgal/Periodic_3_mesh_3/test/Periodic_3_mesh_3/../../include/CGAL/Periodic_3_mesh_3/Protect_edges_sizing_field.h
3293: Line: 776
```

lrineau:
The full log with the macros is at https://gist.github.com/b1279f712f7492702067049a9258c102.

lrineau:
The Gist web UI truncates the file. The really full log is at https://gist.githubusercontent.com/lrineau/b1279f712f7492702067049a9258c102/raw/59cac6a3cc15c7c974e9ab8d62476d47269e6343/LastTest.log

MaelRL:
Thanks, but the log is missing all the debug output related to protection (`CGAL_MESH_3_PROTECTION_DEBUG`) which seems to be what's useful since it crashes in `Protect_edges_sizing_field.h`.

lrineau:
Sorry, I did not set the macro correctly. Here is the new log: https://gist.github.com/lrineau/435ebcd57f8064e6af4cd3e481e48484

MaelRL:
It's my fault this time with a bug in the c3t3_dump code that caused that second log to crash on another issue... I fixed it and it's in integration.

lrineau:
Now, with the macros defined, the test passes, but without the macros it still fails.

MaelRL:
https://github.com/CGAL/cgal-dev/commit/b27a690a6a91f14b01e6e2850750994d90e7ec2e seems to have been removed from integration by a later push so I can't do much with the P3M3 red boxes.

TDS_3/include/CGAL/Triangulation_data_structure_3.h
```diff
@@ -936,24 +936,40 @@ class Triangulation_data_structure_3
     }
 
     void operator()(Cell_handle c) {
+      std::cout <<  " [] Visiting cell: " << &*c << std::endl;
```
lrineau: I suggest surrounding all use of `std::cout` by
```C++
#ifdef CGAL_DEBUG_TDS_3_<SOMETHING>
  std::cout << "bla";
#endif //  CGAL_DEBUG_TDS_3_<SOMETHING>
```

That way you can actually keep the debugging code in CGAL, for a later use.

Note also that in my opinion `std::cerr` or `std::clog` should be preferred instead of `std::cout`.

MaelRL: This commit was just to run it on Andreas' machine and I'll reset hard after.

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt
```diff
@@ -0,0 +1,705 @@
...
+the method of protecting balls \cgalCite{cgal:cdl-pdma-07}, \cgalCite{cgal:cdr-drpsc-07}
+is used to achieve an accurate representation of those features in the mesh
+and to guarantee that the refinement process terminates
+whatever may be the dihedral angles formed by input surface patches incident to a
+given 1-feature or the angles formed by two 1-features incident to a 0-feature.
+See Section \ref Mesh_3Protectionof0and1dimensionalExposed in the documentation
+of the package \ref PkgMesh_3Summary for further information.
+
+Section \ref Periodic_3_mesh_3MeshingDomainswithSharpFeatures details how
+to prescribe sharp features and examples of periodic meshes with features
```
lrineau: Missing: the full-stop at the end of the sentence.

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt
```diff
@@ -0,0 +1,705 @@
...
+The template parameter `C3T3` is required to be a model of
+the concept `MeshComplex_3InTriangulation_3`. This data structure is devised to
+represent a three-dimensional complex embedded in a periodic 3D triangulation.
+In both functions, an instance of type `C3T3` is used to maintain the current
+approximating simplicial mesh and to represent a single copy of
+the final periodic 3D mesh at the end of the procedure.
+
+The embedding periodic 3D triangulation is required to be the nested type
+`CGAL::Periodic_3_mesh_triangulation_3::type`, provided by the class
+`CGAL::Periodic_3_mesh_triangulation_3`. The type for this triangulation is a
```
lrineau: `CGAL::Periodic_3_mesh_triangulation_3` is not a class. That is a class template. You can also call it a meta-functor, but that wording would need a link, because it is not well known.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -785,7 +785,7 @@ polylines_to_protect(const CGAL::Image_3& cgal_image,
      word_type,
      null,
      CGAL::Identity<Image_word_type>(),
-     internal::Mesh_3::Returns_midpoint<K, Image_word_type>(),
+     Mesh_3::internal::Returns_midpoint<K, Image_word_type>(),
```
lrineau: I do not understand what happened! `CGAL::internal::Mesh_3` should be the right namespace, according to our coding guidelines.

lrineau:
There is now a conflict with `master`, about `.travis.yml`.

... It looks like that sort of conflict will arise as soon as two different PRs modify the list of packages: once one is merged in `master`, then the second one will conflict with `master`. Can we find a solution for that issue?

MaelRL:
TODO: symmetry with #2739

lrineau:
Oops. There is now a conflicts with `master` on a lot of files. For most of them that seems trivial: things added by @afabri, like `#include <CGAL/disable_warnings.h>`, in PR #2712.

MaelRL:
It has been the case for the last couple waves of PR merging, so I'll take care of it when incoming reviews are there and master is 4.13.

MaelRL:
@maxGimeno @lrineau Is there now a script to generate the dependencies of a package? I can't seem to find anything in the wiki.

maxGimeno:
Yes, the CHECK job of  in travis is supposed to display the command but I think it doesn't work for some reason. 
You need to run cmake with the flags CHECK_HEADERS and COPY_DEPENDENCIES for CGAL, it will compute the dependencies for each package and update the list in package_info.

lrineau:
As far as I can see in the [raw log], `CGAL/Periodic_3_mesh_3/IO/File_medit.h` does not compile:
```
/home/travis/build/CGAL/cgal/Installation/../Periodic_3_mesh_3/include/CGAL/Periodic_3_mesh_3/IO/File_medit.h:59:16:
  error: no member named 'make_array' in namespace 'CGAL'

    return CGAL::make_array(std::make_pair(pt[0].first, o0 - diff_off),
           ~~~~~~^

  
  /home/travis/build/CGAL/cgal/Installation/../Periodic_3_mesh_3/include/CGAL/Periodic_3_mesh_3/IO/File_medit.h:80:16:
  error: no member named 'make_array' in namespace 'CGAL'

    return CGAL::make_array(std::make_pair(pt[0].first, o0 - diff_off),
           ~~~~~~^

  2 errors generated.
```

[raw log]: https://api.travis-ci.org/v3/job/345358955/log.txt

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2658)

MaelRL:
> Yes, the CHECK job of in travis is supposed to display the command but I think it doesn't work for some reason. You need to run cmake with the flags CHECK_HEADERS and COPY_DEPENDENCIES for CGAL, it will compute the dependencies for each package and update the list in package_info.

Isn't there a way for packages_dependencies to be run on a single package ? It fails on multiple packages because I didn't link with Metis, or didn't say I allowed cpp11, etc. It's really tedious...

lrineau:
The doc of the CMake option says:
> Enable the special targets `packages_dependencies`, and `pkg_<package>_deps` for each package.

So probably there are targets like `pkg_Triangulation_2_deps`.

MaelRL:
@lrineau @MoniqueTeillaud 

I was looking into adding just the default face base for `Delaunay_triangulation_2` (for #3040). I was looking at `Delaunay_triangulation_3` to see if those changes should also be there and I noticed some issues (mostly the fact that `Delaunay_triangulation_3` should use `Delaunay_triangulation_cell_base_3` as base class, which was not the case often). 

Please look at the following commits, which introduce some significant changes. Although it might seem like strong breaking changes, there were a lot of inconsistencies in the documentation (for example, DT3 is supposed to default to `Delaunay_triangulation_cell_base_3` in the doc), and code would not have compiled if you actually had your own models of the triangulation concepts. Any change in the concepts is also reflected by a change in the default class so nothing changes there so overall there shouldn't be much to be added in the changelog (only the removal of some deprecated classes/functions (?)).

Commit messages give more details.

0e7266b , 1301a6f, 3c85ebe, 1378b32 

Note to Laurent: I went against removing T2_face_base and T3_cell_base because I find that the symmetry with vertex classes is useful and I don't like exposing TDS classes and concepts in the triangulation packages.



Thanks.

PS: done in this branch because it (this branch) already introduces a lot of changes in triangulations.

MoniqueTeillaud:
I am afraid that I missed things, or I forgot... 
I cannot suddenly jump into this thread without recalling the logic that led to introducing the concept DTCellBase_3. 

These changes should be read together with
https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/circumcenter()_in_RT3_cell_base#Concepts
as a reminder. 

Questions:
- Doesn’t this deserve a small feature? I cannot get an overview of what is going on with just a bunch of commits. Or is this just a fix of the above mentioned small feature? 
- What is the relation with P3M3? (see title of this email)
- Was there a proposal to remove T3_cell_base...? I don’t remember this. How would we do without? 
- is mirror_index actually deprecated? 
- I am quite surprised when I read the intro here: https://github.com/CGAL/cgal/commit/3c85ebeda0331f39d7791ad082513e4c663257a3 that mentions circumcenters in T_cell_base_3… 
Was this problem introduced by the above mentioned small feature?

MaelRL:
> I am afraid that I missed things, or I forgot...
> I cannot suddenly jump into this thread without recalling the logic that led to introducing the concept DTCellBase_3.

In the small feature you quote, there is "A concept DelaunayTriangulationCellBase_3 which refines TriangulationCellBase_3 has been added. It requires the circumcenter function to be implemented." So it was added because RegularTriangulationCellBase_3 was added.

> These changes should be read together with
> cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/circumcenter()_in_RT3_cell_base#Concepts
> as a reminder.
> 
> Questions:
> - Doesn’t this deserve a small feature? I cannot get an overview of what is going on with just a bunch of commits. Or is this just a fix of the above mentioned small feature?

I don't know if it requires a small feature but it's mostly fixing the small feature and fixing inconsistencies between code and doc. I'm only removing old deprecated code and ensuring everywhere in the doc/code that `Delaunay_triangulation_3` is used with its intended cell base, which is `Delaunay_triangulation_cell_base_3` and not `Triangulation_cell_base_3` (because `dual()` requires a cell with circumcenter, which `Triangulation_cell_base_3` as a model of `TriangulationCellBase_3` should not provide).

> - What is the relation with P3M3? (see title of this email)

P3M3 already cleaned some of the errors in triangulation-related concepts/doc and I didn't want to fix it twice.

> - Was there a proposal to remove T3_cell_base...? I don’t remember this. How would we do without?

`T3_cell_base` and `T2_face_base` are now basically empty shells, the concepts do not require anything and the classes do not implement anything (well, technically T3_cell_base_3 was implementing a `circucmenter()` function that was used by DT3 but it shouldn't have been the case). We were thus talking about possibly deprecating them in favor of `T3_ds_cell_base_3` and `T2_ds_face_base` directly (but I don't think it's a good idea anymore).

> - is mirror_index actually deprecated?

It was deprecated code in `T2_face_base`. Note that this code has long been replaced by triangulations' `mirror_index`. `ccw`/`cw` should also have been replaced and deprecated but for some reason it was not. I'm leaving it there because it is used in a couple of classes (like AS_2).

> - I am quite surprised when I read the intro here: 3c85ebe that mentions circumcenters in T_cell_base_3…

`T3_cell_base_3` used to implement `circumcenter()`, then it was replaced by `RT3_cell_base_3` and `DT3_cell_base_3` but the default cell class and some other stuff was not properly changed in DT3.

> Was this problem introduced by the above mentioned small feature?

Most likely.

MaelRL:
> btw, if I am not mistaken, P3T3 only has a ds_cell_base, it does not have a cell_base (to answer my own question, we could do without).
> Why do you think that it is not a good idea?

P3T3 indeed only has a P3T3_ds_cell_base. However, I've always wrapped it in a T3_cell_base and the TDS doesn't use the periodic ds_cell_base_3 directly there either.

> Is there a specific question for me?

Not really, it was some significant changes to some concepts/classes so I wanted some feedback (and I didn't want to make it look like I was sneaking it through).

MoniqueTeillaud:
On 27 Apr 2018, at 10:46, Mael <notifications@github.com> wrote:
> btw, if I am not mistaken, P3T3 only has a ds_cell_base, it does not have a cell_base (to answer my own question, we could do without).
> Why do you think that it is not a good idea?
> 
> P3T3 indeed only has a P3T3_ds_cell_base. However, I've always wrapped it in a T3_cell_base and the TDS doesn't use the periodic ds_cell_base_3 directly there either.
> 
> Is there a specific question for me?
> 
> Not really, it was some significant changes to some concepts/classes so I wanted some feedback (and I didn't want to make it look like I was sneaking it through).
> 
thank you for the fixes in T3, and for the information. 

--
Monique Teillaud 
https://members.loria.fr/Monique.Teillaud/
INRIA Nancy - Grand Est, LORIA 
Institut National de Recherche en Informatique et Automatique

janetournois:
@MaelRL @lrineau I added commit 0991bb4df9d84324e33a8fe7036b55be99c2b06e to improve the doc about how to use the optimizers, both in Mesh_3 and P3M3, as we discussed offline.

lrineau:
#2739 is now merged into `master`. That triggered a lot of conflicts.

MaelRL:
Should be ready to be tested once again.

MaelRL:
Travis is fine, but there is a new kind of "can't connect to whatever" type of problem that plagues a handful of batches every push.

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+Illustration in 2D (cut view) of a domain defined by an implicit function that is transformed
+into a periodic implicit function.
+Only the values of the implicit function that are in the canonical cube are used:
+the values of the implicit function at \f$ P \f$ and \f$ Q \f$ are obtained by evaluating
+instead at \f$ P' \f$ and \f$ Q' \f$, as shown on the right.
+\cgalFigureEnd
+
+In practice, the implicit function provided by the user is likely defined
+over a larger domain than the canonical cube (in general, it is \f$ \mathbb R^3\f$).
+Note that -- when constructing artifically periodic functions -- all the values of the implicit function
```
janetournois: "artifically" should be "artificially"

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+for points outside this canonical cube are unused since queries are always answered by looking at the canonical representative.
+\cgalFigureRef{Periodic_3_mesh_3FromCanonicalToWholeDiscard} gives an example of such domain where some information is discarded.
+
+\cgalFigureBegin{Periodic_3_mesh_3FromCanonicalToWholeDiscard, periodicity.svg}
+Illustration in 2D (cut view) of a domain defined by an implicit function made artifically periodic.
```
janetournois: same typo

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+Any value of the function outside of the canonical cube is ignored.
+\cgalFigureEnd
+
+Note also that when construction artificially periodic functions, it is the responsability of the user
```
janetournois: "when constructing", or "at construction"

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+to provide an input function that is compatible with the canonical cube (that is, whose isovalues
+are <em>periodicially</em> continuous and without intersections).
```
janetournois: "periodicially" should be "periodically"

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+\cgalFigureRef{Periodic_3_mesh_3ContinuityIssue} is an example of a bad choice
+of input function and canonical cube: there is no continuity of the isovalues
+at the border of the canonical cube. In such configuration, the mesher might
+or might not finish and the result is likely to be non-manifold and to contain self-intersections.
+
+\cgalFigureBegin{Periodic_3_mesh_3ContinuityIssue, periodicity_issue.svg}
+Illustration in 2D (cut view) of a domain defined by an implicit function made artifically periodic.
```
janetournois: artificially made

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
MaelRL: Ah, here is artificial's "i" thief!

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt
```diff
@@ -73,9 +73,10 @@ are inserted at the beginning of the meshing process, ensuring that the projecti
 of the periodic triangulation into the flat torus \f$ \mathbb T_c^3\f$ forms at all times a simplicial complex
 (see Sections \ref P3Triangulation3secspace and \ref P3Triangulation3secintro
 of the manual of 3D periodic triangulations).
-The meshing process can be exclusively conducted within the canonical cube.
-The mesh is created using the \ref PkgMesh_3Summary package of %CGAL,
-which originally aims to mesh non-periodic domains of \f$ \mathbb R^3\f$, and so, an interface
+Thanks to this "scaffholding" structure, the meshing process can be exclusively conducted
```
janetournois: "scaffholding" should be "scaffolding"

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt
```diff
@@ -365,10 +323,9 @@ the point lies, if inside.
 If the domain description includes 0 and 1-dimensional features
 that have to be accurately represented in the final mesh,
 the template parameter `PeriodicMeshDomain` is required to be
-of a model of the refined concept `Periodic_3MeshDomainWithFeatures_3`.
-The concept `Periodic_3MeshDomainWithFeatures_3` mainly provides the incidence
-graph of 0, 1 and 2-dimensional features, and a member function to construct
-sample points on curves.
+of a model of the refined concept `Periodic_3MeshDomainWithFeatures_3`,
```
janetournois: "to be a model of" is enough, imho

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/Periodic_3_mesh_3.txt
```diff
@@ -503,17 +460,23 @@ Cut view (middle). Another cut is shown, using 8 copies (right).
 \cgalFigureCaptionEnd
 
 While the implicit function used in the previous example is defined and periodic
-over the complete space, it is also possible to consider an implicit function
-defined entirely within the canonical cube. It will then be periodically
-duplicated. For example, replacing the previous domain and with the following
-implicit function (a sphere):
+over the complete space, it is also possible to consider non-periodic implicit functions
+defined entirely within the canonical cube (or over the whole space) by using the wrapper class
```
janetournois: entirely defined

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
janetournois: both are fine indeed, but imho the adverb+verb version sounds more correct

Periodic_3_mesh_3/doc/Periodic_3_mesh_3/CGAL/Periodic_3_function_wrapper.h
janetournois: :-D

lrineau:
Mael, do you think the PR is ready to be tested in `integration`? Why is there still the "TODO" label?

MaelRL:
Yes.

Regarding `TODO`, it's just the `CHANGES.md` that is still to be edited, but I doubt that anything will change from the testing so I'll commit it in a moment.

Installation/CHANGES.md
```diff
@@ -23,11 +30,17 @@ Release date: September 2018
...
+### 2D and 3D Triangulations
 
 -   Added a new type of intersection to handle the insertion of intersecting constraints
     in a `Constrained_triangulation_2`.
 
+-   The long-deprecated class `Triangulation_cell_base_with_circumcenter_3` and its associated
+    concept have been removed. Users should use the classes `Delaunay_cell_base_with_circumcenter_3`
+    or `Regular_cell_base_with_circumcenter_3`, depending on which type of triangulation they are using.
+-   The deprecated functions `mirror_index` and `mirror_vertex` of the class `Triangulation_face_base_2`
+    have been removed. Users should use the equivalent functions from the class `Triangulation_2`.
```
lrineau: I think you should add  `**Breaking change**: ` for those two items, because that removes deprecated class and functions.

afabri:
Isn't a deprecation already the breaking change?

MaelRL:
Depreciation doesn't "break" the code, it's still usable, but the removal does. Thus, depreciation is not flagged, but removal is, as far as I understand.

sloriot:
I guess errors in the `mesh_3` plugin of the demo are related to this PR.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -42,8 +42,8 @@
 #include <CGAL/Orthogonal_incremental_neighbor_search.h>
 
 namespace CGAL {
-
-namespace internal { namespace Mesh_3 {
+namespace Mesh_3 {
+namespace internal {
```
lrineau: That changes broke the `mesh_3_plugin` of the demo.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jun 19, 2018, 10:01 AM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-398310780 "2018-06-19T08:01:32Z - Replied by Github Reply Comments"):
> I guess errors in the `mesh_3` plugin of the demo are related to this PR.

Except for this plugin (and anyway the demo was also red because of another plugin), this PR did a pretty good job, for its first test in `integration`. Congratulation!

lrineau:
Once #3170 is in `master`, you should merge it and apply
```patch
--- a/Periodic_3_mesh_3/test/Periodic_3_mesh_3/test_triply_periodic_minimal_surfaces.cpp
+++ b/Periodic_3_mesh_3/test/Periodic_3_mesh_3/test_triply_periodic_minimal_surfaces.cpp
@@ -177,7 +177,7 @@ FT split_p (const Point& p)
 }
 
 struct Segments_function
-  : public CGAL::unary_function<Point, FT>
+  : public CGAL::cpp98::unary_function<Point, FT>
 {
   typedef std::vector<Segment> Segments;
 
```

MaelRL:
I don't really understand: `C:\CGAL\test\CGAL-4.13-Ic-41\include\CGAL/Periodic_3_function_wrapper.h(51) : warning C4180: qualifier applied to function type has no meaning; ignored` (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4180) on
```
Periodic_3_function_wrapper(const Function_& f, const Iso_cuboid_3& domain) 
  : f_(f), gt_(domain) 
{ }
```
(file: https://github.com/CGAL/cgal/pull/2658/files#diff-28acd94512691fa2d3ac2e3a75ccf56cR36) with
```
typedef FT (Function)(const Point&);
```

The function is also grabbed by const& when there is no wrapper (that is, directly in `Labeled_mesh_domain_3` and it's fine then...

MaelRL:
For the runtime errors on MSVC (in P3M3 and two in Mesh_3), I'll need to check on a Windows machine, I can't tell what's is wrong like that.

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 19, 2018, 11:12 AM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-398331080 "2018-06-19T09:12:03Z - Replied by Github Reply Comments"):
> I don't really understand: `C:\CGAL\test\CGAL-4.13-Ic-41\include\CGAL/Periodic_3_function_wrapper.h(51) : warning C4180: qualifier applied to function type has no meaning; ignored` (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4180) on
> ```
> Periodic_3_function_wrapper(const Function_& f, const Iso_cuboid_3& domain) 
>   : f_(f), gt_(domain) 
> { }
> ```
> (file: https://github.com/CGAL/cgal/pull/2658/files#diff-28acd94512691fa2d3ac2e3a75ccf56cR36) with
> ```
> typedef FT (Function)(const Point&);
> ```
> 
> The function is also grabbed by const& when there is no wrapper (that is, directly in `Labeled_mesh_domain_3`) and it's fine then...

1. I am not sure the `Periodic_3_function_wrapper` should take the function parameter by `const&`. Anyway, the function object will be copied.
2. Even if you do not change the argument type of the constructor, probably the warning will disappear if the typedef is for a function *pointer* instead of a function type. Because a pointer can be const or not, whereas a function type cannot be const.

MaelRL:
> 1. I am not sure the `Periodic_3_function_wrapper` should take the function parameter by `const&`. Anyway, the function object will be copied.
> 2. Even if you do not change the argument type of the constructor, probably the warning will disappear if the typedef is for a function *pointer* instead of a function type. Because a pointer can be const or not, whereas a function type cannot be const.

So, I just drop the `const&` (as it seems tedious to impose function pointers)? Should it be the same in other wrappers of Mesh_3/Periodic_3_mesh_3 that most likely avoided this warning because the function is always passed by function pointer somewhere else (I actually put function types here to make sure it'd work too).

lrineau:
Function types are very special types. They are not object types (the only other non-object types are references and `void`). Applying `const` to a function type is the same as trying to have a "const reference": you cannot (`const int&` is a reference to a `const int`, but `int & const` is ill-formed).

In the following:
```c++
using T = int(int);

struct S {
  T f;
};
```
`T f` does not declare a data member of `S`, but a function of the class `S`. That is why the `Periodic_3_function_wrapper` and `Implicit_to_labeling_function_wrapper` have use meta-programming to avoid "storing" an "object of function type" (I already said that it does not exist).

And any time you have a function parameter of a function type, its type is actually converted to "pointer to function". 

For example:
```c++
using T = int(int);

void f(T) {}
void f(T*) {}
```
The compiler says:
```
essai.cpp: In function ‘void f(int (*)(int))’:
essai.cpp:4:6: error: redefinition of ‘void f(int (*)(int))’
 void f(T*) {}
      ^
essai.cpp:3:6: note: ‘void f(int (*)(int))’ previously defined here
 void f(T) {}
```
because `void f(T)` and `void f(T*)` defines the same function when `T` is a function type.

So:
- one cannot store a "function type object",
- a function parameter type cannot be a function type.

Any time you think you pass a "function type" as argument to a function, actually you pass a function pointer, thanks to the [function-to-pointer][1] implicit conversion.

[1]: https://en.cppreference.com/w/cpp/language/pointer#Pointers_to_functions

MaelRL:
Thanks for the explanation.

There was a runtime error on [CentOS 32 bits yesterday](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-41/Mesh_3/TestReport_lrineau_CentOS6-32.gz) but not present today. Not sure if it's related to this PR.

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 20, 2018, 11:05 AM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-398678632 "2018-06-20T09:05:17Z - Replied by Github Reply Comments"):
> There was a runtime error on [CentOS 32 bits yesterday](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-41/Mesh_3/TestReport_lrineau_CentOS6-32.gz) for Mesh_3 but not present today. Not sure if it's related to this PR.

It is not related. I will register an issue for it. It is probably related to 32 bits, and with that particular seed.

lrineau:
Please fix the conflicts (with a merge of `master`), and at the same time apply the patch about `CGAL::cpp98` (see https://github.com/CGAL/cgal/pull/2658#issuecomment-398330658).

lrineau:
The examples and tests from Periodic_3_mesh_3 should use Eigen (to be compatible with the PR #3128). There is patch for CMake, like that one:
https://github.com/CGAL/cgal/pull/3128/commits/06c0835a3bca07284f89660f64416113e05bc328#diff-2c287fd42769e0ea9866eae0ece04c70
and also a patch in Doxygen, like that one:
https://github.com/CGAL/cgal/pull/3128/commits/ae193e5af35f3603900ffd45a0cab471c46149bc#diff-27611f477dd2a31bd3a70fa76231b390

Otherwise, there are deprecation warnings:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-43.shtml#Periodic_3_mesh_3

MaelRL:
Eigen stuff was done above.

Warnings about function shenanigans are gone.

There is a runtime error in Mesh_3 for `test_c3t3_to_face_graph` on Windows that this PR is most likely responsible for since there were changes to the function (https://github.com/CGAL/cgal/pull/2658/commits/2e26cc55e781814762e8e0efd22bc82ec3857b60).

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 19, 2018, 11:14 AM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-398331672 "2018-06-19T09:14:06Z - Replied by Github Reply Comments"):
> For the runtime errors on MSVC (in P3M3 and two in Mesh_3), I'll need to check on a Windows machine, I can't tell what's is wrong like that.

(I am currently reviewing all pull-requests targeting CGAL-4.13).

The code freeze is approaching. Was it the planning for that?

MaelRL:
I thought the new machine would arrive and I would debug on that, but... it's not there.

@janetournois: Could we please look into it this week?

janetournois:
sure, I'm testing it

lrineau:
I have resolved conflicts, with the merge commit 030970405271dd1499a595ed1ffd981e299a48d0.

janetournois:
I cannot reproduce the crash anymore!

MaelRL:
Last (Windows) error still todo is:

https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-54/Periodic_3_mesh_3_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz

lrineau:
`ctest -R mesh_implicit_shape_with_optimizers` passes on my Linux machine, even with the compilation flag `-fsanitize=address`.

janetournois:
I am currently testing the example `mesh_implicit_shape_with_optimizers.exe`
with some extra expensive assertions on VS2013 Debug and VS2017 Release.

sloriot:
Still failing on windows.

MaelRL:
It's in the process of being fixed, but when nothing has changed, it usually stays the same, yes.

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jul 6, 2018, 10:10 AM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-402960684 "2018-07-06T08:10:03Z - Replied by Github Reply Comments"):
> It's in the process of being fixed, but when nothing has changed, it usually stays the same, yes.

What should I do with this PR?

The current status is that there is [a runtime assertion triggered on all our Windows platforms][2] (MSVC 2013, 2015 or 2017), when CGAL assertions are enabled:
```
- ProgramOutput.mesh_implicit_shape_with_optimizers.x64_Cygwin-Windows10_MSVC2013-Debug-64bits
CGAL::Random()::get_seed() = 1530663075
CGAL::Random()::get_seed() = 1530663075
CGAL error: precondition violation!
Expression : orientation(p0, p1, p2, p3, o0, o1, o2, o3) == POSITIVE
File       : C:\CGAL\test\CGAL-4.13-Ic-54\include\CGAL/Periodic_3_regular_triangulation_3.h
Line       : 710
Explanation: 
Refer to the bug-reporting instructions at https://www.cgal.org/bug_report.html
```

[2]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-54.shtml#Periodic_3_mesh_3_Examples

MaelRL:
It works just fine on my Windows laptop, so I haven't been able to look at the issue precisely. 

@janetournois and I were running some code to debug it, but she had to go before we could solve it. I will look at it when @afabri is back.

Mesh_3/include/CGAL/Mesh_3/Triangulation_helpers.h
```diff
@@ -185,6 +185,9 @@ no_topological_change(Tr& tr,
                       const Point& p,
                       Cell_vector& cells_tos) const
 {
+  if(tr.point(v0) == p)
```
lrineau:
```C++
typename Gt::Equal_3 equal = tr_.geom_traits().equal_3_object();
```
et caetera.

Periodic_3_triangulation_3/include/CGAL/Periodic_3_regular_triangulation_3.h
```diff
@@ -817,6 +817,9 @@ class Periodic_3_regular_triangulation_3
     CGAL_triangulation_precondition(new_position.y() >= domain().ymin());
     CGAL_triangulation_precondition(new_position.z() >= domain().zmin());
 
+    if(new_position == v->point())
```
lrineau: `Equal_3`

Mesh_3/include/CGAL/Mesh_3/Triangulation_helpers.h
MaelRL: Equal_3 doesn't exist for weighted points, see https://github.com/CGAL/cgal/issues/2639.

lrineau: But, what matter here is the equality of the 3D bare points, no?

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
```diff
@@ -28,37 +28,33 @@
 #include <CGAL/array.h>
 #include <CGAL/boost/graph/Euler_operations.h>
 #include <CGAL/Hash_handles_with_or_without_timestamps.h>
+#include <CGAL/Polygon_mesh_processing/orientation.h>
```
sloriot: FYI, this will add a dependency to PMP-orientation. Not sure it was there before. The soup -> triangle mesh might be left to the user is we want to avoid that.

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
```diff
@@ -71,59 +67,120 @@ void facets_in_complex_3_to_triangle_mesh(const C3T3& c3t3, TriangleMesh& graph)
...
+  // Whether all the facets of the c3t3 are being exported, or only those of a particular subdomain
+  const bool export_all_facets = (sd_index == -1);
+
   for(Ficit fit = c3t3.facets_in_complex_begin(),
             end = c3t3.facets_in_complex_end(); fit != end; ++fit)
   {
     Cell_handle c = fit->first;
     int s = fit->second;
-    Face f;
+    Face f(3);
```
sloriot: see for example `read_OFF_internal::resize()`. Maybe a global more standard solution should be found.

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
```diff
@@ -71,59 +67,120 @@ void facets_in_complex_3_to_triangle_mesh(const C3T3& c3t3, TriangleMesh& graph)
...
-  points.reserve(nf / 2); // approximating Euler
+  size_type nf = c3t3.number_of_facets_in_complex();
+  faces.reserve(faces.size() + nf);
+  points.reserve(points.size() + nf/2); // approximating Euler
 
   VHmap vh_to_ids;
   std::size_t inum = 0;
 
+  // Whether all the facets of the c3t3 are being exported, or only those of a particular subdomain
+  const bool export_all_facets = (sd_index == -1);
```
sloriot: What the idea in case `sd_index` is -1? Only dump faces with no special care? One alternative could be to export one (possibly non-manifold) soup per sub-domain.

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
```diff
@@ -1270,64 +1268,13 @@ double Scene_c3t3_item_priv::complex_diag() const {
...
-    polygons[index] = facet;
-  }
-
-  namespace PMP = CGAL::Polygon_mesh_processing;
   Polyhedron outmesh;
+  CGAL::facets_in_complex_3_to_triangle_mesh(c3t3(), outmesh);
+
+  Scene_polyhedron_item* item = new Scene_polyhedron_item(outmesh);
+  item->setName(QString("%1_%2").arg(this->name()).arg("facets"));
+  scene->addItem(item);
```
sloriot: as I told you in PM, when the output is a soup, it should be preserved IMO (the user is responsible to call orient). @lrineau  what do you think?

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
MaelRL: This is changed locally but I haven't pushed it yet (I added another boolean in parameter with a default value).

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
MaelRL: No, this function must return a `triangle_mesh`, it is called `_to_triangle_mesh`.

Besides, `orientation.h` is not needed for this conversion, only `orient_polygon_soup.h` and `polygon_soup_to_polygon_mesh.h`. `orientation.h` was only added because I added
```
  if(CGAL::is_closed(graph) && !PMP::is_outward_oriented(graph))
    PMP::orient_to_bound_a_volume(graph);
```
which can certainly be removed (I just got a bit annoyed that it seemed to always pick the wrong orientation and didn't think about dependencies).

lrineau:
The testsuite of Mesh_3 now has a compilation error:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-67.shtml#Mesh_3

MaelRL:
Fixed by b72bbda .

Mesh_3/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h
lrineau: Why is that an issue to depend on `PMP::orient_to_bound_a_volume`?

Polyhedron/demo/Polyhedron/Scene_c3t3_item.cpp
lrineau: I do not have a strong opinion on that point. Both behaviors are correct, in my opinion.

sloriot:
The branch is rather heavy: 13Mo
I've pushed a filtered version with a last commit adding all files removed during the filtering in `sloriot/Periodic_3_mesh_3-Feature-MBogdanov`. The diff is empty and the size is 6.8Mo.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jul 19, 2018, 9:53 PM GMT+2](https://github.com/CGAL/cgal/pull/2658#issuecomment-406394564 "2018-07-19T19:53:54Z - Replied by Github Reply Comments"):
> The branch is rather heavy: 13Mo
> I've pushed a filtered version with a last commit adding all files removed during the filtering in `sloriot/Periodic_3_mesh_3-Feature-MBogdanov`. The diff is empty and the size is 6.8Mo.

I think your push did not work.

lrineau:
Warning:
```
/mnt/testsuite/include/CGAL/IO/facets_in_complex_3_to_triangle_mesh.h:147:3: warning: 'useless' is used uninitialized in this function [-Wuninitialized]
   facets_in_complex_3_to_triangle_soup(c3t3, useless, points, faces,
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

MaelRL:
There is a runtime error left in `test_meshing_3D_gray_image` but if I remember correctly it's not from this PR (?).

lrineau:
Successfully tested in [CGAL-4.13-Ic-72][*] (and -Ic-71), and then merged.

Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h
```diff
@@ -680,79 +715,77 @@ smart_insert_point(const Bare_point& p, Weight w, int dim, const Index& index,
...
-        insert_a_special_ball = true;
+      for(int i=0, d=tr.dimension(); i<=d; ++i)
+      {
+        const Vertex_handle v = (*it)->vertex(i);
+        if(c3t3_.triangulation().is_infinite(v))
+          continue;
+        if(!vertices_in_conflict_zone_set.insert(v).second)
+          continue;
+
+        const FT sq_d = tr.min_squared_distance(p, cp(c3t3_.triangulation().point(v)));
```
lrineau, line 760: @MaelRL Since that change, here in PR #2658, the file `<CGAL/Mesh_3/Protect_edges_sizing_field.h>` requires that `Tr` is a `Mesh_triangulation_3`.

MaelRL: It's not restricted to feature protection, there are these "generic" functions in optimizers as well, for example.

It can probably be relaxed back to CGAL triangulations using Periodic_tag + triangulation traits / helpers.

PR #2659, opened 2017-12-06
Fix header only when CGAL is installed

lrineau:
After PR #1436 was merged, the installed CGAL was no longer working correctly, and Travis detected it.

This PR will fix the issue.


lrineau:
@maxGimeno This PR will fix the Cmake issue.

lrineau:
I have merged that PR very fast, because all Travis builds (for pull-requests based on recent `master`) were failing without that patch.

PR #2660, opened 2017-12-06
Use documented API

sloriot:
Code was using `add_vertex(pt,graph)` while it is not required by the concept


lrineau:
Should not that function be removed, then, to avoid incorrect use of it?

sloriot:
@afabri what do you think?

lrineau:

PR #2661, opened 2017-12-06
fix named parameter functions

sloriot:
Fix errors introduced by https://github.com/CGAL/cgal/pull/1891


lrineau:
This PR will be delayed until #2649 is merged. And then it will have to be rebased/reimplemented.

lrineau:
Well, this PR is based on 4.11... Sébastien, what would you recommend?

One option:
1. Merge this PR without testing into `releases/CGAL-4.11-branch`.
2. And then merge the patch with #2649 for the integration into `master`.

Actually, I am surprised that it is sufficient to change the name of the parameters, and nothing else in CGAL? Does it mean that they are both:
- undocumented,
- and untested?

MaelRL:
The documentation was correct so there is nothing to change. It was untested, it'll be fixed for master in #2649.

lrineau:
Good. So, I just merged this PR blindly.

PR #2662, opened 2017-12-07
Polyhedron demo: Fix polyhedron flat shading and silent make install in travis

maxGimeno:
## Summary of Changes
For very small polyhedron items, the flat shading was failing, I think because of dFdx() and dFdy(). 
This PR also add the flag --silent to the make install in travis, to avoid the log to be unreadable because of its size.

### DIFF
--- a/.travis/template.txt
+++ b/.travis/template.txt
@@ -15,7 +15,7 @@ before_script:
 - cd build 
 - cmake -DCGAL_HEADER_ONLY=ON -DQt5_DIR="/opt/qt55/lib/cmake/Qt5" -DQt5Svg_DIR="/opt/qt55/lib/cmake/Qt5Svg" -DQt5OpenGL_DIR="/opt/qt55/lib/cmake/Qt5OpenGL" -DCMAKE_CXX_FLAGS_RELEASE=-DCGAL_NDEBUG .. 
 - make 
-- sudo make install
+- sudo make install &>/dev/null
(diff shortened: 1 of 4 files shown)


.travis/template.txt
```diff
@@ -15,7 +15,7 @@ before_script:
 - cd build 
 - cmake -DCGAL_HEADER_ONLY=ON -DQt5_DIR="/opt/qt55/lib/cmake/Qt5" -DQt5Svg_DIR="/opt/qt55/lib/cmake/Qt5Svg" -DQt5OpenGL_DIR="/opt/qt55/lib/cmake/Qt5OpenGL" -DWITH_demos:BOOL=TRUE -DWITH_examples:BOOL=true -DWITH_tests:BOOL=TRUE -DCMAKE_CXX_FLAGS_RELEASE=-DCGAL_NDEBUG .. 
 - make 
-- sudo make install
+- sudo make install --silent
```
lrineau: Have you tried locally? I think the output is made by the commands called my `make`, and not by `make` itself.

maxGimeno: Indeed.

lrineau:
Sébastien @sloriot, can you test with that version?

sloriot:
This fixed the issues I observed.

lrineau:
This PR should be rebased on `master` to fix the Travis issues.

PR #2669, opened 2017-12-08
Use reference type of the pmap

sloriot:
If the point map returns a temporary a temporary was returned.


lrineau:

PR #2670, opened 2017-12-08
Polyhedron demo: small fixes for io_image_plugin

### LINKED ISSUE #2668: 3D demo, segmented images: small issues with the dockwidget
I open a 3D segmented image, and the dockwidgets have small issues:

spectacle w55701

Two issues:
1. The sliders are disabled. They are only enabled once I move a plane using the colored spheres.
2. The editbox displaying  the coordinates of the planes, on the right of sliders, are too short to display coordinates even with only 3 digits. They are only for two digits but they should be wide enough for 4 digits.

maxGimeno:
## Summary of Changes
Fixes some stuf in io_image_plugin.
## Release Management
* Issue(s) solved (if any): fix #2668


lrineau:

PR #2678, opened 2017-12-12
Polygon Mesh Processing documenation - Correct spelling mistake

GilesBathgate:
## Summary of Changes
Change the word 'scenarii' to 'scenarios'

## Release Management

* Affected package(s): PMP documentation
* Feature/Small Feature (if any): none

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
@@ -225,7 +225,7 @@ Mathematically speaking, the intersection with an infinitesimally small ball
 centered in the output volume is a topological ball. At the surface level this means
 that no non-manifold vertex or edge is allowed in the output. For example, it
 is not possible to compute the union of two cubes that are disjoint but sharing an edge.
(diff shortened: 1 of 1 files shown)


janetournois:
Because [wiktionnary](https://en.wiktionary.org/wiki/scenario#English) also mentions "scenarios" as valid, let's keep this merged, but please note that the plural of "scenario" is "scenarii", see [wiktionnary](https://en.wiktionary.org/wiki/scenarii) (though using only one "i" seems to be the modern version).

lrineau:
> (nonstandard, rare) plural form of scenario

The officiel plural is "scenarios": https://en.wiktionary.org/wiki/scenario#Noun

GilesBathgate:
@janetournois I think this is known as a "hypercorrection", but you are right is listed on wiktionary, all be it as a rare nonstandard word. I did check before making the PR https://english.stackexchange.com/questions/11775/what-is-the-plural-of-scenario
It's not in the Oxford English dictionary, nor the American ones listed on the link above, but who knows maybe it will come into common use. I guess as an English speaker I wouldn't expect this word, and perhaps I can save people looking it up to check (which is what I did when I saw it ;) )

janetournois:
I understand your point @GilesBathgate.
Let's keep it this way then, thank you.

PR #2682, opened 2017-12-14
Add an option to triangulate all the faces

sloriot:

### DIFF
--- a/BGL/include/CGAL/boost/graph/convert_nef_polyhedron_to_polygon_mesh.h
+++ b/BGL/include/CGAL/boost/graph/convert_nef_polyhedron_to_polygon_mesh.h
@@ -83,11 +83,14 @@ struct Shell_polygons_visitor
   typedef boost::unordered_map<typename Nef_polyhedron::Vertex_const_handle, std::size_t> Vertex_index_map;
   Vertex_index_map& vertex_indices;
   std::vector< std::vector<std::size_t> >& polygons;
+  bool triangulate_all_faces;
 
   Shell_polygons_visitor(Vertex_index_map& vertex_indices,
-                         std::vector< std::vector<std::size_t> >& polygons)
+                         std::vector< std::vector<std::size_t> >& polygons,
+                         bool triangulate_all_faces)
     : vertex_indices( vertex_indices )
     , polygons(polygons)
+    , triangulate_all_faces(triangulate_all_faces)
   {}
 
   void visit(typename Nef_polyhedron::Halffacet_const_handle opposite_facet)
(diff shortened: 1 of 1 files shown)


sloriot:
Tested in [CGAL-4.12-Ic-124](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-124)

BGL/include/CGAL/boost/graph/convert_nef_polyhedron_to_polygon_mesh.h
```diff
@@ -247,11 +345,11 @@ void collect_polygon_mesh_info(
 /// Converts an objet of type `Nef_polyhedron_3` into a polygon mesh model of `MutableFaceGraph`.
 /// Note that contrary to `Nef_polyhedron_3::convert_to_polyhedron()`, the output is not triangulated
 /// (but faces with more than one connected component of the boundary).
-/// The polygon mesh can be triangulated using the function `triangulate_faces()`.
+/// The polygon mesh can be triangulated by setting `triangulate_all_faces` to `true` or by calling the function `triangulate_faces()`.
```
lrineau: We should start a discussion about Boolean parameter. That is a know issue in an API: https://softwareengineering.stackexchange.com/q/147977/118556

Enums with dedicated names would be better.

See also that very good discussion on the Qt wiki: https://wiki.qt.io/API_Design_Principles#The_Boolean_Parameter_Trap

lrineau: I have added a topic for next CGAL meeting: https://cgal.geometryfactory.com/CGAL/Members/wiki/Meetings/Nancy18#Topics_for_discussions

GilesBathgate: A simple solution would be to create two functions in the public API:
`convert_nef_polyhedron_to_polygon_mesh`
and
`convert_nef_polyhedron_to_triangulated_polygon_mesh`

Furthermore, I think it's quite likely you would always want to use the latter since the former creates undesirable geometry like this:

https://github.com/GilesBathgate/cgal/blob/ada918f5b61162df1aceaafb7940a9a2197bb5f2/Nef_3/examples/Nef_3/cube1.stl

GilesBathgate: @lrineau Thinking about it, its probably erroneous conversion of quads to the STL file format in my test code that generates that output. I still think that the two function approach is a clearer API though.

PR #2686, opened 2017-12-15
Add plugin to duplicate non-manifold vertices

sloriot:


lrineau:
I was about to say that we should try to limit the number of new plugins, but actually the title of the PR is wrong: you added a new *action* in an existing plugin.

sloriot:
Tested in [CGAL-4.12-Ic-124](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-124)

PR #2688, opened 2017-12-15
Fix Convex_hull_3 quickhull, with coplanar point

lrineau:
## Summary of Changes

`CGAL::convex_hull_3` documents that the output is a triangulation
polyhedron. That is true... but for coplanar input points! In case of
coplanar input point, the output before this patch was a polyhedron with a
single polygonal face. The patch triangulates the face using a pivot point
and Euler operations.

## Release Management
* For CGAL-4.10.2
* Affected package(s): Convex_hull-3


lrineau:

PR #2691, opened 2017-12-19
Installation: fix the detection of C++11 and the need for Boost Thread

lrineau:
## Summary of Changes

Since recent changes to CMake (PR #1436), it was no longer possible on Linux to use C++11 to build CGAL libraries, and C++03 for a user-project. This PR fixes the issue.

## Release Management
* Affected package(s): CMake/Installation


sloriot:
Tests failing see for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-122/AABB_tree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz)

lrineau:
@efifogel I had to modify this PR. Now it will be possible to compile CGAL libraries with C++11 and a user program with C++03, *only if* CMake is version 3.5 or later. Your machine uses Ubuntu Xenial, that has CMake 3.5.1, so that will work for you.

PR #2707, opened 2018-01-03
Improve release creation cmake script

sloriot:


Installation/include/CGAL/version_macros.h
```diff
@@ -33,6 +33,11 @@
 #define CGAL_STR_STR(X) #X
 #endif
 
+#ifndef CGAL_str
+#define CGAL_xstr(s) #s
+#define CGAL_str(s) CGAL_xstr(s)
+#endif
```
lrineau: Those macros are not used anywhere, and are identical to `CGAL_STR` and `CGAL_STR_STR` defined just a few lines above.

sloriot: I just copy pasted what the previous release creation script was adding in `version.h`

lrineau: :-)

lrineau:
I ran that script with a checkout of 9c6b131 (and a merge of this PR), and compared it with https://cgal.geometryfactory.com/CGAL/Members/Releases/CGAL-4.12-I-131.tar.gz, and the results are really different.

Our old release script did a lot of strange work in test/:
 - copy, for example, `demo/Triangulation_2/` into `test/Triangulation_2_Demo`,
 - add `cgal_test_with_cmake` scripts in test directories where it is missing.

And maybe something else.

I find the behavior of the old script quite buggy, but if we want to use the new CMake script instead, then we will have to deal with those differences:
  - either modify our testsuite process,
  - or modify the CMake script to mimic the strange behavior of the old perl script.

sloriot:
I think we should be good now.

lrineau:
With `TESTSUITE=ON`, the produced internal release is incorrect. The `cgal_test_with_cmake` that is copied in all `test/*/` sub-directory is not right, and triggers a fork bomb (by calling itself recursively), if called.

PR #2721, opened 2018-01-09
Automatically create a bundle of the Polyhedron demo for Linux.

### LINKED ISSUE #2650: (Automatic?) build of demo/Polyhedron into a AppImage
[AppImage] is a way to provide a self-contained application for Linux, that does not require any dependency, because all the dependencies are embedded in the AppImage file. That is similar to [Snaps] or [FlatPak].

[AppImage]: https://appimage.org/
[Snaps]: https://snapcraft.io/
[Flatpak]: https://flatpak.org/

The standard libraries of Linux (libc, among others) are not in the set of embedded dependencies. That is why, when you want to distribute an AppImage that will work on almost all Linux distribution, you need to compile on an old one, and use the fact that the set of standard libraries is backward compatible: a build on an old Linux will run fine on a more recent Linux.

From our Docker image `cgal/testsuite-docker:centos6` (that supports RHEL 6 and CentOS6, two ancient Linux distributions), I have derived an image with the tools needed to create `AppImage` files for a Qt5 demo: https://github.com/lrineau/bundle-CGAL-3D-demo.

This task is to use that image (maybe recompile it if necessary) and build the `master` version of the Polyhedron demo into an `AppImage` file.

If that process could be automatize to that every Sunday we have a new version of the Polyhedron demo, that would be even better.

maxGimeno:
## Summary of Changes

Add a script in Scripts/developer_scripts that bundles the Polyhedron demo with AppImage Linux.
We should discuss where to put the said bundle to download and modify the script accordingly.

## Release Management

* Issue(s) solved (if any): fix #2650

### DIFF
--- a/Scripts/developer_scripts/Bundle_polyhedron_demo_with_appimage.sh
+++ b/Scripts/developer_scripts/Bundle_polyhedron_demo_with_appimage.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 if [ "$1" == '--help' ]; then
-  echo "Usage: $0 <path to CGAL> <path to output>"
+  echo "Usage: $0 <path to CGAL> <path to output> <number of cores to dedicate>"
   echo "Builds and packages the Polyhedron demo form the CGAL dir."
   exit 0
 fi
-docker run --rm -v "$2":/results:Z -v "$1":/cgal:ro docker.io/cgal/bundle-3d-demo '/scripts/build.sh -j6 && /scripts/deploy.sh'
+docker run --rm -v "$2":/results:Z -v "$1":/cgal:ro docker.io/cgal/bundle-3d-demo "/scripts/build.sh -j$3 && /scripts/deploy.sh"
 
--- a/Scripts/developer_scripts/Bundle_polyhedron_demo_with_appimage.sh
+++ b/Scripts/developer_scripts/Bundle_polyhedron_demo_with_appimage.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+if [ "$1" == '--help' ]; then
+  echo "Usage: $0 <path to CGAL> <path to output>"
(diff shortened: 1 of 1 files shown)


maxGimeno:
@lrineau we still haven't discussed where the bundle should be put to download.

lrineau:
That is really a side-question.

I have tried the script on `master`, and the build failed. That is something to fix before we consider the non-important details.

maxGimeno:
Well I just tested it and it worked perfectly. Can you be more specific about what failed ?

lrineau:
It seems it failed because I have tried to compile a branch that was not `master` (the branch `maxGimeno/New_generation_demo-GF`).

One first remark: the script hardcodes `-j6`, and that is certainly not the optimal on my small laptop.

sloriot:
@maxGimeno is this branch ready to be tested?

maxGimeno:
You can ask @lrineau, but from what I remember it is finished, yes.

lrineau:
Most of the work is about testing the Docker image, see if it works, and then install a cron job that creates the Linux AppImage regularly/ Weekly?

maxGimeno:
It could be created every time the last release is tested (every week-end).

maxGimeno:
I tested it again with the release 4.12, it produces a fine app. I did not test everything but it seems to work.

lrineau:
So, what would be the command of the cron job, if we want to build such an image every week?

maxGimeno:
Something like 

>0 22 * * SUN bash CGAL_ROOT/Scripts/developer_scripts/Bundle_polyhedron_demo_with_appimage.sh <path to the latest release> <path to output> <number of cores to dedicate>

sloriot:
@lrineau any update?

lrineau:
Just a lack of time, or priority, to make that script integrated in our testing process.

maxGimeno:
@lrineau still no time ?

PR #2724, opened 2018-01-10
Fix for doxygen 1.8.14

sloriot:
\n in aliases seems to be an issue with doxygen 1.8.14
Replacing them by ^^ seems to solved the isssue


lrineau:
Tested in [CGAL-4.12-Ic-142].

  [CGAL-4.12-Ic-142]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-142.shtml

Issue #2733, opened 2018-01-12
bug with Plane_3.to_2d()?


hugoledoux:
## Issue Details

I first define 5 points in 3D (forming an horizontal plane), then fit a plane through them with the least-square function, and then project each of the original points to the plane with the function Plane_3.to_2d(). The projection returns wrong values, such as:
```
Projection of points to the plane:
1.38039e+27 -379197
1.38039e+27 -379207
...
```

Maybe because the coefficients of the plane are not exactly 0, but -4.3906e-26 for instance?

If I translate the plane by z=-200 then it is okay...

Without Eigen this problem doesn't exists, with it always. Tested on macOS and Ubuntu with CGAL 4.11 and Eigen 3.3.4

## Source Code

https://gist.github.com/hugoledoux/f763636492dc72fb9f3fb77932b84c56

## Environment
 
macOS 10.13.2

with CGAL v4.11 installed with brew
$ brew info cgal
cgal: stable 4.11 (bottled)
Computational Geometry Algorithm Library

$ brew info eigen
eigen: stable 3.3.4 (bottled), HEAD

lrineau:
Fixed by #2735. Will be in CGAL-4.11.1, and CGAL-4.12.

PR #2735, opened 2018-01-15
Cartesian kernel: improve plane point() computation

sgiraudot:
## Summary of Changes

This fix is similar to https://github.com/CGAL/cgal/pull/2122: when computing `Plane_3::point()` (which is supposed to return one point lying on the plane - any point, there is no more constraint), the current code uses a division by the first non-null coordinate: if a coordinate is close to 0 (but non-null), the point's corresponding coordinate goes far away and becomes badly defined.

This leads, for example, to wrong and clearly unusable values for the `Plane_3::to_2d()` method (which internally uses `Plane_3::point()` as the origin of the 2D base), as was reported in https://github.com/CGAL/cgal/issues/2733.

In this fix, we simply check for the largest coordinate (in norm) and use it, to make sure that we always get the coordinate that is the closest to the origin (this is, again, very similar to what I did in https://github.com/CGAL/cgal/pull/2122). This fixes cases like https://github.com/CGAL/cgal/issues/2733 without affecting the other cases

## Release Management

* Affected package(s): Cartesian Kernel
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2733

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -1762,10 +1762,10 @@ namespace CartesianKernelFunctors {
         // to avoid badly defined vectors with coordinates all close
         // to 0 when the plane is almost horizontal, we ignore the
         // smallest coordinate instead of always ignoring Z
-        if (CGAL::possibly(a <= b && a <= c))
+        if (CGAL::possibly(CGAL_AND (a <= b, a <= c)))
           return Vector_3(FT(0), -h.c(), h.b());
 
-        if (CGAL::possibly(b <= a && b <= c))
+        if (CGAL::possibly(CGAL_AND (b <= a, b <= c)))
           return Vector_3(-h.c(), FT(0), h.a());
 
         return Vector_3(-h.b(), h.a(), FT(0));
--- a/Cartesian_kernel/include/CGAL/constructions/kernel_ftC3.h
+++ b/Cartesian_kernel/include/CGAL/constructions/kernel_ftC3.h
@@ -264,9 +264,20 @@ point_on_planeC3(const FT &pa, const FT &pb, const FT &pc, const FT &pd,
                  FT &x, FT &y, FT &z)
 {
   x = y = z = 0;
-  if (! CGAL_NTS is_zero(pa))      x = -pd/pa;
-  else if (! CGAL_NTS is_zero(pb)) y = -pd/pb;
-  else                             z = -pd/pc;
+  
+  FT abs_pa = CGAL::abs(pa);
+  FT abs_pb = CGAL::abs(pb);
+  FT abs_pc = CGAL::abs(pc);
+  
+  // to avoid badly defined point with an overly large coordinate when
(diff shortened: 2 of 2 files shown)


Cartesian_kernel/include/CGAL/constructions/kernel_ftC3.h
```diff
@@ -264,9 +264,20 @@ point_on_planeC3(const FT &pa, const FT &pb, const FT &pc, const FT &pd,
...
-  else                             z = -pd/pc;
+  
+  FT abs_pa = CGAL::abs(pa);
+  FT abs_pb = CGAL::abs(pb);
+  FT abs_pc = CGAL::abs(pc);
+  
+  // to avoid badly defined point with an overly large coordinate when
+  //  the plane is almost orthogonal to one axis, we use the largest
+  //  scalar coordinate instead of always using the first non-null
+  if (CGAL::possibly(abs_pa >= abs_pb && abs_pa >= abs_pc))
```
mglisse: The use of `possibly` with `&&` seems strange to me. `&&` is not overloaded for `Uncertain` so it converts to bool, so `possibly` should be useless. I think you want either `possibly(a>=b)&&possibly(a>=c)` or `possibly(a>=b&a>=c)`.

lrineau: I wonder why `&&` and `||` are not overloaded for `Uncertain<bool>` (but only `&` and `|`). That would allow the writing of generic code deal either with `bool` or `Uncertain<bool>`. Actually, we have such generic code, and it uses the conversion to `bool`.

.. Well, the reason we do not have `&&`/`||` is actually documented:
https://github.com/CGAL/cgal/blob/e2f89318e57d9428e8a221a7d337bab4c5cd51d5/STL_Extension/include/CGAL/Uncertain.h#L338-L360

It was added ten years ago, in 449195eb9d15dac762e620104e2a076ce669203c.

sgiraudot: Ok, so I can use `CGAL_AND`. I suppose I also correct it in this other place: https://github.com/CGAL/cgal/pull/2122/commits/62df1454e3cefe26c03f9e5d2ca418d1eec95413

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
```diff
@@ -1762,10 +1762,10 @@ namespace CartesianKernelFunctors {
         // to avoid badly defined vectors with coordinates all close
         // to 0 when the plane is almost horizontal, we ignore the
         // smallest coordinate instead of always ignoring Z
-        if (CGAL::possibly(a <= b && a <= c))
+        if (CGAL_AND (CGAL::possibly(a <= b), CGAL::possibly(a <= c)))
```
mglisse: That's weird again, you are using CGAL_AND for 2 bools here...

lrineau:
Tested in [CGAL-4.12-Ic-152], successfully, and then merged.

  [CGAL-4.12-Ic-152]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-152.shtml

PR #2737, opened 2018-01-15
BGL: Isotropic_remeshing() keeps constrained vertices intact

maxGimeno:
## Summary of Changes

This PR modifies `CGAL::Euler::collapse_edge()` and `CGAL::Polygon_mesh_processing::isotropic_remeshing()` so that it is always the target vertex of a collapsed_edge that is returned.
This way isotropic_remeshing() keeps the constrained vertices intact, not only their position.
## Release Management

* Affected package(s):BGL, PMP


Polyhedron/demo/Polyhedron/Scene_polyhedron_selection_item.cpp
```diff
@@ -1247,7 +1247,7 @@ bool Scene_polyhedron_selection_item:: treat_selection(const std::set<fg_edge_de
           S = get(vpm, source(targt, *polyhedron()));
           T = get(vpm, target(targt, *polyhedron()));
 
-          put(vpm, CGAL::Euler::collapse_edge(ed, *polyhedron()), Point(0.5*(S.x()+T.x()), 0.5*(S.y()+T.y()), 0.5*(S.z()+T.z())));
+          get(vpm, CGAL::Euler::collapse_edge(ed, *polyhedron())) = T;
```
sloriot: Why would you no longer take the midpoint?

maxGimeno:
Because it is more explicit to take the real remaining point. It is not
really related to the other changes thiugh, just correcting a mistake I
made originally.

Le 15 janv. 2018 1:36 PM, "Sebastien Loriot" <notifications@github.com> a
écrit :

sloriot:
@maxGimeno  I changed the logic of the modification in `collapse_edge()` so as to keep the guarantee that always the same halfedges are removed. To do so I simply swapped the edge vertices in the case `v1` was about to be removed.
TODO: The BGL package is missing a testsuite, it would be nice to add a complete test for the function (and check that all branching point of the functions are tested). I guess the graph [here](https://doc.cgal.org/latest/Surface_mesh_simplification/classEdgeCollapsableSurfaceMesh.html) are the examples that need to be tested.

sloriot:
@janetournois I wonder how you handle the collapse in case there are some constrained edges incident to the edge to be collapsed.

janetournois:
you mean the edge is not constrained, but adjacent to constrained edges at its endpoints?

sloriot:
@janetournois yes

janetournois:
in `is_collapse_allowed()`, there is
```
if (is_on_patch(he)) //hopp is also on patch
      {
        CGAL_assertion(is_on_patch(hopp));
        if (is_on_patch_border(target(he, mesh_)) && is_on_patch_border(source(he, mesh_)))
          return false;//collapse would induce pinching the selection
```
which prevents this edge from collapsing

sloriot:
What if I have constrained edges in the interior of a patch?

janetournois:
the `PATCH_BORDER` status also concerns the constrained halfedges that are in the middle of a patch

sloriot:
Then `if (is_on_patch_border(target(he, mesh_)) && is_on_patch_border(source(he, mesh_)))` protects only if the two endpoints are on the boundary not if only one of them is. I'm mentioning this because with the overload of `edge_collapse()` without constrained, prev edges are removed which can be problematic if one of them is constrained.

janetournois:
previously, when an edge with only one constrained endpoint was to be collapsed, I was making sure it got collapsed TO the constrained vertex.
Not sure what happens with the new implementation...

sloriot:
I'm not talking about constrained vertices but constrained edges here. Let's have a meeting to discuss that when you have time.

lrineau:
Bug fix for CGAL-4.12, or CGAL-4.11?

sloriot:
I prefer delaying it to 4.13

sloriot:
I've added an additional fix so that constrained halfedges are present if the output (when not refined of course).

sloriot:
@janetournois If I try to call the isotropic remeshing with 0.1 as edge length with the following
[edge selection](https://github.com/CGAL/cgal/files/1636585/elephant.selection.selection.txt), some constrained edges are not in the output. It is observed with master and this branch. They should be preserved, right?

janetournois:
Did you click "protect features"?
Otherwise, the selected polylines can be resampled. Only corners and endpoints are guaranteed to be preserved.

BGL/test/BGL/test_Collapse_edge.cpp
```diff
@@ -53,14 +53,19 @@ collapse_edge_test()
...
     BOOST_FOREACH(halfedge_descriptor it, CGAL::halfedges_around_target(v1,test_mesh))
     {
-
+      if(it == eno
+         || it == eno_prime){
+        ++found;
+      }
     }
-    assert(CGAL::Euler::collapse_edge(edge(e, test_mesh), test_mesh) == v1);
+    assert(found == 2);
```
sloriot: This is only true if e and its opposite or not boundary edges, otherwise it is 1.

sloriot: Sorry, I though you were testing next of e and next of opposite(e) (which you should be testing).

sloriot:
That was because of the smoothing along 1D lines.

sloriot:
Actually it's not see:
Before:
screenshot from 2018-01-18 21-11-51
After:
screenshot from 2018-01-18 21-11-39

using 0.1, no border smoothing and no border/cst protection

sloriot:
OK I've just re did the tests. It is OK with master but no longer with this branch. I'm investigating why.

sloriot:
64d96f5 fixes the aforementioned issue.

BGL/test/BGL/test_Collapse_edge.cpp
```diff
@@ -0,0 +1,226 @@
...
+      else
+        return opposite(h, m);
+    }
+  }
+  return boost::graph_traits<Mesh>::null_halfedge();
+}
+
+template <typename Mesh>
+void
+collapse_edge_test()
```
janetournois: everywhere in this test, I suggest to call `h` or `he` the `halfedge_descriptor`, instead of `e`, which seems more to be an edge.

The reason is that when I read `e = find_halfedge(...)` or `opposite(e, g)`, it seems to be a bad use of the types.

This remark is for future debugging

Installation/CHANGES.md
```diff
@@ -1,6 +1,23 @@
...
+Release date: September 2018
+
+### Polygon Mesh Processing
+
+-  Fix a bug in `isotropic_remeshing()` making constrained vertices missing in the output
+
+### CGAL and the Boost Graph Library (BGL)
+
+-   Update the function `CGAL::Euler::collapse_edge` so that the target
+    vertex of the edge collapsed is always kept after the collapse.
```
janetournois: - I would replace "Update" with "Improve"
- "edge collapsed" should better be "collapsed edge", imho

Installation/CHANGES.md
```diff
@@ -1,6 +1,23 @@
...
+
+-   Guarantee that constrained vertices are kept in the remeshed mesh, 
+    and not only constrained points like before.
```
janetournois: does this concern BGL?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -743,34 +742,55 @@ namespace internal {
...
+              CGAL_assertion(mesh_.is_valid());
+            }
           }
 
           //perform collapse
-          Point target_point = get(vpmap_, vb);
+          CGAL_assertion(target(halfedge(e, mesh_), mesh_) == vb);
           vertex_descriptor vkept = CGAL::Euler::collapse_edge(e, mesh_);
-          put(vpmap_, vkept, target_point);
+          CGAL_assertion(mesh_.is_valid());
```
janetournois: this is not BGL

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -743,34 +742,55 @@ namespace internal {
...
+              set_face(ep_p, face(eno_p, mesh_), mesh_);
+              set_face(en_p, face(epo_p, mesh_), mesh_);
+              set_face(epo_p, face(he_opp, mesh_), mesh_);
+              set_face(eno_p, face(he_opp, mesh_), mesh_);
+              set_halfedge(face(he_opp, mesh_), he_opp, mesh_);
+              if (face(ep_p, mesh_) != boost::graph_traits<PM>::null_face())
+                set_halfedge(face(ep_p, mesh_), ep_p, mesh_);
+              if (face(en_p, mesh_) != boost::graph_traits<PM>::null_face())
+                set_halfedge(face(en_p, mesh_), en_p, mesh_);
+              CGAL_assertion(mesh_.is_valid());
```
janetournois: this is not BGL

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
sloriot: yes `is_valid_polygon_mesh()` actually. Were you talking of something else?

janetournois: Ok, but then it should not be there... otherwise we may have issues with non-CGAL meshes.

maxGimeno: So what do we do ? we remove the assertion, making the function vulnerable ? Or we wait for the new mesh validity function to be integrated ?

janetournois: Why not use 
```
CGAL_assertion(is_valid(mesh_));
CGAL_assertion(is_triangle_mesh(mesh_));
```
as everywhere else in the file?

At least `is_valid(mesh_)` is in the `graph_traits_**.h` files

BGL/include/CGAL/boost/graph/Euler_operations.h
```diff
@@ -1021,21 +1021,21 @@ add_face_to_border(typename boost::graph_traits<Graph>::halfedge_descriptor h1,
  *
  * \tparam Graph must be a model of `MutableFaceGraph`
  * Let `h` be the halfedge of `e`, and let `v0` and `v1` be the source and target vertices of `h`.
- * Let `ep` and `e'p` be respectively the edges of `prev(h,g)` and `prev(opposite(h, g), g)`.
+ * Let `hp` and `e'p` be respectively the edges of `prev(h,g)` and `prev(opposite(h, g), g)`.
```
janetournois: Imho that's not any better... Now we have `h` and `e` mixed...
In the first line, "prime" indicates that we are taking the opposite, and in the second line, it does not have the same meaning

BGL/include/CGAL/boost/graph/Euler_operations.h
```diff
@@ -1021,21 +1021,21 @@ add_face_to_border(typename boost::graph_traits<Graph>::halfedge_descriptor h1,
...
  * After the collapse of edge `e` the following holds:
  *   - The edge `e` is no longer in `g`.
- *   - The faces incident to edge `v0v1` are no longer in `g`.
+ *   - The faces incident to edge `e` are no longer in `g`.
  *   - `v0` is no longer in `g`.
- *   - If `h` is not a border halfedge, `ep` is no longer in `g` and is replaced by `en`.
+ *   - If `h` is not a border halfedge, `hp` is no longer in `g` and is replaced by `en`.
  *   - If the opposite of `h` is not a border halfedge, `e'p` is no longer in `g` and is replaced by `e'n`.
- *   - The halfedges kept in `g` that had `v0` as target now have `v1` as target, and similarly for the source.
+ *   - The halfedges kept in `g` that had that had v0 as target (resp. source) now have v1 as target (resp. source).
```
janetournois: missing backquotes around `v0` and `v1`

janetournois: and "that had" is duplicated

BGL/include/CGAL/boost/graph/Euler_operations.h
maxGimeno: so hp and h'p ?

Installation/CHANGES.md
```diff
@@ -29,6 +27,19 @@ Release date: September 2018
     to reflect the real needs of the code (some types and operators were used
     in the code but did not appear in the concepts).
 
+### Polygon Mesh Processing
+
+-  Fix a bug in `isotropic_remeshing()` making constrained vertices missing in the output
+-  Guarantee that constrained vertices are kept in the mesh after calling `isotropic_remeshing()`
+   (and not only constrained points like it was before).
```
janetournois: What do you mean by "constrained points"?

maxGimeno: vertices on a constraint edge, I guess.

sloriot: In the current version (master), the only thing that is guaranteed if you have constrained vertices is that there will exist in the output some vertices with the same coordinates.

janetournois: Oh ok I understand !

So what about changing the parenthesis by
"(and not only the associated constrained points as it was before)."

Installation/CHANGES.md
```diff
@@ -31,7 +31,7 @@ Release date: September 2018
 
 -  Fix a bug in `isotropic_remeshing()` making constrained vertices missing in the output
 -  Guarantee that constrained vertices are kept in the mesh after calling `isotropic_remeshing()`
-   (and not only constrained points like it was before).
+   (and not only the points associated to constrained vertices as it was befor).
```
janetournois: before

sloriot:
TODO: `collapse_edge(e, g, ecm)` must also be updated.

afabri:
The [VC++ testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-14/BGL/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz) fails. 
The reason is that the file `data/flat_hexahedron.off` has comments at the end of the lines with vertices and the read_off function of OpenMesh cannot parse that.
The crash comes from the fact that `copy_face_graph()` does not work for an empty OpenMesh.

lrineau:
@sloriot Do you remember why you set the `TODO (web)` flag?

lrineau:
Should I keep testing the PR in `integration`, or should I way until it is ready?

sloriot:
I think the current version is fine. I remove the ready flag and I'll put it as soon as I'm doing with the TODO.

lrineau:
The PR "Property_map: Fix make_property_map(std::vector&) for an empty vector" #3102 is marked as "fixing" this PR. I do not know what that means. Anyway, that PR #3102 is now merged in both `master` and `releases/CGAL-4.12-branch`.

Note that this PR has a conflict in `CHANGES.md`, that should be easy to fix.

sloriot:
FYI reference of PR #3102 was because @maxGimeno did not commit an test file and a segfault occurred because a vector was empty.

lrineau:

PR #2738, opened 2018-01-15
Make the update of the max value correct

sloriot:
In Haussdorf distance computation in parallel, while the operations are atomic there could be an update of the value and the real max would be overwritten.

In passing it also fixes this [issue](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-I-141/Polygon_mesh_processing/TestReport_lrineau_CentOS6-CXX11-Boost157.gz).

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
@@ -111,8 +111,12 @@ struct Distance_computation{
       if (d>hdist) hdist=d;
     }
 
-    if (hdist > distance->load())
-      distance->store(hdist);
+    // update max value stored in distance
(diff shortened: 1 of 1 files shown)


lrineau:
Do you really need it for 4.10.2?

sloriot:
I don't, I simply apply the patch as _far_ as I could.

lrineau:
The closing was a consequence of removing the branch `releases/CGAL-4.10-branch`. I have re-applied the PR to `releases/CGAL-4.11-branch` and re-opened it.

lrineau:
Tested in [CGAL-4.12-Ic-152], successfully, and then merged.

  [CGAL-4.12-Ic-152]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-152.shtml

PR #2741, opened 2018-01-16
Cleanup documentation

### LINKED ISSUE #2397: clean collapse_edge and surface mesh simplification docs
Copied from cgal-discuss:

Hello,
here
<https://doc.cgal.org/latest/Surface_mesh_simplification/classEdgeCollapsableSurfaceMesh.html>  
it is actually described what results the collapse_edge function produces,
therefore I believe it would be helpful that at least this explanation is
copied (if not moved) to the  collapse_edge doc
<https://doc.cgal.org/latest/BGL/group__PkgBGLEulerOperations.html#gab97c0d071b0d768e22a4fc171a8ca664> 
.

Best Regards,
IM

I even think that the doc is not up-to-date as IIRC @maxGimeno did a modification so that the same vertex is always kept. The doc of Surface mesh simplification must also be cleaned up.

sloriot:
Remove no longer needed concepts and use the one from the BGL package.

@MaelRL I also picked the changes you did in the user manual. Let me know if I forgot something.

Fixes #2397


MaelRL:
- Why not grab the trailing white space commits along the way ?
- We were talking about remaking / moving the figures about edge collapse to BGL. Are you just removing them ?

sloriot:
About whitespaces: your patches could not be applied by cherry-pick (too many conflicts) + I would prefer a PR fixing them all at once where I have nothing to check (I simply call `git diff -w` on the branch merged into master). Otherwise you have to dig into the diff to catch the important pieces.
About the fig, PR #2737 will make them useless.

sloriot:
Successfully tested in CGAL-4.13-Ic-3

PR #2742, opened 2018-01-16
Travis: retry on apt-get failure

### LINKED ISSUE #2740: Travis fails way too often on `apt-get`
## Issue Details

As we have lots of lines in our Travis testsuite, the probability that at least one fails at `apt-get` is quite high (I have no memory of a recent pull request where I did not have to restart it by hand at least 3 or 4 times). It's quite annoying as the PR appears red and we have to check by hand every red cross to see if it does correspond to an `apt-get` failure.

Would it be possible to improve the scripts such that this happens less often? I don't know if it's possible, but what I have in mind:

1. Do the `apt-get` command
2. Check if it executed properly (I imagine this is possible)
3. If not, sleep for like 10 seconds and try again (do that until it works)

I think that could help a lot. What do you think?

maxGimeno:
## Summary of Changes
Manually call apt-get install on each package and wrap it in a test to make it try again when it fails.
## Release Management

* Issue(s) solved (if any): fix #2740

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,10 +69,10 @@ script:
 - cd ./.travis
 - bash -x -e ./build_package.sh $PACKAGE
 addons: 
-  apt:
-    sources:
-      - sourceline: 'ppa:ppsspp/cmake' 
-      - sourceline: 'ppa:hedges/qt5.5'
+#  apt:
+#    sources:
+#      - sourceline: 'ppa:ppsspp/cmake' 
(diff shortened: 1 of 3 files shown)


maxGimeno:
It seems to work. Only error remaining is a timeout for the polyhedorn demo.

lrineau:
That does not seem to work. There is a job that failed anyway: https://travis-ci.org/CGAL/cgal/jobs/344220532

PR #2757, opened 2018-01-22
BGL:  Add a  FaceGraph validity check function

### LINKED ISSUE #2401: Add a function to check a FaceGraph is valid
In the BGL package there are some undocumented functions (`is_valid()` and `is_valid_polygon_mesh()`) that have some assumptions about the FaceGraph model. It would be nice if only one exists and that it uses the methods provided by the FaceListGraph concept.

Subdivision package has some post condition that I commented due to the lack of validity checking function documented. They should be re-enabled.

maxGimeno:
## Summary of Changes

This PR adds a documented CGAL::is_valid_polygon_mesh() function that replaces the undocumented CGAL::is_valid() that worked on some FaceGraphs.
## Release Management

* Affected package(s):BGL
* Issue(s) solved (if any): fix #2401
* Feature/Small Feature (if any): [This one](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Polygon_Mesh_Validity_Check)


BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,207 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_descriptor     vertex_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::face_descriptor       face_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_iterator       vertex_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::halfedge_iterator     halfedge_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::face_iterator         face_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::vertices_size_type    vertex_size_type;
+  typedef typename boost::graph_traits<PolygonMesh>::halfedges_size_type   halfedges_size_type;
+  typedef typename boost::graph_traits<PolygonMesh>::faces_size_type       faces_size_type;
+  Verbose_ostream verr(verb);
+  int num_v(std::distance(boost::begin(vertices(g)), boost::end(vertices(g)))),
```
sloriot: `int` will trigger a conversion warning, use `std::size_t` for example. The real type is `std::iterator_traits<Iterator>::difference_type`

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,207 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+    // opposite integrity.
+    valid = valid && ( opposite(begin, g) != begin);
+    valid = valid && ( opposite(opposite(begin, g), g) == begin);
+    if ( ! valid) {
+      verr << "    opposite pointer integrity corrupted."
+           << std::endl;
+      break;
+    }
+    // previous integrity.
+    valid = valid && ( prev(next(begin, g), g) == begin);
```
sloriot: you can also test `next(prev())`

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,207 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+    }
+
+    ++n;
+    if ( is_border(begin, g))
+      ++nb;
   }
-  BOOST_FOREACH(halfedge_descriptor h, halfedges(g)){
-    if(! is_valid_halfedge_descriptor(h,g)){
-      return false;
+  verr << "summe border halfedges (2*nb) = " << 2 * nb << std::endl;
```
sloriot: sum of ?

sloriot:
Does it make senses to keep `is_valid()` in `graph_traits_Surface_mesh.h`, `graph_traits_Linear_cell_complex_for_combinatorial_map.h`, `graph_traits_Polyhedron_3.h`, and `graph_traits_OpenMesh.h`? IMO it is confusing.

cc @afabri

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,207 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
 
-
-template <typename FaceGraph>
-bool is_valid_polygon_mesh(const FaceGraph& g)
+/*!
+  \ingroup PkgBGLHelperFct
+ * \brief is_valid_polygon_mesh checks the integrity of a `PolygonMesh`.
+ * \param g the `PolygonMesh` to test.
+ * \param verb : if `true`, the details of the check will be written in the standard output.
+ * \return `true` if the `PolygonMesh` is valid, `false` otherwise.
```
sloriot: \tparam `PolygonMesh` a model of `FaceListGraph` and `HalfedgeListGraph`

BGL/include/CGAL/boost/graph/helpers.h
maxGimeno: Probably. It is a copy/paste from Polyhedron_3::is_valid(). I guess I should change this one too

sloriot:
ping @afabri

lrineau:
ping @afabri, again: Sébastien asked a question at https://github.com/CGAL/cgal/pull/2757#issuecomment-359437380, 11 days ago.

afabri:
I guess that the intention of the `is_valid()` function is to do more internal consistency checks than just if the graph is a correct face graph.   But I agree that we can safely remove it, as it is not documented.

lrineau:
Could not we:
  - add and document that `CGAL::is_valid_polygon_mesh()`,
  - add a `CGAL::is_valid(mesh)` global function template, that would just call `CGAL::is_valid_polygon_mesh(mesh)`, and document it as well,
  - let each implementation of polyhedron mesh overload `CGAL::is_valid` with their own implementation if they want to check something more than `CGAL::is_valid_polygon_mesh()`.

Instead of overloading, we could even let implementation add `bool is_valid(Mesh)` and call it using ADL everywhere we want to use it:
```c++
  using CGAL::is_valid;
  CGAL_assertion(is_valid(mesh)); // call by ADL
```
That combination `using`+ADL is also how we are supposed to use `get` on a property-map:
```
 using boost::get;
 get(pmap, key);
```
For example:
https://github.com/CGAL/cgal/blob/2e7c3ee5b8c8bd2b91539636d52607abc6c607f4/Surface_mesh_shortest_path/include/CGAL/Surface_mesh_shortest_path/Surface_mesh_shortest_path.h#L2467-L2471

sloriot:
what's the point of having both?

lrineau:
As Andreas said in https://github.com/CGAL/cgal/pull/2757#issuecomment-363165181, the validity of a data structure model of `FaceGraph` could be more than just the result of `CGAL::is_valid_polygon_mesh()`. So `is_valid(mesh)` called by ADL could implement those extra checks (probably by calling an internal validity-testing method).

sloriot:
IMO it will simply introduce confusion. What should I call as a developer at the end of my algorithm?
`is_valid()`? Then why can't we simply overload `is_valid_polygon_mesh()` and always use that one?
Moreover, I don't think it makes sense because in practice the `is_valid()` member functions might assumes invariant that cannot be maintain only following face graph concepts (like `normalize_borders()` for `Polyhedron_3` for example or all border vertices having a border halfedge for `OpenMesh` and `Surface_mesh`).

lrineau:
You are right one the first point: `is_valid_polygon_mesh` should be the function that is overloaded, because `is_valid` is a too-generic name to be overloaded by a single argument.

---
For the second point:
> Moreover, I don't think it makes sense because in practice the `is_valid()` member functions might assumes invariant that cannot be maintain only following face graph concepts (like `normalize_borders()` for `Polyhedron_3` for example or all border vertices having a border halfedge for `OpenMesh` and `Surface_mesh`).

If CGAL BGL algorithms can break invariants that are so important the mesh would no longer be considered valid by the native validity check of the data structure, then I think we should know it, and try to fix the bug.

afabri:
Let's just remove it.

lrineau:
Is there still a discussion? Last comment https://github.com/CGAL/cgal/pull/2757#issuecomment-363388184 said:
> Let's just remove it.

but then there was no new commits. Do you all agree on what should be implemented?

maxGimeno:
There was no new commit because I removed it from the beginning, and the discusison was about wether I should put it back or not. I think he Small Feature was accepted yesterday or something.

lrineau:
Thanks. The PR will be tested tonight...

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,210 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
-      return false;
+  typedef typename boost::graph_traits<PolygonMesh>::halfedge_descriptor   halfedge_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_descriptor     vertex_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::face_descriptor       face_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_iterator       vertex_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::halfedge_iterator     halfedge_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::face_iterator         face_const_iterator;
+  typedef typename boost::graph_traits<PolygonMesh>::vertices_size_type    vertex_size_type;
+  typedef typename boost::graph_traits<PolygonMesh>::halfedges_size_type   halfedges_size_type;
+  typedef typename boost::graph_traits<PolygonMesh>::faces_size_type       faces_size_type;
```
lrineau: The [testsuite is a lot yellow, today](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-175/Convex_decomposition_3/TestReport_gimeno_Debian-Stable.gz):
```
In file included from /mnt/testsuite/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h:35:0,
                 from /mnt/testsuite/include/CGAL/Nef_polyhedron_3.h:57,
                 from /home/cgal_tester/build/src/cmake/platforms/Debian-Stable/test/Convex_decomposition_3/check_decomposition.cpp:3:
/mnt/testsuite/include/CGAL/boost/graph/helpers.h: In function 'bool CGAL::is_valid_polygon_mesh(const PolygonMesh&, bool)':
/mnt/testsuite/include/CGAL/boost/graph/helpers.h:374:76: warning: typedef 'vertex_const_iterator' locally defined but not used [-Wunused-local-typedefs]
   typedef typename boost::graph_traits<PolygonMesh>::vertex_iterator       vertex_const_iterator;
                                                                            ^~~~~~~~~~~~~~~~~~~~~
/mnt/testsuite/include/CGAL/boost/graph/helpers.h:375:76: warning: typedef 'halfedge_const_iterator' locally defined but not used [-Wunused-local-typedefs]
   typedef typename boost::graph_traits<PolygonMesh>::halfedge_iterator     halfedge_const_iterator;
                                                                            ^~~~~~~~~~~~~~~~~~~~~~~
/mnt/testsuite/include/CGAL/boost/graph/helpers.h:376:76: warning: typedef 'face_const_iterator' locally defined but not used [-Wunused-local-typedefs]
   typedef typename boost::graph_traits<PolygonMesh>::face_iterator         face_const_iterator;
                                                                            ^~~~~~~~~~~~~~~~~~~
```

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,29 +356,210 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+      break;
+  }
+  ++n;
+}
+valid = valid && (n == num_h);
+if ( n != num_h)
+  verr << "counting halfedges failed." << std::endl;
+
+verr << "end of CGAL::Polyhedron_3<...>::is_valid(): structure is "
+   << ( valid ? "valid." : "NOT VALID.") << std::endl;
```
lrineau: > end of CGAL::Polyhedron_3<...>::is_valid()

That message does not seem right.

maxGimeno:
@gdamiand This PR has revealed a bug in LCC : when making an LCC out of BGL/test/BGL/data/7_faces_triangle.off, there is a face that is incident to itself. Can you look into it please ?

lrineau:
This PR will probably not make it into CGAL-4.12...

gdamiand:
@maxGimeno I can have a look; where is the code that shows the bug please ?

maxGimeno:
@gdamiand in the tetss of BGL, in that branch, if you run test_graph_traits it should fail.

gdamiand:
@maxGimeno I had a look at the bug; still searching why there is an edge in LCC incident twice to the same face.

BUT: I think you need to be careful when adding new constraints. For example, an edge incident twice to a same face is valid, this is a dangling edge and sometimes used. Another example, take one triangle, and close the HDS, you obtain two triangles superposed. This is a valid HDS (maybe not a valid polyhedron, but you need to define precisely what is a valid polyhedron, and I think we do not test auto-intersection as validity constraint in Polyhedral surface, do we ?)

I think it is a good thing to differentiate real validity constraints, imposed by the definition of the data structures, and other constraints required by some algorithms. Maybe this can be discussed during the cgal dev meeting ?

gdamiand:
I found the problem (a bug when copying of a combinatorial with a special type of attribute; I am creating a pr for that).

By the way, I suggest to add a vs.reserve(sizeof(data)) [here](https://github.com/CGAL/cgal/blob/master/BGL/test/BGL/test_Prefix.h#L135) to avoid copies during vector re-allocation.

maxGimeno:
@gdamiand I added some doc defining the validity of a PolygonMesh to the function. A PolygonMesh is a model of FaceListGraph, HalfedgeListGrpah and VertexListGraph. An HDS is not necessarily a PolygonMesh then, and the presence of a dangling edge makes a FaceGraph invalid (although it is currently missing from the FaceGraph concept for some reason)

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -362,8 +362,43 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+- std::distance(halfedges(g).first, halfeges(g).second)) is the number of halfedges of `g`
+- for each vertex `v` of `g`:
+  - halfedge(v,g) != null_halfedge()
+  - target(halfedge(v,g),g) == v
+
+- the sum of the halfedges around all the vertices of `g` is 
+std::distance(halfedges(g).first, halfeges(g).second))
+- std::distance(vertices(g).first, vertices(g).second)) is the number of vertices of `g`
+- for each face `f` of `g`:
+  - halfedge(f,g) != null_face()
```
MaelRL: null_halfedge()

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -362,8 +362,43 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+  - face(halfedge(f,g),g) == f
+
+- the sum of halfedges around all the faces of `g` + the border halfedges of `g` is 
+std::distance(halfedges(g).first, halfeges(g).second))
+- The faces of `g` are at least triangles.
+- `g` has distinct faces on each side of an edge.
```
lrineau: Do we really want that level of precision? That description is almost as precise as the code of `is_valid_polygon_mesh` itself. Is not there a description of a polygon mesh somewhere in the BGL package?!

lrineau:
Travis error because of a missing `#include`:
```
[ 55%] Building CXX object CMakeFiles/test_pmp_remove_border_edge.dir/test_pmp_remove_border_edge.cpp.o
/home/travis/build/CGAL/cgal/Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_remove_border_edge.cpp:35:27: error: 
      use of undeclared identifier 'is_valid_polygon_mesh'
  CGAL_assertion(is_valid(is_valid_polygon_mesh));
                          ^
/usr/local/lib/cmake/CGAL/../../../include/CGAL/assertions.h:102:20: note: 
      expanded from macro 'CGAL_assertion'
   (CGAL::possibly(EX)?(static_cast<void>(0)): ::CGAL::assertion_fail( ...
                   ^
```
See https://travis-ci.org/CGAL/cgal/jobs/343275618#L7846

lrineau:
The Travis check might be fixed after a rebase on `master`.

maxGimeno:
rebased

lrineau:
https://github.com/CGAL/cgal/pull/2851 has been merged. So now the CMap test should pass.

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2757)

MaelRL:
Could it be possible to split the validity function to separate a valid Halfedge graph from a valid Facegraph (the latter naturally including the former)?

It would be useful to have to test the validity of upcoming structures like the motorcycle graphs (which are not models of Facegraph), and might also solve some problems described above?

Definition of a valid Facegraph can be found in PMP, and for a valid halfedge graph, it is something like pointer coherence and finite cycles.

sloriot:
I agree with @MaelRL

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -358,12 +358,12 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
 
 /*!
   \ingroup PkgBGLHelperFct
- * \brief is_valid_face_graph checks the integrity of a `HalfedgeGraph`.
- * \param g the `HalfedgeGraph` to test.
+ * \brief checks the integrity of `g`.
+ * \param g the HalfedgeGraph to test.
```
MaelRL: `HalfedgeListGraph` ?

Anyway, just call the template `Graph`, and then
```
\brief Test the validity of the graph `g`. [...]
\tparam Graph is a model of `HalfedgeListGraph`
\param g is an object of type `Graph` whose validity is being tested.
```

The `\brief` also needs to describe what is a valid halfedge(list)graph: finite halfedge loops, and reciprocity of associations (see for example the invariants in the BGL doc).

Same for others below.

maxGimeno: I thought the very interest of having links to a Concept in the doc was to avoid having to re-define it every time we refer to it. Why should I "describe what is a valid halfedge(list)graph" when it is already done in the Concept ?if it is not, maybe it is the Concept page that should be modified.

sloriot: Basically the concept will only tell you what type and functions should be defined for the graph type. I think mentioning the reciprocity with just one example (next <-> prev for example) is sufficient (you get the finite loop for free).

sloriot:
conflicts

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -355,31 +356,293 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
...
+  if ( valid && f != num_f)
+    verr << "counting faces failed." << std::endl;
+  
+  BOOST_FOREACH(halfedge_descriptor i, halfedges(g)){
+    //counting borders
+    if ( is_border(i, g))
+      ++nb;
+    // face integrity.
+    
+    valid = valid && ( face(i, g) == face(next(i, g), g));
```
maxGimeno: @sloriot That should check that all halfedges in a loop point to the same face, shouldn't it ?

sloriot: Great, thanks!

lrineau:

PR #2759, opened 2018-01-22
Polyhedron demo: small new features

### LINKED ISSUE #2390: Polyhedron demo: minor feature requests
@maxGimeno, please find below a bunch of feature requests / bug reports for the polyhedron demo. Thanks!

* It would be nice to be able to reload all selected items.
* More generally, right click should apply -- if possible -- the desired action  (e.g. set colors, set flat/edge/etc., ...) to all selected items, not just whichever is under the cursor.
* ctrl-R / F5 as shortcut to reload items ?
* Some better management of many small items would be nice: handling 100 items which all contain a single segment is very slow.
* It would be helpful to have the possibility to visualize the canonical axes and the origin; similarly, to be able to draw a (xOy) grid (see medit shortcuts 'A' and 'G') through the origin (same thing as the cut plane, without cutting) 
* The color preview is rather unfaithful to the final result. Maybe an option to disable whichever lighting changes the colors so much so that, for example, the white color actually produces white and not dark grey (especially true when `draw two side` is unselected). 
* Clicking a point (loaded from a `.xyz)` with ctrl-click should give the position of the point, not the position of w/e sub-part of the drawn square that represents the point I am clicking (something like that should exist in the selection plugin, I guess).

Minor bugs:
* View > Dock windows > Some windows are duplicated ?
* On the slider for point size, the first click seems bugged (it just closes the slider) and one must first slide before clicking.

### LINKED ISSUE #2768: Polyhedron demo: clipping box surprising behaviors
## Issue Details

I'm using the Clipping Box plugin with a point set.

1) To move the clipping box faces/edges/etc., there is no modifier key. It could be nice that it only moved when holding `shift` for example (holding `ctrl` makes it turn around, which is good). I don't know if it is a wanted behavior or not, but that is kind of related to the next point.

2) I move the clipping box around, click on "clip", then hide the clipping box, when I try to turn around my point set with click+hold, it doesn't work because it actually moves the hidden clipping box. This is not really user-friendly, it would be better if the box was ignored if it is hidden (note that if the box was only moved when using a modifier key, it would not really be a problem, see previous point).

3) If I deleted the clipping box, the clipping remains active: it can be quite surprising, for example when you delete all items of a scene (including a clipping box) and load another object, you can sometimes see _nothing_ (if the new object is out of the clipping box). Clipping should be switched-off if the clipping box is deleted.

maxGimeno:
## Summary of Changes
- [x] It would be nice to be able to reload all selected items.
- [x] More generally, right click should apply -- if possible -- the desired action (e.g. set colors, set flat/edge/etc., ...) to all selected items, not just whichever is under the cursor.
- [x] ctrl-R / F5 as shortcut to reload items ? -> F5, ctrl-R is taken.
- [ ] Some better management of many small items would be nice: handling 100 items which all contain a single segment is very slow. -> It will have to wait for the new API, that comes with a better loading system.
- [x] It would be helpful to have the possibility to visualize the canonical axes and the origin; similarly, to be able to draw a (xOy) grid (see medit shortcuts 'A' and 'G') through the origin (same thing as the cut plane, without cutting)
- [ ] The color preview is rather unfaithful to the final result. Maybe an option to disable whichever lighting changes the colors so much so that, for example, the white color actually produces white and not dark grey (especially true when draw two side is unselected). -> The dark gray is just what white looks light without lighting, we (at least Sebastien, which made it pretty clear) want to keep the difference between inside and outside faces when "draw two sides" is off. For the color being too whight, it is the visual for the selection that makes that. New things are being tested in the new API branch.
- [x] Clicking a point (loaded from a .xyz) with ctrl-click should give the position of the point, not the position of w/e sub-part of the drawn square that represents the point I am clicking (something like that should exist in the selection plugin, I guess). -> Added as a feature in the Point_set_selection-plugin.
- [x] View > Dock windows > Some windows are duplicated ?
- [ ] On the slider for point size, the first click seems bugged (it just closes the slider) and one must first slide before clicking. --> That I don't reproduce.
- [x] Zoom to index should not prompt the error when "cancel" is used
- [x] Display [vertices/faces/etc] IDs are marked as selected when some (e.g. OFF) item is loaded, but no items are shown. Requires to deselect and reselect for the IDs to show up.
- [x] When an item is refreshed, indices disappear (as they should), but Right click > Display all IDs is not switched to off (and so you have to click twice to enable them on the new item).
- [x] If you choose to show tetrahedra on a loaded .mesh, you cannot create a Bbox of the mesh anymore (which should probably be called Create Mesh Bbox rather than Create Bbox Mesh (?)).
Actually, the Create Mesh Bbox is moved to the sub (tetrahedra) item. Clicking it gives:
bbox dimensions: -inf
                 -inf
                 -inf
Floating point exception (core dumped) --> Create Bbox Mesh creates a mesh of the Bbox, I find it more adapted than create mesh bbox, which makes me think more of creating a bbox for the mesh.
- [x] After an item has been reloaded, the cut plane cannot be moved without deselecting and reselecting the item.
- [x] When "Saving as", the file type OFF files to Surface_mesh doesn't really make sense. I guess it's a remnant from reading.
- [x] Surface mesh items have number of faces while Polyhedron items have number of facets.
- [x] Polyhedron items have an extra field number of isolated vertices that is not mirrored for surface mesh items.
- [x] I feel like the cut plane's Show Tetrahedra should be enabled by default when generating a 3D mesh.
- [ ] It would be nice to have some more operations on the c3t3 items, such as extract point set, etc.
- [x] Output formats should always show the file extension or accept anything: ascii/binary should show that the extension is .mesh or not reject me when I try to save in a .txt.
- [x] you can't save a c3t3 item if Show Tetrahedra is enabled, you've got to deselect it first (otherwise save doesn't do anything). Maybe related to that bug with show bbox in a previous comment.
- [x]  refreshing a point set resets the point size.
## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #2390 fix #2768


Polyhedron/demo/Polyhedron/Plugins/PCA/Clipping_box_plugin.cpp
```diff
@@ -88,6 +88,12 @@ dock_widget->show();
     item = new Scene_edit_box_item(scene);
   connect(item, SIGNAL(destroyed()),
           this, SLOT(enableAction()));
+  connect(item, &Scene_edit_box_item::aboutToBeDestroyed,
+          [this]()
+  {
+    clip(false);
+    dock_widget->pushButton->setChecked(false);
+  });
```
lrineau: When you use that new version of `connect`, use anyway the version with four arguments:
```c++
connect(item, &Scene_edit_box_item::aboutToBeDestroyed,
        this, [this]()
{
  clip(false);
  dock_widget->pushButton->setChecked(false);
});
```

The extra argument (the third, here `this`), ensures that if the targeted
object of the connection is destroyed, then the connection is removed just
before. Otherwise the application could run a lambda expression with
dandling references/pointers to objects that were destroyed.

Well, in this case, that was safe, because it is unlikely, because `this`
is a plugin of the application, and we do not have any way to unload and
destroy a plugin before the exit of the application.

lrineau:
New warnings in the Polyhedron demo:
```
C:\CGAL\test\CGAL-4.12-Ic-163\cmake\platforms\x64_Cygwin-Windows10_MSVC2013-Debug-64bits\test\Polyhedron_Demo\Viewer.cpp(1812) : warning C4267: 'argument' : conversion from 'size_t' to 'int', possible loss of data
C:\CGAL\test\CGAL-4.12-Ic-163\cmake\platforms\x64_Cygwin-Windows10_MSVC2013-Release-64bits\test\Polyhedron_Demo\Viewer.cpp(1812) : warning C4267: 'argument' : conversion from 'size_t' to 'int', possible loss of data
C:\CGAL\test\CGAL-4.12-Ic-163\cmake\platforms\x64_Cygwin-Windows10_MSVC2017-Debug-64bits\test\Polyhedron_Demo\Viewer.cpp(1812): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
C:\CGAL\test\CGAL-4.12-Ic-163\cmake\platforms\x64_Cygwin-Windows10_MSVC2017-Release-64bits\test\Polyhedron_Demo\Viewer.cpp(1812): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
/home/cgal_tester/build/src/cmake/platforms/Debian-Stable/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Debian-stable-Release/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Debian-Testing/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Fedora-32/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Fedora-32-Release/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Fedora/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Fedora-Release/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Ubuntu-latest-GCC6-CXX1z/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Ubuntu-latest-GCC6/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
/home/cgal_tester/build/src/cmake/platforms/Ubuntu-latest-GCC6-Release/test/Polyhedron_Demo/Scene_points_with_normal_item.cpp:173:3: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
```
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-163.shtml#Polyhedron_Demo

afabri:
I would even advocate for always putting parenthesis.

lrineau:
Tested in [CGAL-4.12-Ic-164][0]. Then merged.

  [0]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-164.shtml

PR #2760, opened 2018-01-22
Restore aliases for doxygen 1.8.13 and add new resources for 1.8.14

sloriot:
The replacement of `\n` by `^^` in aliases was not compatible with doxygen 1.8.13.

The doc testsuite should probably also report error and warnings for the other tested versions.


lrineau:
Cannot Doxygen 1.8.13 use resources from 1.8.14?

sloriot:
Not `BaseDoxyfile.in`, it seems `^^` was introduced in 1.8.14

lrineau:
Tested in [CGAL-4.12-Ic-159][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-159.shtml

Documentation/doc/resources/1.8.14/deprecated.html
```diff
@@ -0,0 +1,146 @@
...
+</div></div><!-- contents -->
+</div><!-- doc-content -->
+<!-- start footer part -->
+<div id="footer">
+<div id="nav-path" class="navpath"><!-- id is needed for treeview function! -->
+  <ul>
+    <li class="footer">Generated on Fri Aug 9 2013 12:26:14 for CGAL 4.3 - Manual by
+    <a href="http://www.doxygen.org/index.html">
+    <img class="footer" src="doxygen.png" alt="doxygen"/></a> 1.8.4 </li>
+  </ul>
```
lrineau: What is the purpose of that static page `deprecated.html`?

I can see that the file is copied from the resources by our postprocessing.
https://github.com/CGAL/cgal/blob/7d3a5ba1742e6ee4a9603cd2e36917a560dc3d76/Documentation/doc/scripts/html_output_post_processing.py#L375-L376

But it seems that page was once (`Fri Aug 9 2013 12:26:14`) generated by Doxygen 1.8.4, from CGAL-4.3. If the page is still useful, could its content could be regenerated instead of being in the resources as three static copies of the same file?

sloriot: IIRC we put some manual redirects when a package, function, or class disappears in the doc website. We added that when we tested that all former pages were pointing to something valid (again IIRC).

PR #2763, opened 2018-01-23
Use new URL for MathJax

### LINKED ISSUE #2633: CGAL Doxygen doc: MathJax CDN shutting down
https://www.mathjax.org/cdn-shutting-down/

> The MathJax CDN hosted at cdn.mathjax.org will be shutting down on April 30, 2017. 
Current and future releases available on other CDN providers.

That did not break about usage because:

> [Updated 2017/04/10]
>
> We will implement a temporary redirect on cdn.mathjax.org to redirect users to cdnjs.

We will have to do something in our Doxygen configuration, because the 
temporary redirection from cdn.mathjax.com to another provider will 
disappear.

As far as I remember, the documentation at https://doc.cgal.org/ will not break, because we ship a copy of MathJax. But all locally compiled/downloaded documentation will have the issue.

sloriot:
## Release Management

* Affected package(s): doc
* Issue(s) solved (if any): fix #2633

### DIFF
--- a/Documentation/doc/CMakeLists.txt
+++ b/Documentation/doc/CMakeLists.txt
@@ -185,7 +185,7 @@ function(configure_doxygen_package CGAL_PACKAGE_NAME)
 endfunction()
 
 # set up the directories and variables
-set(CGAL_DOC_MATHJAX_LOCATION "https://cdn.mathjax.org/mathjax/latest"
+set(CGAL_DOC_MATHJAX_LOCATION "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2"
     CACHE STRING "The location of MathJax to be used for the documentation.")
 
 set(CGAL_DOC_OUTPUT_DIR "${CMAKE_BINARY_DIR}/doc_output")
--- a/Documentation/doc/resources/1.8.13/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.8.13/BaseDoxyfile.in
@@ -1568,7 +1568,7 @@ MATHJAX_FORMAT         = HTML-CSS
 # Content Delivery Network so you can quickly see the result without installing
 # MathJax. However, it is strongly recommended to install a local copy of
 # MathJax from http://www.mathjax.org before deployment.
-# The default value is: http://cdn.mathjax.org/mathjax/latest.
+# The default value is: https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2.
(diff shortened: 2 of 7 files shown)


Documentation/doc/CMakeLists.txt
```diff
@@ -185,7 +185,7 @@ function(configure_doxygen_package CGAL_PACKAGE_NAME)
 endfunction()
 
 # set up the directories and variables
-set(CGAL_DOC_MATHJAX_LOCATION "https://cdn.mathjax.org/mathjax/latest"
+set(CGAL_DOC_MATHJAX_LOCATION "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2"
```
lrineau: Do we really want to set that variable? Recent versions of Doxygen have a default for `MATHJAX_RELPATH` that is already right. And we can trust the Doxygen author better than us for upgrading that version when new versions of MathJax are published.

sloriot: AFAIS, this is only use in resource files for doxygen 1.8.4, so this should be fine.

lrineau: If I grep the variable name, I see:
```
doc/resources/1.8.13/BaseDoxyfile.in
1574:MATHJAX_RELPATH        = ${CGAL_DOC_MATHJAX_LOCATION}
```
Is not that Doxygen configuration variable used?

sloriot: You're right, I fixed that in #2760 in 48658fe5f6 (since new resources files were added for 1.8.14).

lrineau:
I do not know how to merge this PR in integration. There is a big conflict with `master` in `Documentation/doc/scripts/test_doxygen_versions.sh`.

lrineau:
I have add a look at the result at https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.12-Ic-159/index.html and I have the impression that the CDN of MathJax is never used.

sloriot:
In the testsuite we are using a local version of MathJax and not the online one.

lrineau:
Alright. Now I understand: the cdn version of MathJax will only be used in:
  - local Doxygen builds by users or developers,
  - in the doc_html tarballs...
but that is not what we test automatically.

lrineau:
@sloriot Can you please merge this PR in `master` and in the release branch?

sloriot:
@lrineau: done.

PR #2765, opened 2018-01-24
Documentation testsuite : Add results and overview for all three doxygen versions

maxGimeno:
## Summary of Changes
The result of the script now stores logs of all three versions and display them.

### DIFF
--- a/Documentation/doc/scripts/process_doc.sh
+++ b/Documentation/doc/scripts/process_doc.sh
@@ -23,24 +23,24 @@ DOXYGEN_2=$($PATH_TO_1_8_13 --version)
 ## download and build doxygen_master ##
 #######################################
 echo "downloading and building master"
-  git clone https://github.com/doxygen/doxygen.git doxygen_master  1> /dev/null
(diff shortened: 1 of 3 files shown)


sloriot:
doc [testsuite page](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.12-Ic-157/) leads to a 404,

lrineau:
With the fix, the overview page seems rather OK:
spectacle y59464

I have to include new title headers.

But the result page does not seem optimal:
spectacle h59652

I suggest having one first column with the package names (without links), and then, for each version of Doxygen that is tested:
- a column with the logs (with links named "logs"),
- a column with the number of warnings,
- a column with the number of errors.
and no need to repeat the version of Doxygen in each column header: just the first.

lrineau:
Tested in [CGAL-4.12-Ic-159][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-159.shtml

PR #2776, opened 2018-02-01
Shape Detection: Fix Infinite Loop

sgiraudot:
## Summary of Changes

There exists a very specific an rare case in RANSAC where no new candidate shapes are generated but no candidates are integrated nor discarded, which leads to a constant stop criterion, which leads to an infinite loop.

This is the least invasive fix that I managed to do: if the search is over but no shape was either integrated or discarded, we increase the stop criterion to unlock the situation. It works experimentally and does not modify other cases (and statistically, RANSAC gives the same distribution of shapes and assigned points with and without the fix).

## Release Management

* Affected package(s): Point Set Shape Detection 3

### DIFF
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
@@ -734,6 +734,9 @@ shape. The implementation follows \cgalCite{schnabel2007efficient}.
   
             candidates.resize(end);
           }
+        else if (!keep_searching)
+          ++ generated_candidates;
+        
         keep_searching = (stop_probability(m_options.min_points,
             m_num_available_points - num_invalid,
             generated_candidates, 


lrineau:
There is a conflict with `master`. I have merged the PR anyway in `integration`, using that conflict resolution:
```patch
diff --cc Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
index 0badac4391e,259124e5d97..00000000000
--- a/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
+++ b/Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Efficient_RANSAC.h
@@@ -794,9 -734,8 +794,11 @@@ shape. The implementation follows \cgal
    
              candidates.resize(end);
            }
+         else if (!keep_searching)
+           ++ generated_candidates;
 +
 +        if (callback && !callback(num_invalid / double(m_num_total_points)))
 +          return false;
          
          keep_searching = (stop_probability(m_options.min_points,
              m_num_available_points - num_invalid,
```

Do you agree with that resolution?

sgiraudot:
Yes, the only thing that changed in `master` is the callbacks line, and they do not affect this fix, so you did the right resolution.

lrineau:
Tested in [CGAL-4.12-Ic-164][0]. Then merged.

  [0]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-164.shtml

PR #2784, opened 2018-02-02
TravisCI:  check package dependencies

### LINKED ISSUE #2730: Check of package dependencies
1. Modify the `check_headers` feature of our CMake scripts:
    1. Doxygen is now required, and the list of documented headers is
    computed at each run.
    1. Replace `ack` by `awk` in the script
    `Scripts/developer_scripts/generate_list_of_documented_headers`, and
    replace that script by a logic directly in CMake (fill a CMake list
    instead), using [`execute_process`]
    1. make `CGAL_ENABLE_CHECK_HEADERS` imply `CGAL_COMPUTE_DEPENDENCIES` (because all the work is already done),
    1. `CGAL_COMPUTE_DEPENDENCIES` should use `-E` instead of `-fsyntax-only`,
      if `CGAL_ENABLE_CHECK_HEADERS` is false.
    1. remove the build directory from the list of dependencies
    1. add an option so that `package_info/*/dependencies` are copied into the
      source directories.
1. Add a script that checks if dependencies have been updated.
1. Call it in Travis
1. Change `.travis/build_package.sh`, so that a package (but Polyhedron_demo) is only built if its dependencies have been modified by the pull-request (only if Travis tests a PR).

[`execute_process`]: https://cmake.org/cmake/help/v3.0/command/execute_process.html

### LINKED ISSUE #2740: Travis fails way too often on `apt-get`

maxGimeno:
## Summary of Changes
This PR adds automation for checking the include logic in the documented headers. 
Using it, it is possible to compute the dependencies for all packages and write the list in each package.
This allow to Travis to check that the list is up to date, and to discard any job that is not concerned by the currently tested PR.

## Release Management

* Issue(s) solved (if any): fix #2730, fix #2740


AABB_tree/package_info/AABB_tree/dependencies
```diff
@@ -0,0 +1,19 @@
+
```
sloriot: Any reason to have an empty line in all the files?

Classification/include/CGAL/Classification/Feature/Vertical_dispersion.h
```diff
@@ -29,6 +29,9 @@
 #include <CGAL/Classification/Image.h>
 #include <CGAL/Classification/Planimetric_grid.h>
 #include <boost/algorithm/minmax_element.hpp>
+#include <CGAL/Classification/Feature_base.h>
+#include <CGAL/int.h>
+#include "boost/tuple/tuple.hpp"
```
sloriot: "" -> <>

Intersections_3/include/CGAL/internal/Intersections_3/Bbox_3_Triangle_3_do_intersect.h
```diff
@@ -54,7 +54,8 @@ namespace internal {
         }
         else {
           if(p[i] <= r[i]) { // prq
-            if((bbox.max)(i) < p[i] || (bbox.min)(i) > q[i])
+            if(
+               (bbox.max)(i) < p[i] || (bbox.min)(i) > q[i])
```
sloriot: ?

maxGimeno: ?????

afabri: ? == Why a newline

AABB_tree/package_info/AABB_tree/dependencies
maxGimeno: Yes, it is because we create the file empty, it creates an empty line. I guess it is to ensure that the file exists even if there is no dependency.

Intersections_3/include/CGAL/internal/Intersections_3/Bbox_3_Triangle_3_do_intersect.h
maxGimeno: ????? == What the hell is this change doing here ?

lrineau:
Does not build:
```
  /home/travis/build/CGAL/cgal/Classification/include/CGAL/Classification/Image.h:39:16:
  error: no template named 'map' in namespace 'std'; did you mean 'max'?

    typedef std::map<std::size_t, Type> Map;
            ~~~~~^~~
                 max
```
https://api.travis-ci.org/v3/job/336540521/log.txt

Cc @sgiraudot

Documentation/doc/Documentation/packages.txt
```diff
@@ -37,136 +37,4 @@ h1 {
...
-\package_listing{Circulator}
-\package_listing{Generator}
-\package_listing{Miscellany}
-\package_listing{Stream_support}
-
-\section PartVisualization Visualization
-
-\package_listing{Geomview}
-\package_listing{GraphicsView}
-\package_listing{CGAL_ipelets}
```
lrineau: I cannot approve that change!

maxGimeno: That's not supposed to be here anyways

maxGimeno:
Fixed

Installation/cmake/modules/process_dependencies.cmake
```diff
@@ -8,13 +8,17 @@ if(NOT CGAL_PACKAGES_PREFIX)
 endif()
 
 #message("regexp: \\. ${CGAL_PACKAGES_PREFIX}/[^/]*/include/CGAL/.*h")
+#TO BE CHECKED
+get_filename_component(INSTALLATION "${CMAKE_BINARY_DIR}" DIRECTORY)
+get_filename_component(BUILD_DIR "${INSTALLATION}" NAME)
```
lrineau: Could you comment what has to be checked?

maxGimeno: I think I just forgot to remove the commentary. It has been checked already.

lrineau: Thanks for the answer. I have pushed 3457d2442b603aac2f643711655268b3ea566345 that removes that line.

maxGimeno:
@lrineau it is explained in the Scripts/developer_scripts/cgal_check_dependencies.sh which is displayed in the travis CHECK job for the users to see : 
" You should run cmake with option CGAL_CHECK_HEADERS ON, make the target packages_dependencies and commit the new dependencies files."

Installation/cmake/modules/process_dependencies.cmake
maxGimeno: I overrid it with more clean-up

Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Region_growing.h
```diff
@@ -34,8 +34,10 @@
 #include <CGAL/squared_distance_3.h>
 #include <CGAL/function.h>
 
-#include <CGAL/boost/iterator/counting_iterator.hpp>
-#include <CGAL/boost/iterator/transform_iterator.hpp>
+#include <boost/iterator/counting_iterator.hpp>
+#include <boost/iterator/transform_iterator.hpp>
```
lrineau: That patch is probably not right. @sgiraudot, can you comment it? I think the new headers `<CGAL/boost/...>` are used to remove warnings...

lrineau:
https://github.com/CGAL/cgal/pull/2784#pullrequestreview-95355794
> Could you explain here what a CGAL developer has to do locally if Travis says that dependencies have changed?

@maxGimeno, https://github.com/CGAL/cgal/pull/2784#issuecomment-364380445:
> @lrineau it is explained in the Scripts/developer_scripts/cgal_check_dependencies.sh which is displayed in the travis CHECK job for the users to see :
" You should run cmake with option CGAL_CHECK_HEADERS ON, make the target packages_dependencies and commit the new dependencies files."

Good. Maybe you should mention the new CMake option `CGAL_COPY_DEPENDENCIES` as well.

Point_set_shape_detection_3/include/CGAL/Shape_detection_3/Region_growing.h
maxGimeno: Looks I made a mistake on a conflict resolution.

maxGimeno:
> Good. Maybe you should mention the new CMake option CGAL_COPY_DEPENDENCIES as well.

You are right !

lrineau:

lrineau:
Now it works correctly. Example of a PR touching only demo/Polyhedron/:
https://travis-ci.org/CGAL/cgal/builds/362173402

Less than 19 minutes (instead slightly less than 2 hours). Unfortunately, our installation script takes between 3 and 4 minutes, for each job. Maybe we should move the part that check for dependencies during the installation step.

maxGimeno:
That seems like a good idea. I'll look into it.

maxGimeno:
But I'm pretty sure the less than 19min is not the whole time, as CHECK alone took 39min. I guess an isolated job has been relaunched at some point.

lrineau:
You are right. Maybe the checking of headers should only check headers from packages in the list of modifies packages (or depending on modified packages).

PR #2785, opened 2018-02-02
Triangulation_2: Make CDT_2 less verbose

### LINKED ISSUE #2752: Why is CT_2 so verbose
In the code of `Constrained_triangulation_2` we have the following:

```
  std::cerr << "You are using an exact number types" << std::endl;
  std::cerr << "using a Constrained_triangulation_plus_2 class" << std::endl;
  std::cerr << "would avoid cascading intersection computation" << std::endl;
  std::cerr << " and be much more efficient" << std::endl;
```

in the body of the insert function. I don't see why we have this on.
I'm in a setting where I'm not supposed to have a lot of intersection of constraints so I'd like to stick to my CDT type and not move to CDT+

Do you agree that we should remove it?

maxGimeno:
## Summary of Changes

This PR makes the CDT_2 less verbose by moving some error streamed information to a compilation warning, and it adds an overload for the `intersect()` function that use a new tag to throw an exception when the constrained edges intersect.
## Release Management

* Affected package(s):Triangulation_2
* Issue(s) solved (if any): fix #2752 
* Feature/Small Feature (if any): [New_Exception_in_Constrained_triangulation_2](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/New_Exception_in_Constrained_triangulation_2)

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -1,6 +1,13 @@
 Release History
 ===============
 
+Release 4.13
+------------
+### 2D Triangulations
+
+-   Added a new type of intersection to deal with insertion of a constraints 
+    intersecting in a Constrained_triangulation_2.
+
 Release 4.12
 ------------
 
--- a/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
+++ b/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
@@ -46,9 +46,12 @@ There are three versions of constrained triangulations
 <LI>In the basic version, the constrained triangulation 
 does not handle intersecting constraints, and the set of input 
 constraints is required to be a set of polylines that do not intersect 
-except possibly at their endpoints. Any number of constrained edges 
-may share the same endpoint. Constrained edges may be vertical or
-have zero length. 
+except possibly at their endpoints.
+An exception of type `Intersection_of_constraints_exception`
+is thrown upon insertion of a constraint intersecting in
+its interior an already inserted constraint.
+Any number of constrained edges may share the same endpoint.
+Constrained edges may be vertical or have zero length.
 <LI>The two other versions support intersecting input constraints. 
 In those versions, input constraints may consist of 
 intersecting, overlapping or partially 
@@ -329,7 +332,9 @@ Note that `t` is first cleared.
 */ 
 istream& operator>>(istream& is,Constrained_triangulation_2<Traits,Tds> Ct& t); 
 
-//! Exception thrown for `intersect()` when called with the tag `No_intersection_tag`.
-//! Then, if constraints intersect, an exception of this type will be thrown.
+/*! Exception used by constrained triangulations configured with
+the tag `No_intersection_tag`. It is thrown upon insertion of a constraint
+that is intersecting an already inserted constraint in its interior.
+*/
 class Intersection_of_constraints_exception;
 } /* end namespace CGAL */
--- a/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
+++ b/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
@@ -13,7 +13,7 @@ struct No_intersection_tag{};
 \ingroup PkgTriangulation2TriangulationClasses
 
 Intersection tag for constrained triangulations, when input constraints should not intersect.
-Will throw a `To_be_renamed_exception` if constraints do intersect.
+Will throw a `Intersection_of_constraints_exception` if constraints do intersect.
 */
 struct Throw_on_intersection_tag{};
 
@@ -338,5 +338,5 @@ istream& operator>>(istream& is,Constrained_triangulation_2<Traits,Tds> Ct& t);
 
 //! Exception thrown for `intersect()` when called with the tag `Throw_on_intersection_tag`.
 //! Then, if constraints intersect, an exception of this type will be thrown.
-class To_be_renamed_exception;
+class Intersection_of_constraints_exception;
 } /* end namespace CGAL */
--- a/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
+++ b/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
@@ -9,6 +9,13 @@ Intersection tag for constrained triangulations, when input constraints do not i
 */
 struct No_intersection_tag{};
 
+/*!
+\ingroup PkgTriangulation2TriangulationClasses
+
+Intersection tag for constrained triangulations, when input constraints should not intersect.
+Will throw a `To_be_renamed_exception` if constraints do intersect.
+*/
+struct Throw_on_intersection_tag{};
 
 /*!
 \ingroup PkgTriangulation2TriangulationClasses
@@ -328,4 +335,8 @@ Note that `t` is first cleared.
 \relates Constrained_triangulation_2 
 */ 
 istream& operator>>(istream& is,Constrained_triangulation_2<Traits,Tds> Ct& t); 
+
+//! Exception thrown for `intersect()` when called with the tag `Throw_on_intersection_tag`.
+//! Then, if constraints intersect, an exception of this type will be thrown.
+class To_be_renamed_exception;
 } /* end namespace CGAL */
--- a/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
+++ b/Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
@@ -9,13 +9,6 @@ Intersection tag for constrained triangulations, when input constraints do not i
 */
 struct No_intersection_tag{};
 
-/*!
-\ingroup PkgTriangulation2TriangulationClasses
-
-Intersection tag for constrained triangulations, when input constraints should not intersect.
-Will throw a `Intersection_of_constraints_exception` if constraints do intersect.
-*/
-struct Throw_on_intersection_tag{};
 
 /*!
 \ingroup PkgTriangulation2TriangulationClasses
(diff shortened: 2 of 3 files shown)


Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -41,6 +42,7 @@
 namespace CGAL {
 
 struct No_intersection_tag{};
+struct Throw_on_intersection_tag{};
```
sloriot: The doc should be updated

Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -101,6 +103,16 @@ class Constrained_triangulation_2
     }
   };
 
+
+  //! Exception thrown for `intersect()` when called with the tag `Throw_on_intersection_tag`.
+  class Is_intersecting_exception : public std::exception
```
sloriot: The class should be updated. I don't  like the name. @lrineau/@afabri (as older maintainers of the package) What do you think of `Intersection_of_constraints_exception`.

sloriot: + it should be documented

lrineau:
For a start, I do not understand the reason why we need a new tag that modifies the behavior of the CDT_2 (throw an exception instead of a `CGAL_assertion`). Sébastien, can you develop why you requested that? In the issue #2752, we were only talking about adding a compile-time warning.

I am not completely against that idea, but I would like to understand the need for it.

sloriot:
I wanted to preserve backward compatibility of the behavior. If you think it is acceptable to throw while using the `No_intersection_tag` it is even better.

lrineau:
Now I understand.

In this PR, you not longer want to fix #2752, but somehow officialize what was already possible using the undocumented macro `CGAL_CDT2_EXTRA_ACTION_FOR_INTERSECTING_CONSTRAINTS` in the code below:
https://github.com/CGAL/cgal/blob/2e7c3ee5b8c8bd2b91539636d52607abc6c607f4/Triangulation_2/include/CGAL/Constrained_triangulation_2.h#L851-L869

Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -881,10 +909,12 @@ intersect(Face_handle f, int i,
...
-  std::cerr << "You are using an exact number types" << std::endl;
-  std::cerr << "using a Constrained_triangulation_plus_2 class" << std::endl;
-  std::cerr << "would avoid cascading intersection computation" << std::endl;
-  std::cerr << " and be much more efficient" << std::endl;
+  CGAL_warning_msg(false,
+  "You are using an exact number types\n"
+  "using a Constrained_triangulation_plus_2 class\n"
+  "would avoid cascading intersection computation\n"
+  " and be much more efficient\n");
+
```
lrineau: In #2752, @sloriot said that there should be a macro to allow the removal of that warning, if the macro is defined.

maxGimeno: CGAL_NO_WARNINGS works to discard it

lrineau: I agree, but I thought that Sébastien had something else in mind, a macro that would disable only that warning. @sloriot?

Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -909,12 +909,15 @@ intersect(Face_handle f, int i,
...
+#ifndef CGAL_NO_WARNING_WHEN_USING_CDT_PLUS_WHILE_INTERSECTING_CONSTRAINTS
   CGAL_warning_msg(false,
-  "You are using an exact number types\n"
+  "You are using an exact number type\n"
   "using a Constrained_triangulation_plus_2 class\n"
   "would avoid cascading intersection computation\n"
-  " and be much more efficient\n");
-
+  " and be much more efficient\n"
+  "defineCGAL_NO_WARNING_WHEN_USING_CDT_PLUS_WHILE_INTERSECTING_CONSTRAINTS "
```
lrineau: > defineCGAL_NO_WARNING_WHEN_USING_CDT_PLUS_WHILE_INTERSECTING_CONSTRAINTS

Missing space

maxGimeno:
fixed

sloriot:
@lrineau  I think you did not answer my previous comment. Is it acceptable to you to throw when using `No_intersection_tag`? If yes then no need to add a new one, we simply document the exception type.

lrineau:
There is already `CGAL_triangulation_assertion(false);` in the code. Just replace it with `CGAL_error_msg("something")` and the code will throw the standard CGAL exception even in Release.

lrineau:
Or do you need to invent a specific exception type for that case?

sloriot:
Not throwing a dedicated exception does not seem very clean to me.

lrineau:
Then for whatever new exception type you want to throw, the user must have the documentation of the exception type.

lrineau:
Unless you document that it is an implementation-defined type, but in that case I do not see what the user can do with it, except a `catch(...)`

sloriot:
Which is what I'm saying in my comment: "we simply document the exception type" :)

sloriot:
Any suggestion for the name of the exception? `Intersection_of_constraints_exceptions`? Nested in the constrained triangulation class?

lrineau:
No idea. For me our current `CGA::Error_exception` thrown by `CGAL_error_msg(msg)` is already good enough.

lrineau:
This PR is stalled. As far as I know, you just have to decide on a name for the exception, and write a small documentation.

sloriot:
I'd go for `Intersection_of_constraints_exception` then, `Error_exception` is to generic. @maxGimeno could you do the required changes?

Triangulation_2/doc/Triangulation_2/CGAL/Constrained_triangulation_2.h
```diff
@@ -9,6 +9,13 @@ Intersection tag for constrained triangulations, when input constraints do not i
 */
 struct No_intersection_tag{};
 
+/*!
+\ingroup PkgTriangulation2TriangulationClasses
+
+Intersection tag for constrained triangulations, when input constraints should not intersect.
+Will throw a `Intersection_of_constraints_exception` if constraints do intersect.
+*/
+struct Throw_on_intersection_tag{};
```
sloriot: Actually @lrineau suggested that we can use the existing `No_intersection_tag` so this tag becomes useless.

lrineau:
Travis results are rather red:
https://travis-ci.org/CGAL/cgal/jobs/349759454#L7331
```
[ 40%] Building CXX object CMakeFiles/test_Face_filtered_graph.dir/test_Face_filtered_graph.cpp.o
In file included from /home/travis/build/CGAL/cgal/BGL/test/BGL/graph_concept_Triangulation_2.cpp:3:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/boost/graph/graph_traits_Constrained_triangulation_2.h:32:
/usr/local/lib/cmake/CGAL/../../../include/CGAL/Constrained_triangulation_2.h:455:17: error: 
      class member cannot be redeclared
  Vertex_handle intersect(Face_handle f, int i,
                ^
/usr/local/lib/cmake/CGAL/../../../include/CGAL/Constrained_triangulation_2.h:451:17: note: 
      previous declaration is here
  Vertex_handle intersect(Face_handle f, int i, 
                ^
```
Even the Triangulation_2 test suite does not pass. Test before pushes...

lrineau:
Should this PR go into CGAL-4.12?

sloriot:
There is no rush.

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2785)

maxGimeno:
Is this PR OK now ? Did I miss something ? Can we remove the TODO label and try to integrate it ?

Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -881,10 +884,15 @@ intersect(Face_handle f, int i,
 // split constraint edge (f,i) 
 // and return the Vertex_handle of the new Vertex
 { 
-  std::cerr << "You are using an exact number types" << std::endl;
-  std::cerr << "using a Constrained_triangulation_plus_2 class" << std::endl;
-  std::cerr << "would avoid cascading intersection computation" << std::endl;
-  std::cerr << " and be much more efficient" << std::endl;
+#ifndef CGAL_NO_WARNING_WHEN_USING_CDT_PLUS_WHILE_INTERSECTING_CONSTRAINTS
```
sloriot: The message is shown only when no using the CDT+  something like `CGAL_NO_CDT_2_WARNING` should be enough.

Triangulation_2/include/CGAL/Constrained_triangulation_2.h
```diff
@@ -101,6 +102,15 @@ class Constrained_triangulation_2
     }
   };
 
+
+  class Intersection_of_constraints_exception : public std::exception
+  {
+    const char* what() const throw ()
+    {
+      return "the constraint and subconstraint edge are intersecting.";
```
sloriot: _Intersection of constraints while using No_intersection_tag_ ?

lrineau:
This PR is selected for 4.13. So the integration cannot be started before we create the release branch for CGAL-4.12.

lrineau:
https://github.com/CGAL/cgal/pull/2785#issue-166800292:
> Feature/Small Feature (if any): Coming soon.

Is it still planned?

maxGimeno:
link to the small feature added

Installation/CHANGES.md
```diff
@@ -0,0 +1,4452 @@
...
+
+    The user can restore old (non-symmetric) behaviour by defining the macro:
+
+    `CGAL_NON_SYMETRICAL_OBSERVER_EDGE_REMOVAL_BACKWARD_COMPATIBILITY`
+
+### 2D Triangulations
+
+-   Added a new type of intersection to deal with insertion of a constraints 
+    intersecting in a Constrained_triangulation_2.
+
```
lrineau: Idem. That should be in a section about 4.13.

sloriot:
Sucessfully tested in CGAL-4.13-Ic-3.

PR #2786, opened 2018-02-02
Triangulation_2: Fix template parameter in RT_2's graph traits

MaelRL:
## Summary of Changes

For some reason, the typedef `Triangulation` in `Regular_triangulation_2` is not the base triangulation `Triangulation_2`, but a typedef to `Self`. Consequently, the `halfedge_descriptor` in the graph traits of `Regular_triangulation_2` is wrong and cannot compile.

Although it's not documented, I left `typedef Self Triangulation` in case it's used somewhere (but I couldn't find it anywhere in CGAL) as it could create a nasty bug. Base is typedef'd by `Tr_Base` for consistency with the base typedef in (P3)T3 classes.

Added `RT2` to `graph_concept_Triangulation_2`, but the T2 graph traits are missing proper tests...

## Release Management

### DIFF
--- a/BGL/test/BGL/graph_concept_Triangulation_2.cpp
+++ b/BGL/test/BGL/graph_concept_Triangulation_2.cpp
@@ -2,6 +2,7 @@
 #include <CGAL/boost/graph/graph_traits_Triangulation_2.h>
 #include <CGAL/boost/graph/graph_traits_Constrained_triangulation_2.h>
 #include <CGAL/boost/graph/graph_traits_Delaunay_triangulation_2.h>
+#include <CGAL/boost/graph/graph_traits_Regular_triangulation_2.h>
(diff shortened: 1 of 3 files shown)


sloriot:
Just to mention it, I've made a generic header for the `graph_traits` for the 2 classes from OpenMesh to avoid this kind of issues (copy-pasting). Maybe it would make sense to have one for triangulation classes two.

MaelRL:
`Tr_Base` creates some conflicts with the T2 hierarchy, so it's not a nice name.

lrineau:
Successfully tested in [CGAL-4.12-Ic-173][*].

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-173.shtml

lrineau:

PR #2794, opened 2018-02-06
Documentation:  Fix postprocessing

### LINKED ISSUE #2750: Check why icons color are missing
Compare [that page](https://doc.cgal.org/latest/Algebraic_foundations/annotated.html) to [that page](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.12-Ic-151/master/Algebraic_foundations/annotated.html). Post processing should be replacing the small icons but it seems it is no longer the case for recent version of doxygen.

maxGimeno:
## Summary of Changes
This PR 
- hides the broken index of the Manual page
- adds some postprocessing for 1.8.13 and 1.8.14 so the colors of the Namespace, Class and Concept nodes are modified the same way it is in 1.8.4
## Release Management

* Issue(s) solved (if any): fix #2750

### DIFF
--- a/Documentation/doc/resources/1.8.14/DoxygenLayout.xml
+++ b/Documentation/doc/resources/1.8.14/DoxygenLayout.xml
@@ -4,7 +4,7 @@
     <tab type="mainpage" visible="yes" title=""/>
     <tab type="modules" visible="yes" title="" intro=""/>
     <tab type="pages" visible="yes" title="" intro=""/>
-    <tab type="classlist" visible="yes" title="Class and Concept List" intro="Here is the list of all concepts and classes of the CGAL Library. Classes are inside the namespace CGAL. Concepts are in the global namespace."/>
+    <tab type="classlist" visible="no" title="Class and Concept List" intro="Here is the list of all concepts and classes of the CGAL Library. Classes are inside the namespace CGAL. Concepts are in the global namespace."/>
     <tab type="examples" visible="no" title="" intro=""/>  
     <!-- <tab type="user" url="@ref how_to_cite_cgal" title="Acknowledging CGAL"/> -->
   </navindex>
--- a/Documentation/doc/scripts/html_output_post_processing.py
+++ b/Documentation/doc/scripts/html_output_post_processing.py
@@ -88,7 +88,7 @@ def clean_doc():
     duplicate_files.extend(package_glob('./*/geom.bib'))
     duplicate_files.extend(package_glob('./*/ftv2cl.png'))
     duplicate_files.extend(package_glob('./*/ftv2ns.png'))
-    
+
     for fn in duplicate_files:
         os.remove(fn)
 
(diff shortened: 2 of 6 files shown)


Documentation/doc/scripts/html_output_post_processing.py
```diff
@@ -150,6 +150,16 @@ def rearrange_img(i, dir_name):
...
 
+def rearrange_icon(i, dir_name):
+    icon = pq(this)
+    if icon.attr("class") == "icon-class":
+        parser=pq(this).parent().parent()
+        for link_class in ['a.el', 'a.elRef']:
+            links=parser(link_class)
+            if links.size()>0 and is_concept_file(path.join(dir_name, pq(links[0]).attr("href"))):
+                icon.attr("class","icon-concept")
+    srcpath=icon.attr("class")
```
sloriot: this line seems suspicious. Is it really needed?

lrineau:
@sloriot said in https://github.com/CGAL/cgal/pull/2794#pullrequestreview-94338878:
> I did not see the output but the modifications seem to be OK

@maxGimeno Could you publish somewhere the result of the compiled documentation, from this branch?

sloriot:
We can also wait for tonight testsuite (since integration will be empty from new changes tonight).

maxGimeno:
The Manual index's page that I left for doxygen 1.8.4 does not have the right color for the Concepts. Do I remove it just like the others or do we just ignore it ?

lrineau:
In my opinion: 
  - you can remove it,
  - but we do not really care, actually.
So do as you want.

maxGimeno:
I guess it would be cleaner to remove it. I'll do that.

2018-02-07 9:53 GMT+01:00 Laurent Rineau <notifications@github.com>:

Documentation/doc/resources/1.8.14/DoxygenLayout.xml
```diff
@@ -4,7 +4,7 @@
     <tab type="mainpage" visible="yes" title=""/>
     <tab type="modules" visible="yes" title="" intro=""/>
     <tab type="pages" visible="yes" title="" intro=""/>
-    <tab type="classlist" visible="yes" title="Class and Concept List" intro="Here is the list of all concepts and classes of the CGAL Library. Classes are inside the namespace CGAL. Concepts are in the global namespace."/>
+    <tab type="classlist" visible="no" title="Class and Concept List" intro="Here is the list of all concepts and classes of the CGAL Library. Classes are inside the namespace CGAL. Concepts are in the global namespace."/>
```
sloriot: Dummy question: why do we even keep that line? Wouldn't we have the same result if we remove it?

maxGimeno: I made the minimal change possible, thinking that if one day we want it back it would be easier like this. But you are technically right, I guess. Removing it or hiding it shouldn't be really different in the user's point of view.

sloriot: Thanks.

lrineau:
Sébastien, do we consider the test as successful? I think we could merge the PR as it is.

sloriot:
Definitely!

PR #2800, opened 2018-02-06
prevent GCC flags propagating to NVCC, fixes #2775

### LINKED ISSUE #2775: Compilation flag "frounding-math" incorrectly propagated to CUDA NVCC compiler
## Environment

* Operating system: Ubuntu 16.04
* Compiler: GCC 5.4
* CGAL version: Latest
* Boost version: 1.58

## Issue Details

For some weird reason the compilation flag `frounding-math` is propagated by CGAL to my CUDA application, and because the NVCC compiler cannot recognize it, it causes a compilation error.

## Source Code

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.9)

project(Example LANGUAGES CXX CUDA)

set(PROJECT_SRCS ${PROJECT_SOURCE_DIR}/main.cu)

find_package( CGAL REQUIRED )

add_executable(${PROJECT_NAME} ${PROJECT_SRCS})

target_link_libraries(${PROJECT_NAME} CGAL::CGAL)
```

main.cu
```
int main(){
return 0;
}
```

After the usual `mkdir build`, `cd build`, `cmake ..`, `make`, here is the error

```
[ 33%] Building CUDA object CMakeFiles/Example.dir/main.cu.o
nvcc fatal   : Unknown option 'frounding-math'
```

I tried to debug myself, according to `/usr/lib/x86_64-linux-gnu/CGAL/CGALConfig.cmake`, there is a line `set(CGAL_CXX_FLAGS_INIT " -frounding-math" )`, so I went back to the source code and trace the variable `CGAL_CXX_FLAGS_INIT` and saw it in `Installation/cmake/modules/CGALConfig_install.cmake.in` and `Installation/cmake/modules/CGAL_SetupFlags.cmake`, and it seems to be related to `CMAKE_CXX_FLAGS` only, nothing suspicious, and I have no idea what to look for afterwards to track down the bug. Can someone look into this?

I also tried modifying the `CMAKE_CXX_FLAGS` manually in the minimal example without linking to CGAL and it doesn't affect NVCC, 

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.9)

project(Example LANGUAGES CXX CUDA)

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -DNDEBUG")

set(PROJECT_SRCS ${PROJECT_SOURCE_DIR}/main.cu)

add_executable(${PROJECT_NAME} ${PROJECT_SRCS})
```

Compiles without a problem. Seems like somewhere in CGAL it forces propagation of the `frounding-math` flag to non C++ compilers that causes the issue.

jasjuang:
## Summary of Changes

This PR fixes #2775


Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake
```diff
@@ -136,7 +136,7 @@ function(CGAL_setup_CGAL_dependencies target)
     endif()
     if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 3)
       message( STATUS "Using gcc version 4 or later. Adding -frounding-math" )
-      target_compile_options(${target} ${keyword} "-frounding-math")
+      target_compile_options(${target} ${keyword} "$<$<COMPILE_LANGUAGE:CXX>:-frounding-math>")
```
lrineau: Unfortunately, [`COMPILE_LANGUAGE` was introduced in CMake version 3.3][changelog], and CGAL is supposed to support CMake version 3.1 and later.

  [changelog]: https://cmake.org/cmake/help/v3.3/release/3.3.html#generator-expressions

I think there should be two different lines, with a conditional on `CMAKE_VERSION`:
```cmake
if(CMAKE_VERSION VERSION_LESS 3.3)
  target_compile_options(${target} ${keyword} "-frounding-math")
else
  target_compile_options(${target} ${keyword} "$<$<COMPILE_LANGUAGE:CXX>:-frounding-math>")
endif()
```

jasjuang:
@lrineau updated as you requested

lrineau:
Thanks. You patch will make into our daily testsuite... Well, too late for tonight: it will be tested tomorrow night, and eventually merged.

Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake
```diff
@@ -136,7 +136,11 @@ function(CGAL_setup_CGAL_dependencies target)
     endif()
     if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 3)
       message( STATUS "Using gcc version 4 or later. Adding -frounding-math" )
-      target_compile_options(${target} ${keyword} "-frounding-math")
+      if(CMAKE_VERSION VERSION_LESS 3.3)
+        target_compile_options(${target} ${keyword} "-frounding-math")
+      else
```
lrineau: That patch cannot be right. The `else` should be `else()`.

jasjuang:
https://cmake.org/cmake/help/v3.10/command/if.html

Quote:The elseif and else sections are optional.

lrineau:
Please re-read my message carefully.

You can also have a look at a Travis build failure, like that one:
  https://travis-ci.org/CGAL/cgal/jobs/338668306#L1775

Our Travis setup uses CMake-3.9.2.

lrineau:

lrineau:
Thanks @jasjuang for the bug report and the patch. Our testing process is tedious but your patch eventually got merged today in our `master` branch. It will be part of the future CGAL-4.12-beta1 and CGAL-4.12 public releases.

PR #2802, opened 2018-02-07
Polyhedron: Add *STL to nameFilters()

afabri:
## Summary of Changes

The STL IO plugin did not read files with *.STL

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1015,7 +1015,7 @@ void MainWindow::open(QString filename)
     Q_FOREACH(CGAL::Three::Polyhedron_demo_io_plugin_interface* io_plugin, io_plugins) {
       if ( !io_plugin->canLoad() ) continue;
       all_items << io_plugin->name();
-      if ( file_matches_filter(io_plugin->loadNameFilters(), filename) )
+      if ( file_matches_filter(io_plugin->loadNameFilters(), filename.toLower()) )
         selected_items << io_plugin->name();
     }
   }
@@ -1642,7 +1642,7 @@ void MainWindow::save(QString filename, CGAL::Three::Scene_item* item) {
   bool saved = false;
   Q_FOREACH(CGAL::Three::Polyhedron_demo_io_plugin_interface* plugin, io_plugins) {
     if(  plugin->canSave(item) &&
-        file_matches_filter(plugin->saveNameFilters(),filename) )
+        file_matches_filter(plugin->saveNameFilters(),filename.toLower()) )
     {
       if(plugin->save(item, fileinfo))
       {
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/Plugins/IO/STL_io_plugin.cpp
```diff
@@ -55,7 +55,7 @@ class Polyhedron_demo_stl_plugin :
 };
 
 QString Polyhedron_demo_stl_plugin::nameFilters() const {
-  return "STL files (*.stl)";
+  return "STL files (*.stl *STL)";
```
lrineau: Missing `.` in `*.STL`

sloriot:
Isn't it possible to do the modification in the function using the output of `nameFilters()`? My idea was that the function would lower the case of the file to be read and all filters would use lower cases. @maxGimeno ?

maxGimeno:
Yes you are right ! I didn't think of that when Andreas asked, but we could pass nameFilters().toLower() to the load function in the mainwindow. That would require a pass over every IO plugin to check no Capital is expected though. I can do that in that branch if every one is OK with that solution.

lrineau:
If you do that, that means that the file dialog will not show `.STL` files (with the capitalized extension).

sloriot:
@lrineau you mean that the files won't be shown? If yes it is an issue only for io plugins that are in competition right?

lrineau:
If you only pass `*.stl` [`QFileDialog::setNameFilters`], then the dialog will not show files that have the uppercase extension `*.STL`. It seems that was what Andreas observed on his Windows machine.

Honestly, I am surprise that `QFileDialog` does not use `toLower()` on files extensions before matching them to the list of name filters, but if Andreas is right, that seems the case, and we do not have the choice: we cannot modify that behavior of `QFileDialog`.


[`QFileDialog::setNameFilters`]: http://doc.qt.io/qt-5/qfiledialog.html#setNameFilters

maxGimeno:
Actually the STL files are displayed, so it is a bug from our side.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1015,7 +1015,7 @@ void MainWindow::open(QString filename)
     Q_FOREACH(CGAL::Three::Polyhedron_demo_io_plugin_interface* io_plugin, io_plugins) {
       if ( !io_plugin->canLoad() ) continue;
       all_items << io_plugin->name();
-      if ( file_matches_filter(io_plugin->loadNameFilters(), filename) )
+      if ( file_matches_filter(io_plugin->loadNameFilters().toLower(), filename) )
```
lrineau: Wrong patch!

That is not the filters you want to `toLower()`, but the file names! Here:

https://github.com/afabri/cgal/blob/0caaf4136e5f8673bad64027a38410d6c25c3f6c/Polyhedron/demo/Polyhedron/MainWindow.cpp#L937-L960

L953 the string `filename_striped` should be `toLower()` before being matched.

maxGimeno:
fixed

afabri:
I confirm that it compiles with VC++ and fixes the problem.

lrineau:

Issue #2787, opened 2018-02-02
Optimal_transportation_reconstruction_2: runtime error of `test_vertex_edge` on Linux 32 bits


lrineau:
## Issue Details

https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-I-102/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
```
------------------------------------------------------------------
- ProgramOutput.test_vertex_edge.Fedora-32-Release
------------------------------------------------------------------
CGAL::Random()::get_seed() = 1511774527
CGAL::Random()::get_seed() = 1511774527
test_edge_collapse
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
./cgal_test_with_cmake: line 70: 10783 Aborted                 (core dumped) ./test_vertex_edge
```

I grepped in the test results for 4.10*, 4.11, and 4.12*, and it seems it
has been recently introduced in 4.10.2 and 4.11.1 as well:
```
../CGAL-4.10.2-I-209/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.10.2-I-210/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11.1-I-110/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-100/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-101/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-102/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-103/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-105/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-106/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
../CGAL-4.11-I-107/Optimal_transportation_reconstruction_2/TestReport_lrineau_Fedora-32-Release.gz
[...]

PR #2803, opened 2018-02-07
OT Reconstruction: Fix test_vertex_edge bug by not copying the TDS

sgiraudot:
## Summary of Changes

Fix for https://github.com/CGAL/cgal/issues/2787, which was much trickier than I originally thought.

- the `std::bad_alloc` was happening in an orientation test in EPICK
- I found that one of the coordinates of the points in this test was `nan`
- I figured that it actually came from the fact that the point of the `infinite_vertex()` of a DT2 was used at some point. So there was also a bug in other platforms, but for some reason it was only producing a `nan` on this 32b architecture (on my personal computer, I got random values close to 0).
- In the tests, some internal functions of the algorithms are called but **using elements of a copy of the DT2** as parameters. So of course, tests like `if (v != m_dt.infinite_vertex())` can't work if `v` is not part of `m_dt`.
- After fixing this by not copying the triangulation, I found that an _assertion fail_ was triggered `CGAL_assertion(v != m_dt.infinite_vertex())`
- The test was actually doing something wrong: some tests should not be tested (namely the ones which are `pinned` and which are then adjacent to the infinite vertex). It is clear if we look at the method `push_to_mindex()` which does these tests before calling the function that is tested in our failing test file.

So I replaced the copy by a const ref and added the tests to avoid testing "bad" edges, and now it runs fine on the 32b platform and does not do anything fishy on the other platforms as well (I'm kind of surprised that it did not crash on more platforms before, but I guess having one random point used was not that important if the value was not a `nan`).

### Side note

I know there was, at some point, a project of initializing the infinite vertex to `nan` coordinates: I don't know if there was a good reason not to do it, or it's just in limbo because nobody had time to do it, but it would be a *really* good idea. This kind of bug would have been much easier to detect.

(What's worrying is we'll probably find more of these bugs if we do it.)

## Release Management

* Affected package(s): Optimal Transportation Reconstruction
* Issue(s) solved: fix https://github.com/CGAL/cgal/issues/2787

### DIFF
--- a/Optimal_transportation_reconstruction_2/include/CGAL/Optimal_transportation_reconstruction_2.h
+++ b/Optimal_transportation_reconstruction_2/include/CGAL/Optimal_transportation_reconstruction_2.h
@@ -1779,6 +1779,8 @@ public:
 
 
   /// \cond SKIP_IN_MANUAL
+  const Triangulation& tds() const { return m_dt; }
+  
   void extract_tds_output(Triangulation& rt2) const {
     rt2 = m_dt;
     //mark vertices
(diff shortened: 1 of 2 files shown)


lrineau:

PR #2807, opened 2018-02-08
Fix Fedora_32 errors in the testsuite

maxGimeno:
## Summary of Changes
Remove Qt5::Help from CMakeLists because it causes errors and it is not used.

## Release Management

* Affected package(s):Periodic_3_Triangulation_3 demos


lrineau:
I remove the `testsuite` label, because that is not a bug of our testsuite infrastructure, but an actual bug in the `CMakeLists.txt`.

lrineau:

PR #2811, opened 2018-02-08
Polyhedron_demo: Fix some crashes when saving an item.

maxGimeno:
## Summary of Changes
If the filters are empty, filters.first().split(";;") will trigger an assertion and probably segfault, so this PR adds a check before that. If it is empty, it will trigger the warning and return.


afabri:
I just tested it and it solves the problem.

lrineau:

PR #2812, opened 2018-02-08
Intersections_3: Replace last case with a default to avoid warning

### LINKED ISSUE #2809: warning: this statement may fall through [-Wimplicit-fallthrough=]
## Issue Details

This [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.12-Ic-168/AABB_tree/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz) warns about _fall throughs_ that cannot happen:  It is a `switch` statement where the developer knows that the cases are exhaustive and each case has a `return`statement.

Shall we nevertheless add it (At the risk that another compiler complains about unreachable code)? Or shall we quiet the compiler with a `#pragma`?

afabri:
## Release Management

* Issue(s) solved (if any): fix #2809

### DIFF
--- a/Intersections_3/include/CGAL/internal/Intersections_3/Triangle_3_Segment_3_intersection.h
+++ b/Intersections_3/include/CGAL/internal/Intersections_3/Triangle_3_Segment_3_intersection.h
@@ -257,7 +257,7 @@ intersection_coplanar(const typename K::Triangle_3 &t,
             case NEGATIVE:
               // c is isolated on the negative side
               return t3s3_intersection_coplanar_aux(a,b,c,p,q,true,k);
-            dafault:
+            default:
               if ( collinear_ordered(p,c,q) ) // c is inside [p,q]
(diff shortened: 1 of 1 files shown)


Intersections_3/include/CGAL/internal/Intersections_3/Triangle_3_Segment_3_intersection.h
```diff
@@ -257,21 +257,19 @@ intersection_coplanar(const typename K::Triangle_3 &t,
             case NEGATIVE:
               // c is isolated on the negative side
               return t3s3_intersection_coplanar_aux(a,b,c,p,q,true,k);
-            case COLLINEAR:
+            dafault:
```
mglisse: I am curious if you get a warning for that typo...

lrineau:
I do not like the patch, because it is less readable. Could it be the following?
```c++
default:
case COLLINEAR:
```
with `default:` falling through the `COLLINEAR` case?

lrineau:

PR #2822, opened 2018-02-12
PMP: bugfix in expand_face_selection_for_removal()

### LINKED ISSUE #2788: BGL: expand_face_selection_for_removal crashed with built-in Surface_mesh property
## Issue Details

The method [CGAL::expand_face_selection_for_removal()](https://doc.cgal.org/latest/BGL/group__PkgBGLSelectionFct.html#ga569fe26f889e5e4eed27746ac921651b) requires a `is_selected` property map that associates a face descriptor to a boolean. If I'm not mistaken, a built-in `Surface_mesh` property map (`Surface_mesh::Property_map<Surface_mesh::Face_index,bool>`) should do the job, but in practice it crashes. In debug, I have an assertion that pretty much says that it's trying to use a `Face_index` larger than the number of faces of the mesh.

In the test that uses it (and in the Polyhedron demo), an external `unordered_map` is used:

```c++
// define my selection of faces to remove
  boost::unordered_map<face_descriptor, bool> is_selected_map;
  expand_face_selection_for_removal(faces_to_remove,
                                    sm,
                                    boost::make_assoc_property_map(is_selected_map));
```

And it indeed does work with this kind of map. If we use a `std::map` on the example hereafter instead of the built-in map and print out the values after running the function, we notice that the face index `f4294967295` (which corresponds to the default `Face_index` value if I'm not mistaken) is present in the map. Of course it is not a problem for a `std::map` as it will just be ignored, but you can't use it the built-in map, which leads to the crash/assertion failed I was talking about.

So my guess is that at some point, the algorithm does a `put()` using an uninitialized `face_descriptor`, which should not happen.

## Source Code

Modified `test_Manifold_face_removal.cpp` from the BGL package:

```c++
#include <CGAL/boost/graph/graph_traits_Surface_mesh.h>
#include <CGAL/Surface_mesh/IO.h>

#include <boost/unordered_map.hpp>
#include <boost/property_map/property_map.hpp>

#include <iostream>
#include <fstream>
#include <set>
#include <CGAL/boost/graph/selection.h>
#include <CGAL/boost/graph/helpers.h>
typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
typedef CGAL::Surface_mesh<Kernel::Point_3> SM;
typedef boost::graph_traits<SM>::face_descriptor face_descriptor;

int main()
{
  SM sm;
  std::ifstream input("data/head.off");
  input >> sm;

// define my selection of faces to remove

  SM::Property_map<face_descriptor, bool> is_selected_map
    = sm.add_property_map<face_descriptor, bool> ("is_selected", false).first;

  int index = 0;
  std::set<SM::Face_index> faces_to_remove;

  // select faces surrounding the border
  BOOST_FOREACH (SM::Vertex_index vi, vertices(sm))
  {
    std::vector<SM::Face_index> incident;
    bool border = false;
      
    SM::Halfedge_index hvi = halfedge (vi, sm);

    BOOST_FOREACH (SM::Halfedge_index hi, halfedges_around_target (hvi, sm))
    {
      if (is_border (hi, sm))
        border = true;
      else
        incident.push_back (face(hi, sm));
    }
    if (border)
      for (std::size_t i = 0; i < incident.size(); ++ i)
      {
        faces_to_remove.insert (incident[i]);
        is_selected_map[incident[i]] = true;
      }
  }

  expand_face_selection_for_removal(faces_to_remove,
                                    sm,
                                    is_selected_map);

  return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Debian 64b
* Compiler: clang
* Release or debug mode: both
* CGAL version: master

afabri:
## Summary of Changes

Deal correctly with border halfedges.

## Release Management

* Issue(s) solved (if any): fix #2788

### DIFF
--- a/BGL/include/CGAL/boost/graph/selection.h
+++ b/BGL/include/CGAL/boost/graph/selection.h
@@ -572,14 +572,14 @@ void expand_face_selection_for_removal(const FaceRange& faces_to_be_deleted,
     // set hd to the last selected face of a connected component
     // of selected faces around a vertex
     halfedge_descriptor hd = halfedge(vd, tm);
-    while( !get(is_selected, face(hd, tm)) )
+    while(is_border(hd,tm) || ( !get(is_selected, face(hd, tm))) )
     {
       hd = opposite( next(hd, tm), tm);
       CGAL_assertion( hd != halfedge(vd, tm) );
     }
     halfedge_descriptor start = hd;
     halfedge_descriptor next_around_vertex = opposite( next(hd, tm), tm);
-    while( get(is_selected, face(next_around_vertex, tm) ) )
+    while(is_border(next_around_vertex,tm) || get(is_selected, face(next_around_vertex, tm) ) )
     {
       hd = next_around_vertex;
       next_around_vertex = opposite( next(hd, tm), tm);


lrineau:

PR #2825, opened 2018-02-13
STL_extension:  Fix #2587 concerning &* 

### LINKED ISSUE #2587: Undefined behaviour reference binding to null pointer in CGAL::internal::vector_iterator
## Issue Details

The CGAL::internal::vector_iterator operator*() seems to provoke undefined behaviour when the iterator refers to an empty vector. This undefined behaviour occurs when begin(), end() or size() is called.

In CGAL::internal::vector_iterator:
```
    reference operator*()                const { return *ptr; }
```
This line creates a reference binding to a null pointer when ptr is nullptr.

As detected by the clang undefined behaviour sanitizer.
It can probably be detected with the gcc undefined behaviour sanitizer as well (see https://gcc.gnu.org/onlinedocs/gcc-6.4.0/gcc/Instrumentation-Options.html -fsanitize=null).

### Runtime error

include/CGAL/vector.h:75:57: runtime error: reference binding to null pointer

## Source Code

Minimal reproducible sample:
https://gist.github.com/bobergj/db615e08f00c7820532f151b880775b3

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): macOS Sierra 10.12.6, 64 bits
* Compiler: Apple LLVM version 9.0.0 (clang-900.0.38)
* Specific flags used (if any): -fsanitize=undefined
* CGAL version: 4.11 built from source 
* Boost version: 1.59.0_3 (macports)

afabri:
... applied to a `CGAL::internal::vector_iterator`

## Release Management

* Affected package(s): STL_extension, HalfedgeDS, Partition_2
* Issue(s) solved (if any): fix #2587

### DIFF
--- a/STL_Extension/include/CGAL/vector.h
+++ b/STL_Extension/include/CGAL/vector.h
@@ -66,7 +66,7 @@ public:
 
     // Allows construction of const_iterator from iterator
     template < class A, class B, class C>
-    vector_iterator( const vector_iterator<A,B,C>& i) : ptr( &*i) {}
+    vector_iterator( const vector_iterator<A,B,C>& i) : ptr(i.operator->()) {}
 
     // OPERATIONS Forward Category
     // ---------------------------


lrineau:
Successfully tested in [CGAL-4.12-Ic-176], and then merged.

[CGAL-4.12-Ic-176]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-176.shtml

Issue #2107, opened 2017-05-11
Document the possible Function types in isosurface example


afabri:
## Issue Details

Alec Jacobson posted [CGAL function pointer callback in isosurface example](http://www.alecjacobson.com/weblog/?p=4034)  so we (@janetournois) might take that up and improve CGAL. 

**Edit**: Note that that discussion started as the issue #2753. --Laurent Rineau, 2018/02/14

lrineau:
Alec Jacobson had to understand how to use `std::bind`. `std::function` is a solution. Usage of `auto` and `decltype` could have been another one. If there anything to change in CGAL, that would only be the documentation. The classes are already templated by the callable type (via the parameter `Function`):
  - [`CGAL::Implicit_surface_3<Traits, Function>`](http://doc.cgal.org/4.10/Surface_mesher/classCGAL_1_1Implicit__surface__3.html)
  - [`CGAL::Implicit_mesh_domain_3<Function, BGT>`](http://doc.cgal.org/4.10/Mesh_3/classCGAL_1_1Implicit__mesh__domain__3.html)

lrineau:
In Mesh_3, [`CGAL::Implicit_mesh_domain_3<Function, BGT>`](http://doc.cgal.org/4.10/Mesh_3/classCGAL_1_1Implicit__mesh__domain__3.html) is dealt with in PR #2739, commit https://github.com/CGAL/cgal/pull/2739/commits/81e1d318dd48c1ffee5ca4e06fe1955fe655bd53 and later ones.

In Surface_mesher... I will try to adopt the same ideas...

lrineau:
#2829 is a fix for Surface_mesher.

lrineau:
#2829 was merged in CGAL-4.12, and fixed this issue.

PR #2829, opened 2018-02-13
Use CGAL::cpp11::function in CGAL::Implicit_surface_3

lrineau:
That will fix the issue #2107.

### DIFF
--- a/Surface_mesher/include/CGAL/Implicit_surface_3.h
+++ b/Surface_mesher/include/CGAL/Implicit_surface_3.h
@@ -28,12 +28,15 @@
 #include <CGAL/Surface_mesher/Implicit_surface_oracle_3.h>
 
 #include <functional>
+#include <CGAL/function.h>
 
 namespace CGAL {
 
   template<
     typename GT,
-    typename Function_
+    typename Function_ = cpp11::function<typename GT::FT(typename GT::Point_3)>
+    // The type of the argument `Function` will be ignored anyway.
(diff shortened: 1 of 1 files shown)


lrineau:
@alecjacobson:
This PR changes `CGAL::Implicit_surface_3<Geom_traits, Functor>` as such:
  - the second parameter `Functor` is now optional, and is anyway always ignored,
  - internally, the object of type `CGAL::Implicit_surface_3<Geom_traits, Functor>` always store a `std::function<FT(Point_3)>` instead of a functor.

As far as I know, that should be backward compatible. What do you think?

Internally, we do not use `std::function`, but a wrapper class template `CGAL::function`, that falls back to `boost::function` is the compiler does not implement `std::function` (C++11-only).

I do not know how to document that change...

afabri:
Do you agree that we do not want to document this?   

Classes that derive from it such as [`Emptyset_iterator`](https://doc.cgal.org/latest/STL_Extension/classCGAL_1_1Emptyset__iterator.html#ae0b023da392ee337e9b20106af1c4918)  do not document that they have `std::iterator as base class`.

alecjacobson:
This looks cool to me.

lrineau:
Successfully tested in [CGAL-4.12-Ic-176], and then merged.

[CGAL-4.12-Ic-176]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-176.shtml

PR #2834, opened 2018-02-14
Polyhedron demo: Save as enhancement

### LINKED ISSUE #2334: Polyhedron demo: Save as
## Issue Details

This is a feature request.

When I want to save a polyhedron it offers me the gocad format.  That will rarely be the first choice. 
Also the directory where I saved should be remebered for the next save as.

maxGimeno:
## Summary of Changes
- Adds a function to IO plugins interface.
  It decides if the plugin should be the default plugin for saving a given item.
- Set off_plugin as default saver for polygon soups and polyhedron_items, surface_mesh_io_plugin for surface_mesh_items and PLY_plugin for point_with_normals_items.
  This means their first nameFilter(OFF for the 2 first, PLY for the latter) will be the first choice when saving an item.

- When saving a file, if it was loaded, will use its loading-path as default for saving. If it was not loaded, will use the last path used for saving in the current session as default path for saving. If none of the previous cases are applicable, use the same default path as before (which must be the build dir of the demo I think).

## Release Management

* Issue(s) solved (if any): fix #2334

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1606,12 +1606,19 @@ void MainWindow::on_actionSaveAs_triggered()
       continue;
     }
     QString caption = tr("Save %1 to File...").arg(item->name());
+    QString dir = item->property("source filename").toString();
+    if(dir.isEmpty())
+      dir = QString("%1/%2").arg(last_saved_dir).arg(item->name());
+    if(dir.isEmpty())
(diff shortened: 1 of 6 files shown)


maxGimeno:
to do : Filter filename : if the right extension is already present in the filename, don't add it a second time.

lrineau:
Maxime, why the `TODO` label?

maxGimeno:
because of the to do : 

> to do : Filter filename : if the right extension is already present in the filename, don't add it a second time.

PR #2837, opened 2018-02-14
Define CGAL_NORETURN also for VC++ 

### LINKED ISSUE #2826: Support CGAL_NORETURN on MSVC
## Issue Details

Followup to https://github.com/CGAL/cgal/pull/2712#pullrequestreview-96144636:

The proper patch would be to qualify `CGAL::assertion_fail` as [`__declspec(noreturn)`][doc]. All versions of MSVC support that attribute. The macro `CGAL_NORETURN` defined here:
https://github.com/CGAL/cgal/blob/a7a4966de82c48df461ed556e27582f527d7b7e2/Installation/include/CGAL/config.h#L527-L532
should be updated, to that `__declspec(noreturn)` is used with MSVC.

[doc]: https://msdn.microsoft.com/en-us/library/k6ktzx3s(v=vs.110).aspx

The documentation is clear:
> This `__declspec` attribute tells the compiler that a function does not return. As a consequence, the compiler knows that the code following a call to a `__declspec(noreturn)` function is unreachable. 

If `CGAL::assertion_fail` is qualified with `noreturn`, then anything after `CGAL_error()` or`CGAL_error_msg("oups")` is unreachable.

## Environment

* Compiler: MSVC
* CGAL version: master (before CGAL-4.12-beta1)

afabri:
...and move the macro before the functions.

## Release Management

* Affected package(s):  STL_extensions (assertions.h)
* Issue(s) solved (if any): fix #2826

### DIFF
--- a/STL_Extension/include/CGAL/assertions.h
+++ b/STL_Extension/include/CGAL/assertions.h
@@ -63,9 +63,9 @@ namespace CGAL {
 // =====================
 // failure functions
 // -----------------
-CGAL_EXPORT void assertion_fail      ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
-CGAL_EXPORT void precondition_fail   ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
-CGAL_EXPORT void postcondition_fail  ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
+CGAL_EXPORT CGAL_NORETURN void assertion_fail      ( const char*, const char*, int, const char* = "") ;
+CGAL_EXPORT CGAL_NORETURN void precondition_fail   ( const char*, const char*, int, const char* = "") ;
+CGAL_EXPORT CGAL_NORETURN void postcondition_fail  ( const char*, const char*, int, const char* = "") ;
 
 // warning function
 // ----------------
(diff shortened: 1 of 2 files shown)


STL_Extension/include/CGAL/assertions.h
```diff
@@ -63,9 +63,9 @@ namespace CGAL {
 // =====================
 // failure functions
 // -----------------
-CGAL_EXPORT void assertion_fail      ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
-CGAL_EXPORT void precondition_fail   ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
-CGAL_EXPORT void postcondition_fail  ( const char*, const char*, int, const char* = "") CGAL_NORETURN;
+CGAL_EXPORT CGAL_NORETURN void assertion_fail      ( const char*, const char*, int, const char* = "") ;
+CGAL_EXPORT CGAL_NORETURN void precondition_fail   ( const char*, const char*, int, const char* = "") ;
+CGAL_EXPORT CGAL_NORETURN void postcondition_fail  ( const char*, const char*, int, const char* = "") ;
```
lrineau: Have you tested? Seeing the patch, I have a doubt: en Windows, `CGAL_EXPORT` is also a `__declspec` (if you compile CGAL as a DLL), and I think there could be only one of them, combining the two attributes.

afabri: I hadn't checked together with CGAL_EXPORT, but did it now, and it works.

lrineau:

PR #2839, opened 2018-02-14
STL_extension: Introduce CGAL::iterator 

### LINKED ISSUE #2832: std::iterator depreciation (c++17)
## Issue Details

`std::iterator<> ` is deprecated with C++17 with no replacement (replacement is "write the typedefs in the class") and is used in a few packages, mainly in `CGAL/iterator.h`.

`CGAL::unary_function` was introduced, so should `CGAL::iterator` be introduced too ?

afabri:
... as replacement for the deprecated `std::iterator`.  It is only an implementation detail, that is we do not document it. 

## Release Management

* Affected package(s): `STL_extension`, ` Algebraic_kernel_d`, `Arrangement_on_surface_2`, `GraphicsView` 
* Issue(s) solved (if any): fix #2832

### DIFF
--- a/STL_Extension/include/CGAL/iterator.h
+++ b/STL_Extension/include/CGAL/iterator.h
@@ -93,20 +93,20 @@ Iterator_range<Prevent_deref<I> > make_prevent_deref_range(const I& begin, const
   return Iterator_range<Prevent_deref<I> >(make_prevent_deref(begin), make_prevent_deref(end));
 }
 
-  template<typename _Category, typename _Tp, typename _Distance = std::ptrdiff_t,
-         typename _Pointer = _Tp*, typename _Reference = _Tp&>
+template<typename Category, typename Tp, typename Distance = std::ptrdiff_t,
+         typename Pointer = Tp*, typename Reference = Tp&>
 struct iterator
 {
   /// One of the iterator_tags tag types.
-  typedef _Category  iterator_category;
+  typedef Category  iterator_category;
   /// The type "pointed to" by the iterator.
-  typedef _Tp        value_type;
+  typedef Tp        value_type;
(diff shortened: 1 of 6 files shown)


STL_Extension/include/CGAL/iterator.h
```diff
@@ -286,7 +303,7 @@ class Counting_output_iterator
 };
 
 template < class I,
-           class Val = typename std::iterator_traits<I>::value_type >
+           class Val = typename CGAL::iterator_traits<I>::value_type >
```
MaelRL: Typo

lrineau: Reminder: https://github.com/CGAL/cgal/wiki/Testing#using-ctest

;-)

lrineau:
Another note, about the first message at https://github.com/CGAL/cgal/pull/2839#issue-169191866:
> ## Release Management
> 
> * Affected package(s): STL_extension
> * Issue(s) solved (if any): fix #2832

`STL_extension` is clearly not the only affected package. If I cannot trust that "Release Management" section, then I (as Release Manager) have to verify more.

STL_Extension/include/CGAL/iterator.h
afabri: Well, ` std::iterator` still exists. Also I tested.  We maybe have to add a `#define` with the guard of the header of  ` std::iterator`. Or do we have a testsuite where the deprecated no longer exists.

lrineau: - `CGAL::iterator_traits` does not exist, does it?
- `std::iterator_traits` is not deprecated.

afabri: you are right.

lrineau:

PR #2841, opened 2018-02-15
Correctly generate cgal_test_with_cmake

### LINKED ISSUE #2827: remaining bug in release creation cmake script
Followup to https://github.com/CGAL/cgal/pull/2707#issuecomment-365240916.

With `TESTSUITE=ON`, the produced internal release is incorrect. The `cgal_test_with_cmake` that is copied in all `test/*/` sub-directory is not right, and triggers a fork bomb (by calling itself recursively), if called.

I have created two internal releases: one with the old script, and one with the new CMake scripts. I have used the followint diff command:
```shellsession
$ diff -x cgal_test_with_cmake -x filelist.txt -I '$URL.*$' -I '$Id.*$' -ru /tmp/CGAL-4.12-dev*
```
to display the differences between releases.
  - The `-I` options remove the diff about `$URL$ and `$Id$.
  - The `-x` options are used to exclude those files:
    - `cgal_test_with_cmake` because they are all wrong, and
    - `filelist.txt` (in doc/ subdirectories), because the order of files in those files is different.

Except for the issue with `cgal_test_with_cmake`, the diffs between the two internal releases were OK.

sloriot:
Fixes #2827

### DIFF
--- a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
+++ b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
@@ -127,14 +127,22 @@ if (CGAL_VERSION_NR)
 endif()
 file(WRITE ${release_dir}/include/CGAL/version.h "${file_content}")
 
-
-#make an additional copy for demos and examples for the testsuite
+# make an extra copy of examples and demos for the testsuite and generate
+# create_cgal_test_with_cmake for tests, demos, and examples
 if (TESTSUITE)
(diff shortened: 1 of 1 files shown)


lrineau:
If you agree, Sébastien, I will merge this PR as it is. No need to test it, because it modifies only one CMake file, that is not tested by any of our testing processes.

sloriot:
Sure

PR #2845, opened 2018-02-15
Improve internal autorefinement function

sloriot:
This brings a few improvements for the non-documented autorefinement functionality in case 3 faces intersect in a point (that is not a common vertex).
More than 3 faces intersecting and coplanar faces are not handled for now.


lrineau:

PR #2851, opened 2018-02-16
CMap bug fix in copy

gdamiand:
Bug fix when copying attributes of a cmap into another cmap, and both have no info.

## Summary of Changes

Copy also attribute in this special case.

## Release Management

* Affected package(s): CMap


lrineau:
As it is a bug-fix, should not the PR targets `releases/CGAL-4.11-branch` instead of `master`? Was the buggy code introduced only after the release of CGAL-4.11?

gdamiand:
@lrineau Yes we can target 4.11; I think the bug was here before. But this is not a critical bug, thus this is really so feel free to choose what is simpler for you.

lrineau:

PR #2852, opened 2018-02-16
do not use uppercase name for variables

sloriot:
upper case variables are more likely to be defined as macro in other code (especially EPSILON).


lrineau:

PR #1267, opened 2016-07-13
Random generator

maxGimeno:
This PR implements 4 classes to uniformly generate random points on meshes.
It also moves the function Prevent_deref from BGL to STL_Extension.

Small feature page:
https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_random_point_generators

PR #2853, opened 2018-02-16
Generator: Fix `random_points_in_triangle_mesh_2`

MaelRL:
## Summary of Changes

`random_points_in_triangle_mesh_2` is supposed to generate points in the interior faces of a 2D triangulation, but it didn't check if faces were interior or not. Tests were also broken.

## Release Management

* Affected package(s): `Generator`

### DIFF
--- a/Generator/examples/Generator/random_points_on_triangle_mesh_2.cpp
+++ b/Generator/examples/Generator/random_points_on_triangle_mesh_2.cpp
@@ -1,8 +1,11 @@
-#include <CGAL/Polygon_2.h>
 #include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+
 #include <CGAL/Constrained_Delaunay_triangulation_2.h>
-#include <CGAL/Triangulation_face_base_with_info_2.h>
 #include <CGAL/Delaunay_mesh_face_base_2.h>
+#include <CGAL/Delaunay_mesh_size_criteria_2.h>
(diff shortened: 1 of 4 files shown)


lrineau:
@maxGimeno That seems to fix examples and tests that were created by your PR #1267 (merged in CGAL-4.10). Please have a look at the patch from Mael, and let us know if the initial intent was different.

lrineau:
In particular, I am not confortable with the introduction of dependencies to Mesh_2 and Mesh_3. The Generator package is supposed to be an LGPL package.

MaelRL:
There are no new dependencies in the includes and `Mesh_3` was already being used in the test.

Mesh_2 is already required by the doc via the concept `DelaunayMeshFaceBase_2` (https://doc.cgal.org/latest/Generator/classCGAL_1_1Random__points__in__triangle__mesh__2.html). It's used in example and the test because I didn't want to bother making a custom face base and manually setting up domain flags.

lrineau:
OK, I understand. So we can assume that the package itself does not really depend on Mesh_2 or Mesh_3, because the user can implement a custom model of faces or cells. Good.

lrineau:
I mark the PR with `TODO`, to remember that Maxime has to comment on the patch.

lrineau:
Thanks Maxime.

lrineau:

PR #2860, opened 2018-02-20
 Fix CGAL_ENABLE_CHECK_HEADERS, again

lrineau:
## Summary of Changes

List of the fixes:

- The headers from the .cpp files were not processed (wrong regular
  expression for `awk`),

- `CGAL_COMPUTE_DEPENDENCIES` is now removed (unified with
  `CGAL_ENABLE_CHECK_HEADERS`).

- All `dependencies` files has an empty line (because of
  `include/CGAL/version.h` in the build directory).

- Fix the name of the target `pkg_<package>_deps`

Fix errors introduced in #2784.

## Release Management

Only the Travis check will be affected.


Installation/CMakeLists.txt
```diff
@@ -1317,12 +1303,12 @@ ${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/${header}"
...
           )
-        add_custom_target(${package}_deps DEPENDS ${CGAL_BINARY_DIR}/package_info/${package}/dependencies)
+        add_custom_target(pkg_${package}_deps DEPENDS ${CGAL_BINARY_DIR}/package_info/${package}/dependencies)
         list(APPEND check_pkg_target_list check_pkg_${package}_headers)
         list(APPEND packages_deps ${CGAL_BINARY_DIR}/package_info/${package}/dependencies)
       endif() # if the package has an include directory
-      if(${package}_deps)
-        list(REMOVE_DUPLICATES ${package}_deps)
+      if(packages_deps)
+        list(REMOVE_DUPLICATES packages_deps)
```
maxGimeno: Does that really work ? Shouldn't it be ${packages_deps} ?

lrineau: For the [if] command, if can test both values or variables:
> `if(<variable>)`
>  if the variable is defined to a value that is not a false constant. False otherwise. (Note macro arguments are not variables.)

[if]: https://cmake.org/cmake/help/v3.10/command/if.html

For the [list] command:
> `list(REMOVE_DUPLICATES <list>)`

That command could not work if you apply it to a list *value*. You have to pass it the list *variable name*, otherwise it cannot modify the variable itself.

[list]: https://cmake.org/cmake/help/v3.10/command/list.html

maxGimeno: Ok, then.

maxGimeno:
Apparently the dependencies of BGL have differences when computed from travis.

lrineau:
Successfully tested in [CGAL-4.12-Ic-194], and then merged.

  [CGAL-4.12-Ic-194]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-194.shtml

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2860)

PR #2864, opened 2018-02-21
Set the default value of CGAL_DEV_MODE from the env variable

lrineau:
If the environment variable `CGAL_DEV_MODE` is set to a "true" value,
according to CMake, then the `CGAL_DEV_MODE` CMake option is `ON` by
default.

https://cmake.org/Wiki/CMake/Language_Syntax#CMake_supports_boolean_variables.

> CMake considers an empty string, "FALSE", "OFF", "NO", or any string
> ending in "-NOTFOUND" to be false. [..] Other values are true.

That improves the PR #2799.


lrineau:

Issue #2205, opened 2017-06-23
Visual improvements when testing


MaelRL:
This issue is about improvement ideas to the way one can test code.

- When test branch packages with `create_cgal_test_with_cmake`, the compilation output is dumped in the console, which is mostly unreadable and difficult to search for warnings. I guess one could dump the output into a file and search the file, but maybe the "successful" / "error" message on the left of the program in `error.txt` could show if a warning popped up? The `grep` to detect the warnings can be picked from `collect_cgal_testresults_from_cmake`.

-The page [test suite result](https://cgal.geometryfactory.com/CGAL/testsuite/) page could use a tool to compare between two test suite runs, or at least a comparison to the fixed test suite that runs the previous release without anything. It would show some kind of (hopefully much smaller) `diff` between the two results.

MaelRL:
Adding to that, some stronger search tools (are there any at all right now?) in the test suite page would be nice.

For example, I remember seeing one example fail with an error that was unique but I can't remember which platform it was. I'm not aware of any other way than clicking all the failing platforms and searching "error", which is time-consuming...

maxGimeno:
The diff tool can be found [here](https://cgal.geometryfactory.com/~mgimeno/testsuite_comparison/diff_testsuites.html)

PR #2872, opened 2018-02-23
Travis: Fake modification of Minkowski_sum_3

lrineau:
@maxGimeno I modify this package on purpose: nothing depends on Minkowski_sum_3, so Travis should only try to build the examples and tests of this package!

### DIFF
--- a/Minkowski_sum_3/include/CGAL/minkowski_sum_3.h
+++ b/Minkowski_sum_3/include/CGAL/minkowski_sum_3.h
@@ -23,6 +23,7 @@
 
 #include <CGAL/license/Minkowski_sum_3.h>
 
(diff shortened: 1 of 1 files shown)


maxGimeno:
From what I see, it works. It is juste that it takes `install.sh` 320s to execute...

MaelRL:
Using dependencies still cuts the overall build time in half or more usually.

Maybe the dependencies could be obtained with only a minimal `install.sh` ? However, there is no really any bloating of the travis test queue ever, so I don't know if it improves testing a lot (compared to e.g. --shameless plug-- #2205).

lrineau:
`install.sh` can be improved *a lot*. I'll see what I can do at the developers meeting.

PR #561, opened 2015-12-17
Bugfix: pair not compatible with Point_3 (clang)

sgiraudot:
Bugfix for the following issues: https://github.com/CGAL/cgal/issues/49 and https://github.com/CGAL/cgal/issues/343

It turns out that in recent versions of C++ (c++11 and c++14), a pair cannot be instantiated on incomplete types. Therefore, using a pair of _Point_3_ would not compile for the following reason:
* _Point_3_ is based on _Vector_3_
* _Vector_3_ has a constructor taking a _Line_3_ as an argument
* Because of the new behavior of pair, all methods of _Vector_3_ are checked, including this constructor
* _Line_3_ is using a _std::pair_ of _Point_3_ and _Vector_3_, resulting in a loop and a compilation error

This bug is fixed by replacing the pair inside _Line_3_ by an internal _struct_ reproducing the same behavior as _std::pair_. This has been tested in [4.8-Ic-75](https://cgal.geometryfactory.com/CGAL/Members/testsuite/results-4.8-Ic-75.shtml): the bug has disappeared from the clang columns with no visible regression on other platforms.

PR #2873, opened 2018-02-26
Bugfix: pair not compatible with Circle_3 (clang)

sgiraudot:
## Summary of Changes

While working on a branch we are currently developing, @danston found a bug in the Cartesian kernel that was actually the same as https://github.com/CGAL/cgal/pull/561 but with `Circle_3`. The fix is very similar, we use an internal `struct` that is the same as a `std::pair` and it fixes it.

## Release Management

* Affected package(s): pretty much all of them…

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/Circle_3.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Circle_3.h
@@ -37,7 +37,14 @@ class CircleC3 {
   typedef typename R_::Direction_3              Direction_3;
   typedef typename R_::FT                       FT;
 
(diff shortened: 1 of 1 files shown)


lrineau:
Only for CGAL-4.12? Not for CGAL-.4.11.2?

sgiraudot:
Sorry, I did base my feature branch on `CGAL-4.11-branch` but I forgot to select it when I did the PR. It's fixed.

lrineau:
Thanks. To late for CGAL-4.12-beta1, and CGAL-4.11.1 (already prepared during the week-end). I plan it for CGAL-4.11.2 (and 4.12).

PR #2657, opened 2017-12-06
Parse headers of binary STL files starting with the keyword 'solid' correctly as binary STL

afabri:
## Summary of Changes

When a binary STL file's header started with the word 'solid' (what is recommended not to be done on [this](https://en.wikipedia.org/wiki/STL_(file_format)) wikipedia page), we parsed the file as an Ascii file.

## Release Management

* Affected package(s):  Polyhedron_IO
* Issue(s) solved: fix #2436  pointed out a file that could not be read.  We still have to document the function.

PR #2874, opened 2018-02-26
Fix reading of ASCII STL

sloriot:
A critical instruction was hidden within the patch of #2657

I do have a question for @afabri:
Your patch will not work as soon as the optional name after solid is too long. Indeed, you require that _facet_ is found before reaching the 80th character (as it's the limit of the header in binary format). Shall we consider this as an issue?

### DIFF
--- a/Polyhedron_IO/include/CGAL/IO/STL_reader.h
+++ b/Polyhedron_IO/include/CGAL/IO/STL_reader.h
@@ -47,6 +47,7 @@ namespace CGAL{
     char line[80];
     int i = 0, ni = 0;
 
+      // read the 5 first characters to check if the first word is "solid"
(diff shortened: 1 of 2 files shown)


sgiraudot:
The binary header is supposed to be exactly 80 characters (see [Wikipedia](https://en.wikipedia.org/wiki/STL_(file_format))), so I guess this should be okay?

But then again, the binary format is supposed to start by anything _except_ `solid`, and this patch is designed to handle binary files which start with `solid` (which apparently happens a lot), so I'm not sure what to think…

sgiraudot:
Ok, nevermind, I did not understand well your bug report. Indeed, if the file __is__ ASCII and the name after `solid` is too long, we have a problem.

PR #2878, opened 2018-02-27
Polyhedron demo: Restore statistics for multiple items

maxGimeno:
## Summary of Changes
After #2759 the statistics action disappeared from the context menu of the sceneview if more than one item was selected. This PR put it back.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1321,6 +1321,7 @@ void MainWindow::showSceneContextMenu(const QPoint& p) {
       else if(scene->selectionIndices().size() > 1 )
       {
         QMap<QString, QAction*> menu_actions;
+        bool has_stats = false;
         Q_FOREACH(QAction* action, scene->item(main_index)->contextMenu()->actions())
         {
           if(action->property("is_groupable").toBool())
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1334,19 +1335,23 @@ void MainWindow::showSceneContextMenu(const QPoint& p) {
           CGAL::Three::Scene_item* item = scene->item(index);
           if(!item)
             continue;
-          QVector<QString> action_names;
-          Q_FOREACH(QAction* action, item->contextMenu()->actions())
-          {
-            action_names.push_back(action->text());
```
sloriot: Was this useless?

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1334,19 +1335,23 @@ void MainWindow::showSceneContextMenu(const QPoint& p) {
...
         {
           QAction* action = menu.addAction(name);
           connect(action, &QAction::triggered, this, &MainWindow::propagate_action);
         }
-
+        if(has_stats)
+        {
+          QAction* actionStatistics =
+              menu.addAction(tr("Statistics..."));
+          actionStatistics->setObjectName("actionStatisticsOnPolyhedron");
```
sloriot: The name of the action makes me think that it will only work with polyhedra, it is just the name of the action that is misleading?

maxGimeno: It is. Originally, only scene_polyhedron_items  had statistics. I didn't change the name of the action when I added stats for others.

Polyhedron/demo/Polyhedron/MainWindow.cpp
lrineau: The variable `action_names` was not used.

maxGimeno: I thought I replied. It was a remain of an earlier attempt for displaying actions for a multiple selection of items that survived the first clean-up.

PR #2883, opened 2018-02-28
PMP: Polyline Snapping

sgiraudot:
## Summary of Changes

This introduces an internal non-documented function to snap a set of polylines under a user-specified tolerance. I also put a test and a plug-in for the demo.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
@@ -1,0 +1,302 @@
+// Copyright (c) 2018 GeometryFactory (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org).
+// You can redistribute it and/or modify it under the terms of the GNU
+// General Public License as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0+
+//
+//
+// Author(s)     : Simon Giraudot
+
+#ifndef CGAL_POLYGON_MESH_PROCESSING_POLYLINE_SNAPPING_H
+#define CGAL_POLYGON_MESH_PROCESSING_POLYLINE_SNAPPING_H
+
+#include <CGAL/license/Polygon_mesh_processing/corefinement.h>
+
+#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
+#include <CGAL/box_intersection_d.h>
+#include <CGAL/tuple.h>
+
+#include <CGAL/boost/iterator/counting_iterator.hpp>
+#include <boost/range/const_iterator.hpp>
+#include <boost/range/value_type.hpp>
+#include <boost/foreach.hpp>
+
+#include <cmath>
+
+namespace CGAL {
+namespace Polygon_mesh_processing{
+namespace internal {
+
+template <typename Exact_kernel>
+bool exact_snapping (const typename Exact_kernel::Segment_3& s0,
+                     const typename Exact_kernel::Segment_3& s1,
+                     typename Exact_kernel::Point_3& result)
+{
+  typedef typename Exact_kernel::Point_3 Point_3;
+  typedef typename Exact_kernel::Segment_3 Segment_3;
+  typedef typename Exact_kernel::Vector_3 Vector_3;
+  typedef typename Exact_kernel::Plane_3 Plane_3;
+
+  CGAL_assertion (s0.source() != s0.target());
+  CGAL_assertion (s1.source() != s1.target());
+
+  Vector_3 v0 = s0.to_vector();
+  Vector_3 v1 = s1.to_vector();
+  Vector_3 normal = CGAL::cross_product (v0, v1);
+  if (normal == CGAL::NULL_VECTOR)
+    return false;
+  
+  Plane_3 plane0 (s0.source(), normal);
+  Plane_3 plane1 (s1.source(), normal);
+
+  Segment_3 s0proj (plane1.projection (s0.source()),
+                    plane1.projection (s0.target()));
+  Segment_3 s1proj (plane0.projection (s1.source()),
+                    plane0.projection (s1.target()));
+
+  CGAL_assertion (s0proj.source() != s0proj.target());
+  CGAL_assertion (s1proj.source() != s1proj.target());
+  
+  if (!CGAL::do_intersect (s0, s1proj) ||
+      !CGAL::do_intersect (s1, s0proj))
+    return false;
+
+  Point_3 p0, p1;
+  
+  typename CGAL::cpp11::result_of<typename Exact_kernel::Intersect_3(Segment_3, Segment_3)>::type
+    intersection0 = intersection(s0, s1proj);
+  if (!intersection0)
+    return false;
+
+  if (const Point_3* p = boost::get<Point_3>(&*intersection0))
+    p0 = *p;
+  else
+    return false;
+
+  typename CGAL::cpp11::result_of<typename Exact_kernel::Intersect_3(Segment_3, Segment_3)>::type
+    intersection1 = intersection(s1, s0proj);
+  if (!intersection1)
+    return false;
+
+  if (const Point_3* p = boost::get<Point_3>(&*intersection1))
+    p1 = *p;
+  else
+    return false;
+
+  result = CGAL::midpoint (p0, p1);
+
+  return true;
+}
+
+template <class Kernel,
+          bool Has_exact_constructions=
+          !boost::is_floating_point<typename Kernel::FT>::value >
+class Exact_snapping;
+
+template <class Kernel>
+class Exact_snapping<Kernel,false>
+{
+//typedefs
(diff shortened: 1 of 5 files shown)


lrineau:
For which release do we want this PR? 4.12 or 4.13?

sgiraudot:
It's honestly no rush at all, so unless @sloriot (or you) can do a review and that the testsuite looks good soon enough for 4.12, it can easily be postponed to 4.13.

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2883)

lrineau:
Warnings:
```
In file included from /Users/cgaltester/cgal_test/CGAL-4.12-Ic-200/cmake/platforms/x86-64_Darwin-13.0_Apple-clang-5.0_Release/test/Polyhedron_Demo/Plugins/PMP/Snap_polylines_plugin.cpp:7:
/Users/cgaltester/cgal_test/CGAL-4.12-Ic-200/include/CGAL/Polygon_mesh_processing/polyline_snapping.h:275:64: warning: unused typedef 'PointIterator' [-Wunused-local-typedef]
  typedef typename boost::range_const_iterator<Polyline>::type PointIterator;
                                                               ^
/Users/cgaltester/cgal_test/CGAL-4.12-Ic-200/include/CGAL/Polygon_mesh_processing/polyline_snapping.h:281:84: warning: unused typedef 'OutputType' [-Wunused-local-typedef]
  typedef cpp11::tuple<std::size_t, std::size_t, std::size_t, std::size_t, FT, FT> OutputType;
                                                                                   ^
/Users/cgaltester/cgal_test/CGAL-4.12-Ic-200/include/CGAL/Polygon_mesh_processing/polyline_snapping.h:273:69: warning: unused typedef 'PolylineIterator' [-Wunused-local-typedef]
  typedef typename boost::range_const_iterator<PolylineRange>::type PolylineIterator;
                                                                    ^
3 warnings generated.
```
Maybe have a look at the other columns, because there might be other warnings:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-200.shtml#Polyhedron_Demo

sgiraudot:
@lrineau I see you push it to the testsuite (there are warnings I'm going to take care of). Do you want this to be in 4.12? As I said, I don't mind waiting for 4.13 for this.

lrineau:
We said we would try to push it in CGAL-4.12. So I included it.

sgiraudot:
Testsuite looks good.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+  
+  typename CGAL::cpp11::result_of<typename Exact_kernel::Intersect_3(Segment_3, Segment_3)>::type
+    intersection0 = intersection(s0, s1proj);
+  if (!intersection0)
+    return false;
+
+  if (const Point_3* p = boost::get<Point_3>(&*intersection0))
+    p0 = *p;
+  else
+    return false;
```
sloriot: I don't know what the function is supposed to do but I'm surprized you return false in case the segments are coplanar and collinear

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+  typedef typename Exact_kernel::Point_3 Point_3;
+  typedef typename Exact_kernel::Line_3 Line_3;
+  typedef typename Exact_kernel::Segment_3 Segment_3;
+  
+  Exact_kernel ek;
+  Exact_to_double exact_to_double;
+
+  Point_3 to_exact(const Inexact_point_3& p) const
+  {
+    return Point_3(p.x(), p.y(), p.z());
```
sloriot: use `Cartesian_converter`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+  }
+  Inexact_point_3 to_inexact(const Point_3& p, const Exact_to_double& etd) const
+  {
+    return Inexact_point_3(etd(p.x()),
```
sloriot: use `Cartesian_converter`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+                           etd(p.y()),
+                           etd(p.z()));
+  }
+
+public:
+
+  Exact_snapping()
```
sloriot: useless

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+  std::pair<Inexact_point_3, bool> intersection (const Inexact_segment_3& s0, const Inexact_segment_3& s1) const
+  {
+    Segment_3 exact_s0 (to_exact (s0.source()), to_exact (s0.target()));
+    Segment_3 exact_s1 (to_exact (s1.source()), to_exact (s1.target()));
+    Point_3 result;
+    
+    if (exact_snapping<Exact_kernel> (exact_s0, exact_s1, result))
+      return std::make_pair (to_inexact (result, exact_to_double), true);
+
+    return std::make_pair (Inexact_point_3(), false);
```
sloriot: not sure it's worth is but it seems that `exact_snapping` should return an `optional<Point_3>`. I agree that you then have the conversion issue so maybe it is already the best choice.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+    , output (output)
+  { }
+
+  void operator()(const Box* b, const Box* c) const
+  {
+    Segment_3 s1 (polylines[b->info().first][b->info().second],
+                  polylines[b->info().first][b->info().second + 1]);
+    Segment_3 s2 (polylines[c->info().first][c->info().second],
+                  polylines[c->info().first][c->info().second + 1]);
+    if (CGAL::squared_distance (s1, s2) > squared_tolerance)
```
sloriot: aren't you somehow doing twice the job with `exact_snapping.intersection()`? I mean couldn't you use `new_point` directly?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polyline_snapping.h
```diff
@@ -0,0 +1,324 @@
...
+                                      coord1, coord2);
+  } 
+};
+
+
+}// namespace internal
+
+
+// Note this is not officially documented  
+/*!
```
sloriot: I think this will show up in the documentation if ! is here. Anyway if not documented, the function should move in the `experimental` namespace.

lrineau:
I have temporarily added the label "not yet approved", to avoid that I accidentally merge this PR, before the discussion between Sébastien and Simon is over.

lrineau:
After a discussion with @sgiraudot, we have decided to postpone the tests of this PR for CGAL-4.13.

To be done:
- handle the case where the segments are colinear,
- even when segment are not colinear, handle the cases where the point of a segment s1 that is closest to a segment s2 is not the intersection of projections.

Maybe Simon you want to add things. I think we should in the end modify the first message of the discussion, with a summary of the to-do list.

lrineau:
@sgiraudot Have a look whether you want to refresh this PR and have it tested for CGAL-4.13.

sgiraudot:
I don't think I'll have time to do the work that has to be done (summed up in your message of March 22nd) before the release, so we should probably postpone it.

sloriot:
Shall @maxGimeno take over this PR?

maxGimeno:
@sgiraudot what do we do with this PR ?

sgiraudot:
I honestly don't know. There's still quite a lot of work to do and it's quite low priority, so I don't see it improving anytime soon…

maxGimeno:
Ok, let's postpone it then

maxGimeno:
Conflicts

sgiraudot:
I rebased the branch and did some more work on it:
- the algorithm now takes a Graph as input and directly modifies it
- degenerate cases (collinear segments or coplanar segments) are well detected, and a first version of proximities detection is done
- so far, some cases are not handled, specifically cases were a segment is completely included in the tolerance cylinder of another coplanar segment

Actually, it is really difficult to figure out a way to deal with all possible cases. The problem seems ill-posed, as the same polyline set will give very different results depending on how densely it's sampled: it seems we should consider polylines as a linear distribution and not a set of segments, otherwise we are dependent on vertex densities/positions. But on the other hand, I don't see an easy way to do it without the points.

Handling collinear segments gives also a sort of instability: two segments almost collinear would snap at 1 point, but as soon as they become collinear, they should snap as a common segment?

Depending on how the handle proximities, we can also get quite different outputs, see this coplanar example:

image

There does not seem to be a right answer. If you start considering cases where N polylines might snap at the same location with a combination of degenerate cases, it quickly becomes a nightmare.

In all cases, we also have the problem that applying the algorithm several times would sequentially collapse the polylines onto each other, which is not the case with non-degenerate cases (that would remain stable).

I'm wondering if we should not just make it a precondition that no segments are collinear/coplanar (or use random perturbations to make sure of that), but that might be too

maxGimeno:
Problems in license :
```
 There are modifications in licenses (check if all modified/added headers files contain a valid SPDX tag compatible with the one in package_info/PKG/license and that no license text is present):
 M Polygon_mesh_processing/package_info/Polygon_mesh_processing/license.txt
 ```

PR #2889, opened 2018-03-02
Use the hash of the last modification per file

sloriot:
The hash used to set $Id$ was HEAD and not the hash of the last modification of each file.


lrineau:

PR #2892, opened 2018-03-05
Polyhedron demo: Fix Region Selection / Position Request

sgiraudot:
## Summary of Changes

Position request (ctrl + click) and region selection (shift + click) on point sets were buggy when an item has very large coordinates (and the `offset` trick was used). This fixes it.

## Release Management

* Affected package(s): Polyhedron Demo


lrineau:
Maxime @maxGimeno, could you verify if the same issue could be in the c3t3 item? I have the same bug with c3t3s: no picking. It might be because of the large coordinates.

maxGimeno:
@lrineau it is worse. There is no offset in the c3t3 item. I'm on it.

maxGimeno:
See #2896 for c3t3

lrineau:

PR #2896, opened 2018-03-05
Polyhedron demo: fix c3t3 display with big coordinates

maxGimeno:
## Summary of Changes

This PR fixes the plane and the display of the c3t3_item when it has big coordinates.


lrineau:
I tested it on my Linux machine and it works well.

lrineau:

PR #2109, opened 2017-05-11
Polyhedron_demo: Make the demo customizable at build time

maxGimeno:
## Summary of Changes
Add options to the CMakeLists.txt that allow to chose easily what will or will not be built in the default target thanks to  cmake-gui.
This is done using checkboxes : one per plugin and one per "category", which regroup related plugins.

Also and add four targets that emulate the categories.

To use this : you can either call cmake-gui for the demo, and clic on configure to apply your changes, then generate and call make in the console;
or you can call make category-[ visualizer | mesh_processing | point_set_processing | meshing ].

@afabri @sgiraudot can you try it and give me some feedback ?

## Release Management

* Affected package(s):Polyhedron_demo
* Issue(s) solved (if any): fix #2040

PR #2897, opened 2018-03-05
Travis: Time enhancement

maxGimeno:
## Summary of Changes
This PR removes the polyhedron_demo from the travis matrix, as it timeouts too often. 
It also optimizes the installation script so it doesn't take 5min+ anymore, reduces the depth of the CGAL clone to 3 instead of default 50, and it only installs openMesh for the jobs that need it. 

A run that doesn't touch any dependency, is done in ~50min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,95 +10,95 @@ matrix:
   - compiler: clang-3.6
     env: PACKAGE='CHECK' 
   - compiler: clang-3.6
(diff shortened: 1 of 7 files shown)


MaelRL:
Could #2109 maybe be used in the future to split the demo in two jobs and make the build time smaller than the timeout ? For some packages, the demo uses features that are not tested in the package examples/tests so it's useful to have.

maxGimeno:
Yes, that's what we want to do actually.

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2897)

PR #2900, opened 2018-03-06
Improve removal of self-intersections for the demo plugin

sloriot:
This is modifying a non-documented function that is used in a plug-in of the demo, marked as experimental.


lrineau:
Successfully tested in [CGAL-4.12-Ic-192], and then merged.

  [CGAL-4.12-Ic-192]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-192.shtml

PR #2912, opened 2018-03-08
More fixes for GNU 8.0.1: fallthrough warnings  and constness

afabri:
## Summary of Changes

- Replace the third `case` with `default` in ternary `switch`. 
- Declare an `operator<()`  const

## Release Management

* Affected package(s): Kernel_23, Intersections_2, Convex_decomposition_3

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_rep_2.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_rep_2.h
@@ -145,19 +145,19 @@ friend class Scaling_repC2<R>;
             {
               case 0: return t11;
               case 1: return t12;
-              case 2: return t13;
+              default: return t13;
             }
     case 1: switch (j)
             {
               case 0: return t21;
               case 1: return t22;
-              case 2: return t23;
+              default: return t23;
             }
     case 2: switch (j)
             {
               case 0: return FT(0);
               case 1: return FT(0);
-              case 2: return FT(1);
+              default: return FT(1);
             }
     }
     return FT(0);
--- a/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_rep_3.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_rep_3.h
@@ -162,28 +162,28 @@ public:
               case 0: return t11;
               case 1: return t12;
               case 2: return t13;
-              case 3: return t14;
+              default: return t14;
             }
     case 1: switch (j)
             {
               case 0: return t21;
               case 1: return t22;
               case 2: return t23;
-              case 3: return t24;
+              default: return t24;
             }
     case 2: switch (j)
             {
               case 0: return t31;
               case 1: return t32;
               case 2: return t33;
-              case 3: return t34;
+              default: return t34;
             }
     case 3: switch (j)
             {
               case 0: return FT(0);
               case 1: return FT(0);
               case 2: return FT(0);
-              case 3: return FT(1);
+              default: return FT(1);
             }
     }
   return FT(0);
--- a/Cartesian_kernel/include/CGAL/Cartesian/Rotation_rep_2.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Rotation_rep_2.h
@@ -149,19 +149,19 @@ public:
             {
               case 0: return cosinus_;
               case 1: return -sinus_;
-              case 2: return FT(0);
+              default: return FT(0);
             }
     case 1: switch (j)
             {
               case 0: return sinus_;
               case 1: return cosinus_;
-              case 2: return FT(0);
+              default: return FT(0);
             }
     case 2: switch (j)
             {
               case 0: return FT(0);
               case 1: return FT(0);
-              case 2: return FT(1);
+              default: return FT(1);
             }
     }
     return FT(0);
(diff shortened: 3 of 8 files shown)


mglisse:
I still don't like it: it generates worse code and is harder to debug than adding `default: CGAL_ASSUME(false);` (possibly spelled `CGAL_EXHAUSTIVE_SWITCH`). But I guess it is too late, might as well continue this way...

afabri:
Why does it generate worse code? At least now we have told the compiler that there are exactly 3 cases.

mglisse:
> Why does it generate worse code? At least now we have told the compiler that there are exactly 3 cases.

With `default: CGAL_ASSUME(false);` we give more information and thus more flexibility to the compiler, it is free to generate the same code as with your version for instance (and it often does). And in some cases it can generate better. For instance, when the cases for `i` are 0, 1 and 2, it is free to generate an array of size 3 and use `array[i]`.

afabri:
It's never to late to improve. (and to learn from you - thanks)

mglisse:
I can't promise that it will always be better with all compilers. My first motivation is `-fsanitize=unreachable` (part of `-fsanitize=undefined`).

lrineau:
Marc is right. You can see the results here, with that wonderful tool "Compiler Explorer": https://godbolt.org/g/WQvVAZ

The left column if with `default:` and two different compilers (gcc/clang), and the right column uses a switch with exactly the three values of the enum, with the same compilers (but maybe older versions).

Note that clang 3.7 and later version does better when `default:` is not used, because the compiler knows that it also has to deal with three values (and the code will segfault if another integer value is casted into the enum). Note also that a very recent gcc trunk version produces the same code as a recent clang versions. I have played a bit with that tool, and it seems MSVC cannot do the optimization, at least with the optimization flags I used.

Anyway I wonder if we can find a use case in CGAL where we can see the difference, because in all use cases such a switch will be mixed up with a lot of heavy computation.`

lrineau:
Should I merge the PR as it is?

lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2912)

lrineau:
The remaining warnings are:
```
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:788:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:794:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:826:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:832:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:855:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:861:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:885:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:891:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:915:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:921:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:945:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:951:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:975:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH2.h:981:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:548:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:555:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:562:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:692:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:699:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Homogeneous/Aff_transformationH3.h:706:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
/mnt/testsuite/include/CGAL/Rotational_sweep_visibility_2.h:211:9: warning: this statement may fall through [-Wimplicit-fallthrough=]
```

afabri:
With what you write about the latest compilers which prefer NO `default:`  and which segfault when using an int that does not fit to the `enum`, I would even suggest to leave the code as it is and to suppress the warning with a `#pragma`.

lrineau:

PR #2918, opened 2018-03-12
Polygon: avoid segfaulting when the polygon/range is empty

MaelRL:
## Summary of Changes

Currently, `is_simple()` and similar segfault if called on an empty polygon / range. This PR checks for emptiness. `True` is chosen for symmetry with `is_convex`, which returns `true` when the polygon/range is empty.

## Release Management

* Affected package(s): `Polygon`

### DIFF
--- a/Polygon/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
+++ b/Polygon/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
@@ -49,6 +49,8 @@ bool is_simple_2(ForwardIterator first,
                       ForwardIterator last,
                       const PolygonTraits& traits)
 {
+    if (first == last) return true;
+
     return is_simple_polygon(first, last, traits);
 }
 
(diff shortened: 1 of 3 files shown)


lrineau:
This change is [Reviewable](https://reviewable.io/reviews/cgal/cgal/2918)

MaelRL:
I think the three red squares for Polygon on Windows in the testsuite are from this branch.

@afabri Could you please check it? The branch is `cgal-dev/testsuite-Monday` or `MaelRL:Polygon-Is_simple_with_empty_polygon-GF` and the executable is `Polygon/test/Polygon/AlgorithmTest.cpp` (some fail in debug, some in release). It's complaining that it cannot dereference an iterator but there should be nothing to dereference since the range is empty, thus `begin==end` and we exit.

afabri:
It's a trivial bug, as you dereference an iterator after a `clean()` of its container.

lrineau:

PR #2926, opened 2018-03-14
Use https for all links to www.cgal.org and doc.cgal.org

sloriot:
patch generated with:
`> ack http://doc.cgal -l | xargs sed -i 's/http:\/\/doc.cgal.org/https:\/\/doc.cgal.org/g'`
`> ack http://www.cgal -l | xargs sed -i 's/http:\/\/www.cgal.org/https:\/\/www.cgal.org/g'`


lrineau:
> patch generated with:
> ```shell
> ack http://doc.cgal -l | xargs sed -i 's/http:\/\/doc.cgal.org/https:\/\/doc.cgal.org/g'
> ack http://www.cgal -l | xargs sed -i 's/http:\/\/www.cgal.org/https:\/\/www.cgal.org/g'
> ```

Just a remark: with [sed you can use `s|||` instead of `s///`][*], to avoid quoting the `/`.

[*]: http://backreference.org/2010/02/20/using-different-delimiters-in-sed/

lrineau:

PR #2846, opened 2018-02-15
Fix CGAL_HEADER_ONLY

lrineau:
## Summary of Changes

Try to fix errors or feature requests of #2656.

## Release Management

* Affected package(s): CMake/Installation
* Issue(s) solved (if any): fix #2656

PR #2927, opened 2018-03-14
Fix CMake issues

### LINKED ISSUE #1580: CMake error when WITH_CGAL_Qt5 is set to ON then OFF
Steps to reproduce, in a empty build directory:
- configure CGAL by default
- _do not compile_ yet,
- add `-DWITH_CGAL_Qt5=OFF`
- then compile CGAL libraries
- verify that the CGAL_Qt5 lib was not created,
- and then compile any CGAL program using the produced`CGALConfig.cmake`.
  That gives a CMake error about the imported target `CGAL::CGALQt5` being missing.

@CGAL/geometryfactory Can you please confirm the bug?

Cc: @afabri

I am not sure if the bug is in `master` or if it was already in CGAL-4.9.

lrineau:
## Summary of Changes

- Fix bug introduced by PR #2846
- Fix issue #1580

## Release Management

* Affected package(s): CGAL Installation
* Issue(s) solved (if any): fix #1580, fix #2846


lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-201.shtml

lrineau:
@CGAL/geometryfactory I will need help to test the feature. On both Linux and Windows.

I just merged this PR in `master`, for convenience.

The documentation, so far, does not exist.

In addition to all the existing use case:
1. **configure and compile CGAL** libraries,
2. **configure CGAL header-only**,
now we have a new use case:
3. **Do not configure CGAL at all**, but only configure your user-program or demo
   To do so, use the *root of your Git working directory* as the `CGAL_DIR` CMake (or environment) variable, when you configure your user-program or demo.

There is also a fourth case, that was already existing but got improved *a lot*:
4. **configure all CGAL with `CGAL_HEADER_ONLY`** (and `WITH_demos`, `WITH_examples`, `WITH_tests`), for example to use CTest easily. That should now be *a lot* faster: only 12 seconds on my Linux laptop with the SSD disk.

Please test anything you like, and report what you tested and the results, to me, or here in this discussion.

gdamiand:
Good job @lrineau for this PR  !!

janetournois:
I just tested, with MSVC 2013, 64 bits
3. for the examples of `Convex_hull_3`, and it works perfectly
4. configuration of **all** CGAL took about 16 seconds on my laptop

lrineau:
Thanks @janetournois, @gdamiand.

@afabri You should try: it seems that the performance improvements of the CMake code are huge, on Windows. Probably because now the `<CGAL/compiler_config.h>` header, previously generated by compiling and testing C++ piece of code, is now pre-generated.

cjamin:
Just tested with MSVC 2017, 64 bits.

3. Mesh_3 examples. CMake configuration works fine. No time to build the C++ now.
4. It takes 75 seconds on my rather old computer, source files being on a HDD. Looks like it spends a bit of time when not finding some librairies (e.g. VTK). Anyway, it is a huge improvement!

Very good job, thank you.

sgiraudot:
> 4. configure all CGAL with CGAL_HEADER_ONLY (and WITH_demos, WITH_examples, WITH_tests), for example to use CTest easily. That should now be a lot faster: only 12 seconds on my Linux laptop with the SSD disk.

This works good! That was actually the reason I stopped using `CGAL_HEADER_ONLY` before (because it took ages to configure it with tests and examples). Now it works well.

> 3. Do not configure CGAL at all, but only configure your user-program or demo
To do so, use the root of your Git working directory as the CGAL_DIR CMake (or environment) variable, when you configure your user-program or demo.

This is nice and it works well too.

I still have some issues/questions:

- for some reason, my programs are now in `Debug` mode by default (I have `CGAL_DEV_MODE=1` in my environment variables), even if CGAL was configured with `Release` (but still in header only). But maybe this is expected? Basically:
  * CGAL compiled with `Release` mode -> program gets `Release` mode (okay)
  * CGAL header only not configured -> program gets `Debug` mode (why not)
  * CGAL header only configured in `Release` mode -> program gets `Debug` mode (seems not okay)
- I used to create some symlinks automatically for commodity (`DEMO -> Polyhedron/demo/Polyhedron`, `PSP -> Point_set_processing_3`, etc.). But now I can't compile the demo if I configure it from this symlink. From what I understand, the paths are computed with respect to the current path, and the number of `../` needed to reach the root is inexact. I have this kind of error:
```bash
/home/simon/CGAL/master/DEMO/build_no_configure/__/GraphicsView/include/CGAL/Qt/moc_DemosMainWindow.cpp:9:10: fatal error: 
      '../../../../../../../GraphicsView/include/CGAL/Qt/DemosMainWindow.h' file not found
#include "../../../../../../../GraphicsView/include/CGAL/Qt/DemosMainWindow.h"
```
(If I don't configure CMake from the symlink but from the full directory, no error)
I have this error in header only (configured or not), not in the compiled mode.

lrineau:
@sgiraudot Could you try again with this patch?
```patch
diff --git a/Installation/lib/cmake/CGAL/CGALConfig.cmake b/Installation/lib/cmake/CGAL/CGALConfig.cmake
index d0fda5865a2..3f7a53b4c47 100644
--- a/Installation/lib/cmake/CGAL/CGALConfig.cmake
+++ b/Installation/lib/cmake/CGAL/CGALConfig.cmake
@@ -36,6 +36,7 @@ endif()
 message(STATUS "Using header-only CGAL")
 
 get_filename_component(CGAL_CONFIG_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
+get_filename_component(CGAL_CONFIG_DIR "${CGAL_CONFIG_DIR}" ABSOLUTE)
 
 set(CGAL_HEADER_ONLY TRUE)
 
```

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Mar 20, 2018, 10:38 AM GMT+1](https://github.com/CGAL/cgal/pull/2927#issuecomment-374532872 "2018-03-20T09:38:06Z - Replied by Github Reply Comments"):
> - for some reason, my programs are now in `Debug` mode by default (I have `CGAL_DEV_MODE=1` in my environment variables), even if CGAL was configured with `Release` (but still in header only). But maybe this is expected? Basically:
>   * CGAL compiled with `Release` mode -> program gets `Release` mode (okay)
>   * CGAL header only not configured -> program gets `Debug` mode (why not)
>   * CGAL header only configured in `Release` mode -> program gets `Debug` mode (seems not okay)

I understand. When you configure CGAL in header-only, nothing is actually registered in `CGALConfig.cmake`, as it was before. That is why all user-program restart from new. Without `CGAL_DEV_MODE`, the default `CMAKE_BUILD_TYPE` is the empty string, and the compilation and linker flags are empty.

I understand that can be a surprise for a CGAL developers, used to the old (and non-standard) behavior  What would you suggest? I have no clue what we should do... Maybe nothing, and get used to the new behavior (standard behavior of CMake, plus the trick done by `CGAL_DEV_MODE`).

afabri:
Maybe a message during configuration.

MaelRL:
Testing point 1. (configure and compile) on Fedora 26 using `dev_mode`, `with_*` and a bunch of clang flags in a directory CGAL_ROOT/build, everything goes fine but:
- The default build type of CGAL is "Debug", not "Release"
- After `cmake ..`, `make`, and moving to a package, the `CGAL_DIR` variable is usually set by default to CGAL_ROOT whereas it should be CGAL_ROOT/build.  Strangely it sometimes even finds the root dir of some of my other CGAL clones.
- If I use `dev_mode` in the package and the wrong `CGAL_DIR` is set (that is, not `CGAL_ROOT/build`) then when I configure, the build type and the flags from `CGAL_ROOT/build` are not forwarded properly and even if I fix the `CGAL_DIR` after, forwarding will not do anything, and I have to clean cmake files.

maxGimeno:
I had the same issue as Mael, I had to delete my CGAL dir in usr/local
because even with CGAL_DIR set to CGAL_ROOT/Build, it was still using
/usr/local/CGAL/ headers.

2018-03-20 11:14 GMT+01:00 Mael Rouxel-Labbé <notifications@github.com>:

sgiraudot:
@lrineau it still doesn't work with your patch.

For the mode, the `CGAL_DEV_MODE` solution was pretty neat, but if it doesn't work similarly in header only mode, I honestly think the best solution would be for developers to get used to defining the mode by hand. Anyway, the old way was not standard at all and users will need to select `Release` by hand, so we might as well do it too. And it's easy to create an alias `alias cmake-release="cmake -DCMAKE_BUILD_TYPE=Release"`, that's what I did before you told me about the `CGAL_DEV_MODE`.

As I said before, I'm more concerned that users won't add `Release` and get crappy performances. In that case, I agree with @afabri that putting a message during compilation (such as `Warning: programs using CGAL should be compiled in Release mode to get the best performances`) would be helpful.

lrineau:
@sloriot 
[**@MaelRL**](https://github.com/MaelRL) commented on [Mar 20, 2018, 11:14 AM GMT+1](https://github.com/CGAL/cgal/pull/2927#issuecomment-374543701 "2018-03-20T10:14:38Z - Replied by Github Reply Comments"):
> Testing point 1. (configure and compile) on Fedora 26 using `dev_mode`, `with_*` and a bunch of clang flags in a directory CGAL_ROOT/build, everything goes fine but:
> - The default build type of CGAL is "Debug", not "Release"

Indeed. With `CGAL_DEV_MODE`, that is hardcoded that the default of CMAKE_BUILD_TYPE is `Debug`. Without `CGAL_DEV_MODE`, then the default is the CMake default: the empty string, with empty flags.

@CGAL/geometryfactory  Is that something we want to change?
@sloriot: You were the one requesting `CGAL_DEV_MODE`. What is you opinion on that?

> - After `cmake ..`, `make`, and moving to a package, the `CGAL_DIR` variable is usually set by default to `CGAL_ROOT` whereas it should be `CGAL_ROOT/build`.  Strangely it sometimes even finds the root dir of some of my other CGAL clones.

I find this strange. If the `CGAL_DIR` environment variable is empty, then the default paths were `find_package(CGAL ...)` searches for CGAL is defined by [the documentation of `find_package`][doc]. Maybe you have `/path/to/CGAL_ROOT/bin` in your PATH environment variable.

[doc]: https://cmake.org/cmake/help/v3.9/command/find_package.html

> - If I use `dev_mode` in the package and the wrong `CGAL_DIR` is set (that is, not `CGAL_ROOT/build`) then when I configure, the build type and the flags from `CGAL_ROOT/build` are not forwarded properly and even if I fix the `CGAL_DIR` after, forwarding will not do anything, and I have to clean cmake files.

Indeed. But that behavior was already that way with CGAL-4.11 and earlier. I am not a big fan of `CGAL_DEV_MODE` (or of the old behavior of CGAL-4.11), and I would not like to workaround to much issues of that behavior.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Mar 20, 2018, 11:18 AM GMT+1](https://github.com/CGAL/cgal/pull/2927#issuecomment-374544692 "2018-03-20T10:18:01Z - Replied by Github Reply Comments"):
> I had the same issue as Mael, I had to delete my CGAL dir in usr/local
> because even with CGAL_DIR set to CGAL_ROOT/Build, it was still using
> /usr/local/CGAL/ headers.

Anyway, as soon as you do not fill the environment or CMake variable `CGAL_DIR` yourself, CMake uses its standard way of searching, as documented in [`find_package`][doc]:

[doc]: https://cmake.org/cmake/help/v3.9/command/find_package.html

`/usr/local/` is one of the prefixes where CMake searches by default. That is the expected behavior.

lrineau:
By the way, if we added `HINTS ../../ ../../..` to all our calls to `find_package(CGAL ...)` in `master`, then CMake would find by default (unless `CGAL_DIR` is specified differently) the `CGALConfig.cmake` at the root of the Git repository, or at the root of a release tarball, and that would use CGAL header-only.

maxGimeno:
> Anyway, as soon as you do not fill the environment or CMake variable CGAL_DIR yourself, CMake uses its standard way of searching, as documented in find_package:

In my CMakeCache it was set as I wanted. I checked 4 times.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Mar 20, 2018, 2:08 PM GMT+1](https://github.com/CGAL/cgal/pull/2927#issuecomment-374590857 "2018-03-20T13:08:50Z - Replied by Github Reply Comments"):
> > Anyway, as soon as you do not fill the environment or CMake variable CGAL_DIR yourself, CMake uses its standard way of searching, as documented in find_package:
> 
> In my CMakeCache it was set as I wanted. I checked 4 times.

Maybe a third-party library, like libQGLViewer, is in /usr/local/ as well, and the `-I` options for libQGLViewer made the compiler find the CGAL headers in `/usr/local/include/`. I do not think that is related to header-only or not, or to the recent development in our CMake script.

Even 15 years ago, when I started developing things in CGAL (probably Mesh_2), I had difficulties because CGAL was install in `/usr/include/`, and sometimes that was preferred to the header files I was developing in the svn repository.

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Mar 20, 2018, 10:38 AM GMT+1](https://github.com/CGAL/cgal/pull/2927#issuecomment-374532872 "2018-03-20T09:38:06Z - Replied by Github Reply Comments"):
> - I used to create some symlinks automatically for commodity (`DEMO -> Polyhedron/demo/Polyhedron`, `PSP -> Point_set_processing_3`, etc.). But now I can't compile the demo if I configure it from this symlink. From what I understand, the paths are computed with respect to the current path, and the number of `../` needed to reach the root is inexact. I have this kind of error:
> ```bash
> /home/simon/CGAL/master/DEMO/build_no_configure/__/GraphicsView/include/CGAL/Qt/moc_DemosMainWindow.cpp:9:10: fatal error: 
>       '../../../../../../../GraphicsView/include/CGAL/Qt/DemosMainWindow.h' file not found
> #include "../../../../../../../GraphicsView/include/CGAL/Qt/DemosMainWindow.h"
> ```
> (If I don't configure CMake from the symlink but from the full directory, no error)
> I have this error in header only (configured or not), not in the compiled mode.

That is a general CMake bug, that has an impact on automoc:
* discussed first in Mantis: [CMake bug 16047](https://public.kitware.com/Bug/view.php?id=16047),
* then converted to a bug on gitlab.kitware.com: [issue #16047](https://gitlab.kitware.com/cmake/cmake/issues/16047),
* closed in favor of [issue #16228: "Fix translation map issues"](https://gitlab.kitware.com/cmake/cmake/issues/16228).

One workaround is that shell alias:
```shell
alias cmake="PWD=$(pwd -P) cmake"
```
so that CMake only see the real current working directory, and not the path with symbolic links.

PR #2931, opened 2018-03-16
Fix the determinism of the computation of point and base vectors

sloriot:
`possibly` will return true in case the evaluation lead to true or false
as possible values. If the exact value is computed or interval refined,
false could be the true value which has a consequence will give a
different result for point or base vector.

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -1771,10 +1771,10 @@ namespace CartesianKernelFunctors {
         // to avoid badly defined vectors with coordinates all close
         // to 0 when the plane is almost horizontal, we ignore the
         // smallest coordinate instead of always ignoring Z
-        if (CGAL::possibly(CGAL_AND (a <= b, a <= c)))
+        if (a <= b && a <= c)
           return Vector_3(FT(0), -h.c(), h.b());
 
-        if (CGAL::possibly(CGAL_AND (b <= a, b <= c)))
+        if (b <= a && b <= c)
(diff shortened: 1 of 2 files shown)


sloriot:
Not necessarily. Inside a construction that needs to be deterministic, it is dangerous. In a predicate to filter out some cases it is not an issue.

lrineau:
You are right. Sometimes constructors are used in the implementation of predicates, though. Then we will do it on case by case.

lrineau:
@maxGimeno The PR is based on CGAL-4.11.1, and Travis tests the merge with `master`. It seems it all goes wrong. That would be nice if you can find a way to fix this, if possible.

maxGimeno:
I don't know why it does that. When i try to PR a dummy branch based on 4.11 into 4.11, Travis uses the yml file from back then, not the one used here. 
> ￼ sloriot changed the base branch from master  to releases/CGAL-4.11-branch  4 days ago

Maybe this is what messed things up.

MaelRL:
Another travis PR went very wrong in https://github.com/CGAL/cgal/pull/2918 (https://travis-ci.org/CGAL/cgal/builds/354321164) and I also changed the base branch there, other PRs based on 4.11 opened around the same time and without base change went fine.

maxGimeno:
I confirmed with a dummy test. Basing the PR directly on 4.11 works. Changed afterwards keeps master's yml file and it doesn't work.

lrineau:
Thanks Maxime.

Now that we know about that bug, we should be careful, and maybe do not change the base branch but open a second PR.

maxGimeno:
Yes, I think it is the easiest way to avoid that behavior

lrineau:

Issue #2430, opened 2017-09-14
Hole fill should refuse filling a hole if it has a null edge


sloriot:
This will end up breaking an assertion if the hole is made of 3 vertices.

sloriot:
@kkatrio When you have time, it would be nice if you can handle this issue.

kkatrio:
ok, will do.

kkatrio:
When trying to fill a hole which is created by at least one null edge, then I get either "Returning no output. No possible triangulation is found!" or "Returning no output. Filling hole with extra triangles is not successful!", depending on if I tried with DT or not. Should I try something in particular to reproduce the issue?

sloriot:
Did you try with a hole made only of 3 vertices?

lrineau:
Fixed by #2932.

PR #2932, opened 2018-03-16
PMP hole filling: null edge

kkatrio:
## Summary of Changes

Deals with issue: #2430 
Refuses attempt to triangulate a hole of 3 vertices with a null edge between them, and fixes the relevant warnings on the polyhedron demo plugin.


sloriot:
The patch only fixes the case of a hole made of 3 vertices, 2 having the same points. In the issue you reported that the case of more that 3 vertices is already handled. Is it handled by checking no consecutive vertices with identical points are present?
Also I wonder if your patch is able to handle _3 1 1 0 -1 -1 0 -1 -1 0 _.

kkatrio:
> The patch only fixes the case of a hole made of 3 vertices, 2 having the same points.

correct. This is what I found to be missing.

The case for more than 3 vertices yes, it is handled already: 

- with DT it returns nothing if the dimension of the convex hull of the DT is less than 2.
- without DT it returns nothing when the points of a triangle of the calculated triangulation are collinear.

> Also I wonder if your patch is able to handle _3 1 1 0 -1 -1 0 -1 -1 0 _.

Here the first point will be added as a fourth, to follow the convention of the implementation. (first point = last point on the list) Thus, this case falls to the existing handling.

sloriot:
With DT: the dimension might be 3 even if I have 2 vertices having identical points.
Without DT: does it means that all possible triangulation are evaluated and all of them will be rejected because a triangle is always degenerate?

kkatrio:
If the dimension of the DT hull is 2 or more, then a triangulation will be calculated. Then, if in that there is a degenerate triangle, nothing is returned.
Similarly without DT, all possible triangulations are evaluated but one is selected (with the min weight). If in that there is a degenerate triangle, nothing is returned.

sloriot:
OK that what I though from your description, but I wanted to be sure. Instead, shouldn't we check at the beginning that no two consecutive vertices have similar points?

lrineau:
@kkatrio @sloriot 
I see you are still discussing. Is this pull-request ready to be tested, and integrated?

kkatrio:
@sloriot You mean refuse hole filling on any dataset that contains two consecutive points which are collinear? Isn't this too restrictive?

sloriot:
No I mean testing if a segment is degenerated explicitly instead of letting the process fail.

lrineau:

PR #2943, opened 2018-03-20
Improve CMake scripts,  continued

lrineau:
## Summary of Changes

- With `CGAL_DEV_MODE`, the default of `CMAKE_BUILD_TYPE` is now "Release" instead of "Debug"
- Add a warning at the end of the configuration, when `CMAKE_BUILD_TYPE` is either empty or "Debug".

## Release Management

* Affected package(s): CMake, all CGAL

@sgiraudot @MaelRL Could you try again, now?

As for the default `CMAKE_BUILD_TYPE`, I found out that we can set a variable `CMAKE_BUILD_TYPE_INIT` to initialize it, but it has to be set before the call to `project(..)`. Do we want that?

### DIFF
--- a/Installation/CMakeLists.txt
+++ b/Installation/CMakeLists.txt
@@ -378,6 +378,7 @@ set( CGAL_MODULES_REL_DIR cmake/modules )
 set( CGAL_MODULES_DIR     ${CGAL_INSTALLATION_PACKAGE_DIR}/${CGAL_MODULES_REL_DIR} )
 
 include(${CGAL_MODULES_DIR}/CGAL_Macros.cmake)
+include(${CGAL_MODULES_DIR}/CGAL_enable_end_of_configuration_hook.cmake)
 cgal_setup_module_path()
 
 message( STATUS "CGAL_REFERENCE_CACHE_DIR=${CGAL_REFERENCE_CACHE_DIR}" )
--- a/Installation/cmake/modules/CGALConfig_binary.cmake.in
+++ b/Installation/cmake/modules/CGALConfig_binary.cmake.in
@@ -190,3 +190,5 @@ if( CGAL_DEV_MODE OR RUNNING_CGAL_AUTO_TEST )
   # Nov. 2017). -- Laurent Rineau
   include(CGAL_SetupFlags)
 endif()
+
+include("${CGAL_MODULES_DIR}/CGAL_enable_end_of_configuration_hook.cmake")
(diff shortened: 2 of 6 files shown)


lrineau:
The warning will be:
```
  =======================================================================

  CGAL performance notice:

  The variable CMAKE_BUILD_TYPE is set to "Debug".  That value should be used
  only for debugging.  For performance reasons, you should set
  CMAKE_BUILD_TYPE to "Release".

  Set CGAL_DO_NOT_WARN_ABOUT_CMAKE_BUILD_TYPE to TRUE if you want to disable
  this warning.

  =======================================================================
```
and it will be a non-warning message, if `RUNNING_CGAL_AUTO_TEST`, to avoid warnings in the test suite.

MaelRL:
Seems to work just fine (including playing with that disabling flag), although the warning when the build type is left empty (`The variable CMAKE_BUILD_TYPE is set to "".  That value should be used only for debugging.`) might confuse some people.

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Mar 20, 2018, 5:28 PM GMT+1](https://github.com/CGAL/cgal/pull/2943#issuecomment-374663768 "2018-03-20T16:28:08Z - Replied by Github Reply Comments"):
> Seems to work just fine (including playing with that disabling flag), although the warning when the build type is left empty (`The variable CMAKE_BUILD_TYPE is set to "".  That value should be used only for debugging.`) might confuse some people.

Do you suggest there should be a different message for the empty string and for "Debug"?

afabri:
if ( VAR == "") message "The variable CMAKE_BUILD_TYPE is empty"

MaelRL:
I think you can just drop the second sentence (which is what is a bit weird since an empty build type is not really adapted to debugging if I understood correctly) and leave the rest as is.

sloriot:
_With CGAL_DEV_MODE, the default of CMAKE_BUILD_TYPE is now "Release" instead of "Debug"_
What does that mean? If I have configured the lib in debug mode will my example be always release?

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Mar 20, 2018, 7:15 PM GMT+1](https://github.com/CGAL/cgal/pull/2943#issuecomment-374703133 "2018-03-20T18:15:45Z - Replied by Github Reply Comments"):
> _With CGAL_DEV_MODE, the default of CMAKE_BUILD_TYPE is now "Release" instead of "Debug"_
> What does that mean? If I have configured the lib in debug mode will my example be always release?

If you have configured the **non-header-only** library in Debug, then your user-project will be in Debug by default. But in all the other cases if will be Release.

I am sorry but with header-only, the configuration of CGAL libraries no longer register any compilation flags (or `CMAKE_BUILD_TYPE`) in the `CGALConfig.cmake`, so that feature `CGAL_DEV_MODE` is almost stillborn. Well:
- at least the addition of `-Wall` will remain (with `CGAL_DEV_MODE`),
- and anyway the default `CMAKE_BUILD_TYPE`, without `CGAL_DEV_MODE`, so for most our users, is still the default *empty string*!

lrineau:

Issue #1317, opened 2016-07-29
Mesh smoothing


palliez:
It is possible to fair - but this requires constraining vertices. 
does it make sense to add a menu for smoothing? Laplacian with or without volume preserving, with feature preserving (aniso smoothing) when detected.

lrineau:
@janetournois Can you comment?

@palliez I must admit I have not fully understood. Can you be a bit more specific?

janetournois:
I absolutely makes sense, but it's a lot more work than adding a menu. 
The `Polygon_mesh_processing::fair()` function requires those constrained points.

I would like to add all the state-of-the-art surface mesh smoothing methods to the PMP package and add them to the demo.

janetournois:
There is a chance we can work on this in 2017. Do you have more suggestions of published/state-of-the-art/prototype code that we would like to have published in CGAL?
Thanks for your input!

lrineau:
Follow-up: it will be [proposed as a GSoC subject in 2017][1]. Let's hope CGAL will be accepted as a GSoC organization this year.

[1]: https://cgal.geometryfactory.com/CGAL/Members/wiki/GSoC_2017/GSOC17_draft_Project#Mesh_Smoothing_and_Shape_Smoothing

sgiraudot:
Was this project accepted for the GSoC this year?

afabri:
Yes. See [here](https://summerofcode.withgoogle.com/organizations/5926060504907776/#4839851680595968).  The mentor is @janetournois

sgiraudot:
Good news!

PR #2951, opened 2018-03-23
PMP: Mesh and shape smoothing

kkatrio:
## Summary of Changes

Motivated by issue #1317 
Affected package: PMP
Small Feature :  [Features/Small_Features/Smoothing](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Smoothing) 
  - pre-approved: 2018/06/07
  - canceling approval due to API changes: 2019/06/04
 - pre-approved 2019/06/24


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/mesh_smoothing.h
```diff
@@ -0,0 +1,346 @@
...
+*    to the vertices of `pmesh`. Instance of a class model of `ReadWritePropertyMap`.
+*  \cgalParamEnd
+*  \cgalParamBegin{number_of_iterations} the number of iterations for the
+*    sequence of the smoothing iterations performed.
+*  \cgalParamEnd
+*  \cgalParamBegin{vertex_is_constrained_map} a property map containing the
+*    constrained-or-not status of each vertex of `pmesh`. A constrained vertex
+*    cannot be modified at all during smoothing.
+*  \cgalParamEnd
+*  \cgalParamBegin{gradient_descent_precision} precision which is achieved by minimizing the energy of each
```
janetournois: This description is a bit too long. In this package, we usually give the full description of the parameters in the "Named Parameters" page (including the default value), and a short description in the function ref manual.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,408 @@
...
+  std::cout<<std::endl;
+  #endif
+}
+
+/*!
+* \ingroup PMP_meshing_grp
+* smooths the overall shape of the mesh by using the mean curvature flow.
+* The effect depends on the curvature of each area and on a time step which
+* represents the amount by which vertices are allowed to move.
+* The result is a conformal map of the initial surface and converges to a sphere.
```
janetournois: "is a conformal map" should be "conformally maps"
Imho, the result is a mesh

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,408 @@
...
+*         and no `face_index_map` is given
+*         as a named parameter, then the internal one should be initialized.
+* @tparam FaceRange range of `boost::graph_traits<PolygonMesh>::%face_descriptor`,
+*         model of `Range`. Its iterator type is `ForwardIterator`.
+* @tparam NamedParameters a sequence of \ref pmp_namedparameters "Named Parameters"
+*
+* @param pmesh a polygon mesh with triangulated surface patches to be smoothed.
+* @param faces the range of triangular faces defining one or several surface patches to be smoothed.
+* @param time a time step that corresponds to the speed by which the surface is smoothed.
+*        A larger time step results in faster convergence but details may be distorted to a greater effect
```
janetournois: "to a greater" should be "to get a greater", or "to have a greater". Right?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,408 @@
...
+*        compared to more iterations with a smaller step. Typical values scale at the region (1e-6, 1).
```
janetournois: "at the region" should be "in the interval", imho (and if `1` if included, the interval should be `[1e-6, 1]`)

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
```diff
@@ -122,6 +123,44 @@ Isotropic remeshing. (a) Triangulated input surface mesh.
...
+shape smoothing is designed to be \e intrinsic, depending as little as possible on the discretization
+and smoothing the shape alone without optimizing the shape of the triangles.
+
+- For mesh smoothing: `CGAL::Polygon_mesh_processing::smooth_angles()` moves vertices so that angles between incident edges equalize,
+and `CGAL::Polygon_mesh_processing::smooth_areas()` moves vertices so that areas of adjacent triangles tend to equalize.
+Border vertices are considered constrained and do not move at any step of the procedure. 
+No new vertex is inserted, although degenerate faces are removed as a first step before smoothing. 
+Angle and area smoothing algorithms are based on Surazhsky and Gotsman \cgalCite{cgal:sg-hqct-04}. 
+Since area smoothing considers only triangle areas as a smoothing criterion, it may result in long and skinny triangles,
+especially for meshes with highly irregular connectivity. However, area smoothing is guaranteed to improve the spatial
```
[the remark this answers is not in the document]
kkatrio: regular = close the the average degree
http://imr.sandia.gov/papers/imr11/surazhsky.pdf
I took the confusing sentence out.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
```diff
@@ -122,6 +123,44 @@ Isotropic remeshing. (a) Triangulated input surface mesh.
...
+(d) Applied twice a sequence of 5 iterations of smoothing based on angles followed by 1 iteration of smoothing based on areas.
+\cgalFigureEnd
+
+- For shape smoothing: `CGAL::Polygon_mesh_processing::smooth_along_curvature_flow()`
+moves vertices incrementally towards a weighted barycenter along the mean curvature flow.
+The curvature flow algorithm for shape smoothing is based on Desbrun et al \cgalCite{cgal:dmsb-ifamdcf-99} and on Kazhdan et al \cgalCite{kazhdan2012can}.
+The algorithm uses the mean curvature flow to calculate the translation of vertices along the surface normal with a speed equal
+to the mean curvature of the area that is being smoothed. This means that vertices on sharp corners slide faster.
+If the region around a vertex is flat, this vertex does not move (zero curvature). To avoid the formation of undesirable neck
+pinches the algorithm slows down the evolution in cylindrical regions. The smoothed shape converges to a sphere while staying conformally equivalent to its original shape.
```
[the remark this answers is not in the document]
kkatrio: cylindrical surface areas that form singularities
https://www.cs.jhu.edu/~misha/MyPapers/SGP12.pdf

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
```diff
@@ -122,6 +123,44 @@ Isotropic remeshing. (a) Triangulated input surface mesh.
...
+moves vertices incrementally towards a weighted barycenter along the mean curvature flow.
+The curvature flow algorithm for shape smoothing is based on Desbrun et al \cgalCite{cgal:dmsb-ifamdcf-99} and on Kazhdan et al \cgalCite{kazhdan2012can}.
+The algorithm uses the mean curvature flow to calculate the translation of vertices along the surface normal with a speed equal
+to the mean curvature of the area that is being smoothed. This means that vertices on sharp corners slide faster.
+If the region around a vertex is flat, this vertex does not move (zero curvature). To avoid the formation of undesirable neck
+pinches the algorithm slows down the evolution in cylindrical regions. The smoothed shape converges to a sphere while staying conformally equivalent to its original shape. 
+(A simple example can be found in `Polygon_mesh_processing/shape_smoothing_example.cpp`).
+
+\cgalFigureBegin{shape_smoothing, shape_smoothing.png}
+(a) Shape smoothing on a dinosaur model using the normalized mean curvature flow. The resulted smoothed shape is colored white.
```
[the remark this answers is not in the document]
kkatrio: there is no time step in the explicit scheme, it's normalized. Should there be more emphasis on that?

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
```diff
@@ -122,6 +123,44 @@ Isotropic remeshing. (a) Triangulated input surface mesh.
...
+(b) Smoothing the shape of the devil using the mean curvature flow with a time step equal to 0.5. The result is a conformal map to a sphere.
```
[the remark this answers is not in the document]
kkatrio: By map I mean transformation. How could I make this less confusing?

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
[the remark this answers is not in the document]
janetournois: you should add a comment about that. It is hard to guess...

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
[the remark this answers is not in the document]
janetournois: ok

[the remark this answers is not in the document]
janetournois: ok

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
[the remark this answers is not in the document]
janetournois: Ok. My bad.
However for uniformity, both (a) and (b) should start with "Shape smoothing", or "Smoothing the shape", but not different titles (it's cosmetic but more comfortable to read)

Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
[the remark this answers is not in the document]
janetournois: The result is not a transformation, it is a transformed mesh. Right?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/smoothing_evaluation.h
```diff
@@ -0,0 +1,149 @@
...
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0+
+//
+//
+// Author(s)     : Konstantinos Katrioplas (konst.katrioplas@gmail.com)
+
+#ifndef SMOOTHING_EVALUATION_H
```
sloriot: should be `CGAL_POLYGON_MESH_PROCESSING_....` for all new files.

Polygon_mesh_processing/examples/Polygon_mesh_processing/mesh_smoothing_example.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
+#include <fstream>
+
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Surface_mesh.h>
+
+#include <CGAL/Polygon_mesh_processing/mesh_smoothing.h>
+#include <CGAL/Polygon_mesh_processing/shape_smoothing.h>
```
janetournois: this include can be removed

Polygon_mesh_processing/examples/Polygon_mesh_processing/shape_smoothing_example.cpp
```diff
@@ -0,0 +1,38 @@
+#include <iostream>
+#include <fstream>
+
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Surface_mesh.h>
+
+#include <CGAL/Polygon_mesh_processing/mesh_smoothing.h>
```
janetournois: this include can be removed

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/curvature_flow_explicit_impl.h
```diff
@@ -0,0 +1,250 @@
...
+  void init_smoothing(const FaceRange& face_range)
+  {
+    check_vertex_range(face_range);
+    BOOST_FOREACH(face_descriptor f, face_range)
+      input_triangles_.push_back(triangle(f));
+  }
+
+  std::size_t remove_degenerate_faces()
+  {
+    std::size_t nb_removed_faces = 0;
```
janetournois: unused variable

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,640 @@
...
+namespace internal {
+
+template<typename PolygonMesh, typename VertexPointMap, typename VertexConstraintMap, typename GeomTraits>
+class Compatible_remesher
+{
+  typedef typename boost::graph_traits<PolygonMesh>::halfedge_descriptor halfedge_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_descriptor vertex_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::face_descriptor face_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::edge_descriptor edge_descriptor;
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_iterator vertex_iterator;
```
janetournois: this typedef is duplicate

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,640 @@
...
+  typedef typename GeomTraits::Point_3 Point;
+  typedef typename GeomTraits::Vector_3 Vector;
+  typedef typename GeomTraits::Segment_3 Segment;
+  typedef typename GeomTraits::Triangle_3 Triangle;
+
+  typedef std::vector<Triangle> Triangle_list;
+  typedef std::pair<halfedge_descriptor, halfedge_descriptor> he_pair;
```
janetournois: please capitalize the "H" : `He_pair` to easily recognize a type

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,640 @@
...
+      if(does_it_impove(vp.first, vp.second))
+      {
+        moved_points++;
+        put(vpmap_, vp.first, vp.second);
+      }
+    }
+
+#ifdef CGAL_PMP_SMOOTHING_DEBUG
+    std::cout<<"moved: "<< moved_points <<" points based on angle."<<std::endl;
+    std::cout<<"not imporved min angle: "<< vrange_.size() - moved_points <<" times."<<std::endl;
```
janetournois: typo : should be "improved"

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,638 @@
...
+      if(!is_border(v, mesh_) && !is_constrained(v))
+      {
+        Point p_query = get(vpmap_, v);
+        Point projected = tree_ptr_->closest_point(p_query);
+        put(vpmap_, v, projected);
+      }
+    }
+  }
+
+private:
```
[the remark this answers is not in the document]
janetournois: this remains to do

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,640 @@
...
+        return false;
+    }
+    return true;
+  }
+
+  double get_angle(const Vector& e1, const Vector& e2)
+  {
+    //double rad_to_deg = 180. / CGAL_PI;
+    double cos_angle = (e1 * e2)
+     / std::sqrt(e1.squared_length() * e2.squared_length());
```
janetournois: use `CGAL::sqrt` or `CGAL::approximate_sqrt`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/smoothing_evaluation.h
```diff
@@ -0,0 +1,149 @@
...
+      Point a = get(vpmap_, source(hi, mesh_));
+      Point b = get(vpmap_, target(hi, mesh_));
+      Point c = get(vpmap_, target(next(hi, mesh_), mesh_));
+      Vector ba(b, a);
+      Vector bc(b, c);
+      double cos_angle = (ba * bc)
+        / std::sqrt(ba.squared_length() * bc.squared_length());
+      angles_.push_back(std::acos(cos_angle) * rad_to_deg);
+    }
+    std::cout<<"angles_ size= "<<angles_.size()<<std::endl;
```
janetournois: here and everywhere else, the verbose code should be protected by a macro, for example
```
#ifdef CGAL_PMP_SMOOTHING_VERBOSE
 std::cout << "blabla" << std::endl;
#endif
```

janetournois: I just saw this macro is already being used elsewhere in this PR, so it is the right one!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,413 @@
...
+*    cannot be modified at all during smoothing.
+*  \cgalParamEnd
+*  \cgalParamBegin{number_of_iterations} the number of iterations for the
+*    sequence of the smoothing iterations performed. Each iteration is performed
+*    with the given time step.
+*  \cgalParamEnd
+*  \cgalParamBegin{use_explicit_scheme} a normalized explicit scheme for smoothing along the curvature flow.
+*    It is not dependent on the time step parameter. Each vertex is moved sequentially without solving
+*    a linear system. It can be useful for subtle noise removal and the result
+*    does not converge to a sphere.
```
janetournois: I suggest "and to avoid the result to converge to a sphere"

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,413 @@
...
+template<typename PolygonMesh>
+void smooth_along_curvature_flow(PolygonMesh& pmesh, const double& time)
+{
+  smooth_along_curvature_flow(faces(pmesh), pmesh, time,
+                              parameters::all_default());
+}
+
+// API for Polyhedron demo plugin
+template<typename PolygonMesh, typename FaceRange, typename NamedParameters>
+void solve_mcf(const FaceRange& faces, PolygonMesh& mesh, const double& time,
```
janetournois: this function should not be in the namespace `CGAL::Polygon_mesh_processing`
It should at least be moved to an `internal` nested namespace

janetournois:
I just updated my review. @kkatrio can you please make the fixes?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/constraints_map.h
```diff
@@ -0,0 +1,100 @@
...
+public:
+  Constrained_vertices_map() : constrained_data(new std::set<Descriptor>) {}
+
+  friend bool get(const Constrained_vertices_map& map, const key_type& d)
+  {
+    typename std::set<Descriptor>::iterator it = map.constrained_data->find(d);
+    return it != map.constrained_data->end() ? true : false;
+  }
+
+  friend void put(Constrained_vertices_map& map, const key_type& d)
```
sloriot: I doubt this will work, you are missing the value type. Anyway this is not needed as you only want a read-only map + it is a duplicate of `CGAL::Boolean_property_map` so the whole class can be removed.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/mesh_smoothing.h
```diff
@@ -0,0 +1,346 @@
...
+  remesher.remove_degenerate_faces();
+
+  #ifdef CGAL_PMP_SMOOTHING_VERBOSE
+  t.stop();
+  std::cout << " done ("<< t.time() <<" sec)." << std::endl;
+  std::cout << "Initializing..." << std::endl;
+  t.reset(); t.start();
+  #endif
+
+  remesher.init_smoothing(faces);
```
sloriot: it seems strange to have remesher instead of smoother.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_descriptor vertex_descriptor;
+  typedef typename boost::lookup_named_param_def <
+      internal_np::vertex_is_constrained_t,
+      NamedParameters,
+      internal::Constrained_vertices_map<vertex_descriptor>
+    > ::type VCMap;
+  VCMap vcmap = choose_param(get_param(np, internal_np::vertex_is_constrained),
+                             internal::Constrained_vertices_map<vertex_descriptor>());
+
```
sloriot: same remark

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
sloriot: ping @kkatrio

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+*  \cgalParamEnd
+*  \cgalParamBegin{use_explicit_scheme} a normalized explicit scheme for smoothing along the curvature flow.
+*    It is not dependent on the time step parameter. Each vertex is moved sequentially without solving
+*    a linear system. It can be useful for subtle noise removal and to avoid the result
+*    to converge to a sphere.
+*  \cgalParamEnd
+* \cgalNamedParamsEnd
+*/
+template<typename PolygonMesh, typename FaceRange, typename NamedParameters>
+void smooth_along_curvature_flow(const FaceRange& faces, PolygonMesh& pmesh, const double& time,
```
sloriot: no reference on a double number

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+  }
+  else
+  {
+    // implicit scheme
+#if defined(CGAL_EIGEN3_ENABLED)
+  #if EIGEN_VERSION_AT_LEAST(3,2,0)
+    typedef typename Eigen::SparseMatrix<double> Eigen_sparse_matrix;
+    typedef typename Eigen::BiCGSTAB<Eigen_sparse_matrix, Eigen::IncompleteLUT<double> > Eigen_solver;
+    typedef CGAL::Eigen_solver_traits<Eigen_solver> Default_solver;
+  #else
```
sloriot: The solver must be a template parameter! You can use a new named parameter for if with Eigen as default and if eigen is not available a static assertion.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+void solve_mcf(const FaceRange& faces, PolygonMesh& mesh, const double& time,
+               std::vector<CGAL::Triple<int, int, double> >& stiffness, bool compute_stiffness,
+               const NamedParameters& np)
+{
+  using boost::choose_param;
+  using boost::get_param;
+
+  //typedef typename GetGeomTraits<PolygonMesh>::type GeomTraits;
+  //typedef typename boost::property_map<PolygonMesh, CGAL::vertex_point_t>::type VertexPointMap;
+  //VertexPointMap vpmap = get(CGAL::vertex_point, mesh);
```
sloriot: comments to be removed

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+
+  typedef typename boost::graph_traits<PolygonMesh>::vertex_descriptor vertex_descriptor;
+  typedef typename boost::lookup_named_param_def <
+      internal_np::vertex_is_constrained_t,
+      NamedParameters,
+      internal::Constrained_vertices_map<vertex_descriptor>
+    > ::type VCMap;
+  VCMap vcmap = choose_param(get_param(np, internal_np::vertex_is_constrained),
+                             internal::Constrained_vertices_map<vertex_descriptor>());
+
```
sloriot: same thing here

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_smoothing.h
```diff
@@ -0,0 +1,415 @@
...
+  // nb_iterations
+  std::size_t nb_iterations = choose_param(get_param(np, internal_np::number_of_iterations), 1);
+
+#if defined(CGAL_EIGEN3_ENABLED)
+  #if EIGEN_VERSION_AT_LEAST(3,2,0)
```
sloriot: same issue here. This is an internal function and the solver must be passed as parameter to the function.

sloriot:
TODO for @janetournois  or me before the integration, test the demo plugin.

lrineau:
There is still a conflict with `master`.

Please, do not use rebase to fix it, but a simple merge of `master` into your branch.

BGL/include/CGAL/boost/graph/parameters_interface.h
```diff
@@ -64,6 +64,8 @@ CGAL_add_named_parameter(nb_points_per_distance_unit_t, nb_points_per_distance_u
 CGAL_add_named_parameter(outward_orientation_t, outward_orientation, outward_orientation)
 CGAL_add_named_parameter(overlap_test_t, overlap_test, do_overlap_test_of_bounded_sides)
 CGAL_add_named_parameter(preserve_genus_t, preserve_genus, preserve_genus)
+CGAL_add_named_parameter(gradient_descent_precision_t, gradient_descent_precision, gradient_descent_precision)
```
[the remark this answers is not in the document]
kkatrio: Is it not better without the "set_" ? For example, we don't use "set_number_of_iterations".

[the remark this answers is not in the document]
sloriot: @janetournois what do you think? We should have a consistent scheme.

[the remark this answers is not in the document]
afabri: Where else do we have `set_` ??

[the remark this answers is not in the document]
sloriot: indeed, we only have use_ and do_ so forget about it.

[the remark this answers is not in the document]
janetournois: it's also my opinion

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/curvature_flow_impl.h
```diff
@@ -0,0 +1,303 @@
...
+  VertexPointMap& vpmap_;
+  VertexConstraintMap& vcmap_;
+  IndexMap vimap_;
+
+  // linear system data
+  std::vector<double> diagonal_; // index of vector -> index of vimap_
+  std::vector<bool> constrained_flags_;
+
+  std::set<face_descriptor> frange_;
+  std::set<vertex_descriptor> vrange_;
```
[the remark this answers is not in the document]
kkatrio: hang on, I am confused: do you mean all vertices of the mesh?
I want only the vertices in that face range.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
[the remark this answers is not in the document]
kkatrio: Yes I did!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/curvature_flow_impl.h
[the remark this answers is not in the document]
sloriot: no only the vertices of the selection

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/mesh_smoothing.h
```diff
@@ -0,0 +1,346 @@
...
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0+
+//
+//
+// Author(s)     : Konstantinos Katrioplas (konst.katrioplas@gmail.com)
+
+#ifndef CGAL_POLYGON_MESH_PROCESSING_MESH_SMOOTHING_H
+#define CGAL_POLYGON_MESH_PROCESSING_MESH_SMOOTHING_H
```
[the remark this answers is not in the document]
kkatrio: mesh_smooth.h and shape_smooth.h ?
Or are you talking about merging to one?
And who introduced the -ing in PMP in the first place anyway?? :)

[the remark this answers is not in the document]
janetournois: Then it would become `smooth_mesh()` and `smooth_shape()`?
It's more consistent with the rest of the package indeed.
@sloriot should we go for that?

I did introduce `isotropic_remeshing()` but it's a mistake. And the header is `remesh.h` anyway...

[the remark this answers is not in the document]
janetournois: Let's go for
- `PMP::smooth_mesh()` in `smooth_mesh.h`, and
- `PMP::smooth_shape()` in `smooth_shape.h`
then!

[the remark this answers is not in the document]
kkatrio: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Smoothing#Summary_of_API_changes

I changed the header names.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/curvature_flow_impl.h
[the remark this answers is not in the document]
kkatrio: AFAIU, we would use std::sort and std::unique on a std::vector so as to avoid the double loop in order to take out duplicates, but can we avoid the double loop when we deal with a subset of the faces of the mesh?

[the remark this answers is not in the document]
sloriot: when you collect the vertices per faces, collect them all, duplicates included. Then sort, call unique and remove the extra copies from the range. Then you're done. No idea if it does worth it in your case.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/smoothing_evaluation.h
```diff
@@ -0,0 +1,155 @@
...
+// Author(s)     : Konstantinos Katrioplas (konst.katrioplas@gmail.com)
+
+#ifndef CGAL_POLYGON_MESH_PROCESSING_INTERNAL_SMOOTHING_EVALUATION_H
+#define CGAL_POLYGON_MESH_PROCESSING_INTERNAL_SMOOTHING_EVALUATION_H
+
+#include <CGAL/license/Polygon_mesh_processing/meshing_hole_filling.h>
+
+#include <vector>
+#include <set>
+#include <fstream>
```
sloriot: In general fstream is not needed but in examples directly

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/smoothing_evaluation.h
```diff
@@ -0,0 +1,155 @@
...
+#include <string>
```
sloriot: same for string

Polygon_mesh_processing/test/Polygon_mesh_processing/test_mesh_smoothing.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <boost/graph/graph_traits.hpp>
+#include <fstream>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
+typedef Kernel::Point_3 Point;
+typedef CGAL::Surface_mesh<Point> SurfaceMesh;
+typedef CGAL::Polyhedron_3<Kernel> Polyhedron;
+
+template<typename Mesh>
+struct Constraints_pmap
```
sloriot: Boolean_property_map?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/smooth_shape.h
```diff
@@ -0,0 +1,415 @@
...
+/*!
+* \ingroup PMP_meshing_grp
+* smooths the overall shape of the mesh by using the mean curvature flow.
+* The effect depends on the curvature of each area and on a time step which
+* represents the amount by which vertices are allowed to move.
+* The result conformally maps the initial surface to a sphere.
+*
+*
+* @tparam PolygonMesh model of `MutableFaceGraph`.
+*         If `PolygonMesh` has an internal property map for `CGAL::face_index_t`,
```
sloriot: if you don't use the face index map, you can remove this sentence

sloriot: I ask the question because I don't see it in the named parameters

kkatrio: was forgotten here when I switched to using the Constant_property_map, thanks!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h
```diff
@@ -0,0 +1,629 @@
...
+    Point_ref equidistant_p1 = get(vpmap_, target(incd_edges.first, mesh_));
+    Point_ref equidistant_p2 = get(vpmap_, source(incd_edges.second, mesh_));
+    CGAL_assertion(target(incd_edges.second, mesh_) == source(incd_edges.first, mesh_));
+
+    Vector edge1(pt, equidistant_p1);
+    Vector edge2(pt, equidistant_p2);
+    Vector vec_main_he(pt, ps);
+
+    // check degenerate cases
+    double precision = 1e-3;
```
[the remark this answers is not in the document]
janetournois: @kkatrio is this really needed?

lrineau:
@janetournois @sloriot you made reviewed of the code, and it is not yet approved (the code is not approved). What is more, there is a TODO at the top (https://github.com/CGAL/cgal/pull/2951#issue-177042660).

sloriot:
I'll compile the plugin and test it

Polyhedron/demo/Polyhedron/Plugins/PMP/Smoothing_plugin.cpp
```diff
@@ -0,0 +1,360 @@
...
+    const double time_step = ui_widget.time_step_spinBox->value();
+    const unsigned int nb_iter = ui_widget.curvature_iter_spinBox->value();
+
+    const bool use_constrained_vertex_map = ui_widget.border_button->isChecked()
+                                            && !CGAL::is_closed(pmesh);
+
+    QApplication::setOverrideCursor(Qt::WaitCursor);
+
+    // explicit scheme
+    if(ui_widget.explicit_checkBox->isChecked())
```
sloriot: I wonder why in case we do not use the explicit scheme we are using some internal functions with some caching and not simply pass false to `use_explicit_scheme()`.
@janetournois any idea?

Polyhedron/demo/Polyhedron/Plugins/PMP/Smoothing_plugin.cpp
```diff
@@ -0,0 +1,360 @@
...
+          // If we changed item or if the stiffness cache is cleared (by the user hitting the button)
+          if(use_constrained_vertex_map){
+            internal::solve_mcf(faces(pmesh), pmesh, time_step, stiffness, true,
+            parameters::vertex_is_constrained_map(get_border_constrained_map(pmesh)) );
+          }
+          else{
+            internal::solve_mcf(faces(pmesh), pmesh, time_step, stiffness, true,
+            parameters::all_default());
+          }
+          last_index_id = index_id;
```
sloriot: this mechanism to use some caching for an item is buggy as it will not take into account the reloading of an item for example

sloriot:
About the plugin:
- [ ] It is written for Polyhedron only. @maxGimeno could you help here?
- [ ] Area smoothing does not seem to do anything on the mesh (it is taking time to compute something but I cannot see any difference)
- [ ] Not using the normalized scheme and constraining border vertices will make the mesh vanish.

sloriot:
Thanks @maxGimeno

sloriot:
Also on `anchor.off` area smoothing is doing something. I only tried on `elephant.off`.

Polyhedron/demo/Polyhedron/Plugins/PMP/Smoothing_plugin.cpp
janetournois: I'm trying to figure this out

janetournois: in the demo, this makes possible to compute the stiffness matrix, and then use it without recomputing it, on the cached item you mention below.
I think we can remove that feature. It may be faster in some cases for demo purposes, but it introduces confusion in the plugin.
I will do it

janetournois:
`smooth_angles()` crashes on some data.
I will try to find why

lrineau:
The code freeze is in four days, and this PR has never been tested in `integration`, and it does not seem ready to be tested either.

Should we postpone it to CGAL 4.14, or do we really it in CGAL 4.13?

janetournois:
I'll try to fix the crash asap. If it's not fixed early enough, we'll postpone it to 4.14

kkatrio:
Can I help? On which data does angle smoothing crash?

janetournois:
in the demo, open anchor.off, and set "10" for the nb of iterations of angle smoothing : it crashes at the second iteration, inside the projection step

kkatrio:
_BGL/test/BGL/test_cgal_bgl_named_params.cpp_ is a bit messy after subsequent merges of other PRs. For example _get_cost_policy_ is not in the main function, and _To-be-documented named parameters_ have the same id as others. Please take a look. Smoothing's np are gradient_descent (id 47) and use_explicit_scheme (id 48).

sloriot:
Should be fixed by b66ffa3

janetournois:
The result of 10 iterations of angle smoothing is not very nice, but I guess it's what it should be.

However, there is a (new?) issue with angle smoothing in the demo:
- open anchor.off
- open the selection plugin, and/or create a selection item,
- select sharp edges (60 degrees),
- run isotropic remeshing with all default parameters on the selection,
- launch the smoothing widget,
- run 10 iterations of angle smoothing with re-projection.
Here is the result 
tr_holes

kkatrio:
>   However, there is a (new?) issue with angle smoothing in the demo:
>     open anchor.off
>     open the selection plugin, and/or create a selection item,
>     select sharp edges (60 degrees),
>     run isotropic remeshing with all default parameters on the selection,
>     launch the smoothing widget,
>     run 10 iterations of angle smoothing with re-projection.
>     Here is the result

@janetournois Hi Jane! I just saw this, why didn't you bring it up while I was working there? I could have fixed it.

jasjuang:
Looking forward to see this PR finished and merged to master!

maxGimeno:
There are conflicts with master.

lrineau:
Travis errors:
```
In file included from /home/travis/build/CGAL/cgal/Polygon_mesh_processing/examples/Polygon_mesh_processing/mesh_smoothing_example.cpp:4:
In file included from /usr/local/include/CGAL/Polygon_mesh_processing/smooth_mesh.h:28:
/usr/local/include/CGAL/Polygon_mesh_processing/internal/Smoothing/mesh_smoothing_impl.h:46:10: fatal error: 
      'ceres/ceres.h' file not found
#include "ceres/ceres.h"
         ^~~~~~~~~~~~~~~
1 error generated.
```
https://travis-ci.org/CGAL/cgal/jobs/547697865

```
[ 96%] Building CXX object CMakeFiles/test_cgal_bgl_named_params.dir/test_cgal_bgl_named_params.cpp.o
/usr/local/clang-7.0.0/bin/clang++  -DCGAL_HEADER_ONLY=1 -DCGAL_USE_OPENMESH -DNOMINMAX -D_USE_MATH_DEFINES -isystem /usr/local/include -isystem /usr/include/x86_64-linux-gnu  -DCGAL_NDEBUG -ftemplate-backtrace-limit=0   -std=gnu++14 -o CMakeFiles/test_cgal_bgl_named_params.dir/test_cgal_bgl_named_params.cpp.o -c /home/travis/build/CGAL/cgal/BGL/test/BGL/test_cgal_bgl_named_params.cpp
/home/travis/build/CGAL/cgal/BGL/test/BGL/test_cgal_bgl_named_params.cpp:266:27: error: 
      no member named 'use_angle_smoothing_t' in
[...]
```
https://travis-ci.org/CGAL/cgal/jobs/547697840

MaelRL:
Sure, I can see the red cross. But Travis status is pretty irrelevant when the whole PR is WIP (and unfortunately, PRs can't be switched back to draft). When it's ready to be tested, it'll be flagged as such.

lrineau:
Maxime and I pointing out the errors in the testsuite or in Travis is just to help developers, because not all developers are attentive of those errors. Do not see that as a criticism of your work.

MaelRL:
Should be good to go modulo some detail about `-pthread`. For some reason on my machine, the Ceres library requires `-pthread` (somehow only in `/examples` and not `/test` nor `/demo`...). I tried to write in [FindCeres.cmake](https://github.com/CGAL/cgal/pull/2951/files#diff-b5d1fa9d1891bde5e6c5a31c7141f688R6) something similar to what's done in [FindTBB.cmake](https://github.com/CGAL/cgal/blob/master/Installation/cmake/modules/FindTBB.cmake#L64) which seems to have the same issue.

However, it doesn't seem to work: `try_CERES_with_pthread(CERES_with_pthread -pthread)` also fails to compile (but it's fine when I put `-pthread` manually).

@lrineau  @maxGimeno  Could you please help ?

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 21, 2019, 8:28 AM GMT+2](https://github.com/CGAL/cgal/pull/2951#issuecomment-504307582 "2019-06-21T06:28:57Z - Replied by Github Reply Comments"):
> Should be good to go modulo some detail about `-pthread`. For some reason on my machine, the Ceres library requires `-pthread` (somehow only in `/examples` and not `/test` nor `/demo`...). I tried to write in [FindCeres.cmake](https://github.com/CGAL/cgal/pull/2951/files#diff-b5d1fa9d1891bde5e6c5a31c7141f688R6) something similar to what's done in [FindTBB.cmake](https://github.com/CGAL/cgal/blob/master/Installation/cmake/modules/FindTBB.cmake#L64) which seems to have the same issue.
> 
> However, it doesn't seem to work: `try_CERES_with_pthread(CERES_with_pthread -pthread)` also fails to compile (but it's fine when I put `-pthread` manually).
> 
> @lrineau  @maxGimeno  Could you please help ?

On my Fedora machine, the package `ceres-solver-devel` comes with a file `/usr/lib64/cmake/Ceres/CeresConfig.cmake`. That means that there is no need for a `FindCeres.cmake` and `find_package` can find Ceres with the MODULE mode, just like CGAL users find `CGALConfig.cmake`. That file `CeresConfig.cmake` handles all the difficulty (or at least is supposed to).

MaelRL:
I did not install it through dnf, I compiled from source and installed it manually. Everything went fine and it did create the `CeresConfig.cmake`, and I could find the lib through find_package. However, when compiling examples I had gotten:

```
/usr/bin/ld: /usr/local/lib64/libceres.a(thread_pool.cc.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/bin/ld: //usr/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
```

So since I'm out of my element, I just searched for `pthread` in CGAL and found `FindTBB.cmake` and naively decided I would just write my own FindCeres.cmake and do the same try-to-compile-with-and-without-pthread.

lrineau:
How is defined the imported target `ceres` in `/usr/local/lib/cmake/Ceres/CeresTargets-release.cmake`? On my machine, it is:
```cmake
# Import target "ceres" for configuration "Release"
set_property(TARGET ceres APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(ceres PROPERTIES
  IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
  IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE "/usr/lib64/libglog.so;/usr/lib64/libopenblas.so;/usr/lib64/libopenblas.so;gomp;-lpthread"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib64/libceres.a"
  )
```
and one can see the `-lpthread`.

MaelRL:
Missing the line `IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE`:
```cmake
# Import target "ceres" for configuration "Release"
set_property(TARGET ceres APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(ceres PROPERTIES
  IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "CXX"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib64/libceres.a"
  )
```

lrineau:
... IRL we found out that @MaelRL installed Ceres 1.14 and I had Ceres 1.13. Contrary to Ceres 1.13, the file Ceres 1.14 `CeresTargets.cmake` itself defines `IMPORTED_LINK_INTERFACE_LIBRARIES`.

Polygon_mesh_processing/examples/Polygon_mesh_processing/CMakeLists.txt
```diff
@@ -119,5 +127,7 @@ create_single_source_cgal_program( "triangulate_faces_example_OM.cpp")
 target_link_libraries( triangulate_faces_example_OM PRIVATE ${OPENMESH_LIBRARIES} )
 endif(OpenMesh_FOUND)
 
-
+if(TARGET ceres)
+  target_link_libraries( mesh_smoothing_example PRIVATE ceres)
+endif(TARGET ceres)
```
lrineau: I would even:
  - put down the `find_package(Ceres QUIET)`, here, just before its use,
  - remove the `if(CERES_FOUND)` block, and then, use `target_compile_definitions`, as below:

```suggestion
  target_compile_definitions( mesh_smoothing_example PRIVATE CGAL_PMP_USE_CERES_SOLVER)
endif(TARGET ceres)
```

MaelRL:
Should be ready to test, now. Thanks for the help!

@maxGimeno, a third of the code requires [Ceres](http://ceres-solver.org/) and so it would be nice if it's installed on a few test platforms please.

maxGimeno:
I installed Ceres on Friedrich and Blake, and added the PR in the testsuite

maxGimeno:
The new pmp paramaters hae conflicts with the ones in Locate. If you could check and change the numbers associated in the params tests so both PRs are compatible, I'd be grateful.

MaelRL:
done in https://github.com/CGAL/cgal/pull/3885 (it only had one new named parameter).

maxGimeno:
Yes, perfect ! Thank you !

MaelRL:
The `Debian-stable` platform is on Friedrich but gives a warning that Ceres is not available ("Area-based smoothing requires the Ceres Library, which is not available."), is this normal?

https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-69/Polygon_mesh_processing/TestReport_gimeno_Debian-Stable.gz

MaelRL:
There seems to be some installation issues on Blake too (https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-69/Polygon_mesh_processing/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz).

maxGimeno:
The only platforms where Ceres is expected to be found are Ubuntu-gcc-7 and the winsdows on Blake. I didn't go through all Docker platforms to add it.

Installation/CHANGES.md
```diff
@@ -6,6 +6,12 @@ Release 5.0
 
 Release date: September 2019
 
+### 2D and 3D Linear Geometry Kernel
+ - Added `ComputeApproximateAngle_3` in the 2D/3D Kernel concept to compute
+   the approximate dihedral angle between 2 vectors. Corresponding functors
+   in the model (`Compute_approximate_angle_3`) and free function (`approximate_angle`)
+   are also added.
+
```
lrineau: I do not understand this changes of the changelog. Does this PR really add `ComputeApproximateAngle_3`?!

sloriot: This comes from here:
https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/approximate_angle

lrineau: So it came from https://github.com/CGAL/cgal/pull/3468 and not from this PR. Why is it in the changes, then?

MaelRL: PMP and Kernel were inverted initially in the PR so in the merge conflict resolution I kept that order, which is more natural than having kernel changes at the bottom. The removal corresponding to this addition is actually 3 lines below. 

It doesn't work so well due to other changes being added to kernel with kernel at a different stop, so now it's a bit messy, I'll clean it.

maxGimeno:
There are [warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-74/Polygon_mesh_processing/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz) on windows IN ceres about eigen :/

MaelRL:
Naively, it seems to be the same error as in this issue: https://github.com/tbennun/ceres-windows/issues/9. However, looking at https://github.com/ceres-solver/ceres-solver/blob/0d3a84fce553c9f7aab331f0895fa7b1856ef5ee/include/ceres/gradient_checker.h, this fix has been integrated in the main code base. Could you please check that the version you have installed is recent enough (the fix is from 2 years ago, so I would imagine so)?

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jul 3, 2019, 10:16 AM GMT+2](https://github.com/CGAL/cgal/pull/2951#issuecomment-507990689 "2019-07-03T08:16:35Z - Replied by Github Reply Comments"):
> Naively, it seems to be the same error as in this issue: https://github.com/tbennun/ceres-windows/issues/9.

Actually no: that issue https://github.com/tbennun/ceres-windows/issues/9 warnings on Ceres classes not having the dll-interface declaration. Now the warnings we have are one STL or Eigen classes not having the dll-interface. I think the MSVC compiler is dump, here because it warns about header-only dependencies (or templates defined in headers) not having dll-interface. That warning is safe to ignore.

MaelRL:
I see, thank you. So I just add [this type](https://github.com/CGAL/cgal/blob/7180fe5bb3da062687af2a378a9d770248241cff/Mesh_3/include/CGAL/perturb_mesh_3.h#L49) of shenanigans? Or is the policy to now regroup in `disable_warnings.h`?

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jul 3, 2019, 10:45 AM GMT+2](https://github.com/CGAL/cgal/pull/2951#issuecomment-508000845 "2019-07-03T08:45:15Z - Replied by Github Reply Comments"):
> I see, thank you. So I just add [this type](https://github.com/CGAL/cgal/blob/7180fe5bb3da062687af2a378a9d770248241cff/Mesh_3/include/CGAL/perturb_mesh_3.h#L49) of shenanigans? Or is the policy to now regroup in `disable_warnings.h`?

`disable_warnings.h` is a huge mistake. It was supposed to limit the warnings of Boost with C++17, but it has been abused. Please do not add anything more in it.

You can add a header like `<CGAL/ceres_support.h>` that includes `ceres.h` surrounded by the needed pragmas.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/ceres_support.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef CGAL_POLYGON_MESH_PROCESSING_INTERNAL_CERES_SUPPORT_H
+#define CGAL_POLYGON_MESH_PROCESSING_INTERNAL_CERES_SUPPORT_H
+
+#include <CGAL/license/Polygon_mesh_processing/meshing_hole_filling.h>
+
+// Ignore warnings on Windows of type
+// " class XXXX needs to have dll-interface to be used by clients of struct YYYY" in an interaction
+// between Ceres and Eigen, which doesn't make much sense as Eigen is header-only.
+
+#if defined(BOOST_MSVC)
```
lrineau: Maybe add 
```c++
#include <boost/config.hpp>
```
to be safe.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Smoothing/ceres_support.h
```diff
@@ -36,15 +36,21 @@
 
 
 // If a user is using minilog instead of glog, the verbosity cannot be controlled via ceres options...
+// Since we really don't want to hear about it, override that value with '0'
 #ifdef MAX_LOG_LEVEL
-#  define MAX_LOG_LEVEL_WAS_PREVIOUSLY_DEFINED
-#else
-  #define MAX_LOG_LEVEL 0
+#  define MAX_LOG_LEVEL_WAS_DEFINED
+#  pragma push_macro("MAX_LOG_LEVEL")
```
lrineau: You need to protect that with `#if defined(BOOST_MSVC)`

MaelRL: minilog, which causes Ceres to set that annoying macro, is multi-platform as far I see so I still need to set it up. These pragma seemed to be supported by gcc and clang, so I didn't protect this code with MSVC checks. Could you please explain what I missed?

I don't really know how else to temporarily store a macro value. Maybe it isn't worth the effort and I should just `#undef #def` and be done with it...

lrineau: You are right. I have verified: [gcc-4.4 and clang-3.0 know `push_macro`](
https://godbolt.org/z/sm8c2U). That is enough for CGAL-5.0. Thanks for you correcting me.

MaelRL:
Can't reproduce windows issues at the same hash (or others), debug or not, different types of differenciation... I've added some optional code to explicitly specify parameters instead of leaving them to be deduced, that might help... If it does not, I'll go ask for help on their mailing list, I suppose.

maxGimeno:

MaelRL:
@maxGimeno Smoothing creates massive logs on some platforms like the MSVC Preview on Blake; maybe Ceres was installed with the wrong Vlog option?

MaelRL:
@maxGimeno Also some third party errors there: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-67/Polygon_mesh_processing/TestReport_monet_x86_64-apple_mojave_clang1001-release-64bits.gz

lrineau:
> @maxGimeno Also some third party errors there: [cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-67/Polygon_mesh_processing/TestReport_monet_x86_64-apple_mojave_clang1001-release-64bits.gz](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-67/Polygon_mesh_processing/TestReport_monet_x86_64-apple_mojave_clang1001-release-64bits.gz)

The error is in `Polygon_mesh_processing/test/Polygon_mesh_processing/CMakeLists.txt`: the target `test_mesh_smoothing` should have the Eigen support. That looks like a bug of PR #4356.

PR #2952, opened 2018-03-23
Mesh_3, T2, T3: fix dependencies by moving files

lrineau:
## Summary of Changes

Before the patch Mesh_3 was dependent on Surface_mesh. And the split between TDS_23 and Triangulation_23 was wrong.

## Release Management

* Affected package(s): T2, T3, Mesh_3


lrineau:

PR #2954, opened 2018-03-23
Spatial searching: rework fuzzy items

### LINKED ISSUE #2729: The manual of the dD Spatial Searching is self-conflicting
## Issue Details

In the user manual (1.2 Range Searching) it says, and I quote:

> The fuzziness of the query object is specified by a parameter ϵ used to define inner and outer approximations of the query object. Points in the interior of the inner approximation are always reported and points that are not in the closure of the outer approximation are never reported.
> Other points may or may not be reported. For exact range searching, the fuzziness parameter ϵ is set to zero.

However, the reference of the Fuzzy_sphere says, and I quote:

> Attention
>     The fuzziness of a Fuzzy_sphere is specified by a parameter ϵ denoting a maximal allowed distance to the boundary of a sphere. If the distance to the boundary is greater than ϵ, points inside the object are always reported and points outside the sphere are never reported. Points whose distance to the boundary is less than or equal to ϵ may or may not be reported. Subsequently, points on the inner and outer spheres may or may not be reported. Specifically when ϵ=0, points on the sphere of radius r may or may not be reported.

So which one is correct?
I assume that the latter is the correct one. However, if this is the case, the former should be rephrased.

MaelRL:
## Summary of Changes

More serious attempt at fixing inconsistencies in the code and documentation of fuzzy items after some bad commits (c0cbdab and 5ed7255) and the revert (#1808).

#### Before:
- Points in the _interior_ of the inner approximation are always returned
- Points in the _interior_ of the outer approximation and not in the interior of the inner approximation may or may not be returned
- Everything else is never returned

When there is no fuzziness, points on the boundary of the object are not returned. Also:
- Half open boxes so the "interior" is a little weird and buggy
- Inconsistencies between the doc and the code (doc was vague before, I made it wrong with commits mentioned above).

#### After:
- Points in the inner approximation are always returned (boundary included)
- Points in the outer approximation and not in the inner approximation may or may not be returned (boundary of the outer approximation included)
- Everything else is never returned

When there is no fuzziness, points on the boundary of the object are returned.

#### Also:
- Fixed a bug in the `Fuzzy_iso_box`
- Minor improvements on speed by caching stuff
- Test fuzzy items exhaustively

cc: @mglisse @efifogel 

## Release Management

* Affected package(s): `Spatial_searching:`
* Issue(s) solved (if any): fix #2729

### DIFF
--- a/Spatial_searching/include/CGAL/Fuzzy_sphere.h
+++ b/Spatial_searching/include/CGAL/Fuzzy_sphere.h
@@ -24,6 +24,7 @@
 
 #include <CGAL/license/Spatial_searching.h>
 
+#include <CGAL/assertions.h>
 #include <CGAL/Kd_tree_rectangle.h>
 #include <CGAL/Search_traits_adapter.h>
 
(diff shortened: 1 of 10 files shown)


lrineau:
This PR is stalled for two weeks. I will probably not be integrated in 4.12.

MaelRL:
I added "not yet approved" because I feel like the PR should be reviewed because it's an unclear basis on which I had done some bad changes (not from a code point of view, but a definition point of view).

@efifogel reported the documentation bug #2729 and @mglisse had said before that Gudhi needed some slightly different fuzzy items to make sure points on the boundaries are included, which this PR adds.

lrineau:
Should Efi and Marc review the PR?

MaelRL:
If they could, please.

mglisse:
The new documentation says surprisingly little about how you handle the boundary (the "summary of changes" in this PR is much clearer). I agree that for most users it is irrelevant, but since this is one of the main topics of this PR...
(in gudhi we currently only use non-fuzzy regions with the kd-tree, IIRC, not that we couldn't use fuzzy but we currently spend our fuzziness budget elsewhere)

MaelRL:
There is nothing general in the user manual because the concept of fuzzy items does not specify how boundaries should be handled. For the documentation of the models, there are some bits e.g. https://github.com/CGAL/cgal/pull/2954/files#diff-5862d39c330c8154555e639d019be6ebL154, which says pretty much what is in the summary above.

Regarding the second part, I was referring to https://github.com/CGAL/cgal/pull/1808#issuecomment-276118760.

mglisse:
The parenthesis "(closed)" in Fuzzy_box is the only thing I could find. Did I miss some other mention of the boundary in the models?

Spatial_searching/include/CGAL/Fuzzy_sphere.h
```diff
@@ -32,134 +33,127 @@
...
+
+  // constructor
+  Fuzzy_sphere_impl(const Point_d& center, FT r, FT eps=FT(0),
+                    const SearchTraits& traits_ = SearchTraits())
+    : traits(traits_), c(center), sq_radius(r*r)
+  {
+    CGAL_precondition(r >= 0);
+    CGAL_precondition(eps >= 0);
+
+    sq_inner_radius = (eps > r) ? 0 : (r - eps) * (r - eps);
```
mglisse: IIUC, this means that you force the inner approximation never to be empty. If we made it empty (for instance using -1 instead of 0 as sq_inner_radius), we could return false more often from inner_range_intersects.

MaelRL: But then, if you take r=eps=0 (that is, a non-fuzzy point), the point itself will not intersect its inner range and you will not find it.

Spatial_searching/include/CGAL/Fuzzy_sphere.h
```diff
@@ -32,134 +33,127 @@
...
+
+  bool contains(const typename SearchTraits::Point_d& p) const {
+    // test whether the distance between c and p is less than the radius
+    FT distance=FT(0);
+    typename SearchTraits::Construct_cartesian_const_iterator_d construct_it=
+        traits.construct_cartesian_const_iterator_d_object();
+    typename SearchTraits::Cartesian_const_iterator_d cit = construct_it(c),
+        pit = construct_it(p), end = construct_it(c, 0);
+    for (; cit != end && (distance <= sq_radius); ++cit, ++pit) {
+      distance += ((*cit)-(*pit))*((*cit)-(*pit));
```
mglisse: Eventually we should generalize the use of CGAL::square... (not this PR's problem, this is pre-existing code)

mglisse:
Overall it looks fine to me. Any behavior on the boundary (open, closed, half-open, random) is fine as long as it is clearly documented so a user with an unusually specific need may know whether they need to implement their own item type.

Spatial_searching/include/CGAL/Fuzzy_sphere.h
mglisse: `0>0` is false, so what I suggested shouldn't change anything for that case. Did I misunderstand your reply?

Spatial_searching/include/CGAL/Fuzzy_sphere.h
```diff
@@ -87,12 +87,12 @@ class Fuzzy_sphere_impl
     typename SearchTraits::Construct_cartesian_const_iterator_d construct_it=
         traits.construct_cartesian_const_iterator_d_object();
     typename SearchTraits::Cartesian_const_iterator_d cit = construct_it(c),
-        end = construct_it(c, 0);
+                                                      end = construct_it(c, 0);
     for (int i = 0; cit != end && (distance <= sq_inner_radius); ++cit, ++i) {
       if ((*cit) < rectangle.min_coord(i))
-        distance += (rectangle.min_coord(i)-(*cit))*(rectangle.min_coord(i)-(*cit));
+        distance += CGAL::square(rectangle.min_coord(i)-(*cit));
       if ((*cit) > rectangle.max_coord(i))
```
mglisse: Not new code, but should this be `else if`?

mglisse:
Thanks, looks good to me.

Spatial_searching/include/CGAL/Fuzzy_sphere.h
MaelRL: Seems safe, yes.

Thanks for the review!

MaelRL:
There is some issue in https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-238/Spatial_searching/TestReport_lrineau_Fedora-32-Release.gz.

Can't reproduce it, I'll clean 281e5a2 once it works.

lrineau:
This PR was tested yesterday, and there was new warnings on Windows in `Point_set_shape_detection_3`:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-238.shtml#Point_set_shape_detection_3

Simon @sgiraudot might be able to help you with those, because I think he has seen that type of warning a lot.
```
C:\CGAL\test\CGAL-4.13-Ic-238\include\CGAL/Fuzzy_sphere.h(66): warning C4244: '=': conversion from 'double' to 'float', possible loss of data
C:\CGAL\test\CGAL-4.13-Ic-238\include\CGAL/Fuzzy_sphere.h(59): note: while compiling class template member function 'CGAL::internal::Fuzzy_sphere_impl<CGAL::Search_traits_adapter<Point_and_size_t,CGAL::Nth_of_tuple_property_map<0,Point_and_size_t>,CGAL::Search_traits_3<Gt>>,CGAL::Point_3<Kernel_>>::Fuzzy_sphere_impl(const Point_d &,float,float,const SearchTraits &)'
        with
        [
            Gt=CGAL::Simple_cartesian<float>,
            Kernel_=CGAL::Simple_cartesian<float>,
            Point_d=CGAL::Point_3<CGAL::Simple_cartesian<float>>,
            SearchTraits=CGAL::Search_traits_adapter<Point_and_size_t,CGAL::Nth_of_tuple_property_map<0,Point_and_size_t>,CGAL::Search_traits_3<CGAL::Simple_cartesian<float>>>
        ]
```

MaelRL:
On a side note, the results with Clang's -Wconversion are rather brutal (lots of warnings)

MaelRL:
Some strange merge happening when this PR is put into integration:

Changes induced by this PR (include boost's `pmap.hpp` to get boost tags, remove `transform_iterator.hpp` which is useless) are discarded to keep the changes from 6269309d86a.

The file is also missing an include for `CGAL::unary_function`, even in master.

Not sure what can be done in this PR though.

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [May 25, 2018, 10:58 AM GMT+2](https://github.com/CGAL/cgal/pull/2954#issuecomment-391988841 "2018-05-25T08:58:44Z - Replied by Github Reply Comments"):
> Some strange merge happening when this PR is put into integration:
> 
> Changes induced by this PR (include boost's `pmap.hpp` to get boost tags, remove `transform_iterator.hpp` which is useless) are discarded to keep the changes from 6269309d86a.
> 
> The file is also missing an include for `CGAL::unary_function`, even in master.

Which file?
 
> Not sure what can be done in this PR though.

Do you see what should be done in `master` or `CGAL-4.12-branch`. Could you prepare another PR so that, when they are both merged in `integration`, the code compiles?

MaelRL:
This file, sorry :
https://github.com/CGAL/cgal/blob/master/Spatial_searching/test/Spatial_searching/Point_with_info.h

From what I see that other commit and CGAL::unary_function are actually all in 4.12, so I should be able to just refresh the branch and it should be fine.

lrineau:
It seems this patch is enough:
```patch
--- a/Spatial_searching/test/Spatial_searching/Point_with_info.h
+++ b/Spatial_searching/test/Spatial_searching/Point_with_info.h
@@ -1,4 +1,6 @@
 #include <CGAL/boost/iterator/transform_iterator.hpp>
+#include <CGAL/functional.h>
+#include <CGAL/property_map.h>
 
 template <class Pt>
 struct My_point_with_info
```
I can prepare a PR for 4.12-branch.

MaelRL:
(I messed up, I thought this PR was for 4.12, I'll fix it.)

MaelRL:
See #3115.

lrineau:
not yet approved: there is still the issue pointed in https://github.com/CGAL/cgal/pull/2954#issuecomment-384975179.

MaelRL:
I can reproduce it compiling in 32bits, but I'm not sure how to fix. 

It seems that there is some round-off/optimization inconsistency between computing a distance between two points as:
`CGAL::squared_distance(p1, p2)` 
used to test the sanity of the result, and what the fuzzy item does, i.e.:
`distance += CGAL::square((*cit)-(*pit))`, with `cit`, `pit` being coordinate iterators.

For example, in the log (the sphere has squared radius 1):
```
contains(-0.52501725631761164692, 0.88194007026768095336) returns FALSE
...
missed point: -0.52501725631761164692 0.88194007026768095336 at sq_distance = 1.00000000000000000000
```

This inconsistency is not so bad, I can simply use `CGAL::squared_distance()` in both places (although someone computing the distance with Cartesian iterators will find an inconsistency in the result of the tree search), but this does not extend to what is done in the other functions such as `inner_intersects` and `outer_contains`...

lrineau:
Try add:
```c++
  CGAL::Set_ieee_double_precision pfr;
```
at the beginning of the `main()` function. That might be sufficient.

MaelRL:
Ha, I should have remembered that from P3T3...

Seems to be sufficient, thanks.

lrineau:
Successfully tested in [CGAL-4.13-Ic-26][*], and then merged (on Friday).

PR #2955, opened 2018-03-24
Do not include some BGL related header files

afabri:
## Summary of Changes

The files `graph_traits_Polyhedron.h`, `properties_Polyhedron.h` and `Polyhedron_iostream.h` are included by `Polyhedon.h`

## Release Management

* Affected package(s): many


lrineau:
Is that a pull-request that should be integrated in CGAL-4.12? As far as I can see, it does not solve any bug, but just cleanup things. If it is true, I would prefer to delay it to CGAL-4.13, to avoid yet another reason to delay the next release (already late by 6 weeks).

afabri:
4.13 is ok.

lrineau:
@maxGimeno A few Travis job ends with a strange error:
https://travis-ci.org/CGAL/cgal/jobs/357804492#L2238
```
sudo bash .travis/install_openmesh.sh
bash: .travis/install_openmesh.sh: No such file or directory
```

Could you have a look and try to understand why the file is not in the merge commit tested by Travis?

maxGimeno:
Yes. We checked with Andreas, the file definitly exists. It must be a location related error, maybe a missing cd. I'm surprised it passed the checks in its own branch though.

afabri:
echo $CWD

lrineau:
`echo $PWD`, on Linux.

maxGimeno:
cd $ROOT is done outside the loop over the arguments, so only the first package of the three is in the right directory when launching the installation script. Fixed in #2957

sloriot:
Conflicts with master

PR #2957, opened 2018-03-26
Travis: Fixes

maxGimeno:
## Summary of Changes
Remove the verbose option in install_openmesh.sh so the logs are less polluted.
Also fix the name of the cmake variable that computes dependences.


lrineau:
The fixes must be backported to the 4.11 branch.

maxGimeno:
Those fixes only concern lines and files that did not exist in 4.11: there were no openmesh, so no tar errors, no dependencies check, so no message error and the CD error did not exist.

PR #2966, opened 2018-03-28
Travis: Fix install.sh

maxGimeno:
## Summary of Changes
Sometimes, install.sh runs fine but exits with code 100. I suppose it is caused by a warning about ppas that are not well secured, which does not trigger a failure for the || operator, but does when exiting the script with a non 0 code. 
I added exit 0 at the end of the file, to insure that if the script executed until the end, then we don't consider any error.


maxGimeno:
@lrineau Is there someting wrong with this PR ?

maxGimeno:
Thank you !

PR #2978, opened 2018-04-03
Property map: fix get() not returning the default value

MaelRL:
## Summary of Changes

`Default_property_map` should return the value that is chosen in the constructor. 

I also extended it to be a writable property map, with the idea that we don't have (or I can't find) a nice dummy property map that is also writable. It would then replace custom pmaps such as [No_mark](https://github.com/CGAL/cgal/blob/2cec8e6aca6cab7b5e733fc88ac6baed99560d0c/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/face_graph_utils.h#L39).

## Release Management

* Affected package(s): `Property_map`

### DIFF
--- a/Classification/include/CGAL/Classification/Point_set_feature_generator.h
+++ b/Classification/include/CGAL/Classification/Point_set_feature_generator.h
@@ -496,7 +496,7 @@ private:
     launch_feature_computation (new Feature_adder_verticality<VectorMap> (this, normal_map, 0));
   }
 
-  void generate_normal_based_features(const CGAL::Default_property_map<Iterator, typename GeomTraits::Vector_3>&)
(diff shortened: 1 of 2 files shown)


lrineau:
[`boost::ref_property_map](https://www.boost.org/doc/libs/1_49_0/libs/property_map/doc/ref_property_map.html) is provided by Boost, and is a read-write property map (and even lvalue property map).

Making `Default_property_map` read-write, but with a fake `put` function, seems a bad idea: the semantic of it would become fishy.

MaelRL:
`ref_property_map` is not very practical to use either. Renamed to `Constant_property_map`, hopefully making it less awkward that there is a `put()` free function.

sloriot:
Tested in CGAL-4.13-Ic-11

PR #2979, opened 2018-04-03
Polyhedron demo: Add a verbose option

### LINKED ISSUE #2439: Polyhedron demo log
The starting log of the polyhedron is rarely used. IMO, we better have it store internally and displayed on demand. The reason for this is that when I load a lot of times the demo my terminal gets polluted and I'd like to avoid redirected the output to `/dev/null` as I might need the output of some actions.

maxGimeno:
## Summary of Changes

With this PR, all the logs for the plugins loading are hidden, unless the demo has been launched with the --verbose option. Logs for the scripts are unaffected.
## Release Management

* Issue(s) solved (if any): fix #2439

### DIFF
--- a/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
+++ b/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
@@ -61,12 +61,7 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
   parser.addOption(verbose);
   parser.addPositionalArgument("files", tr("Files to open"), "[files...]");
   parser.process(*this);
-  if(parser.isSet(verbose)) {
-    d_ptr->mainWindow.reset(new MainWindow(true));
-  }
-  else{
-    d_ptr->mainWindow.reset(new MainWindow);
-  }
+  d_ptr->mainWindow.reset(new MainWindow(parser.isSet(verbose)));
   MainWindow& mainWindow = *d_ptr->mainWindow;
   
   mainWindow.setWindowTitle(main_window_title);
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp
```diff
@@ -56,12 +56,19 @@ Polyhedron_demo::Polyhedron_demo(int& argc, char **argv,
...
   parser.addPositionalArgument("files", tr("Files to open"), "[files...]");
   parser.process(*this);
-
-  d_ptr->mainWindow.reset(new MainWindow);
+  if(parser.isSet(verbose)) {
+    d_ptr->mainWindow.reset(new MainWindow(true));
+  }
+  else{
+    d_ptr->mainWindow.reset(new MainWindow);
+  }
```
lrineau: ?!
```c++
d_ptr->mainWindow.reset(new MainWindow(parser.isSet(verbose));
```

lrineau:
@sloriot Could you verify if the new behavior of this PR corresponds to what you wanted (in #2439)?

sloriot:
With the current Qt config I have, I no longer have any qt_debug message printed on the console so I cannot really check.

lrineau:
#2439 was initially fill by Sébastien, who thereafter found a workaround. Is there still anyone interested by this PR and the `--verbose` option?

maxGimeno:
I am. And I think @janetournois might be too

maxGimeno:
@janetournois without the --verbose, at the start of the demo you still get the display of the scripts that are called, so it is not exactly nothing. But it is like 4 lines instead of 80.

janetournois:
Ok.
What worries me is more that I want to be able to have the 80 lines easily :-)

maxGimeno:
They are displayed if you use --verbose, yes.

PR #2980, opened 2018-04-03
Polyhedron demo: Fix NM simplices extraction

maxGimeno:
## Summary of Changes

Don't add NM edges extremities to NM vertices item when extracting the Non-Manifold Simplices of a polygon soup.


lrineau:
Successfully tested in [CGAL-4.12-Ic-219][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.12-Ic-219.shtml

PR #3004, opened 2018-04-10
Shape detection: fix undocumented breaking change

sgiraudot:
## Summary of Changes

`Efficient_RANSAC_traits` was renamed `Shape_detection_traits`, but this breaking change was not documented and no backward-compatibility was provided.

This PR:
- updates `CHANGES.md`
- reintroduces back the class with the old name along with a deprecated warning

## Release Management

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -144,6 +144,9 @@ Release date: April 2018
 
 ### Point Set Shape Detection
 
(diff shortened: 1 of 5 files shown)


lrineau:
That also means that the deprecated name is not tested either. Deprecated code should be tested.

sgiraudot:
I added a test for the deprecated class.

PR #3019, opened 2018-04-13
Polyhedron demo: Fix reload item

### LINKED ISSUE #3017: Reload with empty file makes the 3D demo segfault
Load a polylines.txt file with e.g. `2 0 0 0 0 0 1`, clear it, reload the item --> the demo segfaults.

maxGimeno:
## Summary of Changes
In the function that reloads an item, it is expected that the file is the same and that loadItem() won't fail. But if the file has been epmtied, for example, or has simply become invalid, then the returned item is NULL and the demo segfault. 
This adds a simple test that returns the function if that occurs, after a warning box is prompted to explain the situation.
## Release Management
* Issue(s) solved (if any): fix #3017


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -909,7 +909,14 @@ void MainWindow::reloadItem() {
     QFileInfo fileinfo(filename);
 
     CGAL::Three::Scene_item* new_item = loadItem(fileinfo, fileloader);
-
+    if(!new_item)
+    {
+      QMessageBox::warning(this, tr("Cannot reload item %1:"),
+                           tr("It seems the original file "
+                              "has been altererd and is now empty.\n"
```
MaelRL: typo

lrineau:
Actually, I wonder why the polylines loader returns the null pointer when the file in empty. It should return an empty item.

maxGimeno:
AFAIK, they all  return the null pointer. It is easier to detect when
loadaing failed, I guess.

2018-04-13 14:42 GMT+02:00 Laurent Rineau <notifications@github.com>:

maxGimeno:
I misunderstood your remark. It is the MainWindow that checks if the file is empty before trying to load.

lrineau:
The code of the load function for polylines is that one:
https://github.com/CGAL/cgal/blob/eb75474b23aecc01e5dccdfc1621e33d9e238361/Polyhedron/demo/Polyhedron/Plugins/IO/Polylines_io_plugin.cpp#L94-L142

In my opinion, the line
```c++
if(counter == 0) return 0; 
```
is wrong: if the file is empty, then the load should return an empty polylines item.

And similarly, if one loads an empty file, then `MainWindow::loadItem` should not return `0` but an empty item of the right type.

maxGimeno:
It is the same for ALL of the IO plugins, because  most of the IO functions return false if the file is empty.

lrineau:
I still think that is an error. If that error is in all our plugins, then that is worse.

If that is not obvious, then maybe we should discuss that matter with the other users of the demo.

@CGAL/geometryfactory When a file is empty, should the loading fail (with a popup), or should it create a new item with 0 elements?

sloriot:
A popup sounds better to me (if it does not make the demo crashing if the arguments are passed on the command line).

lrineau:
The popup also prevents the use of Qt Scripts...

lrineau:
This PR is still marked under discussion. It has not been changed for 6 weeks. What do we do with it? Postpone to 4.14?

maxGimeno:
@CGAL/geometryfactory up: Loading an empty file, should it create an empty item or fail with a pop-up ? Please comment.

sloriot:
Pop up: file empty

afabri:
or just a message

lrineau:
In my opinion it should create an empty item. And not a popup (because that would block scripts). Maybe display a warning message in the console. But anyway, it should create the empty item.

sloriot:
With the scenario of @lrineau  I agree with the empty file solution.

maxGimeno:
Ok, but whatever I do, an empty PLY file will be very tricky to handle. If the file is empty there is no way to know if it is a point set or a mesh that is expected. Even a reloading will be dangerous.

Three/include/CGAL/Three/Three.h
```diff
@@ -48,6 +48,21 @@ class THREE_EXPORT Three{
...
+   */
+  virtual void information(QString) = 0;
+  /*!
+   * Displays a blue text preceded by the mention "WARNING :".
+   */
+  virtual void warning(QString) = 0;
+  /*!
+   * Displays a red text preceded by the mention "ERROR :".
+   */
+  virtual void error(QString) = 0;
```
lrineau: Now the interface class `Message_interface` should be removed.

maxGimeno: That would require the modification of all the plugins, and the modification of the API of the polyhedron_demo_plugin_interface.

lrineau: Then I do not understand why you have added those three virtual function to `Three`. `MainWindow` already derives from `Message_interface` and now, for example, `MainWindow` has two virtual function named `information`.

Instead, make `Three` derives from `Message_interface`, and remove the direct derivation to it from `MainWindow`.

Actually, the whole thing about `Three::Three` seems wrong.
1. `MainWindow` derives from `Three`.
2. `MainWindow` has a `Three*` as data member.
3. `Three::Three` was supposed to be a class with only static members.

maxGimeno: 3. : The probleme here is addDockWidget(), which is an enhancement for the original qmainwindow::addDockWidget(), but is not used yet. I think it is ultimately  supposed to replace the one from plugin_helper.

maxGimeno: I don't think it is possible to make those 3 functions static. They need access to member functions from MainWindow (not QMainWindow).

lrineau: As soon as the `MainWindow` object is supposed to be a singleton, then those functions can be implemented as static ones. That is similar to `qDebug()`.

maxGimeno: done.

maxGimeno:
The last warning I don't have locally.

Polyhedron/demo/Polyhedron/Three.cpp
```diff
@@ -130,3 +128,16 @@ void Three::autoConnectActions(Polyhedron_demo_plugin_interface *plugin)
              action->objectName().toUtf8().data());
   } // end foreach action of actions()
 }
+
+void Three::s_information(QString s)
```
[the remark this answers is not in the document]
maxGimeno: s stands for static, and I added it to avoid conflicts with the MainWindow implementation, as it inheritates Three and Message_interface. I figured it was better if the names differed.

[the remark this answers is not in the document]
lrineau: Yet those names are ugly. We need to find a solution.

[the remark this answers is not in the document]
maxGimeno: we could rename the ones from message_interface, as they shouldn't appear anymore. Something like message_warning() etc. And keep warning() for the static functions of Three.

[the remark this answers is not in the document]
lrineau: That is a good solution. Thanks.

maxGimeno:

maxGimeno:
Careful with #3617 , they are not compatible. One of the two should be merged in master first and then the other one should be adapted.

PR #3034, opened 2018-04-20
Polyhedron demo: Fix the orientation of polygon soup to surface_mesh

maxGimeno:
## Summary of Changes

If the isolated vertices are removed from the mesh after its creation , its garbage needs to be collected, or the edges will be messy.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
@@ -683,8 +683,10 @@ Scene_polygon_soup_item::exportAsSurfaceMesh(SMesh *out_surface_mesh)
   CGAL::Polygon_mesh_processing::polygon_soup_to_polygon_mesh< CGAL::Surface_mesh<Point_3> >(
     d->soup->points, d->soup->polygons, *out_surface_mesh);
   std::size_t rv = CGAL::Polygon_mesh_processing::remove_isolated_vertices(*out_surface_mesh);
-  if(rv > 0)
+  if(rv > 0){
     std::cerr << "Ignore isolated vertices: " << rv << std::endl;
+    out_surface_mesh->collect_garbage();
+  }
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
```diff
@@ -683,8 +683,10 @@ Scene_polygon_soup_item::exportAsSurfaceMesh(SMesh *out_surface_mesh)
   CGAL::Polygon_mesh_processing::polygon_soup_to_polygon_mesh< CGAL::Surface_mesh<Point_3> >(
     d->soup->points, d->soup->polygons, *out_surface_mesh);
   std::size_t rv = CGAL::Polygon_mesh_processing::remove_isolated_vertices(*out_surface_mesh);
-  if(rv > 0)
+  if(rv > 0){
     std::cerr << "Ignore isolated vertices: " << rv << std::endl;
+    out_surface_mesh->collect_garbage();
```
sloriot: should be _Ignored_

sloriot:
what if you have a plugin modifies the number of vertices, will it create the same issue?

Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
maxGimeno: I don't think so. When a plugin ;odifies an item, it calls invalidateOpenGLBuffers() to apply the changes, and collect_garbage() is called in there. It means another fix would have been to call invalidateOpenGLBuffers() instead of collect_garbage(), but as the item has not been drawn  yet I think is more optimized to do like that.

maxGimeno:
I don't think so. When a plugin ;odifies an item, it calls invalidateOpenGLBuffers() to apply the changes, and collect_garbage() is called in there. It means another fix would have been to call invalidateOpenGLBuffers() instead of collect_garbage(), but as the item has not been drawn yet I think is more optimized to do like that.

PR #3038, opened 2018-04-20
Polyhedron demo: Make UI for Lighting Settings

maxGimeno:
## Summary of Changes
This PR adds a dialog (accessible from the View menu) that allows to customize the lighting of the scene.
It means you can change the color and intensity of the 3 components of the light (ambient, diffuse and specular), but mostly it allows you to move the position of the light source, which is now camera position dependant.
screenshot_20180420_155404


maxGimeno:
Why is there a 'not yet approved' tag here ? isn't it specific for Small Features ? 
Also, what remains to be discussed ?

maxGimeno:
How are world coordinates "so unintuitive that it is not usable" ? You have acces to the Bbox of your items, it makes it pretty easy to find the right position for your light, I think.

lrineau:
This PR has a small conflicts with `master`. Please merge `master` to fix it.

lrineau:
Thanks for the changes.

Now the default lighting parameters are changed, compared to `master`.

Here is the sphere, not-selected, with `master`:
screenshot_20180705_165235

Now here is the same sphere, not selected, with the default lighting parameters in this PR:
screenshot_20180705_165310

The new settings seem to have more specular light and less uniform light. I would prefer we keep the current default lightning, not to surprise users (including us).

lrineau:
And the position of the main light seems to move when the camera moves. Try with a sphere.

maxGimeno:
I changed the default values. Please aknoledge that the last values you validated in the UI are saved, so if you test again now chances are you'll get the exact same values as before. 

For the moving lighting, I don't understand what you want. When I fix the light to the camera so it moves with it you say it's not good, and when I fix it to the World so it doesn't move with the camera you also say it's not good. What is the behavior you are expecting with this sphere ?

maxGimeno:
Indeed, it was a mistake in a previous merge with master.

maxGimeno:
@lrineau you did not answer my last question about the lighting position. What behavior do you expect ? Should the light be fixed in  Camera coordinates  or in World Coordinates ? 
Camera is basically what we have in master. it is always what you are looking at that is lit.
World  basically means that it is possible to look at the dark side of the moon, what is currently implemented in this branch.

lrineau:
There is a conflict.

Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -1365,7 +1536,112 @@ void Viewer::messageLogged(QOpenGLDebugMessage msg)
...
+  connect(dialog->buttonBox->button(QDialogButtonBox::StandardButton::RestoreDefaults), &QPushButton::clicked,
+          [this, dialog](){
+    d->position = QVector4D(0,0,1,1);
+    d->ambient=QVector4D(77.0/255,77.0/255,77.0/255, 1.0);
+    d->diffuse=QVector4D(204.0/255,204.0/255,204.0/255,1.0);
+    d->specular=QVector4D(0,0,0,1.0);
+    d->spec_power = 51;    
+    update();
+    
+  });
```
lrineau: There is [a warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-75/Polyhedron_Demo/TestReport_lrineau_ArchLinux-clang-CXX14.gz):
```
.../ArchLinux-clang-CXX14/test/Polyhedron_Demo/Viewer.cpp:1621:18: warning: lambda capture 'dialog' is not used [-Wunused-lambda-capture]
          [this, dialog](){
                 ^
1 warning generated.
```

lrineau:

PR #3059, opened 2018-04-26
Fix LEDA support with gcc

lrineau:
## Summary of Changes

With CGAL-4.12, if you try to compile libCGAL with leda-numbers instead of GMP, with gcc, then you get a compilation error, telling that "  -ffriend-injection -fno-strict-aliasing" is not a valid (single) flag. The variable `LEDA_CXX_FLAGS` has to be split into a CMake list, before it can be passed to `target_compile_options`.

## Release Management


lrineau:
TODO: rebase onto `releases/CGAL-4.12-branch`.

lrineau:
Successfully tested in [CGAL-4.13-Ic-238][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-238.shtml

PR #3061, opened 2018-04-26
A fix for MSVC 2017 15.7 with /permissive-

### LINKED ISSUE #2414: Test /permissive- with Visual Studio 2017
Recent versions of Visual C++ 2017 have a new flag to enable a better conformance to the C++ norm: `/permissive-`.

https://blogs.msdn.microsoft.com/vcblog/2016/11/16/permissive-switch/

I think we should try this flag, in on of our VC++2017.

IMO, that is not a priority task, but it could be done during the reorganization of our Windows testsuite.

lrineau:
## Summary of Changes

A fix for MSVC 2017 15.7 (Preview 4) with `/permissive-`

MSVC is probably right that it was not a valid C++ code: the previous
version called a static member function of the enclosing class, without
qualification, and that static member function was defined *after*.

@mglisse An interesting exercise of reading the norm. From what I ready [in cppreference][1]:
> c) if this class is nested, the body of the enclosing class until the declaration of this class and the entire body of the base class(es) of the enclosing class.

The part "until the declaration of this class" is the critical argument, in my opinion. As far as I understand, in the first pass of the Two Pass Name Lookup, the code is ill-formed.

[1]: http://en.cppreference.com/w/cpp/language/unqualified_lookup#Class_definition

## Release Management

* Issue(s) solved (if any): fix #2414

### DIFF
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -306,13 +306,13 @@ public:
         switch(Traits::longest_axis(bbox))
         {
         case AT::CGAL_AXIS_X: // sort along x
-          std::nth_element(first, middle, beyond, boost::bind(less_x,_1,_2,m_traits));
+          std::nth_element(first, middle, beyond, boost::bind(Traits::less_x,_1,_2,m_traits));
           break;
         case AT::CGAL_AXIS_Y: // sort along y
-          std::nth_element(first, middle, beyond, boost::bind(less_y,_1,_2,m_traits));
+          std::nth_element(first, middle, beyond, boost::bind(Traits::less_y,_1,_2,m_traits));
           break;
         case AT::CGAL_AXIS_Z: // sort along z
-          std::nth_element(first, middle, beyond, boost::bind(less_z,_1,_2,m_traits));
+          std::nth_element(first, middle, beyond, boost::bind(Traits::less_z,_1,_2,m_traits));
           break;
         default:
           CGAL_error();
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -291,7 +291,8 @@ public:
    */
   class Sort_primitives
   {
-  const AABB_traits<GeomTraits,AABBPrimitive,BboxMap>& m_traits;
+    typedef AABB_traits<GeomTraits,AABBPrimitive,BboxMap> Traits;
+    const Traits& m_traits;
   public:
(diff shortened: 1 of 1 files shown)


tomkulaga:
I've found another in 

cgal\surface_sweep_2\arr_overlay_traits_2.h

Same compile commands as #2414 but slightly modified test code:

```
#include <QDebug>
#include <CGAL/Surface_sweep_2/Arr_overlay_traits_2.h> //<=== NOW THIS IS THE CULPRIT

// New working set
using K_AABB = CGAL::Exact_predicates_exact_constructions_kernel;



{
	qDebug() << "Hello World" << endl;
	return 0;
}
```

lrineau:
@tomkulaga You are right. There are *a lot* of compilation errors with MSVC 2017 version 15.7 Preview 4. See the test results here, published today:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-238.shtml

The column corresponding to the platform named `x64_Cygwin-Windows10_MSVC2017-Debug-64bits` shows a lot of errors.

Our Windows developers @janetournois and @afabri will continue the work, and try to support that version of Visual, with `/permissive-` and `/std:c++latest`.

There are a lot of day off, in May, in France, and a lot of us will be in vacations. So that work may be slow. Let's hope it can be finished by the end of May 2018, so that I can publish a new version CGAL-4.12.1 that would fully support MSVC 2017 15.7.

afabri:
Well, it is "only" a preview release. Note that CGAL supports the latest official version with -permissive-

afabri:
We also have a problem with `std::nth_element` in the [Minkowski_sum_2 testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-3/Minkowski_sum_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz)

C++17 introduces  a parameter [ExecutionPolicy](http://en.cppreference.com/w/cpp/algorithm/nth_element)

mglisse:
> We also have a problem with std::nth_element in the Minkowski_sum_2 testsuite

Seems to be the same issue with the order of declarations.

> C++17 introduces a parameter ExecutionPolicy

That seems unrelated.

afabri:
Fixed and pushed - thank you Marc!!

lrineau:
We do not see the commit. Maybe you pushed to you fork instead of mine (`lrineau`).

afabri:
I pushed to #3071

lrineau:
We fixed that issue in CGAL-4.12.1 and CGAL-4.13. CGAL-4.12 was already released and it was too late.

PR #3069, opened 2018-05-02
Fix constrained edge map update and add user projection functor as input

sloriot:
In the current operations, not only input edges are marked as constrained in the output (but patch boundary edge too for example). That might be fixing issues observed in #2737. I'll check that.

I also added an option to no use the reprojection on the input surface but use a functor provided by the user. In this case the AABB-tree and the connected components are not computed.


Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt
```diff
@@ -299,15 +299,22 @@ If this parameter is not provided, the perturbation is not deterministic
...
 of close meshes should be done in addition to surface intersection tests.
 \n
-\b Type : `bool` \n
-\b Default value is `false`
+<b>Type:</b> `bool` \n
+<b>Default:</b> `false`
+\cgalNPEnd
+
+\cgalNPBegin{projection_functor} \anchor PMP_projection_functor
+Parameter used in isotropic remeshing function to specify an alternative vertex projection method.
```
janetournois: In other named parameters descriptions, we use the function name (`isotropic_remeshing()`) to get the link to the function. Can you please use it here?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -107,6 +107,12 @@ namespace Polygon_mesh_processing {
 *    constrained in `edge_is_constrained_map` and boundary edges move along the
 *    constrained polylines they belong to.
 *  \cgalParamEnd
+*  \cgalParamBegin{projection_functor}
+*  A function object used to project vertices created. It must have
```
janetournois: It's not only created vertices, but all non-constrained-non-boundary vertices that are projected. The smoothing step moves input vertices, that are also projected by the functor. I suggest : "A function object used to project moving vertices" (this includes input vertices and created vertices)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -239,8 +249,8 @@ void isotropic_remeshing(const FaceRange& faces
     }
     remesher.equalize_valences();
     remesher.tangential_relaxation(smoothing_1d, nb_laplacian);
-    remesher.project_to_surface();
```
janetournois: can you please keep `project_to_surface()` on the same line to keep code clean? :-)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -125,6 +125,8 @@ namespace internal {
       : border_edges_ptr(new std::set<edge_descriptor>() )
       , pmesh_ptr_(NULL)
     {}
+
+    template <class FaceRange>
```
janetournois: I thought having a template parameter on a constructor was forbidden (at compile time)??

janetournois:
I don't understand why storing a set of constrained edges is necessary.
The halfedges status should be enough...?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
sloriot: No it is perfectly fine.

sloriot:
The status is not enough because patch border edges are considered as constrained edges while they were not in the input map.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -634,7 +636,7 @@ namespace internal {
       BOOST_FOREACH(edge_descriptor e, edges(mesh_))
       {
         double sqlen = sqlength(e);
-        if( (sqlen < sq_low) && is_collapse_allowed(e) )
+        if( (sqlen < sq_low) && is_collapse_allowed(e) ) //SL I wonder if we shouldn't delay the is_collapse_allowed test
```
janetournois: it is re-tested in the `while (!short_edges.empty())` loop, so it is an option, indeed

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -81,7 +81,6 @@ namespace Polygon_mesh_processing {
 *  \cgalParamBegin{edge_is_constrained_map} a property map containing the
 *    constrained-or-not status of each edge of `pmesh`. A constrained edge can be split
 *    or collapsed, but not flipped, nor its endpoints moved by smoothing.
-*    Sub-edges generated by splitting are set to be constrained.
```
janetournois: @sloriot did you remove this on purpose, or is it a git merge mistake?

janetournois:
@sloriot I do not understand your commit f03f217
Constrained edges can be collapsed, as far as `protect_constraints == false`
Why do you want to prevent it?

sloriot:
@janetournois Actually the commit message is not correct. The change has do do with the two pairs (in general) of edges that will be merged by the collapse. The call to the function with the constrained edge map will favor to keep constrained edges after the merge (you already prevent to collapse the edge if both edges of the pair were constrained). Do you agree?

janetournois:
Ok!
I thought the call to the function with the constrained edge map was preventing the collapse from happening **at all** for constrained edges.
Thank you!

lrineau:
There are conflicts with the new version of `master`. @sloriot, can you fix them? Let me know once that is done, so that I can merge the PR into `integration`.

sloriot:
@lrineau it was tricky but I think I'm good.

sloriot:
@janetournois I have a serious doubt about one point. `merge_status(h, s1, s2)` that is used just before `collapse_edge()`. Since the function is now not collapsing constrained edges incident to `e`, I think the overload of `merge_status()` used here (which is the only place where it is used) should be updated to update the status of both edges as it can be any of them that will be removed. What do you think?

lrineau:
There is a Travis error:
https://travis-ci.org/CGAL/cgal/jobs/388842842#L2205
```
[ 31%] Building CXX object CMakeFiles/isotropic_remeshing_of_patch_example.dir/isotropic_remeshing_of_patch_example.cpp.o
In file included from /home/travis/build/CGAL/cgal/Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp:1:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/Exact_predicates_inexact_constructions_kernel.h:29:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/Simple_cartesian.h:29:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/Cartesian/Cartesian_base.h:29:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/basic.h:43:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/assertions.h:362:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/assertions_impl.h:35:
In file included from /usr/local/lib/cmake/CGAL/../../../include/CGAL/exceptions.h:29:
In file included from /usr/include/boost/lexical_cast.hpp:169:
In file included from /usr/include/boost/range/iterator_range_core.hpp:23:
/usr/include/boost/iterator/iterator_facade.hpp:551:22: error: no member named
      'distance_to' in 'CGAL::Surface_mesh<CGAL::Point_3<CGAL::Epick>
      >::Index_iterator<CGAL::SM_Face_index>'
          return -f1.distance_to(f2);
                  ~~ ^
```

sloriot:
I don't have it locally, I suspect an issue with boost. Let see if I can workaround it. Did you already saw that somewhere?

sloriot:
With boost 1.55 I can reproduce it.

sloriot:
Forgot to mention that I fixed the issue.

lrineau:
The CTest of `remeshing_test` fails in that PR:
```shellsession
$ ctest -V -R remeshing_test -RE remeshing_test_P_SM_OM
[...]
3414: Remeshing done (size = 0.079, #iter = 1, 0.002997 sec ).
3414: terminate called after throwing an instance of 'CGAL::Assertion_exception'
3414:   what():  CGAL ERROR: assertion violation!
3414: Expr: exception_caught
3414: File: /home/lrineau/Git/cgal/Polygon_mesh_processing/test/Polygon_mesh_processing/remeshing_test.cpp
3414: Line: 103
3414: 
3414: MEM: 7304 TIME: 0.81      /home/lrineau/Git/cgal/build/test/Polygon_mesh_processing/remeshing_test
3/6 Test #3414: execution___of__remeshing_test .................***Failed    0.82 sec
```

sloriot:
@lrineau  I fixed the issue and `ctest` now runs without errors for tests and examples.

sloriot:
Actually it was only for the isotropic remeshing plugin, there are other errors not related to this PR. It is not fixed.

lrineau:
Successfully tested in [CGAL-4.13-Ic-42][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-42.shtml

PR #3075, opened 2018-05-04
Travis: Make the dependencies check NOT order dependant.

maxGimeno:
## Summary of Changes
The current dependencies are checked with a simple diff between the committed and the re-computed dependecies files, which means that if the dependencies are not found in the same order, it counts like an error. With this PR it is not like that anymore, which should allow to manually add or remove the problematic dependencies without having to run the 20 min process to try and find the problems.
## Release Management

* Affected package(s):Scripts

### DIFF
--- a/Scripts/developer_scripts/cgal_check_dependencies.sh
+++ b/Scripts/developer_scripts/cgal_check_dependencies.sh
@@ -28,19 +28,22 @@ do
   fi
 done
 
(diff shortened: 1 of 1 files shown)


sloriot:
@lrineau shall we merge it as is?

lrineau:
I think you can merge it. From what I can see, there is only one modified file, and it is tested only by Travis. And it was tested by Travis successfully 11 days ago.

PR #3079, opened 2018-05-08
Alpha_shapes: class template has already been defined.

afabri:
## Summary of Changes

`CGAL::internal::Is_traits_point_convertible` is defined twice, namely in `Lazy_alpha_nt_2/3.h` and the two are slighlty different. The fix is to postfix with the dimension.

## Release Management

* Affected package(s): Alpha_shape_2/3
* Issue(s) solved (if any): bugreport by a user


sloriot:
Tested in CGAL-4.13-Ic-11

PR #3081, opened 2018-05-09
Documentation overview: enhancement

maxGimeno:
## Summary of Changes
Specifies in the result page that Doxygen 1.8.13 is the one used for the CGAL documentation and adds a header to the result table to regroup the columns by version in a cleaner way.
## Release Management

### DIFF
--- a/Documentation/doc/scripts/testsuite.py
+++ b/Documentation/doc/scripts/testsuite.py
@@ -78,7 +78,8 @@ body  {color: black; background-color: #C0C0D0; font-family: sans-serif;}
 </style>
 </head><body>
 <h1 id="maintitle">Doxygen Manual Results</h1>'''
-    page_footer='''<table class="test-results">
+    page_footer='''<table border="1" cellspacing="2" cellpadding="5" class="test-results">
+    <tr><td/><th colspan="3">Doxygen 1.8.4</th><th colspan="3">Doxygen 1.8.13(official)</th><th colspan="3">Doxygen master</th></tr>
 <tr>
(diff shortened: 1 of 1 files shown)


sloriot:
@lrineau mentionned in a private email that we can stop tested 1.8.4, shall we use this PR to make this change too?

lrineau:
There is actually one thing that could prevent us to upgrade. So far, only the documentation of CGAL-4.12 and `master` (future 4.13) was using Doxygen 1.8.13. If we drop 1.8.4, that means that 4.11.2 will have to be documented with 1.8.13 as well. That means backporting the script to that release branch. As far as I know, that should be fine.

By the way, the PR should be rebased on the branch for 4.12.1.

maxGimeno:
> By the way, the PR should be rebased on the branch for 4.12.1.

Why ? What difference does it make?

lrineau:
If you don't, then the branch `releases/4.12-branch` will not get the PR, and every week-end when 4.12 is tested instead of 4.13 it will use the old script.

lrineau:
Then could you please rebase the PR on 4.12?

lrineau:
This PR is now merged, but I do not see the changes in the result page:
  https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/

Do you know why?!

maxGimeno:
I do ! Compare Ic12 and 14 it should strike you enough :)

lrineau:
Sorry I did not look the right page. All is fine. Thanks.

PR #3087, opened 2018-05-11
Triangulation_2: Fix Constrained Delaunay demo. 

afabri:
## Summary of Changes

The `hint` was outdated after `cdt.clear()`, leading to a segfault.
In order to reproduce: generate points, switch on circumcircle, clear, add points.

## Release Management


sloriot:
_Tested_ in CGAL-4.13-Ic-11 but the patch is trivial.

PR #456, opened 2015-11-02
Fix warnings: use CGAL_CORE_EXPORT instead of CGAL_EXPORT

cjamin:
This was a typo introduced by the header-only branch.

PR #3092, opened 2018-05-16
Infrastructure : Test compilation of headers

### LINKED ISSUE #3089: Check that one can include all CGAL documented headers in one cpp file
## Issue Details

See PR #3079, that fixes an issue reported by a user: if one includes both Alpha_shape_2 and _3, then there is a compilation error.

We should check that a .cpp including all documented headers of CGAL does compile. It could be an extra check to our verification "check headers". All documented from the CMake list `list_of_documented_headers` (but those that are excluded/whitelisted) should be `#include`d in a .cpp file, created in `CMAKE_BINARY_DIR`, and then the file should be compile with `-fsyntax-only`.

I have make a quick and dirty attempt, and:
  - we document non-existing headers,
  - there are other headers than `CGAL/Alpha_shape_2.h` and `_3.h` that conflict.

(I would prefer that the PR is based on 4.12.)

maxGimeno:
## Summary of Changes

This adds a step to check_headers that creates a cpp file including all checked headers and tries to build it.
Here is the said file on my machine :
[test_headers.cpp](https://gist.github.com/maxGimeno/ac6f69b63f4bccfe8cfa297d6c2a53a8)
@CGAL/geometryfactory  We should work to fix the errors that are generated by the compilation of that file. There are quite a few.
## Release Management

* Issue(s) solved (if any): fix #3089


afabri:
For VC++ you have [/Zs](https://msdn.microsoft.com/en-us/library/we9a1ahs.aspx) for syntax-only

Mesh_2/include/CGAL/lloyd_optimize_mesh_2.h
```diff
@@ -40,27 +41,13 @@
...
   
 
 BOOST_PARAMETER_NAME( cdt )
-BOOST_PARAMETER_NAME( (max_iteration_number, tag) max_iteration_number_ )
-BOOST_PARAMETER_NAME( (convergence, tag) convergence_)
-BOOST_PARAMETER_NAME( (time_limit, tag) time_limit_ )
-BOOST_PARAMETER_NAME( (freeze_bound, tag) freeze_bound_)
 BOOST_PARAMETER_NAME( (seeds_begin, tag) seeds_begin_)
 BOOST_PARAMETER_NAME( (seeds_end, tag) seeds_end_)
 BOOST_PARAMETER_NAME( (mark, tag) mark_)
```
MaelRL: All the named parameters were regrouped in /BGL even if they are only used in a single function of PMP. Maybe it ought to be the same with the parameters.

BGL/include/CGAL/boost/parameter.h
```diff
@@ -0,0 +1,45 @@
...
+// 
+//
+// Author(s)     : Andreas Fabri
+
+#ifndef CGAL_BOOST_PARAMETER_H
+#define CGAL_BOOST_PARAMETER_H
+
+#ifdef BOOST_PARAMETER_MAX_ARITY
+#  if (BOOST_PARAMETER_MAX_ARITY < 12)
+#    error "BOOST_PARAMETER_MAX_ARITY must be at least 12 for CGAL::Mesh_3"
```
MaelRL: That looks a little weird as this file might be included by Mesh_2 and if the arity is defined to 8 manually ahead of this file (which is done in the mesh_2 plugin for example), it shouldn't complain that Mesh_3 needs more.

Mesh_2/include/CGAL/lloyd_optimize_mesh_2.h
lrineau: I do not understand you choose BGL.

For PMP, there was a reason: you are using the named parameters of the BGL, but for Mesh_2 and Mesh_3 we use the named parameters from the Boost Parameters library.

BGL/include/CGAL/boost/parameter.h
afabri: Shall I write "at least 12 for CGAL" ?

maxGimeno:
@lrineau Isn't check_headers a dependency of packages_dependencies ?

afabri:
At least the fixes in CGAL code could make it into 4.12

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [May 23, 2018, 9:33 AM GMT+2](https://github.com/CGAL/cgal/pull/3092#issuecomment-391249453 "2018-05-23T07:33:19Z - Replied by Github Reply Comments"):
> @lrineau Isn't check_headers a dependency of packages_dependencies ?

I have not check the recent modifications you made. I'll try.

lrineau:

Actually, the last commit (https://github.com/CGAL/cgal/pull/3092/commits/ee02f3588808eca1fc47c37b0f065eb88b41f271) was not really tested, but it is trivial. I checked it locally on my Linux machine.

lrineau:
There is a problem with this PR. The file `test_headers.cpp` is not tested by Travis.

CGAL_Core/include/CGAL/CORE/CoreDefs.h
```diff
@@ -55,7 +55,7 @@
 #else // CGAL_HEADER_ONLY
 
   #define CGAL_GLOBAL_STATE_VAR(TYPE, NAME, VALUE)  \
-    CGAL_CORE_EXPORT extern TYPE NAME;                   \
+    CGAL_EXPORT extern TYPE NAME;                   \
```
lrineau: @afabri Do you remember having committed that change? It is dated May 16 2018, and it reverts #456.

See https://github.com/CGAL/cgal/pull/456#issuecomment-396274500

afabri: I don't . sorry.

PR #3098, opened 2018-05-18
Fix CDT_2 errors, using snapping of intersection points

### LINKED ISSUE #2999: [Crash] Constrained 2D Delaunay triangulation: assertion violation!
## Issue Details

A crash occurs when inserting a set of constraints that touch each-other (no "full" intersection nor common point)
config

On a debug executable, the following message is displayed:

> CGAL error: assertion violation!
> Expression : false
> File       : C:\Sdk\CGAL\4.11.1\include\CGAL/Polyline_constraint_hierarchy_2.h
> Line       : 1013
> Explanation:
> Refer to the bug-reporting instructions at http://www.cgal.org/bug_report.html

The defect is very sensitive to there being three constraints, and the order in which the points are given when inserting a particular constraint.

## Source Code
Here is the smallest code snippet I could come up with that reproduces the problem.

```
#include <CGAL/Triangulation_face_base_2.h>
#include <CGAL/Triangulation_vertex_base_2.h>
#include <CGAL/Delaunay_triangulation_2.h>

using Vb2 = CGAL::Triangulation_vertex_base_2<K>;
using Fb2 = CGAL::Constrained_triangulation_face_base_2<K>;
using TDS2 = CGAL::Triangulation_data_structure_2<Vb2,Fb2>;
using Itag2 = CGAL::Exact_predicates_tag;
using CDT2_Base = CGAL::Constrained_Delaunay_triangulation_2<K,TDS2,Itag2>;

using Triangulation = CGAL::Constrained_triangulation_plus_2<CDT2_Base>;
using point = Triangulation::Point;

void main(void)
{
  Triangulation t;

  {
    point const p1(942455, 2306674.6);
    point const p2(942452.3, 2306671.9);
    t.insert_constraint(p1,p2);
  }

  point const p3(942453.8, 2306673.4);
  {
    point const p4(942441.1, 2306667.8);
    t.insert_constraint(p4,p3);
  }

  {
    point const p5(942447.7, 2306686);
    t.insert_constraint(p3,p5);
  }

  return;
}

```

## Environment

* Windows (64 bits)
* Compiler: vs 2013 32 and 64
* Release or debug mode: crash in release, assert in debug
* CGAL version: 4.11.1 (reproduced with 4.11)
* Boost version: 1.65.1

lrineau:
## Summary of Changes

When `Constrained_triangulation_2` computes the intersection of two
segments, with a floating-point number type, and with
`Exact_predicates_tag`, the computed intersection point is snapped to an
extremity of the two segments, if it is closest to 4 ulp (with the l-inf
distance).

That value `4` can be changed by defining the macro
`CGAL_CDT_2_INTERSECTION_SNAPPING_ULP_DISTANCE` to another value.

I have benched on a crazy data set of 56000 segments, intersecting, and with duplicated (or almost duplicated) segments. I did not found more than 0.1s of difference, in a run of 50.0s, and the difference was not always in the same direction.

## Release Management

That is a bug-fix, but I wonder if it should be backported to 4.11.2 and 4.12.1, because that modifies the behavior of the code.

What is more, that choice of the distance (4 ulp in the l-infinity distance) is kind of arbitrary. That means about 3 bits of precision, out of 53 (if the number type is `double`).

Anyway, if one wants to backport the patch, it is only one commit, affecting only one file (and only one function `intersection(geom_traits, a,b, c, d, tag)` in that file.

* Issue(s) solved (if any): fix #2999

@mglisse Do you have any clue why that distance of 4 ulp fixed all the problematic data sets we had, but not 3? I imagine that there is real reason it worked with that value, and there might data sets where it fails.

Cc @afabri, how worked with me on the issue.

### DIFF
--- a/Kernel_23/include/CGAL/Bbox_2.h
+++ b/Kernel_23/include/CGAL/Bbox_2.h
@@ -164,8 +164,8 @@ Bbox_2::dilate(int dist)
   using boost::math::float_advance;
   float_advance(rep[0],-dist);
   float_advance(rep[1],-dist);
+  float_advance(rep[2],dist);
   float_advance(rep[3],dist);
-  float_advance(rep[4],dist);
 }
   
 inline
--- a/Kernel_23/include/CGAL/Bbox_2.h
+++ b/Kernel_23/include/CGAL/Bbox_2.h
@@ -29,6 +29,7 @@
 #include <CGAL/IO/io.h>
 #include <CGAL/Dimension.h>
 #include <CGAL/array.h>
+#include <boost/math/special_functions/next.hpp>
 
 namespace CGAL {
 
@@ -75,6 +76,7 @@ public:
   inline Bbox_2     operator+(const Bbox_2 &b) const;
   inline Bbox_2&     operator+=(const Bbox_2 &b);
 
+  inline void dilate(int dist);
 };
 
 inline
@@ -155,7 +157,17 @@ Bbox_2::operator+=(const Bbox_2& b)
   rep[3] = (std::max)(ymax(), b.ymax());
   return *this;
 }
-
+inline
+void
+Bbox_2::dilate(int dist)
+{
+  using boost::math::float_advance;
+  float_advance(rep[0],-dist);
+  float_advance(rep[1],-dist);
+  float_advance(rep[3],dist);
+  float_advance(rep[4],dist);
+}
+  
 inline
 bool
 do_overlap(const Bbox_2 &bb1, const Bbox_2 &bb2)
--- a/Triangulation_2/doc/Triangulation_2/Concepts/ConstrainedTriangulationTraits_2.h
+++ b/Triangulation_2/doc/Triangulation_2/Concepts/ConstrainedTriangulationTraits_2.h
@@ -80,7 +80,8 @@ typedef unspecified_type Compute_squared_distance_2;
 A function object whose
 `operator()` computes the bounding box of a point.
 
-`Point_2 operator()(Point_2 p);` Returns the bounding box of `p`. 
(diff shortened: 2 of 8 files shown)


Triangulation_2/doc/Triangulation_2/Concepts/ConstrainedTriangulationTraits_2.h
```diff
@@ -76,6 +76,14 @@ between `p` and `l`.
 */ 
 typedef unspecified_type Compute_squared_distance_2; 
 
+/*!
+A function object whose
+`operator()` computes the bounding box of a point.
+
+`Point_2 operator()(Point_2 p);` Returns the bounding box of `p`. 
+*/
```
lrineau: The documentation is wrong:
```
`Point_2 operator()(Point_2 p);` Returns the bounding box of `p`. 
```

lrineau:
After a discussion with Andreas, I have rebased the branch onto `releases/CGAL-4.11-branch`.

lrineau:
Successfully tested in [CGAL-4.13-Ic-26][*], and then merged on Friday.

Kernel_23/include/CGAL/Bbox_2.h
```diff
@@ -155,7 +157,17 @@ Bbox_2::operator+=(const Bbox_2& b)
...
+inline
+void
+Bbox_2::dilate(int dist)
+{
+  using boost::math::float_advance;
+  float_advance(rep[0],-dist);
+  float_advance(rep[1],-dist);
+  float_advance(rep[2],dist);
+  float_advance(rep[3],dist);
+}
```
lrineau: `float_advance` does not modify its arguments, but returns a value!

That is why the bug #2999 is still there.

PR #3101, opened 2018-05-19
Fix conversion from mpq to interval

### LINKED ISSUE #3099: to_interval(Gmpq) underflow with mpfr-3+
## Issue Details

`CGAL::to_interval(Gmpq(std::numeric_limits<double>::denorm_min())/2)` returns the interval [denorm_min, denorm_min] instead of [0, denorm_min] with the new variant of the code for mpfr-3+.

A workaround is to call `mpfr_set_emin (-1073);`. While we are at it, we might as well also call `mpfr_set_emax (+1024);` which should make the is_finite test unnecessary. A drawback is that this is a global variable, and if the user also directly uses MPFR, it might interfere with their use. So we may have, each time we call to_interval, to get the current emin, store it somewhere, set -1073, do the computation, then restore the original emin...

While changing this, we should check if we also need a call to mpfr_subnormalize, for a number like `Gmpq(DBL_TRUE_MIN)*3/2`.

## Environment

* CGAL version: 4.12
* Other libraries versions if used (Eigen, TBB, etc.): MPFR-3.*

mglisse:
## Release Management

* Issue(s) solved (if any): fix #3099

### DIFF
--- a/Number_types/include/CGAL/Gmpq.h
+++ b/Number_types/include/CGAL/Gmpq.h
@@ -90,9 +90,14 @@ template <> class Real_embeddable_traits< Gmpq >
       public:
         std::pair<double, double> operator()( const Type& x ) const {
 #if MPFR_VERSION_MAJOR >= 3
+	  mpfr_exp_t emin = mpfr_get_emin();
+	  mpfr_set_emin(-1073);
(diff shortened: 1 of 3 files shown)


lrineau:

PR #3104, opened 2018-05-22
Infrastructure: Enhance dependency managing in Travis

maxGimeno:
## Summary of Changes
The errors due to dependency modification are painful to detect in the over spammed report of travis. 
With this PR, each change reported by the check is listed in the end, so we don't have to grep PKG_DIFF and search for one that is not empty anymore.

### DIFF
--- a/.travis/template.txt
+++ b/.travis/template.txt
@@ -25,7 +25,7 @@ before_script:
 - cd ..
 script: 
 - cd ./.travis
-- bash -x -e ./build_package.sh $PACKAGE
+- if [ -n "$CGAL_DEBUG_TRAVIS" ]; then  bash -x -e ./build_package.sh $PACKAGE; else bash -e ./build_package.sh $PACKAGE; fi
 notifications:
   email:
     on_success: change # default: always
--- a/.travis/template.txt
+++ b/.travis/template.txt
@@ -25,7 +25,7 @@ before_script:
 - cd ..
 script: 
 - cd ./.travis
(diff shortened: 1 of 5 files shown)


lrineau:
While you're at it, I would remove all the `-x` options from Travis scripts:
```shellsession
$ ack -- -x .travis*
.travis/template.txt
28:- bash -x -e ./build_package.sh $PACKAGE

.travis/install.sh
3:set -x

.travis.yml
71:- bash -x -e ./build_package.sh $PACKAGE
```
And replace them by:
```bash
[ -n "$CGAL_DEBUG_TRAVIS" ] && set -x
```
in both `.travis/build_package.sh` and `.travis/install.sh`.

That way, Travis scripts will not use `-x`  by default, but only if we set `CGAL_BUILD_TRAVIS=y` in `.travis.yml` or in settings of the Travis web interface.

maxGimeno:
That seems like an excellent idea !

.travis/template.txt
```diff
@@ -25,7 +25,7 @@ before_script:
 - cd ..
 script: 
 - cd ./.travis
-- bash -x -e ./build_package.sh $PACKAGE
+- if [ -n "$CGAL_DEBUG_TRAVIS" ]; then  bash -x -e ./build_package.sh $PACKAGE; else bash -e ./build_package.sh $PACKAGE; fi
```
lrineau: The Travis `CHECK` says that this template and the file `.travis.yml` do not agree:
https://travis-ci.org/CGAL/cgal/jobs/382181056#L1849

lrineau:
Be careful: I have rebased this PR on `releases/CGAL-4.12-branch`.

PR #3109, opened 2018-05-22
add missing const to compare_endpoints_xy_2 () operator taking x-mono…

psarahdactyl:
…tone-curve as a parameter

## Summary of Changes

Allows use of the () operator for compare_endpoints_xy_2 functor

## Release Management

* Affected packages: Arrangement_on_surface_2 (Arr_Bezier_curve_traits_2)

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_Bezier_curve_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_Bezier_curve_traits_2.h
@@ -784,7 +784,7 @@ public:
      * \return SMALLER if the curve is directed right;
      *         LARGER if the curve is directed left.
      */
-    Comparison_result operator() (const X_monotone_curve_2& cv)
+    Comparison_result operator() (const X_monotone_curve_2& cv) const
     {
       if (cv.is_directed_right())
         return (SMALLER);


lrineau:
@psarahdactyl @efifogel 

Thanks for the patch. How did you detect the issue? I wonder why the CGAL testsuite does not catch it.

Which CGAL version is impacted? If this pull-request is merged as it is, the patch will only go into our `master` branch, and then into CGAL-4.13 (published Summer 2018), and not in our bug-fix releases CGAL-4.11.2 and CGAL-4.12.1.

psarahdactyl:
I am using CGAL 4.12, which is affected. I was using the CGAL::join on a set of polygons in Visual Studio 2017 and got a compiler error ( [Compiler Error C3848](https://msdn.microsoft.com/en-us/library/t209x37w.aspxl) ).

lrineau:
TODO for me: this patch should also be applied to the `CGAL-4.12-branch`.

lrineau:
Thanks @psarahdactyl for the patch. It has been merged in both `master` and the release branch for CGAL-4.12.1.

(Successfully tested in [CGAL-4.13-Ic-30][*], and then merged.)

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-30.shtm

PR #3124, opened 2018-05-28
Travis: Add  check_headers in the CHECK pass of travis

maxGimeno:
## Summary of Changes

This should trigger the test of the test_headers.cpp file.

### DIFF
--- a/Scripts/developer_scripts/cgal_check_dependencies.sh
+++ b/Scripts/developer_scripts/cgal_check_dependencies.sh
@@ -10,6 +10,8 @@ do
         echo "0 otherwise."
         exit 0
             ;;
+        --check_headers) DO_CHECK_HEADERS="True"
+            ;;
         --*) echo "bad option $1"
             ;;
         *) DOX_PATH="$1"
@@ -29,6 +31,9 @@ do
 done
 
 cmake -DCGAL_HEADER_ONLY=FALSE -DCGAL_ENABLE_CHECK_HEADERS=TRUE -DDOXYGEN_EXECUTABLE="$DOX_PATH" -DCGAL_COPY_DEPENDENCIES=TRUE -DCMAKE_CXX_FLAGS="-std=c++11" ..
+if [ -n "$DO_CHECK_HEADERS" ]; then
+    make -j$(nprc --all) -k check_headers
+fi
 make -j$(nproc --all) -k packages_dependencies
 echo " Checks finished"
 for pkg_path in $CGAL_ROOT/*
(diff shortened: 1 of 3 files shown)


lrineau:
If this PR pass the test, we will have to rebase it onto `CGAL-4.12-branch`.

maxGimeno:
The travis fail is fixed by #3113

Scripts/developer_scripts/cgal_check_dependencies.sh
```diff
@@ -29,6 +31,9 @@ do
 done
 
 cmake -DCGAL_HEADER_ONLY=FALSE -DCGAL_ENABLE_CHECK_HEADERS=TRUE -DDOXYGEN_EXECUTABLE="$DOX_PATH" -DCGAL_COPY_DEPENDENCIES=TRUE -DCMAKE_CXX_FLAGS="-std=c++11" ..
+if [ -n "$DO_CHECK_HEADERS" ]; then
+    make -j$(nprc --all) -k check_headers
```
lrineau:
```
Scripts/developer_scripts/cgal_check_dependencies.sh: line 35: nprc: command not found
```
https://travis-ci.org/CGAL/cgal/jobs/385068433#L4956

lrineau: I think that error lead to a run of `make -j`, without any number after the `-j`, and that lead Travis to fork thousands of compilation processes in parallel.

maxGimeno: there is a 'o' missing :/ I fix that immediatly

maxGimeno:
From what I understand, we have a conflict between an enum called EXIT and a macro in opencv also called EXIT.

lrineau:
My suggestion is the following patch:
```patch
--- a/STL_Extension/include/CGAL/assertions_behaviour.h
+++ b/STL_Extension/include/CGAL/assertions_behaviour.h
@@ -28,6 +28,11 @@
 #ifndef CGAL_ASSERTIONS_BEHAVIOUR_H
 #define CGAL_ASSERTIONS_BEHAVIOUR_H
 
+// workaround against the definition of EXIT in <opencv2/core/internal.hpp>
+#ifdef EXIT
+#  undef EXIT
+#endif
+
 namespace CGAL {
 
 enum Failure_behaviour { ABORT, EXIT, EXIT_WITH_SUCCESS, CONTINUE,
```

It's self-defense against the aggressive bad practice in opencv. And anyway that bad `#define EXIT ...` has been removed from OpenCV [5 years ago](https://github.com/opencv/opencv/commit/d62bc8cfbf4a17e9d3a06f81164d8554cf7a14a0#diff-4459a768435f4c7ed078550dadb5b11eL53).

Issue #2676, opened 2017-12-11
bug with linear_least_squares_fitting_3() ?


hugoledoux:
## Issue Details

The function linear_least_squares_fitting_3() does not behave as it should, I believe. See a small example:
https://gist.github.com/hugoledoux/08bde5bf82a5aef7398b5a8d53b2c61b

I first define 5 points in 3D, then fit a plane through the first 3, and finally calculate distance from each point to the plane. All distances <1cm and thus we can say the plane is planar.

After that I fit a plane with least-square with the same points, and calculate the distances: ~1.5m for a few, while the plane is planar!

More worryingly, if I define only 3 points (last example in my CPP), and use least-square then the distances are also 1.5m... Observe that not all 3 points give this result, I tried with simple examples and it works fine (eg distances are 0m, as it should be). Someone using my program based on CGAL for several of these tests (https://github.com/tudelft3d/val3dity) said this happens "when there are super long surfaces". A precision issue?

## Source Code

https://gist.github.com/hugoledoux/08bde5bf82a5aef7398b5a8d53b2c61b

## Environment
 
macOS 10.13.1 

with v4.11 installed with brew
$ brew info cgal
cgal: stable 4.11 (bottled)
Computational Geometry Algorithm Library

sgiraudot:
I can't seem to reproduce your bug. Using CGAL 4.11, I get:

```
collinear= 0
plane= -107.932 -413.802 565.784 294927
dist= 6.15546e-14
dist= 2.05182e-14
dist= 4.10364e-14
dist= 0.00504542
dist= 0.00845212
lq-plane= -0.152458 -0.584439 0.796987 416.993
dist= 0.00314225
dist= 0.00313344
dist= 0.000944076
dist= 0.00413228
dist= 0.00319702
lq-plane-3-pts= -0.152185 -0.583462 0.797754 415.849
dist= 6.32177e-06
dist= 6.49187e-06
dist= 1.70105e-07
dist= 0.00504567
dist= 0.00843927
==========
```

PCA is based on computation of eigenvalues which use imprecise double float numbers, which explains why the 3 point planes is not as precise in the last case than in the first (in the first, as you define the plane directly with the 3 points, there is no eigenvalue computation, they are used directly to create the equation of the plane).

For the other values, unless there is something I did not understand from your explanations, I don't see what is the problem.

hugoledoux:
huh, this is interesting. Your results are indeed as they should. 

I get the following:
```
$ ./test
[...]

PR #3096, opened 2018-05-18
Fix some deprecation-related issues

MaelRL:
## Summary of Changes

Most tests that intentionally use deprecated code define `CGAL_NO_DEPRECATION_WARNINGS`, which does not do much to prevent the compilation from failing if `CGAL_NO_DEPRECATION_CODE` is enabled (e.g. [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-6/Mesh_3/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz)).

- A file to disable deprecation warnings/errors is added rather than simply guarding the tests with `#ifndef CGAL_NO_DEPRECATION_CODE` because the test plateform with `CGAL_NO_DEPRECATED_CODE` seems different from others. The code is tested then.
- 4.12 because a lot of deprecated tests were added with 4.12 (`PSP_3`).

Fixed some deprecated code being used along the way.

## Release Management

* Affected package(s): `Convex_hull_d`, `Circular_kernel_2`, `Installation`, `Mesh_3`, `Point_set_processing_3/shape_detection_3`, `Triangulation_3`
* Issue(s) solved (if any): --
* Feature/Small Feature (if any): --

PR #3128, opened 2018-05-29
Solver Interface: deprecate CGAL::Diagonalize_traits

sgiraudot:
## Summary of Changes

Approved [small feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Remove_CGAL::Diagonalize_traits) to remove `CGAL::Diagonalize_traits` (deprecating it first).

## Release Management

* Affected package(s): Solver Interface, Bounding Volumes, Jet Fitting, Mesh 3, Classification, PCA, PSP, Scale Space
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/2676
* pre-approved Laurent Rineau 11:04, 30 May 2018 (CEST)

### DIFF
--- a/Bounding_volumes/doc/Bounding_volumes/CGAL/Approximate_min_ellipsoid_d.h
+++ b/Bounding_volumes/doc/Bounding_volumes/CGAL/Approximate_min_ellipsoid_d.h
@@ -137,6 +137,8 @@ for display in Maplesoft's Maple.
 
 \cgalExample{Approximate_min_ellipsoid_d/ellipsoid_for_maple.cpp} 
 
+\note This class requires the \ref thirdpartyEigen library.
+
 */
 template< typename Traits >
 class Approximate_min_ellipsoid_d {
--- a/Principal_component_analysis/test/Principal_component_analysis/test_linear_least_squares_fitting_circles_2.cpp
+++ b/Principal_component_analysis/test/Principal_component_analysis/test_linear_least_squares_fitting_circles_2.cpp
@@ -1,3 +1,5 @@
+#define CGAL_NO_DEPRECATION_WARNINGS
+
 // Example program for the linear_least_square_fitting function on a set of circles in 2D
 
 #include <CGAL/Cartesian.h>
--- a/Principal_component_analysis/test/Principal_component_analysis/test_linear_least_squares_fitting_circles_2.cpp
+++ b/Principal_component_analysis/test/Principal_component_analysis/test_linear_least_squares_fitting_circles_2.cpp
@@ -1,4 +1,4 @@
-#define CGAL_NO_DEPRECATION_WARNINGS
+#include <CGAL/internal/disable_deprecation_warnings_and_errors.h>
 
 // Example program for the linear_least_square_fitting function on a set of circles in 2D
 
--- a/Bounding_volumes/doc/Bounding_volumes/PackageDescription.txt
+++ b/Bounding_volumes/doc/Bounding_volumes/PackageDescription.txt
@@ -13,6 +13,7 @@
 \cgalPkgSummaryEnd
 \cgalPkgShortInfoBegin
 \cgalPkgSince{1.1}
+\cgalPkgDependsOn{\ref thirdpartyEigen}
 \cgalPkgBib{cgal:fghhs-bv}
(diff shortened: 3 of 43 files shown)


Principal_component_analysis/test/Principal_component_analysis/test_linear_least_squares_fitting_circles_2.cpp
```diff
@@ -1,3 +1,5 @@
+#define CGAL_NO_DEPRECATION_WARNINGS
```
MaelRL: This will create errors on the no-deprecated-code test suite, please see https://github.com/CGAL/cgal/pull/3096.

sgiraudot: Thanks, I fixed it.

Bounding_volumes/doc/Bounding_volumes/CGAL/Approximate_min_ellipsoid_d.h
```diff
@@ -137,6 +137,8 @@ for display in Maplesoft's Maple.
 
 \cgalExample{Approximate_min_ellipsoid_d/ellipsoid_for_maple.cpp} 
 
+\note This class requires the \ref thirdpartyEigen library.
```
lrineau: Should not we also say that, in case the user does not use our CMake scripts, then the macro `CGAL_EIGEN3_ENABLED` must be defined. Maybe that could be added to [the page about Eigen](https://cgal.geometryfactory.com/~sgiraudot/Solver_interface_remove_diagonalize_traits/Manual/installation.html#thirdpartyEigen).

sloriot: I don't think so, the code should explicitly instantiate the code with the traits class using Eigen. The macro definition is here to filter out overloads where the default of the traits is used.

sloriot:
You should check if the installation manual should be updated as the status of Eigen is changing. Also this will impact package maintainers since they will need to add a hard dependency on Eigen (@lrineau as rpm maintainer, do you agree?)

Bounding_volumes/doc/Bounding_volumes/CGAL/Approximate_min_ellipsoid_d.h
lrineau: OK, I understand. Than maybe our CMake scripts should not add `-DCGAL_EIGEN3_ENABLED` everywhere, to test that it compiles correctly.

lrineau:
I think I will add Eigen as a weak dependency, because it is not required by the code of CGAL, but only by a few CGAL package.

sloriot:
The `PackageDescription.txt` files of the packages impacted must be modified to mention the hard dependency to Eigen.

Bounding_volumes/doc/Bounding_volumes/CGAL/Approximate_min_ellipsoid_d.h
sgiraudot: Here there is a problem: there is no possible way to instantiate the code with a traits class or another as it is all internal. So here it all depends on the macro `CGAL_EIGEN3_ENABLED`: if it is not used, then the fallback mode is used and with this PR, the user will get a deprecation warning.

The clean thing I can do is to remove the fallback and only keep Eigen, but that will probably mean error messages for people who do not include Eigen in their CMakeLists.

lrineau:
This PR triggers deprecation warnings in three lines of the testsuite:
```
CGAL-4.13-Ic-39/CGAL_ipelets_Demo/TestReport_lrineau_ArchLinux-clang.gz
CGAL-4.13-Ic-39/Polygon_Demo/TestReport_lrineau_ArchLinux-clang.gz
CGAL-4.13-Ic-39/Principal_component_analysis_Demo/TestReport_lrineau_ArchLinux-clang.gz
```
Either we switch to the new traits class, or we decide to test the deprecated class, but in that case we must disable the warning.

sgiraudot:
The deprecated class is tested elsewhere. Apart from that, we should switch to the new traits class everywhere in CGAL: I will update the code (only the CMakeLists, in fact) to use the new traits class in the demo.

PR #3131, opened 2018-05-29
Use a directory that exists

sloriot:
There is no straight skeleton demo + the pattern described to build a demo of a package is not correct (the target named should be used)


lrineau:
Successfully tested in [CGAL-4.13-Ic-26][*], and then merged (on Friday).

(The test was only about the compilation by Doxygen. I have not check the produced documentation.)

PR #3139, opened 2018-05-31
Mesh_3: fix verbose mode (when the manifold criterion is used)

lrineau:
## Summary of Changes

When `CGAL_MESH_3_VERBOSE` is defined, the code path is different and, before this patch, the manifold criterion was without any effect, resulting in a possible non-manifold output surface mesh.

The fix in itself is https://github.com/CGAL/cgal/commit/335034e3cf49ff0764c0f5d5f20a3e20dbc1f813. This PR also adds two tests for two different code paths:
  - one for `CGAL_MESH_3_VERBOSE`,
  - and one for the use of a loop on `mesher.one_step()`.

## Release Management

* Affected package(s): Mesh_3


lrineau:

PR #3142, opened 2018-06-01
Spatial Searching: bugfix disable_if

sgiraudot:
## Summary of Changes

When using a `Search_traits_adapter` with a `key_type` that is the same as the point type of the `Search_traits` base, there was a compilation error:
```sh
(...)/Spatial_searching/include/CGAL/Fuzzy_sphere.h:161:69: error: no type named
      'type' in 'boost::disable_if<boost::is_same<CGAL::Point_2<CGAL::Simple_cartesian<double> >,
      CGAL::Point_2<CGAL::Simple_cartesian<double> > >, void>'
                                 typename SearchTraits::Point_d> >::type* = 0)
```

The `disable_if` did not work because none of its arguments was template (so it was instantiated anyway). This PR:
- changes the constructor on which `disable_of` is applied: by default, if the adapter and the base have the same type, we enforce passing by the property map (in case this property map does have the same type but is not `CGAL::Identity_property_map`)
- makes this constructor template so that `disable_of` works
- adds the same behavior to the methods of `Search_traits_adapter` that otherwise create invalid overloads

This solves the problem of using `Search_traits_adapter` with the same `key_type` as the point type of `Search_traits` base (for example using `CGAL::Identity_property_map`: in that case it is useless to use the adapter, but the code should compile and work anyway). Tested locally, this does not break any test/example of `Spatial_searching`.

## Release Management

### DIFF
--- a/Spatial_searching/include/CGAL/Fuzzy_sphere.h
+++ b/Spatial_searching/include/CGAL/Fuzzy_sphere.h
@@ -144,14 +144,18 @@ class Fuzzy_sphere< Search_traits_adapter<K,PM,Base_traits> >
   typedef typename Base_traits::FT FT;
 public:
   // constructors
-  Fuzzy_sphere(const SearchTraits& traits_=SearchTraits()):Base(traits_){}
-  Fuzzy_sphere(const typename Base_traits::Point_d& center, FT radius, FT epsilon=FT(0),const SearchTraits& traits_=SearchTraits()) :
-    Base(center,radius,epsilon,traits_) {}
-  Fuzzy_sphere(const typename SearchTraits::Point_d& center, FT radius, FT epsilon=FT(0),
-               const SearchTraits& traits_=SearchTraits(),
+  Fuzzy_sphere(const SearchTraits& traits_=SearchTraits()):Base(traits_){};
+
+  // Constructor for any point type that is not `SearchTraits::Point_d`
+  template <typename Point> // boost::disable_if requires a template argument to work
+  Fuzzy_sphere(const Point& center, FT radius, FT epsilon=FT(0),const SearchTraits& traits_=SearchTraits(),
                typename boost::disable_if<
-                 boost::is_same<typename Base_traits::Point_d,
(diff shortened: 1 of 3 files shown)


MaelRL:
This will probably conflict with #2954. There's no conflict in the changes, just that I fixed the wonky indentation in the other PR.

Also, might be worth testing whatever was thought to work.

lrineau:
#2954 was just merged, and indeed there is a conflict.

sgiraudot:
I rebased and also added a test as Mael suggested.

Spatial_searching/include/CGAL/Fuzzy_sphere.h
```diff
@@ -144,14 +144,18 @@ class Fuzzy_sphere< Search_traits_adapter<K,PM,Base_traits> >
   typedef typename Base_traits::FT FT;
 public:
   // constructors
-  Fuzzy_sphere(const SearchTraits& traits_=SearchTraits()):Base(traits_){}
-  Fuzzy_sphere(const typename Base_traits::Point_d& center, FT radius, FT epsilon=FT(0),const SearchTraits& traits_=SearchTraits()) :
-    Base(center,radius,epsilon,traits_) {}
-  Fuzzy_sphere(const typename SearchTraits::Point_d& center, FT radius, FT epsilon=FT(0),
-               const SearchTraits& traits_=SearchTraits(),
+  Fuzzy_sphere(const SearchTraits& traits_=SearchTraits()):Base(traits_){};
```
MaelRL: the extra ";" sneaked through the rebase.

lrineau:
There are compilation error in old platforms using Boost 1.48.
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-29.shtml#Scale_space_reconstruction_3_Examples

I think you should include the Boost header `<boost/enable_if.hpp>`.

lrineau:
Successfully tested in [CGAL-4.13-Ic-35][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-35.shtml

PR #3144, opened 2018-06-04
Polyhedron demo: Automatically decide between selection_io_plugin or selection_io_sm_plugin

maxGimeno:
## Summary of Changes

With this PR, static functions to access the scene and the mainwindow are added.
It is used to automatically decide if the selection_io_plugin or the selection_io_sm_plugin should be used.

### DIFF
--- a/Polyhedron/demo/Polyhedron/CMakeLists.txt
+++ b/Polyhedron/demo/Polyhedron/CMakeLists.txt
@@ -185,6 +185,11 @@ if(CGAL_Qt5_FOUND AND Qt5_FOUND AND OPENGL_FOUND )
   target_link_libraries(demo_framework
     PUBLIC Qt5::OpenGL Qt5::Widgets Qt5::Gui Qt5::Script
     )
+  # Let's define `three_EXPORT` during the compilation of `demo_framework`,
+  # in addition of `demo_framework_EXPORT` (defined automatically by
+  # CMake). That is to deal with the visibility of symbols of
+  # `Three.h`/`Three.cpp`.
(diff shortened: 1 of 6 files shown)


lrineau:
That patch seems really ugly, because it breaks the separation between the application and the plugin: with this patch, there would be a piece of code in the `MainWindow` that is specific to two plugins. Could not it be fixed in the `canSave` function of the two plugins?!

maxGimeno:
You mean the canLoad() function. Maybe.

maxGimeno:
You need access to the scene for that, which you don't have in an IO plugin, as they don't have an init() function. So the plugin must be turned into a Polyhedron_demo_plugin_interface, which I know you don't like. What's the better option ?

lrineau:
I think it is time to merge into `master` the static accessor to the scene...

maxGimeno:
That's fine by me.

lrineau:
There is still a license issue in Three:
https://travis-ci.org/CGAL/cgal/jobs/388148748
```
.. Test merging of the branch "HEAD"...
Already up to date.
HEAD detached at FETCH_HEAD
nothing to commit, working tree clean
.. Testing permissions...
.. Testing licenses...
=> There are modifications in licenses:
 M Three/package_info/Three/license.txt
(aborting the merge now)
fatal: There is no merge to abort (MERGE_HEAD missing).
```

You can run `./Scripts/developer_scripts/detect_packages_licenses` locally. It should not modify the license files. If you have doubts on the real reason, you can run `./Scripts/developer_scripts/licensecheck` on all the header files of Three.

lrineau:
Does not compile. See https://github.com/CGAL/cgal/pull/3165#issuecomment-396651537.

PR #3148, opened 2018-06-04
Polyhedron demo:Fix_orthographic selection

### LINKED ISSUE #3140: Polyhedron demo: Point Set Selection broken with orthographic projection
## Issue Details

1. Load a point set (i.e. `cube.pwn` in the data directory of the examples of the PSP package)
2. Switch to orthographic projection
3. Point set selection
4. Try to select a part of the point set: the selection does not correspond to the user-drawn rectangle

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): GNU/Linux Mesa
* CGAL version: master
* Boost version:

maxGimeno:
## Summary of Changes

Part of the problem came from the fact that ray shooting must be performed differently in orthographic projection, 
the other part of the problem came from an "optimisation" in the projection code I used to replace gluProject, that assumed the projection matrix is always perspective, and ignored some values. This is why the selection in PSP was buggy, even with no ray shooting.
## Release Management

* Issue(s) solved (if any): fix #3140


lrineau:
Successfully tested in [CGAL-4.13-Ic-37][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-37.shtml

PR #3150, opened 2018-06-04
Polyhedron demo: Remove conflicting 'Point' typedef

afabri:
## Summary of Changes

As `Point` seems to be a type that is defined by VC++ we get for VS2017 with `-permissive -std:c++latest` an  [error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-26/Polyhedron_Demo/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz)

include\CGAL/Delaunay_triangulation_3.h(1413): error C2244: 'CGAL::Delaunay_triangulation_3<Gt,Tds_,CGAL::Default,Lock_data_structure_>::side_of_oriented_sphere': unable to match function definition to an existing declaration

Adding `typedef Point Point_;` and using the latter in the method definition solves the problem.

## Release Management

* Affected package(s): Polyhedron demo.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.h
+++ b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.h
@@ -162,9 +162,6 @@ typedef CGAL::Surface_mesh_deformation<SMesh, CGAL::Default, CGAL::Default, CGAL
   ,CGAL::Default, CGAL::Default, CGAL::Default,
   Array_based_vertex_point_map<SMesh> > Deform_sm_mesh;
 
-typedef Deform_mesh::Point  Point;
-typedef Deform_sm_mesh::Point  SM_Point;
-
 /// For storing associated data with a group of control vertices
 template<typename Mesh>
 class Control_vertices_data
--- a/Surface_mesher/demo/Surface_mesher/volume.cpp
+++ b/Surface_mesher/demo/Surface_mesher/volume.cpp
@@ -769,9 +769,9 @@ void Volume::display_marchin_cube()
       const unsigned int nbt = facets.size() / 9;
       for(unsigned int i=begin;i<nbt;i++)
       {
-        const Point a(facets[9*i],   facets[9*i+1], facets[9*i+2]);
-        const Point b(facets[9*i+3], facets[9*i+4], facets[9*i+5]);
-        const Point c(facets[9*i+6], facets[9*i+7], facets[9*i+8]);
+        const Point_3 a(facets[9*i],   facets[9*i+1], facets[9*i+2]);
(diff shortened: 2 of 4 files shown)


lrineau:
I think MSVC 2017 is wrong about its interpretation of the code. Let's use this workaround.

MaelRL:
It's kind of strange, DT3 and even T3 are riddled with `const Point&` in function arguments.

Could it be disturbed by the `Point` typedef in the global namespace at https://github.com/CGAL/cgal/blob/33f98a35e560a123f29fbcec51eef321e9a9318b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.h#L165 ?

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 5, 2018, 10:32 AM GMT+2](https://github.com/CGAL/cgal/pull/3150#issuecomment-394627131 "2018-06-05T08:32:05Z - Replied by Github Reply Comments"):
> It's kind of strange, DT3 and even T3 are riddled with `const Point&` in function arguments.
> 
> Could it be disturbed by the `Point` typedef in the global namespace at https://github.com/CGAL/cgal/blob/33f98a35e560a123f29fbcec51eef321e9a9318b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh_deformation/Scene_edit_polyhedron_item.h#L165 ?

That is certainly due to that typedef, but the global scope should be considered last, in the lookup, and typedef in the class scope should be found before. No?

Anyway, maybe the workaround could be used inside `Scene_edit_polyhedron_item.h` instead of modifying `Delaunay_triangulation_3`.

afabri:
You are right with your observation. And the `typedef` can be removed as it is not used. 

The fact that a simple `struct Point { double x,y,z;};`  added in the file `Triangulation_3/test/Triangulation_3/test_delaunay_3.cpp` produces the same error I think it would still make sense to keep my workaround.

I will boil it down to  a minimal piece of code.

lrineau:
[**@lrineau**](https://github.com/lrineau) commented on [Jun 5, 2018, 11:09 AM GMT+2](https://github.com/CGAL/cgal/pull/3150#issuecomment-394638373 "2018-06-05T09:09:55Z - Replied by Github Reply Comments"):
> the global scope should be considered last, in the lookup, and typedef in the class scope should be found before. No?

See [that StackOverflow question](https://stackoverflow.com/q/31690353/1728537), and its answer. It explains why in the following declaration, the two `Type` names can be different types:
```C++
Type Exercise::setVal(Type parm)
```
The answer refers to the paragraph 3.4.1/8 of the norm. The current draft is (in LaTeX):
https://github.com/cplusplus/draft/blob/57de3af9d24037d60d996b255c1ded8ee65387d2/source/basic.tex#L1387-L1455

lrineau:

PR #3156, opened 2018-06-06
Polyhedron demo: degree sphere plugin

maxGimeno:
## Summary of Changes
This adds a plugin that displays spheres with color depending on the degree of the corresponding vertex on the targeted surface_mesh. It is possible to change the color associated with a degree, and to move the spheres.


maxGimeno:
@janetournois Here is my POC, I hope it does what you want. We'll have to discuss how to connect this to your work, if the interface suits you.

lrineau:
@janetournois @maxGimeno, do we agree that this PR should be canceled?

PR #3157, opened 2018-06-06
Surface mesher demo: Another problem with VS2017 -permissive-

afabri:
## Summary of Changes

Similar to #3150  only with `Facet` instead of `Point`.

## Release Management

* Affected package(s): Surface_mesher demo


lrineau:
Successfully tested in [CGAL-4.13-Ic-35][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-35.shtml

PR #2804, opened 2018-02-07
IO: Add IO functions for WKT format

maxGimeno:
## Summary of Changes

This PR adds IO functions to deal with WKT
It also makes some clean-up in the GraphicsView demos.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #1590
* Feature/Small Feature (if any): [`Features/Small_Features/WKT_Support_For_2D_Structures`](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/WKT_Support_For_2D_Structures)
* Pre-approved. @gdamiand  17:02, 26 June 2018 (CEST)

PR #3169, opened 2018-06-13
Polyhedron_IO: Improve STL reader

### LINKED ISSUE #3119: STL file input issue
## Issue Details

Looking at the model database https://ten-thousand-models.appspot.com, most of the models load perfectly, but there are some `.stl` files that can't seem to be loaded in the demo (it just holds forever). For example, [this one](https://ten-thousand-models.appspot.com/detail.html?file_id=36069).

## Environment

* CGAL version: master

MaelRL:
## Summary of Changes

CGAL's STL reader was failing to read some `.stl` files from some data sets such as those from https://ten-thousand-models.appspot.com/. This PR tries to improve that situation (and it can now read those from that particular data set).

## Release Management

* Affected package(s): `Polyhedron_IO`
* Issue(s) solved (if any): fix #3119


MaelRL:
Marked as "not yet approved" because there were talks of documenting this function. However, it is a bit unclear where it should be put, since Polyhedron_IO is not a real package. In `PMP` because they only yield a polygon soup? Possibly in `Stream_support` that would sort of become generalized into some `CGAL_IO`? It seems a bit weird but WKT read/write functions are being put there currently (https://github.com/CGAL/cgal/pull/2804 -- admittedly not approved yet).

Similarly, there is `read_OFF` that is used in examples (for example in PMP) but is not documented.

lrineau:
It seems nobody stands up to talk about the documentation of this function. Anyway now that is too late for a small feature for CGAL-4.13. So, my proposal is to have this PR integrated as it is, in `integration` and then `master`. And once you agree on the documentation, submit a small feature for CGAL-4.14.

lrineau:
Successfully tested in [CGAL-4.13-Ic-47][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-47.shtml

PR #3173, opened 2018-06-15
Interpolation: replace usages of Functor::result_type and Functor::argument_type

MaelRL:
## Summary of Changes

Using `::result_type` and `::argument_type` prevents from passing more generic value/gradient functors (such as lambdas).

- `::result_type` can be easily replaced by `result_of`.
- `::argument_type`, which was needed to distinguish between two types (say `T1` & `T2`), is a bit trickier. It is replaced by some SFINAE to check whether one can construct a `std::function<ret(T1/2)>` or not. [Only in C++11]

## Release Management

* Affected package(s): `Interpolation`

### DIFF
--- a/Interpolation/doc/Interpolation/CGAL/sibson_gradient_fitting.h
+++ b/Interpolation/doc/Interpolation/CGAL/sibson_gradient_fitting.h
@@ -86,8 +86,7 @@ function \ref PkgInterpolationNaturalNeighborCoordinates2.
 \tparam OutputFunctor must be a functor with argument type `std::pair<Dt::Vertex_handle, Traits::Vector_d>`.
         Note that the result type of the functor is not specified and is chosen by users to fit their needs.
 \tparam ValueFunctor must be a functor where:
-- `ValueFunctor::argument_type` must be either `std::pair<Dt::Vertex_handle, Dt::Geom_traits::FT>`
-or `std::pair<Dt::Point, Dt::Geom_traits::FT>`.
+- `ValueFunctor::argument_type` must be either `Dt::Vertex_handle` or `Dt::Point`.
 - `ValueFunctor::result_type` is a pair of the function value type and a Boolean.
 The function value type must provide a multiplication and addition operation with the type
 `Traits::FT` as well as a constructor with argument `0`.
@@ -124,8 +123,7 @@ functions \ref PkgInterpolationRegularNeighborCoordinates2.
 \tparam OutputFunctor must be a functor with argument type `std::pair<Rt::Vertex_handle, Traits::Vector_d>`.
         Note that the result type of the functor is not specified and is chosen by users to fit their needs.
 \tparam ValueFunctor must be a functor where:
-- `ValueFunctor::argument_type` must be either `std::pair<Rt::Vertex_handle, Rt::Geom_traits::FT>`
-or `std::pair<Rt::Weighted_point, Rt::Geom_traits::FT>`.
+- `ValueFunctor::argument_type` must be either `Rt::Vertex_handle` or `Rt::Weighted_point`.
 - `ValueFunctor::result_type` is a pair of the function value type and a Boolean.
 The function value type must provide a multiplication and addition operation with the type
 `Traits::FT` as well as a constructor with argument `0`.
--- a/Interpolation/examples/Interpolation/interpolation_vertex_with_info_2.cpp
+++ b/Interpolation/examples/Interpolation/interpolation_vertex_with_info_2.cpp
@@ -94,6 +94,7 @@ int main()
 
   Delaunay_triangulation T;
 
+  // Note that a supported alternative to creating the functors below is to use lambdas
   Value_function<Vertex_handle, Coord_type> value_function;
   Gradient_function<Vertex_handle, Vector> gradient_function;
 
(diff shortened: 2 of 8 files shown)


lrineau:
Travis says:
```
[  8%] Building CXX object CMakeFiles/sibson_interpolation_vertex_with_info_2.dir/sibson_interpolation_vertex_with_info_2.cpp.o
In file included from /home/travis/build/CGAL/cgal/Interpolation/examples/Interpolation/sibson_interpolation_vertex_with_info_2.cpp:9:
/home/travis/build/CGAL/cgal/Interpolation/examples/Interpolation/../../include/CGAL/sibson_gradient_fitting.h:35:10: fatal error: 
      'boost/type_traits/is_constructible.hpp' file not found
#include <boost/type_traits/is_constructible.hpp>
         ^
1 error generated.
```
https://travis-ci.org/CGAL/cgal/jobs/392666080#L1834

Maybe that was introduced only in recent Boost versions?

lrineau:
https://svn.boost.org/trac10/ticket/8802

That was added in Boost 1.55.

MaelRL:
Looks more like 1.60 using https://www.boost.org/doc/libs/1_60_0/libs/type_traits/doc/html/boost_typetraits/reference/is_constructible.html.

is_copy_constructible does arrive with 1.55 (https://www.boost.org/doc/libs/1_55_0/libs/type_traits/doc/html/boost_typetraits/reference/is_copy_constructible.html).

MaelRL:
I see 
```
Compiler Compatibility: This trait requires the C++11 features decltype variadic templates and SFINAE-expression support for full support. While there is some fallback code for cases where this is not the case, the trait should really be considered broken in that case. The header will define the macro BOOST_TT_IS_CONSTRUCTIBLE_CONFORMING when the full implementation is available.
```
on that page that I hadn't seen. That's not too good.

lrineau:
Maybe the new way could be protected by the macro `CGAL_CXX11`, and then you have `decltype`, `std::function`, and `std::is_constructible`.

MaelRL:
If I do that, it won't work with functions that do not define ::argument_type when using cpp98.

I could just not use SFINAE and `is_constructible` and simply use implicit conversions to `CGAL::cpp11::function` from whatever is passed to me. I have then something like
```
void neighbors(..., cpp11::function<boost::any(Vertex_handle)> f, ...) { ... }
void neighbors(..., cpp11::function<boost::any(Point)> f, ...) { ... }
```
and I can call from whatever
```
void my_function(const Point_3&) { }
return neighbors(..., [](Vertex_handle)->double { return 0;}, ...  );
return neighbors(..., my_function, ...  );
etc.
```

 I'm not sure what's the overhead cost of that, though...

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 21, 2018, 10:34 AM GMT+2](https://github.com/CGAL/cgal/pull/3173#issuecomment-399021535 "2018-06-21T08:34:04Z - Replied by Github Reply Comments"):
> If I do that, it won't work with functions that do not define ::argument_type when using cpp98.

It just means that, for users using C++98/03, the behavior will be the same as the current one: `result_type` and `argument_type` are mandatory.

MaelRL:
Well, the point of this PR was to get rid of `argument_type` and `result_type` regardless of C++11.

It's doable even without adding overhead using a dummy functor like in https://gist.github.com/MaelRL/0356eef441d1c40359ceca861332d1fc; this works fine for `std::function`... but not for `boost::function`, which gives an ambiguous call error...

lrineau:
[**@MaelRL**](https://github.com/MaelRL) commented on [Jun 21, 2018, 12:53 PM GMT+2](https://github.com/CGAL/cgal/pull/3173#issuecomment-399060548 "2018-06-21T10:53:53Z - Replied by Github Reply Comments"):
> Well, the point of this PR was to get rid of `argument_type` and `result_type` regardless of C++11.
> 
> It's doable even without adding overhead using a dummy functor like in https://gist.github.com/MaelRL/0356eef441d1c40359ceca861332d1fc; this works fine for `std::function`... but not for `boost::function`, which gives an ambiguous call error...

I disagree. The goal was to fix the issue of the CGAL user, wanting to use a C++11 lambda expression instead of a functor. The nested type `result_type` can easily be replaced by `result_of` (from boost or C++11), but the for the `argument_type` that is difficult without C++11, as we saw.

The removal of `argument_type` was not the goal, but if you found a solution that works fine with both C++03 and C++11, that is good.

lrineau:
[**@lrineau**](https://github.com/lrineau) commented on [Jun 21, 2018, 1:00 PM GMT+2](https://github.com/CGAL/cgal/pull/3173#issuecomment-399062609 "2018-06-21T11:00:19Z - Replied by Github Reply Comments"):
> The removal of `argument_type` was not the goal, but if you found a solution that works fine with both C++03 and C++11, that is good.

Did you find a solution for that issue? The code freeze is in four days, and I would like now to prepare the list of pull-requests that will make into CGAL-4.13, as opposed to those that will be postponed to CGAL-4.14.

MaelRL:
No, I tried some other things but there always were some limitations. I have the commit for the workaround with CGAL_CXX11 ready but I just had not pushed, I will do it now.

lrineau:
Successfully tested in [CGAL-4.13-Ic-54][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-54.shtml

PR #3187, opened 2018-06-22
Polyhedron Demo: Fix edge selection

maxGimeno:
## Summary of Changes

Use a shader for wide lines to replace the deprecated glLineWidth() function that doesn't work on multiple platforms. Also fix the snapshots of selection items.
## Release Management


janetournois:
That's a lot better!

I have this kind of pollution in the console though
```
GL ERROR : (APISource : ErrorType)
 Error has been generated. GL error GL_INVALID_OPERATION in DisableVertexAttribArray: (ID: 3860911800) Generic error
```

lrineau:
Maxime, the error says there is an error in a use of `glDisableVertexAttribArray`. Can you see where it is?

maxGimeno:
My guess is it relates to the "color" attribute, but it should only be outputed in the initializeBuffers(), not in draw. Here they say this error comes form a VAO not bound, but all VAOs are bound here. 
In the beginning of the log, it is said that the version number in the shaders are deprecated. Maybe upgrading the shaders to GLSL 140 (opengl 3.1) will be sufficient to fix this. We'll see.

lrineau:
@maxGimeno Please merge `master` to fix the non-trivial conflicts. I would like to test this PR.

maxGimeno:
What for ? It is not ready. It should go into the fix for windows, that will put the base context to opengles2

lrineau:
Is this PR integrated in #3198, and should be canceled?

maxGimeno:
Yes it is.

PR #3190, opened 2018-06-26
Fix non-documented IO functions

### LINKED ISSUE #3189: write_off on non-garbage collected Surface_mesh objects
`write_off()` is using `num_vertices()` and `num_faces()` which leads to errors if the surface mesh contains garbage.

sloriot:
## Summary of Changes

Call documented functions in convenience functions.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #3189

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/IO.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/IO.h
@@ -327,45 +327,14 @@ bool read_off(Surface_mesh<K>& mesh, const std::string& filename)
 template <typename K>
 bool write_off(const Surface_mesh<K>& mesh, const std::string& filename)
 {
-    typedef Surface_mesh<K> Mesh;
-    typedef typename Mesh::Point Point_3;
-
-    FILE* out = fopen(filename.c_str(), "w");
-    if (!out)
+    std::ofstream out(filename.c_str());
+    if (out.fail())
         return false;
 
+    out << std::setprecision(17);
+    write_off(out, mesh);
 
-    // header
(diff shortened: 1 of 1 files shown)


Surface_mesh/include/CGAL/Surface_mesh/IO.h
```diff
@@ -267,51 +267,12 @@ bool read_off_ascii(Surface_mesh<Point_3>& mesh,
...
-    }
-
-    // read as ASCII or binary
-    bool ok = (is_binary ?
-               internal::read_off_binary(mesh, in, has_normals, has_texcoords) :
-               internal::read_off_ascii(mesh, in, has_normals, has_texcoords));
+  std::ifstream in(filename.c_str());
+  if (in.fail())
+    return false;
+  in >> mesh;
```
lrineau: The new code certainly cannot read binary file (because of the `ifstream` open mode. That seems to be a regression.

sloriot: Indeed. What should be done here? The real bug was only in the write function but I though that maybe we could get rid of duplicated code. I can simply undo the change for the read function.

lrineau: I think you should indeed undo the change for the read function, otherwise the demo will no longer read the colors of colored OFF.

sloriot: Amended

sloriot:
Successfully tested in CGAL-4.13-Ic-57

PR #3203, opened 2018-07-04
Polyhedron_demo: Fix triangulate_convex_facet in Scene_surface_mesh_item

### LINKED ISSUE #3202: Polyhedron demo: `std::bad_alloc` when loading Surface Mesh
## Issue Details

When loading the following one-facet OFF file:
```
OFF
5 1 0
-219.189358781456065 163.382991740848979 37.9785643023864168
-206.982635130200748 132.554175983545463 37.9785643023864168
-247.535946042803033 98.8178673201488209 37.9785643023864168
-248.009216608833725 99.386770355796898 37.9785643023864168
-267.685527043317165 123.039018224108759 37.9785643023864168
5 0 1 2 3 4
```

The program runs for a while and then returns a `std::bad_alloc`. This happens with `surface_mesh_io_plugin`, _not_ with `off_plugin` which loads the facet fine (so it's specific to the surface mesh item, not the polyhedron one).

From what I understand, the program goes to an infinite loop in the method `triangulate_convex_facet()` of the surface mesh item. Using `triangulate_facet()` instead works, but as @maxGimeno pointed out to me, the method is then suboptimal.

Note that the bug-generating facet seems to have a three aligned vertices, so it's exactly on the limit of being non-convex (don't know if this could be a trigger of this error).

(@janetournois maybe you can also help on this issue?)

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Debian 64 bits
* Compiler: clang
* CGAL version: master

maxGimeno:
## Summary of Changes

A while loop had a typo, making it infinite.

## Release Management

* Issue(s) solved (if any): fix #3202

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
@@ -832,7 +832,7 @@ void Scene_surface_mesh_item_priv::triangulate_convex_facet(face_descriptor fd,
   SMesh::Halfedge_around_face_circulator he(halfedge(fd, *smesh_), *smesh_);
   SMesh::Halfedge_around_face_circulator he_end = he;
   
-  while(next(*he, *smesh_) != prev(*he, *smesh_))
+  while(next(*he, *smesh_) != prev(*he_end, *smesh_))
   {
     ++he;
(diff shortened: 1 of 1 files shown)


sloriot:
Successfully tested in CGAL-4.13-Ic-57

PR #3213, opened 2018-07-09
Stitch bug fix 

sloriot:
Do not delete twice a vertex in case it was already updated


lrineau:
Successfully tested in [CGAL-4.13-Ic-60][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-60.shtml

PR #3218, opened 2018-07-10
Polyhedron Demo: Fix path to external plugins

### LINKED ISSUE #3204: Polyhedron demo: various problems with POLYHEDRON_DEMO_PLUGINS_PATH
There are several problems, when a plugin is loaded from a directory specified by the environment variable `POLYHEDRON_DEMO_PLUGINS_PATH`.

1. We assume that the value of `POLYHEDRON_DEMO_PLUGINS_PATH` is a string, that encodes a list, with `:` as the separator character. On Windows that is obviously wrong, because all paths start with `C:/` or something similar. We need to use [`QDir::listSeparator()`] instead of hard-coding `:`...
  > [static] QChar QDir::listSeparator()
  > 
  > Returns the native path list separator: ':' under Unix and ';' under Windows.
  > 
  > This function was introduced in Qt 5.6.
  
  That is only since Qt 5.6, so we will have to use preprocessor conditionals to continue hardcoding `:`  for Qt<5.6.

2. When an external plugin (in an external directory) has a shared library as a dependency (a `.dll` or a `.so` file), then that dependency is not found by the dynamic loader, and the plugin does not load.

[`QDir::listSeparator()`]: http://doc.qt.io/qt-5/qdir.html#listSeparator

maxGimeno:
Choose etween ':' and ':' as separator in the path to the plugins dir and add it to the global PATH if running on Windows so the potential plugin dependency dlls are found by the linker.
## Release Management

* Issue(s) solved (if any): fix #3204

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -631,6 +631,11 @@ void MainWindow::loadPlugins()
   #endif
 #endif
   if(!env_path.isEmpty()) {
+#if defined(_WIN32)
+    QString path = qgetenv("PATH");
+    QByteArray new_path = path.append(env_path.prepend(separator)).toUtf8();
+    qputenv("PATH", new_path);
+#endif
     Q_FOREACH (QString pluginsDir,
                env_path.split(separator, QString::SkipEmptyParts)) {
       QDir dir(pluginsDir);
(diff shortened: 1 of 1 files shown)


lrineau:
Successfully tested in [CGAL-4.13-Ic-60][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-60.shtml

PR #3219, opened 2018-07-10
Add CGALConfig.cmake to the Windows installer

lrineau:
## Summary of Changes

Add `CGALConfig.cmake` to the Windows installer. The real `CGALConfig.cmake` is in `lib/cmake/CGAL/`, but that is more convenient if the redirection is also installed by the Windows installer.

## Release Management

* Affected package(s): windows installer

### DIFF
--- a/wininst/developer_scripts/script_cgal.nsi
+++ b/wininst/developer_scripts/script_cgal.nsi
@@ -164,6 +164,7 @@ Section "!Main CGAL" MAIN_Idx
 
   SetOutPath "$INSTDIR"
   File "${CGAL_SRC}\AUTHORS"
+  File "${CGAL_SRC}\CGALConfig.cmake"
   File "${CGAL_SRC}\CHANGES.md"
   File "${CGAL_SRC}\CMakeLists.txt"
   File "${CGAL_SRC}\INSTALL.md"


afabri:
I would suggest that we just fix the 4.12 installer in the download area.

lrineau:
Currently [the documentation says]:
> **7.3.2 Header-only without CMake Configuration**
> 
> Since CGAL 4.12, CGAL can be used in header-only mode, without even configuring CGAL. Programs using CGAL (examples, tests, demos, etc.) must be directly configured using CMake. In this case, CGAL will be configured at the same time. The variable CGAL_DIR must point to the root directory of the CGAL source code (either the root of the unpacked release tarball, or the root of the Git working directory).
> 
> So, using CGAL becomes now:
> ```
> cd /path/to/your/code # go to the directory of the code source using CGAL
> cmake -DCGAL_DIR=<CGAL-root> .
> ```

I think that, instead of `cmake -DCGAL_DIR=<CGAL-root>`, we should say: `cmake -DCGAL_DIR=<CGAL-root>/lib/cmake/CGAL`, because the later will also be fine for an installed version of CGAL (either with package manager or with the Windows installer), whereas the former is only good for a tarball or Git.

[the documentation says]: https://doc.cgal.org/4.12/Manual/installation.html#title16

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Jul 10, 2018, 4:40 PM GMT+2](https://github.com/CGAL/cgal/pull/3219#issuecomment-403847271 "2018-07-10T14:40:42Z - Replied by Github Reply Comments"):
> I would suggest that we just fix the 4.12 installer in the download area.

That bug is really a minor bug, with a trivial workaround. As I said in my previous message, the CGAL documentation should be modified.

And trying to launch `nsis` by hand to recreate a Windows installer for CGAL-4.12 will be a lot of work, while our release scripts will do it for free for CGAL-4.12.1 and CGAL-4.13. That is why I prefer to wait.

sloriot:
@lrineau it seems that the issue is still here in 4.13-beta1 (see [this post](https://stackoverflow.com/questions/52014224/configure-of-cgal)). @afabri could you give it a try?

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Aug 26, 2018, 3:06 PM GMT+2](https://github.com/CGAL/cgal/pull/3219#issuecomment-416037749 "2018-08-26T13:06:48Z - Replied by Github Reply Comments"):
> @lrineau it seems that the issue is still here in 4.13-beta1 (see [this post](https://stackoverflow.com/questions/52014224/configure-of-cgal)). @afabri could you give it a try?

I have tried the Windows installer with Wine, on my Linux machine, and the `CGALConfig.cmake` at the root of the installation of `C:\dev\CGAL-4.13-beta1\` is present. Why did you think the issue was still there?

sloriot:
Somebody complained on SO and cgal-discuss that the file was missing.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Aug 27, 2018, 10:20 AM GMT+2](https://github.com/CGAL/cgal/pull/3219#issuecomment-416152124 "2018-08-27T08:20:04Z - Replied by Github Reply Comments"):
> Oh my bad, it is another one:
> `D:/dev/CGAL-4.13-beta1/lib/cmake/CGAL/CGALConfig.cmake`

Indeed, the `lib/` directory is missing.

afabri:
@sloriot is there still a need to try something?

sloriot:
I don't think so.

PR #3220, opened 2018-07-11
Polyhedron demo: Fix normal facet computation for triangulation

sgiraudot:
## Summary of Changes

When loading a polygon soup with non-triangular facets, these facets are triangulated. For triangulation, the points are projected in 2D after computing the normal vector of the facet.

In the code, the normal is computed using the cross product of the 3 first points of the facet. But it can happen that these 3 points are aligned, leading to an infinite normal and to an assertion failed in the triangulation construction.

This PR tests other combinations of vertices to find the first non-collinear triple. Tested locally, it fixes the bug I had with a specific dataset with aligned vertices.

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
@@ -268,7 +268,6 @@ Scene_polygon_soup_item_priv::triangulate_polygon(Polygons_iterator pit, int pol
     if (normal == CGAL::NULL_VECTOR) // No normal could be computed, return
       return;
 
-    normal = normal / std::sqrt(normal * normal);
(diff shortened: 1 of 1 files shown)


Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
```diff
@@ -249,10 +249,25 @@ void
...
+       {
+          normal = CGAL::cross_product(pb-pa, pc -pa);
+          break;
+       }
+    }
+
+    if (normal == CGAL::NULL_VECTOR) // No normal could be computed, return
+      return;
+
     normal = normal / std::sqrt(normal * normal);
```
lrineau: No need to normalize the normal: it is used only in the projection traits. The only important thing is that it must not be zero, and not coplanar with the triangulation.

PR #3221, opened 2018-07-11
Polyhedron Demo: minor fixes and enhancements

maxGimeno:
## Summary of Changes
This PR
- fixes the behavior of the camera copy/paste function
- Reset the viewer's scene bbox when zooming to an item (from the context menu)
- Makes the transparency propagable to all selected items

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -1871,9 +1871,9 @@ void MainWindow::on_actionLookAt_triggered()
   if( i == QDialog::Accepted &&
       dialog.has_correct_coordinates() )
   {
-    viewerShow((float)dialog.get_x(),
-               (float)dialog.get_y(),
-               (float)dialog.get_z());
+    viewerShow((float)dialog.get_x()+viewer->offset().x,
+               (float)dialog.get_y()+viewer->offset().y,
+               (float)dialog.get_z()+viewer->offset().z);
   }
 }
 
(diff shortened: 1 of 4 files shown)


MaelRL:
Thanks!

In minor stuff that I've also just noticed, the "Select All Items" selects from first inserted to last inserted, but if you change the respective positions of the items in the list (with the arrows), then it does not select all the items anymore. For example, if you load `1-10.off`, close the 5th, re-open the 5th and move it to the 5th position in the list, if you Ctrl+A, you get items 1-5 and not 1-10.

maxGimeno:
For the polylines there is already a PR fixing that.
I'll look into the ctrl+A stuff.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -857,7 +857,7 @@ void MainWindow::updateViewerBBox(bool recenter = true)
   double l_dist = (std::max)((std::abs)(bbox_center.x - viewer->offset().x),
                       (std::max)((std::abs)(bbox_center.y - viewer->offset().y),
                           (std::abs)(bbox_center.z - viewer->offset().z)));
-  if((std::log2)(l_dist) > 13.0 )
+  if((std::log2)(l_dist) > 5.0 )
```
lrineau: Why?

MaelRL:
Adding to minor things, the visualization of element IDs could use some looking into, it has some really behavior when you zoom (just take any mesh, show IDs, and zoom with the scrollwheel).

maxGimeno:
You mean when the camera is inside the mesh ?

MaelRL:
Yes, or on the other side, facing away. Basically it seems that the IDs of all the elements are shown, even those behind the camera.

maxGimeno:
I added a test to check the TextItem is not behind the nearPlane, it seems to work great.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -1337,8 +1345,48 @@ void MainWindow::showSceneContextMenu(const QPoint& p) {
         QMenu menu;
         Q_FOREACH(QString name, menu_actions.keys())
         {
-          QAction* action = menu.addAction(name);
-          connect(action, &QAction::triggered, this, &MainWindow::propagate_action);
+          if(name == QString("alpha slider"))
+            continue;
+          if(name == QString("Alpha value"))
+          {
+            QWidgetAction* sliderAction = new QWidgetAction(&menu);
```
lrineau: Missing `#include <QWidgetAction>` in `MainWindow.cpp`. Does it really compile on your machine? It fails on all test platforms:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-62.shtml#Polyhedron_Demo

maxGimeno: Yes it really compiles on my machine, I wonder why. I guess my include tree in built differently or something.

PR #3224, opened 2018-07-11
Fix GetGeomTraits

sloriot:
It was ignoring the fact that a vpm could have been given by the user


lrineau:
Successfully tested in [CGAL-4.13-Ic-63][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-63.shtml

PR #3116, opened 2018-05-25
Extreme points

kkatrio:
## Summary of Changes

convex_hull_3 overload which voids the construction of a polyhedron.
* Affected package(s):  convex_hull_3
* Small Feature:

PR #3226, opened 2018-07-12
Convex_hull_3: Make it minimal.

### LINKED ISSUE #3192: Convex_hull_3 is not minimal
## Issue Details

The function CGAL::convex_hull_3(InputIterator first, InputIterator beyond, Polyhedron_3& polyhedron,  const Traits& traits) does not create a minimal convex_hull, in the sense that some points are not extreme. 
## Source Code
call it on the cross.off, the result contains 20 vertices, where only 16 are expected.
## Environment

* CGAL version: ALL OF THEM

maxGimeno:
## Summary of Changes

This PR adds some initialization steps to the convex_hull_3 function so it returns a minimal hull. 
For this, it also adds a need for Has_on_negative_side_3 to the ConvexHullTraits_3 Concept.
## Release Management

* Affected package(s):Convex_hull_3
* Issue(s) solved (if any): fix #3192 

### DIFF
--- a/Convex_hull_3/doc/Convex_hull_3/Concepts/ConvexHullTraits_3.h
+++ b/Convex_hull_3/doc/Convex_hull_3/Concepts/ConvexHullTraits_3.h
@@ -84,13 +84,6 @@ Predicate object type that provides
 `q` is on the positive side of the halfspace `h`. 
 */ 
 typedef unspecified_type Has_on_positive_side_3; 
-  
-  /*!
-  Predicate object type that provides 
-  `bool operator()(Plane_3 h, Point_3 q)`, which determines if the point 
-  `q` is on the negative side of the halfspace `h`. 
-  */ 
-  typedef unspecified_type Has_on_negative_side_3; 
 
 /*!
 Predicate object type that provides 
--- a/Convex_hull_3/doc/Convex_hull_3/Concepts/ConvexHullTraits_3.h
+++ b/Convex_hull_3/doc/Convex_hull_3/Concepts/ConvexHullTraits_3.h
@@ -84,6 +84,13 @@ Predicate object type that provides
 `q` is on the positive side of the halfspace `h`. 
 */ 
 typedef unspecified_type Has_on_positive_side_3; 
+  
+  /*!
+  Predicate object type that provides 
+  `bool operator()(Plane_3 h, Point_3 q)`, which determines if the point 
+  `q` is on the negative side of the halfspace `h`. 
+  */ 
+  typedef unspecified_type Has_on_negative_side_3; 
 
 /*!
 Predicate object type that provides 
--- a/Convex_hull_3/include/CGAL/convex_hull_3.h
+++ b/Convex_hull_3/include/CGAL/convex_hull_3.h
@@ -715,56 +715,6 @@ ch_quickhull_polyhedron_3(std::list<typename Traits::Point_3>& points,
   
 }
 
-template <class P3_iterator, 
-          class Point_3_list>
-void init_iterators(P3_iterator& minx, 
-                    P3_iterator& maxx, 
-                    P3_iterator& miny,
-                    const P3_iterator& start,
-                    const Point_3_list& points)
-{
-  P3_iterator it = start;
-  for(; it != points.end(); ++it){
-    if(it->x() < minx->x())  minx = it;
-    else if(it->x() ==  minx->x())
-    {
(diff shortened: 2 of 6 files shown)


Convex_hull_3/doc/Convex_hull_3/Concepts/ConvexHullTraits_3.h
```diff
@@ -84,6 +84,13 @@ Predicate object type that provides
 `q` is on the positive side of the halfspace `h`. 
 */ 
 typedef unspecified_type Has_on_positive_side_3; 
+  
+  /*!
+  Predicate object type that provides 
+  `bool operator()(Plane_3 h, Point_3 q)`, which determines if the point 
+  `q` is on the negative side of the halfspace `h`. 
+  */ 
+  typedef unspecified_type Has_on_negative_side_3;
```
lrineau: Now the traits requires both `Has_on_positive_side_3` and `Has_on_negative_side_3`. I suggest that both are replaced by [`Oriented_side_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1OrientedSide__3.html) that provides:
> `Oriented_side operator() (const Kernel::Plane_3 &h, const Kernel::Point_3 &p)`

maxGimeno: @sloriot, do you agree ?

sloriot: No, I'm working on it

Convex_hull_3/include/CGAL/convex_hull_3.h
```diff
@@ -893,8 +936,20 @@ void convex_hull_3(InputIterator first, InputIterator beyond,
...
   // result will be a polyhedron
-  internal::Convex_hull_3::ch_quickhull_polyhedron_3(points, point1_it, point2_it, point3_it,
-                                                     polyhedron, traits);
+  P3_iterator minx, maxx, miny, it;
+  minx = maxx = miny = it = points.begin();
+  ++it;
+  
+  
+  //take extreme points to begin with.
+  internal::Convex_hull_::init_iterators(minx, maxx, miny, it, points);
```
lrineau: compilation error in Travis:
https://travis-ci.org/CGAL/cgal/jobs/403155410#L1983

The minimal testing after such a change is running the test suite of Convex_hull_3!

sloriot:
I think the patch is correct.

I focus only on the case when the convex hull is 3D as otherwise the 2D algorithm is used. The algorithm works as follow: 4 points with a high chance to be on the convex hull are first selected and an initial convex hull is built using these 4 points (these points are not coplanar). Then, iterating over the 4 faces of the convex hull, we select points that are on the positive side of the supporting plane of each face and assign those points to the face. The set of faces with points assigned are collected in the set of so-called pending faces. Then the algorithm simply picks one pending face from the set while it is not empty. For each picked face, it looks for the point assigned to the face that is the furthest from the supporting plane of the face. Then this point is used in `find_visible_set()` to look for all the faces visible (on or on the positive side) from the further point, by diffusion starting from the current face.  Visible faces are removed from the set of pending faces and new faces are created by staring the hole of faces removed using the furthest point. All the points assigned to the faces removed must then be assigned to the new faces like at initialization, and potentially new pending faces  are added to the set.

The reason why only `find_visible_set()` needs to be modified (by changing `has_on_positive_side()` to `!has_on_negative_side()`) is that it is in this function that a face is removed from the current convex hull. The `has_on_positive_side()` test used to assigned a point to a current convex hull face does not need (and must not use) the boundary condition because if a point that is on the convex hull is coplanar with a current face, then it has to be on the positive side of another face (otherwise it is not on the convex hull). Then the current face, it is coplanar with will be invalidated at a future call to `find_visible_set()` using the non-strict predicate version.

sloriot:
By reworking the order of the points when creating the plane, I could get rid of the `has_on_negative()` so no change in the traits is required.

lrineau:
This PR conflicts with #3116. Which one should be tested and integrated first?

sloriot:
They are independent, but this one being smaller it is easier to start by this one.

lrineau:
Successfully tested in [CGAL-4.13-Ic-66][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-66.shtml

PR #3230, opened 2018-07-13
Polygon: Fix `filter_collinear_points()` early return

MaelRL:
## Summary of Changes

When the range of points was of size 3, nothing was returned.

Also, because this is meant to be filtering points in a polygon, added some preconditions:
- there must be at least three points.
- there must be at least three non-collinear points.

## Release Management

* Affected package(s): `Polygon`


lrineau:
Successfully tested in [CGAL-4.13-Ic-63][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-63.shtml

PR #3234, opened 2018-07-17
Fix copy_face_graph with target mesh having garbage

sloriot:
The algorithm relies on a feature that is not always true if the target mesh has garbage. We make sure the feature is valid.


lrineau:
You should run the BGL testsuite. It's pretty bad:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-67.shtml#BGL_LCC_Examples

lrineau:

PR #3240, opened 2018-07-18
Build the kd-tree only if the AABB-tree is built

sloriot:
This avoid issues when using thread where the kd-tree is used while being reconstructed.

The issue I observed was that when `squared_distance()` was called by several threads but the tree was not already built and `accelerate_distance_computations()` has been called, some threads were looking in the kd-tree built in `best_hint()` while another one was actually rebuilding it.

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -230,6 +230,7 @@ namespace CGAL {
       clear_nodes();
 			m_primitives.clear();
 			clear_search_tree();
+			m_default_search_tree_constructed = false;
 		}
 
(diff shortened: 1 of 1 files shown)


sloriot:
@lrineau  I think we can remove the `kd_tree_mutex` if we change the behavior of `accelerate_distance_computations(first, end)` to do nothing if the AABB-tree was not already built. On the other end, the mutex does not hurt compared to building the kd-tree. What do you think?

lrineau:
Successfully tested in [CGAL-4.13-Ic-68][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-68.shtml

PR #3241, opened 2018-07-19
Polyhedron Demo: Hide ids when modifying the mesh

maxGimeno:
## Summary of Changes

As I don't think it is possible to find which ids are impacted by a modification of the mesh (invalidateOpenGLBuffers() only re-compute everything, it is not possible to track which face(s), edge(s) or vertex(ices) is/are precisely modified), I think the easiest way of not having Ids in the wrong place is to simply hide them all.


afabri:
Successfully tested manually with VC2013

lrineau:

PR #3242, opened 2018-07-19
Mesh 3 fix: add the possibility to stop the algorithm, when Parallel_tag is used

lrineau:
## Summary of Changes

A user using the `Sequential_tag` version of Mesh_3 was able to use 
```C++
while(!mesher.algorithm_done())
  mesher.one_step()
```
and then add extra logic to control the loop and allow the stop of the algorithm before its end. With `Parallel_tag` that was not possible. This PR adds the possibility via a (yet) undocumented feature.

## Release Management

* Affected package(s): Mesh_3


lrineau:

lrineau:
There is a still an issue with the stop feature of Mesh_3: the file `Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h` does not handle the `stop` atomic Boolean at all. In the Polyhedron demo, if one starts the meshing of a data set, for example `man.off`, with features, whereas the 1D feature detection creates noisy sharp edges, then the Mesh_3 thread is not interruptible at all: one has to wait for `Protect_edges_sizing_field` to finish its jobs, no matter what.

Cc: @maxGimeno

PR #3243, opened 2018-07-19
Fix CGAL::clear

sloriot:
Do not remove element you iterate on

@bo0ts booh! ;)

### DIFF
--- a/BGL/include/CGAL/boost/graph/helpers.h
+++ b/BGL/include/CGAL/boost/graph/helpers.h
@@ -1123,15 +1123,18 @@ clear_impl(FaceGraph& g)
   typedef typename boost::graph_traits<FaceGraph>::edge_descriptor     edge_descriptor;
   typedef typename boost::graph_traits<FaceGraph>::vertex_descriptor   vertex_descriptor;
   typedef typename boost::graph_traits<FaceGraph>::face_descriptor     face_descriptor;
-  BOOST_FOREACH(edge_descriptor ed, edges(g)) {
+
+  std::vector<edge_descriptor> edges_to_rm(edges(g).first, edges(g).second);
+  BOOST_FOREACH(edge_descriptor ed, edges_to_rm)
     remove_edge(ed, g);
-  }
-  BOOST_FOREACH(vertex_descriptor vd, vertices(g)) {
+
+  std::vector<vertex_descriptor> vertices_to_rm(vertices(g).first, vertices(g).second);
+  BOOST_FOREACH(vertex_descriptor vd, vertices_to_rm)
     remove_vertex(vd, g);
-  }
-  BOOST_FOREACH(face_descriptor fd, faces(g)) {
+
+  std::vector<face_descriptor> faces_to_rm(faces(g).first, faces(g).second);
+  BOOST_FOREACH(face_descriptor fd, faces_to_rm)
     remove_face(fd, g);
-  }
 }
 
 } //end of internal namespace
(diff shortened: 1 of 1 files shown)


BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -1123,15 +1123,18 @@ clear_impl(FaceGraph& g)
...
+  BOOST_FOREACH(vertex_descriptor vd, vertices_to_rm)
     remove_vertex(vd, g);
-  }
-  BOOST_FOREACH(face_descriptor fd, faces(g)) {
+
+  std::vector<face_descriptor> faces_to_rm(faces(g).first, faces(g).second);
+  BOOST_FOREACH(face_descriptor fd, faces_to_rm)
     remove_face(fd, g);
-  }
 }
```
lrineau: I think that this implementation is as wrong as the previous one, because a face graph could have its descriptors invalidated by `remove_*. For example the `CGAL::Polyhedron_3` with a vector as container.

sloriot: It should not otherwise almost nothing would work.

sloriot: This means that the polyhedron based on vector is not a valid model.

sloriot: However I agree that nothing is say in the boost [concept](https://www.boost.org/doc/libs/1_67_0/libs/graph/doc/MutableGraph.html)

lrineau: [`adjacency_list`] invalidates its descriptors in `remove_vertex`, when the `VertexList` template parameter is `vecS`. See the section "Iterator and Descriptor Stability/Invalidation" in the documentation.

[`adjacency_list`]: https://www.boost.org/doc/libs/1_66_0/libs/graph/doc/adjacency_list.html

sloriot: I continue to think the same thing but I've changed the patch since it was making it simpler :)

sloriot:
@gdamiand is there any `clear()` function for LCC? The generic `clear()` version is currently used (it removes one element at a time).

gdamiand:
@sloriot Yes there is one clear() function for CMap, and thus for LCC [here](https://doc.cgal.org/latest/Combinatorial_map/classGenericMap.html#ade40ca606988d75b8a9f26d0d9fb75ab)

sloriot:
@gdamiand 
I mentioned that because `CGAL::clear()` calls the generic version.

If I add:
```
void clear()
{
  Base::clear();
}
```
in the class `Linear_cell_complex_for_combinatorial_map`, it calls the member function.
If I replace the _hack_ in `Has_member_clear` by C++17 `std::is_member_function_pointer<decltype(&T::clear)>::value` it is picking the right version.

I guess this is what due to what @lrineau warned at some point.
Anyway, shall we add the C++17 (when available) implementation in `Has_member_clear()` (and other similar utility classes) as well as the `clear()` function in LCC class?

sloriot:
You did right. There is no chance to do something better provided the constrained of the concept. Since a descriptor might be invalidated by a erase. Note that it is somehow documented : "If the graph has a member function clear(), it will be called instead."

sloriot:
My point is in the comment above is about the hack to check if a member function is present.

sloriot:
Actually, we could have removed the descritpor before the end...

PR #3254, opened 2018-07-24
Mesh 3 fix: add the possibility to stop the algorithm, when Parallel_tag is used (for CGAL-4.12-branch)

lrineau:
## Summary of Changes

A user using the `Sequential_tag` version of Mesh_3 was able to use 
```C++
while(!mesher.algorithm_done())
  mesher.one_step()
```
and then add extra logic to control the loop and allow the stop of the algorithm before its end. With `Parallel_tag` that was not possible. This PR adds the possibility via a (yet) undocumented feature.

## Release Management

* Affected package(s): Mesh_3


lrineau:

PR #3255, opened 2018-07-24
Remove unused file

sloriot:
The file is a remnant from the Polyhedron-only era.

Not used anywhere in CGAL.


lrineau:

PR #3256, opened 2018-07-26
CGAL_imageIO: Fix .inr writing without zlib

maxGimeno:
## Summary of Changes
When linking with zlib, always use the gz API, as gzFiles are not compatible with fopen(), fread(),etc. 
## Release Management

* Affected package(s):CGAL_imageIO

### DIFF
--- a/CGAL_ImageIO/include/CGAL/ImageIO_impl.h
+++ b/CGAL_ImageIO/include/CGAL/ImageIO_impl.h
@@ -194,6 +194,12 @@ size_t ImageIO_write(const _image *im, const void *buf, size_t len) {
       fprintf(stderr, "zlib error: %s\n", gzerror(im->fd, &errnum));
     }
     return ( len - to_be_written );
+  case OM_FILE :
+    while ( (to_be_written > 0) && ((l = gzfwrite( b, sizeof(char), ImageIO_limit_len(to_be_written), im->fd )) > 0) ) {
+      to_be_written -= l;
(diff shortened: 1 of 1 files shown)


lrineau:
There is a problem. [`gzfwrite` was only added in zlib recently](https://zlib.net/ChangeLog.txt):
>  Changes in 1.2.9 (31 Dec 2016)
>  - [...]
>  - Add `gzfwrite()`, duplicating the interface of `fwrite()`

Your new code will have to use preprocessor conditionals.

CGAL_ImageIO/include/CGAL/ImageIO_impl.h
```diff
@@ -63,6 +63,12 @@
 #include <CGAL/ImageIO/mincio.h>
 #endif
 
+
+#ifdef CGAL_USE_ZLIB
+#define ZLIB_MAJOR_VERSION TEST_VERNUM>>8
```
lrineau: What is `TEST_VERNUM`?

maxGimeno: A copy/paste error :(

maxGimeno:
from zlib manual : 
#define ZLIB_VERNUM 0x12b0

So it is 1.2.11->1211->x012b0.
With bit shifting I can split it in a major-minor/super-minor versions and test like this.

lrineau:

PR #3265, opened 2018-07-31
Polyhedron demo: Fix selection plugin

maxGimeno:
## Summary of Changes
Call filter_operations() after a selection using the lasso so the operations on selection are displayed.


lrineau:

PR #3289, opened 2018-08-20
Be more permissive for collapsing constrained edges in Isotropic remeshing

sloriot:
The only non-collapsable constrained edges are those with
both endpoints being explicitly constrained vertices or corners


lrineau:

PR #3292, opened 2018-08-22
Remove Operations_on_polyhedra

sloriot:
The last functionality not moved to PMP was the initial corefinement version, that have been officially ported to PMP since CGAL 4.10


.travis/packages.txt
```diff
@@ -68,8 +68,7 @@ Nef_3
 Nef_S2 
 NewKernel_d 
 Number_types 
-OpenNL 
-Operations_on_polyhedra 
+OpenNL
```
lrineau: There is one more trailing space, after `OpenNL `. Our tools may be stupid and report an error about it.

lrineau:
There was two trailing spaces, instead of one. And that is what could cause an error, in my opinion. Now you have completely modified the file `packages.txt`, whereas it is supposed to be generated.

Use:
```shellsession
$ cd .travis
$ ./generate_travis.sh
```
and commit+push the result.

lrineau:

PR #3293, opened 2018-08-23
Polyhedron Demo:  Mesh Engraving Plugin

maxGimeno:
## Summary of Changes
In this PR, we add a plugin based on selection, parameterization, extrusion and corefinement, that allows to engrave a text on a closed mesh, assuming the extruded text  does not self-intersect. 
The plugin provides different tools to visually insure the text mesh is correct.


afabri:
@sloriot do we want this only as plugin and not as a library function?

Polyhedron/demo/Polyhedron/Plugins/PMP/Engrave_text_plugin.cpp
```diff
@@ -514,7 +514,7 @@ public Q_SLOTS:
       if (A.x()<0) A=-A;
           angle = std::acos(A.x()/CGAL::sqrt(A.squared_length()));
           if ( A.y()<0 ) angle+=3*CGAL_PI/2.;
-          if (angle>2*CGAL_PI) angle-=CGAL_PI;
+          if (angle>2*CGAL_PI) angle-=2*CGAL_PI;
```
lrineau: Note that you can use `atan2` instead of `acos`. It returns directly an angle in (-π , +π], taking the sign of `y` into account.

sloriot:
@afabri I'm not sure about what the API would look like since we need as input a text polygon.

As possible improvement we could have:
-  the automatic fitting of the text in the parametrized region (right now the minimum was done and humain interaction is needed)
- use a walk in the 2D triangulation to refine the polygon letters with the intersection with the mesh edges

afabri:
The problem I see is that it is hard to use code which is only in a plugin.  To give you an example I needed to find the face id of the face with the smallest angle which is buried in the display_properties plugin.

maxGimeno:
@sloriot 
> the automatic fitting of the text in the parametrized region

 I think a 2D version of what the Kernel Plugin does would be a good method. It would give an inscribed quad contained entirely inside the 2D mesh,  that could be used as bbox for the text.

sloriot:
I propose we try to integrate the plugin first and then make the function when we have time. @maxGimeno could you update the branch (since it is older than the removal of Polyhedron_item, I'm not sure it is 100% correct)?

lrineau:
Warnings on Windows: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-46.shtml#Polyhedron_Demo

PR #3294, opened 2018-08-27
Maintenance: Fix the Release Mode for testsuite platforms 

### LINKED ISSUE #3063: Our old  autotest_cgal testsuite driver is not compatible with CGAL header-only mode
## Issue Details

Take the column `x64_Cygwin-Windows10_MSVC2017-Release-64bits` of
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-238.shtml#Installation

Its [Installation line][1] shows the setup of a Release platform:
```
-- Build type: Release
-- USING CXXFLAGS = '/DWIN32 /D_WINDOWS /W3 /GR /EHsc -D_CRT_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_WARNINGS /fp:strict /fp:except- /wd4503 /bigobj /MD /O2 /Ob2 /DCGAL_NDEBUG'
-- USING LDFLAGS = '/machine:x64 /debug:NONE /INCREMENTAL:NO '
```
but all the [other lines][2]

[1]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-238/Installation/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz
[2]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-Ic-238/AABB_tree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz
```
-- Build type: Debug
-- USING CXXFLAGS = '/DWIN32 /D_WINDOWS /W3 /GR /EHsc /MDd /Zi /Ob0 /Od /RTC1'
-- USING EXEFLAGS = '/machine:x64 /debug /INCREMENTAL'
```

That flags chosen as a definition of that test platforms are no used for the examples, tests, and demos. Because in header-only there is no copy-pasting of compilation and linker flags from the `CGAL_DIR` to the using projects.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): all (but Linux tests now use `test_cgal.py` for Docker, that does not have that issue)
* CGAL version: CGAL-4.11 is already impacted

## Proposal

I think that issue can be reproduced even on Linux, using the `autotest_cgal`, with a reference platform that has `CGAL_HEADER_ONLY:BOOL=ON` and `CMAKE_BUILD_TYPE:STRING=Release` in the reference cache.

Instead of using the CMake command [`load_cache`](https://cmake.org/cmake/help/v3.0/command/load_cache.html) with `${CGAL_REFERENCE_CACHE_DIR}/CMakeCache.txt`:
https://github.com/CGAL/cgal/blob/4485412a52414ef2aeef3e75ac44f4fca5de0576/Installation/CMakeLists.txt#L386-L429

All our calls to `cmake` should use the [`-C` option](https://cmake.org/cmake/help/v3.11/manual/cmake.1.html) to preload a small set of variables (such as `CMAKE_CXX_FLAGS_INIT`).

For the moment, there is one call to `cmake` in `autotest_cgal` (easy to change), but there is also calls in all `cgal_test` shell scripts. Some of them are generated, but some of them are committed in CGAL sources.

Those `cgal_test` shell script should use an optional variable, that would indicate which initial cache to load with `-C`. As I said earlier, those initial caches should be smaller than our current "reference caches": with only the small set of important variables (mostly flags). Most, if not all, of the paths to dependencies can be specified using environment variables, that we stored in `setup` shell scripts.

In the Docker testsuite, I used a trick, once, abusing `CMAKE_GENERATOR`:

https://github.com/CGAL/cgal-testsuite-dockerfiles/blob/765fb48ffbf3e88272340573f2a12e40e2057806/ArchLinux/Dockerfile#L128-L131

https://github.com/CGAL/cgal-testsuite-dockerfiles/blob/765fb48ffbf3e88272340573f2a12e40e2057806/ArchLinux/init.cmake

Maybe that trick can be used again, without changing the existing `cgal_test`
shell scripts.

maxGimeno:
## Summary of Changes

Add -C option to all `cmake` calls in the testsuite mechanism to avoid using a CMakeCache. Moreover, this permits the use of the 'Release' build type, contrary to the current system, as explained in #3063.

To put this fix in place, it will require a pass on all the testmachines to add a corresponding init.cmake file and the env variable to define it. 
## Release Management

* Affected package(s):All packages with tests.
* Issue(s) solved (if any): fix #3063


lrineau:
I disagree with the change you made on `Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test_base`. I am pretty sure that now the test is broken, because the variable `TEST_WITH_CMAKE` is not filled. Did you test the change?!

maxGimeno:
Because the argument -cmake is missing in the call ?

PR #3298, opened 2018-08-28
Classification Plugin Bugfix: do not compute simple properties in parallel

sgiraudot:
## Summary of Changes

Simple features (based on point set properties, without any computation) were computed in parallel. The problem is that it resulted in temporary references being passed to threads, which lead to segfaults later.

As these features do not require any computation, it's anyway useless to include them in the parallel processing, so I just move them out of the parallel part.

(This bug is not in `4.12`.)

## Release Management


lrineau:
Successfully tested in [CGAL-4.13-Ic-103][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-103.shtml

PR #3300, opened 2018-08-28
Polyhedron demo: Fix statistics for facegraph items

maxGimeno:
## Summary of Changes

Fix headers fot statistics in polyhedorn and surface_mesh items so the Null Length is not in the Angles part anymore.


lrineau:
Successfully tested in [CGAL-4.13-Ic-103][*], and then merged.

  [*]: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13-Ic-103.shtml

Issue #569, opened 2015-12-21
Improve robustness of release creation scripts


bo0ts:
Currently three scripts are involved in the process of creating releases:

- Scripts/developer_scripts/create_internal_release
- Scripts/developer_scripts/create_new_release
- Maintenance/infrastructure/cgal.geometryfactory.com/bin/create_release

The third scripts kicks-off the whole process, create_new_release sets up the correct branch, and create_internal_release does the dirty work. All three scripts still contain references to svn, the old doc tools and have a habit of not checking return values. They could do with an overhaul to make it easier for developers to understand this process and to make them easier to maintain.

sgiraudot:
@lrineau what do you think of this issue?

lrineau:
I do agree with Philipp (@bo0ts). But we do not have the manpower to improve or rewrite those tools. There was some hope, in a branch started by Philipp (see #23) but the branch was including two different features, he left before both features were ready to be integrated. (@sloriot is also somehow involved in #23.)

sloriot:
Things are improving with the removal of dont_submit files. We are one step closer to generate the release with cmake directly.

lrineau:
For the records: the removal of dont_submit files was #1446.

sloriot:
releases are created with cmake since some time now.

PR #3307, opened 2018-09-03
Maintenance: Use cmake to create a release

maxGimeno:
## Summary of Changes
create_new_release has been updated to get rid of rpm and svn, and use a cmake script instead of the pearl script that creates a release.

PS: the calls to that script will probably have to be updated, as new options have been added.

Also, I could not test the [public] option, so it will have to be tested before integration.
## Release Management

* Affected package(s):Scripts

### DIFF
--- a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
+++ b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
@@ -1,4 +1,5 @@
 #option :
+# GIT_REPO the path to the Git repository, default is the current working directory
 # DESTINATION the path where the release is created, default is /tmp
 # PUBLIC=[ON/OFF] indicates if a public release should be built, default is OFF
 # VERBOSE=[ON/OFF] makes the script more verbose, default is OFF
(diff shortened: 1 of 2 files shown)


MaelRL:
Related to https://github.com/CGAL/cgal/issues/569?

maxGimeno:
I guess so. But it still takes several scripts to create a release, so I don't know if it should "fix" the issue.

lrineau:
I agree with Maxime: this PR proposed to deprecate the Perl script `create_internal_release`, in favor of a CMake script, that is also faster. But the other two scripts `create_release` and `create_new_release` remain. Maybe the logic of `create_release` should be merged into `create_new_release`.

Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```diff
@@ -200,7 +217,7 @@ if (TESTSUITE)
     endif()
   endforeach()
   file(REMOVE_RECURSE "${release_dir}/tmp")
-endif()
+endif() //TESTSUITE
```
lrineau: Comments in CMake start with `#`, and not `//`. Does not it give you a syntax error?!

maxGimeno: woops

PR #3310, opened 2018-09-04
Handle the case of degenerate volumes in the link condition

### LINKED ISSUE #2431: Polyhedron demo:  Filling small holes can be surprising
## Issue Details

The hole filling in the Polyhedron demo will also consider the border of an isolated triangle (or the border any small connected component) as a hole.   For a triangle this leads to two triangular faces sharing three vertices and three edges.  And this surface is not dealt with, for example in the surface mesh simplification.

sloriot:
Somehow fixes #2431


lrineau:

PR #3057, opened 2018-04-25
Fix doc warnings (and bugs)

sloriot:
Fix warnings and bugs discovered by doxygen improvements

PR #3315, opened 2018-09-05
Fix PMP user manual

sloriot:
The end of the user manual was clipped due to the usage of `\cond`. Using `\if` fixed it.

Introduced by #3057


lrineau:

Issue #3320, opened 2018-09-06
Wrong signatures in LAS


appanacca:
Hello,
I had a problem using the point_set_processign library. In particular with writing a .las file.
Following the example in the documentation for reading a .las file with colors I have tried to write a new file using this syntax:

```
 std::ofstream out_stream(file_name);
CGAL::write_las_points_with_properties(
      out_stream,
      pts,
      CGAL::make_las_point_writer(
          CGAL::Nth_of_tuple_property_map<0, PointWithColor>()),
      std::make_pair(
          CGAL::Nth_of_tuple_property_map<1, PointWithColor>(),
          CGAL::LAS_property::R()),
      std::make_pair(
          CGAL::Nth_of_tuple_property_map<2, PointWithColor>(),
          CGAL::LAS_property::G()),
      std::make_pair(
          CGAL::Nth_of_tuple_property_map<3, PointWithColor>(),
          CGAL::LAS_property::B()));
```

But this code does not compile due to this error: 
***l/include/CGAL/IO/write_las_points.h:176:15: error: binding reference of type ‘short unsigned int&’ to ‘const value_type {aka const short unsigned int}’ discards qualifiers
   output_value(point, get(current.first, *it), current.second);

After a check on the error I have found indeed that the problem is with the signature of all the **output_value** functions overload in the write_las_points.h.
The second function parameter should be declared const everywhere like this:
```
 
void output_value(LASpoint &r, const unsigned short &v, LAS_property::R &) {
  r.set_R(v);
}
```

Since the **get** method used below return a not const reference:

```
template <typename ForwardIterator, typename PropertyMap, typename T>
void output_properties(LASpoint &point, ForwardIterator it,
                       std::pair<PropertyMap, T> &&current) {
  output_value(point, get(current.first, *it), current.second);
}
```

I have already tested that changing the **output_value** signature works well *, and I can write my .las file correctly.

Can someone confirm my analysis ? So it could be corrected ?

Thank you

sgiraudot:
You are indeed correct. I'll make a fix. Thank you for the report!

appanacca:
Glad to help the project!

lrineau:
Now fix in #3321. The PR will be in CGAL-4.13, and in CGAL-4.12.2 (if that bug-fix release is eventually published).

PR #3321, opened 2018-09-06
Point set processing IO: Several bugfixes

sgiraudot:
## Summary of Changes

This PR fixes several bugs/badly done things:
 - The constness of LAS output functions (issue https://github.com/CGAL/cgal/issues/3320)
 - The index of a new point in a point set was not correctly setted when reusing existing memory
 - When writing a point set with points using `float`, they were still written as `double` in the PLY output (which is bad because in binary, it doubles the size of the file with no added value)
 - One of the three `char` types (`signed char`, `char` and `unsigned char`) was missing for PLY IO
 - `char` type appeared as characters in the ASCII PLY output (whereas they should just be numbers)

## Release Management

* Affected package(s): Point Set Processing, Point Set 3
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/3320

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -418,7 +418,9 @@ public:
       {
         -- m_nb_removed;
         iterator out = m_indices.end() - m_nb_removed - 1;
+        Index idx = *out;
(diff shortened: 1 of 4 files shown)


lrineau:

PR #3323, opened 2018-09-07
Remove distance check in triangulate()

### LINKED ISSUE #3212: Polyhedron demo: Thin quad facets not displayed
## Issue Details

In a polygon soup with some quad facets, some are not displayed (`Facet not displayed` is outputted on the terminal), see the three white holes on this screenshot:

missing facets

It's because of [this line of code](https://github.com/CGAL/cgal/blob/master/Polyhedron/demo/Polyhedron/triangulate_primitive.h#L124) that triangulates facets with >3 vertices: it ignores vertices if they are closer than a threshold `CGAL::square(0.0001*item_diag)`. I'm sure there is a reason behind this hack, but it really seems like a bad idea (facets _can be_ long and there's no reason not to display them - there's nothing wrong with the three undisplayed facets on my screenshot).

I can fix my case using 0.000001 instead of 0.0001 in the threshold (I tested and it works) but that's just as arbitrary as the original code. What happens if this filtering is not done at all and the facets are _just_ triangulated?

## Environment

* CGAL version: master

maxGimeno:
## Summary of Changes
Removes the distance check in FacetTriangulator::triangulate(). 

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #3212


lrineau:
@sgiraudot Let me know if you tested it and think that it should be pushed to the 4.13 release.

MaelRL:
@maxGimeno, you can clean up whitespace on save in the lines you have modified in Qt Creator (the setting is in `Text Editor` > `Behavior` > `Cleanups Upon Saving`).

sloriot:
@maxGimeno Conflicts please rebase.

sloriot:
@maxGimeno Did you handle the issue reported by @sgiraudot ?

maxGimeno:
Yes I did, in 91ff7c2. It was the "bug in soup", reffered in the commit message, IIRC. @sgiraudot should probably test to check if it works for him.

maxGimeno:
@lrineau This PR conflicts with #3452. I remove it from the testsuite for now, as it is still in discusison and it blocks the results of the other branch.

PR #3329, opened 2018-09-11
QGLViewer bugfix 

### LINKED ISSUE #3327: operator << for Vec
## Issue Details

`operator <<` for `qglviewer::vec` class does not work.

## Source Code

The following simple code raises a link error: `undefined operator << for CGAL::qglviewer::Vec`.

```
#include <iostream>
#include <CGAL/Qt/qglviewer.h>
int main()
{
  CGAL::qglviewer::Vec v(1., 0., 0.);
  std::cout<<v<<std::endl;  
  return EXIT_SUCCESS;
}
```

I think `operator <<` in file `vec_impl.h` should be defined outside `CGAL::qglviewer` namespace.

## Environment

* Compiler:
* Release or debug mode:
* CGAL version:
* Boost version:

gdamiand:

## Summary of Changes

operators << for vec and quaternion should not be in CGAL::qglviewer namespace (otherwise there is a link error).

## Release Management

* Issue(s) solved (if any): fix #3327

### DIFF
--- a/GraphicsView/include/CGAL/Qt/quaternion.h
+++ b/GraphicsView/include/CGAL/Qt/quaternion.h
@@ -323,7 +323,7 @@ public:
           Quaternion rot(...);
           cout << "Rotation=" << rot << endl;
           \endcode */
-  std::ostream &operator<<(std::ostream &o, const CGAL::qglviewer::Vec &);
+  std::ostream &operator<<(std::ostream &o, const CGAL::qglviewer::Quaternion &);
 //@}
 #endif
 
--- a/GraphicsView/include/CGAL/Qt/quaternion.h
+++ b/GraphicsView/include/CGAL/Qt/quaternion.h
@@ -333,8 +333,8 @@ private:
   qreal q[4];
 };
 
-}} // namespace CGAL::qglviewer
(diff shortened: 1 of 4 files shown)


lrineau:
The PR looks wrong. Both the declaration and the definition should be in the namesapce `::CGAL::qglviewer`.

The reason of bug #3327 was because there was a definition in a namespace (`::CGAL::qglviewer`) and a declaration in another one. The solution is not to move the definition to the empty namespace, but to move the declaration in `::CGAL::qglviewer`.

gdamiand:
This is done like that in the original qglviewer, this is why I do the update in this way (see for example https://github.com/GillesDebunne/libQGLViewer/blob/master/QGLViewer/vec.h).

gdamiand:
@lrineau  You can pass this PR in the testsuite; but since there is no test for our version of qglviewer, the code will obviously be correct (the previous code has linker error and this is not detected...).

lrineau:
Our demos are compiled. There could be a compilation error on Windows that you do not know.

gdamiand:
Since a bug was there before this PR, this shows that these operators are not compiled...

lrineau:

PR #3341, opened 2018-09-19
Mesh_3, parallel: fix the stop feature

lrineau:
## Summary of Changes

When the stop variable is flipped, cancel unprocessed tasks, instead of just waiting for them to finish.

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): fix  #3242


afabri:
tested on Windows/VC2018

PR #3345, opened 2018-09-20
Improve Projection of 2D point on line

### LINKED ISSUE #3301: Bug in `line_project_pointC2` when the line is almost horizontal
Using kernel `Simple_cartesian<double>`.

I have a `Line_2`:
a = -4.2885603045067812e-18
b = 1
c = 250.73609999999996

When I try to project the point (35.306000000000004, 250.69800000000001) onto the line, I get the following projected point : (0, 250.73609999999996).

afabri:
## Summary of Changes

The implementation had a "better", that is more robust version in an ` #else`, but it had a bug. 
We keep the vertical and horizontal special cases, and use fixed ` #else`. Add an ` issue_3301.cpp`   in the testsuite, that at the same time has a performance test. The new code is slightly faster.

## Release Management

* Issue(s) solved (if any): fix #3301

### DIFF
--- a/Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h
+++ b/Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h
@@ -313,8 +313,7 @@ line_project_pointC2(const FT &la, const FT &lb, const FT &lc,
 		     const FT &px, const FT &py,
 		     FT &x, FT &y)
 {
-#if 1 // FIXME
-  // Original old version
+  std::cout << la << " "<< lb << " "<< lc << " "<< px << " " << py << std::endl;
   if (CGAL_NTS is_zero(la)) // horizontal line
   {
     x = px;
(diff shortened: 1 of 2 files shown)


lrineau:
bad alloc on a 32 bits platform:
```
- ProgramOutput.issue_3301.CentOS6-32
CGAL::Random()::get_seed() = 1538608670
35.306000000000004 -250.73609999999996
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
./cgal_test_with_cmake: line 70: 26894 Aborted                 (core dumped) ./issue_3301
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-9/Kernel_23/TestReport_lrineau_CentOS6-32.gz

afabri:
@maxGimeno  can we have a look at it in docker?

lrineau:
The issue is there:
https://github.com/CGAL/cgal/blob/d49c0f9e611dbd82505efdb53da2b085703b75e2/Kernel_23/test/Kernel_23/issue_3301.cpp#L23-L33

A Point_2 is two `double`, that is 16 bytes, so the vector of 100000000 points is 1.6GiB, but that is the size of it, and maybe the capacity is bigger. Imagine the new capacity is about 2GiB. Then remember that during a reallocation of the vector, the old storage and the new one are active at the same time, during the copy. Probably that got beyond the limit of 3GB.

lrineau:
https://github.com/CGAL/cgal/blob/dc3f0e940040ce50e84ba721340182c43e137b19/Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h#L309-L334

- One can remove the `CGAL_NTS`.
- I suggest to modify the two first tests of nullity to add a call to `certainly()`:
```c++
   if (certainly(is_zero(la))) // horizontal line 
   { 
     x = px; 
     y = -lc/lb; 
   } 
   else if (certainly(is_zero(lb))) // vertical line 
   { 
     x = -lc/la; 
     y = py; 
   }
   else
```

that way, if the constructor is used in `Epeck`, that will avoid filter failures in case the one of the coefficient is close to zero. And the code would use the first two cases only if the coefficient are 100% sure of being zero (because they are input coordinates).

PR #3349, opened 2018-09-24
Classification: Workaround serialization bug with null pointers

sgiraudot:
## Summary of Changes

_Boost Serialization_ is supposed to take into account NULL pointers, but for some reason, this is broken when compiling with CLang. I managed to reduce the bug to a very simple non-CGAL-related program and I reported the bug to the Boost tracker (see https://github.com/boostorg/serialization/issues/119).

In the mean time, I suggest that we integrate this workaround that simply checks if the pointers are NULL before calling the serialization framework on them (which, again, is supposed to be useless – and which is not needed on GCC or MSVC from what we experimented).

I also added a test for the IO functions: currently, only the compilation is tested (through the plugin in the demo), not the execution, otherwise we would have detected this bug earlier. The test writes the configuration in a file, loads it again and checks if the classification result remains the same.

## Release Management


lrineau:
`test_io` does not compile:

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-8.shtml#Classification

As for the runtime error in Classification Examples, it can be another error.

sgiraudot:
I based this branch on `4.12`, and parts of the API (`Local_eigen_analysis` in particular) have changed from `4.12` to `4.13` (which lead to the error on the testsuite), so this PR can't work on both branches.

Should I create another PR for `4.13` or should I update this PR so that it works on `4.13` (and we just forget about `4.12`)?

This bug happens for only one platform for one specific function of a relatively recent package that does not have that many users, so I would be tempted to just fix for `4.13` and forget about `4.12`.

lrineau:
Indeed that seems easier to only support the branch for 4.13.1. I agree.

sgiraudot:
Okay, now this PR should work fine for tests with `4.13`.

Classification/test/Classification/test_io.cpp
```diff
@@ -0,0 +1,100 @@
+#if defined (_MSC_VER) && !defined (_WIN64)
```
lrineau: I have an issue with this file:
```
-- Configuring test in test/Classification
-- Boost version: 1.66.0
-- Boost version: 1.66.0
-- Found the following Boost libraries:
--   serialization
--   iostreams
--   regex
-- Found Eigen3: /usr/include/eigen3 (found suitable version "3.3.5", minimum required is "3.1.0") 
-- Found Intel TBB
CMake Error at Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake:49 (add_executable):
  add_executable cannot create target "test_io" because another target with
  the same name already exists.  The existing target is an executable created
  in source directory
  "/home/lrineau/Git/other-cgal/Arrangement_on_surface_2/test/Arrangement_on_surface_2".
  See documentation for policy CMP0002 for more details.
  Classification/test/Classification/CMakeLists.txt:108 (create_single_source_cgal_program)
```

lrineau: It should be renamed.

lrineau: @maxGimeno The `cmake` command in `.travis.yml` should use `-DWITH_examples=ON`, `-DWITH_demos=ON`, and `-DWITH_tests=ON`, to check that no two targets have the same names.

sgiraudot: Okay, I've used a more specific name.

sgiraudot:
Of course… with my fix, the serialized files have slightly changed format and are no longer compatible with the new version. So now, reading the old files that were provided with examples lead to a crash I'll have to change these files…

sgiraudot:
Done.

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Oct 5, 2018, 1:05 PM GMT+2](https://github.com/CGAL/cgal/pull/3349#issuecomment-427327692 "2018-10-05T11:05:13Z - Replied by Github Reply Comments"):
> Of course… with my fix, the serialized files have slightly changed format and are no longer compatible with the new version. So now, reading the old files that were provided with examples lead to a crash I'll have to change these files…

Is that something that could impact our users? Maybe they should be told in the changelog (and the PR cannot go the release branch)?

sgiraudot:
Yes it will definitely impact users who previously saved configuration files and who won't be able to use them… I can provide a conversion tool, but we don't really have a place/way to provide such things.

When I discussed the bug with @afabri, we kind of agreed that classification so far does not have many users and that it was unlikely that someone would run into this problem. If I understand correctly, if we want to warn the users in the changelog, we have to wait for the next release? I'm not sure what's the best thing to do here…

lrineau:
Classification is still in a bad shape:

https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-15.shtml#Classification

sgiraudot:
The configuration files are fixed using Boost 1.55.0 for retro-compatibility (I amended the commit where I updated them to the new format). Now they should work on all versions we support.

lrineau:

PR #3350, opened 2018-09-25
Classification plugin bugfix: don't use temp refs in parallel code

sgiraudot:
## Summary of Changes

When calling methods to add features in a parallel scope, no temporary object should be used (as documented). This was not the case in the plug-in where the normal map was created on the fly (and destroyed before the launched thread had the chance to use it). This PR fixes it by creating the point map outside of the method.

## Release Management

* Affected package(s): Polyhedron demo only


lrineau:

PR #3352, opened 2018-09-26
Add warning about MacOS Dark node and demo

sloriot:

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -6,6 +6,11 @@ Release 4.13
 
 Release date: September 2018
 
+### Installation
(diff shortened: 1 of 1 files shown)


lrineau:
Qt5 bug: https://bugreports.qt.io/browse/QTBUG-68850

lrineau:
It will be fixed by Qt 5.12-beta1, in a week. Do we really want to put that in our release notes?

sloriot:
Since I noticed it I wanted to write it somewhere. At least I'm no longer the only one aware of it so we can close the PR and use it for later reference.

PR #3353, opened 2018-09-26
Interpolation: Slightly change linear_interpolation() so that it can also interpolate vectors

afabri:
## Summary of Changes

* Generalize the wording in the documentation
* Don't initialize the result with `0`, but with the first term of a sum computed in a loop
* Add an example

## Release Management

* Affected package(s): Interpolation
* License and copyright ownership: stays at inria


Interpolation/include/CGAL/interpolation_functions.h
```diff
@@ -73,13 +72,14 @@ linear_interpolation(ForwardIterator first, ForwardIterator beyond,
   typedef typename boost::result_of<ValueFunctor(arg_type)>::type                 result_type;
   typedef typename result_type::first_type                                        Value_type;
 
-  Value_type result(0);
-  result_type val;
+  result_type val = value_function(first->first);
```
MaelRL: Need to check that the range is not empty somewhere.

Kernel_23/include/CGAL/Vector_3.h
```diff
@@ -370,7 +370,7 @@ operator>>(std::istream& is, Vector_3<R>& v)
 {
   return extract(is, v, typename R::Kernel_tag() );
 }
-
+
```
MaelRL: some bad whitespace changes throughout the diff.

Interpolation/include/CGAL/interpolation_functions.h
afabri: I added  preconditions for all interpolation functions in the code and the manual. 
@lrineau: Note that this is not a change as we already now have the precondition that the sum of weights `norm` is larger than zero, which implies a non-empty range.

lrineau:

PR #3357, opened 2018-09-29
Convex_hull_3: Fix so that static filters are used

afabri:
## Summary of Changes

Replace "hand written" filtering with just calling the functor of the kernel chosen by the user so that static filters are used.

## Release Management

### DIFF
--- a/Convex_hull_3/include/CGAL/Convex_hull_traits_3.h
+++ b/Convex_hull_3/include/CGAL/Convex_hull_traits_3.h
@@ -147,15 +147,8 @@ public:
       const Point_3& hp = h.p();
       const Point_3& hq = h.q();
       const Point_3& hr = h.r();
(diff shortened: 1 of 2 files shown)


sloriot:
Do you get any speed up?
You should either see no difference or a slow down in practice. Indeed,  there is a static filtering layer for cartesian filtered kernel in the functor `Is_on_positive_side_of_plane_3` from `convex_hull_3.h`. This functor is doing some pre-computations when the plane is used several times in the predicate.

If you do get a speed up then there is probably a bug in the mechanism.

afabri:
The only reason for having done it is the speedup.

sloriot:
Ok then there is a bug somewhere. I can reproduce the speed-up with your fix. I'm investigating.

sloriot:
Filters were indeed deactivated due to a typo. Fixed in #3359

lrineau:

PR #3358, opened 2018-10-01
Polyhedron demo:  Several fixes

### LINKED ISSUE #3356: Several problems in the demo
## Issue Details

* [x] Group items mess up the recentering (especially with an object far away): its seems that the bbox of the group is considered at the origin, whereas it should be the union of bboxes of its items

* [x] Selecting points from different views makes the previous selection appear while selecting a new one:

  1. Load a point set
  2. Selection tools, select a rectangle
  3. Move around the object
  4. Select rectangles several times: the original selection (before moving) appears everytime

* [x] When estimating the normals of a point set, the display switches to pts+normals but when clicking on the Mode column, only "Points" are available (not Points and Normals or Shaded Points)

* [x] _(By the way, if this is corrected, we might as well make shaded points the default view for point sets with normals, either when estimating or when loading a point set with normals. Points with normals display is useful, but using it as default is annoying, it's long to display and often looks messy when looking at the big picture.)_

* [x] When performing Advancing Front Surface Reconstruction, the polyhedron produced does not appear.

  1. Load the `sphere_1k.xyz` file from the Point Set Processing data examples
  2. `Surface Reconstruction` with default parameters
  3. An item is produced but looks empty (although it's not if you look at the infos)
  4. `Operation on polyhedra -> Join polyhedra` makes it appear (or `Inside Out`, etc.)

* [ ] When performing Point Set Shape Detection, sometimes the results does not appear either (similarly as with polyhedron above). It seems pretty random, but it seems that if you _did_ the reconstruction steps before, you can reproduce this way:

  5. Discard the `sphere_1k.xyz` and the rest
  6. Load the `cube.pwn` model from the Point Set Shape Detection package
  7. Launch shape detection plugin with default parameters
  8. An item is produced but looks empty (but again, it's not)
  9. `Select duplicated points` makes it appear…

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux Mint 64b
* Compiler: clang
* CGAL version: master
* Other libraries versions if used : Nvidia drivers

### LINKED ISSUE #3360: Polyhedron demo: Bugs in display of some surface mesh
## Issue Details
With some not triangule surface mesh, facets are not displayed.
## Source Code
OFF
4 1 0

0 0 1
1 0 0
1 1 1
2 1 1

4 0 2 1 3

## Environment

* Compiler:
* Release or debug mode:
* CGAL version:
* Boost version:

maxGimeno:
## Summary of Changes
Fixes some trivial bugs in the Polyhedron demo.
## Release Management
* Issue(s) solved (if any): fix #3356, fix #3360

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_normal_estimation_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_normal_estimation_plugin.cpp
@@ -275,7 +275,7 @@ void Polyhedron_demo_point_set_normal_estimation_plugin::on_actionNormalEstimati
                                                   << (memory>>20) << " Mb allocated"
                                                   << std::endl;
     }
-
+    item->resetMenu();
     item->setRenderingMode(PointsPlusNormals);
 
     //***************************************
--- a/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_normal_estimation_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_normal_estimation_plugin.cpp
@@ -276,7 +276,7 @@ void Polyhedron_demo_point_set_normal_estimation_plugin::on_actionNormalEstimati
                                                   << std::endl;
     }
     item->resetMenu();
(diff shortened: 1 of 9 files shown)


lrineau:

PR #3359, opened 2018-10-01
Fix activation of local static filters

sloriot:
Fix typos preventing activation of local static filters

### DIFF
--- a/Convex_hull_3/include/CGAL/convex_hull_3.h
+++ b/Convex_hull_3/include/CGAL/convex_hull_3.h
@@ -182,13 +182,13 @@ struct Is_cartesian_kernel< Convex_hull_traits_3<Kernel, PolygonMesh, Tag_true> 
 {
   // Rational here is that Tag_true can only be passed by us since it is not documented
   // so we can assume that Kernel is a CGAL Kernel
-  typedef boost::is_same<typename Kernel::Kernal_tag, Cartesian_tag> type;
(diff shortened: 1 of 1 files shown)


lrineau:
I agree that this version should be faster, because the "predicate" is instantiated once per plane, and stores precomputed sub-expressions.

However, I disagree that is replaces #3357: if the kernel is `CGAL::Epick`, then the static filtered version of `Is_on_positive_side_of_plane_3` will be used, and will compute all the three steps of filtering. However, for all types of kernels, then the version of #3357 is simpler (less line of code), and even more efficient with `CGAL::Epeck`, because the static filters of `Epeck` will be used.

So both #3357 and this PR #3359 are relevant.

sloriot:
You're missing something. This will be only used when the traits is `Convex_hull_traits_3` which should basically be used only if you have a kernel with exact predicates and inexact constructions. Indeed this traits presents a plane as a triple of points rather than a `CGAL::Plane_3`. This is not relevant to you if you have exact constructions or use inexact predicates.
If you keep the patch on #3357 , this will imply that the static filters will be done twice with Epick in case of failure.

lrineau:
https://github.com/CGAL/cgal/blob/48bb90a36e42cb4f97bad676f57bae66fc32ccd7/Convex_hull_3/include/CGAL/convex_hull_3.h#L326-L347

sloriot:
@lrineau I don't know what you want to show with the snapped code. Let's discuss this live to avoid useless rounds.

lrineau:

Issue #731, opened 2016-02-03
A doubt on the right choice for ring and field type in the SDG_Linf_2 testsuite


afabri:
The file test_sdg_linf_2_et.cpp  defines `Gmpq` as ring and as field type.
When GMP is not used we use `MP_Float` instead, but define
`MP_Float` as ring type and `Quotient<MP_Float>` as field type.
This leads to a failing [testcase](https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-4.8-Ic-115/Segment_Delaunay_graph_Linf_2/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.1_Release-LEDA-without-GMP.gz)

Making both `Quotient<MP_Float>`  (as in the SDG_2 testsuite) compiles, but then the names seem
not to be choses correctly. 

If the types should be different the testcase shows that we have to make fixes in the predicates.

lrineau:
`CGAL::Simple_cartesian<CGAL::MP_Float>` cannot compile, because `CGAL::MP_Float` is not a field type, but only a ring type.

The following error:
```
CGAL/constructions/kernel_ftC2.h:40:15: error: invalid operands to binary expression ('CGAL::MP_Float' and 'int')
  x = (px+qx) / 2;
      ~~~~~~~ ^ ~
```
shows that the implementation of `midpoint` for the cartesian kernel does not use our concepts. The division by the `int` 2 is not allowed. It should be `x = (px+qx) / FT(2)`.

lrineau:
`K_ring`, the kernel using a rind type, should use the homogeneous kernel.

lrineau:
Try this patch:
```
--- a/Segment_Delaunay_graph_Linf_2/test/Segment_Delaunay_graph_Linf_2/test_sdg_linf_2_et.cpp
+++ b/Segment_Delaunay_graph_Linf_2/test/Segment_Delaunay_graph_Linf_2/test_sdg_linf_2_et.cpp
@@ -37,11 +37,12 @@ typedef exact_ring_t   ring_number_t;
 typedef exact_field_t  field_number_t;
 
 #include <CGAL/Simple_cartesian.h>
+#include <CGAL/Simple_homogeneous.h>
 
 #include <CGAL/Segment_Delaunay_graph_Linf_2.h>
 #include <CGAL/Segment_Delaunay_graph_Linf_traits_2.h>
 
[...]

PR #3362, opened 2018-10-01
Fix when not using Gmp

sloriot:
Should fix [this](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-I-125/Segment_Delaunay_graph_Linf_2/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.1_Release-LEDA-without-GMP.gz) and [that](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.13-I-125/Segment_Delaunay_graph_Linf_2_Examples/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.1_Release-LEDA-without-GMP.gz).

### DIFF
--- a/Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_filtered_traits_2.h
+++ b/Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_filtered_traits_2.h
@@ -221,7 +221,7 @@ template<class CK,
 #ifdef CGAL_USE_GMP
 	 class EK      = Simple_cartesian< Gmpq >,
 #else
-	 class EK      = Simple_cartesian< MP_Float >,
+	 class EK      = Simple_cartesian< CGAL::Quotient<MP_Float> >,
 #endif
 	 class EK_MTag = Integral_domain_without_division_tag,
 	 class FK      = Simple_cartesian< Interval_nt<false> >,
--- a/Segment_Delaunay_graph_Linf_2/include/CGAL/Side_of_bounded_square_2.h
+++ b/Segment_Delaunay_graph_Linf_2/include/CGAL/Side_of_bounded_square_2.h
@@ -70,7 +70,7 @@ namespace CGAL {
         bool samepq = false;
         bool samepr = false;
         bool sameqr = false;
-        const Comparison_result cmppq = compare(p, q);
+        const Comparison_result cmppq = CGAL::compare(p, q);
         switch(cmppq) {
           case SMALLER:
             min_p = &p;
@@ -86,7 +86,7 @@ namespace CGAL {
             samepq = true;
             break;
         }
-        const Comparison_result cmppr = compare(p, r);
+        const Comparison_result cmppr = CGAL::compare(p, r);
         Comparison_result cmpqr;
         if (samepq) {
           cmpqr = cmppr;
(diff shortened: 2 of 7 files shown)


sloriot:
@philaris I noticed in `Side_of_bounded_square_2.h` division by two (like `fix1 = (*bot_p + *top_p - *rgt_p + *lft_p)/two;`). AFAIU, this code is called with a ring type and is working with `Gmpz` and `Leda::integer` but not with `MP_Float` (integral division is not present). Do you see a possible fix for that?

lrineau:
The testsuite is red:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-8.shtml#Segment_Delaunay_graph_Linf_2

philaris:
Hi Laurent, thanks for bringing this to my attention.

I am wondering if it is related to this issue:
https://github.com/CGAL/cgal/issues/731

I can have a longer look during the weekend.

Best,
PanosOn Wed, Oct 3, 2018 at 10:47 AM Laurent Rineau

lrineau:
@sloriot You are the author of this PR. Did you have time to have a look at the issues in the testsuite?

philaris:
> @philaris I noticed in `Side_of_bounded_square_2.h` division by two (like `fix1 = (*bot_p + *top_p - *rgt_p + *lft_p)/two;`). AFAIU, this code is called with a ring type and is working with `Gmpz` and `Leda::integer` but not with `MP_Float` (integral division is not present). Do you see a possible fix for that?

@sloriot I checked in the code that variables `fix1` and `two` are of type `FT`, i.e., Field Type, so they should support division. I am not very familiar with number types and maybe `FT` in that piece of code under some circumstances is not really a type that supports division.

Changing the code so that it does not use division requires a lot of work.

On the other hand, keeping division and taking care so that the `FT` supports division, e.g., by ensuring that `FT` is a quotient of `RT` in the case of `MP_Float` might be a solution.

All the above comments are just from having a look at the code. I plan to also checkout the latest master and I will try to run the tests.

sloriot:
I put back a rational type in all cases. It fixes the runtime issue locally.

sloriot:
Things are getting better but tests are still failing. See [here](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-33.shtml)

lrineau:
@sloriot Will you have time to work on this PR again for 4.14, or:
  - do we delay it?
  - or do we try to find somebody else to take it over?

sloriot:
I thought using ring type was not correct but even rational does not seem to be correct. So I don't really know. There is something strange as `MP_Float` is used for the ring type while `Gmpq` in used when present. Maybe someone should read the diff again and comment in case I missed something.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Nov 2, 2018, 10:21 AM GMT+1](https://github.com/CGAL/cgal/pull/3362#issuecomment-435318827 "2018-11-02T09:21:07Z - Replied by Github Reply Comments"):
> Things are getting better but tests are still failing. See [here](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-33.shtml)

Which tests are failing? I have seen timeouts. Was there anything else?

sloriot:
Maybe we should test it again because I don't remember what was the remaining issue.

maxGimeno:

PR #3363, opened 2018-10-02
Point set shape detection: Fix VC2017 bug

afabri:
## Summary of Changes

Work around a bug that only happens with `_MSC_VER == 1915`.
See [bug report](https://developercommunity.visualstudio.com/content/problem/340310/unaccepted-typename-that-other-compilers-require.html)

## Release Management

* Affected package(s): Point Set Shape Detection


lrineau:

PR #3367, opened 2018-10-03
Triangulation_3: Fix insertion and removal of ranges of weighted points (for 4.12)

### LINKED ISSUE #3354: Segmentation fault in parallel (TBB) regular triangulation 3d
## Issue Details

Segmentation fault happens when inserting points in a parallel regular triangulation object with varying weights. Running with `gdb` gives the following:

```
[Switching to Thread 0x7ffff5833700 (LWP 20752)]
0x0000555555586084 in CGAL::Spatial_lock_grid_base_3<CGAL::Spatial_lock_grid_3<CGAL::Tag_priority_blocking> >::try_lock<false, CGAL::Weighted_point_3<CGAL::Epick> > (this=0x7fffffffda38, point=..., lock_radius=0) at /usr/include/CGAL/Spatial_lock_grid_3.h:221
221       bool try_lock(const P3 &point, int lock_radius = 0)
```

## Source Code

Using an adaptation of the TBB enabled example with varying weights.

```
#include <CGAL/Regular_triangulation_3.h>
#include <iostream>
#include <fstream>
#include <vector>
#include <random>
#include <functional>

// Regular T3
typedef CGAL::Triangulation_data_structure_3< 
  CGAL::Regular_triangulation_vertex_base_3<K>, 
  CGAL::Regular_triangulation_cell_base_3<K>, 
  CGAL::Parallel_tag>                                       Tds;
typedef CGAL::Regular_triangulation_3<K, Tds>               Rt;
typedef Rt::Bare_point                                      Bare_point;
typedef Rt::Weighted_point                                  Weighted_point;
typedef Rt::Vertex_handle                                   Vertex_handle;

constexpr int N = 1000000;

struct MyRT
{
  Rt::Lock_data_structure lock;
  std::vector<Weighted_point> points;
  Rt rt;

  MyRT(std::function<double ()> w)
    : lock(CGAL::Bbox_3(-1., -1., -1., 1., 1., 1.), 50)
    , rt(K(), &lock)
  {
    CGAL::Random_points_in_cube_3<Bare_point> rnd(1.);
    points.reserve(N);
    for (int i = 0; i < N; ++i)
      points.push_back(Weighted_point(*rnd++, w()));
    std::cout << "Inserting points.\n";
    rt.insert(points.begin(), points.end());
  }
};

int main()
{
  std::mt19937 random(0);
  std::normal_distribution<double> normal;

  MyRT myrt(std::bind(normal, random));
  assert(myrt.rt.is_valid());

  return 0;
}
```

compiled with

```shell
g++ -frounding-math -lm -lCGAL -lgmp -lboost_thread -lmpfr \
    -DCGAL_LINKED_WITH_TBB -pthread -ltbb -latomic -ltbbmalloc -pthread \
    cgal-parallel.cc
```
## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): GNU/Linux Debian (testing) 64bit
* Compiler: confirmed with GCC 6.4.0 and GCC 8.2.0
* Release or debug mode: both
* CGAL version: 4.12.0 and 4.12.1
* Boost version: 1.62.0
* Other libraries versions if used (Eigen, TBB, etc.): TBB

MaelRL:
## Summary of Changes

- In a parallel insertion of a range of weighted points, the 'hint' is a vertex handle that is unsafe to use because the insertion of a weighted point by another thread might hide (delete) the hint.
  _Solution_: check the validity of the hint before and after trying to lock it.
- The removal of a range of weighted points while storing hidden points means that we re-insert hidden points after a successful removal. However, since we have only locked what was needed to remove the vertex and not to re-insert the hidden points, deadlocks can appear.
  _Solution_: unlock everything after successful removal and threads can then wrestle normally for control of zones to re-insert hidden points. Same check is needed on the hint.

## Release Management

* Affected package(s): `Triangulation_3`
* Issue(s) solved (if any): fix #3354

### DIFF
--- a/Triangulation_3/include/CGAL/Regular_triangulation_3.h
+++ b/Triangulation_3/include/CGAL/Regular_triangulation_3.h
@@ -2481,7 +2481,7 @@ namespace CGAL {
 
       if(*could_lock_zone && removed)
       {
-        // The vertex has been removed, try to re-insert the points that 'v' was hiding
+        // The vertex has been removed, re-insert the points that 'v' was hiding
 
         // Start by unlocking the area of the removed vertex to avoid deadlocks
         this->unlock_all_elements();
--- a/Triangulation_3/include/CGAL/Regular_triangulation_3.h
+++ b/Triangulation_3/include/CGAL/Regular_triangulation_3.h
@@ -763,6 +763,25 @@ namespace CGAL {
       return std::copy(vertices.begin(), vertices.end(), res);
     }
 
+    // In parallel operations, we need to be able to check the health of the 'hint' vertex handle,
+    // which might be invalided by other threads. One way to do that is the 'is_vertex()' function
+    // of the TDS, but it runs in O(sqrt(n)) complexity. When we are using our TDS, we can use
+    // a lower level function from the compact container, which runs in constant time.
+    template <typename TDS_>
+    struct Vertex_validity_checker
+    {
+      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) { return tds_.is_vertex(vh_); }
+    };
+
+    template <typename T1, typename T2, typename T3>
+    struct Vertex_validity_checker<CGAL::Triangulation_data_structure_3<T1, T2, T3> >
+    {
+      typedef CGAL::Triangulation_data_structure_3<T1, T2, T3>   TDS_;
+
+      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) {
+        return tds_.vertices().is_used(vh_); }
+    };
+
     void remove (Vertex_handle v);
     // Concurrency-safe
     // See Triangulation_3::remove for more information
@@ -1289,25 +1308,6 @@ namespace CGAL {
       void hide_point(Cell_handle c, const Weighted_point &p) {
         c->hide_point(p);
       }
-    };
-
-    // In parallel operations, we need to be able to check the health of the 'hint' vertex handle,
-    // which might be invalided by other threads. One way to do that is the 'is_vertex()' function
-    // of the TDS, but it runs in O(sqrt(n)) complexity. When we are using our TDS, we can use
-    // a lower level function from the compact container, which runs in constant time.
-    template <typename TDS_>
-    struct Vertex_validity_checker
-    {
-      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) { return tds_.is_vertex(vh_); }
-    };
-
-    template <typename T1, typename T2, typename T3>
-    struct Vertex_validity_checker<CGAL::Triangulation_data_structure_3<T1, T2, T3> >
-    {
-      typedef CGAL::Triangulation_data_structure_3<T1, T2, T3>   TDS_;
-
-      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) {
-        return tds_.vertices().is_used(vh_); }
     };
 
   // Functor for parallel insert(begin, end) function
--- a/Triangulation_3/include/CGAL/Regular_triangulation_3.h
+++ b/Triangulation_3/include/CGAL/Regular_triangulation_3.h
@@ -767,19 +767,32 @@ namespace CGAL {
     // which might be invalided by other threads. One way to do that is the 'is_vertex()' function
     // of the TDS, but it runs in O(sqrt(n)) complexity. When we are using our TDS, we can use
     // a lower level function from the compact container, which runs in constant time.
+    BOOST_MPL_HAS_XXX_TRAIT_DEF(Is_CGAL_TDS_3)
+
+    template <typename TDS_,
+              bool has_TDS_tag = has_Is_CGAL_TDS_3<TDS_>::value>
+    struct Is_CGAL_TDS_3 : public CGAL::Tag_false
+    { };
+
     template <typename TDS_>
+    struct Is_CGAL_TDS_3<TDS_, true> : public CGAL::Boolean_tag<TDS_::Is_CGAL_TDS_3::value>
+    { };
+
+    template <typename TDS_,
+              bool is_CGAL_TDS_3 = Is_CGAL_TDS_3<TDS_>::value>
     struct Vertex_validity_checker
     {
-      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) { return tds_.is_vertex(vh_); }
+      bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) {
+        return tds_.is_vertex(vh_);
+      }
     };
 
-    template <typename T1, typename T2, typename T3>
-    struct Vertex_validity_checker<CGAL::Triangulation_data_structure_3<T1, T2, T3> >
+    template <typename TDS_>
+    struct Vertex_validity_checker<TDS_, true /* is_CGAL_TDS_3 */>
     {
-      typedef CGAL::Triangulation_data_structure_3<T1, T2, T3>   TDS_;
-
       bool operator()(const typename TDS_::Vertex_handle vh_, const TDS_& tds_) {
-        return tds_.vertices().is_used(vh_); }
+        return tds_.vertices().is_used(vh_);
+      }
     };
 
     void remove (Vertex_handle v);
--- a/Triangulation_3/include/CGAL/Regular_triangulation_3.h
+++ b/Triangulation_3/include/CGAL/Regular_triangulation_3.h
@@ -1324,14 +1324,48 @@ namespace CGAL {
 #endif
 
       Vertex_handle &hint = m_tls_hint.local();
+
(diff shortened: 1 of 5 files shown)


Triangulation_3/include/CGAL/Regular_triangulation_3.h
```diff
@@ -1324,14 +1324,48 @@ namespace CGAL {
...
+          if(!m_rt.tds().is_vertex(hint))
+#endif
+          {
+            hint = m_rt.finite_vertices_begin();
+            continue;
+          }
+
+          // We need to make sure that while are locking the position P1 := hint->point(), 'hint'
+          // does not get its position changed to P2 != P1.
+          const Weighted_point hint_point_mem = hint->point();
```
lrineau: The `point()` method is not thread-safe. You should store `hint_point_mem` the same way you store `hint`: only refresh it once the lock is obtained.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
```diff
@@ -1324,14 +1324,48 @@ namespace CGAL {
...
+            // that its position hasn't changed to ensure that we will start the locate from where
+            // we have locked.
+#ifdef CGAL_RT3_IGNORE_TDS_CONCEPT
+            if(!m_rt.tds().vertices().is_used(hint) ||
+#else
+            if(!m_rt.tds().is_vertex(hint) ||
+#endif
+               hint->point() != hint_point_mem)
+            {
+              hint = m_rt.finite_vertices_begin();
```
lrineau: `m_rt.finite_vertices_begin()` is not locked, and may be under destruction as well. In my opinion, only the infinite vertex is safe.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
```diff
@@ -2388,40 +2453,96 @@ namespace CGAL {
...
     else
     {
-      Vertex_handle hint = (v->cell()->vertex(0) == v ?
-        v->cell()->vertex(1) : v->cell()->vertex(0));
+      // Check that the vertex hasn't be deleted from the TDS while we were locking it
+#ifdef CGAL_RT3_IGNORE_TDS_CONCEPT
+      if(!tds().vertices().is_used(v))
+#else
+      if(!tds().is_vertex(v))
+#endif
```
lrineau: I am surprised that a vertex that we managed to lock can be destroyed by another thread. That would defeat all the purpose of the lock.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
MaelRL: I think this whole `hide_point_mem` shenanigan is actually just useless. In other contexts, we need to ensure that the position of the lock is valid by checking the erase counter. Here, I just need to ensure that I locked _some_ valid place, and it doesn't matter if it's not where I thought it was.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
MaelRL: First thing it'll do when it enters the loop again is to check the validity of the `hint` so it should be safe.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
MaelRL: I'm checking here that `v` is a valid vertex in case it was destroyed _before_ it was locked. I could also check the validity before the `try_lock_vertex`, but there's no real point since it might be destroyed in between that check and the lock.

Triangulation_3/include/CGAL/Regular_triangulation_3.h
lrineau: I agree, but the value of `hint_point_mem` cannot be trusted, unless you assign it once in the `try_lock` block.

MaelRL: Sure, but I don't care about its value: of course it would ideally be where I think it is, but if it is not and it locked somewhere else, then the `locate` will just start from another position and that's alright as the hint might be wrong anyway in the first place. 

All I care about is that the `vh->cell()` is safe to use, not whether it is at an interesting position.

lrineau:
@MaelRL: why the flag `not yet approved`?

MaelRL:
Just to emphasis the code review. If you're okay with the diff and my last replies, it can be removed.

MaelRL:
Nevermind, I forgot also the fact that currently it has this ugly macro: see [this comment](https://github.com/CGAL/cgal/pull/3367#discussion_r222331973). Do I just go through with ignoring the TDS concept?

Triangulation_3/include/CGAL/Regular_triangulation_3.h
```diff
@@ -1324,14 +1324,48 @@ namespace CGAL {
...
-          if (m_rt.try_lock_vertex(hint) && m_rt.try_lock_point(p))
+          // The 'hint' is unsafe to use immediately because we are in a regular triangulation,
+          // and the insertion of a (weighted) point in another thread might have hidden (deleted)
+          // the hint.
+#define CGAL_RT3_IGNORE_TDS_CONCEPT
+#ifdef CGAL_RT3_IGNORE_TDS_CONCEPT
+          if(!m_rt.tds().vertices().is_used(hint))
+#else
+          if(!m_rt.tds().is_vertex(hint))
+#endif
```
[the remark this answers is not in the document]
lrineau: In my opinion, the concepts `TDS_2` are `TDS_3` are just a way to document the actual concrete implementation. I am pretty sure that there are already a lot of things that are not really documented in those concepts, and we can go on and add new undocumented stuff, in my opinion.

On the other hand, you can also document the new requirement.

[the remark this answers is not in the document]
lrineau: Your bench times are probably not reliable: the "system time" of the first result probably indicates a loaded machine.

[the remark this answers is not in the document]
MaelRL: I do get the same results consistently and the machine has nothing else running. I know practically nothing about kernel mode instructions, but won't going through all the items of the compact container require a lot of caching, which would then create this dramatic increase in system time?

Triangulation_3/include/CGAL/Regular_triangulation_3.h
MaelRL: Documenting it is difficult because we already don't expose the vertex/cell containers. Also, there are indeed some calls to undocumented functions from the TDS at a few points, but nothing "as bad" as calling the containers.

aa1e0acbeef802a1ae8cbdfeaec2f0808211d4d5 does something legal and the time loss compared to directly calling the function is hardly noticeable. Is this ok?

lrineau: I really do not like specializations, because that breaks easily. We have seen that on several occasions (for example the first version of the implementation of the structural filtering was almost only used in the test cases, but not in real uses cases).

Could it be related to the detection of a tag, instead?

MaelRL: How would a tag ensure that other TDS without such tag be properly tested or used in real cases ? In any case, we don't have any tests checking that T3 concepts match the actual code requirements (and they don't).

lrineau: Anyway, nothing will ensure that "other TDS" will be properly tested, because there exists only one concept `TDS_3` in the world.

But the use of a tag (instead of a specialization) will ensure that the trick will not break if:
  - one derives from a TDS, or if
  - one adds new templates parameters to TDS_3, will default values,
  - other maybe other patterns that can break easily with a specialization, and that I do not know of.

The truth is just that specialization on full classes is generally bad, whereas specialization on tags is acceptable.

And the worst is that the trick could turn unused without any compiler warning or failure. That is exactly what happened with the structural filtering, that was triggered by a full-specialization on `CGAL::Epick`

sloriot:
@MaelRL  what is the status of this PR wrt the Tests failing?

MaelRL:
Tests failing are irrelevant, I have to change the way the detection of CGAL::TDS_3 is coded (see https://github.com/CGAL/cgal/pull/3367#discussion_r224474228), but I haven't gotten around to it yet.

lrineau:
You could also assume that `CGAL::TDS_3` is mandatory. Anyway, that is the only thing we have been testing for years.

MaelRL:
I have used a tag in 82e458d. Should be good to test now if CR / Travis are okay.

lrineau:
Travis is OK.

lrineau:
Warnings on Windows: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-46.shtml#Triangulation_3

lrineau:
@MaelRL Can you please merge this PR in `integration`, and fix the conflicts with `master`?

Maybe you should create another PR for 4.13 or `master`, to avoid those conflicts.

sloriot:
Successfully tested in CGAL-4.14-Ic-87

PR #3368, opened 2018-10-03
Mesh_3 is now deterministic with TBB when only one thread is used

lrineau:
## Summary of Changes

- `Time_stamper` uses an atomic counter, to be usable with threads.
- Changes to `Compact_container` and `Concurrent_compact_container`:
  - Remove `CCC_iterator`, because a slightly modified `CC_iterator` can be used instead
  - `Concurrent_compact_container` now uses also the time stamper.

## Release Management

* Affected package(s): STL_Extension, Triangulation_3, Mesh_3


lrineau:

PR #3172, opened 2018-06-15
Fix vertex istream operator

sloriot:
Reported on cgal-discuss.

PR #3374, opened 2018-10-05
Triangulation: Manage binary mode in Triangulation IO

maxGimeno:
## Summary of Changes

Only the first line of the operator<< tested for ascii mode. Now everything that is output does.
## Release Management

* Affected package(s):Triangulation


maxGimeno:
Done in #3172

PR #3378, opened 2018-10-05
Polyhedron demo: Accelerate the creation of items.

### LINKED ISSUE #3373: Polyhedron demo: Deleting many items takes too much time
## Issue Detail
When we delete more than 10 items it becomes obvious that something is wrong with the process, because it takes too much time for something that should be instant.

maxGimeno:
## Summary of Changes

Always use the same member QSettings in the mainwindow makes it 200ms faster to build an item. This is really significant when loading hundreds of small items.
## Release Management

* Issue(s) solved (if any): fix #3373

### DIFF
--- a/GraphicsView/include/CGAL/Qt/DemosMainWindow.h
+++ b/GraphicsView/include/CGAL/Qt/DemosMainWindow.h
@@ -35,6 +35,7 @@
 #include <CGAL/auto_link/Qt.h>
 #include <CGAL/export/Qt.h>
 #include <CGAL/Qt/resources.h>
+#include <QSettings>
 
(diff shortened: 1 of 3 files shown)


lrineau:
The downside is that, if the demo crashes, the settings are not saved.

Maybe we should trigger a oneshot QTimer asking to save the settings are soon as possible...

... Actually Qt5 already does that:
http://doc.qt.io/qt-5/qsettings.html#sync

> `void QSettings::sync()`
> 
> [...]
> 
> This function is called automatically from QSettings's destructor and by the event loop at regular intervals, so you normally don't need to call it yourself.

PR #3386, opened 2018-10-08
AABB_tree: Add an overload to build() for shared data

### LINKED ISSUE #1166: AABB tree usability: allow to pass the "shared data" to `AABB_tree::build()`
For mechanism for "shared data" in the AABB_tree is horrible!

Let's say we cannot use the constructor, or the `insert` function, from a range of primitives, then there is the only documented way to pass the shared data is to construct the AABB traits and call its `set_shared_data()` function.

I suggest that, by similarity with the other usages (`insert(begin, begin, ...)`), the `build()` function is modified so that one can pass the shared data to it.

maxGimeno:
## Summary of Changes
This adds a parameter pack to the function build() that calls set_shared_data(t...) if specified.
## Release Management

* Affected package(s):AABB_tree
* Issue(s) solved (if any): fix #1166 
* documentation: 
```c++
/// After one or more calls to `insert()` the internal data
/// structure of the tree must be reconstructed. This procedure
/// has a complexity of \f$O(n log(n))\f$, where \f$n\f$ is the number of
/// primitives of the tree.  This procedure is called implicitly
/// at the first call to a query member function. You can call
/// `build()` explicitly to ensure that the next call to
/// query functions will not trigger the reconstruction of the
/// data structure.
/// The parameter pack `T` is of any types
/// such that `Primitive` has a constructor with the following signature:
/// `Primitive(%InputIterator, T...)`. A call to `AABBTraits::set_shared_data(t...)`
/// is made using the internally stored traits.
/// For compilers that do not support variadic templates,
/// overloads up to 5 template arguments are provided.
template<typename ... T>
void build(T&& ...);
```

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -145,7 +145,7 @@ namespace CGAL {
     template<typename InputIterator, typename T1, typename T2, typename T3, typename T4>
 		AABB_tree(InputIterator first, InputIterator beyond,T1&,T2&,T3&,T4&);
     template<typename InputIterator, typename T1, typename T2, typename T3, typename T4, typename T5>
-		AABB_tree(InputIterator first, InputIterator beyond,T1&,T2&,T3&,T4&,T5&);
+                AABB_tree(InputIterator first, InputIterator beyond,T1&,T2&,T3&,T4&,T5&);
     #endif
 
     /// After one or more calls to `insert()` the internal data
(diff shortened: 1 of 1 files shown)


lrineau:
The indentation of the patch is really bad!

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -156,7 +156,31 @@ namespace CGAL {
     /// `build()` explicitly to ensure that the next call to
     /// query functions will not trigger the reconstruction of the
     /// data structure.
+    /// The parameter pack `T` is of any types
+    /// such that `Primitive` has a constructor with the following signature:
+    /// `Primitive(%InputIterator, T...)`. A call to `AABBTraits::set_shared_data(t...)`
```
sloriot: no need to mention primitive here

maxGimeno: How do I explain `T` then ?

sloriot: With [`set_shared_data()`](https://doc.cgal.org/latest/AABB_tree/classAABBTraits.html#a6dba5130cc78298fb7039f479f657fe4) directly since it is was it is called in practice.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -1062,7 +1130,7 @@ namespace CGAL {
 
 			// constructs the tree
 			m_p_root_node->expand(m_primitives.begin(), m_primitives.end(),
-														m_primitives.size(), m_traits);
+			m_primitives.size(), m_traits);
```
lrineau: The indentation here is wrong.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -156,7 +156,29 @@ namespace CGAL {
...
+    /// For compilers that do not support variadic templates,
+    /// overloads up to 5 template arguments are provided.
+    
+#if !defined(CGAL_CFG_NO_CPP0X_VARIADIC_TEMPLATES) && !defined(CGAL_CFG_NO_CPP0X_RVALUE_REFERENCE)
+    template<typename ... T>
+    void build(T&& ...);
     void build();
+#else
+    void build();
+    template
```
sloriot: ...

sloriot:
Successfully tested in [4.14-Ic-32](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-32.shtml).

PR #3389, opened 2018-10-10
P3T3: Minor improvements (random_shuffle + determinism)

### LINKED ISSUE #3346: Periodic Alpha Shape 3 is not deterministic
## Issue Details

When parsing a periodic alpha shape 3, CGAL::dispatch_output does not give a deterministic list of objects
The bug is in file https://github.com/CGAL/cgal/blob/master/Periodic_3_triangulation_3/include/CGAL/Periodic_3_Delaunay_triangulation_3.h
on the line:
`CGAL::cpp98::random_shuffle (points.begin(), points.end());`
Moreover, a spatial_sort is performed just after.
If I remove the random_shuffle, then the object parsing is deterministic.

## Source Code

https://gudhi.inria.fr/doc/latest/_alpha_complex_2periodic_alpha_complex_3d_persistence_8cpp-example.html

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Ubuntu 18.04 x64
* Compiler: gcc 7.3.0
* CGAL version: 4.13.beta1
* Boost version: 1.65.1
* Other libraries versions if used (Eigen, TBB, etc.): Eigen 3.1.0

MaelRL:
## Summary of Changes

- Call to `random_shuffle` if `is_point_set_large` is `true` is redundant since the next thing done is to call `spatial_sort`, which calls `random_shuffle` itself.
- `make_canonical` was needlessly using addresses to compare `Vertex_handle`s. With that change, P3T3 is deterministic when using time stamped vertex/cell base classes, but for now requires to always be in 1-cover.

## Release Management

* Affected package(s): `Periodic_3_triangulation_3`
* Issue(s) solved (if any): fix #3346 

### DIFF
--- a/Periodic_3_triangulation_3/include/CGAL/Periodic_3_Delaunay_triangulation_3.h
+++ b/Periodic_3_triangulation_3/include/CGAL/Periodic_3_Delaunay_triangulation_3.h
@@ -490,21 +490,27 @@ public:
       is_large_point_set = false;
 
     std::vector<Point> points(first, last);
-    CGAL::cpp98::random_shuffle (points.begin(), points.end());
-    Cell_handle hint;
     std::vector<Vertex_handle> dummy_points, double_vertices;
     typename std::vector<Point>::iterator pbegin = points.begin();
     if(is_large_point_set)
+    {
       dummy_points = insert_dummy_points();
-    else while(!is_1_cover()) {
-      insert(*pbegin);
-      ++pbegin;
-      if(pbegin == points.end())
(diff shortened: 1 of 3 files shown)


lrineau:
Is not it a bit harsh to require the 1-cover property? That is a new behavior, and will break users code.

lrineau:
Right. I misunderstood the post-condition.

lrineau:

VincentRouvreau:
+1 !
I have tested it and it works fine.
Thank you for the fix !

PR #3394, opened 2018-10-11
Polyhedron demo: Add an action to save a scene

### LINKED ISSUE #3340: Feature Proposition: loading a "scene" in Polyhedron demo
## Issue Details

I thought about a feature that would be useful for me and maybe for other people: being able to write/load full "scenes" in the Polyhedron demo. By "scene", I mean a set of items (of potentially different types) with some additional informations (color, rendering mode, etc.).

What I have in mind is a text file that just refers to other files, similarly to a M3U playlist that refers to existing music files. One example could be:

```
~/data/my_mesh.off surface_mesh_io_plugin "Mesh Faces" #FF0000 Flat Visible
~/data/aux/edges.polylines.txt "Meta-Edges" polylines_io_plugin #000000 FlatPlusEdges Visible
../debug/something.ply ply_plugin "Debug Points" #00FF00 ShadedPoints Hidden
```

_(Maybe using a standard file format such as XML or Json could make sense.)_

In that case, I assume that each line contains the path to the file, the plugin to use to open it, the name of the item, the color of the item, the rendering mode and the view mode (but it's just an example, it could be different). When I load this file, it automatically loads all the files listed and sets up the property as defined (if the file doesn't exist or the IO plugin to use is not available, an error should be displayed - again, similarly to a M3U file that would refer a non-existent MP3 file).

Being able to load something like that would allow to easily recover an exact same scene (instead of loading items one by one and resetting the properties the way we want). To save such a file, one option would be to make a special IO plugin that is triggered when hitting "Save" when several items are selected: the problem in that case is that it might get complicated afterwards (= does the user need to select each IO plugin for each item and assign a name to them? Or is it done automatically using the default IO plugin for each type?).

That might be an overly complicated feature for a small added value, so this is not exactly a feature _request_, just something I thought about and that I wanted to write down in order not to forget it. Feel free to comment and tell me what you think about it.

maxGimeno:
## Summary of Changes
In the file menu, an action is added that exports all the items in the scene that have a source name and a loader (saying the ones that have been loaded, not those that have been generated) to a .js file. 
Loading this .js file will add all the items to the scene, having the same color and the same rendering mode, and move the camera to where it was when it has been saved.

## Release Management
* Issue(s) solved (if any): fix #3340


Three/include/CGAL/Three/Scene_item.h
```diff
@@ -316,7 +316,7 @@ public Q_SLOTS:
   virtual void setColor(QColor c) { color_ = c;}
   //!Setter for the RGB color of the item. Calls setColor(QColor).
   //!@see setColor(QColor c)
-  void setRbgColor(int r, int g, int b) { setColor(QColor(r, g, b)); }
+  virtual void setRgbColor(int r, int g, int b) { setColor(QColor(r, g, b)); }
```
lrineau: No, not virtual!

sloriot:
Please fix conflicts.

sloriot:

@lrineau could you test this PR as @sgiraudot who reported the issue is not available?

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2369,6 +2366,79 @@ void MainWindow::propagate_action()
...
   }
 }
+
+void MainWindow::on_actionSa_ve_Scene_as_Script_triggered()
+{
+  QString filename =
+      QFileDialog::getSaveFileName(this,
+                                   "Save the Scene as a Script File",
+                                   last_saved_dir,
+                                   "*.js");
```
lrineau: The filter is not right, and not not work as expected. It should be `Qt Script files (*.js)`.

Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2369,6 +2366,79 @@ void MainWindow::propagate_action()
...
+  std::vector<QString> loaders;
+  std::vector<QColor> colors;
+  std::vector<int> rendering_modes;
+  for(int i = 0; i < scene->numberOfEntries(); ++i)
+  {
+    Scene_item* item = scene->item(i);
+    QString loader = item->property("loader_name").toString();
+    QString source = item->property("source filename").toString();
+    if(loader.isEmpty())
+      continue;
```
lrineau: Maybe the list of items that cannot be saved should be pushed into a list, so that a popup can explain which items are not saved.

lrineau:
I approved the new changes. Once the PR passes the test, it can be merged.

maxGimeno:

PR #3411, opened 2018-10-18
Add an option to disable the deprecation warning

sloriot:
@sgiraudot could you confirm that Eigen is used by default only if `CGAL_USE_EIGEN` is not defined?

### DIFF
--- a/Solver_interface/include/CGAL/Diagonalize_traits.h
+++ b/Solver_interface/include/CGAL/Diagonalize_traits.h
@@ -26,9 +26,13 @@
 #include <CGAL/number_type_config.h>
 #include <CGAL/double.h>
 
+#ifndef CGAL_I_WANT_TO_USE_DIAGONALIZE_TRAITS
 #define CGAL_WARNING_DIAGONALIZE_TRAITS \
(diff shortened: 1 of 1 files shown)


Solver_interface/include/CGAL/Diagonalize_traits.h
```diff
@@ -26,9 +26,11 @@
 #include <CGAL/number_type_config.h>
 #include <CGAL/double.h>
 
+#ifndef CGAL_I_WANT_TO_USE_DIAGONALIZE_TRAITS
 #define CGAL_WARNING_DIAGONALIZE_TRAITS \
   CGAL_DEPRECATED_MSG("CGAL::Diagonalize_traits is a deprecated class that can \
 lead to precision issues, please use CGAL::Eigen_diagonalize_traits")
+#endif
```
lrineau: You have not tested your patch. There must be an `#else` that defines the macro to an empty value.

sgiraudot:
That's the opposite, Eigen is used by default _IF_ `CGAL_USE_EIGEN` is defined.

sloriot:
Yeah that's what I meant :)

PR #3412, opened 2018-10-18
Polyhedron demo: Make the demo customizable at configure and start time

### LINKED ISSUE #2040: Polyhedron_demo: Make the demo configurable at build time
## Issue Details
Find a way to chose which plugins should be compiled and added at the build time to make portable versions of the demo.

maxGimeno:
## Summary of Changes

This PR introduces a mechanism that uses Qt's Metadata system to find out if a plugin should be loaded or not at the start. It also modifies the macro that creates plugins to allow the definition of Keywords that will be used to classify plugins, and create associated targets. 

For example, building the target IO will build all the plugins that have been declared with the keyword IO in their cmakelists, and their dependencies. It will also create JSON files associated to those plugins in their build dir, and those JSON files will be used to get metadata retrievable in the Preferences of the application. 
The Preferences page has also been modified. Itgets rids of the unintuitive Tooltips for the report of loading errors and introduces instead a button Details. When clicking on this button, a dialog pops up with the name of the selected plugins, the keywords that they are associated to and the date corresponding to the day they have been configured (allows to quicly know if a plugin is outdated and probably binary incompatible). Clicking on one of those plugins will dosplay their status (successfully loaded, blacklisted, ignored because of the keywords or the error text  if any.)

using the argument --keyword when launching the app allows to choose a category of plugin. Following the previous example, ./Polyhedron_demo --keyword IO will only load the IO plugins. All of this allows to build and/or use lighter versions of the demo, and after some more work, will allow to easily create specific applications based on the demo, like a demo specific for polygonmesh or point_set_processing.

image

image

image

image

image

## Release Management
* Issue(s) solved (if any): fix #2040


Polyhedron/demo/Polyhedron/implicit_functions/CMakeLists.txt
```diff
@@ -39,9 +39,9 @@ if(CGAL_Qt5_FOUND AND Qt5_FOUND)
   include(AddFileDependencies)
   remove_definitions(-DQT_STATICPLUGIN)
 
-  polyhedron_demo_plugin(p_sphere_function_plugin Sphere_implicit_function)
-  polyhedron_demo_plugin(p_tanglecube_function_plugin Tanglecube_implicit_function)
-  polyhedron_demo_plugin(p_klein_function_plugin Klein_implicit_function)
+  polyhedron_demo_plugin(p_sphere_function_plugin Sphere_implicit_function KEYWORDS Mesh_3)
+  polyhedron_demo_plugin(p_tanglecube_function_plugin Tanglecube_implicit_function KEYWORDS Mesh_3)
+  polyhedron_demo_plugin(p_klein_function_plugin Klein_implicit_function KEYWORDS Mesh_3)
```
lrineau: I love that API. It is so simple that it is frightening ! :smile:

maxGimeno:
What's wrong with travis ? I tested the permissions localy and nothing is wrong, and the file name just before the error does not even exit. I'm lost.

maxGimeno:

lrineau:
This PR is now merged. Probably it created conflicts in several other PRs about `demo/Polyhedron/`.

PR #3417, opened 2018-10-22
Polyhedron demo: Enhance clipping plugin

maxGimeno:
## Summary of Changes
Adds a tab to the clipping-box plugin. It allows to use a 2D rectangle selection instead of a selection box.


lrineau:
It seems this PR depends on another one. I will wait before the integration.

PR #3426, opened 2018-10-29
BGL: Fix make_grid() documentation

### LINKED ISSUE #3422: BGL: inconsistency between doc/code in make_grid's CoordinateFunctor
## Issue Details

[Doc](https://doc.cgal.org/latest/BGL/group__PkgBGLHelperFct.html#ga5a41c8f7154516b2f6f58937136e4bb3) says that `CoordinateFunctor` takes two `size_type` values (the grid indices). [Code](https://github.com/CGAL/cgal/blob/master/BGL/include/CGAL/boost/graph/helpers.h#L1295) expects a ternary functor.

maxGimeno:
## Summary of Changes
Adds missing information in the doc of make_grid and adds the 3rd argument of the functor.

## Release Management

* Affected package(s):BGL
* Issue(s) solved (if any): fix #3422

### DIFF
--- a/BGL/include/CGAL/boost/graph/helpers.h
+++ b/BGL/include/CGAL/boost/graph/helpers.h
@@ -1011,6 +1011,7 @@ make_icosahedron(
  *
  * \brief Creates a row major ordered grid with `i` cells along the width and `j` cells
  * along the height and adds it to the graph `g`.
+ * An internal property map for `CGAL::vertex_point_t` must be available in `Graph`.
  *
  * \param i the number of cells along the width.
  * \param j the number of cells along the height.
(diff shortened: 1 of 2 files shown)


MaelRL:
Shouldn't it rather be that the doc is correct and the code is wrong? Why is there a need for a third argument to parameterize the position of a point on a 2D grid?

maxGimeno:
Yeah well I was specifically asked to use that functor that won't work with the write API, so... 
Apart from that, i agree that a special default functor should be (and was) written to fit a decent API that won't take `0` as argument. 
@CGAL/geometryfactory Who is against a new Functor that takes only 2 coords as input ?

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -1258,8 +1258,9 @@ make_icosahedron(
  * If `triangulated` is `true`, the diagonal of each cell is oriented from (0,0) to (1,1)
  * in the cell coordinates.
  *
- * \tparam CoordinateFunctor that takes two `boost::graph_traits<Graph>::%vertices_size_type`
+ * \tparam CoordinateFunctor that takes three `boost::graph_traits<Graph>::%vertices_size_type`
```
sloriot: a function object providing `Point_3 operator()(size_type I, size_type J, size_type)` with `Point_3` being ....
Also backquote the entire interval not just `i` and `j`.

sloriot:
Is the function new in 4.14? if not if should be based on 4.12 or 4.13

sloriot:
@afabri  I guess you are the one that requested 3 arguments. Any reason not to have only 2?

afabri:
Even I have to guess, but probably I had a use case with a terrain model in mind.

sloriot:
Travis error is stranger. Did you already get this?

maxGimeno:
Yes, it means the package_info/dependencies is empty, but it is not. I rrestarted the job, maybe it's just a temporary bug.

sloriot:

sloriot:
Since there is somehow an API change I changed the base to master.

PR #3427, opened 2018-10-29
Fix clipping of open meshes

sloriot:
The orientation of the clipper was ignored.


sloriot:
Successfully tested in [4.14-Ic-30](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-30.shtml).

PR #3429, opened 2018-10-29
Make the degenerate edge removal an iterative process

sloriot:
An edge impossible to remove might be removable after some degenerate edge removals.

diff is only 5 lines using `git show -w 3be4dd0`


sloriot:
Successfully tested in [4.14-Ic-30](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-30.shtml).

PR #3437, opened 2018-11-03
Workaround -Wmaybe-uninitialized for GCC

sloriot:
I started this PR to discuss about how we can get rid of all those warnings triggered by GCC about maybe uninitialized variables.
I'm not saying that what I started to do is the solution, but I least I could get rid of all the warnings for the polyhedron demo.

A simpler alternative can be to simply disable the warning if `CGAL_TESTSUITE` is defined but as a developer we still would be polluted. However, we have the warnings only in release build (I guess it is due to optimizations) so it's not that polluting as most of us develop in debug mode.

cc @lrineau @maxGimeno


lrineau:
I do not like hiding those warnings without knowing why they trigger.

And I do not like having normal values as default values: if there is a real bug, then those normal values will trigger any assertion, or raise any suspicion.

mglisse:
This warning is known for having a lot of false positives, and it is often hard to silence the warning. Most users of boost::optional eventually get the warning, for instance. I regularly try to convince people at gcc that it should not be part of -Wall (maybe -Wextra). It isn't a useless warning, it does find real bugs, but if we insist on 0 warning, we shouldn't be afraid to disable it with pragmas or whatnot.

mglisse:
> we have the warnings only in release build

IIRC gcc deactivates the warning if optimizations are turned off. But yes, different optimization levels can produce different warnings.

sloriot:
The following [program](https://gist.github.com/sloriot/41a90a614aa656f7f608cf59bcf25632) generates some `maybe-unintialized` warnings and I have no idea how to silence them properly (ie no pragma or no init in the class).

Short-term, locally using the pragma when `CGAL_TESTSUITE` is `ON` seems like a good alternative to me.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Nov 12, 2018, 4:50 PM GMT+1](https://github.com/CGAL/cgal/pull/3437#issuecomment-437930359 "2018-11-12T15:50:17Z - Replied by Github Reply Comments"):
> The following [program](https://gist.github.com/sloriot/41a90a614aa656f7f608cf59bcf25632) generates some `maybe-unintialized` warnings and I have no idea how to silence them properly (ie no pragma or no init in the class).
> 
> Short-term, locally using the pragma when `CGAL_TESTSUITE` is `ON` seems like a good alternative to me.

For that warning, the patch is this one:
```patch
modified   TDS_2/include/CGAL/Triangulation_data_structure_2.h
@@ -296,7 +296,8 @@ public:
 // 		  FaceIt face_begin,
 // 		  FaceIt face_end);
   
-  Vertex_handle create_vertex(const Vertex &v = Vertex());
+  Vertex_handle create_vertex();
+  Vertex_handle create_vertex(const Vertex &v);
   Vertex_handle create_vertex(Vertex_handle v); //calls copy constructor 
   Face_handle create_face(const Face& f = Face());
   Face_handle create_face(Face_handle f); //calls copy constructor 
@@ -1353,6 +1354,15 @@ make_hole(Vertex_handle v, List_edges& hole)
 }
 
 
+template <class Vb, class Fb>
+inline
+typename Triangulation_data_structure_2<Vb,Fb>::Vertex_handle
+Triangulation_data_structure_2<Vb,Fb>::
+create_vertex()
+{
+  return vertices().emplace();
+}
+
 template <class Vb, class Fb>
 inline
 typename Triangulation_data_structure_2<Vb,Fb>::Vertex_handle

```

Instead of copying a default-constructed (and uninitialized) `Vertex`, construct it directly in the container using `emplace()`.

sloriot:
OK, I have reset my branch with new patches, including your.

sloriot:
Still WIP as not all warnings are gone.

lrineau:
Why `work in progress`? The PR was tested and it does reduce the number of warnings. Do you expect more work on it soon? Maybe we should merge this PR, and then later create another one.

lrineau:
Correctly tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-48.shtml

sloriot:
You can merge it, I'll open a new one a soon as I can.

PR #3438, opened 2018-11-05
Fix bug in clipping code

### LINKED ISSUE #3428: invalid clip result
If a mesh to be clipped has a patch without original vertex after corefinement, its classification is not valid.

There are several ways to fix that:
* pass an exact vpm and generate a point inside the patch to be used by side-of (the interval version should be considered to guarantee that the interval bbox of the point does not intersect the clipper).
* adapt the Face_graph_output_builder to only classify patches
* ...

[br.zip](https://github.com/CGAL/cgal/files/2533977/br.zip)

sloriot:
Fixes a bug when a patch with no input vertex must be classified wrt the clipper
The patch uses existing code with a special switch to disable some part of it.

Fix #3428


sloriot:

PR #3445, opened 2018-11-08
Use correct vpm

sloriot:


sloriot:
Successfully tested in [4.14-Ic-40](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-40.shtml).

PR #3450, opened 2018-11-11
Copy relative_precision_of_to_double from Lazy_exact_nt to Lazy_alpha_nt_3

### LINKED ISSUE #3153: to_double(Lazy_alpha_nt_3) relative precision
`to_double(Lazy_exact_nt)` has a mechanism with `get_relative_precision_of_to_double`, etc so that it only uses to_double on the interval if the precision is good enough. Otherwise it computes the exact value and converts that one to double. `to_double(Lazy_alpha_nt_3)` seems to be missing a similar scheme and returns the middle of the interval, even if the interval is [-inf,+inf].

As a workaround, it is possible for users to do it themselves, calling `a.approx()`, computing the relative precision, etc. But this all seems undocumented.

mglisse:
Only minimally tested.

## Release Management

* Affected package(s): Alpha_shapes_3, Number_types
* Issue(s) solved (if any): fix #3153


sloriot:
Successfully tested in [CGAL-4.14-Ic-43](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-43.shtml).

PR #3451, opened 2018-11-12
Fix doc warning

sloriot:
Fixes [this warning](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.14-Ic-40/logs_master/AABB_tree.log)

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -160,11 +160,12 @@ namespace CGAL {
     /// is made using the internally stored traits.
     /// For compilers that do not support variadic templates,
     /// overloads up to 5 template arguments are provided.
-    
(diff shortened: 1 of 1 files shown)


lrineau:
@maxGimeno: Why is there a `build()` without parameter and a variadic version? The variadic version can be called with zero arguments, no?

maxGimeno:
I just reproduces a similar pattern on some other doc. The variadic version is dependent on some configuration, so if no variadic templates are allowed I have to define an overload for 0, 1, 2,.. 5 args or something like that.

lrineau:
Here that is different: if variadic templates are allowed, then there is a `build()` function with zero arguments, *plus* a variadic version `build(T&&...)`.

sloriot:
I'm not 100% sure but it would assume that `set_shared_data()` can be called which is only possible for primitive with shared data.

maxGimeno:
@sloriot Yes that actually rings a bell !

sloriot:
Successfully tested in [CGAL-4.14-Ic-43](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-43.shtml).

PR #3454, opened 2018-11-14
Polyhedron demo: Fix visibility change

maxGimeno:
## Summary of Changes

This PR introduces two new functions to the scene in the demo API: allItemsChanged() and cutDataUpdate(bool). 
The first one updates all the items in the scene view at once, which only requires 2 ModelIndex to be created instead of 2 by item, and the second one keeps the scene from updating the sceneview.
Using those two functions, this PR reduces the time it takes to hode or show all items to almost 0.
For example, it took my machine 5 secs to hide a scene of 240 polylines, and now it is instant.


lrineau:
This PR now has a conflict with `master`.

Three/include/CGAL/Three/Scene_interface.h
```diff
@@ -144,11 +144,22 @@ class Scene_interface {
   //! Updates the information about `item` in the
   //! Geometric Objects list and redraws the scene.
   virtual void itemChanged(CGAL::Three::Scene_item* item) = 0;
+  //!
+  //! \brief Updates all the items in the SceneView.
+  //!
+  virtual void allItemsChanged() = 0;
```
lrineau: Each time you add a virtual method, anywhere, that is not the last one, then you break the ABI compatibility.

You should always add virtual functions after the other ones.

Three/include/CGAL/Three/Scene_interface.h
```diff
@@ -144,11 +144,22 @@ class Scene_interface {
...
   //! Used to update the selection in the Geometric Objects view.
-  virtual void setSelectedItem(Item_id) = 0;  
+  virtual void setSelectedItem(Item_id) = 0;
+  //! \brief ignore data updating.
+  //! 
+  //! This will ignore all the individual calls to `itemChanged()` until 
+  //! `cutDataUpdate()` is called whith `b` being `false`.
+  //! Activate this when you know you will call allItemsChanged() after a loop. 
+  //!
+  virtual void cutDataUpdate(bool b) =0;
```
lrineau: I should prefer the name [`setUpdatesEnabled`, like in Qt5](http://doc.qt.io/qt-5/qwidget.html#updatesEnabled-prop).

And, in my opinion, `setUpdatesEnabled(true)` should call `allItemsChanged()` itself.

sloriot:
It seems that no issue was detected by CGAL-4.14-Ic-45.

maxGimeno:
I remove it from the testsuite because it strongly conflicts with the multiple viewers and I don't want to deal with it right now.

maxGimeno:

PR #3413, opened 2018-10-18
Clean CMakeLists.txt of examples and tests

gdamiand:
## Summary of Changes

With the last updates of our cmake machinery, our CMakeLists.txt can be simplified.

We don't need anymore to do:
```cmake
 include( CGAL_CreateSingleSourceCGALProgram )
 include_directories(BEFORE ../../include)
```

and for examples and tests that do not use cgal_core nor cgal_imageio we either don't need do to
```cmake
 include(${CGAL_USE_FILE})
```

Moreover, the cmake minimal required version is now 3.1.

This PR update almost all CMakeLists.txt accordingly.

(Note that maybe some other things could be cleanup up...)

## Release Management

* Affected package(s): many (all ?)

PR #3463, opened 2018-11-19
Fix CTest after #3413

lrineau:
## Summary of Changes

Fix a bug with CTest in non-header-only

The pull-request #3413 has introduced a side-effect: CTest was broken when `CGAL_HEADER_ONLY=OFF`, for all tests using a `.cin` or `.cmd` file. That is due to the variable `CGAL_CURRENT_SOURCE_DIR` that was only filled by `UseCGAL.cmake`.

Now that we forbid CGAL directories without `CMakeLists.txt`, that variable `CGAL_CURRENT_SOURCE_DIR` is now useless, and can be removed: `CMAKE_CURRENT_SOURCE_DIR` is sufficient. That fixes the bug with CTest, and also makes the CMake code a bit simpler.

Cc @gdamiand, @sloriot 

## Release Management

* Affected package(s): Installation, CMake scripts, Arr_2


gdamiand:
Do we really want to keep these mechanisms ? (.cin or .cmd)

Indeed, it seems that recent examples/tests use usually default command line parameters. Maybe we can update all tests to use also default parameter ?

afabri:
I also do not like .cin

maxGimeno:

PR #3472, opened 2018-11-22
Fix clipping on border edges

sloriot:
Fix special cases involving border edges in the clipping.


sloriot:
@maxGimeno you added the "Tested" label without indicating in which internal release it was tested.

maxGimeno:
@sloriot yes sorry, I keep forgetting that :/

It was in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-61.shtml

PR #3477, opened 2018-11-25
Add OpenGR wrapper

sloriot:
Add a CGAL wrapper to [OpenGR](https://github.com/STORM-IRIT/OpenGR) providing methods for registering point clouds.

cc @nmellado 

**TODO:**
- [x] get rid of `FindOpenGR.cmake`
- [ ] small feature
- [ ] add support for other registration methods (through named parameters?)
- [x] add interface to other functions (`compute_transformation()`, ...)
- [x] check potential issue of compiling OpenGR library with a given version of Eigen and using CGAL with another version of Eigen.


sloriot:
@nmellado I think it would make think simpler on the cmake side for CGAL if when building OpenGR in addition to `generated/OpenGRConfig.cmake` (that is generated for the purpose of `make install` IIUC) you also generate a `OpenCRConfig.cmake` directly in the build directory that enables anyone to use the current build directory.

Basically, this means making a copy of libs in a `lib` directory in the build dir and updating
```
set(OpenGR_INCLUDE_DIR "/usr/local/include/")
set(OpenGR_LIB_DIR "/usr/local/lib/")
set(OpenGR_LIBRARIES opengr_accel opengr_io opengr_algo)
```
using `${CMAKE_SOURCE_DIR}/src` and `${CMAKE_BUILD_DIR}/lib` for exemple.

afabri:
Do we really have to copy the points?

nmellado:
Thanks @sloriot for preparing this PR.
This remind me I've still not merged my OpenGR branch with the stable API (https://github.com/STORM-IRIT/OpenGR/pull/13)

I've just started, I've a couple of things to fix but will try to do it today.

@afabri I have to improve a bit the OpenGR API so points does not need to be copied from CGal. But as discussed with @sloriot, we can first try with the current version where points are copied.

sloriot:
@nmellado I did the update.

nmellado:
Regarding this point:
> check potential issue of compiling OpenGR library with a given version of Eigen and using CGAL with another version of Eigen.
Having OpenGR as header only would solve this. RIght now OpenGR is header only, except the `io` package. 

One solution would be to add package components to OpenGR, i.e. having one component per target that exists today. 
Thus you could do something:
```cmake
find_package(OpenGR COMPONENTS algo accel utils )
```
Of course this would also require an option in OpenGR to build the library (ie. generate the cmake package) without compiling IO.

Sounds good to you ?

sloriot:
I think right now it just does not link with the IO lib. So it is probably no longer an issue.

nmellado:
Fair enough.

sloriot:
@nmellado does it make sense to offer an interface to another method than super4PCS for now?
Could you point me to the documentation of the options?

nmellado:
We can stick to Super4PCS now, the other approaches might need more testing.
Supporting 4PCS should come at no cost however.

Parameter setting is described here: https://storm-irit.github.io/OpenGR/a00012.html

sgiraudot:
I'm also in favor of adding new parameters, especially because they can be re-used by other algorithms (something like accuracy is general enough to have another usage somewhere else later on). Which wouldn't be the case with the option struct. Also NP is already a quite complex/flexible way to handle parameters, using them to carry a struct of several parameters seems like adding an unnecessary layer of complexity.

By the way, we could already thinking of recycling some existing named parameters, for example `number_of_output_points` could be renamed `number_of_samples`, be also used for OpenGR and be documented more generally in the NP page (and more specifically on the functions that use it).

nmellado:
@sloriot it's ok you can copy/paste the doc, and link to the OpengGR version, in case I update it.

The parameters are stables and should not move that much.
Actually, there is no any parameters that is specific to Super4PCS itself:
 - Overlap: specific to RANSAC-based algorithms (e.g. 3PCS, (Super)4PCS). Other kinds of algorithms might use similar parameter to sub-sample the input clouds.
 - Number of samples: related to the sub-sampling stage. Note that for CGAL's wrapper, we could skip this step and assume the user gives a subsampled point cloud, obtained by CGAL algorithms. It is maybe more generic and powerful than using the simple quantification approach used in current version of OpenGR. 
 - Registration accuracy: related to the alignment metric. This metric could be used in other contexts (to compute distances between models) or may be replaced by another metric. 
**Note**: this is potential change for next release of OpenGR, where the metric can be changed (e.g. set as a template parameter of the registration method).

BGL/include/CGAL/boost/graph/parameters_interface.h
```diff
@@ -118,3 +118,7 @@ CGAL_add_named_parameter(plane_index_t, plane_index_map, plane_index_map)
 CGAL_add_named_parameter(select_percentage_t, select_percentage, select_percentage)
 CGAL_add_named_parameter(require_uniform_sampling_t, require_uniform_sampling, require_uniform_sampling)
 CGAL_add_named_parameter(point_is_constrained_t, point_is_constrained, point_is_constrained_map)
+CGAL_add_named_parameter(number_of_samples_t, number_of_samples, number_of_samples)
+CGAL_add_named_parameter(accuracy_t, accuracy, accuracy)
+CGAL_add_named_parameter(maximum_running_time_t, maximum_running_time, maximum_running_time)
+CGAL_add_named_parameter(overlap_t, overlap, overlap)
```
sloriot: new parameters should be tested in `test_cgal_bgl_named_params.cpp`

BGL/test/BGL/test_cgal_bgl_named_params.cpp
```diff
@@ -97,6 +97,7 @@ void test(const NamedParameters& np)
   assert(get_param(np, CGAL::internal_np::face_normal).v == 36);
   assert(get_param(np, CGAL::internal_np::random_seed).v == 37);
   assert(get_param(np, CGAL::internal_np::do_project).v == 38);
+  assert(get_param(np, CGAL::internal_np::opengr_options).v == 9001);
```
sloriot: this one no longer exists

BGL/include/CGAL/boost/graph/parameters_interface.h
sgiraudot: That is correct, I'll do it.

BGL/test/BGL/test_cgal_bgl_named_params.cpp
sgiraudot: I'll update the test with the new NP and remove this one.

Point_set_processing_3/doc/Point_set_processing_3/Point_set_processing_3.txt
```diff
@@ -318,7 +318,24 @@ points in the domain.
...
+\cgal provides two wrappers for the \ref thirdpartyOpenGR library
+\cgalCite{cgal:m-ogr-17} :
+
+- `CGAL::OpenGR::compute_registration_transformation()` computes the
+registration of one point set w.r.t. another in the form of a
+`CGAL::Aff_transformation_3` object, using the Super4PCS algorithm
+\cgalCite{cgal:mam-sffgp-14};
+
+- `CGAL::OpenGR::register_point_sets()` computes the registration of
+one point set w.r.t. another and directly aligns it to it.
```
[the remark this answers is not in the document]
sgiraudot: Sounds better, yes.

Point_set_processing_3/include/CGAL/OpenGR/register_point_sets.h
```diff
@@ -0,0 +1,236 @@
...
+     `geom_traits::Point_3`.  If this parameter is omitted,
+     `CGAL::Identity_property_map<geom_traits::Point_3>` is used.\cgalParamEnd
+     \cgalParamBegin{normal_map} a model of `ReadablePropertyMap` whose key
+     type is the value type of the iterator of `PointRange2` and whose value
+     type `geom_traits::Vector_3`.\cgalParamEnd
+   \cgalNamedParamsEnd
+
+   \return the registration score.
+*/
+template <class PointRange1, class PointRange2,
```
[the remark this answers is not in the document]
sgiraudot: It's a good idea, avoiding duplicating the full documentation of the function.

maxGimeno:
Travis report : In file included from /home/travis/build/CGAL/cgal/Installation/../Point_set_processing_3/include/CGAL/OpenGR/register_point_sets.h:31:
/home/travis/build/CGAL/cgal/Point_set_processing_3/include/CGAL/OpenGR/compute_registration_transformation.h:33:10: fatal error: 'gr/algorithms/FunctorSuper4pcs.h' file not found
#include <gr/algorithms/FunctorSuper4pcs.h>
         ^

Point_set_processing_3/include/CGAL/OpenGR/compute_registration_transformation.h
```diff
@@ -0,0 +1,307 @@
...
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0+
+//
+// Author(s) : Sebastien Loriot
+
+#ifndef CGAL_OPENGR_COMPUTE_REGISTRATION_TRANSFORMATION_H
```
afabri: This file needs an `#ifdef CGAL_OPEN_GR_ENABLED` otherwise travis individual header compilation will fail.

Point_set_processing_3/include/CGAL/OpenGR/compute_registration_transformation.h
```diff
@@ -0,0 +1,307 @@
...
+#define CGAL_OPENGR_COMPUTE_REGISTRATION_TRANSFORMATION_H
+
+#include <CGAL/license/Point_set_processing_3.h>
+
+#include <CGAL/Aff_transformation_3.h>
```
afabri: No need for this inclusion as later we have `Kernel::Aff_transformation_3`.

maxGimeno:
The other possibility to avoid a CHECK failure is to add the  headers that use openGR to the whitelist so that they are not tested.

afabri:
@maxGimeno do we have a testsuite where you have installed OpenGR?

maxGimeno:
I don't think so.I can't find any trace of it in the DockerFiles or in any issue on github.

lrineau:
@sloriot You asked this morning about this pull-request. It is stalled:
 - waiting for small feature page, and that is probably what blocked it at the first time,
 - Travis tests did not pass,
 - there is a conflict with `master`.

sloriot:
Replaced by https://github.com/CGAL/cgal/pull/4153

PR #3486, opened 2018-11-27
Fixed face filtering in case one of the summands does not have an out…

efifogel:
…er boundary

## Summary of Changes

This fixes a bug in the reduced-convolution Minkowski sum.
In particular, the bug occurred when one of the summands did not have an outer-boundary.
I've added several test cases that covers all these cases that  used to fail.

## Release Management

* Affected package(s): Minkowski_sum_2
* Link to compiled documentation (obligatory for small feature) [link](https://mydoc/Manual/Pkg)


lrineau:
Would it be worth to backport the fix to release branches?

lrineau:
[**@lrineau**](https://github.com/lrineau) commented on [Nov 27, 2018, 4:12 PM GMT+1](https://github.com/CGAL/cgal/pull/3486#issuecomment-442093834 "2018-11-27T15:12:47Z - Replied by Github Reply Comments"):
> Would it be worth to backport the fix to release branches?

Efi, do you see this message? Could you please comment on my comment above?

maxGimeno:
errors on Apple:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-66.shtml#Minkowski_sum_2

efifogel:
It seems that I've missed this for some reason, sorry.

@Laurent Rineau <Laurent.Rineau@geometryfactory.com> Yes, I think it is
worth backporting, as it's a fix to a real bug.
Naturally, now that a month has passed, and it seems that I still need to
fix something, by the time it is fixed, you may decide that it's too late.

BW, I don't have an immediate access to an Apple comp, so if you think you
help me trace it down, let me know.
According to the error message it shouldn't be hard, and it might even be
not in the 2D Minkowski sum or even 2D Arrangement packages.

Efi
                         _/

maxGimeno:
I cannot reproduce on Darwin either. Let's test it again in integration tonight, if it still doesn't work I'll be able to test the integration branch tomorrow.

maxGimeno:
The bugs on magritte come from my branch on WKT after all. This one seems ok.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-101.shtml

PR #3487, opened 2018-11-27
Cartesian_kernel: Update kernel_ftC2.h

afabri:
## Summary of Changes

Simplify expression as we know that `b == 0`

## Release Management

### DIFF
--- a/Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h
+++ b/Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h
@@ -284,7 +284,7 @@ line_get_pointC2(const FT &a, const FT &b, const FT &c, int i,
 {
   if (CGAL_NTS is_zero(b))
     {
-      x = (-b-c)/a + i * b;
+      x = -c/a;
       y = 1 - i * a;
     }
   else


lrineau:
I have pushed a commit, to fix a warning.

Cartesian_kernel/include/CGAL/constructions/kernel_ftC2.h
```diff
@@ -284,7 +285,16 @@ line_get_pointC2(const FT &a, const FT &b, const FT &c, int i,
 {
   if (CGAL_NTS is_zero(b))
     {
-      x = (-b-c)/a + i * b;
+      // Laurent Rineau, 2018/12/07: I had this CGAL_assume to calm
```
MaelRL:
```suggestion
      // Laurent Rineau, 2018/12/07: I add this CGAL_assume to calm
```

maxGimeno:

lrineau:
@maxGimeno @afabri This PR is one example of a PR that was merged in `master` whereas it did not solve anything. ... And that is my fault, because I did not flag the PR as a work-in-progress.

Maybe I should revert that change: it is useless.

lrineau:
Useless because the warning is still there: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-105/Arrangement_on_surface_2_Examples/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz.

MaelRL:
Would removing `boost::is_integral<FT>::value` fix the warning? Could you please explain why it is important from a performance point of view?

lrineau:
> Would removing `boost::is_integral<FT>::value` fix the warning? Could you please explain why it is important from a performance point of view?

No. `boost::is_integral<FT>::value` is here for performance reason. We do not want to test `a != 0` if `a` is a constructed Lazy_exact_nt: that could trigger the recomputation of the exact version of the number.

Two options:
  - use instead `std::is_integral` from `<type_traits>` (in `master` or 5.0.x),
  - remove the `CGAL_assume` completely, because the warning is still there anyway: see [this result in CGAL-5.1-I-71](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-I-71/Arrangement_on_surface_2_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz).

MaelRL:
Right, I forgot about lazy types. 

Doesn't really justify the "No" answer though: I meant that if performance reasons were ignored, would the use of `__asume` actually remove the warning? Toying on godbolt, I still get the warning even if a `__assume(x!=0)`, see https://godbolt.org/z/agNLiN. 

Note that if I put `int a = 2`, the warning goes away, so it seems it's more like compiler optimisation have allowed it to deduce that `a` is actually `0` rather than it being afraid of whether it could be `0` or not. The warning page also says `The second operand in a divide operation evaluated to zero at compile time, giving undefined results.`, which goes in that direction.

Why would using `std::is_integral` change something?

lrineau:
It seems that I already knew the warning was still there: https://github.com/CGAL/cgal/pull/3487#issuecomment-457214870.

The MSVC compiler is not smart enough to realize that the "potential divide by 0" is not possible with the assumption we give to it. :disappointed:

lrineau:
> Note that if I put `int a = 2`, the warning goes away, so it seems it's more like compiler optimisations have allowed it to deduce that `a` is actually `0` rather than it being afraid of whether it could be `0` or not. The warning page also says `The second operand in a divide operation evaluated to zero at compile time, giving undefined results.`, which goes in that direction.

So that means there might actually be a real bug in the CGAL code, or at least in the test code?

> Why would using `std::is_integral` change something?

Forget that. I thought you were chasing a warning in Boost.

MaelRL:
> > Note that if I put `int a = 2`, the warning goes away, so it seems it's more like compiler optimisations have allowed it to deduce that `a` is actually `0` rather than it being afraid of whether it could be `0` or not. The warning page also says `The second operand in a divide operation evaluated to zero at compile time, giving undefined results.`, which goes in that direction.
> 
> So that means there might actually be a real bug in the CGAL code, or at least in the test code?

From what I understand, it means that the compiler does enter the function `line_get_pointC2` having deduced at compilation time that `line.a() == 0`, but since it ignores branching (and it wouldn't be able to know that we can't have `l.a()` and `l.b()` both being `0` anyway), it throws a warning. However in practice, we are not going into the branch `b == 0` where a division by `a` is performed since the program runs fine.

I am not sure how it could be resolved: putting `__assume(a != 0)` or even `if(a != 0)` (which would be annoying because it would cost some operations for no reason) doesn't seem to resolve the warning. Maybe the test's data can be changed or obfuscated so that the compiler cannot deduce that `a` is `0`...

lrineau:
The easiest might be to disable that warning in `Arrangement_on_surface_2/examples/Arrangement_on_surface_2/unbounded_non_intersecting.cpp`, using MSVC's pragmas.

PR #3499, opened 2018-11-30
Fix for stitching

sloriot:
Fix a case that were incorrectly classified


lrineau:
I have merged this PR in #3500, for practical reasons.

PR #3500, opened 2018-11-30
Fix for stitching (4.13)

sloriot:
same as #3499 but for 4.13+master


lrineau:
This PR now conflicts with `master`. I probably did something wrong.

sloriot:
The conflict is trivial to fix and is in the test file.

maxGimeno:

PR #3516, opened 2018-12-07
Polyhedron Demo: Fixes.

### LINKED ISSUE #3430: CGAL demo issues / feature requests
## Issue Details
- [x] Using edition mode on a non-triangle mesh crashes because of is_degen_triangle().
- [x] If the surface_mesh_item is selected during a collapse_edge, the wrong edge is collapsed.
- [x] Reloading a c3t3 with children crashes the demo.
- [x] Deleting / Moving an item scrolls the item list back to the top.

- [x] Deletion changes the order of the items if those items have been moved in the list prior to that deletion.

- <s>Can't change the color of the tentative hole-filling mesh.</s>

- [x] Detect sharp features + mesh editing results in a lot of wrong colors + some caching issues (recomputing the detect sharp features just loads the same wrong colors).

- [x] Move point + scrollwheel crashes the demo.

- [x] The console message when you call `remove_degenerate_faces` is wrong (looks like it interprets a Boolean return type as a number of faces removed).

- [x] Euler operations' "split edges" should be "split edge".

- [x] Color items when a selection item is present colors everything gray (possibly the same problem as with polylines item?).

- <s> Selecting both the mesh and the selection hides indices.  </s> 

- [x] Load file, select an edge, reload both ---> crash

- [x] NaN in an input file makes the demo crash. Fixed in #3532

- [x] Operation filters shouldn't be based only on the lowest items but also on menus/submenus names (e.g. if you search "coref" with a single mesh, you don't get anything because this is a submenu and not an operation).

- [x] "Tick plugins you don't want to load at start up" --> no "don't"

- [x] Shift-clicking coordinates depend on the camera position (when using transparency).

- [x] Collision detection: test items have a "reload from file" but it does nothing.

- [x] The length of normals in the mode "Pts + Normals" seems messed up: it seems to scale down the normals assuming the normals from the file have length 1, which might not be the case. Might need a setting "real length / scaled length".

- [x] Move an item in the list, and, without deselecting it, shift-click to another item to get a range. The result is not what is expected.

- [x] Using the operation search function and clearing the search results in missing operations whereas everything should be back.

- [x] Flip-edge selection (for example) is still visible after closing the selection widget.

- [x] "Look at" while the mesh is in a continuous rotation makes funky movements (just stop the rotation, I guess).

- [x] Degenerate triangles in an .OFF file (e.g. two points +  `3 0 1 0`) make the demo crash. see #3531 

- [x] `Set different colors` for a bunch of polylines use very dark colors hardly distinguishable.

- [x] Changing the transparency sucks all the RAM in before crashing.

- [x] Line thickness (and probably other things) do not actualize until camera moves.

- [x] The demo crashes when trying to load [this](https://gist.github.com/MaelRL/0d4da0f4a3a5b40867f18e84012b1746) (admittedly, it's an abomination and probably fails to triangulate its faces... but it shouldn't crash).

- [x] Some parameterizers, when failing, do not restore cursor properly (my bad, most likely...).

- [x] Fails to open [this file](https://gist.github.com/MaelRL/843a3be96f2d4256ca4ba348cb6d6150).

---
Some feature requests (developped in #3355 for convenience):

- [x] Add a Polygon generator that creates a polyline in Z=0 from 2D coords of a polygon. Don't forget to add the last point = the first point.

- [x] Be able to manipulate and save parameterization's UV mesh.
- <s> Add "is_valid_pm()" to statistics + a way to visualize non-valid parts?</s>
- [x] The selection of vertices/edges/faces could show the same highlighting (yellow-ish) than when using the edition mode.

- [x] Would be neat to have a "revert last operation" for all (or just some of) the Euler operations.
- [x] Change the thickness of multiple polylines items at a time (and other similar properties for similar items...).
- [x] It would be nice to be able to move multiple items up and down (at once) in the list of items.

### LINKED ISSUE #3548: Several memory leaks in the polyhedron demo
## Issue Details

While trying to debug a modification to polyhedron demo, I have tried the compilation flag `-fsanitize=address` (that can be used with both gcc and cland). It activates [AddressSanitizer] (also known as "ASAN").

[AddressSanitizer]: https://github.com/google/sanitizers/wiki/AddressSanitizer

And I found out several memory leaks in the demo:
```
Direct leak of 6080 byte(s) in 95 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb3caf2 in Viewer::drawVisualHints() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:978:30
    #2 0x7f841fb94523 in CGAL::QGLViewer::postDraw() /home/lrineau/Git/cgal/GraphicsView/include/CGAL/Qt/qglviewer_impl.h:578:3
    #3 0x7f841fb3f24c in Viewer::paintGL() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:1272:5
    #4 0x7f841ec255c7 in QWidget::event(QEvent*) (/lib64/libQt5Widgets.so.5+0x1a85c7)

Direct leak of 144 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb34f64 in Viewer::init() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:397:27
    #2 0x7f841fb8f3ab in CGAL::QGLViewer::initializeGL() /home/lrineau/Git/cgal/GraphicsView/include/CGAL/Qt/qglviewer_impl.h:250:3
    #3 0x7f841ec45e2e  (/lib64/libQt5Widgets.so.5+0x1c8e2e)

Indirect leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb3caf2 in Viewer::drawVisualHints() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:978:30
    #2 0x7f841fb94523 in CGAL::QGLViewer::postDraw() /home/lrineau/Git/cgal/GraphicsView/include/CGAL/Qt/qglviewer_impl.h:578:3
    #3 0x7f841fb3f24c in Viewer::paintGL() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:1272:5
    #4 0x7f841ec255c7 in QWidget::event(QEvent*) (/lib64/libQt5Widgets.so.5+0x1a85c7)

Indirect leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb31e46 in Viewer::Viewer(QWidget*, bool) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:256:21
    #2 0x7f841ffed188 in Ui_MainWindow::setupUi(QMainWindow*) /home/lrineau/Git/cgal/build/demo/Polyhedron/ui_MainWindow.h:174:22
    #3 0x7f841ffb22d1 in MainWindow::MainWindow(QStringList const&, bool, QWidget*) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/MainWindow.cpp:141:7
    #4 0x7f842003c51c in Polyhedron_demo::Polyhedron_demo(int&, char**, QString, QString, QStringList) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp:79:31
    #5 0x518565 in main /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Polyhedron_3.cpp:20:21
    #6 0x7f841cb1511a in __libc_start_main (/lib64/libc.so.6+0x2311a)

Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb0529d in Scene::initializeGL(CGAL::Three::Viewer_interface*) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Scene.cpp:439:9
    #2 0x7f841fb06063 in Scene::draw_aux(bool, CGAL::Three::Viewer_interface*) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Scene.cpp:672:9
    #3 0x7f841fb34c2a in Viewer_impl::draw_aux(bool, Viewer*) /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp
    #4 0x7f841fb3f21c in Viewer::paintGL() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:1271:5
    #5 0x7f841ec255c7 in QWidget::event(QEvent*) (/lib64/libQt5Widgets.so.5+0x1a85c7)

Indirect leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841fb358c4 in Viewer::init() /home/lrineau/Git/cgal/Polyhedron/demo/Polyhedron/Viewer.cpp:515:16
    #2 0x7f841fb8f3ab in CGAL::QGLViewer::initializeGL() /home/lrineau/Git/cgal/GraphicsView/include/CGAL/Qt/qglviewer_impl.h:250:3
    #3 0x7f841ec45e2e  (/lib64/libQt5Widgets.so.5+0x1c8e2e)
```

Among all the memory leaks reported by ASAN, there are false positives, when Qt5 (for example) was not compiled with ASAN activated. For example:
```
Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x515060 in operator new(unsigned long) (/home/lrineau/Git/cgal/build/demo/Polyhedron/Polyhedron_3+0x515060)
    #1 0x7f841e955f9f in QAbstractOpenGLFunctions::QAbstractOpenGLFunctions() (/lib64/libQt5Gui.so.5+0x3d5f9f)
    #2 0x7f841fb8f3ab in CGAL::QGLViewer::initializeGL() /home/lrineau/Git/cgal/GraphicsView/include/CGAL/Qt/qglviewer_impl.h:250:3
    #3 0x7f841ec45e2e  (/lib64/libQt5Widgets.so.5+0x1c8e2e)
```

All the reports that are interesting for us are when `new` or `malloc` are called by a function we wrote. That means it is when the frame `#1` is the report is from one of our files.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux Fedora 28
* Compiler: clang/gcc
* Release or debug mode: any of them, as soon as ASAN is activated
* Specific flags used (if any): `-fsanitize=address`
* CGAL version: `master` (950bd428ede9)
* Boost version: ?

maxGimeno:
## Summary of Changes
More fixes and feature requests.

## Release Management
* Issue(s) solved (if any): fix #3430, fix #3548, fi #3640


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
```diff
@@ -1466,7 +1466,7 @@ bool remove_degenerate_faces(      TriangleMesh& tmesh,
 }
 
 template<class TriangleMesh>
-std::size_t remove_degenerate_faces(TriangleMesh& tmesh)
+bool remove_degenerate_faces(TriangleMesh& tmesh)
```
lrineau: @sloriot @MaelRL Please review this change.

MaelRL: @sloriot changed the return type in e4ad5d96a7afd1f2cca3fbc7cb2661a1c1be7659, but forgot to change the return type in the named parameter-less call.

Polyhedron/demo/Polyhedron/Scene.cpp
```diff
@@ -683,6 +683,23 @@ Scene::draw_aux(bool with_names, CGAL::Three::Viewer_interface* viewer)
         transparent_items.push_back(id);
     }
     renderScene(children, viewer, picked_item_IDs, with_names, -1, false, NULL);
+    if(with_names)
+    {
+      //here we get the selected point, before erasing the depth buffer. We store it 
+      //in a dynamic property as a QRectF because there is no QVector3D. If not found, we erase the property.
+      bool found = false;
```
lrineau: http://doc.qt.io/qt-5/qvector3d.html

lrineau: You meant that `QVector3D` is not compatible with `QVariant`.

lrineau: But even that is wrong:
http://doc.qt.io/qt-5/qvector3d.html#operator-QVariant

maxGimeno: I meant there is no toVector3D() function to convert a QVariant to a QVector3D.

lrineau: There is `QVector3D` is supported by `QVariant`, with `QVariant::fromValue` and `variant.value<QVector3D>()`. That API is valid for all types recognized by `QMetaType`.

maxGimeno: Ok, but I need a QList<QVariant> after all.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Parameterization_plugin.cpp
```diff
@@ -824,6 +824,7 @@ void Polyhedron_demo_parameterization_plugin::parameterize(const Parameterizatio
       if (i == QDialog::Rejected)
       {
         std::cout << "Aborting parameterization" << std::endl;
+        QApplication::restoreOverrideCursor();
```
lrineau: You should create a [guard class](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Scope_Guard):
  - its constructor would call `setOverrideCursor(cursor)`, `cursor` being the argument to the constructor, and
  - its destructor would call `restoreOverrideCursor`.

I am not saying that you should then replace all calls to those functions with the guard class, but a few of them, to get used to it, and then, every new usage would use the guard.

Three/include/CGAL/Three/Three.h
```diff
@@ -88,6 +89,19 @@ class THREE_EXPORT Three{
...
   static int default_lines_width;
 
+public:
+  struct CursorScopeGuard
+  {
+    CursorScopeGuard(){}
+    void setCursor(QCursor cursor)
+    {
+      QApplication::setOverrideCursor(cursor);
+    }
```
lrineau: The real guard idiom would be to merge `setCursor` into the constructor.

maxGimeno: Which would trigger an unused variable warning. What they do in the example you cited is call release() in the end, but it's kind of the same, isn't it?

lrineau: Compiler are smart enough not to warn about that case. I think they warn about unused variables when they can prove that the constructor and the destructor have no side-effect.

[std::lock_guard](https://en.cppreference.com/w/cpp/thread/lock_guard) is an example, from the STL. Have a look at the example usage: only the constructor is called.

sloriot:
Successfully tested in CGAL-4.14-Ic-87. @lrineau please check that all your comments have been taken into account.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jan 4, 2019, 5:02 PM GMT+1](https://github.com/CGAL/cgal/pull/3516#issuecomment-451485816 "2019-01-04T16:02:48Z - Replied by Github Reply Comments"):
> @lrineau please check that all your comments have been taken into account.

They have. :+1:

lrineau:
I have removed the `Under testing` because of a non-trivial conflict without another PR.

maxGimeno:
I merged master again and it seems it solved all the conflicts. At least I don't see any in testsuite-Saturday anymore. I put back the flag

maxGimeno:

PR #3518, opened 2018-12-08
XHTML address of dtd

albert-github:
The address of the dtd for checking has changed from http -> https


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-67.shtml

PR #3519, opened 2018-12-08
XHTML problem regarding TOC definition

albert-github:
Form the doxygen documentation:
Normally the contents between `\htmlonly` and `\endhtmlonly` is inserted as-is. When you want to insert a HTML fragment that has block scope like a table or list which should appear outside `<p>..</p>`, this can lead to invalid HTML. You can use `\htmlonly[block]` to make doxygen end the current paragraph and restart it after `\endhtmlonly`.


albert-github:
An issue that occurs here is also "Why use this construct" and not the doxygen command `tableofcontents`.
(introduced in version 1.8.0 of doxygen).

sloriot:
screenshot from 2018-12-12 07-50-54
Here is a comparison of (left) `\tableofcontents` and (right) `\cgalAutotoc`. Our js is taking care of automatically generate the toc + assign a number to each section. I guess the latter is most probably the reason why we don't use the default one.

sloriot:
@albert-github  What is the issue with our current toc?

albert-github:
There is no problem with your usage of the TOC (except for the `[block]` part), but I was just wondering.
A bit strange that your picture depicts the TOC  below the text whilst the `\tableofcontents places it besides the text (looks nicer in my opinion)

maxGimeno:

PR #3520, opened 2018-12-08
XHTML incorrect tag sequence with figure

albert-github:
Due to the extra empty line an wrong sequence of tags appeared in the HTML output.


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-67.shtml

PR #3526, opened 2018-12-10
Subdivision_methods_3: Use deprecation warning

### LINKED ISSUE #3497: Missing deprecation warnings in subdivision
The old API of  subdivision functions should get deprecation warnings. It is a real issue because for example the demo is using the old API and it can silently break user code if the nb of iteration is not exactly the type expected (because the version with named parameters is called and the value is ignored).

afabri:
## Summary of Changes

Add deprecation warning and upgrade the plugin.

## Release Management

* Affected package(s): Subdivision_methods,  Polyhedron demo.
* Issue(s) solved (if any): fix #3497


maxGimeno:
Warnings https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-68.shtml#Polyhedron_Demo

maxGimeno:

PR #3530, opened 2018-12-11
Handle case of empty meshes (4.13+)

sloriot:
Same as #3517 but for 4.13+


maxGimeno:

PR #3533, opened 2018-12-11
XHTML Problems regarding tag sequences

albert-github:
XHTML is strict regarding tag sequences (contrary to HTML) in out case most of the times missing or superfluous p tags.

## Summary of Changes

- Documentation/doc/resources/1.8.14/BaseDoxyfile.in
  Normally the contents between `\htmlonly` and `\endhtmlonly` is inserted as-is. When you want to insert a HTML fragment that has block scope like a table or list which should appear outside `<p>..</p>`, this can lead to invalid HTML. You can use `\htmlonly[block]` to make doxygen end the current paragraph and restart it after `\endhtmlonly`.
- Documentation/doc/CMakeLists.txt
   Reason as with `BaseDoxyfile.in`, the usage of the `div` tag is more consistent and doesn't suffer from the `p` tag problem  and also the  part about bibtex, license etc. remains on the right (didn't happen with an empty ALIAS).
- Arrangement_on_surface_2/doc/Arrangement_on_surface_2/Arrangement_on_surface_2.txt
  doxygen doesn't see the `<` as the end of the word `points` (could be a template name) in `<b>A range of \a points</b>` and therefore creates invalid XHTML code. By using the `<em> ... </em>` construct this problem can be overcome.
- Algebraic_kernel_d/doc/Algebraic_kernel_d/Algebraic_kernel_d.txt
  Reason as with `BaseDoxyfile.in`, also the `<SMALL>` is only used with examples  in this part and not with other examples in the project and therefore removed


Algebraic_kernel_d/doc/Algebraic_kernel_d/Algebraic_kernel_d.txt
```diff
@@ -325,27 +325,27 @@ efficiency.
 
 The following example illustrates the construction of `AlgebraicKernel_d_1::Algebraic_real_1` 
 using `AlgebraicKernel_d_1::Construct_algebraic_real_1`:
-<SMALL>\cgalExample{Algebraic_kernel_d/Construct_algebraic_real_1.cpp}</SMALL>
+\cgalExample{Algebraic_kernel_d/Construct_algebraic_real_1.cpp}
```
sloriot: Could you please explain again why this is needed, and only for this file?

albert-github: I only found this construct only in this file, all other files straight away used `\cgalExample`.

sloriot: I actually miss-read the diff and though you were adding the SMALL. I indeed don't see why we have it. Thanks!

maxGimeno:

lrineau:
Thanks Albert for the pull-request. It has now been tested, and then merged in `master`, for CGAL-4.14. Note that, unless the other Doxygen fixes, it was not merged in the releases branches, because your pull-request was against `master`.

albert-github:
Thanks. I'm concentrating on the master version. It is up to the CGAL team whether or not they want to back port it to older versions.

PR #3534, opened 2018-12-12
avoid extra </p> after figures

sloriot:
The extra `</p>` can only be seen if the post processing is not done.


maxGimeno:

PR #3538, opened 2018-12-12
BGL: Fix Euler::add_face()

### LINKED ISSUE #3531: Surface_mesh: Crash on collect_garbage()
## Issue Details

When calling collect_garbage() on a degenerate triangle in a surface_mesh, it crashes.
## Source Code
```

#include <sstream>
#define CGAL_USE_BASIC_VIEWER 1
typedef K::Point_3 Point;
typedef CGAL::Surface_mesh<Point> Mesh;

int main()
{
  Mesh m;
  std::stringstream poly("OFF\n"
                         "2 1 0\n"
                         "0 0 0\n"
                         "0 0 1\n"
                         "3  0 1 0\n");
  poly >> m;
  m.collect_garbage();
  
  return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):Linux 64 bits
* Compiler:clang
* CGAL version:4.13
* Boost version:1.65

maxGimeno:
## Summary of Changes
in Euler::add_face(), check if the vertices of the new face are unique. If not, return null_face. 
## Release Management

* Affected package(s):BGL
* Issue(s) solved (if any): fix #3531


BGL/include/CGAL/boost/graph/Euler_operations.h
```diff
@@ -585,6 +585,12 @@ add_face(const VertexRange& vr, Graph& g)
 
   std::vector<vertex_descriptor> vertices(vr.begin(), vr.end()); // quick and dirty copy
   unsigned int n = (unsigned int)vertices.size();
+  //check that every vertex is unique
+  std::sort(vertices.begin(), vertices.end());
+  if(std::unique(vertices.begin(), vertices.end()) != vertices.end())
+    return boost::graph_traits<Graph>::null_face();
+  vertices.clear();
+  vertices=std::vector<vertex_descriptor>(vr.begin(), vr.end());
```
lrineau: Your code does not avoid the allocation of another vector. What I suggested was:
```c++
vertices.clear();
vertices.insert(vr.begin(), vr.end());
```
But actually, we can avoid `clear()`, with just:
```c++
// re-fill the vector with the initial (unsorted) values
std::copy(vr.begin(), vr.end(), vertices.begin());
```

afabri: And this should be benchmarked.  Note also  that most often it will be a triangle mesh.

BGL/include/CGAL/boost/graph/Euler_operations.h
```diff
@@ -585,6 +585,12 @@ add_face(const VertexRange& vr, Graph& g)
 
   std::vector<vertex_descriptor> vertices(vr.begin(), vr.end()); // quick and dirty copy
   unsigned int n = (unsigned int)vertices.size();
+  //check that every vertex is unique
+  std::sort(vertices.begin(), vertices.end());
+  if(std::unique(vertices.begin(), vertices.end()) != vertices.end())
```
lrineau: Actually, [`std::adjacent_find`] is better than [`std::unique`], because it stops as soon as it has found two adjacent identical values:
```suggestion
  if(std::adjacent_find(vertices.begin(), vertices.end()) != vertices.end())
```

[`std::adjacent_find`]: https://en.cppreference.com/w/cpp/algorithm/adjacent_find
[`std::unique`]: https://en.cppreference.com/w/cpp/algorithm/unique

maxGimeno:
I benched on iphigenia, there is no real impact. Both take between 3.4 and 3.6 seconds to load.

maxGimeno:
I think a function checking for the validity of a face would be even more impacting than that, but it would probably be cleaner and safer. I don't really have an opinion.

sloriot:
What I mean is that usually in CGAL we are trying to have function that expect the input is clean to not impact users that have clean data. Anyway in this case, you are not going to use this function if you look for performance so I don't really mind either (but changing the n>2 into a if).

maxGimeno:
I agree for the `n>2`, I'll do that today.

maxGimeno:
Travis doesn't like rebasing, it will complain about packages that didn't exist in 4.12 but we don't care.

maxGimeno:

lrineau:
@sloriot @afabri This pull-request was tested correctly in CGAL-4.14-Ic-70. But there was a discussion. Do you agree with the last version of it?

afabri:
> 
> 
> @sloriot @afabri This pull-request was tested correctly in CGAL-4.14-Ic-70. But there was a discussion. Do you agree with the last version of it?

I approve the pull request.

PR #3544, opened 2018-12-13
Parameterization: No need to disable warnings in CMakeLists.txt

afabri:
## Summary of Changes

The example gives the wrong impression that some warnings must be suppressed.

@MaelRL does the Orbifold really need SuiteSparse??

## Release Management

* Affected package(s): Suerface_mesh_parameterization

### DIFF
--- a/Surface_mesh_parameterization/test/Surface_mesh_parameterization/CMakeLists.txt
+++ b/Surface_mesh_parameterization/test/Surface_mesh_parameterization/CMakeLists.txt
@@ -14,24 +14,7 @@ find_package(CGAL QUIET)
 
 if ( CGAL_FOUND )
 
-  # VisualC++ optimization for applications dealing with large data
-  if (MSVC)
 
-    # Allow Windows applications to use up to 3GB of RAM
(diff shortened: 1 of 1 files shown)


MaelRL:
Orbifold parameterization by default would like to use the UmfPackLU solver (which is part of suitesparse). It can go ahead with only SparseLU if SuiteSparse is not available, but the results are much worse.

MaelRL:
Also, I'm not really fit to review this diff: those lines are apparently from the dawn of SMP and for warnings on MSCV, I don't know if those lines are still relevant (likely not).

afabri:
> 
> 
> Orbifold parameterization by default would like to use the UmfPackLU solver (which is part of suitesparse). It can go ahead with only SparseLU if SuiteSparse is not available, but the results are much worse.

Can you check with @maxGimeno if we have at least one testsuite with SuireSparse installed.

afabri:
> 
> 
> Also, I'm not really fit to review this diff: those lines are apparently from the dawn of SMP and for warnings on MSCV, I don't know if those lines are still relevant (likely not).

I checked it locally, and it was not necessary for my VC, but the testsuite will tell us.

maxGimeno:
Every docker image based on Ubuntu has Suitesparse installed.

PR #3545, opened 2018-12-13
Move the definition of CGAL_USE_FILE in case of early return()

lrineau:
## Summary of Changes

In the `CGALConfig.cmake`, move the definition of `CGAL_USE_FILE` in case of early `return()`. That may happen if `find_package(CGAL)` is called in a sub-directory *before* it is called in the current `CMakeLists.txt`.

I know the bug is already in CGAL-4.13, and I know other bugs about other variables set by `CGALConfig.cmake`, after the early `return()`. But I prefer to fix that only in `master`.

## Release Management


maxGimeno:

PR #3549, opened 2018-12-17
Point Set Processing: Fix PLY unknown element reading

sgiraudot:
## Summary of Changes

Currently, our PLY reader only handles elements of type `vertex` and `face`. Though this is in itself acceptable, a problem occurs if another type of element is used in the PLY input: in that case, our reader might read garbage (as the properties of the elements are not processed at all instead of being processed but ignored, which would be the right behaviour).

This PR fixes it. It also makes the handling of PLY element much better: the internal API (not documented) of our PLY reader now allows to have access to _all_ elements and their respective properties, which can be handy if we want to read other types of PLY elements.

## Release Management

* Affected package(s): Point Set Processing, Point Set 3, Polyhedron IO, Polyhedron demo

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3/IO.h
+++ b/Point_set_3/include/CGAL/Point_set_3/IO.h
@@ -135,7 +135,7 @@ private:
   struct Abstract_ply_property_to_point_set_property
   {
     virtual ~Abstract_ply_property_to_point_set_property() { }
-    virtual void assign (PLY_reader& reader, typename Point_set::Index index) = 0;
+    virtual void assign (PLY_element& element, typename Point_set::Index index) = 0;
   };
 
   template <typename Type>
@@ -154,10 +154,10 @@ private:
       m_pmap = ps.push_property_map (m_map);
     }
     
-    virtual void assign (PLY_reader& reader, typename Point_set::Index index)
+    virtual void assign (PLY_element& element, typename Point_set::Index index)
     {
       Type t;
-      reader.assign (t, m_name.c_str());
+      element.assign (t, m_name.c_str());
       put(m_pmap, index, t);
     }
   };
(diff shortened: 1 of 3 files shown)


Point_set_3/include/CGAL/Point_set_3/IO.h
```diff
@@ -401,24 +399,33 @@ read_ply_point_set(
...
+      for (std::size_t k = 0; k < element.number_of_properties(); ++ k)
+      {
+        internal::PLY::PLY_read_number* property = element.property(k);
+        property->get (stream);
 
-      filler.process_line (reader);
-      
-      ++ points_read;
+        if (stream.eof())
+          return false;
```
lrineau: In that case you should set the `failbit` of the stream.

sgiraudot: Actually, I just realized that this is completely wrong (if the input only has points, the _end of files_ is reached at the last iteration, and it's a normal behaviour, we don't want to return false in that case). I'll fix it (and I think I'll use `stream.bad()` as I did in the mesh reader).

lrineau: Be careful with `good()` and `bad()`. Their meaning is not always what one expects. In particular `bad()` is not the opposition of `good()`.

What is more, after `stream >> something`, the success of the operation if neither `good()` nor `!bad()`, but it is `!fail()`.

sgiraudot: Alright, I knew about `bad()` not being the opposite of `good()` (I used to do `return !stream.bad()` at the very end of the function), not about the `fail()` part.

For the above case, I think the best solution would be to do, _before_ attempting to read a number, this:
```c++
        if (stream.eof())
        {
          stream.setstate(std::ios::failbit);
          return false;
        }
```
Do you agree?

lrineau: - Returning `!bad()` is incorrect, because the stream could in a state with only the `failbit` set.

- Notice that there are cases were the `eofbit` is not set when reaching the end of a stream. For example, when `std::getline` can read the final `\n` character, it does not set the `eofbit`. It is only the next reading that will set it (and probably the `failbit` at the same time). That is why it is rather tricky to test `eof()` before reading.

A useful documentation page: https://en.cppreference.com/w/cpp/io/ios_base/iostate

If you have extra questions, I suggest we discuss together before a screen.

sloriot:
Successfully tested in CGAL-4.14-Ic-87

PR #3561, opened 2018-12-20
More modular implementation of nef to face graph

sloriot:
This allows an undocumented use of nef -> polygon soup


sloriot:
Successfully tested in CGAL-4.14-Ic-76

PR #3570, opened 2018-12-21
Fix some -Wmaybe-uninitialized warnings

sloriot:


NewKernel_d/include/CGAL/NewKernel_d/Vector/array.h
```diff
@@ -65,13 +65,14 @@ template<class NT_,class Dim_,class Max_dim_=Dim_> struct Array_vector {
...
+				Vector operator()(unsigned d,Iter const& f,Iter const& e) const {
 					CGAL_assertion(d==(unsigned) std::distance(f,e));
 					CGAL_assertion(d<=d_);
 					//TODO: optimize for forward iterators
 					Vector a;
-					CGAL_assume(f!=e);
-					std::copy(f,e,a.begin());
+					Iter it = f;
+					for(unsigned i=0; i<d ; ++i) a[i]=*it++;
+					for(unsigned i=d; i<d_; ++i) a[i] = 0;
```
mglisse: Why? If size<maxsize, the last coordinates are meant to be unused. Not only does the patch do (probably negligible) useless work, it also hides potential bugs from runtime checkers or from more clever static checkers.

sloriot: I agree for the extra work but I expected that `d==d_` most of the time so I expected the patch to be as efficient as the `std::copy`. I didn't find a better way to fix the warning, maybe we'll have to simply switch it off for this case if we can't find a better workaround.

lrineau: @sloriot Please add a comment, with the warning you are trying to fix.

afabri: Why not a `CGAL::use()` and sticking to `std::copy()` ?

sloriot:
I have removed the commit from the history

PR #3572, opened 2018-12-30
Fix Surface_mesh::join()

afabri:
## Summary of Changes

Fix how the freelists get joined.

## Release Management


afabri:
@sloriot as you reported the bug, can you please proof-read my fix.

sloriot:
Succesfully tested in CGAL-4.14-Ic-87

PR #3580, opened 2019-01-07
Fix warnings (Add/remove copy constructors, and other warnings)

afabri:
## Summary of Changes

Fix some (tip of the ice berg?) warnings in [TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-88/Property_map/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz).

## Release Management

* Affected package(s): Stream_support, Convex_hull_d


sloriot:
I think the critical part of the warning message is:
```
/mnt/testsuite/include/CGAL/IO/Color.h:64:10: note: because 'CGAL::Color' has user-provided 'CGAL::Color& CGAL::Color::operator=(const CGAL::Color&)'
   64 |   Color& operator=(const Color &c)
```
Removing the useless `operator=` should be the correct solution.

mglisse:
Doesn't Sebastien's comment also apply to most other cases? i.e. remove the copy constructor of Point_d instead of adding a defaulted assignment operator? Or does that break something?

afabri:
I simplified some.  What about this one:
```
    Polynomial(const Self& p) : Base(static_cast<const Base&>(p)) {}
```

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Jan 10, 2019, 2:10 PM GMT+1](https://github.com/CGAL/cgal/pull/3580#issuecomment-453090490 "2019-01-10T13:10:08Z - Replied by Github Reply Comments"):
> I simplified some.  What about this one:
> ```
>     Polynomial(const Self& p) : Base(static_cast<const Base&>(p)) {}
> ```

If `Base` is the first base class of `Self`, then the explicit `static_cast` is just redundant with the implicit cast of a reference to class object to a reference of its base. That copy constructor is the same as the one generated by the compiler.

maxGimeno:
/home/cgal-testsuite/cgal_doc_build/CGAL-4.14-Ic-96/include/CGAL/Polygon_2.h:169: warning: Member operator=(const Polygon_2 &)=default (function) of class CGAL::Polygon_2 is not documented.

afabri:
@mglisse and @sloriot when we add a function with `=default`, shall we document it with doxygen?   It was not documented before.

sloriot:
I don't think so. It is strange that you have the warning provided that the config of `Polygon` says:
```
EXTRACT_ALL                =  false
HIDE_UNDOC_MEMBERS         =  true
HIDE_UNDOC_CLASSES         =  true
```

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jan 15, 2019, 3:00 PM GMT+1](https://github.com/CGAL/cgal/pull/3580#issuecomment-454401627 "2019-01-15T14:00:48Z - Replied by Github Reply Comments"):
> I don't think so. It is strange that you have the warning provided that the config of `Polygon` says:
> ```
> EXTRACT_ALL                =  false
> HIDE_UNDOC_MEMBERS         =  true
> HIDE_UNDOC_CLASSES         =  true
> ```

Should we report a bug to Doxygen?

sloriot:
@albert-github can you confirm that we should not get undocumented members warnings with such options?

albert-github:
@sloriot, @lrineau,

I did some small tests, the message present when the code:
`Polygon_2& operator=(const Polygon_2&)=default;`
is at its current place (with the constructors) but disappears when I place the code in the 'Global Operators' section.
I would not have expected a message in both cases as `EXTRACT_ALL` is set, but I'm not sure about this (I myself never used this type of constructs).

Best is (as suggested by @lrineau) to file a doxygen issue and probably Dimitri (@doxygen) can shine a light on it.

afabri:
If I get it right the problem is that if there is only one of  copy constructor / assignment, and it is there on purpose because the default generated one does not do the job, then the other must also be written, either a real implementation or with `=default;`.   The question for me is what we have to document. Systematically both?  Systematically none? Or selectively as we do currently?  

A workaround is to use an `#ifdef DOXYGEN_RUNNING`.

albert-github:
Regarding the not documented operator I would anyway document it in the code (not necessarily in the documentation to the customer) as, as far as I understand, there were some special reasons to create the operator (and otherwise this reasoning might get lost for future maintenance).

lrineau:
I have fixed the strange warning in https://github.com/CGAL/cgal/pull/3580/commits/d2eb07bbcdb7d3bac435dd3bd3aa08f04cc3ff93.

But that file does something strange: it specializes [`std::swap`] for the (non-template) class `CGAL::Exponent_vector`. That is allowed by the C++ norm, but that does not look robust, to me. The warning here was because the [`std::swap`] of C++11 has a `noexpect` specification, and the specialization had not that `noexpect` specification.

[`std::swap`]: https://en.cppreference.com/w/cpp/algorithm/swap

Specializing `std::swap` does not seem robust for me. In C++17, the noexcept specification changes, and I wonder if we can get another warning about the fact that my patch in d2eb07bbcdb7d3bac435dd3bd3aa08f04cc3ff93 copies the specification of C++11.

I think the code should not specialize `std::swap` but instead define an overload to `CGAL::swap`, that will be found by ADL. That would satisfy the requirements for [`Swappable`](https://en.cppreference.com/w/cpp/named_req/Swappable).

afabri:
Reading the comment in [ `SM_list.h`](https://github.com/CGAL/cgal/blob/master/Nef_S2/include/CGAL/Nef_S2/SM_list.h#L64)  I am wondering if want to define the `=default` copy constructor, or at best `private:`

afabri:
@MaelRL can you have a look at [`Quadratic_program_solution`](https://github.com/CGAL/cgal/blob/master/QP_solver/include/CGAL/QP_solution.h#L317).

maxGimeno:

lrineau:
I think it is time to stop adding any new commit to this PR, and open another one. With the commit https://github.com/CGAL/cgal/pull/3580/commits/a7d65b7913d2b0a3cd5f311728b4a79e6a70d737, you open a new work in progress that can last for a long time.

The response from Qt5 team to GCC9 new warnings was this one: https://github.com/qt/qtbase/commit/76f11b0eda50f05fb0912738be41b4333a82f748. They have added `-Wno-deprecated-copy` to their own build system, instead of trying to calm down the warnings.

I think the right move for us would be to make the compiler consider Qt5 headers as system headers, and ignore the warning emitted from Qt5 headers.

afabri:
I added it only in the examples using the basic viewer and when we are happy this should fix ALL examples.
Shouldn't we add the `-Wno-deprecated-copy` in all `CMakeLists.txt` of the demos.

afabri:
@maxGimeno Looking at [this testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-112/AABB_tree_Demo/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz)  I think the `#ifdefs` are not correct as I can't see the `-Wno-deprecated-copy`.

afabri:
@maxGimeno I don't see `-Wno-deprecated-copy` in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-113/AABB_tree_Demo/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz)

afabri:
@lrineau can you have a look as the last commit seems not to solve the problem. The warning is still [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-113/AABB_tree_Demo/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz) and I do not see the option set.

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Feb 6, 2019, 8:02 AM GMT+1](https://github.com/CGAL/cgal/pull/3580#issuecomment-460920986 "2019-02-06T07:02:38Z - Replied by Github Reply Comments"):
> @lrineau can you have a look as the last commit seems not to solve the problem. The warning is still [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-113/AABB_tree_Demo/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz) and I do not see the option set.

The commit https://github.com/CGAL/cgal/pull/3580/commits/ff4f13e82cae2704c86140b2866d99bf1acc7520 was not yet in the testsuite for CGAL-4.14-Ic-113, but it was in CGAL-4.14-Ic-115, and [the warning was still there][1]. That is because, when the CMake variables `CGAL_DEV_MODE` or `RUNNING_CGAL_AUTO_TEST` are set, then the `CMAKE_CXX_FLAGS` used during the configuration of the CGAL library is force-copied when `UseCGAL.cmake` is included. I will change commit https://github.com/CGAL/cgal/pull/3580/commits/ff4f13e82cae2704c86140b2866d99bf1acc7520 and push the fix directly in the setup of CGAL_Qt5, so that all CGAL qt5 demos are impacted...

[1]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-115/AABB_tree_Demo/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz

lrineau:
Last commit https://github.com/CGAL/cgal/pull/3580/commits/2455affba0f5bbca3f87157978ddc9586c1c4695 moved the logic to the setup of CGAL_Qt5. That will quiet the warning of CGAL_Qt5 and all CGAL Qt5 demos. I have verified on my machine and in the Docker container that g++ >= 9.0 is correctly detected.

afabri:
Cool !!! Saves us time to patch lot's of CMakeLists.txt

lrineau:
I suggest we merge this PR, now. If extra work is needed on warnings, then it will be in a new PR.

afabri:
@lrineau before merging this PR let's test my fix concerning algebraic_kernel_d and please help me to fix STL_Extensions.

afabri:
@lrineau can you or @maxGimeno  please check with gcc9 if the missing initialization of `full` in this [constructor ](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L221)  is the reason for this [warning ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-119/Circular_kernel_3/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz)

maxGimeno:
There are several constructors that are missing initialization, not only this one.

afabri:
Not really. Those with `*this = Circular_arc_3(..)`  initialize on the rhs and then copy in the assignment.

maxGimeno:
At least there is the default constructor and the one you pointed.

afabri:
@MoniqueTeillaud can  you please help us to fix this.  The class  `Circular_arc_3` has a Boolean data member `_sign_cross_product` which is not initialized in [this ](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L143) and [this ](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L163) constructor.   We think that this is a mistake as for example the function [`approximate_angle()`](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L272) uses this Boolean.   
Can you please fix the initialization (either in the code, or a code snipped in a mail or directly as a reply in this conversation.

afabri:
@MoniqueTeillaud and can you confirm that my latest commit is correct.

MoniqueTeillaud:
I must dig into this old code. 
Added to my todo-list...

MoniqueTeillaud:
If I trust the comment here
      // it saves execution time for the has_on functor
      Sign _sign_cross_product;
my first impression is that there is actually a mistake in each constructor when _sign_cross_product IS initialized... this computation has a cost, it should probably be initialized using a lazy approach: compute the cross product in the functions only when it is used (e.g., approximate_angle()) and cache it in _sign_cross_product for further uses. 
That said, I might be wrong (I only had a quick look during a boring meeting...)

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Feb 11, 2019, 3:07 PM GMT+1](https://github.com/CGAL/cgal/pull/3580#issuecomment-462339521 "2019-02-11T14:07:56Z - Replied by Github Reply Comments"):
> @MoniqueTeillaud can  you please help us to fix this.  The class  `Circular_arc_3` has a Boolean data member `_sign_cross_product` which is not initialized in [this ](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L143) and [this ](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L163) constructor.   We think that this is a mistake as for example the function [`approximate_angle()`](https://github.com/CGAL/cgal/blob/master/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L272) uses this Boolean.   
> Can you please fix the initialization (either in the code, or a code snipped in a mail or directly as a reply in this conversation.

I think that in both cases, the `_sign_cross_product` could be set to 0. That is consistent with the comment documenting that data member:
https://github.com/CGAL/cgal/blob/73a22a8655433e3c3c8108141f376a235ff24ad9/Circular_kernel_3/include/CGAL/Circular_kernel_3/Circular_arc_3.h#L89-L95

MoniqueTeillaud:
> I think that in both cases, the `_sign_cross_product` could be set to 0. That is consistent with the comment documenting that data member:

I don't think so: _sign_cross_product == 0 has a geometric meaning, it means that s,t,c are collinear, it is not a safe default value.

lrineau:
[**@MoniqueTeillaud**](https://github.com/MoniqueTeillaud) commented on [Feb 13, 2019, 8:57 AM GMT+1](https://github.com/CGAL/cgal/pull/3580#issuecomment-463096620 "2019-02-13T07:57:45Z - Replied by Github Reply Comments"):
> > I think that in both cases, the `_sign_cross_product` could be set to 0. That is consistent with the comment documenting that data member:
> 
> I don't think so: _sign_cross_product == 0 has a geometric meaning, it means that s,t,c are collinear, it is not a safe default value.

The two cases we are talking about are two cases where the circular arc is full, and thus `s == t`, so `_sign_cross_product` can be set to `0`, even if it is probably not used by the code when `is_full()` returns `true`.

MoniqueTeillaud:
Maybe... I have not investigated each use case.

lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-122/

PR #3588, opened 2019-01-11
Triangulation_2: Fix stack order in non-recursive conflict computations

### LINKED ISSUE #3380: Infinite loop in nn2
Initially reported on cgal-discuss (2014/07/04)

It seems that there is an infinite loop:
[inter_br.zip](https://github.com/CGAL/cgal/files/2454861/inter_br.zip)

MaelRL:
## Summary of Changes

Boundary edges of the conflict zones must be output in a CCW order, which is achieved by walking ccw-ly first. In the recursive function, we correctly call `propagate(..., ccw(j), ...)` first. For the non-recursive version, which uses a stack, then we must add the call with `..., ccw(j), ...` last, since the stack is a LIFO structure.

`CDT_2` was already correct.

## Release Management

* Issue(s) solved (if any): fix #3380


maxGimeno:
There is a test failing in the testsuite :  https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-105.shtml#Periodic_2_triangulation_2_Examples

I'm not sure it is related but I couldn't find another PR touching Triangulation_2, so...

MaelRL:
It's unrelated: this PR is just swapping two lines in the class Triangulation_2 (which P2T2 doesn't use anyway), and adding tests to T2.

I'll fix the error in P2T2 in another PR.

maxGimeno:
Thanks !

maxGimeno:
Then, 
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-105.shtml

PR #3595, opened 2019-01-14
Triangulation_2: revert small feature DT2_strict_and_weak_conflict_zone

MaelRL:
## Summary of Changes

https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/DT2_strict_and_weak_conflict_zone is a small feature that was integrated a while ago without approval has some legitimate concerns in it (see discussion in the small feature page). This PR reverts its changes, which are entirely code (no doc).

This small feature was integrated to fix some bug in the Interpolation package, possibly due to the presence of potentially null coordinates in the output when called on degenerate configurations. This PR filters irrelevant (`==0`) or illogical (`<0`) coordinates after the computation rather than modifying the triangulation's conflict zone.

No test failed after I reverted the small feature and before I added the filters, so it is difficult to evaluate whether whatever bug forced the small feature in the first place is still fixed by just filtering the output.

## Release Management

* Affected package(s): `Triangulation_2`, `Interpolation`
* Putting it in master and 4.15 because - in theory - it's a small feature change, but it could be back-ported onto any other release.

### DIFF
--- a/Interpolation/include/CGAL/natural_neighbor_coordinates_2.h
+++ b/Interpolation/include/CGAL/natural_neighbor_coordinates_2.h
@@ -99,20 +99,27 @@ natural_neighbors_2(const Dt& dt,
     Point_2 p1(v1->point()), p2(v2->point());
 
     Coord_type coef1(0);
-    Coord_type coef2(0);
     Equal_x_2 equal_x_2;
     if(!equal_x_2(p1,p2))
-    {
       coef1 = (p.x() - p2.x()) / (p1.x() - p2.x());
-      coef2 = 1 - coef1;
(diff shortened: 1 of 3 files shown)


afabri:
I approve this PR.

MaelRL:
@sloriot the requested review is limited to the file in the Interpolation package (checking that the early exits make sense) and does not concern the small feature revert.

afabri:
@maxGimeno can you please add to the testsuite.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-10.shtml

PR #3596, opened 2019-01-14
PMP:: speed up PMP::do_intersect()

### LINKED ISSUE #3195: Speeding up PMP::do_intersect
If we have two large models that are far apart, we still pay for the initialization of the box vectors. A first test using the global bbox of the objects divided the run time by 40 on a model of 260000 faces.

maxGimeno:
## Summary of Changes

Test if the global bboxes of the two elements in PMP::do_intersect() intersect before computing the boxes of the faces/segments.

## Release Management

* Affected package(s):PMP
* Issue(s) solved (if any): fix #3195

benchmark:

Number of vertices in meshes: 639832
Number of points in polylines: 3838896

_Before PR_:

With Intersections:
TM-TM : 0.260919
Poly-Poly: 1.66438
TM-Poly: 0.226532

Without intersection : 
TM-TM : 2.16355
Poly-Poly: 3.12714
TM-Poly: 2.71119

_After PR_::

With Intersections:
TM-TM : 0.289844
Poly-Poly: 1.71594
TM-Poly: 0.242312

Without intersection : 
TM-TM : 0.00642
Poly-Poly: 0.113171
TM-Poly: 0.034315


lrineau:
I wonder why that step of optimization is not done in the `Box_intersection_d` package. This PR probably slows down a bit the query for intersecting objects.

maxGimeno:
@lrineau it is shown in the attached benchmark, yes.

lrineau:
Then this filtering should probably go into the `Box_intersection_d` package, to that it avoids the recursion if the boxes do not intersect at the highest level. Actually, that is really strange if it does not do exactly that. That looks like a bug.

sloriot:
+ the allocation of the container to store the boxes. I guess if we run the benchmark on two small non-intersecting models we would see no improvement.

maxGimeno:
It is true that if the test was performed in Box_intersection_d, even before this PR, the case without intersection should be quicker than the case with intersection, and it is not the case. Really not. 
Box_intersection_d computes all the intersections between the boxes of the range, and the call_back stops it after the first one is found.

That makes the case with no intersection the worst case, as we have to check them all. But it is not a tree, it is nothing more than a collection  of boxes, if I am not mistaking. So there is no hiearchy to check that we won't find anything if this and that boxes do not intersect. 
At least that's what I understood.

PR #3597, opened 2019-01-14
Update cmake script to create a release with only a subset of packages

sloriot:


Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```diff
@@ -7,6 +7,70 @@
 #   Must be followed by -beta<beta_number> if the release is a beta.
 # CGAL_VERSION_NR=release string used to update version.h. Must be something like 1041200033 , or 10412009<beta number>0
 # TESTSUITE=indicate if the release is meant to be used by the testsuite, default if OFF
+# GPL_PACKAGE_LIST=path to a file containing the list of GPL packages to include in the release. If not provided all of them are.
+
+if(POLICY CMP0009)
+  # http://www.cmake.org/cmake/help/v3.2/policy/CMP0009.html
+  cmake_policy(SET CMP0009 NEW)
+endif()
```
lrineau: I suggest to add, instead:
```cmake
cmake_minimum_required(VERSION 3.1)
```
(same as for CGAL).

That way, all policies from `CMP0000` to `CMP0057` are set to `NEW`.

lrineau:
The CMake script changes by this PR is the one that is now used to create our internal releases. I trust Sébastian to have tested correctly that the new version of the script produces exactly the same tarballs as before (when the new option is not used).

@maxGimeno I suggest to merge this PR in `integration`. If something goes wrong, we should now it pretty fast.

PR #3602, opened 2019-01-17
Fix for VC2017 with /arch=AVX

afabri:
## Summary of Changes

It seems that VC2017 has a problem with default generated assignment operator when it contains two 64bit data members. Providing the implementation fixes this.

## Release Management

* Affected package(s): QP_solver, Spatial_searching


afabri:
The testsuite [4.14-Ic-101](https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-101.shtml) column with arch:AVX is green.

Issue #3583, opened 2019-01-09
MP_float runtime error


maxGimeno:
## Issue Details
Execution bug in the testsuite : https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-91.shtml#Number_types

lrineau:
Your report is really bad. You should at least give the direct link:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-91/Number_types/TestReport_lrineau_ArchLinux-clang-CXX17-Release.gz

And copy-paste the runtime log:
```
------------------------------------------------------------------
- ProgramOutput.MP_Float.ArchLinux-clang-CXX17-Release
------------------------------------------------------------------
CGAL::Random()::get_seed() = 1547016968
 base = 65536
 trunc_min = -32768.500007629510947
 trunc_max = 32767.499992370489053
 limbs_per_double = 5
bug = 0.49999237058324297323
sizeof(long) = 8
sizeof(int) = 4
sizeof(short) = 2
sizeof(char) = 1
Checking MP_Float(float) constructor.
Checking MP_Float(double) constructor.
Checking MP_Float(long double) constructor.
CGAL error: assertion violation!
Expression : CGAL::abs(frac) < T(1.0)
File       : /mnt/testsuite/include/CGAL/MP_Float_impl.h
Line       : 52
Explanation: 
Refer to the bug-reporting instructions at https://www.cgal.org/bug_report.html
terminate called after throwing an instance of 'CGAL::Assertion_exception'
  what():  CGAL ERROR: assertion violation!
Expr: CGAL::abs(frac) < T(1.0)
File: /mnt/testsuite/include/CGAL/MP_Float_impl.h
Line: 52
./cgal_test_with_cmake: line 70: 23362 Aborted                 (core dumped) ./MP_Float
```

Issue #3303, opened 2018-08-31
MP_Float assertion violation


valerashm:
Issue Details

I have the next error:

CGAL error: assertion violation!
Expression : CGAL::abs(frac) < T(1.0)
File       : e:\projects2\libs\src\cgal-4.11.2_from_installer\include\cgal\mp_float_impl.h
Line       : 51
Explanation:
Refer to the bug-reporting instructions at http://www.cgal.org/bug_report.html

I was reading code of MP_Float class to see how it is implemented and then tried to create an object of MP_Float and I have got an assertion violation.

Source Code to repeat the issue

Source code to reproduce the error is

#include "CGAL/MP_Float.h"

int main() {
  CGAL::MP_Float f( CGAL::INTERN_MP_FLOAT::trunc_max );
  return 0;
}

Environment

* Operating system (Windows 7 Home Premium SP1 64 bits):
* Compiler: Microsoft Visual Studio Community 2017 15.6.6
* Release and debug mode.
* No specific flags
* CGAL version: 4.11.2
* Boost version: 1.67
* No other libraries

Issue #1815, opened 2017-01-10
Rewrite to_double(Quotient)


mglisse:
I didn't notive when filing issue #1053, but today this function looks like complete nonsense to me. It checks if x.num or x.den is infinite (was that supposed to be nd and dd?), and in that case (at least one is infinite) proceeds to do divisions and various other operations... It was supposed to have some code to handle overflow (why not underflow as well?) and perform a division of the bigger of num and den by the smaller (careful with ring types...), and it even knows that this division might not give such a good precision when the quotient is close to 1, but in that case it prefers to fall back to a likely NaN than give the low precision quotient or refine the approximation.

Issue #1053, opened 2016-04-30
to_double(Quotient<RT>) divides RT


mglisse:
(copy-paste from https://sympa.inria.fr/sympa/arc/cgal-develop/2013-02/msg00134.html)

I noticed that the default implementation of to_double for a Quotient<RT> uses a division on the type RT. However, in the documentation, I can only see a requirement that RT should be a model of IntegralDomainWithoutDivision which I assume doesn't need to have a division ;-)

Existing types like Gmpzf explicitly specialize Real_embeddable_traits on Quotient<Gmpzf> to redefine To_double and work around it.

lrineau:
What do you suggest for the `to_double` of `CGAL::Quotient`? It could apply `to_double` on the numerator and denominator, and then divide.

mglisse:
Applying `to_double` to numerator and denominator is already what it tries first, the RT division is a fallback for cases where that fails. I don't have a particular suggestion and I don't consider this particularly urgent, I just wanted to record the issue in a better place than the mailing list.

PR #3614, opened 2019-01-22
Rewrite to_interval for boost.multiprecision types

mglisse:
## Summary of Changes

We rely a lot on intervals of length 0, for performance. I had written a simple to_interval for boost types (essentially to_double enlarged, so it never has length 0), but I have to copy the one from Gmpq/mpq_class to avoid regressions. I am also adding the boost types to the utilities.cpp test, which required fixing Coercion_traits.
Note that cpp_int/cpp_rational (as opposed to mpz_int/mpq_rational) will still use the simple to_interval, which may make them slower for some applications. But then we currently only use them if GMP and LEDA are not available, so it doesn't matter.

## Release Management

### DIFF
--- a/Number_types/include/CGAL/boost_mp.h
+++ b/Number_types/include/CGAL/boost_mp.h
@@ -210,10 +210,14 @@ struct RET_boost_mp_base
             double s = i;
             double inf = std::numeric_limits<double>::infinity();
             int cmp = x.compare(i);
-            if (cmp > 0)
+            if (cmp > 0) {
               s = nextafter(s, +inf);
-            else if (cmp < 0)
+              CGAL_assertion(x.compare(s) < 0);
+            }
+            else if (cmp < 0) {
               i = nextafter(i, -inf);
+              CGAL_assertion(x.compare(i) > 0);
+            }
             return std::pair<double, double> (i, s);
         }
     };
--- a/Number_types/include/CGAL/boost_mp.h
+++ b/Number_types/include/CGAL/boost_mp.h
@@ -203,11 +203,18 @@ struct RET_boost_mp_base
         : public CGAL::cpp98::unary_function< Type, std::pair< double, double > > {
         std::pair<double, double>
         operator()(const Type& x) const {
+            // See if https://github.com/boostorg/multiprecision/issues/108 suggests anything better
             // assume the conversion is within 1 ulp
             // adding IA::smallest() doesn't work because inf-e=inf, even rounded down.
-            double d = x.template convert_to<double>();
+            double i = x.template convert_to<double>();
+            double s = i;
             double inf = std::numeric_limits<double>::infinity();
-            return std::pair<double, double> (nextafter (d, -inf), nextafter (d, inf));
+            int cmp = x.compare(i);
+            if (cmp > 0)
+              s = nextafter(s, +inf);
+            else if (cmp < 0)
+              i = nextafter(i, -inf);
+            return std::pair<double, double> (i, s);
         }
     };
 };
@@ -234,28 +241,28 @@ struct RET_boost_mp <boost::multiprecision::mpz_int>
         std::pair<double, double>
         operator()(const Type& x) const {
 #if MPFR_VERSION_MAJOR >= 3
-	  MPFR_DECL_INIT (y, 53); /* Assume IEEE-754 */
-	  int r = mpfr_set_z (y, x.backend().data(), MPFR_RNDA);
-	  double i = mpfr_get_d (y, MPFR_RNDA); /* EXACT but can overflow */
-	  if (r == 0 && is_finite (i))
-	    return std::pair<double, double>(i, i);
-	  else
-	  {
-	    double s = nextafter (i, 0);
-	    if (i < 0)
-	      return std::pair<double, double>(i, s);
-	    else
-	      return std::pair<double, double>(s, i);
-	  }
+          MPFR_DECL_INIT (y, 53); /* Assume IEEE-754 */
+          int r = mpfr_set_z (y, x.backend().data(), MPFR_RNDA);
+          double i = mpfr_get_d (y, MPFR_RNDA); /* EXACT but can overflow */
+          if (r == 0 && is_finite (i))
+            return std::pair<double, double>(i, i);
+          else
+          {
+            double s = nextafter (i, 0);
+            if (i < 0)
+              return std::pair<double, double>(i, s);
+            else
+              return std::pair<double, double>(s, i);
+          }
 #else
-	  mpfr_t y;
-	  mpfr_init2 (y, 53); /* Assume IEEE-754 */
-	  mpfr_set_z (y, x.backend().data(), GMP_RNDD);
-	  double i = mpfr_get_d (y, GMP_RNDD); /* EXACT but can overflow */
-	  mpfr_set_z (y, x.backend().data(), GMP_RNDU);
-	  double s = mpfr_get_d (y, GMP_RNDU); /* EXACT but can overflow */
-	  mpfr_clear (y);
-	  return std::pair<double, double>(i, s);
+          mpfr_t y;
+          mpfr_init2 (y, 53); /* Assume IEEE-754 */
+          mpfr_set_z (y, x.backend().data(), GMP_RNDD);
+          double i = mpfr_get_d (y, GMP_RNDD); /* EXACT but can overflow */
+          mpfr_set_z (y, x.backend().data(), GMP_RNDU);
+          double s = mpfr_get_d (y, GMP_RNDU); /* EXACT but can overflow */
+          mpfr_clear (y);
+          return std::pair<double, double>(i, s);
 #endif
         }
     };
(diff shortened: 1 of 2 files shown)


lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Jan 22, 2019, 3:47 PM GMT+1](https://github.com/CGAL/cgal/pull/3614#issue-246632284 "2019-01-22T14:47:36Z - Replied by Github Reply Comments"):
> Note that cpp_int/cpp_rational (as opposed to mpz_int/mpq_rational) will still use the simple to_interval, which may make them slower for some applications. But then we currently only use them if GMP and LEDA are not available, so it doesn't matter.

If I remember well, `mpq_rational` is chosen before `CGAL::MP_Float`, if the ther exact number types are absent. So it matters if `mpq_rational` is slower than `CGAL::MP_Float`.

mglisse:
> If I remember well, `mpq_rational` is chosen before `CGAL::MP_Float`, if the ther exact number types are absent. So it matters if `mpq_rational` is slower than `CGAL::MP_Float`.

@lrineau You are replying to a sentence where I say that I have made `mpq_rational` fast but not `cpp_rational`...

mglisse:
Ah, yes, `cpp_rational` is chosen before `Quotient<MP_Float>`. Let me check the performance on Efi's testcase.

lrineau:
Indeed, I mixed up `mpq_rational` (that is a sort of `Gmpq` with expression templates), and `cpp_rational` (that is... what? an equivalent of `MP_Float`?).

mglisse:
`cpp_rational` is expression templates on top of https://www.boost.org/doc/libs/1_69_0/libs/rational/rational.html, a generic rational type on top of an integer, where the integer here is essentially `cpp_int`, a bigint written entirely in boost.

mglisse:
Apparently I am misremembering. boost.rational sucks enough that I can't find an example where cpp_rational is faster than `Quotient<MP_Float>`... So we should disable it in Exact_field_selector (although MP_Float seems to have more correctness issues).
`Quotient<cpp_int>` is more than 3 times faster than `Quotient<MP_Float>` on arbitrary fractions, but if the numbers are converted from `double` and we don't do divisions, then the comparison is reversed (I guess the denominator is 1 for all numbers with `Quotient<MP_Float>` in that case). I am thinking of changing BOOST_cpp_arithmetic_kernel from `cpp_int+cpp_rational` to `cpp_int+Quotient<cpp_int>`...
Anyway, for the coming release, I think disabling cpp_rational makes sense.

mglisse:
The conversions from boost types to intervals should now be tight. This only leaves the issue that `cpp_rational` is super slow. That should probably be handled as a separate issue. Option 1 is to replace it with `Quotient<cpp_int>`. There are still cases where that will be slower than `Quotient<MP_Float>`, but there are also cases where it is faster (in particular Efi's testcase). Option 2 is to change Exact_field_selector (&co) never to select `cpp_int` and `cpp_rational`. Other options seem to involve significantly more work (possibly create a new backend to replace boost.rational). I am in favor of option 1 at this point.

efifogel:
Yes, me to (option 1) :)
                         _/

mglisse:
Comparing the number of bug reports, `MP_Float` has #3303 and #3583 (and I believe more that were not copied to github), while `Quotient` of something that is not MP_Float, Gmpzf or Gmpz has #1053, #1815. Neither `Quotient<MP_Float>` nor `Quotient<cpp_int>` feel super safe. I guess I could specialize to_double (and to_interval?) for `Quotient<cpp_int>` to create a temporary `mpq_rational` and get its to_double or to_interval, that's more or less what we do for `to_double(Quotient<Gmpz>)`...
NB: I am purposefully not doing any of that in this PR, it should be a separate PR.

mglisse:
@lrineau What does the label TODO mean? I can't seem to find a description of the labels. As I said above, I'd rather we integrate this branch (if the tests look ok) and work on the rest in a separate branch. I can copy the remarks here in a new issue if that helps.

maxGimeno:

lrineau:
I set the label `TODO` as a reminder not to merge this PR blindly without a second thought.

This PR is clearly an improvement over `master` (that contains #3406), but I wonder if it is better than CGAL-4.13 in all use-cases. I have to have a sharp look at the code. And maybe the solution is to revert the merge of #3406 until you fix all the uses cases. For now I do not know.

mglisse:
If you want to go back to `Quotient<MP_Float>`, there is #3627 for you.

mglisse:
(Commenting here rather than in #3627 so it is all in one place)
The speed difference between cpp_rational and `Quotient<MP_Float>` is actually a bit unfair. Looking at `Quotient<cpp_int>`, which is faster than `Quotient<MP_Float>`, if I add
```c++
namespace CGAL { 
  void simplify_quotient(cpp_int&a,cpp_int&b){
    cpp_int c=gcd(a, b); a/=c; b/=c;
  }
}
```
it becomes as slow as cpp_rational. I had only done a few tests with numbers that do not grow very large... In other contexts with larger numbers, cpp_rational becomes faster. So the question is actually about how large the rationals are that we typically encounter in CGAL, and whether **reducing fractions** helps (keeps the numbers smaller) more than it hurts (it is costly, gcd+2*div). `simplify_quotient` for MP_Float actually has 3 different implementations, and if I enable the third one (that does a true reduction), `Quotient<MP_Float>` becomes slower than cpp_rational for small numbers as well. The status quo is that `Quotient<MP_Float>` purposefully does not reduce.

efifogel:
This is probably a naive question, but isn't it possible to attempt the
simplification conditionally based on the number of operations the number
goes through (perhaps with a different weight assigned to multiplications
and additions) or the bit length of the nominator and denominator (or both)?
                         _/

mglisse:
> This is probably a naive question, but isn't it possible to attempt the simplification conditionally based on the number of operations the number goes through (perhaps with a different weight assigned to multiplications and additions) or the bit length of the nominator and denominator (or both)?

Yes, you could. You could store a counter in the rational, use the sum of the 2 counters whenever you compute a+b or a*b, and have simplify_quotient do nothing if the counter is less than say 10. Or simplify_quotient could check the "size" of the numerator and denominator and do nothing unless the min of the 2 has more than 128 bits. Or any number of heuristics. But how do we tune that? We would need a representative set of benchmarks where we could objectively say: this is faster. Also, this is a slippery slope towards reimplementing an efficient rational type, whereas here we are just trying to provide a fallback for when GMP is unavailable, so it doesn't need to be as fast as GMP and we shouldn't invest too much time in it. Note that GMP always reduces fractions, but its reduction code is efficient enough that this doesn't seem to be a problem.
So yes, maybe a simple heuristic will work, but I don't know which one or how to validate that it works (or even define "work"). Anyone is welcome to give it a try.

afabri:
When we do not have GMP/Leda but define `CGAL_USE_BOOST_MP`, then the  Exact_rational and the Exact_integer are not compatible, because we have `Quotient<MP_Float>`  and `boost::multiprecision::cpp_int`.  @mglisse Isn't this incompability a bug?

afabri:
Another strange thing is that we test for  `#ifdef CGAL_USE_GMP  #elif  CGAL_USE_CORE`.  I guess the latter implies the former, so we can remove the latrter.   If for any reason it was possible, then the same must be done for `Exact_integer`, again for the compatibility between `Exact_rational` and `Exact_integer`.

mglisse:
> When we do not have GMP/Leda but define `CGAL_USE_BOOST_MP`, then the Exact_rational and the Exact_integer are not compatible, because we have `Quotient<MP_Float>` and `boost::multiprecision::cpp_int`. @mglisse Isn't this incompability a bug?

In previous releases, we had no Exact_integer at all in that case, so at least this isn't a regression. I don't know if MP_Float would work as Exact_integer, it behaves quite differently from a true integer, so we probably don't have any type compatible with MP_Float that we could use. Better would be to replace `Quotient<MP_Float>` in Exact_field_selector with `Quotient<cpp_int>` or `cpp_rational`, but the discussion here concluded that this should wait. Or do you want to make Exact_rational different from `Exact_field_selector<double>`?

> Another strange thing is that we test for `#ifdef CGAL_USE_GMP #elif CGAL_USE_CORE`. I guess the latter implies the former, so we can remove the latter.

Yes. I didn't clean it up but did notice that it seemed useless.

afabri:
Reading this discussion I am wondering how much is to gain by not simplifying when we are interested in the sign of a determinant.  At least the simplification for the last number is not necessary if we are only interested in the sign.

danston:
@mglisse Could you please explain what you mean with "Efi's testcase" that you were using to test performance? Thanks!

mglisse:
"Efi's testcase" is a testcase that Efi sent in private email to Andreas, Laurent, Sebastien and me on 22/1/2019, so I cannot post it, although it is rather short. It calls minkowski_sum_2, and the input polyhedra have integer coordinates.

danston:
Great. This is enough. We will find it. Thanks!

PR #3616, opened 2019-01-22
Reset the halfedge pointer of the faces before setting them

sloriot:
Before the patch, when used with `Bounded_kernel` the halfedge pointer of the unbounded face could be a removed halfedge.


maxGimeno:

PR #3618, opened 2019-01-23
Polyhedron demo: Fix normal estimation parameter

sgiraudot:
## Summary of Changes

Copy/paste error, the jet estimation was using the PCA neighbor parameter instead of its own.

## Release Management

* Bug introduced in 4.13


maxGimeno:

PR #3624, opened 2019-01-24
P2T2: Fix not using periodic vertex bases in P2T2 examples

MaelRL:
## Summary of Changes

Some examples in P2T2 were using insufficient vertex base classes (only T2 vertex classes and not P2T2 vertex classes). 

It was most likely only seen on some Windows platforms (e.g. https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-105/Periodic_2_triangulation_2_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz), because that compiler was over-zealous and compiling functions not used by the examples (whatever the P2T2 vertex base class brings is rarely used).

## Release Management

* Affected package(s): `Periodic_2_triangulation_2`


maxGimeno:
There is one error at runtime here: https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-106.shtml#Periodic_2_triangulation_2_Examples

PR #3627, opened 2019-01-24
Prefer Quotient<MP_Float> to cpp_rational for now.

mglisse:
As discussed in #3614, to avoid the temptation to revert #3406.


maxGimeno:

PR #3628, opened 2019-01-25
Surface mesh approximation: Fix demo crash

### LINKED ISSUE #3617: VSA demo crash
In the demo:

1. Load any file
2. Generate seeds
3. Try to un-hide the polygon soup or the patch polygons
4. ???
5. <s>Profit!!!</s> Crash!!!

Likely because they haven't been generated yet, but then they shouldn't be visible.

aboudev:
## Fix demo crash #3617 

Generate visualize items after meshing.

## Release Management

* Affected package(s): Surface mesh approximation
* Issue(s) solved (if any): fix #3617


sloriot:
I don't know if you already did it but It would be nice to hide the widget and only display it when the action is called from the menu. @maxGimeno can probably help if you don't know how to do it.

aboudev:
I think we've done that in commit 0cbb170f6e4619bb432e38f4ee9da192de6660cf.

sloriot:
I think `dock_widget->setVisible(false);` in the `init()` function must be added.

maxGimeno:
@sloriot more precisely, in the init() function there is 
>   dock_widget->setVisible(true);

that should be replaced by 
>   dock_widget->setVisible(false);

aboudev:
The visibility of the widget was set in commit 493b1b5ec1c80c3a8704e70e24458d94da6c91b5 and is still in the code.
https://github.com/CGAL/cgal/blob/cb402f1e91e685383d2d5e333e12def2b6d9b91b/Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Surface_mesh_approximation_plugin.cpp#L81

maxGimeno:
Yes, you are right, I was looking at the previously stated commit. It should be good with the current code.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Surface_mesh_approximation_plugin.cpp
```diff
@@ -404,19 +365,18 @@ void Polyhedron_demo_surface_mesh_approximation_plugin::on_buttonFit_clicked() {
...
-
   sm_item->color_vector() = approx.proxy_colors();
   sm_item->setItemIsMulticolor(true);
   sm_item->computeItemColorVectorAutomatically(false);
   sm_item->invalidateOpenGLBuffers();
   scene->itemChanged(scene->item_id(sm_item));
 
   if (ui_widget.checkAutomatic->isChecked()) {
+    if (!approx.visual_items().has_meshing_items) {
+      mi->information(QString("Please meshing before checking auto meshing."));
```
maxGimeno: "Please mesh" if I'm not mistaken.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Surface_mesh_approximation_plugin.cpp
```diff
@@ -456,19 +416,18 @@ void Polyhedron_demo_surface_mesh_approximation_plugin::on_buttonAdd_clicked() {
...
```
maxGimeno: Same here

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Surface_mesh_approximation_plugin.cpp
```diff
@@ -508,19 +467,18 @@ void Polyhedron_demo_surface_mesh_approximation_plugin::on_buttonTeleport_clicke
...
```
maxGimeno: And here.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Surface_mesh_approximation_plugin.cpp
```diff
@@ -562,19 +520,18 @@ void Polyhedron_demo_surface_mesh_approximation_plugin::on_buttonSplit_clicked()
...
```
maxGimeno: And here.

maxGimeno:
I took the liberty of fixing the errors with the new master.

maxGimeno:

PR #3635, opened 2019-01-27
Drop unused template parameter from Lazy.

mglisse:
* Affected package(s): Filtered_kernel, Number_types

(I also removed some parameters from Lazy_functor_2_1, but since it looks unused... I guess testing will show)

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -61,42 +61,42 @@ template <class E,
           class K>
 class Lazy_kernel_base;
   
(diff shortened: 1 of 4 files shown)


mglisse:
This is included in the Epeck_d branch, no need to keep both.

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Feb 12, 2019, 8:07 AM GMT+1](https://github.com/CGAL/cgal/pull/3635#issuecomment-462642401 "2019-02-12T07:07:12Z - Replied by Github Reply Comments"):
> This is included in the Epeck_d branch, no need to keep both.

Indeed. Replaced by #3660.

PR #3637, opened 2019-01-28
fix wrong ref

sloriot:
Fixes a broken link

@MaelRL shouldn't we move the function (file) to BGL before the release?

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
@@ -258,7 +258,7 @@ namespace Polygon_mesh_processing {
     return border_counter;
   }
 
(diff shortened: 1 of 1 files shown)


MaelRL:
It belongs with other `border.h` functions. Since `border.h` has some documented functions, it will need compatibility headers, etc. I would say just do it in the next cycle, there is no real hurry (is there?).

sloriot:
Well, the function would be new to 4.14 so it would be better to have it in BGL IMO to avoid the deprecation procedure. There is indeed the function `border_halfedges()` that should be deprecated and renamed `patch_border_halfedges()` IMO. I would also suggest to update the implementation to use a dynamic property in case no face index map is provided. BTW, the doc should be updated (must be a model of `FaceGraph`). I can do the change if you want, let me know.

maxGimeno:

PR #3642, opened 2019-01-30
Convex_hull_d: Initialize to avoid warning

afabri:
## Summary of Changes

Fix warning in the testsuite [4.14-Ic-110](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-110/Convex_hull_d/TestReport_lrineau_ArchLinux-CXX17-Release.gz)
I initialize now an `int` with `0`, but didn't check if the before uninitialized member was ever used.

## Release Management

* Affected package(s): Convex_hull_d

### DIFF
--- a/Convex_hull_d/include/CGAL/Regular_complex_d.h
+++ b/Convex_hull_d/include/CGAL/Regular_complex_d.h
@@ -85,7 +85,7 @@ class RC_vertex_d
 public:
   RC_vertex_d(Simplex_handle s, int i, const Point_d& p) :
     s_(s), index_(i), point_(p) {}
-  RC_vertex_d(const Point_d& p) : point_(p), index_(-42), pp(NULL) {}
+  RC_vertex_d(const Point_d& p) : index_(-42), point_(p), pp(NULL) {}
   RC_vertex_d() :  s_(), index_(-42), pp(NULL) {}
   // beware that ass_point was initialized here by nil_point
   ~RC_vertex_d() {}
--- a/Convex_hull_d/include/CGAL/Regular_complex_d.h
+++ b/Convex_hull_d/include/CGAL/Regular_complex_d.h
@@ -85,8 +85,8 @@ class RC_vertex_d
 public:
   RC_vertex_d(Simplex_handle s, int i, const Point_d& p) :
     s_(s), index_(i), point_(p) {}
-  RC_vertex_d(const Point_d& p) : index_(-42), point_(p), pp(NULL) {}
-  RC_vertex_d() :  s_(),  index_(-42), pp(NULL) {}
+  RC_vertex_d(const Point_d& p) : point_(p), pp(NULL) {}
+  RC_vertex_d() :  s_(), pp(NULL) {}
   // beware that ass_point was initialized here by nil_point
   ~RC_vertex_d() {}
 
--- a/Convex_hull_d/include/CGAL/Regular_complex_d.h
+++ b/Convex_hull_d/include/CGAL/Regular_complex_d.h
@@ -86,7 +86,7 @@ public:
   RC_vertex_d(Simplex_handle s, int i, const Point_d& p) :
     s_(s), index_(i), point_(p) {}
   RC_vertex_d(const Point_d& p) : index_(-42), point_(p), pp(NULL) {}
-  RC_vertex_d() :  s_(), pp(NULL) {}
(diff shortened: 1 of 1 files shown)


afabri:
For me 0 was as good as anything as it is uninitialized, and there is no code  `if(is_uninitialized(index_))`.  If I interpret right what you write the `-42`  is somehow a representative for "uninitialized" as it is a negative number,so I better do the same.

mglisse:
> For me 0 was as good as anything as it is uninitialized,

Not initializing gives the best uninitialized value :wink: 

 > If I interpret right what you write the `-42` is somehow a representative for "uninitialized" as it is a negative

I think (again I don't know this code) it is just more likely to be detected if someone mistakenly uses an "uninitialized" index, maybe they will use `vector[-42]` and get a warning/error, while 0 looks like a valid index. Setting a value instead of no value makes it harder to detect uninitialized uses, either at compile time with static analysis or at runtime with sanitizers/valgrind. Using a value that is as likely to cause issues as possible is a way to mitigate this.

For this particular warning, the issue is that we create a Vertex and immediately copy it, including the uninitialized member. I think the following would make sense and possibly remove the warning:
```diff
--- a/Convex_hull_d/include/CGAL/Regular_complex_d.h
+++ b/Convex_hull_d/include/CGAL/Regular_complex_d.h
@@ -524,9 +524,7 @@ Vertex_handle  new_vertex(const Point_d& p)
         has |p| as the associated point, but is has no associated
         simplex nor index yet.}*/
 { 
-  Vertex v(p);
-  Vertex_handle h = vertices_.insert(v);
-  return h;
+  return vertices_.emplace(p);
 }
 
 void associate_vertex_with_simplex(Simplex_handle s, int i, Vertex_handle v)
```

afabri:
@lrineau who do you expect to approve this?

Convex_hull_d/include/CGAL/Regular_complex_d.h
```diff
@@ -86,7 +86,7 @@ class RC_vertex_d
   RC_vertex_d(Simplex_handle s, int i, const Point_d& p) :
     s_(s), index_(i), point_(p) {}
   RC_vertex_d(const Point_d& p) : index_(-42), point_(p), pp(NULL) {}
-  RC_vertex_d() :  s_(), index_(-42), pp(NULL) {}
+  RC_vertex_d() :  s_(), pp(NULL) {}
```
mglisse: Your earlier patch introduced `index_(-42)` for the other constructor, not this one. The combined diff then shows a strange move of this initialization from one line to the previous line...

Convex_hull_d/include/CGAL/Regular_complex_d.h
```diff
@@ -524,9 +524,7 @@ Vertex_handle  new_vertex(const Point_d& p)
         has |p| as the associated point, but is has no associated
         simplex nor index yet.}*/
 { 
-  Vertex v(p);
-  Vertex_handle h = vertices_.insert(v);
-  return h;
+  return vertices_.emplace(p);;
```
mglisse: I think you could do the same for the other overload of `new_vertex` right above, while you are there.

lrineau:
It's @mglisse who objected to the changes. I set the flag to remember that there is an ongoing discussion. 

Marc, what do you think of the latest version of the changes?

mglisse:
I am ok with it. I would have preferred not to introduce new unnecessary initializations, but at least we analyzed the issue, found the root cause (copy in insert) and fixed it (use emplace), so I don't mind as much.

mglisse:
By the way, should we wrap the extra initializations that are done just to quiet warnings in some macro, as documentation, to remember where they are, and so we can disable them if we want to?

lrineau:
I think the variable should not be initialized at all. With a variable initialized to a stupid valid (like `-42`), the compilers static analyzers cannot do any useful analysis. Being non-initialized is better, in my opinion. The call to `emplace` instead of `insert` has already fixed the warning.

afabri:
So you also suggest to remove the pre-existing `(-42)` ?

mglisse:
> So you also suggest to remove the pre-existing `(-42)` ?

If that does not cause any new warning or error...

afabri:
Successfully tested in [CGAL-4.14-Ic-119](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-119/Convex_hull_d/TestReport_lrineau_ArchLinux-CXX17-Release.gz)

PR #3644, opened 2019-01-30
Give Eigen a scoring strategy for pivot selection with intervals

mglisse:
## Summary of Changes

Without this, when computing some decompositions for a matrix of intervals, Eigen looks for the largest element in a column (for instance). There may easily be 2 equal, slightly imprecise numbers that could equally well be used as pivots, but Eigen ends up spuriously throwing in the comparison between them. So we provide a different strategy for picking the pivot.

This is a functionality that I added to Eigen years ago, but for some reason I completely forgot to push the CGAL part using it :disappointed: . On a basic test of computing alpha-complexes in Gudhi, this divides the computation time by 8.

## Release Management

* Affected package(s): Number_types (as far as I know only NewKernel_d uses Eigen with intervals)

### DIFF
--- a/Number_types/include/CGAL/Interval_nt.h
+++ b/Number_types/include/CGAL/Interval_nt.h
@@ -52,6 +52,7 @@
 #include <CGAL/FPU.h>
 #include <CGAL/IO/io.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #3647, opened 2019-01-31
Polyhedron Demo: Fix mesh_3 without edge protection.

maxGimeno:
## Summary of Changes
With no edge protection, detect_features() is not called, so no face-patch_id map was created. And yet this map is always used, so it crashed. I added a default map before the meshing to fix that.
## Release Management

* Issue(s) solved (if any): fix ##3638


maxGimeno:
I don't think so. Or just a doc issue, maybe, if it is not specifyed that the mesh must have a f:pid map available.

maxGimeno:
Ok, so I'd say the polyhedral_mesh_domain_3 should document its need for an initialized property_map in the graph for face_patch_id_t<Patch_id>(). 
@lrineau, do you agree ?

lrineau:
I wrote a test. And launched the testsuite. And then I have modified the patch because the testsuite failed.

Now, everything has been tested correctly on my machine, with:
```shell-session
$ ctest -L Mesh_3 -j5
```

maxGimeno:

PR #3651, opened 2019-01-31
CGAL_Core: Remove call of setprecision()

### LINKED ISSUE #3606: Illegal use of std::setprecision
https://github.com/CGAL/cgal/blob/daa1493e984a61e0d221b54a4dd97fec6db0a3d9/CGAL_Core/include/CGAL/CORE/CoreDefs.h#L325

in STL it's iomanip and not a global function, therefore return value cannot be ignored and must be used in stream. I don't know where and how it worked as intended, but it goes against standard

afabri:
## Summary of Changes

A call to `std::setprecision()` does not affect any stream. In order to do so the returned object would have to be inserted into a stream. So we can just remove the line.  Thank you for pointing this out @ysyrov.

## Release Management

* Affected package(s): Core
* Issue(s) solved (if any): fix #3606

### DIFF
--- a/CGAL_Core/include/CGAL/CORE/CoreDefs.h
+++ b/CGAL_Core/include/CGAL/CORE/CoreDefs.h
@@ -322,7 +322,6 @@ inline bool setRationalReduceFlag(bool f) {
 inline void CORE_init(long d) {
   get_static_defAbsPrec() = CORE_posInfty;
   get_static_defOutputDigits() = d;
-  std::setprecision(get_static_defOutputDigits());
 }
 
 /// change to scientific output format


maxGimeno:

MaelRL:
milestone says 4.12 but the merge is into master

maxGimeno:
I set the milestone because I remembered somtehing about bugfixes that should target 4.12 too because of support stuff, but I'm not sure at all if it should go into 4.12 or not.

lrineau:
Andreas has create the branch against `master`, and not a release branch. As the fix is not an important one, let's not postpone. I will merge the PR in `master` only.

Issue #3605, opened 2019-01-18
Test Triangulation package with clang


mglisse:
Most tests in the Triangulation package start with a check that disables them for old versions of gcc. Sadly, this check also matches clang, no matter how recent, so we are missing some testing there...

`test_delaunay.cpp`:
```c++
#if defined(__GNUC__) && defined(__GNUC_MINOR__) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
```

afabri:
@maxGimeno can you please fix this.

mglisse:
The test in NewKernel_d seems to use BOOST_GCC to detect the real gcc, it probably just needs to be copied.

lrineau:
Fixed by #3654.

PR #3654, opened 2019-02-01
Triangulation: Fix the tests for GCC<4.4 so it wont match clang.

maxGimeno:
## Summary of Changes
Clang also defines `__GNUC__`, but it defines `__clang__` too, so I also check if `__clang__` is not defined to only match GCC < 4.4.

## Release Management
* Affected package(s):Triangulation
* Issue(s) solved (if any): fix ##3605

### DIFF
--- a/Triangulation/examples/Triangulation/delaunay_triangulation.cpp
+++ b/Triangulation/examples/Triangulation/delaunay_triangulation.cpp
@@ -1,4 +1,5 @@
-#if defined(__GNUC__) && ! defined (__clang__) && defined(__GNUC_MINOR__) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
+#include <CGAL/config.h>
+#if defined(BOOST_GCC) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
 
 #include <iostream>
 int main()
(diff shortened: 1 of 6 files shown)


Triangulation/examples/Triangulation/delaunay_triangulation.cpp
```diff
@@ -1,4 +1,4 @@
-#if defined(__GNUC__) && defined(__GNUC_MINOR__) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
+#if defined(__GNUC__) && not defined (__clang__) && defined(__GNUC_MINOR__) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
```
lrineau: The use of `not` is not portable. It is in the standard, as [Alternative operator representations](https://en.cppreference.com/w/cpp/language/operator_alternative), but not implemented everywhere. Please replace by `!`.

lrineau:
As Marc said in https://github.com/CGAL/cgal/issues/3605#issuecomment-459462468, I think that would be better to use `BOOST_GCC`:
```c++
#include <boost/config.hpp>
#if defined(BOOST_GCC) && (__GNUC__ <= 4) && (__GNUC_MINOR__ < 4)
```

maxGimeno:
I'm not really confortable with adding an include to boost when it is not necessary, especially in examples where it might be a little confusing for users that might think boost is needed when it is not.
I don't see why the `! defined (__clang__) wouldn't work, so I'll stick with that for now, and if it is really mot enough, I'll come to terms with boost and use it.

maxGimeno:
Then I can include CGAL/config.h instead of boost/config, to hide it from the user :P

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Feb 4, 2019, 2:07 PM GMT+1](https://github.com/CGAL/cgal/pull/3654#issuecomment-460243730 "2019-02-04T13:07:39Z - Replied by Github Reply Comments"):
> Then I can include CGAL/config.h instead of boost/config, to hide it from the user :P

Exactly... but not so hidden: the macro `BOOST_GCC` would still start with `BOOST`.

maxGimeno:
BOOST_GCC is pretty straight forward, it seems obvious enough what it is here for. include boost/smth is less. Anyway, I'm probably overthinking it.

maxGimeno:

PR #3655, opened 2019-02-01
Polyhedron demo: First draft for a  WRL plugin

maxGimeno:
## Summary of Changes
This adds a basic WRL reader to the demo. It won't read a file without a "Shape" keyword, and should be enhanced in the future.
## Release Management

* Affected package(s):Polyhedron_Demo

### DIFF
--- a/Polyhedron_IO/include/CGAL/IO/WRL_reader.h
+++ b/Polyhedron_IO/include/CGAL/IO/WRL_reader.h
@@ -1,0 +1,340 @@
+// Copyright (c) 2018 GeometryFactory (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org).
+// You can redistribute it and/or modify it under the terms of the GNU
+// General Public License as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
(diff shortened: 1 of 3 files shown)


Polyhedron_IO/include/CGAL/IO/WRL_reader.h
```diff
@@ -0,0 +1,340 @@
...
+      face.push_back(new_id);
+    }
+
+    all_faces.push_back(face);
+  }
+}
+
+// The function below does NOT merge the shapes together
+template <typename Point, typename Face>
+void write_shapes_to_OFF(const std::string filename,
```
MaelRL: You can remove this function and the function 'convert_WRL_to_OFF'

maxGimeno:

MaelRL:
I think it would be best to postpone that to 4.15, Andreas already had some files that couldn't be opened with just this code. (The issue was that the geometry was only put in a single shape, and all the other shapes were just combinatorics; I guess that in that case you're supposed to use the coordinates from the first shape in the other shapes too...)

sloriot:
@MaelRL @afabri do you have an updated version of the reader that could be contributed here?

MaelRL:
As of now, no.

lrineau:
Do we want to close this pull-request? It seems the current code is not good enough: 

[**@MaelRL**](https://github.com/MaelRL) commented on [Feb 7, 2019, 10:25 AM GMT+1](https://github.com/CGAL/cgal/pull/3655#issuecomment-461344330 "2019-02-07T09:25:30Z - Replied by Github Reply Comments"):
> I think it would be best to postpone that to 4.15, Andreas already had some files that couldn't be opened with just this code. (The issue was that the geometry was only put in a single shape, and all the other shapes were just combinatorics; I guess that in that case you're supposed to use the coordinates from the first shape in the other shapes too...)

sloriot:
Closing it. We will reopen/finish it if we need to.

PR #3658, opened 2019-02-04
Polyhedron Demo: Fix isotropic remeshing plugin

### LINKED ISSUE #3649: Issue with isotropic remeshing plugin
[mesh.off.gz](https://github.com/CGAL/cgal/files/2817542/mesh.off.gz)
[isotropic_remeshing_example.cpp.gz](https://github.com/CGAL/cgal/files/2817545/isotropic_remeshing_example.cpp.gz)
If I'm trying to reproduce what is done in the attached example with the attached mesh and in the demo, I cannot obtain the same result.
I tried both detect-sharp feature 30⁰ + remeshing with protect-edges and using a select of sharp edges.

maxGimeno:
## Summary of Changes
For some reason, the results in and out the demo differs. This should unify them.
Also :
- protection can use detected shard edges and not only selected edges
- Split only can be used with only a selection of edges and not a selection of faces
- When trying to remesh with a selection of edges and a protection, if the constrained are too long, a pop up asks if the user wants to split instead of simply aborting. 

## Release Management

* Issue(s) solved (if any): fix #3649

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Isotropic_remeshing_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Isotropic_remeshing_plugin.cpp
@@ -534,23 +534,35 @@ public Q_SLOTS:
       }
       else if (poly_item)
       {
+        boost::property_map<FaceGraph, CGAL::edge_is_feature_t>::type eif
+          = get(CGAL::edge_is_feature, pmesh);
         if (edges_only)
         {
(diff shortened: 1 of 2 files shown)


lrineau:
@maxGimeno Why the TODO label?

sloriot:
The TODO is for me as @maxGimeno is stuck on the issue and does not understand what's is going on.

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-122.shtml

PR #3660, opened 2019-02-06
Epeck_d

mglisse:
## Summary of Changes

New lazy exact kernel in arbitrary dimensions, reusing the infrastructure from Epick_d and Epeck.

## Release Management

* Affected package(s): NewKernel_d, Kernel_d (doc), Filtered_kernel (internals only)
* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Epeck_d
* License and copyright ownership: Inria, LGPL


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: cpp 
-dist: trusty 
+dist: xenial
```
mglisse: It seems that this requires updating the list of versioned dependencies (libboost1.55-dev, qt55base, etc). I would guess libboost-dev (i.e. 1.58) works. And probably Ubuntu's qt packages should be fine, instead of 3rd party.

lrineau: @maxGimeno Could you please have a look and make our Travis script work with "dist: xenial"? We will need that to switch to C++14.

lrineau: xenial already have a recent toolchain: https://docs.travis-ci.com/user/reference/xenial/#compilers-and-build-toolchain
> clang and llvm 7
> cmake 3.12.4
> gcc 5.4.0

mglisse:
I am going to remove the "(WIP)" from the title because I think the feature is ready (that doesn't mean I won't add more patches, I don't know, but it could go in as it is). I still have to remove the cgalmodif tags (should probably wait 15 days after pre-approval for that). And it would be best if some C++14 infrastructure patches went in first (enable C++14 by default, travis (that one will require a manual merge, or possibly some reverts)).

maxGimeno:
I'm currently  working on uprading travis to Xenial.

mglisse:
> I'm currently working on uprading travis to Xenial.

:+1: Yes, I noticed, thanks :smiley:

lrineau:
The upgrade of our Travis configuration to Xenial (Ubuntu 16.04 LTS) is now finished and merged in `master`. I have merged `master` in this PR, and resolved the conflicts. We can hope that the Travis tests will go fine, now.

mglisse:
I am seeing many mentions of qt4 in the travis logs, that can't be right...

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Feb 12, 2019, 9:56 AM GMT+1](https://github.com/CGAL/cgal/pull/3660#issuecomment-462673515 "2019-02-12T08:56:24Z - Replied by Github Reply Comments"):
> I am seeing many mentions of qt4 in the travis logs, that can't be right...

I do not understand how qt4 got installed. I seems that it is in the base Xenial image of Travis.

The test of #3666 was incomplete because of our dependency testing:
https://travis-ci.org/CGAL/cgal/jobs/491682188
```
Package ignored because none of its dependencies has been modified.
```

We build QGLViewer, for pull-requests from the `releases/CGAL-4.12-branch`, and it seems during that build qmake-qt4 was chosen. ... Actually, pull-requests from 
`releases/CGAL-4.12-branch` are tested by Travis against `releases/CGAL-4.12-branch` itself. In `master` we could completely remove that compilation of QGLViewer. @maxGimeno, what do you think?

Surface_mesh/benchmark/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ include_directories(BEFORE "../include")
 #add_definitions("-pg")
 #SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pg")
 # add_definitions("-g")
-add_definitions("-std=c++11")
+add_definitions("-std=c++1y")
```
lrineau: I think none of that is relevant. CMake handles the dependency on C++14 here:
https://github.com/CGAL/cgal/blob/328f2bdbc6f9d2056e967b8d5fbc5ce025dbe361/Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake#L104-L106

mglisse: Well, it seemed to be causing errors in travis, with this C++11 flag overriding the C++14 one from cmake, and it seemed easier for someone who doesn't know this configuration to replace 11 with 1y than remove the line completely, but if you want to remove it, please do.

lrineau: You are right that:
  - our CMake scripts add `-std=gnu++14` (or similar),
  - but maybe the flags specified on the command line, or with `add_definitions` are after.

mglisse:
It looks like P3T3 requires libqt5sql5-sqlite?

maxGimeno:
The missing qt5 dependency should be fixed now, but there is a new dependency that should be added : 
>Differences in Triangulation_3: NewKernel_d are new and not committed.
 Differences in Triangulation_2: NewKernel_d are new and not committed.
 Differences in Surface_sweep_2: NewKernel_d are new and not committed.
 Differences in Surface_mesh_skeletonization: NewKernel_d are new and not committed.
 Differences in Surface_mesher: NewKernel_d are new and not committed.
 Differences in Straight_skeleton_2: NewKernel_d are new and not committed.
 Differences in Snap_rounding_2: NewKernel_d are new and not committed.
 Differences in Skin_surface_3: NewKernel_d are new and not committed.
 Differences in Segment_Delaunay_graph_Linf_2: NewKernel_d are new and not committed.
 Differences in Segment_Delaunay_graph_2: NewKernel_d are new and not committed.
 Differences in Scale_space_reconstruction_3: NewKernel_d are new and not committed.
 Differences in Polytope_distance_d: NewKernel_d are new and not committed.
 Differences in Polyline_simplification_2: NewKernel_d are new and not committed.
 Differences in Polygon_mesh_processing: NewKernel_d are new and not committed.
 Differences in Poisson_surface_reconstruction_3: NewKernel_d are new and not committed.
 Differences in Point_set_processing_3: NewKernel_d are new and not committed.
 Differences in Periodic_3_triangulation_3: NewKernel_d are new and not committed.
 Differences in Periodic_3_mesh_3: NewKernel_d are new and not committed.
 Differences in Number_types: NewKernel_d are new and not committed.
 Differences in Nef_3: NewKernel_d are new and not committed.
 Differences in Minkowski_sum_3: NewKernel_d are new and not committed.
 Differences in Minkowski_sum_2: NewKernel_d are new and not committed.
 Differences in Mesh_3: NewKernel_d are new and not committed.
 Differences in Linear_cell_complex: NewKernel_d are new and not committed.
 Differences in Kernel_23: NewKernel_d are new and not committed.
 Differences in Heat_method_3: NewKernel_d are new and not committed.
 Differences in Generalized_map: NewKernel_d are new and not committed.
 Differences in Filtered_kernel: NewKernel_d are new and not committed.
 Differences in Envelope_3: NewKernel_d are new and not committed.
 Differences in Convex_hull_3: NewKernel_d are new and not committed.
 Differences in Convex_decomposition_3: NewKernel_d are new and not committed.
 Differences in Combinatorial_map: NewKernel_d are new and not committed.
 Differences in Circular_kernel_3: NewKernel_d are new and not committed.
 Differences in Circular_kernel_2: NewKernel_d are new and not committed.
 Differences in Boolean_set_operations_2: NewKernel_d are new and not committed.
 Differences in BGL: NewKernel_d are new and not committed.
 Differences in Arrangement_on_surface_2: NewKernel_d are new and not committed.
 Differences in Alpha_shapes_3: NewKernel_d are new and not committed.
 Differences in Alpha_shapes_2: NewKernel_d are new and not committed.
 Differences in Algebraic_kernel_for_spheres: NewKernel_d are new and not committed.
 Differences in Algebraic_kernel_for_circles: NewKernel_d are new and not committed.
 Differences in Advancing_front_surface_reconstruction: NewKernel_d are new and not committed.
  You can run cmake with options CGAL_ENABLE_CHECK_HEADERS and CGAL_COPY_DEPENDENCIES ON, make the target packages_dependencies and commit the new dependencies files,
 or simply manually edit the problematic files.

Just add those in the package_info of NewKernel_d and it should be ok.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Feb 13, 2019, 10:13 AM GMT+1](https://github.com/CGAL/cgal/pull/3660#issuecomment-463119365 "2019-02-13T09:13:29Z - Replied by Github Reply Comments"):
> Just add those in the package_info of NewKernel_d and it should be ok.

Please don't. That new dependency is something we want to fix.

Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -35,8 +35,11 @@
 #include <CGAL/Bbox_2.h>
 #include <CGAL/Bbox_3.h>
 #include <vector>
+#include <type_traits>
 #include <CGAL/Default.h>
 #include<CGAL/tss.h>
+#include <CGAL/is_iterator.h>
+#include <CGAL/transforming_iterator.h>
```
lrineau: This include of `<CGAL/transforming_iterator.h>` introduced a dependency to `NewKernel_d`. Could this header be moved to `STL_Extension`, and even merged in `<CGAL/iterator.h>`?

afabri:
What is the transforming_iterator about, and is there nothing in boost that does the job?

Filtered_kernel/include/CGAL/Lazy.h
mglisse: Oh yes, I should move it. Not sure about merging it though.

mglisse:
> What is the transforming_iterator about, and is there nothing in boost that does the job?

As noted inside, this is indeed similar to boost's transform_iterator, except that it wasn't doing exactly what I wanted (wasting space, being too strict about iterator category).

Filtered_kernel/include/CGAL/Lazy.h
lrineau: Having a look at `CGAL::transforming_iterator`, I have two remarks...

  - In `<CGAL/transforming_iterator.h>`, `<CGAL/transforming_pair_iterator.h>`, and `<CGAL/typeset.h>`, you do not follow the CGAL naming guidelines, of having types named starting with a capital letter. I imagine you embraced the coding convention of the STL. I kind of agree with that convention for CGAL things that belong to STL_Extension. Would you like to propose a change of our coding convention?

  - I have seen usages of `transforming_iterator` is a few headers without the corresponding `#include`. You should run `clang/gcc -fsyntax-only` on your headers. The `CHECK` part of our Travis checks only verify the documented headers... and in `NewKernel_d` that means only `<CGAL/Epick_d.h>` and `<CGAL/Epeck_d.h>`.

mglisse: I don't mind adding a capital letter if that helps...
I guess I could add more #includes, but only the 2 documented headers really matter.

afabri: > Not sure about merging it though.

Any good reason for not merging it?

mglisse: Any good reason why packages are split in several files instead of having everything in one huge file?

lrineau: Actually, I do not really care about the naming with capital letter or not. As long as it belongs to a kind of extension to the STL, I would agree that we can following the STL naming conventions.

What really matters is the move of the header <CGAL/transforming_iterator.h>, or its code, to STL_Extension, to avoid adding a dependency.

mglisse:
> @mglisse Can you please fix the conflicts with `master`. That would be better to test this PR as soon as possible.

Couldn't GeometryFactory handle it? They seem to be responsible for them (merging the PRs in FIFO order would have avoided it).

lrineau:
@mglisse is right. @sloriot or @maxGimeno should solve the conflicts.

maxGimeno:
@mglisse I'll need you to add me as collaborator to push a rebase

maxGimeno:
All done.

afabri:
@mglisse shall I readd the pragma push or remove the pop?  I guess the former as I get the warning with VC2015.

afabri:
@mglisse the almost compleytely red line in `Triangulation_2`  happens in the call of `std::swap` in `Handle.h`.

mglisse:
Do what you want with the warning pragmas. I'll take a look at the Handle.h assertion failures (if necessary the Handle.h changes can be reverted, they are only a minor performance help, but the errors could point at something suspicious).

afabri:
@mglisse can you please make me collaborator.

mglisse:
The error in Triangulation_2 is because CGAL's random_shuffle does self swaps. Testing for them to avoid calling iter_swap works. We could also handle copying null Handles. Replacing random_shuffle with the C++11 shuffle seems like the way forward.

afabri:
CGAL's random shuffle is there for making packages like `Triangulation_{2,3}` cross platform deterministic.

mglisse:
I am just going to revert the Handle.h patch, it can be handled (pun!) separately later (most likely by removing the assertions and handling copies of a null handle).

afabri:
@mglisse can you also have a look at:

> In file included from /Users/cgaltester/cgal_test/CGAL-5.0-Ic-13/include/CGAL/Lazy_exact_nt.h:47:
/Users/cgaltester/cgal_test/CGAL-5.0-Ic-13/include/CGAL/Lazy.h:132:25: warning: unused parameter 't' [-Wunused-parameter]
unsigned depth(T const& t) {return 1;} // FIXME: depth(*t) would be better when t is valid, but not for end iterators, and the true answer would iterate on the range, but we can't do that with only one iterator... We need to replace iterators with ranges to solve that.

[link](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-13/Mesh_2/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.0_Release.gz)

afabri:
@mglisse can you also have a look at:

> In file included from /Users/cgaltester/cgal_test/CGAL-5.0-Ic-13/include/CGAL/Lazy_exact_nt.h:47:
/Users/cgaltester/cgal_test/CGAL-5.0-Ic-13/include/CGAL/Lazy.h:132:25: warning: unused parameter 't' [-Wunused-parameter]
unsigned depth(T const& t) {return 1;} // FIXME: depth(*t) would be better when t is valid, but not for end iterators, and the true answer would iterate on the range, but we can't do that with only one iterator... We need to replace iterators with ranges to solve that.

[link](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-13/Mesh_2/TestReport_cgaltester_x86-64_Darwin-13.0_Apple-clang-5.0_Release.gz)

afabri:
And any idea what might go wrong for [VC2015](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-13/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)

mglisse:
> And any idea what might go wrong for [VC2015](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-13/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)

No, sorry, looks like a compiler bug but it isn't clear from the message what exactly confuses it. Since it only affects new functionality on a single old compiler, I guess it isn't that bad.

afabri:
> 
> 
> > And any idea what might go wrong for [VC2015](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-13/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)
> 
> No, sorry, looks like a compiler bug but it isn't clear from the message what exactly confuses it. Since it only affects new functionality on a single old compiler, I guess it isn't that bad.

It is not an old compiler, but one of the compiler we support.

mglisse:
> It is not an old compiler, but one of the compiler we support.

I didn't say it was unsupported, but IIRC it is the oldest version supported (newer versions of that compiler work). Obviously it would be better if it could be made to work, but I don't have time to work on that. And since the issue is only with a new functionality, I believe we can live with it. But feel free to debug this issue, if you can find a simple workaround that would be nice.

afabri:
This  [gist](https://gist.github.com/afabri/0ab6008b3d53b07c1b06347917aad007) fails, and it has a problem with the   `l(Lazy_internal::do_extract(T{},lll)` [here ](https://github.com/CGAL/cgal/pull/3660/files#diff-f50fa2aab278634fb9d68bf3954354eaR150).   What is the `{}` about - sorry about my ignorance.

mglisse:
> What is the `{}` about

`T{}` is essentially the same thing as `T()`, just more modern and with less risk of ambiguity.

maxGimeno:
There are [Errors](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-15.shtml#NewKernel_d) in the testsuite, I revove it from integration.

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Apr 16, 2019, 5:56 PM GMT+2](https://github.com/CGAL/cgal/pull/3660/#issuecomment-483722596 "2019-04-16T15:56:16Z - Replied by Github Reply Comments"):
> > What is the `{}` about
> 
> `T{}` is essentially the same thing as `T()`, just more modern and with less risk of ambiguity. If it works with parentheses, that's great.

@afabri Have you tried to replace `{}` with `()`?

afabri:
Yes, with the same error.

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Apr 16, 2019, 5:10 PM GMT+2](https://github.com/CGAL/cgal/pull/3660#issuecomment-483703027 "2019-04-16T15:10:18Z - Replied by Github Reply Comments"):
> This  [gist](https://gist.github.com/afabri/0ab6008b3d53b07c1b06347917aad007) fails, and it has a problem with the   `l(Lazy_internal::do_extract(T{},lll)` [here ](https://github.com/CGAL/cgal/pull/3660/files#diff-f50fa2aab278634fb9d68bf3954354eaR150).   What is the `{}` about - sorry about my ignorance.

I had a look at the code, and the error message, and it appears that MSVC 2015 is just bad at parsing that C++14 valid code, with fold expressions and a slightly complicated matching. As @mglisse said, that is a new feature... and we know the support for C++14 in MSVC 2015 is not 100% relyable. I suggest we "disable" the feature `Epeck_d` for MSVC 2015: we add preprocessor conditionals so that MSVC 2015 does not see the code.

lrineau:
[**@lrineau**](https://github.com/lrineau) commented on [Apr 30, 2019, 12:33 PM GMT+2](https://github.com/CGAL/cgal/pull/3660#issuecomment-487903431 "2019-04-30T10:33:40Z - Replied by Github Reply Comments"):
> As @mglisse said, that is a new feature... and we know the support for C++14 in MSVC 2015 is not 100% relyable. I suggest we "disable" the feature `Epeck_d` for MSVC 2015: we add preprocessor conditionals so that MSVC 2015 does not see the code.

Well... we will have to make sure `Epick_d` keeps working correctly with MSVC 2015.

mglisse:
I added a sentence in the doc and disabled the Epeck_d tests for VS 2015 (I think). I assume that the compiler only fails if something instantiates the code. If you want more warning / disabling, please go ahead and push what you want to the branch.
Can I get this branch tested and merged?

maxGimeno:
There is [a strange warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-42/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz) in MSVC 2015

mglisse:
> There is [a strange warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-42/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz) in MSVC 2015

Yes, that warning makes no sense to me, the line says `return a*b;` and the compiler complains that I am not doing anything with the result of the multiplication???

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [May 22, 2019, 4:32 PM GMT+2](https://github.com/CGAL/cgal/pull/3660#issuecomment-494828766 "2019-05-22T14:32:00Z - Replied by Github Reply Comments"):
> There is [...] and [other warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-42/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz) in MSVC 2017

For the second warning, it can be reproduced with `-Wconversion -Wno-sign-conversion` on Linux with gcc. The second parameter of `Compute_cartesian_coordinate::operator()` has type `int`:
https://github.com/CGAL/cgal/blob/2f709a40e51515da76dc3bbb0b9e0298354acffc/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h#L143-L155

but it is called in `Iterator_from_indices::dereference` with a `std::ptrdiff_t`:
https://github.com/CGAL/cgal/blob/2f709a40e51515da76dc3bbb0b9e0298354acffc/NewKernel_d/include/CGAL/iterator_from_indices.h#L39-L46
https://github.com/CGAL/cgal/blob/2f709a40e51515da76dc3bbb0b9e0298354acffc/NewKernel_d/include/CGAL/iterator_from_indices.h#L59-L63

mglisse:
When accessing the k-th coordinate of a point, k is a small integer, `int` is a sensible type.
When making a generic wrapper for iterators, the collection we are iterating on could be large, a natural type is `ptrdiff_t`.
If I use the generic wrapper with coordinate accesses, there is a conversion or two between `int` and `ptrdiff_t`. Nothing wrong there.

From the comments, it looks like I did consider using int in the wrapper, although that would mean we cannot use it for large collections. I could also take the coordinate index via an unnecessarily large integer type. I guess I'll have to pick one of them...

lrineau:
Instead of hard-coding `ptrdiff_t`, maybe you can use `std::iterator_traits<Iterator>::difference_type`, and make it the right type is the various cases.

lrineau:
[**@lrineau**](https://github.com/lrineau) commented on [May 24, 2019, 11:10 AM GMT+2](https://github.com/CGAL/cgal/pull/3660#issuecomment-495534604 "2019-05-24T09:10:00Z - Replied by Github Reply Comments"):
> Instead of hard-coding `ptrdiff_t`, maybe you can use `std::iterator_traits<Iterator>::difference_type`, and make it the right type is the various cases.

@mglisse Marc, does that suggestion sound right for you?

As for the other warning from VC++ 2015, I think my commit 45efa5a6c03 will quiet it with just a spurious pair of parenthesis.

mglisse:
> @mglisse Marc, does that suggestion sound right for you?

Not directly, there is no iterator to take it from. I looked into adding a template parameter, it is doable, but it complicates the code. I would rather add a #pragma to disable the warning for now, and think about it later...

> As for the other warning from VC++ 2015, I think my commit [45efa5a](https://github.com/CGAL/cgal/commit/45efa5a6c0335fddf3c54e2de5dfc67b1f29508c) will quiet it with just a spurious pair of parenthesis.

Thanks.

maxGimeno:
Another warning : 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-59/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Jun 17, 2019, 9:09 AM GMT+2](https://github.com/CGAL/cgal/pull/3660#issuecomment-502561595 "2019-06-17T07:09:15Z - Replied by Github Reply Comments"):
> Another warning : 
> https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-59/NewKernel_d/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

That is not another warning. That is the warning debate [four weeks ago](https://github.com/CGAL/cg`al/pull/3660#issuecomment-495501304), that can be reproduced with clang `-Wconversion -Wno-sign-conversion`.

lrineau:
https://github.com/CGAL/cgal/blob/2f709a40e51515da76dc3bbb0b9e0298354acffc/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h#L143-L155

@mglisse, would you mind if the `operator()` was templated by the index type? That would solve the warning:
```patch
diff --git a/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h b/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h
index e7a61332e57..65f73a2a182 100644
--- a/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h
@@ -149,7 +149,8 @@ template<class R_> struct Compute_cartesian_coordinate {
 	typedef Tag_true Is_exact;
 	typedef decltype(std::declval<const first_argument_type>()[0]) result_type;
 
-	result_type operator()(first_argument_type const& v,int i)const{
+	template <typename index_type>
+	result_type operator()(first_argument_type const& v,index_type i)const{
 		return v[i];
 	}
 };
```

mglisse:
> @mglisse, would you mind if the `operator()` was templated by the index type? That would solve the warning:

Does that make the warning disappear? ok then, can you push it?

lrineau:
Done. Pushed in 1957ecc26343281ce411888d40efc67132954feb. I have also pushed a merge of `master`, to solve a minor conflict: 2a5b558eb5288d9ab6dbbc9a55b9f97b8ba2dcb1

lrineau:
[All is green now, in CGAL5.0-Ic-65](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-65.shtml#NewKernel_d).

maxGimeno:
I am not that sure it's all green, check the [Peridoc_4_triangulation_2 line](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-71.shtml#Periodic_4_hyperbolic_triangulation_2), there's a lot of assertions in core. That could come from one of the many PRs about Triangulation_2 too, I'm a little lost in there.

lrineau:
The issues with CORE a because of by PR https://github.com/CGAL/cgal/pull/4013. :disappointed:

mglisse:
`\cgalModifBegin` and `\cgalModifEnd` were apparently still in there :disappointed:

sloriot:
Fixed by 86752ea880 in master.

Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -878,30 +835,23 @@ struct Lazy_construction_nt {
...
-    }                                                                   \
-  }                                                                     \
-
-  BOOST_PP_REPEAT_FROM_TO(1, 6, CGAL_NT_OPERATOR, _)
+  template<class...L>
+  auto operator()(L const&...l) const ->
+  Lazy_exact_nt<std::remove_cv_t<std::remove_reference_t<decltype(ec(CGAL::exact(l)...))>>>
+  {
+    typedef std::remove_cv_t<std::remove_reference_t<decltype(ec(CGAL::exact(l)...))>> ET;
+    typedef std::remove_cv_t<std::remove_reference_t<decltype(ac(CGAL::approx(l)...))>> AT;
```
lrineau: Why not `std::decay_t`? Is there any risk is can be different from `std::remove_cv_t<std::remove_reference_t<T>>``?

mglisse: I don't remember. I probably used this form because there were already some `typename boost::remove_cv< typename boost::remove_reference < typename boost::result_of<` (introduced by Philipp Möller a while ago). And boost's decay used to be very different from the standard one. I also think the standard takes care to use remove_cvref in cases where the difference with decay is irrelevant. In this case, I don't think it matters which one is used (but I didn't look deeply).

PR #3665, opened 2019-02-07
Regular_triangulation_3: Add Regular_triangulation_cell_base_3::hidden_points()

afabri:
## Summary of Changes

The [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-117/Periodic_3_triangulation_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz) for 3D periodic regular triangulations calls `std::distance(b,e)` where the iterators are from a list. The function that returns a range (something we want later anyways) permits to call `.size()`. 

## Release Management

* Affected package(s): Triangulation_3 and Periodic_3_triangulation_3

### DIFF
--- a/Periodic_3_triangulation_3/test/Periodic_3_triangulation_3/test_p3rt3_insert_remove_point_set.cpp
+++ b/Periodic_3_triangulation_3/test/Periodic_3_triangulation_3/test_p3rt3_insert_remove_point_set.cpp
@@ -79,13 +79,13 @@ public:
 
       std::size_t hidden_point_count = 0;
       for (Cell_iterator iter = p3rt3.cells_begin(), end_iter = p3rt3.cells_end(); iter != end_iter; ++iter)
-        hidden_point_count += std::distance(iter->hidden_points_begin(), iter->hidden_points_end());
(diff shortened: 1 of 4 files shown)


afabri:
@MoniqueTeillaud do you think we can reduce the size the point set of the 3D periodic regular triangulation test?

mglisse:
> `std::distance(b,e)` where the iterators are from a list

Too bad other implementations did not copy gcc, where std::distance(list.begin(),list.end()) has constant complexity (it is optimized to a call to `size()`).

afabri:
Successfully tested in [CGAL-4.14-Ic-119](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-119/Periodic_3_triangulation_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2013-Debug-64bits.gz)

afabri:
The conclusion of a discussion with @MoniqueTeillaud was that we better do  tests with random point sets and not with two stored point sets.

maxGimeno:

PR #3666, opened 2019-02-07
Travis: Upgrade travis to xenial

maxGimeno:
## Summary of Changes
Use xenial instead of trusty in travix.


lrineau:
Almost! Now I think we should modify the line
```
compiler: clang-3.6
```
and the other references to `clang-3.6` or `clang++-3.6`.

maxGimeno:
Indeed, I forgot about that

.travis.yml
```diff
@@ -51,27 +51,4 @@ env:
...
-- make 
-- sudo make install &>/dev/null
-- cd ..
-script: 
-- cd ./.travis
-- bash ./build_package.sh $PACKAGE
-notifications:
-  email:
-    on_success: change # default: always
-    on_failure: always # default: always
```
lrineau: That looks wrong. You have removed almost everything from the `.travis.xml`.

maxGimeno: Yes, the script that generates it was expecting the -3.6, i didn't see

lrineau:
You can have a look at the Docker image `ubuntu:xenial`, to get a look at the installation paths in Xenial.

.travis/install.sh
```diff
@@ -2,23 +2,13 @@
...
-while [ $DONE = 0 ]
-do
-  DONE=1 && sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install clang-3.6 zsh \
-flex bison cmake graphviz libgmp-dev libmpfr-dev libmpfi-dev zlib1g-dev libeigen3-dev  libboost1.55-dev \
-libboost-system1.55-dev libboost-program-options1.55-dev libboost-thread1.55-dev libboost-iostreams1.55-dev \
-qt55base qt55script qt55svg qt55tools qt55graphicaleffects libopencv-dev mesa-common-dev libmetis-dev libglu1-mesa-dev \
+  DONE=1 && sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install clang zsh \
+flex bison cmake graphviz libgmp-dev libmpfr-dev libmpfi-dev zlib1g-dev libeigen3-dev  libboost-dev \
+libboost-system-dev libboost-program-options-dev libboost-thread-dev libboost-iostreams-dev \
+qtbase5-dev qtscript5-dev libqt5svg5-dev qttools5-dev qttools5-dev-tools qml-module-qtgraphicaleffects libopencv-dev mesa-common-dev libmetis-dev libglu1-mesa-dev \
```
lrineau: Missing `libqt5opengl5-dev`.

See the list of package of Qt5 on Xenial:
https://packages.ubuntu.com/source/xenial/qtbase-opensource-src

lrineau: `qtbase5-dev` "recommends" `libqt5opengl5-dev`, but that is not a hard requirement. That must be the difference of behavior between the Docker image `ubuntu:xenial` and the Travis image based on Xenial: `apt` on Travis is probably configured to install the minimum.

maxGimeno: Thanks a lot. I've been looking for a pkg like this but without the "lib" prefix, as the other ones don't have it -_-.

PR #3668, opened 2019-02-07
PSP: Reduce the size of the data set in an example

afabri:
## Summary of Changes

10K instead of 1M points is enough to illustrate the callback mechanism.    And it avoids a timeout in the [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-4.14-Ic-117/Point_set_processing_3_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)

## Release Management

* Affected package(s): Point_set_processing

### DIFF
--- a/Point_set_processing_3/examples/Point_set_processing_3/callback_example.cpp
+++ b/Point_set_processing_3/examples/Point_set_processing_3/callback_example.cpp
@@ -44,7 +44,7 @@ struct Progress_to_std_cerr_callback
     // Avoid calling time() at every single iteration, which could
     // impact performances very badly
     ++ nb;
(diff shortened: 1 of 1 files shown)


sgiraudot:
Not really. If I launch this code on my machine, everything is computed in a microsecond and I don't see any advancement (everything is 100% done immediately), which is the point of the example. Maybe the testsuite should use a different value than the user?

afabri:
@maxGimeno please integrate it into the testsuite.  This change was discussed with @sgiraudot

maxGimeno:

PR #3671, opened 2019-02-08
Use only initialized values

sloriot:
Fix for non documented autorefinement functions


maxGimeno:

PR #3675, opened 2019-02-11
fix navtree hack

### LINKED ISSUE #3562: Online Documentation: sidebar is mislinked in Mesh_3 Reference Manual
## Issue Details

When clicking a section in the documentation (latest) sidebar, the correct section is loaded but then the sidebar jumps to a different section, making us lose context since the sidebar no longer shows the section we're reading. This is possibly a problem in other documentation chapters as well. 

Going back to CGAL documentation version 4.11.3, this issue is no longer present. The issue starts appearing in doc version 4.12.1 and is also present in 4.13, "latest" and "master".

Steps to reproduce:
* Navigate to: https://doc.cgal.org/latest/Mesh_3/group__PkgMesh__3.html
* Click the `Reference Manual - Mesh Classes` section in the sidebar. (The correct section text is loaded on the right, but the sidebar jumps to `Reference Manual - Main Concepts - MeshComplexWithFeatures_3InTriangulation_3`.)

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Tested in Chrome, Edge, Firefox on Windows 10
* CGAL version: documentation versions master, latest, 4.13, 4.12.1

sloriot:
Fix #3562

### DIFF
--- a/Documentation/doc/resources/1.8.13/hacks.js
+++ b/Documentation/doc/resources/1.8.13/hacks.js
@@ -46,6 +46,10 @@ $(document).ready(function() {
         // override gotoNode from navtree.js
         gotoNode = function (o,subIndex,root,hash,relpath) {
             var nti = navTreeSubIndices[subIndex][root+hash];
+            if (!nti)
+            {
+              nti = navTreeSubIndices[subIndex][root];
+            }
             if(nti && (nti[0] === 1 && nti[0])) {
                 nti.splice(1, 1);
             }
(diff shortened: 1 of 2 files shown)


sloriot:
@albert-github I wonder whether I should report the behavior change  (`hash` was an empty string before and now is `#` in case no anchor is given).

albert-github:
Yes, I think you should report it together with an example / the place where we can find the problem.
The problem might be due to the extra processing as done by CGAL.

lrineau:
The bug #3562 is fixed, in the testsuite: see at https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.14-Ic-121/master/Mesh_3/group__PkgMesh3MeshClasses.html.

maxGimeno:

PR #3676, opened 2019-02-12
Travis: Remove QGLViewer from Travis

maxGimeno:
## Summary of Changes
Remove QGLViewer from travis. It removes the call to qmake and should fix the problem in xenial.


maxGimeno:
Travis passed on my own account. It should fix the bugs

PR #3679, opened 2019-02-13
P3t3 demo: Fix crash 

maxGimeno:
## Summary of Changes
In find_conflict, if the request point is a vertex of the request cell, there is no conflict and that triggers an assertion. I added a test before calling it in the demo to avoid this situation, that was systematic when adding the moving point to the triangulation.

### DIFF
--- a/Periodic_3_triangulation_3/demo/Periodic_3_triangulation_3/Scene.cpp
+++ b/Periodic_3_triangulation_3/demo/Periodic_3_triangulation_3/Scene.cpp
@@ -1463,15 +1463,10 @@ void Scene::gl_draw_conflict() {
     std::vector<Cell_handle> cic;
     std::vector<Facet> boundary_facets;
     // Find the conflict region
-    Cell_handle c = p3dt.locate(moving_point);
-    Point _a(c->vertex(0)->point()),
-          _b(c->vertex(1)->point()),
-          _c(c->vertex(2)->point()),
-          _d(c->vertex(3)->point());
-    if(!( moving_point == _a
-       || moving_point == _b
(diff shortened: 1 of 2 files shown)


Periodic_3_triangulation_3/demo/Periodic_3_triangulation_3/Scene.cpp
```diff
@@ -1463,7 +1464,15 @@ void Scene::gl_draw_conflict() {
...
-    p3dt.find_conflicts(moving_point,c,std::back_inserter(boundary_facets),std::back_inserter(cic),CGAL::Emptyset_iterator());
+    Point _a(c->vertex(0)->point()),
+          _b(c->vertex(1)->point()),
+          _c(c->vertex(2)->point()),
+          _d(c->vertex(3)->point());
+    if(!( moving_point == _a
+       || moving_point == _b
+       || moving_point == _c
+       || moving_point == _d))
+      p3dt.find_conflicts(moving_point,c,std::back_inserter(boundary_facets),std::back_inserter(cic),CGAL::Emptyset_iterator());
```
lrineau: You are redoing part of the job of `locate()`, here.

Instead, use the [full locate with the `Locate_type`][1], and verify that the locate type is not `VERTEX`.

[1]: https://doc.cgal.org/4.13/Periodic_3_triangulation_3/classCGAL_1_1Periodic__3__triangulation__3.html#ae0e1ff13ddae388a580212f2645886ac

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-124.shtml

PR #3680, opened 2019-02-13
Fix warnings in testsuite

maxGimeno:
## Summary of Changes

Fix a comparison warning in CGAL_ipelets. And other warnings.

## Release Management

* Affected package(s):CGAL_ipelets


maxGimeno:
I added a fix for Intersections_2. 

> switch ( linepair.intersection_type()) {
    case Line_2_Line_2_pair<K>::NO_INTERSECTION:
        _result = NO_INTERSECTION;
        break;
    case Line_2_Line_2_pair<K>::POINT:
        _intersection_point = linepair.intersection_point();
        _result = (_seg->collinear_has_on(_intersection_point) )
                ? POINT : NO_INTERSECTION;
        break;
    case Line_2_Line_2_pair<K>::LINE:
        _result = SEGMENT;
        break;
    }

There can be only 3 possibilities for linepair.intersection_type(), and they are treated. Yet, on some platforms, it is considered that _result might not be initialized.

mglisse:
Some other places in CGAL have `default: CGAL_assume(false);` to let the compiler know that there aren't any other cases.

maxGimeno:
Thanks @mglisse

CGAL_ipelets/include/CGAL/CGAL_Ipelet_base_v7.h
```diff
@@ -66,20 +66,20 @@ namespace CGAL{
...
     void transform_selected_objects_(const IpeMatrix& tfm) const {
-      for (int i=0;i<get_IpePage()->count();++i)
+      for (int i=0;i<static_cast<int>(get_IpePage()->count());++i)
         if (get_IpePage()->select(i)!=ipe::ENotSelected)
           get_IpePage()->transform(i,tfm);      
     }
  
     void delete_selected_objects_() const {
-      for (unsigned i=get_IpePage()->count();i>0;--i)
+      for (unsigned i=static_cast<int>(get_IpePage()->count());i>0;--i)
```
mglisse: Really, even here?

lrineau: That cast is ugly, but:
  - the return type of `count()` changes from one version of Ipe to another,
  - that return type is not a nested typedef (like `size_type` in STL containers),
  - and `INT_MAX` is enough for the page count of any document.

mglisse: But why not be consistent and either cast to unsigned (the type of i) or make i an int (the type you are casting to)? Or wait a month and use `auto`...

lrineau: You are right!...

@maxGimeno Marc is right: in that line the loop variable `i` has type `unsigned int`, and that is not consistent with the cast. And anyway that loop is different from the other ones. Be careful.

maxGimeno: Oh yes I see. Sorry, I just used "replace all" without reading carefully... My bad.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-124.shtml

PR #3681, opened 2019-02-13
Polyhedron Demo: new batch of bug fixes

### LINKED ISSUE #3629: Batch of bugs/feature requests for the CGAL demo
### Heat Method bugs ###

- [x] Heat method: if you duplicate the item, it also seems to duplicate the selected vertices (thus, if you duplicate the item, and select new seeds, old seeds are also used in the coloring process)
- [x] Color range in "Display property" should immediately actualize when colors are changed
- [x] Heat method: reload and statistics do nothing on the "Heat Visualizer" item
- [x] There should be a simple way to go back from the heat visualization group to the source mesh ("reset colors" seems gone)
- [x] Heat Method being an option in Display Property is not very intuitive: it is not a property of the mesh and it should have its own widget (I get it that it's to re-use the code for coloring, but it's just not at the right place imo)

### Bugs ###
- [x] Surfaceic_Mesh_3 is all broken
- [x] Loading a .off with 2D coordinates instead of 3D usually crashes the demo (might have been fixed already)
- [x] Affine transformation cursor restoration seems messed up
- [x] "may occur for operations other tha<b>n</b> move_point"
- [x] Loading a selection without all the lines (e.g. forgetting to put an empty line for the edges) crashes the demo
- [x] Split/Join polyhedra is surprisingly slow
- [x] Color preview (ambiant/diffuse/specular) in "Lighting Setting" is not updated when changed. Also, maybe a value would make more sense than a degree of gray (?).
- [x] Get rid of "(as Surface Mesh)" in the orient operation
- [x] The result of offset meshing should be oriented outwards
- [x] Statistics > `#Null Length` should say `#Degenerate edges` (and `#degenerated faces` should be `#degenerate faces`)
- [x] Polyline stats should not show (n+1) vertices when it's just a closed polyline with the same vertex duplicated at the front and at the end
- [x] Small facet lasso selections freeze the demo (input mesh does not need to be large)

### Feature Requests ###
- [x] Interrupt on Offset Meshing
- [x] Create point sets from polyline items
- [x] "Apply affine transformation" should be in the same widget, not back within the "operations" menu

maxGimeno:
## Summary of Changes

More bug fixes in the demo.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #3629

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -2433,7 +2433,7 @@ QString MainWindow::get_item_stats()
 {
   //1st step : get all classnames of the selected items
   QList<QString> classnames;
-  Q_FOREACH(int id, getSelectedSceneItemIndices())
+  Q_FOREACH(int id, scene->selectionIndices())
   {
     QString classname = scene->item(id)->metaObject()->className();
     if(!classnames.contains(classname))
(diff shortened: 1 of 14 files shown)


maxGimeno:

PR #3683, opened 2019-02-13
CGAL: clean up CMakeLists.txt files in examples and tests

afabri:
## Summary of Changes

Remove unnecessary `include_directories()`.  
If anybody has more suggestions what should get cleaned up it could go into this PR.

For example removing `SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4244")`  from [this ](https://github.com/CGAL/cgal/blob/master/Spatial_searching/examples/Spatial_searching/CMakeLists.txt#L27) file, first of all as it does not work as @lrineau told me, and secondly because we shouldn't suppress warnings this way.

## Release Management

* Affected package(s): several

### DIFF
--- a/Surface_mesh_parameterization/examples/Surface_mesh_parameterization/CMakeLists.txt
+++ b/Surface_mesh_parameterization/examples/Surface_mesh_parameterization/CMakeLists.txt
@@ -6,9 +6,6 @@ project( Surface_mesh_parameterization_Examples )
 cmake_minimum_required(VERSION 3.1)
 
 
-# Include this package's headers first
-include_directories (BEFORE . include)
-
 # Find CGAL
 find_package(CGAL QUIET)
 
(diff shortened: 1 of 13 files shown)


Surface_mesh_parameterization/examples/Surface_mesh_parameterization/CMakeLists.txt
```diff
@@ -6,9 +6,6 @@ project( Surface_mesh_parameterization_Examples )
 cmake_minimum_required(VERSION 3.1)
 
 
-# Include this package's headers first
-include_directories (BEFORE . include)
```
lrineau: That should break the build. `./` and `./include/` are not by default.

lrineau: Well.. no, because `./include` does not exist, and `./` does not contain any header.

afabri: Only if there is such a directory named `include`, and if there is  something that gets included from "."

maxGimeno:
I put the TODO tag as you seem to expect more commits. Tell me if I'm wrong

maxGimeno:

PR #3692, opened 2019-02-19
Polyhedron demo: Fix crash in triangulation of non convex facets.

maxGimeno:
## Summary of Changes
Test if the two constraints vertices are different before inserting to avoid crash.

### DIFF
--- a/Polyhedron/demo/Polyhedron/triangulate_primitive.h
+++ b/Polyhedron/demo/Polyhedron/triangulate_primitive.h
@@ -141,8 +141,8 @@ private:
     }
     if(last_inserted == typename CDT::Vertex_handle())
       return false;
-    
-    cdt->insert_constraint(previous, first);
+    if(previous != first)
+      cdt->insert_constraint(previous, first);
     // sets mark is_external
     for(typename CDT::All_faces_iterator
         fit2 = cdt->all_faces_begin(),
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/triangulate_primitive.h
```diff
@@ -141,8 +141,8 @@ class FacetTriangulator
     }
     if(last_inserted == typename CDT::Vertex_handle())
       return false;
-    
-    cdt->insert_constraint(previous, first);
+    if(previous != first)
+      cdt->insert_constraint(previous, first);
```
lrineau: I am really surprised, that the code of the CDT does not handle that case by its own. Are you sure that this test fixes a bug?

maxGimeno: Well, the code of the CDT only uses an assertion. Before this test it crashed, after it doesn't, so I'm pretty sure, yes.

lrineau: Then that might be a bug of the code of CDT, because [the documentation does not document that precondition][1].

[1]: https://doc.cgal.org/4.13/Triangulation_2/classCGAL_1_1Constrained__Delaunay__triangulation__2.html#acd1f7edafecd48ea978e43c98c87172e

maxGimeno: Right here: https://github.com/CGAL/cgal/blob/0d4b7185bd7d73d21e96ea8f706c9a9d2ed0d269/Triangulation_2/include/CGAL/Constrained_triangulation_2.h#L296

maxGimeno:

PR #3693, opened 2019-02-19
STL_Extension: Use Boost Heap 

maxGimeno:
## Summary of Changes
Replace the use of Modifiable_priority_queue with Boost::heap::fibonacci_queue.
EDIT: The reason is that  Modifiable_priority_queue is an undocumented header that we want to avoid now that Boost provides a reliable replacement for it. 

## Release Management

* Affected package(s):AABB_tree, Polyline_simplifiaction_2, STL_Extension, Surface_mesh_simplification


afabri:
Can you explain why you do this change?  Deprecation of  `Modifiable_..`,  a fix of a bug, a convincing benchmark,....

maxGimeno:
@afabri see the Edit in the description

afabri:
> 
> 
> @afabri see the Edit in the description

Can you please make a performance test, for example for surface mesh simplication.

maxGimeno:
I used the test_edge_collapse_Polyhedron_3 with a 340 000 vertices (1919448 edges) chinese dragon, here are the results:
Before PR : 41.71s
After PR: 36.85s

afabri:
> 
> 
> I used the test_edge_collapse_Polyhedron_3 with a 340 000 vertices (1919448 edges) chinese dragon, here are the results:
> Before PR : 41.71s
> After PR: 36.85s

That's good news!!!

sloriot:
Quick question: Instead of modifying the code of various packages, would it be possible to update only the implementation of `Modifiable_priority_queue` using the boost documented heap?

maxGimeno:
I don't know. It wouldn't be trivial, that's all I can say without trying.

sloriot:
With an external view, that's a bit surprising.

afabri:
I guess what Sebastien want's to say that it is probably easy to use the fibonacci_heap as an implementation detail which can be de/activated with a `#define`, right as it is done now  with a [boost::relaxed_heap](https://github.com/CGAL/cgal/blob/master/STL_Extension/include/CGAL/Modifiable_priority_queue.h#L83)

maxGimeno:
I looked back at the diff, it seems easier than I thought. Theproblem is we need to embark the map and maintain it, where it is not always needed, so there will be a loss of performance in Edge_collapse.

Polyline_simplification_2/test/Polyline_simplification_2/simplify_polygon_test.cpp
```diff
@@ -46,7 +46,7 @@ int main(int argc, char* argv[])
 {
 
   for(int i= 1;i < argc; i++){
-    test<Cost1>(argv[i]);
```
sloriot: is this wanted?

maxGimeno: Probably not :/

sloriot:
My guess is that the use in polyline simplification was not optimal and those type of heaps are bad for contains query. In simplification, the handle are already store in the algorithm. So maybe the implementation of polyline simplification should be updated.

maxGimeno:
I restarted everything, only replacing the relaxed_heap by a fibonacci_heap, which is documented. Before I start a series of benchmarks between the different heaps in Boost.Heap, I'd like to expose a problem to @CGAL/geometryfactory :

As exposed by @afabri , the boost specific code is only activated if the flag -DCGAL_SURFACE_MESH_SIMPLIFICATION_USE_RELAXED_HEAP is passed to the CXX flags. But it only works for Surface_mesh_simplification, and as it is done in Modifiable_priority_queue.h, the flag will also erase the definition of the `mutable_queue_with_remove`. **The problem is that `mutable_queue_with_remove` is used in two other packages, that won't compile with the flag.**

My question is: Do I let it as it is, do I split the definitions in Modifiable_priority_queue.h so that `mutable_queue_with_remove` is always usable even if not used in SMS, or do I adapt all the code to switch to Boost.Heap if the flag is passed ?

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -273,16 +273,20 @@ operator()()
       boost::optional<FT> dist = cost(pct, u);
       if(! dist){
         // cost is undefined
-        if( mpq->contains(u) ){
-          mpq->erase(u);
+        if( h.find(u.base()->id) != h.end()){
+          mpq->erase(u, h[u.base()->id]);
```
sloriot: avoid the double find by storing the result of find

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -273,16 +273,20 @@ operator()()
...
+          h.erase(u.base()->id);
         }
       } else {
-        (*u)->set_cost(*dist);
-        if(mpq->contains(u)){
-          mpq->update(u, true);
+        if(h.find(u.base()->id) != h.end()){
```
sloriot: same remark here

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -273,16 +273,20 @@ operator()()
...
+          typename MPQ::handle ex_h = h[u.base()->id];
```
sloriot: and here

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -291,16 +295,20 @@ operator()()
       boost::optional<FT> dist = cost(pct, w);
       if(! dist){
         // cost is undefined
-        if( mpq->contains(w) ){
-          mpq->erase(w);
+        if(h.find(w.base()->id) != h.end()){
```
sloriot: and here

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -291,16 +295,20 @@ operator()()
...
+          mpq->erase(w, h[w.base()->id]);
+          h.erase(w.base()->id);
         }
       } else {
-        (*w)->set_cost(*dist);
-        if(mpq->contains(w)){
-          mpq->update(w, true);
+        if(h.find(w.base()->id) != h.end()){
```
sloriot: and here

STL_Extension/include/CGAL/Modifiable_priority_queue.h
```diff
@@ -80,35 +44,50 @@ class Modifiable_priority_queue
   typedef Compare_     Compare;
   typedef ID_          ID ;
   
-  #ifdef CGAL_SURFACE_MESH_SIMPLIFICATION_USE_RELAXED_HEAP
+#ifdef CGAL_SURFACE_MESH_SIMPLIFICATION_USE_RELAXED_HEAP
```
sloriot: is the relaxed heap now officially documented in boost? I wonder whether we should add as template parameter the possibility to have the relaxed heap. If the performance gain is noticeable it might be interesting for some application (those targeting an approximation error for example)

maxGimeno: No, it is still in /boost/pending/pending

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -271,14 +271,16 @@ operator()()
...
         // cost is undefined
-        if( h.find(u.base()->id) != h.end()){
+        if( find_result_u != h.end()){
           mpq->erase(u, h[u.base()->id]);
           h.erase(u.base()->id);
         }
       } else {
-        if(h.find(u.base()->id) != h.end()){
+        if(find_result_u != h.end()){
           typename MPQ::handle ex_h = h[u.base()->id];
```
afabri: No need for `h[]`.

maxGimeno: Can you elaborate ?

maxGimeno:
Conflicts.

@sloriot I'm still waiting for your review on this

lrineau:
Travis is fully red. It probably does not compile either on your machine.
```
/home/travis/build/CGAL/cgal/STL_Extension/include/CGAL/iterator.h:1383:3: error: constructor cannot be redeclared
  Dispatch_output_iterator(const Dispatch_output_iterator&)=default;
  ^
/home/travis/build/CGAL/cgal/STL_Extension/include/CGAL/iterator.h:1379:3: note: previous definition is here
  Dispatch_output_iterator(const Dispatch_output_iterator&)=default;
  ^
```

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -272,36 +273,48 @@ operator()()
...
       boost::optional<FT> dist = cost(pct, w);
+      typename boost::unordered_map<int,typename MPQ::handle>::iterator find_result_w = 
+          h.find(w.base()->id);
       if(! dist){
         // cost is undefined
-        if( mpq->contains(w) ){
-          mpq->erase(w);
+        if(find_result_w != h.end()){
+          mpq->erase(w, find_result_w->second);
+          h.erase(w.base()->id);
```
sloriot: I'm pretty sure you can use the iterator for doing the removal...

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -103,7 +103,7 @@ class Polyline_simplification_2
   typedef CGAL::Modifiable_priority_queue<Vertices_in_constraint_iterator,Compare_cost,Id_map> MPQ ;
   
   MPQ* mpq;
-
+  boost::unordered_map<int,typename MPQ::handle> h;
```
sloriot: now std::unordered_map can be used

sloriot:
The PR looks good now. However I still have some questions:

- I'd like to see the result of benchmarks on the modified packages (Mesh_3, Polyline simplification, and SMS).
- I see that in Mesh_3 and Polyline simplification you needed to introduce a hash map to map an item to its handle. AFAIS, it is not needed (there are alternative functions without the handles) but I can imagine that there might be a noticeable speed up.
- If using the hash map does worth it, I wonder if we couldn't have an API where it is handled internally. For example, we could use a key property map that is responsible for storing the handles, and we would have function overloads taking the pmap as parameter (I don't think the pmap needs to be a data member). For example, you would have `pqueue.update(pv, pmap);` etc... I guess a benchmark with and without the hash map would be needed to see if it is a good choice.

maxGimeno:
Benchmark code is [here](https://gist.github.com/maxGimeno/34469bf83c7234832bbc5c11ddcae20b)
My results on my machine : 
Branch:
=== Polyline Simplification ===
0.0041625 seconds.
=== Mesh_3 ===
0.0180625 seconds.
=== SMS ===
0.0980775 seconds.

Master:
=== Polyline Simplification ===
0.00344075 seconds.
=== Mesh_3 ===
0.0182057 seconds.
=== SMS ===
0.0875535 seconds.

afabri:
What are the total times of the executables?  How often are they executed?

maxGimeno:
You can checkout out the  code of the test to find that out. 
They are executed 5 times each and the printed time is the average.

maxGimeno:
> * (there are alternative functions without the handles)

If they exist, they are not documented :/

maxGimeno:
The map management is too costly, even with abseil's flat_hash_map.

maxGimeno:
> The map management is too costly, even with abseil's flat_hash_map.

afabri:
@maxGimeno can you check if the new function [`std::map<>::extract()` ](https://en.cppreference.com/w/cpp/container/map/extract) is a possible replacement.

sloriot:
@afabri meant using `std::map` that has a new function in c++17 that could be used to update the priority. (Maybe you can keep the wrapper to simplify the testing?)

maxGimeno:
so IIUC, the task is to design a heap based on a c++17 `std::map<cost, type>` to replace the `mutable_queue_with_remove`, use this custom heap internally  to the Modifiable_priority_queue and finally retry the bench ?

sloriot:
Something like that, maybe hidden behind a `#ifdef CGAL_USE_FOOBAR`

maxGimeno:
[Here](https://github.com/maxGimeno/cgal/blob/Use_cxx17_set/STL_Extension/include/CGAL/Modifiable_priority_queue.h) I tried an implementation based on a cxx17 set, using the function `extract()` for `update()`, and it is still slower than the master version.

sloriot:
Replaced by #6155

Issue #3696, opened 2019-02-20
Read_ply_points_with_properties() with Exact Kernel


danston:
## Issue Details

When I try to read points with normals from .ply using CGAL::Exact_predicates_exact_constructions_kernel and the function CGAL::read_ply_points_with_properties(), I get an assertion (reader != NULL). Changing exact kernel to inexact produces the correct result.

### example:
```
#include <fstream>
#include <utility>

#include <CGAL/property_map.h>
#include <CGAL/Simple_cartesian.h>
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
#include <CGAL/IO/read_ply_points.h>

using Kernel = CGAL::Exact_predicates_exact_constructions_kernel;

using Point_3 = typename Kernel::Point_3;
using Vector_3 = typename Kernel::Vector_3;

using Point_with_normal = std::pair<Point_3, Vector_3>;

using Point_map = CGAL::First_of_pair_property_map<Point_with_normal>;
using Normal_map = CGAL::Second_of_pair_property_map<Point_with_normal>;

int main(int argc, char **argv) {

  const auto file_name = argv[1];
  std::vector<Point_with_normal> data;
  std::ifstream in(file_name);

  if (
    !in ||
    !CGAL::read_ply_points_with_properties(in,
    std::back_inserter(data),
    CGAL::make_ply_point_reader(Point_map()),
    CGAL::make_ply_normal_reader(Normal_map()))) {

    std::cerr << "Error: cannot read file " << file_name << std::endl;
    return EXIT_FAILURE;
  }

  const Point_3& p = data[0].first;
  const Vector_3& n = data[0].second;
    
  std::cout.precision(15);
  std::cout << "Point (" << p << ") with normal (" << n << ")" << std::endl;

  return EXIT_SUCCESS;
}
```

### cmake:
```
project(test)

cmake_minimum_required(VERSION 2.8.10)
set(CMAKE_CXX_STANDARD 11)

find_package(CGAL QUIET COMPONENTS Core)
if(CGAL_FOUND)

  include(${CGAL_USE_FILE})
  include(CGAL_CreateSingleSourceCGALProgram)

  create_single_source_cgal_program(
    "test.cpp")

else()
  message(WARNING 
  "This program requires the CGAL library, and will not be compiled.")
endif()
```

### ply:
```
ply
format ascii 1.0
comment Generated by the CGAL library
[...]

PR #3636, opened 2019-01-28
[Small Feature] Surface Mesh PLY IO

sgiraudot:
## Rationale

The PLY format has 3 main advantages:
- it can embed any type of item that can be described by a sequence of numbers, and is in particular widely used by geometry processing software to store point sets and meshes
- it can read/write binary files, which is more efficient than parsing ASCII
- it can embed any property the user wants, which makes it very powerful especially when plugged-in to the `Surface_mesh` dynamic property maps.

The `Point_set_3` class (which is very similar to `Surface_mesh`) already uses this format to save and load point sets with properties. This small feature adds similar IO functions for `Surface_mesh`.

## Summary of API changes

- Methods `CGAL::read_ply()` and `CGAL::write_ply()` are added to the `Surface_mesh` package, following the same terminology as the existing `CGAL::read_off()` and `CGAL::write_off()` functions.
- An extra parameter is provided for the PLY IO functions of `Point_set_3` to load/save the comments embedded in the PLY header (the same parameter is put in the 2 new `Surface_mesh` functions described above).

## License and copyright ownership

(no change)

## CHANGES.md

TODO

## Status

-   Developed and locally tested (GNU/Linux)
-   (Empty) small feature page: [Surface Mesh PLY IO](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Surface_Mesh_PLY_IO)

## Submission

* [`CGAL::read_ply()`](https://cgal.geometryfactory.com/~sgiraudot/Surface_mesh-PLY/Surface_mesh/group__PkgSurface__mesh.html#ga42f6ad486ddab74e13d3dc53f511c343): new surface mesh reading function
* [`CGAL::write_ply()`](https://cgal.geometryfactory.com/~sgiraudot/Surface_mesh-PLY/Surface_mesh/group__PkgSurface__mesh.html#ga77bbb79d449c981895eedb6c3c23bd14): new surface mesh writing function
* [`CGAL::read_ply_point_set()`](https://cgal.geometryfactory.com/~sgiraudot/Surface_mesh-PLY/Point_set_3/group__PkgPointSet3IO.html#ga0d80f23b142818bba05869ee1bfc8c41): new documentation for point set reading function
* [`CGAL::write_ply_point_wet()`](https://cgal.geometryfactory.com/~sgiraudot/Surface_mesh-PLY/Point_set_3/group__PkgPointSet3IO.html#ga4dafa5ad1b9a33b84ad3467510791875): new documentation for pont set writing function

PR #3703, opened 2019-02-22
Point Set Processing: Fix PLY FT type detection

sgiraudot:
## Summary of Changes

In order to avoid reading/writing point coordinates that are `float` as `double` when using the PLY reader with `CGAL::Simple_cartesian<float>`, I introduced an automatic detection of the FT type based on the value type of the point map. This detection fails if the kernel is exact, see issue https://github.com/CGAL/cgal/issues/3696.

This fixes it by declaring the type to read/write (from a PLY point of view) as double by default, _except_ when using the float kernel.

(I'll also have to fix https://github.com/CGAL/cgal/pull/3636 similarly.)

## Release Management

* Affected releases: from 4.13 (the automatic FT detection was not in 4.12)
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/3696

### DIFF
--- a/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
+++ b/Point_set_processing_3/include/CGAL/IO/read_ply_points.h
@@ -92,10 +92,19 @@ namespace CGAL {
   };
 
   /// \cond SKIP_IN_MANUAL
+
+  // Use a double property for all kernels...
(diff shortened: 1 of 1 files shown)


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-135.shtml

PR #3704, opened 2019-02-25
Update STL_writer.h

vbchrist:
Current implementation produces an unreadable format due to incorrect number formatting for ASCII output.
Included sign-mantissa-"e"-sign-exponent format on line 88 by setting notation and precision.
Additionally added some formatting to output for readability and to match common STL file formats.
Works with most STL readers.

## Summary of Changes

_Describe what your pull request changes to CGAL (this can be skipped if it solves an issue already in the tracker or if it is a Feature or Small Feature submitted to the CGAL Wiki)._

## Release Management

* Affected package(s):
* Link to compiled documentation (obligatory for small feature) [link](https://mydoc/Manual/Pkg)

### DIFF
--- a/Polyhedron_IO/include/CGAL/IO/STL_writer.h
+++ b/Polyhedron_IO/include/CGAL/IO/STL_writer.h
@@ -85,11 +85,12 @@ write_STL(const TriangleMesh& tm, std::ostream& out)
 
       Vector_3 n = collinear(p,q,r) ? Vector_3(1,0,0):
                                       unit_normal(p,q,r);
-      out << "facet normal " << n << "\nouter loop\n";
-      out << "vertex " << p << "\n";
-      out << "vertex " << q << "\n";
-      out << "vertex " << r << "\n";
-      out << "endloop\nendfacet\n";
+      out << std::scientific << std::setprecision(6);
+      out << "facet normal " << n << "\n    outer loop\n";
+      out << "        vertex " << p << "\n";
+      out << "        vertex " << q << "\n";
+      out << "        vertex " << r << "\n";
+      out << "    endloop\nendfacet\n";
     }
     out << "endsolid\n";
   }
--- a/Polyhedron_IO/include/CGAL/IO/STL_writer.h
+++ b/Polyhedron_IO/include/CGAL/IO/STL_writer.h
@@ -75,7 +75,8 @@ write_STL(const TriangleMesh& tm, std::ostream& out)
   }
   else
   {
-    out << "solid\n";
+    out << "solid\n";    
+    out << std::scientific << std::setprecision(17);
     BOOST_FOREACH(face_descriptor f, faces(tm))
     {
       halfedge_descriptor h = halfedge(f, tm);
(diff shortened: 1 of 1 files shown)


Polyhedron_IO/include/CGAL/IO/STL_writer.h
```diff
@@ -85,11 +85,12 @@ write_STL(const TriangleMesh& tm, std::ostream& out)
 
       Vector_3 n = collinear(p,q,r) ? Vector_3(1,0,0):
                                       unit_normal(p,q,r);
-      out << "facet normal " << n << "\nouter loop\n";
-      out << "vertex " << p << "\n";
-      out << "vertex " << q << "\n";
-      out << "vertex " << r << "\n";
-      out << "endloop\nendfacet\n";
+      out << std::scientific << std::setprecision(6);
```
lrineau: I disagree with `std::setprecision(6)`. Would you accept to change it into `std::setprecision(17)`, so that the double precision (with a 53 bits mantissa) is preserved during the input/ouputs?

vbchrist: Agree, wasn't sure if there was a character limit on paraview. Adjusting to 

```suggestion
      out << std::scientific << std::setprecision(17);
```

should work.

afabri:
Two remarks:
- It has probably no big overhead, but I would suggest to set the precision once outside the loop.
- @lrineau wasn't there a recent commit where you recommended not to change precision at all?  After all the stream is an argument of the function, so the user should set the precision.  Now the function has a side effect, that should be documented.

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Feb 28, 2019, 8:26 AM GMT+1](https://github.com/CGAL/cgal/pull/3704#issuecomment-468166763 "2019-02-28T07:26:11Z - Replied by Github Reply Comments"):
> - @lrineau wasn't there a recent commit where you recommended not to change precision at all?  After all the stream is an argument of the function, so the user should set the precision.  Now the function has a side effect, that should be documented.

I agree. The function should store the current precision before changing it, and then restore it at the end. And same for 

@vbchrist Actually, I am really surprised that the patch is needed at all. Which STL reader is only able to recognize floating point numbers when they are written in `scientific` format?

vbchrist:
I rechecked my work...  

The error I was receiving was NOT due to the number formatting, but due to the initial and final lines of the .STL file which do not include a solidname (first line of STL file after the key word solid).  I had added a default name in my offline code.  I checked this in SolidWorks 2019, the latest release of paraview, and MS Paint 3D.  Paint 3D is the only reader which fails.

There are other software packages I'm aware of that use solidname (OpenFOAM), but may not be req.

It may still be useful to output the sign-mantissa-"e"-sign-exponent format, however, I was wrong and these readers DO NOT seem to require it.

On the other hand, perhaps adding in a place holder for the solidname would be useful to ensure compatibility.

afabri:
I think that this PR shows nevertheless that there is an issue with the lack of documentation and the fact that even [examples](https://doc.cgal.org/latest/Point_set_3/Point_set_3_2point_set_read_xyz_8cpp-example.html#a7) we give, where we do not set the precision.

PR #3705, opened 2019-02-26
add missing function required by the concept

sloriot:
That was missing to make `write_off()` working with a `Face_filtered_graph`


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-135.shtml

PR #3718, opened 2019-02-28
Allow to have different SOVERSION for each CGAL libraries

lrineau:
## Summary of Changes

In CGAL-4.14, I would like to have `libCGAL.so.13` (to keep
compatibility), but `libCGAL_Image.so.14` (to break compatibility).

This PR allows to have different `SOVERSION` and `SONAME` for each libraries. And it also change the SOVERSION and SONAME for `liCGAL_ImageIO`, while keeping the versions same as for CGAL-4.13 for all the other CGAL libraries.

@maxGimeno I need this PR before CGAL-4.14-beta1 is published, so that one can have CGAL-4.14 in recent Linux distributions as soon as possible.

## Release Management


maxGimeno:
Successfully tested by https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-136.shtml

PR #3719, opened 2019-02-28
Invalid (X)HTML sequence

albert-github:
In packages.html we get:
`Element style is not declared in p list of possible children`
as the `style` tag is only allowed in the `head` part of a file.

The specific changes are only needed for this one file, so it has been implemented using the post processing facility of te CGAL documentation generation.

### DIFF
--- a/Documentation/doc/Documentation/packages.txt
+++ b/Documentation/doc/Documentation/packages.txt
@@ -2,15 +2,6 @@
 
 \page packages Package Overview
 
-\htmlonly
-<style type="text/css">
-h1 {
-	font-size: 150%;
-    background-color: #EEEDF2;
-}
-</style>
-\endhtmlonly
-
 \section PartAlgebraicFoundations Arithmetic and Algebra
 
 \package_listing{Algebraic_foundations}
--- a/Documentation/doc/Documentation/packages.txt
+++ b/Documentation/doc/Documentation/packages.txt
@@ -2,7 +2,7 @@
 
 \page packages Package Overview
 
-\cgalPackagesSection{PartAlgebraicFoundations,Arithmetic and Algebra}
+\cgalPackageSection{PartAlgebraicFoundations,Arithmetic and Algebra}
 
 \package_listing{Algebraic_foundations}
 \package_listing{Number_types}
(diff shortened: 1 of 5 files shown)


Documentation/doc/Documentation/packages.txt
```diff
@@ -2,15 +2,6 @@
...
 
-\htmlonly
-<style type="text/css">
-h1 {
-	font-size: 150%;
-    background-color: #EEEDF2;
-}
-</style>
-\endhtmlonly
-
```
lrineau: I think it would be easier to have a big `<div style="…">` containing all the content of the file, to avoid having to hard-code styling in the post-processing.

albert-github: I'll give it a try.

albert-github: When having `<div style="...` the result would be valid for the entire file (so in this case complete gray background).
A possibility would be to wrap each `<h1>`  into a `<div style="...`, this is quite feasible (with an ALIAS) but has at the moment 1 small problem:
 `<div` would set the font size top 150% and the `<h1>` would set it back to 180%
when it is acceptable to have 180% for these headings (I think this would not be a problem) I can implement it.

The example with the 180%:
example180

The example with the 150%:
example150

albert-github:
Update: an updated implementation dropping the post processing step in favor of a `<div style=..` implementation

sloriot:
I have renamed the macro (removed the s) and updated the two other resources files and checked the output (1.8.4 and 1.8.13).

albert-github:
@sloriot Thanks. No objections, looks OK. I only did the 1.8.14 version as I don't know which parts will and will not land in other versions.

sloriot:
I realized that `[block]` is not supported by doxygen 1.8.4 so some other fixes you did will break but I guess we can simply drop the support for that version.

albert-github:
@sloriot , @lrineau 
I think too that still trying to support 1.8.4 (Release May,19 2013) is not necessary anymore.
Best to remove the `[block]` from the 1.8.4 version of the Doxyfile till it is decided whether or not 1.8.4 still has to be supported or not..

sloriot:
I was not talking of this PR (I already fixed the macro) but for example #3724. But as I said support of 1.8.4 can be dropped.

lrineau:
@sloriot In which release do we want this PR?

sloriot:
It does not worth back-porting IMO, let's keep master.

maxGimeno:

PR #3723, opened 2019-03-01
Documentation italic in italics

albert-github:
When having italics in italics and (X)HTML warning is given about  Opening and ending tag mismatch.
The `<EM>` as well as `<I>` result in emphasized text and would map in doxygen both to `<i>` but will confuse doxygen. output

The problematic code can be found in `Ridges_3/doc/Ridges_3/Ridges_3.txt`

The original text looks like:
italics

The updated text:
new_version_italics


maxGimeno:

PR #3724, opened 2019-03-01
Inconsistent tags for (X)HTML output (Manual)

albert-github:
The xhtml lint checker gave:
`parser error : Opening and ending tag mismatch: div line 375 and p`

Solved by making the `\htmlonly` into `\htmlonly[block]` to prevent invalid HTML

### DIFF
--- a/Documentation/doc/Documentation/main.txt
+++ b/Documentation/doc/Documentation/main.txt
@@ -47,7 +47,7 @@ For releases X.Y, with 3.1 <= X.Y <= 4.1 visit
 
 
 
-\htmlonly
+\htmlonly[block]
 <div style="display:none">
 \endhtmlonly
 
@@ -56,7 +56,7 @@ For releases X.Y, with 3.1 <= X.Y <= 4.1 visit
 \subpage packages
 \subpage dev_manual
 
-\htmlonly
+\htmlonly[block]
 </div>
 \endhtmlonly
 


albert-github:
As mentioned with #3719 `[block]` is not supported for doxygen 1.8.4 a solution for this is to define an `ALIASES` that has in 1.8.4 version just `\htmlonly` and in the 1.8.14 version `\htmlonly[block]` but as said in #3719 this is dependent whether or not 1.8.4 still has to be supported or not.

lrineau:
We use Doxygen 1.8.13 since 4.12, and we do not support previous release branches. So for our testsuite and releases, we can drop the support for Doxygen 1.8.4.

Actually, the question is more which Doxygen versions we want to support, for CGAL users and developers:
  - [all supported Fedora version ship Doxygen 1.8.14][fedora],
  - [Debian stable ships Doxygen 1.8.13, but Debian oldstable uses 1.8.8][debian],
  - [Ubuntu 18.04 LTS ships 1.8.13, but 16.04 LTS ships 1.8.11][ubuntu].

[fedora]: https://apps.fedoraproject.org/packages/doxygen
[ubuntu]: https://packages.ubuntu.com/search?keywords=doxygen
[debian]: https://packages.debian.org/search?keywords=doxygen

That would be convenient if Doxygen 1.8.8 or 1.8.11 could be supported. At least if one can compile the documentation without errors, but maybe with formatting glitches.

Maybe the resources for 1.8.13 already works correctly with 1.8.11 or 1.8.8. That would be the easiest way to support those versions.

albert-github:
A wild though: 
Wouldn't it be better to support only version 1.8.14 (or 1.8.15) and suggest to the user that in case he uses an older version it is at his own risk (this is more or less the same as @lrineau suggests, at least in my opinion, but even goes one step further) .
Updating doxygen on a Linux system is not a big deal for a user.

maxGimeno:

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Mar 5, 2019, 3:59 PM GMT+1](https://github.com/CGAL/cgal/pull/3724#issuecomment-469711723 "2019-03-05T14:59:05Z - Replied by Github Reply Comments"):
> Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-139.shtml

I am surprised. That PR was supposed to break Doxygen 1.8.4.

lrineau:
Well...

https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-4.14-Ic-139/output1/Manual/index.html

screenshot_20190305_161011

The "error" is just a buggy `[block]` inserted in that page. Let's say that is so minor that we can merge this PR. Anyway, we want to drop support for older versions of Doxygen.

PR #3726, opened 2019-03-04
Classification: Fix bbox precision

### LINKED ISSUE #3720: classify.h  Precision lost in float value
## Issue Details
Unexpected behavior when I run the example(example_ethz_random_forest.cpp)
## Source Code
The program exception occurs in the classify_with_graphcut function of classify.h. It divides the data according to min_number_of_subdivisions and and recalculates the bounding box of the block using the float type. The float type has a loss of precision problem, which causes the new block bounding box (bboxes) to be smaller than the bounding box of the original data(bbox). The end result is that the size of the indices (L598) is smaller than the size of the input, which causes some points to be missed. So I recommend using the double type in bounding box calculations.
## Environment
* Operating system (Windows10 64 bits):
* Compiler: visual studio 2015
* Specific flags used (if any): WITH_TBB
* CGAL version: 4.13
* Boost version: 1.60

sgiraudot:
## Summary of Changes

To subdivide an input set, a set of bounding boxes is used. Because the coordinates are converted to floats, some points may end up not being assigned to any subdivision (with the precision loss, they can end up outside of all bboxes).

This fixes it and also adds an assertion to check that all points have been assigned to a subdivision.

## Release Management

* Issue(s) solved (if any): fix #3720

### DIFF
--- a/Classification/include/CGAL/Classification/classify.h
+++ b/Classification/include/CGAL/Classification/classify.h
@@ -480,11 +480,11 @@ namespace internal {
       (boost::make_transform_iterator (input.begin(), CGAL::Property_map_to_unary_function<ItemMap>(item_map)),
        boost::make_transform_iterator (input.end(), CGAL::Property_map_to_unary_function<ItemMap>(item_map)));
 
-    float Dx = float(bbox.xmax() - bbox.xmin());
-    float Dy = float(bbox.ymax() - bbox.ymin());
-    float A = Dx * Dy;
-    float a = A / min_number_of_subdivisions;
-    float l = std::sqrt(a);
+    double Dx = double(bbox.xmax() - bbox.xmin());
+    double Dy = double(bbox.ymax() - bbox.ymin());
+    double A = Dx * Dy;
+    double a = A / min_number_of_subdivisions;
+    double l = std::sqrt(a);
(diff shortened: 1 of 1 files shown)


Classification/include/CGAL/Classification/classify.h
```diff
@@ -495,11 +495,11 @@ namespace internal {
...
       {
         bboxes.push_back
-          (CGAL::Bbox_3 (bbox.xmin() + Dx * (x / float(nb_x)),
-                         bbox.ymin() + Dy * (y / float(nb_y)),
+          (CGAL::Bbox_3 (bbox.xmin() + Dx * (x / double(nb_x)),
+                         bbox.ymin() + Dy * (y / double(nb_y)),
                          bbox.zmin(),
-                         bbox.xmin() + Dx * ((x+1) / float(nb_x)),
-                         bbox.ymin() + Dy * ((y+1) / float(nb_y)),
+                         bbox.xmin() + Dx * ((x+1) / double(nb_x)),
```
sloriot: I think you might not be able to achieve xmax/ymax due to rounding and you'll have an issue if the bound is tight. You should have a special case I guess.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-140.shtml

PR #3727, opened 2019-03-04
Polyhedron demo: Fix snapshot

### LINKED ISSUE #3710: Polyhedron demo: bugs in `saveSnapshot()`
## Issue Details

I have tried the command `File`->`Save snapshot` to get a snapshot with a bigger resolution than my screen, and a lot got wrong.

The object I wanted to draw: a c3t3, with triangles and no tets.

screenshot_20190226_161406

I have used Save snapshot:

screenshot_20190226_161458

Then I changed the width to 2000, and the oversampling to 2:

screenshot_20190226_161544

The ratio seems to be preserved.

But the image I got was:
test

Two issues:

1. Wrong ratio et wrong frustrum.

2. The anti-aliasing is wrong. In the following screenshot, I have turned the background color of the image viewer to yellow, and then zoomed a lot on the image. See that the rendering of the anti-aliased edges is wrong: it is blended with the background color, instead of being blended with the color of the triangles that are behind.

screenshot_20190226_162447

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux Fedora 29
* Specific flags used (if any): `-Og -g`
* CGAL version: `master` (fe5c2d0bb8a)

maxGimeno:
## Summary of Changes
hanges the order of blending so that the edges are blended with the triangles and not the background.
## Release Management

* Issue(s) solved (if any): fix #3710

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.cpp
+++ b/Polyhedron/demo/Polyhedron/Viewer.cpp
@@ -689,14 +689,16 @@ void Viewer_impl::draw_aux(bool with_names, Viewer* viewer)
     viewer->glEnable(GL_BLEND);
     viewer->glEnable(GL_LINE_SMOOTH);
     viewer->glHint(GL_LINE_SMOOTH_HINT, GL_NICEST);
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #3729, opened 2019-03-04
T2,T3,Mesh_2: backticks, lowercase, capitalize

afabri:
## Summary of Changes

tiny fixes.

## Release Management

* Affected package(s): Triangulation_2, Triangulation_3, Mesh_2

### DIFF
--- a/Mesh_2/doc/Mesh_2/Concepts/DelaunayMeshFaceBase_2.h
+++ b/Mesh_2/doc/Mesh_2/Concepts/DelaunayMeshFaceBase_2.h
@@ -32,7 +32,7 @@ typedef std::pair<DelaunayMeshFaceBase_2, int> Edge;
 /// @{
 
 /*!
-returns true if this face is in the domain to be refined.
+returns `true` if this face is in the domain to be refined.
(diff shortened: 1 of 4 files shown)


albert-github:
Wouldn't it be better to use instead of the text `returns`  the doxygen command `\returns` or use the `\retval` command (might apply to a number of other functions as well, also outside Mesh_3).

sloriot:
if we use \return we will always have the long version of the doc of a function and somtimes it is just useless to have it.

albert-github:
Indeed, for small descriptions this will indeed be a problem and the case(s) I saw are indeed just a "lookup" function where the return value us just the (brief) description of the function.

maxGimeno:

afabri:
> 
> 
> Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-143.shtml

I don't agree, as I made commits after that.

maxGimeno:
@afabri I think you posted in the wrong PR.There was no activity since I flagged it as Under Testing 4 days ago.

PR #3731, opened 2019-03-04
IO: Set precision of the ostream when writing a point set or face graph

afabri:
## Summary of Changes

Examples which write a point set or mesh to a stream should set the precision.
Additionally I added a \note in the manual.

## Release Management

* Affected package(s): Point_set_3, Point_set_processing, Surface_mesh_simplification.


maxGimeno:

PR #3741, opened 2019-03-07
PMP debug code - use the new API of `PMP::remove_degenerate_faces()`

janetournois:
If there is no degenerate faces to deal with, i.e. if the function has nothing to do, it now returns `true`, and not `0`
This change was introduced by commit e4ad5d96a7afd1f2cca3fbc7cb2661a1c1be7659

## Summary of Changes

we now use the new API in the debug code of `isotropic_remeshing()`

## Release Management

* Affected package(s): PMP


maxGimeno:

PR #3759, opened 2019-03-11
Testsuite : Use ctest

maxGimeno:
## Summary of Changes

Add a script to launch a testsuite with ctest. 
This script has an option for cross-compilation, that uses the new docker image Debian-stable-cross-compilation-for-arm.

It also makes the needed modifications to scripts and infrastructure to make ctest work.

TODO: Update the doc with the validated toolchain and init files.
## Release Management

* Affected package(s): Installation, Arrangements_on_surface_2


Installation/cmake/modules/CGAL_add_test.cmake
```diff
@@ -43,6 +43,11 @@ if(ANDROID)
   set(ANDROID_DIR_PREFIX /data/local/tmp/)
   find_program(adb_executable adb)
 endif()
+if(SSH)
```
lrineau: I think that option `SSH` should be renamed to something else.

Installation/cmake/modules/CGAL_add_test.cmake
```diff
@@ -43,6 +43,11 @@ if(ANDROID)
...
+  set(SSH_DIR_PREFIX /home/pi/CGAL/)
```
lrineau: I think that should be a cache variable, with a documentation (and a better name). Maybe unify with `ANDROID_DIR_PREFIX`, and rename it to `CGAL_REMOVE_TEST_DIR_PREFIX` (and the default value would be different on Android and rPI).

Scripts/developer_scripts/global_run_testsuite.sh
```diff
@@ -0,0 +1,265 @@
+#!/bin/bash
```
lrineau: Why the name `global_run_testsuite.sh`?

Installation/cmake/modules/CGAL_add_test.cmake
maxGimeno: Like EXECUTE_REMOTELY ?

Scripts/developer_scripts/global_run_testsuite.sh
maxGimeno: It is just the first thing that crossed my mind when I was looking for a a name for a the script that runs the testsuite remotely and locally.

Installation/cmake/modules/CGAL_add_test.cmake
lrineau: I think the name should start with `CGAL_`... maybe:

    CGAL_RUN_TESTS_THROUGH_SSH

Scripts/developer_scripts/global_run_testsuite.sh
maxGimeno: It could be autotest_cgal_with_cmake, wichi sounds like a more logical choice when compared with run_testsuite_with_cmake.

Installation/cmake/modules/CGAL_add_test.cmake
maxGimeno: Ok, I like it, it makes it clear what it does.

maxGimeno:
The doc for the new script is here : https://github.com/CGAL/cgal/wiki/Linux-X86-ARM-Cross-Compilation-(WIP)#use-the-script

Scripts/developer_scripts/global_run_testsuite.sh
lrineau: For the record, this morning we said: `autotest_cgal_with_ctest`

Scripts/developer_scripts/autotest_cgal_with_ctest
```diff
@@ -0,0 +1,268 @@
...
+    DECOMPRESSOR="${GUNZIP}"
+    log_done "${ACTUAL_LOGFILE}"
+  fi
+
+  if [ "${USE_TARBZ}" = "y" ]; then
+    DECOMPRESSOR="bunzip2"
+  fi
+
+  log "${ACTUAL_LOGFILE}" "untarring CGAL"
+  ${DECOMPRESSOR} -c "${CGAL_ZIPFILE}" > /dev/null
```
lrineau: This line must be removed.
```suggestion
```

lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-159.shtml.

PR #3765, opened 2019-03-12
Prevent infinite loop in Advancing front surface reconstruction. 

dangerski:
In some cases K can become really large but doesn't get to infinity and instead gets into a situation where K + epsilon == K and the the while loop never exits. 

## Summary of Changes

Check for the case where K stops changing and exit the loop.  This happens when K + step == K for large values of K.

## Release Management

* Affected package(s):  Advancing front surface reconstruction
* Issue(s) solved (if any): 
* Feature/Small Feature (if any): Small bug fix
* Link to compiled documentation (obligatory for small feature) [link](https://doc.cgal.org/latest/Advancing_front_surface_reconstruction/index.html)
* License and copyright ownership: CGAL

### DIFF
--- a/Advancing_front_surface_reconstruction/include/CGAL/Advancing_front_surface_reconstruction.h
+++ b/Advancing_front_surface_reconstruction/include/CGAL/Advancing_front_surface_reconstruction.h
@@ -2005,6 +2005,7 @@ namespace CGAL {
     {
       // initilisation de la variable globale K: qualite d'echantillonnage requise
       K = K_init; // valeur d'initialisation de K pour commencer prudemment...
+      coord_type K_prev = K;
 
       Vertex_handle v1, v2;
       if (_ordered_border.empty()){
@@ -2069,12 +2070,12 @@ namespace CGAL {
             }
           while((!_ordered_border.empty())&&
                 (_ordered_border.begin()->first < STANDBY_CANDIDATE_BIS));
-
+          K_prev = K;
           K += (std::max)(K_step, min_K - K + eps);
           // on augmente progressivement le K mais on a deja rempli sans
           // faire des betises auparavant...
         }
-      while((!_ordered_border.empty())&&(K <= K)&&(min_K != infinity()));
+      while((!_ordered_border.empty())&&(K <= K)&&(min_K != infinity())&&(K!=K_prev));
 
 #ifdef VERBOSE
       if ((min_K < infinity())&&(!_ordered_border.empty())) {


sloriot:
@afabri @sgiraudot could you review this PR? You know better than me the internal. If it is a bug it would be good to have it integrated in the next release.

afabri:
@dangerski do you have an example data set where that problem shows up?

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-151.shtml says it doesn't break anything.

dangerski:
@afabri Yes I have an example dataset:

[https://github.com/dangerski/cgal/blob/example_file/Advancing_front_surface_reconstruction/test/Advancing_front_surface_reconstruction/data/afsr_infinite_loop.xyz](https://github.com/dangerski/cgal/blob/example_file/Advancing_front_surface_reconstruction/test/Advancing_front_surface_reconstruction/data/afsr_infinite_loop.xyz)

And a somewhat minimal example to reproduce the error.
```c++
#include <iostream>
#include <fstream>

#include <CGAL/Advancing_front_surface_reconstruction.h>
#include <CGAL/IO/read_xyz_points.h>


using Kernel = CGAL::Exact_predicates_inexact_constructions_kernel;
using CGSurfaceMesh = CGAL::Surface_mesh<Kernel::Point_3>;
using Facet = CGAL::cpp11::array<std::size_t, 3>;

struct Construct {
	CGSurfaceMesh& mesh;
	template < typename PointIterator>
	Construct(CGSurfaceMesh& mesh,PointIterator b, PointIterator e) : mesh(mesh)
	{
		for(; b!=e; ++b){
			boost::graph_traits<CGSurfaceMesh>::vertex_descriptor v;
			v = add_vertex(mesh);
			mesh.point(v) = *b;
		}
	}
	Construct& operator=(const Facet f)
	{
		typedef boost::graph_traits<CGSurfaceMesh>::vertex_descriptor vertex_descriptor;
		typedef boost::graph_traits<CGSurfaceMesh>::vertices_size_type size_type;
		mesh.add_face(vertex_descriptor(static_cast<size_type>(f[0])),
		              vertex_descriptor(static_cast<size_type>(f[1])),
		              vertex_descriptor(static_cast<size_type>(f[2])));
		return *this;
	}
	Construct& operator*() { return *this; }
	Construct& operator++() { return *this; }
	Construct operator++(int) { return *this; }
};

struct Perimeter {
	double bound;

	Perimeter(double bound) : bound(bound){}

	template <typename AdvancingFront, typename Cell_handle>
	double operator() (const AdvancingFront& adv, Cell_handle& c, const int& index) const
	{
		if (bound != 0) {
			double d = 0;
			d = sqrt(squared_distance(c->vertex((index + 1) % 4)->point(),
			                          c->vertex((index + 2) % 4)->point()));
			if (d > bound) return adv.infinity();
			d += sqrt(squared_distance(c->vertex((index + 2) % 4)->point(),
			                           c->vertex((index + 3) % 4)->point()));
			if (d > bound) return adv.infinity();
			d += sqrt(squared_distance(c->vertex((index + 1) % 4)->point(),
			                           c->vertex((index + 3) % 4)->point()));
			if (d > bound) return adv.infinity();
		}
		// Otherwise, return usual priority value: smallest radius of
		// delaunay sphere
		return adv.smallest_radius_delaunay_sphere (c, index);
	}
};

int main() {

	std::list<Kernel::Point_3> points_list;
	std::ifstream fin;
	fin.open("afsr_infinite_loop.xyz");
	CGAL::read_xyz_points(fin, std::back_inserter(points_list));
	std::cout << "Number of points: " << points_list.size() << std::endl;
	fin.close();

	CGSurfaceMesh mesh;

	Construct construct(mesh, points_list.begin(), points_list.end());
	Perimeter perimeter(3.5);
	std::cout << "Advancing front..." << std::endl;
	CGAL::advancing_front_surface_reconstruction(points_list.begin(), points_list.end(), construct, perimeter);

}
```

afabri:
@dangerski which version of CGAL and which compiler did you use. I ask as I cannot reproduce the infinite loop with VC2017 and neither with master nor with CGAL-4.13.

dangerski:
@afabri This was on OSX Mojave with both CGAL-4.13 and master(b2b0d9d486).  Here is the CMake file.  I was also able to reproduce it on a Centos7 machine.  I'll see if I can put together a docker image for that.
```
cmake_minimum_required(VERSION 3.13)
project(afsr)

set(CMAKE_CXX_STANDARD 11)

#Eigen
find_package(Eigen3 3.3 REQUIRED NO_MODULE)
include_directories(${EIGEN3_INCLUDE_DIRS})

#Boost
find_package( Boost 1.67.0 )
include_directories( ${Boost_INCLUDE_DIRS} )

#CGAL
set(CGAL_DIR "/usr/local/")
find_package(CGAL COMPONENTS Core REQUIRED)
include(${CGAL_USE_FILE})
include( CGAL_CreateSingleSourceCGALProgram )
message("CGAL DIR ${CGAL_DIR}")

add_executable(afsr main.cpp)
```

sgiraudot:
I can't reproduce the bug either, tested on Linux Mint with both 4.13 and master.

dangerski:
I made some docker images but am unable to reproduce it on Ubuntu.  However, I ran into extremely long run times on Centos7 that seem to be a different issue.  I am able to consistently reproduce this issue on OSX but there is no docker image for it.  I'll put up my centos7 image.

sloriot:
I'll give it a try on Mac OSX

sloriot:
Can't reproduce it using a debug build on:
`Darwin cezanne.local 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64`

using

```
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
Output in lldb:
```
(lldb) r afsr_infinite_loop.xyz
Process 1810 launched: './main' (x86_64)
Number of points: 202700
Advancing front...
Process 1810 exited with status = 0 (0x00000000)
```

However, I ran it again in release mode. It seems to loop. @sgiraudot , did you try in release mode or only debug?

sloriot:
@sgiraudot where do we stand with this issue? (see also my previous question)

sgiraudot:
Sorry, I think I missed your message.

I tried it both in release, debug, with both Clang and GCC, I could not reproduce. I really don't know what to do with this bug: we could integrate the PR as it seems harmless anyway, but I'd prefer if we at least understood what's going on…

dangerski:
I was able to reproduce it in an Ubuntu docker image. Haven't had time to make a minimal reproducible example yet.

lrineau:
I consider this pull-request to be stalled, until one of you manage to reproduce the bug on a small test-case.

dangerski:
@afabri Ok I now have a minimal reproducible example in a docker image.  It will run AFSR on several data sets because some examples will reproduce on one OS but not the other.  I am able to consistently get at least one of those examples to reproduce the bug on OSX and Ubuntu.  Here is the docker file:

[https://github.com/dangerski/cgal/blob/example_file/Advancing_front_surface_reconstruction/test/docker/Dockerfile](https://github.com/dangerski/cgal/blob/example_file/Advancing_front_surface_reconstruction/test/docker/Dockerfile)

Download the docker file and run:
`$ docker build . -t cgal_afsr_repro_ubun -f Dockerfile`
`$ docker run --rm -it cgal_afsr_repro_ubun:latest`

maxGimeno:
I can't reproduce. I keep getting [ next K required = 1.39493e+15 ][ current K = 1.39493e+15 ][ K_prev = 1.39493e+15] and it won't crash...

dangerski:
@maxGimeno the bug is an infinite loop when K + step == K for large values of K so it looks like you reproduced it.

maxGimeno:
Ow OK, sorry, I was expecting a crash for some reason. So yeahm I reproduce :)

dangerski:
@afabri @lrineau Any chance we can get this merged as it has now been reproduced?  Thanks

lrineau:
> @afabri @lrineau Any chance we can get this merged as it has now been reproduced? Thanks

The persons  in charge are @afabri, @sgiraudot, and maybe @sloriot.

sgiraudot:
I haven't had the chance to investigate on why this bug happens, and although the proposed fix shouldn't do any harm in any case, I'd still like to understand and what is going, especially considering this apparently happens on _one very specific platform_ (so specific that we couldn't do better than a Docker image to reproduce).

But again, as the fix is quite harmless, I have no opposition if we want to integrate it now, as it fixes one problem while not changing anything to the already working cases.

dangerski:
@sgiraudot I should clarify that this can be reproduced on OSX, Ubuntu, and Centos.  The docker image was just created to ease reproduction here with minimal effort.  The minimal example program and datasets used in the docker image _should_ reproduce this issue on just about any machine.  One dataset may cause the issue to happen on one machine and not another which is why I iterate over several in the example program.  Each dataset in the example has reproduced the issue on at least one test machine (inside and outside of docker).

There may be another way to fix this to prevent K from blowing up but I would think, even then, you would still want to make sure to avoid the infinite loop.

Infinite loop bugs are nefarious when they happen randomly(never know which dataset will cause it) and are difficult to detect.  I figured other users may find this useful which is why I created the pull request.

sloriot:
@sgiraudot you can open an issue from a relevant comment from this PR if you want to investigate later what's going on.
@lrineau I think it make sense to have this patch integrated in the upcoming release (provided it has already been tested).

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Sep 25, 2019, 6:32 AM GMT+2](https://github.com/CGAL/cgal/pull/3765#issuecomment-534846756 "2019-09-25T04:32:02Z - Replied by Github Reply Comments"):
> @lrineau I think it make sense to have this patch integrated in the upcoming release (especially because it has already been tested).

Probably too late for CGAL-5.0-beta1, but it will be in CGAL-5.0. finale version.

maxGimeno:

lrineau:
This PR is a bug-fix, that was created during the release cycle of 5.0. I will merge it in 5.0.1.

PR #3771, opened 2019-03-15
Use a generated version of deprecated.html

### LINKED ISSUE #3753: Obsolete documentation file
During the post processing phase of `make doc_with_postprocessing` the file `deprecated.html` is copied to the `Manual` output directory.
- as far as I can see the is never referenced
- the file refers to CGAL 4.3, independent of the used version
- the file is not XHTML compliant

The original file can be found, for the different versions, at:
```
./doc/resources/1.8.13/deprecated.html
./doc/resources/1.8.14/deprecated.html
./doc/resources/1.8.4/deprecated.html

```

the copying is done nearly at the end of the file `Documentation/doc/scripts/html_output_post_processing.py` in the lines:
```
    #copy deprecated.html
    shutil.copy(path.join(resources_absdir,"deprecated.html"),path.join("Manual/", "deprecated.html"))
```

As I'm not 100%sure about the usage with doxygen version 1.8.4 I have not created a pull request to remove the files and code, but when the support for 1.8.4 is removed I think these files / code should also be removed.

sloriot:
Fix #3753 

Put in dev man to avoid being at the root of the navtree


Documentation/doc/Documentation/deprecated.txt
```diff
@@ -0,0 +1,11 @@
+/*!
+
+\page deprecated Deprecation Notice
+
+In \cgal, we are trying as much as possible not to break the backward compatibility. If we have to do it, it will be announced in the <a href="https://www.cgal.org/releases.html">release notes</a> ,
+and a solution will always be provided. When possible, we keep the deprecated API for two releases before removing it.
+If you reached this page after clicking on a link, it must probably mean that the documentation page you were trying to reach
+is now deprecated. Have a look at the <a href="https://www.cgal.org/releases.html">release notes</a> for more information.
```
lrineau: This text is wrong. `deprecated.html` can only be reached by following a link that points to a page that has been removed in CGAL-4.3.

sloriot:
As I said the page appear in the navtree. So I decided to put it in the devman to be the less cumbersome.

maxGimeno:

PR #3772, opened 2019-03-15
Basic viewer: use compatibility shaders in basic_viewer

maxGimeno:
## Summary of Changes
Remove remaining `#version` from compatibility glsl and use shaders without #version in basic viewer in case the ES context is used to avoid glsl error.


MaelRL:
It's better: I get the vertices and edges on each call of draw(), but on the other hand I have lost face colors at all iterations :).

Here is the error log I get now: https://gist.github.com/MaelRL/126694fb91f3332a3e1918df11c2da2a.

CC @gdamiand for info.

maxGimeno:
Yes ! I knew there was more to it than I remembered. 
This should do it .

MaelRL:
Some different error now: https://gist.github.com/MaelRL/656d0ab81b4b0d673fec21b86235e097

maxGimeno:
Oops, sorry.

MaelRL:
Colors and no errors on the first display, no color from 2nd iteration (but still vertices/edge).

Log: https://gist.github.com/MaelRL/73c9d6fd53304f5e43d5f6b63d0ab850

maxGimeno:
Ok, I forgot a line. You can try again

MaelRL:
Same: https://gist.github.com/MaelRL/994ef770363d33959cc89b0f1e458c4c

maxGimeno:
Fixed. It's a real shame that I can't test by myself, it twould take a lot less time :(

MaelRL:
Let's just settle this directly on my machine at some point.

lrineau:
It is really annoying that our big 3D demo and the basic viewers do not use the same shaders.

I really would like to know why.

maxGimeno:
The equivalent shaders in the demo are far more complicated, so I don't want to complexify everything in the Basic_viewer code to use them, and the shaders in the basic_viewer have some specificities (like the point size management) that we don't want in the demo for performence purpose.

gdamiand:
I agree with @lrineau ; it could be better to factorize more code between the demo and the basic viewer.

maxGimeno:
@maelRL we really need to find nome time to finish this.

Conflicts.

maxGimeno:

PR #3774, opened 2019-03-15
Fix error in outer CCB setting

### LINKED ISSUE #3747: Polygon_with_holes_2 difference fails for valid input polygons with holes
## Issue Details

I try to compute an intersection on polygons with holes. 
- The first polygon (from which we subtract) has two holes, one of which is intersecting the outline at one vertex. 

screenshot 2019-03-08 at 12 05 39

- The second polygon (which we subtract) intersects the hole which is touching the outline at one vertex and it intersects the other hole of the first polygon at two vertices.

screenshot 2019-03-08 at 12 01 19

Both polygons are valid polygons with holes.

It fails with the error:

> CGAL error: assertion violation!
> Expression : h->flag()!=NOT_VISITED
> File       : /usr/local/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
> Line       : 1207
> Explanation: 
> Refer to the bug-reporting instructions at https://www.cgal.org/bug_report.html
> libc++abi.dylib: terminating with uncaught exception of type CGAL::Assertion_exception: CGAL ERROR: assertion violation!
> Expr: h->flag()!=NOT_VISITED
> File: /usr/local/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
> Line: 1207
> Abort trap: 6

I'm assuming that when the input is valid, CGAL should not fail because the operations will produce valid polygons again since those polygon(-sets) are closed under these operations. Am I doing something wrong?

## Source Code
```
#include <sstream>
#include <iterator>

#include <CGAL/Boolean_set_operations_2.h>

typedef CGAL::Polygon_with_holes_2<Kernel>                Polygon_with_holes_2;
typedef std::list<Polygon_with_holes_2>                   Pwh_list_2;

int main() {
  std::stringstream ps("4  -2 2  -2 -2  2 -2  2 2  2  3  -2 -2  -1 1  -1 -1  3  0 -1  0 1  1 1");
  Polygon_with_holes_2 P;
  ps >> P;

  std::stringstream qs("3  0 -1  0 1  -1 1  0");
  Polygon_with_holes_2 Q;
  qs >> Q;

  Pwh_list_2 result;
  CGAL::difference(P, Q, std::back_inserter(result));

  return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): MacOS 10.14
* Compiler: clang/LLVM
* CGAL version: 4.13 stable (via homebrew)
* Boost version: 1.68.0_1

sloriot:
Fixes #3747

@efifogel would you mind add the failing test in the testsuite ?

### DIFF
--- a/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
+++ b/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
@@ -1112,6 +1112,15 @@ protected:
       }
     }
     while(something_was_updated);
+    // last loop, if some tags are not set it means that they are the only ccb
+    // of the face and that they have to be the outer ccb
+    BOOST_FOREACH(Halfedge_handle h, halfedges_that_was_on_an_outer_ccb)
+    {
+      if (h->flag()!=NOT_VISITED) continue;
+      std::size_t face_master_id=(*uf_faces.find(face_handles[h->face()->id()]))->id();
+      set_flag_of_halfedges_of_final_argt(h,ON_OUTER_CCB);
+      face_outer_ccb_set[face_master_id]=true;
+    }
     // at this position there might be some bits in face_outer_ccb_set not set
     // but they are corresponding to the unbounded face
   // End tagging ccbs


maxGimeno:

PR #3775, opened 2019-03-15
Doc fixes

MaelRL:
## Summary of Changes

@sloriot @albert-github Would you please have any suggestion on how to:
- remove the link to `Vector_3` on https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__meshing__grp.html#ga6cf0f4aaf1f304d129d76b9d20a4e909 (`%` does not seem to work: it's just printed next to the linked `Vector_3`).
- have code in a snippet (e.g. via `\code{.sh} ... \endcode`) while still resolving aliases such as `\cgal` within said snippet

### DIFF
--- a/Periodic_2_triangulation_2/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt
+++ b/Periodic_2_triangulation_2/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt
@@ -234,7 +234,7 @@ is also a model for the concept `TriangulationTraits_2`.
 
 The kernels `Cartesian`, `Homogeneous`,
 `Simple_cartesian`, `Simple_homogeneous` and
-`Filtered_kernel` can all be used as \cgalModels for
+`Filtered_kernel` can all be used as models for
 `Traits`. `Periodic_2_triangulation_traits_2` provides exact
 predicates and exact constructions if `Traits` does. It provides
 exact predicates but not exact constructions if
--- a/Polygon/include/CGAL/Polygon_2.h
+++ b/Polygon/include/CGAL/Polygon_2.h
@@ -516,18 +516,17 @@ transform(const Transformation& t, const Polygon_2<Traits_P,Container_P>& p);
 /// followed by the output of the coordinates of the vertices.
 /// @{
 
-/// Inserts the polygon `p` into the stream `os`. \pre The insert
-/// operator must be defined for `Point_2`.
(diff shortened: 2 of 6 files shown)


albert-github:
We have here the `@tparam` description and the definition in the function (I looked at the current master). We currently have:
```
 * @tparam Vector_3 vector type from the same CGAL kernel as the point of the vertex point map used for `OutputMesh`.
```
and
```
                  #ifdef DOXYGEN_RUNNING
                  Vector_3 v,
                  #else
                  typename GetGeomTraits<OutputMesh, NamedParameters2>::type::Vector_3 v,
                  #endif
```

Why not drop the `@tparam Vector_3` and in the function call remove the doxygen specific part, as the real argument contains no template parameter other than the already described. The result (including the removing of the empty comment lines as described in the note below) looks like:
result_extrude_mesh

*Note*
the parameter description is also a bit strange. The `*` problem has been handled in this pull request. I think it is also wise to remove the empty comment line between:
```
* @param np_in ...
*
* \cgalNamedParamsBegin
```
analogous for `np_out` and on other places.

MaelRL:
`GetGeomTraits` is undocumented, which is probably why this `#ifdef DOXYGEN_RUNNING` was put in the first place. However, its name could be considered clear-enough.

maxGimeno:

sloriot:
You can rename `Vector_3` to `Vector` or `Vector_3_`. Or we get document the helper class...

sloriot:
For the snippet question, I don't think it is possible.

lrineau:
Can I merge this tested pull-request?

And then Mael will open an issue for the remaining issues:

[**@MaelRL**](https://github.com/MaelRL) commented on [Mar 15, 2019, 9:20 PM GMT+1](https://github.com/CGAL/cgal/pull/3775#issue-261686936 "2019-03-15T20:20:11Z - Replied by Github Reply Comments"):
> @sloriot @albert-github Would you please have any suggestion on how to:
> - remove the link to `Vector_3` on https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__meshing__grp.html#ga6cf0f4aaf1f304d129d76b9d20a4e909 (`%` does not seem to work: it's just printed next to the linked `Vector_3`).
> - have code in a snippet (e.g. via `\code{.sh} ... \endcode`) while still resolving aliases such as `\cgal` within said snippet

@MaelRL, do you agree?

MaelRL:
Yes.

PR #3778, opened 2019-03-19
Fix the *installed* header-only CGALConfig.cmake

lrineau:
## Summary of Changes

The previous installed version of `CGALConfig.cmake`, when
`CGAL_HEADER_ONLY` is true, was hard-coding a path `../../../`,
relative to the location of itself. It might be that the CMake modules
are not installed in a sub-directory with three components, relatively
to the prefix.

That way of doing it buggy. And that triggered issue with the cgal
port in vcpkg: https://github.com/Microsoft/vcpkg/pull/5734.

This patch is just another hack: it hardcodes that the relative path
can be any of:

  - `../`,
  - `../../`, or
  - `../../../`.

## Release Management


maxGimeno:

PR #3779, opened 2019-03-19
Polyhedron:  read_OFF() must not set the badbit if there are 0 vertices

### LINKED ISSUE #3108: Inconsistence of read_off(istream, FaceGraph)
## Issue Details

The overloads of the function `read_off()` do not the same for an OFF file with "0 0 0" for the simplex counters.
For[ `read_off(istream&,Polyhedron&)`](https://doc.cgal.org/latest/Polyhedron/group__PkgPolyhedronIOFunc.html#ga9791a3720c7a3e9e32e43aed28cd9695) the stream is `bad` after reading.  The manual says  "If the stream in does not contain a permissible polyhedral surface the ios::badbit of the input stream in is set and P remains unchanged.", but I would consider an empty Polyhedron as "permissible".

## Environment

* CGAL version: 4.12

afabri:
## Summary of Changes

When reading an OFF file with 0 vertices the `std::ios::badbit` was set.  This was documented behavior.
I suggest that we change this with CGAL 5.0.

Is this a Breaking Change?   I didn't modify changes.md yet, as it makes merging painful.

## Release Management

* Issue(s) solved (if any): fix #3108


maxGimeno:

sloriot:
retested successfully in CGAL-5.0-Ic-10

PR #3783, opened 2019-03-20
BGL:  Seam_mesh::add_seams() requires .selection.txt

### LINKED ISSUE #3782: Bug while checking filename format in Seam_mesh::add_seams
Hello everyone, I believe I have found a little bug.

## Issue Details

Function `Seam_mesh::add_seams(const char* filename, const VdContainer& tm_vds)` assumes `filename` to be of the form `*.selection.txt`. When called with a sufficiently long `filename`, a reasonable error message is printed:

    > ./CGAL-4.13-build/examples/Surface_mesh_parameterization/seam_Polyhedron_3 surface.off very_long_filename.txt
    Error: seams must be given by a *.selection.txt file

However, when `filename` is too short, CGAL crashes:

    > ./CGAL-4.13-build/examples/Surface_mesh_parameterization/seam_Polyhedron_3 surface.off short.txt
    terminate called after throwing an instance of 'std::out_of_range'
      what():  basic_string::substr
    Aborted

## Source Code

The culprit seems to be the line 1078 in `include/CGAL/boost/graph/Seam_mesh.h`:

     if(str.substr(str.length() - 14) != ".selection.txt") {

Instead, the correct check (in my view) would be something like

     if(str.length() < 14 || str.substr(str.length() - 14) != ".selection.txt") {

However, I would go as far as questioning whether the requirement on `filename` to be of the form `*.selection.txt` is really necessary. After all, we just need a text file with space-separated non-negative integers. Whether it's called `seams.selection.txt` or `seams.txt` or even `numbers.abc` seems irrelevant to the problem in question.

If there is a reason for it, I would recommend mentioning the requirement in the documentation as well.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux 64 bits
* Compiler: g++
* CGAL version: 4.13
* Boost version:

afabri:
## Summary of Changes

Clarify that we require that the selection comes from a file with the extension `*.selection.txt`. 

## Release Management

* Affected package(s): BGL
* Issue(s) solved (if any): fix #3782

### DIFF
--- a/BGL/include/CGAL/boost/graph/Seam_mesh.h
+++ b/BGL/include/CGAL/boost/graph/Seam_mesh.h
@@ -1107,7 +1107,7 @@ public:
   ///
   /// \returns one of the halfedges of the seam mesh that is on a seam.
   ///
-  /// \pre filename should be the name of a \cgal selection filee with file extension "*.selection.txt":
+  /// \pre filename should be the name of a \cgal selection file with file extension "*.selection.txt":
   ///      edges are described by pairs of integers, on the third line of the file.
   TM_halfedge_descriptor add_seams(const char* filename)
   {
--- a/BGL/include/CGAL/boost/graph/Seam_mesh.h
+++ b/BGL/include/CGAL/boost/graph/Seam_mesh.h
@@ -1065,7 +1065,7 @@ public:
   /// \tparam VdContainer must be a model of <a href="http://en.cppreference.com/w/cpp/concept/SequenceContainer"><tt>SequenceContainer</tt></a> (that is, provide
   ///         the functions: `operator[]` and `at()`).
   ///
-  /// \pre filename should be the name of a \cgal selection file with the file extension "*.selection.txt":
+  /// \pre filename should be the name of a \cgal selection file with file extension "*.selection.txt":
   ///      edges are described by pairs of integers, on the third line of the file.
   template<typename VdContainer>
   TM_halfedge_descriptor add_seams(const char* filename,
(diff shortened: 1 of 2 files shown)


BGL/include/CGAL/boost/graph/Seam_mesh.h
```diff
@@ -1107,8 +1107,8 @@ class Seam_mesh
   ///
   /// \returns one of the halfedges of the seam mesh that is on a seam.
   ///
-  /// \pre filename should be the name of a CGAL selection file: edges are
-  ///      described by pairs of integers, on the third line of the file.
+  /// \pre filename should be the name of a CGAL selection filee with the file extension "*.selection.txt":
```
MaelRL:
```suggestion
  /// \pre filename should be the name of a \cgal selection file with file extension "*.selection.txt":
```

BGL/include/CGAL/boost/graph/Seam_mesh.h
```diff
@@ -1107,8 +1107,8 @@ class Seam_mesh
   ///
   /// \returns one of the halfedges of the seam mesh that is on a seam.
   ///
-  /// \pre filename should be the name of a CGAL selection file: edges are
-  ///      described by pairs of integers, on the third line of the file.
+  /// \pre filename should be the name of a \cgal selection filee with file extension "*.selection.txt":
```
MaelRL:
```suggestion
  /// \pre filename should be the name of a \cgal selection file with file extension "*.selection.txt":
```

maxGimeno:
Successfullty tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-155.shtml

PR #3794, opened 2019-03-25
PMP: Fix in normalize()

maxGimeno:
## Summary of Changes

If the normal vector is small enough, it might happen that the vector is considered not null, but the approximate_sqrt of its norm is. Then the returned normalized vector is Nan. To avoid that, i added a check that the norm is not null.

## Release Management

* Affected package(s):PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
@@ -51,7 +51,10 @@ namespace internal {
         traits.compute_squared_length_3_object()(v));
     //If the vector is small enough, approx_sqrt might return 0, and then we get nan values. 
     //To avoid that, we check the resulted norm. If it is 0, we don't normalize.
-    v = traits.construct_divided_vector_3_object()(v, norm == 0 ? 1 : norm);
+    if(norm != 0)
+    {
+      v = traits.construct_divided_vector_3_object()(v, norm );
+    }
   }
 
   template<typename Point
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
```diff
@@ -266,7 +266,10 @@ compute_vertex_normal(typename boost::graph_traits<PolygonMesh>::vertex_descript
     he = opposite(next(he, pmesh), pmesh);
   } while (he != end);
 
-  if ( ! traits.equal_3_object()(normal, CGAL::NULL_VECTOR))
+  if(normal.x() != normal.x()
```
sloriot: You are normalizing only if one coord is a NAN?

maxGimeno: Arf I thought I removed that.

maxGimeno: It is acually absolutely not the right commit

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
```diff
@@ -49,7 +49,9 @@ namespace internal {
   {
     typename GT::FT norm = CGAL::approximate_sqrt(
         traits.compute_squared_length_3_object()(v));
-    v = traits.construct_divided_vector_3_object()(v, norm);
+    //If the vector is small enough, approx_sqrt might return 0, and then we get nan values. 
+    //To avoid that, we check the resulted norm. If it is 0, we don't normalize.
+    v = traits.construct_divided_vector_3_object()(v, norm == 0 ? 1 : norm);
```
janetournois: why don't you just use a `if(norm != 0)` ?
It avoids doin another construction

maxGimeno: Indeed.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.14-Ic-157.shtml

PR #3805, opened 2019-03-27
Number_types: Add support for arm and aarch64

maxGimeno:
## Summary of Changes
Define the floating point control word values for arm architectures (for gcc).
## Release Management

* Affected package(s):Number Types

### DIFF
--- a/Number_types/include/CGAL/FPU.h
+++ b/Number_types/include/CGAL/FPU.h
@@ -383,8 +383,23 @@ typedef unsigned int FPU_CW_t;
 #define CGAL_FE_TOWARDZERO   _RC_CHOP
 #define CGAL_FE_UPWARD       _RC_UP
 #define CGAL_FE_DOWNWARD     _RC_DOWN
-# elif (defined __VFP_FP__ && !defined __SOFTFP__) || defined __aarch64__
-//to do
+# elif defined  __arm__
+#define CGAL_IA_SETFPCW(CW) asm volatile ("VMSR FPSCR, %0" : :"r" (CW))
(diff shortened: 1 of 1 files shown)


Number_types/include/CGAL/FPU.h
```diff
@@ -383,6 +383,22 @@ typedef unsigned int FPU_CW_t;
 #define CGAL_FE_TOWARDZERO   _RC_CHOP
 #define CGAL_FE_UPWARD       _RC_UP
 #define CGAL_FE_DOWNWARD     _RC_DOWN
+# elif defined  __arm__
```
mglisse: There are several possible FPUs for ARM (plus softfloat), and I think you don't have `__arm__` when using thumb, so I don't know if that is the right test.

lrineau: Thanks. I have fixed that in https://github.com/CGAL/cgal/pull/3805/commits/1aee92af751a7021c64082c69dfd2ca6c569c67c. For the moment, I only care about VFP and Neon.

mglisse: https://wiki.debian.org/ArmEabiPort suggests `#if (defined(__VFP_FP__) && !defined(__SOFTFP__))` like we already use elsewhere in the file.

lrineau: OK. Thanks for the reference. I have committed https://github.com/CGAL/cgal/pull/3805/commits/42cbce9fbf9ad3becb5d1c20ee2eb106b1de7c61.

lrineau:

PR #3806, opened 2019-03-27
Polyline_simplification_2: Fix for Projection_traits_xy_3

afabri:
## Summary of Changes

Adding the distance computation of the projection of a 3D point and a 3D segment to the class `Projection_traits_3` enables to simplify 3D polylines.  I added an example and a subsection in the manual. 

## Release Management

* Affected package(s): Polyline_simplification_2
* Issue(s) solved (if any): fix an [issue reported on stackoverflow](https://stackoverflow.com/questions/55366527/is-it-possible-to-use-the-projection-traits-xy-3-adapter-with-the-2d-polyline-s).
* License and copyright ownership:  GF


Polyline_simplification_2/examples/Polyline_simplification_2/simplify.cpp
```diff
@@ -8,7 +9,8 @@
 
 namespace PS = CGAL::Polyline_simplification_2;
 
-typedef CGAL::Exact_predicates_inexact_constructions_kernel K;
+typedef CGAL::Exact_predicates_inexact_constructions_kernel Epic;
+typedef CGAL::Projection_traits_xy_3<Epic>  K;
```
sloriot: This is probably a left-over...

lrineau:

PR #3807, opened 2019-03-27
Fix the order of link when GMP is a static lib

lrineau:
## Summary of Changes

The test of CGAL-4.14-beta3 with vcpkg (on Linux) showed that there is an error in the order of links of GMP and MPFR, when GMP is a static library.

## Release Management


Installation/cmake/modules/CGAL_SetupGMP.cmake
```diff
@@ -61,15 +61,14 @@ function(use_CGAL_GMP_support target)
...
   else()
     target_include_directories(${target} SYSTEM ${keyword}
       $<BUILD_INTERFACE:${MPFR_INCLUDE_DIR}>
       $<INSTALL_INTERFACE:include>)
   endif()
-  target_link_libraries(${target} ${keyword} ${MPFR_LIBRARIES})
+  target_link_libraries(${target} ${keyword} ${MPFR_LIBRARIES} ${GMP_LIBRARIES})
   if(WITH_GMPXX OR CGAL_WITH_GMPXX)
     target_include_directories(${target} SYSTEM ${keyword}  ${GMPXX_INCLUDE_DIR})
     target_link_libraries(${target}  ${keyword} ${GMPXX_LIBRARIES})
```
mglisse: Don't you have the same problem here that gmpxx should come before gmp?

lrineau: You are right. That setup is never tested because `CGAL_WITH_GMPXX` is `OFF` by default, and we only set it on a few test platforms where GMP, MPFR, and GMPXX are *shared libraries*.

lrineau: Fixed by https://github.com/CGAL/cgal/pull/3807/commits/4b7cc4a896f1ed3f1047b96bb5f90646970aba4a.

lrineau: Actually, I have tried to use `-DCGAL_USE_GMPXX=1` without the patch of https://github.com/CGAL/cgal/commit/4b7cc4a896f1ed3f1047b96bb5f90646970aba4a, with the vcpkg package (that has static MPIR/MPFR libraries), and that compiled fine. It seems that all the undefined symbols of `libgmpxx.a` are also undefined in `libmpfr.a` (and thus pulled by the linker).

mglisse: And gmpxx is very small and we don't use much from it. Anyway, better do it right. Note that if doing the inclusions in the right order becomes complicated, you can take advantage of the fact that it is perfectly ok to link multiple times with the same library and use something like `target_link_libraries(${target}  ${keyword} ${GMPXX_LIBRARIES} ${GMP_LIBRARIES})`.

lrineau:

PR #3814, opened 2019-03-28
Surface Mesh: Fix not reseting edge property when a new element re-uses memory

MaelRL:
## Summary of Changes

https://github.com/CGAL/cgal/pull/2358 made it so new elements of the `Surface_mesh` use default property map values instead of values from the deleted element, but edge properties were forgotten.

## Release Management

* Affected package(s): `Surface_mesh`


lrineau:

PR #3820, opened 2019-03-31
Correction of links

albert-github:
Correction of redirected links and links that moved to another place.

### DIFF
--- a/Solver_interface/include/CGAL/Eigen_diagonalize_traits.h
+++ b/Solver_interface/include/CGAL/Eigen_diagonalize_traits.h
@@ -52,7 +52,7 @@ namespace CGAL {
 ///
 /// \cgalModels `DiagonalizeTraits`
 ///
-/// \sa http://eigen.tuxfamily.org
+/// \sa http://eigen.tuxfamily.org/index.php?title=Main_Page
 template <typename FT, unsigned int dim = 3>
 class Eigen_diagonalize_traits
 {
--- a/.gitignore
+++ b/.gitignore
@@ -1063,7 +1063,7 @@ cmake_install.cmake
 *~
 .#*
 
-# MacOS file http://en.wikipedia.org/wiki/.DS_Store(
+# MacOS file https://en.wikipedia.org/wiki/.DS_Store
 .DS_Store
 
 # Binaries:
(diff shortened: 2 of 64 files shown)


Solver_interface/include/CGAL/Eigen_diagonalize_traits.h
```diff
@@ -52,7 +52,7 @@ namespace CGAL {
 ///
 /// \cgalModels `DiagonalizeTraits`
 ///
-/// \sa http://eigen.tuxfamily.org
+/// \sa http://eigen.tuxfamily.org/index.php?title=Main_Page
```
sloriot: why is such a change needed?

albert-github: When I use in the browser:
http://eigen.tuxfamily.org
I get the page but it is automatically redirected to:
http://eigen.tuxfamily.org/index.php?title=Main_Page

sloriot: I know but why is it an issue? Does it trigger a false positive in your test?

albert-github: Indeed in the link test it gives that a redirect takes place, one can discuss about whether or not it is a false positive or not but also changes from http -> https give this type of message:

```
http://eigen.tuxfamily.org/
-> http://eigen.tuxfamily.org/index.php?title=Main_Page
  Line: 403
  Code: 301 -> 200 OK
 To do: This is a permanent redirect. The link should be updated.
```
and:
```
http://www.threadingbuildingblocks.org/
-> https://www.threadingbuildingblocks.org/
  Line: 413
  Code: 301 -> 200 OK
 To do: This is a permanent redirect. The link should be updated.
```

sloriot: 100% agree for the https, my comment was only about the extra part.

albert-github: I had another look at the code change:
```
The \sc{Eigen} web site is <A HREF="http://eigen.tuxfamily.org/index.php?title=Main_Page">`http://eigen.tuxfamily.org/index.php?title=Main_Page`</A>.
```
I think I get a bit what the main objection is, the part shown to the user. I think that in this case it might be better to have:
```
The \sc{Eigen} web site is <A HREF="http://eigen.tuxfamily.org/index.php?title=Main_Page">`http://eigen.tuxfamily.org`</A>.
```
meaning that the user sees the "short" name and the reference is directly to the correct place.

sloriot:
I fixed the Eigen link, tested locally and merged the branch in 4.14 after rebasing it.

PR #3830, opened 2019-04-03
Fix CGAL DLLs verinfo on Windows

lrineau:
## Summary of Changes

Detail to fix: the product version was `CGAL_VERSION`.

image001 png@01D4E980

This PR fixes that.

Well, I hope, because I was only able to check the compilation, using Mingw on Linux.

## Release Management


lrineau:
Actually, we should use the method proposed here: https://stackoverflow.com/a/36731183/1728537.

maxGimeno:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-5.shtml

PR #3842, opened 2019-04-04
Fix Iterator_range for Apple Clang

maxGimeno:
## Summary of Changes

Add an conversion operator to tuple into Iterator_range to satisfy clang used with c++11 and libc++

## Release Management

* Affected package(s):STL_Extension


lrineau:
It should be based on 4.13.

maxGimeno:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-5.shtml

PR #3846, opened 2019-04-09
PoC C++11 result_of

sloriot:
@mglisse proposal for the removal of `CGAL::cpp11::result_of`

@maxGimeno This one also needs to be refreshed and tested. Let me know if you have issue with the refresh.


maxGimeno:
@sloriot it's done

afabri:
In Travis I see an error in[ ` Extreme_points_traits_adapter.h` ](https://github.com/CGAL/cgal/blob/master/Convex_hull_3/include/CGAL/Extreme_points_traits_adapter_3.h#L52).    The types of the arguments of ` F`  should be `boost::property_traits<PointPropertyMap>::value_type`   instead of ` Vertex`, but that does not compile either.    
Replacing the ` result_of`   with ` declspec(auto)`  works, but it might make sense to understand which ` result_of`  would be the correct one.

lrineau:
@sloriot You said this PR is important for CGAL-5.0.

@mglisse Would you have time to have a look at the issue, and try to find the appropriate fix?


Anyway, I am a bit worried about changing that so late in the release cycle of CGAL-5.0: that change of `result_of` could introduce nasty errors, like new implicit conversions, and then maybe dandling references to temporary objects.

mglisse:
> @mglisse Would you have time to have a look at the issue, and try to find the appropriate fix?

Andreas already has the right fix (modulo the misspelling of `decltype`), I don't see the point of looking further (well, the maintainer of the package is always welcome to look further, but that's not me :wink: )

maxGimeno:
We just realized that `result_of` was removed from c++20

Sameeer12:
Hi, my name is Sameer and I'm applying for more C++ 11 project in GSoC. I stumbled upon this issue while going through the code. Intrigued by the character of the issue, I digged deeper. Here are my 2 cents on the issue:

It is true that `boost::property_trait<PointPropertyMap>::value_type` is same as what you get from `get` function. However, there seems to be an issue with `const` or `reference`, since a function return type making use of `get` works perfectly while one with `property_traits` leads to compilation error
i.e. 

```
decltype(std::declval<F>()(get(std::declval<PointPropertyMap>(),std::declval<const Vertex&>),get(std::declval<PointPropertyMap>(),std::declval<const Vertex&>))

or

std::result_of_t<F(decltype(get(std::declval<PointPropertyMap>(),std::declval<const Vertex&>)),decltype(get(std::declval<PointPropertyMap>(),std::declval<const Vertex&>)))>
```

works perfectly. While using `boost::property_traits<PointPropertyMap>::value_type` the compiler points out that there is no `type` named type which is an usual occurrence when either the callable is not invokable or the arguments types are not valid. Again, since `property_traits`'s value_type gives exactly the same type as `get`, it leads me to belive that we're doing something wrong with const references.

Finally, I see the code in master is working with `F(const Vertex&, const Vertex&)`. Given `Vertex` is completely orthogonal to the type of `boost::property_traits<PointPropertyMap>::value_type`, I can see some implicit conversions occurring throughout the CGAL code base, which should not occur.

Also, I had a curious question. Why don't you guys employ perfect forwarding in these situations to maintain the signature across the code base.

@maxGimeno About `result_of`, it is not removed from C++20 but deprecated. This situation can be dealt with by utilizing `boost::tr1_result_of` for the time being while converting the `std::result_of` to `std::invoke_result`. Also, I believe that `decltype(auto)` as a function signature is better as it returns the return type of the `return` statement (too many returns there ;) ). This will also help you guys figure out all the implicit conversions that were occurring in the code.

One last thing, I see you replace `cpp11::result_of` with `result`. What was the rationale behind this decision? Also, I see `result` as an empty templated struct in the code. What is it trying to achieve?

lrineau:
> @maxGimeno About `result_of`, it is not removed from C++20 but deprecated. 

https://en.cppreference.com/w/cpp/types/result_of

It was deprecated in C++17, and then removed from C++20.

Sameeer12:
Ohh! I didn't know. I apologize for not realizing it.

MaelRL:
Replaced by https://github.com/CGAL/cgal/pull/4878.

PR #3849, opened 2019-04-10
STL_Extension: Fix Dispatch_output_iterator test 

maxGimeno:
Add a function to_tuple() to the Dispatch_output_iterator and use it in the test to avoid a clang error on old macs


STL_Extension/include/CGAL/iterator.h
```diff
@@ -1408,6 +1408,16 @@ class Dispatch_output_iterator < std::tuple<V...>, std::tuple<O...> >
...
   }
+  
+  std::tuple<O...>& to_tuple()
+  {
+    return static_cast<std::tuple<O...>&>(*this);
+  }
+  
+  const std::tuple<O...>& to_tuple() const
+  {
+    return static_cast<std::tuple<const O...>&> (*this);
```
sloriot:
```suggestion
    return static_cast<const std::tuple<const O...>&> (*this);
```

STL_Extension/include/CGAL/iterator.h
```diff
@@ -1449,6 +1459,7 @@ class Dispatch_or_drop_output_iterator < std::tuple<V...>, std::tuple<O...> >
 
   template <class T>
   Self& operator=(const T&) { return *this; }
+
```
sloriot:
```suggestion
```

STL_Extension/include/CGAL/iterator.h
```diff
@@ -1408,6 +1408,16 @@ class Dispatch_output_iterator < std::tuple<V...>, std::tuple<O...> >
     tuple_dispatch(t);
     return *this;
   }
+  
+  std::tuple<O...>& to_tuple()
```
lrineau: This patch is strange: the user has to know one has to call `to_tuple()`. Why not renaming this `to_tuple()` into a real implicit conversion operator: `operator std::tuple<0...>&`.

maxGimeno: To make it a working operator() it would require to make it a std::tuple<Arg&...>(args...), which is exactly what doesn't work for some reason. I tried for some time to make it work but @sloriot told me to simply use a function to_tuple() as the type is easier to deal with.

afabri:
And did you try non-variadic operators?

STL_Extension/include/CGAL/iterator.h
lrineau: I am not talking about `operator()` but [conversion operator]. That is what we used in #3842 to fix `Iterator_range` with clang lib++ and `std::tie`.

[conversion operator]: https://en.cppreference.com/w/cpp/language/cast_operator

maxGimeno: Yes, but in #3842 we had a std::tuple<I,I>, here it is a std::tuple<O...>, and IIUC the problem is that the problematic compiler has troubles with the syntax XXX<O&...>

maxGimeno:
@afabri no I didn't.

STL_Extension/include/CGAL/iterator.h
maxGimeno: It is not tested with gcc, because gcc does the deduction well, and uses the std::tuple<> type inherited in the iterator class.

lrineau: Actually I understand the issue: that is difficult to write `get<0>(*this), get<1>(*this), get<2>(this*)...` in a fold expression. If you want to try, that is doable in C++14, using the tools given in that Stack Overflow answer: https://stackoverflow.com/a/45284891/1728537.

lrineau: If you test with gcc, then you can be convinced that the issue if the AppleClang compiler, and not tour code being incorrect.

maxGimeno: It is not called. get<> or no get<>, the downcast is prioritary. As long as the class derives from std::tuple, no operator std::tuple will be called with gcc. I simply cannot test with gcc.

maxGimeno:

PR #3853, opened 2019-04-15
CMakeLists.txt: Remove unnecessary include directories

afabri:
## Summary of Changes

just a cleanup

### Todo:
- [x] I see 

>   Make sure the compiler can find generated .moc files
  include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
  include_directories( ${QT_INCLUDE_DIR} )

in a few files. Either this is really needed and then we have to add it in other demos, or we can remove it.

## Release Management

* Affected package(s): demos

### DIFF
--- a/Boolean_set_operations_2/archive/demo/Boolean_set_operations_2_GraphicsView/CMakeLists.txt
+++ b/Boolean_set_operations_2/archive/demo/Boolean_set_operations_2_GraphicsView/CMakeLists.txt
@@ -15,11 +15,10 @@ set( QT_USE_QTMAIN   TRUE )
 
 find_package(Qt5 QUIET COMPONENTS Widgets Script Svg)
 
-include_directories (BEFORE ${CMAKE_CURRENT_BINARY_DIR})
 include_directories (BEFORE ${CMAKE_CURRENT_SOURCE_DIR}/include )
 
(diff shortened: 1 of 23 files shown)


afabri:
@maxGimeno you are more CMake-savy than me.   Are the include_directories mentioned in the Todo needed?

maxGimeno:
I don't think so, but I'm not sure. Well, for recent cmake syntax ( the CGAL::CGAL stuff) I'm pretty sure it is not needed, but before that I don't know

afabri:
@maxGimeno What is the recommended destination directory for moc files, when we run the command `qt_5_generate_moc()`.  In some demo CMakeLists.txt its the source dir and in others the binary dir.

maxGimeno:
it should go in the binary dir. I even consider it an error to put it in the source dir.

afabri:
@maxGimeno more to do for you

afabri:
@maxGimeno unless there is more to do from your side  I would say Succesfully tested in [5.0-Ic-23](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-23.shtml)

maxGimeno:
I confirm.

sloriot:
What about the TODO filed in the description?

maxGimeno:
it is done

maxGimeno:
I think it is done. (That is what 62f0a73 does IIRC)

PR #3854, opened 2019-04-15
Fix forwarding of projection traits

sloriot:


maxGimeno:

PR #3858, opened 2019-04-17
BGL: Fix is_valid_polygon_mesh not breaking away from potential infinite loops

MaelRL:
If the mesh being tested is actually invalid, infinite loops are possible when walking around vertices (for example). Loops must thus break after a while. Added a few more tiny tests too (number of edges vs number of halfedges).

Also reduced verbosity (don't care about things that go well), and improved indentation to make it readable.

## Release Management

* Affected package(s): `BGL`

### DIFF
--- a/BGL/include/CGAL/boost/graph/helpers.h
+++ b/BGL/include/CGAL/boost/graph/helpers.h
@@ -359,152 +359,174 @@ bool is_valid_face_descriptor( typename boost::graph_traits<FaceGraph>::face_des
 /*!
   \ingroup PkgBGLHelperFct
  * \brief checks the integrity of `g`.
- * 
+ *
  * `g` is valid if it follows the rules of the `HalfedgeListGraph` concept,
  * and all of its associations are reciprocal.
- * For example, `prev(next(h, g), g)` must be `h`, 
+ * For example, `prev(next(h, g), g)` must be `h`,
  * and `next(prev(h, g), g)` must be `h`.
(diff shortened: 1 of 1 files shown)


BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -514,134 +536,159 @@ return valid;
...
-   if ( ! valid) {
-       verr << "    incident facet is not at least a triangle."
-            << std::endl;
-       break;
-   }
- }
- return valid;
-}
+  bool valid = is_valid_face_graph(g, verb);
+  if(!valid)
```
sloriot: Is this the only fix in the patch? putting all in the same commit makes it hard to spot the real fixes.

MaelRL: https://github.com/CGAL/cgal/pull/3858/files?utf8=%E2%9C%93&diff=split&w=1 and it's pretty clear. What you're quoting is just an early exit. What was broken is not having a break in loops such as [here](https://github.com/CGAL/cgal/pull/3858/files#diff-dbc8316e37114ea33573d304bd560faaR482).

sloriot: That's what I did and I can't agree. Having two commits, one with the extra breaks and one with the cosmetic changes would have made the change obvious. Thanks for pointing the line.

lrineau:
The corresponding PR #3916 was tested in `integration`.

PR #3862, opened 2019-04-18
BGL: Fix `Face_filtered_graph`'s `is_selection_valid()`

MaelRL:
## Summary of Changes

Mainly, the walking was wrong:
```
  hd = next(opposite(hd, _graph), _graph);
```
does not walk around `target(hd, _graph)` (`next` and `opposite` should be swapped). It was also missing some manifoldness test and was overall too complicated compared to what's required, so I just rewrote it to be closer to PMP's `is_non_manifold(vertex_descriptor, graph)`.

Also some tests were added.

## Release Management

* Affected package(s): `PMP`


maxGimeno:
Conflicts with master

MaelRL:
The conflict only comes from `BOOST_FOREACH(,)` ---> `for(:)`, and it will show up in a lot of bug fixes unfortunately. You can just overwrite whatever master has, here. Or do you prefer another PR for master?

maxGimeno:
I'd rather have a second PR, as i messed everything up last time I put it in integration...

PR #3878, opened 2019-04-24
Fix for less

sloriot:
From @mhsaar 

Some STL implementations may call comparator(x, x)
to verify irreflexivity.  Don't violate less_than_in_tree's
in such an environment.

### DIFF
--- a/Generator/include/CGAL/Random_polygon_2_sweep.h
+++ b/Generator/include/CGAL/Random_polygon_2_sweep.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2001  
+// Copyright (c) 2001
 // Utrecht University (The Netherlands),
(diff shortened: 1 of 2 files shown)


afabri:
Again, why this, and why this only at two spots?

sloriot:
`less_than_in_tree()` has the following preconditions:
```
    CGAL_polygon_precondition(
       m_vertex_data->edges[tree_edge.as_int()].is_in_tree);
    CGAL_polygon_precondition(
       !m_vertex_data->edges[new_edge.as_int()].is_in_tree);
```
so as stated in the comment, it called with `x`, x` it will break. `Less_segments` is called by `std::set` so it is STL implementation dependent.

maxGimeno:

PR #3879, opened 2019-04-24
Polyhedron Demo: fixes for multiviewer

### LINKED ISSUE #3875: CGAL demo issues batch
### Bugs ###

- [x] indices only show up on the first viewer (other viewers say "too many elements")
- [x] the selection highlighting is not working properly: open a selection item, select some face, move to another viewer, the highlighting does not correspond to the cursor position
- [x] lcc being the default plugin to open .off meshes is unpractical because no operation is available
- [x] create clipping box crashes the demo
- [x] "Lighting..." doesn't do anything
- [x] Shouldn't be able to call functions if the input mesh does not meet preconditions, e.g. calling "does_self_intersect" on a quad mesh will return that everything is fine because the mesh is not triangulated, regardless of the geometry.
- [x] Filter items and shift+click do not play well together (segfaults). I can't really reproduce it, but maybe if you shift+click between filtered items...
- [x] If you try to offset mesh an item that is hidden, you can't change the values and the automatically computed suggestions are 0 (I imagine this is not restricted to offset meshing).
- [x] Failing to reload an item crashes the demo (take an .off, load it, change the number of a vertices to 1 fewer in the file, reload, kaputt).
- [x] recenter applies to all viewers (might be a feature?)
- [x] Camera twitching within 1s after a recenter is still present
- [x] Can't load polygon soups
- [x] Selection highlight doesn't move with the cursor when shift+clicking a selection
- [x] Selection highlight is wrong when the brush size is greater than a single simplex (ideally the selection highlight would be adapted to the brush's size)
- [x] Can't open OBJ
### Feature Requests ###
- [x] Nicer palette colors for some plugins. For example in VSA, the colors are objectively not aesthetic. Rather than purely random, which yields all type of ugly pinks and browns, it would be nicer to use some restricted color palette.

## Environment

* CGAL version: master

maxGimeno:
## Summary of Changes
Some fixes.
Among them, a change of color generation : 
before
befor

after:

after

## Release Management

* Issue(s) solved (if any): fix #3875


Polyhedron/demo/Polyhedron/resources/solid_wireframe_shader.geom
```diff
@@ -27,7 +27,7 @@ void main(void)
   //linearized arbitrary Z offset to keep the widelines in front of the edges as much as possible 
   //Problem if coef is not 0, edges might be visible through faces.
   float coef = 0.0f;
-  float z_offset = (2.0 * near) / (far + near - coef * (far - near)) - (2.0 * near) / (far + near);
+  float z_offset = 0;//(2.0 * near) / (far + near - coef * (far - near)) - (2.0 * near) / (far + near);
```
lrineau: Each time I see a modification to the OpenGL offset of polylines, the modification seems random. What is the rational for the modification?

maxGimeno: the result of that formula should be 0 with that coef, but it isn't for some reason. I replaced it with a real 0 and it fixes the bug. There shouldn't be any offset there, from an earlier decision.

maxGimeno:

lrineau:
@maxGimeno Maxime, you set the label "work in progress". Let me know if it can be removed.

maxGimeno:
Sorry, i forgot to remove it

PR #3880, opened 2019-04-24
Change return type to avoid raising a precondition

sloriot:
Return `false` instead of raising an exception.


maxGimeno:

PR #3884, opened 2019-04-26
BGL: Fix `Face_filtered_graph`'s `is_selection_valid()` (5.0)

MaelRL:
See #3862.

Compared to the other PR, commit e17f10f (deletes a useless data file) is not present there, because it will not create any conflict anyway.


maxGimeno:

PR #3893, opened 2019-04-30
Fix the prev/next pointers after polyline removal

sloriot:


lrineau:

PR #3899, opened 2019-05-03
Scripts: Update create_package_dir

afabri:
## Summary of Changes

Cleanup, as it currently creates a little bit too much files and directories.

## Release Management

* Affected package(s): developer_scripts


lrineau:
I do not see what Travis or our testsuite can test. So I will merged this PR as it is, without testing. Andreas has probably tested it locally.

PR #3912, opened 2019-05-07
Simplify user scripts (cgal_create_CMakeLists and cgal_create_cmake_script)

lrineau:
## Summary of Changes

This pull-request simplifies and updates the user scripts `cgal_create_CMakeLists` and `cgal_create_cmake_script` for CGAL-5.0.

## Release Management

@afabri, @maxGimeno: I do not know how to test those scripts automatically. Maybe our Travis scripts could call them somewhere in a simple directory, and verify that one can compile the targets.


maxGimeno:
>Maybe our Travis scripts could call them somewhere in a simple directory, and verify that one can compile the targets.

Like re-creating some simple CGAL tests with the scripts and have travis test them ?

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [May 7, 2019, 1:19 PM GMT+2](https://github.com/CGAL/cgal/pull/3912#issuecomment-490037899 "2019-05-07T11:19:58Z - Replied by Github Reply Comments"):
> >Maybe our Travis scripts could call them somewhere in a simple directory, and verify that one can compile the targets.
> 
> Like re-creating some simple CGAL tests with the scripts and have travis test them ?

Exactly. But I think the script should also check that the targets have been created (after CMake). Because it is possible that the run of the `cgal_create_cmake_script` (or the other) creates a valid `CMakeLists.txt` but without any target in it.

maxGimeno:
That's why I mentioned re-creating existing tests, so that we can compare the results. If the results differ, then there is a problem. We could also just create tests from scratch, knowing the expected output though. And travis would compare the returned value of the test with the expected result. This way, if the target in not runnable, the there will be an error anyway. And if the result is wrong, there is also a problem.

Scripts/scripts/cgal_create_CMakeLists
```diff
@@ -161,120 +153,35 @@ fi
...
+  if [ ! -z "$BOOST_COMPONENTS" ]; then
+    cat << 'EOF'
 
 # Boost and its components
 EOF
-  #---------------------------------------------------------------------------
-
-  if [ ! -z "$BOOST_COMPONENTS" ]; then
-
     echo "find_package( Boost REQUIRED COMPONENTS ${BOOST_COMPONENTS//:/ } )"
```
maxGimeno: Shouldn't it be Boost QUIET COMPONENTS  here ?

lrineau: Boost headers are required for CGAL. But components can be optional. Now that we require CMake 3.1, it could be:
```shell
echo "find_package( Boost REQUIRED OPTIONAL_COMPONENTS ${BOOST_COMPONENTS//:/ })"
```

lrineau: I pushed that modification.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [May 7, 2019, 1:49 PM GMT+2](https://github.com/CGAL/cgal/pull/3912#issuecomment-490046593 "2019-05-07T11:49:45Z - Replied by Github Reply Comments"):
> That's why I mentioned re-creating existing tests, so that we can compare the results. If the results differ, then there is a problem. We could also just create tests from scratch, knowing the expected output though. And travis would compare the returned value of the test with the expected result. This way, if the target in not runnable, the there will be an error anyway. And if the result is wrong, there is also a problem.

Actually, our `CMakeLists.txt` are not generated with those scripts. Maybe they were, long time ago, but no longer.

What I suggest it recreating the `CMakeLists.txt` of a directory, and then with `make help` check if the list of targets contains the targets that are expected.

maxGimeno:
The branch has conflicts with master

lrineau:
I have pushed a merge, to solve the conflicts.

maxGimeno:
You removed the find_package(openGL) from the 3D demo, but you left the condition OPENGL_FOUND in the end, deactivating the demo for all platforms.

lrineau:
I just fixed that in https://github.com/CGAL/cgal/pull/3912/commits/4af7e7dcdca1002d05521430dd2916abdd767514.

maxGimeno:

PR #3916, opened 2019-05-08
BGL: Fix is valid polygon mesh (5.0)

MaelRL:
See #3858.


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-36.shtml.

PR #3930, opened 2019-05-14
T2: Fix insertion of range of constraints

sloriot:
Fix initialization of the vector of indices fill directly instead of resizing it and inserting back...


maxGimeno:
Seems ok in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-39.shtml

PR #3935, opened 2019-05-15
[Classification] Add missing "algorithm" include

sgiraudot:
## Summary of Changes

Missing `#include <algorithm>`.

The incriminated file moved and changed quite a bit in `CGAL 4.14`, that's why I don't do the fix for `4.13` (to avoid wasting time with conflicts for a minor bug).

## Release Management


lrineau:
I think I still have to merge that PR into CGAL `master`. My fault.

Issue #2791, opened 2018-02-04
FPU: Support platforms without round-towards-infinity


mglisse:
The platform I have in mind is WebAssembly, which does not support non-default rounding modes. This platform may still be worth targeting, for small examples or demos where performance is not that important (or for applications where static filters are sufficient).

* First idea would be a software emulation of round-towards-infinity, or more roughly replace a +up b with nextafter(a +near b, +infinity). Looks doable, nextafter is not so cheap but probably still better than full rationals.
* Second idea is to disable intervals. Note that static filters are still ok (they use round-to-nearest), so we would likely want Epeck to be Simple_cartesian<rational> with static filters on top, and Epick predicates would try static filters and then go directly to the exact computation.
* I am not sure how well Boost supports it for now, which may cause other issues. GMP/MPFR also require unofficial patches IIRC, although the branch Number_types-boost_mp-glisse makes them not-so-necessary.

lrineau:
@mglisse, @horasio Will you be at next developers meeting? That could be an opportunity to start and work on that topic.

mglisse:
I am unlikely to be at the next meeting.

sloriot:
@mglisse did you already start a branch for option 1, I'd be curious to see the slow down in practice.

mglisse:
@sloriot no, I haven't experimented with anything like that. For a quick try, I guess you could just redefine CGAL_IA_ADD, etc.

lrineau:
LLVM 8, just released, has now an official WebAssembly target:
http://releases.llvm.org/8.0.0/docs/ReleaseNotes.html#changes-to-the-webassembly-target

pentacular:
I'm taking a look at option 1.

But I'm not sure I understand how Interval_nt.h is supposed to be working.

I see
```
  compare (const Interval_nt<Protected> & d, const Interval_nt<Protected> & e)
  {
    if (d.inf() > e.sup()) return LARGER;
    if (e.inf() > d.sup()) return SMALLER;
    if (e.inf() == d.sup() && d.inf() == e.sup()) return EQUAL;
    return Uncertain<Comparison_result>::indeterminate();
  }
```

These looks safe for round-to-nearest to me (please tell me if I am wrong)
```
    if (d.inf() > e.sup()) return LARGER;
    if (e.inf() > d.sup()) return SMALLER;
```

But I don't understand
```
    if (e.inf() == d.sup() && d.inf() == e.sup()) return EQUAL;
```

In is_same I see:
```
[...]

PR #3939, opened 2019-05-16
Fixes for Convex_hull_3: use `is_certain` to avoid exceptions throwing

lrineau:
## Summary of Changes

Add an alternative to `Epick` without the dynamic filters: only the static filters are applied before the exact computation.

## Release Management

* Affected package(s): Kernel_23m Convex_hull_3

### DIFF
--- a/Kernel_23/test/Kernel_23/Simple_cartesian.cpp
+++ b/Kernel_23/test/Kernel_23/Simple_cartesian.cpp
@@ -22,7 +22,6 @@
  
 
 #include <CGAL/Simple_cartesian.h>
-#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
 #include <CGAL/Quotient.h>
 #include <cassert>
 
@@ -91,7 +90,5 @@ main()
 
   typedef   CGAL::Simple_cartesian<CGAL::Quotient<Precise_integer> >     Cls;
   test_kernel("Simple_cartesian<Quotient<Precise_integer>>", Cls());
-  typedef   CGAL::Simple_cartesian<CGAL::Quotient<Precise_integer> >     Cls;
-  test_kernel("Epick_without_intervals", CGAL::Epick_without_intervals());
   return 0;
 }
--- a/Kernel_23/test/Kernel_23/Simple_cartesian.cpp
+++ b/Kernel_23/test/Kernel_23/Simple_cartesian.cpp
@@ -22,6 +22,7 @@
  
 
 #include <CGAL/Simple_cartesian.h>
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
 #include <CGAL/Quotient.h>
 #include <cassert>
 
@@ -46,30 +47,26 @@
 
 #include "CGAL/_test_mf_plane_3_to_2d.h"
 
-int
-main()
-{
-  typedef   CGAL::Simple_cartesian<double>     Clsd;
-  std::cout << "Testing IO with Simple_cartesian<double> :" << std::endl;
-  _test_io( Clsd() );
+#include <string>
 
-  typedef   CGAL::Simple_cartesian<CGAL::Quotient<Precise_integer> >     Cls;
-  std::cout << "Testing 2d with Simple_cartesian<Quotient<Precise_integer>> :";
+template <typename Cls>
+void test_kernel(std::string kernel_name, Cls) {
+  std::cout << "Testing 2d with "+kernel_name+" :";
   std::cout << std::endl;
   _test_2( Cls() );
 
-  std::cout << "Testing 3d with Simple_cartesian<Quotient<Precise_integer>> :";
+  std::cout << "Testing 3d with "+kernel_name+" :";
   std::cout << std::endl;
   _test_3( Cls() );
 
-  std::cout << "Testing new 2d with Simple_cartesian<Quotient<Precise_integer>>:";
+  std::cout << "Testing new 2d with "+kernel_name+" :";
   std::cout << std::endl;
   test_new_2( Cls() );
-  std::cout << "Testing new 3d with Simple_cartesian<Quotient<Precise_integer>>:";
+  std::cout << "Testing new 3d with "+kernel_name+" :";
   std::cout << std::endl;
   test_new_3( Cls() );
 
-  std::cout << "Testing new parts with Simple_cartesian<Quotient<Precise_integer>> :";
+  std::cout << "Testing new parts with "+kernel_name+" :";
   std::cout << std::endl;
   _test_orientation_and_bounded_side( Cls() );
   _test_fct_points_implicit_sphere( Cls() );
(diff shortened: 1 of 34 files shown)


lrineau:
@afabri Now you should bench it with the real use case.

mglisse:
Related to #2791 .

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [May 17, 2019, 12:44 PM GMT+2](https://github.com/CGAL/cgal/pull/3939#issuecomment-493410257 "2019-05-17T10:44:57Z - Replied by Github Reply Comments"):
> Related to #2791 .

I agree. It might be the right moment for us to study the various alternatives to `Interval_nt`.

Kernel_23/test/Kernel_23/Simple_cartesian.cpp
```diff
@@ -80,9 +77,21 @@ main()
...
+main()
+{
+  typedef   CGAL::Simple_cartesian<double>     Clsd;
+  std::cout << "Testing IO with Simple_cartesian<double> :" << std::endl;
+  _test_io( Clsd() );
+
+  typedef   CGAL::Simple_cartesian<CGAL::Quotient<Precise_integer> >     Cls;
+  test_kernel("Simple_cartesian<Quotient<Precise_integer>>", Cls());
+  typedef   CGAL::Simple_cartesian<CGAL::Quotient<Precise_integer> >     Cls;
+  test_kernel("Epick_without_intervals", CGAL::Epick_without_intervals());
```
maxGimeno: fails.

lrineau: Indeed the tests are completely wrong. We have nothing in CGAL to test correctly `CGAL::Epick`!

lrineau: I have eventually managed to set up a testsuite for `CGAL::Epick`. And then also for `CGAL::Epick_without_intervals`, in https://github.com/CGAL/cgal/pull/3939/commits/7d543d5bf80ae5fe3d0075d52b9750e3a0115fbc.

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-58.shtml#Kernel_23

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Jun 14, 2019, 10:14 AM GMT+2](https://github.com/CGAL/cgal/pull/3939#issuecomment-502014695 "2019-06-14T08:14:02Z - Replied by Github Reply Comments"):
> Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-58.shtml#Kernel_23

Whouah. It seems that instantiating the kernel test-suite with `Epick` instead of just `Epeck` triggered a lot of new warnings, about uninitialized variables. I was able to reproduce on my machine, and think I have fixed all of them, in https://github.com/CGAL/cgal/pull/3939/commits/30b25a4d58684681e04d21fffa3143dc383d9763.

maxGimeno:
I don't really know if it is related to this PR or if it was simply revealed by it, but https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-59/Kernel_23/TestReport_Friedrich_Ubuntu-gcc-7.gz

lrineau:
I have pushed commit 34a9b246989, that might fix the remaining warnings.

lrineau:

mglisse:
Can you explain what this PR was about? I see 2 independent things:

1. a tweak to the ad hoc, Epeck-like caching in convex hull traits
2. a new kernel, without doc, only used in a couple places in the testsuite, without explanation of what the purpose is. Is `Epick_without_interval` only meant for wider coverage in the testsuite? Was it added to speed up something?

(to clarify, I am not protesting or asking for changes or anything)

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Jun 20, 2019, 3:17 PM GMT+2](https://github.com/CGAL/cgal/pull/3939#issuecomment-504022165 "2019-06-20T13:17:54Z - Replied by Github Reply Comments"):
> Can you explain what this PR was about? I see 2 independent things:
> 
> 1. a tweak to the ad hoc, Epeck-like caching in convex hull traits
> 2. a new kernel, without doc, only used in a couple places in the testsuite, without explanation of what the purpose is. Is `Epick_without_interval` only meant for wider coverage in the testsuite? Was it added to speed up something?
> 
> (to clarify, I am not protesting or asking for changes or anything)

Actually there are three different features:
- add a testsuite for `Epick`,
- add a new kernel `Epick_without_interval`, rather useless actually,
- fix of the convex hull traits in 3D, to remove the throw of `Uncertain_conversion_exception`.

The history of the development is actually quite different:
1. A user complained that Convex_hull_3 with Epick throws a lot on degenerated data set (the actual complain was "CGAL is dramatically slow in the debugger, even in Release").
2. That led to the creation of `Epick_without_intervals`, to try benchmarks.
3. Then we found out why `Convex_hull_3` was throwing that much on degenerated data sets (and in passing we found the bug fixed by #3941): the ad-hoc code in `Convex_hull_traits_3` for `Epick` was not using `is_certain` to check the validity of the computation with `Interval_nt` but only relied on the try/catch. Thus the small fix: use `is_certain`, see https://github.com/CGAL/cgal/pull/3939/files#diff-05e072598bf144f1bfdb9dc3ca0c24daR358.
4. After that I found out that there was ways to improve the ad-hoc solution in `Convex_hull_traits_3`, and thus the bigger fix in `Is_on_positive_side_of_plane_3`, see https://github.com/CGAL/cgal/pull/3939/files#diff-05e072598bf144f1bfdb9dc3ca0c24daR248.
5. A few days after, I was trying to cleanup the branch, and wondered if I should keep the strange kernel `Epick_without_intervals`, that was actually useless for the user complaining about the exceptions, after the better fix we found. I have tried to "test" it... and I found out that there was no testsuite for `Epick`!!
6. I have tweaked the testsuite for `Epeck`, by somehow disabling the `assert` involving inexact constructions. That is the bigger set of changes, all having a lot of `|| inexact` in the asserted expressions.
7. Then I have decided to keep `Epick_without_intervals`, in case we might want it (maybe for issue #2791).

mglisse:
Thank you, it all makes sense now.
For #2791, I think I'd rather have Epick itself not use intervals on platforms where they don't exist.
I'll have to check how Epick_without_intervals works. AFAICS, it indirectly derives from Filtered_kernel_base so I would have expected it to still use intervals (basically you hard-coded that it does use static filters without defining Has_static_filters).

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Jun 20, 2019, 4:01 PM GMT+2](https://github.com/CGAL/cgal/pull/3939#issuecomment-504037526 "2019-06-20T14:01:13Z - Replied by Github Reply Comments"):
> I'll have to check how Epick_without_intervals works. AFAICS, it indirectly derives from Filtered_kernel_base so I would have expected it to still use intervals (basically you hard-coded that it does use static filters without defining Has_static_filters).

You are right. Probably that kernel does not even avoid `Interval_nt`.

afabri:
> You are right. Probably that kernel does not even avoid `Interval_nt`.

I confirm experimentally that it still uses `Interval_nt`.

MaelRL:
https://github.com/MaelRL/cgal/commit/6dceaaea4e70efaac1bf6d277025ae6e9efb9d70 would fix that.

However, `EPICK_without_intervals` would then not provide exact predicates since its base is going to be `Simple_cartesian<double>`. It seems to me that the static filters ought to have two template parameters : one kernel that it evaluates the value with (like `Simple_cartesian<double>`), and then an another kernel, which it defaults to if it cannot conclude from comparison value. That kernel then probably would be exact, either because it's a filtered kernel, or because it's using an exact number type. 

Finally, `EPICK_without_intervals` would be simply something like `static_filters<simple_cartesian<double>, simple_cartesian<exact_number_type>`.

mglisse:
In the mean time, for the imminent release, it seems best to remove it, lest someone get confused by its name.

PR #3943, opened 2019-05-17
Misc fixes for expression templates

mglisse:
## Summary of Changes

There hasn't been much testing of `Simple_cartesian<mpq_class>` as an exact kernel...

Since this touches a number of places, you are welcome to take a look and shout if you notice anything suspicious.

## Release Management

* Affected package(s): many

### DIFF
--- a/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Mean_value_2.h
+++ b/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Mean_value_2.h
@@ -336,8 +336,8 @@ private:
         // Following section 4.2 from [2] we denote P_j = r_j*r_{j+1} + dot_product(d_j, d_{j+1}).
         // Vector s_i from [1] corresponds to that one with the name d_i in [2].
         for(int j = 0; j < n-1; ++j)
-            P[j] = (CGAL::max)(r[j]*r[j+1] + scalar_product_2(s[j], s[j+1]), FT(0));
-        P[n-1] = (CGAL::max)(r[n-1]*r[0] + scalar_product_2(s[n-1], s[0]), FT(0));
+            P[j] = CGAL::max<FT>(r[j]*r[j+1] + scalar_product_2(s[j], s[j+1]), 0);
+        P[n-1] = CGAL::max<FT>(r[n-1]*r[0] + scalar_product_2(s[n-1], s[0]), 0);
 
         // Compute mean value weights using the formula (16) from [2].
         // Since the formula (16) always gives positive values, we have to add a proper sign to all the weight functions.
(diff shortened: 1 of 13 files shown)


mglisse:
If I read the report correctly, it is complaining that because I am including a file from Kernel_23 in Interpolation, the package now depends on Interval_support and Modular_arithmetic. That's all GF internal stuff that I'll let you handle.

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [May 20, 2019, 12:12 AM GMT+2](https://github.com/CGAL/cgal/pull/3943#issuecomment-493796345 "2019-05-19T22:12:16Z - Replied by Github Reply Comments"):
> If I read the report correctly, it is complaining that because I am including a file from Kernel_23 in Interpolation, the package now depends on Interval_support and Modular_arithmetic. That's all GF internal stuff that I'll let you handle.

I'll deal with that.

afabri:
> That's all GF internal stuff that I'll let you handle.

@mglisse it is not GF internal stuff, but a long term goal towards modularity, and hence a dependency tracking.

mglisse:
> modularity

Do any of the opensource users care about that? It is quite possible that they do, Boost also insists on going modular although in their case they don't even have the licensing business as a reason and that always mystified me. So I am aware that my understanding if very limited here and I'll take your word for it.

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [May 20, 2019, 10:06 AM GMT+2](https://github.com/CGAL/cgal/pull/3943#issuecomment-493881547 "2019-05-20T08:06:44Z - Replied by Github Reply Comments"):
> > modularity
> 
> Do any of the opensource users care about that? It is quite possible that they do, Boost also insists on going modular although in their case they don't even have the licensing business as a reason and that always mystified me. So I am aware that my understanding if very limited here and I'll take your word for it.

I think the CGAL project cares. The Editorial Board, at least.

There are a lot of critics against CGAL, and some of them are easy to fix.

- hard to install -> now as much header-only as possible,
- hard to learn... well, we try to improve the user manual,
- a huge beast, compared to "my little need" (say "triangulations").

For the last one, what we envisioned is a set of modules, header-only, instead of that huge library. I have started that mouvement alone, several years ago, at the same time I worked on:
  - using a non-ancient CMake syntax,
  - moving to the adoption of C++11 in CGAL,
  - making CGAL header-only,
  - remove the dependency to Boost binaries (and use only Boost as header-only when possible).

Among all those work-in-progress, the only one that is not yet achieved is the modularity, but I must say I am still waiting for help from Geometryfactory. At the beginning, I planned the modularity for CGAL-4.11... and now we know it will be 5.1 at best.


Do the opensource user care about all that? I think my answer was convincing! I can even say that only the open source users really care about that, because the commercial users have the resources to deal with big libraries with hard requirements.

lrineau:
Travis is fine with this PR now. @maxGimeno, please add it in your set of PR to-be-tested in `integration`.

maxGimeno:

PR #3949, opened 2019-05-20
Fix CTest CGAL tests when binary dir is a sub-directory

### LINKED ISSUE #2865: CTest should not use the SetupFixture when the CMake binary dir is in source
With @afabri setup, the build directory is, for example, `examples/Mesh_3/VC12/`. The `SetupFixture` step of CTest does not work. Anyway I think that we should run tests in-sources, if the build directory is in-sources.

TODO for next meeting.

lrineau:
## Summary of Changes

When the current binary dir is a sub-directory of the current sources
dir, then the recursive copy of the sources dir to `__exec_test_dir/`
cannot be done. In that case, do not setup the fixture
directory. Tests can still be run in the source directory (and pollute
it).

@afabri: That is an old bug that prevented you from using CTest. You reported it more than one year ago. It will be soon fixed in release branches and `master`.

## Release Management

* Affected package(s): CMake
* Issue(s) solved (if any): fix #2865

### DIFF
--- a/Installation/cmake/modules/CGAL_add_test.cmake
+++ b/Installation/cmake/modules/CGAL_add_test.cmake
@@ -105,7 +105,12 @@ function(cgal_setup_test_properties test_name)
       APPEND PROPERTY DEPENDS "compilation_of__${exe_name}")
   endif()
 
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #3961, opened 2019-05-27
[Classification] Make cluster lightweight to copy + fix bbox computation

sgiraudot:
## Summary of Changes

Because they use vectors internally (that can grow big), clusters are heavy to copy. I fixed this by using shared pointers (instantiated only when explicitly constructed).

There was also a bug in the computation of the bbox: it was the full bounding box of all items no matter what the actual cluster was. This is also fixed.

## Release Management


maxGimeno:
Errors in [the 3D demo](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-57.shtml#Polyhedron_Demo).

maxGimeno:

PR #3964, opened 2019-05-29
Fix invalid index

sloriot:
The bug was hard to spot because in most cases the intersection point was recomputed instead of using the existing value.


sloriot:
It will conflict with master but it boils down to replacing cpp11 with std

maxGimeno:

Issue #1116, opened 2016-05-28
Using boost::hash with Point


freud14:
Hi,

I don't know if this is the best place to post this as it is more of a feature request (that I'm willing to program) than an issue but here we go.

I would like to be able to use a boost unordered_map (or STL's) but the boost::hash traits is not implemented in CGAL. I was thinking of implementing it but I don't know how to deal with the different numeric types. I was thinking to use the CGAL::to_double function to convert the field type of the point to double and then passing each coordinate to boost::combine_hash to get the final hash. 

The only problem with this method is the lost of precision with double. However, it will respect the fact that if two points are equal, then the hash will be the same. 

What do you think?

Thank you.

Frédérik

mglisse:
I am not convinced with using to_double generically. I believe you should rely on the number type being hashable, and hash_combine the lot. This would already work for double (and thus Epick::Point_2). It would then move the issue to hashing other number types, which is relevant as well. We might decide to use to_double on a case by case basis then (we may need CGAL_IA_FORCE_TO_DOUBLE to guarantee consistent results). Note that for intervals, hashing seems quite hard...

freud14:
Thank you for your answer and sorry for the late answer. So, a general implementation with hash_combine would work for double and would later need to implement hashing functions for the other number types. I do not know what is CGAL_IA_FORCE_TO_DOUBLE and how intervals are used in CGAL but I suppose that it is not needed for a general hashing function. 

The only other problem I see is if the hashing should take into account if the point is in homogenous coordinates or Cartesian coordinates or just hashing the Cartesian coordinates would work without taking into account the coordinates type of the point.

Thank you.

mglisse:
I don't think we can hash homogeneous points, since we don't canonicalize their representation. If the 2D points (1,2,3) and (2,4,6) are the same, they should have the same hash, which seems hard to guarantee. Asking for the corresponding Cartesian coordinates would in most cases use type Quotient, and we would need to make sure that Quotient(1,3) and Quotient(2,6) have the same hash, which means canonicalizing them. But that's not part of the concept for RT. Converting to double as you suggested might be the only general way, but we don't guarantee enough precision for it to really work.

[...]

PR #3973, opened 2019-06-03
[Small Feature] Add hash functions for kernel objects

sgiraudot:
## Rationale

With CGAL supporting C++11/14, users should be able to use some CGAL Kernel objects with `std::unordered_set` and `std::unordered_map` without too much trouble (points, segments, etc.).

All these objects already have `operator==()`, but to be usable with the unordered structure, they also need to come with a specialization of `std::hash`: this is what this PR does.

## Summary of API changes

The following Kernel objects are given a partial specialization of `std::hash`, in the cases where `Kernel::FT` is a floating-point type:

- `Aff_transformation_2.h`
- `Aff_transformation_3.h`
- `Bbox_2.h`
- `Bbox_3.h`
- `Circle_2.h`
- `Iso_cuboid_3.h`
- `Iso_rectangle_2.h`
- `Point_2.h`
- `Point_3.h`
- `Segment_2.h`
- `Segment_3.h`
- `Sphere_3.h`
- `Vector_2.h`
- `Vector_3.h`
- `Weighted_point_2.h`
- `Weighted_point_3.h`

Objects such as lines, planes or triangles are not handled because comparing the values of their internal numbers is not enough to test equality, which means that it's complicated to create a hash function that ensures that two equal objects are given the same hash.

The following line is added to the documentation of all the previous classes:

```
\cgalModels `Hashable` if `Kernel` is a cartesian kernel and if `Kernel::FT` is `Hashable`
```

With the exception of `Bbox_2` and `Bbox_3` which are simply documented as:

```
\cgalModels `Hashable`
```

as they are not templated by a Kernel and are always hashable.

LGPL GeometryFactory

## CHANGES.md

```markdown
### 2D and 3D Linear Geometry Kernel
 - The following objects are now hashable and thus trivially usable
   with `std::unordered_set` and `std::unordered_map`:
   `CGAL::Aff_transformation_2`, `CGAL::Aff_transformation_3`,
   `CGAL::Bbox_2`, `CGAL::Bbox_3`, `CGAL::Circle_2`,
   `CGAL::Iso_cuboid_3`, `CGAL::Iso_rectangle_2`, `CGAL::Point_2`,
   `CGAL::Point_3`, `CGAL::Segment_2`, `CGAL::Segment_3`,
   `CGAL::Sphere_3`, `CGAL::Vector_2`, `CGAL::Vector_3`,
   `CGAL::Weighted_point_2` and `CGAL::Weighted_point_3`.
```

## Submission

* [__Updated Reference Manual__](https://cgal.geometryfactory.com/~sgiraudot/Kernel_hash_functions/Kernel_23/group__PkgKernel23Ref.html)

## Status

- [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_hash_functions_for_kernel_objects)
- pre-approuved by @lrineau 2019/06/04

Partial fix of #1116.

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -45,6 +45,16 @@ Release date: September 2019
    the approximate dihedral angle between 2 vectors. Corresponding functors
    in the model (`Compute_approximate_angle_3`) and free function (`approximate_angle`)
    are also added.
+ - The following objects are now hashable and thus trivially usable
+   with `std::unordered_set` and `std::unordered_map`:
+   `CGAL::Aff_transformation_2`, `CGAL::Aff_transformation_3`,
+   `CGAL::Bbox_2`, `CGAL::Bbox_3`, `CGAL::Circle_2`,
+   `CGAL::Direction_2`, `CGAL::Direction_3`,
+   `CGAL::Iso_cuboid_3`, `CGAL::Iso_rectangle_2`,
+   `CGAL::Point_2`, `CGAL::Point_3`, `CGAL::Ray_2`,
+   `CGAL::Ray_3`, `CGAL::Segment_2`, `CGAL::Segment_3`,
+   `CGAL::Vector_2`, `CGAL::Vector_3`, `CGAL::Weighted_point_2`
+   and `CGAL::Weighted_point_3`.
 
 ### IO Streams
 
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -49,12 +49,10 @@ Release date: September 2019
    with `std::unordered_set` and `std::unordered_map`:
    `CGAL::Aff_transformation_2`, `CGAL::Aff_transformation_3`,
    `CGAL::Bbox_2`, `CGAL::Bbox_3`, `CGAL::Circle_2`,
-   `CGAL::Direction_2`, `CGAL::Direction_3`,
-   `CGAL::Iso_cuboid_3`, `CGAL::Iso_rectangle_2`,
-   `CGAL::Point_2`, `CGAL::Point_3`, `CGAL::Ray_2`,
-   `CGAL::Ray_3`, `CGAL::Segment_2`, `CGAL::Segment_3`,
-   `CGAL::Vector_2`, `CGAL::Vector_3`, `CGAL::Weighted_point_2`
-   and `CGAL::Weighted_point_3`.
+   `CGAL::Iso_cuboid_3`, `CGAL::Iso_rectangle_2`, `CGAL::Point_2`,
+   `CGAL::Point_3`, `CGAL::Segment_2`, `CGAL::Segment_3`,
+   `CGAL::Vector_2`, `CGAL::Vector_3`, `CGAL::Weighted_point_2` and
+   `CGAL::Weighted_point_3`.
 
 ### IO Streams
 
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -51,8 +51,8 @@ Release date: September 2019
    `CGAL::Bbox_2`, `CGAL::Bbox_3`, `CGAL::Circle_2`,
    `CGAL::Iso_cuboid_3`, `CGAL::Iso_rectangle_2`, `CGAL::Point_2`,
    `CGAL::Point_3`, `CGAL::Segment_2`, `CGAL::Segment_3`,
-   `CGAL::Vector_2`, `CGAL::Vector_3`, `CGAL::Weighted_point_2` and
-   `CGAL::Weighted_point_3`.
+   `CGAL::Sphere_3`, `CGAL::Vector_2`, `CGAL::Vector_3`,
+   `CGAL::Weighted_point_2` and `CGAL::Weighted_point_3`.
 
 ### IO Streams
 
--- a/Kernel_23/include/CGAL/Kernel/hash_functions.h
+++ b/Kernel_23/include/CGAL/Kernel/hash_functions.h
@@ -1,0 +1,54 @@
+// Copyright (c) 2019
+// GeometryFactory (France)
+//
+// This file is part of CGAL (www.cgal.org); you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public License as
+// published by the Free Software Foundation; either version 3 of the License,
+// or (at your option) any later version.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING THE
+// WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: LGPL-3.0+
+// 
+//
+// Author(s)     : Simon Giraudot
+ 
+#ifndef CGAL_KERNEL_HASH_FUNCTIONS_H
+#define CGAL_KERNEL_HASH_FUNCTIONS_H
+
+namespace CGAL
+{
+
+template <typename K>
+inline std::enable_if_t<std::is_floating_point<typename K::FT>::value, std::size_t>
+hash_value (const Point_3<K>& point)
+{
+  std::size_t result = boost::hash_value(point.x());
+  boost::hash_combine(result, boost::hash_value(point.y()));
+  boost::hash_combine(result, boost::hash_value(point.z()));
+  return result;
+}
+
+} //namespace CGAL
+
+// overloads of std::hash used for using std::unordered_[set/map] on CGAL Kernel objects
+namespace std
+{
+
+template <typename K> struct hash<CGAL::Point_3<K> >
+{
+  std::size_t operator() (const CGAL::Point_3<K>& point) const
+  {
+    return CGAL::hash_value<K> (point);
+  }
+};
+
+}
+
+#endif  // CGAL_KERNEL_HASH_FUNCTIONS_H
--- a/Kernel_23/include/CGAL/Kernel/hash_functions.h
+++ b/Kernel_23/include/CGAL/Kernel/hash_functions.h
@@ -27,11 +27,196 @@ namespace CGAL
 
 template <typename K>
 inline std::enable_if_t<std::is_floating_point<typename K::FT>::value, std::size_t>
+hash_value (const Aff_transformation_2<K>& transform)
+{
+  std::size_t result = boost::hash_value(transform.cartesian(0,0));
+  for(int i=0; i < 3; ++i)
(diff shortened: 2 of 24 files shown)


lrineau:
For the documentation, you can document
> `Point_3<K>` is a model of [`Hashable`] if `K::FT` is a floating-point number.

[`Hashable`]: https://doc.cgal.org/latest/STL_Extension/classHashable.html

sgiraudot:
I've completed the work on other objects, created the small feature and updated `CHANGES.md`.

I did not judge it necessary to upload the doc, as the changes are quite trivial (just adding a `\cgalModels` to several objects), but let me know if you want me to do it.

Installation/CHANGES.md
```diff
@@ -45,6 +45,16 @@ Release date: September 2019
...
+ - The following objects are now hashable and thus trivially usable
+   with `std::unordered_set` and `std::unordered_map`:
+   `CGAL::Aff_transformation_2.h`, `CGAL::Aff_transformation_3.h`,
+   `CGAL::Bbox_2.h`, `CGAL::Bbox_3.h`, `CGAL::Circle_2.h`,
+   `CGAL::Direction_2.h`, `CGAL::Direction_3.h`,
+   `CGAL::Iso_cuboid_3.h`, `CGAL::Iso_rectangle_2.h`,
+   `CGAL::Point_2.h`, `CGAL::Point_3.h`, `CGAL::Ray_2.h`,
+   `CGAL::Ray_3.h`, `CGAL::Segment_2.h`, `CGAL::Segment_3.h`,
+   `CGAL::Vector_2.h`, `CGAL::Vector_3.h`, `CGAL::Weighted_point_2.h`
+   and `CGAL::Weighted_point_3.h`.
```
lrineau: I think you did not want those `.h` extensions after the object types.

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Jun 4, 2019, 10:18 AM GMT+2](https://github.com/CGAL/cgal/pull/3973#issuecomment-498571255 "2019-06-04T08:18:00Z - Replied by Github Reply Comments"):
> I've completed the work on other objects, created the small feature and updated `CHANGES.md`.

Please provide the link to the small feature. I cannot see it.

> I did not judge it necessary to upload the built doc, as the changes are quite trivial (just adding a `\cgalModels` to several objects), but let me know if you want me to do it.

Our testsuite does not check the compilation of the documentation (it does, but Doxygen rarely report errors), and that is really easy to mess the documentation with a Doxygen syntax error. Please provide a built version of the documentation, so that anybody can visually checks that everything is OK.

Installation/CHANGES.md
sgiraudot: Oops, yes, that was a badly processed copy-paste... I'll correct it.

sgiraudot:
Ok, I've updated the doc and added the link to the small feature.

mglisse:
> in the cases where `Kernel::FT` is a floating-point type

why not in the cases where `Kernel::FT` is hashable?

mglisse:
It seems that the `Direction_2` given by a vector and by twice that vector are the same for `==` but have different hashes?

lrineau:
[**@mglisse**](https://github.com/mglisse) commented on [Jun 4, 2019, 1:02 PM GMT+2](https://github.com/CGAL/cgal/pull/3973#issuecomment-498624914 "2019-06-04T11:02:31Z - Replied by Github Reply Comments"):
> It seems that the `Direction_2` given by a vector and by twice that vector are the same for `==` but have different hashes?

I agree with Marc about `Direction_2` and `Direction_3`: their `operator==` is non-trivial, and those types should not be hashable. And, anyway, who wants to bet on the usefulness of hashing directions?...

sgiraudot:
Yes, you are right about directions, I did not notice that. I'll remove them.

About the `Kernel::FT` being hashable, it's a bit more complicated, because the requirement is also that the hash does not change over time: as the representation of exact numbers may change if computations are performed on them, it seems non-trivial to create a consistent hash from them, unless I'm mistaken (if there is a way to do it, then we should include it).

sgiraudot:
I'll also remove rays which depend on direction…

lrineau:
We should also restrict the hashes to object of Cartesian kernels, if possible.

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Jun 4, 2019, 1:17 PM GMT+2](https://github.com/CGAL/cgal/pull/3973#issuecomment-498629207 "2019-06-04T11:17:39Z - Replied by Github Reply Comments"):
> About the `Kernel::FT` being hashable, it's a bit more complicated, because the requirement is also that the hash does not change over time: as the representation of exact numbers may change if computations are performed on them, it seems non-trivial to create a consistent hash from them, unless I'm mistaken (if there is a way to do it, then we should include it).

Is not that a moral contract on hashes that hashes are consistent with the value-semantic of objects? And in particular
  - if `a==b`, then `hash(a)==hash(b)`.
  - and thus hashes of const objects should not change.

sgiraudot:
Yes, exactly, see [the doc of `std::hash`](https://en.cppreference.com/w/cpp/utility/hash):

> 4. For two parameters k1 and k2 that are equal, std::hash<Key>()(k1) == std::hash<Key>()(k2).

mglisse:
> the requirement is also that the hash does not change over time: as the representation of exact numbers may change if computations are performed on them,

Then that number type should not be hashable? Gmpq or mpq_class seem very hashable to me, although the hash function should be provided by GMP and isn't :disappointed:. Hopefully `mpq_rational` is hashable? `Lazy_exact_nt` can only be sensibly hashed by hashing the result of `exact()`, so it is debatable if the hashing is useful, but at least it seems possible. What did you have in mind?

afabri:
Note that `Rational` has the same problem as `Direction`.

sgiraudot:
I have a very loose knowledge of exact numbers, so I think I was mainly thinking about `Lazy_exact_nt`. I have no idea if the other types you mention are easily hashable or not (not to mention how to do it), I only took care of the `double`/`float` case as it was the only case I felt confident in what I was doing.

lrineau:
[**@sgiraudot**](https://github.com/sgiraudot) commented on [Jun 4, 2019, 1:40 PM GMT+2](https://github.com/CGAL/cgal/pull/3973#issuecomment-498635784 "2019-06-04T11:40:45Z - Replied by Github Reply Comments"):
> I have a very loose knowledge of exact numbers, so I think I was mainly thinking about `Lazy_exact_nt`. I have no idea if the other types you mention are easily hashable or not (not to mention how to do it), I only took care of the `double`/`float` case as it was the only case I felt confident in what I was doing.

The issue https://github.com/CGAL/cgal/issues/1116 has been stalled for three years, because there was that discussion about hashing number types. The proposal of Simon is sound: provide the hash functions of number types that are know to have hashes (actually, the hashing of a `double` is not completely trivial, because `-0==0` whereas they have two distinct bits representation, but `std::hash` deals with that for us).

I think we can discuss about extending the proposal of Simon, if that is easy, but that should not be a reason to block the integration of the version limited to `float`/`double`..

mglisse:
> Note that `Rational` has the same problem as `Direction`.

i.e. they need to be normalized for the hash to make sense. Good thing that mpq_t are always normalized. Quotient isn't, which is why I was saying that it requires a call to `normalize()` in the hasher (or a promise from the user that they did normalize all their numbers first).

mglisse:
> provide the hash functions of number types that are know to have hashes

That's pretty much what I was suggesting, except that I was thinking of making it automatic instead of hard-coding some types. In some sense I wanted to make it **more** independent of the problem of hashing numbers. I didn't ask to reject or block this PR.

lrineau:
How can it be automatic? No `enable_if` and we trust that if `std::hash<FT>` compiles then it is good?

As I said, I would like to `enable_if` the `Rep_tag` of the kernel is `Cartesian_tag`. I do not think we can hash correctly homogeneous types, because of quotient by `hw`.

Kernel_23/include/CGAL/Kernel/hash_functions.h
```diff
@@ -0,0 +1,275 @@
...
+#define CGAL_KERNEL_HASH_FUNCTIONS_H
+
+namespace CGAL
+{
+
+template <typename K>
+inline std::enable_if_t<std::is_floating_point<typename K::FT>::value, std::size_t>
+hash_value (const Aff_transformation_2<K>& transform)
+{
+  std::size_t result = boost::hash_value(transform.cartesian(0,0));
```
lrineau: `hash_value` must be called by ADL, without the namespace. Or, if you want to specify a namespace, it must be `CGAL`.

Travis tests are fully red: https://travis-ci.org/CGAL/cgal/builds/541190492. It probably does not compile on your machine either.

sgiraudot: It does compile using Clang (I just double-checked), but fails with GCC.

But in that case, `hash_value` is applied on a double/float type, so it does come from Boost, doesn't it? I pretty much copied what I did for [`CGAL::grid_simplify_points()`](https://github.com/CGAL/cgal/blob/master/Point_set_processing_3/include/CGAL/grid_simplify_point_set.h#L80) a few years ago.

mglisse:
> How can it be automatic? No `enable_if` and we trust that if `std::hash<FT>` compiles then it is good?

Sure, why not? Unless I am missing something, the main reason not to would be so users can SFINAE-detect if Point_2 is hashable, and if that seems important, we could enable_if it.

Kernel_23/include/CGAL/Kernel/hash_functions.h
lrineau: Actually my suggestion was wrong. The correct one is:
  - call by ADL (without the namespace),
  - but add a `using boost::hash_value;` before, for the cases where the FT does not have a namespace, like `double`.

Maybe the errors in Travis are just about a missing `#include` for `boost::hash_value`. But in general `hash_value` must be called by ADL (with the additional `using`).

sgiraudot:
Alright, so I fixed the issues and after discussion with Laurent, we changed the condition: now the kernel needs to be cartesian and the FT just needs to be hashable (which means if the FT of Epeck becomes hashable, the objects of Epeck will be usable with unordered set/map without changing anything).

sgiraudot:
Just for information, I tested the following:
- Using `Simple_homogeneous` returns an error that says that no `hash_value` can be found for the wanted object (`Point_2`, etc.)
- Using `Exact_predicates_exact_constructions` returns an error that says that no `std::hash<FT>` can be found.

So I think these are pretty clear errors and that match the documentation quite well, which should prevent any bad surprise for the user.

maxGimeno:
There are many errors in Travis, I cannot add this PR into integration until it is all clear.

lrineau:
The functions `hash_value` for Bbox_23 are not template functions, and must be declared `inline`.

lrineau:

PR #3981, opened 2019-06-05
Surface_mesh: No need to call add_property_map()

afabri:
## Summary of Changes

... because we add the property in the constructor and never remove it.

## Release Management


afabri:
@maxGimeno ready for testing.

lrineau:
Is that a bug-fix, or something we want only for `master`?

afabri:
I can't imagine that it is critical, so targeting 5.0 is fine with me.

maxGimeno:

PR #3988, opened 2019-06-07
Polyhedron Demo: Enhance doc

maxGimeno:
## Summary of Changes

Add a warning about the Polyhedron_demo_DIR in the doc to explain it needs to be from the same CGAL version as the CGAL_DIR
## Release Management

* Affected package(s):Three


maxGimeno:

PR #3996, opened 2019-06-13
Put declaration before the call to the function

sloriot:


maxGimeno:

PR #4003, opened 2019-06-14
3D Demo: Fix Points Selection when Clipping

### LINKED ISSUE #3955: [Demo] Clipping+point set selection not working for data far away from origin
## Issue Details

1. Open/generate a point set far away from origin
2. Clip something using orthographic mode (probably works with the box mode too)
3. Try to select something: nothing is selected

I guess this has to do with the trick used to avoid bad display with data far away from origin (the same thing works with points centered on the origin).

## Environment

* CGAL version: master

maxGimeno:
## Summary of Changes
Take the offset into account when testing if the point is clipped during the selection.

## Release Management
* Issue(s) solved (if any): fix #3955


lrineau:
That is a bug fix, and the patch is trivial. I am wondering why it applies only to `master`.

maxGimeno:
Because I'm almost certain it appeared with  the big branch of the multiviewers

maxGimeno:
I was wrong, it is already in 4.14. Somefeatures are older than I thought

lrineau:

PR #1888, opened 2017-02-07
Followup for PR #1744

lrineau:
The commit 6c7774048521d2779d1657871f476624a46d220b was not complete:
> CORE MemoryPool<T> has to be destroyed last
>
> If Boost implementation of thread local storage is used, the order of
> destructors is reversed, compared to C++11 `static thread_local`. The
> solution for CORE `MemoryPool<T>` is to make the static variable a
> static member of the class, and initialize the pointer only inside the
> function. That ensures that the destructor will be called after the
> destructor of local static variables.

Actually the explanation about the reverse order is not right, and even
with C++ `thread_local`, we have to ensure that the static data member
of `MemoryPool<T>` is created before any other CORE static variable.

This PR is a followup of the commit
6c7774048521d2779d1657871f476624a46d220b: even in C++11, the `memPool`
variable becomes a thread-local data member, instead of a thread-local
variable at function scope.

Fix #1844.

PR #1744, opened 2016-12-05
CGAL_CORE:  Make it thread safe

afabri:
Towards a fix of Issue #1397

PR #4012, opened 2019-06-18
Fix a stack-use-after-scope in Arr_polycurve_basic_traits_2

lrineau:
## Summary of Changes

Before this patch, `Arr_polycurve_basic_traits_2` was holding a const
reference to a temporary object.

Here was the result of ctest, with the address sanitizer:
```shellsession
$ make && ASAN_OPTIONS=detect_leaks=0 ctest -L Arr -j6
[...]
The following tests FAILED:
        144 - execution___of__generic_curve_data (Failed)
        170 - execution___of__polycurve_circular_arc (Failed)
        172 - execution___of__polycurve_conic (Failed)
        176 - execution___of__polycurves_basic (Failed)
        178 - execution___of__polylines (Failed)
        2166 - test_traits_test_polylines__data_polylines_compare_y_at_x__polyline_traits (Failed)
        2168 - test_traits_test_polylines__data_polylines_intersect__polyline_traits (Failed)
        2169 - test_traits_test_polylines__data_polylines_split__polyline_traits (Failed)
        2171 - test_traits_test_polylines__data_polylines_assertions__polyline_traits (Failed)
        2175 - test_traits_conic_polycurve__data_polycurves_conics_compare_y_at_x__polycurve_conic_traits (Failed)
        2176 - test_traits_conic_polycurve__data_polycurves_conics_compare_y_at_x_left__polycurve_conic_traits (Failed)
        2177 - test_traits_conic_polycurve__data_polycurves_conics_compare_y_at_x_right__polycurve_conic_traits (Failed)
        2179 - test_traits_conic_polycurve__data_polycurves_conics_intersect__polycurve_conic_traits (Failed)
        2180 - test_traits_conic_polycurve__data_polycurves_conics_split__polycurve_conic_traits (Failed)
        2193 - test_traits_circular_arc_polycurve__data_Polycurves_circular_arcs_compare_y_at_x__polycurve_circular_arc_traits (Failed)
        2198 - test_traits_circular_arc_polycurve__data_Polycurves_circular_arcs_split__polycurve_circular_arc_traits (Failed)
        2221 - test_traits_non_caching_polylines__data_polylines_compare_y_at_x__non_caching_polyline_traits (Failed)
        2223 - test_traits_non_caching_polylines__data_polylines_intersect__non_caching_polyline_traits (Failed)
        2224 - test_traits_non_caching_polylines__data_polylines_split__non_caching_polyline_traits (Failed)
[...]

PR #4013, opened 2019-06-18
Remove the need for Boost.Thread (but for CGAL_Core and gcc<9.1)

lrineau:
## Summary of Changes

Now that CGAL uses C++14, Boost.Thread should no longer be used.

Actually, after #1744 and #1888, it was documented that CGAL_Core still need Boost.Thread, with GCC, because of a bug. <s>I can no longer reproduce it, so let's try to remove Boost.Thread entirely, so that CGAL is header-only, with Boost header-only, and requires only GMP/MPFR as binary dependencies.</s>
**Update:** there are still bugs, at least until gcc 9.1. For gcc 9.1, I think that [bug fix PR 60702](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60702) fixed what is needed for CGAL_Core.

I have tested on my machine with both 64 and 32 bits builds, the ctest testsuite of the following packages:
  - `Algebraic_foundations_Tests`
  - `Algebraic_kernel_d_Tests`
  - `Algebraic_kernel_for_circles_Tests`
  - `Algebraic_kernel_for_spheres_Tests`
  - `Arithmetic_kernel_Tests`
  - `Arrangement_on_surface_2_Tests`
  - `Number_types_Tests`
  - `Polynomial_Tests`
  - `Straight_skeleton_2_Tests`

## Release Management

* Affected package(s): Installation, CGAL_Core


lrineau:
I have tested with the AddressSanitizer, to be sure. That is why this PR has cherry-picked the commit from PR #4012.

lrineau:
Unfortunately, that does not work. The following assertion was triggered a lot:
```
CGAL error: assertion violation!
Expression : ! blocks.empty()
File       : C:\CGAL\test\CGAL-5.0-Ic-64\include\CGAL/CORE/MemoryPool.h
Line       : 130
```

in:
```
CGAL-5.0-Ic-64/Algebraic_kernel_d/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Arithmetic_kernel/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2_Examples/TestReport_cgaltester_x86-64_Darwin-18.2_Apple-clang-default_Release.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2/TestReport_cgaltester_x86-64_Darwin-18.2_Apple-clang-default_Release.gz
CGAL-5.0-Ic-64/Arrangement_on_surface_2/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Boolean_set_operations_2_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Core_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Hyperbolic_triangulation_2_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Hyperbolic_triangulation_2/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Interval_support/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Number_types/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Periodic_3_triangulation_3/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2_Examples/TestReport_cgaltester_x86-64_Darwin-18.2_Apple-clang-default_Release.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2/TestReport_cgaltester_x86-64_Darwin-18.2_Apple-clang-default_Release.gz
CGAL-5.0-Ic-64/Periodic_4_hyperbolic_triangulation_2/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Segment_Delaunay_graph_2/TestReport_lrineau_ArchLinux-CXX17-Release.gz
CGAL-5.0-Ic-64/Segment_Delaunay_graph_Linf_2/TestReport_lrineau_ArchLinux-CXX17-Release.gz
```

sloriot:
Could you refresh my memory? It is due to the wrong delete order, right?

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jun 22, 2019, 10:50 AM GMT+2](https://github.com/CGAL/cgal/pull/4013#issuecomment-504645663 "2019-06-22T08:50:19Z - Replied by Github Reply Comments"):
> Could you refresh my memory? It is due to the wrong delete order, right?

Well... that complicated. Very complicated. And when I wrong PR #1744 and #1888 I should have documented the code a lot more. Because it took me several days (and long thoughts during my evenings) to rediscover the exact chain of issues... I will write it there, and then, copy that long text verbatim in the source code of `<CGAL/Core/MemoryPool.h>`, once you confirm that the text is correct and understandable...

GCC bugs with `thread_local`
----------------------------
gcc has bugs with `thread_local`. At lot! See the meta-bug about `thread_local` in gcc's bugzilla: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59994.

And even in very recent gcc versions. On my desktop machine, I use gcc-9.1.1, and they have fixed [at least one bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60702) that impacted CGAL code in gcc-9.1, and future bug-fix gcc-8.4 (not yet published).

Now, knowing that gcc<9.1 cannot be trusted completely with gcc<9.1, one has to keep using Boost.Thread at least for CGAL_Core.

Use of Boost.Thread with CGAL_Core
----------------------------------
And then with Boost.Thread, there is one extra more problem with CGAL_Core... let have a look at `BigFloat::getZero()`:
```C++
const BigFloat& BigFloat::getZero() {
  CGAL_STATIC_THREAD_LOCAL_VARIABLE(BigFloat, Zero,0);
  return Zero;
}
```
with Boost.Thread, the macro `CGAL_STATIC_THREAD_LOCAL_VARIABLE` evaluates to:
```C++
const BigFloat& BigFloat::getZero() {
  static boost::thread_specific_ptr<BigFloat> Zero_ptr;
  if(Zero_ptr.get() == nullptr) {Zero_ptr.reset(new BigFloat());}
  BigFloat& Zero =  * Zero_ptr.get()
  return Zero;
}
```
That's rather easy to read, and looks correct. ... Now, consider that `BigFloat` contains a shared pointer to a `BigFloatRep`, and that `BigFloatRep` defines `operator new`, that way:
```c++
    void *BigFloatRep::operator new( size_t size)
    { return MemoryPool<BigFloatRep>::global_allocator().allocate(size); }
```
and `MemoryPool<T>::global_allocator()` also uses a `thread_local` object... Now, at the first call of `BigFloat::getZero()` in the program (or in a thread), then the `boost::thread_specific_ptr<BigFloat> Zero_ptr` is created first, with value `nullptr`, and then during the call `Zero_ptr.reset(new BigFloat())` the `MemoryPool<BigFloatRep>::global_allocator()` is called.

That means that the object returned by `MemoryPool<BigFloatRep>::global_allocator()` might be create **second**, after the creation of the `thread_specific_ptr` `Zero_ptr`... and that means that during the destruction of the thread, or at the end of the program, `MemoryPool<BigFloatRep>::global_allocator()` might be destroyed **first** (because the destructors of static objects are called in reversed order of the calls to constructors), and that means that the memory pool will be destroyed before the destructor of `Zero`! That is why the memory pool returned by `MemoryPool<BigFloatRep>::global_allocator()` has to be created *before* any call to `getZero` or similar static functions.

Solution to create MemoryPool<BigFloatRep>::global_allocator()` earlier
-----------------------------------------------------------------------
To ensure that the memory pools are created before any call to static functions, one can define then as static data members of the class template `MemoryPool<T>`.

Static non-local variables [are initialized very early][nonlocal]:
> All non-local variables with static storage duration are initialized as part of program startup, before the execution of the main function begins

[nonlocal]: https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables

whereas [as for local static variables][localstatic]:
> Variables declared at block scope with the specifier static have static storage duration but are initialized the first time control passes through their declaration

[localstatic]: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

That actually exposes [another bug of gcc with `thread_local`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66944), that is not yet fixed, not even in gcc-9.1.

Why is that different with `thread_local`?
------------------------------------------
And actually that order issue would be different with `thread_local` was used: in `getZero`, the object would be declared `thread_local Zero(0)`, and its constructor would call `BigFloatRep->operator new` as well. But, [according to the norm][atexit]:
> destructors of objects with static storage duration are called in *reverse order* of **completion** of their constructors

and the *completion* of the constructor of `Zero` is *after* the completion of the construction of the memory pool returned by `MemoryPool<BigFloatRep>::global_allocator()`.

[atexit]: https://en.cppreference.com/w/cpp/utility/program/exit

lrineau:
So, now, after this PR, assuming it is correct. We will no longer document that CGAL depends on Boost.Thread (because with C++11 we can use `thread_local`). But we document that CGAL_Core will need Boost.Thread when gcc<9.1 is used.

lrineau:
It seems it eventually looked good, in [CGAL-5.0-Ic-68](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-68/).

maxGimeno:
/mnt/testsuite/include/CGAL/CORE/MemoryPool.h:124:25: error: redefinition of ‘bool __tls_guard’

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Jun 26, 2019, 3:30 PM GMT+2](https://github.com/CGAL/cgal/pull/4013#issuecomment-505875381 "2019-06-26T13:30:28Z - Replied by Github Reply Comments"):
> /mnt/testsuite/include/CGAL/CORE/MemoryPool.h:124:25: error: redefinition of ‘bool __tls_guard’

I mixed-up things. With non-gcc compilers, or with gcc>=9.1, one has to use the function-local `thread_local` variable, to avoid the use of Boost.Thread. Probably fixed in https://github.com/CGAL/cgal/pull/4013/commits/9eed06477dfe366ee08fde1eb22dcd0e51a77666. I swear I tested it... but with clang. :disappointed:

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-71.shtml

lrineau:
I have understood why that patch fails on two packages of CGAL, with AppleClang (using libc++) or MSVC:
- https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-71.shtml#Arrangement_on_surface_2
- https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-71.shtml#Periodic_4_hyperbolic_triangulation_2

In:
https://github.com/CGAL/cgal/blob/9eed06477dfe366ee08fde1eb22dcd0e51a77666/Periodic_4_hyperbolic_triangulation_2/include/CGAL/Hyperbolic_octagon_translation.h#L64-L74

there is an object (a matrix) that is created empty (and thus the completion of its constructor is done), and then filled with numbers.

That means that it is possible that its constructor is completed **before** any CORE numbers are created, and thus **before** the corresponding `CORE::MemoryPool` are created. Then at the exit of the thread/program, the memory pool is destroyed before the number contained in that matrix.

Similarly, in Ak_2 used by the algebraic traits class of Arr_2:
https://github.com/CGAL/cgal/blob/6639a69869952d28bd87d334414285460f48e2ef/Algebraic_kernel_d/include/CGAL/Algebraic_kernel_d/Algebraic_curve_kernel_2.h#L411-L415

That thread local object is contains a cache of algebraic curves, created empty. Then that cache is filled with curves containing numbers from CORE. It is possible that the memory pools of CORE are destroyed before that object.

lrineau:
@maxGimeno Do not add this PR for tonight tests.

maxGimeno:

maxGimeno:
There are now problems with magritte, probably due to this PR : 
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-I-84.shtml

It is strange, because it has worked for some time with the last commit before suddenly breaking with no apparent reason. Maybe it was reverted or hidden by another PR in the testsuite

maxGimeno:
It might cause [this](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-83/Installation/TestReport_lrineau_Ubuntu-no-gmp-no-leda.gz) too, I'm not sure.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Jul 15, 2019, 1:17 PM GMT+2](https://github.com/CGAL/cgal/pull/4013#issuecomment-511362328 "2019-07-15T11:17:44Z - Replied by Github Reply Comments"):
> There are now problems with magritte, probably due to this PR : 
> https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-I-84.shtml
> 
> We didn't see it before because Magritte was off since Ic-67.

Hopefully these platforms with old versions of XCode will disappear. Let's say we no longer support them.

PR #4023, opened 2019-06-21
Mesh_3: remove the attempt for `os << std::pair`

lrineau:
## Summary of Changes

`std::cout << std::make_pair(0, 1)` does not compile... I have tried in the past to add an `operator<<` for `std::pair`, but that does not work. Let's remove that old attempt, now.

## Release Management

* Affected package(s): Mesh_3


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-73.shtml

PR #4039, opened 2019-06-27
update marks on removed edges

sloriot:
Needed if you have garbage or use a (hash) map

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/Face_graph_output_builder.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/Face_graph_output_builder.h
@@ -560,9 +560,15 @@ public:
         ++epp_it;
     }
     BOOST_FOREACH(edge_descriptor ed, inter_edges_to_remove1)
+    {
+      put(marks_on_input_edges.ecm1, ed, false);
       intersection_edges1.erase(ed);
+    }
     BOOST_FOREACH(edge_descriptor ed, inter_edges_to_remove2)
+    {
+      put(marks_on_input_edges.ecm2, ed, false);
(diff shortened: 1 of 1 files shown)


sloriot:
Since it is conflicting with master (why?) I have merged the PR in integration.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jun 27, 2019, 8:48 AM GMT+2](https://github.com/CGAL/cgal/pull/4039#issuecomment-506216167 "2019-06-27T06:48:55Z - Replied by Github Reply Comments"):
> Since it is conflicting with master (why?)

Because of the changes from `BOOST_FOREACH` to C++11 range-based for-loop.

sloriot:
I know but I was expecting that since the change was on a different line it wouldn't be a problem. Maybe it is too close and git prefers to be on a safe side.

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jun 27, 2019, 9:49 AM GMT+2](https://github.com/CGAL/cgal/pull/4039#issuecomment-506234330 "2019-06-27T07:49:00Z - Replied by Github Reply Comments"):
> I know but I was expecting that since the change was on a different line it wouldn't be a problem. Maybe it is too close and git prefers to be on a safe side.

The `diff` and `patch` work on chunks. You can see the size of the chunk `git diff` or `git show`. And also in Github: https://github.com/CGAL/cgal/pull/4039/files

I think Git is right to be safe, even if that is annoying. That is why I was not particularly happy to see changes in `master` about C++11 without any particular reason but "it is more pretty" or "more uniform". Any of those changes is a maintenance burden.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-71.shtml

PR #4042, opened 2019-06-28
Surface mesh topology (new package)

gdamiand:

## Summary of Changes

Add a new package: surface mesh topology.

This feature introduces a new package to compute topological invariants on surfaces. The first method provides allows to test if two cycles on a surface are homotopic. This method is used to provide two more functions: a first one to test if a cycle is contractible, a second one to test if two paths are based points homotopic. 

TODO:
- [ ] Code review (?)
- [x] Update CHANGES.md
- [x] Check branch size
- [ ] News entry

## Release Management

* Affected package(s): Surface_mesh_topology
* Feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Surface_mesh_topology
* License and copyright ownership: GPL; CNRS and Liris establishements


gdamiand:
Is there a guideline somewhere that explains what to do to add a new package ? I didn't find it.

MaelRL:
https://github.com/CGAL/cgal/wiki/Travis-CI-Explained-and-Check-List-for-a-New-Package is supposed to regroup the necessary information

lrineau:
https://travis-ci.org/CGAL/cgal/jobs/551799243#L6324
```
In file included from /home/travis/build/CGAL/cgal/Installation/../Combinatorial_map/include/CGAL/Face_graph_wrapper.h:24:
In file included from /home/travis/build/CGAL/cgal/Combinatorial_map/include/CGAL/Functors_for_face_graph_wrapper.h:24:
/home/travis/build/CGAL/cgal/Combinatorial_map/include/CGAL/Iterators_for_face_graph_wrapper.h:99:14: error: use of undeclared identifier 'boost'
    typename boost::graph_traits<typename Map_::HEG>::halfedge_iterator m_it;
             ^
```
Probably a missing `#include`.

lrineau:
https://travis-ci.org/CGAL/cgal/jobs/551799285
```
No dependencies found for Surface_mesh_topology
```
@maxGimeno Can you have a look? Maybe the Travis script should display what should be the content of the file, so that package authors can be independent.

maxGimeno:
package_info's dependencies are missing.
I'll change the script so it computes and display the dependencies instead of just failing.

maxGimeno:
done in #4045

gdamiand:
Instead of comparing in travis the contend of dependencies files and the result of packages_dependencies, why not generating directly the dependencies files ? This should simplify the submission of pull request, avoiding several errors ?

lrineau:
[**@gdamiand**](https://github.com/gdamiand) commented on [Jul 3, 2019, 9:12 AM GMT+2](https://github.com/CGAL/cgal/pull/4042#issuecomment-507970429 "2019-07-03T07:12:03Z - Replied by Github Reply Comments"):
> Instead of comparing in travis the contend of dependencies files and the result of packages_dependencies, why not generating directly the dependencies files ? This should simplify the submission of pull request, avoiding several errors ?

The main idea of those files is that we want to monitor, and approve, when they are modified. It would be an error if, for example, Mesh_3 became a dependency of Triangulation_3. That is why it has to be modified by a human.

Combinatorial_map/package_info/Combinatorial_map/dependencies
```diff
@@ -14,8 +17,15 @@ Intersections_3
...
 Modular_arithmetic
 Number_types
+Polyhedron
+Polyhedron_IO
 Profiling_tools
+Property_map
 STL_Extension
 Stream_support
+Surface_mesh
+Surface_mesh_topology
```
lrineau: What happended?! Why are there suddenly so many new dependencies? We cannot accept that because `Combinatorial_map` is licensed LGPLv+3, and it cannot depend on GPL package.

lrineau:
TODO: solve the dependency issue. https://github.com/CGAL/cgal/pull/4042#pullrequestreview-257299220

Combinatorial_map/package_info/Combinatorial_map/dependencies
gdamiand: I don't know; this is the result of the make packages_dependencies !!
Maybe this is due to some links in the doc ?

Combinatorial_map/include/CGAL/Combinatorial_map.h
```diff
@@ -50,6 +50,11 @@
 #include <deque>
 #include <boost/type_traits/is_same.hpp>
 #include <boost/unordered_map.hpp>
+#include <boost/graph/graph_traits.hpp>
+#include <CGAL/boost/graph/helpers.h>
+#include <CGAL/boost/graph/graph_traits_Polyhedron_3.h>
+#include <CGAL/boost/graph/graph_traits_Surface_mesh.h>
```
lrineau: This file should be include anything specific to `Polyhedron_3` or `Surface_mesh`. That is why the dependencies increased that much, including bad dependencies to GPL headers.

sloriot: I don't see why you would need to include those graph_traits files anyway.

gdamiand: For the method "import_from_halfedge_graph", which takes an halfedge graph as input and build a CMap.  Without these include I have error, for example 

‘next’ is not a member of ‘CGAL’
                         (*origin_to_copy)[CGAL::next(dartmap_iter->first, heg)],

gdamiand: I can remove the #include <CGAL/boost/graph/graph_traits_Surface_mesh.h>, keeping the others. It should be ok since BGL and Polyhedron are both LGPL.

lrineau: The functions described in the BGL concepts must be called by [ADL], not qualified by `CGAL::`. Otherwise the code cannot compile on a non-CGAL face graph where the `next` function can be in the `openmesh` namespace, for example.

[ADL]: https://en.cppreference.com/w/cpp/language/adl

BGL/package_info/BGL/dependencies
```diff
@@ -1,19 +1,11 @@
...
-Distance_3
-Filtered_kernel
 Hash_map
-Homogeneous_kernel
 Installation
-Intersections_2
-Intersections_3
 Interval_support
 Kernel_23
-Kernel_d
```
lrineau: If you run `make packages_dependencies` without openmesh found by CMake, then the dependencies for BGL are wrong.
The header `BGL/include/CGAL/boost/graph/properties_OpenMesh.h` includes `<CGAL/Exact_predicates_inexact_constructions_kernel.h>` and that triggers those dependencies to the CGAL core packages. Travis will probably complain.

Combinatorial_map/include/CGAL/Combinatorial_map.h
gdamiand: I am not able to remove the CGAL::
Indeed, in the CMap class, I have a method next(Dart_handle). Without CGAL::next the compiler complains that there is no next function with two parameters.

@lrineau Do you have any solution to this problem ?

lrineau: Having a member function named `next` blokes the ADL lookup. I am looking for a solution...

gdamiand: One possibility is to say that my method that converts a model of face graph into a combinatorial maps works only for CGAL models. Then I can keep the CGAL::

sloriot: `Surface_mesh` also has a `next` member function and I'm not aware of any issue due to that. The issue should be somewhere else?

gdamiand: Does a method of Surface_mesh calls a next of bgl? Probably not.

afabri: A member function or a free function ?

gdamiand: A member function of Surface_mesh which calls a free function next ?

lrineau: In Surface_mesh, there is nowhere where `CGAL::next` is called on a face graph type that is not `Surface_mesh` itself. So probably the calls are qualified, with `CGAL::`.

sloriot: Ah I understand now. Making your import function a free function should solve the issue. I don't have time now but I'm willing to help if needed.

lrineau: I have a small patch in hand. I am testing it fully...

Combinatorial_map/test/Combinatorial_map/Combinatorial_map_test.cpp
```diff
@@ -193,8 +198,60 @@ bool test_get_new_mark()
   return true;
 }
 
+bool test_face_graph_wrapper()
+{
+  bool res=true;
+
+  typedef CGAL::Surface_mesh<CGAL::Simple_cartesian<double>::Point_3> SMesh;
+  SMesh m;
+  std::ifstream in1("data/head.off");
```
lrineau: Guillaume: you forgot to commit and push that file `data/head.off`. CTest tests fail with:
```
23/24 Test #2705: execution___of__Combinatorial_map_test ..............***Failed    0.01 sec
Error: impossible to open 'data/head.off'
ERROR during test_face_graph_wrapper.
```

Combinatorial_map/include/CGAL/Combinatorial_map.h
lrineau: I have pushed a small fix in 7959a40e68f. Guillaume, do you agree with the patch?

In passing, I have added a test of the function `import_from_halfedge_graph` with an alien face graph (from OpenMesh). CTest tests pass on my machine, and I have verified that `import_from_halfedge_graph` does not compile with OpenMesh if the patch to `Combinatorial_map/include/CGAL/Combinatorial_map.h` is reverted. So the fix does fix something.

I have not fixed the `dependencies` files. For the moment they are wrong no matter what. Let's wait for Travis to tell us what to modify.

gdamiand: Thanks @lrineau  !
About your patch, I suppose I agree if it works both for Surface_mesh and OpenMesh. 
Thanks again !

Combinatorial_map/test/Combinatorial_map/Combinatorial_map_test.cpp
gdamiand: Will add.

lrineau:
@maxGimeno Please deal with this PR, to fix the `dependencies` files.

Combinatorial_map/include/CGAL/Face_graph_wrapper.h
```diff
@@ -0,0 +1,899 @@
...
+
+#include <CGAL/Functors_for_face_graph_wrapper.h>
+#include <CGAL/Iterators_for_face_graph_wrapper.h>
+#include <CGAL/Combinatorial_map.h>
+#include <CGAL/Linear_cell_complex_for_combinatorial_map.h>
+#include <CGAL/Generalized_map.h>
+#include <CGAL/Linear_cell_complex_for_combinatorial_map.h>
+#include <CGAL/Linear_cell_complex_for_generalized_map.h>
+#include <CGAL/Surface_mesh.h>
+#include <CGAL/Polyhedron_3.h>
```
lrineau: These two lines introduce a dependency on Polyhedron_3 and Surface_mesh, which are not LGPL.

lrineau:
In addition to my comment in https://github.com/CGAL/cgal/pull/4042#pullrequestreview-258936368, Travis still disagrees with the `dependencies`.

gdamiand:
What is the problem with dependencies of Algebraic_foundations and Distance_3 ?

Looking at the errors message, it seems dependencies that disappeared and the ones which are new are the same for Distance_3 ? Thus why this is an error ?

And for Algebraic_foundations, the error is 
Distance_2
Distance_3 have disappeared.
but they don't ?

gdamiand:
Easy ;)

Combinatorial_map/package_info/Combinatorial_map/dependencies
```diff
@@ -14,8 +16,13 @@ Intersections_3
...
+Modifier
 Modular_arithmetic
 Number_types
+Polyhedron
+Polyhedron_IO
 Profiling_tools
+Property_map
 STL_Extension
 Stream_support
+Surface_mesh
```
lrineau: I am glad you managed to make Travis happy.

But the changes of `dependencies` for CMap are still problematic: you cannot depend on `Polyhedron` or `Surface_mesh`.

Combinatorial_map/include/CGAL/Functors_for_face_graph_wrapper.h
```diff
@@ -0,0 +1,98 @@
...
+//
+// Author(s)     : Guillaume Damiand <guillaume.damiand@liris.cnrs.fr>
+//
+#ifndef CGAL_FUNCTORS_FOR_FACE_GRAPH_WRAPPER_H
+#define CGAL_FUNCTORS_FOR_FACE_GRAPH_WRAPPER_H 1
+
+#include <boost/graph/graph_traits.hpp>
+#include <CGAL/boost/graph/helpers.h>
+#include <CGAL/boost/graph/graph_traits_Polyhedron_3.h>
+#include <CGAL/boost/graph/graph_traits_Surface_mesh.h>
```
lrineau: The bad dependencies to Polyhedron and Surface_mesh are here.

Combinatorial_map/include/CGAL/Face_graph_wrapper.h
```diff
@@ -0,0 +1,944 @@
...
+  class Linear_cell_complex_for_generalized_map;
+
+  template<class P>
+  class Surface_mesh;
+
+  template<class PolyhedronTraits_3,
+           class PolyhedronItems_3,
+           template<class T, class I, class A> class T_HDS,
+           class Alloc>
+  struct Polyhedron_3;
```
lrineau: Forward-declarations are not robust to refactoring. Include `<CGAL/Polyhedron_3_fwd.h>` and `<CGAL/Surface_mesh/Surface_mesh_fwd.h>` instead. They are in `Installation/include/` to avoid any trouble with our dependencies checker.

Combinatorial_map/include/CGAL/Functors_for_face_graph_wrapper.h
gdamiand: Thanks @lrineau ; I am working on it.

Combinatorial_map/include/CGAL/Face_graph_wrapper.h
gdamiand: Ok, thanks !

gdamiand:
Not yet ready for testing; but soon.

gdamiand:
@MaelRL CHANGES.ms is updated  cf. [here](https://github.com/gdamiand/cgal/blob/Surface_mesh_topology-gdamiand/Installation/CHANGES.md).

MaelRL:
Still needs to be updated to say 5.1, put the news entry if there's any, etc., which is why I put the flag.

gdamiand:
This PR is ready for testing into integration.

sloriot:
Here are the [warnings](https://gist.github.com/sloriot/b7453e9bdce16858cc73fe58f526d106) I get when I compile the examples of the new package using `-Wall -Wextra`.

gdamiand:
@maxGimeno Lot of errors; is it possible to merge first #4224 before this PR (to simplify the integration process) ? (I think #4224 was tested successfully)

maxGimeno:
#4224 is successfully tested, I guess @lrineau could merge it quicly enough, it shouldn't be a problem.

gdamiand:
Ok @maxGimeno. 
Please do not add this PR in integration before #4224 will be integrated and I will have resolved conflicts. Thanks.

maxGimeno:
#4224 is now merged, please fix the new conflicts.

lrineau:
Note that last night the testsuite reported documentation issues:
```
CGAL-5.1.1-Ic-21/doc/Surface_mesh_topology/CGAL/Polygonal_schema_min_items.h:38: warning: unbalanced grouping commands
CGAL-5.1.1-Ic-21/doc/Surface_mesh_topology/PackageDescription.txt:35: warning: unable to resolve reference to 'licensesGPL' for \ref command
```
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1.1-Ic-21/logs_master/Surface_mesh_topology.log

gdamiand:
I don't know how to solve the warning 

CGAL-5.1.1-Ic-21/doc/Surface_mesh_topology/PackageDescription.txt:35: warning: unable to resolve reference to 'licensesGPL' for \ref command

Thanks for some help.

Surface_mesh_topology/doc/Surface_mesh_topology/dependencies
```diff
@@ -0,0 +1,6 @@
+Combinatorial_map
+Generalized_map
+Linear_cell_complex
+BGL
+Surface_mesh
+Polyhedron
```
lrineau: @gdamiand Add `Manual` in that list, because `licensesGPL` is defined in `Documentation/doc/Documentation/License.txt` inside the `Manual` package.

gdamiand: Thanks @lrineau  !!

maxGimeno:
Conflicts in CHANGES.md

sloriot:
@gdamiand The branch history is more than 6Mo. At some point, I'll have to filter the history to reduce the size. Please let me know when you consider the branch as stable.

gdamiand:
Ok @sloriot  (asap the test suite will be green).

maxGimeno:
@lrineau [errors in FindOpenmesh](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-33/BGL_OpenMesh_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz) from a commit of yours

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-33.shtml#Surface_mesh_topology : 
There are runtime errors on windows.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Dec 17, 2019, 3:01 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-566553205 "2019-12-17T14:01:46Z - Replied by Github Reply Comments"):
> https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-33.shtml#Surface_mesh_topology : 
> There are runtime errors on windows.

```
Assertion failed: list iterator not decrementable
```

```
Assertion failed: list iterator not dereferencable
```

Guillaume, you could enable the iterator debugging on Linux, to debug that locally on your machine, by adding `-D_GLIBCXX_DEBUG` in your compilation flags.

gdamiand:
Thanks @lrineau ; I will try. This error is very strange since this code works perfectly on other configurations.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Dec 17, 2019, 3:01 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-566553205 "2019-12-17T14:01:46Z - Replied by Github Reply Comments"):
> https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-33.shtml#Surface_mesh_topology : 
> There are runtime errors on windows.

Fixed in https://github.com/CGAL/cgal/pull/4042/commits/5565f6ce2df220bddd9f494b19edc07205e594e1.

lrineau:
I have also added the commit https://github.com/CGAL/cgal/pull/4042/commits/fd92f3ad7619a93ea2e3acdcf130b8e388651131, to support the `CONFIG` mode that will appear in OpenMesh-8.1.

gdamiand:
Thanks @lrineau .

Surface_mesh_topology/include/CGAL/Polygonal_schema.h
```diff
@@ -14,10 +14,11 @@
 
 #include <CGAL/license/Surface_mesh_topology.h>
 
-#ifdef _MSC_VER
+// TODO ADD SOME TEST IF WE ARE ON WINDOWS ?
+// THIS SEEMS A WRONG TEST =>  #ifdef _MSC_VER
 // To remove the warning on windows about unsafe strncpy function.
 #define _CRT_SECURE_NO_WARNINGS
-#endif
+// #endif
```
lrineau: [Code smell]!

1. That is very nasty, for our users, to deactivate a security feature of MSVC silently.
2. Why don't you use `strncpy_s`?
3. Actually, why are you using raw C strings in the first place? In C++, we have `std::string` to avoid a lot of nasty bugs.

[Code smell]: https://en.wikipedia.org/wiki/Code_smell

gdamiand: 1. My goal is to remove this warning only in my file. But for now it does not remove the warning, this is why I did some tests.
2. I was not able to use strncpy_s on my os; even with "__STDC_WANT_LIB_EXT1__ ". 
3. For memory reason; these labels are constants thus I don't need all the possibilities of std::string.

lrineau: I also personally think that we should reject any new code in CGAL that uses `new`/`delete`.

You should definitely use `std::string` instead of `const char*` and `new char[s.size()+1]` everywhere. At least with `std::string` you will benefit from improvements like the "small buffer optimization". The downside is that `sizeof(std::string)` is bigger that `sizeof(char*)`. But you might recover that memory with the small buffer optimization, if your labels are short enough.

gdamiand: Ok will do...

lrineau: If your labels are constants, then do not copy them with `strncopy`, and just hold the pointer to the constant strings.

Do you really gain something by not using `std::string`? Even if you do not "need all the possibilities". That is not a good reason: I have never needed all the possibilities of `std::vector`, but I will never use a C-array instead.

gdamiand: Still not convinced. One ptr: 8 bytes; One std::string between 24 and 40 bytes. For 500000 elements, I waste between 7 Mb and 15 Mb of memory for nothing...

lrineau: If you use-case is constant strings anyway, then only hold the pointers to constant strings, instead of copying the string in all nodes.

gdamiand: They are not constant strings; they are constant, i.e. never updated during their life.

lrineau: Is there any reason the node should hold a copy of the string, instead of referencing existing copies? If the answer is yes, then I would say "use `std::string`. If the answer is no, then you should use non-owning pointers to raw C-strings, without the `new/strncopy/delete` machinery.

gdamiand:
CHANGES.md is now updated.
One question remains: where to add this new package in the doc.
Maybe in the group "Geometry Processing" ?

sloriot:
About the group, you can indeed put it in Geometry Processing. @afabri  what to clean the group at some point but for the time being I think it is OK.

gdamiand:
Thanks @sloriot .

gdamiand:
@lrineau Still a compiler error with openmesh.
Do I need to update the cmake to use 
   include( UseOpenMesh )
  add_definitions( -DCGAL_USE_OPENMESH )
like done for BGL in master ?

(cf https://github.com/CGAL/cgal/blob/master/BGL/test/BGL/CMakeLists.txt)

gdamiand:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-36.shtml

lrineau:
[**@gdamiand**](https://github.com/gdamiand) commented on [Dec 22, 2019, 4:40 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-568273515 "2019-12-22T15:40:46Z - Replied by Github Reply Comments"):
> Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-36.shtml

Maxime @maxGimeno, can you confirm that?

Let's try to get the PR merged soon.... What about the remaining TODOs?

- I see that `CHANGES.md` has been changed (https://github.com/CGAL/cgal/pull/4042/files#diff-7714bacaf2a1d6fc7cf1462bad8f4fc2), but its texts is different from the text that was approved [in the feature page](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Surface_mesh_topology#CHANGES.md). @MaelRL, @afabri, @mglisse do you approved the new text? 
- Code review? How added that TODO? Or how should review the code?
- What about the branch size? @sloriot, can you have look?

gdamiand:
Thanks @lrineau.

About CHANGES.md, I am not sure the first version given in the feature page was approved by someone. I think the new version is better but I can revert and move to the initial version if you prefer.

About code review, it seems new for cgal packages ? Who is supposed to do the review ?

sloriot:
@gdamiand do you have another branch based on this one?

MaelRL:
"Code review" is there because I copied the complete TODO list from another feature (other entries in the list being relevant).

mglisse:
> do you approved the new text?

Sure, looks ok to me.
(164 files changed? That's one big feature...)

maxGimeno:
> Maxime @maxGimeno, can you confirm that?

Well it's all green, so I guess it is, yes.

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Jan 7, 2020, 12:54 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-571556593 "2020-01-07T11:54:09Z - Replied by Github Reply Comments"):
> > Maxime @maxGimeno, can you confirm that?
> 
> Well it was ok in 36, but it is not in 48, so...
> 
> https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-48.shtml#Surface_mesh_topology
> 
> 
> If the two failures are considered as " those tests that randomly fail depending on the seed but it's ok"
> I guess we can consider this PR as successfully tested

In both cases, it is the test `path_with_rle_deformation_tests` that fails on Windows, with: 
> ....................[Test 19] ERROR: we obtained +(2 2 2 2)  -(1 2 2 2) instead of (1 2 2 2)
> .TEST RANDOM PATHS ON CUBE FAILED.

@gdamiand What do you think of those results?

sloriot:
I pushed the filtered version [here](https://github.com/sloriot/cgal/pull/new/Surface_mesh_topology
). It weights 1.4Mo. I mainly removed unused files from the history (added and then removed) and some files that are not used.

@gdamiand  my question on whether another branch contains that branch still hold (I'm thinking of the gsoc branch).

sloriot:
Any reason not to put the doc in the code?

gdamiand:
> > do you approved the new text?
> 
> Sure, looks ok to me.
> (164 files changed? That's one big feature...)

Yes a lot of work !!

gdamiand:
> @gdamiand do you have another branch based on this one?

@sloriot Probably the branch gsoc2019-nccycles-thien (but I am not really sure since we started with the initial branch from the CNRS repo which is now deleted).

gdamiand:
> Any reason not to put the doc in the code?

@sloriot As discussed several times, I think having the doc in the code pollute the code and leads to more complicated files. Personal point of view...

gdamiand:
> [**@maxGimeno**](https://github.com/maxGimeno) commented on [Jan 7, 2020, 12:54 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-571556593):
> 
> > > Maxime @maxGimeno, can you confirm that?
> > 
> > 
> > Well it was ok in 36, but it is not in 48, so...
> > https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-48.shtml#Surface_mesh_topology
> > If the two failures are considered as " those tests that randomly fail depending on the seed but it's ok"
> > I guess we can consider this PR as successfully tested
> 
> In both cases, it is the test `path_with_rle_deformation_tests` that fails on Windows, with:
> 
> > ....................[Test 19] ERROR: we obtained +(2 2 2 2)  -(1 2 2 2) instead of (1 2 2 2)
> > .TEST RANDOM PATHS ON CUBE FAILED.
> 
> @gdamiand What do you think of those results?

@lrineau These results are really strange. I think between 36 and 48 there were no modification of this package; thus I don't understand where this error comes from.

Moreover this test works perfectly on other systems, and locally on my computer, even with the flag -D_GLIBCXX_DEBUG.

Is it possible this comes from a compiler bug ? Is there some modifications of the test suite between 36 and 48 ?

lrineau:
It looks like the test is randomized. Probably the only difference with the other test results is the seed of the random generator.

gdamiand:
There is indeed a CGAL::random but the seeds are fixed; and thus the test is not randomized.

lrineau:
[**@gdamiand**](https://github.com/gdamiand) commented on [Jan 7, 2020, 2:40 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-571590127 "2020-01-07T13:40:10Z - Replied by Github Reply Comments"):
> There is indeed a CGAL::random but the seeds are fixed; and thus the test is not randomized.

Is there a way you can display the random paths, to debug that issue?

gdamiand:
> [**@gdamiand**](https://github.com/gdamiand) commented on [Jan 7, 2020, 2:40 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-571590127):
> 
> > There is indeed a CGAL::random but the seeds are fixed; and thus the test is not randomized.
> 
> Is there a way you can display the random paths, to debug that issue?

Yes; I will add the code in this test and we will see tomorrow the results on the new test suite.

lrineau:
[**@gdamiand**](https://github.com/gdamiand) commented on [Jan 7, 2020, 3:11 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-571602421 "2020-01-07T14:11:50Z - Replied by Github Reply Comments"):
> Yes; I will add the code in this test and we will see tomorrow the results on the new test suite.

Actually tonight will test CGAL-5.0.1, because it has not been tested for a long time. You will see those results in two days.

gdamiand:
@sloriot My last push is not tested in travis; I imagine this is because you created a new branch based on this one ?

sloriot:
The new branch is in my remote, I'm not going to overwrite your branch before getting your green light (and that I checked I can handle the gsoc branch with the update branch)

afabri:
> 
> 
> > Any reason not to put the doc in the code?
> 
> @sloriot As discussed several times, I think having the doc in the code pollute the code and leads to more complicated files. Personal point of view...

But @gdamiand,  you PUT doxygen comments in the code, for example in [`Path_on_surface`](https://github.com/gdamiand/cgal/blob/Surface_mesh_topology-gdamiand/Surface_mesh_topology/include/CGAL/Path_on_surface.h#L104)

gdamiand:
> > > Any reason not to put the doc in the code?
> > 
> > 
> > @sloriot As discussed several times, I think having the doc in the code pollute the code and leads to more complicated files. Personal point of view...
> 
> But @gdamiand, you PUT doxygen comments in the code, for example in [`Path_on_surface`](https://github.com/gdamiand/cgal/blob/Surface_mesh_topology-gdamiand/Surface_mesh_topology/include/CGAL/Path_on_surface.h#L104)

@afabri To simplify the life of users who read the code. 

BUT this is totally different than writing the documentation for the reference manual. We usually need to create fake methods/classes only for doxygen; we also need to hide some parts ... At the end the file is a big mess, it is extremely hard to understand what is real code and what is here only for the manual. Of course, this is not true for some basic cases, for example a non template class with few methods...

sloriot:
The GSoC rebased branch is [here](https://github.com/sloriot/cgal/tree/gsoc2019-nccycles-thien).

@gdamiand There is no blocking point on my side if you agree with my updated branch. List of files removed (ignoring your last commit):
```
> git diff gdamiand/Surface_mesh_topology-gdamiand..sloriot/Surface_mesh_topology --name-only 
Surface_mesh_topology/doc/Surface_mesh_topology/fig/Figure_1.svg
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p38crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p39crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p40crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p41crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p42crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p43crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p44crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p45crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p46crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p47crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p48.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p48crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/p79crop.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/push-right-0-b.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/push-right-0.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/push-right-0b.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/push-right-1.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/push-right-2.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/quadSystem.svg
Surface_mesh_topology/doc/Surface_mesh_topology/fig/shortening-0.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/shortening-1t.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/shortening-3.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/shortening-3b.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/shortening-4.pdf
Surface_mesh_topology/doc/Surface_mesh_topology/fig/surface-mesh-topology-example1.png
Surface_mesh_topology/test/Surface_mesh_topology/data/case4-right-shift-squared.moka
Surface_mesh_topology/test/Surface_mesh_topology/data/case7-right-shift-squared.moka
Surface_mesh_topology/test/Surface_mesh_topology/data/cases5-6-right-shift-squared.moka
Surface_mesh_topology/test/Surface_mesh_topology/data/double-torus-example.off
Surface_mesh_topology/test/Surface_mesh_topology/data/elephant.off
Surface_mesh_topology/test/Surface_mesh_topology/data/small-cylinder-two-borders.off
Surface_mesh_topology/test/Surface_mesh_topology/data/sphere.off
```

Doc and test passes in my branch.

gdamiand:
Thanks @sloriot .
If possible I would like to keep the following files, which can be used in some tests/examples, even if they are not used by default.

Surface_mesh_topology/test/Surface_mesh_topology/data/double-torus-example.off
Surface_mesh_topology/test/Surface_mesh_topology/data/elephant.off
Surface_mesh_topology/test/Surface_mesh_topology/data/small-cylinder-two-borders.off
Surface_mesh_topology/test/Surface_mesh_topology/data/sphere.off

sloriot:
Sure but I'll do it in a commit on top of my rebase (because I don't want to redo the job on the gsoc branch). Let me know when you are done with your checks and I'll overwrite both branches (you'll then have to reset your local branches to avoid a merge of the old history).

gdamiand:
@sloriot  Ok.

gdamiand:
@lrineau What about the compiler error with OpenMesh ?
(supposed to be solved by https://github.com/CGAL/cgal/commit/5565f6ce2df220bddd9f494b19edc07205e594e1 ?)

(cf. https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-48/Combinatorial_map/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz)

Installation/cmake/modules/FindOpenMesh.cmake
```diff
@@ -44,7 +46,34 @@ endif()
...
+  add_library(OpenMesh::OpenMesh UNKNOWN IMPORTED)
+
+  if(TARGET OpenMeshCore)
+    target_link_libraries(OpenMesh::OpenMesh PUBLIC OpenMeshCore)
+    return()
+  endif()
+
+  set_target_properties(OpenMesh::OpenMesh PROPERTIES
+    INTERFACE_COMPILE_DEFINITIONS "CGAL_USE_OPENMESH;NOMINMAX;_USE_MATH_DEFINES"
+    INTERFACE_INCLUDE_DIRECTORIES  "${OPENMESH_INCLUDE_DIRS}")
```
lrineau: There was a type in my previous commits:
```suggestion
    INTERFACE_INCLUDE_DIRECTORIES  "${OPENMESH_INCLUDE_DIR}")
```
I will fix it as soon as I know which version of the branch is the right one.

sloriot:
The branches are now update. I created the following tags to keep the old branches:
[branch](https://github.com/CGAL/cgal-public-dev/releases/tag/archive%2FSurface_mesh_topology-gdamiand_before_history_filtering) and [gsoc](https://github.com/CGAL/cgal-public-dev/releases/tag/archive%2Fgsoc2019-nccycles-thien_before_history_filtering).

@gdamiand : you have to fetch `gdamiand` and hard reset your local branches.
I'll push the commit with the remove off files in a minute.

gdamiand:
Thanks @sloriot  and @lrineau  !

gdamiand:
The test suite https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-49.shtml confirms that the test path_with_rle_deformation_tests generates different paths on the os that can have an error during the execution of the program.

Is it a normal behavior of CGAL::Random ? More precisely if I use a CGAL::Random with a given seed, is it supposed to give me the same sequence of numbers ?

lrineau:
[**@gdamiand**](https://github.com/gdamiand) commented on [Jan 9, 2020, 3:45 PM GMT+1](https://github.com/CGAL/cgal/pull/4042#issuecomment-572592826 "2020-01-09T14:45:45Z - Replied by Github Reply Comments"):
> The test suite https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-49.shtml confirms that the test `path_with_rle_deformation_tests` generates different paths on the os that can have an error during the execution of the program.
> 
> Is it a normal behavior of `CGAL::Random` ? More precisely if I use a `CGAL::Random` with a given seed, is it supposed to give me the same sequence of numbers ?

I do not think that our `CGAL::Random` guaranties the same sequence with a given seed across different OS or architectures.

Anyway, that shows a bug in the implementation of the package. You can now debug on you machine, with the path that exposes the bug.

gdamiand:
@lrineau No this does not show a bug in the package. After the simplification on the initial path, I obtain a different simplified path than with my computer: this is normal since the first path is not the same.

I just need to use a different method to generate the initial path which is not based on random.

gdamiand:
@lrineau Should be good with https://github.com/CGAL/cgal/pull/4042/commits/2ca8651c661cce170715c9f2bdfec28da4c6ae74.

@maxGimeno Is it possible to add this pr in the test suite tonight ? Thanks.

sloriot:
Just did.

gdamiand:
Thanks @sloriot.

maxGimeno:
It looks all good : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-50.shtml#Surface_mesh_topology

gdamiand:
> It looks all good : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-50.shtml#Surface_mesh_topology

Thanks !
This package will be integrated in CGAL 5.1 ? (I need to update this number in PackageDescription.txt)

maxGimeno:
It should be, yes.

PR #4043, opened 2019-06-28
CTest: Add a new ctest fixture, that checks the build system

lrineau:
## Summary of Changes

Add a new ctest fixture, that checks the build system. That will avoid that `cmake` is run in parallel several times, when `ctest -j` is used.

@maxGimeno That may explain why you found CTest unreliable, so far. Now, CTest will always start by checking if `cmake` must be run before starting the tests.

## Release Management
* Affected package(s): CMake

I think this pull-request cannot really be tested, unless perhaps in the test platform using CTest instead of `autotest_cgal`.


lrineau:
I have modified the PR to apply it to CGAL-4.13-branch.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-73.shtml

PR #4050, opened 2019-07-03
[Surface Mesh / PLY] Fix uninitialized warning and error on unhandled properties

sgiraudot:
## Summary of Changes

* Fix uninitialized warning (see [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-I-72/Polyhedron_Demo/TestReport_lrineau_ArchLinux-CXX17-Release.gz)). Code introduced in 5.0.
* Fix error https://github.com/CGAL/cgal/issues/4067 that happens when a property is found on edges and halfedges, but then its type is not handled, which leads to the header being filled _but_ the properties not written (which leads to an error when reading the produced PLY). Now, the header for edges and halfedges are only written if the properties were actually found with an handled type.

## Release Management

### DIFF
--- a/Stream_support/include/CGAL/IO/PLY.h
+++ b/Stream_support/include/CGAL/IO/PLY.h
@@ -375,6 +375,7 @@ public:
         t = property->buffer();
         return;
       }
+    t = {};
   }
 
   template <typename Type>
@@ -389,6 +390,7 @@ public:
         t = property->buffer();
         return;
       }
+    t = {};
   }
 
   void assign (double& t, const char* tag)
@@ -410,6 +412,7 @@ public:
           
         return;
       }
+    t = {};
   }
   
 };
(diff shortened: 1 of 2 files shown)


Surface_mesh/include/CGAL/Surface_mesh/IO/PLY.h
```diff
@@ -305,7 +305,7 @@ class Surface_mesh_filler
 
     if (m_vcolors == 3)
     {
-      unsigned char r, g, b;
+      unsigned char r = 0, g = 0, b = 0;
```
lrineau: I think the overloads of `PLY_element::assign` should all end by `t = {}` to fill `t` with a default-constructed value, in case the property is missing. That would also fix that sort of warning.

sgiraudot: That's a good idea, I changed my fix by that.

sgiraudot:
This PR now also fixes https://github.com/CGAL/cgal/issues/4067.

maxGimeno:

PR #4052, opened 2019-07-03
Polyhedron_Demo: Fixes

### LINKED ISSUE #4019: Polyhedron demo issues
## Issue Details

- [x] Open mesh with self-inters, create a selection item, select a few faces, call "self-intersections detection" --> the action from the selection item is being applied on the mesh where your mouse is (e.g. if you're still on face selection, it'll apply face selections without having to shift-click).
- [x] Load mesh with self-intersections, create selection item, call "Self-inter Test", Reload mesh -> Crash
- [x] Calling "Self-inter Test" shows the highlights as if you were manipulating a selection item (might need to have constructed a selection item on the object before).
- [x] Order of items in the "Operations" menu changes arbitrarily every time the demo is recompiled.
- [x] Give a summary of what "clean polygon soup" achieved (it's so silent currently)
- [x] Generate objets instead of ObjeCts
- [x] Load a far triangle, call cut plugin -> crash
- [x] holes selection with CTRL is not working very well
- [x] Selection item doesn't actualize when the mesh is modified by something other than the selection item itself.
- [x] join faces in selection item is very broken

maxGimeno:
## Summary of Changes

A few fixes.

## Release Management

* Affected package(s): CGAL 3D demo
* Issue(s) solved (if any): fix #4019


Polyhedron/demo/Polyhedron/Scene_polygon_soup_item.cpp
```diff
@@ -25,6 +26,8 @@
 #include <CGAL/Polygon_mesh_processing/orient_polygon_soup.h>
 #include <CGAL/Polygon_mesh_processing/orientation.h>
 #include <CGAL/Polygon_mesh_processing/repair.h>
+
+#define CGAL_PMP_REPAIR_POLYGON_SOUP_VERBOSE 1
```
lrineau: Is that wanted?

maxGimeno: Yes, that triggers the debug info in the repair functions.

MaelRL: That was my suggestion for a zero-effort way to at least get a little bit of info on what is happening when you click "repair polygon soup" in the demo (otherwise it's just completely silent and it's difficult to see what was done - or not). Note that I use another macro for "spammy" debug output in `repair_polgyon_soup.h`.

lrineau: Thanks for the explanation.

Polyhedron/demo/Polyhedron/Plugins/IO/CMakeLists.txt
```diff
@@ -76,9 +76,15 @@ endif()
...
+  PATHS /usr/include
+        /usr/local/include
   DOC "Path to lib3MF headers"
   )
-find_library(3MF_LIBRARIES NAMES 3MF DOC "Path to the lib3MF library")
+find_library(3MF_LIBRARIES
+  NAMES 3MF DOC
+  PATHS /usr/lib
+  /usr/local/lib
+  /usr/lib64"Path to the lib3MF library")
```
lrineau: This patch is wrong:
  - syntax error,
  - anyway you are added hints to places that already searched for by CMake.

https://cmake.org/cmake/help/v3.9/command/find_package.html

maxGimeno: @sloriot complained that they were missing, so I added the same as the ones in our FindIPE. If those are useless, then FindIPE has the same problem.

maxGimeno:

PR #4053, opened 2019-07-04
BGL: Fix output of "is_valid_polygon_mesh" function family

MaelRL:
## Summary of Changes

Some error messages were not printed despite failure. Some error messages were printed wrongly since as soon as 'valid' was wrong, everything below was printed, even if the test was not actually failing.

## Release Management

* Affected package(s): `BGL`


MaelRL:
Although it was marked as not being tested, this seems to have been put in the testsuite, but with some problems during the merge into integration.

maxGimeno:
I removed the tag this morning. Yes I messed up the integration, sorry :/

lrineau:
Tested through its equivalent PR #4054.

PR #3987, opened 2019-06-07
Polyhedron demo:  mesh several surfaces at a time

lrineau:
## Summary of Changes

Make possible to mesh a bounded volume with extra surfaces inside, like the example `examples/Mesh_3/mesh_polyhedral_domain_with_surface_inside.cpp`.

## Release Management

* Affected package(s): Mesh_3, Polyhedron Demo
* License and copyright ownership: maintenance

PR #4056, opened 2019-07-05
Fix PR #3987 with "mesh surface only"

lrineau:
## Summary of Changes

Fix PR #3987 with "mesh surface only". That was a bug in #3987 merged in 4.14.1-dev. If one tries "mesh surface" then there is an assertion `!tree.empty()`:
```
Expr: !tree_.empty()
File: [...]/Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
Line: 649
```

## Release Management


maxGimeno:

PR #4063, opened 2019-07-08
Remove CGAL_CXX11 macro usage

sloriot:
Warning: not tested

### DIFF
--- a/NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Cartesian_wrap.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Cartesian_wrap.h
@@ -65,25 +65,6 @@ template<class T> struct Is_wrapper_iterator<T,true> :
 { };
 
 struct Forward_rep {
-//TODO: make a good C++0X version with perfect forwarding
-//#ifdef CGAL_CXX11
-//template <class T,class=typename std::enable_if<!Is_wrapper<typename std::decay<T>::type>::value&&!Is_wrapper_iterator<typename std::decay<T>::type>::value>::type>
-//T&& operator()(typename std::remove_reference<T>::type&& t) const {return static_cast<T&&>(t);};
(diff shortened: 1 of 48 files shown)


maxGimeno:

NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Cartesian_wrap.h
```diff
@@ -65,25 +65,6 @@ template<class T> struct Is_wrapper_iterator<T,true> :
 { };
 
 struct Forward_rep {
-//TODO: make a good C++0X version with perfect forwarding
-//#ifdef CGAL_CXX11
```
mglisse: Er, there is indeed a test on CGAL_CXX11 here, but it is commented out (the C++11 version doesn't quite work), and you are removing this C++11 version and the TODO. That doesn't seem right.

Polyhedron/demo/Polyhedron/Plugins/Point_set/Point_set_normal_estimation_plugin.cpp
```diff
@@ -24,12 +24,6 @@
 
 #include "ui_Point_set_normal_estimation_plugin.h"
 
-#if BOOST_VERSION == 105700
```
sloriot: I think this is not correct too. @sgiraudot do you remember why we disable this plugin with boost 1.57 and C++11? Is it still valid? Actually @lrineau did bd435457178c23a2363eb9e05df1f8f63175660d. Shall I revert the fix?

lrineau: Four years ago I should I known better and write comments so that even myself can understand the patch a few years later. But did not, and I do not remember the issue.

I suspect it was a runtime error somewhere in Boost-1.57.

sloriot: I decided to ignore the issue, it is in the demo.

PR #4068, opened 2019-07-09
Scripts: Fix autotest_with_ctest

maxGimeno:
Move logging functions in the script to avoid having to ask for CGAL-I before it is set.

## Release Management

* Affected package(s):Scripts


lrineau:
I think that should be against 4.14-branch.

maxGimeno:

PR #4069, opened 2019-07-09
Polyhedron_IO: OBJ_reader reads relative indices

### LINKED ISSUE #4066: Handle negative indices while reading OBJ
I think it is:
https://en.wikipedia.org/wiki/Wavefront_.obj_file#Vertex_indices
> If an index is positive then it refers to the offset in that vertex list, starting at 1. If an index is negative then it relatively refers to the end of the vertex list, -1 referring to the last element.

_Originally posted by @lrineau in https://github.com/CGAL/cgal/issues/4062#issuecomment-509539166_

maxGimeno:
## Summary of Changes

If an index is negative in a face of an OBJ file, reads it relatively to the current state of the points vector instead of brutally crashing.
## Release Management

* Affected package(s):Polyhedron_IO
* Issue(s) solved (if any): fix #4066

### DIFF
--- a/Polyhedron_IO/include/CGAL/IO/OBJ_reader.h
+++ b/Polyhedron_IO/include/CGAL/IO/OBJ_reader.h
@@ -48,7 +48,13 @@ read_OBJ( std::istream& input,
       faces.push_back( std::vector<std::size_t>() );
       while(iss >> i)
       {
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #4075, opened 2019-07-10
BGL: fix memory leak in test

lrineau:
## Summary of Changes

Fix a memory leak in `test/BGL/`, so that one can run CTest with ASAN enabled without false-positives.

## Release Management

* Affected package(s): BGL

For `master` only, because I have used C++14 (`std::make_unique`) to fix the issue.


maxGimeno:

PR #4077, opened 2019-07-11
Demo 3D : Optimize the "Move point" operation

### LINKED ISSUE #4072: CGAL Demo: Quad meshes and selection item interactions are very slow
For example, `move point` pretty much freezes the demo. It would be nice if those operations were usable on quad meshes.

`move_point` is still also crashing the demo a good 20% of the moves I attempt.

maxGimeno:
Only update locally the geometry of the item when moving a vertex around. Great gain in cpu time, but still only working in Release Mode for big items, because of the aabb_tree. (Any call is to it is really long in debug, so highlighting alone is a killer for the plugin in debug mode).

TODO: Factorize
## Release Management
* Issue(s) solved (if any): fix #4072


maxGimeno:
[warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0-Ic-82/Polyhedron_Demo/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz)

lrineau:
The code of this PR looks unmaintainable. At least all the almost identical blocks should be in a function or a lambda. And maybe the factorization should be at a higher lever.

```c++
          vbo = &(getTriangleContainer(1)->getVbo(Tri::Flat_normals)->vbo);
          getTriangleContainer(1)->getVbo(Tri::Flat_normals)->bind();
          vbo->write(static_cast<int>((3*size*id+vid)*sizeof(cgal_gl_data)),new_normal, static_cast<int>(3*sizeof(cgal_gl_data)));
          getTriangleContainer(1)->getVbo(Tri::Flat_normals)->release();
          vbo = &(getTriangleContainer(1)->getVbo(Tri::Flat_vertices)->vbo);
          getTriangleContainer(1)->getVbo(Tri::Flat_vertices)->bind();
          vbo->write(static_cast<int>((3*size*id+vid)*sizeof(cgal_gl_data)),new_point, static_cast<int>(3*sizeof(cgal_gl_data)));
          getTriangleContainer(1)->getVbo(Tri::Flat_vertices)->release();
```

[Always code as if the future maintainer of the code is a violent psychopath who knows where you live][1]. Even if that future maintain is yourself a few months later. You will be angry with that piece of code the same way as if it was not yours.

[1]: https://stackoverflow.com/q/876089/1728537

Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
```diff
@@ -843,7 +844,13 @@ public Q_SLOTS:
   {
     Scene_polyhedron_selection_item* selection_item = getSelectedItem<Scene_polyhedron_selection_item>();
     if(selection_item)
+    {
       selection_item->on_Ctrlz_pressed();
+      if(mode == 11)
```
lrineau: Is not it dangerous to hardcode a value `11`, instead of using a enumerator?

lrineau:
The code is easier to read, now. Thanks.

Polyhedron/demo/Polyhedron/Plugins/PMP/Selection_plugin.cpp
maxGimeno: Using 11 is dangerous because if the list of parameters of the widget is modified it will become the wrong number. Using an enum would do exactly the same. It would just make it clearer what 11 is supposed to be, but I commented everywhere what number is supposed to do.

maxGimeno:

PR #4081, opened 2019-07-12
Move store outside the loop.

mglisse:
## Summary of Changes

Looks like a typo kind of mistake: the value was computed and stored repeatedly in the inner loop instead of once in the outer loop... At least it was correct, if wasteful.

## Release Management


maxGimeno:

sloriot:
@mglisse should it go into 4.14.1 or even 4.13.2?

mglisse:
I don't think it needs a backport. As I said, the code was correct, it was just uselessly doing the same operation several times. Fixing it for 5.0 seems sufficient.

PR #4083, opened 2019-07-15
Named function parameters

### LINKED ISSUE #3706: Replace/reimplement boost BGL named parameters
I've read this morning that there is an on going effort to replace boost named parameters in the BGL using "revamped Boost.Parameter interface".

Someone has to investigate what is the "revamped Boost.Parameter interface" and worse case, we should have our implementation of named parameters to avoid breaking when boost will remove the base class.

sloriot:
Get away from boost named function parameters than are not compatible with CGAL anymore in the upcoming release

This is a branch against 4.13 that we have to test first, I'll do the upcoming PRs for 4.14 and master after this one is integrated.

@maxGimeno : I suggest that this branch is tested alone in integration (i.e. reset integration to this branch)

Fixes #3706


maxGimeno:
[errors](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0.2-Ic-85.shtml)

MaelRL:
ae44cf3 should fix the errors.

BGL/include/CGAL/boost/graph/Named_function_parameters.h
```diff
@@ -204,8 +204,7 @@ get_parameter(const Named_function_parameters<T, Tag, Base>& np, Query_tag tag)
 }
 
 template <typename D>
-const D&
-choose_parameter(const internal_np::Param_not_found&, const D& d)
+D choose_parameter(const internal_np::Param_not_found&, const D& d)
 {
   return d;
 }
```
lrineau: Is not that a place where we should use move-semantic and perfect forwarding, like in the new version of Boost.Parameters?

With the current implementation, the argument passed in the parameter `d` is guaranteed to be copied. Were `d` passed by value, instead of reference to const, then the move semantic could be used if `D` is heavy.

On the other hand, in CGAL BGL API, I think all our named parameters are actually either lightweight property maps, or trivial parameters (type `int`, `double`, or at worse a small struct).

MaelRL: I was thinking about it some other day with the fact that you could have as default value a traits class that wouldn't like copies due to some internal members. It was quite specific, but I guess moving it would be a good solution.

In any case, this targets 4.13 so it just needs to copy here.

lrineau: It targets 4.13 but 4.13 does not prevent users from using C++11 or later. C++14 is the default since gcc 6 and MSVC 2015, for example.

But I agree with can think about that later, once we have a real use case to test with.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-4.13.2-Ic-94.shtml

lrineau:
This branch conflicts a lot with `master`. I cannot merge it until we have an equivalent for the other branches.

Issue #3684, opened 2019-02-13
Bug in PCA 3D for tetrahedra


palliez:
## Issue Details

The fitting plane is undefined for a union of tets that are well defined!!
when the tag dimension is 3.
when the dimension is 2 it works.. 

## Source Code
```c++
Point a(0.0, 0.0, 0.0);
Point b(0.0, 1.0, 0.0);
Point c(1.0, 1.0, 0.0);
Point d(1.0, 0.0, 0.0);
Point e(0.5, 0.5, 4.0);
Tetrahedron tet1(a, c, b, e);
Tetrahedron tet2(a, c, d, e);
std::list<Tetrahedron> tets;
tets.push_back(tet1);
tets.push_back(tet2);

Plane plane;
CGAL::linear_least_squares_fitting_3(tets.begin(), tets.end(), plane, CGAL::Dimension_tag<3>());
Vector n = plane.orthogonal_vector();
std::cout << "n: " << n << std::endl;
std::cout << "plane: " << plane << std::endl;
```
## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): windows
* Compiler: Visual C++ 2017
* Release or debug mode: release
* Specific flags used (if any):
* CGAL version: 4.13
* Boost version: 1,67
* Other libraries versions if used (Eigen, TBB, etc.): eigen 3

sgiraudot:
Ok, this is worse than what I originally thought: I think I can safely say that PCA for Tetrahedra never worked in CGAL. 

I already found 2 major bugs due to the fact that `Tetrahedron_3::volume()` is considered as unsigned in the PCA package, whereas it really isn't (so there should be a `CGAL::abs()` around it):
* The error is done in `assemble_covariance_matrix_3()` for Tetrahedron
* It's also done in `CGAL::centroid()` for Tetrahedron (yes, even the centroid of a range of tetrahedra did not work and could produce `nan` coordinates…), which is used for assembling the covariance matrix too.

What worries me is: I fixed these errors, the `nan` are thus gone and the returned plane has valid attibrutes… but if we display the plane, it clearly _isn't_ the PCA of the tetrahedra (even when reducing to a single tetrahedron very elongated/flat, the plane does not seem related at all to it). So my guess is, there is something wrong in the way we compute the covariance matrix.

palliez:
Thanks much Simon for your time on this.
Let me investigate too.... I have a prototype code too that worked for sure for unions of tets.

PR #4093, opened 2019-07-17
[PCA] Fix PCA on tetrahedra

sgiraudot:
## Summary of Changes

The variant of PCA on tetrahedra is broken, the returned plane/line has no meaning and is obviously wrong. It's been this way for a very long time, I tracked the bug back to CGAL 4.7 (I did not have the faith to go farther...). The problem is we only test compilation+execution on this package, not the validity of the result from a geometric point of view. We know for sure that the variants of PCA on points and triangles work because we (I for starters) use them regularly, but the other variants might be broken too as far as we know.

At some point, we want to add validity tests, but I discussed it with @palliez in March and we did not have time to work on those tests since then. In the mean time, Pierre found back his old prototype code which works well, so we thought we should at least push this fix to CGAL even without additional tests.

## Release Management

* Affected package(s): PCA
* Fix https://github.com/CGAL/cgal/issues/3684

### DIFF
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -37,12 +37,12 @@ template < typename InputIterator,
 void
 assemble_covariance_matrix_3(InputIterator first,
                              InputIterator beyond,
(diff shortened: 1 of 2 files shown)


MaelRL:
Fix for https://github.com/CGAL/cgal/issues/3684?

sgiraudot:
Indeed. I did not remember that Pierre filled a public issue. Thanks!

maxGimeno:

PR #4095, opened 2019-07-18
Box_intersection_d: Add an example that implements a do_intersect for two polylines

afabri:
## Summary of Changes

We add an example that tests if two polylines intersect. As the Polygon package is LGPL, we cannot add a dependency on a GPLed package.   

## Release Management

* Affected package(s): Box_intersection_d


maxGimeno:

PR #4099, opened 2019-07-18
Speed up comparison of points

sloriot:
With this simple trick I speed-up an example sorting EPECK points from 15s to 12s (and the sorting is for sure not the central part of the bench program).

First question, is it the right place for this fix?
Do you think we could have something more general, as other primitive could benefit from that?

cc @afabri @mglisse


mglisse:
> With this simple trick I speed-up an example sorting EPECK points from 15s to 12s (and the sorting is for sure not the central part of the bench program).

In what context do you end up sorting points where there are many copies of the same points? Do the copies appear as vertices of different edges/triangles? My first reaction is to wonder if this case could be handled at the combinatorial level, without calling the predicate at all.

This kind of check makes sense for Epeck, and indeed we already do it when comparing Lazy_exact_nt, so why not when comparing points. It makes less sense for Epick. Maybe it could look like `if(AST<K::FT>::Is_exact::value && identical(p,q))` with Is_exact as a proxy for Is_expensive, so it is done for Epeck but not Epick?

I don't immediately see a better place to add this.

(I noticed that `u<0` gives smaller code than `u==-1` for u of type `Uncertain<Sign>`, because the compiler doesn't know that Sign cannot be < -1)

sloriot:
> > With this simple trick I speed-up an example sorting EPECK points from 15s to 12s (and the sorting is for sure not the central part of the bench program).
> 
> In what context do you end up sorting points where there are many copies of the same points? Do the copies appear as vertices of different edges/triangles? My first reaction is to wonder if this case could be handled at the combinatorial level, without calling the predicate at all.
> 
Of course if I could do this combinatorially that would be best but for now I don't have this information and have to rely on comparing the points (I have triangles meshes with some identical vertices. Think about meshes that were corefined for exemple).


> This kind of check makes sense for Epeck, and indeed we already do it when comparing Lazy_exact_nt, so why not when comparing points. It makes less sense for Epick. Maybe it could look like `if(AST<K::FT>::Is_exact::value && identical(p,q))` with Is_exact as a proxy for Is_expensive, so it is done for Epeck but not Epick?
> 
> I don't immediately see a better place to add this, although there are a number of choices. For Epeck, do you want to check for identical before the static filter, before the interval filter, or before the exact computation?
> 
My guess is that is should be the first thing to be done.

> (I noticed that `u<0` gives smaller code than `u==-1` for u of type `Uncertain<Sign>`, because the compiler doesn't know that Sign cannot be < -1)

mglisse:
> Of course if I could do this combinatorially that would be best but for now I don't have this information and have to rely on comparing the points (I have triangles meshes with some identical vertices. Think about meshes that were corefined for exemple).

Well, sometimes points are actually indices, and then it is obvious which points are the same. Not here though, ok.

> My guess is that is should be the first thing to be done.

Then a static filter might be where you want to do it? Unless you remove the static filter.

I am not sure about using `identical` on points with interval coordinates. It seems a bit unsafe, but I can't think of circumstances where that would matter.

Looking a bit at the code generated by Epeck::Less_xyz_3, I notice a few things. The static filter seems almost redundant with what the dynamic filter should be doing. However, the current dynamic filter sucks. First, I guess setting the rounding mode is not necessary if there are only comparisons and no arithmetic (although we never know if implementing operator< with a subtraction might become interesting in the future, but then maybe it would work with any rounding mode?). Second, if the x are not comparable, instead of returning uncertain, it throws. I think Equal_3 could use CGAL_AND; for Less_xyz_3 it is a bit more awkward.

mglisse:
The following does not remove the need for something like your approach, but it seems like a step in the right direction.
```patch
--- a/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
+++ b/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
@@ -70,8 +70,10 @@ compare_lexicographically_xyzC3(const FT &px, const FT &py, const FT &pz,
 {
   typedef typename Compare<FT>::result_type Cmp;
   Cmp c = CGAL_NTS compare(px, qx);
+  if (is_indeterminate(c)) return indeterminate<Cmp>();
   if (c != EQUAL) return c;
   c = CGAL_NTS compare(py, qy);
+  if (is_indeterminate(c)) return indeterminate<Cmp>();
   if (c != EQUAL) return c;
   return CGAL_NTS compare(pz, qz);
 }
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -3102,7 +3102,7 @@ namespace CommonKernelFunctors {
     result_type
     operator()(const Iso_rectangle_2& i1, const Iso_rectangle_2& i2) const
     {
-      return ((i1.min)() == (i2.min)()) && ((i1.max)() == (i2.max)());
+      return CGAL_AND ( (i1.min)() == (i2.min)() , (i1.max)() == (i2.max)() );
     }
   };
 
@@ -3129,7 +3129,7 @@ namespace CommonKernelFunctors {
     result_type
     operator()(const Point_3 &p, const Point_3 &q) const
     {
-      return p.x() == q.x() && p.y() == q.y() && p.z() == q.z();
+      return CGAL_AND_3( p.x() == q.x() , p.y() == q.y() , p.z() == q.z() );
     }
 
     result_type
```

afabri:
I just added the following test:   I generate a vector of points with blocks of 10 consecutive points, with decreasing x-coordinate, and I call `std::sort()`.     I see no difference, and if I understand @mglisse comments right that is not surprising as the static filters already do the job, and the identity test is never performed.   This would probably be different for points which got really constructed (what I test next).

afabri:
Now, with points where `inf() != sup()`  for the approximated points identical kicks in and we speak about 12 sec instead of 70sec.

afabri:
Hello @mglisse, After a discussion with Sebastien our proposal is to add a function similar to `identical()` which checks if the pointee of the `Handle`  is identical. We could call it `has_identical_pointee()`. At the same time we offer an overload for all other types that is a `constexpr`  and returns `false`  so that we can add the tests in  `CGAL/Cartesian/function_objects.h` and the compiler can optimize them away.
We propose to be pragmatic and add them where we see a performance problem.  It does not seem important to us to do the identity check on the level of the static or dynamic filters as they are not that costly.

We will look at your other suggestions for improvements in independent PRs.   Concerning `Less_xyz_3` we should write a static filter for it, as we did for example for `Compare_x_3`.  Remark:  The fact that it is not there shows that a systematic approach would have been better than a pragmatic one, but we have to manage resources.

mglisse:
> After a discussion with Sebastien our proposal is to add a function similar to `identical()` which checks if the pointee of the `Handle` is identical. We could call it `has_identical_pointee()`.

Some differences with `identical` IIUC:
- the default is false instead of an address comparison
- it does not look at Handle_for (?)
- maybe it handles types that derive from Handle, not just Handle itself?

I think it is good **not** to do the address comparison for points with interval coordinates since the result might be different (I hope we don't have a variant of `Cartesian<Interval_nt>::Point_3` using Handle instead of Handle_for...). Checking for identity with Cartesian<Gmpq> would also be fine.

> At the same time we offer an overload for all other types that is a `constexpr` and returns `false`

`constexpr` is nice, but I don't think it would affect codegen here (well, maybe in debug mode if the code is strangely written).

> so that we can add the tests in `CGAL/Cartesian/function_objects.h` and the compiler can optimize them away.
> We propose to be pragmatic and add them where we see a performance problem. It does not seem important to us to do the identity check on the level of the static or dynamic filters as they are not that costly.

Your testcase becomes epsilon faster if you do the test before static filters, but that's not very significant.
I am not sure I understand the suggestion exactly. I didn't check, but I think with Sébastien's current patch the `identical` test happens by checking the address of the approximate points. With your new identical_handle, I expect the test will never happen anymore. Or are you planning to change other things?

> Concerning `Less_xyz_3` we should write a static filter for it, as we did for example for `Compare_x_3`.

I looked at Compare_x_2, which has a static filter, but when I use Epeck, it actually goes through Epic_converter instead, which also checks if both `y` coordinates can be converted to `double` exactly, that doesn't look very useful.

afabri:
I didn't understand a word of what you wrote Marc, before I today started looking at it under a debugger... and I was surprised.  In fact the overload for handles is not used at all.  Instead the version that compares addresses is used for the `Simple_cartesian<Interval_nt>`  point that serves as `approx`,  and these addresses are equal for two handles of `Epeck`  points with the same `Rep`.

Filtered_kernel/include/CGAL/Lazy_kernel.h
```diff
@@ -480,6 +480,18 @@ class Lazy_kernel_base
 
   };
 
+  struct Less_xyz_3 : public BaseClass::Less_xyz_3
+  {
+    typedef typename Kernel_::Point_3 Point_3;
+
+    using BaseClass::Less_xyz_3::operator();
```
mglisse, line 487: What other calls are you trying to catch with this?

mglisse (approved):

Filtered_kernel/include/CGAL/Lazy_kernel.h
sloriot: Indeed it is useless since I expended it in the operator().

maxGimeno:

PR #4101, opened 2019-07-19
Demo 3D:Diff Between Meshes Plugin

### LINKED ISSUE #4091: Mesh diff plugin
Add a plugin that is the version for meshes of  `diff`. Of course, we do not care about the reordering of simplices only the final mesh.

Basically, you take two meshes, and 2 set of simplices are generated: those that are not in the first and vice-versa. Maybe the common part is interesting too.

The way to implement it is to first match vertices (the new hash function might help here!) then I think you only need to collect incident faces (I did not think too much about it so I might miss some).

Usage: after simplification, after mesh repair, ...

maxGimeno:
## Summary of Changes

Add a plugin that computes the diff between two meshes, and creates 3 meshes: the difference between m1 and m2, the difference between m2 and m1, and the common part.

## Release Management

* Issue(s) solved (if any): fix #4091


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-100.shtml

PR #4102, opened 2019-07-19
3D Demo: Fixes

### LINKED ISSUE #4110: Reload issue with the CGAL demo
Steps: Load polygon mesh, shuffle, orient, reload doesn't reload (can be seen easily if you modify the mesh)

### LINKED ISSUE #4127: Flipping boundary edges crashes the demo
cf title.

maxGimeno:
Fixes #4110, fixes #4127 
- Fix the reloading of an oriented soup. 
- Restore highlighting of self-intersections.

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.h
+++ b/Polyhedron/demo/Polyhedron/MainWindow.h
@@ -360,7 +360,7 @@ protected Q_SLOTS:
     QString cameraString(CGAL::Three::Viewer_interface *v) const;
   //!Hides not available operations and show available operations in all the
   //!menus.
-  void filterOperations();
+  void filterOperations(bool hide);
   //!Updates the bounding box and moves the camera to fits the scene.
   void recenterScene();
   //!Resizes the header of the scene view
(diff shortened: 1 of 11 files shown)


maxGimeno:

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -360,7 +360,7 @@ protected Q_SLOTS:
     QString cameraString(CGAL::Three::Viewer_interface *v) const;
   //!Hides not available operations and show available operations in all the
   //!menus.
-  void filterOperations();
+  void filterOperations(bool hide);
```
sloriot: @lrineau do you agree with this API change?

lrineau: I generally no longer like Boolean parameters. But this one is not part of any API. That is more like an ugly implementation detail. I cannot say I agree with the change, but I do not object to its merge.

maxGimeno: It IS an ugly workaround for a HELL of a bug, where one implementation only works on Windows and the other doesn't, and I couldn't find any reason why it doesn't work on windows, so I had to resolve to do it like that.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-110.shtml

PR #4108, opened 2019-07-24
Demo: update reconstruction plugin

sgiraudot:
## Summary of Changes

This PR mainly aims at adding the possibility to use the new [Polygonal Surface Reconstruction package](https://github.com/CGAL/cgal/pull/3984) in the demo, but I also took advantage of it to:
- clean up the reconstruction plugin that really needed it… I removed many experimental things that never really turned out to be useful, and recentered it on reconstruction only, removing artificial dependencies to shape detection, etc. The plugin compiles much faster now.
- also reworked the widget to be less of a mess
- added a marching tets implementation as an alternative to the Surface Mesher for Poisson, as I had it in my personal files and found it quite practical as it allows to quickly get a (low-to-medium quality) mesh with guarantee of success. I know that @palliez wants to, at some point, add a generalized version of the marching tets to CGAL: when that happens, we'll just replace this internal implementation

## Release Management

* Affected package(s): Demo


sgiraudot:
It's not critical at all, but it would be nice to have this PR in 5.0 (it only affects the demo). I just flagged it as well (feel free to postpone to 5.1 if it's not possible).

maxGimeno:
[Errors](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-104.shtml#Polyhedron_Demo).

sgiraudot:
The error (+ a warning) should be fixed.

sgiraudot:
Many warnings are coming from the new _Polygonal Surface Reconstruction_ package which is barely tested (as it relies on very specific dependencies that are not installed everywhere). As only one function relies on these dependencies, it's possible to test the rest even on platform that don't have them (which is what happens with the reconstruction plugin that makes these warnings appear).

I modified the test+CMakeLists of the new package so that it is the case. Hopefully we'll get a more useful testsuite for it now.

Polyhedron/demo/Polyhedron/Plugins/Point_set/CMakeLists.txt
```diff
@@ -1,8 +1,30 @@
...
+
+  find_package(SCIP QUIET)
+  if(SCIP_FOUND)
+    include_directories(BEFORE ${SCIP_INCLUDE_DIRS})
+    set(polyfit_linked_libraries ${polyfit_linked_libraries} ${SCIP_LIBRARIES})
+    set(polyfit_compile_definitions ${polyfit_compile_definitions} "-DCGAL_USE_SCIP")
+  endif()
+  find_package(GLPK QUIET)
+  if(GLPK_FOUND)
+    include_directories(BEFORE ${GLPK_INCLUDE_DIRS})
```
maxGimeno: it is GLPK_INCLUDE_DIR, no `S`. That's why it doesn't work on my machine.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-107.shtml

PR #4116, opened 2019-07-26
Change ray_23 and line_23's point(int) function to point(FT)

### LINKED ISSUE #4113: Ray_2's point generation interface, `point(...)`, is too restrictive.
## Issue Details

Ray_2's point generation interface, `point(...)`, is too restrictive -- it only takes an integer parameter as input. This is also the case with Ray_3.

## Source Code

Please have a look at line 107 of Ray_2.h and observe that the interface is overly restrictive, especially given the fact that on line 119, the parameter, `int i`, is converted to `FT(i)` anyway.

## Environment

* All

MaelRL:
## Summary of Changes

There is no real requirement on `i` being an 'int', except for some symmetry with the segment. On the other hand, it restricts heavily what you can do with the function for no reason.

Ideally, segment_23 would also return a point on the segment with `0<=i<=1`, but it would be a strong breaking change (right now it expects an int and applies a %2 before calling source() or target()). There are not enough reasons to break such low level code.

Put as a bug/cleaning fix, but could (should be a small feature ?)

## Release Management

* Issue(s) solved (if any): fix #4113

### DIFF
--- a/Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
+++ b/Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
@@ -3098,7 +3098,16 @@ namespace HomogeneousKernelFunctors {
     }
 
     Point_2
-    operator()(const Line_2& l, const FT i) const
+    operator()(const Line_2& l, const RT& i) const
+    {
+      Point_2 p = K().construct_point_2_object()(l);
+      Vector_2 v = K().construct_vector_2_object()(l);
+      return K().construct_translated_point_2_object()
+                 (p, K().construct_scaled_vector_2_object()(v, i));
+    }
+
+    Point_2
+    operator()(const Line_2& l, const FT& i) const
     {
       Point_2 p = K().construct_point_2_object()(l);
       Vector_2 v = K().construct_vector_2_object()(l);
--- a/Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
+++ b/Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
@@ -3098,12 +3098,12 @@ namespace HomogeneousKernelFunctors {
     }
 
     Point_2
-    operator()(const Line_2& l, int i) const
+    operator()(const Line_2& l, const FT i) const
     {
       Point_2 p = K().construct_point_2_object()(l);
       Vector_2 v = K().construct_vector_2_object()(l);
       return K().construct_translated_point_2_object()
-                 (p, K().construct_scaled_vector_2_object()(v, RT(i)));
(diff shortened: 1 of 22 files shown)


afabri:
Why is the change for segment big?  0 and 1 have still the same meaning.

Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
```diff
@@ -3098,12 +3098,12 @@ namespace HomogeneousKernelFunctors {
     }
 
     Point_2
-    operator()(const Line_2& l, int i) const
+    operator()(const Line_2& l, const FT i) const
```
lrineau: Is not it an issue to rely on `FT` in the homogeneous kernel?

MaelRL:
> Why is the change for segment big? 0 and 1 have still the same meaning.

Because it says that it takes `i%2`, so it's not just an extrapolation of the values at `O` and `1` if `s.point(2) == s.point(0)`

Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
MaelRL: It could be... Could you please precise what you mean? In my head, `FT` was just `Quotient<template_number_type>` so here the user gives me some quotient, and I do some vector scaling? I guess I might change the `w` in the homogeneous point in output, but it doesn't seem like an issue?

lrineau: I think there should be an overload for `int` or `RT` as well, so that the code involved is not complicated by the `FT` case. It seems one can use `typename First_if_different<FT_,RT>::Type` to have SFINAE one of the two overloads when `FT` and `RT` are equal.

MaelRL: Done in a45ccab. Can't do anything for other types because impressingly `ray_23/line_23.point(i)` all manage to take different approaches to achieve the same construction. Probably could do some kernel-wide uniformizing!

MaelRL: Needed some more work because there is some code using `FT := int` (which doesn't seem very legal, `int` is not a model of `FieldNumberType`...?)

maxGimeno:

lrineau:
This PR is `Tested` but also `not yet approved`. @MaelRL, how should approve it?

MaelRL:
It was just marked as `not yet approved` to have someone do a swift code review, but it seems you already did that a while ago since you've left remarks above. So, it can be considered reviewed, if you're ok with it.

PR #4124, opened 2019-07-30
minor fix: add const specifier

aboudev:
## Add const specifier

Add const spedifier to access the function `number_of_unassigned_points`.

## Release Management


maxGimeno:
succesfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-99.shtml

PR #4128, opened 2019-08-01
PMP: Fix ctestsuite

maxGimeno:
## Summary of Changes
Lock resources between the 2 triangulate_hole tests to prevent the results from generate_mesh_with_hole to be cleaned before the end.
Also fixes the recurring bug in the ctestsuite for Polyhedorn Demo about truncated files. (Time will tell if it works.)
- [x] TODO : keep editing the ctesting of the demo to get rid of the broken files errors
## Release Management

* Affected package(s):PMP


lrineau:
It could be backported to release branches. If we want to opt for CTest-like testsuite on Windows, we will need that.

Polyline_simplification_2/demo/Polyline_simplification_2/Polyline_simplification_2.ui
```diff
@@ -1,268 +1,221 @@
...
    </rect>
   </property>
-  <property name="windowTitle" >
+  <property name="windowTitle">
    <string>CGAL Polyline Simplification</string>
   </property>
-  <property name="windowIcon" >
-   <iconset resource="../resources/CGAL.qrc" >
+  <property name="windowIcon">
+   <iconset resource="../../../../CGAL/GraphicsView/demo/resources/CGAL.qrc">
```
lrineau: How can this be a valid path?

maxGimeno: It's probably only valid on my current config, indeed. Still better than the current path that leads to a file that doesn't exist :)

lrineau: At least it you remove the inner `../CGAL/` it would work in all Git-layouts.

lrineau:
@maxGimeno Is this PR ready to be tested, now?

PR #4129, opened 2019-08-01
Fix segfault in `PMP::isotropic_remeshing()`

### LINKED ISSUE #4112: isotropic_remeshing segfaults even when is_valid_polygon_mesh is true
##  Issue Details

I ran the function `CGAL::Polygon_mesh_processing::isotropic_remeshing` for a particular mesh that I have and it causes a segfault, and I am not sure why. What's even more confusing is when I check it with `CGAL::is_valid_polygon_mesh` it appears to be true. Please see below minimal code to reproduce the problem. 

```
typedef CGAL::Surface_mesh<K::Point_3> Mesh;

Mesh m;

std::ifstream input("test.off");
input >> m;
input.close();

std::cout << CGAL::is_valid_polygon_mesh(m) << std::endl;

CGAL::Polygon_mesh_processing::isotropic_remeshing(faces(m), 5.0, m);
```

Here is the mesh that triggers the failure. [test.zip](https://github.com/CGAL/cgal/files/3432733/test.zip) Additionally, this raw mesh is reconstructed from multiple depth sensors, so when I run `isotropic_remeshing` on the previous frame where the raw mesh is nearly identical as this one (the only difference is a slight head movement), it actually doesn't segfault, which makes it even more confusing. Please let me know if you can reproduce the problem.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Ubuntu 18.04
* Compiler: GCC 7
* CGAL version: Latest master
* Boost version: 1.68

janetournois:
## Summary of Changes

The collapse step of `PMP::isotropic_remeshing()` has a post processing step that removes degenerate faces.
We had forgotten to deal with degenerate faces that are on a boundary, and this lead to a segfault.
It's fixed now.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #4112

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -1638,6 +1638,9 @@ private:
         BOOST_FOREACH(halfedge_descriptor hf,
                       halfedges_around_face(h, mesh_))
         {
+          if(face(opposite(hf, mesh_), mesh_) == boost::graph_traits<PM>::null_face())
+            continue;
+
           vertex_descriptor vc = target(hf, mesh_);
           vertex_descriptor va = target(next(hf, mesh_), mesh_);
           vertex_descriptor vb = target(next(next(hf, mesh_), mesh_), mesh_);


jasjuang:
Another question, I was testing `isotropic_remeshing` on other meshes and I ran into a case where `isotropic_remeshing` runs for an abnormally long time, currently 30 minutes and counting. So I tested the mesh with `is_valid_polygon_mesh` and it returns false. What are some functions I can call to repair the mesh so `is_valid_polygon_mesh` becomes true? So far I tried, `duplicate_non_manifold_vertices`, `remove_isolated_vertices ` and `keep_largest_connected_components` but none of them solves it. Is there a way to let CGAL report why `is_valid_polygon_mesh` is false for a particular mesh?

MaelRL:
There is a verbose boolean in `is_valid_polygon_mesh`.

There are a lot more repair functions in the `PMP` package. However, except for manifoldness issues, they mostly try to fix the geometry or the topology of the mesh, not the fact that the actual storing structure of the mesh is broken, which is a usually bug. Can you open another issue and upload the mesh and what you did with it please?

jasjuang:
@MaelRL thanks, please see #4133

maxGimeno:

Issue #4133, opened 2019-08-02
duplicate_non_manifold_vertices causes isotropic_remeshing to never return 


jasjuang:
## Issue Details

For this specific mesh [test1.zip](https://github.com/CGAL/cgal/files/3462248/test1.zip) I worked with, if I run `duplicate_non_manifold_vertices` first then `isotropic_remeshing` will be stuck somewhere internally and never return. The reason I ran `duplicate_non_manifold_vertices` is because in a previous issued I filed #3565, running `duplicate_non_manifold_vertices` first actually is required in order for `isotropic_remeshing` to work. The dilemma for me is I can't have different codes for different meshes. Ideally, I need to decide to add `duplicate_non_manifold_vertices` or not, and then `isotropic_remeshing` should work for all kinds of meshes. Please see below minimal code to reproduce the problem

```
typedef CGAL::Simple_cartesian<double> K;
typedef CGAL::Surface_mesh<K::Point_3> Mesh;

Mesh m;

std::ifstream input("test1.off");
input >> m;
input.close();

// false
std::cout << CGAL::is_valid_polygon_mesh(m) << std::endl; 

CGAL::Polygon_mesh_processing::duplicate_non_manifold_vertices(m);

// false
std::cout << CGAL::is_valid_polygon_mesh(m) << std::endl;

// never returns
CGAL::Polygon_mesh_processing::isotropic_remeshing(faces(m), 5.0, m)
```

## Environment

* Operating system: Ubuntu 18.04
* Compiler: GCC 7
* Release or debug mode: Release
* CGAL version: Latest master
* Boost version: 1.68

MaelRL:
There was indeed some issues with the manifold vertex duplication function. This mesh really was bending the rules (e.g. the same vertex being used in two stars while being technically a single flat star, etc.), so it was interesting to improve the function, thanks! Do you mind / is it possible from a license point of view to please use it in the CGAL testsuite? Otherwise, I will just extract the local combinatorics and change the geometry, but it's tedious :).

Things should be fixed with https://github.com/MaelRL/cgal/commit/814f8c67e56297167e9a6f57d90fa166fabf5a4a, but I worked in an unrelated branch and I have to backport things. Isotropic remeshing runs fine afterwards. [Here](https://gist.github.com/MaelRL/961367357931dcfd1aec203e24af43d8) is what I get, in case you want to already progress with it.

Note that you can also call `remove_isolated_vertices()` initially, you were carrying about a hundred useless points.

[...]

PR #4137, opened 2019-08-05
PMP: Fix improper halfedge incidence state in post non-manifold vertex duplication mesh (4.14)

MaelRL:
## Summary of Changes

There was an issue with keeping a valid incident halfedge (i.e. non-null) after non-manifold vertex duplication. Now, we ensure that the halfedge incident to the initial vertex that we do not duplicate is properly set.

Not done for CGAL 4.13 because the version in 4.13 was really different and it's not worth it.

## Release Management

* Affected package(s): `PMP`
* Issue(s) solved (if any): #4133
* version for 5.0: #4139


sloriot:
Successfully tested in [CGAL-5.0.1-Ic-123](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0.1-Ic-123)

PR #4152, opened 2019-08-13
Nef_S2: Fix memory corruption

maxGimeno:
## Summary of Changes
Fix a memory corruption due to const refs on temporary objects.
## Release Management

* Affected package(s):Nef_S2


maxGimeno:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-113.shtml

PR #4160, opened 2019-08-19
Do not collapse an edge is the placement cannot be computed

sloriot:
Simple fix making it obvious that the placement must be valid.

My guess is that it was the job of the cost to return `boost::none` is the placement was invalid (but nothing was documented).


maxGimeno:
Successfully tested i nhttps://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-118.shtml

PR #4172, opened 2019-08-21
PMP: Fix distance range type

sgiraudot:
## Summary of Changes

The function `approximate_Hausdorff_distance_impl()` has a `PointRange` as parameter but then calls a functor with `std::vector<Point_3>` as parameter, which makes the function unable to handle anything else than a vector of Point_3 (which makes the template useless).

This PR makes the functor us a `PointRange` too.

The range used by `max_distance_to_triangle_mesh()` is now documented as having random access iterators.

## Release Management

* Affected package(s): PMP


maxGimeno:
Successfully tested i nhttps://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-118.shtml

PR #4174, opened 2019-08-21
PMP: Fix boundary cycle when non manifold vertices exist in the cycle (5.0)

MaelRL:
## Summary of Changes

Boundary cycle stitching was done by "zipping" the cycle up from all potential starting points (a pair of consecutive & geometrically equal border halfedges). When there are non manifold vertices, this doesn't work because zipping from one side can zip part of another starting pair.

The function is completely rewritten to zip compatible ranges one-by-one instead (theoretically worse complexity, but unnoticeable in practice).

## Release Management

* Issue(s) solved (if any): -


sloriot:

PR #4176, opened 2019-08-21
 PMP: Fix boundary cycle when non manifold vertices exist in the cycle (4.14)

MaelRL:
## Summary of Changes

See #4174.

## Release Management

* Issue(s) solved (if any): -

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h
@@ -34,6 +34,7 @@
 #include <CGAL/Polygon_mesh_processing/internal/named_params_helper.h>
 #include <CGAL/Polygon_mesh_processing/border.h>
 #include <CGAL/Polygon_mesh_processing/connected_components.h>
+#include <CGAL/Polygon_mesh_processing/shape_predicates.h>
 
 #include <CGAL/array.h>
 #include <CGAL/Dynamic_property_map.h>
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h
```diff
@@ -572,57 +559,152 @@ std::size_t stitch_boundary_cycles(PolygonMesh& pm,
...
-  cpp11::unordered_set<halfedge_descriptor> stitched_hedges;
+  //
+  // As long as we find vertices on the boundary with both halfedges being geometrically equal,
+  // we zip it up as much as we can.
 
-  BOOST_FOREACH(halfedge_descriptor h, boundary_cycles)
+  // not everything is always stitchable
+  std::set<halfedge_descriptor> unstitchable_halfedges;
+
+  halfedge_descriptor null_h = boost::graph_traits<PolygonMesh>::null_halfedge();
```
sloriot:
```suggestion
  const halfedge_descriptor null_h = boost::graph_traits<PolygonMesh>::null_halfedge();
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h
```diff
@@ -572,57 +559,152 @@ std::size_t stitch_boundary_cycles(PolygonMesh& pm,
...
+
+    if(!hedges_to_stitch.empty())
+    {
+      std::size_t local_stitches = internal::stitch_borders_impl(pm, hedges_to_stitch);
+      stitched_boundary_cycles_n += local_stitches;
+
+      if(local_stitches == 0) // refused to stitch this halfedge pair range due to manifold issue
+      {
+        BOOST_FOREACH(const Halfedge_pair& hp, hedges_to_stitch)
+          unstitchable_halfedges.insert(hp.first);
```
sloriot: why only first?

MaelRL: In the loop that tries to find a starting stitchable pair (above), the geometries of `hn` and `hnn` are compared and I need only to check that at least one of them is not marked as `unstitchable` (both is just redundant information). I chose the first one arbitrarily.

Actually, this probably could simply be `unstitchable_halfedges.insert(hedges_to_stitch.front().first);` because the other halfedges pairs are not formed by consecutive halfedges so these pairs will never be considered in the starting loop.

sloriot:
Successfully tested in [CGAL-5.0.1-Ic-123](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.0.1-Ic-123)

PR #4178, opened 2019-08-22
`PMP::isotropic_remeshing()` - prevent creating non-manifold configurations

### LINKED ISSUE #4148: isotropic_remeshing never returns even when is_valid_polygon_mesh is true
## Issue Details

For this specific mesh [test2.zip](https://github.com/CGAL/cgal/files/3490335/test2.zip) if I run the below code `isotropic_remeshing` will never return. Note that I already applied PR #4139 and #4129 and the problem is still there. Please let me know if you can replicate the problem.

## Source Code

```
typedef CGAL::Surface_mesh<K::Point_3> Mesh;

Mesh m;

std::ifstream input("test2.off");
input >> m;
input.close();

std::cout << CGAL::is_valid_polygon_mesh(m) << std::endl;

// never returns
CGAL::Polygon_mesh_processing::isotropic_remeshing(faces(m), 5.0, m)
```

## Environment

* Operating system: Ubuntu 18.04
* Compiler: GCC 7
* CGAL version: PR #4139 and manually applied #4129
* Boost version: 1.68

janetournois:
## Summary of Changes

Internal code of `PMP::isotropic_remeshing()` uses `CGAL::Euler::flip_edge()` to fix degenerate faces that can be created during the short edges collapse step.

This PR fixes this use of `flip_edge()` by adding a topological condition that prevents from creating non-manifold configurations (and a segfault in the following steps).
The condition was there but only as an assertion, and has been turned into a strong condition.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #4148

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -787,19 +787,20 @@ namespace internal {
           //fix constrained case
           CGAL_assertion((is_constrained(vkept) || is_corner(vkept) || is_on_patch_border(vkept)) ==
                          (is_va_constrained || is_vb_constrained || is_va_on_constrained_polyline || is_vb_on_constrained_polyline));
-          fix_degenerate_faces(vkept, short_edges, sq_low, collapse_constraints);
-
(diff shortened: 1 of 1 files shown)


janetournois:
@sloriot I locally tested this PR on the 4.14 branch but it does not work. There must be some other changes in `master` that help it work. So let's target 5.0.

sloriot:

PR #4184, opened 2019-08-24
Problem with single quote in code span.

albert-github:
In case we have a single quote in a code span it is better to have the start and end backtics of the code span as double backticks.
Sometimes the constructs lead to warnings or to garbled code. This happens in the newer versions of doxygen where some new heuristics are used to recognize texts line "it's".
See also "Warning in case of usage of a single quote in a code span." (https://github.com/doxygen/doxygen/pull/7209).

The wrong result (see 2nd section the text: However....), as obtained from https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.0-Ic-119/master/Advancing_front_surface_reconstruction/index.html#Chapter_Advancing_Front_Surface_Reconstruction

wrong

the correct / updated result:
correct


sloriot:

PR #4187, opened 2019-08-26
Spatial Searching: Search Traits Adapter without lvalue point map

sgiraudot:
## Summary of Changes

`Search_traits_adapter` requires an lvalue point map because, internally, the `cartesian_const_iterator` is used to iterate on the coordinates of points. In practice, this is a very annoying limitation, as it means one cannot construct a kd tree with points generated on-the-fly (for example).

This PR adds an alternative to the `cartesian_const_iterator` defined by the search traits adapter in the case where the point map is not lvalue (by storing the point in a shared pointer to avoid iterating on a temporary object). If the point map is lvalue, the original behavior is kept.

Now the point map can be only readable and still work with `Search_traits_adapter`.

## Release Management


sloriot:

PR #4191, opened 2019-08-27
Formulas shown as code

albert-github:
Some formulas are shown as code due to the fact that the indentation of the start of the formula is 4 or more compared to the paragraph it is part of (handling of code sections has been improved from e.g. 1.8.13 to master).
the formula construct `\f$ .. \f$` should not be used when using e.g. the `equation` environment. With MathJax this works but in standard LaTeX an error is thrown (not yet the focus of CGAL, but might be of interest in the future / newer implementations of MathJax when more compatible).

See: Polygonal_surface_reconstruction/index.html#title2


sloriot:

PR #4192, opened 2019-08-27
WIP - PMP - fix `isotropic_remeshing()` self intersections

### LINKED ISSUE #4144: Isotropic remeshing creates self intersections
## Issue Details

* Load the mesh in the Polyhedron demo
* Select 60 degree edges
* Isotropic remesh :  0.0015,  3 iterations, 3 smoothings, protect, allow

[cheese_transformed-facets.zip](https://github.com/CGAL/cgal/files/3485856/cheese_transformed-facets.zip)

The resulting mesh has self intersections as well as triangles with very small angles

## Environment

* CGAL version: master

janetournois:
## Summary of Changes

During the edge collapse step, almost degenerate faces tend to create self-intersections in the output mesh.
With this PR, the function `fix_degenerate_faces()` now also deals with almost degenerate
faces, namely the faces with their area lower than 1% * average area of neighbors
(we only compute squared areas though it is not the exact same criterion).

About `fix_degenerate_faces()`:
There also are some configurations that this function is not able to fix :
sometimes, flipping an edge to fix a degenerate face creates a new
degenerate face. And flipping an edge to fix it comes back to the
initial configuration.
After 10 attempts around `v`, we give up and just return `true`
because the mesh has been modified and the calling code needs to know about it.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #4144


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -1598,33 +1626,70 @@ namespace internal {
...
+        else
+        {
+          double sqa = squared_area(h);
+          sum_sq_areas += sqa;
+          sq_areas[h] = sqa;
+        }
       }
 
+      //now collect almost degenerate faces
+      const double sq_max_ratio = 1e-4;
```
janetournois: @MaelRL @sloriot what do you think about this 1% "magic ratio"?

sloriot:
If you are trying to handle self-intersections created during the local operations, I suggest that you first try to look at what is done in `Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h` in the function `remove_almost_degenerate_faces()`. In their there are two criteria that are used to prevent bad collapsed and bad flip. Maybe this would solve the issues you observed.

maxGimeno:
@janetournois what is the status of this PR ? Should it be postponed ?

janetournois:
Yes, you can postpone it
Thank you!

lrineau:
Replaced by https://github.com/CGAL/cgal/pull/5266.

PR #4193, opened 2019-08-28
Kernel_23: Fix documentation of functor `ConstructLiftedPoint_3`

MaelRL:
## Summary of Changes

Documentation of the functor seems to indicate the projection of the lifted point along `Oz` would be the 2D point, but this is not what this functor is about. Rather, it maps the `xOy` plane and the user-provided plane via an affine transformation (in the CGAL implementation, it maps `Ox` and `Oy` to the vectors given by `plane.base1()` and `plane.base2()`.

## Release Management

* Issue(s) solved (if any): -


sloriot:

PR #4194, opened 2019-08-28
Demo: Improve affine transform on point sets

sgiraudot:
## Summary of Changes

The affine transform plugin's behavior when creating a new point set item from the transformation was quite bad: all properties were lost in the process (normals, colors, etc.), and by looking at the code, I realized that the point set was copied twice for no reason (the first copy being also never properly deleted, leading to a memory leak).

This PR removes the useless copies + memory leak and does a full copy of the point set to avoid loosing properties.

## Release Management

* Affected package(s): demo


sloriot:

PR #4198, opened 2019-08-29
Make 2d segment segment distance more robust for kernel with inexact constructions

### LINKED ISSUE #4189: Wrong result of squared_distance of two Segment_2 objects
Wrong result of squared_distance of two Segment_2 objects.

## Issue Details

Using `CGAL::squared_distance` to compute the closest distance between two **almost collinear** segments may produce wrong results in some cases. e.g.
```
# Wavefront
v -4.0380854964382 -1.9947196614192 0.0
v 10.43442091460618 -0.5886833953492263 0.0
v -11.5138934277993 -2.721011070186227 0.0
v -8.822747585009402 -2.459560251317805 0.0
l 1 2
l 3 4
```
The computed result is 0, which is not right if we visualize the .obj file.

snapshot00

## Source Code

```cpp
#include <iostream>

#include <CGAL/squared_distance_2.h>

typedef CGAL::Simple_cartesian<double> Kernel;
typedef Kernel::Segment_2 Segment_2;

int main() {
  const Segment_2 segi = {
    { -4.0380854964382, -1.9947196614192 },
    { 10.43442091460618, -0.5886833953492263 } };
  const Segment_2 segj = {
    { -11.5138934277993, -2.721011070186227 },
    { -8.822747585009402, -2.459560251317805 } };

  const auto dist = CGAL::squared_distance(segi, segj);
  std::cout << "#dist: " << dist << std::endl;

  return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows and Linux, 64 bits
* Compiler: VC 2015, gcc 5.5.0
* CGAL version: 4.14
* Boost version: 1.65

afabri:
## Summary of Changes

Avoiding subtractions fixes the issues when using `Simple_cartesian<double>`, and to use `orientation()` should make it robuster with `Epick`. 

The same should be done for Ray_2/Segment_2.   
We could introduce a predicate for 4 points that computes all orientation tests and determines at the same time which is the closer endpoint. This would make it yet robuster and it would avoid redundant computations.

I accidentally committed also a slightly faster implementation of  compare_signed_distance_to_line()`.

TODO:
- [x] Small feature

## Release Management

* Affected package(s): Distance_2
* Issue(s) solved (if any): fix #4189
* Small Feature: [CompareSignedDistanceToLine_2](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Kernel::CompareSignedDistanceToLine_2) I pre-approuve. -- Laurent Rineau 10:34, 12 March 2020 (CET)

image

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -533,7 +533,7 @@ namespace CartesianKernelFunctors {
     typedef typename K::Line_2    Line_2;
     typedef typename K::Equal_2   Equal_2;
     typedef typename K::Less_signed_distance_to_line_2 Less_signed_distance_to_line_2;
-    
+
   public:
     typedef typename K::Comparison_result     result_type;
 
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -527,6 +527,39 @@ namespace CartesianKernelFunctors {
   };
 
   template <typename K>
+  class Compare_signed_distance_to_line_2
+  {
+    typedef typename K::Point_2   Point_2;
+    typedef typename K::Line_2    Line_2;
+    typedef typename K::Equal_2   Equal_2;
+    typedef typename K::Less_signed_distance_to_line_2 Less_signed_distance_to_line_2;
+    
+  public:
+    typedef typename K::Comparison_result     result_type;
+
+    result_type
+    operator()(const Point_2& a, const Point_2& b,
+               const Point_2& c, const Point_2& d) const
+    {
+      CGAL_kernel_precondition_code(Equal_2 equal;)
+          CGAL_kernel_precondition(! equal(a,b));
+      return cmp_signed_dist_to_lineC2( a.x(), a.y(),
+                                        b.x(), b.y(),
+                                        c.x(), c.y(),
+                                        d.x(), d.y());
+    }
+
+    result_type
+    operator()(const Line_2& l, const Point_2& p, const Point_2& q) const
+    {
+      Less_signed_distance_to_line_2 less = K().less_signed_distance_to_line_2_object();
+      if (less(l, p, q)) return SMALLER;
+      if (less(l, q, p)) return LARGER;
+      return EQUAL;
+    }
+  };
+
+  template <typename K>
   class Compare_squared_radius_3
   {
     typedef typename K::Point_3            Point_3;
@@ -3884,7 +3917,7 @@ namespace CartesianKernelFunctors {
     }
   };
 
-  // TODO ...
+
   template <typename K>
   class Less_signed_distance_to_line_2
   {
--- a/Distance_2/include/CGAL/squared_distance_2_1.h
+++ b/Distance_2/include/CGAL/squared_distance_2_1.h
@@ -199,7 +199,7 @@ namespace internal {
       return internal::squared_distance(seg1.source(), seg2, k);
     if (seg2.source() == seg2.target())
       return internal::squared_distance(seg2.source(), seg1, k);
-    
+
     Orientation o1s = orientation(seg2.source(), seg2.target(), seg1.source());
     Orientation o1e = orientation(seg2.source(), seg2.target(), seg1.target());
     if (o1s == RIGHT_TURN) {
@@ -231,7 +231,7 @@ namespace internal {
     if (crossing1) {
       if (crossing2)
         return (FT)0;
-      
+
       c2s = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.source(), k));
       c2e = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.target(), k));
       Comparison_result dm = compare(c2s,c2e);
@@ -269,11 +269,11 @@ namespace internal {
         min1 = (dm == SMALLER) ?
                  internal::squared_distance(seg1.source(), seg2, k):
                  internal::squared_distance(seg1.target(), seg2, k);
-        
+
         c2s = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.source(), k));
         c2e = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.target(), k));
         dm = compare(c2s,c2e);
-        
+
         if (dm == EQUAL)  // should not happen.
           return internal::squared_distance_parallel(seg1, seg2, k);
         min2 = (dm == SMALLER) ?
@@ -378,7 +378,7 @@ namespace internal {
     }
   }
 
-  
+
   template <class K>
   inline typename K::RT
   _distance_measure_sub(const typename K::RT &startwcross,
--- a/Distance_2/include/CGAL/squared_distance_2_1.h
+++ b/Distance_2/include/CGAL/squared_distance_2_1.h
@@ -188,7 +188,108 @@ namespace internal {
   typename K::FT
   squared_distance(const typename K::Segment_2 &seg1,
                    const typename K::Segment_2 &seg2,
-                   const K& k)
+                   const K& k,
+                   const Cartesian_tag&)
+  {
+    typedef typename K::RT RT;
+    typedef typename K::FT FT;
+    bool crossing1, crossing2;
+    RT c1s, c1e, c2s, c2e;
+    if (seg1.source() == seg1.target())
+      return internal::squared_distance(seg1.source(), seg2, k);
+    if (seg2.source() == seg2.target())
+      return internal::squared_distance(seg2.source(), seg1, k);
+    
+    Orientation o1s = orientation(seg2.source(), seg2.target(), seg1.source());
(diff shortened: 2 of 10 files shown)


mglisse:
What is the impact on Epeck?

afabri:
If I measure the time of 1mio distance computations for pairs of segments that go in some different branches the performance drops from 10.5 to 11 seconds.   
Ideally, we had a filtered predicate that computes the orientation of two points with respect to a segment and at the same time which of the two is  closer.  That would remove the re-computation of sub-determinants computed in the orientation test, and it would make it robust for further corner cases.  
As an alternative  we could also specialize for `Epick` and `Simple_cartesian<double>`.

mglisse:
If the impact on Epeck (which is not the goal of this patch) is only about 5%, that's not so bad, I think further improvements can wait and don't need to block this.

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
```diff
@@ -535,6 +535,40 @@ namespace CartesianKernelFunctors {
...
+    typedef Comparison_result   result_type;
+
+    result_type
+    operator()(const Point_2& a, const Point_2& b,
+               const Point_2& c, const Point_2& d) const
+    {
+      CGAL_kernel_precondition_code(Equal_2 equal;)
+      CGAL_kernel_precondition(! equal(a,b));
+      return cmp_signed_dist_to_lineC2( a.x(), a.y(),
+					b.x(), b.y(),
```
MaelRL: The indentation of the whole patch is pretty bad

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
```diff
@@ -535,6 +535,40 @@ namespace CartesianKernelFunctors {
...
+					c.x(), c.y(),
+					d.x(), d.y());
+    }
+    /*
```
MaelRL: Why is this commented? If it's useless or broken, it should probably be removed.

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
```diff
@@ -3893,7 +3927,7 @@ namespace CartesianKernelFunctors {
     }
   };
 
-  // TODO ...
+  // TODO:  Add functor Compare_signed_distance_to_line_2 and use this one in compare_signed_distance_to_line
```
MaelRL: Obsolete

Distance_2/include/CGAL/squared_distance_2_1.h
```diff
@@ -197,7 +197,109 @@ namespace internal {
...
+	crossing2 = true;
+      } else {
+	crossing2 = (o2s == COLLINEAR);
+      }
+    }
+
+    c1s = CGAL::abs(wcross(seg2.source(), seg2.target(), seg1.source(), k));
+    c1e = CGAL::abs(wcross(seg2.source(), seg2.target(), seg1.target(), k));
+    c2s = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.source(), k));
+    c2e = CGAL::abs(wcross(seg1.source(), seg1.target(), seg2.target(), k));
```
MaelRL: There is no need to compute all four values. For example, the first case only uses `c2s` and `c2e`.

Distance_2/include/CGAL/squared_distance_2_1.h
```diff
@@ -197,7 +197,109 @@ namespace internal {
   typename K::FT
   squared_distance(const typename K::Segment_2 &seg1,
 		   const typename K::Segment_2 &seg2,
-		   const K& k)
+		   const K& k,
+                   const Cartesian_tag&)
```
MaelRL: Is there still a need to disjoint Cartesian and Homogeneous kernels?

Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
```diff
@@ -770,6 +770,64 @@ namespace HomogeneousKernelFunctors {
     }
   };
 
+    template <typename K>
+  class Compare_signed_distance_to_line_2
```
MaelRL: Needs to be tested and documented in the Kernel concept... or just removed since it's not used by the PR.

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
afabri: removed.

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
afabri: removed

Distance_2/include/CGAL/squared_distance_2_1.h
afabri: changed.

Distance_2/include/CGAL/squared_distance_2_1.h
afabri: I don't want to work on Homogeneous.

Homogeneous_kernel/include/CGAL/Homogeneous/function_objects.h
afabri: `Less_signed_distance_to_line_2`  now uses `Compare_..`.

Distance_2/include/CGAL/squared_distance_2_1.h
MaelRL: My point is that the code for the Cartesian kernel doesn't seem to be specific to a Cartesian kernel anymore (?)

afabri:
@MaelRL what remains to be done?

MaelRL:
https://github.com/CGAL/cgal/pull/4198#discussion_r319916927 (is there still a need to distinguish the Cartesian case?) and https://github.com/CGAL/cgal/pull/4198#discussion_r319917347 (must document Compare_signed_distance_to_line_2 in the Kernel concept/models) are still relevant.

MaelRL:
For tracking, the current code does not fix the issue as the configuration reported in the issue still yields a distance of 0.

afabri:
> 
> 
> For tracking, the current code does not fix the issue as the configuration reported in the issue still yields a distance of 0.

With VC++ it makes a difference. I added the issue.cpp to the tests, so I would be glad to see it in the testsuite

Distance_2/include/CGAL/squared_distance_2_1.h
lrineau: I agree with Mael, here: it looks like the the code with `Cartesian_tag` could be applied to the homogeneous case as well. I also think we could merge the two overloads of `squared_distance(Segment_2, Segment_2, K, Tag)` into one.

afabri:
I had used the supposed generic code for the homogeneous kernel, but the testsuite "proves" that it does not work.    I prefer switching back to two different codes that work, instead of understanding what goes wrong for the homogeneous kernel, as I consider the homogeneous kernel as legacy.

maxGimeno:
runtime errors in a Distance_2 test : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-73.shtml#Distance_2

not entirely sure this PR is to blame as those are quite recent though...

Distance_2/test/Distance_2/issue_4189.cpp
```diff
@@ -0,0 +1,34 @@
...
+  return 0;
+}
+
+int main()
+{
+	fct<SC>();
+	fct<CGAL::Epick>();
+	fct<EC>();
+	return 0;
+}
```
lrineau: This file has several issues:
  1. It hardcodes `CGAL::Gmpq` unconditionally, and fails to compile on the few platforms where we have disabled GMP.
  2. It does not test anything: there are no `assert` to check that the results are the expected ones. It looks more like a debugging tool.

afabri: I changed it to `Exact_rational`.     The goal is not to assert whatsoever, but to see what the result is on different platforms.  It seems that on some platforms we will still see a really wrong distance.

lrineau: Then try to measure what it a "really wrong distance", and write the corresponding `assert`. Otherwise that test will be useless: we will just check that it compiles, but if, in a few years, suddenly the distance is wrong again, we will not notice.

maxGimeno:
Travis is unhappy : https://travis-ci.org/CGAL/cgal/jobs/655368799#L2834

afabri:
I find it rather annoying that `<CGAL/Filtered_kernel_selector.h>` includes the homogenous kernel.

Distance_2/test/Distance_2/issue_4189.cpp
MaelRL: Is this resolved?

afabri: I just want to see  the difference but neither yellow nor red in the testsuite.

maxGimeno:

sloriot:
Tested in CGAL-5.1-Ic-114

@afabri `CHANGE.md` not updated

PR #4203, opened 2019-09-02
Polyhedron Demo: Fix  stitching plugin 

maxGimeno:
## Summary of Changes

For now, loading an OFF as a Surface_mesh does not check for the validity of the Polygon Mesh. It means that it is possible to load a data set that is not a 2-manifold in a surface_mesh, which can cause problems when iterating the structure. 
This PR fixes it by adding a check in the load function. If it is not valid, then the set is loaded as a soup, which forces the user to orient it, which will duplicate the non-manifold vertices. 

EDIT: This PR now only adds stats for nm vertices to the scene_surface_mesh_item and modify the stitching plugin to not use a filter graph in the "Detect Boundaries" operation, so that it doesn't use the `is_border(vertex)` function that doesn't work on nm vertices.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
@@ -2,6 +2,7 @@
 #include "Scene_polygon_soup_item.h"
 #include "Scene_points_with_normal_item.h"
 #include <CGAL/Three/Three.h>
+#include <CGAL/Polygon_mesh_processing/repair.h>
 
 
 #include <CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h>
@@ -25,17 +26,17 @@ class Polyhedron_demo_off_plugin :
   Q_PLUGIN_METADATA(IID "com.geometryfactory.PolyhedronDemo.IOPluginInterface/1.90" FILE "off_io_plugin.json")
 
 public:
-  bool isDefaultLoader(const Scene_item *item) const 
-  { 
+  bool isDefaultLoader(const Scene_item *item) const
+  {
     if(qobject_cast<const Scene_surface_mesh_item*>(item)
-       || qobject_cast<const Scene_polygon_soup_item*>(item)) 
-      return true; 
+       || qobject_cast<const Scene_polygon_soup_item*>(item))
+      return true;
     return false;
   }
-  bool isDefaultLoader(const QString& name) const 
-  { 
-    if(name == QString("off")) 
-      return true; 
+  bool isDefaultLoader(const QString& name) const
+  {
+    if(name == QString("off"))
+      return true;
     return false;
   }
   QString name() const { return "off_plugin"; }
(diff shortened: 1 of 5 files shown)


sloriot:
I don't like the patch. I suggest that the mesh can be loaded, indicate in the statistics the presence of non-manifold vertices, and add (if it does not exist) a plugin to call duplicate non-manifold vertices. (remember that the demo is also used as a diagnostic tool). 

About the display of boundary edges, I don't understand why a `Filtered_face_graph` is required to display boundary edges.

sloriot:
And there is no iteration issue, only possible circulation around a non-manifold vertex

maxGimeno:
It is not required, it is juste used to simplify the creation of the polylines, using `split_graph_into_polylines()`

maxGimeno:
Remains the issue about [is_border(Vertex_index)](https://doc.cgal.org/latest/Surface_mesh/classCGAL_1_1Surface__mesh.html#a587a5b95dab4077e28b4b6db8d947f1b), which returns a wrong result if the surface_mesh is not a 2-manifold.

sloriot:
you cannot circulate around a non-manifold vertex. `is_border` for a vertex must document that if the vertex is non-manifold the result might not be correct.

Polyhedron/demo/Polyhedron/Plugins/IO/OFF_io_plugin.cpp
```diff
@@ -164,12 +165,15 @@ Polyhedron_demo_off_plugin::load_off(QFileInfo fileinfo) {
...
   item->setName(fileinfo.completeBaseName());
   std::size_t isolated_v = 0;
+  std::size_t nm_v = 0;
   for(vertex_descriptor v : vertices(*surface_mesh))
   {
     if(surface_mesh->is_isolated(v))
     {
       ++isolated_v;
     }
+    if(CGAL::Polygon_mesh_processing::is_non_manifold_vertex(v, *surface_mesh))
```
sloriot: use `non_manifold_vertices()` that takes an output iterator as paramater.

maxGimeno: No. `non_manifold_vertices()` stores one halfedge per umbrella incident to the vertex, so you have no way to know the actual number of nm_vertices. You may have 4 umbrellas incident to vertex A, 2 to vertex B and 2 to vertex C, or 6 to vertex A and 2 to vertex B, per example.

sloriot: so what? even a post processing using a `std::set` will be faster than something that is O(n^2).

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-127.shtml.

@sloriot, @MaelRL , do you agree with those changes ?

lrineau:
[**@maxGimeno**](https://github.com/maxGimeno) commented on [Sep 3, 2019, 1:10 PM GMT+2](https://github.com/CGAL/cgal/pull/4203#issuecomment-527413319 "2019-09-03T11:10:41Z - Replied by Github Reply Comments"):
> Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-127.shtml.
> 
> @sloriot, @MaelRL , do you agree with those changes ?

Ping @sloriot, @MaelRL

sloriot:
From the description in the EDIT, yes. I did not look at the code in detail.

PR #4204, opened 2019-09-02
Polyhedron Demo: Check the Polygon Mesh validity after loading an OFF (4.13)

maxGimeno:
## Summary of Changes
4.13 version of #4203


maxGimeno:
Rejected patch.

PR #4205, opened 2019-09-02
Surface_mesh: Make usage more explicit in doc

maxGimeno:
## Summary of Changes

We regularly have issues on SO or Github from users that complain about add_face(0 not working correctly, when they actually face orientation or manifoldness issues. I suppose it comes from the fact that the examples that show how to create a surface_mesh do not specify that add_face() may return null_face. This PR exposes this behaviour by adding `if`s around `add_face()` calls in the examples.
## Release Management

* [Compiled documentation](https://cgal.geometryfactory.com/~mgimeno/Surface_mesh/Surface_mesh/index.html#sectionSurfaceMeshUsage)


afabri:
I am not sure that it makes sense for code where we know that it is correct.

maxGimeno:
It doesn't, but an example is only here to make it clear how to use a function. Would you prefer commentaries instead of `if`s ?

afabri:
I think an example  check-non-manifoldness.cpp  might be a good idea.  In the text that comes with the example you can then refer to the   soup2surface function.

maxGimeno:
@afabri can you please check the [Compiled documentation](https://cgal.geometryfactory.com/~mgimeno/Surface_mesh/Surface_mesh/index.html#sectionSurfaceMeshUsage) and make comments ? Is it what you had in mind ?

afabri:
Looks good to me.   You might rename the exampe to  `check_orientation.cpp` which is more positive.

Surface_mesh/doc/Surface_mesh/Surface_mesh.txt
```diff
@@ -81,7 +81,7 @@ The following example shows how to create a very simple `Surface_mesh`
 by adding 2 faces, and how to check that a face is correctly added
 to the mesh.
 
-\cgalExample{Surface_mesh/sm_check_wrong_orientation.cpp}
+\cgalExample{Surface_mesh/check_orientation.cpp}
```
lrineau: You must also rename `examples/Surface_mesh/sm_check_wrong_orientation.cpp` itself, in Git.

maxGimeno: ? It is done, I made a git mv

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-129.shtml

PR #4207, opened 2019-09-03
Polyhedron Demo: Fix Split Polyhedra

maxGimeno:
## Summary of Changes

There was a conflict between maps in the split_polyhedra operations, causing a crash when re-computing the original item. 
This removes the patch_id property_map of the item after creating it to avoid tricking the item into thinking it has an associated color_map when it really has not.


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-129.shtml

PR #4209, opened 2019-09-04
PMP: Default Property_maps for Index_maps

### LINKED ISSUE #1918: Better handling of the initialization of index map
When using polyhedron with ids, the default index property map do not have their ids initialized. This is an issue in algorithms than rely on it being initialized.

Recently, in `copy_face_graph()` I've introduced a mechanism that always initialize the id map (if writable) which is a bad thing if the user does not expect the ids of its const polyhedron to be modified.

We need to find a workaround for this.

### LINKED ISSUE #2814: Default property maps for named parameters
Now that we have dynamic property maps, I wonder whether we should use them for functions that require properties.
If we take the example of `copy_face_graph()` or `connected_components()`, they require some index property maps. Everything goes well if the graph type has internal maps but this will lead to a compilation issue with `Polyhedron_3` if no named parameter is passed.

What we could do is to check if there is an internal property map (using `CGAL::graph_has_property`) and if not use a dynamic property instead.  If we want to prevent this from being automatic (so that the user is aware that an unordered map will be used) we could make the named parameter accept `Default` to trigger the creation of a dynamic property map.

Implementation wise, all that mechanism could be hidden in a helper (responsible for choosing either the internal property map or the dynamic one).

Opinions?

CC @CGAL/geometryfactory 

Remember that there is also #1918 that is somehow related (because dynamic pmap must be initialized).

maxGimeno:
## Summary of Changes

Introduces a helper class for index maps in PMP, that uses dynamic_property_maps in case there is no internal map for indices in the mesh. Specifies if the map needs initialization or not.

TODO:
- [x] Add get_initialized(mesh&)
- [x] Tests (is_writable, all get_initialized_...)
- [x] check #1918 
- [x] Rewrite doc
- [x] Fix all usages
- [x] Add perf warning when the dpmap is selected
- [x] Solve remaining todos
- [x] Enhance doc in NP.txt

TODO after merge:
- [ ] https://github.com/CGAL/cgal/pull/4209#discussion_r392344896

## Release Management

* Issue(s) solved (if any): Fix #1918, Fix #2814
* Feature/Small Feature (if any): [Initialized_index_property_maps](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Initialized_index_property_maps)
* **Pre-approval date:** 2020/03/13


maxGimeno:
@sloriot I need your approbation about that mechanism before I integrate it in all incvolved functions.

Polygon_mesh_processing/test/Polygon_mesh_processing/connected_component_polyhedron.cpp
```diff
@@ -71,17 +71,17 @@ void mesh_no_id(const char* argv1, const bool save_output)
   PMP::connected_component(fd, sm, std::back_inserter(cc));
 
   std::cerr << cc.size() << " faces in the CC of " << &*fd << std::endl;
-  boost::property_map<Mesh,boost::vertex_external_index_t>::type vim 
+  boost::property_map<Mesh,boost::vertex_external_index_t>::type vim
     = get(boost::vertex_external_index,sm);
 
-  boost::property_map<Mesh,boost::face_external_index_t>::type fim 
+  boost::property_map<Mesh,boost::face_external_index_t>::type fim
```
sloriot: this will trigger a unused variable warning

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -232,9 +232,18 @@ connected_components(const PolygonMesh& pmesh,
...
-                         get_const_property_map(boost::face_index, pmesh));
+  typedef CGAL::Polygon_mesh_processing::GetMapFromNP<boost::face_index_t,
+      CGAL::dynamic_face_property_t<int>,
+      PolygonMesh, NamedParameters, internal_np::face_index_t> MapGetter;
+
+  MapGetter get_map(boost::face_index_t(),
+                    CGAL::dynamic_face_property_t<int>(),
+                    pmesh, np, internal_np::face_index);
+
+  bool need_init = false;
```
sloriot: the init could be done directly in the `get` I think

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,8 +113,82 @@ namespace CGAL {
...
+    need_init = false;
+    return get(t,m);
+  }
+
+  // create a dynamic property and initialize it
+  template <class Dynamic_tag, class Mesh>
+  typename boost::property_map<Mesh, Dynamic_tag >::const_type
+  get_map(CGAL::internal_np::Param_not_found, Dynamic_tag t, Dynamic_tag , const Mesh& m, bool& need_init)
+  {
+    need_init = true;
```
sloriot: I think another case when init is needed is when the index map is writable (like Polyhedron with items with ids).

Polygon_mesh_processing/test/Polygon_mesh_processing/connected_component_polyhedron.cpp
maxGimeno: How so ? It did not before, I only changed a whitespace.

sloriot: the map is no longer used AFAIS

Polygon_mesh_processing/test/Polygon_mesh_processing/connected_component_polyhedron.cpp
```diff
@@ -71,17 +71,17 @@ void mesh_no_id(const char* argv1, const bool save_output)
...
     = get(boost::face_external_index,sm);
 
   boost::vector_property_map<int,
     boost::property_map<Mesh, boost::face_external_index_t>::type>
       fccmap(fim);
```
maxGimeno: @sloriot right here

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
sloriot: This is not resolved

sloriot: You did not get the point. The whole block should go away and the map should be initialized without the developer having to write anything (init directly in the get).

maxGimeno: Ah. More overloads and specializations then.

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,8 +113,156 @@ namespace CGAL {
...
+  namespace internal_np
+  {
+  template<typename key_value, class PMap, class Graph, typename Tag>
+  struct MapInitializer{
+    void operator()(PMap, const Graph& )
+    {}
+  };
+
+  template<typename key_value, class PMap, class Graph>
+  struct MapInitializer<key_value, PMap, Graph, CGAL::Tag_false>{
```
sloriot: this one is not needed (the first specialization works)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -232,10 +232,15 @@ connected_components(const PolygonMesh& pmesh,
...
-                         get_const_property_map(boost::face_index, pmesh));
+  typedef CGAL::Polygon_mesh_processing::Get_index_map_from_NP<boost::face_index_t,
+      CGAL::dynamic_face_property_t<int>,
+      PolygonMesh, NamedParameters, internal_np::face_index_t> MapGetter;
 
+  MapGetter get_map(boost::face_index_t(),
+                    CGAL::dynamic_face_property_t<int>(),
+                    pmesh, np, internal_np::face_index);
+
+  typename MapGetter::PropertyMapType fimap = get_map.property_map();
```
sloriot: you can hide the need of the struct in a function doing the cooking.
If you want to be convenient for the developers (and for the guy that is going to replace all the usage of index maps), you should have something like `typename Default_face_index_map<NP, TM>::const_type fimap = get_initialized_face_index_map(np, pmesh)`. I'm pretty sure a macro can be used to define the 4 possible functions/classes.

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,8 +113,156 @@ namespace CGAL {
     typedef typename boost::property_traits<PMap>::value_type type;
   };
 
+  namespace internal_np
+  {
+  template<typename key_value, class PMap, class Graph, typename Tag>
+  struct MapInitializer{
```
sloriot: you must respect CGAL naming conventions, something like `Index_map_initializer`

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,8 +113,156 @@ namespace CGAL {
...
+    typedef CGAL::Tag_false result;
+  };
+
+  template<>
+  struct Is_pmap_writable<boost::read_write_property_map_tag>{
+    typedef CGAL::Tag_true result;
+  };
+
+  template<>
+  struct Is_pmap_writable<boost::writable_property_map_tag>{
```
sloriot: you're missing mutable lvalue pmap, see [here](https://github.com/CGAL/cgal/blob/master/BGL/include/CGAL/boost/graph/properties.h#L126) (you can also probably save space with a macro if you want). BTW writable tag does not really makes sense if you can't read them...

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,8 +113,156 @@ namespace CGAL {
...
+    typedef CGAL::Tag_true result;
+  };
+  //overloads used to select a default map:
+  // use the one passed in the named parameters (user must have initialized it)
+  template <class MapFromNP, class Default_tag, class Dynamic_tag, class Mesh>
+  MapFromNP
+  get_map(MapFromNP m, Default_tag, Dynamic_tag, const Mesh&)
```
sloriot: the name is too generic IMO

BGL/include/CGAL/boost/graph/named_params_helper.h
maxGimeno: Yes but a function `is_writable()` that reeturns false for `writable_tag` doesn't seem very trustable :)

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,9 +113,196 @@ namespace CGAL {
...
+  //overloads used to select a default map:
+  // use the one passed in the named parameters (user must have initialized it)
+  template <class MapFromNP, class Default_tag, class Dynamic_tag, class Mesh>
+  MapFromNP
+  get_ndi_map(MapFromNP m, Default_tag, Dynamic_tag, const Mesh&)
+  {
+    return m;
+  }
+
+  // use the one internal to the mesh (user must have initialized it)
```
sloriot: Looks like the comment is no longer valid.

sloriot:
The doc of the updated functions needs to be updated I think.

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,9 +113,196 @@ namespace CGAL {
...
+  struct Is_pmap_writable<TAG>{ typedef CGAL::Tag_true result; };
+
+  template<typename PMapCategory>
+  struct Is_pmap_writable{
+    typedef CGAL::Tag_false result;
+  };
+
+  CGAL_IS_PMAP_WRITABLE(boost::read_write_property_map_tag)
+  CGAL_IS_PMAP_WRITABLE(boost::writable_property_map_tag)
+  CGAL_IS_PMAP_WRITABLE(boost::lvalue_property_map_tag)
```
sloriot: This is true only if `reference` is not const

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,9 +113,196 @@ namespace CGAL {
...
+  };
+
+  template< class PMap, class Graph>
+  struct Index_map_initializer<
+      typename boost::graph_traits<Graph>::vertex_descriptor,
+      PMap, Graph,
+      CGAL::Tag_true>{
+    void operator()(PMap map, const Graph& g)
+    {
+      CGAL::helpers::init_vertex_indices(g, map);
```
sloriot: I'm still wondering if it makes sense to overwrite all the time the indices of the user map. I think it would be safer to first check that indices are not correctly initialized (by using a `std::vector<bool>(num_vertex(g))` to check for duplicate indices+each index < `num_vertex(g)`).

sloriot: @MaelRL do you agree?

MaelRL: It's a problem (see issue #1918), but it seems the initialization is here only performed when a dynamic pmap has been requested (i.e. no index map in the NP); @maxGimeno?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -152,9 +152,9 @@ namespace Polygon_mesh_processing {
   * but `face(h, pmesh)` does not belong to the patch.
   *
   * @tparam PolygonMesh model of `HalfedgeGraph`. If `PolygonMesh`
-  *  has an internal property map
+  *  has an internal not writable property map
```
MaelRL: `read-only` maybe ?

BGL/include/CGAL/boost/graph/named_params_helper.h
MaelRL: Actually it does seem wrong: https://github.com/CGAL/cgal/pull/4209/files#diff-aad529ba52ba39aa444683bce2b8c679R179-R195. The internal pmap should not be overwritten, whether it has already been initialized or not.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
lrineau: The wording I see the most is "non modifiable", but in the documentation of Boost property map, the wording is "readable", "writable":
https://www.boost.org/doc/libs/1_71_0/libs/property_map/doc/property_map.html

BGL/include/CGAL/boost/graph/named_params_helper.h
sloriot: I think it should work as follow:

- if the user pass an index map, we do nothing about it but to check that the indices are valid. 
- if the default map is used and it is writable, then we must initialize it if not already initialized
- if there is no default map and nothing is passed, use an internal dynamic one that we of course initialize. Note that If an algorithm is updating the mesh, and the index map is used after the update then we cannot use it safely.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
afabri: Amnd non instead of not

sloriot: for lvalue maps, mutable is also used in boost.

BGL/include/CGAL/boost/graph/named_params_helper.h
maxGimeno: Currently: 
- If there is a map in the np, nothing is done to it.
- Else if there is a writable map in the mesh, we initialize it
- else we create a dynamic pmap and initialize it.

So we never overwrite the indices of the user map, only the ones of the default internal maps.

maxGimeno: > Actually it does seem wrong: https://github.com/CGAL/cgal/pull/4209/files#diff-aad529ba52ba39aa444683bce2b8c679R179-R195. The internal pmap should not be overwritten, whether it has already been initialized or not.

Why is that ?

MaelRL: Because you can imagine a situation where a user has an existing, initialized ID map but forgot to pass it as NP. You then do not want to mess up with this ID map.

maxGimeno: Then you won't ever be able to use this mechanism with a Polyhedron if it has an internal id map, like in the demo.

sloriot: Let's close the loop here and implement what I suggest (safest option IMO). You have all what I suggested but when the is-initialized test for writable internal index map (I already gave the way to do it using a bitset in this thread).

MaelRL: My first remark (https://github.com/CGAL/cgal/pull/4209#discussion_r357528453) is wrongly worded : I only meant "shouldn't overwrite the internal ID map if it's already been initialized" (which is @sloriot's second point). Obviously, if it's not initialized, you can use it.

maxGimeno:
@MaelRL @sloriot I addressed your change-requests. If you are both ok with that version, I suggest we start testing it as soon as possible, as this PR is quite important and blocking for some others.

BGL/include/CGAL/boost/graph/named_params_helper.h
sloriot: @maxGimeno I don't think the is_intialized has been implemented

BGL/include/CGAL/boost/graph/named_params_helper.h
sloriot: Ping!

BGL/include/CGAL/boost/graph/properties.h
```diff
@@ -114,13 +115,73 @@ using boost::face_external_index;
...
+}
+
 // matches read-write property maps
 template <class PolygonMesh, class FaceIndexMap, class Tag>
 void init_face_indices(PolygonMesh& pm,
                        FaceIndexMap& fid,
                        boost::read_write_property_map_tag,
                        Tag)
 {
+  if(is_face_index_map_valid(pm, fid))
```
maxGimeno: @sloriot the is_initialized is here

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,204 @@ namespace CGAL {
...
+    void operator()(PMap map,const Graph& g)
+    {
+      CGAL::helpers::init_face_indices(g, map);
+    }
+  };
+
+#define CGAL_IS_PMAP_WRITABLE(TAG) template<typename IsRefConst> \
+  struct Is_pmap_writable<TAG, IsRefConst>{ typedef CGAL::Tag_true result; }; \
+  template<> \
+  struct Is_pmap_writable<TAG,std::integral_constant<bool, true> >{ typedef CGAL::Tag_false result; };
```
maxGimeno: @sloriot now it takes the constness into accont

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,204 @@ namespace CGAL {
...
+    return m;
+  }
+
+  // use the one internal to the mesh (it will be init if writable)
+  template <class Default_tag, class Dynamic_tag, class Mesh>
+  typename boost::property_map<Mesh, Default_tag >::const_type
+  get_ndi_map(CGAL::internal_np::Param_not_found, Default_tag t, Dynamic_tag , const Mesh& m)
+  {
+    typename boost::property_map<Mesh, Default_tag >::const_type map = get(t, m);
+    Index_map_initializer<
```
sloriot: don't you think that some intermediate typedef would help reading?

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,204 @@ namespace CGAL {
...
+    {
+      CGAL::helpers::init_halfedge_indices(g, map);
+    }
+  };
+
+  template< class PMap, class Graph>
+  struct Index_map_initializer<
+      typename boost::graph_traits<Graph>::face_descriptor,
+      PMap, Graph,
+      CGAL::Tag_true>{
```
lrineau: Any code design that supposes that `vertex_descriptor`, `halfedge_descriptor`, and `face_descriptor` are distinct types is broken by design. I have already said so multiple times, and we should stop assuming that the three types are distinct!

You should:
  - use the `Default_tag` or `Dynamic_tag` to distinguish between the three cases,
  - and use free functions overloads, instead of template specialization.

maxGimeno: free functions overloads that would decide vertex/halfedge/face initialization based on the named_parameter they receive, right ?

BGL/doc/BGL/CGAL/boost/graph/properties.h
```diff
@@ -34,15 +34,6 @@ enum face_index_t { face_index };
...
 enum vertex_point_t { vertex_point };
 
-
-/// @}
-} // namespace boost
-
-namespace CGAL {
-
-
-/// @{
```
maxGimeno: Error here

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,205 @@ namespace CGAL {
...
+    const Mesh& m;
+    const NamedParameters& np;
+    const Parameter p;
+
+  public:
+    //get the Default tag :
+    //if Mesh has an internal property map for Tag, use Tag, else use the Dynamic_tag.
+    typedef typename boost::mpl::if_c<CGAL::graph_has_property<Mesh, Tag>::value
+    , Tag
+    , Dynamic_tag
```
MaelRL: indentation

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,205 @@ namespace CGAL {
...
+  template <class Default_tag, class Dynamic_tag, class Mesh>
+  typename boost::property_map<Mesh, Default_tag >::const_type
+  get_ndi_map(CGAL::internal_np::Param_not_found, Default_tag t, Dynamic_tag , const Mesh& m)
+  {
+    typedef typename boost::property_map<Mesh, Default_tag >::const_type Map_const_type;
+    typedef typename boost::property_traits
+        <typename boost::property_map<Mesh, Default_tag >
+        ::const_type>::category Category;
+    typedef typename boost::property_traits<typename boost::property_map<Mesh, Default_tag >
+        ::const_type>::reference Reference;
```
MaelRL: You have just defined `Map_const_type` above, can just use it

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,205 @@ namespace CGAL {
...
+
+    PropertyMapType property_map()
+    {
+      return internal_np::get_ndi_map(
+            parameters::get_parameter(np, p),
+            Final_tag(),
+            dtag,
+            m);
+    }
+  };
```
MaelRL: sudden switch to 40 characters wrapping

BGL/include/CGAL/boost/graph/properties.h
```diff
@@ -86,13 +91,37 @@ using boost::face_external_index;
...
+
+
 // matches read-write property maps
 template <class PolygonMesh, class FaceIndexMap, class Tag>
 void init_face_indices(PolygonMesh& pm,
                        FaceIndexMap& fid,
                        boost::read_write_property_map_tag,
                        Tag)
 {
+  if(is_index_map_valid(num_faces(pm), faces(pm), fid))
```
MaelRL: `num_faces(pm)` is not necessarily equal to the size of `faces(pm)`

BGL/include/CGAL/boost/graph/properties.h
```diff
@@ -86,13 +91,37 @@ using boost::face_external_index;
...
+//check that an existing map is initialized/valid
+template <class SimplexRange, class IndexMap>
+bool is_index_map_valid( std::size_t num_simplices,
+                        const SimplexRange& range,
+                        const IndexMap& idmap)
+{
+  
+  typedef typename boost::property_traits<IndexMap>::value_type Id_type;
+  Id_type max_id = static_cast<Id_type>(num_simplices);
+  std::vector<bool> indices(max_id + 1);
```
MaelRL: Why max_id +1, the max_id that an element can have is size-1, here you seem to be at size+1?

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -112,6 +114,205 @@ namespace CGAL {
...
+  {
+    typedef typename boost::property_map<Mesh, Dynamic_tag >::const_type Map_const_type;
+    Map_const_type map = get(t,m);
+    initialize_index_map(map, m, t, CGAL::Tag_true());
+    return map;
+  }
+
+  }//end of internal_np
+
+  namespace Polygon_mesh_processing
```
MaelRL: Why is this in PMP and not directly in CGAL? Seems like it should be useful at plenty of other places too, no?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -171,8 +171,10 @@ connected_component(typename boost::graph_traits<PolygonMesh>::face_descriptor s
  * \ingroup keep_connected_components_grp
  *  computes for each face the index of the corresponding connected component.
  *
- *  A property map for `CGAL::face_index_t` must be either available as an internal property map 
- *  to `pmesh` or provided as one of the \ref pmp_namedparameters "Named Parameters".
+ * If `PolygonMesh`
+  *  has an internal not writable property map
```
MaelRL: see https://github.com/CGAL/cgal/pull/4209#discussion_r357536415

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -171,8 +171,10 @@ connected_component(typename boost::graph_traits<PolygonMesh>::face_descriptor s
...
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL: else -> otherwise (same at other places)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -213,13 +215,15 @@ connected_components(const PolygonMesh& pmesh,
...
     = choose_parameter(get_parameter(np, internal_np::edge_is_constrained),
                    internal::No_constraint<PolygonMesh>());
 
-  typedef typename GetFaceIndexMap<PolygonMesh, NamedParameters>::const_type  FaceIndexMap;
-  FaceIndexMap fimap = choose_parameter(get_parameter(np, internal_np::face_index),
-                                        get_const_property_map(boost::face_index, pmesh));
+  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FaceIndexMap;
+  
+  FaceIndexMap  fimap
+      = get_initialized_face_index_map(pmesh, np);
```
MaelRL: indentation

Polygon_mesh_processing/include/CGAL/Rigid_triangle_mesh_collision_detection.h
```diff
@@ -537,11 +537,9 @@ class Rigid_triangle_mesh_collision_detection
       std::vector<std::size_t> cc_ids(num_faces(tm));
 
       // face index map
-      typedef typename CGAL::GetFaceIndexMap<TriangleMesh, NamedParameters>::type Fid_map;
 
-      Fid_map fid_map =
-        parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                            get_const_property_map(boost::face_index, tm));
+      auto fid_map =
+        CGAL::Polygon_mesh_processing::get_initialized_face_index_map(tm, np);
```
MaelRL: Are there really only places in PMP where this changes stuff? IIRC, SMP also wants nicely initialized index maps too, no?

BGL/include/CGAL/boost/graph/properties.h
maxGimeno: ?

MaelRL: You're checking if the faces are well indexed by defining it as a unique ID between 0 and num_faces(pm), but well indexed should likely be defined as each face has a unique ID between 0 and faces(pm).size(). Not sure if it'd actually mess up some algorithms, but it just seems safer.

sloriot: That's not true, the code is correct and it should be between 0 and num_faces-1 (think about defaut index map for a surface mesh with garbage).

MaelRL: I understand that it looks correct from the data structure POV, but looking from the algorithm POV, don't algorithms expect the face IDs to respresent a unique ID between 0 and the size of the range of faces being passed? For example if I make a vector `vector<whatever> v(faces(pm).size())` and decide to do `v[get(fidmap, f)] = my_whatever;` then it'll explode. Sure I should have known better than the size of the range is smaller than `num_faces(pm)`, but it's creating an opening for user failure.

On the other side, I imagine that in cases where the fidmap is based on just the index of the element in an index-based data structures, then it would be annoying to have to recreate a "proper" ID map.

Polygon_mesh_processing/include/CGAL/Rigid_triangle_mesh_collision_detection.h
maxGimeno: I've seen traces of index_maps in shortest_path, but the whole class is templated by them, and there are no named parameters. It would require a big effort to make it use this mechanism, and the small feature that goes with it. I'm not doing it. At least not in that PR.

BGL/include/CGAL/boost/graph/properties.h
sloriot: This does come from boost and it actually make sense if you have an algorithm removing elements (you do not want to corrupt properties of element that were not touched). If you need such a strict index, we could have a wrapper with an extra indirection through a vector.

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -238,58 +234,53 @@ namespace CGAL {
   NP,                                                         \
   typename boost::property_map<TM, Final_tag >::const_type    \
   > ::type  type;                                             \
-  };
-
+};
+
```
MaelRL: Please, just use the following setting please

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/border.h
```diff
@@ -178,20 +178,9 @@ namespace Polygon_mesh_processing {
...
-    {
-      //face index map is not given in named parameters, nor as an internal property map
-      return internal::border_halfedges_impl(faces, out, pmesh);
-    }
 
-    //face index map given as a named parameter, or as an internal property map
-    FIMap fim = parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                                             get_const_property_map(CGAL::face_index, pmesh));
+    auto fim =
+        CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
    auto fim = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -171,8 +171,10 @@ connected_component(typename boost::graph_traits<PolygonMesh>::face_descriptor s
  * \ingroup keep_connected_components_grp
  *  computes for each face the index of the corresponding connected component.
  *
- *  A property map for `CGAL::face_index_t` must be either available as an internal property map 
- *  to `pmesh` or provided as one of the \ref pmp_namedparameters "Named Parameters".
+ * If `PolygonMesh`
+ *  has an internal non writable property map
```
MaelRL:
```suggestion
 * If `PolygonMesh` has an internal non modifiable property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -213,14 +215,15 @@ connected_components(const PolygonMesh& pmesh,
...
     = choose_parameter(get_parameter(np, internal_np::edge_is_constrained),
                    internal::No_constraint<PolygonMesh>());
 
-  typedef typename GetFaceIndexMap<PolygonMesh, NamedParameters>::const_type  FaceIndexMap;
-  FaceIndexMap fimap = choose_parameter(get_parameter(np, internal_np::face_index),
-                                        get_const_property_map(boost::face_index, pmesh));
-
+  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FaceIndexMap;
+  
+  FaceIndexMap  fimap = get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FaceIndexMap  fimap = get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -323,9 +326,10 @@ std::size_t number_of_connected_components(const PolygonMesh& pmesh)
  * By default, the size of a face is `1` (and thus the size of a connected component is the number
  * of faces it contains), but it is also possible to pass custom sizes, such as the area of the face.
  *
- * Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
- * must be either available as internal property maps
- * to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -368,11 +372,9 @@ std::size_t keep_largest_connected_components(PolygonMesh& pmesh,
   using parameters::choose_parameter;
   using parameters::get_parameter;
 
-  // FaceIndexMap
-  typedef typename GetFaceIndexMap<PM, NamedParameters>::type            FaceIndexMap;
-  FaceIndexMap fimap = choose_parameter(get_parameter(np, internal_np::face_index),
-                                        get_property_map(boost::face_index, pmesh));
+  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FaceIndexMap;
 
+  FaceIndexMap fimap = get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FaceIndexMap;
  FaceIndexMap fimap = get_initialized_face_index_map(pmesh, np);

```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -456,9 +458,10 @@ std::size_t keep_largest_connected_components(PolygonMesh& pmesh,
  * the size of a connected component is the number of faces it contains), but it is also possible
  * to pass custom sizes, such as the area of the face.
  *
- * Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
- * must be either available as internal property maps
- * to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -505,10 +508,9 @@ std::size_t keep_large_connected_components(PolygonMesh& pmesh,
   using parameters::choose_parameter;
   using parameters::get_parameter;
 
-  // FaceIndexMap
-  typedef typename GetFaceIndexMap<PM, NamedParameters>::type             FaceIndexMap;
-  FaceIndexMap fim = choose_parameter(get_parameter(np, internal_np::face_index),
-                                      get_property_map(boost::face_index, pmesh));
+  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FaceIndexMap;
+  FaceIndexMap fim =
+      CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FaceIndexMap fim = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -721,9 +722,10 @@ void keep_or_remove_connected_components(PolygonMesh& pmesh
 * \note If the removal of the connected components makes `pmesh` a non-manifold surface,
 * then the behavior of this function is undefined.
 *
-* Property maps for `CGAL::vertex_index_t`
-* must be either available as internal property map
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -764,9 +766,10 @@ void keep_connected_components(PolygonMesh& pmesh
...
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -805,9 +808,10 @@ void remove_connected_components(PolygonMesh& pmesh
 *  keeps the connected components not designated by the faces in `components_to_remove`,
 *  and removes the other connected components and all isolated vertices.
 *
-* Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
-* must be either available as internal property maps
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -842,9 +846,9 @@ void remove_connected_components(PolygonMesh& pmesh
   using parameters::get_parameter;
 
   //FaceIndexMap
-  typedef typename GetFaceIndexMap<PM, CGAL_PMP_NP_CLASS>::type FaceIndexMap;
-  FaceIndexMap fim = choose_parameter(get_parameter(np, internal_np::face_index),
-                                  get_property_map(boost::face_index, pmesh));
+  typedef typename Default_face_index_map<CGAL_PMP_NP_CLASS, PolygonMesh>::type FaceIndexMap;
+  FaceIndexMap fim =
+      CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FaceIndexMap fim = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -863,9 +867,10 @@ void remove_connected_components(PolygonMesh& pmesh
 *  keeps the connected components designated by the faces in `components_to_keep`,
 *  and removes the other connected components and all isolated vertices.
 *
-* Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
-* must be either available as internal property maps
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
```
MaelRL:
```suggestion
* If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -900,9 +905,9 @@ void keep_connected_components(PolygonMesh& pmesh
   using parameters::get_parameter;
 
   //FaceIndexMap
-  typedef typename GetFaceIndexMap<PM, CGAL_PMP_NP_CLASS>::type FaceIndexMap;
-  FaceIndexMap fim = choose_parameter(get_parameter(np, internal_np::face_index),
-                                  get_property_map(boost::face_index, pmesh));
+  typedef typename Default_face_index_map<CGAL_PMP_NP_CLASS, PolygonMesh>::type FaceIndexMap;
+  FaceIndexMap fim =
+      CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FaceIndexMap fim = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/corefinement.h
```diff
@@ -187,8 +185,8 @@ bool does_bound_a_volume(const TriangleMesh& tm, const NamedParameters& np)
   Vpm vpm = parameters::choose_parameter(parameters::get_parameter(np, internal_np::vertex_point),
                                 get_const_property_map(boost::vertex_point, tm));
 
-  Fid_map fid_map = parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                                        get_const_property_map(boost::face_index, tm));
+  auto fid_map =
+      CGAL::get_initialized_face_index_map(tm, np);
```
MaelRL:
```suggestion
  auto fid_map = CGAL::get_initialized_face_index_map(tm, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/corefinement.h
```diff
@@ -1014,10 +1011,10 @@ namespace experimental {
   Vpm vpm = parameters::choose_parameter(parameters::get_parameter(np, internal_np::vertex_point),
                                 get_property_map(boost::vertex_point, tm));
 // Face index map
-  typedef typename GetFaceIndexMap<TriangleMesh,
-                                   NamedParameters>::type Fid_map;
-  Fid_map fid_map = parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                                        get_property_map(boost::face_index, tm));
+  typedef typename Default_face_index_map<NamedParameters, TriangleMesh>::type Fid_map;
+  Fid_map fid_map =
+      CGAL::get_initialized_face_index_map(tm, np);
```
MaelRL:
```suggestion
  Fid_map fid_map = CGAL::get_initialized_face_index_map(tm, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/detect_features.h
```diff
@@ -139,9 +139,8 @@ detect_surface_patches(PolygonMesh& p,
                        const NamedParameters& np)
 {
   //extract types from NPs
-  typename GetFaceIndexMap<PolygonMesh, NamedParameters>::const_type
-          fimap = parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                                      get_const_property_map(boost::face_index, p));
+  auto fimap =
+      CGAL::get_initialized_face_index_map(p, np);
```
MaelRL:
```suggestion
  auto fimap = CGAL::get_initialized_face_index_map(p, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/detect_features.h
```diff
@@ -373,8 +372,10 @@ namespace internal
  * computing a
  * surface patch id for each face.
  *
- * A property map for `CGAL::face_index_t` must be either available
- * as an internal property map to `pmesh` or provided as one of the Named Parameters.
+ * If `PolygonMesh`
+  *  has an internal not writable property map
```
MaelRL:
```suggestion
 * If `PolygonMesh` has a non modifiable internal property map
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/detect_features.h
```diff
@@ -373,8 +372,10 @@ namespace internal
...
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
  *  as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -484,8 +484,8 @@ void orient(TriangleMesh& tm, const NamedParameters& np)
   Vpm vpm = CGAL::parameters::choose_parameter(get_parameter(np, internal_np::vertex_point),
                          get_const_property_map(boost::vertex_point, tm));
 
-  Fid_map fid_map = CGAL::parameters::choose_parameter(get_parameter(np, internal_np::face_index),
-                                 get_const_property_map(boost::face_index, tm));
+  auto fid_map =
+      CGAL::get_initialized_face_index_map(tm, np);
```
MaelRL:
```suggestion
  auto fid_map = CGAL::get_initialized_face_index_map(tm, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -591,8 +591,8 @@ void orient_to_bound_a_volume(TriangleMesh& tm,
   Vpm vpm = CGAL::parameters::choose_parameter(get_parameter(np, internal_np::vertex_point),
                          get_const_property_map(boost::vertex_point, tm));
 
-  Fid_map fid_map = CGAL::parameters::choose_parameter(get_parameter(np, internal_np::face_index),
-                                 get_const_property_map(boost::face_index, tm));
+  auto fid_map =
+    CGAL::get_initialized_face_index_map(tm, np);
```
MaelRL:
```suggestion
  auto fid_map = CGAL::get_initialized_face_index_map(tm, np);
```

Polygon_mesh_processing/include/CGAL/Rigid_triangle_mesh_collision_detection.h
```diff
@@ -537,11 +537,9 @@ class Rigid_triangle_mesh_collision_detection
       std::vector<std::size_t> cc_ids(num_faces(tm));
 
       // face index map
-      typedef typename CGAL::GetFaceIndexMap<TriangleMesh, NamedParameters>::type Fid_map;
 
-      Fid_map fid_map =
-        parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                            get_const_property_map(boost::face_index, tm));
+      auto fid_map =
+        CGAL::get_initialized_face_index_map(tm, np);
```
MaelRL:
```suggestion
      auto fid_map = CGAL::get_initialized_face_index_map(tm, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h
```diff
@@ -185,9 +185,9 @@ collect_duplicated_stitchable_boundary_edges
   if(per_cc)
   {
     cc = get(Face_property_tag(), pmesh);
-    typedef typename GetFaceIndexMap<PM, CGAL_PMP_NP_CLASS>::const_type FIMap;
-    FIMap fim = parameters::choose_parameter(parameters::get_parameter(np, internal_np::face_index),
-                                             get_const_property_map(face_index, pmesh));
+
+    auto fim =
+      CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
    auto fim = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -338,9 +339,9 @@ void split_long_edges(const EdgeRange& edges
   VPMap vpmap = choose_parameter(get_parameter(np, internal_np::vertex_point),
                              get_property_map(vertex_point, pmesh));
 
-  typedef typename GetFaceIndexMap<PM, NamedParameters>::type FIMap;
-  FIMap fimap = choose_parameter(get_parameter(np, internal_np::face_index),
-                             get_property_map(face_index, pmesh));
+  typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FIMap;
+  FIMap fimap =
+    CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FIMap fimap = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -159,10 +161,9 @@ void isotropic_remeshing(const FaceRange& faces
   typedef typename GetVertexPointMap<PM, NamedParameters>::type VPMap;
   VPMap vpmap = choose_parameter(get_parameter(np, internal_np::vertex_point),
                              get_property_map(vertex_point, pmesh));
-
-  typedef typename GetFaceIndexMap<PM, NamedParameters>::type FIMap;
-  FIMap fimap = choose_parameter(get_parameter(np, internal_np::face_index),
-                           get_property_map(face_index, pmesh));
+typedef typename Default_face_index_map<NamedParameters, PolygonMesh>::type FIMap;
+  FIMap fimap =
+      CGAL::get_initialized_face_index_map(pmesh, np);
```
MaelRL:
```suggestion
  FIMap fimap = CGAL::get_initialized_face_index_map(pmesh, np);
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -41,9 +41,11 @@ namespace Polygon_mesh_processing {
 *         The descriptor types `boost::graph_traits<PolygonMesh>::%face_descriptor`
 *         and `boost::graph_traits<PolygonMesh>::%halfedge_descriptor` must be
 *         models of `Hashable`.
-*         If `PolygonMesh` has an internal property map for `CGAL::face_index_t`,
-*         and no `face_index_map` is given
-*         as a named parameter, then the internal one must be initialized
+*         If `PolygonMesh`
+  *  has an internal not writable property map
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
*         If `PolygonMesh` has a non modifiable internal property map
*         for `CGAL::face_index_t` and no `face_index_map` is given
*         as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -542,8 +542,10 @@ void orient(TriangleMesh& tm)
  * See \ref coref_def_subsec for a precise definition.
  *
  * @tparam TriangleMesh a model of `MutableFaceGraph`, `HalfedgeListGraph` and `FaceListGraph`.
- *                      If `TriangleMesh` has an internal property map for `CGAL::face_index_t`,
- *                      as a named parameter, then it must be initialized.
+ *                      If `TriangleMesh`
+  *  has an internal not writable property map
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
 *                      If `TriangleMesh` has a non modifiable internal property map
 *                      for `CGAL::face_index_t` and no `face_index_map` is given
 *                      as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
```diff
@@ -438,8 +438,10 @@ void recursive_orient_volume_ccs( TriangleMesh& tm,
 * inward or outward oriented.
 *
 * @tparam TriangleMesh a model of `FaceListGraph` and `MutableFaceGraph` .
-*                      If `TriangleMesh` has an internal property map for `CGAL::face_index_t`,
-*                      as a named parameter, then it must be initialized.
+*                      If `TriangleMesh`
+  *  has an internal not writable property map
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
*                      If `TriangleMesh` has a non modifiable internal property map
*                      for `CGAL::face_index_t` and no `face_index_map` is given
*                      as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -863,9 +867,10 @@ void remove_connected_components(PolygonMesh& pmesh
 *  keeps the connected components designated by the faces in `components_to_keep`,
 *  and removes the other connected components and all isolated vertices.
 *
-* Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
-* must be either available as internal property maps
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
+*  for `CGAL::face_index_t` or `CGAL::vertex_index_t` and no `face_index_map` (respectively `vertex_index_map`) is given
+*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp.  `vertex_index_map`) is given
*  as a named parameter, then the internal one must be initialized. Otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -805,9 +808,10 @@ void remove_connected_components(PolygonMesh& pmesh
 *  keeps the connected components not designated by the faces in `components_to_remove`,
 *  and removes the other connected components and all isolated vertices.
 *
-* Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
-* must be either available as internal property maps
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
+*  for `CGAL::face_index_t` or `CGAL::vertex_index_t` and no `face_index_map` (respectively `vertex_index_map`) is given
+*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp. `vertex_index_map`) is given
*  as a named parameter, then the internal one must be initialized. Otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -764,9 +766,10 @@ void keep_connected_components(PolygonMesh& pmesh
 * \note If the removal of the connected components makes `pmesh` a non-manifold surface,
 * then the behavior of this function is undefined.
 *
-* Property maps for `CGAL::vertex_index_t`
-* must be either available as internal property map
-* to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
+*  for `CGAL::face_index_t` or `CGAL::vertex_index_t` and no `face_index_map` (respectively `vertex_index_map`) is given
+*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp. `vertex_index_map`) is given
*  as a named parameter, then the internal one must be initialized. Otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -721,9 +722,10 @@ void keep_or_remove_connected_components(PolygonMesh& pmesh
...
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp. `vertex_index_map`) is given
*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -456,9 +458,10 @@ std::size_t keep_largest_connected_components(PolygonMesh& pmesh,
  * the size of a connected component is the number of faces it contains), but it is also possible
  * to pass custom sizes, such as the area of the face.
  *
- * Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
- * must be either available as internal property maps
- * to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
+*  for `CGAL::face_index_t` or `CGAL::vertex_index_t` and no `face_index_map` (respectively `vertex_index_map`) is given
+*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp. `vertex_index_map`) is given
*  as a named parameter, then the internal one must be initialized. Otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/clip.h
```diff
@@ -274,8 +274,9 @@ clip_to_bbox(const Plane_3& plane,
   * \pre \link CGAL::Polygon_mesh_processing::does_bound_a_volume() `CGAL::Polygon_mesh_processing::does_bound_a_volume(clipper)` \endlink
   *
   * @tparam TriangleMesh a model of `MutableFaceGraph`, `HalfedgeListGraph` and `FaceListGraph`.
-  *                      If `TriangleMesh` has an internal property map for `CGAL::face_index_t`,
-  *                      as a named parameter, then it must be initialized.
+  * If `TriangleMesh` *  has an internal not writable property map
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
  * If `TriangleMesh` has an internal not modifiable property map
  *  for `CGAL::face_index_t` and no `face_index_map` is given
  *  as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/clip.h
```diff
@@ -365,8 +366,9 @@ bool dispatch_clip_call(TriangleMesh& tm, TriangleMesh& clipper,
   * \pre \link CGAL::Polygon_mesh_processing::does_self_intersect() `!CGAL::Polygon_mesh_processing::does_self_intersect(tm)` \endlink
   *
   * @tparam TriangleMesh a model of `MutableFaceGraph`, `HalfedgeListGraph` and `FaceListGraph`.
-  *                      If `TriangleMesh` has an internal property map for `CGAL::face_index_t`,
-  *                      as a named parameter, then it must be initialized.
+  * If `TriangleMesh` has an internal not writable property map
+  *  for `CGAL::face_index_t` and no `face_index_map` is given
+  *  as a named parameter, then the internal one must be initialized; else, it will be.
```
MaelRL:
```suggestion
  * If `TriangleMesh` has an internal non modifiable property map
  *  for `CGAL::face_index_t` and no `face_index_map` is given
  *  as a named parameter, then the internal one must be initialized; otherwise, it will be.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -323,9 +326,10 @@ std::size_t number_of_connected_components(const PolygonMesh& pmesh)
  * By default, the size of a face is `1` (and thus the size of a connected component is the number
  * of faces it contains), but it is also possible to pass custom sizes, such as the area of the face.
  *
- * Property maps for `CGAL::face_index_t` and `CGAL::vertex_index_t`
- * must be either available as internal property maps
- * to `pmesh` or provided as \ref pmp_namedparameters "Named Parameters".
+* If `PolygonMesh`
+*  has a non writable internal property map
+*  for `CGAL::face_index_t` or `CGAL::vertex_index_t` and no `face_index_map` (respectively `vertex_index_map`) is given
+*  as a named parameter, then the internal one(s) must be initialized. Otherwise, it will be.
```
MaelRL:
```suggestion
*  for `CGAL::face_index_t` (resp. `CGAL::vertex_index_t`) and no `face_index_map` (resp. `vertex_index_map`) is given
*  as a named parameter, then the internal one must be initialized. Otherwise, it will be.
```

maxGimeno:
[Travis error](https://travis-ci.org/CGAL/cgal/jobs/658213025#L4277)

MaelRL:
So, the code now works, and you can do something like:
```
  typedef typename GetInitializedFaceIndexMap<TriangleMesh, NamedParameters>::const_type Fid_map;
  Fid_map fid_map = get_initialized_face_index_map(tm, np);
```
which will:
- look if there's a face index map in the named parameters, and return it if it's there;
- if not, it will check if there is an internal property map (initializing it if needed);
- if no internal pmap, it will construct a dynamic pmap and initialize it.

However, I think that the construction is not fully sound. In my opinion we should not divert from the `choose_param(get_param(), default)` that is used for everything else. Instead it should be `choose_param(get_param(index_tag, np), get_initialized_index_map())`.

Avoiding the lengthy and repetitive `choose_param(get_param(...), default)` is not specific to this particular property tag, and it should be solved another way: for example, rename `get_param()` to `extract_param()` and use `get_param(tag, np, default) := choose_param(extract_param(tag, np), default))`.

To summarize, my proposition is to have `get_initialized_face_index_map` not handle the possibility of a named parameters being included, and instead create a new `parameters::` convenience function, which will make the call something like:
```
  typedef typename GetInitializedFaceIndexMap<TriangleMesh, NamedParameters>::const_type Fid_map;
  Fid_map fid_map = parameters::get_param(internal_np::face_index_t, np, get_initialized_face_index_const_property_map(tm));

  typedef typename GetVertexPoint<TriangleMesh, NamedParameters>::const_type VPM;
  VPM vpm = parameters::get_param(internal_np::vertex_point_t, np, get_const_property_map(boost::vertex_point, tm));
```
Actually we can even imagine having `get_initialized_face_index_const_property_map(tm)` simply be called `get_const_propertymap(boost::initialized_face_index_t, tm)` which whould give:
```
  typedef typename GetInitializedFaceIndexMap<TriangleMesh, NamedParameters>::const_type Fid_map;
  Fid_map fid_map = parameters::get_param(internal_np::face_index_t, np, get_const_property_map(boost::initialized_face_point, tm));

  typedef typename GetVertexPoint<TriangleMesh, NamedParameters>::const_type VPM;
  VPM vpm = parameters::get_param(internal_np::vertex_point_t, np, get_const_property_map(boost::vertex_point, tm));
```
which can then be easily extended to other property tags by just providing overloads of `get(_const)_property_map`.

Note that these functions are only meant to be used by devs.

cc: @CGAL/geometryfactory

sloriot:
Why is `GetInitializedFaceIndexMap` not named `Get_initialized_face_index_map`?

sloriot:
`GetVertexPoint` -> `Get_vertex_point_map`?

I'm  not against sticking with the class `property_map` + new tags for extracting the default maps. For naming convention purpose, we might want to introduce our own `Property_map` extending the boost one. For the extraction function, I really think that we should have some short functions doing the job for us instead of having to repeat again and again the same thing, so a new convenience in `parameters` would indeed be nicer.

maxGimeno:
I don't understand why we should split a mechanism that now works in one call into several. I understand that the new parameters mechanism would avoid a call to chose_param(), but it doesn't seem that much of a gain to me, and removing te np from get_initialized_xxx_map would make it heavier to use, which removes a third of the interest to me :/

maxGimeno:
[Travis error](https://travis-ci.com/CGAL/cgal/jobs/295434009#L3353)

maxGimeno:
[Runtime error]( https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-95/Polygon_mesh_processing/TestReport_gimeno_Fedora-rawhide.gz)

BGL/include/CGAL/boost/graph/Face_filtered_graph.h
```diff
@@ -71,9 +72,9 @@ namespace CGAL
    * \cgalModels \bgllink{VertexListGraph}
    */
 template<typename Graph,
-         typename FIMap = typename boost::property_map<Graph, CGAL::face_index_t>::type,
-         typename VIMap = typename boost::property_map<Graph, boost::vertex_index_t>::type,
-         typename HIMap = typename boost::property_map<Graph, CGAL::halfedge_index_t>::type>
+         typename FIMap = typename CGAL::GetInitializedFaceIndexMap<Graph>::const_type,
+         typename VIMap = typename CGAL::GetInitializedVertexIndexMap<Graph>::const_type,
```
sloriot: Get_initialized_vertex_index_map?

BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -175,37 +177,105 @@ namespace CGAL {
...
+// GetInitializedFaceIndexMap
+
+#define CGAL_DEF_GET_INDEX_TYPE(CTYPE, TYPE)                                                       \
+template <typename Graph,                                                                          \
+          typename NamedParameters =                                                               \
+            CGAL::Named_function_parameters<bool, CGAL::internal_np::all_default_t> >              \
+struct GetInitialized##CTYPE##IndexMap                                                             \
+  : public BGL::internal::GetInitializedIndexMap<internal_np::TYPE##_index_t,                      \
+                                                 boost::TYPE##_index_t,                            \
+                                                 CGAL::dynamic_##TYPE##_property_t<int>,           \
```
sloriot: any reason not to use `unsigned int`?

Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt
```diff
@@ -32,20 +32,27 @@ is the property map with the points associated to the vertices of the polygon me
 \cgalNPEnd
 
 \cgalNPBegin{vertex_index_map} \anchor PMP_vertex_index_map
-is the property map containing the index of each vertex of the input polygon mesh.\n
+is the property map associating a unique index to each vertex of a polygon mesh,
+between `0` and `num_vertices(g)-1`.
+If this parameter is not passed, internal machinery will create and initialize a vertex index
+property map, either using the internal property map if it exists or using an external map. The latter
+might result in  - slightly - worsened performance if case of non-constant complexity for index access.\n
```
sloriot:
```suggestion
might result in  - slightly - worsened performance in case of non-constant complexity for index access.\n
```

Polygon_mesh_processing/doc/Polygon_mesh_processing/NamedParameters.txt
```diff
@@ -32,20 +32,27 @@ is the property map with the points associated to the vertices of the polygon me
...
+<b>Default:</b> an initialized vertex index property map
 \cgalNPEnd
 
 \cgalNPBegin{face_index_map} \anchor PMP_face_index_map
-is the property map containing the index of each face of the input polygon mesh.\n
+is the property map associating a unique index to each face of a polygon mesh,
+between `0` and `num_faces(g)-1`.
+If this parameter is not passed, internal machinery will create and initialize a face index
+property map, either using the internal property map if it exists or using an external map. The latter
+might result in  - slightly - worsened performance if case of non-constant complexity for index access.\n
```
sloriot:
```suggestion
might result in  - slightly - worsened performance in case of non-constant complexity for index access.\n
```

BGL/include/CGAL/boost/graph/named_params_helper.h
MaelRL: Should it actually even be `vertices_size_type`, `halfedges_size_type`, ...?

BGL/include/CGAL/boost/graph/Face_filtered_graph.h
MaelRL: Every other getter in named_params_helper is using caml case for some reason. This can be changed but should be done CGAL-wide in a different PR, I'd say.

BGL/include/CGAL/boost/graph/named_params_helper.h
sloriot: Agreed

maxGimeno:

sloriot:
As this feature is central for almost all PRs around PMP/BGL I'm merging it now without waiting the 2 weeks approval delay as I don't see somebody complaining that we are removing a constraint.

sloriot:
After merge TODO mentioned in the 5.1 release check list

lrineau:
@MaelRL @sloriot As the TODO-after-merge been done?

sloriot:
Yes and it works: `https://doc.cgal.org/5.1.1/BGL/group__PkgBGLHelper.html` will redirect you to `https://doc.cgal.org/5.1.1/BGL/group__BGLGraphExternalIndices.html`.

nawaid-ahmed98:
Hi Everyone My name is Nawaid Ahmed and i am new in open source projects. currently i am doing compititve programing But Now i want to get started open source .If anyone Guide me i will be be mean a lot to me . you can also contact me on twitter: https://twitter.com/nawaidahmed98 . Thankyou

PR #4208, opened 2019-09-03
Add code to prevent silent bad usage of boost::get_param

sloriot:

PR #4211, opened 2019-09-09
available only if c++11 is available

sloriot:
Fix for #4208 

The change shall make it in 4.14 branch but not master.


lrineau:
Our testsuite cannot test this PR in `integration`. I will merge it without testing. At least Travis checks are OK.

PR #4212, opened 2019-09-09
CGAL 3D Demo: Enhancement of the Scene Saving System

### LINKED ISSUE #3802: Improve sharing of a demo setting
Right now sharing a complex scene (say more that one file) is complicated.
One step that would simplify that is to have an option to make the scene script file embed the files instead of pointing to existing local files.

If we go one step further, we could even imagine having a way to make the demo host the file online (gist/pastebin/...) so that we could share a URL that you open in the demo and get the scene loaded directly on the fly.

While writing that, I wonder whether we could not even have a way to _connect_ to another demo to get the current camera position to make 1 user remotely showing something.

maxGimeno:
## Summary of Changes

Saves the whole scene into the .js file. It means it is whole items that are stored in the .js, not only paths to existing files. 
The items content are stored as base64 binary data, so it doesn't matter the file format (ascii or binary or both) used.

## Release Management

* Issue(s) solved (if any): fix #3802


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2777,61 +2777,130 @@ void MainWindow::propagate_action()
...
+        //OFF Files (*.off)
+        QRegularExpression re("\\(\\*\\.(.*)\\)");
+        QRegularExpressionMatch rem = re.match(sf);
+        if(!rem.hasMatch())
+          continue;
+        ext = rem.captured(1);
+        QList<Scene_item*>to_save;
+        to_save.append(item);
+        QString savename(tr("%1.%2").arg(item->name()).arg(ext));
+        iop->save(QFileInfo(savename), to_save);
```
lrineau: If the current working directory is not writable, that will fail.

It should use [`QTemporaryFile`](https://doc.qt.io/qt-5/qtemporaryfile.html) within the temporary directory (`QDir::tempPath()`).

sloriot:
Could we add this PR in the testsuite so that it is merged rapidly? Having the new functionality would be very nice.

lrineau:
I still do not feel comfortable with the security issues that pull-request may create. I would like to know more about the attack surface. From far, I can see that the pull-request plays with SSH, and also deals with sockets. All lot of things may go wrong. If that pull-request must be a priority, I suggest we start a meeting to discuss it.

sloriot:
I understand your concerns but any security issue would involve that the demo is ran. So we could set the SSH feature OFF by default (but for the testsuite) and we will switch it on internally for our daily work. I would be worried if I would interact with a random person's demo but not with someone I know. We of course can open an issue to review the security part of the feature more in details but since we lack time I still would like to use it.

maxGimeno:
@lrineau @sloriot  We need to discuss this and find a solution at some point

lrineau:
> @lrineau @sloriot We need to discuss this and find a solution at some point

Then please list the services/ports the demo will start listening to, after this PR, and then we can start the discussion.

Polyhedron/demo/Polyhedron/Use_ssh.cpp
```diff
@@ -0,0 +1,247 @@
...
+namespace ssh_internal{
+
+bool establish_ssh_session(ssh_session &session,
+                           const char* user,
+                           const char* server,
+                           const char* pub_key_path,
+                           const char* priv_key_path,
+                           const char* priv_key_password)
+{
+  int port = 22;
```
maxGimeno: @lrineau here is the port we use for SSH

maxGimeno:
> > @lrineau @sloriot We need to discuss this and find a solution at some point
> 
> Then please list the services/ports the demo will start listening to, after this PR, and then we can start the discussion.

SSH on port 22 and WebSockets are the new protocols used. For the websocket, the user specifies the whole adress for the server to be used, including the port.

Polyhedron/demo/Polyhedron/MainWindow.h
```diff
@@ -1,5 +1,6 @@
 #ifndef MAINWINDOW_H
 #define MAINWINDOW_H
+#undef CGAL_USE_WEBSOCKETS
```
sloriot: fishy

Polyhedron/demo/Polyhedron/include/CGAL/Use_ssh.h
```diff
@@ -0,0 +1,25 @@
+#ifndef USE_SSH_H
```
sloriot: license?

Polyhedron/demo/Polyhedron/MainWindow.ui
```diff
@@ -463,10 +465,25 @@
...
+  </action>
+  <action name="action_Start_a_Session">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>&amp;Start a Session</string>
+   </property>
+   <property name="toolTip">
+    <string>Start a WebSocket Server to share your camera with others on your network.</string>
```
sloriot:
```suggestion
    <string>Start a WebSocket Server to Share your Camera with Others on your Network.</string>
```

Polyhedron/demo/Polyhedron/Use_ssh.cpp
```diff
@@ -0,0 +1,247 @@
+#ifdef CGAL_USE_SSH
```
sloriot: license?

lrineau: @maxGimeno Most of that file is probably a copy-paste from somewhere. We should check on which license we can put this file.

lrineau: Do you remember the source of the copy-paste?

Polyhedron/demo/Polyhedron/Use_ssh.cpp
```diff
@@ -0,0 +1,247 @@
...
+    if( ssh_write_knownhost(session) != SSH_OK )
+    {
+      std::cerr << "writeKnownHost failed" << std::endl;
+      return false;
+    }
+    else
+    {
+      ssh_connect(session);
+    }
+  }
```
lrineau: That `if` should pop a dialog warning the user and checking what to do.

Polyhedron/demo/Polyhedron/Use_ssh.cpp
maxGimeno: It is all base don the guide provided by libssh

lrineau: The put the file under the license `LGPLv3+`.

maxGimeno:

PR #4242, opened 2019-09-23
Demo/classif: Fix unresized ranges

sgiraudot:
## Summary of Changes

Some ranges were sent empty to an algorithm that expects allocated ranges.

## Release Management

* Affected package(s): Demo
* Introduced in CGAL 4.14


lrineau:
Fast-merging because we really would like to publish the release soon.

PR #4243, opened 2019-09-24
General CMake improvements

### LINKED ISSUE #3529: Improvements to be done for CGAL CMake scripts
## Issue Details

@afabri and [Max Lyon] (not on Github) showed me how hard it can be to configure the polyhedron demo of CGAL, on Windows.

[Max Lyon]: https://www.graphics.rwth-aachen.de/person/99/

### Various issues
- [x] `FindTBB` displays a message starting with `ERROR: `, for messages that are just warnings. It should check the variables `TBB_FIND_REQUIRED` and `TBB_FIND_QUIET`, and display messages accordingly.
- [x] Qt5 is hard to configure. If CMake does not auto-detect the install of Qt5, that is rather difficult to understand how should the CMake variable `Qt5_DIR` be filled. It is not the root of the installation of Qt5 on Windows.
- [ ] GMP/MPFR: if one uses a version of CGAL installed by the CGAL Windows installer, that is fine, but as soon as you use a git checkout of CGAL, things get complicated.
- [x] Boost... Boost installation layout is strange, on Windows. It is not strait forward to know what to fill in CMake.

### Various proposals

- All CGAL `CMakeLists.txt` should start with `find_package(CGAL REQUIRED)`, before any other `find_package`, so that CMake forces the users to first configure the path to `CGAL_DIR`. That would ensure that `CGALConfig.cmake` (and all the CGAL CMake machinery) is already included for all the remaining lines of the `CMakeLists.txt`.
- That way, we could fill `CMAKE_PREFIX_PATH` with the known standard installation paths for Qt5 and Booost.
- I also suggest that the `else()` sections of our `CMakeLists.txt` explain how to configure the missing dependencies. That could probably be easily done with the usage of the [`FeatureSummary`] module of CMake. OpenFlipper [already uses it](https://bitbucket.org/fenics-project/dolfin/pull-requests/275/use-cmakes-featuresummary-for-generating-a/diff).

[`FeatureSummary`]: https://cmake.org/cmake/help/v3.13/module/FeatureSummary.html

maxGimeno:
## Summary of Changes
Makes CGAL REQUIRED in all CMakeLists.

Add a FeatureSummary in the end of the Polyhedron/demo/Polyhedron/CMakeLists.txt, that lists all missing packages/components, and explains for what purpose they are needed.

## Release Management

* Affected package(s):Installation, Polyhedron
* Issue(s) solved (if any): fix #3529


Polyhedron/demo/Polyhedron/CMakeLists.txt
```diff
@@ -451,7 +451,7 @@ endif (CGAL_Qt5_FOUND AND Qt5_FOUND)
 
 feature_summary(WHAT OPTIONAL_PACKAGES_NOT_FOUND
                 INCLUDE_QUIET_PACKAGES
-                DESCRIPTION "NOTICE: Missing required packages that prevent the demo from being compiled:"
+                DESCRIPTION "NOTICE : Missing required packages that prevent the demo from being compiled:"
```
lrineau, line 454: Why would you want to remove the information about missing dependencies?

It helps see if there exist a platform where all dependencies are installed.

maxGimeno: Because I don't think any platform has ALL of the dependencies installed at once, and so the whole line would turn blue. I don't want that.

sloriot: `if (TESTSUITE_RUNNING)`? (don't know the precise syntax)

maxGimeno: It doesn't change anything to have this whitespace or not if we are not inside the testsuite

lrineau: Actually, I do not see the issue with having that line fully blue. If you manage to get it fully blue, then that would mean that warnings and errors are gone (because otherwise the yellow and red colors have the priority). And one it is fully blue, we can try add dependencies to a few test platforms, to have green cells. I really see any benefit in this patch. TO me, that looks like removing tests because there are red cells.

But maybe I am wrong, and you should explain why that matters.

maxGimeno: Well, at a timùe we are trying to get as green cells as possible in the testsuite, I am not encline to add some blue. And don't expect to see any green cell in that line if I leave it like that, as dependencies like OpenGR come with a bagful of warnings. To me, that will give the impression that most of the platforms don't test the demo, and the very few that do show it doesn't work correctly, when it is not the point.

lrineau: What is true is that originally, the blue cells of 3D demos was when Qt5 was missing. Now that our big 3D demos have a lot of features, and tones of dependencies, we have a lot more opportunities for missing optional dependencies.

maxGimeno:
Strange. In the testsuite some platforms still return a green cell despite having a "NOTICE: " printed out in the results...

lrineau:
> Strange. In the testsuite some platforms still return a green cell despite having a "NOTICE: " printed out in the results...

That probably depends on what follows `NOTICE:`. Remember that the regular expression is a bit more complex:
https://github.com/CGAL/cgal/blob/e6e1c8b55795c94bd02b34e6ff84caa3706bdf12/Testsuite/test/collect_cgal_testresults_from_cmake#L56-L61

maxGimeno:
Oooh yeah.

maxGimeno:

sloriot:
What is the purpose of adding `REQUIRED`, especially if we still have the `if(CGAL_FOUND)`?

maxGimeno:
I think it make the failure more obvious if CGAL is not found.

sloriot:
I mean AFAIU `if (CGAL_FOUND)` is useless then, isn't it?

maxGimeno:
Yes, probably, as there is an erreur if CGAL is not found, anyway

sloriot:
@lrineau can you confirm that we still want to keep `if (CGAL_FOUND)` and that it's not redundant with the fact that `REQUIRED` was added.

lrineau:
> @lrineau can you confirm that we still want to keep `if (CGAL_FOUND)` and that it's not redundant with the fact that `REQUIRED` was added.

No, I disagree. The test of `CGAL_FOUND` is completely useless. If CGAL is not found, `find_package(CGAL REQUIRED)` will terminate the configuration with a fatal error.

Algebraic_kernel_for_circles/test/Algebraic_kernel_for_circles/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(Algebraic_kernel_for_circles_Tests)
 
 find_package(CGAL REQUIRED)
 
-include_directories(BEFORE "include")
```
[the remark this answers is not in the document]
maxGimeno: Is this a special case ? or is it general and we should keep the include_derectories(BEFORE...) everywhere ?

Polyhedron/demo/Polyhedron/CMakeLists.txt
```diff
@@ -114,6 +114,10 @@ endif()
 
 #find libssh for scene sharing
 find_package(LibSSH)
+set_package_properties(
+  LibSSH PROPERTIES
+  DESCRIPTION "A library used to enable the SSH features. "
+  PURPOSE "Requiered for loading (saving) a scene to (from) a distant server.")
```
lrineau, line 120:
```suggestion
  PURPOSE "Required for loading (saving) a scene to (from) a distant server.")
```

AABB_tree/demo/AABB_tree/CMakeLists.txt
```diff
@@ -17,58 +17,55 @@ if(POLICY CMP0071)
 endif()
 
 # Include this package's headers first
-include_directories( BEFORE ./ ./include )
+include_directories(BEFORE ./ ./include)
 
 # Find CGAL and CGAL Qt5
-find_package(CGAL COMPONENTS Qt5)
+find_package(CGAL REQUIRED COMPONENTS Qt5)
```
lrineau, line 23: If you add `REQUIRED` before `COMPONENTS Qt5`, then what happens if CGAL is found but not `CGAL_Qt5`?

maxGimeno: The doc statse "A package-specific list of required components may be listed after the COMPONENTS option (or after the REQUIRED option if present). Additional optional components may be listed after OPTIONAL_COMPONENTS. Available components and their influence on whether a package is considered to be found are defined by the target package."
So as I understand, if Qt5 is not found, there is a CMake error.

lrineau: Then you must remove the `REQUIRED`, here. And re-add the test on `CGAL_FOUND`.

Because Qt5 is not a required dependency for CGAL and, for the CTest testsuite to work correctly, the whole configuration of CGAL (with `-DWITH_tests=ON -DWITH_examples=ON -DWITH_demos=ON`) must finish without errors.

Actually, you should verify. I do not know if our `CGALConfig.cmake` files have a correct behavior with `REQUIRED` and with components.

maxGimeno: I think you are wrong. I had to make all the CGAL components optional instead, and keep the tests as they were. I checked, an OPTIONAL_COMPONENT works as expected even with a REQUIRED CGAL, as stated in the piece of doc I pasted above.

lrineau:
@maxGimeno You should update this PR.

Installation/test/Installation/CMakeLists.txt
```diff
@@ -83,15 +83,6 @@ if(WITH_CGAL_Qt5)
...
 
-if(WIN32 OR CMAKE_SYSTEM_NAME STREQUAL Windows)
-  add_executable(display_dll_version_info display_dll_version_info.cpp)
-  target_link_libraries(display_dll_version_info version)
-  add_executable(test_gmp_mpfr_dll test_gmp_mpfr_dll.cpp)
-  target_link_libraries(test_gmp_mpfr_dll version)
-  cgal_add_test(test_gmp_mpfr_dll)
-  add_to_cached_list(CGAL_EXECUTABLE_TARGETS test_gmp_mpfr_dll)
-endif()
-
```
lrineau: Please do not remove the test `test_gmp_mpfr_dll`.

maxGimeno: It is not removed, this block is already present a few lines before, it has probably been duplicated during a merge.

Polyhedron/demo/Polyhedron/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1...3.15)
+cmake_minimum_required(VERSION 3.1...3.20)
```
lrineau, line 1: That will break the build.

The warning is right.

Polyhedron/demo/Polyhedron/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ if(POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW)
 endif()
 
+if(POLICY CMP0115)
+  cmake_policy(SET CMP0115 OLD)
+endif()
```
maxGimeno, line 22: @lrineau better this way ?

lrineau: The warning related to [`CMP0115`](https://cmake.org/cmake/help/git-stage/policy/CMP0115.html) explains that our CMake code is doing something fishy. Instead of hidden that warning by using the `OLD` behavior (that is fishy, and deprecated by design), the solution is to actually fix the issue. I have stated that in https://github.com/CGAL/cgal/pull/4243/commits/25e65bcedacdad2f5f4645f695cda65da0994b31.

lrineau:
As for the fixes for `CMP0115`, I think we should apply the patch to a release branch, otherwise we will have the warnings in the test results for release branches.

lrineau:
I have added https://github.com/CGAL/cgal/pull/4243/commits/20619a586b33c313ebaacc93544924d3e6c6be30 and tested with a build of CMake-3.20 that was on my machine.

lrineau:
This PR has conflicts with `master`.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-50.shtml

PR #4245, opened 2019-09-24
Surface_mesh: Disable comparison of mixed index types

afabri:
# Summary of Changes

Add `= delete`'d generic comparison templates so that we only compare indices of exactly the same type.  Without this the compiler used the conversion to the underlying integer type and compared them.

## Release Management


afabri:
@maxGimeno ready to test

sloriot:
This PR is particularly interesting as it might spot bugs in the testsuite.

maxGimeno:

PR #4250, opened 2019-09-25
PMP: Make Side_of_triangle_mesh faster for disjoint meshes

### LINKED ISSUE #4247: Speeding up side of triangle mesh
I notice that the runtime of doing a Boolean operation of two disjoint and far apart meshes was actually larger than when they intersect. My guess is that it is coming from the need to build the AABB-tree to decide if one mesh is inside the other.

I think we should only build the tree if the bbox of the meshes (or even obb as soon as we integrate it) are not disjoint.
The same kind of trick can be apply to `Side_of_triangle_mesh`, that is use the bbox of the mesh before building the tree to answer a query.

@maxGimeno or @afabri. you might want to have a look.

maxGimeno:
## Summary of Changes
Use the mesh's bbox until a request falls inside, only then use the AABB_tree.
## Release Management

* Affected package(s):PMP
* Issue(s) solved (if any): fix #4247

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
@@ -76,7 +76,7 @@ namespace CGAL {
  */
 template <class TriangleMesh,
           class GeomTraits,
-          class VertexPointMap = Default
+          class VertexPointMap_ = Default
 #ifndef DOXYGEN_RUNNING
           , class AABBTree = Default
 #endif
@@ -86,26 +86,35 @@ class Side_of_triangle_mesh
   // typedefs
   template <typename TriangleMesh_,
             typename GeomTraits_,
-            typename VertexPointMap_>
+            typename VertexPointMap__>
   struct AABB_tree_default {
     typedef CGAL::AABB_face_graph_triangle_primitive<TriangleMesh_,
-                                                     VertexPointMap_> Primitive;
+                                                     VertexPointMap__> Primitive;
     typedef CGAL::AABB_traits<GeomTraits_, Primitive> Traits;
     typedef CGAL::AABB_tree<Traits> type;
   };
   typedef typename Default::Lazy_get<AABBTree,
                                      AABB_tree_default<TriangleMesh,
                                                        GeomTraits,
-                                                       VertexPointMap>
+                                                       VertexPointMap_>
                                      >::type AABB_tree_;
+  typedef typename Default::Get<VertexPointMap_,
+                                typename boost::property_map<TriangleMesh,
+                                                             vertex_point_t>::const_type>::type
+                                  VertexPointMap;
   typedef typename GeomTraits::Point_3 Point;
 
   //members
   typename GeomTraits::Construct_ray_3     ray_functor;
   typename GeomTraits::Construct_vector_3  vector_functor;
-  const AABB_tree_* tree_ptr;
+  mutable const AABB_tree_* tree_ptr;
+  const TriangleMesh* tm_ptr;
+  boost::optional<VertexPointMap> opt_vpm;
   bool own_tree;
   CGAL::Bbox_3 box;
+#ifdef CGAL_HAS_THREADS
+  mutable CGAL_MUTEX tree_mutex;
+#endif
 
 public:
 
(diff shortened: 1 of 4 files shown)


lrineau:
Could not it have a side-effect for a user construction the `Side_of_triangle_mesh` object and then spawning a pool of threads working with the object?

maxGimeno:
I don't think so. The first time a thread performs a query inside the box it will switch to the tree queries for all of them, and the potential still unfinished test shouldn't be affected by that behavior. The worst thing I could think of is the boolean being set to `true` twice, or a test that should have directly used the tree that first tested the box. As `tree_built` is never set back to false I'm quite confident this is thread safe.

lrineau:
It is true that, for the moment, this PR does not build the AABB tree lazily. The current version seems thread-safe, assuming the operation `tree_built = true;` is atomic. But I am pretty sure the issue #4250 suggested to make the construction of the tree lazy. And in that case, there will be thread-safety issue to deal with.

maxGimeno:
Actually the function `build()` is only called if a query falls inside the bbox, but this thread-safety issue was already here before.

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -195,8 +203,21 @@ class Side_of_triangle_mesh
...
+                    || point.x() > box.xmax()
+                    || point.y() < box.ymin()
+                    || point.y() > box.ymax()
+                    || point.z() < box.zmin()
+                    || point.z() > box.zmax());
+      if(is_outside)
+        return CGAL::ON_UNBOUNDED_SIDE;
+    }
+
+    tree_built = true;
```
sloriot: the Boolean seems useless.

maxGimeno: It enables/disables the first test, I don't see how it is useless.

lrineau: @sloriot suggest that it could be:
```c++
if(!tree_built) {
  if(point.x() < box.xmin()
     || point.x() > box.xmax()
     || point.y() < box.ymin()
     || point.y() > box.ymax()
     || point.z() < box.zmin()
     || point.z() > box.zmax()) return CGAL::ON_UNBOUNDED_SIDE;
}
tree_built = true;
```

maxGimeno: Ok from where he put his comment I thought he was refering to tree_built

sloriot: Well what I was saying was not to refactor the check but that the bool was never used because you don't built the tree lazily for now (AFAIS)

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -195,8 +203,21 @@ class Side_of_triangle_mesh
...
     return internal::Point_inside_vertical_ray_cast<GeomTraits, AABB_tree>()(
-      point, *tree_ptr, ray_functor, vector_functor);
+          point, *tree_ptr, ray_functor, vector_functor);
```
lrineau: That call (in the return) triggers a call on the AABB_tree, that is then built lazily.

sloriot: yes but `tree_built` has nothing to do with it so it was useless. And the previous patch is not needed IMO. When I'm saying lazy construction, I also mean lazy allocation/creation of the AABB-tree. But it requires some tricks since the AABB-tree might not be built from a mesh (so it might not even be possible).

maxGimeno: Why is it faster like this then ?

sloriot: I don't see any use of the value of `tree_built` that is meaningful. Disabling the check using the bbox is not going to speed anything significantly.

maxGimeno: So it's better to perform systematically one to six tests than to perform one ?

sloriot:
The other test is a ray shooting in an AABB-tree so even if it's only on the root node, I'm pretty confident it will  be less expensive.

lrineau:
Travis issue. I remove this PR from `integration`.

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -175,11 +177,12 @@ class Side_of_triangle_mesh
   , tree_ptr(&tree)
   , own_tree(false)
   {
+    box = tree.bbox();
   }
 
   ~Side_of_triangle_mesh()
   {
-    if (own_tree)
+    if (own_tree && tree_ptr!=nullptr)
```
lrineau: For the record, `delete nullptr;` is a no-op, and that is documented in the C++ norm.

maxGimeno:

Issue #2196, opened 2017-06-21
IO for Polygon Soups


sgiraudot:
## Issue Details

I'm opening this issue just to start a discussion on the input/output functions for polygons soups, especially:
- the current OFF reader
- a PLY reader to develop

So far, we already have a function that can interpret OFF input as a polygon soup, namely `read_OFF()`. But this function is in the `Polyhedron_IO` package (although it never actually refers to a Polyhedron), and it is not documented. Note that it is used in [the example for polygon soup orientation](http://doc.cgal.org/latest/Polygon_mesh_processing/index.html#PolygonSoupExample), which makes me think it _should_ be documented.

For PLY, we already have a PLY reader (with a new version in [pull request](https://github.com/CGAL/cgal/pull/2138) along with LAS) but for points only. There is not that much of work to do if we want to handle face graphs and polygon soups in addition (basically, write something to handle the PLY lists of indices).

### Questions

1. Do you think it would make sense to move `read_OFF()` into Polygon Mesh Processing? There could be an IO section in there (and polygon soups are used there, for example for orientation)

2. In that case, do we also want to add `read_PLY()`? I'm not really fond of these names, it seems to me that they should be more precise, for example `read_off_polygon_soup()` and `read_ply_polygon_soup()` (similarly to what we have with `read_off_points()` and `read_ply_points()`).

3. For `Point_set_3`, I made an `operator>>()` that automatically calls the right reader (between `XYZ`, `OFF`, `PLY` and soon `LAS`) by analysing the first line of the stream (that contains a tag, except for `XYZ`). Do we want the same thing for `Surface_mesh`, for example? It could select between `OFF` and `PLY`. And as it works similarly to `Point_set_3`, we could similarly recover additional properties in `PLY` in automatically generated property maps.

gdamiand:
1. I agree that this function could be move into Polygon Mesh Processing and documented.

2. I also agree that the name read_OFF is not correct (moreover it seems in conflict with the small feature add generic read and write functions). Why not define a unique function read_model_as_polygon_soup that analyse the header of the file ? (and you can do the same for read_model_as_points)

sgiraudot:
(I'm reopening it, it's only half-solved: there is still a small feature to do to document and unify what exists in non-documented code.)

afabri:
The PR #2657 is related. There as well I am looking for the right spot  where to document `read_STL()`.

lrineau:
Is that solved?

sloriot:
[...]

PR #4255, opened 2019-09-27
Re-organize I/O

maxGimeno:
## Summary of Changes
- Robustify all I/O functions
- Add missing overloads
- Uniformize calls of I/O functions to read_XXX/write_XXX, deprecate old stuff
- Regroup functions into only a few packages + remove duplicates + remove useless packages
- Add new convenience functions : `read_points`, `read_polygon_soup`, `read_polygon_mesh`, ...
- Document stuff in IO Stream

## Release Management
* Issue(s) solved (if any): 
* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Polgyon_soup_mesh_IO

Relates : #2196, #4332, 

## TODO

Mael:
- [x] Check all @todo / @fixme in BGL/IO and Stream_support/IO
- [x] Re-introduce functions that were documented in CGAL 5.0 but that have been removed, and deprecate them all
- [x] Go over read/write_xxx (small xxx) and replace with new API + deprecate all old ones
- [x] Get rid of the various local "read_mesh" functions (check for e.g. "off")
- [x] Fix documentation of operator<< >> CGAL::Color
- [x] Binary shenanigans, LAS is always binary
- [x] Put links to LAS_with_properties in `PSP3::read/write_LAS()`
- [x] Don't use `Graph` in `PMP`
- [x] partition.h and split_graph_into_polylines into BGL > Partitioning operations
- [x] Hide `CGAL_DEPRECATED` from the doc
- [x] Fix "polgyon"
- [x] Add an "extension: " in Stream support
- [x] Fix XYZ and Gocad tables in stream support
- [x] check flushing of writers
- [x] check that all readers use `reserve()` when they ought to
- [x] fix std::precision interaction with geomview_stream
- [x] fix OFF reading requiring larger kernels than before
- [x] fix broken logic of `&& !std::distance < 0`
- [x] Add note about std::binary to lower level overloads for PLY.
- [ ] Nef_3/Nef_3_problematic_construction.cpp runtime error.
- [ ] Go over the issues of the project (https://github.com/CGAL/cgal/projects/3), check if changes from related issues haven't been lost / fixed

Maxime:
- [ ] Polyhedron "save as XXX" but filename "asd.YYY" --> saves in YYY format
- [x] Can't load .obj if it's not a polygon mesh (can be a simple soup if there are multiple objects). Probably can merge most readers (like `load_off` and `load_obj` to be the same function using PMP::read_polygon_mesh()`).
- [x] Convex_hull_3/demo is broken at runtime (even in master).

major:
- [x] Fix `skip_to_next_vertex` in File_scanner_OFF that should also have has_textures() has_normals() checks
- [x] Make sure all OFF reading works (everything now uses the File_scanner_OFF rather than having 10 different implementations).).
- [x] The property maps in BGL::write_PLY use the Surface_mesh API only (it should use NP like in read_off to read them) + add overload in SM/IO like `read_off(surface_mesh)`
- [x] Extend IO that uses std::vector (SS/IO) to ranges
- [x] Don't `return in.good()` otherwise it fails if eof bit is set -> return !in.fail() instead
- [x] Make sure all facegraph IO (except overloads for a particular class) is in BGL
- [x] Check if `read_point_set()` exists and add it if not (get advice from @sgiraudot on what he thinks is best)
- [x] Introduce `CGAL::read_polygon_soup` in Stream_support/IO that dispatchs based on the file extension 
- [x] Introduce `CGAL::read_polygon_mesh` in BGL/IO that dispatchs based on the file extension (gotta mind .off, .coff etc.)
- [x] Introduce `PMP::IO::read_polygon_mesh` in PMP/IO that dispatchs based on the file extension (gotta mind .off, .coff etc.) and which call orient_PS if CGAL::read_polygon_mesh fails
- [x] Move as much code away from the demo and into the correct SS/IO (polygon soup) BGL/IO (polygon mesh) SM (specific overloads) files
- [x] Make sure all readers/writers actually work (not just that it doesn't crash but that the data read is the one that is expected)
- [x] Document IO: section in BGL pointing to SS/IO, all classes that can be read from files have an IO section and point to SS/IO, etc.
- [x] Mind CGAL_LINKED_WITH_3MF (I used this macro to hide 3MF code, but it needs to be added to CMakeLists.txt)
- [x] Make sure that all readers fail graciously, there are still many `in >> x >> y` that should be `if(in >> x >> y)` (and similar)
- [x] Fix broken links to IO::Mode
- [x] Go over the operator<< >> for FaceGraphs and make sure they are working and document what they do
- [x] Check that I properly re-applied https://github.com/CGAL/cgal/pull/4212 (manually re-applied changes to 3MF)
- [x] Check if changes from https://github.com/CGAL/cgal/pull/4501 should be applied to the OFF scanner
- [x] Check that #4674 was correctly applied
- [x] import #4727
- [x] Use `iformat`/`oformat` (see e.g. STL reader)
- [x] Fix point sets sometimes being read as `read_PLY_points` and sometimes `read_PLY`
- [x] Use binary for default input / output format when it makes sense (e.g. PLY) + document 
- [x] Fix `read/write_point_set` extension extraction (see `read_polygon_soup`)
- [x] Test all IO, all overloads, binary/ascii, all extensions (SS, BGL, PMP, SM, Polyhedron, etc.). Check the output are the correct geometry, combinatorics. Check for bad inputs. See `test_bgl_read_write.cpp`.
- [x] Fully document IO functions (most are missing template parameters / parameters documentation). Document INP?
- [x] Use the `read_point_set` / `read_polygon_soup` / `read_polygon_mesh` in examples / demos (not fully done, see TriMesh.cpp, face_selection_borders_regularization_example.cpp, etc.).

minor:
- [ ] Fix VTK reading (check the vtp/vtu consistency)
- [ ] `write_vtu`/`output_to_avizo` / `output_to_tetgen` etc. (newly documented) don't fit the standard names
- [ ] Bench flat_map vs CGAL::Inverse_index vs std::unordered_map (generic_facegraph_builder)

extra:
- [x] Also rename Point_set I/O functions?
- [ ] read_XXX_with_properties() function could not be simply renamed as read_XXX() because then, the overload with properties and the one with NP conflict. We could probably find a way to deal with it at some point.
- [ ] Go over Triangulations IO (mind the IO from tetrahedral_remeshing)
- [ ] np::point_map in polygon soup readers/writers
- [ ] Document NPs in SS/IO ?

extra extra:
- [ ] Factorize VTU code (mesh_2/mesh_3) and move it into IO
- [ ] Go over C3T3 IO
- [ ] Go over LCC IO
- [ ] Create GOCAD file writer in SS and use it in BGL
- [ ] Kernel objects IO
- [ ] Add << and >> operators for LCC and BGL Facegraphs.


MaelRL:
TODO:

major:
- [ ] Fix VTK reading
- [ ] Make sure that all readers fail graciously, there are still many "in >> x >> y" that should be if(in >> x >> y)
- [ ] Introduce CGAL::read_polygon_soup in Stream_support/IO that dispatchs based on the file extension (gotta mind .off, .coff etc.)
- [ ] Introduce CGAL::read_polygon_mesh in BGL/IO that dispatchs based on the file extension (gotta mind .off, .coff etc.)
- [ ] Introduce PMP::IO::read_polygon_mesh in PMP/IO that dispatchs based on the file extension (gotta mind .off, .coff etc.) and which call orient_PS if CGAL::read_polygon_mesh fails
- [ ] Use the functions above in examples / demos
- [ ] Fix "skip_to_next_vertex" in File_scanner_OFF that should also have has_textures() has_normals_check()
- [ ] Test all overloads
- [ ] Make sure all readers/writers actually work
- [ ] Re-introduce functions documented in CGAL 5.0 and deprecate them all
- [ ] Fix documentation of operator<< >> CGAL::Color
- [ ] Document IO
- [ ] Fix broken links to IO::Mode
- [ ] Go over the issues of the project (https://github.com/CGAL/cgal/projects/3)
- [ ] Go over the operator<< >> for FaceGraphs and make sure they are working and document what they do

minor:
- [ ] Bench flat_map vs CGAL::Inverse_index vs std::unordered_map
- [ ] Property maps in write_PLY seem based on Surface_mesh only

extra:
- [ ] Move as much code away from the demo and into the correct SS/IO BGL/IO files
- [ ] Go over C3T3 IO
- [ ] Go over LCC IO
- [ ] Create GOCAD file writer in SS and use it in BGL
- [ ] Factorize VTU code (mesh_2/mesh_3) and move it into IO

afabri:
@MaelRL  I don't see why `is >> x >> y` is not correct and needs parenthesis. First I never saw that, and second in case the fail bit is set as one cannot read `x` it will stay set when we come back from `>> y`.

MaelRL:
Godspeed @maxGimeno !

maxGimeno:
Thx a lot !

Linear_cell_complex/include/CGAL/Linear_cell_complex_for_combinatorial_map.h
```diff
@@ -205,7 +205,7 @@ namespace CGAL {
         import_from_halfedge_graph(heg, default_point_converter,
                                    origin_to_copy, copy_to_origin);
       }
-
+      using Base::clear;
```
maxGimeno, line 208: @gdamiand Is this OK ? My intention is to fix the helper using `Has_member_clear<>` that doesn't detect LCC::clear() in clear(FaceGraph). 
With this it does, and I think it's OK because the General_map concept contains a clear() function, and a LCC_for_combinatorial_maps refines the concept.

gdamiand: @maxGimeno Yes this is ok ! Thanks.

BGL/include/CGAL/boost/graph/IO/STL.h
```diff
@@ -238,13 +243,13 @@ bool write_STL(std::ostream& os,
...
-      os << "facet normal " << n << "\nouter loop\n";
-      os << "vertex " << p << "\n";
-      os << "vertex " << q << "\n";
-      os << "vertex " << r << "\n";
-      os << "endloop\nendfacet\n";
+      os << "facet normal " << n << "\nouter loop"<<std::endl;;
+      os << "vertex " << p <<std::endl;
+      os << "vertex " << q <<std::endl;
+      os << "vertex " << r <<std::endl;
+      os << "endloop\nendfacet"<<std::endl;
```
lrineau, line 245: I wonder why you insist on turning `\n` into `std::endl`. You should read [the documentation of `std::endl`](https://en.cppreference.com/w/cpp/io/manip/endl).

maxGimeno: With the \n, after around 14K lines, the stream seems to be full or something, and the end of the file is not written. endl flushes the stream, that's what I was going for. And indeed, I don't have the writing bug after I turned the \n into endl. But if you really prefer \n, I can absolutely only keep the last endl and turned back to \n for the rest.

lrineau: You could keep one `std::endl` (the last one) for each iteration in the loop. That will already save a lot of `std::flush`.

MaelRL:
@lrineau Could you please check 1732010 ?

sgiraudot:
[This version of `read_PLY()`](https://cgal.geometryfactory.com/~mrouxell/IO/V3/Point_set_processing_3/group__PkgPointSetProcessing3IOPly.html#ga587cc261f094e7dcc3d4fbc0d58a72e1) adds a note that warns users they should use `std::ios::binary` if binary is used. It's very good, but it should be done for all functions that can/must use binary (the other `read_PLY()` variant, `read_PLY_with_properties()`, all variants of `write_PLY()`, and in other packages as well, etc.).

MaelRL:
The other variant uses a filename so it doesn't make sense to add the warning. (Internally I do use the flag when creating the stream if the binary NP is chosen.)

Forgot to mirror it on other functions, indeed, thanks.

maxGimeno:
@janetournois @afabri The test that fails is `BGL/test/BGL/test_bgl_read_write.cpp`, while trying to read the file `mesh_with_colors.off`. The 2nd face reading is complete, but the next `read_facet()` in the leading `skip_comment_OFF`.

maxGimeno:
Trying to read ascii files in binary mode works for some, fails for some, in BGL functions. I don't know what to do with that information.
@afabri @maelRL

maxGimeno:
Ready to integrate after https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-40.shtml
(trivial warnings fixed by c06dde1)

lrineau:
@MaelRL I have [updated the `CHANGES.md` file](https://github.com/CGAL/cgal/blob/dffa89b2cd21bf0b6c4fe4ae213812c9320591ae/Installation/CHANGES.md), but two links are missing.

MaelRL:
It should be:
`
A comprehensive list of the supported file formats is available in the Stream_support package [here](https://doc.cgal.org/5.3/Stream_support/index.html#IOstreamSupportedFormats); inversely, the following [page](https://doc.cgal.org/5.3/Stream_support/IOStreamSupportedFileFormats.html) can be used to find out which CGAL data structures can be used given a specific file format.
`

lrineau:
> It should be:
> `A comprehensive list of the supported file formats is available in the Stream_support package [here](https://doc.cgal.org/5.3/Stream_support/index.html#IOstreamSupportedFormats); inversely, the following [page](https://doc.cgal.org/5.3/Stream_support/IOStreamSupportedFileFormats.html) can be used to find out which CGAL data structures can be used given a specific file format.`

Done in 287836254f9.

lrineau:
> > It should be:
> > `A comprehensive list of the supported file formats is available in the Stream_support package [here](https://doc.cgal.org/5.3/Stream_support/index.html#IOstreamSupportedFormats); inversely, the following [page](https://doc.cgal.org/5.3/Stream_support/IOStreamSupportedFileFormats.html) can be used to find out which CGAL data structures can be used given a specific file format.`
> 
> Done in [2878362](https://github.com/CGAL/cgal/commit/287836254f9541aa2b8faedbe08a7304d6097b7f).

But those links will only become valid on Sunday night, during the testsuite of `master`.

lrineau:
> > > It should be:
> > > `A comprehensive list of the supported file formats is available in the Stream_support package [here](https://doc.cgal.org/5.3/Stream_support/index.html#IOstreamSupportedFormats); inversely, the following [page](https://doc.cgal.org/5.3/Stream_support/IOStreamSupportedFileFormats.html) can be used to find out which CGAL data structures can be used given a specific file format.`
> > 
> > 
> > Done in [2878362](https://github.com/CGAL/cgal/commit/287836254f9541aa2b8faedbe08a7304d6097b7f).
> 
> But those links will only become valid on Sunday night, during the testsuite of `master`.

I have verified: the links work as expected. Thanks.

PR #4256, opened 2019-09-27
Fix CGAL_TOOLSET for latest visual compilers

### LINKED ISSUE #4206: Compiler detection fails for MSVC version > 14.0
## Issue Details

When building using the included CMAKE scripts, the resulting library filenames contain an incorrect vc version number. 
f.e. CGAL_[Core|ImageIO]-**vc140**-[mt]-[gd]-1.14.dll where vc142 would be expected.

## Environment

* Operating system: Windows:
* Compiler: MSVC 14.1, MSVC 14.2
* Release or debug mode:
* CGAL version:
* Boost version:

maxGimeno:
## Summary of Changes
Replace vc150 by vc141 for 2017 and add vc142 for 2019

## Release Management

* Issue(s) solved (if any): fix #4206


Installation/cmake/modules/CGAL_GeneratorSpecificSettings.cmake
```diff
@@ -8,8 +8,11 @@ if ( NOT CGAL_GENERATOR_SPECIFIC_SETTINGS_FILE_INCLUDED )
     set(CGAL_AUTO_LINK_ENABLED TRUE)
   endif()
 
-  if     ( MSVC15 )
-    set(CGAL_TOOLSET "vc150")
+  if     ( MSVC16 )
+    set(CGAL_TOOLSET "vc142")
+    message( STATUS "Using VC16 compiler." )
+  elseif ( MSVC15 )
```
lrineau: `MSVC15` and `MSVC16` [do not exist], and never will. [`MSVC14`] and previous numbers have been deprecated. [`MSVC_VERSION`] (support since CMake-3.0 at least) is supposed to be used instead.

Since CMake 3.12, [`MSVC_TOOLSET_VERSION`] can be used directly to get the toolset. I think we could documented that recent MSVC versions require recent CMake versions. That is already true:
  - MSVC 2017 is supported since CMake 3.8,
  - MSVC 2019 is supported since CMake 3.14 only.

[`MSVC_TOOLSET_VERSION`]: https://cmake.org/cmake/help/v3.12/variable/MSVC_TOOLSET_VERSION.html

[do not exist]: https://cmake.org/cmake/help/v3.15/manual/cmake-variables.7.html#variables-that-describe-the-system
[`MSVC14`]: https://cmake.org/cmake/help/v3.15/variable/MSVC14.html
[`MSVC_VERSION`]: https://cmake.org/cmake/help/v3.15/variable/MSVC_VERSION.html

maxGimeno: It should be good now

lrineau: You can remove the test of MSVC15, because that variable does not exist.

Installation/cmake/modules/CGAL_GeneratorSpecificSettings.cmake
```diff
@@ -8,9 +8,9 @@ if ( NOT CGAL_GENERATOR_SPECIFIC_SETTINGS_FILE_INCLUDED )
     set(CGAL_AUTO_LINK_ENABLED TRUE)
   endif()
 
-  if     ( MSVC15 )
-    set(CGAL_TOOLSET "vc150")
-    message( STATUS "Using VC15 compiler." )
+  if ( MSVC_TOOLSET_VERSION )
+    set(CGAL_TOOLSET "vc${MSVC_TOOLSET_VERSION}")
+    message( STATUS "Using Visual toolset." )
```
lrineau:
```suggestion
    message( STATUS "Using VC toolset ${MSVC_TOOLSET_VERSION}." )
```

PR #4257, opened 2019-09-28
Replace vertex/cell member function calls in TDS_3

afabri:
## Summary of Changes

Replace member function calls of vertex and cell with member functions of TDS_3

## Release Management

* Affected package(s):  TDS_3


sloriot:
@afabri  superseded by #8909 ?

PR #4259, opened 2019-09-30
PMP: Improvements for snapping

MaelRL:
## Summary of Changes

Some improvements for mesh snapping:
- [x] Non conformal snapping should not modify the geometry of the target mesh
- [x] Give an option to do non-conformal snapping with the opposite behavior (target projects to the source) and add an overload that does "constrained symmetric snapping" (source --> target and then target --> source with  only the source being modified). Will need to re-use code removed in e37d708.
- [x] Fix foldings (forbid stitchables edge? Match polylines?)
- [x] Re-add previously existing cases and check behavior
- [x] Box_with_info_d and parallelize vertex-vertex snapping
- [x] Bench
- [x] Check if everything has been ported here correctly
- [x] Fix the cases where the range is not the full border
- [x] Fix tests and look into border
- [x] Add proper data files
- [x] Compute_squared_distance >> predicates

## Release Management

* Affected package(s): `PMP`


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Snapping/snap.h
```diff
@@ -639,7 +639,7 @@ std::size_t split_edges(EdgesToSplitContainer& edges_to_split,
 
     // Inserting new points ordered from the source to the target of (the initial) 'h'
     bool first_split = true;
-    Point previous_split_position;
+    Point previous_split_position = Point();
```
lrineau: That is not really important, here, but what you wrote:
  - creates a temporary, using the default-constructor of `Point`,
  - and then copy-initialize `previous_split_position` with that temporary,
  - and then destroy the temporary.

If the kernel type uses `Lazy`, all those operations could be non-trivial.

What is more... the line `Point previous_split_position;` already created the object `previous_split_position` using its default-constructor... If that default-constructor does not initialize the data members, then `Point()` will create a temporary `Point` with uninitialized data-members, and then you will get a warning that uninitialized values are used.

If you really want to initialize the data members, then use the constructor from `ORIGIN`.
```suggestion
    Point previous_split_position{ORIGIN};
```

MaelRL: I had used ORIGIN initially, but there's no really any reason it should be constructible from ORIGIN?

lrineau: `CGAL::Point_3<K>` is constructible from `ORIGIN`, but in the traits class that is used, it might be that the point type is not.

Maybe the use of `boost::optional<Point>` instead of:
```c++
bool first_split = true;
Point previous_split_position;
```
could help the compiler understand that the value is never used uninitialized.

MaelRL: Used that. Thanks for the suggestions.

lrineau: Unfortunately, one can see that did not work because [the warning is still there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-84/Polygon_mesh_processing/TestReport_lrineau_Debian-stable-Release.gz).

I saw you eventually resorted to initialize the point with a real one.

maxGimeno:

PR #4265, opened 2019-10-01
Do not require -DCGAL_HEADER_ONLY

### LINKED ISSUE #4264: header-only not quite the default
## Issue Details

I downloaded the beta1 version of CGAL-5.0. In it, I ran `cmake . && make` (probably useless, I ran it just in case it might still create `compiler_config.h`). Now I am trying to use CGAL without cmake. `g++ -c file.cpp -I /path/to/include` produces
> fatal error: CGAL/compiler_config.h: No such file or directory

Apparently I need `-DCGAL_HEADER_ONLY`, which is strange for what is supposed to be the default. Shouldn't it be the other way around, people who do not want header-only need to define some macro?

## Source Code

```c++
#include <CGAL/config.h>
```

## Environment

* Compiler: g++-9.2.1
* CGAL version: 5.0-beta1

lrineau:
## Summary of Changes

Do not require `-DCGAL_HEADER_ONLY`. Now, that macro is set unless `CGAL_NOT_HEADER_ONLY` is 1.

For the moment, using CGAL without CMake at least is not tested, but at least users can try.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #4264


maxGimeno:
Some dependencies errors, plus this PR fails on not header-only platforms:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-147.shtml#AABB_tree_Demo

joachim-reichel:
The modification to CGAL/config.h should happen at the top of the file since the macro is already used just after the include guards.

PR #4270, opened 2019-10-03
PMP: Fix doc macro-master

maxGimeno:
## Summary of Changes

master version of #4269  that adds the missing macro definition to more recent files.

## Release Management

* Affected package(s):PMP


sloriot:
Actually the issue is here since several CGAL releases for some functions (the smooth one are new in 5.0). To make it simple we can keep it like that.

PR #4271, opened 2019-10-03
Set GMPXX ON if detected

### LINKED ISSUE #4266: Enable GMPXX automatically
## Issue Details

With old CGAL, I could configure with `-DWITH_GMPXX`, which defined some macro in `compiler_config.h` and enabled GMPXX for all programs using CGAL. It was already sad that I had to specify `-DWITH_GMPXX` once when configuring, since CMake could have looked for it and enabled it if available. Now with header-only, I actually need to do that every time I use CGAL. Would it be acceptable to enable GMPXX automatically if it is detected? One way would be to use `find_package(GMPXX QUIET)` in `CGAL_SetupGMP.cmake`. Another one would rely on `__has_include`, which is a bit more dangerous because it does not affect the linker.

I understand that some users might want tight control of exactly what is used, but I believe most just want CGAL to sort its implementation details itself, which means enabling GMPXX. Or the documentation on how to use CGAL should explain very clearly that users are supposed to enable everything they have available on their machine.

## Environment

* Compiler: g++-9
* CGAL version: 5.0-beta1
* Other libraries versions if used (Eigen, TBB, etc.): GMP-6.1.2

maxGimeno:
## Summary of Changes

With this PR, the default of the WITH_GMPXX option will be set to ON if GMPXX is found.

## Release Management

* Issue(s) solved (if any): fix #4266

### DIFF
--- a/Installation/CMakeLists.txt
+++ b/Installation/CMakeLists.txt
@@ -568,10 +568,6 @@ message("== Detect external libraries ==")
 #  Coin is used in KDS, but no FindCoin or FindCOIN exists
 #  There exists FindIPE, FindMKL, but they are only used to support supporting libs
 list (INSERT CGAL_SUPPORTING_3RD_PARTY_LIBRARIES 0 GMP MPFR ZLIB OpenGL LEDA MPFI RS RS3 OpenNL Eigen3 ESBTL Coin3D NTL IPE)
-if (NOT WIN32)
-  # GMPXX is not supported on WIN32 machines
-  list (INSERT CGAL_SUPPORTING_3RD_PARTY_LIBRARIES 1 GMPXX)
-endif()
 
 # Where CMake is run several times, to avoid duplicates
 list(REMOVE_DUPLICATES CGAL_SUPPORTING_3RD_PARTY_LIBRARIES)
@@ -611,8 +607,6 @@ if(CGAL_DISABLE_GMP)
   unset(WITH_GMP CACHE)
   unset(CGAL_USE_GMP)
   unset(CGAL_USE_GMP CACHE)
-  unset(WITH_GMPXX)
-  unset(WITH_GMPXX CACHE)
   unset(WITH_MPFR)
   unset(WITH_MPFR CACHE)
 
@@ -643,7 +637,6 @@ foreach (lib ${CGAL_SUPPORTING_3RD_PARTY_LIBRARIES})
     option(WITH_${lib} "Select external library ${lib}" ON)
   else()
     option(WITH_${lib} "Select external library ${lib}" OFF)
-  else() #GMPXX set in CGAL_SetupGMP.cmake
   endif()
 
 endforeach()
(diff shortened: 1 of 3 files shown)


Installation/CMakeLists.txt
```diff
@@ -644,8 +644,11 @@ foreach (lib ${CGAL_SUPPORTING_3RD_PARTY_LIBRARIES})
   list( FIND CGAL_ESSENTIAL_3RD_PARTY_LIBRARIES "${lib}" POSITION )
   if ( "${POSITION}" STRGREATER "-1" ) # if lib is essential
    option(WITH_${lib} "Select external library ${lib}" ON)
-  else()
-    option(WITH_${lib} "Select external library ${lib}" OFF)
+  else() #GMPXX set in CGAL_SetupGMP.cmake
+    string(COMPARE EQUAL "${lib}" "GMPXX" IS_LIB_GMPXX)
+    if(NOT ${IS_LIB_GMPXX})
```
lrineau: Ugly.
```cmake
if(lib STREQUAL "GMPXX")
```

mglisse:
Thanks. Do you agree that it is a good idea? Y'all are much more familiar with what CMake users expect.

maxGimeno:
I personally have no problem  with that.

maxGimeno:

Installation/CMakeLists.txt
```diff
@@ -643,9 +640,8 @@ foreach (lib ${CGAL_SUPPORTING_3RD_PARTY_LIBRARIES})
   # add option
   list( FIND CGAL_ESSENTIAL_3RD_PARTY_LIBRARIES "${lib}" POSITION )
   if ( "${POSITION}" STRGREATER "-1" ) # if lib is essential
-   option(WITH_${lib} "Select external library ${lib}" ON)
-  else()
-    option(WITH_${lib} "Select external library ${lib}" OFF)
+
+  else() #GMPXX set in CGAL_SetupGMP.cmake
   endif()
```
lrineau: The patch does not look good. It has removed all the options `WITH_${lib}`. I agree that those options should eventually disappear, but I do not think that is the intent of this PR.

maxGimeno: Indeed, this is an error

maxGimeno:

maxGimeno:
With GMPXX ON, there is an [error on MacOS in the intersection of Line_3_triangle_3](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-53/Intersections_3/TestReport_cgaltester_x86-64_Darwin-18.2_Apple-clang-default_Release.gz).

Intersections_3/test/Intersections_3/call_test.cpp
```diff
@@ -264,10 +266,6 @@ int main()
     call_do_intersect_with_kernel(Bbox_3(), S(), K());
     call_do_intersect_with_kernel(Bbox_3(), Sph(), K());
     call_do_intersect_with_kernel(Bbox_3(), Tr(), K());
-    // There is no kernel to call
-    // call_do_intersect_with_kernel(Bbox_3(), Bbox_3(), K());
-  } catch(...) {
-    // as long as this test compiles, it is fine
-  }
+}
```
lrineau:
```suggestion
  }
```

maxGimeno:
I'm quite sure this PR is responsible for [the errors on 32bit](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-1/Surface_mesh_approximation/TestReport_lrineau_Fedora-32-Release.gz)

maxGimeno:
Actually it wasn't.

lrineau:

PR #4277, opened 2019-10-08
PMP: Replace `Location_traits` by a simpler API

MaelRL:
## Summary of Changes

Users have rightfully pointed out that the API used in PMP Location function was not practical, sometimes requiring the user to pass the type of the named parameters, which is very unpractical.

Instead, template aliases for most types, and template by the field type (FT) when necessary, which is much easier to write for the user. Since template aliases cannot automatically deduce template parameters, a non-Doxygen version of a function is written without them from time to time.

updated doc: https://cgal.geometryfactory.com/~mrouxell/PMP_locate_v3/Polygon_mesh_processing/group__PMP__locate__grp.html

This PR also adds an example for the PMP Location functions.

## Release Management

* Affected package(s): `Polygon_mesh_processing `


sloriot:
This PR targets the 5.0 release because the feature is new in this release.

lrineau:
locate was introduced by https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/PMP-Locate. @janetournois, can you please have a look at the new API, and approve it again?

Polygon_mesh_processing/examples/Polygon_mesh_processing/locate_example.cpp
```diff
@@ -0,0 +1,113 @@
...
+typedef K::Point_2                                                      Point_2;
+typedef K::Ray_2                                                        Ray_2;
+typedef K::Point_3                                                      Point_3;
+typedef K::Ray_3                                                        Ray_3;
+
+typedef CGAL::Surface_mesh<Point_3>                                     Mesh;
+typedef typename boost::graph_traits<Mesh>::vertex_descriptor           vertex_descriptor;
+typedef typename boost::graph_traits<Mesh>::face_descriptor             face_descriptor;
+
+namespace CP = CGAL::parameters;
```
janetournois: I don't mind too much, but it seems to me that everywhere else (at least in this package) we call this namespace `NP` for "Named Parameters".
Is there a reason to change that?

janetournois:
I re-approve the new documentation (and wrote it in the small feature).

There is only one detail I already mentioned in the my first review is : for template parameters descriptions, I would stick to "a model of" everywhere, and avoid mixing with "must be a model of", which makes the reader wonder why this one is different, imho.

lrineau:
Travis still find errors:
https://travis-ci.org/CGAL/cgal/jobs/598094517#L7619

lrineau:
Actually, I do not understand why Travis sees new dependencies in the Property_maps package.

MaelRL:
Yes, I couldn't understand either, so I thought it might have been an issue present in master when I forked (doesn't ring a bell, though), and ignored it. I can try updating it with master.

maxGimeno:

PR #4232, opened 2019-09-19
SMS: Add Garland-Heckbert simplification

MaelRL:
## Summary of Changes

This PR adds a new duo of cost and placement oracles to the Surface Mesh Simplification package, based upon the paper of Garland && Heckbert and their concept of quadric errors. Initial work by @baskinburak, during the 2019 Season of the Google Summer of Code (https://cgal.geometryfactory.com/CGAL/Members/wiki/GSoC2019/Simplification).

In addition, this PR:
- Completely cleans the package (no more Camel case, incomprehensible variable names, weird files, useless `_impl.h`, subdirectory /Details/, wonky indentation, needlessly complicated assertion system, etc.).
- Upgrade to C++11 / BGL functions.
- Properly (un)document everything.
- Fixes the concept and models: `EdgeProfile` is not a real concept, it is hard coded in the class `Edge_collapse`.
- Fixes the absence of a traits class, and thus the fact you couldn't actually use a custom VPM.

No breaking changes other than fixing the documentation to reflect the code.

Not done: fixing LT_core to properly use Eigen::Matrix and geom traits.

TODO:
- [x] Clean branch
- [x] Check doc
- [x] fix "It returns an optional value, which can be absent, in which case the remaining vertex is kept in its place."
- [x] Small feature
- [x] Check that the Eigen requirement is clear
- [x] read full diff
- [x] Benchmarks / Vtune
- [ ] news entry (?)

## Release Management

* Affected package(s): `Surface_mesh_simplification`
* Issue(s) solved (if any): Fixes #3835
* Feature/Small Feature (if any): [GH_Simplification](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/GH_Simplification) -- **pre-approval date: 2019/10/21**

PR #4279, opened 2019-10-09
Surface_mesh: Fix Property_container::swap()

afabri:
## Summary of Changes

The swap function swapped the content, but not the variable storing the size of the content.
The fix is not worth backporting, as the function  is not documented publicly and only used in a future PR.

## Release Management

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
@@ -486,6 +486,7 @@ public:
     void swap (Property_container& other)
     {
       this->parrays_.swap (other.parrays_);
+      std::swap(this->size_, other.size_);
     }
   
 private:


lrineau:
This PR looks like a fix. Why is it only for `master`? Was the bug introduced in `master` recently? Otherwise please rebase the PR onto `releases/CGAL-4.14-branch`.

afabri:
The fix is not worth backporting, as the function is not documented publicly and only used in a future PR.

maxGimeno:
I strongly suspect this PR is responsable for those warnings : 
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-11.shtml#Surface_mesh_simplification_Examples

afabri:
Wrong suspicion.  It is the `/arch:AVX2`  which triggers the warning, when you compile in the branch.
@lrineau why do we have this switch in the testsuite (independently from adressing the warning) ?

maxGimeno:
`...when you compile in the branch.`
How is the branch not responsible for warnings that don't occur in other branches ?

afabri:
The warning comes from the branch SMS-Add_GH_simplification-GF  not this one.  It is probably about a  `std::vector<Eigen::Matrix<..>>`.

maxGimeno:
Ok

afabri:
I have isolated the problem to this [gist](https://gist.github.com/afabri/a6df1813b8d13fabe1e7d1b889d84180).   When compiled with VC++ 2019 and   `/O2 /W3  /arch:AVX2` I get the warning:

> warning C4316: 'Derived<Matrix>': object allocated on the heap may not be aligned 32

This stackoverview [page](https://stackoverflow.com/questions/20104815/warning-c4316-object-allocated-on-the-heap-may-not-be-aligned-16) might explain it. 
Do we have to do something in CGAL or should it be fixed in Eigen. @ggael

lrineau:
[**@afabri**](https://github.com/afabri) commented on [Nov 22, 2019, 8:39 AM GMT+1](https://github.com/CGAL/cgal/pull/4279#issuecomment-557426099 "2019-11-22T07:39:35Z - Replied by Github Reply Comments"):
> Wrong suspicion.  It is the `/arch:AVX2`  which triggers the warning, when you compile in the branch.
> @lrineau why do we have this switch in the testsuite (independently from adressing the warning) ?

We want to test CGAL with AVX2, on Windows. The file `<CGAL/Interval_nt.h>` behaves differently with SSE2, SSE4, AVX, etc., and that is why we have activated those features in a few test platforms.

ggael:
Sadly, we cannot do anything about that on Eigen's side. In your case you have two different issues:

1. Heap allocation through `operator new`: [uptodate doc](https://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html)
2. Heap allocation through `std::vector`: [uptodate doc](https://eigen.tuxfamily.org/dox-devel/group__TopicStlContainers.html)

Those issues are not limited to MSVC.

You have three solutions:

1. All those issue have been fixed by c++17, so if you compile with `/std:c++17`  or `-std=c++17` with a recent compiler (GCC>=7, clang>=5, MSVC>=19.12) you're all good!

2. Use an unaligned type for storage within a `Property_array`:  `Eigen::Matrix<double, 4, 4, Eigen::DontAlign>`

3. Add `EIGEN_MAKE_ALIGNED_OPERATOR_NEW_IF( some_compile_time_condition_on_value_type_to_enable_it_for_Eigen_type_only)`  **and** use an aligned allocator for the nested `std::vector` like `boost::alignment::aligned_allocator` or `Eigen::aligned_allocator`.

Solution 2 is likely the most general and less intrusive one.

afabri:
Thank you @ggael.    What is the penalty for using the second solution?

lrineau:
It seems we are discussing in this PR #4279, but the issue in in PR #4232 because 

We could say that the PR https://github.com/CGAL/cgal/pull/4232 requires C++17.

ggael:
> Thank you @ggael. What is the penalty for using the second solution?

Solution 2 will imply unaligned vector loads instead of aligned ones. That's all. In many cases, the performance drop is not even measurable.

maxGimeno:

PR #4282, opened 2019-10-10
Fix doc style issue

### LINKED ISSUE #2095: Bug in doc in blocks with only template functions
## Issue Details

I noticed a strange bug in the CGAL documentation: most blocks of function headers are justified (the background takes all the width of the page) but some are just aligned to the left (the background stops after the text). For example, [see in Surface Mesh](http://doc.cgal.org/latest/Surface_mesh/classCGAL_1_1Surface__mesh.html), the bug happens for the section _Related Functions_.

I've made some tests, and it seems that this happens for sections that _only_ contain template functions or methods (I have checked and this is not only a bug with related functions). As soon as you add a function (or method) that is not template to the section, then the bug disappears (or similarly, go to other sections and remove all functions/methods that are not template, the bug appears).

I'm not sure if this is a bug of Doxygen or of our own usage of Doxygen (I did not find an issue related to this in the Doxygen tracker, although I'm not sure what keywords to use to describe it).

## Environment

* CGAL version: latest (official documentation)

sloriot:
Fixes #2095

@albert-github running with a recent doxygen (9a23681bc4322a1912fc98f683bcf3e49a85f47b) with option -w to generate the stylesheet I see the issue is still here. Would you have time to fix it or do you prefer I open a PR?

### DIFF
--- a/Documentation/doc/resources/1.8.13/stylesheet.css
+++ b/Documentation/doc/resources/1.8.13/stylesheet.css
@@ -498,7 +498,7 @@ table.memberdecls {
         white-space: nowrap;
 }
 
-.memItemRight {
+.memItemRight, .memTemplItemRight {
 	width: 100%;
 }
 
--- a/Documentation/doc/resources/1.8.14/stylesheet.css
+++ b/Documentation/doc/resources/1.8.14/stylesheet.css
@@ -498,7 +498,7 @@ table.memberdecls {
         white-space: nowrap;
 }
 
-.memItemRight {
+.memItemRight, .memTemplItemRight {
 	width: 100%;
 }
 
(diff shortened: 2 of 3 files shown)


albert-github:
I had to see where the `stylesheet.css` was located that was used (not on the `Surface_mesh` was used but the one from the `Manual` directory; so less trivial than the label indicates  ;-) ).

I see the possible improvement now. I'll create a pul request on doxygen for it (normally I would prefer an issue, but I use this issue for it).

albert-github:
I've just pushed a proposed patch, pull request https://github.com/doxygen/doxygen/pull/7311

sloriot:
Thanks for the PR. Sorry about the css location, I should have mentioned that it is the one in `Manual` that is used.

sloriot:
#4282 should fix it.

albert-github:
The doxygen pull request https://github.com/doxygen/doxygen/pull/7311 has been integrated into the doxygen master.

PR #4096, opened 2019-07-18
Kernel 23: remove epick without intervals

MaelRL:
## Summary of Changes

`EPICK_without_intervals` actually inherits intervals, defeating its purpose. It might be useful and can be fixed e.g. by changing the way the static filter wrapper works (see https://github.com/CGAL/cgal/pull/3939); however, the release is nigh, and so this PR removes it.

## Release Management

* Affected package(s): `Kernel_23`
* Issue(s) solved (if any): -
* License and copyright ownership: no change

PR #4306, opened 2019-10-21
Re-add Epick_without_intervals

lrineau:
## Summary of Changes

New attempt for `Epick_without_intervals`. First attempt was #3939, reverted by commit bb640d2883be715e8e5727333ad54ef6b1fc6808 in #4096.

This time, I made it with the right kernel base, I think.

## Release Management

* Affected package(s): all

### DIFF
--- a/Convex_hull_3/test/Convex_hull_3/test_extreme_points.cpp
+++ b/Convex_hull_3/test/Convex_hull_3/test_extreme_points.cpp
@@ -14,7 +14,7 @@
 #include <iostream>
 
 
-typedef CGAL::Exact_predicates_inexact_constructions_kernel K;
+typedef CGAL::Epick_without_intervals K;
 typedef CGAL::Polyhedron_3<K> Polyhedron_3;
 typedef K::Point_3 Point_3;
 
--- a/Convex_hull_3/test/Convex_hull_3/test_halfspace_intersections.cpp
+++ b/Convex_hull_3/test/Convex_hull_3/test_halfspace_intersections.cpp
@@ -81,4 +81,6 @@ int main()
   test<Epec,CGAL::Polyhedron_3<Epec> >();
   test<Epic,CGAL::Surface_mesh<Epic::Point_3> >();
   test<Epec,CGAL::Surface_mesh<Epec::Point_3> >();
+  test<CGAL::Epick_without_intervals,
+       CGAL::Surface_mesh<CGAL::Epick_without_intervals::Point_3> >();
 }
--- a/Kernel_23/include/CGAL/Exact_predicates_inexact_constructions_kernel.h
+++ b/Kernel_23/include/CGAL/Exact_predicates_inexact_constructions_kernel.h
@@ -45,7 +45,29 @@ class Epick
 #endif
 {};
 
-typedef Epick Exact_predicates_inexact_constructions_kernel;
+template < typename CK >
+struct Static_filters_without_intervals_base
+  : public internal::Static_filters< CK >
+{
+    template < typename Kernel2 >
+    struct Base {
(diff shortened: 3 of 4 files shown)


mglisse:
Is the 'E' for "exact" true? It isn't clear where bignums appear in here.

lrineau:
You are right. I messed up again.

mglisse:
Mael had suggestions in his last comment for #3939.

MaelRL:
https://github.com/MaelRL/cgal/commit/88ee987eeda18b0c62a164900320c5d197aceba0 is a proof of concept of what I meant. I only modified `Orientation_3`.

Running 'Kernel_23/test/Kernel_23/Cartesian' gives:
```
typeid k: N4CGAL5EpickE
static call in orientation_3
Calling base of static...
Calling a filtered predicate
Could not resolve with filtered predicate, calling base of filtered
typeid kiwi: N4CGAL23Epick_without_intervalsE
static call in orientation_3
Calling base of static...
```

mglisse:
> [MaelRL@88ee987](https://github.com/MaelRL/cgal/commit/88ee987eeda18b0c62a164900320c5d197aceba0) is a proof of concept of what I meant.

Makes sense, but you will need a (trivial) functor for all remaining predicates to replace them with an exact version, not just a fallback for those that have a static filter. An alternative would be to decompose it into 2 wrappers: one that replaces all predicates with their exact counterpart, and one that interposes static filters on known predicates, but this split is probably not necessary.

mglisse:
One simple way to get something like this is to modify Filtered_predicate and put everything but the last line of the body between `#ifndef CGAL_NO_INTERVALS` and `#endif`. Then you can just use Epick. (ok, probably this needs to be done in a few more places as well)

MaelRL:
PR #4495 would also benefit from the same mechanism of `Static_filters<CK>` not automatically wrapping with `Filtered_kernel_base`. For now, we have added (commit https://github.com/CGAL/cgal/pull/4495/commits/46876c8395ba13535126a7dfd4fda848b7f0b522) another class, but if changes from above were to be used, it can be simply `Static_filters<CK>`.

@mglisse Since `CGAL::Filtered_predicate` is its own class, I don't know if we can handle everything with macros. Maybe just a template parameter to `Filtered_kernel_base`? But then wouldn't it look a bit weird with `CGAL::Filtered_kernel_base<CK, DONT_FILTER>`?

mglisse:
> @mglisse Since `CGAL::Filtered_predicate` is its own class, I don't know if we can handle everything with macros.

The idea of a macro was first as a quick proof of concept, and second that the only point of Epick_without_interval (for users) is as a replacement for Epick on platforms where we cannot use intervals (no rounded operations), so it might as well be called Epick, controlled by some macros.

> Maybe just a template parameter to `Filtered_kernel_base`? But then wouldn't it look a bit weird with `CGAL::Filtered_kernel_base<CK, DONT_FILTER>`?

You can call it `CGAL::Filtered_kernel_base<CK, NO_INTERVAL>` if that looks less weird. In software pipelines, I have seen a few times functions that take an argument enable=true/false, so that the pipeline is structurally constant, but the behavior can be controlled. So I don't find it shocking anymore...

(I haven't looked at this since October, so I don't remember much, and I haven't looked at your other branch either, so don't take my comment as something insightful that needs to be analyzed)

maxGimeno:
Is this PR ready for testing ?

sloriot:
Travis fails

MaelRL:
The implementation is wrong. If I have time, I will enhance internal::Static_filters like I said above (I did part of it in the Filtered rational kernel), and it can be finished that way.

PR #4293, opened 2019-10-15
Basic viewer bugfix

gdamiand:
_Please use the following template to help us managing pull requests._

## Summary of Changes

Update cgal basic viewer to solve a linker error when used with nef.

## Release Management

* Affected package(s): Cgal basic viewer
* Issue(s) solved (if any): fix #4291

Issue #4291, opened 2019-10-15
draw_polyhedron.h does not compile when Nef_polyhedron_3.h is included.


ChamzasKonstantinos:
## Issue Details
A compilation error occurs when I include the Nef_polyhedron_3.h and the draw_polyhedron.h headers in the same executable. The problem seems to be a namespacing issue of the newell_single_step_3 function which exists both in polygon_mesh_to_nef_3.h and the Qt/Basic_viewer_qt.h . 

The full error I get is the following;

```
[ 92%] Building CXX object CMakeFiles/test_.dir/test.cpp.o
In file included from /usr/local/include/CGAL/Qt/Basic_viewer_qt.h:56:0,
                 from /usr/local/include/CGAL/draw_polyhedron.h:25,
                 from /home/chamzask/test/test.cpp:6:
/usr/local/include/CGAL/Buffer_for_vao.h: In function ‘CGAL::Local_vector CGAL::internal::compute_normal_of_face(const std::vector<CGAL::Point_3<CGAL::Epick> >&)’:
/usr/local/include/CGAL/Buffer_for_vao.h:68:74: error: call of overloaded ‘newell_single_step_3(const value_type&, const value_type&, CGAL::Local_vector&)’ is ambiguous
       newell_single_step_3(points[i], points[(i+1)%points.size()], normal);
                                                                          ^
/usr/local/include/CGAL/Buffer_for_vao.h:50:8: note: candidate: void CGAL::internal::newell_single_step_3(const Point&, const Point&, Vector&) [with Point = CGAL::Point_3<CGAL::Epick>; Vector = CGAL::Vector_3<CGAL::Epick>]
   void newell_single_step_3(const Point& p, const Point& q, Vector& n)
        ^~~~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h:32:0,
                 from /usr/local/include/CGAL/Nef_polyhedron_3.h:57,
                 from /home/chamzask/test/test.cpp:5:
/usr/local/include/CGAL/normal_vector_newell_3.h:40:6: note: candidate: void CGAL::newell_single_step_3(const Handle&, const Handle&, Vector&) [with Handle = CGAL::Point_3<CGAL::Epick>; Vector = CGAL::Vector_3<CGAL::Epick>]
 void newell_single_step_3( const Handle& p, const Handle& q, Vector& n )
      ^~~~~~~~~~~~~~~~~~~~
[...]

PR #4310, opened 2019-10-22
Namespaced newell function

ChamzasKonstantinos:
Hello all,

Thank you for you swift reply the previous time @gdamiand  and @lrineau .  This is related to issue #4291 and its solution proposed #4293.  The minimal example provided in #4291 does indeed compile now succefully. However  if I create a Nef polyhedron and convert it to a Polygon  polyhedron there is another namespacing issue. I made this small fix that successfully compiled for all my usecases. This  a minimal example that illustrates the other namespacing issue. 

```
#include <CGAL/IO/Polyhedron_iostream.h>

#include <CGAL/draw_polyhedron.h>

#include <fstream>
typedef CGAL::Polyhedron_3<Kernel> Polyhedron;
typedef CGAL::Nef_polyhedron_3<Kernel> Nef_polyhedron;
int main(int argc, char* argv[]) {
    Polyhedron P;
    std::ifstream in1((argc > 1) ? argv[1] : "compl_scene.off");
    in1 >> P;
    auto nP = Nef_polyhedron(P);
    CGAL::draw(P);
    return EXIT_SUCCESS;
}
```
## Summary of Changes
Simple added a namespace for the newell_single_step_3 to be used internally inside. I am not sure though if this  is the recommended  way to fix this problem. 

## Release Management
* Affected package(s):Nef

### DIFF
--- a/Nef_3/include/CGAL/normal_vector_newell_3.h
+++ b/Nef_3/include/CGAL/normal_vector_newell_3.h
@@ -35,6 +35,8 @@
 
 namespace CGAL {
 
+namespace internal_nef 
+{
 template <class Handle, class Vector>
 CGAL_MEDIUM_INLINE
 void newell_single_step_3( const Handle& p, const Handle& q, Vector& n )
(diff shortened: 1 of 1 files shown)


gdamiand:
Thanks @ChamzasKonstantinos for the patch. For me it is ok.

Another possibility: I can change the name used in the basic viewer to use a different name.

 @lrineau what do you think ?

lrineau:
I do not care. The patch is correct, as far as I know. The naming conflict has been solved by a namespace. I will merge it as it is.

PR #4317, opened 2019-10-26
Wrong interval multiplication with NaN

### LINKED ISSUE #4312: Wrong interval multiplication with NaN
## Issue Details

The following program prints
> [nan;-nan]

instead of
> [-inf;inf]

This issue seems to have been present for many years, possibly since the introduction of intervals in CGAL. `[0,0]*[1,inf]` also has a NaN.

In master, the new SSE code is also affected by this issue, but the set of inputs that cause problems is a bit different (larger?) because of NaNs appearing in intermediate computations. In particular, `[1,INFINITY]*[0,1]` gives [1;inf]. This can only happen if 0 is an extremity of one interval and ±inf of the other. I think that only multiplication is affected, although I would need to double-check division.

It would have been super convenient for interval arithmetic if IEEE-754 had defined 0*inf as ±inf when the rounding mode is towards ±inf, but they defined it as NaN for all rounding modes, so we have to handle it somehow. (Intel's maxpd's handling of NaN is also a bit strange, but I don't think any behavior of max could fix what multiplications produce).

As an experiment, I created [a branch](https://github.com/mglisse/cgal/tree/Number_types-interval_NaN-glisse) where I check the floating point environment to know if any NaN appeared. For my favorite example, the simple_2 benchmark in Triangulation_3 with structural and static filters disabled, the running time in 4.14 was 1.6s, the current master is 0.87s, and the branch increases it to 0.93s (on a nice Xeon processor). Significant, but it does not kill the benefits of SSE. Note that the branch is not ready. Setting/clearing exception flags is not done cleanly, and the non-SSE path is still as broken as in 4.14. Also, this is only one approach, others are possible (I have some ideas I want to try).

The problem was reported to me by @MaelRL .

## Source Code

```c++
#include <CGAL/Interval_nt.h>
#include <iostream>
typedef CGAL::Interval_nt<> IA;
int main(){
  std::cout << IA(0) * IA::largest() << '\n';
}
```

mglisse:
## Summary of Changes

Handle cases where interval multiplication produces NaN. For SSE, I use `min(x,infinity)` as an equivalent of `isnan(x)?infinity:x`. For non-SSE, I explicitly check for the problematic case.

This seems safer, close to the release, than the other branch where I was checking the exception flag.

## Release Management

* Issue(s) solved (if any): fix #4312

### DIFF
--- a/Number_types/include/CGAL/Interval_nt.h
+++ b/Number_types/include/CGAL/Interval_nt.h
@@ -551,13 +551,23 @@ private:
     __m128d bz = _mm_xor_pd(bb, m);                       // {bi,bs}
     bz = IA_opacify128(bz);
     __m128d c = swap_m128d (bz);                          // {bs,bi}
+
+    // The multiplications could produce some NaN, with 0 * inf. Replacing it with inf is safe.
+    // min(x,y) (the order is essential) returns its second argument when the first is NaN.
+    __m128d big = IA::largest().simd();
     __m128d x1 = _mm_mul_pd(aa,bz);                       // {-ai*bi,as*bs}
+    x1 = _mm_min_pd(x1,big); // no NaN
     __m128d x2 = _mm_mul_pd(aa,c);                        // {-ai*bs,as*bi}
+    x2 = _mm_min_pd(x2,big); // no NaN
     __m128d x3 = _mm_mul_pd(ap,bz);                       // {-as*bi,ai*bs}
+    x3 = _mm_min_pd(x3,big); // no NaN
     __m128d x4 = _mm_mul_pd(ap,c);                        // {-as*bs,ai*bi}
+    x4 = _mm_min_pd(x4,big); // no NaN
+
     __m128d y1 = _mm_max_pd(x1,x2);
     __m128d y2 = _mm_max_pd(x3,x4);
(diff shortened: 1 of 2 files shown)


mglisse:
For the SSE case, I think we can do with fewer _mm_min_pd, this patch is just the safe version while I am convincing myself that I can do better.

lrineau:
Thanks Marc for this PR. I will apply its non-SSE part to 4.14.1 as well.

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.0-Ic-176.shtml shows warnings on Fedora32, saying braces should be added after an if. I find it strange as there is no error in the indentation...

lrineau:
I'll deal with that...

lrineau:
Done in https://github.com/CGAL/cgal/pull/4317/commits/81c7f53d20f2643d42fa987a3450a42c9582265c.

lrineau:
The new testsuite for Interval_nt fails when SSE2 is not available (with `-m32`, for example).

mglisse:
> The new testsuite for Interval_nt fails when SSE2 is not available (with `-m32`, for example).

Je croyais avoir testé pourtant... Je regarde ça ce soir.

maxGimeno:

PR #4324, opened 2019-10-28
Cgal 3D Demo: Two-sided shading

### LINKED ISSUE #4320: Polyhedron demo: 2-color mode for point set & mesh
I suggest to add the following feature to the demo: a 2-material/color rendering mode for meshes or point sets, such that we can eg view each point as grey if front-facing, and green otherwise. Same for surface meshes. Colors can be combined with shading when using the ColorMaterial option: https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glColorMaterial.xml

The goal is to ease the interactive orientation of point sets & meshes.

Pierre

maxGimeno:
## Summary of Changes

Add a Viewer option to use a specific shader to make more obvious the orientation of the fragments in a shaded program.

To use it, enable/disable it from the `View` menu :
View_screen

You can edit the colors from the Edit->Preferences menu:
Pref_screen

## Release Management

* Issue(s) solved (if any): fix #4320

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -2275,6 +2275,11 @@ void MainWindow::on_actionPreferences_triggered()
     this->s_defaultPSRM = CGAL::Three::Three::modeFromName(text);
   });
   
+  connect(prefdiag.backFrontColor_pushButton, &QPushButton::clicked,
+          this, [this](){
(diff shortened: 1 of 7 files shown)


maxGimeno:
I benched with both versions (one with an one without all the if loops that can be modularised), and the difference is minim. for 2M triangles : 29fps with, 30fps without, for 6M triangles: 12 fps with, 15fps without.
I don't think it is worth the (considerable) effort of making the shaders programmable.

maxGimeno:

PR #4329, opened 2019-10-30
Set to nullptr after reset

sloriot:


STL_Extension/include/CGAL/Handle.h
```diff
@@ -71,7 +71,10 @@ class Handle
       if (PTR)
       {
         if (--PTR->count==0)
+        {
           delete PTR;
+          PTR=nullptr;
+        }
```
lrineau: What is the gain?

efifogel:
Just in case you invoke the destructor when the count is already zero.
Without it, the count will simply continue to decrease, which is also
harmless, but I think cleaner this way.
                         _/

STL_Extension/include/CGAL/Handle.h
lrineau: Actually, the next line, after the `if`, is already `PTR=0`.

sloriot: Right, I should have looked more closely, we can probably close the PR. @efifogel is it fine with you?

efifogel:
Yep
                         _/

PR #4333, opened 2019-10-31
Better IO example in Point_set_3

sgiraudot:
## Summary of Changes

@afabri noticed that `CGAL::Point_set_3` IO methods were not clear (especially the binary behavior), so this PR details a bit more the exampe.

## Release Management

* Affected package(s): Point_set_3 (example only)


maxGimeno:

PR #4335, opened 2019-10-31
Check the presence of <CGAL/Qt/*.h> header

lrineau:
# Summary of Changes
When CGAL is installed by package-managers, CGAL_Qt5 headers maybe
installed separately:
  - `libcgal-qt5-dev` on Debian,
  - `CGAL-qt5-devel` on Fedora, for example.

## Release Management


maxGimeno:

PR #4336, opened 2019-10-31
Take the version number from `public_release_name`

lrineau:
## Summary of Changes

Try to fix that issue: the version number of CGAL in the documentation of CGAL-5.0-beta2 is `5.0`, instead of `5.0-beta2`.

Also fix the download URLs.


maxGimeno:

PR #4344, opened 2019-11-06
PMP: Self_intersections_with_tree

maxGimeno:
## Summary of Changes

Use a custom traaits for aabb_tree to compute self_intersections.

Bench [here](https://gist.github.com/maxGimeno/fdea8214214b624a2fff4385d7388d81)
## Release Management

* Affected package(s):PMP


sloriot:
The timing are not favorable to this implementation because `Box_intersection_d` test box vs box instead of face vs tree in the current implementation. I guess we have to use the tree/tree intersection.

sloriot:
For archive my timing with 289ce74 

Input has 703512 faces.
Building the tree takes: 0.451887s.
New version in parallel takes: 15.6277s (0).
New version in sequential takes: 90.8133s (0).
Old version takes: 1.21053s (0).

PR #4346, opened 2019-11-06
TDS_3: Add an edge iterator that puts marks in Cell_data 

afabri:
## Summary of Changes

The edge iterator  uses a cell iterator and for each edge of the cell gravitates around the edge. In case it finds a cell with a smaller address it it not the "representative" of the edge. 

This PR proposes as alternative an edge iterator that stores in `TDS_3::Cell_data` if an edge was already seen before. If not seen one gravitates around to find the  cell with smallest address, and marks the edge in all incident cells.  

This is faster (In vtune I saw that on one data set `make_mesh_3()` passed 7% in the `operator++()` of the edge iterator while protecting edges)  and the bits of `Cell_data` were not used so far.
However, after a loop over edges one must loop over all cells to reset the mark.  Also one cannot loop over all edges in two threads.

I added new names for the iterators which are up to discussion, but before that it would be good to see if there is a degradation of performance of large 3D triangulations, as setting the old marks are now bit setting operations and not just an assignment of 0, 1, or 2.

## Release Management

* Affected package(s):

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h
+++ b/Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h
@@ -1334,9 +1334,9 @@ refine_balls()
     ++refine_balls_iteration_nb;
     restart = false;
     boost::unordered_map<Vertex_handle, FT, Hash_fct> new_sizes;
-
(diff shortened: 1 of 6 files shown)


sloriot:
@afabri  If you have an id per cell, then the iterator can store a bitset encoding the visited status. This solves the multi-thread issues and removes the need of resetting since the bitset is simply dropped.

Issue #4215, opened 2019-09-09
How to configure CGAL to use laslib


sloriot:
There is nothing in the documention. I expected setting `LASLIB_DIR` to be working but it does not for now and anyway there is nothing in the installation manual.

afabri:
Have a look at the [CMakeLists.txt](https://github.com/CGAL/cgal/blob/master/Point_set_processing_3/examples/Point_set_processing_3/CMakeLists.txt#L62) file of the Point Set Processing package.

sloriot:
That's what I did. What I'm saying is that is should be documented like the other dependencies.

sgiraudot:
What do you mean "documented like the other dependencies"? From what I see, [the LASlib paragraph in the doc](https://doc.cgal.org/latest/Manual/installation.html#thirdpartyLASlib) is pretty much the same as the other ones. Is there any place I missed where we document the CMake variables?

sloriot:
https://doc.cgal.org/latest/Manual/installation.html#installation_summary

sloriot:
I guess the same hold for OpenCV, TensorFlow (and upcoming OpenGR, libpointmatcher)

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Sep 23, 2019, 11:34 AM GMT+2](https://github.com/CGAL/cgal/issues/4215#issuecomment-534026312 "2019-09-23T09:34:57Z - Replied by Github Reply Comments"):
> https://doc.cgal.org/latest/Manual/installation.html#installation_summary

So you think those tables have a value? For the documentation of CGAL-5.0, we were considering removing all that.

sloriot:
Well, if `LibFooBar_DIR` is not supported it is hard to guess what to use. Is the official way to configure a lib to look at the `FindLibFooBar.cmake`?

PR #4356, opened 2019-11-12
Third party libs: improve documentation + introduce new CMake functions

sgiraudot:
## Summary of Changes

This PR improves how third party libraries for Point Set Processing related packages are handled:
1. Generalize the `CGAL_target_use_[lib](<target>)` introduced with TBB. This allows to be cleaner in terms of inclusion (use `target_include_directories()` instead of `include_directories()` to avoid polluting other targets with useless includes), handle the macros (like `CGAL_LINKED_WITH_LASLIB`) and linking. For Eigen, it may look less user-friendly because you now have to specify which target use Eigen, but it seems much much cleaner to me: I found out that some targets were not in the `if(EIGEN3_FOUND)` section whereas they _did_ require Eigen. It only worked before because the Eigen files were included globally.
2. Document missing libraries in the advanced section "_Summary of CGAL's Configuration Variables_" of the manual
3. Document the `CGAL_target_use_[lib](<target>)` functions in the "_Optional Third Party_" section. It's debatable, but I think these should be documented as they make the life of the users much simpler (so far, they have to look in the `CMakeLists.txt` of the examples to have an idea of what's going on).  I put the documentation in this section because it's the section where we usually redirect users when we document that a feature depends on a third party library, so it should be the most easy to find.

I'm putting the _Under discussion_ tag as I think this should be discussed a bit before being integrated (or not). The scripts look now much cleaner to me (especially if you look at the Classification example script, it was barely readable before, and now it's quite okay), but it might not be the case for everybody.

## Release Management

* Affected package(s): pretty much every package that used Eigen + all PSP related packages
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4215


sgiraudot:
As no one complained about this feature (and @lrineau and I discussed it AFK and agreed on it), I propose to test it for integration.

afabri:
I am not sure if we want to  add CMake functions.  Real user have to integrate CGAL components into their applications so it should be "raw CMake" that they can copy paste (the non-CMake scenario is yet another story).

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/CMakeLists.txt
```diff
@@ -51,7 +51,14 @@ polyhedron_demo_plugin(mesh_3_optimization_plugin Optimization_plugin
...
 
-
+# Use Eigen
+ find_package(Eigen3 3.1.0 REQUIRED) #(3.1.0 or greater)
+ 
+ if (EIGEN3_FOUND)
+   CGAL_target_use_Eigen(mesh_3_optimization_plugin)
+ else() #eigen
+   message(STATUS "The Mesh_3_optimization_plugin requires Eigen, which was not found, and will use a deprecated class to replace it. Warnings are to be expected.")
+ endif()#eigen
```
lrineau: With `REQUIRED` in the `find_package`, the `if(EIGEN3_FOUND)` is useless. Are you sure you wanted that `REQUIRED`?

maxGimeno:
@sgiraudot maybe this PR needs some updates with master, to add the missing Eigen managemen, at least in mesh smoothing

sgiraudot:
Sure, I'll take care of it.

maxGimeno:
thx !

sgiraudot:
I updated with master and run the full testsuite on my machine, but I did not find any missing Eigen management…

What I did find was that a test in `Number_types` called `eigen.cpp` was actually never compiled with Eigen and only displayed `Eigen is not configured!`. And of course, when I actually included Eigen… it did not compile well. I had to remove to tests because apparently, Eigen does not compute determinants with integer matrices. It should work now.

maxGimeno:
Thank you ! We'll see that tomorrow

maxGimeno:
[Errors](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-72.shtml#Number_types) 
plus [travis errors](https://travis-ci.org/CGAL/cgal/jobs/648447615#L3703), which are actually the same.

Polygonal_surface_reconstruction/test/Polygonal_surface_reconstruction/CMakeLists.txt
```diff
@@ -24,58 +24,38 @@ include( ${CGAL_USE_FILE} )
...
-    include_directories( BEFORE ${GLPK_INCLUDE_DIR} )
-
-    target_link_libraries( polygonal_surface_reconstruction_test PRIVATE ${GLPK_LIBRARIES} )
-
-    target_compile_definitions(polygonal_surface_reconstruction_test PRIVATE -DCGAL_USE_GLPK)
-
-    message("GLPK found and used")
-
+  if ( NOT GLPK_FOUND )
+    message( STATUS "NOTICE: This project requires either SCIP or GLPK, and will not be compiled.")
```
maxGimeno:
```suggestion
    message( STATUS "NOTICE : This project requires either SCIP or GLPK, and will not be compiled.")
```

maxGimeno:

lrineau:
> Successfully tested in [cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-88.shtml](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-88.shtml)

This PR was **not** tested in [CGAL-5.1-Ic-88.shtml](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-88.shtml)!

Number_types/test/Number_types/eigen.cpp
```diff
@@ -59,7 +59,7 @@ int main(){
   //check<CGAL::Lazy_exact_nt<CGAL::Quotient<CGAL::MP_Float> > >();
   check<CGAL::Sqrt_extension<double,double> >();
 #ifdef CGAL_USE_GMP
-  check<CGAL::Gmpz>();
+//  check<CGAL::Gmpz>();
```
maxGimeno: ???

maxGimeno:
[Errors](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-95/Number_types/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz) on MSVC 2015

Number_types/test/Number_types/eigen.cpp
sgiraudot: Eigen does not work on integer types. As this file was never tested before, this never raised an error but it does not work.

I could have removed it completely but I figured I'd just comment it (similarly to line 59) as someone might think the test is just missing and no deactivated voluntarily.

maxGimeno:
That last commit is not correct :/

sgiraudot:
It should be fixed now.

maxGimeno:

PR #4364, opened 2019-11-15
CGAL:: Add Parallel_if_available_tag

### LINKED ISSUE #1658: Add a tag Parallel_if_available?
When using parallelism with CGAL and TBB, if often find myself using the following piece of code to handle it:

```c++
#ifdef CGAL_LINKED_WITH_TBB
typedef CGAL::Parallel_tag Concurrency_tag;
#else
typedef CGAL::Sequential_tag Concurrency_tag;
#endif
```

Is it something that could be useful in CGAL? I understand that in some cases, one might want to execute some code in sequential mode even if TBB is present, but I'm pretty sure that in many cases, a user (like me) would simply want to activate parallelism as soon as TBB is available (which is what my piece of code above does) and fall back to sequential mode if it is not.

We can imagine adding a tag that does this, for example `CGAL::Parallel_if_available_tag`, `CGAL::Parallel_if_possible_tag` or `CGAL::Parallel_if_tbb_enabled_tag` (not sure about the name to choose).

afabri:
## Summary of Changes

Added  `Parallel_if_available_tag`.

## Release Management

* Affected package(s):   
* Issue(s) solved (if any): fix #1658
* Feature/Small Feature (if any):  [Patch_to_concurrency_in_CGAL](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Patch_to_concurrency_in_CGAL)  -- **Pre-approval date:** 2020/01/14


maxGimeno:
I think this is responsible for [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-13/Polygon_mesh_processing_Examples/TestReport_lrineau_Debian-stable-Release.gz), although I have no idea how this PR got into integration at that time.

afabri:
@maxGimeno can you please test in the docker with the testsuite that failed.

maxGimeno:
Ok we found out that the container used in the testsuite is not up to date, and in that old version of Debian Stable, the tbb package has an error in its configuration file that leads to this error. The PR is not to blame after all.

sgiraudot:
Just to summarize the discussion we just had with @afabri and @lrineau, this is the behavior we want to implement eventually:

|                             | `CGAL_LINKED_WITH_TBB` | `!CGAL_LINKED_WITH_TBB` |
|-----------------------------|------------------------|-------------------------|
| `Sequential_tag`            | OK                     | OK                      |
| `Parallel_tag`              | OK                     | Error                   |
| `Parallel_if_available_tag` | OK                     | OK / Runtime warning*   |

\* Warning can be deactivated using a macro such as `CGAL_DO_NOT_WARN_ME_ABOUT_SEQUENTIAL_FALLBACK`.

MaelRL:
For the `Parallel_tag` giving an error, this is already the case but it is done at the algorithm-level with something like
```
#ifndef CGAL_LINKED_WITH_TBB
if(concurrencytag == parallel_tag) --> #error "you messed up blabla"
#endif
```
I can put the `struct Parallel_tag {}` behind a `#ifdef CGAL_LINKED_WITH_TBB`, but then the error is just lot less readable (`Parallel_tag not defined` or whatever the compiler says). Is this what you had in mind? I don't see how I can give a clear error at the declaration (but only if it is then used somewhere else).

As for the warning, I think it's redundant with the idea of the tag itself.

sgiraudot:
No, what I had in mind is exactly what you quoted, which is what is done in the PSP package for the example: this behavior should just be generalized and maybe standardized (like a CGAL function `check_concurrency_tag()` which could be called by any function using concurrency).

The warning would be here to reduce the main source of problems that can arise: silent deactivation of the parallelism. Of course we can always say that it's the user's responsibility, but it's just an additional way to make sure users know what they are doing (if you had to define `CGAL_DO_NOT_WARN_ME_ABOUT_SEQUENTIAL_FALLBACK`, it becomes quite complicated to complain afterwards).

MaelRL:
If you are afraid of losing parallelism, shouldn't you be just using `CGAL::Parallel_tag` in the first place?

maxGimeno:

lrineau:
This PR has a conflict.

And `CHANGES.md` should be modified, so it can be merged into `master`.

lrineau:

sloriot:
I'm a bit disappointed that this PR did not actually implement the most useful feature that warns the user in case the parallel is passed but the sequential is actually used (because some TBB things went wrong).

MaelRL:
This is already the case for most algorithms that have parallel versions, e.g. https://github.com/CGAL/cgal/blob/master/Spatial_sorting/include/CGAL/Hilbert_sort_median_3.h#L177.

I asked in https://github.com/CGAL/cgal/pull/4364#issuecomment-562499288 if (and how) this should be done at a global level, but there no voices in favor of implementing this.

sloriot:
It's not the case in `Mesh_3`.  @maxGimeno could you please implement that ?

maxGimeno:
probably

sloriot:
Side note, maybe adding it to Triangulation_3 is sufficient.

lrineau:
The small feature planed to rename the processor macro `CGAL_LINKED_WITH_TBB` to `CGAL_TBB_IS_AVAILABLE`, but it was not implemented. Is that something we want to "fix" as well?

sgiraudot:
If we do rename it, I think the fix should be a global rework of the 3rd party macros to make them uniform. I think most of them use `CGAL_LINKED_WITH_[library]`, but of course that would be incorrect for header only libraries (eigen uses `CGAL_EIGEN3_ENABLED` if I remember correctly).

It would be nice to have one and only way to write such macros, to make it simple to generalize when we add new dependencies (and avoid surprises).

afabri:
We have to distinguish between something that is available and something that we want to use.

lrineau:
> We have to distinguish between something that is available and something that we want to use.

If I follow your thought, the name should not contain `HAS_` or `_IS_AVAILABLE`. It seems `CGAL_TBB_SUPPORT_ENABLED` fits the requirements.

lrineau:
Actually, I think just `CGAL_TBB_SUPPORT` is enough. The preprocessor code would then be:
```c++
#if CGAL_TBB_SUPPORT
//...
#else
//...
#endif
```
and same for the other third-party libraries.

sloriot:
We have:
-`CGAL_EIGEN3_ENABLED`
-`CGAL_PMP_USE_CERES_SOLVER`
-?
so you would use another naming convention?

sloriot:
Not sure if `CGAL_USE_ZLIB` and `CGAL_USE_VTK` are for internal use only.

afabri:
When I discussed it with Laurent 5min ago we thought that `CGAL_TBB_SUPPORT`  matches with the CMake `CGAL::TBB_support`  we already have.   So we introduce less new words.

sloriot:
The other should be renamed then.

lrineau:
@sloriot, https://github.com/CGAL/cgal/pull/4364#issuecomment-764625445:
> We have:
> -`CGAL_EIGEN3_ENABLED`
> -`CGAL_PMP_USE_CERES_SOLVER`
> -?
> so you would use another naming convention?

For the moment, that is only a brainstorming. All those macros are there to inform the preprocessor that an optional third-party dependency is detect and made available by the build system.

The naming for those macro, so far, is not uniform in the CGAL code base, as you pointed out. At the CMake level, a recent effort has come up with the uniform convention to enable an optional third-party: `CGAL::<lib>_support`, like `CGAL::TBB_support`, and `CGAL::Eigen_support`. We are discussing a way to have a uniform naming also for the preprocessor macros.

At https://github.com/CGAL/cgal/pull/4364#issuecomment-764573871 @afabri said:
> We have to distinguish between something that is available and something that we want to use.

I think that is the very reason why we have so many different naming conventions.


## `CGAL_USE_<something>`
Imagine a CGAL code has different implementations, one using a `std::list`, and another one a `std::unordered_set`. Both class templates are "available" in the STL of any C++11 compiler. Then I can imagine a macro name like `CGAL_MESH_3_USE_UNORDERED_SET`: that is an injunction to use that implementation instead of the other one.

## `CGAL_<lib>_ENABLED`
For Eigen3 and TBB, their activation in the build system only allow to use of C++ functions or classes using them (specializations/overloads with `Parallel_tag` for TBB, `CGAL::Eigen_diagonalize_traits` for Eigen3). So that is not an injunction to the code to use it, but just a new possibility.

## Mixed situations
In CGAL we have the preprocessor macros like `CGAL_USE_GMPXX`, so that the CGAL kernels can use GMPXX as the exact number type. That one is a mixed situation: the build system must ensure that GMPXX is available before the preprocessor can enable the corresponding code.

## Discussion of naming
I think we want to discuss here only about the naming of the macros of the second type: the ones that enable the use of specific C++ code, because a third-party library is available. In CGAL-5.2 we have documented the CMake API with the interface libraries `CGAL::<lib>_support`. We plan to document how to do the same without CMake in a future release. It would be better if the naming was uniform.

That is why Andreas and I are proposing the macro name `CGAL_<lib>_SUPPORT`. The meaning would be that the build system has set up the support for the optional library `<lib>`, and only that.

PR #4375, opened 2019-11-26
Avoid a construction in segment traits

sloriot:


efifogel:
Just to make sure that I understand: we actually save the comparison of the
2 representative points on the line.
(Cause we already have a left and right points)
Right?
                         _/

sloriot:
The old code construct the equation of the line + evaluate the equation at a given while here we do a direct computation on the input (which are segment endpoints).

efifogel:
Ah, even better!
I approved, of course.
                         _/

maxGimeno:

PR #4395, opened 2019-12-05
Stream_support: Fix reading WKT POLYGON

### LINKED ISSUE #4379: WKT  Is the data set buggy?
## Issue Details

Why is the first point repeated [here](https://github.com/CGAL/cgal/blob/master/Stream_support/examples/Stream_support/data/polygons.wkt#L3)?  When I read it into a Polygon with holes it is not simple because there is a zero length segment.

@maxGimeno please check.

afabri:
## Summary of Changes

As a file with a  `POLYGON` repeats the first point of the sequence we ended up with a zero length segment inside a `CGAL::Polygon_2`, so we have to remove the last point (or write our own WKT parser).

I fixed at the same time that we set the `failbit` of the `istream&`  in case the boost reading function throws an exception.

TODO: 
  - [x] @maxGimeno  please do the same fixes for the `MULTIPOLYGON`. 

## Release Management

* Issue(s) solved (if any): fix #4379


maxGimeno:
You forgot `typename`s

afabri:
> 
> 
> You forgot `typename`s

Please fix it at the same time as you fix the other issuer.

lrineau:
Travis error: https://travis-ci.org/CGAL/cgal/jobs/632220960#L5707

maxGimeno:

PR #4396, opened 2019-12-05
ImageIO and Polyhedron demo - Add error message when `io_image_plugin` can't save

janetournois:
## Summary of Changes

The ImageIO saving functions cannot save images as `.inr` when Zlib version is lower than 1.2.9,
even if `CGAL_USE_ZLIB` is true.

This PR introduces an error message to help the user understand what's going wrong.

It also fixes a seg fault in the demo, that happened when saving an image as `.inr` had failed.

This PR follows PR #3256

## Release Management

* Affected package(s): ImageIO, Polyhedron demo

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -2130,10 +2130,13 @@ void MainWindow::save(QString filename, QList<CGAL::Three::Scene_item*>& to_save
     }
   }
   if(!saved)
+  {
     QMessageBox::warning(this,
                          tr("Cannot save"),
                          tr("The selected object %1 was not saved. (Maybe a wrong extension ?)")
                          .arg(to_save.front()->name()));
+    to_save.pop_front();
+  }
 }
 
 void MainWindow::on_actionSaveSnapshot_triggered()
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/MainWindow.cpp
```diff
@@ -2130,10 +2130,13 @@ void MainWindow::save(QString filename, QList<CGAL::Three::Scene_item*>& to_save
     }
   }
   if(!saved)
+  {
     QMessageBox::warning(this,
                          tr("Cannot save"),
                          tr("The selected object %1 was not saved. (Maybe a wrong extension ?)")
                          .arg(to_save.front()->name()));
+    to_save.pop_front();
```
maxGimeno: Good catch !

Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Io_image_plugin.cpp
```diff
@@ -292,7 +292,8 @@ class Io_image_plugin :
 
     point_image p_im = *im_item->image()->image();
     bool ok = _writeImage(&p_im, fileinfo.filePath().toUtf8()) == 0;
-    items.pop_front();
+    if(ok)
+      items.pop_front();
```
maxGimeno: here too

maxGimeno:

PR #4401, opened 2019-12-08
Fix set_error_handler() header-only-mode issue (#4400)

### LINKED ISSUE #4400: set_error_handler() is per-translation-unit instead of global in header-only mode
## Issue Details

When building an application with CGAL 5.0 in header-only mode, `set_error_handler()` (and similar from assertions_behavior.h) sets handlers on a per-translation unit basis instead of globally. 

## Source Code

Please see example code at https://gist.github.com/fritzm/761eb27c50b4b1a4525cce08b08f7124.

This is a simple application which sets the CGAL warning handler once from each of two translation units, then invokes a warning once from each translation unit.

Expected: the most recently set handler function should be invoked from both translation units.
Observed: each translation unit invokes its own handler.

## Environment

* Operating system: macOS Catalina 10.15.1
* Compiler: Apple clang version 11.0.0
* Specific flags used (if any): --std=c++11 -O3
* CGAL version: 5.0
* Boost version: 1.71.0
* Other libraries versions if used (Eigen, TBB, etc.): eigen 3.3.7

fritzm:
## Summary of Changes

Move error handler/behaviour accessor inlines out of anonymous namespace when compiling in header-only mode.  This allows statics in the handler/behaviour accessor inlines to be globally unique in header-only mode.

## Release Management

* Issue(s) solved (if any): fix #4400

### DIFF
--- a/STL_Extension/include/CGAL/assertions_impl.h
+++ b/STL_Extension/include/CGAL/assertions_impl.h
@@ -30,8 +30,6 @@
 
 namespace CGAL {
 
-namespace {
-
 #ifdef CGAL_HEADER_ONLY
 
 inline Failure_behaviour& get_static_error_behaviour()
@@ -47,6 +45,8 @@ inline Failure_behaviour& get_static_warning_behaviour()
 
 #else // CGAL_HEADER_ONLY
 
+namespace {
+
 // behaviour variables
 // -------------------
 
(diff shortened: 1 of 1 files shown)


lrineau:
TODO after merge, for myself: I will have to cherry-pick the patch to 5.0.1.

maxGimeno:

sloriot:
Cherry-picking done.

PR #4407, opened 2019-12-11
Doxygen: More space for class names

afabri:
## Summary of Changes

Give the class name more space by modifying the CSS.

Before:
image

After:

image

### DIFF
--- a/Documentation/doc/resources/1.8.13/cgal_stylesheet.css
+++ b/Documentation/doc/resources/1.8.13/cgal_stylesheet.css
@@ -343,5 +343,5 @@ div.fragment {
 /* Make summary smaller to avoid wrapping of classes and concepts */
 div.summary
 {
-  width: 20%;
+  width: auto;
 }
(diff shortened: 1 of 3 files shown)


sloriot:
The change must be applied in the 3 resource directories.

MaelRL:
Could wonder if the `Class Template Reference` is really necessary too...

Documentation/doc/resources/1.8.13/cgal_stylesheet.css
```diff
@@ -339,3 +339,9 @@ div.fragment {
     padding: 4px;
     margin: 1em 4px 1em 4px;
 }
+
+/* Make summary smaller to avoid wrapping of classes and concepts */
+div.summary
+{
+  width: 20%;
```
lrineau: Why `20%`? With `auto` instead, it will only take the size it needs, and nothing more.

```suggestion
  width: auto;
```

Documentation/doc/resources/1.8.13/cgal_stylesheet.css
```diff
@@ -339,3 +339,9 @@ div.fragment {
...
+}
```
lrineau: This missing end-of-line might be an issue with certain tools.

maxGimeno:

PR #4408, opened 2019-12-11
Typos: splitted -> split

afabri:
## Summary of Changes

The word _splitted_ is ["non-standard or archaic"](https://www.yourdictionary.com/splitted)


afabri:
@maxGimeno I accidentally clicked on a review request.    But it might really make sense to look at the diff and confirm that it is so trivial that it can be merged without testing.

PR #4414, opened 2019-12-12
Set the precision when saving a mesh

oleg-alexandrov:
## Summary of Changes

I know that this is a trivial thing, yet, CGAL is a very high precision numerical software, and surely the default way of saving a mesh should use full precision, especially that some algorithms, eg., mesh smoothing, can have very subtle effects, which won't be recorded if a mesh is saved simply with 8 digits of precision. 

## Release Management

* Affected package(s): Mesh processing
* Issue(s) solved (if any): Fix for mesh not being saved with sufficient precision. 
* License and copyright ownership: Assigned to CGAL under any desired license

### DIFF
--- a/Polyhedron/examples/Polyhedron/polyhedron_prog_subdiv.cpp
+++ b/Polyhedron/examples/Polyhedron/polyhedron_prog_subdiv.cpp
@@ -102,6 +102,6 @@ int main(int argc, char* argv[]) {
         std::exit(1);
     }
     subdiv( P);
-    std::cout << std::setprecision(17)<<P;
+    std::cout << std::setprecision(17) << P;
     return 0;
 }
--- a/Polyhedron/examples/Polyhedron/polyhedron_prog_subdiv.cpp
+++ b/Polyhedron/examples/Polyhedron/polyhedron_prog_subdiv.cpp
@@ -102,6 +102,6 @@ int main(int argc, char* argv[]) {
         std::exit(1);
     }
     subdiv( P);
-    std::cout << P;
+    std::cout << std::setprecision(17)<<P;
     return 0;
 }
--- a/Surface_mesh/examples/Surface_mesh/sm_join.cpp
+++ b/Surface_mesh/examples/Surface_mesh/sm_join.cpp
@@ -34,5 +34,5 @@ int main(int argc, char* argv[])
   }
   
   
-  std::cout <<std::setprecision(17)<< sm1 << std::endl;
+  std::cout << std::setprecision(17)<< sm1 << std::endl;
 }
--- a/Surface_mesh/examples/Surface_mesh/sm_join.cpp
+++ b/Surface_mesh/examples/Surface_mesh/sm_join.cpp
@@ -34,5 +34,5 @@ int main(int argc, char* argv[])
   }
   
   
-  std::cout << sm1 << std::endl;
+  std::cout <<std::setprecision(17)<< sm1 << std::endl;
 }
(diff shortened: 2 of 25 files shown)


lrineau:
I wonder: maybe `CGAL::write_ply` should do that itself. @sgiraudot, what do you think? Do we let that responsibility to users, or should `CGAL::write_ply` modify temporarily the precision?

afabri:
I think it is a bad idea that we start setting the precision internally. Imagine a user who knows that he only needs a certain precision because he knows the precision of a measuring device.  It is more educative to add it in each example even if this seems redundant.

oleg-alexandrov:
I agree the precision should not be set internally but that the examples should show how it is done.

sgiraudot:
I agree with @afabri, we shouldn't handle that (also for PLY, you don't necessarily need to set precision as `CGAL::set_binary_mode()` will make PLY be written in binary).

It would indeed be good idea to add the precision line in an example, but here I don't really see the point in putting it in a test which is not user-oriented.

oleg-alexandrov:
I agree that this test is not user-facing. Yet, from my experience, sometimes documentation or user examples are not complete, and a user can dig through the code to see what existing logic can address their needs, and tests often show useful APIs. I think it is a good practice to set the precision each time a text file is written when possible.

maxGimeno:
@oleg-alexandrov Can you please add me as a collaborator so I can edit some examples in your branch, instead of creating a new one ?

oleg-alexandrov:
I just did add you as a collaborator. Though I thought it was the default to allow people from the parent repo of a fork to access the fork.

maxGimeno:
Thank you. 
It is not, we always have to ask for permission :)

lrineau:
Among the [default settings when creating a stream (deriving from `std::basic_ios<CharT,Traits>`][init], I wonder also about the locale.  By default, the locale is a copy of the global locale, that may have been changed by earlier parts of the program. In particular, for Qt5 programs on Unix/POSIX systems, the locale may be changed to the preferred locale of the user, from the `LC_ALL` or `LC_NUMERIC` environment variable. In my opinion, as soon as we are outputting anything to a known file format, it should be using the classic locale. That can be done with:

```c++
out.imbue(std::locale::classic());
```

[init]: https://en.cppreference.com/w/cpp/io/basic_ios/init

afabri:
@lrineau I think we should not overdo it.  Precision issues are a recurrent issue in geometric computing and limited precision may lead to degenerate edges or faces, whereas setting the locale at least in our examples without Qt adress a non-existing problem.

Surface_mesh/examples/Surface_mesh/sm_join.cpp
```diff
@@ -34,5 +34,5 @@ int main(int argc, char* argv[])
   }
   
   
-  std::cout << sm1 << std::endl;
+  std::cout <<std::setprecision(17)<< sm1 << std::endl;
```
afabri: Can you please put a blank around the `<<` in all your changes.

Polyhedron/examples/Polyhedron/polyhedron_prog_subdiv.cpp
```diff
@@ -102,6 +102,6 @@ int main(int argc, char* argv[]) {
         std::exit(1);
     }
     subdiv( P);
-    std::cout << P;
+    std::cout << std::setprecision(17)<<P;
```
lrineau:
```suggestion
    std::cout << std::setprecision(17)<< P;
```

lrineau:

PR #4417, opened 2019-12-13
Remove call to boost in PMP::connected_components

sloriot:
This is a proposal to remove the call to the boost function.
Pros:
- remove a requirement of a defaut constructable map (with a cryptic error that I'm sure I'm not the only one that have lost time to understand why it failed).
- we have the hand on the code if we want to to the collection of faces per components

I did not bench the new version against the old one. @maxGimeno when you have time could you please give it a try?


maxGimeno:
I just benched it:

13M of vertices and triangles
bench code : 
```
Mesh::Property_map<face_descriptor,std::size_t> fccmap;
  fccmap = m.add_property_map<face_descriptor,std::size_t>("f:CC").first;
  
  timer.start();
  CGAL::Polygon_mesh_processing::connected_components(m, fccmap);
  timer.stop();
```

results : mean of 10 runs
  master : 3.15383 seconds.
  branch : 1.00775 seconds.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -224,22 +226,40 @@ connected_components(const PolygonMesh& pmesh,
...
+  for (face_descriptor f : faces(pmesh))
+  {
+    if (handled[get(fimap,f)]) continue;
+    std::vector<face_descriptor> queue;
+    queue.push_back(f);
+    while(!queue.empty())
+    {
+      face_descriptor fq = queue.back();
+      queue.pop_back();
+      typename boost::property_traits<FaceComponentMap>::value_type  fq_id = get(fimap,fq);
```
MaelRL: value type of `FaceIndexMap`, not `FaceComponentMap`

lrineau:

PR #4153, opened 2019-08-13
[Small Feature] OpenGR and PointMatcher wrappers

necipfazil:
## Summary

Adds wrappers for OpenGR and pointmatcher libraries.

Both wrappers allow computing registration transformation of two CGAL point sets, where OpenGR wrapper uses Super4PCS algorithm while pointmatcher wrapper uses ICP algorithm.

OpenGR (Super4PCS) does a coarse registration while pointmatcher (ICP) does a fine registration when compared. Therefore, those two wrappers together can be pipelined to have precise registration results in an efficient manner. 

TODO:

- [x] Changes.md to 5.1
- [ ] news entry (?)

## Release Management

* Affected package(s): Point set processing
* Small Feature:  [Features/Small_Features/OpenGR_and_PointMatcher_wrappers](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/OpenGR_and_PointMatcher_wrappers)
* Link to compiled documentation: [Documentation/Point_set_processing](http://www.necipyildiran.com/other/cgal_opengr_pointmatcher_wrappers_doc/Point_set_processing_3/)
* License and copyright ownership: [CLA-GF](https://www.cgal.org/CLA/CLA-GF.html), GeometryFactory
* Developed and locally tested (GNU/Linux) with CGAL 5.0

* GSoC: https://cgal.geometryfactory.com/CGAL/Members/wiki/GSoC2019/OpenGR

PR #4418, opened 2019-12-13
PMP: Add `point_map` and `vertex_point_map` to `polygon_soup_to_polygon_mesh`

sloriot:
... to `polygon_soup_to_polygon_mesh()`

Also add `CGAL::Polygon_mesh_processing::polygon_mesh_to_polygon_soup()`

TODO:
- [x] Fix ps_to_pm not working for "array of double" point types
- [x] Check that all is well with the demo
- [x] Update and submit small feature
- [x] Make it work with an array

## Release Management

* Feature/Small Feature (if any): [PM_to_PS and NPs for PS_to_PM](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/NP_for_PS_to_PM)
* **Pre-approval date:** 2020/03/13


MaelRL:
I think it would be best to give it directly named parameters rather than just adding an optional VPM (template) parameter.

It should also be a point_map like in PSP_3 and stuff rather than a VPM since there are no vertices.

sloriot:
The vpm is for the output mesh, but you're right that there could be a map to access the points in `PointRange`.

MaelRL:
So:
```
void polygon_soup_to_polygon_mesh( PointRange& points,
	 							   const PolygonRange& polygons,
							       PolygonMesh& out,
							       const NP_IN& np_in, const NP_OUT& np_out)
```
with NP_IN offering `point_map` and NP_OUT offering `vertex_point_map` 
?

STL_Extension/include/CGAL/Container_helper.h
```diff
@@ -0,0 +1,80 @@
...
+{
+}
+
+// Plenty of times we ask for a model of RandomAccessContainer and use reserve, but some models
+// of the concept do not have a reserve (e.g. deque)
+CGAL_GENERATE_MEMBER_DETECTOR(reserve);
+
+// Container with 'reserve'
+template <class Container>
+void reserve(Container& c, std::size_t size,
```
MaelRL: Wouldn't mind some feedback on that one. 

I added reserve here (size/resize was also added by me a while ago in IO for similar reasons). because I noticed that the concepts are requiring say RandomAccessContainer, but the models do not necessarily have a reserve function (a lot would fail if we provided strict concepts and not std::vector). However, when it's there, it's nice to be able to use it. Therefore this addition, but it's a bit awkward.

sloriot: I think it's fine. It's just hard to extend if the user comes with its own container (because it would mean defining an overload) but `reserve()` is a pretty good name so for most users it will be fine. An alternative could be to provide those function from a documented struct but I don't think it does worth it.You might want to rename internal into container_helper just to make easier to match to this header when reading the implementation.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_mesh_to_polygon_soup.h
```diff
@@ -0,0 +1,117 @@
...
+///                    `BackInsertionSequence` whose value type can be constructed from
+///                    the point type of the polygon mesh
+/// \tparam PolygonRange a model of the concepts `RandomAccessContainer` and `BackInsertionSequence` whose
+///                      value type is itself a model of the concepts `RandomAccessContainer` and
+///                      `BackInsertionSequence` whose value type is `std::size_t`
+/// \tparam NamedParameters a sequence of \ref pmp_namedparameters "Named Parameters"
+///
+/// \param mesh the mesh whose faces are being put in the polygon soup
+/// \param points points making the polygons of the soup
+/// \param polygons each element in the vector describes a polygon using the index of the points in `points`
```
janetournois: "index" should be "indices"

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_mesh_to_polygon_soup.h
```diff
@@ -0,0 +1,117 @@
...
+/// \param np optional sequence of \ref pmp_namedparameters "Named Parameters" among the ones listed below
+///
+/// \cgalNamedParamsBegin
+///   \cgalParamBegin{vertex_point_map}
+///     a model of `ReadablePropertyMap`, the property map with the points associated to the vertices of `mesh`.
+///     If this parameter is omitted, an internal property map for
+///     `CGAL::vertex_point_t` must be available in `PolygonMesh`.
```
janetournois: should it be available in `PolygonMesh`, or in `out`?
It seems to me that with dynamic property maps, that it is `out`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
```diff
@@ -10,225 +10,278 @@
...
+* @tparam PolygonMesh a model of `MutableFaceGraph`
+* @tparam PointRange a model of the concept `RandomAccessContainer`
+* whose value type is the point type
+* @tparam PolygonRange a model of the concept `RandomAccessContainer` whose
+* value type is a model of the concept `RandomAccessContainer` whose value type is `std::size_t`
+* @tparam NamedParameters_PS a sequence of \ref pmp_namedparameters "Named Parameters"
+* @tparam NamedParameters_PM a sequence of \ref pmp_namedparameters "Named Parameters"
+*
+* @param points points of the soup of polygons
+* @param polygons each element in the vector describes a polygon using the index of the points in `points`
```
janetournois: "index" should be "indices"

maxGimeno:
This creates conflicts with #4153 which uses PSP::GetPointMap

MaelRL:
I moved `GetPointMap` into `CGAL::` since it now being used in other packages (PMP).

maxGimeno:
@MaelRL the messages were a note for integration. We should chose one to remove, and modify the other once it is master

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-102.shtml

PR #4428, opened 2019-12-20
Handle "border" case

sloriot:
I cannot produce a test for that case because it depends on the iteration order in the model (detected on a model I cannot share)


maxGimeno:

PR #4429, opened 2019-12-20
PMP:  deterministic corefinement

maxGimeno:
## Summary of Changes

This PR adds the maccro CGAL_FORCE_DETERMINISTIC_COREFINEMENT to the code of corefinement, that force it to be deterministic using a flat_set. 

## Release Management


maxGimeno:
@sloriot it is not working after all...

sloriot:
@lrineau could you confirm that in contrast with a `std::map` the iteration over elements of an `std::unordered_set` is not deterministic from one run to another  (if the sequence of insertions of elements is not always the same).

lrineau:
[**@sloriot**](https://github.com/sloriot) commented on [Jan 8, 2020, 10:17 AM GMT+1](https://github.com/CGAL/cgal/pull/4429#issuecomment-571960130 "2020-01-08T09:17:54Z - Replied by Github Reply Comments"):
> @lrineau could you confirm that in contrast with a `std::map` the iteration over elements of an `std::unordered_set` is not deterministic from one run to another  (if the sequence of insertions of elements is not always the same).

Indeed, I do not think the standard says anything about the ordering of elements for an `unordered_set`.

maxGimeno:
@sloriot What do we do with that PR ? I propose to close it as the concept  doesn't work.

PR #4439, opened 2020-01-01
Add to the user manual example a parameter described in the text

oleg-alexandrov:
The user manual at 

https://doc.cgal.org/latest/Advancing_front_surface_reconstruction/index.html 

goes to great lengths to describe the parameter radius_ratio_bound. Yet, the example right below that is meant to illustrate the code is missing it. One of course visit the reference documentation and dig up what it is and where it should go, but it can make the example more useful to actually use the parameter that is discussed. I am setting it to the value 5 that is used as the default value per the user and reference manuals.


lrineau:

Issue #4442, opened 2020-01-03
Link CGAL with Eigen library!


KeepFaithMe:
_Please use the following template to help us solving your issue._

## Issue Details

when I use the parameterization algorithm,some problem "**_You must either provide 'SolverTraits_' or link CGAL with the Eigen library_**".But W hen I compile the CGAL ,I add the Eigen library to Cmake,and success.So how can I handle this problem.The Parameteriztion algroithm is from https://doc.cgal.org/latest/Surface_mesh_parameterization/Surface_mesh_parameterization_2simple_parameterization_8cpp-example.html

CGAL versioin 4.13.1
Operating system          Win10  x64

## Source Code

_If your issue arises by using CGAL in your own source code, please provide a minimalist example that we can compile easily to reproduce the bug. If your issue arises from using a CGAL program (demo, example, etc.), please let us know which one. Helping you solving an issue is much easier and efficient if we can reproduce it._

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):
* Compiler:
* Release or debug mode:
* Specific flags used (if any):
* CGAL version:
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

lrineau:
Your report is not really good, because you do not tell us how you have used CMake. Without that, I have to guess, and here is my advice: make sure you enable Eigen, in CMake, when you configure the example program, and not when you configure the CGAL libraries.

KeepFaithMe:
I am so sorry&nbsp;I didn't make myself clear.I mean when I use Cmake 3.9.0 compile the CGAL library ,there is an option 'WITH_Eigen3',and I selected it,and I add the Eigen path into the 'EIGEN3_INCLUDE_DIR'. The Eigen version is 3.7.So when I use the example&nbsp;https://doc.cgal.org/latest/Surface_mesh_parameterization/Surface_mesh_parameterization_2discrete_authalic_8cpp-example.html, The Visual studio 2017 show: You must either provide 'SolverTraits_' or link CGAL with the Eigen library.I saw the source code:&nbsp;https://github.com/CGAL/releases/blob/master/include/CGAL/Surface_mesh_parameterization/Discrete_conformal_map_parameterizer_3.h&nbsp; and define the 'CGAL_EIGEN3_ENABLED',but it is still wrong.So how can it deal with it!
looking forward your reply!
good luck!




as following is my testing code:


#include <CGAL/Eigen_solver_traits.h&gt;
[...]

PR #4443, opened 2020-01-03
SMP: Fix bad usage of pragma message

MaelRL:
## Summary of Changes
I added a pragma message in 9eebcfb so that the error is not just something like `CGAL::Default has no type Vector`. However, it was badly done since the pragma message will show up as soon as Eigen is not linked, even if the user has correctly provided a traits class.

Reported in https://github.com/CGAL/cgal/issues/4442#issuecomment-570511774.

## Release Management

* Affected package(s): `Surface_mesh_parameterization`
* Issue(s) solved (if any):


maxGimeno:

PR #4446, opened 2020-01-05
Straight_skeleton_builder_2_impl.h: Get more precision for free.

### LINKED ISSUE #4445: Pi has more digits than 6
https://github.com/CGAL/cgal/blob/76ff945bf47e9d6a123bf51670ba2f5fc88c32ea/Straight_skeleton_2/include/CGAL/Straight_skeleton_2/Straight_skeleton_builder_2_impl.h#L1424

CGAL overall has a few different definitions of PI that have better accuracy than this one...  I would suggest using one or just us M_PI from math.h

fenrus75:
## Summary of Changes

The compute speed of double precision does not depend on the number of decimals,
so it's free and better to use more than 5 or 6 decimals for PI.

## Release Management

* Affected package(s): `Straight_skeleton_2`
* Issue(s) solved (if any): fix #4445


MaelRL:
Thanks for the patch. `CGAL_PI` should probably be used instead.

On a sidenote, there are actually quite a few places where pi is hard-coded with various degrees of precision...

MaelRL:
Replaced by #4450 that generalize the patch. I kept the commit so that it appears in history.

PR #4450, opened 2020-01-07
Use `CGAL_PI` where relevant

MaelRL:
## Summary of Changes

Use `CGAL_PI` instead of hardcoding pi (usually with a lower precision).

## Release Management

* Affected package(s): Multiple
* Issue(s) solved (if any): #4445


lrineau:

Issue #4410, opened 2019-12-11
Size of holes to fill


oleg-alexandrov:
## Issue Details

In https://doc.cgal.org/latest/Polygon_mesh_processing/index.html it is described how to fill holes. 

Unfortunately, not all meshes are closed. This algorithm will try to fill any hole, even one at the outer boundary of a mesh. Imagine here a mesh that represents a portion of the terrain surface, such a mesh has a boundary where it terminates, and it also may have some holes.

In such a situation the algorithm hangs, and likely will give a junk result.

There is likely no fool-proof solution here. At least, the algorithm could choose not to fill holes whose boundary have more than a given number of edges, or a hole with given diameter. 

# Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux
* Compiler: gcc 
* Release or debug mode:
* Specific flags used (if any):
* CGAL version: 5
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

sloriot:
The algorithm is working on boundary cycles that the user provides as input. So you are free to measure the size of the boundary cycle (#edges or length) before trying to fill it.

MaelRL:
There is also some code that does just that in the following file: https://github.com/CGAL/cgal/blob/master/Polyhedron/demo/Polyhedron/Plugins/PMP/Hole_filling_plugin.cpp as the CGAL demo already has a button "Select small holes".

oleg-alexandrov:
Thank you. That example at 

https://doc.cgal.org/latest/Polygon_mesh_processing/Polygon_mesh_processing_2hole_filling_example_8cpp-example.html

is still however not very helpful. It does not explain how to get a boundary cycle that has this particular edge and how one would find its number of edges. I guess it is simple if you already know that, and it could be simple for you to add that, but as  a new user I find this very intimidating.

In short, it could be little work for this example to be improved as the user does not get sufficient knowledge of how a hole could be understood before deciding to fill it or not. And for me the whole code just hanged, I guess when it got to a huge hole when it would give junk results anyway.

PR #4455, opened 2020-01-09
Add control over hole size in the hole-filling example

oleg-alexandrov:

## Summary of Changes

I found the hole-filling functionality of CGAL to be very powerful, but problematic in practice, since there is no control over the hole size. Many real-world meshes are not closed, or otherwise have huge holes, and the algorithm either hangs on those or can give junk results.

Here I am making a small addition to the hole-filling example by allowing the user to control the diameter and number of edges in a hole to fill.

I added this to the surface mesh example (there are 4 hole-filing examples in the repo), because the Surface_mesh datastructure seems to be more natural when dealing with meshes than the Polyhedron_3 datastructure, and I mention in the doc that examples exist for the other ones as well (but those I did not modify).

I tested this extensively, including with very large and noisy meshes. A simple way of testing the changed example is to invoke it with the  mech-holes-shark.off dataset in the repo, while setting the maximum hole diameter to 10 and the maximum number of edges in a hole to 80. Then the largest of the 4 holes won't be filled, as desired. 

It took me very many hours to develop this simple change. I hope it can be useful to some folks. 

Lastly, I raised this first at: https://github.com/CGAL/cgal/issues/4410 and I followed the suggestions outlined there.

## Release Management

* Affected package(s): `Polygon Mesh Processing`
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4410
* License and copyright ownership: Same as the CGAL codebase

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
@@ -412,10 +412,13 @@ one incident non-null face) after each hole filling step.
 Holes are filled one after the other, and the process stops when there is no border edge left.
 
 This process is illustrated by the example below, where holes are
-iteratively filled, refined and faired to get a faired mesh with no hole.
+iteratively filled, refined and faired. Optionally, only holes
+not exceeding a certain diameter or number of edges can be filled.
+This example assumes that the mesh is stored in a `Surface_mesh`
+datastructure. Analogous examples when using the `Polyhedron_3`
+class and a few others are part of the code base.
 
-
-\cgalExample{Polygon_mesh_processing/hole_filling_example.cpp}
+\cgalExample{Polygon_mesh_processing/hole_filling_example_SM.cpp}
 
  \cgalFigureBegin{Triangulated_fork, fork.jpg}
  Holes in the fork model are filled with triangle patches.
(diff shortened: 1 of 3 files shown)


oleg-alexandrov:
The checkbox "Allow edits from maintainers." was and is checked. Is there something else I need to do? Thanks.

oleg-alexandrov:
I added you as an actual collaborator to my fork from the settings. I never used this before and I am not sure how it relates to the "Allow edits from maintainers".  Hopefully one of these can do the job.

oleg-alexandrov:
Thanks. I also tested the updated code on my side and it does well.

maxGimeno:
Warnings in the doc : https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1-Ic-52/logs2/Polygon_mesh_processing.log

maxGimeno:
Also errors : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-52.shtml#Polygon_mesh_processing_Examples

oleg-alexandrov:
Maybe the doc error can be fixed by replacing in line:

https://github.com/CGAL/cgal/blob/fe321e686270eddde25a4a06d099488654f1e455/Polygon_mesh_processing/doc/Polygon_mesh_processing/examples.txt#L4

hole_filling_example.cpp with hole_filling_example_SM.cpp

I am not sure. When I modified the doc, I did replace hole_filling_example.cpp with hole_filling_example_SM.cpp, and the CGAL doc processing system should have found it, since both of them are in the same directory. Maybe it does not because I should have also made that change that I am suggesting above. I don't know how to trigger building the CGAL doc. 

I don't know about the other error in the regressions. The modified example should have given the same answer as the original, since the new parameters in it were set to -1 which should have triggered the original behavior.

maxGimeno:
@oleg-alexandrov  yes, that is a change that needs to be done anyway.

lrineau:

PR #4471, opened 2020-01-15
PMP: Enable enforcing which edge is kept during stitching

### LINKED ISSUE #3863: PMP: add a way to enforce which edge is preserved during stitching
For the user, which edge is preserved during stitching is seemingly random. It would be nice to have a way to enforce it. 

Here is a [gist](https://gist.github.com/MaelRL/638fbf638a233c3114d958af78fc9489) with some code allowing to do it in `collect_duplicated_stitchable_boundary_edges`, passing a halfedge comparer as an additional named parameter.

Would need to pass it through the main functions (`stitch_borders()`, `stitch_boundary_cycles()`) directly, if it's the chosen way to go.

MaelRL:
## Summary of Changes

Add a named parameter and code which enables passing a halfedge comparer such that it is possible to choose which edges are kept during stitching (combinatorially speaking). An example of application is the case where some edges are constrained (a comparer is provided for this case).

All internal and undocumented for now.

## Release Management

* Issue(s) solved (if any): fix #3863


MaelRL:
There is already a `LessHalfedge` in `stitch_borders()` (but in a geometrical sense whereas this new one is more of a combinatorial sense). I do agree that the current name is not pleasant.

afabri:
Comparator.  But what is the result_type?

lrineau:
In the [documentation of the STL, like in `std::set`](https://en.cppreference.com/w/cpp/container/set):
- those functors are defined as "key comparison function",
- the name of the template parameter is `Compare`,
- and the name of the parameter in the constructor is `comp`.

So maybe `halfedges_comp` or `halfedges_comparator`?

[Here in the STL](https://en.cppreference.com/w/cpp/container/map/value_compare) we can see occurrence of the wording "comparator".

sloriot:
Actually, the whole thing is not about comparing but which halfedge should be kept. We should use another name `Halfedge_keeper` ?

sloriot:
`Halfedge_merger`?

MaelRL:
Some issue during merge into `integration` is responsible for the red PMP (https://github.com/CGAL/cgal-dev/blob/testsuite-Monday/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h#L975, actualy code: https://github.com/MaelRL/cgal/blob/PMP-Halfedge_comparer_for_stitch_borders-GF/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/stitch_borders.h#L970)

MaelRL:
Still causing timeouts on Windows somehow

maxGimeno:

PR #4472, opened 2020-01-15
Fix iterator type required

sloriot:
detected by the bindings

no need to backport IMO


lrineau:

PR #4483, opened 2020-01-21
PMP: Add a function for hole_filling based on a cdt_2

### LINKED ISSUE #4464: Delaunay triangulation goes outside border
## Issue Details

If I try to fill the holes in the object in the gist, then I end up with a shape that includes triangles that are outside the border of the object.

Disabling Delaunay triangulation fixes the problem.

## Source Code

Source and object are at https://gist.github.com/benlaurie/1ed073fa9777a53993967de5fd77b7f7

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): FreeBSD 64 bit
* Compiler: clang
* CGAL version: 5.0
* Boost version: 1.72.0

maxGimeno:
## Summary of Changes
In the case a hole is entirely in a plane, the basic triangulation_3 may fail. 
This PR is a first step to an alternative based on a cdt2 and hole_filling.

## Release Management

* Affected package(s):PMP
* Fix #4464


sloriot:
I wonder if using the projection traits (extended if needed) to call the `is_simple` test for a polygon with holes is not exactly what is needed. Maybe we can also find this information while building the CDT. cc @danston

danston:
Hi all. I will investigate this and other possible solutions asap and let you know what could be done.

lrineau:
Replaced by #4749.

danston:
I have proposed an extended solution in the PR #4749.

PR #4484, opened 2020-01-21
Also test Triangulation with Epeck_d

mglisse:
## Summary of Changes

More testing for Epeck_d would be good. We are getting compilation issues on windows in gudhi, we'll see if they already appear in CGAL this way.

## Release Management

### DIFF
--- a/Triangulation/test/Triangulation/test_delaunay.cpp
+++ b/Triangulation/test/Triangulation/test_delaunay.cpp
@@ -10,6 +10,7 @@ int main()
 #else
 
 #include <CGAL/Epick_d.h>
+#include <CGAL/Epeck_d.h>
 #include <CGAL/point_generators_d.h>
 #include <CGAL/Delaunay_triangulation.h>
 #include <CGAL/algorithm.h>
@@ -112,13 +113,21 @@ void test(const int d, const string & type, const int N)
 template< int D >
 void go(const int N)
 {
-    typedef CGAL::Epick_d<CGAL::Dimension_tag<D> > FK;
-    typedef CGAL::Delaunay_triangulation<FK> Triangulation;
+    typedef CGAL::Epick_d<CGAL::Dimension_tag<D> > KI;
+    typedef CGAL::Delaunay_triangulation<KI> Triangulation;
     test<Triangulation>(D, "static", N);
 
-    typedef CGAL::Epick_d<CGAL::Dynamic_dimension_tag> FK_dyn;
-    typedef CGAL::Delaunay_triangulation<FK_dyn> Triangulation_dyn;
+    typedef CGAL::Epick_d<CGAL::Dynamic_dimension_tag> KI_dyn;
+    typedef CGAL::Delaunay_triangulation<KI_dyn> Triangulation_dyn;
     test<Triangulation_dyn>(D, "dynamic", N);
+
+    typedef CGAL::Epeck_d<CGAL::Dimension_tag<D> > KE;
+    typedef CGAL::Delaunay_triangulation<KE> TriangulationE;
+    test<TriangulationE>(D, "static", N);
+
+    typedef CGAL::Epeck_d<CGAL::Dynamic_dimension_tag> KE_dyn;
+    typedef CGAL::Delaunay_triangulation<KE_dyn> TriangulationE_dyn;
+    test<TriangulationE_dyn>(D, "dynamic", N);
 }
 
 int main(int argc, char **argv)
--- a/Triangulation/test/Triangulation/test_delaunay.cpp
+++ b/Triangulation/test/Triangulation/test_delaunay.cpp
@@ -115,19 +115,19 @@ void go(const int N)
 {
     typedef CGAL::Epick_d<CGAL::Dimension_tag<D> > KI;
     typedef CGAL::Delaunay_triangulation<KI> Triangulation;
-    test<Triangulation>(D, "static", N);
+    test<Triangulation>(D, "inexact static", N);
 
     typedef CGAL::Epick_d<CGAL::Dynamic_dimension_tag> KI_dyn;
     typedef CGAL::Delaunay_triangulation<KI_dyn> Triangulation_dyn;
-    test<Triangulation_dyn>(D, "dynamic", N);
+    test<Triangulation_dyn>(D, "inexact dynamic", N);
 
     typedef CGAL::Epeck_d<CGAL::Dimension_tag<D> > KE;
     typedef CGAL::Delaunay_triangulation<KE> TriangulationE;
-    test<TriangulationE>(D, "static", N);
+    test<TriangulationE>(D, "exact static", N);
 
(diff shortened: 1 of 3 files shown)


Triangulation/test/Triangulation/test_delaunay.cpp
```diff
@@ -112,13 +113,21 @@ void test(const int d, const string & type, const int N)
...
+    typedef CGAL::Delaunay_triangulation<KI_dyn> Triangulation_dyn;
     test<Triangulation_dyn>(D, "dynamic", N);
+
+    typedef CGAL::Epeck_d<CGAL::Dimension_tag<D> > KE;
+    typedef CGAL::Delaunay_triangulation<KE> TriangulationE;
+    test<TriangulationE>(D, "static", N);
+
+    typedef CGAL::Epeck_d<CGAL::Dynamic_dimension_tag> KE_dyn;
+    typedef CGAL::Delaunay_triangulation<KE_dyn> TriangulationE_dyn;
+    test<TriangulationE_dyn>(D, "dynamic", N);
```
lrineau: You forgot `inexact` and `exact`, in the names.

mglisse: Indeed, done. Maybe I started with this file and only thought about adding (in)exact while doing the second file, I don't remember.

MaelRL:
There are some errors in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-68.shtml#Triangulation.

mglisse:
> There are some errors in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-68.shtml#Triangulation.

A snapshot of gcc-10 crashes (the compiler itself) during optimization? If I have time I'll try to reproduce and report a bug, but that shouldn't block the branch (which only adds tests).

MaelRL:
I mostly meant the MSVC one (https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-68/Triangulation/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz).

mglisse:
Ah, great, that's exactly the error I got in gudhi and was hoping to reproduce.
I don't intend to work on it (people who want the new Epeck_d can use a more recent compiler where that bug is fixed), but at least now the error is visible somewhere...

lrineau:
> I mostly meant the MSVC one ([cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-68/Triangulation/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-68/Triangulation/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)).

Probably a compiler bug (from MSVC 2015), because it does not seem to occur with MSVC 2017 or MSVC 2019.

lrineau:
> > There are some errors in [cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-68.shtml#Triangulation](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-68.shtml#Triangulation).
> 
> A snapshot of gcc-10 crashes (the compiler itself) during optimization? If I have time I'll try to reproduce and report a bug, but that shouldn't block the branch (which only adds tests).

@mglisse Have you reported the bug? Can you please quote it the URL of it?

lrineau:
> Ah, great, that's exactly the error I got in gudhi and was hoping to reproduce.
> I don't intend to work on it (people who want the new Epeck_d can use a more recent compiler where that bug is fixed), but at least now the error is visible somewhere...

@mglisse Would you allow a patch with workaround?
```patch
--- include/CGAL/NewKernel_d/Lazy_cartesian.h	2020-02-03 21:01:20.000000000 +0100
+++ include/CGAL/NewKernel_d/Lazy_cartesian.h	2020-02-07 15:36:38.175976400 +0100
@@ -109,6 +109,12 @@
 class Lazy_rep_XXX :
   public Lazy_rep< AT, ET, E2A >, private EC
 {
+  // `default_construct<T>()` is the same as `T{}`. But, this is a
+  // workaround to a MSVC-2015 bug: its parser seemed confused by
+  // `T{}` somewhere below.
+  template <typename T>
+  static T default_construct() { return T(); }
+
   // Lazy_rep_0 does not inherit from EC or take a parameter AC. It has different constructors.
   static_assert(sizeof...(L)>0, "Use Lazy_rep_0 instead");
   template <class Ei, class Ai, class E2Ai, class Ki> friend class Lazy_kernel_base;
@@ -138,7 +144,7 @@
   // Currently we construct the vectors, then move them into the tuple. It would be nicer to construct them in their final destination, because eventually we will also have arrays instead of vectors.
   template<class...T,class LLL,class...LL>
   Lazy_rep_XXX(Lazy_internal::typelist<T...>, const AC& ac, const EC& ec, LLL const&lll, LL const&...ll) :
-    Lazy_rep<AT, ET, E2A>(ac(CGAL::approx(ll)...)), EC(ec), l(Lazy_internal::do_extract(T{},lll)...)
+    Lazy_rep<AT, ET, E2A>(ac(CGAL::approx(ll)...)), EC(ec), l(Lazy_internal::do_extract(default_construct<T>(),lll)...)
   {
     //this->set_depth(std::max({ -1, (int)CGAL::depth(ll)...}) + 1);
     this->set_depth(1); // FIXME: now that we have ranges, we could actually compute the depth if we cared...

```

I have tested it on our test platform.

mglisse:
> @mglisse Have you reported the bug? Can you please quote it the URL of it?

I didn't manage to reproduce it quickly the other day, and haven't had time to try again.

mglisse:
As far as I can see, T is always a trivial empty type, so if MSVC prefers `T()` that would be fine. Is such a workaround sufficient? Making T a template parameter of do_extract (instead of an argument) also looks like a possibility.
Your default_construct function doesn't look like it default constructs...

lrineau:
I have tried `T()` over `T{}` and that was not sufficient to please the parser of MSVC 2015.

> Your default_construct function doesn't look like it default constructs...

What do you mean?

mglisse:
> What do you mean?

It was just a terminology detail. It does value initialization, which calls the default constructor. Ignore it.
Ok for your patch if you add in the comment that the bug is already fixed in 2017 (make it easier to remove the workaround when we drop 2015), thank you.

mglisse:
Now that this is turning into a bugfix, do you want to retarget to CGAL-5.0-branch?

lrineau:
> Now that this is turning into a bugfix, do you want to retarget to CGAL-5.0-branch?

Yes, good idea. Do you want me to do it?

mglisse:
> Do you want me to do it?

Yes please. Thank you.

maxGimeno:

PR #4260, opened 2019-09-30
SMS: Bounded_input_distance_placement

maxGimeno:
## Summary of Changes
Add a policy in Edge_collapse based on a distance threshold. If a placement would be too far from the mesh, it is discarded.

TODO:
-  [x]  Undocument

## Release Management

* Affected package(s):Surface_mesh_simplification
* <s>Small Feature: [SMS_Bounded_distance_placement](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/SMS_Bounded_distance_placement) -- **Pre-approval date:** 2020/01/07</s>

PR #4486, opened 2020-01-22
SMS: Improve edge_collapse()

### LINKED ISSUE #4485: Improve edge_collapse API
* When passing a constrained edge map to `edge_collapse()` the user has to manually provide a placement wrapped into `Constrained_placement`. It would be much more user friendly to do it automatically as soon as a constrained edge map as been provided (+ check if not already done since we want to keep backward compatibility). We should also deprecate the public definition of the class.

* Having also a `constrained_vertex_map` would also be interesting. Care has to be taken for the placement that will potentially have to handle both a vertex and a edge constrained map

* `Bounded_normal_placement` should be replaced by a named parameter, deprecate the class + handle backward compatibility

maxGimeno:
## Summary of Changes

Make some improvements to edge_collapse() :

 - [x] When passing a constrained edge map to edge_collapse() the user has to manually provide a placement wrapped into Constrained_placement. It would be much more user friendly to do it automatically as soon as a constrained edge map as been provided (+ check if not already done since we want to keep backward compatibility). Use a tag to detect whether the constraint_placement is already present.

 - [x] Check if `const EdgeIsConstrainedMap map = EdgeIsConstrainedMap(),` can be come simply `const EdgeIsConstrainedMap map,` in `Constrained_placement` (that is, do not allow a default pmap because it's pointless and error inducing).
- [x] Add new NP "constrain_geometry" (default true (tiny breaking change to mention)), false = don't wrap, true = wrap.

 - [x] We should also deprecate the public definition of the class.

 - [x] Bounded_normal_placement should be replaced by a named parameter (w/ angle parameter) + handle backward compatibility

 - [x] deprecate the class Bounded_normal_placement

- [x] add unit tests for the new parameters

 - [x] Do the same with #4260 once it is integrated. Don't document the NP.
 - [ ] Before integration : replace numbers in test_named_parameters for better ones.

## Release Management

**Integrate after #4260**

* Affected package(s):Surface_mesh_simplification
* Issue(s) solved (if any): fix #4485


maxGimeno:
@MaelRL it should be ready for a first review.

Polyhedron/demo/Polyhedron/Plugins/Surface_mesh/Mesh_simplification_plugin.cpp
```diff
@@ -153,8 +153,8 @@ void Polyhedron_demo_mesh_simplification_plugin::on_actionSimplify_triggered()
 
     if (selection_item)
       {
-        CGAL::Surface_mesh_simplification::Constrained_placement
-          <CGAL::Surface_mesh_simplification::Bounded_normal_change_placement
+        CGAL::Surface_mesh_simplification::internal::Constrained_placement
+          <CGAL::Surface_mesh_simplification::internal::Bounded_normal_change_placement
```
MaelRL: The bounded normal change placement must check with the finite position. It pre-dates your changes, but still should be fixed. Please check other places too.

Surface_mesh_simplification/examples/Surface_mesh_simplification/edge_collapse_bounded_normal_change.cpp
```diff
@@ -37,7 +36,7 @@ int main(int argc, char** argv)
   const std::size_t stop_n = (argc > 2) ? std::stoi(argv[2]) : num_halfedges(surface_mesh)/2 - 1;
   SMS::Count_stop_predicate<Surface_mesh> stop(stop_n);
 
-  typedef SMS::Bounded_normal_change_placement<SMS::LindstromTurk_placement<Surface_mesh> > Placement;
+  typedef SMS::LindstromTurk_placement<Surface_mesh> Placement;
```
MaelRL: You can even use the default cost / placement now; it needed to be explicit because you had to wrap, but now this example is just about using the named parameter.

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/edge_collapse.h
```diff
@@ -87,25 +91,64 @@ int edge_collapse(TM& tmesh,
...
-                                 choose_parameter(get_parameter(np, internal_np::get_placement_policy),
-                                                  LindstromTurk_placement<TM>()),
-                                 choose_parameter(get_parameter(np, internal_np::graph_visitor),
-                                                  internal::Dummy_visitor()));
+  typedef typename GetGeomTraits<TM, NamedParameters>::type       Geom_traits;
+  
+  
+  typedef typename internal_np::Lookup_named_param_def <
+    internal_np::edge_is_constrained_t,
+    NamedParameters, internal_np::Param_not_found > ::type        ConstrainedMapOriginalType;
```
MaelRL: No camel case in typedef please (and consistent indentation)

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/edge_collapse.h
```diff
@@ -87,25 +91,64 @@ int edge_collapse(TM& tmesh,
...
+
+  bool do_constrain = choose_parameter(get_parameter(np, internal_np::constrain_geometry),
+                                       true);
+
+  typename TempPlacement::type tmp_placement = TempPlacement::get_placement(c_map, placement,
+                                                                       !internal::Has_nested_type_constrained_tag<PlacementType>::value && do_constrain);
+  double max_angle = choose_parameter(get_parameter(np, internal_np::max_normal_angle_change),
+                                      CGAL_PI);
+  typename FinalPlacement::type final_placement = FinalPlacement::get_placement(max_angle, tmp_placement);
+
```
MaelRL: indentation

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Placement_wrappers.h
```diff
@@ -0,0 +1,212 @@
...
+  {
+    if(!m_constrain_geom)
+    {
+      return BasePlacement::operator()(profile);
+    }
+    typedef typename Profile::TM                                    TM;
+    typedef typename boost::graph_traits<TM>::halfedge_descriptor   halfedge_descriptor;
+
+    for(halfedge_descriptor h : halfedges_around_target(profile.v0(), profile.surface_mesh()))
+    {
```
MaelRL: @sloriot @afabri What should happen here if the base placement returns `boost::none` ? Right now, if one of the vertices is constrained, we just ignore the fact that the base failed and use that vertex, but should it just also return boost::none instead?

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Placement_wrappers.h
```diff
@@ -0,0 +1,212 @@
...
+           Vector eq2r = gt.construct_vector_3_object()(q2, r);
+
+           Vector n1 = gt.construct_cross_product_vector_3_object()(eqp, eqr);
+           Vector n2 = gt.construct_cross_product_vector_3_object()(eq2p, eq2r);
+
+           const FT sp = gt.compute_scalar_product_3_object()(n1, n2);
+           const FT cos_bound = std::cos(m_angle);
+           bool is_bound_pos = is_positive(cos_bound);
+           bool is_pos = is_positive(sp);
+           const FT sq_cos_bound = CGAL::square(cos_bound);
```
MaelRL: Store this

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Placement_wrappers.h
```diff
@@ -0,0 +1,212 @@
...
+           if(!is_pos && is_bound_pos
+              || (!is_pos && !is_bound_pos && CGAL::square(sp) > n1.squared_length() * n2.squared_length() * sq_cos_bound)
+              || (is_pos && is_bound_pos && CGAL::square(sp) < n1.squared_length() * n2.squared_length() * sq_cos_bound))
+             return boost::optional<typename Profile::Point>();
```
MaelRL: You can simplify and cache this big if()

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Placement_wrappers.h
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+/*******************
+ * helper classes *
+ ******************/
+
+template<class Placement, class IsConstrainedMap>
+struct GetPlacementType{
+  typedef Constrained_placement<Placement, IsConstrainedMap> type;
```
MaelRL: indentation

Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Placement_wrappers.h
sloriot: In order not to surprise users, it should return `boost::none` as well IMO. Some users will use the cost, others the placement to prevent an edge to be collapsed.

maxGimeno: What about the cost ? It must have a cost to test every not constrained result when it was not done before.

sloriot: The cost is using the placement.

maxGimeno: yes

sloriot:
@MaelRL do you also think we should not integrate this PR?

MaelRL:
I wasn't so convinced by some of the API changes in the end, but some features can (should) be extracted.

Issue #4487, opened 2020-01-23
Demo frozen after resetting selection


sloriot:
Open a mesh -> Point inside Polyhedron -> Sample bbox -> Tick "inside" -> Select -> left click on the point set + reset selection -> Tick "Outside" -> Select ---> demo is frozen.

Select requires the point set and the mesh to be selected

PR #4488, opened 2020-01-23
[Demo] Fix point selection in Point_inside_polyhedron_plugin

sgiraudot:
## Summary of Changes

I could reproduce https://github.com/CGAL/cgal/issues/4487 but I don't quite understand why it becomes longer the second time...

Anyway, I've replaced the selection section by a more efficient mechanism using `std::partition` similarly to what I did in the point set selection plugin, it seems to work well from what I tested.

## Release Management

* Affected package(s): demo
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4487


maxGimeno:

PR #4489, opened 2020-01-23
PMP: Fixes for compute_normals

MaelRL:
## Summary of Changes

The assertion checked that the scalar product of a bisector with one of the vectors that it bisects was positive. While this is logically true, numerical errors can sneak in and obtain tiny negative values.

This for example arises when computing the normal of a star that has faces whose normals are opposite, or almost opposite.

Also fixes some corner cases involving degenerate faces.

## Release Management

* Affected package(s): `PMP`


lrineau:

PR #4492, opened 2020-01-28
Remove undocumented CGAL::Point_with_normal_3 from examples

sgiraudot:
## Summary of Changes

There is a structure `CGAL::Point_with_normal_3` that is internally used by the bilateral smoothing algorithm (and some tests). Though undocumented, the header was included in several examples (probably as a result of a large copy-paste, as the structure in itself was not used), and the structure was used in 2 examples of Poisson reconstruction.

I cleaned these examples of it as we should not use undocumented classes in public examples.

## Release Management

* Affected package(s): AFSR, PSP, Shape Detection, Poisson


maxGimeno:

PR #4493, opened 2020-01-28
[Surface Mesh / Point Set] Fix capacity handling in properties

sgiraudot:
## Summary of Changes

The properties used in Surface Mesh and Point Set 3 are internally a vector of vectors with dynamic type. Most operations on vectors are available (`resize(), `reserve()`, etc.), but I noticed that the capacity was not handled well.

This is what can typically happen:
1. Add some properties
2. Call `resize(s)` and `reserve(c)`
3. -> all properties have size `s` and capacity `c`
4. Add a new property
5. -> this new property is given size `s` and capacity `s` (not `c`), because only `resize()` is called when adding a new property, not `reserve()`

It seems to me that once the user does `reserve()`, then a property added afterwards should also be given the wanted capacity, similarly to how size is handled. This PR mimics the behavior of `resize()/size` with `reserve()/capacity`.

## Release Management

* Affected package(s): Surface Mesh + Point Set 3

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Properties.h
@@ -445,7 +445,7 @@ public:
 
 
     // reserve memory for n entries in all arrays
-    void reserve(size_t n) const
+    void reserve(size_t n)
     {
         for (std::size_t i=0; i<parrays_.size(); ++i)
             parrays_[i]->reserve(n);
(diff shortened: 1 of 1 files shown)


Surface_mesh/include/CGAL/Surface_mesh/Properties.h
```diff
@@ -491,6 +500,7 @@ class Property_container
 private:
     std::vector<Base_property_array*>  parrays_;
     size_t  size_;
+    mutable size_t  capacity_;
```
lrineau: `shrink_to_fit()` should not be a `const` method.

maxGimeno:

PR #4495, opened 2020-01-28
Add a kernel that stores objects twice: with Intervals and Gmpq

afabri:
## Summary of Changes

Add a kernel and a numbertype that stores interval approximations for Gmqs.

## Release Management.

* Affected package(s):

### DIFF
--- a/Number_types/include/CGAL/Filtered_rational.h
+++ b/Number_types/include/CGAL/Filtered_rational.h
@@ -1,0 +1,961 @@
+// Copyright (c) 2005-2007  
+// Utrecht University (The Netherlands),
+// ETH Zurich (Switzerland),
+// INRIA Sophia-Antipolis (France),
+// Max-Planck-Institute Saarbruecken (Germany),
+// and Tel-Aviv University (Israel).  All rights reserved. 
+//
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: LGPL-3.0-or-later OR LicenseRef-Commercial
+//
+//
+// Author(s)     : Sylvain Pion, Michael Hemmer
+
+#ifndef CGAL_FILTERED_RATIONAL_H
+#define CGAL_FILTERED_RATIONAL_H
+
+#include <CGAL/Gmpq.h>
+#include <CGAL/Interval_nt.h>
+#include <CGAL/number_type_basic.h>
+#include <sstream>
+
+// A number type class, parameterized by 2 number types NT1 and NT2.
+// It runs all operations on parallel over NT1 and NT2.
+// It is also parameterized by a comparator which compares the values
+// of NT1 and NT2 after all arithmetic operations.
+
+// #define CGAL_NT_CHECK_DEBUG(s) std::cerr << s << std::endl
+#define CGAL_NT_CHECK_DEBUG(s)
+
+namespace CGAL {
+
+namespace internal {
+
+// std::equal_to<T> sticks things to one type, so I define my own.
+struct Equal_to
+{
+  template < typename NT1, typename NT2 >
+  bool operator()(const NT1& a, const NT2& b) const
+  { return true; }
+};
+
+} // namespace internal
+
+
+  template < typename NT1 = Interval_nt<true>, typename NT2=Gmpq, typename Cmp = internal::Equal_to >
+class Filtered_rational
+{
+  NT1 _n1;
+  NT2 _n2;
+  Cmp cmp;
+public:
+
+  typedef Filtered_rational<NT1, NT2, Cmp> Self;
+
+  Filtered_rational() {}
+  
+  Filtered_rational(int i)
+    : _n1(i), _n2(i)
+  { CGAL_assertion(is_valid()); }
+
+  Filtered_rational(double d)
+    : _n1(d), _n2(d)
+  { CGAL_assertion(is_valid()); }
+  
+  Filtered_rational(const NT1 &n1, const NT2 &n2)
+    : _n1(n1), _n2(n2)
+  { CGAL_assertion(is_valid()); }
+
+  Filtered_rational(const std::pair<NT1,NT2> &np)
+    : _n1(np.first), _n2(np.second)
+  { CGAL_assertion(is_valid()); }
+
+  Filtered_rational(const NT2 &n2)
+    : _n1(to_interval(n1)), _n2(n2)
+  { CGAL_assertion(is_valid()); }
+
+  Self& operator=(const std::pair<NT1,NT2> &np)
+  {
+    _n1 = np.first;
+    _n2 = np.second;
+  }
+  
+  // The following need to be dependant on NT1 != {NT2,int,double} ...
+  //Filtered_rational(const NT1 &n1) : _n1(n1), _n2(n1) {}
+  //Filtered_rational(const NT2 &n2) : _n1(n2), _n2(n2) {}
+
+  Self& operator+=(const Self &a)
+  { n2() += a.n2(); n1() = to_interval(n2()); return *this; }
+
+  Self& operator-=(const Self &a)
+  { n2() -= a.n2(); n1() = to_interval(n2()); return *this; }
+
+  Self& operator*=(const Self &a)
+  { n2() *= a.n2(); n1() = to_interval(n2()); return *this; }
+
+  Self& operator/=(const Self &a)
+  { n2() /= a.n2(); n1() = to_interval(n2()); return *this; }
+
+  // Accessors and setters.
+  const NT1& n1() const { return _n1; }
+  const NT2& n2() const { return _n2; }
+
+  NT1& n1() { return _n1; }
+  NT2& n2() { return _n2; }
+
+  // Validity checking: the exact number must be in the interval
+
+  bool is_valid() const
+  {
+    NT1 n2i = to_interval(n2());
+    return (n2i.inf() >= n1().inf())  && (n2i.sup() <= n1().sup());
+  }
+};
+
+  
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator+(const Filtered_rational<NT1, NT2, Cmp> &a)
+{
+   CGAL_NT_CHECK_DEBUG("operator+");
+   return Filtered_rational<NT1, NT2, Cmp>(+ a.n1(), + a.n2() );
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator+(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator+");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() + b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator+(int a, const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator+");
+   return Filtered_rational<NT1, NT2, Cmp>(a + b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator+(const Filtered_rational<NT1, NT2, Cmp> &a, int b)
+{
+   CGAL_NT_CHECK_DEBUG("operator+");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() + b);
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator-(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator-");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() - b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator-(int a, const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator-");
+   return Filtered_rational<NT1, NT2, Cmp>(a - b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator-(const Filtered_rational<NT1, NT2, Cmp> &a, int b)
+{
+   CGAL_NT_CHECK_DEBUG("operator-");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() - b);
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator-(const Filtered_rational<NT1, NT2, Cmp> &a)
+{
+   CGAL_NT_CHECK_DEBUG("unary operator-");
+   return Filtered_rational<NT1, NT2, Cmp>(-a.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator*(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator*");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() * b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator*(int a, const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator*");
+   return Filtered_rational<NT1, NT2, Cmp>(a * b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator*(const Filtered_rational<NT1, NT2, Cmp> &a, int b)
+{
+   CGAL_NT_CHECK_DEBUG("operator*");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() * b);
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator/(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator/");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() / b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator/(int a, const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+   CGAL_NT_CHECK_DEBUG("operator/");
+   return Filtered_rational<NT1, NT2, Cmp>(a / b.n2());
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+Filtered_rational<NT1, NT2, Cmp>
+operator/(const Filtered_rational<NT1, NT2, Cmp> &a, int b)
+{
+   CGAL_NT_CHECK_DEBUG("operator/");
+   return Filtered_rational<NT1, NT2, Cmp>(a.n2() / b);
+}
+
+// arithmetic operators end
+// compare operators begin
+
+template < typename NT1, typename NT2, typename Cmp >
+bool
+operator==(const Filtered_rational<NT1, NT2, Cmp> &a,
+           const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+  if(a.n1().is_point() && (b.n1().is_point())){
+    return a.n1().inf() == b.n1().inf();
+  }
+  
+  if(a.n1().do_overlap(b.n1())){
+    return a.n2() == b.n2();
+  }
+  return false;
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+bool
+operator!=(const Filtered_rational<NT1, NT2, Cmp> &a,
+           const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+  
+  return ! (a == b);
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+bool
+operator<(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+  if(a.n1().sup() < b.n1().inf()){
+    return true;
+  }
+  if(a.n1().inf() > b.n1.sup){
+    return false;
+  }
+  return a.n2() < b.n2();
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+bool
+operator>(const Filtered_rational<NT1, NT2, Cmp> &a,
+          const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+  return b < a;
+}
+
+template < typename NT1, typename NT2, typename Cmp >
+bool
+operator<=(const Filtered_rational<NT1, NT2, Cmp> &a,
+           const Filtered_rational<NT1, NT2, Cmp> &b)
+{
+  if(a.n1().sup() <= b.n1().inf()){
+    return true;
+  }
+  if(a.n1().inf() >= b.n1.sup){
+    return false;
+  }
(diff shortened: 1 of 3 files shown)


mglisse:
Ah, yes, that's a good idea, I did think about it when I noticed how much time a filtered rational kernel spends converting the same rational to an interval again and again.

sloriot:
TODO: Once all tests pass, add specialization of the construction wrapper to use approx input when possible (to avoid non needed conversion to interval).

afabri:
> 
> 
> TODO: Once all tests pass, add specialization of the construction wrapper to use approx input when possible (to avoid non needed conversion to interval).

After the exact construction we create the approximation.  However, we have "constructions" such as  `Construct_source_2(Segment_2)` where we can just take the approximation of the point instead of regenerating it again from the exact result. Question:  Isn't it enough to distinguish between result types that are copies and result types that are references?

MaelRL:
@sloriot @afabri 

2490190 and feae9aa need reviewing. 

Former is to get `CGAL::Iso_rectangle_2<FRK>::Rep::first_type` and `CGAL::Getter` to play nice together.

Latter is because `const CGAL::Getter<Args>::first_type&...` gives a `const&` and then it doesn't match the specific overloads for `result_of`, for example for `Construct_vertex_2` which by default has result type `const Point_2&`, but if matched with `(Iso_rectangle_2, int)`, it becomes simply `Point_2`. So, `result_of<CV2(const Iso_rectangle_2&, int)> won't see `Point_2` but `const Point_2&` and then you have `const&` of temporaries.

afabri:
As we do constructions always with the exact time, we get lots of mallocs and reallocs of `Gmpq`.  @mglisse  I saw this [newsgroup entry](https://gmplib.org/list-archives/gmp-discuss/2012-March/004950.html)  and I am wondering if you still have your trivial allocator.

mglisse:
> As we do constructions always with the exact time, we get lots of mallocs and reallocs of `Gmpq`. 

Did you try the latest version of GMP? Did you try mpq_class (or boost) instead? It might reduce a bit the overhead (no miracle though).

> @mglisse I saw this [newsgroup entry](https://gmplib.org/list-archives/gmp-discuss/2012-March/004950.html) and I am wondering if you still have your trivial allocator.

What "trivial allocator" are you talking about? The pool mentioned in the linked message sounds similar to the old branch cgal-public-dev/STL_Extension-recycle_Handle_for-glisse .

mglisse:
One point about static filters (I don't know if this is the right place to discuss it, but it is relevant to this branch). The way they are engineered is, we store for each number a bound on the absolute value of the number and a bound epsilon on the absolute error. Then we propagate this and get a bound on the final error. One important aspect is how we initialize the error. We can assume that the input doubles are exact, which leads to a small epsilon, but can only be used if `fit_in_double`. Or we can assume that the input double has an error smaller than 1 ulp (.5 for some types) and get a larger epsilon (fails more often), but be able to use the filters in more cases. In particular, here, if we do constructions with mpq_t and create an interval from that, most intervals should have length 1ulp and thus be good candidates for the relaxed static filters.

afabri:
Looking at cgal-public-dev/STL_Extension-recycle_Handle_for-glisse   I am wondering if we do not speak about different things here.   I thought one would have to use a pool on a lower level, namely [Custom Allocation](https://gmplib.org/manual/Custom-Allocation.html).   Your goal is to avoid deleting reps, right?

mglisse:
> Looking at cgal-public-dev/STL_Extension-recycle_Handle_for-glisse I am wondering if we do not speak about different things here. I thought one would have to use a pool on a lower level, namely [Custom Allocation](https://gmplib.org/manual/Custom-Allocation.html).

I read the mailing list discussion you linked to in the previous message, and I don't see any mention of an allocator there, so I am confused... You can pass your own allocation function to GMP, I probably tried it at some point but I don't remember it.

> Your goal is to avoid deleting reps, right?

That branch never destructs any Gmpq, they get reused, which limits allocations a lot.

afabri:
The `Filtered_rational_kernel`  is used as `Exact_predicates_exact_constructions_kernel` when `CGAL_USE_FILTERED_RATIONAL_KERNEL` is defined. 
I compared the performance of three programs
- Polygon_mesh_processing/examples/Polygon_mesh_processin/corefinement_mesh_union_with_evpm.cpp
  takes 6.9 sec instead of 5.1 sec
- Boolean_set_operations/benchmark/Boolean_set_operations/polygon_set_2_join.cpp takes 17 sec instead of 7 sec.
- A 3D Delaunay triangulation with Epeck takes 18 sec instead of 11 sec

mglisse:
The time for 3D Delaunay is surprising. There is no construction involved, this version of Epeck should do pretty much the same computations as the lazy Epeck. The memory is arranged slightly less efficiently (intervals and Gmpq interspersed), but I wouldn't expect it to have such a large effect. Or is the extra time spent in the initial conversion from double to Gmpq to construct the points, where the lazy kernel only constructs a node with the double, while the actual Delaunay computation time remains the same?

afabri:
Everything becomes slighlty slower. For example the fit_in_double inside the static filter runs on an interval and has to determine again and again that inf == sup , the hilbert sort compares intervals, std::nth, and even the assignment of a point when std::swapped while doing the std::random_shuffle.

afabri:
We probably should focus on examples where we really make constructions.

mglisse:
> the fit_in_double inside the static filter runs on an interval and has to determine again and again that inf == sup

How does the lazy kernel avoid doing the same thing? fit_in_double(Lazy_exact_nt) calls fit_in_double(Interval_nt).

> the hilbert sort compares intervals, std::nth,

What does the lazy kernel do differently there?

> and even the assignment of a point when std::swapped while doing the std::random_shuffle.

Ok, you assign 3 Handle_for and 3 Interval_nt (1 per coordinate) instead of 1 Handle (for the whole point), although you could use Cartesian instead of Simple_cartesian if that's what you want. On the other hand, I thought that the goal of this kernel would be thread-safety, so even using Gmpq (as opposed to mpq_class or mpq_rational) is strange, but those would be way more costly to copy.

NB: std::swap doesn't seem to give very good code for Gmpq or lazy point, we may want to do something about it at some point.

afabri:
I looked again at the 3D Delaunay with the `Lazy_kernel` vs the `Filtered_rational_kernel`. The former does not construct a single `Gmpq`. The latter constructs them and then in the static filter when calling  `p .x()` every time makes a pair of an interval and a copy of a `Gmpq` (though just a handle).

afabri:
@sloriot and @MaelRL could you please have a look at `Coercion_traits` etc.

Number_types/include/CGAL/Filtered_rational.h
```diff
@@ -414,17 +414,17 @@ public :
     }
 };
 
-namespace NTC_INTERN{
+namespace FR_INTERN{
```
lrineau: Would you mind respecting the naming guideliens while you are at renaming the namespace?

mglisse:
> calling `p .x()` every time makes a pair of an interval and a copy of a `Gmpq`

Ah, I hadn't noticed yet that you used a pair<approx_point,exact_point> as point, and not an array of 3 pair<interval,Gmpq>.

afabri:
@sloriot could you have a look at [this line in travis](https://travis-ci.org/CGAL/cgal/jobs/650455544#L3648).

afabri:
> 
> 
> I looked again at the 3D Delaunay with the `Lazy_kernel` vs the `Filtered_rational_kernel`. The former does not construct a single `Gmpq`. The latter constructs them and then in the static filter when calling `p .x()` every time makes a pair of an interval and a copy of a `Gmpq` (though just a handle).

What about having a function `fit_in_double_x(const Point

> 
> 
> > calling `p .x()` every time makes a pair of an interval and a copy of a `Gmpq`
> 
> Ah, I hadn't noticed yet that you used a pair<approx_point,exact_point> as point, and not an array of 3 pair<interval,Gmpq>.

@lrineau why don't we use `get_approx(p)`  in the static filter of `Orientation_3`  as we do in [`Equal_2.h`](Filtered_kernel-Filtered_rational_kernel-GF)?  Wouldn't that also be better for `Epeck`?

Filtered_kernel/include/CGAL/internal/Static_filters/Angle_3.h
```diff
@@ -71,16 +71,14 @@ class Angle_3
   {
     CGAL_BRANCH_PROFILER_3("semi-static failures/attempts/calls to   : Angle_3", tmp);
 
-    Get_approx<Point_3> get_approx; // Identity functor for all points
-                                    // but lazy points
```
lrineau: Should not you also remove the definition of `Get_approx` in `./tools.h`?

afabri:
> > 
> 
> @lrineau why don't we use `get_approx(p)` in the static filter of `Orientation_3` as we do in [`Equal_2.h`](Filtered_kernel-Filtered_rational_kernel-GF)? Wouldn't that also be better for `Epeck`?

In fact it turns out that the `Get_approx` was a way to avoid constructing a `Lazy_exacxt_nt` when `p.x()` gets called., but it was superseded by the usage of `Static_filtered_predicate`.  However we still had lot's of `Get_approx`  because only one commit that had added them was reverted.  In this PR we remove them from all the static filters.

Filtered_kernel/include/CGAL/internal/Static_filters/Angle_3.h
afabri: No, as it is still used in `Triangulation_2/3` and `Robust_intersection_traits_3`.

afabri:
Here are the latest timings with each time two runs:
Delaunay_3
|V| = 1000000 |C| = 6748016

Epick
8.353 sec
8.28 sec

Epeck - Lazy_exact_nt<Gmpq>
11.295 sec
11.176 sec

Epeck - Lazy_exact_nt<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational,1> >
11.201 sec
11.824 sec

Epeck - Filtered_rational<Interval_nt<0>,Gmpq>
13.48 sec
13.872 sec

Epeck - Filtered_rational<Interval_nt<0>,boost::multiprecision::number<boost::multiprecision::backends::gmp_rational,1> >
14.306 sec
14.408 sec

afabri:
image

`Filtered_rational_kernel` creates in any case a GMP rational number, which explains the runtime difference, which is essentially the malloc and free.   We could make the generation of the rationals lazy, but that is probably not worth it,  Can we gain sonething on the swaps?

mglisse:
`dt.insert(std::make_move_iterator(points.begin()), std::make_move_iterator(points.end()));` may remove a few allocations.

afabri:
My points now only store a default constructed rational.  And the rational is constructed from the interval (with `inf==sup`) if needed.  @mglisse do you have any idea how to avoid the `mpq_init`, that to even initialize on an is needed base.

mglisse:
I though the point of this kernel was to avoid mutable lazy stuff that is thread-unsafe?
If you have a bool to indicate if the exact part is initialized, this is manually reimplementing optional<exact_part>, which can handle not initializing just fine. If you want to save the space of the bool, I guess you could replace the exact object with some suitably aligned buffer, memset(0) it when you don't want it, test if the _mp_d of _mp_den is 0 (this part is not exactly casher), use placement new to construct an exact object when needed, explicitly copy/destruct in constructors/destructors.

afabri:
Looking at the assignment operator of gmp_rational

`gmp_rational& operator = (const gmp_rational& o)`
  ` {`
     ` if(m_data[0]._mp_den._mp_d == 0)`
        `mpq_init(m_data);`
      `mpq_set(m_data, o.m_data);`
     ` return *this;`
`   }`

why is that multi threading safe?  As `mpq_init`  changes `m_data[0]._mp_den._mp_d`  to nonzero, can't that pose a problem @mglisse ?

mglisse:
> why is that multi threading safe? As `mpq_init` changes `m_data[0]._mp_den._mp_d` to nonzero, can't that pose a problem @mglisse ?

You are saying that `x=y` modifies x, which is true, but not really an issue. There are several levels of safety. Concurrent data-structures want to allow to threads to modify the structure at the same time. Simple thread-safety lets several threads read a variable at the same time, but if you are modifying a variable, reading it at the same time in another thread is unsafe. The issue with lazy* is that reading is actually a mutating operation.

afabri:
Stupid me.  By the way do you have any idea when `m_data[0]._mp_den._mp_d` equals zero?  Reading the code I had the impression that it is always initialized.

mglisse:
> Stupid me. By the way do you have any idea when `m_data[0]._mp_den._mp_d` equals zero? Reading the code I had the impression that it is always initialized.

After a move constructor maybe?

Number_types/include/CGAL/boost_mp.h
```diff
@@ -20,7 +20,7 @@
 // easy solution.
 // TODO: MSVC has trouble with versions <= 1.69, reenable once 1.70 has been
 // tested. https://github.com/boostorg/multiprecision/issues/98
-#if !defined CGAL_DISABLE_GMP && !defined CGAL_DO_NOT_USE_BOOST_MP && BOOST_VERSION >= 106300 && !defined _MSC_VER
+#if !defined CGAL_DISABLE_GMP && !defined CGAL_DO_NOT_USE_BOOST_MP && BOOST_VERSION >= 106300 // && !defined _MSC_VER
```
lrineau, line 23: Are we sure of that change?!

Polygon_mesh_processing/examples/Polygon_mesh_processing/corefinement_mesh_union.cpp
```diff
@@ -1,11 +1,11 @@
-#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
 #include <CGAL/Surface_mesh.h>
 
 #include <CGAL/Polygon_mesh_processing/corefinement.h>
 
 #include <fstream>
 
-typedef CGAL::Exact_predicates_inexact_constructions_kernel K;
+typedef CGAL::Exact_predicates_exact_constructions_kernel K;
```
lrineau, line 8: Is that a change you want to push in CGAL?

lrineau:
Seeing the `DO NOT PUSH TO MASTER`, in last commit (https://github.com/CGAL/cgal/pull/4495/commits/8e89bfcb858383522decbfc2861b78180d0103b1), I have converted the PR to a draft.

MaelRL:
Rough benchmark: computing the DT3 of 1000000 random 3D points; average of at least 5 runs.

|   | EPECK | FRK   | FRK Lazy, no mutex member  | FRK Lazy + mutex member  |
|---|-----|---|---|---|
| Seq  |  6.2    | 8.00  | 7.5  | 7.75  |
| Par  |   2.05    | 3  | 3.15  |  3.35 |

Some remarks:
- DT3 does not involve any construction, so exact values never need be built and predicates of EPECK and FRK both just call EPICK. Nevertheless, EPECK is 25% faster than FRK.
- Since there is no construction, the only construction that matters is at the very beginning, but it's negligible (hence very little time difference between FRK and FRK lazy)
- There is a hidden construction in parallel T3: finding the index of the cell to lock, this requires computex_x/y/z_3 and kills FRK's runtime because it requires the construction of the exact value for the points, every time. Going around that code (which seems a bit broken too) with:
```
-    int index_z = static_cast<int>( (CGAL::to_double(point.z()) - m_bbox.zmin()) * m_resolution_z);
+    int index_z = static_cast<int>( (CGAL::to_double(CGAL::approx(point).z()) - m_bbox.zmin()) * m_resolution_z);
```
- Strange (and repeatable) time differences when mutexes are present or not, even though we never use them.

sloriot:
@gdamiand I'd be interested to know what this Kernel is doing in your application that has constructions and that can run in parallel. You have to use something like `mp_class` as NT to be thread-safe.

afabri:
In order to use FRK in parallel we have to use `boost::mutiprecision`.   @mglisse do you know how things evolved around your [issue](https://github.com/boostorg/multiprecision/issues/98) mentioned in a todo in [`CGAL/boost_mp.h`](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/boost_mp.h#L21)

sloriot:
or as I said `mpq_class` on supported platforms.

mglisse:
The comment says to re-enable (I guess that is changing `!defined _MSC_VER` to `!defined _MSC_VER || BOOST_VERSION >= 107000`) once someone has tested that it works with boost 1.70 and MSVC, so maybe someone working on windows can do that...

afabri:
What exactly do I have to test?  The Kernel_23 testsuite?

gdamiand:
> @gdamiand I'd be interested to know what this Kernel is doing in your application that has constructions and that can run in parallel. You have to use something like `mpq_class` as NT to be thread-safe.

@sloriot I will try; but not now (dead line in 1 week).
If the meantime, you can try this simple code which construct exact number in parallel

https://gist.github.com/gdamiand/530700bb8640ecf6a502b312a7fccab3

mglisse:
> What exactly do I have to test? The Kernel_23 testsuite?

Ideally the whole testsuite. But according to https://github.com/CGAL/cgal/pull/3406#issuecomment-444091335 the issue was visible in Circular_kernel_3.

gdamiand:
> @gdamiand I'd be interested to know what this Kernel is doing in your application that has constructions and that can run in parallel. You have to use something like `mpq_class` as NT to be thread-safe.

@sloriot  I tried; I have compiler errors when defining typedef CGAL::Filtered_kernel<mpq_class> MyKernel

Could you give me the kernel type to define please?

MaelRL:
```
typedef CGAL::Simple_cartesian<CGAL::Interval_nt<false> >                    FRK_IA;
typedef CGAL::Simple_cartesian<internal::Exact_field_selector<double>::Type> FRK_EA;
typedef CGAL::Filtered_rational_kernel<FRK_IA, FRK_EA>                       FRK;
```

gdamiand:
Thanks @MaelRL .

MaelRL:
@gdamiand have you had time to look into it?

gdamiand:
> @gdamiand have you had time to look into it?

I tried, but I have compiler errors with my prog since I use #4519.  I need to merge both pr locally.

gdamiand:
> > @gdamiand have you had time to look into it?
> 
> I tried, but I have compiler errors with my prog since I use #4519. I need to merge both pr locally.

When adding the two for_compact_container() methods in Handle.h and updating Lazy.h does not solve my error. 

@afabri , @lrineau or @MaelRL Can you suggest me what to modify (only locally) to be able to compile my prog and test this kernel ? Thanks.
@lrineau

lrineau:
I have merged `master` and (thus PR #4519) in this PR. That way you have only one PR to merge.

gdamiand:
Thanks @lrineau , but this does not solve my errors (I already tested these modifications locally before your merge).

const class CGAL::Segment_2<CGAL::Filtered_rational_kernel<CGAL::Simple_cartesian<CGAL::Interval_nt<false> >, CGAL::Simple_cartesian<__gmp_expr<__mpq_struct [1], __mpq_struct [1]> > > >’ has no member named ‘for_compact_container’

lrineau:
Anyway, this PR "Add a kernel that stores objects twice: with Intervals and Gmpq #4495" has compilation issues, shown by Travis. They should be fixed, first.

gdamiand:
Ok I will wait. Thanks.

MaelRL:
I have removed the hack from T3 which travis was complaining about, but it's inconsequential to what you're trying to do so nothing will change.

Even when FRK is lazy (which you have to enable with the macro `CGAL_LAZY_FILTERED_RATIONAL_KERNEL`), it doesn't use `Handle.h` so it doesn't have the `for_compact_container()` that is required from a type `T` used in a compact container.

Adding [`Compact_container_base`](https://doc.cgal.org/latest/STL_Extension/classCGAL_1_1Compact__container__base.html) as a base of FRK's objects will solve that:
```
--- a/Filtered_kernel/include/CGAL/Filtered_rational_kernel.h
+++ b/Filtered_kernel/include/CGAL/Filtered_rational_kernel.h
@@ -18,6 +18,7 @@
 #include <CGAL/Filtered_rational.h>
 
 #include <CGAL/Cartesian_converter.h>
+#include <CGAL/Compact_container.h>
 #include <CGAL/Filtered_kernel/Cartesian_coordinate_iterator_2.h>
 #include <CGAL/Filtered_kernel/Cartesian_coordinate_iterator_3.h>
 #include <CGAL/intersections.h>
Stage this hunk [y,n,q,a,d,j,J,g,/,e,?]? y
@@ -57,7 +58,7 @@ struct Infimum_to_double
 
 template <typename T1, typename T2>
 struct Approximate_exact_pair
-  : public std::pair<T1, T2>
+  : public std::pair<T1, T2>, public CGAL::Compact_container_base
 {
   typedef Approximate_exact_pair<T1, T2>          Self;
   typedef std::pair<T1, T2>                       Base;

```
but I don't know if it's the best way to go at it as I am not super familiar with the compact container inner workings.

gdamiand:
Thanks @MaelRL , will try.

gdamiand:
First (preliminary tests) with my program that computes arrangement of segments, on two datasets:

=======================================
File1:

EPECK: 9.61331 seconds
FRK: 68.5651 seconds

Last step:
sequential: 10,7 seconds
in parallel: 7,3962 seconds


File2:

EPECK: 53.8384 seconds
FRK: 302.304 seconds

Last step: 
sequential: 47,622 seconds
in parallel: 38,873 seconds
====================================

My algorithm has 3 steps. In this test, I only parallelized the last step. It works with FRK while I have a core dump with EPECK in the destructor (corrupted double-linked list).

I can try more tests if you want; and I can try to re-add the parallel version for my second step if you think it is interesting.

Note that cgal arrangement does not compile with this kernel, nor my basic viewer.

Hope this help.

MaelRL:
FRK might have such a bad runtime because of `Compute_x/y/z` and similar constructions.

gdamiand:
I mainly use the following predicates:
* do_intersect and intersection (between pair of segments, and one segment and one rectangle)
* has_on_unbounded_side, bounded_side, has_on, y_at_x

mglisse:
> FRK might have such a bad runtime because of `Compute_x/y/z` and similar constructions.

Couldn't that return a proxy instead of the heavy object?

But in any case, FRK is supposed to be significantly slower than Epeck, that was the whole point of the lazy kernel...

> Adding [`Compact_container_base`](https://doc.cgal.org/latest/STL_Extension/classCGAL_1_1Compact__container__base.html) as a base of FRK's objects will solve that:

If we start adding various things in Compact containers, I guess an FRK segment should look for a pointer in the FRK point, which should look in the exact point, which should look in its first coordinate, which should look in its numerator, etc, stopping whenever there is already a pointer available.

MaelRL:
> > FRK might have such a bad runtime because of `Compute_x/y/z` and similar constructions.
> 
> Couldn't that return a proxy instead of the heavy object?

I don't really see how I could return a proxy. With the current definition, `FRK::Point` is `pair<Ak::Point, EK::Point>` and `FRK::FT` is `pair<AK::FT, EK::FT>`, so when you call `Compute_x_3` you are pretty much forced to construct the pair. When this type of function is called within a predicate or a "bigger" construction, things are however fine because predicates/constructions are filtered hence you first extract `AK/EK::Point`. However, raw calls to `Compute_x/y/z_3` and similar normally cheap constructions cost here a lot.

> 
> But in any case, FRK is supposed to be significantly slower than Epeck, that was the whole point of the lazy kernel...
> 
> > Adding [`Compact_container_base`](https://doc.cgal.org/latest/STL_Extension/classCGAL_1_1Compact__container__base.html) as a base of FRK's objects will solve that:
> 
> If we start adding various things in Compact containers, I guess an FRK segment should look for a pointer in the FRK point, which should look in the exact point, which should look in its first coordinate, which should look in its numerator, etc, stopping whenever there is already a pointer available (and avoiding shared objects?).

I see, thanks.

mglisse:
> I don't really see how I could return a proxy.

You could return an object that contains just pointers to the approximate and exact numbers, and has a conversion operator to a true FT. If this object provides exact, to_interval, etc, in some cases you may avoid the conversion to FT (in less lucky cases you may end up converting several times to FT...). It depends how the result of Compute_x_3 is used.

MaelRL:
I thought about this, but then you need a mechanism to know whether your pointers are still valid because you could have a function that constructs a point and returns its `x.()`. Maybe you can still track this, but it rather feels like the design of this kernel objects as `pair<ak, ek>` was just too flawed from the start.

mglisse:
> I thought about this, but then you need a mechanism to know whether your pointers are still valid because you could have a function that constructs a point and returns its `x.()`.

Same for kernels where x() returns a reference?

MaelRL:
Sure, if that function returns a reference; but if you return a copy, you are still going to have a pair with a pointer to a deleted object whereas it is safe for other kernels.

mglisse:
So the case that works with references but not this shallow pair is
```c++
auto x = get_temporary_point().x();
```
If I write FT instead of auto, it is safe. If I write auto const& the reference case is broken as well.

MaelRL:
My point was something like:
```
FT compute_x_for_some_reason() { 
  Point p = construct_some_point();
  return p.x();
}
```
If you have a shallow pair, your FT will be pointing to a deleted object, even with the return by copy.

mglisse:
Since the return type is explicitly FT, which is a heavy pair, the shallow pair returned by p.x() gets immediately converted to a heavy pair. It doesn't save anything, but it seems safe to me. The issue would be if the function returned auto (or decltype(auto)).

Issue #34, opened 2015-04-01
Move enable CGAL types


bo0ts:
When working with CGAL in C++11 and upwards one of the biggest issues is that most types cannot be efficiently moved even if they should. Users end up writing code that works around that instead of using the straight forward approach. This primarily affects types which carry large geometric structures.

A non-exhaustive list:
- [x] Compact_container and Concurrent_compact_container (PR #4496)
- [x] Surface_mesh
- [ ] HDS and Polyhedron_3: movability depends on underlying storage
- [x] Triangulation_data_structure_2 and Triangulation_2: depends on Compact_container (PR #4496)
- [x] Triangulation_data_structure and Triangulation: depends on Compact_container (PR #4496)

lrineau:
Who could spend time on that task, among the @CGAL/developers ?

efifogel:
I'm not volunteering my self, but could you give a specific example.


   ____  _        ____             _
  /_____/_) o    /__________  __  //
 (____ (   (    (    (_/ (_/-(-'_(/
                         _/

bo0ts:
@efifogel Polygon_2 is a very simple example: it is templated by the container. All possible containers support move construction. For some unknown reason a dubious copy-constructor is provided, which prevents generation of a move constructor etc. Fixing this would make it possible to cheaply pass a Polygon_2 by value.

lrineau:
Everybody should know something about the C++ Rule of 3 (in C++98), that became Rule of 5 or Rule of 0, in C++11:
- Here is a simplified reference, in Wikipedia: [Rule of three (C++ programming](https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29).
- For better examples, but a longer explanation: http://en.cppreference.com/w/cpp/language/rule_of_three
- if you want more, the Wikipedia page has the useful reference.

I personally prefer the Rule of 0: do not deal with resources yourself, but we could apply the Rule of 5 for our own containers.

bo0ts:
One problem is that we need to identify classes with unnecessary
constructors and sometimes even destructors. They didn't do much harm in
C++03 but are actively harmful in C++11.

I guess one starting point is to raise awareness and fix it as we go along.

lrineau:
Le Friday 10 April 2015 05:34:23 Philipp Moeller a écrit :
> One problem is that we need to identify classes with unnecessary
> constructors and sometimes even destructors. They didn't do much harm in
> C++03 but are actively harmful in C++11.
> 
[...]

PR #4496, opened 2020-01-29
Add move-semantic to Compact_container and Triangulations

lrineau:
## Summary of Changes

Add move-semantic to:
  - `Compact_container` and `Concurrent_compact_container`,
  - all triangulation classes in Triangulation_2,
  - all triangulation classes in Triangulation_3.

## Release Management

* Affected package(s): Triangulation_2, Triangulation_3
* Issue(s) solved (if any): partially fix of #34


lrineau:
@mglisse Do you mind having a look at what I did?

How can we actually check, in a test, that the move-semantic is correctly implemented?

I plan to implement move-semantic for 2D and 3D triangulations (by the rule-of-zero, if possible), and then bench the reallocation of a vector of CDT_2 (my actual need). Can I do better tests than benchmarks?

mglisse:
Looks good to me. I ignored the Time_stamper thing, let me know if that was supposed to be the key thing to look at. Implementing everything in terms of swap is a bit out of fashion, but remains very convenient. Sometimes people put all the members in a simple base class (without constructors), which avoids writing 10 lines of std::swap in a row, you can just use the default std::swap on the base, or in a move constructor you can use the base move, then fix up what the base thing did wrong (call init() on the argument?). But that doesn't change much.

I don't have particularly good techniques to recommend for testing. One important point is to decide how much guarantee you want to provide on moved-from objects. Should I be able to use the moved-from Compact_container like a default-initialized one? Or can I only destruct it or use it in assignments? That also guides what you want to test. If being able to use it efficiently in a vector is all you want to guarantee, then...

lrineau:
Thanks @mglisse for you feedback.

I have amended my commit (now https://github.com/CGAL/cgal/pull/4496/commits/e82ea5de96bc1beb186449cff4998ebe7cfa5b2b) with an explanation about the time stamper stuff. In short, I needed to rewrite its implementation to allow the `noexcept` of the move operations (cstr and assignment).

I am willing to do better than copy and swap. What would you suggest? ... Actually, I have a proposal: I can delegate the move-constructor and move-assignment operator to the copy-constructor and copy-assignment operator, and then call `init()` on the moved-from object. Would it be better?

As for moved-from objects, they are supposed to be valid objects, possibly empty. It seems to mean that one should be able to call any operations on a moved-from objects, as soon as the operation does not have preconditions. For containers and triangulations, I think that means the moved-from objects should behave like an empty container or triangulation.

In my last version of the PR, commit https://github.com/CGAL/cgal/pull/4496/commits/290c3a2011c08a7a9b8f18a0880b1828fdb8001e, note that I know that a moved-from `Triangulation_hierarchy_3` is invalid: one can only call the destructor on it, and maybe the assignment operators. That was to preserve the `noexcept`. I think one can rewrite that class so that its default constructor does not allocate on the heap. And then a moved-from object from be equivalent to the default-constructed one.

Please give me feedbacks. I value them.

mglisse:
> I am willing to do better than copy and swap. What would you suggest?

I am fine with copy and swap, it is quite simple, and if you use the same idiom everywhere, it makes it easy to follow. In most cases, it should generate code (almost) as good as something more specific. With the atomic, there is probably a little overhead, for instance if you swap 2 atomics a and b twice in a row, I don't think gcc simplifies that to nothing, but I strongly doubt it matters to your overall performance.

> ... Actually, I have a proposal: I can delegate the move-constructor and move-assignment operator to the copy-constructor and copy-assignment operator, and then call `init()` on the moved-from object. Would it be better?

Uh, which copy-constructor are you talking about? If it is the one from Compact_container that does a deep copy, that's counter-productive. You would have to create a base class with dumb constructors for that (as mentioned in my previous message).

> As for moved-from objects, they are supposed to be valid objects, possibly empty. It seems to mean that one should be able to call any operations on a moved-from objects, as soon as the operation does not have preconditions. For containers and triangulations, I think that means the moved-from objects should behave like an empty container or triangulation.

Then maybe you can test that. Pick some arbitrary existing test that tests 2 triangulations, and instead of constructing a second one, move the first on the side to tmp and reuse it for the second one. Then maybe in the middle of the second one, delete tmp, and keep using the triangulation. That's going to look a bit messy though. Usually I would just be lazy and not test very extensively...

> In my last version of the PR, commit [290c3a2](https://github.com/CGAL/cgal/commit/290c3a2011c08a7a9b8f18a0880b1828fdb8001e), note that I know that a moved-from `Triangulation_hierarchy_3` is invalid: one can only call the destructor on it, and maybe the assignment operators. That was to preserve the `noexcept`. I think one can rewrite that class so that its default constructor does not allocate on the heap. And then a moved-from object from be equivalent to the default-constructed one.

For current libstdc++, you can cheat and specialize std::__is_bitwise_relocatable (no promise that it will stay there) to let the compiler know that a move construction immediately followed by a destruction of the source is equivalent to memcpy, specifically to optimize std::vector. It is done for std::deque, which has the same issue of always allocating. Some people are trying to introduce things in the standard to help with it, but it isn't for soon...
You aren't the only one restricting what can be done with moved-from objects. While the standard requires them to be fully valid objects if you are going to use them with any function in the standard library, I think boost.multiprecision crashes if you add 2 moved-from numbers.
Removing the invariant of always having something allocated would indeed make the situation simpler.

lrineau:
I think I have finished the pull-request, now: I have implemented move-semantic for all of Triangulation_2 and Triangulation_3, as far I as know. I have left aside other triangulations in CGAL (like alpha shapes, periodic triangulations...). We hope to have a GSoC internship to continue the work.

I have tried to avoid swap as much as possible. So far, I think only `Compact_container` and `Concurrent_compact_container` implements move-semantic using swap. I have also tried to `=default` the special members when possible, and to fix a few warnings issues by clang-tidy (about unnecessary declaration of copy-constructors).

The testsuite will probably pass, because the testsuite so far does not test much of the move-semantic. I'll try to extend the testsuite like you propose in comment https://github.com/CGAL/cgal/pull/4496#issuecomment-580741236, but later. Anyway I will probably discover bugs as soon as I try the move-semantic in my own code.

maxGimeno:
[Errors](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-72.shtml#Advancing_front_surface_reconstruction) on windows

lrineau:
I hope that should be fixed, now.

STL_Extension/include/CGAL/Compact_container.h
```diff
@@ -328,17 +339,8 @@ class Compact_container
...
-    std::swap(alloc, c.alloc);
-    std::swap(capacity_, c.capacity_);
-    std::swap(size_, c.size_);
-    std::swap(block_size, c.block_size);
-    std::swap(first_item, c.first_item);
-    std::swap(last_item, c.last_item);
-    std::swap(free_list, c.free_list);
-    all_items.swap(c.all_items);
-    std::swap(time_stamper, c.time_stamper);
+  friend void swap(Compact_container& a, Compact_container b) {
```
mglisse: `noexcept` as well?

lrineau: Yes, it is `noexcept`. I have not tried to add `noexcept`-specifications on all functions, but only on move-constructors and move-assignments. In the end, I no longer use swap for the move-semantic, so there is not need to add `noexcept` on `swap` methods or `swap` functions.

I know the performance benefits of `noexcept` move-constructors and move-assignements. What are the benefits of `noexcept` swap functions?

mglisse: > What are the benefits of `noexcept` swap functions?

They make is_nothrow_swappable_v true? They probably had a reason to create that trait, and I could imagine using it for optimizations similar to noexcept move, in particular when move construction is not noexcept (like std::list with msvc or std::deque with libstdc++), but in practice libstdc++ only uses that trait so it can mark other swap functions noexcept.
It just feels strange that std::swap is noexcept for that type (based on moves) but the swap found by ADL is not.

lrineau: There is a really good reason to create `std::is_nothrow_swappable`, and it is explained in [the "motivation" section of N3619](https://wg21.link/n3619#motivation). But, as you say, it seems to be designed to be used to write `noexcept` specifications for swap functions using other swap functions.

I agree that might be nice to have `noexcept` specifications for more functions in CGAL, but without the immediate performance benefit, I am not willing to do it. Even for adding `noexcept` to `Compact_container::swap`, I have to verify if `std::vector` is nothrow_swappable (it is), and for the concurrent compact container, I do not know if the TBB vector is nothrow swappable. So... I postpone...

mglisse: > I no longer use swap for the move-semantic

You do, the move assignment and move constructor both use swap (the member function).

lrineau: Oh, you are right! For `Compact_container` and `Concurrent_compact_container`, I lazily still rely on `swap`... :cry:

mglisse: > I lazily still rely on `swap`... :cry:

There's nothing wrong with that...

maxGimeno:
[Errors on windows](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-77.shtml#Advancing_front_surface_reconstruction)

lrineau:
Fix by 0c40f7a458e. Verified on blake.

maxGimeno:
C:\CGAL\test\CGAL-5.1-Ic-79\cmake\platforms\x64_Cygwin-Windows10_MSVC2017-Debug-64bits\test\Triangulation_2\include\CGAL/_test_cls_const_Del_triangulation_2.h(33): error C2338: move cstr is missing

maxGimeno:
In the testsuite : 
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-79.shtml#BGL 
and https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-79.shtml#Triangulation_2

lrineau:
Both issues will be fixed by https://github.com/CGAL/cgal/pull/4496/commits/a060fb639644ca2742e68b832545c140d03f89e9.

maxGimeno:
Errors on Blake : https://cgal.geometryfactory.com/CGAL/testsuite/#Skin_surface_3

lrineau:
Fixed by https://github.com/CGAL/cgal/pull/4496/commits/e3d7f7946db42f52317b23fcc150a39dcde76e53.

maxGimeno:
I suspect this PR causes the errors in MSVC2015 in [Ic-86](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-86.shtml)

lrineau:
As I was tired of compilation/executions issues with MSVC, I have eventually changed a lot:
  - now the pair of function `for_compact_container` are a regular pair of getter/setter, instead of returning a const/non-const pointer to `void`,
  - `CC_iterator` no longer use a union to cast a pointer to a pointer to `void`, but a plain regular cast, and
  - `CC_iterator` now uses the rule-of-zero instead of the rule-of-five.

That will break the source-compatibility for any user implementing her own type of vertex/face/cell, without deriving from one of CGAL's.

Tested locally on Linux:
```
Combinatorial_map_Examples             =   5.75 sec*proc (16 tests)
Combinatorial_map_Tests                =  28.56 sec*proc (8 tests)
Installation                           =   0.68 sec*proc (1 test)
Linear_cell_complex_Demo               =   5.83 sec*proc (1 test)
Linear_cell_complex_Examples           =  15.35 sec*proc (24 tests)
Linear_cell_complex_Tests              =  87.74 sec*proc (10 tests)
Mesh_2_Demo                            =   2.22 sec*proc (2 tests)
Mesh_2_Examples                        =  13.90 sec*proc (12 tests)
Mesh_2_Tests                           =  21.86 sec*proc (24 tests)
STL_Extension_Examples                 =   4.17 sec*proc (14 tests)
STL_Extension_Tests                    =  22.70 sec*proc (54 tests)
TDS_2_Tests                            =   0.73 sec*proc (4 tests)
TDS_3_Examples                         =   1.62 sec*proc (6 tests)
TDS_3_Tests                            =   0.88 sec*proc (4 tests)
Triangulation_2_Demo                   =  12.52 sec*proc (3 tests)
Triangulation_2_Examples               =  18.93 sec*proc (44 tests)
Triangulation_2_Tests                  =  23.59 sec*proc (42 tests)
Triangulation_3_Demo                   =   2.97 sec*proc (1 test)
Triangulation_3_Examples               =  25.17 sec*proc (38 tests)
Triangulation_3_Geomview_demos_Demo    =   4.00 sec*proc (4 tests)
Triangulation_3_Tests                  = 169.17 sec*proc (28 tests)
```

TODO: should I need a small feature?

sloriot:
break at compile time?

lrineau:
> break at compile time?

Yes, that will break at compile-time.

gdamiand:
@lrineau Do you have the same running times for the tests with the current version (to compare) ?

maxGimeno:

lrineau:
I have added "not yet approuved" because we need benchmarks.

lrineau:
@maxGimeno Would you mind benchmarking this PR, against `master`?

maxGimeno:
here's my quick bench : 
```bash
#!/bin/bash
cd /tmp
mkdir -p build_branch
cd build_branch
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DWITH_examples=ON -DWITH_tests=ON ~/CGAL &> /dev/null
ctest -j4 -L Triangulation -E execution___of__ &> /dev/null
cd ..
mkdir -p build_master
cd build_master
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DWITH_examples=ON -DWITH_tests=ON ~/Master_cgal &> /dev/null
ctest -j4 -L Triangulation -E execution___of__ &> /dev/null


echo "Testing PR..."
cd ../build_branch
START=$SECONDS
for t in {1..10}; do
  ctest -j4 -L Triangulation_2_ -E compilation_of__ &> /dev/null
done
RES=$(( SECONDS - START ))
echo "\n\n--------------===========================------------------\n\nTriangulation_2 tests took ${RES} seconds.\n\n--------------===========================------------------\n\n"
START=$SECONDS
ctest -j4 -L Triangulation_3_ -E compilation_of__ &> /dev/null
RES=$(( SECONDS - START ))
echo "\n\n--------------===========================------------------\n\nTriangulation_3 tests took ${RES} seconds.\n\n--------------===========================------------------\n\n"



echo "Testing Master..."
cd ../build_master
START=$SECONDS
for t in {1..10}; do
  ctest -j4 -L Triangulation_2_ -E compilation_of__ &> /dev/null
done
RES=$(( SECONDS - START ))
echo "\n\n--------------===========================------------------\n\nTriangulation_2 tests took ${RES} seconds.\n\n--------------===========================------------------\n\n"
START=$SECONDS
ctest -j4 -L Triangulation_3_ -E compilation_of__ &> /dev/null
RES=$(( SECONDS - START ))
echo "\n\n--------------===========================------------------\n\nTriangulation_3 tests took ${RES} seconds.\n\n--------------===========================------------------\n\n"
```

The results are : 
```
Testing PR...
\n\n--------------===========================------------------\n\nTriangulation_2 tests took 5 seconds.\n\n--------------===========================------------------\n\n
\n\n--------------===========================------------------\n\nTriangulation_3 tests took 7 seconds.\n\n--------------===========================------------------\n\n
Testing Master...
\n\n--------------===========================------------------\n\nTriangulation_2 tests took 6 seconds.\n\n--------------===========================------------------\n\n
\n\n--------------===========================------------------\n\nTriangulation_3 tests took 7 seconds.\n\n--------------===========================------------------\n\n
```

gdamiand:
Hum, without additional digits we can not conclude anything... 5s and 6s, this is 20% more !!!!  You also need to give results for several run (3?).

maxGimeno:
so I timed with time -f "%U"  and ran 5 times the bench. Here are the results :

```
run number 1
 ==================
Testing PR...
Triangulation_2...
Command exited with non-zero status 8
	4.59 seconds
Triangulation_3...
20.42 seconds
Testing Master...
Triangulation_2 ...
Command exited with non-zero status 8
	4.66 seconds
Triangulation_3...
18.42 seconds
run number 2
 ==================
Testing PR...
Triangulation_2...
Command exited with non-zero status 8
	4.58 seconds
Triangulation_3...
22.50 seconds
Testing Master...
Triangulation_2 ...
Command exited with non-zero status 8
	4.57 seconds
Triangulation_3...
17.07 seconds
run number 3
 ==================
Testing PR...
Triangulation_2...
Command exited with non-zero status 8
	4.52 seconds
Triangulation_3...
21.19 seconds
Testing Master...
Triangulation_2 ...
Command exited with non-zero status 8
	4.60 seconds
Triangulation_3...
19.11 seconds
run number 4
 ==================
Testing PR...
Triangulation_2...
Command exited with non-zero status 8
	4.58 seconds
Triangulation_3...
19.64 seconds
Testing Master...
Triangulation_2 ...
Command exited with non-zero status 8
	4.59 seconds
Triangulation_3...
15.57 seconds
run number 5
 ==================
Testing PR...
Triangulation_2...
Command exited with non-zero status 8
	4.68 seconds
Triangulation_3...
16.32 seconds
Testing Master...
Triangulation_2 ...
Command exited with non-zero status 8
	4.65 seconds
Triangulation_3...
17.28 seconds
```

So it seems to be longer in the branch after all.

gdamiand:
Thanks @maxGimeno . Indeed it seems longer...

lrineau:
@sloriot @afabri Can you please suggest good benchmarks for all (or most) classes of triangulations we have in CGAL?

afabri:
I may miss something here, but I do not see the need to bench all packages. The impact should be the same for 2D and 3D, and for periodic and Euclidean space,

What has to be tested?  Constructions?  Or also removal?  Or even traversal of an otherwise unmodified triangulation?  I guess that the change has no impact on traversal. 

If only constructions are impacted then just a call of a constructor with a range should do it.  

If a sequence of insertion and removal are important to test, what about the optimization step of `Mesh_3`?

lrineau:
> I may miss something here, but I do not see the need to bench all packages. The impact should be the same for 2D and 3D, and for periodic and Euclidean space,
> 
> What has to be tested? Constructions? Or also removal? Or even traversal of an otherwise unmodified triangulation? I guess that the change has no impact on traversal.
> 
> If only constructions are impacted then just a call of a constructor with a range should do it.
> 
> If a sequence of insertion and removal are important to test, what about the optimization step of `Mesh_3`?

This PR is not supposed to modify at all the performance of current CGAL code. As far as I know, there is no real code in CGAL that pass triangulation objects by copy, or copy explicitly triangulations or compact containers.

The only change that may be different for compilers is that the implementation of `CC_iterator` is different, probably simpler for compilers to optimize. And that may impact any code that deals a lot with "handles" in triangulations.

maxGimeno:

lrineau:
Should we assume that this PR was sufficiently tested and benched?

maxGimeno:
if you are satisfied with the bench results, yes

lrineau:
> if you are satisfied with the bench results, yes

How could I be satisfied? The only bench that was done is yours (https://github.com/CGAL/cgal/pull/4496#issuecomment-608337139) and it seemed to show a slowdown. 

@afabri in https://github.com/CGAL/cgal/pull/4496#issuecomment-609646740 asks for what to bench, and I do not know if any benches were actually done.

maxGimeno:
Then how could we consider it sufficently benched ?

lrineau:
> @afabri in [#4496 (comment)](https://github.com/CGAL/cgal/pull/4496#issuecomment-609646740) asks for what to bench, and I do not know if any benches were actually done.

@afabri Have your benchmarked the effects of this branch? Or should we merge it now, and see what it does?

lrineau:
I have eventually merged it. We'll see if future benchmarks show issues.

gdamiand:
No small feature for a breaking change ? Nor a message in CHANGES.md ?

lrineau:
> No small feature for a breaking change ? Nor a message in CHANGES.md ?

As far as I know, this PR should not change the behavior of existing CGAL code. It is not a breaking change. It is a big improvement, though, because it now allows, for example, to have a `std::vector` of triangulations, without caring of the cost of the reallocation of the vector.

mglisse:
> this PR should not change the behavior of existing CGAL code

I believe it is about for_compact_container. The PR touches a number of concepts, in particular.

PR #4508, opened 2020-02-07
BGL: fix doc of graph_has_property

MaelRL:
## Summary of Changes

The package BGL has documentation split between `doc/CGAL` and `include/CGAL`. The doc for `graph_has_property` was erroneously put into `include/CGAL` and thus it didn't show up in the doc.

External indices property tags are also undocumented despite being used in most `Polyhedron_3` examples, but I don't know if we really want to document them (@sloriot ?); sounds to me like it would be better to phase out `Polyhedron_3` from example.

## Release Management

* Affected package(s): `BGL`

### DIFF
--- a/BGL/doc/BGL/CGAL/boost/graph/properties.h
+++ b/BGL/doc/BGL/CGAL/boost/graph/properties.h
@@ -4,11 +4,6 @@ namespace boost {
 /// \ingroup PkgBGLProperties
 /// @{
 
-/// The constant `edge_index` is a property tag which identifies the <i>index</i> property of an edge of a \sc{Bgl}
-/// <a href="https://www.boost.org/libs/graph/doc/Graph.html"><code>Graph</code></a>.
-/// \cgalModels <a href="https://www.boost.org/libs/graph/doc/PropertyTag.html"><code>PropertyTag</code></a>
-enum edge_index_t { edge_index};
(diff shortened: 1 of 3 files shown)


lrineau:
Why not add that header to `INPUT` in `BGL/doc/BGL/Doxyfile.in`?

MaelRL:
I would have to hide code that in ::CGAL:: and which is not documented. Since the doc header already exists, I'd rather use it than pollute the header in `include/CGAL` with more `ifndef DOXYGEN_RUNNING`.

Ideally the doc wouldn't be split in BGL...

MaelRL:
https://travis-ci.org/CGAL/cgal/jobs/647383669 travis failed on its own, but I don't have the option to restart jobs anymore.

lrineau:
I have restarted the Travis job. I do not know why you could not. Anyway, the Travis checks are now green.

MaelRL:
Thanks.

Restart might require push access for some reason, or some privilege that was removed when you created the new groups.

sloriot:
I don't like the external tags but we should at least document them so that we can deprecate them, or they were never documented and we just deprecate them. Maybe they are now useless anyway since we have dynamic maps (except the initialization issue).

maxGimeno:

PR #4513, opened 2020-02-10
add arithmetic operations

Mr-Leshiy:
Add some simple arithmetic operations into the Point_2 and Point_3 classes.

P.S:
It is my first pull request into this repo, now I am in a learning process of the project codebase. So I am open to any comments :). I have some ideas what to add in this library. With this PR I want to understand how to work with this repo.

### DIFF
--- a/Kernel_23/include/CGAL/Point_2.h
+++ b/Kernel_23/include/CGAL/Point_2.h
@@ -174,6 +174,48 @@ public:
                   R().construct_opposite_vector_2_object()(v));
     return *this;
   }
+  
+  Point_2<R_>&
+  operator+=(const typename Point_2<R_> &p)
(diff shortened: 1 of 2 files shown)


sloriot:
Thanks for the PR but we won't be able to merge it as it breaks [one fundamental rule of kernel](https://doc.cgal.org/latest/Kernel_23/index.html):  *...the kernel, which consists of constant-size non-modifiable geometric primitive...*. Even turning those into global operators will not make it into CGAL as (even if convenient) this would also break mathematical rules: You cannot multiply a point by a scalar, add two points,... You can do that with a vector however.

For those reasons I'm closing this PR. Thanks anyway for taking the time to contribute.

PR #4519, opened 2020-02-12
Enable the use of Lazy objects in Compact_container

lrineau:
## Summary of Changes

This PR allows to use kernel objects from `CGAL::Epeck` (or `CGAL::Lazy_kernel<...>`) in `CGAL::Compact_container`.

See also discussion on cgal-dev started by @gdamiand, 2020-02-11 15:07 CET.

## Release Management

* Affected package(s): Kernel_23, Filtered_kernel, STL_Extension

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_X_trapezoid.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_X_trapezoid.h
@@ -6,9 +6,9 @@
 // $URL$
 // $Id$
 // SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
-// 
 //
-// Author(s)	 : Oren Nechushtan <theoren@math.tau.ac.il>
+//
+// Author(s)         : Oren Nechushtan <theoren@math.tau.ac.il>
 //               updated by: Michal Balas <balasmic@post.tau.ac.il>
 
 #ifndef CGAL_TD_X_TRAPEZOID_H
@@ -38,24 +38,24 @@ namespace CGAL {
  * Implementation of a pseudo-trapezoid as two halfedges(top,bottom)
  * and two curve-ends(left,right).
  * Trapezoids are represented as two curve-ends called right and left and
- * two halfedges called top and bottom. The curve-ends (points) lie on the 
- * right and left boundaries of the trapezoid respectively and the halfedges 
+ * two halfedges called top and bottom. The curve-ends (points) lie on the
+ * right and left boundaries of the trapezoid respectively and the halfedges
  * bound the trapezoid from above and below.
- * There exist degenerate trapezoids called infinite trapezoid; this happens 
+ * There exist degenerate trapezoids called infinite trapezoid; this happens
  * when one of the four sides is on the parameter space boundary.
  * Trapezoids are created as active and become inactive when Remove() member
  * function called.
  * Each trapezoid has at most four neighbouring trapezoids.
- * X_trapezoid structure can represent a real trapezoid, a Td-edge or an 
+ * X_trapezoid structure can represent a real trapezoid, a Td-edge or an
  * edge-end (end point).
  */
 template <class Td_traits_>
 class Td_X_trapezoid : public Handle
 {
 public:
-  
+
   //type of trapezoid type
-  enum Type 
+  enum Type
   {
       TD_TRAPEZOID,
       TD_EDGE,
@@ -64,7 +64,7 @@ public:
 
   //type of traits class
   typedef Td_traits_                                   Traits;
-  
+
   //type of point (Point_2)
   typedef typename Traits::Point                       Point;
 
@@ -76,13 +76,13 @@ public:
 
   //type of Halfedge_const_handle (trapezoid edge)
   typedef typename Traits::Halfedge_const_handle  Halfedge_const_handle;
-  
+
   //type of Vertex_const_handle (trapezoid vertex)
   typedef typename Traits::Vertex_const_handle    Vertex_const_handle;
 
   //type of Trapezoid (Self)
   typedef typename Traits::X_trapezoid            Self;
-  
+
   //type of Trapezoid parameter space
   // Ninetuple which represents the Trapezoid:
   //  - for regular & edge trapezoids or active point trapezoids:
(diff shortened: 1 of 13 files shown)


lrineau:
@mglisse I do not like at all the use of `union` for type punning. Do you know what would be best in modern C++? [Core Guidelines recommends `reinterpret_cast`](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ru-pun) and I agree. Is there something better, to play with the object-representation of pointers, nowadays? I would prefer a implementation-defined behavior, instead of the well-known- but undefined- behavior we currently use.

mglisse:
It would be more convenient if for_compact_container had a setter and a getter instead of providing a reference (or at least if it accepted any kind of pointer). Assuming that different pointers have the same size, that objects are aligned enough, that pointers to aligned objects end in 0, that's all going to be implementation-defined anyway, and most implementations define union punning. I don't have anything better than casts with uintptr_t in mind at the moment. `std::align` & co may be a way to delegate these tricks to the implementation if you really want.
I also don't like the union here much, unions have ways of confusing optimizers, although I may be too pessimistic.

gdamiand:
Tested successfully locally.

lrineau:
Note that the current PR breaks `Arrangement_on_surface_2`.

gdamiand:
@lrineau If you give me the right, I can update this PR to solve the problems in Arrangement_on_surface_2.

lrineau:
Hi @gdamiand. You can modify this PR as you want. You already have the push access to my fork https://github.com/lrineau/cgal.

gdamiand:
@lrineau problems in Arrangement_on_surface_2 are solved by https://github.com/CGAL/cgal/pull/4519/commits/5c9bb6f016704f09f06902e553229539158c62dd.

lrineau:
@CGAL/geometryfactory Shall we go one with this change?

sloriot:
I don't have a strong opinion. It seems to be backward compatible but as pointed out by Marc we should be sure that there are no performance regression.

gdamiand:
If we think using an union is not a good idea, why not replacing this by the use of reinterpret cast ?

    void*  for_compact_container() const { return reinterpret_cast<void*>(PTR); }
    void*& for_compact_container() { return reinterpret_cast<void*&>(PTR); }


I tested locally and it works ?

mglisse:
> why not replacing this by the use of reinterpret cast ?

That looks dangerous, we would be writing a T* and reading it as a U*, which I think violates strict aliasing. Gcc currently seems to disable this optimization when one of the types is `void*`, but still... If we don't want the union, it seems best to replace the version of for_compact_container that returns a reference with a setter.

lrineau:
I think our union trick is not conformant. That is an undefined behavior. But one that is documented to work with all known compilers, because in the past that was the way people recommended to do type-punning. That is why the compiler optimisers are voluntary shy when they deal with unions.

And I do think that we should instead `reinterpret_cast` the pointer to `std::uintptr_t`, do the arithmetic we need with it to do the bit-squatting. As far as I understand the norm, if our bit-squatting code is correct, the value of the pointer is restored before it is dereferenced, and that seems to be a well defined behavior.

I also agree with Marc that there should be a getter and a setter to read/write that pointer.

gdamiand:
Do not mix things.

The goal of this PR is not to solve existing problem with compact container, but only to add the possibility to store Lazy objects in Compact_container using the same method that the one use in other CGAL packages.

If other objects stored in CC used union, thus let us do the same for lazy objects.

And if you think this is a problem, let us open another issue and another PR to solve this problem globally for CGAL.

lrineau:
We have at least to check, with benchmarks, that the introduction of that union in `CGAL::Handle` does not reduce the performance of computations using `CGAL::Epeck`.

I agree we do not want to solve the issue of `Compact_container` in this PR, but if there is an issue in the design of it, with a performance penatly, then we do not want to extend the performance penatly to other parts of CGAL, and hope to find a solution soon to that penalty.

If this PR does not slow down `CGAL::Epeck` or other parts of CGAL using `CGAL::Handle`, then we can merge it as it is.

maxGimeno:
Conflicts

sloriot:
Most probably responsible for the many red boxes, for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-101/Generator/TestReport_gimeno_Debian-Stable.gz)

gdamiand:
> Most probably responsible for the many red boxes, for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-101/Generator/TestReport_gimeno_Debian-Stable.gz)

Should be corrected by https://github.com/CGAL/cgal/pull/4519/commits/f3bb8f1100a84a190fa77802d6b878cd8dffe6bd

mglisse:
Note that this PR and #4496 are strongly related (people involved know it, but it helps to have a link).

gdamiand:
@lrineau @maxGimeno Please, what is your plan for this PR ?

maxGimeno:
Sorry, I didn't see the bugfix

maxGimeno:

sloriot:
Tested in CGAL-5.1-Ic-114

@lrineau nothing more to add?

lrineau:
> Tested in CGAL-5.1-Ic-114
> 
> @lrineau nothing more to add?

As I said in https://github.com/CGAL/cgal/pull/4519#issuecomment-589133822, proper benchmarks should be done before we can merge this PR. @afabri Do you have a benchmark that involves a lot of computations with `Epeck`? It would be better if that benchmark involves a lot of nodes in DAGs.

maxGimeno:
I benched the branch using this code:
```c++
#include <cassert>
#include <iostream>
#include <CGAL/Timer.h>

typedef CGAL::Exact_predicates_exact_constructions_kernel K;
typedef CGAL::Exact_intersections_tag                     Itag;
typedef CGAL::Constrained_Delaunay_triangulation_2<K,CGAL::Default,Itag> CDT;
typedef CGAL::Constrained_triangulation_plus_2<CDT>       CDTplus;
typedef CDTplus::Point                                    Point;

double do_test()
{
  CDTplus cdt;
  CGAL::Timer timer;
  std::cout  << "Starting." << std::endl;
  int max_nb=400;
  timer.start();
  for (int i = 1; i < max_nb; ++i)
    cdt.insert_constraint( Point(0,i), Point(max_nb,i));
  for (int j = 1; j < max_nb; ++j)
    cdt.insert_constraint( Point(j,0), Point(j,max_nb));
  timer.stop();
  assert(cdt.is_valid());
  return timer.time();
}
int main( )
{
  int nb_test = 5;
  double total_time = 0;
  for(int i=0; i < nb_test; ++i)
    total_time += do_test();
  std::cout<<"took "<<total_time/nb_test<<" s."<<std::endl;

  return 0;
}

```

The results are :
 - Master :  2.78543 s.
 - branch :  2.70006 s.

sloriot:
@lrineau I guess it is fine to merge the branch, right?

lrineau:
That is right. It can be merged.

gdamiand:
Thanks to all; nice new feature ;)

PR #4520, opened 2020-02-12
Fix LCC import_from_triangulation_3

gdamiand:

## Summary of Changes

Bug fix in the function import_from_triangulation_3 that transforms a T3 into a LCC.

When the T3 use weighted points, the function does not compile.

## Release Management

* Affected package(s): T3 ? (or LCC)

### DIFF
--- a/Triangulation_3/include/CGAL/Triangulation_3_to_lcc.h
+++ b/Triangulation_3/include/CGAL/Triangulation_3_to_lcc.h
@@ -15,9 +15,27 @@
 
 #include <CGAL/assertions.h>
 #include <map>
+#include <CGAL/Weighted_point_3.h>
 
 namespace CGAL {
 
+  namespace internal 
+  {
+    template<typename Point>
+    struct Get_point
+    {
+      static const Point& run(const Point& p)
+      { return p; }
+    };
+
+    template<typename Kernel>
+    struct Get_point<CGAL::Weighted_point_3<Kernel> >
+    {
+      static const typename Kernel::Point_3& run(const CGAL::Weighted_point_3<Kernel>& p)
+      { return p.point(); }
+    };
+  }
+  
   /** Convert a given Triangulation_3 into a 3D linear cell complex.
    * @param alcc the used linear cell complex.
    * @param atr the Triangulation_3.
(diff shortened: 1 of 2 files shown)


lrineau:
Everytime I see a PR named "bugfix", I ask the same question: should not the PR apply to one of the release branches?

gdamiand:
No one complain about this bug; it only restricts the use of the function import_from_triangulation_3 to triangulation 3 having non weighted points.

I suggest to simplify our life and keep the bug fix only in master.

maxGimeno:

Triangulation_3/include/CGAL/Triangulation_3_to_lcc.h
```diff
@@ -52,7 +70,7 @@ namespace CGAL {
     for (TVertex_iterator itv = atr.vertices_begin();
          itv != atr.vertices_end(); ++itv)
     {
-      TV[itv] = alcc.create_vertex_attribute(itv->point());
+      TV[itv] = alcc.create_vertex_attribute(internal::Get_point<typename Triangulation::Point>::run(itv->point()));
```
MaelRL: I only notice now, but this seems like it should just have been `atr->construct_point(atr->point(itv))` (might be worth it to add an overload to have `Tr::construct_point(Vertex_handle)` directly) without requiring that full class.

gdamiand: I don't find the construct_point method in any triangulation 3 concept, but only in Periodic triangulation. Is this method exists for all triangulation 3 in cgal ?

MaelRL: https://github.com/CGAL/cgal/blob/master/Triangulation_3/include/CGAL/Triangulation_3.h#L568 is the particular function. It is just a call to the model of `ConstructPoint_3` provided in the traits, and this functor is required by the concept `TriangulationTraits_3`.

gdamiand: Ok, thanks @MaelRL ; will try...

PR #4521, opened 2020-02-12
Fix CGAL_Qt5_FOUND issue

### LINKED ISSUE #4517: CGAL_Qt5_FOUND is wrong
## Issue Details

Currently, if dependencies for CGAL_Qt5 are found, we set CGAL_Qt5_FOUND to true, and then, if it is true, we propose the option for WITH_CGAL_Qt5.

But if WITH_CGAL_Qt5 is set to false, `create_link_to_program(CGAL_Qt5) ` is not performed, which is supposed to mean that when trying to build a graphical example, link errors will occur (e.g. qrc_CGAL.cpp not found or something like that).

The real problem here is that when cgal is installed from a package_manager, without the option for qt5, then CGAL_Qt5_FOUND will still return TRUE if Qt5 is installed, when CGAL_Qt5 is absolutely not working, and instead of having nice NOTICE messages in cmake indicating that a dependency is misisng, the user gets cryptic moc errors.

I think that system about WITH_CGAL_Qt5 and CGAL_Qt5_FOUND is not correct, OR we shouldn't test CGAL_Qt5_FOUND but WITH_CGAL_Qt5 or something. Or maybe we should do the same locally and in the package managers, i.e if Qt is already installed, install CGAL_With_Qt5 by default.

maxGimeno:
## Summary of Changes
At the installation step, if WITH_CGAL_Qt5 was not checked, we remove the GraphicsView package from the configured packages, so no CGAL/Qt header is installed, which should fix the CGAL_Qt5_FOUND issue.

## Release Management

* Issue(s) solved (if any): fix #4517


maxGimeno:

PR #4523, opened 2020-02-13
[Surface Mesh] Fix PLY writing by reindexing vertices

sgiraudot:
## Summary of Changes

The Surface Mesh PLY writer used the vertices indices directly, which could lead to ill-formed files in the presence of garbage for example. This PR introduces a reindexing of vertices, similarly to what was already done in the OFF writer of the same class.

PR targets the 5.0 branch as the PLY writer was introduced in 5.0.

## Release Management

* Affected package(s): Surface Mesh


maxGimeno:

PR #4524, opened 2020-02-13
Handle the case of interior vertex

### LINKED ISSUE #4522: Infinite loop in PMP::corefine_and_compute_union since 4.14.2
When corefining and unioning two closed, non-self-intersecting meshes, an infinite loop is occasionally seen. This is new with 4.14.2 and also in 5.x (it did not happen in 4.12). It only happens (apparently) when the output mesh is the same as one of the inputs.

OFF files and demo code are in
https://www.dropbox.com/sh/0q1i263zkjjjv3o/AAAz27BUXGYUBqRGZDnTi_v3a?dl=0
The demo code can be built in the PMP examples environment (it's a mod of an existing sample)

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 10
* Compiler: Visual Studio 2019, 32-bit, tool set 1.42
* CGAL version: 4.14.2 or later
* Boost version: 1.67.0

sloriot:
A vertex can become interior after the stitch of the patches kept
while being the endpoint of a polyline to be removed

Completes #3893
Fixes #4522

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/face_graph_utils.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/face_graph_utils.h
@@ -1640,9 +1640,22 @@ void remove_unused_polylines(
     halfedge_descriptor h = halfedge(v, tm), start=GT::null_halfedge();
 
     do{
+
+      halfedge_descriptor tmp_start = h;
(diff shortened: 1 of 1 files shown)


lrineau:
@maxGimeno This new bug-fix PR does not look dangerous. I think you can test it tonight with the other buf-fix PRs, if Travis checks are fine.

maxGimeno:

PR #4531, opened 2020-02-17
Bug fix in selection expansion for removal

sloriot:
Handle situation when the range of faces kept is split by a boundary:
Screenshot from 2020-02-17 19-15-02
In such case, the range must be  split and considered as two ranges


maxGimeno:

PR #4532, opened 2020-02-18
3mf <2.0

maxGimeno:
## Summary of Changes
We recently realized that the whole API of the lib3MF changed in the version 2.0. This PR adds doc and cmake checks to be sure that the user is not using a version > 2. 

## Release Management

* Affected package(s): BGL


Stream_support/include/CGAL/IO/read_3mf.h
```diff
@@ -589,6 +589,8 @@ int read_from_3mf(const std::string& file_name, PointRanges& all_points,
  * \param names will contain the name of each mesh in `file_name` if any.
  *  If the i'th mesh has no name, it will be called "Unknown Mesh" in names.
  * \return the number of soups read.
+ * 
+ * \attention Versions superior to 2.0 of lib3mf are not supported.
```
sloriot: if 2.0 is not supported it should be "superior or equal to "

sloriot:
Tested locally and version 2.0 is now ignored (at least in the demo)

sloriot:
A bit surprizing that `test/Stream_support/CMakeLists.txt` is not update...

maxGimeno:

PR #4538, opened 2020-02-20
Fix Yade high precision Real compilation

### LINKED ISSUE #4527: Incorrect use of std::sqrt, std::pow, std::abs.
## Issue Details

I have implemented in [YADE](https://yade-dem.org/doc/) support for [high precision calculations](https://yade-dev.gitlab.io/-/trunk/-/jobs/438538107/artifacts/install/share/doc/yade-ci/html/HighPrecisionReal.html). The problem is that I had to pollute `std::` namespace with this:

```CPP
namespace std {
using ::yade::math::pow;
using ::yade::math::sqrt;
using ::yade::math::abs; // This one is actually written elsewhere in the code, in specific function calls.
}
```

See the file [CgalNumTraits.hpp](https://gitlab.com/yade-dev/trunk/-/blob/master/lib/high-precision/CgalNumTraits.hpp#L161). Best programming practices require that mathematical functions are always called unqualified and let [ADL](https://en.cppreference.com/w/cpp/language/adl) take care of this. See a discussion about this problem in these references (search for word "unqualified"):

1. [boost float128](https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/float128_hints.html)
2. [boost multiprecision](https://www.boost.org/doc/libs/1_60_0/libs/multiprecision/doc/html/boost_multiprecision/tut/floats/fp_eg/jel.html)
3. [boost real concepts](https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/real_concepts.html)

You can write `using std::pow` or `using std::abs` inside a function which is going to call this. But write the function call itself unqualified, so that ADL will work.

There are several places in CGAL that use `std::sqrt`, for example `PCA_util_Eigen.h` or `
centroid.h`. Maybe you should be using [Sqrt function](https://gitlab.com/yade-dev/trunk/-/blob/master/lib/high-precision/CgalNumTraits.hpp#L75) from numerical traits?

The same applies to `std::pow` and `std::abs` functions. Launching command in CGAL sources:

```sh
grep -E "std::pow|std::sqrt|std::abs" . -r --color
grep -E "=.*std::pow|=.*std::sqrt|=.*std::abs" . -r --color
```

has a lots of hits.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux, see [gitlab-CI pipeline](https://gitlab.com/yade-dev/trunk/-/merge_requests/383/pipelines).
* Compiler: g++, clang
* Release or debug mode: both
* Specific flags used (if any): ` -Werror -Wformat -Wformat-security -Wformat-nonliteral -Wall -Wextra -Wnarrowing -Wreturn-type -Wuninitialized -Wfloat-conversion -Wcast-align -Wdisabled-optimization -Wtrampolines -Wpointer-arith -Wswitch-bool -Wwrite-strings -Wnon-virtual-dtor -Wreturn-local-addr -Wno-error=cpp`
* CGAL version: 4.13
* Boost version: 1.67.0
* Other libraries versions if used (Eigen, TBB, etc.): that's for one particular run. Gitlab pipeline covers more versions.

| library       | cmake                | C++                |
| ------------- | -------------------- | ------------------ |
| boost         | 106700               | 1.67.0             |
| cgal          |                      | 4.13               |
| clp           | 1.16.11              | 1.16.11            |
| cmake         | 3.13.4               |                    |
| coinutils     | 2.10.14              | 2.10.14            |
| compiler      | /usr/bin/g++ 8.3.0   | gcc 8.3.0          |
| eigen         | 3.3.7                | 3.3.7              |
| freeglut      | 2.8.1                |                    |
| gl            |                      | 20180725           |
| ipython       | 5.8.0                |                    |
| metis         |                      | 5.1.0              |
| mpi           | 3.1                  | ompi:3.1.3         |
| mpi4py        | 2.0.0                |                    |
| openblas      |                      |  OpenBLAS 0.3.5    |
| python        | 3.7.3                | 3.7.3              |
| qglviewer     |                      | 2.6.3              |
| qt            |                      | 5.11.3             |
| sphinx        | 1.8.4-final-0        |                    |
| sqlite        |                      | 3.27.2             |
| suitesparse   | 5.4.0                | 5.4.0              |
| vtk           | 6.3.0                | 6.3.0              |

Linux version: Devuan GNU/Linux 3 (beowulf)

**EDIT:** You may want to have a look at Eigen, I reported there the [same bug](https://gitlab.com/libeigen/eigen/issues/1823) but they had this problem only in four lines in entire codebase.

cosurgi:
This one partially fixes #4527 - first commit is all that is necessary to fix compiling Yade. Second commit is because the changes were extremely similar to the first commit, so I felt they were appropriate.

The `std::abs` though looks very strange there, it's called on an always positive value of `squared_length`, unless you expect complex numbers to be in play?

## Release Management

* Affected package(s): PCA

### DIFF
--- a/Principal_component_analysis/include/CGAL/PCA_util.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util.h
@@ -657,7 +657,8 @@ assemble_covariance_matrix_3(InputIterator first,
        t[0].y(), t[1].y(), 0.0,
                    t[0].z(), t[1].z(), 1.0};
     Matrix transformation = init_matrix<FT>(3,delta);
-    FT length = std::sqrt(t.squared_length());
+    using std::sqrt;
+    FT length = sqrt(t.squared_length());
 
 		// skip zero measure primitives
     if(length == (FT)0.0)
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -575,7 +575,8 @@ assemble_covariance_matrix_3(InputIterator first,
     transformation << t[0].x(), t[1].x(), 0.0, 
                       t[0].y(), t[1].y(), 0.0,
                       t[0].z(), t[1].z(), 1.0;
-    FT length = std::sqrt(t.squared_length());
+    using std::sqrt;
+    FT length = sqrt(t.squared_length());
 
     // skip zero measure primitives
     if(length == (FT)0.0)
--- a/Principal_component_analysis/include/CGAL/linear_least_squares_fitting_segments_2.h
+++ b/Principal_component_analysis/include/CGAL/linear_least_squares_fitting_segments_2.h
@@ -91,7 +91,8 @@ linear_least_squares_fitting_2(InputIterator first,
     FT delta[4] = {t[0].x(), t[1].x(), 
 		   t[0].y(), t[1].y()};
     Matrix transformation = init_matrix<FT>(2,delta);
-    FT length = std::sqrt(t.squared_length());
+    using std::sqrt;
+    FT length = sqrt(t.squared_length());
     CGAL_assertion(length != 0.0);
 
     // Find the 2nd order moment for the segment wrt to the origin by an affine transformation.
--- a/Principal_component_analysis_LGPL/include/CGAL/centroid.h
+++ b/Principal_component_analysis_LGPL/include/CGAL/centroid.h
@@ -129,7 +129,9 @@ centroid(InputIterator begin,
       it++)
   {
     const Segment& s = *it;
-    FT length = std::sqrt(std::abs(s.squared_length()));
+    using std::abs;
+    using std::sqrt;
+    FT length = sqrt(abs(s.squared_length()));
     Point c = K().construct_midpoint_2_object()(s[0],s[1]);
     v = v + length * (c - ORIGIN);
     sum_lengths += length;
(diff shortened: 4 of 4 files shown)


lrineau:
Hi Janek. Thanks for your contribution.

The PR you created applies to `master`. That means the fix will only be in the future CGAL release CGAL-5.1. (planned in June). If you want the fix to apply to release branch, then retarget your PR for the branch `releases/CGAL-4.14-branch`. If you add me as collaborator for your branch, I can do the rebasing.

cosurgi:
Hi Laurent,

please let me know if you received the collaborator invite. I couldn't find the option to give you a role with write access. Maybe it's automatic now? I hope this works ;)

lrineau:
@cosurgi I have managed to rebase and push the PR. Thanks.

@maxGimeno Be careful: I have rebased this PR. If you have already merged it in `integration`, please redo it.

cosurgi:
uh, what was wrong with these checks, can I help here somehow?

lrineau:
I have pushed the rebased branch, and then changed the base branch of the PR. That seems to confuse the Travis CI. I have repushed the branch. That will probably fix the issue.

lrineau:
The Travis results are OK, now.

mglisse:
> The `std::abs` though looks very strange there, it's called on an always positive value of `squared_length`, unless you expect complex numbers to be in play?

For centroid.h, one of the std::abs was removed in 0c408aeee22f6db9a198e97b77772940d3571c4c, but the others remain, probably they could be removed as well. The only reason I can think of for abs here would be numerical instability and avoiding sqrt(negative), but the squared length of a segment is a sum of squares, not something that can easily be negative, however approximate the computation may be...

maxGimeno:

cosurgi:
Great! Thanks for the info.

PR #4540, opened 2020-02-25
Some noexcept, swap, etc to help Handle* types

mglisse:
## Summary of Changes

The most important part is noexcept on the copy constructors. I assumed that the destructors of the types we put in there don't throw, it would be possible to make it conditionally noexcept, but that doesn't seem worth the trouble.
There is an issue with the precondition in the copy constructor of Handle: noexcept means that it cannot throw an exception it will terminate instead. I think the gains of noexcept far outweigh the possibility to catch such an exception.

## Release Management

* Affected package(s): Number_types, STL_Extension

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/Point_2.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Point_2.h
@@ -50,6 +50,15 @@ public:
   PointC2(const FT &hx, const FT &hy, const FT &hw)
     : base(hx, hy, hw) {}
 
+  friend void swap(Self& a, Self& b)
+#ifdef __cpp_lib_is_swappable
+    noexcept(std::is_nothrow_swappable_v<Vector_2_>)
+#endif
+  {
+    using std::swap;
+    swap(a.base, b.base);
+  }
+
   const FT& x() const
   {
       return base.x();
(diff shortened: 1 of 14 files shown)


mglisse:
Question: ideally, all kernel objects, both those in Kernel_23 and in Cartesian/Homogeneous, could do with a swap function. Is it worth the trouble to add it to all of them? I only did Point_3 as a proof of concept, and because it seems to be the most used object and thus the most likely to be swapped.

lrineau:
What perf benefit do you expect from swap functions? As I said in #4496, I see the benefit of `noexcept` move-constructors and move-assignements, but that is less clear for swaps.

mglisse:
> What perf benefit do you expect from swap functions? As I said in #4496, I see the benefit of `noexcept` move-constructors and move-assignements, but that is less clear for swaps.

It isn't so much about having them noexcept, but just having them small and a bit faster. IIRC, some implementations of std::sort use (iter_)swap a lot, so if you want to sort Lazy_exact_nt numbers, that can help a bit. And Andreas was complaining recently about slow swapping in random_shuffle for #4495.

lrineau:
That is right. STL sorting/shuffling algorithms use `swap` with ADL. Now I see the need to have `swap` for `Point_3`. For other object types... maybe, but the use-case is less clear. More like a nice-to-have.

mglisse:
> Now I see the need to have `swap` for `Point_3`.

Probably Point_2 as well then, but it will wait, this PR is already under testing.

maxGimeno:

PR #4542, opened 2020-02-25
Use FT instead of RT

sloriot:
`Compare_distance_3` is a generic functor that is comparing
the returned value of Squared_distance_3. If called to
compare the distance of a point to a plane/segment/triangle
a division is used leading to runtime error.

Links in the code to the division:
https://github.com/CGAL/cgal/blob/master/Distance_3/include/CGAL/squared_distance_3_1.h#L140
https://github.com/CGAL/cgal/blob/master/Distance_3/include/CGAL/squared_distance_3_0.h#L202


maxGimeno:

PR #4547, opened 2020-03-01
Enable move semantics to AABB trees and use smart pointers

### LINKED ISSUE #2925: AABB tree: add move semantic (Why is copying in AABB_tree made private?)
https://github.com/CGAL/cgal/blob/72a6f22a382b4662196b6e495805ac5631aadf73/AABB_tree/include/CGAL/AABB_tree.h#L682

The copy constructor in `AABB_tree` is not public. It seems like making `std::vector<AABB_tree>` is not possible without some kind of hack. Any reason why it's there and should not be made `public`?

theartful:
## Summary of Changes

This addresses issue #2925. I also took the liberty to change the following:
1. Use `std::vector` and `std::unique_ptr` instead of manually managing array memory.
2. Move `expand` function from `AABB_node` to `AABB_tree` because it dealt with memory in an unsafe way + It made sense from a design point of view.

## Release Management

* Issue(s) solved (if any): fix #2925
## TODO
- [x] tests
- [x] update CHANGES.md
- [x] CLA?


AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -555,8 +577,8 @@ namespace CGAL {
     AABBTraits m_traits;
 		// set of input primitives
 		Primitives m_primitives;
-		// single root node
-		Node* m_p_root_node;
```
theartful: Using raw pointer to manage an array. Also it's a little misleading as it's a pointer to an array not a single node.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -670,25 +710,51 @@ namespace CGAL {
...
 	template<typename Tr>
 	void AABB_tree<Tr>::build()
 	{
     clear_nodes();
+
     if(m_primitives.size() > 1) {
 
 			// allocates tree nodes
-			m_p_root_node = new Node[m_primitives.size()-1]();
-			if(m_p_root_node == nullptr)
```
theartful: If `new` failed, it will throw an exception (even with `-fno-exceptions` unless libc++ is compiled with the flag, which in this case will call abort), and will not return nullptr. So the check is superfluous.

AABB_tree/include/CGAL/internal/AABB_tree/AABB_node.h
```diff
@@ -109,49 +109,8 @@ class AABB_node
...
-    break;
-  case 3:
-    m_p_left_child = &(*first);
-    m_p_right_child = static_cast<Node*>(this)+1;
-    right_child().expand(first+1, beyond, 2,traits);
-    break;
-  default:
-    const std::size_t new_range = range/2;
-    m_p_left_child = static_cast<Node*>(this) + 1;
-    m_p_right_child = static_cast<Node*>(this) + new_range;
```
theartful: A little presumptuous memory handling, as it assumes there are other nodes, and they are in contiguous memory. Also bug prone.

AABB_tree/include/CGAL/internal/AABB_tree/AABB_search_tree.h
```diff
@@ -83,7 +85,7 @@ namespace CGAL
                 typedef typename CGAL::Orthogonal_k_neighbor_search<TreeTraits> Neighbor_search;
                 typedef typename Neighbor_search::Tree Tree;
         private:
-                Tree* m_p_tree;
+                std::unique_ptr<Tree> m_p_tree;
```
sloriot: I wonder if we really need yet another pointer here (rather than a plain `Tree` object)

sloriot:
@theartful Thanks for the submission, it's the best *proof-of-concept* application to a GSoC I ever saw. Did you do some benchmark on the tree construction (using [this example](https://doc.cgal.org/latest/AABB_tree/AABB_tree_2AABB_face_graph_triangle_example_8cpp-example.html) with [this model](http://www.pmp-book.org/download/meshes/iphi_fullres.zip) for example)?

sloriot:
cc @danston

sloriot:
A test in `AABB_tree/test/AABB_tree/` should be added to test the new feature.

AABB_tree/include/CGAL/internal/AABB_tree/AABB_search_tree.h
theartful: I don't think using a pointer is really necessary. Will change it.

AABB_tree/benchmark/AABB_tree/CMakeLists.txt
```diff
@@ -23,9 +23,18 @@ if ( NOT Boost_FOUND )
   return()  
 endif()
 
+# google benchmark
+find_package(benchmark REQUIRED)
+if ( NOT benchmark_FOUND )
+  message(STATUS "This project requires Google benchmarks, and will not be compiled.")
+endif()
```
lrineau: [Doc of `REQUIRED`](https://cmake.org/cmake/help/latest/command/find_package.html):
>  The `REQUIRED` option stops processing with an error message if the package cannot be found.

That means that, if `benchmark` is not found, then the next lines will not be processed anyway. You can remove the `if(NOT benchmark_FOUND)`.

theartful:
I added some benchmarks, and the performance is the same:
```
NEW
====
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_TreeCreation      963083 ns       963068 ns          693
BM_Intersections       8776 ns         8776 ns        78002


OLD
====
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_TreeCreation      971003 ns       970954 ns          681
BM_Intersections       8815 ns         8815 ns        76547

```
I'm not sure however how to go about testing the move assignments/constructor. Checking the the member variables seems like the wrong way arorund it.

theartful:
I added some unit tests for move semantics which were mostly copy pasted from already made ones. I'm not sure about the code redundancy.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -555,8 +577,8 @@ namespace CGAL {
     AABBTraits m_traits;
     // set of input primitives
     Primitives m_primitives;
-    // single root node
-    Node* m_p_root_node;
+    // tree nodes. first node is the root node
+    std::vector<Node> m_p_nodes;
```
lrineau, line 581: I suggest to rename `m_p_nodes` to `m_nodes`.

Installation/CHANGES.md
```diff
@@ -6,6 +6,9 @@ Release History
 
 Release date: June 2020
 
+### Add Move Semantics to AABB Trees
+ -   Added move constructor and assignment operator to AABB Tree class.
+
```
lrineau, line 11: Should be move to the new section about CGAL-5.2, after a merge of `master`.

lrineau (approved):

AABB_tree/include/CGAL/AABB_tree.h
sloriot: The name comes from the original one and there is `m_p_search_tree` (and maybe others) so I'm not in favor of the renaming (if we don't do it for the whole package, which is out of the scope of the PR).

lrineau: That is different: `m_p_nodes` is no longer a pointer. `m_p_search_tree` is still a pointer, and that justifies its prefix `m_p_`.

sloriot: `p_` stands for pointer so you are right.

Installation/CHANGES.md
sloriot: @theartful Since I'm responsible for the conflicts, let me try to do the merge and I'll let you verify I did not do anything wrong.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -632,6 +656,30 @@ namespace CGAL {
...
+  template <typename Tr>
+  typename AABB_tree<Tr>::Self& AABB_tree<Tr>::operator=(Self&& tree) noexcept
+  {
+    m_traits = std::move(tree.m_traits);
+    m_primitives = std::move(tree.m_primitives);
+    m_nodes = std::move(tree.m_nodes);
+    m_p_search_tree = std::move(tree.m_p_search_tree);
+    m_use_default_search_tree = std::exchange(tree.m_use_default_search_tree, true);
+#ifdef CGAL_HAS_THREADS
+    m_atomic_need_build = tree.m_atomic_need_build.load(std::memory_order_relaxed);
```
sloriot, line 668: I'm not sure that what I did is the best choice.

lrineau: That is correct as far as I can see.

maxGimeno:
@sloriot is this ready ?

sloriot:
yep

lrineau:
@MaelRL [You added](https://github.com/CGAL/cgal/pull/4547#event-3792505387) "not yet approved". Can we remove it now?

MaelRL:
It was pending your review, feel free to remove it if you're happy with the changes.

lrineau (approved):

lrineau:
@maxGimeno The PR is ready.

maxGimeno:
Warnings : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-30/AABB_tree/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz,
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-30/AABB_tree/TestReport_lrineau_ArchLinux-clang.gz,
and runtime error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-30/AABB_tree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

sloriot:
@janetournois could you please give a try for the error?

maxGimeno:
Also doc warning : https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.2-Ic-30/logs_master/AABB_tree.log

theartful:
I think this shouldn't fail now.

maxGimeno:

PR #4549, opened 2020-03-01
Shape_detection: Correcting comment in efficient_RANSAC_with_parameters example

thopiekar:
Just because 200 != 500.


thopiekar:
Just found this while reading the docs:
https://doc.cgal.org/latest/Shape_detection/index.html#Shape_detection_RANSACExample_with_parameters

PR #4552, opened 2020-03-02
[Point Set Processing] Big factorization + cleanup

sgiraudot:
## Summary of Changes

This PR introduces many factorizations and cleaning in the PSP package (internal code only, no API change):
* In lots of places, points were copied in a first container then passed to the KD tree (which copied them again), which seems quite inefficient. The KD tree now works on the input range (using iterators to limit copies if the data set is big)
* The neighbor search that I already factorized some time ago is even more factorized now (even the tree type/instantiation is now handled in a single class, removing many code duplication)
* TBB functors are removed and replaced by lambdas (which are much more self-contained as no typedefs/templates are required)
* The `Parallel_callback` I introduced to handle callbacks in parallel mode is replaced by a `Callback_wrapper` which indifferently works with sequential and parallel
* A function (so far undocumented) `CGAL::for_each` is introduced to run a functor on a range either sequentially or in parallel with TBB. It checks that no parallel tag is used without TBB being linked and it handles the cases of parallel loop on non-random access iterators (by copying iterators in a vector, which is equivalent to what was done in PSP in general, only now it only happens in the worst case scenario)

In general, the code of PSP is largely reduced and much more readable. Many functions were written ike that:

```c++
#ifndef CGAL_LINKED_WITH_TBB
  CGAL_static_assertion_msg (!(boost::is_convertible<ConcurrencyTag, Parallel_tag>::value),
			     "Parallel_tag is enabled but TBB is unavailable.");
#else
   if (boost::is_convertible<ConcurrencyTag,Parallel_tag>::value)
   {
     CGAL::internal::Tbb_functor_defined_somewhere_else</* many template paremeters */>
       f (/* many parameters */);
     tbb::parallel_for(tbb::blocked_range<size_t>(0, kd_tree_points.size ()), f);
   }
   else
#endif
     {
       for(typename PointRange::const_iterator it = points.begin(); it != points.end(); it++, nb++)
       {
         /* The actual algorithm, often duplicated in TBB functor */
       }
     }
```

Now they look like that:
```c++
  CGAL::for_each<ConcurrencyTag>
    (points,
     [&](const typename PointRange::const_iterator::value_type& t)
     {
       /* The actual algorithm */
     });
```

## Release Management

* Affected package(s): PSP, Demo


maxGimeno:
[travis error](https://travis-ci.org/CGAL/cgal/jobs/657339097#L3208)

maxGimeno:
[still](https://travis-ci.org/CGAL/cgal/jobs/658593727#L3279) :(

sgiraudot:
Sorry, I should have spotted those… I rerun PSP examples + tests locally, it should be good now.

lrineau:
A license issue, now: https://travis-ci.com/CGAL/cgal/jobs/294843121#L2868

STL_Extension/include/CGAL/for_each.h
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2020 GeometryFactory (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org).
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
```
lrineau: Here is the license issue: STL_Extension` is supposed to be **L**GPLv3+

sgiraudot:
Please do not merge this PR yet, I just found a bug.

sgiraudot:
Fixed.

sloriot:
[Warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-101/Scale_space_reconstruction_3_Examples/TestReport_gimeno_Debian-Stable.gz)

sgiraudot:
I also found [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-101/Point_set_processing_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz) which I'll check.

STL_Extension/include/CGAL/iterator.h
```diff
@@ -1471,6 +1471,17 @@ dispatch_or_drop_output(O... o)
   return Dispatch_or_drop_output_iterator<std::tuple<V...>, std::tuple<O...> >(o...);
 }
 
+
+// Trick to select iterator or const_iterator depending on the range constness
+template <typename RangeRef>
+struct Range_iterator_type                  { typedef int                               type; }; // Should generate errors
```
sloriot:
```suggestion
struct Range_iterator_type;
```

maxGimeno:
warning here :https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-102.shtml#Optimal_transportation_reconstruction_2_Demo

maxGimeno:
[error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-103/Optimal_transportation_reconstruction_2_Demo/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)

STL_Extension/include/CGAL/for_each.h
```diff
@@ -0,0 +1,155 @@
...
+         <typename Range_iterator_type<RangeRef>::type>::reference r : range)
+    try
+    {
+      functor(r);
+    }
+    catch (const stop_for_each&)
+    {
+      break;
+    }
+}
```
lrineau: The use of exceptions here seems a huge hammer for a small task. The functors could return somethings, instead of `void`, to signal need for break. It could be a simple `bool`, or better: an enum-class.

sgiraudot: Indeed… I remember that it somehow made sense when I first used that, but a return value is much simpler/efficient/better. I fixed it.

maxGimeno:
warnings : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-106/AABB_tree/TestReport_gimeno_Fedora-rawhide-Release.gz

maxGimeno:
Errors on for_each :

Poisson_surface_reconstruction_3/include/CGAL/Poisson_reconstruction_function.h
```diff
@@ -106,13 +106,13 @@ struct Poisson_visitor {
...
-    : ratio(ratio), m_random(random) {}
+  Poisson_skip_vertices(const double ratio = 0)
+    : ratio(ratio) {}
 
   template <typename Iterator>
-  bool operator()(Iterator) const {
-    return m_random.get_double() < ratio;
+  bool operator()(Iterator it) const {
+    // make the result deterministic for each iterator
+    return Random((std::size_t)(&*it)).get_double() < ratio;
```
sloriot: @lrineau : this is not something we should do, right?
@sgiraudot  what are you trying to achieve? you need different seeds per iterator? This won't be deterministic accross runs AFAIU.

lrineau: First, that does not seem efficient, to create an object of type `Random`, at each call, just to get a few random bits.

Second, I am surprised, because I think the previous version (before the patch) seemed deterministic, and then the second one seems to be non-deterministic, because that depends on the memory layout. And the comment says the contrary.

sgiraudot: The previous version produced a different range everytime a `for()` loop was used to iterator from begin to end, which created problems in the new version of `compute_average_spacing()` where the range is iterated twice. This patch was made to make the range consistently the same if iterated several times.

sloriot: can't you init the random with the size of the range as seed once?

sgiraudot: It's not a matter of initialization.
Do this:
```c++
for(const auto& : range)
  // something

for(const auto& : range)
  // something
```
and you don't get the same range on the second loop, no matter the initialization. Maybe I could use a trick to initialize a shared `Random` object when calling `begin()`…

maxGimeno:
The following files have trailing whitespaces:
Poisson_surface_reconstruction_3/include/CGAL/Poisson_reconstruction_function.h

sloriot:
Tested in CGAL-5.1-Ic-114

PR #4556, opened 2020-03-04
Fix Bezier intersection bug around the boundaries (issue #3381)

### LINKED ISSUE #3381: Issue with Bezier traits
Here is an old bug-report by @alextsui05

"Attached, I will leave a small program and two data files that show the problem.

Running the test with a.dat will demonstrate that we unexpectedly find an intersection point between the upper quarter circle curve and the lower quarter circle curve after being made x-monotone. We expect them to be disjoint because there is a tiny subcurve separating them.

b.dat replaces the lower quarter circle with a straight-line segment with the same endpoints. Running the test with this shows that no intersection point is found, as we expect.

I tracked the source of the intersection point to a call on line 2442 in the Arr_geometry_traits/Bezier_x_monotone_2.h file, in the _intersect method. The details are beyond me at the moment, but I guess a way to proceed would be to compare why a.dat and b.dat yield different results."

[argt_bezier.zip](https://github.com/CGAL/cgal/files/2454875/argt_bezier.zip)

theartful:
## Summary of Changes

Compute all roots of one of the resultants to do successful matching.

Two methods are used for computing the intersections:
1. de Casteljau subdivision.
2. Analytic solution by solving their equations.

In case de Casteljau fails, the analytic method is used. It's easy to reproduce the bug by disabling de Casteljau subdivision (changing line 2355 in `Bezier_x_monotone_2.h` for example to `if (false)`)
and  trying these curves:
https://www.desmos.com/calculator/f9efwzbb3k
In which, there are two roots, both out of range, but it'll match them as if they're one root resulting in an incorrect root which is in range.
## Release Management

* Affected package(s): 2D Arrangements
* Issue(s) solved (if any): fix #3381

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_cache.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_cache.h
@@ -258,7 +258,7 @@ private:
                              const Polynomial& polyY_1, const Integer& normY_1,
                              const Polynomial& polyX_2, const Integer& normX_2,
                              const Polynomial& polyY_2, const Integer& normY_2,
-                             Parameter_list& s_vals) const;
+                             Parameter_list& s_vals, bool find_out_of_range=false) const;
 
   /*!
    * Compute all s-parameter values of the self intersection of (X(s), Y(s))
@@ -414,7 +414,7 @@ _Bezier_cache<NtTraits>::get_intersections
 
   do_ovlp = _intersection_params (polyX_2, normX_2, polyY_2, normY_2,
                                   polyX_1, normX_1, polyY_1, normY_1,
-                                  t_vals);
+                                  t_vals, true);
 
   CGAL_assertion (! do_ovlp);
 
@@ -464,21 +464,14 @@ _Bezier_cache<NtTraits>::get_intersections
   const Algebraic           one (1);
   unsigned int              k;
 
-  //pointers are used to set the list pts1_ptr as the one with the less values
-  Point_list* pts1_ptr=&pts1;
-  Point_list* pts2_ptr=&pts2;
-  bool swapt=pts1.size() > pts2.size();
-  if (swapt)
-    std::swap(pts1_ptr,pts2_ptr);
-
-  for (pit1 = pts1_ptr->begin(); pit1 != pts1_ptr->end(); ++pit1)
+  for (pit1 = pts1.begin(); pit1 != pts1.end(); ++pit1)
   {
     // Construct a vector of distances from the current point to all other
     // points in the pts2 list.
-    const int                     n_pts2 = static_cast<int>(pts2_ptr->size());
+    const int                     n_pts2 = static_cast<int>(pts2.size());
     std::vector<Distance_iter>    dist_vec (n_pts2);
 
-    for (k = 0, pit2 = pts2_ptr->begin(); pit2 != pts2_ptr->end(); k++, ++pit2)
+    for (k = 0, pit2 = pts2.begin(); pit2 != pts2.end(); k++, ++pit2)
     {
       // Compute the approximate distance between the teo current points.
       dx = pit1->app_x - pit2->app_x;
@@ -515,7 +508,7 @@ _Bezier_cache<NtTraits>::get_intersections
           pit1->y = pit2->y;
 
         // Remove this point from pts2, as we found a match for it.
-        pts2_ptr->erase (pit2);
+        pts2.erase (pit2);
         found = true;
       }
     }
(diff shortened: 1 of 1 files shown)


Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_cache.h
```diff
@@ -606,21 +598,21 @@ bool _Bezier_cache<NtTraits>::_intersection_params
 
   for (k = degX_2; k >= 0; k--)
   {
-    coeff = nt_traits.get_coefficient (polyX_2, k) * normX_1;
+    coeff = nt_traits.get_coefficient (polyX_2, k);
```
theartful: Multiplying the polynomials with these huge normalizing factors really degraded the root finding algorithm. I can't see any reason why it is done. Shouldn't it be divided by it to get the original rational polynomial?

Removing these seemed to solve the problem. If they're necessary, I have another solution that works.

Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_cache.h
```diff
@@ -170,6 +170,7 @@ template <class Nt_traits_> class _Bezier_cache
                                           tangency parameters. */
   Intersect_map     intersect_map;    /*! Maps curve pairs to their
                                           intersection parameter pairs. */
+  std::pair<Polynomial, Polynomial>     resultant_pair;
```
theartful: The resultant polynomial of the two bezier curves. Need to store them to do the boundary check.

Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
```diff
@@ -1720,6 +1731,37 @@ bool _Bezier_x_monotone_2<RatKer, AlgKer, NtTrt, BndTrt>::_is_in_range
...
+  else
+  {
+    Originator_iterator  t_org = _pt.get_originator (_curve, _xid);
+    at_boundary = t_resultant.eval(t_org->point_bound().t_max);
+  }
+  at_t = at_t.abs();
+  at_boundary = at_boundary.abs();
+
+  if (CGAL::compare(at_t, at_boundary) == SMALLER)
+    return true;
```
theartful: polynomial value check to see whether the boundary is closer to the real root than the calculated root

Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
```diff
@@ -1720,6 +1731,37 @@ bool _Bezier_x_monotone_2<RatKer, AlgKer, NtTrt, BndTrt>::_is_in_range
...
+
+  Nt_traits nt_traits;
+  Polynomial derivative = nt_traits.derive(t_resultant);
+  Algebraic deriv_at_t = derivative.eval(t);
+  return (is_min && deriv_at_t.sign() < 0) || (!is_min && deriv_at_t.sign() > 0);
```
theartful: first derivative check to see the direction on which the root resides

Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
```diff
@@ -2639,6 +2681,17 @@ bool _Bezier_x_monotone_2<RatKer, AlgKer, NtTrt, BndTrt>::_intersect
     if (_is_in_range (iit->s, cache) &&
         cv._is_in_range (iit->t, cache))
     {
+      if (iit == inter_list.begin())
+      {
+        if (!_resultant_boundary_check(iit->t, cache, true))
```
theartful: these boundary checks are only tested at the first and last calculated roots

lrineau:
Hi @theartful, would that patch also apply to release branches `releases/CGAL-4.14-branch` and `releases/CGAL-5.0-branch`? If that is the case, the PR should be rebased onto `releases/CGAL-4.14-branch`.

theartful:
Sorry about the hassle over such a small fix. Third time is the charm!
The problem was that roots were matched incorrectly. Specifically this case:
```
 (root1)      (root2)       (root3)
-------------=============================
                      curve2
=====================------------------------      
       curve1

===== : valid range
----- : invalid range
```
The previous method would calculate the roots in the valid range for each curve, which in this case would be:
```
curve1_roots: {root1_sval, root2_sval}
curve2_roots: {root2_tval, root3_tval}
```
And then would match t-s pair with the curve having the least roots to the one with the biggest number of roots. Clearly here `root2` and `root3` cannot be matched, but it would match them anyway resulting in erroneous results.
The solution I did is to calculate all the roots for one of the curves, then the matching will always be successful.

lrineau:
@theartful Why do you frequently close and then reopen the PR?

theartful:
Yeah, sorry about that. I found out that my previous attempt didn't solve
the root cause, and I thought that it will be a while till I find out
what's the problem. So I close it not to waste any reviewer's time.
Probably not the best move. Last time was just rebasing error.

maxGimeno:

sloriot:
@theartful I've rebased your branch, please hard reset you local copy.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-187.shtml

PR #4559, opened 2020-03-05
[Small Feature] Parallelize KD Tree build

sgiraudot:
## Rationale

Many algorithms (notably in the _Point Set Processing_ package) rely on the _Spatial Searching_ package for fast neighbor queries. Although many PSP algorithms take advantage of trivial parallelisation to speed-up computation by querying the `kd` tree in parallel for each point, a substantial time can be spend _building_ the `kd` tree, especially for very large point clouds. Up to a third of the computation time of algorithms like `CGAL::jet_estimate_normals()` can be spent building the `kd` tree when queries are done in parallel.

This feature adds a parallel variant of the `KD_tree::build()` member function. On a quad-core processor, this parallel algorithm is experimentally 2 to 3 times faster than the sequential one, depending on the point cloud.

## Summary of API changes

A template parameter `ConcurrencyTag` is added to the `KD_tree:build()` member function, with possible values being `CGAL::Sequential_tag`, `CGAL::Parallel_tag` and `CGAL::Parallel_if_available_tag`.

For backward compatibility, an overload of `KD_tree::build()` without template parameter is provided, falling back to the sequential version. The `build()` called internally if the tree is not built at first query is also the sequential one. The rational is that parallelism should always be explicitly specified by users (also to avoid hidden changes of behavior in existing codes).

(No change)

## CHANGES.md

(todo)

### Version 1 (_outdated_)

* __User manual:__
  - The [first example](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build/Spatial_searching/index.html#title5) is updated with an explicit call to `build()` with `CGAL::Parallel_if_available_tag`
  - A paragraph is added to [the `kd` tree design section](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build/Spatial_searching/index.html#title20)
* [__Reference manual__](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build/Spatial_searching/classCGAL_1_1Kd__tree.html#a8559dbe4d7136fbc8ebab5ee290cbe06)

### Version 2

* __User manual:__
  - An [additional example](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build_v2/Spatial_searching/index.html#title17) is given with an explicit call to `build()` with `CGAL::Parallel_tag` and a TBB loop for queries
  - [The Performance section](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build_v2/Spatial_searching/index.html#title18) is updated and a new array for the parallel version is added
* [__Reference manual__](https://cgal.geometryfactory.com/~sgiraudot/SpatialSearching-Parallelize_kd_tree_build_v2/Spatial_searching/classCGAL_1_1Kd__tree.html#a8559dbe4d7136fbc8ebab5ee290cbe06)

## Status

- [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Parallel_KD_Tree_build) Pre-approved -- Andreas Fabri 2 April 2020 (CEST)

### DIFF
--- a/Spatial_searching/include/CGAL/Kd_tree.h
+++ b/Spatial_searching/include/CGAL/Kd_tree.h
@@ -27,12 +27,35 @@
 #include <CGAL/Splitters.h>
 #include <CGAL/internal/Get_dimension_tag.h>
 
-#include <deque>
 #include <boost/container/deque.hpp>
 #include <boost/optional.hpp>
 
 #ifdef CGAL_HAS_THREADS
 #include <CGAL/mutex.h>
+#endif
+
+/*
+  For building the KD Tree in parallel, TBB is needed. If TBB is
+  linked, the internal structures `deque` will be replaced by
+  `tbb::concurrent_vector`, even if the KD Tree is built in sequential
+  mode (this is to avoid changing the type of the KD Tree when
+  changing the concurrency mode of `build()`).
+
+  Experimentally, using the `tbb::concurrent_vector` in sequential
+  mode does not trigger any loss of performance, so from a user's
+  point of view, it should be transparent.
+
+  However, in case one wants to compile the KD Tree *without using TBB
+  structure even though CGAL is linked with TBB*, the macro
+  `CGAL_DISABLE_TBB_STRUCTURE_IN_KD_TREE` can be defined. In that
+  case, even if TBB is linked, the standard `deque` will be used
+  internally. Note that of course, in that case, parallel build will
+  be disabled.
+ */
+#if defined(CGAL_LINKED_WITH_TBB) && !defined(CGAL_DISABLE_TBB_STRUCTURE_IN_KD_TREE)
+#  include <tbb/parallel_invoke.h>
+#  include <tbb/concurrent_vector.h>
+#  define CGAL_TBB_STRUCTURE_IN_KD_TREE
 #endif
 
 namespace CGAL {
@@ -77,14 +100,13 @@ public:
   typedef EnablePointsCache Enable_points_cache;
 
 private:
+
   SearchTraits traits_;
   Splitter split;
 
-
-  // wokaround for https://svn.boost.org/trac/boost/ticket/9332
-#if   (_MSC_VER == 1800) && (BOOST_VERSION == 105500)
-  std::deque<Internal_node> internal_nodes;
-  std::deque<Leaf_node> leaf_nodes;
+#if defined(CGAL_TBB_STRUCTURE_IN_KD_TREE)
+  tbb::concurrent_vector<Internal_node> internal_nodes;
+  tbb::concurrent_vector<Leaf_node> leaf_nodes;
 #else
   boost::container::deque<Internal_node> internal_nodes;
   boost::container::deque<Leaf_node> leaf_nodes;
@@ -119,7 +141,6 @@ private:
     : traits_(tree.traits_),built_(tree.built_),dim_(-1)
   {};
 
-
   // Instead of the recursive construction of the tree in the class Kd_tree_node
   // we do this in the tree class. The advantage is that we then can optimize
   // the allocation of the nodes.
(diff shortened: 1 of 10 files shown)


Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -35,6 +35,33 @@
 #include <CGAL/mutex.h>
 #endif
 
+#ifdef CGAL_LINKED_WITH_TBB
```
sloriot: ?

Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -35,6 +35,33 @@
...
+  point of view, it should be transparent.
+
+  However, in case one wants to compile the KD Tree *without using TBB
+  structure even though CGAL is linked with TBB*, the macro
+  `CGAL_DISABLE_TBB_STRUCTURE_IN_KD_TREE` can be defined. In that
+  case, even if TBB is linked, the standard `deque` will be used
+  internally. Note that of course, in that case, parallel build will
+  be disabled.
+ */
+#if defined(CGAL_LINKED_WITH_TBB) && !defined(CGAL_DISABLE_TBB_STRUCTURE_IN_KD_TREE)
```
sloriot: CGAL_DISABLE_TBB_STRUCTURE_IN_KD_TREE vs CGAL_TBB_STRUCTURE_IN_KD_TREE

Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -77,12 +104,15 @@ class Kd_tree {
...
   SearchTraits traits_;
   Splitter split;
 
-
-  // wokaround for https://svn.boost.org/trac/boost/ticket/9332
-#if   (_MSC_VER == 1800) && (BOOST_VERSION == 105500)
+#if defined(CGAL_TBB_STRUCTURE_IN_KD_TREE)
+  tbb::concurrent_vector<Internal_node> internal_nodes;
+  tbb::concurrent_vector<Leaf_node> leaf_nodes;
+#elif (_MSC_VER == 1800) && (BOOST_VERSION == 105500)
```
sloriot: you should use the boost macro instead of `_MSC_VER`

sgiraudot: Not my code but I'll look into it.

lrineau: I can save you some time: CGAL-5.0 only supports Boost 1.57 and later. No need to test `BOOST_VERSION == 105500`.

Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -128,50 +149,69 @@ class Kd_tree {
...
   {
-    return create_internal_node_use_extension(c);
-  }
-
-  Node_handle
-  create_internal_node(Point_container& c, const Tag_false&)
-  {
-    return create_internal_node(c);
+#ifdef CGAL_TBB_STRUCTURE_IN_KD_TREE
+    return &*(internal_nodes.push_back(Internal_node()));
```
MaelRL: emplace_back (for both) ?

afabri:
I would suggest to remove the parallel tag in the very first example.  There should be a section at the end on parallelism, where you build with the parallel tag and at the same time show a parallel for loop for performing parallel queries.

afabri:
Why is `build()` called before a removal function?  I know this is also written in master.

sgiraudot:
I updated the doc with a Version 2 following @afabri's review.

I'm currently investigating the performances behavior with respect to the bucket size: we present results with the default value (10) and with 20, and the performances with 20 are _always_ better in our array (it was already the case before this small feature), which probably makes users wonder why 20 is not the default value.

sgiraudot:
So I did some more experiment regarding the computation time VS bucket size, and it turns out that for the 800k points case used in the manual, the bucket size that gives the shortest computation time is more around 100:

time_vs_bucket_size_800k_pts

(Violet is sequential, green is parallel.)

The manual does explain that the bucket size should increase with the size of the point cloud, which is confirmed by the following experiment: if I randomly simplify the point cloud to keep only 10% of points (80k), then the optimal bucket size drops to around 50.

time_vs_bucket_size_80k_pts

I'm not sure what we should do about this. It looks like the bucket size could be roughly estimated as a factor of the logarithm of the number of points, but do we want to include such an automatic selection in the KD Tree? Or do we keep it as is, meaning it's the responsibility of users to choose the best bucket size according to their needs?

afabri:
@sgiraudot  I pre-approved but only for the parallelism.  Let's keep the PR open for the bucket size.  Also please look at Mael's comment, and my question about the call to `build()` for removal.

sgiraudot:
Mael's comment was addressed in https://github.com/CGAL/cgal/pull/4559/commits/9ab9081038d578eff8b7c4fd28183a3df64e0ae9 (the `emplace_back()` of TBB has a different signature that forces the construction of the object inside the function, which is not what we want here – it would only block the vector longer for other threads).

The reason why the tree should be built when `remove()` is called is that it's a function specifically designed to remove a point without having to recompute the tree entirely. If you want to remove points before building the tree, then just instantiate the tree once you have removed your points…

Problem: I've checked the code, and `remove()` does not actually call `build()`, but it's a mistake because it tries to remove the point from the tree, which will obviously lead to a segfault if the tree is not built. There's also this code which is disabled because its complexity is too high, but I'm wondering if we should not activate it as it would avoid the segfault:

```c++
#if 0
    // This code could have quadratic runtime.
    if (!is_built()) {
      std::vector<Point_d>::iterator pi = std::find_if(pts.begin(), pts.end(), equal_to_p);
      // Precondition: the point must be there.
      CGAL_assertion (pi != pts.end());
      pts.erase(pi);
      return;
    }
#endif
```

sloriot:
This branch might be responsible for [these warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-122/Point_set_processing_3/TestReport_lrineau_Fedora-Release.gz)

Spatial_searching/include/CGAL/Kd_tree.h
```diff
@@ -128,50 +149,69 @@ class Kd_tree {
...
-
-  Node_handle
-  create_internal_node(Point_container& c, const Tag_false&)
+  Node_handle new_internal_node()
   {
-    return create_internal_node(c);
+#ifdef CGAL_TBB_STRUCTURE_IN_KD_TREE
+    return &*(internal_nodes.push_back(Internal_node()));
+#else
+    internal_nodes.emplace_back (Internal_node());
```
lrineau, line 170: I do not understand that use of `emplace_back`: why use the copy-constructor from a default-constructed `Internal_node`? I suggest:
```suggestion
    internal_nodes.emplace_back ();
```
That passes the empty set of argument to the constructor of `Internal_node`, constructing a default-constructed `Internal_node` inside the container.

mglisse:
The optimal bucket size may also depend on the kernel (and in particular the dimension).

mglisse:
> I've checked the code, and `remove()` does not actually call `build()`,

`remove` calls `root` which calls `build` if necessary.

maxGimeno:

Spatial_searching/include/CGAL/Kd_tree_node.h
```diff
@@ -51,15 +51,12 @@ namespace CGAL {
     typedef typename Kdt::iterator iterator;
     typedef typename Kdt::D D;
 
-    bool leaf;
```
sloriot, line 54: is that change required in this PR ?

sgiraudot: No.

sloriot: it is not clear to me that this will not negatively impact the performance so except if you show that it is faster you should undo that change.

mglisse:
> I agree that the commented code in remove should be enabled.

Did you read my comment in this PR? Do you disagree with it?

sloriot:
> > I agree that the commented code in remove should be enabled.
> 
> Did you read my comment in this PR? Do you disagree with it?

Sorry I missed it.

maxGimeno:

PR #4560, opened 2020-03-05
PMP::isotropic_remeshing() - fix remeshing with patch ids

janetournois:
## Summary of Changes

This PR fixes `PMP::isotropic_remeshing()` when it is used with a `face_patch_map` and no constrained edges.
It should consider interface polylines as constrained edges, so that they can be (or not, depending on the protection parameter) resampled but never "broken".

Moreover, it is what is documented.

Until now, the polyline interfaces were simply not taken into account.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -1523,14 +1523,9 @@ private:
       // tag patch border halfedges
       BOOST_FOREACH(halfedge_descriptor h, halfedges(mesh_))
       {
-        if (status(h)==PATCH && status(opposite(h, mesh_))!=PATCH)
-        {
-          set_status(h, PATCH_BORDER);
-          has_border_ = true;
-        }
         if (status(h) == PATCH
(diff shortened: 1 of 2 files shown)


janetournois:
@sloriot please note that commit f4665a8 modifies your commit 0099e12246c1836ad9b708858b7845d8eb253932
Did I miss something about the case you were handling?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -1528,6 +1528,13 @@ namespace internal {
           set_status(h, PATCH_BORDER);
           has_border_ = true;
         }
+        if (status(h) == PATCH
```
sloriot: you could only update the first if with
` if (status(h)==PATCH && (status(opposite(h, mesh_))!=PATCH || get_patch_id(face(h, mesh_)) != get_patch_id(face(opposite(h, mesh_), mesh_)))`

janetournois: right, thank you!

maxGimeno:

PR #4561, opened 2020-03-05
Travis: Don't compute dependencies if a package is missing.

maxGimeno:
## Summary of Changes
The point is to avoid getting dependencies differencies in a package if some 3rd party lib is not install for that specific job.


maxGimeno:

PR #4562, opened 2020-03-06
BGL: Add missing const (Euler operations)

sloriot:


maxGimeno:

PR #3864, opened 2019-04-18
[Feature] Shape detection with region growing

danston:
# Shape detection with generic region growing - version 1.0

This submission proposes a major update to the `Point_set_shape_detection_3` package by extending it with the generic region growing algorithm that works on a set of arbitrary items. We contribute the generic version of the region growing algorithm and its three particular instances: region growing on points in 2D and 3D, and region growing on a polygon mesh.

## Summary of Changes

- Renaming the package;
- Refactoring the docs;
- Removing current version of region growing on 3D points;
- Adding a generic region growing algorithm;
- Adding three instances of region growing: on 2D points, on 3D points, on a polygon mesh;
- Adding all related region growing examples, tests, and benchmarks;
- Cleaning all related RANSAC files.

## Status

- Proposed for acceptance by reviewers;
- Developed and locally tested (MacOS/AppleClang).

## Release Management

* Affected package(s): `Point_set_shape_detection_3`
* Feature: [Generalized Region Growing](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Shape_detection) Proposed for acceptance. --Andreas Fabri 16:20, 20 March 2019 (CET) 
* Link to compiled documentation (obligatory for small feature) [link](http://anisimovdmitry.com/region_growing/doc_output/Shape_detection-final/index.html)
* License and copyright ownership: GPL (v3 or later)

**TODO:** 
- [ ] After merge, redirection should be put in place
- [ ] News entry for cgal.org

PR #4563, opened 2020-03-06
shape detection: Fix plane normals being arbitrary in region growing.

nh2:
Fixes the bug pointed out in https://github.com/CGAL/cgal/pull/3864#discussion_r389110391.

## Summary of Changes

In #3864 region growing was introduced, but the
`linear_least_squares_fitting_3()` used in `update()` cannot
estimate a correct face normal direction from just unoriented vertex points
(the plane normal's orientation signum may be flipped).

This commit flips the estimated plane normal to face into the
same direction as the majority of the faces it is estimated from.

With the estimated planes now being "the right way around",
this commit subsequently also fixes the incorrect use of `abs()`
in `is_part_of_region()` when comparing a face's normal to the
estimated plane normal (direction matters here).

This fixes parallelly-opposing planar regions (means: with opposite normals),
connected with some triangles in between, like this:

      | ->      <- |
      | ->      <- |      Legend:
      | ->      <- |        face:   |
      | ->      <- |        normal: ->
      | ->      <- |
       \__________/

being detected as 1 region.

This should have been impossible according to the docs
("... and if the angle between its normal and the plane's normal is within
the `angle_threshold`"), given the documented requirement
("\pre `angle_threshold >= 0 && angle_threshold <= 90`").
Now it is impossible.

## Release Management

* Affected package(s): **not sure what to write here**
* License and copyright ownership: **not sure what to write here**, this is my first nontrivial contribution to CGAL, but if in doubt, I'm `Niklas Hambüchen <mail@nh2.me>`, own this change, and am happy to contribute under CGAL's current licenses

### DIFF
--- a/Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h
+++ b/Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h
@@ -208,8 +208,11 @@ namespace Polygon_mesh {
       const FT distance_to_fitted_plane = 
       get_max_face_distance(face);
       
+      // The sign of this scalar product is important, as it indicates
+      // into which side of the plane the face's normal points.
       const FT cos_value = 
-      CGAL::abs(m_scalar_product_3(face_normal, m_normal_of_best_fit));
+      m_scalar_product_3(face_normal, m_normal_of_best_fit);
+
 
       return (( distance_to_fitted_plane <= m_distance_threshold ) && 
         ( cos_value >= m_normal_threshold ));
@@ -284,6 +287,11 @@ namespace Polygon_mesh {
 
         // The best fit plane will be a plane fitted to all vertices of all
         // region faces with its normal being perpendicular to the plane.
+        // Given that the points, and no normals, are used in estimating
+        // the plane, the estimated normal will point into an arbitray
+        // one of the two possible directions.
+        // We flip it into the correct direction (the one that the majority
+        // of faces agree with) below.
         CGAL::linear_least_squares_fitting_3(
           points.begin(), points.end(), 
           fitted_plane, fitted_centroid, 
@@ -291,18 +299,46 @@ namespace Polygon_mesh {
           Local_traits(), 
           CGAL::Eigen_diagonalize_traits<Local_FT, 3>());
 
-        m_plane_of_best_fit = 
+        Plane_3 unoriented_plane_of_best_fit =
         Plane_3(
           static_cast<FT>(fitted_plane.a()), 
           static_cast<FT>(fitted_plane.b()), 
           static_cast<FT>(fitted_plane.c()), 
           static_cast<FT>(fitted_plane.d()));
 
-        const Vector_3 normal = m_plane_of_best_fit.orthogonal_vector();
-        const FT normal_length = m_sqrt(m_squared_length_3(normal));
+        Vector_3 unoriented_plane_normal =
+        unoriented_plane_of_best_fit.orthogonal_vector();
+
+        const FT normal_length =
+        m_sqrt(m_squared_length_3(unoriented_plane_normal));
 
         CGAL_precondition(normal_length > FT(0));
-        m_normal_of_best_fit = normal / normal_length;
+
+        unoriented_plane_normal /= normal_length;
+
+        // Compute actual direction of plane's normal sign
+        // based on faces belonging to that region.
+        // Approach:
+        // Each face gets one vote to keep or flip the current plane normal.
+        int64_t votes_to_keep_normal = 0;
+        for (std::size_t i = 0; i < region.size(); ++i) {
+          const auto face = *(m_face_range.begin() + region[i]);
+          Vector_3 n;
+          get_face_normal(face, n);
+          bool agrees = n * unoriented_plane_normal > 0;
+          votes_to_keep_normal += (agrees ? 1 : -1);
+        }
+        bool flip_normal = (votes_to_keep_normal < 0);
+
+        m_plane_of_best_fit =
+        flip_normal
+          ? unoriented_plane_of_best_fit.opposite()
+          : unoriented_plane_of_best_fit;
+
+        m_normal_of_best_fit =
+        flip_normal
+          ? (-1 * unoriented_plane_normal)
+          : unoriented_plane_normal;
       }
     }
 


nh2:
Some more questions that I haven't managed to answer yet:

* Both this code and the seed sorting code at https://github.com/CGAL/cgal/blob/master/Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_sorting.h#L212-L220 work on plain unoriented points for fitting the plane, and determining a plane fit score, in the first place. During the fitting they completely ignore orientation, so they would compute a good fit even if half the triangles disagree with the plane (normal). That seems odd.
* Why is the `fit plane to whole triangles` (shown in the example in https://doc.cgal.org/latest/Principal_component_analysis/index.html#Principal_component_analysisLeastSquares_1) not used instead? Also, I don't quite understand what that does: What objective/error function does a whole-plane PCA minimise? Does it take into account the face position/orientation/area? If yes, how does it weigh them to compute a single distance metric?

sgiraudot:
Just for information: region growing started as a point cloud algorithm that was later extended to other types (like surface meshes). In most cases, normals in point clouds are randomly oriented (especially if they are estimated and not measured), which is why most shape detection algorithm (region growing but also RANSAC) ignore the orientation.

I do agree that for surface meshes, the orientation seems more important to avoid propagating a shape through a very sharp feature (a feature whose angle is smaller than the tolerance angle), so at first sight I would agree that this change makes sense (the PR still needs to be reviewed of course).

danston:
Hi @nh2. I am currently reviewing your change and I will answer to your other questions a bit later today. But as Simon has already pointed out, the original intention of the code was with unoriented normals that has an impact on your test cases. If you could also show me the visual output of your problem with two parallel planes and even better send me a minimum reproducible example that I could play around with, it would be great. Thanks.

Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h
```diff
@@ -284,25 +287,58 @@ namespace Polygon_mesh {
...
         CGAL_precondition(normal_length > FT(0));
-        m_normal_of_best_fit = normal / normal_length;
+
+        unoriented_plane_normal /= normal_length;
+
+        // Compute actual direction of plane's normal sign
+        // based on faces belonging to that region.
+        // Approach:
+        // Each face gets one vote to keep or flip the current plane normal.
+        int64_t votes_to_keep_normal = 0;
```
danston: Is there any reason for using `int64_t` here instead of just `long`?

Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h
```diff
@@ -284,25 +287,58 @@ namespace Polygon_mesh {
...
+        for (std::size_t i = 0; i < region.size(); ++i) {
+          const auto face = *(m_face_range.begin() + region[i]);
```
danston: I would do here simply: 
```C++
for (const std::size_t face_index : region) {
const auto face = *(m_face_range.begin() + face_index);
...

File: Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h [hunk 3]
```diff
@@ -284,25 +287,58 @@ namespace Polygon_mesh {
...
+          Vector_3 n;
+          get_face_normal(face, n);
+          bool agrees = n * unoriented_plane_normal > 0;
```
danston: That should be:
```C++
const bool agrees = m_scalar_product_3(face_normal, unoriented_plane_normal) > FT(0);

File: Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h [hunk 4]
```diff
@@ -284,25 +287,58 @@ namespace Polygon_mesh {
...
+
+        // Compute actual direction of plane's normal sign
+        // based on faces belonging to that region.
+        // Approach:
+        // Each face gets one vote to keep or flip the current plane normal.
+        int64_t votes_to_keep_normal = 0;
+        for (std::size_t i = 0; i < region.size(); ++i) {
+          const auto face = *(m_face_range.begin() + region[i]);
+          Vector_3 n;
+          get_face_normal(face, n);
```
danston: rename `n` -> `face_normal`

Shape_detection/include/CGAL/Shape_detection/Region_growing/Region_growing_on_polygon_mesh/Least_squares_plane_fit_region.h
```diff
@@ -284,25 +287,58 @@ namespace Polygon_mesh {
...
+          bool agrees = n * unoriented_plane_normal > 0;
+          votes_to_keep_normal += (agrees ? 1 : -1);
+        }
+        bool flip_normal = (votes_to_keep_normal < 0);
```
danston:
```C++
const bool flip_normal = (votes_to_keep_normal < 0);

danston:
Ok, to answer some of your questions:

- Currently, the angle between each face normal and estimated plane normal is always constrained to the interval [0, 90] degrees that is the cos value used internally always belongs to [0, 1]. It indicates that two normals with the opposite directions will have an angle of 0 instead of 180 degrees. That leads to the problem in your case with two noisy parallel and connected planes. The problem comes from the fact that the code was originally designed to work with data, which lack robust normals. However, you are right, in case of a polygon mesh, we should be more precise.

- PCA does not require any normals and does not estimate them explicitly. In fact, we use PCA to estimate a plane and derive an unoriented normal from it.

- We do not use `fit_plane_to_triangles` from PCA, because our class is designed for an arbitrary polygon mesh, that is a face is not necessary a triangle. Moreover, the main region growing class is designed such that if you do not like the default behavior, you can easily change it. For example, if you know that your normals are correctly oriented in the point cloud, you can create your own `Least_squares_plane_fit_region` for points where you set the correct signs, distances, add new region criteria, etc.

- About what PCA does exactly in the code, you should better address the paper [you cited](https://hal.inria.fr/inria-00327027/document). It has all details. But note that no PCA function returns a normal and regarding your question on internal weights, the weighting scheme depends on the chosen input element. For triangles, the weights are triangle areas rather than normals. This is why, I believe, in [this figure](https://doc.cgal.org/latest/Principal_component_analysis/index.html#fig__PCAfit) the plane is estimated correctly, because the dense area is weighted differently.

- I do not think we should mention the PCA paper in the docs, but we should probably mention that the PCA package is used internally to estimate the plane.

nh2:
@danston Thanks for the responses! I will reply to it soon and also incorporate the review feedback.

Another question, before I forget it:

* Given that we use vertex-based fitting, should we not deduplicate the vertices? A given vertex may be part of multiple faces, so the current use of `vertices_around_face()` will make that vertex count multiple times.

danston:
I thought about that, but I am not sure if this is necessary. If we do it, it is again a problem of sacrificing time versus space. In this case, we need a more difficult algorithm to extract only unique vertices and pass them to the PCA, however the PCA results are no different if we have these points duplicated. Now, we have a simple `push_back` operation, which is super efficient. So, the whole thing requires testing. The test should check which option is faster/better. I may do it for the next version of the package.

danston:
Hi @nh2. You do not have to do your changes that I have mentioned in my prior review. Another user found a different imprecision in the polygon mesh class here #4570. So, I have fixed that issue and I also included your changes in that PR. It also contains changes for docs (that we discussed) and many other little changes. Note, that I mentioned you in the comment inside the code that the fix for normal orientation is proposed by you. You can see the full PR here #4574. We though can still continue the discussion here. I may add these new issues (if any) from the discussion to that PR later.

lrineau:
Replaced by #4574.

PR #4573, opened 2020-03-11
Fix issue with border vertices in Euler::add_face

sloriot:
The bug was here because of an assumption that was wrong (border vertex has a border halfedge pointer)


maxGimeno:

sloriot:
@MaelRL why not yet approved? Do you want to review it before it is merged?

PR #4574, opened 2020-03-11
Shape detection - bug fixes

### LINKED ISSUE #4564: Shape detection: Attempt to detesselate mesh // Unclear points in the manual
## Issue Details

I'm trying to use efficient_RANSAC with Python bindings to detesselate a mesh. Right now, I'm focussing to do this on a simple cube (10x10x10):
image
So the expected output of the efficient_RANSAC algorithm would be 6 planes.

From what I see in the docs, the algorithm expects surface points (location and normal) as you would get from a laser tracker or 3D scanner. Therefore, I implemented an algorithm which randomly and equally fills a triangle with surface points. Their normal is the same as the one from the triangle. Currently, efficient_RANSAC can't find any surfaces, which is surprising since my point cloud seems to be fine:
Screenshot_20200307_174833

So something I can understand is, that the parameters I use are not optimal for this use case. 
Therefore, I took a look into the documentation and I have trouble to understand what the efficient_RANSAC function expects as parameters.
Screenshot_20200307_183503

When working with any kind of functions, I expect something like units mentioned. For example, "normal_threshold" is 0.9 by default which shall be around 25deg, but what is 0.9 then? 25deg as  radians? AFAIK, the radians of 25deg is ~0.44, so the threshold is twice the radians?
Same for cluster epsilon: According to the docs it is "1% of bounding box diagonal". So is the input done on percent? In decimal values? Or the absolute maximum distance as described?

I have to admit that I'm a bit confused at this point and it would be great, if you could take a look onto my code and/or give me some help here. Because right now, I can't be sure how the detection is influenced by changing the values.

Thanks! :wink: 

## Source Code

https://github.com/tetonsim/is-cura-ui/blob/r20.0-tp-selection_using_cgal/SmartSlicePlugin/shape_detection/ShapeDetectedNode.py#L88

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux (doesn't matter I think)
* Compiler: Clang
* Specific flags used (if any): <Nope>
* CGAL version: 5.0.1
* Boost version: Whatever comes with Ubuntu Focal at the moment

## Files
[Generated points](https://github.com/CGAL/cgal/files/4302343/points.csv.log)
[Used mesh (STL)](https://github.com/CGAL/cgal/files/4302344/test.10.stl.log)

-------------------------------

PS: Something I can imagine is, that my points are "too optimal" in my case epsilon is often zero, if the plane is not slightly curved. However, that is only a wild guess.

### LINKED ISSUE #4570: [Shape Detection - Region growing] issue with degenerate faces

## Issue Details

I encountered a problem, when I try to run the example code of region_growing_on_polygon_mesh.cpp
in the shape detection package for mesh segmentation. 

It was stopped at the algorithm running step, which is region_growing.detect(std::back_inserter(regions)).

Still, using the example dataset is fine, but I tried to input another test dataset myself, it turns to collapse. This time I just compared the .off format of the files, they have definitely the same. I have tried several different datasets, and had the same errors. So it seems to be bugs in the codes, isn’t it?

Please see the error information in the attachment. Thank you very much!
[mesh_section_YA.zip](https://github.com/CGAL/cgal/files/4313565/mesh_section_YA.zip)

## Source Code

region_growing_on_polygon_mesh.cpp

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows10
* Compiler: Visual Studio 2017
* CGAL version:5.0.1
* Boost version:1.71.0

danston:
This is a small PR that fixes multiple little issues in the package.

## Summary of Changes

- includes multiple fixes from the PRs: #3864; #4563
- includes a test with the mesh that contains degenerated faces;
- improves docs;
- cleans up examples and tests.

## Release Management

* Issue(s) solved (if any): fix #4564, fix #4570 
* Feature/Small Feature (if any): not a feature
* Link to compiled documentation (obligatory for small feature) [docs](https://cgal.geometryfactory.com/~danston/Shape_detection/index.html)


sgiraudot:
Nice! Since you gathered some fixes here, would you mind adding one for https://github.com/CGAL/cgal/issues/4564#issuecomment-596395632?

What's missing is: in the efficient RANSAC's parameters reference manual, it should be said clearly that the `normal_threshold` parameter is a threshold on the dot product of normals (not a threshold directly on the angle). Thanks!

danston:
@sgiraudot Sure, I am going to add it right now.

lrineau:
As the PR is a bug-fix, I think you should consider basing it on a release branch.

danston:
@lrineau Ok, no problem.

lrineau:
Does it mean that PR #4563  should be closed, and replaced by this one?

danston:
If you know any other bugs/improvements/corrections, which I have not noticed and which should be included in this PR, please let me know. Thanks!

danston:
> Does it mean that PR #4563 should be closed, and replaced by this one?

@lrineau Yes, I have included all those changes and fixed a few other related issues in that class, so it is all in this PR now.

maxGimeno:

PR #4576, opened 2020-03-12
CDT2: Base CDT2's edge sets on geometry to ensure determinism

MaelRL:
## Summary of Changes

When using face addresses, we do get the same combinatorics and geometry in the result, but it is possible to create different TDS' from one run to another, with a given face not having the same order of points in memory.

## Release Management

* Issue(s) solved (if any): -

### DIFF
--- a/Triangulation_2/include/CGAL/Constrained_Delaunay_triangulation_2.h
+++ b/Triangulation_2/include/CGAL/Constrained_Delaunay_triangulation_2.h
@@ -34,6 +34,7 @@
 #include <CGAL/internal/info_check.h>
 #include <CGAL/is_iterator.h>
 
+#include <boost/container/flat_set.hpp>
 #include <boost/iterator/zip_iterator.hpp>
 #include <boost/mpl/and.hpp>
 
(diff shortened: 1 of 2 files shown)


MaelRL:
Runtime takes a little bit of a hit when running `T2/benchmarks/CDT_with_intersections`, roughly 5% slower. The benchmark however has an artificially high number of intersections and thus of flips (which is where the edge set changes matter) and there is no runtime different on `T2/benchmarks/CDT_insert_on_constraints` and `norway.cin`.

I used geometry similarly to 0991efefeee3100c715c820102ee56232a3bb498, but maybe this should be done differently, via IDs or timestamps?

MaelRL:
Seems like you had already made `CDT2` / `CDT_plus_2` pretty much deterministic a little while ago, see e.g. the commit above (0991efe) and similar. This particular `std::set` probably escaped because this precise configuration isn't tested. (I am not even sure how to make it fit into a single test as it shows up only when compiling twice and forcing the memory to change in the second case.)

afabri:
> I am not even sure how to make it fit into a single test as it shows up only when compiling twice and forcing the memory to change in the second case.

You could have two triangulations with the same input, and separated by a line with the memory changer.

MaelRL:
For the record, `flat_set` does provide a speed-up on the same benchmarks. The gain is roughly 5% in favor of the new code on `CDT_with_intersections`, and no change `CDT_insert_on_constraints`.

maxGimeno:

PR #4577, opened 2020-03-12
Fix bug in point_position in the bezier traits

### LINKED ISSUE #94: Bézier Arrangements bug report - segmentation fault upon insertion of a Bézier curve
I provide a simple program which yields a segmentation fault when inserting Bézier curves to a Bézier Arrangement. I believe this error is related to the kernel used within the Bézier arrangement. I've shared the details at the following link which includes the detailed information, a very simple program, an image to display the error condition and the stack trace.

Please check the link for the shared files. For anything related to this issue, you can contact to me via e-mail: myirci@gmail.com

https://drive.google.com/file/d/0ByXhUOvsZStITS1iaDFFT2NxVjg/view?usp=sharing

theartful:
## Summary of Changes

In `point_position` in the case when `p` is rational, the following equation is solved:
```
curve.x_polynomial() / norm_x = numerator(p.x) / denominator(p.x) 
<=>
curve.x_polynomial() *  denominator(p.x) = numerator(p.x) * norm_x 
```
to find the point where `X(t) = p.x`, and then compute `Y(t)` and compare it to `p.y` to determine whether `p` lies above or below the curve. The bug was that `norm_x` was not multiplied to `numerator(p.x)`

## Release Management

* Affected package(s):  2D Arrangements
* Issue(s) solved (if any): fix #94
* License and copyright ownership: The license used by CGAL

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
@@ -959,11 +959,12 @@ _Bezier_x_monotone_2<RatKer, AlgKer, NtTrt, BndTrt>::point_position
 
 
   if ( p.is_rational() ){
-    const Rational& px = ((Rat_point_2) p).x();
+    const Rational px = ((Rat_point_2) p).x();
 
     Integer denom_px=nt_traits.denominator(px);
     Integer numer_px=nt_traits.numerator(px);
-    Polynomial poly_px = CGAL::sign(numer_px) == ZERO ? Polynomial() : nt_traits.construct_polynomial(&numer_px,0);
+    Integer poly_px_scale = numer_px * _curve.x_norm();
+    Polynomial poly_px = CGAL::sign(numer_px) == ZERO ? Polynomial() : nt_traits.construct_polynomial(&poly_px_scale,0);
     Polynomial poly_x = nt_traits.scale(_curve.x_polynomial(),denom_px) - poly_px;
 
     std::vector <Algebraic> roots;


Arrangement_on_surface_2/include/CGAL/Arr_geometry_traits/Bezier_x_monotone_2.h
```diff
@@ -959,11 +959,12 @@ _Bezier_x_monotone_2<RatKer, AlgKer, NtTrt, BndTrt>::point_position
 
 
   if ( p.is_rational() ){
-    const Rational& px = ((Rat_point_2) p).x();
```
theartful: Rat_point_2::x() returns a value, so capturing it by const ref is UB, and it indeed caused problems.

sloriot: Strange. Were you able to reproduce the crash and was this diff fixing the segfault?

theartful: With some test cases, it would segfault, and in others it would get assertion error, I didn't check the cause of segfault before, but yes, now I checked it. The crash is reproduced consistently, and this line was the cause of the segfault.

theartful: However, this change alone doesn't fix the bug.

theartful: I checked out the reason of this behavior. It comes out to the fact that returning a `const ref` from a `pr value` is not safe. The `pr value` may be destructed, and with it its member variables.
Example: 
https://gist.github.com/theartful/d3a8b85e8d8e9cbe465bf570b740f8ab

lrineau: The problem here is that it is a const ref to a **sub-object** of a PR-value. If it was to the PR-value itself, then the lifetime extension would have applied. See https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary. But binding a reference to a sub-object does not extend the lifetime of the object.

theartful: Nice! Thanks for clearing that up.

maxGimeno:

sloriot:
Thanks for the patch!

PR #4579, opened 2020-03-13
Handle non-manifold vertices in copy_face_graph

### LINKED ISSUE #3206: Handling of non-manifold vertices
The way I wrote `copy_face_graph()` and internal functions used in Boolean operations through corefinement implies that if a mesh to be copied features a non-manifold vertex, only one umbrella of faces will have a new vertex created and assigned.

What do we want to do?
- keep the non-manifold vertex
- duplicate the vertex and resolved at the same time the non-manifold situation
- ignore the issue and requires that the mesh does not contain non-manifold vertices

sloriot:
There should not be an overhead if there are no non-manifold umbrellas (only a extra loop over halfedges).

I also:
- removed the internal mecanism for the hash-map selection and replaced it with a dynamic map.
- fix compilation issues of the dynamic map with `Face_filter_graph`
- fix compilation issues of the dynamic map with LCC. @gdamiand : there is probably a better solution if you want to update the code (instead of using the default I mean, not the trick to avoid the ambiguity)

Possible TODO: update in the corefinement code in face_graph_utils.h the method importing a patch.

Fixes #3206
No backport to 4.14 on purpose.

### DIFF
--- a/BGL/examples/BGL_polyhedron_3/copy_polyhedron.cpp
+++ b/BGL/examples/BGL_polyhedron_3/copy_polyhedron.cpp
@@ -79,7 +79,6 @@ int main(int argc, char* argv[])
     boost::unordered_map<source_vertex_descriptor, tm_vertex_descriptor> v2v;
     boost::unordered_map<source_halfedge_descriptor, tm_halfedge_descriptor> h2h;
     boost::unordered_map<source_face_descriptor, tm_face_descriptor> f2f;
-    CGAL::copy_face_graph(T1, S, std::inserter(v2v, v2v.end()), std::inserter(h2h, h2h.end()));
     std::ofstream out("reverse.off");
     out << S;
     CGAL::copy_face_graph(T1, S, CGAL::parameters::vertex_to_vertex_map(boost::make_assoc_property_map(v2v))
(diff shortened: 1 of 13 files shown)


sloriot:
Issue with dynamic maps and Triangulation_2. A fix similar to what I did is needed for other graph classes is needed.

sloriot:
At least `remesh_polyhedral_surface_sm` does not compile.

BGL/examples/BGL_polyhedron_3/copy_polyhedron.cpp
```diff
@@ -79,7 +79,6 @@ int main(int argc, char* argv[])
     boost::unordered_map<source_vertex_descriptor, tm_vertex_descriptor> v2v;
     boost::unordered_map<source_halfedge_descriptor, tm_halfedge_descriptor> h2h;
     boost::unordered_map<source_face_descriptor, tm_face_descriptor> f2f;
-    CGAL::copy_face_graph(T1, S, std::inserter(v2v, v2v.end()), std::inserter(h2h, h2h.end()));
     std::ofstream out("reverse.off");
     out << S;
```
lrineau, line 83: That seems to change the semantic of the example a lot. Now that the first call to `    CGAL::copy_face_graph` is removed, then `S` is empty in `out << S;`.

sloriot: the next two lines are useless and I'll remove them.

maxGimeno:

PR #4584, opened 2020-03-18
[Small Feature] Custom Property Map Example

sgiraudot:
## Rationale

CGAL uses quite extensively property maps. I noticed that we often advise users to write their own property maps to apply CGAL algorithms to their data, but we don't give any indication on how to do it. Of course, because property maps are originated from Boost, indications are technically available on the Boost Manual. But it can't hurt to add more information to our own manual, just to have somewhere where we can orient users when we suggest they should write their own property map.

This (very) small feature simply adds a section with example to the _Property Map_ package.

## Summary of API changes

No API change, just a new example and section in the manual.

(no change)

## CHANGES.md

No need

## Submission

* [__User manual__](https://cgal.geometryfactory.com/~sgiraudot/Pmap_example/Property_map/index.html#title5)

## Status

- [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Custom_Property_Map_Example) **Pre-approved** 2020/03/24

### DIFF
--- a/Property_map/examples/Property_map/custom_property_map.cpp
+++ b/Property_map/examples/Property_map/custom_property_map.cpp
@@ -16,7 +16,7 @@ struct Custom_point_map
   using value_type = Point_3;   // The object manipulated by the algorithm is a Point_3
   using reference = Point_3;    // The object does not exist in memory, so there's no reference
   using category = boost::readable_property_map_tag; // The property map is only used for reading
-    
+
   double *x, *y, *z;
 
   Custom_point_map (double* x = nullptr, double* y = nullptr, double* z = nullptr)
@@ -65,7 +65,7 @@ struct Custom_normal_map
 int main()
 {
   constexpr std::size_t nb_points = 1000;
-  
+
   // in this example, points are stored as separate coordinate arrays
   double x[nb_points];
   double y[nb_points];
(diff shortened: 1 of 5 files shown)


Property_map/examples/Property_map/custom_property_map.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+
+int main()
+{
+  std::size_t nb_points = 1000;
+
+  // in this example, points are stored as separate coordinate arrays
+  double* x = new double[nb_points];
+  double* y = new double[nb_points];
+  double* z = new double[nb_points];
```
lrineau: I think we should avoid any use of `new` and `delete` in examples. The explicit use of `new` and `delete` is strongly discouraged by any tutorial or guidelines about C++11. We should follow those guidelines.

lrineau: And so are macros to define constants. `constexpr auto nb_points = 1000;` is preferable to the `#define`.

maxGimeno:

sloriot:
Tested in CGAL-5.1-Ic-114

The new warning is due to Eigen.

sloriot:
IMO, there is no need to update changes.md here.

PR #4588, opened 2020-03-20
Polyhedron Demo: Fix ssh scene 

maxGimeno:
## Summary of Changes

When pushing a file with scp, it seems the permissions were not right. This PR replaces 0707 with 0777
## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Use_ssh.cpp
+++ b/Polyhedron/demo/Polyhedron/Use_ssh.cpp
@@ -182,7 +182,7 @@ bool push_file(ssh_session &session,
     return false;
   }
   //push a file to /tmp
-  res = ssh_scp_push_directory(scp, ".", 0707);
+  res = ssh_scp_push_directory(scp, ".", 0777);
   if (res != SSH_OK)
   {
     std::cerr<<"Can't create remote directory: %s\n"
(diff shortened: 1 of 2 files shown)


sloriot:
Tested locally between me and @maxGimeno

Polyhedron/demo/Polyhedron/Use_ssh.cpp
```diff
@@ -182,7 +182,7 @@ bool push_file(ssh_session &session,
     return false;
   }
   //push a file to /tmp
-  res = ssh_scp_push_directory(scp, ".", 0707);
+  res = ssh_scp_push_directory(scp, ".", 0777);
```
lrineau: Do you have any idea what 0777 mean?!

It means that everybody on the system can read, write, access to it with `cd`, like in `/tmp`. The problem read is the "write" access: that is a serious security issue!

A safe mode for a directory is 0755 (aka `rwxr-xr-x`): everybody can read and enter the directory,  but not write to it. I will push a commit to `master` directly to fix that issue.

Polyhedron/demo/Polyhedron/Use_ssh.cpp
```diff
@@ -191,7 +191,7 @@ bool push_file(ssh_session &session,
     return false;
   }
   res = ssh_scp_push_file
-      (scp, dest_path, size, 0707);
+      (scp, dest_path, size, 0777);
```
lrineau: And then, for a file, the meaning of `0777` is a bit different, but the security issue is the same: the "write" permission should not be given to everybody. The correct mode is `0644`.

PR #4591, opened 2020-03-21
Improve NP documentation

### LINKED ISSUE #4300: Cleaning Named Parameter Documentation
[The named parameter page](https://doc.cgal.org/latest/Polygon_mesh_processing/group__pmp__namedparameters.html) (and similarly for PSP and other packages) contains the documentation of parameters that are used by only one function. There is no need to documented them on that page and the relevant information should be moved to the function.

Adding the documentation of one parameter on that page should be motivated by a good enough reason. For example for `vertex_point_map` we document that the default is the internal vertex point map of the mesh. For `geom_traits` we can says that the default is deduced from the vertex point map and that the value type of the vertex point map should be the nested `Point_3` type of the class.

Opinions?

MaelRL:
## Summary of Changes

The documentation of named parameters is tedious to use: a good part of the documentation is left on another page (NamedParameters.txt), so important details or default values are often not easily understandable without looking at the code.

This PR aims to improve this by putting the full documentation of each named parameter into the function that makes use of this paramter, but making it collapsible such that the documentation of a function does not become its own chapter.

TODO:
- [x] BGL
- [X] PMP
- [X] PSP
- [x] Get rid of Pkg/NamedParameters.txt
- [x] proof read. VPM types, callback parallel/return types.
- [x] Fix "\param np an optional \ref bgl_namedparameters "Named Parameters" described below"
- [x] Others
- [x] Duplicate css/js into other doxygen versions

## Release Management

Waiting for other 5.1 PRs that modify NPs to be integrated...

* Affected package(s): All
* Issue(s) solved (if any): fix #4300


MaelRL:
Below is the current look, with actual code and not just photoshopped. 

@maxGimeno @janetournois I don't know about showing a description even when not expanded, I thought that actually hiding everything by default both makes NPs look more optional as a user doesn't have to lose time by reading stuff which he will not use 90+% of the time, and makes the rest of the doc seems lighter. Therefore, I thought it was a good idea to hide, but if everyone thinks it's better with the description, it can be added.

> Is it possible to create and use a doxygen macro like
> `\CGAL_np{np_name}{np_short_description}{np_type}{np_default}{np_extra}`
> that would be used as follows :
> -- if np_name is known (like geom_traits), its description is used, and the writer can simply write `\CGAL_np{geom_traits}`. The description > would be automatically added (and consistent with other functions)
> -- otherwise, all the parameters are needed (but np_extra which is optional). The writer needs to write a short description, a type, and a default. This would force everybody to follow the documentation rules? Then it would look like : `\CGAL_np{my_new_np}{"my short description"}{double}{1.1}{"this is optional extra information"}`
> 

Right now, I use an alias `cgal_np{4-5}` which is then used like `cgal_np{parameter, description, type, default, optional extra}` (see [this PR's files modified](https://github.com/CGAL/cgal/pull/4591/files)). I don't know if it's possible to match based on the parameter to another alias like you suggest (@sloriot ?). My plan was to create individual macros such as `\cgal_np_VPM{2}` (2 for type and variable name) and use it like that instead. It is then always possible to use `\CGAL_np{}` if your function is very special

I'm anyway not very happy with the current aliases anyway because parameters in an alias call are separated by commas, and so if you want to have a comma in your description, you have to escape it like `\,`, which is too tedious to be used in practice (@sloriot maybe a suggestion here too please?).


new_np

janetournois:
The new look is very nice!
Indeed it does not matter much to have no short description displayed by default, because it is easy to "unfold".

I agree that the issue about commas in the macro is not very easy to use. If you manage to find a better option, it would be great!

maxGimeno:
Yes it does look nice ! 
About the small description, maybe you are right, especially as most of the time the parameter name is enough to understand what it is about.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/clip.h
```diff
@@ -688,6 +753,7 @@ void split(TriangleMesh& tm,
   *
   * \note In the current implementation it is not possible to set the vertex point map and the default will be used. `Iso_cuboid_3` must be
   * from the same %Kernel as the point of the vertex point map.
```
MaelRL: @sloriot can you please comment on what that note means? The mesh's vpm seems to be used just fine. Does this refer to an hypothetical VPM associated to the iso cuboid?

MaelRL:
This PR is just documentation improvements, but I feel it could deserve at least a night of testing before being pushed to 5.2: it was ready before the code freeze and I was (still am) just waiting for PRs modifying named parameter documentation of 5.1 to be integrated so that late PRs do not in addition have to deal with conflicts coming from the new doc macros / formatting.

maxGimeno:
Warnings : https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1-Ic-199/index.html

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-200.shtml

sloriot:
@lrineau If this PR is tested I see no reason not to merge it. It is only a documentation improvement.

PR #4595, opened 2020-03-23
PMP: Add a named_parameter to split_connected_components()

maxGimeno:
## Summary of Changes
Add `face_patch_map` as NP to `split_connected_components()` to allow giving an existing map to the function and avoid having to recompute it.
## Release Management

* Affected package(s):Polygon_mesh_processing


Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_clip.cpp
```diff
@@ -528,9 +528,15 @@ void test_split()
...
+  //try with np
+  typename boost::template property_map<
+  TriangleMesh, CGAL::dynamic_face_property_t<int > >::type
+      pidmap = get(CGAL::dynamic_face_property_t<int>(), tm1);
+  CGAL::Polygon_mesh_processing::connected_components(
+        tm1, pidmap, CGAL::parameters::all_default());
   PMP::split_connected_components(tm1,
                                   meshes,
-                                  params::all_default());
+                                  params::face_patch_map(pidmap));
```
MaelRL: Might want to check the result and not just the call.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -986,6 +1011,9 @@ void split_connected_components_impl(FIMap fim,
  *   \cgalNPBegin{halfedge_index_map}
  *     a property map containing a unique index for each halfedge initialized 0 to `num_halfedges(pm)`
  *   \cgalNPEnd
+ *  \cgalParamBegin{face_patch_map} a property map with the patch id's associated to the
+     faces of `pm`. Instance of a class model of `ReadPropertyMap`.
```
sloriot:
```suggestion
     faces of `pm`. Instance of a class model of `ReadPropertyMap`. If not provided, an internal map will be filled with a call to `connected_components()` with `edge_is_constrained_map()` (if provided).
```

sloriot: You might want to add something similar in the function doc too.

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_clip.cpp
sloriot: For example put several time the same `off` in a given mesh and check that you have all of them as output + validity + #face/#verteces

maxGimeno: the result IS tested. Ok, it is expected to be the same result as the default call, but a result nontheless.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
maxGimeno: If I also add it to the function, it will be said twice, and then somebody else will tell me to remove it.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/connected_components.h
```diff
@@ -931,23 +931,48 @@ struct No_mark
...
       PolygonMesh, CGAL::dynamic_face_property_t<int > >::const_type
-      pidmap = get(CGAL::dynamic_face_property_t<int>(), tm);
+      > ::type Fpm;
+
+  using parameters::choose_parameter;
+  using parameters::get_parameter;
+  using parameters::is_default_parameter;
+
+  Fpm pidmap = choose_parameter(get_parameter(np, internal_np::face_patch),
+                                get(CGAL::dynamic_face_property_t<int>(), tm));
```
MaelRL, line 955: the face component map of PMP::CC must have value type `boost::graph_traits<PolygonMesh>::faces_size_type`

sloriot:
One last warning:
```
CGAL-5.1-Ic-114/Polygon_mesh_processing/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz:C:\CGAL\test\CGAL-5.1-Ic-114\include\CGAL/Polygon_mesh_processing/connected_components.h(980): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
```

maxGimeno:

PR #4599, opened 2020-03-25
Bugfix : Spatial Searching Interruptible Distance

sgiraudot:
## Summary of Changes

`Orthogonal_k_neighbor_search` is broken when dimension is 6 or higher (segfault). The reason is there's an optimization for `dim>=6` that misuses an iterator. 

## Release Management

* Introduced on master


lrineau:
Is that a bug-fix only for the `master` branch?

sgiraudot:
It was introduced after 5.0.

maxGimeno:

PR #4604, opened 2020-03-31
[Doc fix] Swapped figures in PSP Registration

sgiraudot:
## Summary of Changes

Spotted by @nmellado: the first two lines of [the array of figures for Registration](https://cgal.geometryfactory.com/CGAL/doc/master/Point_set_processing_3/index.html#fig__Point_set_processing_3tableRegistrationRegistration_visualization_table) in the doc of PSP are swapped (the OpenGR line is supposed to show coarsely registered point clouds). This PR swaps them back.

## Release Management

__Note:__ I swapped the files using `git mv` so that the naming is consistent (I could have just moved the figures but someone checking the Doxygen source might wonder why the names are not fitting the columns). I don't know if it is possible to make this efficient from a file weight point of view (I think git treats this as removal+addition of files, not just renaming).

* Affected package(s): PSP (doc only)


sloriot:
Without trying to compile, if you rename files without changing the names in .txt I don't see it can work.

sgiraudot:
The names are just swapped, there are no "new names", they fit the names already existing in the .txt.

lrineau:
> I think git treats this as removal+addition of files, not just renaming

Git always treat renaming as removal+addition, internally, but the diffing tools in Git sometimes detect that it corresponds to a renaming.

PR #4606, opened 2020-03-31
Deduce reference return type in approx()

mglisse:
## Summary of Changes

In Epeck_d, `*approx(iterator)` was copying the approx part of the element instead of returning a reference. With a dynamic dimension, this caused a ton of allocations. The patch saves 12% of running time on the testcase I was looking at.

## Release Management

* Affected package(s): Filtered_kernel, but really NewKernel_d


sloriot:
Tested in CGAL-5.1-Ic-114

PR #4611, opened 2020-04-02
replaced throw() with noexcept

rathod-sahaab:
## Summary of Changes
**Small change:** replaced all `throw()` with `noexcept`

_throw() is equivalent to noexcept(true) but was deprecated with C++11 and will be removed with C++20._ [source](https://www.modernescpp.com/index.php/c-core-guidelines-the-noexcept-specifier-and-operator)

### DIFF
--- a/STL_Extension/include/CGAL/Uncertain.h
+++ b/STL_Extension/include/CGAL/Uncertain.h
@@ -67,7 +67,7 @@ public:
   Uncertain_conversion_exception(const std::string &s)
     : std::range_error(s) {}
 
-  ~Uncertain_conversion_exception() throw() {}
+  ~Uncertain_conversion_exception() noexcept {}
 };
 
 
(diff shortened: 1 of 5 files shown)


STL_Extension/include/CGAL/Uncertain.h
```diff
@@ -67,7 +67,7 @@ class Uncertain_conversion_exception
   Uncertain_conversion_exception(const std::string &s)
     : std::range_error(s) {}
 
-  ~Uncertain_conversion_exception() throw() {}
+  ~Uncertain_conversion_exception() noexcept {}
```
mglisse, line 70: Isn't `noexcept` the default for destructors? So we could remove `throw()` there instead of replacing it with `noexcept`. Quite possibly we could even remove the destructor...

lrineau: Indeed.

https://en.cppreference.com/w/cpp/language/destructor
> if no exception specification is explicitly provided, the exception specification is considered to be one that would be used by the implicitly-declared destructor. In most cases, this is noexcept(true).

rathod-sahaab: Sorry I have been a bit busy with my GSoC project at Inkscape, and as it's hosted on GitLab I rarely visit GH.
Will do this soon :heart:

maxGimeno:

PR #4612, opened 2020-04-02
Fix some tests intializing or changing data in assertions

### LINKED ISSUE #4609: test_constrained_triangulation_2 does not pass in Release
Since the merge of PR #4163 in `master`, the test `test_constrained_triangulation_2` does not pass with default CMake option in `Release`.

@maxGimeno and I detected that while trying to benchmark triangulations (see https://github.com/CGAL/cgal/pull/4496#issuecomment-607814445), because `ctest -L Triangulation_2` failed, in `master`, if cmake was configured with `-DCMAKE_BUILD_TYPE=Release`. We found strange that the test suite did not caught that error.

After use of git-bisect, I found that PR #4163 introduced the issue. But, that did not explain why the CGAL testsuite was fine with it. Eventually, I found that in `master` (23aca30a0493081478181a72a9572bc67f9d90b0):
  - `ctest -R test_constrained_triangulation_2` *passes* if one compiles with `-O3 -DCGAL_NDEBUG` (like in the testsuite)
  - `ctest -R test_constrained_triangulation_2` *does not* pass if one compiles with `-O3 -DNDEBUG` (like in the default CMake flags for Release).

MaelRL:
## Summary of Changes

I only checked BGL (because I vaguely remembered seeing that before) and T2 (to fix #4609)

## Release Management

* Affected package(s): `BGL`, `Triangulation_2`
* Issue(s) solved (if any): fix #4609

### DIFF
--- a/BGL/test/BGL/test_Prefix.h
+++ b/BGL/test/BGL/test_Prefix.h
@@ -156,7 +156,8 @@ std::vector<LCC> lcc_data()
 template <typename Graph>
 struct Surface_fixture_1 {
   Surface_fixture_1() {
-    assert(read_a_mesh(m, "data/fixture1.off"));
+    const bool is_reading_successful = read_a_mesh(m, "data/fixture1.off");
+    assert(is_reading_successful);
     assert(CGAL::is_valid_polygon_mesh(m));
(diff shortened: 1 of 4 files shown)


maxGimeno:

PR #4619, opened 2020-04-06
[Bugfix] PSP Jet Smoothing do not alter points when reading them

sgiraudot:
## Summary of Changes

I introduced a bug with https://github.com/CGAL/cgal/pull/4552: because points are not longer copied in the KD Tree, altering the point cloud means altering the KD Tree. In the case of the jet smoothing, that means that the projection was computed on a subset of points that could contain already altered points (whereas it should be computed on the original points). In parallel mode, it's even worse because some points could be read and written at the same time, I did not spot this possibility when I wrote it…

Anyway, this PR separates computation of new points from their update and fixes it.

## Release Management

### DIFF
--- a/Point_set_processing_3/include/CGAL/jet_smooth_point_set.h
+++ b/Point_set_processing_3/include/CGAL/jet_smooth_point_set.h
@@ -192,26 +192,42 @@ jet_smooth_point_set(
   Point_set_processing_3::internal::Callback_wrapper<ConcurrencyTag>
     callback_wrapper (callback, nb_points);
 
+  std::vector<typename Kernel::Point_3> smoothed (points.size());
+
+  typedef boost::zip_iterator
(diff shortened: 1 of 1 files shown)


maxGimeno:
- error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-118/Point_set_processing_3/TestReport_lrineau_Ubuntu-latest-GCC6.gz
 - warning : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-118/Polyhedron_Demo/TestReport_lrineau_Ubuntu-Latest.gz

maxGimeno:
Success in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-119.shtml

PR #4622, opened 2020-04-06
Fix reference-to-null-pointer UB

### LINKED ISSUE #4617: Undefined behavior: dereferencing null pointer in hash function
## Issue Details

[The default hash function](https://github.com/CGAL/cgal/blob/8795725dba418060a92759d94cb6160742951879/Hash_map/include/CGAL/Handle_hash_function.h#L31) for `Hash_map` hashes all types with this code:
```c++
return std::size_t(&*h) /
    sizeof( typename std::iterator_traits<H>::value_type);
```
When `H` is a pointer type the value `h` will occasionally be null - usually because of default-constructed Handle objects. At that point the execution of the above code is undefined behavior and is flagged by gcc's UB sanitizer.

I found that adding a pointer specialization that detects `nullptr` cases and returns 0 eliminates the errors, though this may not be the optimal approach.

## Related bugs

- @bobergj pointed out this issue in a comment on #2587
- similar issues were noted in #139

## Environment

* Linux (Ubuntu 19.04)
* gcc 8.3.0
* debug mode
* -fsanitize=undefined
* CGAL 5.0.2
* Boost 1.67

jefftrull:
## Summary of Changes

Hash compact_container iterators directly from their underlying
pointers, rather than taking the address of their dereferenced objects.

This avoids situations where a reference may be constructed from a null pointer.

## Release Management

* Affected package(s):
* fixes #4617
* License and copyright ownership: As CGAL sees fit.
----

### DIFF
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -39,7 +39,6 @@
 #include <CGAL/iterator.h>
 #include <CGAL/CC_safe_handle.h>
 #include <CGAL/Time_stamper.h>
-#include <CGAL/Handle_hash_function.h>
 
 #include <boost/mpl/if.hpp>
 
@@ -1316,6 +1315,10 @@ namespace internal {
 
 namespace handle {
   // supply a specialization for Hash_functor
+
+  // forward declare base template
+  template <class H> struct Hash_functor;
+
   template<class DSC, bool Const>
   struct Hash_functor<CC_iterator<DSC, Const>>{
     std::size_t
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -39,6 +39,7 @@
 #include <CGAL/iterator.h>
 #include <CGAL/CC_safe_handle.h>
 #include <CGAL/Time_stamper.h>
+#include <CGAL/Handle_hash_function.h>
 
 #include <boost/mpl/if.hpp>
 
@@ -1310,8 +1311,20 @@ namespace internal {
   std::size_t hash_value(const CC_iterator<DSC, Const>&  i)
   {
     typedef Time_stamper_impl<typename DSC::value_type> Stamper;
-    return Stamper::hash_value(&*i);
+    return Stamper::hash_value(i.operator->());
   }
+
+namespace handle {
+  // supply a specialization for Hash_functor
+  template<class DSC, bool Const>
+  struct Hash_functor<CC_iterator<DSC, Const>>{
+    std::size_t
+    operator()(const CC_iterator<DSC, Const>& i)
+    {
+      return hash_value(i);
+    }
+  };
+} // namespace handle
 
 } // namespace internal
 


STL_Extension/include/CGAL/Compact_container.h
```diff
@@ -1109,9 +1110,21 @@ namespace internal {
   std::size_t hash_value(const CC_iterator<DSC, Const>&  i)
   {
     typedef Time_stamper_impl<typename DSC::value_type> Stamper;
-    return Stamper::hash_value(&*i);
+    return Stamper::hash_value(i.operator->());
```
afabri, line 1113: This seems a valid fix to me, as [`Stamper::hash_value()`](https://github.com/CGAL/cgal/blob/master/STL_Extension/include/CGAL/Time_stamper.h#L84) checks if the pointer is a null pointer.

STL_Extension/include/CGAL/Compact_container.h
```diff
@@ -31,6 +31,7 @@
 #include <CGAL/CC_safe_handle.h>
 #include <CGAL/Time_stamper.h>
 #include <CGAL/Has_member.h>
+#include <CGAL/Handle_hash_function.h>
```
afabri, line 34: In case this include is really needed, and cannot be replace with a forward declaration (@lrineau can you tell us), then in several  files `Package/package_info/Package/dependencies` you must add `Hash_map`.  I copy paste from travis.

Differences in Triangulation: Hash_map are new and not committed.
Differences in Surface_mesh_simplification: Hash_map are new and not committed.
Differences in Surface_mesh_shortest_path: Hash_map are new and not committed. 
Differences in Surface_mesh_segmentation: Hash_map are new and not committed.
Differences in STL_Extension: Hash_map are new and not committed.
Differences in Spatial_searching: Hash_map are new and not committed.
Differences in Snap_rounding_2: Hash_map are new and not committed.
Differences in AABB_tree: Hash_map are new and not committed.

MaelRL:
Could you please rebase the fix on the [CGAL 4.14 branch](https://github.com/CGAL/cgal/tree/releases/CGAL-4.14-branch) please?

jefftrull:
> Could you please rebase the fix on the [CGAL 4.14 branch](https://github.com/CGAL/cgal/tree/releases/CGAL-4.14-branch) please?

Just to be sure - you would like me to produce a second PR that addresses the 4.14 branch?

STL_Extension/include/CGAL/Compact_container.h
jefftrull: From my experiments with gcc it looks like multiple forward declarations of a class template are not an issue so I will try that route.

MaelRL:
No, the branch 4.14 will be merged into newer release branches and into master, so a single PR based on 4.14 is sufficient.

Just a call to `git rebase --onto cgal/releases/CGAL-4.14-branch cgal/master` (and a force push) should do the job.

Thank you!

jefftrull:
> No, the branch 4.14 will be merged into newer release branches and into master, so a single PR based on 4.14 is sufficient.

I see. Do I need to close this PR and create a new one targeted to 4.14 instead?

MaelRL:
This isn't necessary : you can force push the rebased branch, and change the base into which you're merging at the top by clicking the `edit` button at the top, right of the PR title.

STL_Extension/include/CGAL/Compact_container.h
lrineau: It can be a (forward-)declaration, without including the header that defines the template, because one can declare and define a partial specialization of a template that is only declared.

jefftrull:
OK I _think_ I did it correctly :) Now rebased to 4.14 and using forward decl instead of adding a header.

afabri:
@maxGimeno is this a timeout in travis?

maxGimeno:
@afabri no it's not, but it is probably a false positive, as the branch has been rebased on 4.14. It always cause problems with travis.

maxGimeno:
Maybe we should check for trailing whitespaces as travis is brokjen, though.

sloriot:
Thanks for the patch!

PR #4623, opened 2020-04-07
Allow Classification package to be used without Boost IO Stream and Serialization

sgiraudot:
## Summary of Changes

First step of an on-going work to get rid of dependencies to _Boost IO Streams_ and _Boost Serialization_. Currently, compiling any program using CGAL Classification without these creates a compilation error (or at least a link error). While this is consistent with the documentation that states that these 2 Boost packages are hard dependencies of Classification, in practice only `CGAL::Classification::ETHZ::Random_forest::save_configuration()` and `CGAL::Classification::ETHZ::Random_forest::load_configuration()` depend on them.

This PR adds some CMake definitions and preprocessor conditions to make Classification usable without _Boost IO Streams_ and _Serialization_ (provided, of course, that the IO functions of the Random Forest are not used). This does not change anything from a user point of view (documentation is unchanged) but it should unlock the pip situation for @maxGimeno until new dependence-free IO functions are introduced.

## Release Management

### DIFF
--- a/Classification/include/CGAL/Classification/ETHZ/Random_forest_classifier.h
+++ b/Classification/include/CGAL/Classification/ETHZ/Random_forest_classifier.h
@@ -38,10 +38,12 @@
 
 #include <CGAL/tags.h>
 
+#if defined(CGAL_LINKED_WITH_BOOST_IOSTREAMS) && defined(CGAL_LINKED_WITH_BOOST_SERIALIZATION)
 #include <boost/archive/text_iarchive.hpp>
 #include <boost/archive/text_oarchive.hpp>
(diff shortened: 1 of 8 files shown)


maxGimeno:
there is still [this kind of error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-121/Polyhedron_Demo/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz) in the testsuite

sgiraudot:
Well, yes, the PR _allows_ the package to be used the dependencies, but on these platforms, they are found and used (and there's apparently a link error), so there's no reason this PR should fix that.

maxGimeno:

PR #4624, opened 2020-04-07
Documentation: Add a page that explains how to use cmake

maxGimeno:
## Summary of Changes

Adds page in the developer manual that explains how to create a CMakeLists to use with CGAL.
## Release Management

* Affected package(s):Manual
* Issue(s) solved (if any): fix #3105
* Link to compiled documentation (obligatory for small feature) 
[Getting Started ](https://cgal.geometryfactory.com/~mgimeno/doc/CMakeTuto/Manual/general_intro.html)
[the new page](https://cgal.geometryfactory.com/~mgimeno/doc/CMakeTuto/Manual/devman_create_and_use_a_cmakelist.html)

PR #4636, opened 2020-04-09
CMake: Use Imported Targets for 3rd Party Libs

sgiraudot:
## Summary of Changes

Following [this discussion](https://github.com/CGAL/cgal/pull/4624), this is a PR to switch from the function-based 3rd party handling (`CGAL_target_use_[lib]()`) to the imported targets framework (`target_link_libraries([myprogram] CGAL::CGAL CGAL::[lib]_support)`).

## Release Management

* Affected package(s): all (cmake)


lrineau:
That looks good, so far. The `CMakeLists.txt` is complex, but it was complex before this set of changes anyway.

sgiraudot:
Alright, I did a full pass on the whole library, all `target_use` should be replaced by imported targets. I updated the doc of 3rd party libraries accordingly.

I completely removed all the `CGAL_target_use_...` functions as they were never in a release (no need to deprecate them). The only exception is `CGAL_target_use_TBB` for which I added a deprecation message.

lrineau:
I think this PR should also change the sentence we find now in the developer manual:
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1-Ic-127/output2/Manual/devman_create_and_use_a_cmakelist.html#seclink
> There are also some cmake macros to link with CGAL dependencies that can be found in the section Essential and Optional Third Party Dependencies.

It was not macros, but functions, and anyway now it will become imported targets.

lrineau:
By the way, do we want this PR in CGAL-5.1? I think so.

sloriot:
Side remark, I noticed that for basic viewer some extra flags must be passed. I wonder whether a similar macro shouldn't be defined . CC @gdamiand

gdamiand:
> Side remark, I noticed that for basic viewer some extra flags must be passed. I wonder whether a similar macro shouldn't be defined . CC @gdamiand

Only add_definitions(-DCGAL_USE_BASIC_VIEWER); not sure it deserves an imported target ?

Maybe we can define this flag automatically when cgal_qt5 is defined ? (and thus automatically enable basic viewer when cgal_qt5 is found) ?

sgiraudot:
> By the way, do we want this PR in CGAL-5.1? I think so.

I strongly agree: the functions `CGAL_target_use...` were never in a release, integrating this PR directly will save users the trouble of changing their cmake scripts twice.

lrineau:
@sgiraudot There are Travis errors to fix before it can be tested.

maxGimeno:
errors : 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-150/Polyhedron_Demo/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz

and I think 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-150/Polyhedron_Demo/TestReport_lrineau_ArchLinux-clang.gz
also

maxGimeno:
Some hints might be found here : https://github.com/PointCloudLibrary/pcl/pull/3341/files

maxGimeno:
Conflicts

lrineau:
@sgiraudot @maxGimeno In my opinion, this PR is one that we want in 5.1 (a blocker). You should fix the conflicts as soon as possible, and test the PR again in `integration`.

sgiraudot:
I'm on it.

sgiraudot:
Rebased.

maxGimeno:
Thank you !

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-168.shtml

Installation/cmake/modules/CGAL_Eigen_support.cmake
```diff
@@ -0,0 +1,9 @@
+if(EIGEN3_FOUND AND NOT TARGET CGAL::Eigen_support)
+  if(NOT TARGET Threads::Threads)
+    find_package(Threads REQUIRED)
+  endif()
```
lrineau, line 4: @sgiraudot If you are still around, do you remember why you search for CMake `Threads` support in a file about Eigen3? Was that a copy-paste error?

Issue #4615, opened 2020-04-05
C++20 vs boost/operators.hpp


mglisse:
## Issue Details

Compiling a random CGAL test file (say Epick_d.cpp) with g++-10 -std=c++2a -Wall -Wsystem-headers, I get:

<pre>
In file included from /home/glisse/repos/cgal2/Number_types/include/CGAL/Interval_nt.h:46,
                 from /home/glisse/repos/cgal2/Number_types/include/CGAL/long_double.h:27,
                 from /home/glisse/repos/cgal2/Number_types/include/CGAL/number_type_basic.h:52,
                 from /home/glisse/repos/cgal2/Kernel_23/include/CGAL/basic.h:33,
                 from /home/glisse/repos/cgal2/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_base.h:15,
                 from /home/glisse/repos/cgal2/NewKernel_d/include/CGAL/Epick_d.h:17,
                 from Epick_d.cpp:13:
/home/glisse/repos/boost/utility/include/boost/operators.hpp: In instantiation of ‘bool boost::operators_impl::operator==(const int&, const CGAL::Lazy_exact_nt<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational> >&)’:
Epick_d.cpp:231:3:   required from ‘void test2() [with Ker = CGAL::Epeck_d<CGAL::Dimension_tag<2> >]’
Epick_d.cpp:733:48:   required from here
/home/glisse/repos/boost/utility/include/boost/operators.hpp:154:64: warning: in C++20 this comparison calls the current function recursively with reversed arguments
  154 |      friend bool operator==(const U& y, const T& x) { return x == y; }
      |                                                              ~~^~~~
/home/glisse/repos/boost/utility/include/boost/operators.hpp: In instantiation of ‘bool boost::operators_impl::operator==(const double&, const CGAL::Lazy_exact_nt<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational> >&)’:
Epick_d.cpp:274:3:   required from ‘void test2() [with Ker = CGAL::Epeck_d<CGAL::Dimension_tag<2> >]’
Epick_d.cpp:733:48:   required from here
/home/glisse/repos/boost/utility/include/boost/operators.hpp:154:64: warning: in C++20 this comparison calls the current function recursively with reversed arguments
</pre>

and the program crashes (-O0) or goes into an infinite loop (-O2).

[...]

PR #4640, opened 2020-04-11
C++20 fixes

mglisse:
## Summary of Changes

The important part is the changes to Quotient and Lazy_exact_nt so the code actually works with -std=c++20 and recent compilers.
The change to Gmpq is an experiment (on a very simple type), I first thought I would handle the others that way, but `operator<=>` is harder to work with than I expected. One idiotic detail is that std::strong_ordering is essentially an int that can be -1, 0, or 1, but they do not provide any direct way to convert to/from such an int. Also, using a non-standard return type for <=> seems less acceptable than for the traditional operators.
I think more work will be needed on this topic, but let's first unbreak things.

## Release Management

* Issue(s) solved (if any): partial #4615

### DIFF
--- a/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Polygon_2_curve_iterator.h
+++ b/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Polygon_2_curve_iterator.h
@@ -60,9 +60,9 @@ public:
     Edge_const_iterator m_curr_edge;   // points to the current edge iterator
 
 public:
-    Polygon_2_curve_iterator< X_monotone_curve_2_, Polygon_ >(){}
+    Polygon_2_curve_iterator(){}
 
-    Polygon_2_curve_iterator< X_monotone_curve_2_, Polygon_ >
+    Polygon_2_curve_iterator
       (const Polygon* pgn, Edge_const_iterator ci) : m_pgn(pgn),
                                                      m_curr_edge(ci) {}
 
--- a/Boolean_set_operations_2/include/CGAL/Gps_circle_segment_traits_2.h
+++ b/Boolean_set_operations_2/include/CGAL/Gps_circle_segment_traits_2.h
@@ -27,7 +27,7 @@ class Gps_circle_segment_traits_2 :
   public Gps_traits_2<Arr_circle_segment_traits_2<Kernel_, Filer_> >
 {
 public:
-  Gps_circle_segment_traits_2<Kernel_, Filer_>(bool use_cache = false) :
+  Gps_circle_segment_traits_2(bool use_cache = false) :
     Gps_traits_2<Arr_circle_segment_traits_2<Kernel_, Filer_> >()
   {
     this->m_use_cache = use_cache;
--- a/Inscribed_areas/include/CGAL/Largest_empty_iso_rectangle_2.h
+++ b/Inscribed_areas/include/CGAL/Largest_empty_iso_rectangle_2.h
@@ -233,12 +233,11 @@ public:
   ~Largest_empty_iso_rectangle_2();
 
   //! An operator=
-  Largest_empty_iso_rectangle_2<T>&
-    operator =(const Largest_empty_iso_rectangle_2<T>& ler);
+  Largest_empty_iso_rectangle_2&
+    operator =(const Largest_empty_iso_rectangle_2& ler);
 
   //! A copy constructor
-  Largest_empty_iso_rectangle_2<T>(
-               const Largest_empty_iso_rectangle_2<T>& ler);
+  Largest_empty_iso_rectangle_2(const Largest_empty_iso_rectangle_2& ler);
 
   struct Internal_point {
     Point_2 x_part;// the x coordinate of the point
--- a/Interpolation/include/CGAL/interpolation_functions.h
+++ b/Interpolation/include/CGAL/interpolation_functions.h
@@ -34,7 +34,7 @@ struct Data_access
   typedef typename Map::mapped_type   Data_type;
   typedef typename Map::key_type      Key_type;
 
-  Data_access<Map>(const Map& m): map(m){}
+  Data_access(const Map& m): map(m){}
 
   std::pair< Data_type, bool>
   operator()(const Key_type& p) const
--- a/Number_types/include/CGAL/GMP/Gmpq_type.h
+++ b/Number_types/include/CGAL/GMP/Gmpq_type.h
@@ -238,9 +238,9 @@ public:
 
   bool operator==(const Gmpq &q) const noexcept { return mpq_equal(this->mpq(), q.mpq()) != 0;}
 #if __cpp_impl_three_way_comparison >= 201907L
-  std::strong_ordering operator<=>(const Gmpq&q) const noexcept { return mpq_cmp(this->mpq(), q.mpq()) <=> 0; }
+  std::strong_ordering operator<=>(const Gmpq&q) const { return mpq_cmp(this->mpq(), q.mpq()) <=> 0; }
 #else
-  bool operator< (const Gmpq &q) const noexcept { return mpq_cmp(this->mpq(), q.mpq()) < 0; }
+  bool operator< (const Gmpq &q) const { return mpq_cmp(this->mpq(), q.mpq()) < 0; }
 #endif
 
   double to_double() const noexcept;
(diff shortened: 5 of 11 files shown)


maxGimeno:
There are errors on windows:

https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-151/Classification_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz

lrineau:
> There are errors on windows:
> 
> [cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-151/Classification_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-151/Classification_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)

That was with MSVC 2015 (to prove that it is not related to C++20), but we can see the same error with MSVC 2019 Preview 16.6 with `/std=c++latest` (C++20):
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-151/Triangulation_2/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz

That seems related to `operator==` being transformed to a friend function of `CGAL::Quotient`. Probably `MP_Float` should get new `operator==(int)` and `operator!=(int)` to disambiguate.

I agree with Visual C++ that the call to `operator==(MP_Float, int)` seems ambiguous: the `int` could be converted to `MP_Float`, or the `MP_Float` could be converted to `Quotient`. As usual, the error message from Visual C++ is very confusing. I see an ambiguity between 2 overloads, by VC++ says that 5 overloads have equally conversions, and shows 11 of them.

@mglisse What do you think?

mglisse:
Same conclusion as you: the error message is crap, I don't know if VS is right about the ambiguity, but we need to fix something because gcc picks the overload Quotient,int when called on MP_Float,int...

mglisse:
> gcc picks the overload Quotient,int when called on MP_Float,int

Uh, no, gcc and clang pick the right MP_Float,MP_Float overload, it was only in a small example trying to reproduce (badly, apparently) the problem that it picked the other one...

mglisse:
I have trouble reproducing the issue with a small self-contained testcase on godbolt, VS uses the right overload in all my tests. I guess I'll have to boot windows, try to reproduce there with boost+cgal, then reduce... (not now)

lrineau:
Why not adding the `MP_Float::operator==(int)`, and `operator!=`? That seems right, anyway. I do not know if Visual or clang/gcc are right, but the current of set overloads *looks* ambiguous, even to C++-guru like you and me.

mglisse:
Then you also need to add a version with double and many useless functions. I would feel better understanding where the issue is coming from before doing that (does Interval_nt play a role? It seems prominent in the error message). This kind of friend function is only supposed to be visible if the type of the class is directly looked up (that's part of their selling point). If the friend operator== of Quotient starts interacting with random non-Quotient types, we may have other trouble, and maybe using such friends is not the best idea.

mglisse:
(test/Number_types/MP_Float.cpp is a more natural reproducer)
So, it is a *feature* :bug: in visual studio that, unless you pass `/permissive-`, it compiles a language that only vaguely looks like C++, and in particular hidden friends are visible. In our case, something causes the instantiation of `Quotient<MP_Float>` (we specialize some operations for it), and that causes its `operator==` to appear in namespace `CGAL`, as a full function (not a template, so it is considered even for unrelated types). Then for `MP_Float==int`, it is equally valid to convert the int to `MP_Float` or `Quotient<MP_Float>`.
I don't expect "make `/permissive-` mandatory" to fly with CGAL maintainers. Adding mixed comparisons for MP_Float and int/double (for Visual Studio only!) looks like one way to paper over the issue. I need to think a bit more.

mglisse:
Since what's broken is boost::equality_comparable2 (https://github.com/boostorg/utility/issues/65), another workaround is, in C++20, to avoid deriving from it. It is a bit inconvenient, for istance instead of ordered_field_operators2 we would need to use field_operators2 and less_than_comparable2, and we would need a reliable way to detect when it should be done (there is a feature macro for this, but visual studio tends to be not so good at implementing those).
I still prefer the hidden friends, but I am trying to list our options.

lrineau:
@mglisse Do you think it is ready to be tested in `integration`, now?

mglisse:
Yes, let's see if this workaround is sufficient, or if something more complicated is needed.

maxGimeno:
Errors on windows qbout operator * : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-154/Algebraic_kernel_for_circles/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz

mglisse:
I'll add the other operations for MP_Float...

mglisse:
I am going to revert the spaceship part of the patch, it seems to be causing trouble and is not as urgent.

lrineau:
> I am going to revert the spaceship part of the patch, it seems to be causing trouble and is not as urgent.

It should only show effects on C++20 platforms? I am strugling to introduce C++20 in our testsuite. That is difficult. So far, the only one that "works" is MSVC 2019 Preview with the compilation option `/std=c++latest`.

I have setup two Linux platforms to test C++20 with both gcc and clang, but those two platforms run during 24 hours, and get killed.

afabri:
You suggested in a skype meeting that one should first run for example the Kernel testsuite.

lrineau:
> You suggested in a skype meeting that one should first run for example the Kernel testsuite.

I have already tested the kernel CTests and all was fine:
```
Kernel_23_Examples    =  85.70 sec*proc (18 tests)
Kernel_23_Tests       = 870.34 sec*proc (44 tests)
Kernel_d_Tests        =  77.91 sec*proc (14 tests)
NewKernel_d_Tests     = 102.87 sec*proc (4 tests)
```

The issues seems to be in the Algebraic stuff, and in the Arr_2 tests dealing with algebraic curves.

mglisse:
<pre>
In instantiation of ‘bool boost::operators_impl::operator==(const CORE::BigInt&, const CGAL::Polynomial<CORE::BigInt>&)’:
boost/operators.hpp:154:64: warning: in C++20 this comparison calls the current function recursively with reversed arguments
</pre>
Polynomial seems to have the same problem as I fixed for Quotient here.

mglisse:
I handled Polynomial in a different way, not as nice for good compilers, but less likely to create issues for visual studio :crossed_fingers: .

mglisse:
It looks like the rules for rewriting operator== that are causing us problems are also causing problems for a number of other projects, and some implementers and members of the committee are trying to change those rules. Most of the changes I did are not specific to C++20 and should thus be "safe", but the version I used for Polynomial explicitly tests for C++20, and would likely break if the rules change. This is useful for testing, but even if it passes the tests, please don't merge too quickly.

mglisse:
Some of the issues with C++20 come from boost and don't seem easy to work around without touching the boost headers:
https://github.com/boostorg/bimap/issues/23
https://github.com/boostorg/format/issues/67

mglisse:
A lot of other issues that appear in the C++20 testsuite fall I think in the topic of @Sameeer12 's GSoC: less use of result_type, replace std::not1 with std::not_fn in C++17+, replace boost::* with std::* (bind for instance), some lambdas to replace complicated binds, etc.
(I hope I got the right person)

maxGimeno:
Note that last night's testsuite was only master

lrineau:
@mglisse Thanks to your effort and this PR (as of https://github.com/CGAL/cgal/pull/4640/commits/7a35fbe6fb9d7baeed51671901dfdc3bdcb30264), the two C++20 test platforms of the testsuite that ran between Friday and Saturday manage to finish: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-158.shtml

The results are not good, but that is a huge improvement.

Thanks for the commits about the strange constructors. I wonder why it was compiling previously with all compilers.

It seems that Boost libraries have errors, in particular Boost.Bimap and Boost.Format.

mglisse:
> The results are not good, but that is a huge improvement.

If the branch doesn't cause regressions on non-C++20 platforms (check visual studio in particular), I think it is in a mergeable state.

> It seems that Boost libraries have errors, in particular Boost.Bimap and Boost.Format.

I posted links somewhere above.

lrineau:
> > The results are not good, but that is a huge improvement.
> 
> If the branch doesn't cause regressions on non-C++20 platforms (check visual studio in particular), I think it is in a mergeable state.

I agree. What do you think @maxGimeno?

> > It seems that Boost libraries have errors, in particular Boost.Bimap and Boost.Format.
> 
> I posted links somewhere above (bimap seems fixed in master).

For bimap, the commits are indeed in `master`. Our C++20 test platforms should use Boost `master` branch (or `develop`) for a while.

maxGimeno:
I don't see any regression in recent testsuites, I think we can merge the branch.

lrineau:
I have installed Boost `develop` branch on the two test platforms running with `-std=c++20`.

maxGimeno:

lrineau:
> I have installed Boost `develop` branch on the two test platforms running with `-std=c++20`.

The results are pretty good. The testsuite of [CGAL-5.1-Ic-161](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-161.shtml) has two columns using C++20 (with `CXX20` in their names): one using clang-10.0.0, the other one using gcc-master (11.0.0 20200526). There are a lot of warnings/errors about `operator==` in arrangements, and kernel_d.

mglisse:
(not committing anything, the branch has been merged already)
CGAL/Surface_sweep_2/Curve_pair.h:160
--> operators are missing `const`

PR #4642, opened 2020-04-13
Incorrect definition of ccProd

albert-github:
The ccProd macro for formulas is defined as `\sum_{#1}^{#2}{#3}` but this should be `\prod_{#1}^{#2}{#3}`, probably a cut and paste error from ccSum.
Only place where ccProd is used in the (code) documentation is: Polynomial/doc/Polynomial/Concepts/PolynomialTraits_d--Resultant.h

### DIFF
--- a/Documentation/doc/resources/1.8.13/header.html
+++ b/Documentation/doc/resources/1.8.13/header.html
@@ -32,7 +32,7 @@ $extrastylesheet
 }
 
 \def\ccProd #1#2#3{
-  \sum_{#1}^{#2}{#3}
+  \prod_{#1}^{#2}{#3}
 }\)
 </span>
 <div id="top"><!-- do not remove this div, it is closed by doxygen! -->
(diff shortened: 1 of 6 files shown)


albert-github:
**Note**
 The proposed pull request #4644 also contains this change, but I think this proposed pull request has its value as it clearly signals the ccProd problem and otherwise it goes under  in the major change of #4644.

lrineau:
This PR is flagged for 4.14.x, but the branch is based on `master`.

sloriot:
@albert-github  I guess we can close this one if the fix is included in the PR updating the location of the macro definitions?

albert-github:
@sloriot  In principle yes, but for the clarity, see my argumentation in my previous comment, I think it would be better to incorporate this proposed pull request before incorporating the proposed pull #4644

albert-github:
As #4644 has been merged into master, so this pull request became obsolete unless it will be  used for older versions.

@sloriot, @lrineau up to you to decide.

sloriot:
Let's close it.

PR #4643, opened 2020-04-13
Three doc should not be built by default in public releases

### LINKED ISSUE #4635: Bibtex reference not defined
## Issue Details

In the file: https://doc.cgal.org/latest/Three/group__PkgThreeRef.html we have a BibTex link:

CGAL_bib

 but this link is pointing more or less into thin air as the BibTex reference: cgal:rlfg-pd-20a does not exist in the available bibtex files:
```
Documentation/doc/biblio/cgal_manual.bib
Documentation/doc/biblio/geom.bib
```

sloriot:
Fixes  #4635


sloriot:
@maxGimeno you might want to update the testsuite to continue testing the Three package doc

albert-github:
Irrespective whether or not the `Three` documentation is build, it is still possible to build the documentation, so I think the reference should be added anyway (for a public build the problem is "solved" but not for internal builds).

sloriot:
It will be automatically fixed after the package is officially part of the doc. Then, it will appear in the package overview and the bibtex entry will be automatically generated.

sloriot:
@maxGimeno Looks like travis also need the option to be switch ON. Would you mind adding it while adding the option to the testsuite?

sloriot:
@maxGimeno  Your patch does not seem sufficient

maxGimeno:

PR #4644, opened 2020-04-13
Introducing central MATHJAX_CODEFILE

albert-github:
Currently the needed extra definitions for MathJax usage (i.e. `\newcommand` and `\def` commands) are distributed over the header file and in a number of source code files.
Doxygen has the possibility for a centralize this by means of the MATHJAX_CODEFILE configuration setting, this has been implemented.


albert-github:
Note: this pull requests also incorporates the change of #4642 (the correction of ccProd).

maxGimeno:

sloriot:
Looking at [here](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1-Ic-126/master/Skin_surface_3/index.html#Chapter_3D_Skin_Surface_Meshing) vs. [here](https://doc.cgal.org/latest/Skin_surface_3/index.html) it seems to me that there are some extra parenthesis. I guess the syntax between the header and codefile are different. Maybe it is enought to remove the `()` from the code file. Shall I try it or do you handle it?

albert-github:
Which formula / text to search for (I don't see it at first glance.

sloriot:
I only checked `ssWpoint` and `ssWeight` but I suspect it is a general issue.

albert-github:
I now saw some places, well spotted.
In the past ssWeight and ssWpoint were `\def` and now they are analogous to the `\newcommand` (all others were already `\newcommand`). I'll dive into it ...

albert-github:
In my original tests I didn't see the round brackets, but when rerunning the build I saw the round brackets as well. I do remember doing something with the `\(` (I think it was adding them, but I don't remember and apparently I didn't pay enough attention to the output). I hope the fix 93e0eb11f9a8f9b627133b32f301688190528c08 fixes the problem.

sloriot:
With some help from python, I identified files that were having a difference in the generated html and inspected those files. The rendering is the same (even better in the new version), the only difference being in `QP_solver` where an extra (useless) empty line was removed and some formula disappear from the short description (why were they there???). It's all good and I'm merging it now.

sloriot:
Thanks again for the patch.

PR #4649, opened 2020-04-14
Try to filter out warnings not in CGAL

sloriot:


lrineau:
With that sort of filtering, that will hide all warnings that are not from CGAL. In particular, that will hide deprecation warnings when CGAL code use deprecated things from the STL or from 3rd party libraries.

There has been in the past important warnings from 3rd party libraries, or from the STL. Do we really want to hide them all?

sloriot:
We can also update `create_testresult_page` to handle a third category `t` that would mean warnings in third parties. At least in the current state it is helpful to prepare the next release IMO.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-137.shtml.
Nice feature btw

maxGimeno:
It seems this PR also hides CMake warnings

lrineau:
Now that we have discovered that this PR hid CMake warnings, should not we think about other tools? For example, what about linker warnings? Or from other tools that are called by CMake, without `cmake` in the warning message?

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-169.shtml

sloriot:
> We can also update `create_testresult_page` to handle a third category `t` that would mean warnings in third parties. At least in the current state it is helpful to prepare the next release IMO.

This is a TODO that must be implemented by someone before thinking about integrating this PR (even if it is not a perfect solution when it comes to cmake but at least no information is lost).

lrineau:
I have also seen warnings like that one:
```
C:\3rdPartyLibs\boost\boost_1_65_1\boost/property_map/property_map.hpp(310): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data
C:\CGAL\test\CGAL-5.1-I-201\include\CGAL/cluster_point_set.h(181): note: see reference to function template instantiation 'void boost::put<CRTP_derived_class,int&,CGAL::Point_set_3<Point_3,CGAL::Vector_3<Kernel_>>::Index,std::size_t>(const boost::put_get_helper<int &,CRTP_derived_class> &,K,const V &)' being compiled
        with
        [
            CRTP_derived_class=CGAL::Point_set_3<Point_3,CGAL::Vector_3<CGAL::Epick>>::Property_map<int>,
            Kernel_=CGAL::Epick,
            K=CGAL::Point_set_3<Point_3,CGAL::Vector_3<CGAL::Epick>>::Index,
            V=std::size_t
        ]
```
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-I-201/Point_set_processing_3_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz

The warning is in Boost, but the origin of the warning is in CGAL.

sloriot:
Screenshot from 2021-04-15 13-59-35

maxGimeno:
I'd go for another color, like 
rgb(0, 210, 255)
image

or rgb(0, 110, 0)
image

lrineau:
I would swap `w` and `t` is the summary table, to follow the priority by order of gravity `y < t < w < n`.

maxGimeno:
Just a last thing, if we take the HUE wheel, it seems logical to have a "t" warning between the green of OK and the yellow of Warning, where it is now between the yellow of warning and the red of error, which makes it look more important than a warning, where it is the opposite. So I would go for a yellow-green : 
the warning color:
rgb(100%,100%,50%)
image
The "t" warning:
rgb(90%,100%,0%)
image

the OK : rgb(50%,100%,50%)
image

But I can't get it to not look too yellow or too green, so I'm not sure about the logic. @CGAL/geometryfactory what do you think ?

lrineau:
In the Hue wheel, our yellow is hue=60, and our green is hue=120, and both have sat=127 and val=255.

So, pick as intermediate color hue=90, sat=127, and val=255: in rgb tat give (75%, 100%, 50%). Do you want me to try and modify the color accordingly in the CSS?

lrineau:
That gives:
Screenshot_20210621_122947

I think the contrast of that color, compare to the green, is not enough.

lrineau:
I have decreased the saturation of that color, to increase the contrast:
Screenshot_20210621_125512

maxGimeno:
I think That would be better than our current orange, that makes the whole testsuite seems far, far worse than it really is.

danston:
Now, to my taste, the dark green looks more like the fully correct output while the light green like a t-warning.

lrineau:
> Now, to my taste, the dark green looks more like the fully correct output while the light green like a t-warning.

So you think the color of `y` should be darker, instead of the one of `t`? We can do that.

danston:
> > Now, to my taste, the dark green looks more like the fully correct output while the light green like a t-warning.
> 
> So you think the color of `y` should be darker, instead of the one of `t`? We can do that.

Yes, we could try to see if it looks better.

maxGimeno:
I agree with that decision.

lrineau:
Screenshot_20210621_131026

Now that the `y` green is darker, the whole page seems darker. We keep that?

maxGimeno:
It looks strange, like the whole think was de-activated or something, doesn't it ?

danston:
To me, it looks quite good. Maybe we could make the dark green a bit lighter but how much lighter is hard to say, to not fall back to the light green.

lrineau:
New saturation:
Screenshot_20210621_132451

danston:
For me, the previous one and this one are both good. In the first one, I think the distinction between two greens and yellow is slightly better visible but the whole page is a bit dark. On the other hand, in the second one, the page as a whole looks very nice, but the distinction is slightly less visible. We should see what other people think.

PR #4650, opened 2020-04-14
Fix is_outward_oriented doc

### LINKED ISSUE #3847: Is_outward_oriented does not work with non triangular meshes
https://github.com/CGAL/cgal/pull/2057 fixed some issues with `PMP::is_outward_oriented`, but it assumes that the mesh is triangulated (specifically, while looping around the maximal vertex in [PMP::internal::is_outward_oriented](https://github.com/CGAL/cgal/blob/releases/CGAL-4.13-branch/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h#L69)). That the mesh is triangulated is not a documented precondition of the function, since `compute_normal` (as broken as it was) was supposed to work with non-triangle faces.

Either the doc should be clarified to restrict inputs to triangle meshes, or the code needs to be fixed (could just call triangulate_faces() on a local copy of the faces incident to the maximal vertex ?).

It should also specify that it doesn't work on self-intersecting meshes.

sloriot:
Fixes  #3847

I'm not doing the backport to release branches as the doc changed and it is not worth handling the conflict.

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orientation.h
@@ -141,50 +141,52 @@ namespace internal{
 
 /**
  * \ingroup PMP_orientation_grp
- * tests whether a closed polygon mesh has a positive orientation.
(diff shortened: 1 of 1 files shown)


maxGimeno:

Issue #4367, opened 2019-11-18
Warning C4244 with Point_set_3 and float kernel on MS Visual Studio


elombardi2:
## Issue Details

Compiling a code using Point_set_3 with a float kernel triggers several "warning C4244:  conversion from 'const double' to 'const float' ..." warnings on MS Visual Studio.

The warnings come from [this line](https://github.com/CGAL/cgal/blob/55a05c5a23c1a5721f8cba3a7b5485b9a1f2f663/Point_set_3/include/CGAL/Point_set_3.h#L322) and [this line](https://github.com/CGAL/cgal/blob/55a05c5a23c1a5721f8cba3a7b5485b9a1f2f663/Point_set_3/include/CGAL/Point_set_3.h#L852). Changing the `0.` into `0.f` turns off the warning, but is not a global fix (works only with a float or a double kernel).

## Source Code
```
#include <CGAL/Cartesian.h>
#include <CGAL/Point_set_3.h>

using Kernel = CGAL::Cartesian< float >;
using Point = Kernel::Point_3;

int main(void)
{
	CGAL::Point_set_3< Point > ps;
	return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 10 64bits
* Compiler: MS Visual Studio 2015 & 2017
* Release or debug mode: both
* Specific flags used (if any):
* CGAL version: 4.14.1
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

gdamiand:
Maybe we need to replace all occurrences of Point (0., 0., 0.) by CGAL::Origin and Vector(0., 0., 0.) by CGAL::Null_vector ?

sgiraudot:
One solution is to do something like `Point(FT(0),FT(0),FT(0))` but I find it quite heavy… I'll have a look into it, maybe `CGAL::Origin` is a good idea indeed.

sloriot:
@sgiraudot would you have time to fix that?

sgiraudot:
Sure, I'll look into it.

PR #4656, opened 2020-04-16
Point Set: Fix warnings with float kernel

sgiraudot:
## Summary of Changes

Using a float kernel is raising warnings on some platforms because points and vectors are initialized from doubles. This fix uses `CGAL::ORIGIN` and `CGAL::NULL_VECTOR` instead.

(I also changed some outdated `boost::tie` lines.)

## Release Management

* Affected package(s): Point Set 3
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4367

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -330,8 +330,8 @@ public:
   void clear()
   {
     m_base.clear();
-    boost::tie (m_indices, boost::tuples::ignore) = this->add_property_map<Index>("index", typename Index::size_type(-1));
-    boost::tie (m_points, boost::tuples::ignore) = this->add_property_map<Point>("point", Point (0., 0., 0.));
+    m_indices = this->add_property_map<Index>("index", typename Index::size_type(-1)).first;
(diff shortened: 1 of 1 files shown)


sgiraudot:
(@janetournois will test this fix on Windows later today.)

Point_set_3/include/CGAL/Point_set_3.h
```diff
@@ -330,8 +330,8 @@ class Point_set_3
   void clear()
   {
     m_base.clear();
-    boost::tie (m_indices, boost::tuples::ignore) = this->add_property_map<Index>("index", typename Index::size_type(-1));
-    boost::tie (m_points, boost::tuples::ignore) = this->add_property_map<Point>("point", Point (0., 0., 0.));
```
lrineau, line 334: Another way to fix that sort of conversion warning: use `Point(0.f, 0.f, 0.f)`. `0.f` and `0.` are the same value, and a float converts to a double without any loss of precision.

But I like your option (`ORIGIN`) best.

janetournois:
I tried with vs 2015 and vs 2019, and could not reproduce the warning, nor before nor after this fix.
Sorry...

maxGimeno:

PR #4657, opened 2020-04-16
Try to make installation relocatable

### LINKED ISSUE #4433: CMake installed package is no longer relocatable
## Issue Details

When the package generated by `cmake install` is moved to another place, `find_package(CGAL REQUIRED)` would fail with `Found package configuration file but it set CGAL_FOUND to FALSE so package "CGAL" is considered to be NOT FOUND.`

It turns out that `CGAL_ROOT` is set to a hard-coded path at configure step in `CGALConfig-installation-dirs.cmake`.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 10 64 bit
* Compiler: Visual Studio Community 2019
* CGAL version: 5.0
* Boost version: 1.70.0
* Other libraries versions if used (Eigen, TBB, etc.): Eigen 3.3.7

sloriot:
Fixes #4433

It is working on Linux, no idea on windows nor whether this is bad cmake practice.


lrineau:
This patch would not solve the issue that the PR https://github.com/CGAL/cgal/pull/4285 solved.

The issue is that:
 - on Debian, the path to `CGALConfig.cmake` is `/usr/lib/x86_64-linux-gnu/cmake/CGAL/`,
 - on most other Linux distributions it is `/usr/lib/cmake/CGAL/` or `/usr/lib64/cmake/CGAL/`.

So the relationship between `CGALConfig.cmake` and the prefix is:
  - `../../../..` on Debian (four components),
  - `../../..` on other distributions (three components).

That is why I made that patch https://github.com/CGAL/cgal/pull/4285 at the last moment before CGAL-4.14.2, CGAL-5.0. It made the package non-relocatable on purpose, by hardcoding the prefix in a file configured by CMake. A real patch could be computing at configure-time the number of components in `CGAL_INSTALL_CMAKE_DIR`.

sloriot:
Reading again #4433, the only thing that is mentioned is "moving the package created by make install", what you are mentioning now is another story to me. I even don't understand why you are saying "This patch would not solve the issue that the PR #4285 solved.". Do you mean that this patch breaks the previous patch?

sloriot:
OK things start to get clearer. AFAIU there are 2 cases: package managers and "regular" make install. I propose to have 2 `CGAL_ROOT` variables,  one for cmake files, one for c++ headers. In the case where they are identical we use the behavior of this branch and if not the behavior currently in master. This implies that package managers will have to update their pipeline.

lrineau:
Package managers are regular users of `make install`. The difference is whether the cache variable `CGAL_INSTALL_CMAKE_DIR` is customized from its default, and if the relative path has three components or not.

lrineau:
https://github.com/CGAL/cgal/pull/4697 will fix the issue #4433.

PR #4397, opened 2019-12-06
Triangulation_3: Accelerate point insertion

afabri:
## Summary of Changes

Filling the cavity of a conflict zone used `next_around_edge()` on the outside of the boundary to identify neighboring new cells.  This got replaced with a small custom hash map, inspired by geogram.  
Additionally, replaced some `std::vector` by `boost::small_vector`.

## Release Management

* Affected package(s): Triangulation_3
* License and copyright ownership: maintenance so IP remains at inria sophia

PR #4661, opened 2020-04-17
Documentation : Update Boost Minimal Version

maxGimeno:
## Summary of Changes

#4397 introduced the small_vector of boost, but it only appeared in the version 1.58 of boost. 
## Release Management

### DIFF
--- a/Documentation/doc/Documentation/Third_party.txt
+++ b/Documentation/doc/Documentation/Third_party.txt
@@ -56,7 +56,7 @@ or <A HREF="https://msdn.microsoft.com/en-us/library/1fe2x6kt(v=vs.140).aspx">`h
 The \stl comes with the compiler, and as such no installation is required.
 
 \subsection thirdpartyBoost Boost
-<b>Version 1.57 or later</b>
+<b>Version 1.58 or later</b>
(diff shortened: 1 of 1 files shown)


lrineau:
We have an issue in our testsuite because https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-127.shtml says that our MSVC2015 platform on gauguin is using Boost-1.57, but it actually uses Boost-1.65 (see in a log, [for example this one](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-127/AABB_tree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)).

lrineau:
And the minimal version of Boost we actually test is 1.62.0, but that is one a test platform we do not really trust: our Raspberry Pi.

We should bump the minimal version of Boost to 1.65.1 (tested on our Ubuntu 18.04 Bionic platforms) (for CGAL-5.1). I think for CGAL-5.0, we cannot really change the minimal version after 1.58 (that is the version in Ubuntu 16.04 Xenial).

PR #4664, opened 2020-04-18
Make orientations_ a local variable, for thread-safety

### LINKED ISSUE #4537: dD Triangulation locate thread safety issue
I noticed a thread safety issue when invoking several Triangulation::locate in parallel. There is the mutable member 'orientations_' which is modified.
Wen moving the member inside the member function in Triangulation.h, it works.

There are also mutable members, but I don't know if they can affect thread safety as well.

Are they intended to be CGAL_STATIC_THREAD_LOCAL_VARIABLEed?

mglisse:
In small dimension, small_vector won't allocate, and in large dimension, orientations are more costly than a single allocation.
I was tempted to go for the simpler and cheaper
```c++
CGAL_assertion(cur_dim <= 31);
signed char orientations_[32];
```
(or even 15 and 16, I don't think Triangulation is very usable in such large dimensions)

I used `signed char` to store Oriented_side because it seemed wasteful to use 32 bits (on some ABIs) just for { -1, 0, 1 }. I considered changing the definition as `enum Sign : signed char { ... };` in enum.h, but it doesn't look like we store those enums in many places, so leaving it to the implementation to decide on the size seems good.

## Release Management

* Issue(s) solved (if any): fix #4537

### DIFF
--- a/Triangulation/include/CGAL/Triangulation.h
+++ b/Triangulation/include/CGAL/Triangulation.h
@@ -27,6 +27,7 @@
 #include <CGAL/Default.h>
 #include <CGAL/Random.h>
 
+#include <boost/container/small_vector.hpp>
 #include <boost/iterator/filter_iterator.hpp>
 #include <CGAL/boost/iterator/transform_iterator.hpp>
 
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #4669, opened 2020-04-20
Remove "unused-but-set-parameter" warnings

gdamiand:
## Summary of Changes

When compiling draw_linear_cell_complex.cpp example with -Wall -Wextra, I have 23 warnings:
   parameter ‘n’ set but not used [-Wunused-but-set-parameter]
 
 This probably comes from the fact that the methods use std::nullptr_t and thus the assertion (n == nullptr) is removed since it is always true.
 
 There are probably other places where this problem occurs.
 
## Release Management

* Affected package(s): STL_Extension and TDS_2

### DIFF
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -882,12 +882,12 @@ namespace internal {
     }
 
     // Construction from nullptr
-    CC_iterator (std::nullptr_t CGAL_assertion_code(n))
+    CC_iterator (std::nullptr_t /*CGAL_assertion_code(n)*/)
 #ifdef CGAL_COMPACT_CONTAINER_DEBUG_TIME_STAMP
       : ts(0)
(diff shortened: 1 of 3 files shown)


lrineau:
I think you are right: `std::nullptr_t` is a special type that can have only one value `nullptr`. So maybe any variable of type `std::nullptr_t` is actually considered as unused.

You are not the only one impacted by that strange warning,  see that on StackOverflow: https://stackoverflow.com/q/40790212/1728537.

gdamiand:
> I think you are right: `std::nullptr_t` is a special type that can have only one value `nullptr`. So maybe any variable of type `std::nullptr_t` is actually considered as unused.
> 
> You are not the only one impacted by that strange warning, see that on StackOverflow: https://stackoverflow.com/q/40790212/1728537.

It's strange that this warning never occurs on our test suite?

maxGimeno:

PR #4244, opened 2019-09-24
AABB_tree: make accelerate distance queries default

maxGimeno:
## Summary of Changes
Make the `accelerate_distance_queries()` lazy and add a function to disable it.

## Release Management

* Affected package(s):AABB_tree
* Issue(s) solved (if any): fix #4008
* Feature/Small Feature (if any):
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership:

PR #4675, opened 2020-04-22
Make the kd-tree build really lazy and fix thread safety of builds

sloriot:
Follow up of #4244 + document explicitely what is done.

The PR is on purpose for 5.1 only.

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -603,7 +603,7 @@ public:
     std::atomic<bool> m_atomic_search_tree_constructed;
 #else
     bool m_need_build = false;
-    bool m_search_tree_constructed = false;
+    mutable bool m_search_tree_constructed = false;
 #endif
 
   private:
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -752,7 +752,7 @@ public:
     if(m_p_search_tree != nullptr)
     {
 #ifdef CGAL_HAS_THREADS
-      m_atomic_search_tree_constructed.store(false, std::memory_order_release); // in case build_kd_tree() is triggered by a call to best_hint()
+      m_atomic_search_tree_constructed.store(true, std::memory_order_release); // in case build_kd_tree() is triggered by a call to best_hint()
 #else
       m_search_tree_constructed = true;
 #endif
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -172,7 +172,7 @@ namespace CGAL {
     }
 
     /// Clears the tree and the search tree if it was constructed,
-    /// and switches on the usage of the search tree to find the hints for the distance queries
+    /// and switches on the usage of the search tree to find the hint for the distance queries
     void clear()
     {
       // clear AABB tree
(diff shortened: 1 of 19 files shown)


AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -484,15 +475,23 @@ namespace CGAL {
     }
 
     // clears internal KD tree
-    void clear_search_tree() const
+    void clear_search_tree()
     {
+#ifdef CGAL_HAS_THREADS
+      if ( m_atomic_search_tree_constructed.load() )
```
lrineau, line 481: As you choose to use the full API of `std::atomic`, you can as well pass the right `std::memory_order` argument:
```suggestion
      if ( m_atomic_search_tree_constructed.load(std::memory_order_acquire) )
```

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -484,15 +475,23 @@ namespace CGAL {
...
+#else
       if ( m_search_tree_constructed )
+#endif
       {
         CGAL_assertion( m_p_search_tree!=nullptr );
         delete m_p_search_tree;
         m_p_search_tree = nullptr;
+#ifdef CGAL_HAS_THREADS
+        m_atomic_search_tree_constructed.store(false);
```
lrineau, line 490:
```suggestion
        m_atomic_search_tree_constructed.store(false, std::memory_order_release);
```

AABB_tree/include/CGAL/AABB_tree.h
sloriot: I think it should be relaxed since the variable is guaranteed not to be modified once we are in this function (which I thought was the default but is seq actually...).

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -213,11 +232,19 @@ class Side_of_triangle_mesh
...
+#ifdef CGAL_HAS_THREADS
+        CGAL_SCOPED_LOCK(tree_mutex);
+        atomic_tree_ptr.store(tree_ptr, std::memory_order_release);
+#endif
         }
       }
-
+#ifdef CGAL_HAS_THREADS
+      return internal::Point_inside_vertical_ray_cast<GeomTraits, AABB_tree>()(
+            point, *atomic_tree_ptr.load(), ray_functor, vector_functor);
```
lrineau, line 243: This `load()` could be relaxed. It might be important for performances, because that `load()` is performed at each call to `operator()`.

Actually, even better for performance, I think you should use:
```suggestion
            point, *tree_ptr, ray_functor, vector_functor);
```

AABB_tree/include/CGAL/AABB_tree.h
lrineau: That is right (relaxed). I had not seen in which function we were. Actually, in this function, the synchronization between threads and the performance do not really matter.

maxGimeno:
ERROR:      execution___of__point_inside_polyhedron_boundary_test in 1200.15 s : Timeout 
   ERROR:      execution___of__point_inside_surface_mesh_test in 1200.08 s : Timeout 
   ERROR:      execution___of__test_pmp_clip in 1200.08 s : Timeout .
and most docker platforms are timing out.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -581,10 +603,15 @@ namespace CGAL {
...
-    mutable bool m_default_search_tree_constructed; // indicates whether the internal kd-tree should be built
-    bool m_need_build;
+    const Search_tree* m_p_search_tree = nullptr;
+    bool m_use_default_search_tree = true; // indicates whether the internal kd-tree should be built
+#ifdef CGAL_HAS_THREADS
+    std::atomic<bool> m_atomic_need_build;
+    std::atomic<bool> m_atomic_search_tree_constructed;
+#else
+    bool m_need_build = false;
+    mutable bool m_search_tree_constructed = false;
```
lrineau, line 613: This variable probably no longer needs the `mutable` keyword.

sloriot: It does, otherwise is does not compile when threads is OFF (it is used in a const function while with threads it is a local variable).

maxGimeno:
I think this PR is responsible for https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-133/Surface_mesh_simplification/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

maxGimeno:
warning in the doc : https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.1-Ic-137/logs_master/AABB_tree.log

maxGimeno:

Issue #4662, opened 2020-04-17
Different behavior of the edge_aware_upsample_point_set?


ibrahimLearning:
@afabri @sgiraudot 
I am trying to do up-sampling using edge_aware_upsample_point_set function. I used it from point_set_processing_3 example. It works for **before_upsample.xyz** data (which is available in the data folder); But when I used my own data it gives the following error:

> edge_aware_upsample_point_set_example.exe (process 17092) exited with code -1073741819.

```
//Algorithm parameters
  const double sharpness_angle = 30;   // control sharpness of the result.
  const double edge_sensitivity = 0.2;    // higher values will sample more points near the edges          
  const double neighbor_radius = 18;  // initial size of neighborhood.
```
Is there any hidden things with the dataset?
My data is look like this:
```
256075.14697445265 3935300.4999999995 -1.9903260605285873 -0.15059442500026382 -0.009264547714967571 -0.98855221779857272
256127.18092914129 3935279.9999999991 1.1806971722817656 0.36255488715508305 -0.28210066175799464 -0.88824161714911032
256124.96664647732 3935273.5000000005 0.018751732199395927 -0.14177362442736433 -0.13905823184364116 0.98008318400697436
256092.491746024 3935308.4999999995 -1.4489924924748039 -0.085803198655671153 0.024081843084161825 0.99602102183343766
256125.86596481747 3935280.5000000009 13.515503257111945 -0.13654115576027245 -0.24797889506659618 0.95909487559114781
256073.15088360262 3935255 -0.93490020406201102 0.59004993642832648 -0.095249131262525044 0.80172855475819382
256141.9517848785 3935322.2500000005 8.2318409757309219 -0.24795966835219574 -0.510449916403933 0.82338137318856486
256096.20889673862 3935319.7500000005 -0.21291360675685664 -0.099540727133083942 -0.024195625188008569 -0.9947392700419434
256107.09924066794 3935294.0000000005 -2.4377930441306859 -0.26315074779627329 -0.097325421959314151 0.95983303036241863
256130.16003200199 3935331.0002575037 14.609926967219579 -0.3588182051513355 -0.18823254749436522 0.9142308262773442
256116.12694613857 3935270.75 -2.0657511148141805 0.24053723579657579 0.070452458792483491 -0.9680796915778348
256089.84995883351 3935295.0000017555 -1.6188145313052476 0.0064104152035124699 0.04608216574353774 0.99891708393505008
[...]

PR #4677, opened 2020-04-23
Fix initialization of Bbox_3 in edge_aware_upsample_point_set()

sgiraudot:
## Summary of Changes

`CGAL::edge_aware_upsample_point_set()` produced a segfault in some cases. This came from 2 problems combined:

1. The size of a `std::vector` was computed using `int` variables: in the bug case, an overly large vector was created, but the `int` silently overflowed with an incorrect size instead of triggering a `std::bad_alloc`
2. At some point, the `CGAL::Bbox_3` of the point cloud was initialized with all values to 0, which is obviously wrong as it means that the origin is always included, which is not necessarily the case.

What happens is that for point clouds far away from the origin, the size of a `std::vector` computed from the bbox size becomes overly large. Because of the `int` overflow, this is not detected, the vector is constructed with the wrong size, and at some point an access to an index greater than this size is done, which leads to a seg fault.

Obviously, fixing the bbox initialization (with default constructor) fixes the bug, but I still changed the construction of the vector using `std::size_t` sizes, just to make sure that if something similary ever happens, we'll at least get a `std::bad_alloc` exception.

## Release Management

* Affected package(s): PSP
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4662

### DIFF
--- a/Point_set_processing_3/include/CGAL/Point_set_processing_3/internal/Rich_grid.h
+++ b/Point_set_processing_3/include/CGAL/Point_set_processing_3/internal/Rich_grid.h
@@ -195,15 +195,19 @@ void Rich_grid<Kernel>::init(std::vector<Rich_point<Kernel> > &vert,
 
   radius = _radius;
 
-  x_side = (unsigned int)ceil((bbox.xmax() - bbox.xmin()) / radius);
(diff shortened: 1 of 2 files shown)


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-137.shtml

PR #4678, opened 2020-04-23
Script: Update the script for the demo bundle

### LINKED ISSUE #2650: (Automatic?) build of demo/Polyhedron into a AppImage

maxGimeno:
## Summary of Changes

Updates for use with new docker image.
## Release Management

* Affected package(s):Scripts
* Issue(s) solved (if any): fix #2650


Polyhedron/demo/Polyhedron/Plugins/Mesh_3/CMakeLists.txt
```diff
@@ -34,13 +34,13 @@ else()
...
 
-find_package(Boost QUIET OPTIONAL_COMPONENTS filesystem)
+find_package(Boost QUIET OPTIONAL_COMPONENTS filesystem system)
 if(Boost_FILESYSTEM_FOUND)
   qt5_wrap_ui( imgUI_FILES Image_res_dialog.ui raw_image.ui)
   polyhedron_demo_plugin(io_image_plugin Io_image_plugin Volume_plane_intersection.cpp Raw_image_dialog.cpp ${imgUI_FILES} ${VOLUME_MOC_OUTFILES} KEYWORDS IO Mesh_3)
   target_link_libraries(io_image_plugin PUBLIC scene_image_item ${VTK_LIBRARIES} CGAL::CGAL_ImageIO)
   if(TARGET Boost::filesystem)
-    target_link_libraries(io_image_plugin PUBLIC Boost::filesystem)
+    target_link_libraries(io_image_plugin PUBLIC Boost::filesystem Boost::system)
```
lrineau, line 43: I do not know why `Boost::system` is now required. Why is that in this PR?

If it is justified, then the `if(Boost_FILESYSTEM_FOUND)` should be updated to add the test for `system`, as well as the message when one of them is not found.

lrineau:
I have tried and it seems that works fine, now.

Note that most of the work is done in the comtainer image. See https://github.com/CGAL/bundle-CGAL-3D-demo/pull/3.

lrineau:
The result is a unique file, named `CGAL_3D_Demo-x86_64.AppImage`. And it should work on most Linux distributions.

1. Is that the right name?
2. If we want to have it built automatically, should it be done by a CI, each time `master` is pushed?
3. Where do we want to publish the file?

maxGimeno:
1. I chose "CGAL_3D_Demo" as it is how it's called in the github flags, but we should probably call it the same name as the windows version : `polyhedron_3`. The  suffix seems to be standard.
2. That seems like a good idea to me.
3. I think it should be on github, like `https://github.com/CGAL/cgal/releases/download/continuous/polyhedron_3-x86_64.Appimage`, if possible

maxGimeno:

PR #4687, opened 2020-04-27
Introducing CMake option CGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY

albert-github:
When (re)generating the documentation from the sources and not using the, internally patched CGAL, doxygen version one gets messages about:
```
warning: ignoring unsupported tag 'EXTRACT_ALL_NO_DETAILED_IF_EMPTY'
```

By introducing the option `CGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY` this doxygen setting can be steered
- option not supplied to cmake, will use as if specified  `-DCGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY=ON`
- option supplied to cmake as `-DCGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY=ON` will write the line `EXTRACT_ALL_NO_DETAILED_IF_EMPTY = YES` in the doxygen configuration
- option supplied to cmake as `-DCGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY=OFF` will write the line `#EXTRACT_ALL_NO_DETAILED_IF_EMPTY = NO` in the doxygen configuration, so just using the default setting of `NO` line could be omitted but it is better to have a placeholder

### DIFF
--- a/Documentation/doc/CMakeLists.txt
+++ b/Documentation/doc/CMakeLists.txt
@@ -201,6 +201,15 @@ if(CGAL_DOC_CREATE_LOGS)
   file(MAKE_DIRECTORY "${CGAL_DOC_LOG_DIR}")
 endif()
 
+option(CGAL_EXTRACT_ALL_NO_DETAILED_IF_EMPTY "Use CGAL special doxygen setting EXTRACT_ALL_NO_DETAILED_IF_EMPTY." ON)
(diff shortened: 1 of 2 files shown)


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-137.shtml

PR #4690, opened 2020-04-28
Mesh_2: Make it deterministic

### LINKED ISSUE #4681: Mesh_2: Meshing is not deterministic
## Issue Details

We have non regression tests randomly diffing between executions. These tests rely on CGAL mesh2 refinement which is not deterministic. This is a problem for us, since we WANT reproducibility!

## Source Code

See https://gist.github.com/manuel-garnier/8214384716971459a8037814b2c5626d
CMakeLists.txt was generated from script provided in CGAL distrib. You could use setup.sh script but it's mainly 2 cmake calls :)
Provided sample test simply tries to do 20 meshing and fails when one is different from first one (in term of number of vertices). It may not seem enough if you integrate this in CGAL test suite though.
In source, you can uncomment output file write to visualize triangulation if needed.

## Environment

* Operating system: Windows/Linux
* Compiler: VS2015 / gcc48 / gcc73
* Release or debug mode: Both
* Specific flags used (if any): 5.0.2 built from scratch without modif
* CGAL version: 4.1.4 and 5.0.2
* Boost version: 1.63.0
* Other libraries versions if used (Eigen, TBB, etc.): gmp-5.0.1 and mpfr-3.0.0

afabri:
## Summary of Changes

Use a time stamper for the mesh face base class.  This makes the construction deterministic.
The `Edge_iterator` is also deterministic as it compares face handles which means comparing time stamps.

We decided not to document the time stamp mechanism, and not to document that the mesh generator is deterministic.

I tried it with two more test cases: the contour of norway, as well as many segments all intersecting in a single point in order to have more complicated clusters.

### Todo: 
- [ ]  We have to look at `create_face()` which makes a copy of a face, which messes up the time stamper.
- [ ] Is there more to make deterministic?  
- [ ] It probably needs more complicated test cases.

## Release Management

* Affected package(s): Mesh_2
* Issue(s) solved (if any): fix #4681

* License and copyright ownership: unchanged INRIA

### DIFF
--- a/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
+++ b/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
@@ -17,6 +17,7 @@
 
 
 #include <CGAL/Constrained_Delaunay_triangulation_face_base_2.h>
+#include <CGAL/Has_timestamp.h>
 
 namespace CGAL {
 
@@ -67,6 +68,18 @@ public:
   /** compatibility with CGAL-3.2 */
   inline
   void set_marked(const bool b) { in_domain=b; }
+
+  
+    typedef Tag_true Has_timestamp;
+
+  std::size_t time_stamp() const {
+    return time_stamp_;
+  }
+  void set_time_stamp(const std::size_t& ts) {
+    time_stamp_ = ts;
+  }
+  std::size_t time_stamp_;
+  
 };
 
 } // namespace CGAL
--- a/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
+++ b/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
@@ -69,8 +69,8 @@ public:
   inline
   void set_marked(const bool b) { in_domain=b; }
 
-  
-    typedef Tag_true Has_timestamp;
+
+  typedef Tag_true Has_timestamp;
 
   std::size_t time_stamp() const {
     return time_stamp_;
@@ -79,7 +79,7 @@ public:
     time_stamp_ = ts;
   }
   std::size_t time_stamp_;
-  
+
 };
 
 } // namespace CGAL
--- a/Mesh_2/test/Mesh_2/reproductibility.cpp
+++ b/Mesh_2/test/Mesh_2/reproductibility.cpp
@@ -1,0 +1,106 @@
+//#define CGAL_MESH_2_DEBUG_BAD_FACES
+//#define CGAL_MESH_2_DEBUG_CLUSTERS
+
+#include <CGAL/Constrained_Delaunay_triangulation_2.h>
+#include <CGAL/Delaunay_mesh_size_criteria_2.h>
+#include <CGAL/Delaunay_mesh_face_base_2.h>
+#include <CGAL/Delaunay_mesher_2.h>
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+
+#include <fstream>
+#include <iostream>
+
+using K = CGAL::Exact_predicates_inexact_constructions_kernel;
+using Vb = CGAL::Triangulation_vertex_base_2<K>;
+using Fb = CGAL::Delaunay_mesh_face_base_2<K>;
+using Tds = CGAL::Triangulation_data_structure_2<Vb, Fb>;
+using CDT = CGAL::Constrained_Delaunay_triangulation_2<K, Tds, CGAL::Exact_predicates_tag>;
+using Criteria = CGAL::Delaunay_mesh_size_criteria_2<CDT>;
+using Mesher =  CGAL::Delaunay_mesher_2<CDT, Criteria>;
+
+using Vertex_handle = CDT::Vertex_handle;
+using Point = CDT::Point;
+
+int main(int argc, char* argv[])
+{
+  std::string path = argv[0];
+  path = path.substr(0, path.rfind('/') + 1);
+
+  std::cout << "Current dir:" << path << std::endl;
+
+  auto triangulate = [&path](int index)
+  {
+    CDT cdt;
+
+    auto write_tr = [&](const std::string& filename)
+    {
+//       std::ofstream file(path + filename + "_" + std::to_string(index) + ".off");
+//
+//       cdt.file_output(file);
+//       file.close();
+    };
+
+    Vertex_handle va = cdt.insert(Point(-0.74397572, -0.54545455));
+    Vertex_handle vb = cdt.insert(Point(-0.13526831, -1));
+    Vertex_handle vc = cdt.insert(Point(0.067634156, -1));
+    Vertex_handle vd = cdt.insert(Point(0.33817078, -0.54545455));
+    Vertex_handle ve = cdt.insert(Point(0.74397572, 0.27272727));
+    Vertex_handle vf = cdt.insert(Point(0.74397572, 0.54545455));
+    Vertex_handle vg = cdt.insert(Point(0.067634156, 1));
+    Vertex_handle vh = cdt.insert(Point(-0.13526831, 1));
+    Vertex_handle vi = cdt.insert(Point(-0.74397572, -0.18181818));
+
+    cdt.insert_constraint(va, vb);
+    cdt.insert_constraint(vb, vc);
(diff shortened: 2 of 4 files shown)


Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
```diff
@@ -67,6 +68,18 @@ class Delaunay_mesh_face_base_2 : public Fb
   /** compatibility with CGAL-3.2 */
   inline
   void set_marked(const bool b) { in_domain=b; }
+
+  
+    typedef Tag_true Has_timestamp;
```
sloriot, line 73:
```suggestion
  typedef Tag_true Has_timestamp;
```

Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
```diff
@@ -67,6 +68,18 @@ class Delaunay_mesh_face_base_2 : public Fb
...
+
+  std::size_t time_stamp() const {
+    return time_stamp_;
+  }
+  void set_time_stamp(const std::size_t& ts) {
```
sloriot, line 78:
```suggestion
  void set_time_stamp(const std::size_t ts) {
```

Mesh_2/test/Mesh_2/reproductibility.cpp
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const size_t ref_number_of_vertices = triangulate(0);
+
+  for (int i = 1; i < 20; ++i)
+  {
+    const size_t number_of_vertices = triangulate(i);
+    if (ref_number_of_vertices != number_of_vertices)
+      return 1;
+  }
```
lrineau, line 102: Instead of comparing just the number of vertices, I think you should compare the triangulations. You can use the I/O, for that:

```c++
  constexpr int number_of_iterations = 20;
  std::vector<std::string> triangulation_strings(number_of_iterations);
  for (int i = 0; i < number_of_iterations; ++i)
  {
    auto cdt = triangulate(i);
    std::stringstream ss;
    ss << cdt;
    triangulation_strings[i] = ss.str();
    if (i > 0 && triangulation_strings[i] != triangulation_strings[0])
      return 1;
  }
```

(I have not tested.) `triangulate(int)` will also have to be changed to return the triangulation.

Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
afabri: It is copy/paste from the 3D case.

afabri:
@lrineau the important change is [here](https://github.com/CGAL/cgal/pull/4690/files#diff-85e543718faf40353d9859d754a325fdR1035).   Calling `create_face(Face_handle)` which makes a copy messes up the time stamper. Do I pass enough arguments when I change the function? The function is not documented.   In the TDS_3 package it is [documented](https://doc.cgal.org/latest/TDS_3/classTriangulationDataStructure__3.html#a1432860206073c24ca43dbbdfb13b26e) as advanced.

afabri:
@lrineau the important change is [here](https://github.com/CGAL/cgal/pull/4690/files#diff-85e543718faf40353d9859d754a325fdR1035).   Calling `create_face(Face_handle)` which makes a copy messes up the time stamper. Do I pass enough arguments when I change the function? The function is not documented.   In the TDS_3 package it is [documented](https://doc.cgal.org/latest/TDS_3/classTriangulationDataStructure__3.html#a1432860206073c24ca43dbbdfb13b26e) as advanced.

Mesh_2/test/Mesh_2/reproductibility.cpp
afabri: I just integrated the comparison of the triangulations.

lrineau: And your implementation is better than mine. :+1:

lrineau:
> @lrineau the important change is [here](https://github.com/CGAL/cgal/pull/4690/files#diff-85e543718faf40353d9859d754a325fdR1035). Calling `create_face(Face_handle)` which makes a copy messes up the time stamper. Do I pass enough arguments when I change the function? The function is not documented. In the TDS_3 package it is [documented](https://doc.cgal.org/latest/TDS_3/classTriangulationDataStructure__3.html#a1432860206073c24ca43dbbdfb13b26e) as advanced.

The current implementation of the time-stamper has a lot of defects:
  - it uses an undefined behavior, by initializing the timestamp of objects of the `Compact_container`, before the objects are created,
  - because of that, the constructors (including the copy-constructors of the object class) **must not** initialize the timestamps themselves.

In the case of `create_face(f)`, I think the copy-constructor of the face class is compiler-generated, and thus copy the timestamp.

afabri:
> In the case of `create_face(f)`, I think the copy-constructor of the face class is compiler-generated, and thus it copies the timestamp of `f` into the new face.

That I found out, and that's why I replace the call. The question is if my replacement is correct.

lrineau:
> That I found out, and that's why I replace the call. The question is if my replacement is correct.

It seems right. Anyway, you have the CGAL test suite (and CTest) to check.

afabri:
@sloriot I consider it no longer as work in progress but ready for testing. Can you please change.

sloriot:
```
/home/sloriot/CGAL/git/branch/Mesh_2/test/Mesh_2/reproductibility.cpp: In lambda function:
/home/sloriot/CGAL/git/branch/Mesh_2/test/Mesh_2/reproductibility.cpp:37:44: warning: unused parameter ‘filename’ [-Wunused-parameter]
     auto write_tr = [&](const std::string& filename)
                         ~~~~~~~~~~~~~~~~~~~^~~~~~~~
/home/sloriot/CGAL/git/branch/Mesh_2/test/Mesh_2/reproductibility.cpp: In function ‘int main(int, char**)’:
/home/sloriot/CGAL/git/branch/Mesh_2/test/Mesh_2/reproductibility.cpp:26:14: warning: unused parameter ‘argc’ [-Wunused-parameter]
          ~~~~^~~~
```

sloriot:
```
The following tests FAILED:
	  3 - execution___of__conform_plus (Child aborted)
	  9 - execution___of__test_conforming (Child aborted)
Errors while running CTest
```

sloriot:
What about the TODOs in the description?

afabri:
> > In the case of `create_face(f)`, I think the copy-constructor of the face class is compiler-generated, and thus it copies the timestamp of `f` into the new face.
> 
> That I found out, and that's why I replace the call. The question is if my replacement is correct.

Well it is not correct, as I only copy connectivity, and not for example the `bool` for marking constraints.

MaelRL:
Discussed semi-offline with Andreas: my point of view is that it's faulty to abuse the fact that `create_face(f)` is going to copy states such as `is_in_domain()`. The face base copy constructor should stay compiler-generated (and thus copying the timestamp), and a fix is required in the triangulations to set the flags of new faces properly.

lrineau:
> ```
> The following tests FAILED:
> 	  3 - execution___of__conform_plus (Child aborted)
> 	  9 - execution___of__test_conforming (Child aborted)
> Errors while running CTest
> ```

If I run the tests of Triangulation_2 as well, I get:
```
The following tests FAILED:
        3135 - execution___of__conform_plus (SEGFAULT)
        3139 - execution___of__test_conforming (Child aborted)
        4474 - execution___of__test_const_del_triangulation_2 (Child aborted)
        4478 - execution___of__test_constrained_triangulation_2 (Child aborted)
```
So that is not specific to Mesh_2.

TDS_2/include/CGAL/Triangulation_data_structure_2.h
```diff
@@ -1002,26 +1003,94 @@ insert_dim_up(Vertex_handle w,  bool orient)
...
+      set_adjacency(le, index, g, 0);
+      index = 1; // for connecting new faces
+      le = g;
+    }
+    set_adjacency(g, 1, right, 2);
+    v->set_face(g);
+    w->set_face(left);
+  }
+  break;
+#if 0
```
lrineau, line 1089: Please avoid committing `#if 0` in CGAL. Just remove the code.

MaelRL:
Replaced by #4726.

PR #4696, opened 2020-04-30
Clarify doc

### LINKED ISSUE #4158: Arr_2: Issue with visitor in post face insertion
## Issue Details

The doc of [after_split_face](https://doc.cgal.org/latest/Arrangement_on_surface_2/classCGAL_1_1Arr__observer.html#ae382d8a32ea9ef3fab11434705f57dea) says `the existing face f1 has been split, such that a portion of it now forms a new face f2`.

In practice, we don't always have that `f1` is the old face and `f2` is the new face.

For example: create a square with edges `<(0,0), (10,0), (10, 10), (0,10), (0,0)>` and then a triangle which partitions the square and the unbounded face into two faces. The triangle’s edges are `<(3,7), (7,7), (5,12), (3,7)>` (inserting segments one by one). For some insertions, `f2` is the old face.

Base code found [here](https://gist.github.com/MaelRL/7a1033ae0712ae992c1d448dfce516b6) (just need to plug-in the segment values).

sloriot:
Fixes #4158

### DIFF
--- a/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/CGAL/Arr_observer.h
+++ b/Arrangement_on_surface_2/doc/Arrangement_on_surface_2/CGAL/Arr_observer.h
@@ -285,8 +285,8 @@ virtual void before_split_face (Face_handle f,
 Halfedge_handle e);
 
 /*!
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #4701, opened 2020-05-05
Fix remaining std::max

maxGimeno:
## Summary of Changes

Fix a remaining std::max in a plugin of the demo
## Release Management

* Affected package(s):Polyhedron demo


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-148.shtml

PR #4702, opened 2020-05-05
Simplify example

sloriot:
also add some debug code


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-148.shtml

PR #4709, opened 2020-05-06
Surface_mesh: fix clear()

maxGimeno:
## Summary of Changes
The function clear() of the Surface_mesh celars the property_maps but does not remove them. 
This PR remove all the maps in clear(), then creates the default maps again.

## Release Management

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -39,7 +39,7 @@ Release History
 
 ### Surface Mesh
 
- - The function `CGAL::Surface_mesh::clear()` now removes all non-default properties instead of just emptying them. 
+ - **Breaking change**: The function `CGAL::Surface_mesh::clear()` now removes all non-default properties instead of just emptying them. 
 
 Release 5.0
 -----------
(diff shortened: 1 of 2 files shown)


sloriot:
You probably want to update the doc by copying what is said in `Point_set_3`.

sloriot:
My feeling about this change is that it should target only 5.1 and be mentioned in the change log as even if it is a bug-fix the doc was not precise enough and users might have understood that maps will be kept.

maxGimeno:
Ok for me if you convince @lrineau

lrineau:
I agree with Sébastien. That is a breaking change.

maxGimeno:
TODO: update doc

maxGimeno:

lrineau:
@maxGimeno You added a TODO: https://github.com/CGAL/cgal/pull/4709#issuecomment-627384452

> TODO: update doc

As far as I know, that is not yet done.

maxGimeno:
Indeed

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-156.shtml

lrineau:
I think this "feature" should be qualified as "breaking change", in the `CHANGES.md`.

sloriot:
I'm actually affected by this breaking change in one of my code. We should have add a bool for example to not remove the properties. I'll open a PR to have a way to restore the old behavior.

PR #4723, opened 2020-05-14
Two bug fixes in Surface mesh topology package

gdamiand:
## Summary of Changes

1. computation of cycle lengths, when some halfedges are flip;
2. when we use of face graph wrapper with surface mesh we can not use nullptr.

## Release Management

* Affected package(s): Surface_mesh_topology


maxGimeno:

PR #4726, opened 2020-05-15
Mesh_2: Make it deterministic  V2

afabri:
## Summary of Changes

Use a time stamper for the mesh face base class. This makes the construction deterministic.
The Edge_iterator is also deterministic as it compares face handles which means comparing time stamps.

We decided not to document the time stamp mechanism, and not to document that the mesh generator is deterministic.

I tried it with two more test cases: the contour of norway, as well as many segments all intersecting in a single point in order to have more complicated clusters.

I replaced the call of `create_face()` which makes a copy of a face and messes up the time stamper, with a call of `create_face()` which only copies the connectivity. 
This breaks the 2D constrained triangulation testsuite because the bool that  marks a constrained edge gets lost.  In PR #4690 I tried to fix this by rewriting `TDS_2::insert_dim_up()`. I would have continued to exploit an implementation detail.   In this PR  I collect the  constrained edges of a 1D triangulation, and once it is 2D I restore the lost information.   

## Release Management

* Affected package(s):  TDS_2, Triangulation_2, Mesh_2
* License and copyright ownership:  Inria

### DIFF
--- a/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
+++ b/Mesh_2/include/CGAL/Delaunay_mesh_face_base_2.h
@@ -17,6 +17,7 @@
 
 
 #include <CGAL/Constrained_Delaunay_triangulation_face_base_2.h>
+#include <CGAL/Has_timestamp.h>
 
 namespace CGAL {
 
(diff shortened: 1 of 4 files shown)


maxGimeno:
Travis shows numerous errors

afabri:
@maxGimeno can you have a look at travis please.

maxGimeno:
The testsuite shows warnings for this PR : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-156.shtml#Mesh_2

afabri:
> 
> 
> From the changes it seems that this PR also fixes an issue in CDT for dimension 1, is that the case?

The additional code in the CDT is necessary as we no longer use the copy constructor for face in the TDS, which teleported the Boolean marker of a constraint. And we do not need additional code for dimension 2, because the Boolean markers on the boundary of the star of the inserted vertex get retrieved from the faces adjacent and outside of the star.

maxGimeno:

lrineau:
Its seems this PR is ready for 5.1, even if it was initially scheduled for 5.2.

PR #4734, opened 2020-05-19
CGAL 3D Demo: Enhancements and fixes

maxGimeno:
## Summary of Changes

- [x] Remove the WS server from the demo and make it a standalone executable
- [x] SSH actions use ssh-agent if able
- [x] Make camera movements from WS smoother
- [x] Filter out some useless GL warnongs
- [x] Fix some typos

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.cpp
+++ b/Polyhedron/demo/Polyhedron/Viewer.cpp
@@ -1691,6 +1691,9 @@ void Viewer::setTotalPass(int p)
 void Viewer::messageLogged(QOpenGLDebugMessage msg)
 {
   //filter out useless warning
+  // From those two links, we decided we didn't care for this warning:
+  // https://community.khronos.org/t/vertex-shader-in-program-2-is-being-recompiled-based-on-gl-state/76019
+  // https://stackoverflow.com/questions/12004396/opengl-debug-context-performance-warning
   if(msg.message().contains("is being recompiled"))
     return;
   QString error;
--- a/Polyhedron/demo/Polyhedron/Viewer.cpp
+++ b/Polyhedron/demo/Polyhedron/Viewer.cpp
@@ -2021,6 +2021,9 @@ void Viewer::onTextMessageSocketReceived(QString message)
     return;
   }
   QStringList sl = position.split(" ");
+  if(sl.size() != 7)
+    return;
+
   CGAL::qglviewer::Vec pos(sl[0].toDouble(),sl[1].toDouble(),sl[2].toDouble());
   CGAL::qglviewer::Quaternion q(sl[3].toDouble(),sl[4].toDouble(),
       sl[5].toDouble(),sl[6].toDouble());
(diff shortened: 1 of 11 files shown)


Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -1690,6 +1690,9 @@ void Viewer::setTotalPass(int p)
 
 void Viewer::messageLogged(QOpenGLDebugMessage msg)
 {
+  //filter out useless warning
+  if(msg.message().contains("is being recompiled"))
+    return;
```
lrineau, line 1695: Before we silence that warning, I would like you to explain why the NVidia OpenGL implementation emits a warning.

sloriot:
Working very well and now the arcball is very smooth.

Polyhedron/demo/Polyhedron/Viewer.cpp
maxGimeno: I've read some things about NVIDIA not liking when a vertex shader is used multiple times, other things about NVidia not liking when a fragment shader is not bound, and mostly ppl saying it is a very useless warning.

Polyhedron/demo/Polyhedron/Viewer.cpp
```diff
@@ -2020,7 +2020,12 @@ void Viewer::onTextMessageSocketReceived(QString message)
   if(session != d->session){
     return;
   }
-  moveCameraToCoordinates(position, 0.05f);
+  QStringList sl = position.split(" ");
+  CGAL::qglviewer::Vec pos(sl[0].toDouble(),sl[1].toDouble(),sl[2].toDouble());
+  CGAL::qglviewer::Quaternion q(sl[3].toDouble(),sl[4].toDouble(),
+      sl[5].toDouble(),sl[6].toDouble());
```
lrineau, line 2026: This code is not safe. It should check that `sl` has indeed 7 elements, otherwise there could be a array bound overflow, triggered by string received from an unsafe channel, possibly controlled by an attacker.

Polyhedron/demo/Polyhedron/Viewer.cpp
lrineau: Can you please give links? I think the links should be in a comment in the code.

maxGimeno: https://community.khronos.org/t/vertex-shader-in-program-2-is-being-recompiled-based-on-gl-state/76019

https://stackoverflow.com/questions/12004396/opengl-debug-context-performance-warning

lrineau: That seems indeed a warning that we can ignore. Please insert those two URLs in the comments of the code.

maxGimeno:
C:\3rdPartyLibs\boost_1_73_0\boost/bind/bind.hpp(75): error C3646: 'type': unknown override specifier

maxGimeno:

PR #4749, opened 2020-05-26
Polygon Mesh Processing triangulate hole with CDT 2

### LINKED ISSUE #4464: Delaunay triangulation goes outside border

danston:
## Summary of Changes

This PR extends the solution for the issue #4464 that was proposed in the PR #4483.

I have updated the Maxime's code and added simplicity and planarity checks. If the hole's boundary is near planar and its projection in 2D is simple, we use CDT2 to close the hole, otherwise we revert back to the original solution with the DT3.

## Release Management

* Issue(s) solved (if any): fix #4464, PR #4483 
* Feature/Small Feature (if any): [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/PMP_triangulate_hole_with_CDT2)
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.github.io/4749/v1/Manual/index.html)


lrineau:
Travis checks failed: PMP now depends on PCA. Is that wanted?

sloriot:
@danston the projection traits I mentioned is not the xx but the one in `CGAL/Triangulation_2_projection_traits_3.h` that works with an arbitrary vector.

sloriot:
I think it is not needed that the hole is planar, it somehow requires that along the chosen projection the polygons are projected as valid polygons (which is why I was mentioning `is_simple`)

sloriot:
This branch should target 5.1 and not 4.14, I don't see it as a bug-fix.

danston:
> @danston the projection traits I mentioned is not the xx but the one in `CGAL/Triangulation_2_projection_traits_3.h` that works with an arbitrary vector.

@sloriot Ok, I did not know about them. I will have a look inside these traits. You are actually right, the planarity is not required for a valid triangulation in this case. @lrineau I think I can get rid of PCA, I did not like this dependency, too.

I will do the new thing asap.

danston:
Hi all. I have updated the code using the Sebastien's suggestion. I removed all dependencies like PCA and extended the `Triangulation_2_projection_traits_3` class with the `Less_xy` object that was the only missing object in order to use the `is_simple_2` function. Please check if the updated traits is ok. It is my first time I modify the global traits class, so I would like to be sure I did not make a stupid mistake or forgot to add something important. Thanks.

sloriot:
1. The modification of the traits looks good. 
2. Did you investigate if the `is_simple` test could be replaced by checking that no constraint intersect while creating the 2D constrained triangulation? In particular I'm referring to this sentence [here](https://cgal.geometryfactory.com/CGAL/doc/master/Triangulation_2/classCGAL_1_1Constrained__triangulation__2.html):
*In the basic version, the constrained triangulation does not handle intersecting constraints, and the set of input constraints is required to be a set of polylines that do not intersect except possibly at their endpoints. An exception of type Intersection_of_constraints_exception is thrown upon insertion of a constraint intersecting in its interior an already inserted constraint. Any number of constrained edges may share the same endpoint. Constrained edges may be vertical or have zero length.* 
I guess a good `Itag` parameter should also be used in order to correctly handle non-manifold vertices.
3. You should add a test where using the cdt is not possible and is correctly detected. The case of the two vertices having with the projection the same coordinate is particularly interesting.
4. Should we add a new named parameter in the main function `triangulate_hole()` to allow the usage of the new method?

danston:
1. Great.
2. Not yet. I know the `Itag` you point to since I am using it quite often in my code, but I have not checked how we can use it here. I will investigate.
3. Good point, I will add a few tests with different configurations.
4. I am not sure about that. As I wrote in the description to this PR, in the new method, we need to use the `triangulate_face()` function that is in the header, which already includes the `triangulate_hole()` function and so we get the circular inclusion. How can I handle that? Or again maybe I am missing something. Thanks.

danston:
I have added a few tests and the current code seems to work on a number of cases (see below). The non-simple hole is correctly detected and the original solution with DT3 is used.

**Horizontal Hole:**
hor_before
hor_after

**Orthogonal Hole:**
orth_before
orth_after

**Simple Hole:**
simple_before
simple_after

**Non-Simple Hole:**
nonsimple_before
nonsimple_after

danston:
I have investigated the `Itag` property. The `No_intersection_tag` is deprecated. The currently used tag is `Exact_intersections_tag`. This is the most robust tag. When using this tag and triangulating a polygon mesh face, which is a hole, using CDT, the intersecting constraints are allowed. In our case, the constraints are the edges of a hole. When the hole is not simple, several edges may intersect each other, which creates extra vertices at the place of intersections. Then checking the condition `cdt.number_of_vertices() != number_of_original_vertices`, where the `number_of_original_vertices` is the number of the hole's boundary vertices, is enough to see if the hole's boundary is not simple, in which case we turn back to the original solution that is `triangulate_hole()` with DT3. When using this condition, all tests above success.

sloriot:
Note that when using `Exact_intersections_tag`, you should get an exception thrown when the constraints intersect in their interior. However, the check on the number of vertices is still needed to make sure no two points on the hole are identical when "projected".

afabri:
In case there are intersections the choice of the tag should depend on if the kernel does exact constructions or not.   Did you hardwire the kernel?

sloriot:
@afabri if they are intersection, the method cannot be used and an exception should be thrown

danston:
The kernel is not hardwired, but the tag is always `Exact_intersections_tag`. I agree with Sebastien that we should detect if we can use the CDT method and if not, turn back to the original solution. Three tests above show that the method can be used and in one test, the hole is not simple, the intersection is detected and we use the DT3 method instead.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
...
+  face_descriptor new_face = add_face(pmesh);
+  set_halfedge(new_face, border_halfedge, pmesh);
+  do {
+    set_face(*circ, new_face, pmesh);
+  } while (++circ != done);
+
+  // Triangulating.
+  const bool use_cdt = true;
+  internal::Triangulate_modifier<PolygonMesh, VPM, Traits> modifier(vpm, traits);
+  const bool success_with_cdt_2 =
```
sloriot, line 719: OK here is the part I missed in my previous review. The intersection of constraints as well has the vertices being identical upon projects is handled in the function `triangulate_face()`.

sloriot:
One thing that we need to decide is if the possibility to use the cdt shouldn't be a named parameter in the general `triangulate_hole()` function, like it is already the case for DT3 search space. Of course, we have to properly handle the extra dependency of the function wrt to CDT2 (like we did with DT3 + cf header `CGAL/Polygon_mesh_processing/internal/do_no_use_CDT2.`)

danston:
@sloriot Thanks for the review. I will take care of it and answer your questions asap.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1203,176 @@ class Triangulate_hole_polyline {
...
+    // Skip in case we have collinear points.
+    if (collinear_3(p1, p2, p3))
+      continue;
+
+    // Computing the normal of a triangle.
+    Vector_3 tri_normal = CGAL::normal(p1, p2, p3);
+    const FT tri_normal_length = static_cast<FT>(
+      CGAL::sqrt(CGAL::to_double(squared_length_3(tri_normal))));
+    if(tri_normal_length <= FT(0))
+      return false;
```
sloriot, line 1262: why not simply `continue`?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1203,176 @@ class Triangulate_hole_polyline {
...
+
+    tri_normal /= tri_normal_length;
+
+    x += tri_normal.x(); y += tri_normal.y(); z += tri_normal.z();
```
sloriot, line 1266: Note that depending on the shape of the hole some normals might be pointing up and some be pointing down. It is probably better to always take a "positive" normal for the sum and then pick the right orientation using the hole orientation (or using the third point for the case of the polyline).

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1203,176 @@ class Triangulate_hole_polyline {
...
+  }
+
+  FT x = FT(0), y = FT(0), z = FT(0);
+  std::size_t num_normals = 0;
+  const Point_3& ref_point = P[0];
+  std::size_t size = P.size()-1;
+  for (std::size_t i = 1; i < size; ++i) {
+    const std::size_t ip = i + 1;
+
+    const Point_3& p1 = ref_point; // 3 points, which form a triangle
```
sloriot, line 1249: Any reason why you use a fixed point for P1? instead of for example taking triples of consecutive points?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1203,176 @@ class Triangulate_hole_polyline {
...
+     return false;
+
+
+  Lookup_table_map<int> lambda(static_cast<int>(size),-1);
+  //Create and fill the cdt_2
+  typedef CGAL::Triangulation_vertex_base_with_info_2<std::size_t, P_traits> Vb;
+  typedef CGAL::Triangulation_face_base_with_info_2<bool, P_traits>          Fb1;
+  typedef CGAL::Constrained_triangulation_face_base_2<P_traits, Fb1>         Fb;
+  typedef CGAL::Triangulation_data_structure_2<Vb,Fb>                        TDS;
+  typedef CGAL::Exact_intersections_tag                                      Itag;
```
sloriot, line 1296: if we keep the `is_simple()` test above I think we should not allow intersection of constraints (because they should simply never happen). I would update that tag then.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1203,176 @@ class Triangulate_hole_polyline {
...
+  typedef CGAL::Constrained_Delaunay_triangulation_2<P_traits, TDS, Itag>    CDT;
+  P_traits cdt_traits(normal);
+  CDT cdt(cdt_traits);
+
+  std::vector< std::pair<Point_3,std::size_t> > points_and_ids;
```
sloriot, line 1301: `std::vector::reserve()`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -29,6 +29,10 @@
 #include <CGAL/use.h>
 #endif
 
+#include <CGAL/assertions.h>
```
sloriot, line 32: AFAIS, all changes in this file can be undo.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -99,13 +102,19 @@ namespace Polygon_mesh_processing {
 #endif
 
     CGAL_precondition(face(border_halfedge, pmesh) == boost::graph_traits<PolygonMesh>::null_face());
+    bool use_cdt =
+    #ifdef CGAL_HOLE_FILLING_DO_NOT_USE_CDT2
+        false;
+#else
+        choose_parameter(get_parameter(np, internal_np::use_2d_constrained_delaunay_triangulation), true);
```
sloriot, line 109: I don't know if the default should be `true`. opinions?

maxGimeno: I'd say it depends of the most usual cases. IMHO, if most of the time, a hole is not planar, it could be debated, but if most holes are small and planar, it should be `true` as it will be more stable.

maxGimeno:
@danston, I took care of some of it, and marked as resolved when that's the case. I leave the rest to you

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: So, the reason is that if the whole has multiple collinear vertices along its boundary, taking the fixed first point should lead to a higher number of non-degenerate triangles. Moreover, in this case, the generated triangles cover the hole and hence approximate its shape somehow, which should lead to better normal estimation. Using only consecutive points takes into account only the region in the vicinity of the hole boundary, but not the hole-itself.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: This is a good idea. I will implement it.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
danston: Probably yes. I will see if this is possible.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
   return triangulate_faces(faces(pmesh), pmesh, CGAL::Polygon_mesh_processing::parameters::all_default());
 }
 
+/*!
+  \ingroup hole_filling_grp
+  \brief triangulates a planar hole in a polygon mesh.
+
+  If the hole is planar, this function uses the 2D constrained Delaunay triangulation
```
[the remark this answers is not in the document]
danston: Right. I actually copy pasted that text from some other function and then forgot to modify it after my recent changes. I will write it better.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
...
+  in order to close the hole. The constraints are the border edges of the hole.
+
+  The hole must not contain any non-manifold vertex, nor self-intersections.
```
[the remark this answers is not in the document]
danston: Exactly. But I should probably describe it better, where it is important and where it is not.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
...
+  \param border_halfedge a border halfedge incident to the hole
+  \param out iterator over patch faces
+  \param np optional sequence of \ref pmp_namedparameters "Named Parameters" among the ones listed below
+
+  \cgalNamedParamsBegin
+    \cgalParamBegin{vertex_point_map} the property map with the points associated to the vertices of `pmesh`.
+        If this parameter is omitted, an internal property map for
+        `CGAL::vertex_point_t` should be available in `PolygonMesh`
+        \cgalParamEnd
+    \cgalParamBegin{geom_traits} a geometric traits class instance \cgalParamEnd
```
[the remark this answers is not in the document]
danston: I am not sure about that. I think what we have is enough to estimate the normal quite well and this normal is only used to check the simplicity of the hole boundary. But adding it to the interface will complicate it and the question here is how often it is going to be used and is it useful at all?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
...
+    // Skip in case we have collinear points.
+    if (collinear_3(p1, p2, p3))
+      continue;
+
+    // Computing the normal of a triangle.
+    Vector_3 tri_normal = CGAL::normal(p1, p2, p3);
+    const FT tri_normal_length = static_cast<FT>(
+      CGAL::sqrt(CGAL::to_double(squared_length_3(tri_normal))));
+    CGAL_assertion(tri_normal_length > FT(0));
+    tri_normal /= tri_normal_length;
```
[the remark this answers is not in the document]
danston: Yes, it could be tested. I agree.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -533,6 +579,159 @@ bool triangulate_faces(PolygonMesh& pmesh)
...
+  // Setting the final normal.
+  const Vector_3 normal = Vector_3(x, y, z);
+  const FT normal_length = static_cast<FT>(
+    CGAL::sqrt(CGAL::to_double(squared_length_3(normal))));
+  const Vector_3 avg_normal = normal / normal_length;
+
+  // Checking the hole simplicity.
+  typedef Triangulation_2_projection_traits_3<Traits> P_traits;
+  const P_traits p_traits(avg_normal);
+  const bool is_simple =
```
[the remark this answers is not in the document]
danston: Actually, it is in the new version of Maxim, if I am not mistaken.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
danston: Yes.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
danston: In my experience, most of the holes can be triangulated with CDT. Even when I was creating a test where it is not the case, I had to take quite a specific part of the mesh. I am not an expert in holes filling, so maybe I miss something, but I would use `true`.

danston:
@sloriot, @maxGimeno I have answered all the comments and I will handle the rest.

danston:
> I was wondering if we should look at the deviation of the normals to make sure that even if the projection is planar there is not too much curvature on the hole boundary. No need to modify the code until the discussion is over.

But if it is, will it lead to what you think?

sloriot:
> > I was wondering if we should look at the deviation of the normals to make sure that even if the projection is planar there is not too much curvature on the hole boundary. No need to modify the code until the discussion is over.
> 
> @sloriot But if it is, will it lead to what you think?

My point was to find a way to avoid using the cdt when it's clearly not a good idea (something with a large global curvature). Maybe it requires some testing in the demo, after @maxGimeno updated the plugin :)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
danston: Done.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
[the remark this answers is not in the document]
danston: Done.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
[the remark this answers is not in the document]
danston: Done.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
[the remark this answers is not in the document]
danston: Done.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: Done.

danston:
A few more test cases.

**Near Flat Hole:**
flat_before
flat_after

**Curved Hole:**
curved_before
curved_after

**Concave Hole + Saddle Region:**
concave_before
concave_after

danston:
I have fixed the estimated normal orientation. It is now always a positive normal that does not affect the orientation of the final patch. The patch has always the correct orientation by construction (performed on the lookup table that is built upon the CDT and then called by a tracer) that is the final mesh is always outward-oriented (I also check for that now). I also added a few extra tests to check different hole configurations. Finally, I have added a simple planarity test that now does not depend on the PCA but is similar in fashion. It enables to filter out a few obviously not planar cases. If you are fine with it, I can add the `max_distance` parameter from this test to the named parameters.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1210,16 +1210,87 @@ class Triangulate_hole_polyline {
...
+
+  // Compute centroid.
+  FT cx = FT(0), cy = FT(0), cz = FT(0);
+  for (std::size_t i = 0; i < n; ++i) {
+    const Point_3& p = points[i];
+    cx += p.x();
+    cy += p.y();
+    cz += p.z();
+  }
+  cx /= static_cast<FT>(n);
```
afabri, line 1240: Can we static cast from an int  to a FT?   Isn't a FT constructible from int? @sloriot 
Don't we have a centroid function?

danston: Actually, we do have the centroid function. You are right. I will change that. It is so simple that I forgot about it.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1210,16 +1210,87 @@ class Triangulate_hole_polyline {
  ************************************************************************/
 // /!\ points.first == points.last
 
+template <typename Traits>
+bool is_planar_2(
```
afabri, line 1214: Would this function not be of general interest? @sloriot

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1279,37 +1341,17 @@ triangulate_hole_polyline_with_cdt(const PointRange1& points,
...
   // Setting the final normal.
   x /= static_cast<FT>(num_normals);
   y /= static_cast<FT>(num_normals);
   z /= static_cast<FT>(num_normals);
   const Vector_3 avg_normal = Vector_3(x, y, z);
-  // std::cout << "avg: " << avg_normal << std::endl;
+  // std::cout << "avg normal: " << avg_normal << std::endl;
+
+  // Checking the hole planarity.
+  const FT max_distance = FT(2) / FT(100);
```
afabri, line 1352: Is this a constant that does not depend on the input?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1318,88 +1360,81 @@ triangulate_hole_polyline_with_cdt(const PointRange1& points,
...
-  std::vector<typename CDT::Vertex_handle> vertices(size);
   for (std::size_t i = 0; i < size; ++i)
-  {
-    points_and_ids.push_back(std::make_pair(P[i],i));
-  }
+    points_and_ids.push_back(std::make_pair(P[i], i));
+
+  std::vector<typename CDT::Vertex_handle> vertices(size);
   cdt.insert(points_and_ids.begin(), points_and_ids.end());
   for (typename CDT::Vertex_handle v : cdt.finite_vertex_handles())
```
afabri, line 1382: I find it a bad idea not to put parenthesis, and here you even replace them with an empty line.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1318,88 +1360,81 @@ triangulate_hole_polyline_with_cdt(const PointRange1& points,
...
     return false;
-  }
 
   // Fill the lambda.
   for (typename CDT::Finite_faces_iterator fit = cdt.finite_faces_begin(),
-    end = cdt.finite_faces_end(); fit != end; ++fit)
-  {
-    if (!fit->info()) // if !is external
-    {
+  end = cdt.finite_faces_end(); fit != end; ++fit) {
```
afabri, line 1416: wrong indentation.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -426,7 +496,7 @@ bool use_dt3 =
 
   /*!
   \ingroup  hole_filling_grp
-  same as above but the range of third points is omitted. They are not
+  Same as above but the range of third points is omitted. They are not
```
afabri, line 499: Check in the html that it is really "above".

Polygon_mesh_processing/test/Polygon_mesh_processing/triangulate_hole_with_cdt_2_test.cpp
```diff
@@ -96,6 +97,8 @@ void test_triangulate_hole_with_cdt_2(
     assert(patch_faces.size() == num_patch_faces);
   }
   assert(pmesh.is_valid() && is_closed(pmesh));
+  assert(CGAL::Polygon_mesh_processing::is_outward_oriented(pmesh,
+  CGAL::parameters::all_default()));
```
afabri, line 101: Please check all indentation.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: It does, but as I wrote in the comment above I have not yet added it to the input, because I would first like to get opinions on whether we need this planarity test at all. If we do not, then we do not need this constant either.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1210,16 +1210,87 @@ class Triangulate_hole_polyline {
...
+  maxv = (CGAL::max)(maxv, y);
+  maxv = (CGAL::max)(maxv, z);
+  if (maxv <= FT(0))
+    return false; // a good plane does not exist for sure!
+
+  const Plane_3 plane = Plane_3(centroid, avg_normal);
+  FT avg_sq_distance = FT(0);
+  for (std::size_t i = 0; i < n; ++i) {
+    const Point_3& p = points[i];
+    const Point_3  q = plane.projection(p);
```
afabri, line 1272: @sloriot @janetournois Is it ok that constructors and member functions are used instead of kernel functors?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1210,16 +1210,87 @@ class Triangulate_hole_polyline {
...
+    avg_sq_distance += squared_distance_3(p, q);
+  }
+  avg_sq_distance /= static_cast<FT>(n);
+  // std::cout << "avg squared distance: " << avg_sq_distance << std::endl;
+
+  const FT sq_tolerance = max_distance * max_distance;
```
afabri, line 1278: `CGAL::square(max_distance)`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: I can add them back. I just shortened the code. The parenthesis should not affect anything if I am not mistaken. And the empty line makes the code more readable IMHO.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
afabri: And why is the square not directly the parameter of the function?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: Fixed.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
danston: This is a change that I included from the new version of PMP with the updated named parameters, but I can change it back.

Polygon_mesh_processing/test/Polygon_mesh_processing/triangulate_hole_with_cdt_2_test.cpp
danston: Checked.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: Ok.

danston: Regarding the question, the same as I explained above, the `max_distance` parameter is not yet finalized.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
sloriot: since the traits is a parameter of the function, we can use the traits to get the functors

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
sloriot: Yes but I don't know where we should put it.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: I will fix that.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
danston: The centroid function requires PCA, which we would like to avoid from the very beginning. So I have aborted this change.

maxGimeno:
@danston is this PR ready for testing ?

danston:
> @danston is this PR ready for testing ?

If @sloriot or @afabri do not have any other comments, which I should take into account, then yes.

maxGimeno:
Warnings : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-20/Surface_mesh_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

danston:
I see. Thanks. I will fix them asap.

sloriot:
Looks like the doc is not up-to-date. @maxGimeno Can we merge master and get the doc automatically built now?

maxGimeno:
You don't need to merge master to do that.

sloriot:
May I ask how to do it then?

sloriot:
build:doc_v2

Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
```diff
@@ -114,6 +114,7 @@ and provides a list of the parameters that are used in this package.
 - `CGAL::Polygon_mesh_processing::triangulate_and_refine_hole()`
 - `CGAL::Polygon_mesh_processing::triangulate_refine_and_fair_hole()`
 - `CGAL::Polygon_mesh_processing::triangulate_hole_polyline()`
+- `CGAL::Polygon_mesh_processing::triangulate_hole_with_cdt_2()`
```
sloriot, line 117: This should be removed, right?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -42,29 +42,45 @@ namespace Polygon_mesh_processing {
   /*!
   \ingroup hole_filling_grp
   triangulates a hole in a polygon mesh.
-  The hole must not contain any non-manifold vertex,
-  nor self-intersections.
-  The patch generated does not introduce non-manifold edges nor degenerate triangles.
+  The hole must not contain any non-manifold vertex, nor self-intersections.
+  When using the 2D constrained Delaunay triangulation, the non-manifold vertex is
+  not an issue. The patch generated does not introduce non-manifold edges nor degenerate triangles.
```
sloriot, line 47: the named parameter is missing

maxGimeno:
There are conflicts now

Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt
danston: Yes. I removed that.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
danston: I added the missing parameters.

danston:
Done. Should be fine now.

sloriot:
build:doc_v3

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1204,250 @@ class Triangulate_hole_polyline {
...
+  typedef typename Traits::FT FT;
+  typedef typename Traits::Point_3 Point_3;
+  typedef typename Traits::Vector_3 Vector_3;
+  typedef typename Traits::Collinear_3 Collinear_3;
+
+  // Compute an average normal of the hole.
+  const Collinear_3 collinear_3 =
+    traits.collinear_3_object();
+
+  std::vector<Point_3> P(boost::begin(points), boost::end(points));
```
sloriot, line 1304:
```suggestion
  std::vector<Point_3> P(std::begin(points), std::end(points));
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
sloriot: To be on the safe side, I'll put it to `false` by default and while we test it on real cases, we can reevaluate the decision and switch to `true` by default.

sloriot (commented):
I think it will not work but could you please try the new code on #4977? Simply stop if the cdt2 does not work (otherwise you'll fall in the very long computation time).

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
```diff
@@ -1194,7 +1204,250 @@ class Triangulate_hole_polyline {
...
+
+  // Setting the final normal.
+  x /= static_cast<FT>(num_normals);
+  y /= static_cast<FT>(num_normals);
+  z /= static_cast<FT>(num_normals);
+  const Vector_3 avg_normal = Vector_3(x, y, z);
+  // std::cout << "avg normal: " << avg_normal << std::endl;
+
+  // Checking the hole planarity.
+  const FT max_sq_distance = FT(4) / FT(10000);
```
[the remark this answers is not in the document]
danston: Yep. I did not change it because we have not yet decided in the discussion above if we should publish the planarity function on its own. If we do not have a clear path now, I can simply put it in the named parameters and document.

danston:
> I think it will not work but could you please try the new code on #4977? Simply stop if the cdt2 does not work (otherwise you'll fall in the very long computation time).

Sure. I will do that asap.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h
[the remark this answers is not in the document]
danston: Ok, done.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polygon_mesh.h
```diff
@@ -95,7 +95,9 @@ triangulate_hole_polygon_mesh(PolygonMesh& pmesh,
             VertexPointMap vpmap,
             bool use_delaunay_triangulation,
             const Kernel& k,
-            const bool use_cdt)
+            const bool use_cdt,
+            const typename Kernel::FT max_squared_distance =
+            typename Kernel::FT(4) / typename Kernel::FT(10000))
```
sloriot, line 100: the default should definitely be related to the input, using the bbox of the polyline for example. These days I'm working on models with a bbox of 1e-07 for example.

danston: I see. You are right.

sloriot: actually the default value here is not needed, right? only the call matters.

danston: Yes, her not really, but I need a default for the named parameters anyway.

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-25/Optimal_bounding_box/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

danston:
Fixed warnings.

maxGimeno:

lrineau:
> Successfully tested in [cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-27.shtml](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-27.shtml)

This PR is actually a small feature, or should have been one:
https://github.com/CGAL/cgal/pull/4749/files#diff-937f09ec7e605b8feb094fc669f95ea4537f90d02ebabf545b0c0d40e1855d08R86-R103

I am waiting for an explicit approval from those who reviewed the code.

lrineau:
I am pretty sure the code and documentation were already reviewed, and approved, long ago. But I would like a confirmation.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -81,6 +83,24 @@ namespace Polygon_mesh_processing {
...
+      \cgalParamExtra{If no valid 2D triangulation can be found, the algorithm
+                      falls back to the 3D Delaunay triangulation search space and/or the general search space to find a solution.
+                      This parameter is a good choice for near planar holes.}
+    \cgalParamNEnd
+
+    \cgalParamNBegin{max_squared_distance}
+      \cgalParamDescription{The maximum squared distance between the vertices of
+                            the hole boundary and the least squares plane fitted to this boundary.}
+      \cgalParamType{double}
+      \cgalParamDefault{squared one quater of the hole bounding box height}
```
afabri, line 100: quater -> quarter

In German you can concatenate nouns to get a new one, but even in German "hole bounding box height"  would be too much :>    What about   "of the height of the bounding box of the hole".   Three times of, but I think clearer.

BGL/include/CGAL/boost/graph/parameters_interface.h
```diff
@@ -47,6 +47,8 @@ CGAL_add_named_parameter(geom_traits_t, geom_traits, geom_traits)
 CGAL_add_named_parameter(vertex_incident_patches_t, vertex_incident_patches, vertex_incident_patches_map)
 CGAL_add_named_parameter(density_control_factor_t, density_control_factor, density_control_factor)
 CGAL_add_named_parameter(use_delaunay_triangulation_t, use_delaunay_triangulation, use_delaunay_triangulation)
+CGAL_add_named_parameter(use_2d_constrained_delaunay_triangulation_t, use_2d_constrained_delaunay_triangulation, use_2d_constrained_delaunay_triangulation)
+CGAL_add_named_parameter(max_squared_distance_t, max_squared_distance, max_squared_distance)
```
sloriot, line 51: Since we have `threshold_distance`, what about `threshold_squared_distance()`? If the square is not "important" you could also use the distance.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
danston: Good point. I will change it.

BGL/include/CGAL/boost/graph/parameters_interface.h
danston: Yes, maybe I should reuse the distance without square. I now think that it is simpler to understand as a parameter name. I will fix that.

danston:
I fixed the docs + params. I think we should better test this PR one more time since I slightly changed the code related to the params. I do not think, it broke anything, but better to test, just in case.

danston:
I have added a small feature and updated the description accordingly.

afabri:
Shouldn't it be `distance_threshold` instead of `threshold_distance` ?

danston:
> Shouldn't it be `distance_threshold` instead of `threshold_distance` ?

I have reused the existing parameter in order to avoid duplication. It is already called `threshold_distance`.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -81,6 +83,24 @@ namespace Polygon_mesh_processing {
       \cgalParamExtra{If no valid triangulation can be found in this search space, the algorithm
                       falls back to the non-Delaunay triangulations search space to find a solution.}
     \cgalParamNEnd
+
+    \cgalParamNBegin{use_2d_constrained_delaunay_triangulation}
+      \cgalParamDescription{If `true`, use the 2D constrained Delaunay triangulation facet search space.}
```
sloriot, line 88: using CDT is not a method using a search space

sloriot: search space means that the algorithm is going to try all possible combinations of triangles from a set of possible triangles (= search space). DT3 is a restricted search space and the full search space is all possible triangles.

danston: You are right. I copy-pasted the text but did not notice that. I will fix that.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -42,8 +43,9 @@ namespace Polygon_mesh_processing {
   /*!
   \ingroup hole_filling_grp
   triangulates a hole in a polygon mesh.
```
afabri, line 45: I would like to suggest some rewording.  And I would remove the "However.." sentence
as it is not about preconditions.

Depending on the choice of the underlying algorithm different preconditions apply.
When using the 2D constrained Delaunay triangulation, the border edges of the hole
must not intersect the surface.  Otherwise,  additionally, the boundary
of the hole must not contain any non-manifold vertex.

danston: Yes, this sounds better. I have fixed that.

maxGimeno:
runtime error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-30/Polygon_mesh_processing/TestReport_lrineau_Fedora-32-Release.gz

danston:
This is weird. I will have a look.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -192,6 +240,24 @@ namespace Polygon_mesh_processing {
                       falls back to the non-Delaunay triangulations search space to find a solution.}
     \cgalParamNEnd
 
+    \cgalParamNBegin{use_2d_constrained_delaunay_triangulation}
+      \cgalParamDescription{If `true`, use the 2D constrained Delaunay triangulation to fill the hole.}
```
sloriot, line 244:
```suggestion
      \cgalParamDescription{If `true`, the points of the boundary of the hole are used to estimate a fitting plane and a 2D constrained Delaunay triangulation is then used to fill the hole projected in the fitting plane.}
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -192,6 +240,24 @@ namespace Polygon_mesh_processing {
...
+      \cgalParamType{Boolean}
+      \cgalParamDefault{`true`}
+      \cgalParamExtra{If no valid 2D triangulation can be found, the algorithm
+                      falls back to the method using the 3D Delaunay triangulation.
+                      This parameter is a good choice for near planar holes.}
+    \cgalParamNEnd
+
+    \cgalParamNBegin{threshold_distance}
+      \cgalParamDescription{The maximum squared distance between the vertices of
```
sloriot, line 253:
```suggestion
      \cgalParamDescription{The maximum distance between the vertices of
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -283,6 +349,24 @@ namespace Polygon_mesh_processing {
                       falls back to the non-Delaunay triangulations search space to find a solution.}
     \cgalParamNEnd
 
+    \cgalParamNBegin{use_2d_constrained_delaunay_triangulation}
+      \cgalParamDescription{If `true`, use the 2D constrained Delaunay triangulation to fill the hole.}
```
sloriot, line 353:
```suggestion
      \cgalParamDescription{If `true`, the points of the boundary of the hole are used to estimate a fitting plane and a 2D constrained Delaunay triangulation is then used to fill the hole projected in the fitting plane.}
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -283,6 +349,24 @@ namespace Polygon_mesh_processing {
...
+      \cgalParamType{Boolean}
+      \cgalParamDefault{`true`}
+      \cgalParamExtra{If no valid 2D triangulation can be found, the algorithm
```
sloriot, line 356:
```suggestion
      \cgalParamExtra{If the boundary of the hole is not planar (according to the parameter `threshold_distance`) or if no valid 2D triangulation can be found, the algorithm
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -283,6 +349,24 @@ namespace Polygon_mesh_processing {
...
+                      falls back to the method using the 3D Delaunay triangulation.
+                      This parameter is a good choice for near planar holes.}
+    \cgalParamNEnd
+
+    \cgalParamNBegin{threshold_distance}
+      \cgalParamDescription{The maximum squared distance between the vertices of
```
sloriot, line 362:
```suggestion
      \cgalParamDescription{The maximum distance between the vertices of
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -409,6 +493,24 @@ namespace Polygon_mesh_processing {
...
```
sloriot, line 507:
```suggestion
      \cgalParamDescription{The maximum distance between the vertices of
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -409,6 +493,24 @@ namespace Polygon_mesh_processing {
       \cgalParamExtra{If no valid triangulation can be found in this search space, the algorithm
                       falls back to the non-Delaunay triangulations search space to find a solution.}
     \cgalParamNEnd
+
+    \cgalParamNBegin{use_2d_constrained_delaunay_triangulation}
+      \cgalParamDescription{If `true`, use the 2D constrained Delaunay triangulation to fill the hole.}
```
sloriot, line 498:
```suggestion
      \cgalParamDescription{If `true`, the points of the boundary of the hole are used to estimate a fitting plane and a 2D constrained Delaunay triangulation is then used to fill the hole projected in the fitting plane.}
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -409,6 +493,24 @@ namespace Polygon_mesh_processing {
...
+      \cgalParamType{Boolean}
+      \cgalParamDefault{`true`}
+      \cgalParamExtra{If no valid 2D triangulation can be found, the algorithm
```
sloriot, line 501:
```suggestion
      \cgalParamExtra{If the boundary of the hole is not planar (according to the parameter `threshold_distance`) or if no valid 2D triangulation can be found, the algorithm
```

danston:
I hope, it should work now.

danston:
It would be nice to see if this version passes the test suite. I remember the last time we had several run-time errors.

lrineau:
@danston There is still a conflict with this PR. Please merge `master` again, so that Maxime can schedule the testing of this PR soon.

maxGimeno:
Errors: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-48/Polygon_mesh_processing/TestReport_lrineau_ArchLinux.gz

The IO functions have changed their API with CGAL_IO, you have to update all the calls to read_off/write_off etc. Now it is read_OFF, with the format in capital.

lrineau:
> The IO functions have changed their API with CGAL_IO, you have to update all the calls to read_off/write_off etc. Now it is read_OFF, with the format in capital.

Did we break the API on purpose, without backward compatibility?! It is not documented in the [`CHANGED.md`](https://github.com/CGAL/cgal/blob/287836254f9541aa2b8faedbe08a7304d6097b7f/Installation/CHANGES.md).

MaelRL:
Looks like the deprecated versions are missing overloads without named parameters.

danston:
I have fixed the IO. In fact, the deprecated version `read_off` is there and there is also a version without NP for the `Surface_mesh` but not for the `Polyhedron`.

maxGimeno:
``` /home/runner/work/cgal/cgal/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Hole_filling/Triangulate_hole_polyline.h:1266:3: error: missing template arguments before ‘(’ token
2021-01-26T09:42:46.2420384Z    CGAL_assertion(avg_normal != Vector_3());
```

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-50/Optimal_bounding_box/TestReport_lrineau_Fedora-Release.gz

maxGimeno:

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
sloriot: Looks like the default is still to true while we decided it should be false.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_hole.h
```diff
@@ -115,13 +140,45 @@ namespace Polygon_mesh_processing {
 #endif
 
     CGAL_precondition(face(border_halfedge, pmesh) == boost::graph_traits<PolygonMesh>::null_face());
+    bool use_cdt =
+    #ifdef CGAL_HOLE_FILLING_DO_NOT_USE_CDT2
+        false;
+#else
+        choose_parameter(get_parameter(np, internal_np::use_2d_constrained_delaunay_triangulation), true);
```
sloriot, line 147:
```suggestion
        choose_parameter(get_parameter(np, internal_np::use_2d_constrained_delaunay_triangulation), false);
```

maxGimeno:

PR #4754, opened 2020-05-29
PMP: Fix parallel hausdorff

### LINKED ISSUE #4750: Fix parallel approximated Haussdorf distance
https://github.com/CGAL/cgal/blob/b8791bbd22804109f8680c5e970acf5f8d32692e/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h#L137-L141

@maxGimeno This code is not the best one to compute a maximum using thread.

Currently:
  - the code uses `tbb::parallel_for`,
  - all threads compute a "local" maximum,
  - then they compare to a shared `atomic<double>` to synchronize the value.

Note that:
  - `atomic<double>` is not guaranteed to be lock-free,
  - there is a compile error with clang+C++20, in recent test-results:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-161/Polygon_mesh_processing/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz

I think the code should use [`tbb::parallel_reduce`](https://software.intel.com/en-us/node/506154): each thread returns a value, and the main thread is responsible for the reduction of all values to the final value.

_Originally posted by @lrineau in https://github.com/CGAL/cgal/pull/1578/review_comment/create_

maxGimeno:
## Summary of Changes

Use `parallel_reduce` instead of `parallel_for` in hausdorff computation.
## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #4750

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
@@ -29,7 +29,7 @@
 #include <CGAL/spatial_sort.h>
 
 #ifdef CGAL_LINKED_WITH_TBB
-#include <tbb/parallel_for.h>
+#include "tbb/parallel_reduce.h"
 #include <tbb/blocked_range.h>
 #include <atomic>
 #endif // CGAL_LINKED_WITH_TBB
@@ -81,21 +81,27 @@ struct Distance_computation{
   const AABB_tree& tree;
   const PointRange& sample_points;
   Point_3 initial_hint;
-  std::atomic<double>* distance;
-
+  double distance;
+  //constro
   Distance_computation(
           const AABB_tree& tree,
           const Point_3& p,
-          const PointRange& sample_points,
-          std::atomic<double>* d)
+          const PointRange& sample_points)
     : tree(tree)
     , sample_points(sample_points)
     , initial_hint(p)
-    , distance(d)
+    , distance(-1)
+  {}
+  //split constro
+  Distance_computation(Distance_computation& s, tbb::split )
+    : tree(s.tree)
+    , sample_points(s.sample_points)
+    , initial_hint(s.initial_hint)
+    , distance(-1)
   {}
 
   void
-  operator()(const tbb::blocked_range<std::size_t>& range) const
+  operator()(const tbb::blocked_range<std::size_t>& range)
   {
     Point_3 hint = initial_hint;
     double hdist = 0;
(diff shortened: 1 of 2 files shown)


maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-167.shtml

sloriot:
Note that nobody reviewed this PR, and AFAIS it is not matching CGAL standards.

lrineau:
> Note that nobody reviewed this PR, and AFAIS it is not matching CGAL standards.

Please do the review now. @maxGimeno will fix in another PR.

sloriot:
You reported the issue and I did not yet understand what the problem was so I'm a poor reviewer for now.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -29,7 +29,7 @@
 #include <CGAL/spatial_sort.h>
 
 #ifdef CGAL_LINKED_WITH_TBB
-#include <tbb/parallel_for.h>
+#include "tbb/parallel_reduce.h"
```
afabri, line 32: Can anybody explain why ""  instead of <>

afabri: Or do I misunderstand in which direction things got changed?

lrineau:
Except for `"..."` instead of `<...>`, I have not seen any issue with this PR. You said:
> AFAIS it is not matching CGAL standards.

so please be more specific, so that Maxime can fix that.

maxGimeno:
the //constro comments, apparently constro is a french short for constructor, I didn't know

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
```diff
@@ -107,15 +113,11 @@ struct Distance_computation{
...
-    while( current_value < hdist )
-    {
-      if(distance->compare_exchange_weak(current_value, hdist))
-        current_value = hdist;
-    }
+    if(hdist > distance)
+      distance = hdist;
   }
+
+  void join( Distance_computation& rhs ) {distance = std::max(rhs.distance, distance); }
```
sloriot, line 120: max is not protected, I guess the testsuite with the check is not using tbb

sloriot:
I only looked at the first diff and it was breaking our rules, I went through the whole patch and the rest seems correct except for the max that is not surrender by ().

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/distance.h
maxGimeno: Indeed.

maxGimeno:
The fixes are made in #4762

PR #4757, opened 2020-06-01
Added off2ply example to Polyhedron_IO.

thebluespecs:
## Summary of Changes

Added an Examples in the Polyhedron_IO package which reads an .OFF and produces a .PLY file with options to have a binary output.

## Release Management

* Affected package(s): Polyhedron_IO
* Issue(s) solved (if any): 


afabri:
@MaelRL  I am wondering if the   `precision` of the stream must be set when writing in ascii?  I didn't check how the other `X2Y.cpp`  do and if you deal with this question in the IO PR.

thebluespecs:
- @afabri There are three formats ASCII, binary and pretty. I didn’t know if ASCII was set as default, hence I set it regardless. Upon checking, I found that IO::ASCII is the default set enum type.

PR #4778, opened 2020-06-10
Fix complexity in case of append

sloriot:
If we append one triangle, the check on for non-manifold vertices was done on the whole mesh.

Not relevant for 5.0-branch


maxGimeno:

PR #4793, opened 2020-06-17
Testsuite rework the timeout of run_testsuite

lrineau:
## Summary of Changes

Use `timeout` if available, instead of a system of sleep/kill

The GNU coreutils software [timeout](https://man7.org/linux/man-pages/man1/timeout.1.html) is a perfect replacement for our buggy function `wait_for_process`.

This patch uses it when available. On Linux, brew, Cygwin, it is available in the `coreutils` package.

## Release Management

* Affected package(s): Testsuite, and thus all


maxGimeno:

PR #4798, opened 2020-06-19
Fix unknown behavior that changed in Visual 2019

### LINKED ISSUE #4796: Mesh_3 parallel - regression
##  Issue Details

I am meshing an implicit domain with Mesh_3 parallel, and facing this assertion :
```CGAL error: assertion violation!
Expression : this->is_facet_on_surface(facet)
File       : C:\Users\Jane\Documents\CGAL\cgal-dev\Mesh_3\include\CGAL/Mesh_3/Refine_facets_3.h
```

Since this crash happens with CGAL-5.0.2 but not with CGAL-5.0.1, I bisected and found that commit 
85712ba28c63b708f106517a78229665d9a3ef23 seems to be responsible for that crash. (Indeed there is no assertion with the previous commit).

## Source Code

@maxGimeno @lrineau I am sending it to you directly (it's simple).

## Environment

* Compiler: vs2019
* Release or debug mode: both
* Specific flags used (if any): `CGAL_LINKED_WITH_TBB` `CGAL_CONCURRENT_MESH_3`
* CGAL version: 5.0.2
* Boost version: 1.67.0
* Other libraries versions if used (Eigen, TBB, etc.): Eigen 3.7.0, TBB 2019 Update 3

maxGimeno:
## Summary of Changes
The Concurrent_compact_container uses some undefined behavior to  maintain time_stamps and some other things, including a counter that is not initialized during construction. 
In Vicual Studio 2019, the behavior of the std::atomic constructor changed, breaking this counter, leading to bugs at least in Mesh_3. This PR fixes it.
## Release Management

* Affected package(s):STL_Extension
* Issue(s) solved (if any): fix #4796

### DIFF
--- a/STL_Extension/include/CGAL/Concurrent_compact_container.h
+++ b/STL_Extension/include/CGAL/Concurrent_compact_container.h
@@ -341,17 +341,25 @@ public:
   iterator
   emplace(const Args&... args)
   {
+    typedef CCC_internal::Erase_counter_strategy<
+      CCC_internal::has_increment_erase_counter<T>::value> EraseCounterStrategy;
     FreeList * fl = get_free_list();
     pointer ret = init_insert(fl);
+    auto erase_counter = EraseCounterStrategy::erase_counter(*ret);;
     new (ret) value_type(args...);
+    EraseCounterStrategy::set_erase_counter(*ret, erase_counter);
     return finalize_insert(ret, fl);
   }
 
   iterator insert(const T &t)
   {
+    typedef CCC_internal::Erase_counter_strategy<
+      CCC_internal::has_increment_erase_counter<T>::value> EraseCounterStrategy;
     FreeList * fl = get_free_list();
     pointer ret = init_insert(fl);
+    auto erase_counter = EraseCounterStrategy::erase_counter(*ret);;
     std::allocator_traits<allocator_type>::construct(m_alloc, ret, t);
+    EraseCounterStrategy::set_erase_counter(*ret, erase_counter);
     return finalize_insert(ret, fl);
   }
 


STL_Extension/include/CGAL/Concurrent_compact_container.h
```diff
@@ -341,17 +341,25 @@ class Concurrent_compact_container
   iterator
   emplace(const Args&... args)
   {
+    typedef CCC_internal::Erase_counter_strategy<
+      CCC_internal::has_increment_erase_counter<T>::value> EraseCounterStrategy;
     FreeList * fl = get_free_list();
     pointer ret = init_insert(fl);
+    auto erase_counter = EraseCounterStrategy::erase_counter(*ret);;
```
lrineau, line 348: Note that we still use an undefined behavior, because:
  - here the code reads the erase counter from `*ret`, that is not yet created,
  - in `allocator_new_block`, the erase counter of uninitialized objects is assigned.

maxGimeno: Should we try to change that now, instead of waiting until it breaks and spend another 2 days trying to understand why ?

lrineau: As we use bit-squatting for the type `USED`/`FREE` of a cell of the container, we have to use undefined-behavior, no matter what we do. That is a property of that container: it uses undefined-behavior for efficiency. That is worth a few days of hard work from time to time.

maxGimeno: Ok, I was right to ask then :)

maxGimeno:

PR #4828, opened 2020-07-02
Rewrite circumcenter computations in NewKernel_d

mglisse:
## Summary of Changes

The linear system used to compute the circumcenter of sets of fewer than d+1 points is not so good, so I am replacing it with one that works better: faster, and numerically more stable (fewer filter failures). I am also adding a special case when the number of points is small.
The formula is not exactly the same as used in Kernel_23, because I wanted it to work independently of the ambient dimension.
And I did the same for the weighted case.

## Release Management

* Affected package(s): NewKernel_d, Number_types

### DIFF
--- a/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_functors.h
@@ -224,6 +224,7 @@ template<class R_> struct Scalar_product {
 };
 
 template<class R_> struct Squared_distance_to_origin_stored {
+        // What about weighted points, should they store sdo-w?
         CGAL_FUNCTOR_INIT_IGNORE(Squared_distance_to_origin_stored)
         typedef R_ R;
         typedef typename R::LA_vector LA;
--- a/NewKernel_d/include/CGAL/NewKernel_d/Types/Weighted_point.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Types/Weighted_point.h
@@ -149,6 +149,7 @@ template <class R_> struct Power_center : Store_kernel<R_> {
   typedef typename Get_type<R_, Weighted_point_tag>::type        WPoint;
   typedef WPoint result_type;
   typedef typename Get_type<R_, Point_tag>::type        Point;
+  typedef typename Get_type<R_, Vector_tag>::type       Vector;
   typedef typename Get_type<R_, FT_tag>::type FT;
   template <class Iter>
   result_type operator()(Iter f, Iter e)const{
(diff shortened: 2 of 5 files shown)


sloriot:
Error: `/mnt/testsuite/include/CGAL/NewKernel_d/function_objects_cartesian.h:690:12: error: ‘vector’ is not a member of ‘std’`

sloriot:
There are other errors but it is not clear it they are triggered by the first one or not.
See [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-217/NewKernel_d/TestReport_lrineau_ArchLinux-CXX14.gz)

mglisse:
I added the missing include. For the other error, you are testing both this one and #4838 together, but as I commented there, they conflict (the change of name of the functors will need to be propagated). I did it separately because I expected this one would be merged long before the other one got reviewed, and didn't want to pollute the small feature with unnecessary changes.
I'll be unavailable for about 10 days starting tomorrow, I would suggest you keep testing this PR (and merge it if there is no other problem) but postpone #4838 until I come back.

maxGimeno:
There is an error on the cxx20 platform : 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-222/NewKernel_d/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz
It seems to be Eigen related...

maxGimeno:
>I think it is acceptable to require a very very recent version of Eigen.

I agree with you. @lrineau, shouldn't we configure this platform to use Eigen master, then ?

lrineau:
> > I think it is acceptable to require a very very recent version of Eigen.
> 
> I agree with you. @lrineau, shouldn't we configure this platform to use Eigen master, then ?

Yes. I have filled https://github.com/CGAL/cgal-testsuite-dockerfiles/issues/108 as a reminder.

mglisse:
There is still a problem with Eigen versions from before April 2018, which use `NumTraits<RealScalar>::highest()`. I added it for Interval_nt, but stupidly stopped there, when obviously it is also needed for the exact type. Except that there is no meaningful value I can give it, the use in old Eigen is such that we would want a value v for which 1/v is 0...
I guess I should check the version of Eigen, and for older ones use a slower method than ldlt (probably lu).

mglisse:
For older eigen, I am now using a slower method that shouldn't have this issue, so it should be good now.

mglisse:
> ERROR:      execution___of__Epick_d in 0.773397 s : NUMERICAL

Any idea what that means? `grep` doesn't show NUMERICAL anywhere in CGAL. I'll ignore this column in the testsuite.

maxGimeno:
My guess is that ctest gathers a result value that is neither 1 nor 0. It is possible as the main() function of this test doesn't return anything. I wonder why there is no warning about that...

mglisse:
> It is possible as the main() function of this test doesn't return anything. I wonder why there is no warning about that...

There is an implicit `return 0;` at the end of main in C++.

maxGimeno:
> There is an implicit return 0; at the end of main in C++.

Ok, I didn't know. I grepped the error in the results of the testsuite and it never appears in master, so it wasn't that anyways :(

mglisse:
Ah, if that's ctest, then it may be how it reports SIGFPE. So the test crashes on that one platform...

lrineau:
> Ah, if that's ctest, then it may be how it reports SIGFPE. So the test crashes on that one platform...

I confirm. That is what I have read in the source code of CTest.

maxGimeno:
I found where the NUMERCIAL comes from : it is the ExitValue written in the xml resulting of `ctest -T Test`.

mglisse:
I am confused about how the internal release was created. I pushed a commit yesterday around 3pm, the internal release was created around 9pm, it includes this branch, but it does not include this latest commit. This is confirmed by the text
> 44f9593 Merge pull request #4828 from mglisse/circumcenter-lowdim
  parents: 1b6610b 334f8ae

(334f8ae is the previous commit)
I thought the process was automatic for PRs with the right labels, but maybe not.

I'll cross my fingers that the commit is picked up next time...

maxGimeno:
Tonight's testsuite was basically the same as yesterday's, wiht only 4 or 5 branches up to date. It's my fault, I forgot to reset it on master.

mglisse:
No problem, thanks. I really thought it was automatic...

maxGimeno:
it is not, someone runs a script every evening before leaving work, and it 's the script that automatically gathers the PRs from the labels.

maxGimeno:

PR #4832, opened 2020-07-06
[PSP] Fix neighbor query fallback

sgiraudot:
## Summary of Changes

When performing a neighbor query using a range with fixed radius, problems may occur if no point was found in the radius, which lead me to add a fallback mode using k=3 if no point was found (which seemed like the minimum number of points to fit a plane for example, provided they are not collinear).

While debugging the bindings of CGAL, @maxGimeno encountered a bug which made me realize that these (internal) parameter should vary depending on the algorithm: in that case, the algorithms that use the _Jet Fitting_ package have a precondition on the minimum number of neighbors, so I fixed the bug by using this value as the fallback value.

The neighbor query class did not exist in CGAL 5.0 so this is a fix for master only.

## Release Management


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-192.shtml

PR #4838, opened 2020-07-07
Document Power_center_d, Power_distance_d (renamed)

mglisse:
## Summary of Changes

Document 2 existing functors of `Ep[ie]ck_d`. To try and make reviewers happy, I renamed `Power_distance_d` to `Compute_power_product_d` so it matches the names in `Kernel_23`. `Power_center_d` does not seem to have an equivalent in 2D/3D, where you can get the point, but you have to compute the radius/weight separately afterwards (computing both together can save a few operations).

## Release Management

* Affected package(s): `NewKernel_d`
* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/More_Weighted_point_d_functors
* Link to compiled documentation (obligatory for small feature) [here](https://geometrica.saclay.inria.fr/team/Marc.Glisse/tmp/cgal-power/Kernel_d/structCGAL_1_1Epick__d.html)

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -15,6 +15,14 @@ Release date: December 2020
 -   Added the function `CGAL::halfspace_intersection_interior_point_3()` that can be used to retrieve
     the point that is the most interior a convex closed volume defined by the intersection of a set of halfspaces.
 
+### [dD Geometry Kernel](https://doc.cgal.org/5.2/Manual/packages.html#PkgKernelD)
+
+-   The kernels [`Epick_d`](https://doc.cgal.org/5.2/Kernel_d/structCGAL_1_1Epick__d.html)
+    and [`Epeck_d`](https://doc.cgal.org/5.2/Kernel_d/structCGAL_1_1Epeck__d.html) gain two new functors:
+    [`Compute_power_product_d`](https://doc.cgal.org/5.2/Kernel_d/classCGAL_1_1Epeck__d_1_1Compute__power__product__d.html)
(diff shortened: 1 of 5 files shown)


lrineau:
Hi Marc, there seems to be a compilation error at the moment. You can see that on Travis: https://travis-ci.com/github/CGAL/cgal/jobs/358600978#L2891.

mglisse:
Uh, fixed. I thought I had tested...
By the way, this branch and #4828 conflict (easy to fix, just rename the new places where the old name appears) and should not be tested at the same time. Priority should be given to #4828 since it doesn't require a small feature.

afabri:
Isn't the concept missing?

mglisse:
> Isn't the concept missing?

The functors documented in Epick_d are precisely those that do not have a matching concept in Kernel_d. If I added them to the Kernel_d concept, I would need either to add an implementation to Cartesian_d and Homogeneous_d or to document that those kernels do not implement the Kernel_d concept anymore. Or I could create a new kernel concept for the new kernels, which might help reduce doc duplication, maybe I should do that if I add a 3rd flavor in addition to Epick_d and Epeck_d, but that would be too much work for just a couple new functors, and if I were to create a new kernel concept, I'd be rather reluctant to keep it similar to Kernel_d, I'd be tempted to make it look more like the internals of Epick_d, which would be even more work...

mglisse:
Now that #4828 has tested fine, I am going to merge it in this branch so I can resolve the conflict and this branch can be tested as well.

mglisse:
> Now that #4828 has tested fine, I am going to merge it in this branch so I can resolve the conflict and this branch can be tested as well.

Done. Maybe rebase would have been cleaner.

maxGimeno:

PR #4851, opened 2020-07-20
Arrangement 2: Enhance CCB merging with union set

sgiraudot:
## Summary of Changes

When merging 2 inner CCBs, all halfedges of one of the CCB are traversed again and updated, which can lead to an explosion of complexity, for example when computing the arrangement of a very large holes with many concavities.

This PR replaces this merging algorithm by the following:
- a `Union_find` structure of inner CCBs is used
- everytime inner CCBs are merged, we use `unify_sets()` on the union find
- the union find is used to access the correct inner CCB of each halfedge (to always fall back to the currently active CCB in the union find)
- after the sweep, the CCB are cleaned up using the union find to recover the correct result where each halfedge points to a valid inner CCB

In practice, this leads to accelerations up to 40x (1sec instead of 40sec) in the case similar to the one I described (large holes with many concavities). For more common data sets, I observed an acceleration of 30%.

As this enhancement comes at the cost of additional memory usage (one union find structure + 1 handle per inner CCB), I added the possibility to deactivate it with the macro `CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB`.

## Release Management

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -34,6 +34,9 @@
 #include <CGAL/Arrangement_2/Arrangement_2_iterators.h>
 #include <CGAL/assertions.h>
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+#include <CGAL/Union_find.h>
+#endif
 
 namespace CGAL {
 
@@ -762,10 +765,18 @@ public:
   typedef Arr_face<V,H,F>                    Face;
   typedef typename Face::Inner_ccb_iterator  Inner_ccb_iterator;
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  typedef CGAL::Union_find<Self*>            Uf;
+  typedef typename Uf::handle                Uf_handle;
+#endif
+
 private:
   Face* p_f;                  // The face the contains the CCB in its interior.
   Inner_ccb_iterator  iter;   // The inner CCB identifier.
   bool iter_is_not_singular;
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  Uf_handle ufh;
+#endif
 
 public:
   /*! Default constructor. */
@@ -814,6 +825,21 @@ public:
     iter = it;
     iter_is_not_singular = true;
   }
+
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  /*! Get the UF handle associated to the inner CCB */
+  Uf_handle uf_handle() const
+  {
+    CGAL_assertion (ufh != Uf_handle());
+    return ufh;
+  }
+
+  /*! Set the UF handle associated to the inner CCB */
+  void set_uf_handle(Uf_handle ufh) const
+  {
+    this->ufh = ufh;
+  }
+#endif
 };
 
 /*! \class
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -916,6 +916,11 @@ public:
 
   typedef Inner_ccb                   Hole;
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  typedef typename Inner_ccb::Uf        Inner_ccb_uf;
+  typedef typename Inner_ccb_uf::handle Inner_ccb_uf_handle;
+#endif
+
 protected:
   // The vetices, halfedges and faces are stored in three in-place lists.
   typedef In_place_list<Vertex, false>           Vertex_list;
@@ -956,6 +961,10 @@ protected:
   Inner_ccb_allocator in_ccb_alloc;         // An allocator for inner CCBs.
   Iso_vert_allocator  iso_vert_alloc;       // Allocator for isolated vertices.
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  Inner_ccb_uf        inner_ccb_uf;         // efficient CCB merging
+#endif
+
 public:
   // Definitions of iterators.
   typedef typename Vertex_list::iterator              Vertex_iterator;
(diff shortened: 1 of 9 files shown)


Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -34,6 +34,9 @@
 #include <CGAL/Arrangement_2/Arrangement_2_iterators.h>
 #include <CGAL/assertions.h>
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
```
lrineau, line 37: Do you really want to keep that macro `CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB` in CGAL?

If you do, you should add a test that actually defines it.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -762,10 +765,18 @@ class Arr_inner_ccb : public In_place_list_base<Arr_inner_ccb<V,H,F> >
   typedef Arr_face<V,H,F>                    Face;
   typedef typename Face::Inner_ccb_iterator  Inner_ccb_iterator;
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  typedef CGAL::Union_find<Self*>            Uf;
+  typedef typename Uf::handle                Uf_handle;
```
sloriot, line 770: you probably want to have a less generic name for the typedefs and variables

Arrangement_on_surface_2/include/CGAL/Surface_sweep_2/Arr_construction_ss_visitor.h
```diff
@@ -142,6 +142,9 @@ class Arr_construction_ss_visitor :
   /* A notification issued before the sweep process starts. */
   inline void before_sweep();
 
+  /*! A notification issued after the sweep process ends. */
+  inline void after_sweep();
```
sloriot, line 146: why is this needed?

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -930,13 +961,18 @@ class Arr_dcel_base {
   Inner_ccb_allocator in_ccb_alloc;         // An allocator for inner CCBs.
   Iso_vert_allocator  iso_vert_alloc;       // Allocator for isolated vertices.
 
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  Inner_ccb_uf        inner_ccb_uf;         // efficient CCB merging
```
sloriot, line 965: you probably want to clear the container when you're done with the construction of the arrangement.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -762,10 +765,18 @@ class Arr_inner_ccb : public In_place_list_base<Arr_inner_ccb<V,H,F> >
...
+  typedef CGAL::Union_find<Self*>            Uf;
+  typedef typename Uf::handle                Uf_handle;
+#endif
+
 private:
   Face* p_f;                  // The face the contains the CCB in its interior.
   Inner_ccb_iterator  iter;   // The inner CCB identifier.
   bool iter_is_not_singular;
+#ifndef CGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB
+  Uf_handle ufh;
```
sloriot, line 778: I was trying to think about a way to avoid storing this extra handles but then I realized that it is only at the CCB level so it is probably useless trying to optimize that. If we want to optimise it, one possible way is to replace `p_f` by `std::union F_or_id { Face* p_f; std::size_t id };` and use the id to also store the face pointer in a vector while sweeping.

lrineau:
You have to declare the new dependency to `Union_find`: https://travis-ci.com/github/CGAL/cgal/jobs/362900640#L7733

Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
```diff
@@ -659,6 +659,17 @@ function(test_construction_segments)
...
+#---------------------------------------------------------------------#
+# construction with segments, without Union Find
+#---------------------------------------------------------------------#
+function(test_construction_segments_no_union_find)
+  set(nt ${CGAL_GMPQ_NT})
+  set(kernel ${CARTESIAN_KERNEL})
+  set(geom_traits ${SEGMENT_GEOM_TRAITS})
+  set(flags "-DTEST_NT=${nt} -DTEST_KERNEL=${kernel} -DTEST_GEOM_TRAITS=${geom_traits} -DCGAL_ARRANGEMENT_DISABLE_UNION_SET_CCB")
+  compile_and_run_with_flags( test_construction segments_no_union_find "${flags}")
+endfunction()
```
lrineau, line 671: 1. You add a function, but that does not add any test if you do not all it.
2. The Arr_2 test suite is complex, and it has also a non-CTest version, in `cgal_test_with_cmake`.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
sgiraudot: Such as?

Arrangement_on_surface_2/include/CGAL/Surface_sweep_2/Arr_construction_ss_visitor.h
sgiraudot: To call [`clean_inner_ccb()`](https://github.com/CGAL/cgal/pull/4851/files/1be446983801c00884d785ff9acc16fe5f913f75#diff-c697bc93f9992191b6ca431c9ee36d37R280).

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
sgiraudot: Indeed, I'll change that.

sgiraudot:
After talking it through with Sebastien, we decided to externalize as much as possible this Union find behavior so that it's only used by the surface sweep (to avoid creating invalid arrangements for users who manage it manually). I'll work on that, the code is likely to change quite a lot, so don't waste time reviewing right now :)

sgiraudot:
Replaced by https://github.com/CGAL/cgal/pull/4875.

PR #4852, opened 2020-07-20
Add Octree Package

JacksonCampolattaro:
## Summary of Changes

[New feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Octree) adding an Octree package to CGAL

This PR adds a new package `Octree`. This is a tree data structure which subdivides 3d space, similar to kD_tree. The package will replace the Octree used in Shape Detection's RANSAC algorithm. When used with a 2d point type, the Octree becomes a Quadtree. See [the GSoC wiki page](https://cgal.geometryfactory.com/CGAL/Members/wiki/GSoC2020/Octree) for more information.

_Note: this is a Google Summer of Code Project_

## Release Management

* Affected package(s): Octree
* Issue(s) solved (if any): _none_
* Feature/Small Feature (if any): [Features/Octree](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Octree)
* Link to compiled documentation (obligatory for small feature): [Manual](https://cgal.github.io/4852/v6/Orthtree/index.html)
* License and copyright ownership: GPL v3, GeometryFactory

## CHANGES.md

```markdown
### [Quadtree, Octree and Orthtree](https://doc.cgal.org/5.3/Manual/packages.html#PkgOrthree) (new package)

-   This package implements a tree data structure in which each node
    encloses a hypercubic section of space and each non-leave node has
    hypercubic children whose edge lengths are half its edge
    length. Such a data structure is known as a quadtree in 2D, an
    octree in 3D and is generalized as an "orthtree" in higher
    dimensions.
```

## TODO
* [x] Replace RANSAC octree
* [x] Complete documentation
* [x] Feature page
* [x] Remove large data file used for benchmarking
* [x] Add Quadtree functionality
* [x] @sloriot: check branch size


sloriot:
Will you replace the RANSAC octree in this PR? If not why?

JacksonCampolattaro:
I was actually wondering about that myself; the code specific to RANSAC is all going to be done in that directory, because we don't want to pollute the Octree package with functionality so specialized to one purpose. Does it make sense to do that in a separate branch/PR? 

Thank you for the contributions, by the way!

sloriot:
Well, AFAIU, RANSAC should be a user of the Octree tree. So the internal Octree in RANSAC should go away and the public API of the new Octree should enable you to update the RANSAC code to get the same thing done. Am I missing something?

JacksonCampolattaro:
That's pretty much it, the internal octree will be replaced by the new one. I might even start by turning it into a wrapper for the new octree as a first step, so that I can test that everything works without making changes in every file where the old octree was used.

JacksonCampolattaro:
Does RANSAC being dependent on the new octree mean that it's best to combine it into the same PR? If I were replacing the internal octree after I knew the new one was completely finished I would definitely make it separate, but that's not necessarily the case here because I'll be doing some of the work in parallel.

sloriot:
I was mentioning that because it might help refine the API and running the testsuite of RANSAC might help find issues. I don't want to change the plan you have with @sgiraudot, I was mostly curious (while also being obsessed with removing duplicated code :) )

JacksonCampolattaro:
That's a good point! Perhaps the new octree shouldn't be merged without first using it to replace the old one. RANSAC's test suite will definitely make for a good integration test, and the way it's used there will also influence a lot of my API decisions. Should I update the PR description to mention this?

Shape_detection/include/CGAL/Shape_detection/Efficient_RANSAC/Octree.h
```diff
@@ -26,699 +26,706 @@
...
+
+//      struct Cell {
+//        std::size_t first, last;
+//        Cell *child[8];
+//        Point_3 center;
+//        std::size_t level;
+//
+//        Cell(std::size_t first, std::size_t last, Point_3 center, std::size_t level)
+//          : first(first), last(last), center(center), level(level) {
+//            memset(child, 0, sizeof(Cell *) * 8);
```
lrineau, line 181: Please do not comment lines to "remove" code. Just remove the lines. We have the Git history to retrieve what has disappeared.

JacksonCampolattaro: Thanks for the advice! After some experimenting, it looks like turning the RANSAC tree into a wrapper won't be a workable intermediate step after all. It exposes its member variables, and there's a lot of code that interacts with them directly.

I have my meeting with Simon tomorrow, and we should be able to find a better way forward! For now I removed the old octree entirely, and I'm spending some time trying to get an understanding for the way it was used.

JacksonCampolattaro:
After some experimenting, it looks like turning the RANSAC tree into a wrapper won't be a workable intermediate step after all. It exposes its member variables, and there's a lot of code that interacts with them directly.

I have my meeting with Simon tomorrow, and we should be able to find a better way forward! For now I removed the old octree entirely, and I'm spending some time trying to get an understanding for the way it was used.

JacksonCampolattaro:
We decided to start by factoring out RANSAC-specific functions from RANSAC's octree, until it's nearly as simple as the new one. `score` and `fullScore` will both be free functions that take a reference to an octree, and they'll be moved to Efficient_RANSAC.h. Once this is done, the tree should be much easier to replace!

Octree/include/CGAL/Octree/IO.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include <CGAL/Octree.h>
+#include <CGAL/Octree/Node.h>
+#include <CGAL/Octree/Walker_criterion.h>
+
+#include <iostream>
+#include <ostream>
+#include "Node.h"
+
+using std::ostream;
```
afabri, line 25: Is that not a mistake, because it globally imports?

Octree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,646 @@
...
+namespace CGAL {
+
+namespace Octree {
+
+/*!
+ * \ingroup PkgOctreeClasses
+ *
+ * \brief Class Octree is a data structure for efficient computations in 3D space.
+ *
+ * \details It builds a heirarchy of nodes which subdivide the space based on a collection of points.
```
afabri, line 57: heirarchy -> hierarchy 
subdivide -> subdivides

Octree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,646 @@
...
+    Point bbox_centroid = midpoint(bbox.min(), bbox.max());
+
+    // scale bounding box to add padding
+    bbox = bbox.transform(Aff_transformation_3<Kernel>(SCALING, enlarge_ratio));
+
+    // Convert the bounding box into a cube
+    FT x_len = bbox.xmax() - bbox.xmin();
+    FT y_len = bbox.ymax() - bbox.ymin();
+    FT z_len = bbox.zmax() - bbox.zmin();
+    FT max_len = std::max({x_len, y_len, z_len});
```
afabri, line 182: `std::max` -> `(std::max)`   to avoid problems with VC++ .   Is the same for `min`

Octree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,646 @@
...
+    refine(Split_criterion::Max_depth_or_bucket_size(max_depth, bucket_size));
+  }
+
+  /// @}
+
+  /// \name Accessors
+  /// @{
+
+  /*!
+   * \brief Provides read-only access to the root node, and by extension the rest of the tree
```
afabri, line 273: It's a non-written convention to start with lower case so that the function signature and the description together form a sentence.

Octree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,646 @@
...
+
+    // Return the result
+    return *node_for_point;
+  }
+
+  /*!
+   * \brief Find the bounding box of a node
+   *
+   * Creates a cubic region representing a node.
+   * The size of the region is dependent on the node's depth in the tree.
```
afabri, line 352: is dependent -> depends

afabri:
Hello, I was just curious looked into the code and left some comments.   I am wondering if Traversal would not be a better choice than Walker.

Octree/include/CGAL/Octree/Node.h
```diff
@@ -0,0 +1,346 @@
...
+#include <boost/range/iterator_range.hpp>
+
+#include <array>
+#include <memory>
+#include <bitset>
+#include <cassert>
+
+namespace CGAL {
+namespace Octree {
+namespace Node {
```
afabri, line 26: Do we really want namespaces for Node and Walker?   `namespace Node` only contains the `class Node`.

Octree/include/CGAL/Octree/Node.h
```diff
@@ -0,0 +1,346 @@
...
+
+public:
+
+  /// \name Types
+  /// @{
+
+  /*!
+   * \brief Array for containing the children of this node
+   */
+  typedef std::array<Node<Point_index>, 8> Child_list;
```
afabri, line 48: What about children?

Octree/include/CGAL/Octree/Node.h
```diff
@@ -0,0 +1,346 @@
...
+   * \brief Read-only access to the content held by this node
+   * \return a read-only reference to the collection of point indices
+   */
+  const Point_range &points() const { return m_points; }
+
+  /*!
+   * \brief Check whether this node contains any points
+   * \return if this node contains no points
+   */
+  bool is_empty() const {
```
afabri, line 284: `empty()`, that is no is_  prefix

Octree/include/CGAL/Octree/Node.h
```diff
@@ -0,0 +1,346 @@
...
+    return m_points.empty();
+  }
+
+  /*!
+   * \brief Count the points contained by this node
+   * \return the number of points this node owns
+   */
+  std::size_t number_of_points() const {
```
afabri, line 292: Maybe just `size()`?

Octree/include/CGAL/Octree/IO.h
JacksonCampolattaro: Good catch!
https://github.com/JacksonCampolattaro/cgal/blob/7674b9f296734369e4e00ef38b11ae8af680d3c5/Octree/include/CGAL/Octree/IO.h#L15-L23
It should be resolved now, I still need to clean up my imports in general.

Octree/include/CGAL/Octree.h
JacksonCampolattaro: I fixed the typo in "hierarchy", but I think "subdivide" might be correct. I guess it could be interpreted either way:
"the (singular) hierarchy of nodes subdivides the space" or "a hierarchy consists of (plural) nodes which subdivide the space."

Octree/include/CGAL/Octree.h
JacksonCampolattaro: This should be fixed now, I think I only use `std::max` in one place. I'm surprised windows.h would contain such an obvious naming conflict!

Octree/include/CGAL/Octree.h
JacksonCampolattaro: Thanks for the tip!

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: I'm a bit uncertain about this too, I agree that the redundant namespaces are a bit ugly.

I originally added `namespace Node` in order to be able to typedef it inside the Octree without introducing a naming conflict:
https://github.com/JacksonCampolattaro/cgal/blob/d290ddcfe792db0b26a462de592054181666134b/Octree/include/CGAL/Octree.h#L96
Is there a better way to avoid that kind of verbosity?

I added the Walker namespace when I noticed that I had several functors with the suffix "walker" on their name. E.g. `Preorder_walker` became `Walker::Preorder`. I think a namespace is appropriate here because it makes it clear that all the structs contained are different versions of the same concept.

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: I'm not sure I understand the question, do I need to make that comment more descriptive?

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: Good point, that definitely makes it more consistent with other parts of CGAL. Does it make sense to change `is_leaf` and `is_root` too?

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: I used `number_of_points` to match [Point_set_3](https://doc.cgal.org/latest/Point_set_3/classCGAL_1_1Point__set__3.html)'s equivalent function (the same reason I used `is_empty`), but it occurs to me that I probably don't need to be so explicit since the octree doesn't have "removed" points.

JacksonCampolattaro:
@afabri Thank you for all the input! 

Replacing "Walk" with "Traverse" might be more appropriate, since it seems more in line with the most commonly used terminology.

Honestly a big reason for the syntax I chose was the grammar: "Walker" is a good description of what functors like `Walker::Preorder` do, and "Traverser" would be as well, but apparently that's not a real word. "Traversal" might communicate the purpose just as well, but it feels less descriptive of what the functor being passed actually does. `Traversal::Preorder` sounds like it could be part of an enum class, where "Walker" or "Traverser" both suggest that the user is passing the thing that does the traversing. I think that difference could matter, because I expect the user to define their own walker functors.

I'll make sure to bring this up with @sgiraudot !

Octree/include/CGAL/Octree/Node.h
afabri: I suggested to replace the word `Child_list` with `Children`.

Octree/include/CGAL/Octree/Node.h
sgiraudot: I agree with Andreas that the namespace Node is not a good idea. For the typedef, I think you can trick it by doing `typedef CGAL::Octree::Node<blabla> Node;` but I'm not sure. Otherwise, just use another name (I agree that it's less satisfying but it's better than adding a namespace just for the sake of typedefing later).

For Walker, as we already discussed, I agree with Jackson, this namespace have / will have several classes and serves for a clear naming `Walker::Preorder`, `Walker::Postorder`, etc.

Octree/include/CGAL/Octree/Node.h
sgiraudot: I agree with Andreas, especially because `list` can be misleading (you are not using a `std::list`).

Octree/include/CGAL/Octree/Node.h
sgiraudot: If you search for `is_` in CGAL, you will find many occurrences, so I think you can keep `is_leaf()` and `is_root()`.

Don't use `is_` if the function obviously returns a boolean (as is the case for `empty()`, which by the way is also more consistent with STL containers).

Functions `leaf()` or `root()` may sound like they would return _the_ leaf object or _the_ root object, and not a boolean, so keeping `is_` makes sense there.

Octree/include/CGAL/Octree/Node.h
sgiraudot: A small subtlety here is that `Point_set_3` models the range concept (which means [all the functions of the range are available while not directly documented](https://doc.cgal.org/latest/Circulator/classRange.html)) and was also written based on `Surface_mesh` which explains why:

- `empty()` and `size()` are available (model of Range)
- `is_empty()` and `number_of_points()` are available (consistency with Surface Mesh) and basically do the same thing

In your case, I think it would actually be a good idea to make Node a model of Range too. You probably already have all the necessary functions (again, see the link above): you have `begin() / end()` for the points, you have the size, the empty test, etc.

And in that case, I would remove `number_of_points()`and just use `size()` (same for `empty()` and not `is_empty()`), as you don't really need consistency with `Point_set_3` or `Surface_mesh`.

Octree/include/CGAL/Octree/Node.h
afabri: "is_"  is fine.

Octree/include/CGAL/Octree/Node.h
afabri: In the AABB Tree package, we have a `Traversal_traits`.   So you might replace `Walker` with `Traversal`.

afabri:
> "Traversal" might communicate the purpose just as well, but it feels less descriptive of what the functor being passed actually does.

image

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: I just removed the `Node` namespace, @sgiraudot Thanks for the suggestion, the trick worked!

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: That makes a lot of sense! In that case, I think this can be resolved.

Octree/include/CGAL/Octree/Node.h
JacksonCampolattaro: Good idea! If I make the Node class satisfy the range concept, I should probably store a `m_begin` and `m_end` iterators rather than storing a `PointRange` like I'm doing now.

JacksonCampolattaro:
@afabri You're right, traversal is by far the more common terminology. 

@sgiraudot Do you think I should replace `Walker` with `Traversal`, or should it be `Traverser`? I think Traverser has a closer meaning to Walker, but somehow it doesn't sound right to me.

sgiraudot:
Traversal sounds good!

It may sound weird to you because you are used to a name that describes an object realizing a procedure (a walker), but here the name describes the procedure itself (a traversal) and it's totally fine.

JacksonCampolattaro:
Great! I'll switch it now.

Shape_detection/include/CGAL/Shape_detection/Efficient_RANSAC/Octree.h
```diff
@@ -84,6 +105,13 @@ class Indexed_octree : public Octree::Octree<typename Traits::Input_range, typen
   std::size_t maxLevel() const {
     return this->max_depth_reached();
   }
+
+  void createTree(double cluster_epsilon_for_max_level_recomputation = -1., std::size_t bucketSize = 2,
```
afabri, line 163: no CamelCase.

JacksonCampolattaro: Fixed! That was a carry-over from the original RANSAC octree.

Note: I intend for `Direct_octree` and `Indexed_octree` to be very temporary. They're acting as thin wrappers for the new octree, but as soon as I get the code to pass tests I'm hoping to refactor it so that it uses the octree more directly.

JacksonCampolattaro:
Will I need to post a link to the compiled documentation?

sloriot:
> Will I need to post a link to the compiled documentation?

Would be better for an internal review round. We can host it if you have no where to put it.

JacksonCampolattaro:
I have no problem hosting it! Is Github-pages okay?

sloriot:
if it works yes.

JacksonCampolattaro:
Simon has written a new feature which allows the Octree to become a Quadtree when used with a 2-dimensional point type, [here's the added code](https://github.com/JacksonCampolattaro/cgal/compare/gsoc2020-Octree-campolattaro...sgiraudot:gsoc2020-Octree-campolattaro). Should I wait until after the first round of review to incorporate the changes?

sloriot:
Go ahead :)

JacksonCampolattaro:
Done! Simon included an example and some documentation for the new behavior, but I'll keep my eye out for places I need to update the Octree's description (like the summary of this PR, for example :) ).

sloriot:
THanks! I'm supposed to do a review of the package. Is the doc up-to-date?

JacksonCampolattaro:
The compiled documentation hasn't been updated with the new functionality yet, I'll be able to do that later this week!

maxGimeno:

maxGimeno:
```
CMakeLists with incorrect project name
Orthtree/test/Orthtree/CMakeLists.txt
```

maxGimeno:
```
Orthtree/include/CGAL/Orthtree_traits_3.h
Orthtree/include/CGAL/Orthtree_traits_2.h
Orthtree/include/CGAL/Orthtree_traits_d.h
```

Shape_detection/test/Shape_detection/test_efficient_RANSAC_scene.cpp
```diff
@@ -11,6 +11,10 @@
 
 template <class K>
 bool test_scene(int argc, char** argv) {
+
+  // TODO: Remove this after testing
```
afabri, line 15: don't forget the todos

lrineau:
> The documentation is built. It will be available, after a few minutes, here : [cgal.github.io/4852/v4/Manual/index.html](https://cgal.github.io/4852/v4/Manual/index.html)

@maxGimeno I thought there was a check to avoid overwriting by error a given version (like here `v4`). Can you verify?

sgiraudot:
I can confirm that there is no check (overwriting does work).

maxGimeno:
I know, I already took note for later. 
I must say that I find it far more useful to be able to overwrite, as most users of the feature don't respect the rule of one round, one build.

lrineau:
> I know, I already took note for later.
> I must say that I find it far more useful to be able to overwrite, as most users of the feature don't respect the rule of one round, one build.

Actually, I would appreciate that we cannot overwrite:
  - it avoids overwrites by error,
  - it enforces the rules.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -172,6 +172,10 @@ class Orthtree
     with all Orthtree functionality, but any performance benefits are
     unlikely to be realized until `refine()` is called.
 
+    \warning the input point range is rearranged by the
```
sloriot, line 175: The kd-tree does copy the points right?

sgiraudot: Yes.

sloriot: Maybe make it even clearer that input are not copied, stressing it is also a difference with the kd-tree.

sloriot: Maybe even in the user manual as it is important.

sgiraudot:
I rebased the PR and fixed conflicts.

The feature was [proposed for acceptance](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Octree#Decision) on the 27th of January by @sloriot, can it be then be considered as accepted now?

sgiraudot:
Rebased.

maxGimeno:
dependencies modifications:
 Shape_detection: Orthtree is new
 Orthtree: Circulator have disappeared.
 Orthtree: Modular_arithmetic is new

maxGimeno:
errors and warnings: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-81.shtml#Orthtree

afabri:
@maxGimeno maybe you can take care of this one:  cl : Command line warning D9025 : overriding '/std:c++17' with '/std:c++14'

maxGimeno:
@afabri done

sgiraudot:
The errors/warnings should be fixed.

maxGimeno:
- Warning in doc : 
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.3-Ic-83/logs2/Documentation.log
 - Warnings : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-83/Orthtree/TestReport_lrineau_Ubuntu-Latest-CXX11.gz and https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-83/Orthtree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz
 - Runtime error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-83/Orthtree/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

sgiraudot:
- the warnings in the doc don't seem to come from this package, unless I've misunderstood something
- I fixed the warnings in the code
- I can't reproduce the runtime error: could someone using Windows try to reproduce it? (the failing platform is `MSVC2017-Debug-64bits_Cygwin-Windows10 `)

maxGimeno:
You are right for the doc, I found the culprit. Sorry.

maxGimeno:
Happy windows-specific min/max errors : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-86/Orthtree_Examples/TestReport_Blake_Windows_MSVC-2019-Community-Release-64bits.gz

sgiraudot:
I fixed it using the same macro used in `Bbox_2/3` as it was a case where I was the one defining `min()/max()` functions.

Orthtree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,50 @@
...
+ * These two types are exactly equivalent:
+ * - `Octree<GeomTraits, PointRange, PointMap>`
+ * - `Orthtree<Orthtree_traits_3<GeomTraits>, PointRange, PointMap>`.
+ *
+ * \warning this is a not a real class but an alias, please refer to
+ * the documentation of `Orthtree`.
+ *
+ * \tparam GeomTraits is a model of Kernel
+ * \tparam PointRange is a range type that provides random access iterators over the indices of a set of points.
+ * \tparam PointMap is a type that maps items in the range to Point data
```
MaelRL, line 36: Need proper requirement

Orthtree/include/CGAL/Octree.h
```diff
@@ -0,0 +1,50 @@
...
+/*!
+ * \ingroup PkgOrthtreeClasses
+ *
+ * \brief Alias that specialized the `Orthtree` class to a 3D Octree.
+ *
+ * These two types are exactly equivalent:
+ * - `Octree<GeomTraits, PointRange, PointMap>`
+ * - `Orthtree<Orthtree_traits_3<GeomTraits>, PointRange, PointMap>`.
+ *
+ * \warning this is a not a real class but an alias, please refer to
```
MaelRL, line 31:
```suggestion
 * \warning This is a not a real class but an alias, please refer to
```

sgiraudot:
Thanks for the review @MaelRL, I'll apply corrections.

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
```diff
@@ -0,0 +1,60 @@
...
+  typedef unspecified_type Sphere_d; ///< The sphere type for neighbor queries.
+
+  /*!
+    A random access iterator type to enumerate the
+    %Cartesian coordinates of a point.
+  */
+  typedef unspecified_type Cartesian_const_iterator_d;
+  typedef std::array<FT, Dimension::value> Array; ///< Array used for easy point constructions.
+
+  typedef unspecified_type Adjacency; ///< Specify the adjacency directions
```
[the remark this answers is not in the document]
sgiraudot: That seems similar to what is done in [the search traits of the KD Tree](https://doc.cgal.org/latest/Spatial_searching/classRangeSearchTraits.html). Unless I need to add the functors to construct points/radius? They are not really well defined either…

Shape_detection/include/CGAL/Shape_detection/Efficient_RANSAC/Octree.h
```diff
@@ -22,703 +22,143 @@
...
+
+// Forward declaration needed for automatic traits detection without
+// including the deprecated header itself…
+template <typename Gt, typename IR, typename IPM, typename INM>
+struct Shape_detection_traits;
+
+namespace internal {
+
+template<typename InputIterator, typename PointMap>
+struct Point_map_to_indexed_point_map {
```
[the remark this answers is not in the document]
sgiraudot: That's a good idea, it doesn't even need to be a point type, it can just be a property adaptor, I'll indeed move it to the Property Maps package.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -0,0 +1,956 @@
...
+    \param output the output iterator to add the found points to (in order of increasing distance)
+   */
+  template<typename OutputIterator>
+  OutputIterator nearest_k_neighbors_in_radius
+  (Sphere& query_sphere,
+   std::size_t k, OutputIterator output) const {
+
+    // Create an empty list of points
+    std::vector<Point_with_distance> points_list;
+    if (k != (std::numeric_limits<std::size_t>::max)())
```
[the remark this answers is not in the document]
sgiraudot: max is used internally to select this behaviour particularly.

Orthtree/include/CGAL/Orthtree/Node.h
```diff
@@ -0,0 +1,597 @@
...
+
+  \brief represents a single node of the tree. Alternatively referred
+  to as a cell, orthant, or subtree.
+
+  A `Node` is a lightweight object and thus generally passed by
+  copy. It is also a `ConstRange` whose value type is `Traits::Point_d`.
+
+  \cgalModels ConstRange
+ */
+template<typename Traits, typename PointRange, typename PointMap>
```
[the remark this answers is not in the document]
sgiraudot: Node is a sub-class of Orthtree, these are the template parameters of the Orthtree class (documented there).

maxGimeno:
Conflicts

sgiraudot:
- I fixed some divergences from the original RANSAC Octree code
- I optimized the memory management (using shared pointers was responsible for making RANSAC run slower)
- I slightly modified the test case to avoid timeouts in the testsuite
- I rebased on master

sgiraudot:
Now it's really rebased.

maxGimeno:
Errors: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-103.shtml#Orthtree

sgiraudot:
They should be fixed now.

maxGimeno:

sloriot:
@lrineau  I removed "ready for master" as there is one TODO (for me actually)

sloriot:
Branch size: 12Mo! Working on it....

sloriot:
Down to 1.5Mb now (`Orthtree/examples/Orthtree/data/cube.pwn` is actually already in the repo so real size should be around 830Kb)

sloriot:
@sgiraudot  @JacksonCampolattaro the "TODO after merge" remaining is about the news entry for the website.

JacksonCampolattaro:
I'm really excited to see this merged! Should I write something for the news entry?

sloriot:
Congrats! About the news entry, you should coordinate with @sgiraudot but you can surely draft something. For example look for posts starting with "New in CGAL:" [here](https://www.cgal.org/news.html).

Repo is [here](https://github.com/CGAL/cgal-web)
See also the PR of the last post: [here](https://github.com/CGAL/cgal-web/pull/8)
Website doc [here](https://github.com/CGAL/cgal/wiki/Tools)

JacksonCampolattaro:
Cool, I'll arrange something with Simon right away. I've been missing out by not watching the news page, that April Fools joke is great!

lrineau:
> @sgiraudot @JacksonCampolattaro the "TODO after merge" remaining is about the news entry for the website.

Done now: https://github.com/CGAL/cgal-web/pull/10

PR #4859, opened 2020-07-21
Fix CMake module for Boost IO Streams

sgiraudot:
## Summary of Changes

The `INTERFACE_LINK_LIBRARIES` option requires a single string, with semicolon-separated values. This fixes it.

## Release Management

* Affected package(s): Classification / Installation
* Bug introduced in master (5.1)


lrineau:
The minimal CMake version that is tested with CGAL 5.0 and later is CMake 3.13.4 on Debian Stable (currently version 10, *Buster*).

Maybe we could acknowledge that, and bump our minimal required version of CMake to 3.13, or at least 3.11.

maxGimeno:
Without surprise, no errors in the testsuite as it only tests  for cmake >3.13.

Note that the cmake version on travis is currently 3.12.4, so if we change the minimal req we should take 3.11 instead of 3.13.

Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-202.shtml

PR #4861, opened 2020-07-21
Arrangement 2: Small Enhancements

sgiraudot:
## Summary of Changes

This PR contains the following enhancements:
- in `Arr_segment_traits_2`, when computing the intersection between 2 segments, an overlap test on the bboxes is done to exit the function early when the segments are obviously distinct
- the bbox computation is now done on the points directly to avoid constructing a segment
- a more efficient specialization of `do_intersect()` is added to take into account the fact the the orientations of the segments are known in this structure
- `boost::fast_pool_allocator` is now used by default instead of `std::allocator` for both `Arrangement_on_surface_2` and `Arr_dcel_base`, as it is a more efficient allocator for list-like structures

Experimentally, all these enhancement lead to a speed-up between 10% and 50% depending on the dataset.

## Release Management

* __Note:__ this PR is based on master as the modified pieces of code have been changed since 5.0 and I want to avoid conflicts/integration problems.

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -34,6 +34,7 @@
 #include <CGAL/Arrangement_2/Arrangement_2_iterators.h>
 #include <CGAL/assertions.h>
 
+#include <boost/pool/pool_alloc.hpp>
 
 namespace CGAL {
 
@@ -876,7 +877,7 @@ public:
  * The arrangement DCEL class.
  */
 template <class V, class H, class F,
-          class Allocator = CGAL_ALLOCATOR(int) >
+          class Allocator = boost::fast_pool_allocator<int> >
 class Arr_dcel_base {
 public:
   // Define the vertex, halfedge and face types.
--- a/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
@@ -672,13 +672,25 @@ public:
 
     /*! Determine whether the bounding boxes of two segments overlap
      */
-    bool do_overlap(const X_monotone_curve_2& cv1,
-                    const X_monotone_curve_2& cv2) const
+    bool do_bboxes_overlap(const X_monotone_curve_2& cv1,
+                           const X_monotone_curve_2& cv2) const
     {
-      const Kernel& kernel = m_traits;
-      auto bbox_ctr = kernel.construct_bbox_2_object();
-      auto bbox1 = bbox_ctr(cv1.source()) + bbox_ctr(cv1.target());
-      auto bbox2 = bbox_ctr(cv2.source()) + bbox_ctr(cv2.target());
+      auto x1_min = CGAL::to_double(cv1.left().x());
+      auto x1_max = CGAL::to_double(cv1.right().x());
+      auto y1_s = CGAL::to_double(cv1.source().y());
+      auto y1_t = CGAL::to_double(cv1.target().y());
+      auto bbox1 = (y1_s < y1_t) ?
+        CGAL::Bbox_2(x1_min, y1_s, x1_max, y1_t) :
+        CGAL::Bbox_2(x1_min, y1_t, x1_max, y1_s);
+
+      auto x2_min = CGAL::to_double(cv2.left().x());
+      auto x2_max = CGAL::to_double(cv2.right().x());
+      auto y2_s = CGAL::to_double(cv2.source().y());
+      auto y2_t = CGAL::to_double(cv2.target().y());
+      auto bbox2 = (y2_s < y2_t) ?
+        CGAL::Bbox_2(x2_min, y2_s, x2_max, y2_t) :
+        CGAL::Bbox_2(x2_min, y2_t, x2_max, y2_s);
+
       return CGAL::do_overlap(bbox1, bbox2);
     }
 
@@ -701,7 +713,7 @@ public:
                                                         Intersection_result;
 
       // Early ending with Bbox overlapping test
-      if (! do_overlap(cv1, cv2)) return oi;
+      if (! do_bboxes_overlap(cv1, cv2)) return oi;
 
       // Early ending with specialized do_intersect
       const Kernel& kernel = m_traits;
--- a/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
@@ -797,7 +797,6 @@ public:
                       const Point_2& B1, const Point_2& B2,
                       const Kernel& k) const
     {
-      auto less_xy = k.less_xy_2_object();
       auto compare_xy = k.compare_xy_2_object();
       namespace interx = CGAL::Intersections::internal;
 
--- a/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
@@ -8,8 +8,9 @@
 // SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
 //
 // Author(s): Ron Wein          <wein@post.tau.ac.il>
-//            Efi Fogel         <efif@post.tau.ac.il>
+//            Efi Fogel         <efif@gmail.com>
 //            Waqar Khan        <wkhan@mpi-inf.mpg.de>
+//            Simon Giraudot    <simon.giraudot@geometryfactory.com>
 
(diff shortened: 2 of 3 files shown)


efifogel:
First, thanks for all this. I know it serves Geometry Factory, but still...
Secondly, we (Danny & I) plan to publish the 2nd addition of the CGAL 2D
Arrangement book, and I plan to mention this and the other enhancement.

1. What does make_certain() do (why is it needed)?

2. Notice, that there is a tag called Has_do_intersect_category. It is set
to false and not used (definitely not documented).
A functor do_intersect(), which returns boolean, should be added to the
traits models in order to enable the use of the Epic kernel when applying
the free function CGAL::do_intersect(P1, P2) on two general polygons.
(Currently, this function reduces to using Traits::Intersect_2, which
requires Epec.) I assume it can be introduced as a sole requirement of an
additional concept. All this may not be directly relevant, but it is
somehow related.
                         _/

lrineau:
> 1. What does make_certain() do (why is it needed)?

https://github.com/CGAL/cgal/pull/4861/files#diff-42a41109855dc310c32d277f293fa830R796

With the `Epick` kernel, a function like:
```c++
    bool do_intersect (const Point_2 &A1,
                       const Point_2 &A2,
                       const Point_2 &B1,
                       const Point_2 &B2,
                       const Kernel& k) const
```
will be instantiated twice:
- once with `CGAL::Interval_nt<false>` as the number types of the kernel,
- and then once with an exact ring type (`CGAL::Mpfz`, `CGAL::Gmpq`, `CGAL::MP_Float`... depending on what is available).

When the function is instantiated with a kernel based on `Interval_nt` when the result type of a function like `compare_xy` is not `Comparison_result`, but `CGAL::Uncertain<Comparison_result>`. That class template `Uncertain` is like `Interval_nt` but for enums or Booleans. `make_certain` is a function to force the conversion from `Uncertain<T>` to `T`:
  - if the predicate is instantiated with `Interval_nt` as the number type, as thus with `Uncertain<Comparison_result>` as the return type of functions like `compare_xy`, then `make_certain(v)` will check if the value `v` of type `Uncertain<Comparison_result>` is certain or not. If it is certain, then it will return the corresponding value of type `Comparison_result`, and if it is uncertain it will throw the exception of type `Uncertain_conversion_exception`. Then that exception is caught by all the predicates of `Epick`, of type `Filtered_predicate<..>` and that will trigger the reevaluation of the predicate using an exact number type.
  - if the predicate is instantiated with any other number type, `make_certain` is just the identify function.

efifogel:
I will review it next week.

efifogel:
Hi Simon,

A. Please apply the attached patch.
Note the changes in lines 800 & 801. (Conceptually, the Kernel may have
state.)
The remaining changes are only formatting.

B. Would you mind using this repo for a few additional fixes?
I don't mind applying them myself, and I'll do it rather quick:

1. Move the code from _Segment_cached_2 to the derived class Arr_segment_2.
I don't see any advantage in having the extra class + derivation.
2. Introduce additional operators to
Arr_segment_traits_2::Construct_x_monotone_curve_2 to match the
constructors of Arr_segment_2.
3. Remove local allocations of kernels.
4. Use make_certain() when possible.

5. If you really want (kidding---if you really don;t mind) add a
functor Do_intersect::operator()(const
X_monotone_curve_2& cv1, const X_monotone_curve_2& cv2) const
This will eventually enable the use of Epic with CGAL::do_intersect(p, q)
from the "2D Regularized Boolean Set Operations" package.
(Currently, we resort to calling Intersect_2::operator()(...) and check
whether the resulting container is empty.)

Thanks,
Efi
                         _/

sgiraudot:
> Hi Simon, A. Please apply the attached patch.

I don't see a patch, did you forget to attach it?

I just gave you writing access to my clone, this way you can push some commits to it if you want.

efifogel:
The patch file was attached. (I can see it...)
If you want, since I have write permissions now (thanks), I can apply it
myself.
I'll go ahead and apply the other changes I suggested. Then, you'll review
my changes...

                         _/

efifogel:
oh, perhaps the attachment is dropped when sent to github.
                         _/

sgiraudot:
Yes, of course feel free to apply the patches you want (it's even better this way as you will be the author of the commit in the git history).

efifogel:
ok, I've applied the patch.
BW, you have added 2 filters to the Intersect_2 functor.
It seems to me that they are equivalent (except, perhaps, that Bbox_2
always uses double).
So, what are differences and are they worth it?
                         _/

sgiraudot:
Bbox overlapping test is very simple and very quick (much quicker than `do_intersect()` on segments), so it allows to exit early if the segments are obviously far away from each other.

The `do_intersect()` test saves the cost of construction the intersection of the support lines if segments do not intersect (the same thing is done for intersecting 2 `Kernel::Segment_2` objects in the `Intersection_2` package).

Both tests do improve performances, especially for large data sets where many segments may cross the sweep line at the same time while only few of them actually intersect.

efifogel:
ok, so I've looked more carefully at the code.
The bbox test is indeed quick because it uses doubles.
Since it uses doubles, if it passes, we cannot even tell whether the true
(precise) bounding boxes overlap.
Are we certain that the true bounding boxes do not overlap when the test
fails?
I assume that the answer is yes (but I rather be sure..)

So, concentrating at the 2nd filter, it seems that if the test passes an
intersection between the segments is guaranteed.
If this is true, we can expedite the code a bit further; we only need to
find the intersection between the underlying lines and return it. (There is
no need to test whether the intersection point or overlapping curve lie in
the respective ranges). What do you say?

                         _/

sgiraudot:
The bbox are computed after converting the coordinates of the points as intervals, so the test is supposed to be reliable:
```c++
  template <typename K>
  class Construct_bbox_2
  {

(...)

    result_type
    operator()(const Point_2& p) const
    {
      std::pair<double,double> xp = CGAL_NTS to_interval(p.x());
      std::pair<double,double> yp = CGAL_NTS to_interval(p.y());
      return Bbox_2(xp.first, yp.first, xp.second, yp.second);
    }
```

I agree with you about removing tests after computing the intersection, but I think we still need to check if the intersection is a single point or a segment.

efifogel:
On Tue, 18 Aug 2020 at 15:42, Simon Giraudot <notifications@github.com>
wrote:

> The bbox are computed after converting the coordinates of the points as
> intervals, so the test is supposed to be reliable:
>
>   template <typename K>
>   class Construct_bbox_2
>   {
>
> (...)
>
>     result_type
>     operator()(const Point_2& p) const
>     {
>       std::pair<double,double> xp = CGAL_NTS to_interval(p.x());
>       std::pair<double,double> yp = CGAL_NTS to_interval(p.y());
>       return Bbox_2(xp.first, yp.first, xp.second, yp.second);
>     }
>
> I agree with you about removing tests after computing the intersection,
> but I think we still need to check if the intersection is a single point or
> a segment.
>
Right.
I'll do it (since I'm going to edit it further anyway).

efifogel:
ok, I'm done with my changes

- I've decided not to collapse
  Arr_segment_2 : public Arr_segment_traits_2<Kernel_>::_Segment_cached_2
into one class template. I think that retaining the main code in a class
nested in the traits is beneficial after all (at least in the current
situation where the constructors below are not deprecated).

- My grand plan is to deprecate the constructors of Arr_segment_2 (and
_Segment_cached_2) and force the user to use the
Traits::Construct_x_monotone_curve_2 functor instead. I've already created
the latter, but deprecation will have to wait for the future.

- I've slightly optimized the bounding box filter. You may want to verify.
                         _/

afabri:
Hi Efi, Did you benchmark your optimization?  What gain did you observe?

sloriot:
Travis is not happy with the latest changes:

```
/home/travis/build/CGAL/cgal/Arrangement_on_surface_2/demo/Arrangement_on_surface_2/./ArrangementGraphicsItem.h:1057:40: error: no member named 'bbox' in 'CGAL::Arrangement_on_surface_with_history_2<CGAL::Arr_segment_traits_2<Kernel>, CGAL::Arr_bounded_planar_topology_traits_2<CGAL::Arr_segment_traits_2<Kernel>, Dcel<CGAL::Arr_segment_traits_2<Kernel> > > >::Curve_halfedges'

        this->curveBboxMap[ it ] = it->bbox( );

                                   ~~  ^
```

efifogel:
ok, I've placed back the bbox() member in Arr_segement_2.
It should be deprecated.
I'll deprecate it in a separate commit and a separate PR.
                         _/

maxGimeno:
Errors on windows : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-226/Arrangement_on_surface_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

efifogel:
Should be fixed now (one hopes).
                         _/

maxGimeno:

PR #4862, opened 2020-07-21
3D Demo: Display Surface_mesh properties

maxGimeno:
## Summary of Changes

This PR modifies the Display_property_plugin to make it automatically detect any property map of the target Surface-mesh_item that has an arithmetic type as value_type, and display it with a color map or ramp.

Properties of the faces are visible in Flat mode, and properties of the vertices on Gouraud mode.

It works better with PLY IO, as it allows to save/load properties witha surface_mesh.
display_screen_1
display_screen_2
display_screen_3
display_screen_4


lrineau:
@maxGimeno I think it would make sense to tag the `display_property_plugin` with the `IO` keyword. Actually, a 'Viewer' keyword would be a better name, in my opinion.

maxGimeno:
@sgiraudot  This now also works for point sets.

maxGimeno:

PR #4864, opened 2020-07-21
CGAL_Core: fix a long-standing bug ( !blocks.empty() )

lrineau:
## Summary of Changes

Fix a long-standing bug (the assertion about  `!blocks.empty()` in `MemoryPool`).

The fix is a call to `init_CORE()` before any thread_local CORE static variables, and the reason is actually simple:
  - `ExprRep::approx` may have to create a `Real` variable of a `double`,
  - that variable needs `MemoryPool<RealDouble>`.

If, for example, `Expr::getZero()` is called first, then the static variable representing that zero is created first, and will be destroyed last.

But if, during the computation, CORE needs to call `approx` on that zero, then the `MemoryPool<RealDouble>` is created. At the end of the program execution, at the destruction of static variables, the static variable for `MemoryPool<RealDouble>` is destroyed. The static variable of `Expr::getZero()` is destroyed last. ... But, as its `approx` function was called, the destructor of that variable may need `MemoryPool<RealDouble>` to call `free`, but that pool has already been destroyed!

The function `init_CORE()`, if called *before* any thread-local local static variable is created, ensured that `MemoryPool<RealDouble>` is created first, and then destroyed last during the execution of the program (or thread).

## Release Management

* Affected package(s): CGAL_Core, Installation
* Issue(s) solved (if any): fix #1888


maxGimeno:
You need to run the clean-up script : https://travis-ci.com/github/CGAL/cgal/jobs/363391911#L2907

maxGimeno:
Also, CGAL_Core is not configurable on Blake anymore : 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-202/Installation/TestReport_Blake_Windows_MSVC-2019-Community-Release-64bits.gz

lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-203.shtml.

afabri:
With a `static CORE::Expr real(0);`   in `main()`,  I still get this error.

lrineau:
@afabri Can you send your mail to gist? I would like to try and reproduce the error.

PR #4865, opened 2020-07-22
flush the buffer

zfb132:
## Summary of Changes

After using the code `CGAL::write_off()`, we should flush the buffer immediately so that the result data can be written into a certain file.  If not, we may get an empty file when reading this file next line.

## Release Management
* Affected package(s):  Polygonal_surface_reconstruction


sloriot:
Any reason not to use `std::flush`?

zfb132:
> Any reason not to use `std::flush`?

I forget the function, thanks for reminding me. I have changed the `std::endl` to `std::flush`.

lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-203.shtml

PR #4784, opened 2020-06-14
Aos_2: Change Make_x_monotone_2 return type from legacy CGAL::Object to boost::variant

efifogel:
## Summary of Changes

Change the type of the result of the Make_x_monotone_2 functor operator. More precisely, the old type was an output iterator that dereferenced CGAL::Object(). The new type is an output-iterator that dereferences variant of the possible types of results. (Both wrap either Point_2 and X_monotone_2 objects.
Also fixed some left overs of the previous change of the result type of the Intersect_2 functor from legacy CGAL::Object to boost::variant.

## Release Management

* Affected package(s): Arrangement_on_surface_2
* Issue(s) solved (if any): fix #0000, fix #0000,...
* Feature/Small Feature (if any): No (backward compatible)
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership: TAU

I named the repo Aos_2-remove_object-efif cause I thought that I would eliminate all uses of CGAL::Object, but now, I've decided to split it into 2 steps, so there will be another round that will take care of the remaining uses in insert(), arr.zone(), and vertical_decomposition().

PR #4875, opened 2020-07-23
Arrangement 2: Enhanced CCB Merging for Surface Sweep

sgiraudot:
## Summary of Changes

This PR replaces https://github.com/CGAL/cgal/pull/4851.

As said in the previous PR, _when merging 2 inner CCBs, all halfedges of one of the CCB are traversed again and updated, which can lead to an explosion of complexity, for example when computing the arrangement of a very large holes with many concavities_.

This new PR uses a similar approach than the previous one (using temporary redirections of inner CCBs towards a valid one and cleaning everything afterwards). The differences are:
- instead of union set, we simply keep, for each inner CCB, a pointer to another CCB: these pointers form chains to a valid CCB. These chains are shortened everytime we access an inner CCB. This is, in a way, a union find in-place. There is no need for a dependency to Union_find anymore
- we use a union of `Face*` / `Inner_ccb*` instead of the current face pointer in the inner CCB class, and we replace a boolean by an enum to determine if the inner CCB is valid (`Face*` is used) or not (it's just a garbage object that chains to a correct one via `Inner_ccb`). This mechanism makes the memory overhead = 0
- the valid inner CCB is always accessed through `Halfedge::inner_ccb()` without having to go through functions of AOS2, which makes the changes far less invasive and ensures we don't break anything (testsuite is fine)
- we use the construction visitor to make sure this mode is only used when sweeping (not when merging manually)

The resulting accelerations are the same (up to 40x faster in master's worst case scenario, 30% in common datasets).

## Release Management

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -477,11 +477,10 @@ public:
     const Inner_ccb* out = reinterpret_cast<const Inner_ccb*>(_clean_pointer(this->p_comp));
     if (out->is_valid())
       return out;
-    // else
-    while (!out->is_valid())
-      out = out->next();
-    const_cast<Halfedge*>(this)->set_inner_ccb(out);
 
+    // else
+    out = const_cast<Inner_ccb*>(out)->reduce_path();
+    const_cast<Halfedge*>(this)->set_inner_ccb(out);
     return out;
   }
 
@@ -495,12 +494,17 @@ public:
     Inner_ccb* out = reinterpret_cast<Inner_ccb*>(_clean_pointer(this->p_comp));
     if (out->is_valid())
       return out;
-    // else
-    while (!out->is_valid())
-      out = out->next();
-    set_inner_ccb(out);
 
+    // else
+    out = out->reduce_path();
+    set_inner_ccb(out);
     return out;
+  }
+
+  Inner_ccb* inner_ccb_no_redirect()
+  {
+    CGAL_precondition(is_on_inner_ccb());
+    return reinterpret_cast<Inner_ccb*>(_clean_pointer(this->p_comp));
   }
 
   /*! Set the incident inner CCB. */
@@ -804,13 +808,25 @@ public:
   { if (other.status == ITER_IS_NOT_SINGULAR) iter = other.iter; }
 
   /*! Get a halfedge along the component (const version). */
-  const Halfedge* halfedge() const { return (*iter); }
+  const Halfedge* halfedge() const
+  {
+    CGAL_assertion (is_valid());
+    return (*iter);
+  }
 
   /*! Get a halfedge along the component (non-const version). */
-  Halfedge* halfedge() { return (*iter); }
+  Halfedge* halfedge()
+  {
+    CGAL_assertion (is_valid());
+    return (*iter);
+  }
 
   /*! Set a representative halfedge for the component. */
-  void set_halfedge(Halfedge *he) { *iter = he; }
+  void set_halfedge(Halfedge *he)
+  {
+    CGAL_assertion (is_valid());
+    *iter = he;
+  }
 
   /*! Get the incident face (const version). */
   const Face* face() const
@@ -850,7 +866,7 @@ public:
   /*! Set the inner CCB iterator. */
   void set_iterator(Inner_ccb_iterator it)
   {
-    CGAL_assertion (status != INVALID);
+    CGAL_assertion (is_valid());
     iter = it;
     status = ITER_IS_NOT_SINGULAR;
   }
@@ -870,6 +886,15 @@ public:
   {
     status = INVALID;
     f_or_icc.icc = next;
+  }
+
+  Arr_inner_ccb* reduce_path()
+  {
+    if (is_valid())
+      return this;
+    // else
+    f_or_icc.icc = f_or_icc.icc->reduce_path();
+    return f_or_icc.icc;
   }
 };
 
@@ -994,6 +1019,7 @@ public:
   typedef typename Face_list::iterator                Face_iterator;
   typedef CGAL::N_step_adaptor_derived<Halfedge_iterator, 2>
                                                       Edge_iterator;
+  typedef typename Inner_ccb_list::iterator           Inner_ccb_iterator;
 
   // Definitions of const iterators.
   typedef typename Vertex_list::const_iterator        Vertex_const_iterator;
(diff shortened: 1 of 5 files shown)


lrineau:
Do we want @efifogel or others to review the PR, before it is merged?

sloriot:
I'll first make a review and then ask @efifogel

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-206.shtml

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1516,6 +1527,37 @@ class Arrangement_on_surface_2 {
...
+        continue;
+
+      DInner_ccb* ic1 = he->inner_ccb_no_redirect();
+      if (ic1->is_valid())
+        continue;
+
+      DInner_ccb* ic2 = he->inner_ccb();
+      if (!ic2->halfedge()->is_on_inner_ccb()
+          || ic2->halfedge()->inner_ccb_no_redirect() != ic2)
+        ic2->set_halfedge(&(*he));
```
sloriot, line 1549: I'm a bit surprised that this is needed.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -763,56 +785,116 @@ class Arr_inner_ccb : public In_place_list_base<Arr_inner_ccb<V,H,F> >
...
+  Arr_inner_ccb* next() const
+  {
+    CGAL_assertion (status == INVALID);
+    return f_or_icc.icc;
+  }
+
+  /*! Set the next CCB to primary chain. */
+  void set_next(Arr_inner_ccb* next)
+  {
+    status = INVALID;
```
sloriot, line 887: You should add an assertion that status is not already `INVALID` (should be guaranteed by the way you use it)

Arrangement_on_surface_2/include/CGAL/Arrangement_2/Arrangement_on_surface_2_impl.h
```diff
@@ -2739,14 +2739,22 @@ _insert_at_vertices(DHalfedge* he_to,
...
 
-      // Make all halfedges along ic2 to point to ic1.
-      DHalfedge* curr;
-
-      for (curr = he2->next(); curr != he1; curr = curr->next())
-        curr->set_inner_ccb(ic1);
+      if (m_sweep_mode)
+      {
+        CGAL_assertion(ic1->is_valid());
+        CGAL_assertion(ic2->is_valid());
```
sloriot, line 2745: You can add a comment that the precondition are valid because both ic1 and ic2 are obtained using `inner_ccb()` and that this function is doing path compression.

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
```diff
@@ -1516,6 +1527,37 @@ class Arrangement_on_surface_2 {
 
   //@}
 
+  /*!
+   * Cleans the inner CCB if sweep mode was used, by removing all
+   * non-valid inner CCBs
+   */
+  void clean_inner_ccbs()
```
sloriot, line 1534: the function name should remind that is it is related to a post sweep

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -763,56 +785,116 @@ class Arr_inner_ccb : public In_place_list_base<Arr_inner_ccb<V,H,F> >
...
+    f_or_icc.icc = next;
+  }
+
+  Arr_inner_ccb* reduce_path()
```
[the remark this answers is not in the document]
sgiraudot: In that current implementation it does update the pointer (see the following lines). But I'll change that accordingly to your comments about recursiveness.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
```diff
@@ -763,56 +785,116 @@ class Arr_inner_ccb : public In_place_list_base<Arr_inner_ccb<V,H,F> >
...
+  {
+    if (is_valid())
+      return this;
+    // else
+    f_or_icc.icc = f_or_icc.icc->reduce_path();
```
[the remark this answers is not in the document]
sgiraudot: I will.

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
sgiraudot: It's actually not, it's remaining from an older version of the code which needed it. I just tested and it never happens (as expected). I'll replace it with an assertion.

Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
sgiraudot: Yes, but that is not going to be true anymore after my upcoming modification.

Arrangement_on_surface_2/include/CGAL/Arrangement_2/Arrangement_on_surface_2_impl.h
sgiraudot: Indeed.

Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
sgiraudot: Indeed.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-222.shtml

lrineau:
As far as I am concerned, this PR is the last one blocking the release of 5.0.4 and 5.1. It seems we are waiting for a review from Efi (@efifogel). Do we wait or can I prepare CGAL-5.1 without this pull-request?

efifogel:
Please don't wait.
It may take me longer to review this one and besides I rather merge the
ones below (which will already have conflicts).

   - https://github.com/CGAL/cgal/pull/4784
   - https://github.com/CGAL/cgal/pull/4942
   - https://github.com/CGAL/cgal/pull/4943
   - https://github.com/CGAL/cgal/pull/4861

If you want, you can merge 4861 for 5.0.4 and 5.1 regardless.
                         _/

lrineau:
I am not sure, but this PR is probably responsible for the runtime errors on Windows in Arr_2, Sweep, Bool_op_2, and Minkovski_2: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-5.shtml#Arrangement_on_surface_2

Tonight, I will merge others PR about Arr_2 instead of this one. I suggest you and @afabri check that this PR run fine on Windows.

efifogel:
I haven't reviewed this one yet.
                         _/

PR #4895, opened 2020-08-03
Surface Sweep 2: Use one (small) unordered set per curve

sgiraudot:
## Summary of Changes

Surface sweep uses an unordered set of pairs of curves to keep track of which intersections were already computed. This set constantly grows (nothing is ever removed from it) and may impact the performances badly when the number of curves is high.

In practice, this unique global set can be replaced by one set per curve to keep track of which _other_ curve was already intersected with that curve. Most sets will remain very small and local, with faster insertions.

In addition, because in many cases, the number of intersected curves per curve remains very small, a mechanism similar to `small_vector` is introduced to avoid paying the cost of instantiating an `unordered_set` if the number of intersections is small.

Experimentally, these modifications decrease computation times of the sweep algorithm by 35% in best cases (many curves with few intersections) while not penalizing "worst" cases (few curves with many intersections).

Packages relying on Surface Sweep (Arrangement, Boolean Operations) also benefit from this.

_(Note: a class badly named `random_access_input_iterator` was included in the file `Curve_pair.h` which is no longer needed. I fixed the name of the class to `Random_access_output_iterator` and changed the name of the file accordingly.)_

## Release Management

* Affected package(s): Surface Sweep (+ packages that depend on it: Arrangement, Boolean Operations, etc.)


STL_Extension/include/CGAL/Small_unordered_set.h
```diff
@@ -0,0 +1,85 @@
...
+  needed and thus not done.
+*/
+template<typename Key, std::size_t MaxSize>
+class Small_unordered_set
+{
+  using Array = std::array<Key, MaxSize>;
+  using Set   = std::unordered_set<Key>;
+
+  Array m_array;
+  std::shared_ptr<Set> m_set;
```
MBkkt, line 53:
```
constexpr static std::size_t MaxSize = sizeof(Set) / sizeof(Key);
// Are you really needed MaxSize? May be this?
union {
    Set set;// typical sizeof(std::unordered_set) is 56~
    Array array;
} impl;
uint8_t size;

Small_unordered_set() {
    size = 0;
    impl.array = Array{};
}

bool insert(const Key &key) {
    if (size <= MaxSize) {
        for (const Key &a : impl.array) {
            if (a == key) {
                return false;
            }
        }
        if (size < MaxSize) {
            impl.array[size++] = key;
            return true;
        } else {
            Array temp = std::move(impl.array);
            impl.set = Set{};
            impl.set.reserve(++size);
            impl.set.insert(temp.begin(), temp.end());
        }
    }
    return impl.array.set.insert(key).second;
}
```

STL_Extension/include/CGAL/Small_unordered_set.h
```diff
@@ -0,0 +1,85 @@
...
+  std::size_t m_size = 0;
+
+public:
+
+  bool insert (const Key& key)
+  {
+    if (m_size != MaxSize)
+    {
+      for (const Key& a : m_array)
```
MBkkt, line 62: This is probably a bug, as the array is fill of garbage.

Surface_sweep_2/include/CGAL/Surface_sweep_2/Random_access_output_iterator.h
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2006,2007,2009,2010,2011 Tel-Aviv University (Israel).
```
MBkkt, line 1: Maybe you should use "git mv"?

STL_Extension/include/CGAL/Small_unordered_set.h
MBkkt: Even if you decide to use the pointer use unique_ptr instead shared

Surface_sweep_2/include/CGAL/Surface_sweep_2/Surface_sweep_2_impl.h
```diff
@@ -520,9 +517,8 @@ template <typename Vis>
 
   CGAL_assertion(c1 != c2);
 
-  // look up for (c1,c2) in the table and insert if doesnt exist
-  Curve_pair cv_pair(c1,c2);
-  if (! (m_curves_pair_set.insert(cv_pair)).second) {
+  // look up for c1 in the table of c2 (or vice versa if c2<c1) and insert if doesnt exist
+  if ((c1 < c2 ? c1->intersection_exists(c2) : c2->intersection_exists(c1))) {
```
MBkkt, line 521: Maybe it is worth calculating in advance some kind of middle address?
Otherwise, there is a skew in filling
```
bool insert;
if (c1 < middle && c2 < middle) {
    insert = c1 < c2 ? c1->intersection_exists(c2) : c2->intersection_exists(c1);
} else {
    insert = c1 < c2 ? c2->intersection_exists(c1) : c1->intersection_exists(c2);
}
```
Maybe
```
if (c1 < middle && c2 < middle) {
    insert = c1 < c2 ? c1->intersection_exists(c2) : c2->intersection_exists(c1);
} else if (c1 < middle) {
    insert = (middle - c1) < (c2 - middle) ? c1->intersection_exists(c2) : c2->intersection_exists(c1);
} else if (c2 < middle) {
    insert = (middle - c2) < (c1 - middle) ? c2->intersection_exists(c1) : c1->intersection_exists(c2);
} else {
    insert = c2 > c1 ? c2->intersection_exists(c1) : c1->intersection_exists(c2);
}
```
Another option is to check for availability in both and insert in the one with less. Then the size will be "uniform"

STL_Extension/include/CGAL/Small_unordered_set.h
sgiraudot: I think you missed the point of this set: it is not to reduce memory usage but to increase speed. Max size should be tuned to minimize computation times (not to match the "weight" of an unordered set). Finding the value in an array is done linearly and we certainly don't want to do it for large arrays.

Using a union just for the sake of saving one pointer seems unnecessary to me. Also using unions of non-simple types also implies calling destructors by hand and overall makes the code less readable and more error-prone.

Shared pointer is used because this structure can be copied.

STL_Extension/include/CGAL/Small_unordered_set.h
sgiraudot: No it's not, at this point it's filled with all the values already inserted.

Surface_sweep_2/include/CGAL/Surface_sweep_2/Random_access_output_iterator.h
sgiraudot: I did, Github interprets this as deleted file + new file, I can't really do much about it.

STL_Extension/include/CGAL/Small_unordered_set.h
MBkkt: for example m_size == 1 m_array.size() always == MaxSize

MBkkt: https://en.cppreference.com/w/cpp/container/array `initializes the array following the rules of aggregate initialization (note that default initialization may result in indeterminate values for non-class T)`

Surface_sweep_2/include/CGAL/Surface_sweep_2/Surface_sweep_2_impl.h
sgiraudot: I don't really understand what you mean by "middle" address (if it is really the "middle", as in `(c1+c2)/2`, then I don't see the point of testing `c1<middle && c2<middle` which will obviously return false everytime).

And overall I don't see the point of this suggested modification. Why would we want size to be "uniform"?

STL_Extension/include/CGAL/Small_unordered_set.h
MBkkt: I understood correctly, you just use MaxSize == 8, and I thought that maybe it is worth just adjusting them

Union is standard practice for sso

If you want easy cope this container, you should use shared_ptr<Container> in curve class, because otherwise the semantics of copying are not obvious, besides, why would you need to copy the array every time

Surface_sweep_2/include/CGAL/Surface_sweep_2/Surface_sweep_2_impl.h
MBkkt: I mean the global median, if you can count it

but it may be difficult, and it looks strange, I like the option with lookup in both maps more

STL_Extension/include/CGAL/Small_unordered_set.h
sgiraudot: My bad, you are correct, I mistakenly replaced an index-based loop by this one. I'll fix it.

Surface_sweep_2/include/CGAL/Surface_sweep_2/Surface_sweep_2_impl.h
MBkkt: we may want a uniform size to increase the speed of the lookup, and to increase the chance that we will avoid allocation.  

For example, now the curve with the smallest address will collect all its intersects

STL_Extension/include/CGAL/Small_unordered_set.h
```diff
@@ -0,0 +1,124 @@
...
+
+  Array m_array;
+  std::unique_ptr<Set> m_set;
+  std::size_t m_size = 0;
+
+public:
+
+  Small_unordered_set() { }
+
+  ~Small_unordered_set() { }
```
sloriot, line 60: useless

STL_Extension/include/CGAL/Small_unordered_set.h
```diff
@@ -0,0 +1,124 @@
...
+    m_size = other.m_size;
+    if (other.m_set)
+      m_set = std::make_unique<Set>(*other.m_set);
+    else
+      m_array = other.m_array;
+  }
+
+  Small_unordered_set& operator= (Small_unordered_set&& other)
+  {
+    if (other.m_set)
```
sloriot, line 91: what about `m_size`?

sgiraudot: That is a bug, I'll fix it.

STL_Extension/include/CGAL/Small_unordered_set.h
sgiraudot: I'll remove them.

lrineau: Useless special member functions have the mean effect of disabling the automatic generation of move-constructors and move-assignment.

maxGimeno:
This PR conflicts in master : the file Surface_sweep_2/include/CGAL/Surface_sweep_2/Curve_pair.h has been modified since 5.0, and it is removed by this PR.

sgiraudot:
Alright, I also rebased this one on master so that it's easier to integrate.

maxGimeno:
Thanks !

maxGimeno:

PR #4899, opened 2020-08-04
Surface Sweep: Enhance memory allocation with Compact Container

sgiraudot:
## Summary of Changes

This PR enhances memory allocation in surface sweep by:
- using a `Compact_container` to allocate elements in `Multiset` (by adding an optional non-documented template argument to `Multiset` defaulting to the default non-CC-based `Multiset` version)
- simplifying the allocation and tracking of allocated events, by merging a multiset (used to allocate events) and a vector (used to keep track of the events allocated) in a single `Compact_container` structure that takes care of both while providing a much better memory allocation (taking advantage of memory freed by previous events)

These enhancements lead to a speed-up of 8% to 25% in best cases.

## Release Management

* Affected package(s): STL Extension + Surface Sweep + packages that rely on it (Arrangement, Boolean Operation, etc.)

### DIFF
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -910,12 +910,11 @@ namespace internal {
       m_ptr = nullptr;
     }
 
-    CC_iterator (pointer p)
-#ifdef CGAL_COMPACT_CONTAINER_DEBUG_TIME_STAMP
-      : ts(0)
-#endif
+    static CC_iterator create_from_pointer (pointer p)
     {
-      m_ptr.p = p;
+      CC_iterator out;
+      out.m_ptr.p = p;
+      return out;
     }
 
   private:
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -910,13 +910,6 @@ namespace internal {
       m_ptr = nullptr;
     }
 
-    static CC_iterator create_from_pointer (pointer p)
-    {
-      CC_iterator out;
-      out.m_ptr.p = p;
-      return out;
-    }
-
   private:
 
     typedef typename DSC::Time_stamper           Time_stamper;
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -910,6 +910,14 @@ namespace internal {
       m_ptr = nullptr;
     }
 
+    CC_iterator (pointer p)
+#ifdef CGAL_COMPACT_CONTAINER_DEBUG_TIME_STAMP
+      : ts(0)
+#endif
+    {
+      m_ptr.p = p;
+    }
+
   private:
 
     typedef typename DSC::Time_stamper           Time_stamper;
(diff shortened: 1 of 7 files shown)


STL_Extension/include/CGAL/Compact_container.h
```diff
@@ -908,6 +908,14 @@ namespace internal {
       m_ptr.p = nullptr;
     }
 
+    CC_iterator (pointer p)
+#ifdef CGAL_COMPACT_CONTAINER_DEBUG_TIME_STAMP
+      : ts(0)
+#endif
+    {
+      m_ptr.p = p;
+    }
```
lrineau, line 917: When I saw that new constructor, I thought about new ambiguities. And I was right, Travis shows a lot of errors:
https://travis-ci.com/github/CGAL/cgal/builds/178369312

maxGimeno:
errors : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-32/Arrangement_on_surface_2/TestReport_lrineau_Fedora-Release.gz

sgiraudot:
The `Compact_container` API changed a bit since 5.0, which made this PR fail on master. I've rebased it on 5.1 and included the API changes, it seems to work now.

gdamiand (commented):
I think https://github.com/CGAL/cgal/blob/6b991a95ad4d449e030af21996ff078f20a1b0ee/Surface_sweep_2/include/CGAL/Surface_sweep_2/No_overlap_event_base.h#L121
is not correct.

I think changing m_point as if it is a pointer, while it is a data member will produce wrong behavior...

Now EPECK objects can be used in compact container, you can directly write:
m_point.for_compact_container(p).

maxGimeno:
@gdamiand @sgiraudot What is the status of this ?

sgiraudot:
As far as I'm concerned, the PR works.

Of course, a `reinterpret_cast` is theoretically undefined behavior, but in practise, the value of the pointer will be written at the memory position of the beginning of the `Event` object, which is large enough to hold a pointer. It is a particularly dirty form of squatting, but I don't see a better way to do it without adding a pointer to the `Event` object and thus increasing memory consumption, which we want to avoid.

The `for_compact_container()` will only work for CGAL EPECK points, do we want to only support this?

gdamiand:
> As far as I'm concerned, the PR works.
> 
> Of course, a `reinterpret_cast` is theoretically undefined behavior, but in practise, the value of the pointer will be written at the memory position of the beginning of the `Event` object, which is large enough to hold a pointer. It is a particularly dirty form of squatting, but I don't see a better way to do it without adding a pointer to the `Event` object and thus increasing memory consumption, which we want to avoid.

I am not sure it works; I think this could be a problem depending on the values in the memory.... Compact container use bit squatting for pointers, knowing the fact that the two least significant bits of a pointer are always zero. But this is not true for other data, and thus using these two bits for something else than a pointer could be source of bug...

sgiraudot:
Alright, so I fixed my code the following way:
- if `Point_2` has a method `void* for_compact_container()`, then it is used
- otherwise, an additional pointer is used

This makes the Epeck version work without increasing memory consumption, while keeping the code clean in other cases (at the cost of an additional pointer).

maxGimeno:
```
The following files have tabs:
Surface_sweep_2/include/CGAL/Surface_sweep_2/No_overlap_event_base.h
```

maxGimeno:

PR #4901, opened 2020-08-04
Point Set Processing: Remove outdated precondition

sgiraudot:
## Summary of Changes

In `edge_aware_upsample_point_set()`, there is a precondition `neighbor_radius > 0` while the default value of `neighbor_radius` is `-1`. It is documented that the neighbor radius is set automatically if lower than a factor of the average spacing, so the default value is correct and the precondition is wrong. I removed it.

## Release Management

* Affected package(s): PSP


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-212.shtml

PR #4902, opened 2020-08-05
PMP: Restore array-based soup to mesh conversion

MaelRL:
## Summary of Changes

This was broken in https://github.com/CGAL/cgal/pull/4418, but untested so unseen.

An example is modified so the code is tested now.

## Release Management

* Issue(s) solved (if any): -

### DIFF
--- a/Polygon_mesh_processing/examples/Polygon_mesh_processing/repair_polygon_soup_example.cpp
+++ b/Polygon_mesh_processing/examples/Polygon_mesh_processing/repair_polygon_soup_example.cpp
@@ -5,30 +5,55 @@
 #include <CGAL/Polygon_mesh_processing/orient_polygon_soup.h>
 #include <CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h>
 
+#include <algorithm>
+#include <array>
 #include <iostream>
 #include <vector>
 
 typedef CGAL::Exact_predicates_inexact_constructions_kernel     K;
+typedef K::FT                                                   FT;
 typedef K::Point_3                                              Point_3;
 
-typedef std::vector<std::size_t>                                CGAL_Polygon;
 typedef CGAL::Surface_mesh<Point_3>                             Mesh;
 
+typedef std::array<FT, 3>                                       Custom_point;
+typedef std::vector<std::size_t>                                CGAL_Polygon;
+
 namespace PMP = CGAL::Polygon_mesh_processing;
+
+struct Array_traits
+{
+  struct Equal_3
+  {
+    bool operator()(const Custom_point& p, const Custom_point& q) {
+      return (p == q);
+    }
+  };
+
+  struct Less_xyz_3
+  {
+    bool operator()(const Custom_point& p, const Custom_point& q) {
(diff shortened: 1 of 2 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
```diff
@@ -37,13 +38,26 @@ namespace CGAL {
...
+{
+  CGAL_static_assertion((std::is_convertible<PS_Point, PM_Point>::value));
+  return PM_Point(p);
+}
+
+// just for backward compatibility reasons
+template <typename PM_Point, typename PS_FT>
+PM_Point convert_to_pm_point(const std::array<PS_FT, 3>& p)
+{
+  return PM_Point(p[0], p[1], p[2]);
```
sloriot, line 52: why not always using this one?

Polygon_mesh_processing/examples/Polygon_mesh_processing/repair_polygon_soup_example.cpp
```diff
@@ -5,30 +5,55 @@
...
 
-typedef std::vector<std::size_t>                                CGAL_Polygon;
 typedef CGAL::Surface_mesh<Point_3>                             Mesh;
 
+typedef std::array<FT, 3>                                       Custom_point;
+typedef std::vector<std::size_t>                                CGAL_Polygon;
+
 namespace PMP = CGAL::Polygon_mesh_processing;
 
+struct Array_traits
```
sloriot, line 24: Since a geom traits is needed, I think I would also add in this example the point map. This gives a nice example for a user defined point. You can then undo the changes in the other file.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h
MaelRL: * This is weaker than the generic one using explicit conversion;
* `operator[]` is in `CGAL::Point_3` but is not part of its concept;
* Might want to do weird stuff with `CGAL::Point_2` for example.

Polygon_mesh_processing/examples/Polygon_mesh_processing/repair_polygon_soup_example.cpp
MaelRL: But that would go against explicitly testing both functions with `std::array<FT, 3>` as point type, wouldn't it ?

sloriot: As soon as the geom_traits is needed, the old behavior is broken. But that's not an issue, the old code was somehow incorrect since it was using free functions/operators.

MaelRL: The traits is there for `repair_polygon_soup`, not `polygon_soup_to_polygon_mesh`.

sloriot:
Tested in CGAL-5.1-Ic-216

PR #4910, opened 2020-08-06
Removing false include CGAL/license/Polyhedron.h

mhsaar:
write_vtk.h is under LGPL and does not require a license file.


afabri:
See #4906

PR #4911, opened 2020-08-06
Fix do_intersect for Tetrahedron_3 Tetrahedron_3

mhsaar:
## Summary of Changes

In Line 65 in function do_intersect_tetrahedron_bounded (tr,tet,p,k) we eval  k.has_on_bounded_side_3_object()(tet, p), where tr=lh_tet and tet=rh_tet.
Thus p must be generated from lh_tet or L65 is always true. 

## Release Management

* Affected package(s): Kernel23


afabri:
[link](https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Tetrahedron_3_Bounded_3_do_intersect.h#L65)

afabri:
You  are right.  This is a copy paste from the test with the sphere.  And it was my fault.  @maxGimeno can you please check in which release it was added, so that the fix makes it into the bugfix releases.   And could you please add test cases in the testsuite.

sloriot:
Tested in CGAL-5.1-Ic-216

PR #4914, opened 2020-08-07
Generic 2D/3D Periodic Triangulations

MaelRL:
## Summary of Changes

This PR adds a new type of 2D and 3D periodic triangulations whose base domain is not an iso_rectangle/cuboid, but a Voronoi face, thus greatly expanding the potential use cases.

This new class also uses the name `Periodic_2_(Delaunay)_triangulation_2`, and the choice between both implementations is done by looking at the domain type.

2D Periodic Triangulations are almost fully rewritten to use Bowyer-Watson rather than flips, and some other issues in periodic classes are fixed along the way (LINK TO ISSUES TODO).

TODO:
- [ ] Reintroduce documented swap functionality in P2T2
- [ ] Insertion of ranges + vertex removal in P2T2
- [ ] Could implement P2RT2 almost zero effort now
- [ ] Mirror all changes from P2T2 to P3T3
- [ ] filter-graph to remove all unintended changes (e.g. README.md)
- [ ] Document + small feature
- [ ] Tests

EXTRA:
- [ ] Generic periodic without exact constructions?

## Release Management

* Affected package(s): `Periodic_2_triangulation_2`, `Periodic_3_triangulation_3`
* Small Feature (if any): TODO
* License and copyright ownership: INRIA


mglisse:
Hi Mael, what is the status of this PR?
We were waiting until it is done to add a python wrapper in gudhi, since the API should be a bit different, but if it is stalled maybe we should just wrap the old P3DT3 (we have demand for the cuboid case)? Is the API in this PR roughly finalized (unless people complain in the small feature of course), so we can judge how to write our wrapper in a way that can be updated without breaking our users' code?

MaelRL:
It's stalled due to lack of time, but there's no theoretical issue.

The API should be relatively stable, and pretty much the same as every other triangulations (even the current "master" periodic triangulations) for the usual functionalities such as `insert`, `remove`, `move`, `number_of_X` etc.

So far this PR is work only on P2T2 and is still missing quite a bit to finish. And then everything needs to be mirrored into P3T3...

mglisse:
> The API should be relatively stable, and pretty much the same as every other triangulations (even the current "master" periodic triangulations) for the usual functionalities such as `insert`, `remove`, `move`, `number_of_X` etc.

Yes, the interesting part is specifying the domain, the need sometimes to call convert_to_1_sheeted_covering or deal with the redundancy, etc.

Thank you for the information, we will probably go ahead without waiting then.

Good luck with finishing the PR when you find the time.

MaelRL:
For the domains, nothing will (should) change for the flat torus.

It is also backward compatible: I replace e.g. `Periodic_2_triangulation_2` by a more generic class that is templated by the domain, and by default the domain is the flat torus.

PR #4921, opened 2020-08-11
STL Extension: Small Stack Allocator

sgiraudot:
## Summary of Changes

When a lot of containers are instantiated and are expected to contain a small number of elements, it can be interesting to use static allocation (through `std::array` or C-array style structures) to avoid the cost of dynamic allocation for every small container instantiated. This is the principle behind `boost::container::small_vector` for example.

This PR introduces `CGAL::Small_stack_allocator` which aims at generalizing this principle to other types of container, like `std::list`.

This PR makes this allocator used in the Event structure of the `Surface_sweep_2` package: in that case, events keep track of their adjacent curves through `std::list` data structures. In practice, lots of events can be instantiated but usually, only a few curves are adjacent to an event, so the `std::list` rarely grow to more than 1 or 2 curves. Using the small stack allocator allows to speed up computation by a factor up to 15% in my experimentation.

Note that this small stack allocator could maybe be used in other places that I don't know of.

## Release Management

* Affected package(s): STL Extensions, Surface Sweep + packages that rely on it (Arrangement, etc.)

### DIFF
--- a/STL_Extension/include/CGAL/Small_stack_allocator.h
+++ b/STL_Extension/include/CGAL/Small_stack_allocator.h
@@ -93,7 +93,11 @@ class Small_stack_allocator
 {
 public:
 
-  using value_type = T;
+  using pointer =               T*;
+  using const_pointer =   const T*;
+  using reference =             T&;
+  using const_reference = const T&;
+  using value_type =            T ;
 
   template <typename T2, std::size_t MaxSize2>
   friend class Small_stack_allocator;
@@ -129,6 +133,25 @@ public:
   {
     m_pool.deallocate (p, n);
   }
+
+  // Functions construct and destroy are not used by modern compilers
+  // and are deprecated from the Allocator concept in C++17. They are
+  // included to comply with older compilers and should be removed at
+  // some point.
+
+  template <typename U, typename ... Args>
+  void construct (U* u, Args&& ... args)
+  {
+    ::new (static_cast<void*>(u))
+      U (std::forward<Args> (args)...);
+  }
+
+  template <typename U>
+  void destroy (U* u)
+  {
+    u->~U();
+  }
+
 };
 
 } // namespace CGAL
--- a/STL_Extension/include/CGAL/Small_stack_allocator.h
+++ b/STL_Extension/include/CGAL/Small_stack_allocator.h
@@ -52,7 +52,7 @@ public:
     }
 
     // Else, fallback to `new` allocation
-    return new T[n];
+    return static_cast<T*>(::operator new(n * sizeof(T)));
   }
 
   void deallocate (T* p, std::size_t n)
@@ -69,7 +69,7 @@ public:
     }
     // Else, delete it
     else
-      delete[] p;
+      ::operator delete(p);
   }
 
 };
--- a/STL_Extension/include/CGAL/Small_stack_allocator.h
+++ b/STL_Extension/include/CGAL/Small_stack_allocator.h
@@ -16,7 +16,6 @@
 
 #include <cstdlib>
 
-
 namespace CGAL
 {
 
@@ -42,6 +41,8 @@ public:
   T* allocate (std::size_t n)
   {
     CGAL_BRANCH_PROFILER("stack allocations / total allocations of Small_stack_allocator", prof);
+
+    // If chunk to allocate does not exceed remaining stack size, use stack
     if (m_next + n < MaxSize)
     {
       CGAL_BRANCH_PROFILER_BRANCH(prof);
@@ -50,17 +51,23 @@ public:
       return out;
     }
 
+    // Else, fallback to `new` allocation
     return new T[n];
   }
 
   void deallocate (T* p, std::size_t n)
   {
+    // If pointer is part of the pool, nothing to do
     if (m_pool <= p && p < m_pool + MaxSize)
     {
       std::size_t pos = static_cast<std::size_t>(p - m_pool);
+
+      // If pointer was the last one allocated, we can use again this
+      // chunk of memory
       if (pos + n == m_next)
         m_next = pos;
     }
+    // Else, delete it
     else
       delete[] p;
   }
@@ -69,6 +76,18 @@ public:
 
 } // namespace internal
 
+/*
+  The small stack allocator holds a statically allocated array of type
+  T. When allocation is required, it first returns pointers from that
+  statically allocated space. When allocation exceeds this small size,
+  it goes to the "normal" mode and allocates with `new`.
+
+  When deallocating, the stack can be used again if the last element
+  allocated on it is deallocated (in the absence of the free list, it
+  is difficult to do better than that). Keep in mind that this
+  allocator is mainly designed for containers that you know are going
+  to do few allocations/deallocations.
+*/
 template <typename T, std::size_t MaxSize>
 class Small_stack_allocator
 {
@@ -96,7 +115,7 @@ public:
 
   // Do not copy pool
   Small_stack_allocator(const Small_stack_allocator&) { }
-  Small_stack_allocator& operator=  (const Small_stack_allocator&)
+  Small_stack_allocator& operator= (const Small_stack_allocator&)
   {
     return Small_stack_allocator();
   }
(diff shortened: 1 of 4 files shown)


sgiraudot:
Travis is failing, I'll have a look.

STL_Extension/include/CGAL/Small_stack_allocator.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // Else, fallback to `new` allocation
+    return new T[n];
+  }
+
+  void deallocate (T* p, std::size_t n)
+  {
+    // If pointer is part of the pool, nothing to do
+    if (m_pool <= p && p < m_pool + MaxSize)
```
mglisse, line 61: The legality of this kind of test regularly comes up...

lrineau: What would be the alternative?

maxGimeno:
Errors : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.1-Ic-227/Arrangement_on_surface_2/TestReport_Blake_Windows_MSVC-2019-Community-Release-64bits.gz

sgiraudot:
This time they should be fixed…

lrineau:
There are errors on Windows: https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.2-Ic-5#STL_Extension

https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.2-Ic-5/STL_Extension/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

sgiraudot:
I've tried again several things like address sanitizer and I can't reproduce on Linux. I'm out of ideas, could someone using Windows test this?

maxGimeno:
The errors only happen on Gauguin, where the visual licences have gone staled. Therefore, I don't have access to any debugger on it until Jane comes back (she's the owner of the license on Gauguin.)
Sorry.

janetournois:
Sorry @sgiraudot I did not see you request.
I'm giving it a try

janetournois:
I can reproduce the runtime error. Trying to understand/fix it

STL_Extension/include/CGAL/Small_stack_allocator.h
```diff
@@ -0,0 +1,168 @@
...
+  template <typename T2>
+  struct rebind
+  {
+    using other = Small_stack_allocator<T2, MaxSize>;
+  };
+
+private:
+
+  using Pool = internal::Small_stack_allocator_pool<T, MaxSize>;
+  Pool m_pool;
```
theartful, line 122: @sgiraudot As far as I know, you can't define allocator resources inside the allocator itself. The allocator should only hold a reference/pointer to the resource (buffer pool). And when you copy assign the allocator using rebound type, the new allocator should share the resource, and hence you can't parameterize the buffer pool on the type.

See here:
https://en.cppreference.com/w/cpp/named_req/Allocator

STL_Extension/include/CGAL/Small_stack_allocator.h
```diff
@@ -0,0 +1,168 @@
...
+  {
+    CGAL_BRANCH_PROFILER("stack allocations / total allocations of Small_stack_allocator", prof);
+
+    std::size_t N = n * m_type_size;
+
+    // If chunk to allocate does not exceed remaining stack size, use stack
+    if (m_next + N < m_pool_size)
+    {
+      CGAL_BRANCH_PROFILER_BRANCH(prof);
+      char* out = m_pool + m_next;
```
theartful, line 54: alignment should be taken into consideration here

STL_Extension/include/CGAL/Small_stack_allocator.h
```diff
@@ -0,0 +1,168 @@
...
+  Small_stack_allocator_pool& operator= (const Small_stack_allocator_pool&) = delete;
+
+  T* allocate (std::size_t n)
+  {
+    CGAL_BRANCH_PROFILER("stack allocations / total allocations of Small_stack_allocator", prof);
+
+    std::size_t N = n * m_type_size;
+
+    // If chunk to allocate does not exceed remaining stack size, use stack
+    if (m_next + N < m_pool_size)
```
theartful, line 51: I think this should be `m_next + N <= m_pool_size`.

STL_Extension/include/CGAL/Small_stack_allocator.h
theartful: For example:
```C++
#include <CGAL/Small_stack_allocator.h>
#include <vector>

constexpr std::size_t max_size = 100;
typedef CGAL::Small_stack_allocator<int, max_size> SSA;

int main()
{
  std::vector<std::vector<int, SSA>, SSA> vec;
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
}
```
This segfaults in gcc, which is due to the allocator can't be relocated (you can't std::move neither std::swap a vector without freeing the stack data in the pool).

theartful: For example this segfaults in gcc since the allocators can't be reloacted without corrupting the memory.
```C++
#include <CGAL/Small_stack_allocator.h>
#include <vector>

constexpr std::size_t nb_tests = 10000000;

constexpr std::size_t max_size = 100;
typedef CGAL::Small_stack_allocator<int, max_size> SSA;

int main()
{
  std::vector<std::vector<int, SSA>> vec;
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
  vec.push_back({1, 2, 3, 4});
}
```

sgiraudot:
Thanks for all these remarks @theartful !

I have tried to modify the structures accordingly, but I quickly realized that it will make using this allocator quite tedious and not reliable from a user's point of view: it means that the size of the pool needs to be fixed, even though the containers will likely use it to store larger data structures.

So instead of `Small_stack_allocator<T, MaxSize>`, to get the intended behavior, users need to know the size of whatever structures the container is going to use it for, like `Small_stack_allocator<T, MaxSize * sizeof(List_element)>`. Except that this structure's size is probably very different depending on STL containers (I expect it to be `sizeof(T)` for `vector` but larger for `list`, for example). In addition, it does not seem to be documented, nor accessible through a constant or typedef, and likely to change from one platform to another (or one version of the STL to another). I don't see a simple to get the intended behavior (= use the stack as long `container.size() < MaxSize` and fallback to dynamic allocation after that) reliably.

At this point, I'm considering forgetting this small stack allocator idea, and just implement directly implement a `CGAL::Small_list` similarly to what is done with `boost::container::small_vector`...

theartful:
I think a way to change the requirement would be to the use the stack as long as allocated data < MaxSize (bytes) with no guarantees on the actual number of the elements.

However, I agree with you that this might be suboptimal, and a `small_vector` implementation might be more convenient (especially since you also have to define an external stack pool and guarantee that it outlive the structures that use it).

lrineau:
As allocators are difficult to set up (and so are containers), should not we rely on libraries instead of doing that on our own?

If we bump the minimal version to 1.68, Boost.Containers has allocators:
https://www.boost.org/doc/libs/1_68_0/doc/html/container/extended_allocators.html

And so has C++17 STL.

That looks like easier to maintain than our own type of allocators or containers. We already have `CGAL::internal::vector<T>` in CGAL, and we really would like to avoid making the same mistakes.

sgiraudot:
I agree but my problem is that what I want to achieve (a list-like structure with a small part statically allocated because most of these lists will in practice have 2 elements at most) does not seem to exist neither in Boost (even 1.68 unless I missed it) nor in C++17 STL. It seems my idea to use an allocator was no that good because of how constrained and complex the API of iterators is, but I'm not sure I can do better than creating a container.

sgiraudot:
I'm closing this PR, it is replaced by https://github.com/CGAL/cgal/pull/5040 introducing a `Small_list` container.

PR #4923, opened 2020-08-13
fix edge_to_faces key in Collect_face_bbox_per_edge_bbox

grassofsky:
## Summary of Changes

fix compile error when using Collect_face_bbox_per_edge_bbox as callback when DO_NOT_HANDLE_COPLANAR_FACES is defined


sloriot:
Could you elaborate on why you use this undocumented flag? I wanted to remove it but maybe you have an interesting use case.

grassofsky:
> anted to remove it but maybe you have an interesting use ca

Sign. There is no interesting use case. When I write my own callback reference to the Collect_face_bbox_per_edge_bbox, the compile error is found.

sloriot:
If you can share some details, I'm interested in knowing what your own callback is doing :star_struck:

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-30.shtml.

PR #4924, opened 2020-08-13
Surface Sweep + Arr Overlay: Init with indexed edges

sgiraudot:
## Summary of Changes

When performing a sweep on a set of connected segments, the connectivity is actually lost (disconnected segments are passed to the sweep) and found again through the event comparator (if two segments have the same endpoint, then this endpoint will only trigger one event), which is an unnecessarily costly operation.

This PR adds a new function `indexed_sweep()` to surface sweep which takes a range of edges with an accessor to get indices and some other information, so that each vertex only trigger one event without the need to use the comparison operator in the multiset of events. A set of indexed events is maintained accordingly during the initialization. The rest of the sweep is the same as without indices.

This indexed sweep if used in `Arr_overlay_2.h` (except when geodesic traits are used as the duplication of top and bottom vertices make it not usable so far). It could be used in other places: namely, we have experimented constructing General Polygon Sets this way and it does work well, but it's not included in this PR as we have a better lead for improving GPS (not using a sweep at all).

In practice, this speeds up boolean operations by up to 30%.

## Release Management

* Affected package(s): Surface Sweep, Arrangement_2, Boolean Operations, etc.

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
@@ -57,14 +57,14 @@ public:
     return arr1.number_of_vertices() + arr2.number_of_vertices();
   }
 
-  std::size_t source_index (const Curve& c) const
-  {
-    return reinterpret_cast<std::size_t>(halfedge(c)->source()->inc());
-  }
-
-  std::size_t target_index (const Curve& c) const
+  std::size_t min_end_index (const Curve& c) const
   {
     return reinterpret_cast<std::size_t>(halfedge(c)->target()->inc());
+  }
+
+  std::size_t max_end_index (const Curve& c) const
+  {
+    return reinterpret_cast<std::size_t>(halfedge(c)->source()->inc());
   }
 
   const Curve& curve (const Curve& c) const
@@ -258,7 +258,7 @@ overlay(const Arrangement_on_surface_2<GeometryTraitsA_2, TopologyTraitsA>& arr1
   if (total_iso_verts == 0) {
     // Clear the result arrangement and perform the sweep to construct it.
     arr.clear();
-    if (std::is_same<typename Agt2::Base_traits_2::Bottom_side_category,
+    if (std::is_same<typename Agt2::Bottom_side_category,
                      Arr_contracted_side_tag>::value)
       surface_sweep.sweep (xcvs_vec.begin(), xcvs_vec.end());
     else
@@ -297,7 +297,7 @@ overlay(const Arrangement_on_surface_2<GeometryTraitsA_2, TopologyTraitsA>& arr1
 
   // Clear the result arrangement and perform the sweep to construct it.
   arr.clear();
-  if (std::is_same<typename Agt2::Base_traits_2::Bottom_side_category,
+  if (std::is_same<typename Agt2::Bottom_side_category,
       Arr_contracted_side_tag>::value)
     surface_sweep.sweep(xcvs_vec.begin(), xcvs_vec.end(),
                         pts_vec.begin(), pts_vec.end());
--- a/Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
@@ -297,8 +297,16 @@ overlay(const Arrangement_on_surface_2<GeometryTraitsA_2, TopologyTraitsA>& arr1
 
   // Clear the result arrangement and perform the sweep to construct it.
   arr.clear();
-  surface_sweep.sweep(xcvs_vec.begin(), xcvs_vec.end(),
-                      pts_vec.begin(), pts_vec.end());
+  if (std::is_same<typename Agt2::Base_traits_2::Bottom_side_category,
+      Arr_contracted_side_tag>::value)
+    surface_sweep.sweep(xcvs_vec.begin(), xcvs_vec.end(),
+                        pts_vec.begin(), pts_vec.end());
+  else
+    surface_sweep.indexed_sweep (xcvs_vec,
+                                 Indexed_sweep_accessor
+                                 <Arr_a, Arr_b, Ovl_x_monotone_curve_2>
+                                 (arr1, arr2),
+                                 pts_vec.begin(), pts_vec.end());
   xcvs_vec.clear();
   pts_vec.clear();
 }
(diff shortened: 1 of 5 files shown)


Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
```diff
@@ -40,6 +40,81 @@
...
+    for (typename Arr2::Vertex_const_iterator vit = arr2.vertices_begin();
+         vit != arr2.vertices_end(); ++vit, ++idx)
+      vit->set_inc (backup_inc[idx]);
+  }
+
+private:
+
+  typename Curve::Halfedge_handle halfedge (const Curve& c) const
+  {
+    if (c.red_halfedge_handle() == typename Curve::Halfedge_handle())
```
sloriot, line 108: travis is not happy with this line

Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
```diff
@@ -40,6 +40,81 @@
...
+    }
+    for (typename Arr2::Vertex_const_iterator vit = arr2.vertices_begin();
+         vit != arr2.vertices_end(); ++vit, ++idx)
+    {
+      backup_inc[idx] = vit->inc();
+      vit->set_inc (reinterpret_cast<void*>(idx));
+    }
+  }
+
+  void after_init() const
```
sloriot, line 93: add a comment saying that you are restoring the property after the squat

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
     m_visitor->after_sweep();
   }
 
+  /*!
+    Run the indexed sweep-line algorithm on a given range of
```
sloriot, line 313:
```suggestion
    runs the indexed sweep-line algorithm on a given range of
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+    x-monotone curves and accessor. The main difference from the
+    original sweep() function is that the accessor allows to get
```
sloriot, line 315:
```suggestion
    original `sweep()` function is that the accessor allows to get
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+    indices of end-points and avoid checking several times in the
+    event queue when the same vertex as several incident edges.
```
sloriot, line 317:
```suggestion
    event queue when the same vertex has several incident edges.
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+
+    \param edges A range of edges
+    \param accessor An object providing, for a value_type `e` of
```
sloriot, line 320: You might want to introduce a concept since you are reusing it

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+    accessor.before_init();
+    _init_indexed_sweep(edges, accessor);
+    accessor.after_init();
+    _sweep();
+    _complete_sweep();
+    m_visitor->after_sweep();
+  }
+
+  /*!
+    Run the indexed sweep-line algorithm on a given range of
```
sloriot, line 343:
```suggestion
    runs the indexed sweep-line algorithm on a given range of
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+    x-monotone curves and accessor. The main difference from the
+    original sweep() function is that the accessor allows to get
```
sloriot, line 345:
```suggestion
    original `sweep()` function is that the accessor allows to get
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+    indices of end-points and avoid checking several times in the
+    event queue when the same vertex as several incident edges.
```
sloriot, line 347:
```suggestion
    event queue when the same vertex has several incident edges.
```

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
```diff
@@ -309,6 +309,78 @@ class No_intersection_surface_sweep_2 {
...
+                      const Accessor& accessor,
+                      PointInputIterator action_points_begin,
+                      PointInputIterator action_points_end)
+  {
+    m_visitor->before_sweep();
+    accessor.before_init();
+    _init_indexed_sweep(edges, accessor);
+    accessor.after_init();
+    _init_points(action_points_begin, action_points_end, Event::ACTION);
+    //m_visitor->after_init();
```
sloriot, line 378: remove?

Surface_sweep_2/include/CGAL/No_intersection_surface_sweep_2.h
sgiraudot: All of these documentations are internal (the sweep classes themselves are completely internal and not publicly documented). Do we have such things as internal concepts?

sloriot: OK I miss that point. No need then.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-33.shtml.

Arrangement_on_surface_2/include/CGAL/Arr_overlay_2.h
lrineau: @sloriot Is this blocking the merge?

sloriot: I think the comment is here now. Marking it as resolved.

PR #4937, opened 2020-08-19
Polyhedron Demo: Fix c3t3 colors after optimization

### LINKED ISSUE #4935: Polyhedron demo - Mesh global optimizers loose their patch IDs
## Issue Details

In the demo :
- open a polyhedron with sharp edges (e.g. data/star.off)
- run Mesh_3 (default parameters are OK) : the surface has one color per patch ID
image

- run `lloyd_optimize_mesh_3()` : the surface has lost (almost all) its patch colors/IDs
image

Note that :
- `odt_optimize_mesh_3()` has the same issue, but `perturb_mesh_3()` and `exude_mesh_3()` don't.
- I have tried in a standalone cpp file (I can send it if it helps) and the patch IDs are valid.

## Environment

* Compiler: MSVC 2019
* CGAL version: master
* Boost version: 1.67.0

maxGimeno:
## Summary of Changes
When performing optimizations on a c3t3, the original domain is lost, and a new one is re-created according to the type of the original meshed item (surface_mesh, segmented image or implicit function), with default values.

In the case of a surface_mesh, the default behavior when claling mesh_3 is to detect features, with a sharp_edges_angle_bound parameter set to 60. This is not the default behaviour in optimizers, so the results clearly differ, as there are no patch ids in the second case. 
To "fix" this behaviour, I added two members (and accessors) to the c3t3_item, to store at the meshing time if we detect the features or not. Using these, those two parameters are retrieved in the optimizer pass, allowing to rebuild the right patch ids in the domain. 
There was also a wrong color set in the new optimized item, and now it is the source item's that is used.

## Release Management

* Affected package(s): Demo
* Issue(s) solved (if any): fix #4935

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Mesh_3_plugin_cgal_code.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Mesh_3_plugin_cgal_code.cpp
@@ -87,6 +87,14 @@ Meshing_thread* cgal_code_mesh_3(QList<const SMesh*> pMeshes,
   std::cerr << " done (" << timer.time() * 1000 << " ms)" << std::endl;
 
   Scene_c3t3_item* p_new_item = new Scene_c3t3_item(surface_only);
+  if(polylines.empty()) {
+    if(protect_features) {
+      p_new_item->set_sharp_edges_angle(sharp_edges_angle);
+    }
+    else if (protect_borders) {
+      p_new_item->set_detect_borders(true);
+    }
+  }
 
   QString tooltip = QString("<div>From \"") + filename +
     QString("\" with the following mesh parameters"
--- a/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Optimization_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Mesh_3/Optimization_plugin.cpp
@@ -524,10 +524,9 @@ treat_result(Scene_c3t3_item& source_item,
                             static_cast<float>(bbox.ymin() + bbox.ymax())/2.f,
                             static_cast<float>(bbox.zmin() + bbox.zmax())/2.f);
 
-    result_item.setColor(QColor(59,74,226));
+    result_item.setColor(source_item.color());
     result_item.setRenderingMode(source_item.renderingMode());
     result_item.set_data_item(source_item.data_item());
-
     source_item.setVisible(false);
 
     const Scene_interface::Item_id index = scene->mainSelectionIndex();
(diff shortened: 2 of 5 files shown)


janetournois:
Actually the color/subdomain/patch id info are stored in the cells of the C3T3, and re-detecting sharp edges of the input before optimization should not make any change.
I think that duplicating the C3T3 for optimization could be responsible for the issue, but I am surprised it does not happen for perturb and exude (which have a similar API).

Here is the output I get for Lloyd : only a few facets (their location is funny though) keep their colors.

image

maxGimeno:
> Actually the color/subdomain/patch id info are stored in the cells of the C3T3

From what I see in the code, during optimization, every impacted facet is removed from the complex and then re-added, given a patch_id according to the domain. If the domain did not compute features, is will give only one patch ID.
see here : https://github.com/CGAL/cgal/blob/master/Mesh_3/include/CGAL/Mesh_3/C3T3_helpers.h#L1279

> re-detecting sharp edges of the input before optimization should not make any change.

Did you try the branch ? Because I tried with and without it, and the change is quite obvious to me. 

> I think that duplicating the C3T3 for optimization could be responsible for the issue

No it is not. When you create the new item from the source item's c3t3, all patches are the same. After optimization, they all changed.

>Here is the output I get for Lloyd 

Here is what I get after my patch : 
snap

janetournois:
> > Actually the color/subdomain/patch id info are stored in the cells of the C3T3
> 
> From what I see in the code, during optimization, every impacted facet is removed from the complex and then re-added, given a patch_id according to the domain. If the domain did not compute features, is will give only one patch ID.
> see here : https://github.com/CGAL/cgal/blob/master/Mesh_3/include/CGAL/Mesh_3/C3T3_helpers.h#L1279
> 
Right.
And local optimizers update the C3T3 facets differently.

> > re-detecting sharp edges of the input before optimization should not make any change.
> 
> Did you try the branch ? Because I tried with and without it, and the change is quite obvious to me.
> 
Yes I did
The screenshot I sent was made with the branch.

> > I think that duplicating the C3T3 for optimization could be responsible for the issue
> 
> No it is not. When you create the new item from the source item's c3t3, all patches are the same. After optimization, they all changed.
> 
Ok

I will have a look inside the optimizers code and come back to you.

janetournois:
With recompiling more plugins that I thought was needed, it's fixed indeed!
Thanks @maxGimeno  :-)

janetournois (approved):

sloriot:
Tested in 5.1-Ic-223 (manual zgrep to exclude clip errors + check there is no warning)

PR #4939, opened 2020-08-19
Arrangement: Avoid constructing Segment_2/Line_2 objects when possible

sgiraudot:
## Summary of Changes

This PR is another small enhancement for arrangement:

- we define an alternative undocumented edge iterator for `Polygon_2` which does not construct any `Segment_2` object but just returns pairs of Points (pairs of reference wrappers to be precise). This allows, when constructing a GPS from a Polygon, to avoid the cost of constructing very short-living `Segment_2`
- when constructing an `Arr_segment_2`, we don't construct the `Line_2` support immediately but delay this construction until it's needed. This is especially useful in conjunction with https://github.com/CGAL/cgal/pull/4861 when intersections of segments are discarded early (some segments' line support may not need to be computed at all). To do so, I replace the boolean `is_vert` by an enum (to have a `UNKNOWN` value when the line is not computed). Of course, it would also be possible to add another boolean but it seems more memory-friendly like this.

These 2 tricks allow to reduce computation time of boolean operations using GPS by around 5%.

## Release Management

* Affected package(s): Polygon 2, Arrangement, etc.

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
@@ -82,17 +82,13 @@ public:
     typedef typename Kernel::Point_2               Point_2;
 
   protected:
-    mutable Line_2 m_l;       // the line that supports the segment.
-    Point_2 m_ps;             // the source point of the segment.
-    Point_2 m_pt;             // the target point of the segment.
-    bool m_is_directed_right; // is (lexicographically) directed left to right.
-    mutable enum              // is this a vertical segment.
-    {
-      UNKNOWN,                  // If the line is not computed
-      REGULAR,                  // If the line is computed and not vertical
-      VERTICAL                  // If the line is computed and vertical
-    } m_is_vert;
-    bool m_is_degen;          // is the segment degenerate (a single point).
+    mutable Line_2 m_l;         // the line that supports the segment.
+    Point_2 m_ps;               // the source point of the segment.
+    Point_2 m_pt;               // the target point of the segment.
+    bool m_is_directed_right;   // is (lexicographically) directed left to right.
+    mutable bool m_is_vert;     // is this a vertical segment.
+    mutable bool m_is_computed; // is the support line computed.
+    bool m_is_degen;            // is the segment degenerate (a single point).
 
   public:
 
@@ -1170,7 +1166,8 @@ public:
 template <typename Kernel>
 Arr_segment_traits_2<Kernel>::_Segment_cached_2::_Segment_cached_2() :
   m_is_directed_right(false),
-  m_is_vert(UNKNOWN),
+  m_is_vert(false),
+  m_is_computed(false),
   m_is_degen(true)
 {}
 
@@ -1178,7 +1175,8 @@ Arr_segment_traits_2<Kernel>::_Segment_cached_2::_Segment_cached_2() :
 template <typename Kernel>
 Arr_segment_traits_2<Kernel>::
 _Segment_cached_2::_Segment_cached_2(const Segment_2& seg)
-  : m_is_vert(UNKNOWN)
+  : m_is_vert(false)
+  , m_is_computed(false)
 {
   Kernel kernel;
   auto vertex_ctr = kernel.construct_vertex_2_object();
@@ -1200,7 +1198,8 @@ _Segment_cached_2::_Segment_cached_2(const Point_2& source,
                                      const Point_2& target) :
   m_ps(source),
   m_pt(target),
-  m_is_vert(UNKNOWN)
+  m_is_vert(false),
+  m_is_computed(false)
 {
   Kernel kernel;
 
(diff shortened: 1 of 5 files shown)


maxGimeno:
This branch is marked 5.2 but is base don 5.0. 
Also, it strongly conflicts with master, so I won't test for now.

sgiraudot:
As the modified files changed a lot since 5.0, I just rebased this branch on master to avoid numerous conflicts and make integration easier.

maxGimeno:
Thanks !

maxGimeno:
I think this PR is responsible for https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-31/Arrangement_on_surface_2/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz but it might be #4924

sgiraudot:
No, it came from the other one :slightly_smiling_face: 
I fixed it.

maxGimeno:
Cool, thx !

efifogel:
When you run the program below you see that the size of the _Segment_cached_2 record is 32 bytes.
It consists of 8 bytes for the underlying line, 16 bytes for the source and target points and 8 additional bytes for the remaining 3 Boolean flags. If you add 1, 2, or 5 additional Boolean flags the size does not change.
Only when you add the 6th flag, which makes it a total of 9 Boolean flags all together, the size jumps to 40 bytes.

1. Can you please verify. (I hadn't been aware of that before I conducted this small experiment.) Debug, Release, Other compilers?
2. If this is indeed the case, I suggest that we simply add an additional Boolean flag.
3. If this is not the case, I would use bit masks instead....
4. In Arr_geometric_traits/Circle_segment_2.h we use bit masks. If the size is affected as I describe, the use of bit masks might be redundant and replacing it with Booleans may expedite computation...

#include <CGAL/Arr_segment_traits_2.h>
#include <CGAL/Arrangement_2.h>
typedef CGAL::Exact_predicates_exact_constructions_kernel   Kernel;
typedef CGAL::Arr_segment_traits_2<Kernel>                  Traits_2;
typedef Traits_2::X_monotone_curve_2                        Segment_2;
typedef Traits_2::Point_2                                   Point_2;
int main() {
  std::cout << "size of Line_2: " << sizeof(Kernel::Line_2) << std::endl;
  std::cout << "size of Point_2: " << sizeof(Point_2) << std::endl;
  std::cout << "size of Segment_2 (_Segment_cached_2): " << sizeof(Segment_2) << std::endl;
  return 0;
}

sgiraudot:
You are correct, indeed the size is 40 with the enum and only 32 with 2 booleans. That's good to know!

I changed to 2 booleans which seems simpler in terms of implementation, efficient and easy to read and understand.

Arrangement_on_surface_2/include/CGAL/Arr_segment_traits_2.h
```diff
@@ -82,17 +82,13 @@ class Arr_segment_traits_2 : public Kernel_ {
...
-    Point_2 m_ps;             // the source point of the segment.
-    Point_2 m_pt;             // the target point of the segment.
-    bool m_is_directed_right; // is (lexicographically) directed left to right.
-    mutable enum              // is this a vertical segment.
-    {
-      UNKNOWN,                  // If the line is not computed
-      REGULAR,                  // If the line is computed and not vertical
-      VERTICAL                  // If the line is computed and vertical
-    } m_is_vert;
-    bool m_is_degen;          // is the segment degenerate (a single point).
```
lrineau: The underlying type of an enum is implementation defined. However, since C++11, one can specify it:
```c++
enum Is_vert : char { /* ... */ ];
```

Actually, the size of a `bool` is also implementation-defined. The version with the `enum` might be smaller.

And be also careful of alignement: the `class` can be smaller if the two or four `bool` are packed together.

sgiraudot: The class being smaller with two of four `bool` packed together is pretty much what Efi analyzed: in the case of this class, there was already 2 `bool`, so adding a third one does not change the size.

I did not know about the specification of the enum type, it's very interesting and good to know! In the case of this PR, as the size won't be smaller, I think using 2 booleans remains more readable (and the change is less invasive than replacing the existing boolean with an enum).

lrineau: No problem. I use feedbacks on issues to "trickle down" things I know on C++. In the context of this issue, I understand why using the booleans can be more useful or efficient.

maxGimeno:
Error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-35/Arrangement_on_surface_2/TestReport_gimeno_Fedora-rawhide.gz but it might be because of #5091

sgiraudot:
Unless I missed something, the latest errors did come from #5091, so I think this PR can be tested again.

maxGimeno:
We'll see that with tomorrow's results :)

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-42.shtml#Arrangement_on_surface_2_Demo

efifogel:
For now, I've un-deprecated Arr_segment_2::bbox().
I was hoping that deprecating it would be easier, but it requires more
work, and it does not belong to this branch.
In case you wonder, bbox() should be deprecated, because it does not belong
to any concept.
Instead, we should introduce a functor (in all traits classes) that
computes the (inexact) bbox.
                         _/

lrineau:
Let's all remember that the code freeze is already over for more than one month. I agreed to push it a bit, but any new important development will remove the pull-request from the set of tested PR for CGAL-5.2.

efifogel:
Sure thing. I'm not touching it further.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-43.shtml.

PR #4933, opened 2020-08-18
Disable boost_serialization and iostreams

maxGimeno:
## Summary of Changes
Use -DCMAKE_DISABLE_FIND_PACKAGE_XXX to keep travis to test the classification IO system, as we don't have a working version of boost for that.

## Release Management

* Issue(s) solved (if any): fix #4929

PR #4940, opened 2020-08-20
Travis: Use another PPA for boost

maxGimeno:
## Summary of Changes
As we found out, the boost versions on the mhier PPA is not fitted for CGAL, as they are not built with zlib and bzip2, which is needed for Classification. 

As a reminder, the problem is that the components iostreams and serialization are found, but they are missing all the parts we use, which leads to linking errors. I don't think there is an easy way to detect if those components are built correctly, so I decided to use another ppa for boost, one that does build the two components correcly. 

In #4933, I tried to use CMAKE_DISABLE_FIND_PACKAGE to block iostreams and serialization, but for some reason it doesn't work on linux. (I say on linux because I use it in the cgal-swig-bindings in the windows CI and it is not ignored like it is on travis) .

### DIFF
--- a/.travis/build_package.sh
+++ b/.travis/build_package.sh
@@ -11,7 +11,7 @@ function mytime {
 function build_examples {
   mkdir -p build-travis
   cd build-travis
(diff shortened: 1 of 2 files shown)


maxGimeno:
Tests here : https://travis-ci.com/github/maxGimeno/cgal/builds/180629004

PR #4947, opened 2020-08-23
Fix error in HSV/RGB color conversion

oboes:
## Summary of Changes

The function `CGAL::Color::set_hsv` ignores (most of) the value of `hue`, because for some reason `(hue/60.)` is converted to an integer. I changed two lines in the color conversion code and it now has the expected behaviour.

I also removed one line of documentation because it was inconsistent with the code.

## Release Management

* Affected package(s): IO Streams


lrineau (approved):

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.1-Ic-228.shtml

lrineau:
Thanks for you patch @oboes

PR #4955, opened 2020-08-26
Use [[no_unique_address]] for info in vertex/cell

mglisse:
## Summary of Changes

In 3d there are separate classes for the case with_info. In dD, it is a single class. Since by default there is no data, avoid wasting space using the relevant attribute. I didn't check if this really saves anything, but it shouldn't hurt.

## Release Management


lrineau (approved):
I approve.

maxGimeno:

PR #4960, opened 2020-08-27
Polyhedron Demo: .mesh reading enhancement

maxGimeno:
## Summary of Changes
When reading a .mesh in the demo, if a subdomain is 0, make it 1.

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
+++ b/Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
@@ -106,7 +106,8 @@ bool build_finite_cells(Tr& tr,
     boost::unordered_map<std::array<typename Tr::Vertex_handle, 3>,
                         std::vector<std::pair<typename Tr::Cell_handle, int> > >& incident_cells_map,
     const std::map<std::array<int,3>, typename Tr::Cell::Surface_patch_index>& border_facets,
-    const bool verbose)
+    const bool verbose,
+                        bool prevent_domain_0 = true)
 {
   typedef std::array<int, 5>              Tet_with_ref; // 4 ids + 1 reference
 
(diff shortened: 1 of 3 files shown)


Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
```diff
@@ -141,7 +151,15 @@ bool build_finite_cells(Tr& tr,
 
     Cell_handle c = tr.tds().create_cell(vs[0], vs[1], vs[2], vs[3]);
     c->set_subdomain_index(tet[4]); // the cell's info keeps the reference of the tetrahedron
-
+    c->info() = tet[4]; // the cell's info keeps the reference of the tetrahedron
```
janetournois, line 154: tetrahedral remeshing uses this function, and we do not have `info()` in cells.
Why isn't `subdomain_index()` enough?

janetournois:
Can you please explain why you make a 0 subdomain_index 1?
0 is for "outside"... it status should not be changed, right?

sloriot:
If you read random medit file 0 is not a special value

janetournois:
However it is special or not, why do we change it?
It amounts to merging 2 subdomains... (number 0 and number 1)

sloriot:
It is done only if 1 is not present IIRC

Mesh_3/include/CGAL/Mesh_3/tet_soup_to_c3t3.h
maxGimeno: It seems it is after all, must have been a reminiscence from an older try

maxGimeno:
> However it is special or not, why do we change it?
> It amounts to merging 2 subdomains... (number 0 and number 1)

It doesn't. It takes the first number that is not used by subdomains

janetournois:
Ok.
Two more questions : 
- what if you have negative subdomain indices?
- `prevent_domain_0 = true` sounds like the subdomain 0 is avoided by default. From what I understand, it is used the over way round. Am I wrong?

maxGimeno:
I did not change anything about the negative subdomain indices.

In the case where prevent_domain_0 is true,  there is a CGAL_precondition(tet[4] > 0)), so the domain 0 is forbidden in the .mesh file. That is why I call it `prevent`.

maxGimeno:
errors https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-2/Tetrahedral_remeshing_Examples/TestReport_lrineau_ArchLinux-clang.gz

maxGimeno:

PR #4972, opened 2020-09-02
Polyhedron Demo: Use PCH

maxGimeno:
## Summary of Changes
Use PCH for plugins to speed up the compilation time.


lrineau:
I think our Travis CI should try to compile the Polyhedron demo in two modes:
  - with, or
  - without PCH.

Otherwise we will have a lot of such issues with missing includes.

maxGimeno:
@lrineau if we compile the polyhedron demo in the CI, it will add around 80min (IIRC) to each run, and most of them will lead to a timeout error.

lrineau:
Could we try to compile on Github Actions?

maxGimeno:
Probably

sloriot:
@maxGimeno  could you please refresh it so that we can test this branch?

maxGimeno:
@sloriot Actually I'd rather drop it. The use of the PCH is invisible in terms of performance, I think I saw a difference of 1.30 min on a 1h30 build, something like that, but it heavily modifies all of the CMakeLists of the demo and plugins, and makes it a lot harder to maintain.

PR #4990, opened 2020-09-10
Do not create isolated vertices

### LINKED ISSUE #4970: facets_in_complex_2_to_triangle_mesh() creates isolated vertices
## Issue Details

[CGAL:facets_in_complex_2_to_triangle_mesh()](https://doc.cgal.org/latest/Surface_mesher/group__PkgSurfaceMesher3FunctionsIO.html#ga39a105bd1817a9200ec5df208888e8aa) creates isolated vertices in the output mesh. This seems like a bug to me, as the function is documented as returning only facets.

The bug comes from the fact that all vertices of the triangulation are added to the mesh, regardless of whether or not they are used by at least one facet or not. It seems to me that a first pass should be done on facets to identify which triangulation vertices are indeed part of the output mesh, and then only instantiate mesh vertices from those.

This bug propagates to the Poisson reconstruction function which returns meshes with isolated vertices, which should not happen.

afabri:
## Summary of Changes

Do not create isolated vertices when creating a `Surface_mesh` from a `C2T3`.

There is nothing to do for the 3D equivalent `facets_in_complex_3_to_triangle_mesh()`.

## Todo 
- [x]  check the other functions which write to vtk, medit,...

## Release Management

* Affected package(s): Surface_mesher  
* Issue(s) solved (if any): fix #4970

### DIFF
--- a/Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
+++ b/Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
@@ -63,8 +63,9 @@ output_surface_facets_to_medit (std::ostream& os, const C2t3& c2t3)
       vit != tr.finite_vertices_end();
       ++vit)
   {
-    if(V.find(vit) != V.end()){
-      V[vit] = inum++;
+    auto it = V.find(vit);
+    if(it != V.end()){
+      *it = inum++;
       Point p = static_cast<Point>(vit->point());
       os << p.x() << " " << p.y() << " " << p.z() << " 0 \n";
     }
--- a/Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
+++ b/Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
@@ -65,7 +65,7 @@ output_surface_facets_to_medit (std::ostream& os, const C2t3& c2t3)
   {
     auto it = V.find(vit);
     if(it != V.end()){
-      *it = inum++;
+      it->second = inum++;
       Point p = static_cast<Point>(vit->point());
       os << p.x() << " " << p.y() << " " << p.z() << " 0 \n";
     }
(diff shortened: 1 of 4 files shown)


afabri:
@MaelRL and @maxGimeno  I hope the merge with your IO PR is smooth.

afabri:
@maxGimeno I have fixed the output to vtk, but  I could not test it  as I haven't  installed vtk. Can you please check.

maxGimeno:
I don't see isolated vertices in master in the output of output_c2t3_to_vtk_polydata(), but after your PR there is no more input at all.

Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
```diff
@@ -41,19 +41,33 @@ output_surface_facets_to_medit (std::ostream& os, const C2t3& c2t3)
...
       ++vit)
   {
-    V[vit] = inum++;
-    Point p = static_cast<Point>(vit->point());
-    os << p.x() << " " << p.y() << " " << p.z() << " 0 \n";
+    if(V.find(vit) != V.end()){
+      V[vit] = inum++;
+      Point p = static_cast<Point>(vit->point());
+      os << p.x() << " " << p.y() << " " << p.z() << " 0 \n";
+    }
```
lrineau, line 70: This code is doing the search twice in `V`. Use the iterator returned by `find` instead:
```suggestion
    auto it = V.find(vit);
    if(it != V.end()){
      *it = inum++;
      Point p = static_cast<Point>(vit->point());
      os << p.x() << " " << p.y() << " " << p.z() << " 0 \n";
    }
```

afabri:
@lrineau checked the medit output and @maxGimeno the vtk output.  Thanks.

Surface_mesher/include/CGAL/IO/Complex_2_in_triangulation_3_to_medit.h
```diff
@@ -63,8 +63,9 @@ output_surface_facets_to_medit (std::ostream& os, const C2t3& c2t3)
       vit != tr.finite_vertices_end();
       ++vit)
   {
-    if(V.find(vit) != V.end()){
-      V[vit] = inum++;
+    auto it = V.find(vit);
+    if(it != V.end()){
+      *it = inum++;
```
lrineau, line 68: Actually, my code it probably wrong:
```suggestion
      it->second = inum++;
```

You should test.

maxGimeno:

PR #4994, opened 2020-09-11
PMP isotropic remeshing - test validity during remeshing only if input mesh was valid

janetournois:
## Summary of Changes

`CGAL::isotropic remeshing()` is a local process, so it does not need the mesh to be globally valid (e.g. it is allowed to have isolated vertices).

This PR removes the global test when it's not needed, i.e. when the input mesh already was not valid.
When the input mesh is valid, we keep the test in debug mode anyhow.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -325,6 +325,9 @@ namespace internal {
                                   pmesh);
       CGAL_assertion(CGAL::is_triangle_mesh(mesh_));
       CGAL_assertion_code(input_mesh_is_valid_ = CGAL::is_valid_polygon_mesh(pmesh));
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -1961,6 +1962,7 @@ namespace internal {
     EdgeIsConstrainedMap ecmap_;
     VertexIsConstrainedMap vcmap_;
     FaceIndexMap fimap_;
+    bool input_mesh_is_valid_;
```
sloriot, line 1965:
```suggestion
    CGAL_assertion_code(bool input_mesh_is_valid_;)
```

sloriot (commented):
I'm not sure I'm in favor of such a change. While for isolated vertices it is fine, for non-manifold vertices or invalid meshes it is not. I would at least add in the constructor a `CGAL_warning_msg` if the input is not valid.

maxGimeno:

Issue #4993, opened 2020-09-11
Invalid results for PCA of segments


sloriot:
When using PCA for a set of 3D segments, I have incorrect results if I'm using the fitting of the segments while the fitting for the points works as expected.
Here is an example (`segments_0.polylines.txt`) where the segments are shown with small sphere as endpoints. The fitting of the segments is shown in green while the fitting of the points is shown in red. I'm calling the function the same way for both, with a range of segments, only the `Dimension_tag` changes.

![Screenshot from 2020-09-11 10-10-59](https://user-images.githubusercontent.com/1224632/92893191-c3873800-f419-11ea-9c47-b46392869bf7.png)

[polys.zip](https://github.com/CGAL/cgal/files/5207657/polys.zip)

Here is a log extracted from my application (`line2` is when I use `Dimenson_tag<1>()`, `line` when I use `Dimension_tag<0>()`.

```
line2: 0.235461 1.01355 0.0922231 0.0170576 0.0484649 -0.0523584 vs line 0.233188 0.994762 0.0700325 0.311133 1.63903 0.830853
edges written to segments_0.polylines.txt
line2: 0.235461 1.01355 0.0922231 0.0170576 0.0484649 -0.0523584 vs line 0.237735 1.03234 0.114414 0.31568 1.67661 0.875234
edges written to segments_1.polylines.txt
line2: 0.0020388 0.349903 0.0209326 -0.217733 -0.57448 0.332727 vs line -0.0093041 0.329981 0.0389645 -0.398203 -0.353054 0.657204
edges written to segments_2.polylines.txt
line2: 0.0020388 0.349903 0.0209326 -0.217733 -0.57448 0.332727 vs line 0.0133817 0.369825 0.00290059 -0.375518 -0.31321 0.62114
edges written to segments_3.polylines.txt
line2: -0.153438 0.330518 0.710462 -0.304608 0.784954 1.58832 vs line -0.151165 0.349309 0.732653 -0.0732194 0.993574 1.49347
edges written to segments_4.polylines.txt
line2: -0.153438 0.330518 0.710462 -0.304608 0.784954 1.58832 vs line -0.155712 0.311727 0.688272 -0.0777662 0.955992 1.44909
edges written to segments_5.polylines.txt
line2: 0.0799842 0.994168 0.781753 0.0112015 0.201589 0.175875 vs line 0.0913271 1.01409 0.763721 -0.297572 0.331054 1.38196
edges written to segments_6.polylines.txt
[...]

PR #4995, opened 2020-09-14
PCA: Add validity test and fix several bugs

sgiraudot:
## Summary of Changes

When converting the deprecated API to the Eigen version a few years ago, I must have missed a division which lead to an incorrect moment matrix initialization in the case of an input with `Segment_3`, which lead to the bug https://github.com/CGAL/cgal/issues/4993.

This PR fixes it. Now the Eigen version (which is the officially documented one) works well.

This PR also adds a validity test to check that the output of each PCA variant is correct. This revealed other errors which I fixed:
- the moment matrix of `Iso_cuboid_3` was wrong too
- the overload of the variant using `Segment_2` with `CGAL::Dimension_tag<0>` was broken due to 2 inverted arguments
- the overloads of the variants using `Iso_rectangle_2` with `CGAL::Dimension_tag<1>` and `<0` were broken due to a missing argument

All is fixed.

## Release Management

* Affected package(s): PCA
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4993

### DIFF
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -144,9 +144,9 @@ assemble_covariance_matrix_3(InputIterator first,
     FT x0 = t[0].x();
     FT y0 = t[0].y();
     FT z0 = t[0].z();
-    FT delta[9] = {t[0].x(), t[1].x(), t[2].x(),
-                   t[0].y(), t[1].y(), t[2].y(),
-                   t[0].z(), t[1].z(), t[2].z()};
+    FT delta[9] = {t[1].x()-x0, t[3].x()-x0, t[5].x()-x0,
+                   t[1].y()-y0, t[3].y()-y0, t[5].y()-y0,
+                   t[1].z()-z0, t[3].z()-z0, t[5].z()-z0};
     Matrix transformation (delta);
     FT volume = t.volume();
 
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -548,9 +548,9 @@ assemble_covariance_matrix_3(InputIterator first,
 
   // assemble 2nd order moment about the origin.
   Matrix moment;
-  moment << 1.0, 0.5, 0.0,
-            0.5, 1.0, 0.0,
-            0.0, 0.0, 0.0;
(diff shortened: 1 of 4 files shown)


afabri:
Can you please add test cases.   My guess is that a better testsuite had avoided the regession.

sgiraudot:
Alright, so I added a validity test (which was quite long to do considering the number of variants).

It actually revealed additional errors which I added to the summary of the PR. So I think it was definitely worth it, it should save us some trouble in the future.

lrineau:
This PR has a conflict with CGAL-5.0-branch.

maxGimeno:
warnings :https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-15/Principal_component_analysis/TestReport_lrineau_Fedora-Release.gz

maxGimeno:
warnings : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-17/Principal_component_analysis/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz

maxGimeno:
Runtime error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-21/Principal_component_analysis/TestReport_lrineau_Ubuntu-Latest.gz

maxGimeno:
Also a warning about diagonalize_traits that is deprecated

sgiraudot:
The error should be fixed.

The warning seems to come from `-- Eigen3 path specified in cmake variable EIGEN3_INCLUDE_DIR is set to /usr/local/include, but that path does not contains the file signature_of_eigen3_matrix_library and is considered as invalid.` but I don't think this is related to this PR (the warning pops if `CGAL_EIGEN3_ENABLED` is not defined, which is probably due to this cmake problem).

maxGimeno:
Ok, thanks

lrineau:
> The error should be fixed.
> 
> The warning seems to come from `-- Eigen3 path specified in cmake variable EIGEN3_INCLUDE_DIR is set to /usr/local/include, but that path does not contains the file signature_of_eigen3_matrix_library and is considered as invalid.` but I don't think this is related to this PR (the warning pops if `CGAL_EIGEN3_ENABLED` is not defined, which is probably due to this cmake problem).

I think:
- our `find_package(Eigen)` should be `REQUIRED`.
- Eigen should be installed on all test platforms.

maxGimeno:

PR #5001, opened 2020-09-17
Fix iterator invalidation bug

oboes:
## Summary of Changes

There is an iterator invalidation bug when looping over `degenerate_face_set` and modifying it at the same time (see [this page on cppreference](https://en.cppreference.com/w/cpp/container#Iterator_invalidation) for details on iterator invalidation).

On some meshes it can causes segmentation faults. If required I can upload a .off file containing a mesh for which it's the case (but I don't have a minimal example, only a 156K .off file).

I fix that by simply iterating over a copy of `degenerate_face_set` rather than the original one (while still modifying the original one). It might not be the most computationally efficient way to do it (it involves creating a copy of the set), but it does fix the bug in one line of code. 

## Release Management

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -1697,7 +1697,9 @@ bool remove_degenerate_faces(const FaceRange& face_range,
     std::map<edge_descriptor, bool> are_degenerate_edges;
 
     // use a copy of the set to avoid iterator invalidation
-    std::set<face_descriptor> degenerate_face_set_copy(degenerate_face_set);
+    std::vector<face_descriptor> degenerate_face_set_copy(degenerate_face_set.size());
+    std::copy(degenerate_face_set.begin(), degenerate_face_set.end(),
+              degenerate_face_set_copy.begin());
     for(face_descriptor fd : degenerate_face_set_copy)
     {
       for(halfedge_descriptor hd : halfedges_around_face(halfedge(fd, tmesh), tmesh))
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -1696,7 +1696,9 @@ bool remove_degenerate_faces(const FaceRange& face_range,
   {
     std::map<edge_descriptor, bool> are_degenerate_edges;
 
-    for(face_descriptor fd : degenerate_face_set)
+    // use a copy of the set to avoid iterator invalidation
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -1696,7 +1696,9 @@ bool remove_degenerate_faces(const FaceRange& face_range,
   {
     std::map<edge_descriptor, bool> are_degenerate_edges;
 
-    for(face_descriptor fd : degenerate_face_set)
+    // use a copy of the set to avoid iterator invalidation
+    std::set<face_descriptor> degenerate_face_set_copy;
+    for(face_descriptor fd : degenerate_face_set_copy)
```
lrineau, line 1701: Problem: `degenerate_face_set_copy` is created empty, and then the loop is on an empty set.

lrineau (changes requested):

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
oboes: Sorry, I of course meant `std::set<face_descriptor> degenerate_face_set_copy(degenerate_face_set)`.

I've now amended my commit.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -1696,7 +1696,9 @@ bool remove_degenerate_faces(const FaceRange& face_range,
   {
     std::map<edge_descriptor, bool> are_degenerate_edges;
 
-    for(face_descriptor fd : degenerate_face_set)
+    // use a copy of the set to avoid iterator invalidation
+    std::set<face_descriptor> degenerate_face_set_copy(degenerate_face_set);
+    for(face_descriptor fd : degenerate_face_set_copy)
```
lrineau, line 1701: Even better, copy in a vector, to avoid the multiple memory allocations:
```suggestion
    std::vector<face_descriptor> degenerate_face_set_copy(degenerate_face_set.size());
    std::copy(degenerate_face_set.begin(), degenerate_face_set.end(),
              degenerate_face_set_copy.begin());
    for(face_descriptor fd : degenerate_face_set_copy)
```

lrineau (approved):

MaelRL:
Thank you for the bug report.

I have rewritten the full loop in https://github.com/CGAL/cgal/pull/5012 to be less intricate, which should fix your issue. Could you please test it on your data set? I would also be interested if you could please share it, degeneracy removal is still quite WIP, and I do not have non-trivial data for this particular configuration.

oboes:
I tried and it does work now. Thanks!

PR #5006, opened 2020-09-21
Windows, FPU.h: replace `_control_fp` by SSE2 intrisics

### LINKED ISSUE #2135: Windows, FPU.h: replace `_control_fp` by intrisics
In CGAL.4.10, in <CGAL/FPU.h>,  the macros `CGAL_IA_SETFPCW` and `CGAL_IA_GETFPCW` are implemented using `_controlfp_s`, instead of using macros from `<xmmintrin.h>`.

If the platform is Windows x64, with `_M_X64` defined, [only the SSE2 FPU is used][*], and we can probably define `CGAL_SAFE_SSE2` and use the macros from `<xmmintrin.h>`.

  [*]: https://msdn.microsoft.com/en-us/library/e9b52ceh.aspx

lrineau:
## Summary of Changes

Windows, FPU.h: replace `_control_fp` by SSE2 intrisics. Fix #2135

## Release Management

* Issue(s) solved (if any): fix #2135 
* License and copyright ownership: N/A

Cc: @afabri please check if that accelerates on Windows

### DIFF
--- a/Number_types/include/CGAL/FPU.h
+++ b/Number_types/include/CGAL/FPU.h
@@ -75,8 +75,8 @@ extern "C" {
 // Pure and safe SSE2 mode (g++ -mfpmath=sse && (-msse2 || -march=pentium4))
 // can be detected by :
 // TODO : see what Intel and VC++ have to say about this.
-#if defined __FLT_EVAL_METHOD__ && defined __SSE2_MATH__ && \
-      (__FLT_EVAL_METHOD__ == 0 || __FLT_EVAL_METHOD__ == 1)
+#if defined _M_X64 || ( defined __FLT_EVAL_METHOD__ && defined __SSE2_MATH__ && \
+                        (__FLT_EVAL_METHOD__ == 0 || __FLT_EVAL_METHOD__ == 1) )
 #  define CGAL_SAFE_SSE2
 #  include <xmmintrin.h>
 #endif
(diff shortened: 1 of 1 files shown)


lrineau:
@afabri and I tested that code on Windows, and it speed up a code using Epeck, with a lot of predicates solved by intervals filter, in loops, from 5s to 1.7s.

mglisse (approved):
Looks ok to me. I am surprised that it gives such a big performance boost, but since it goes in the right direction, that's good. Are there still big perf differences between VS and gcc/clang for this kind of code? VS should still be a bit slower.

mglisse:
There may be other compilers that define `_M_X64` (Intel? clang in some windows compatibility mode?). Do they offer the same guarantees?

lrineau:
> There may be other compilers that define `_M_X64` (Intel? clang in some windows compatibility mode?). Do they offer the same guarantees?

Which guarantees? Anyway, I would say that, if they define `_M_X64`, then it would be a bug if they do not comply to MSVC ABI and "guarantees" (whatever you call with that word).

mglisse:
> Which guarantees?

Use only SSE and never x87 for operations on double.

lrineau:
> > Which guarantees?
> 
> Use only SSE and never x87 for operations on double.

That is part of the ABI. Other compilers should comply.

maxGimeno:

PR #5007, opened 2020-09-21
Fix overload of remove_invalid_polygons_in_polygon_soup

mhsaar:
##  Summary of Changes

Fix name in overload of `remove_invalid_polygons_in_polygon_soup`.

Also removed points as argument from the internal implementation. 

Should it be `remove_invalid_polygons_in_polygon_soup`, `remove_degenerate_polygons_in_polygon_soup`, or both?

It certainly needs fixing as `remove_degenerate_polygons_in_polygon_soup` doesn't seem to exist.

## Release Management

* Issue(s) solved (if any): -


MaelRL:
Thanks, I have opened #5070 to back port the fix to earlier releases.

I have only removed the meaningless overload in that PR because I have a local branch where the function is enhanced to treat other bad configurations (and points are needed).

PR #5023, opened 2020-09-25
Use property map in std::nth_element call

### LINKED ISSUE #5008: Compilation error when using `CGAL::remove_outliers` with `CGAL::named_parameters::point_map`
Looks like the changes introduced by commit https://github.com/CGAL/cgal/commit/82479e780bdaeb4c7257a3a1593f2e68e638d24e forbid the use of `CGAL::named_parameters::point_map` as a parameter of `CGAL::remove_outliers`.
The compilation produces an error at the call to `nth_elements` on line 257 of `remove_outliers.h`.

I'm using a `std::vector<std::pair<Point_3, Vector3>>` and I get the following error with gcc: "error: no match for ‘operator<’ (operand types are ‘const CGAL::Vector_3<CGAL::Epick>’ and ‘const CGAL::Vector_3<CGAL::Epick>’)"

sloriot:
Fixes #5008

### DIFF
--- a/Point_set_processing_3/include/CGAL/remove_outliers.h
+++ b/Point_set_processing_3/include/CGAL/remove_outliers.h
@@ -257,9 +257,9 @@ remove_outliers(
     std::nth_element (f2r,
                       sorted_points.begin() + first_index_to_remove,
                       sorted_points.end(),
-                      [&point_map](const std::pair<FT, value_type>& v1, const std::pair<FT, value_type>& v2)
+                      [](const std::pair<FT, value_type>& v1, const std::pair<FT, value_type>& v2)
                       {
-                        return get(point_map, v1.second) < get(point_map, v2.second);
+                        return v1.first<v2.first;
                       });
     f2r = sorted_points.begin() + first_index_to_remove;
   }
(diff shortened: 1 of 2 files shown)


Point_set_processing_3/include/CGAL/remove_outliers.h
```diff
@@ -256,7 +256,11 @@ remove_outliers(
   {
     std::nth_element (f2r,
                       sorted_points.begin() + first_index_to_remove,
-                      sorted_points.end());
+                      sorted_points.end(),
+                      [&point_map](const std::pair<FT, value_type>& v1, const std::pair<FT, value_type>& v2)
+                      {
+                        return get(point_map, v1.second) < get(point_map, v2.second);
+                      });
```
sgiraudot, line 263: Actually the `nth_element` is supposed to be done on the computed distance, so it should `return v1.first < v2.second`. I don't understand how I did not see that this code was obviously wrong (on master) and did not detect errors with it before…

sgiraudot (approved):

cjamin:
Thank you for the fix, it works!

maxGimeno:

PR #5035, opened 2020-09-29
Poisson Reconstruction: Fix thread safety

sgiraudot:
## Summary of Changes

Poisson reconstruction had a couple of problems when used with parallel `Mesh_3` as some structures were not thread-safe:
- a lazy cache mechanism to store barycentric coordinates of cells
- a global cell hint for fast locate

This PR fixes these 2 problems using atomic variables. It also adds a test for this specific case. This test triggers and assertion fail on master and succeeds on this branch.

## Release Management

* Affected package(s): Poisson Surface Reconstruction
* Issue(s) solved (if any): [cgal-discuss:  make_mesh_3 assertion violation in Triangulation_ds_cell_base_3](http://cgal-discuss.949826.n4.nabble.com/make-mesh-3-assertion-violation-in-Triangulation-ds-cell-base-3-td4665230.html)

### DIFF
--- a/Poisson_surface_reconstruction_3/include/CGAL/Poisson_reconstruction_function.h
+++ b/Poisson_surface_reconstruction_3/include/CGAL/Poisson_reconstruction_function.h
@@ -225,18 +225,33 @@ private:
     std::array<double, 9> m_bary;
   public:
     Cached_bary_coord() : m_state (UNINITIALIZED) { }
-    Cached_bary_coord(const Cached_bary_coord& other)
-      : m_state (other.m_state.load()) // not atomic
-    { }
 
-    Cached_bary_coord& operator= (const Cached_bary_coord& other)
+    // Copy operator to satisfy vector, shouldn't be used
+    Cached_bary_coord(const Cached_bary_coord&)
     {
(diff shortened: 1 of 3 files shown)


maxGimeno:
Note for integration : in 5.0 and in master, the cmake command for linking with TBB is not the same.

lrineau:
> Note for integration : in 5.0 and in master, the cmake command for linking with TBB is not the same.

Should not we have a backward compatibility layer? I have the same issue with external code, when I want to test it with various CGAL versions.

sgiraudot:
Well, the old function `CGAL_target_use_TBB` is still available, so the code won't break in master, which sounds like backward compatibility to me. It will raise a deprecation warning though. Do you suggest we remove that warning?

maxGimeno:
@sgiraudot I got an error message, not a warning, when trying to merge into master. Maybe an `include` is missing to use `CGAL_target_use_TBB`.

maxGimeno:

lrineau:
@sgiraudot Now you can announce in the discussion on cgal-discuss, that the bug is fixed.

PR #5036, opened 2020-09-29
Basic_viewer: Minor fixes

maxGimeno:
## Summary of Changes

 - Use O instead of < as it doesn't work on some keyboards (mine, for example)
 - Remove the Camera state from the settings saved, as it is supposed to be overridden by the recenter, and sometimes, if the said state is broken enough, the user is unable to visualize anything even after re-launching the app.


maxGimeno:

PR #5040, opened 2020-09-30
STL Extensions: Small List

sgiraudot:
## Summary of Changes

This PR replaces https://github.com/CGAL/cgal/pull/4921: the goal is still the same, having an equivalent of `boost::container::small_vector` but for a `std::list`-like structure. A small memory chunk is statically allocated, which can improve performances in cases where users know at compile-time that their list will rarely exceed a certain size.

Instead of a general allocator which turned out to be an overly-complicated and unsatisfying solution due to the constraints of the STL Allocator API, this is "just" a `Small_list` data structure. It provides pretty much the same gains of performances while being less general.

I added a test to STL Extension package, and the testsuites of Surface Sweep, Arrangement and Boolean Operations (which all use it) all pass without errors. Not all methods of `std::list` were implemented, only those necessary for surface sweep, but the other ones can always be added later if someone sees some interest in them.

## Release Management

* Affected package(s): STL Extensions, Surface Sweep + packages that rely on it (Arrangement, etc.)

### DIFF
--- a/STL_Extension/include/CGAL/Small_list.h
+++ b/STL_Extension/include/CGAL/Small_list.h
@@ -226,7 +226,10 @@ public:
     if (next == m_first)
       push_front(t);
     else if (next == nullptr)
+    {
       push_back(t);
+      return iterator(m_last, m_last);
+    }
     else
     {
       Node* n = m_pool.allocate();
@@ -238,7 +241,7 @@ public:
       ++ m_size;
     }
 
-    return next->prev();
+    return iterator(m_last, next->prev());
   }
 
   iterator erase(iterator first, iterator last)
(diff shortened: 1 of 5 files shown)


lrineau (changes requested):
Like I said in https://github.com/CGAL/cgal/pull/4921#issuecomment-700532486, I strongly oppose to having new allocators or containers in CGAL when valid alternatives exist. That would be the same mistake as what CGAL did 25 years ago.

Boost Container from version 1.68 has all the API of PMR: https://www.boost.org/doc/libs/1_68_0/doc/html/container/cpp_conformance.html#container.cpp_conformance.polymorphic_memory_resources

You are reimplementing `boost::pmr::list` that uses a `boost::pmr::monotonic_buffer_resource`. What would be the benefit of trying to implement our own similar classes in CGAL?

sgiraudot:
None, but as I said in my answer to your comment, I did not find what I was looking for in Boost. I'll have a look at this monotonic buffer resource.

sgiraudot:
After more performance tests, this feature does not seem to provide significant enhancement for surface sweep (I have also tested the monotonic buffer of Boost), so I'm cancelling this PR.

PR #5042, opened 2020-09-30
Viewer zoom to rectangle

maxGimeno:
## Summary of Changes

Add a "real zoom" as new feature for the basic viewer and demo. It changes te field of view in order to emulate an optical zoom (instead of simply translating the camera closer as we currently do). It allows to bypass the clipping problem (when the camera is too close of an item and it goes through it).

It is triggered by maintaining the Z key and using the wheel.

## Release Management

### DIFF
--- a/GraphicsView/include/CGAL/Qt/camera_impl.h
+++ b/GraphicsView/include/CGAL/Qt/camera_impl.h
@@ -895,13 +895,17 @@ Vec Camera::pointUnderPixel(const QPoint &pixel, bool &found) const {
 
 /*! Moves the Camera so that the entire scene is visible.
 
- Simply calls fitSphere() on a sphere defined by sceneCenter() and
- sceneRadius().
+ Calls fitSphere() on a sphere defined by sceneCenter() and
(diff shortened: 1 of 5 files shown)


lrineau:
Is there a way to reset the zoom factor? Does the recenter_scene method reset also the zoom?

maxGimeno:
Just unzoom as far as you can, the minimum zoom factor is the actual value of the FOV (45 degrees)

maxGimeno:
build:test

maxGimeno:
build:test

maxGimeno:
build:test

maxGimeno:

lrineau:
It seems that that new zoom feature zooms to the center of the screen. But when one want to zoom, it might be on something that is not exactly at the center of the screen. I think that would easier to use if at the same time the camera was slightly rotated to focus on where the mouse pointer is.

That would be something similar to z+click, but with the "real zoom" instead of the movement of the camera. Maybe 'Z'+click could be a real zoom, and 'z'+click would be the zoom-using-movement.

PR #5045, opened 2020-10-01
Add missing overload

sloriot:


lrineau:
Note that I would never classify the addition of one overload in an overload-set as "trivial". That can cause of lot of different issues. :smile:

maxGimeno:

PR #5047, opened 2020-10-01
Add a way to not remove properties after clearing the mesh

sloriot:
Small feature: [Features/Small Features/SM more clear functions](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/SM_more_clear_functions), pre-approved by @afabri, 15 October 2020 (CEST).

Add the following functions:
- [`clear_without_removing_property_maps()`](https://cgal.github.io/5047/doc_v5/Surface_mesh/classCGAL_1_1Surface__mesh.html#aad000a07a5ada30536f194b28b59d111)
- [`remove_property_maps<Index_type>()`](https://cgal.github.io/5047/doc_v5/Surface_mesh/classCGAL_1_1Surface__mesh.html#a2a3dd8c01f7fba7b640d85bfd1c41d90)
- [`remove_all_property_maps()`](https://cgal.github.io/5047/doc_v5/Surface_mesh/classCGAL_1_1Surface__mesh.html#a5696da09300f3d0eafed117668bb3bec)

Update doc:
- [`clear()`](https://cgal.github.io/5047/doc_v5/Surface_mesh/classCGAL_1_1Surface__mesh.html#a247d4ad3e6b106ae22e5306203812642)

User manual update: [here](https://cgal.github.io/5047/doc_v5/Surface_mesh/index.html#title7)

In passing, note that I fixed the fact that the `vpm` after a call to `clear()` changes, so it is safe to keep a reference on it.

Follow up of #4709


sloriot:
Note that in my patch I kept `clear()` on purpose because there is a c++ trick for the free function `CGAL::clear()`.

sloriot:
I noticed another issue I have in a generic function that calls `clear(tm)` but uses the `vpm` passes as parameter. With the new behavior of `Surface_mesh` the `vpm` becomes invalid.

lrineau:
Boolean parameters in API are usually a bad idea, and an anti-pattern. I agree there should be a documented function name equivalent to `clear(true)`.

sloriot:
I've added `remove_all_simplices()`, the rational being that there are `remove_vertex()`, `remove_edge()`, ...
I've also updated the test to check that the property is either kept or re-created.

sloriot:
build:doc_v1

sloriot:
Here is the link to the doc:
https://cgal.github.io/5047/doc_v1/Surface_mesh/classCGAL_1_1Surface__mesh.html#aee6a0d0c8d0c827f33003747fe9abda9
@afabri do you agree with that addition?

lrineau:
> Here is the link to the doc:
> [cgal.github.io/5047/doc_v1/Surface_mesh/classCGAL_1_1Surface__mesh.html#aee6a0d0c8d0c827f33003747fe9abda9](https://cgal.github.io/5047/doc_v1/Surface_mesh/classCGAL_1_1Surface__mesh.html#aee6a0d0c8d0c827f33003747fe9abda9)
> @afabri do you agree with that addition?

> Collects garbage but keeps all properties. 

I find that ambiguous: does it mean that the list of properties is preserved, or is it also the contents (the arrays) of the properties that are preserved?

sloriot:
properties are kept not their values. That's the same logic when you remove an element, upon addition of a new element (be it from garbage or totally new, the property is init to the default value).

sloriot:
Of course something could be added but it should be general, not specific to this function.

afabri:
I do not really like the name.  It could be `reinit()` or `reinitialize()`.  It could also be an optional `bool` parameter of the function `clear()`.

afabri:
I just saw that my `bool`  is a bad idea.

sloriot:
I like `remove_all_simplices()` because it describes what it does. `reinit()` sounds like`clear()` to me. `remove_all()` could work if you don't like long function names.

sloriot:
well `clear()` sounds like `remove_all_properties()` and `reinit()` like it's put in the same state as when created.

afabri:
On the phone with Sebastien we realized that there is a function `remove_property(..)` but not a function `remove_properties()` which removes all of them (all but the connectivity and the point property).       What about adding such a function and to add `clear_without_removing_properties()`?   It has a lengthy name, and maybe someone can improve on the name, but the idea of the name is that it explicitly states what it does, instead of making me deduce from `remove_all_simplices()`, that this does clear the properties of the simplices (because there are 0 simplices) without removing the properties.

sloriot:
The existing function is `remove_property_map()` so @afabri 's proposal is to add:
- `remove_property_maps()`
- `clear_without_removing_property_maps()`

I still prefer `remove_all_simplices()` but we need to move forward ...

lrineau:
> It could also be an optional `bool` parameter of the function `clear()`.

We [already said that is a bad idea (https://github.com/CGAL/cgal/pull/5047#issuecomment-702840677): boolean parameters are a bad idea, and a well know antipattern.

lrineau:
> The existing function is `remove_property_map()` so @afabri 's proposal is to add:
> 
>     * `remove_property_maps()`
>     * `clear_without_removing_property_maps()`
> 
> I still prefer `remove_all_simplices()` but we need to move forward ...

Sounds good.

sloriot:
build:doc_v2

Installation/CHANGES.md
```diff
@@ -32,6 +32,12 @@ Release date: December 2020
       As explained above, the code is backward compatible. However, it is recommended
       that all calls to `decompose()` are fixed to use the new interface.
 
+### [Surface Mesh](https://doc.cgal.org/5.2/Manual/packages.html#PkgSurfaceMesh)
+
+-   Added the function `clear_without_removing_property_maps()` to clear a mesh but keep all the created property maps added.
+-   Added the functions `remove_property_maps()` and `remove_all_property_maps()` to remove all added property maps by index type or all of them respectively.
```
lrineau, line 38: `remove_property_maps` is a function template, and not a function. I suggest to explicit that using:

```suggestion
-   Added the functions `remove_property_maps<Index_type>()` and `remove_all_property_maps()` to remove all added property maps by index type or all of them respectively.
```

(I want to say, again, that it is a bad idea to discrimine on index types. It should be allowed that all index types are identical (for example `int`)/ Instead, it should be different names for functions, or using tags to discriminate on element types.)

afabri: The different index types were introduced  for more type safety.

sloriot: @lrineau this is a pure `Surface_mesh` function and the index types are different. Index type here mean simplex type.

sloriot:
- [`clear_without_removing_property_maps()`](https://cgal.github.io/5047/doc_v2/Surface_mesh/classCGAL_1_1Surface__mesh.html#aad000a07a5ada30536f194b28b59d111)
- [`clear()`](https://cgal.github.io/5047/doc_v2/Surface_mesh/classCGAL_1_1Surface__mesh.html#a247d4ad3e6b106ae22e5306203812642)
- [`remove_property_maps<Index_type>()`](https://cgal.github.io/5047/doc_v2/Surface_mesh/classCGAL_1_1Surface__mesh.html#a2a3dd8c01f7fba7b640d85bfd1c41d90)
- [`remove_all_property_maps()`](https://cgal.github.io/5047/doc_v2/Surface_mesh/classCGAL_1_1Surface__mesh.html#a5696da09300f3d0eafed117668bb3bec)

In passing, note that I fixed the fact that the `vpm` after a call to `clear()` changes, so it is safe to keep a reference on it.

sloriot:
build:doc_v3

sloriot:
build:doc_v4

maxGimeno:
build:doc_v5

maxGimeno:
I suspect this branch for https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-31/Polygon_mesh_processing/TestReport_lrineau_Ubuntu-Latest-CXX11.gz and https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.2-Ic-31/BGL/TestReport_lrineau_Ubuntu-Latest-CXX11.gz

sloriot:
Tests are fine now locally

maxGimeno:

sloriot:
@lrineau pre-approval was 15 days ago

sloriot:
Wrong line sorry was pre-approved on the 15th.

citystrawman:
> I've added `remove_all_simplices()`, the rational being that there are `remove_vertex()`, `remove_edge()`, ... I've also updated the test to check that the property is either kept or re-created.

Hi, I have a question about `remove_vertex()`, if I execute `mesh1.remove_vertex(v)`, does this just remove vertex v? for example as below
image
if I remove p0, then the 6 triangles are also deleted and the mesh may contains a hexagon?

sloriot:
It 

> > I've added `remove_all_simplices()`, the rational being that there are `remove_vertex()`, `remove_edge()`, ... I've also updated the test to check that the property is either kept or re-created.
> 
> Hi, I have a question about `remove_vertex()`, if I execute `mesh1.remove_vertex(v)`, does this just remove vertex v? for example as below image if I remove p0, then the 6 triangles are also deleted and the mesh may contains a hexagon?

No, its a low level operation, it only remove the vertex from the mesh and leave the rest unchanged (halfedges then have invalid reference to the removed vertex).

You have to use Euler operations like [remove_center_vertex()](https://doc.cgal.org/latest/BGL/group__PkgBGLEulerOperations.html#ga8ec295610396e837258997c435b7a75f)

citystrawman:
> It
> 
> > > I've added `remove_all_simplices()`, the rational being that there are `remove_vertex()`, `remove_edge()`, ... I've also updated the test to check that the property is either kept or re-created.
> > 
> > 
> > Hi, I have a question about `remove_vertex()`, if I execute `mesh1.remove_vertex(v)`, does this just remove vertex v? for example as below image if I remove p0, then the 6 triangles are also deleted and the mesh may contains a hexagon?
> 
> No, its a low level operation, it only remove the vertex from the mesh and leave the rest unchanged (halfedges then have invalid reference to the removed vertex).
> 
> You have to use Euler operations like [remove_center_vertex()](https://doc.cgal.org/latest/BGL/group__PkgBGLEulerOperations.html#ga8ec295610396e837258997c435b7a75f)

So  method `remove_vertex()` must be used with other operations that keeps the validity of halfedge structure , if not,  executing mesh.collect_garbage()  will report error?

sloriot:
you would need to collect all incident faces and halfedges and call `remove_face()`, `remove_edge()` and also update the incident face on the boundary of the link. Otherwise `is_valid_polygon_mesh()` will return `false`

citystrawman:
> you would need to collect all incident faces and halfedges and call `remove_face()`, `remove_edge()` and also update the incident face on the boundary of the link. Otherwise `is_valid_polygon_mesh()` will return `false`

does "update the incident face" refer to` add_face()` method?

sloriot:
no set the incident face to the halfedge to the null_face so that it is on the boundary. You also need to link border halfedges.
You should use [remove_center_vertex()](https://doc.cgal.org/latest/BGL/group__PkgBGLEulerOperations.html#ga8ec295610396e837258997c435b7a75f) that is already doing that

PR #5062, opened 2020-10-08
Fix move vertex on the boundary

sloriot:


maxGimeno:

PR #5063, opened 2020-10-08
CMake modernization

### LINKED ISSUE #4815: Finish the modernization of CGAL CMake scripts
## What is modern CMake?

Since CMake-2.8, CMake has introduced a new way to create express dependencies between targets and external libraries, using "properties" and "imported targets", instead of relying of variables.

The web seems to consider [that gist](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1) to be a good introduction to what is "modern CMake".

Two enlightening talks in 2017:
 - [Mathieu Ropert “Using Modern CMake Patterns to Enforce a Good Modular Design” at CppCon 2017](https://www.youtube.com/watch?v=eC9-iRN2b04),
 - [Daniel Pfeifer “Effective CMake" at C++Now 2017](https://www.youtube.com/watch?v=bsXLMQ6WgIk).

## Where are we, now?

In CGAL-5.1, most of the CMake code of CGAL uses modern CMake. 
  - the PR https://github.com/CGAL/cgal/pull/1436 (4.12) was the beginning of using modern CMake in CGAL: the CGAL libraries were exported, even in header-only,
  - since PR https://github.com/CGAL/cgal/pull/4636 (5.1), our external dependencies are now used through pseudo-targets instead of using functions or variables.

So, essentially, CGAL CMake code already uses modern CMake. ...But not everywhere.

## What should be done?

- [x] Decide which use-cases we want to support. For example, do we still want to support non-header-only CGAL libraries? See also [that (private) Gitlab issue about the subject](https://gitlab.com/cgal/GeometryFactory/-/issues/71).
- [x] Add an automatic testsuite of those use-cases: see issue #4826.
- Remove legacy code:
  - [x] Remove everything about preconfigured 3rd party libraries. Their semantic is fuzzy, and we no longer document them.
  - [x] Remove all variable `WITH_*` : their semantic is also fuzzy.
- Simplify the code so that it is sufficient for the use-cases we support, and nothing more. Ideally, the code should be as if it was written from scratch using only modern CMake.
  - [x] remove non-header-only support
  - [ ] check that there is no call to general functions of the type `add_definitions` or `link_libraries`, only the target_xx versions
  - [ ] replace the PUBLIC by PRIVATE or INTERFACE where it is possible
  - [ ]  dependencies for 3rdPartyLibs should be declared in the library Finders, not in the CMakeLists that call the 3rdPartyLibs
  - [ ] avoid variables in target_link_libraries(), they should be targets in the form Foo::Bar most of the time

maxGimeno:
## Summary of Changes
Remove most of WITH_XXX options, remove conditional HEADER_ONLY code, attemps to replace some file(GLOB) calls and remove conditional CMP00XX code from Installation cmake scripts.
## Release Management

* Issue(s) solved (if any): fix #4815


lrineau:
I think everything from the `src/` directory should disappear, once we stop supporting the non-header-only version of CGAL: in header-only, nothing can be compiled.

Installation/CMakeLists.txt
```diff
@@ -8,27 +8,16 @@ if(NOT PROJECT_NAME)
...
 #
-#                                    -= HEADER ONLY =-
+# Fix a bug: `try_compile` should use `CMAKE_EXE_LINKER_FLAGS`.
+# That variable can contain important flags like
 #
-#--------------------------------------------------------------------------------------------------
-option(CGAL_HEADER_ONLY "Enable header-only mode of CGAL (no compilation of CGAL libraries)" TRUE)
+#     -static-libstdc++ -static-libgcc
+#
+cmake_policy(SET CMP0056 NEW)
```
lrineau, line 18: I think for now one, we should require CMake>=3.12, and use `cmake_minimum_required(VERSION 3.12...3.18)` everywhere.

That will remove all policies before `CMP0076`.

Installation/CMakeLists.txt
```diff
@@ -526,7 +510,7 @@ message("== Detect external libraries ==")
 #  External libs configured when Qt5 lib of cgal are required
 #  Coin is used in KDS, but no FindCoin or FindCOIN exists
 #  There exists FindIPE, FindMKL, but they are only used to support supporting libs
-list (INSERT CGAL_SUPPORTING_3RD_PARTY_LIBRARIES 0 GMP MPFR ZLIB OpenGL LEDA MPFI RS RS3 OpenNL Eigen3 ESBTL Coin3D NTL IPE)
+list (INSERT CGAL_SUPPORTING_3RD_PARTY_LIBRARIES 0 LEDA )
```
lrineau, line 513: That variable `CGAL_SUPPORTING_3RD_PARTY_LIBRARIES` should disappear. I understand why we still use `CGAL_ESSENTIAL_3RD_PARTY_LIBRARIES`, but `CGAL_SUPPORTING_3RD_PARTY_LIBRARIES` should be useless, now.

Installation/CMakeLists.txt
```diff
@@ -1178,38 +1089,36 @@ You must disable CGAL_ENABLE_CHECK_HEADERS.")
...
-        else() # do not recurse
-          file(GLOB ${package}_HEADERS
+          file(GLOB_RECURSE ${package}_HEADERS #don't use a hard coded list because the ones for the documented headers is enough.
             RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include"
             "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
+        else()
+        file(GLOB ${package}_HEADERS
+                    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include"
+                    "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
         endif()
```
lrineau, line 1115: I do not remember: what is the mean of that variable `has_list_of_documented_headers`?

Installation/CMakeLists.txt
```diff
@@ -1178,38 +1089,36 @@ You must disable CGAL_ENABLE_CHECK_HEADERS.")
...
-            "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
-        else() # do not recurse
-          file(GLOB ${package}_HEADERS
+          file(GLOB_RECURSE ${package}_HEADERS #don't use a hard coded list because the ones for the documented headers is enough.
             RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include"
             "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
+        else()
+        file(GLOB ${package}_HEADERS
+                    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include"
+                    "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
```
lrineau, line 1114:
```suggestion
          file(GLOB ${package}_HEADERS
                    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include"
                    "${CMAKE_CURRENT_SOURCE_DIR}/../${package}/include/CGAL/*.h*")
```

lrineau (approved):
The modifications are good.

I think we can simplify the code even further, but it can be in a second step.

Installation/CMakeLists.txt
maxGimeno: it means that the awk research reported things, which should always be the case.

lrineau:
@maxGimeno It is time to refresh this PR and work on its integration. That should be the next big change to merge just after CGAL I/O.

maxGimeno:
```
CMake Error at CMakeLists.txt:11 (add_executable):
  Target "link_to_CGAL_Core" links to target "CGAL::CGAL_Core" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
  CMakeLists.txt:59 (create_link_to_program)
  ```

maxGimeno:
CGAL_Qt5 is not found and all demos are disabled in the testsuite

maxGimeno:

PR #5069, opened 2020-10-09
Polyhedron_demo: Merge hole_filling and hole_filling_polylines plugins

maxGimeno:
## Summary of Changes

Both plugins are very similar and therefore are now merged into one.


maxGimeno:

PR #5077, opened 2020-10-14
Add const qualifier for getters in MCF skeletonization

FabienPean:
## Summary of Changes
Add const qualifier for getters in class Mean_curvature_flow_skeletonization

## Release Management
* Affected package(s): Triangulated Surface Mesh Skeletonization

### DIFF
--- a/Surface_mesh_skeletonization/include/CGAL/Mean_curvature_flow_skeletonization.h
+++ b/Surface_mesh_skeletonization/include/CGAL/Mean_curvature_flow_skeletonization.h
@@ -867,9 +867,6 @@ private:
     m_vertex_id_count = static_cast<int>(num_vertices(m_tmesh));
     m_max_id = m_vertex_id_count;
 
-    if (m_is_medially_centered)
-      compute_voronoi_pole();
-
     init_args();
   }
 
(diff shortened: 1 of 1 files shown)


FabienPean:
Another potential bug or dead code

The private function `void init(const TriangleMesh& tmesh)` is only called in the constructor, therefore this part of the method
```c++
void init(const TriangleMesh& tmesh) 
{
    /*...*/
    if (m_is_medially_centered)
      compute_voronoi_pole();

    init_args();
```
will never compute the voronoi poles because m_is_medially_centered is always false at this point. (default-constructed)
Similarly, if it were placed after the `init_args()`, poles would always be computed because m_is_medially_centered is set to true in `init_args()`.
It could be a typo and meant to be `if(!m_are_poles_computed)` but then it would again always compute the poles at this point anyway.

Depending on the original intention, either voronoi poles should always be computed (`if(m_is_medially_centered)`unneeded) or the block if+statement should be removed because it is dead code currently.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2-Ic-33.shtml.

PR #5085, opened 2020-10-17
Fix warnings with Arrangement Demo

theartful:
## Summary of Changes

Initialize missing member in `_Linear_object_cached`, and workaround a faulty gcc warning with boost variant.

I reproduced the warnings locally with boost 1.69, 1.70, and couldn't with boost 1.71 building with `-m32` flag. Trunk version of gcc pointed at this line as the source of the warning:
https://github.com/CGAL/cgal/blob/b4a536047d098d586a89982146fd094dd97b41aa/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h#L236-L237
and at boost::variant move constructor. Avoiding boost::variant move constructor removed the warnings with gcc trunk and gcc 10.

I checked the constructors of the "maybe uninitialized" classes, and couldn't find any uninitialized members except for `_Linear_object_cached` (which did not reduce the number of warnings), so I think these warnings are false positives.

The warnings are not related to the demo, I reproduced them using a simple example that computes the overlay of two arrangements with history, so I think the changes to the package are warranted.

## Release Management

* License and copyright ownership: the license used by CGAL

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
@@ -233,8 +233,8 @@ public:
         }
         // The current intersection object is an intersection point:
         // Copy it as is.
-        const Intersection_point* ip = boost::get<Intersection_point>(&item);
-        *oi++ = Intersection_result(*ip);
+        Intersection_result ip = *boost::get<Intersection_point>(&item);
+        *oi++ = ip;
       }
 
       return oi;
--- a/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
@@ -190,20 +190,29 @@ public:
   private:
     const Base_traits_2& m_base;
 
-    template <typename Wrapper, typename OutputIterator, typename T>
+    typedef std::pair<Point_2, Multiplicity>          Intersection_point;
+    typedef boost::variant<Intersection_point, X_monotone_curve_2>
+                                                      Intersection_result;
+    typedef boost::variant<Intersection_point, Base_x_monotone_curve_2>
+                                                      Intersection_base_result;
+
+    template <typename OutputIterator, typename T>
     inline std::enable_if_t<std::is_assignable<
         decltype(*std::declval<OutputIterator>()), T&&>::value>
-    assign(OutputIterator& oi, T&& x) const
+    assign_iterator(OutputIterator oi, T&& x) const
     {
+      // avoid maybe-uninitialized warnings with boost <= 1.70.0
+      // when move constructing Intersection_result(variant)
       (*oi = {}) = std::forward<T>(x);
     }
 
-    template <typename Wrapper, typename OutputIterator, typename T>
+    template <typename OutputIterator, typename T>
     inline std::enable_if_t<!std::is_assignable<
         decltype(*std::declval<OutputIterator>()), T&&>::value>
-    assign(OutputIterator& oi, T&& x) const
+    assign_iterator(OutputIterator oi, T&& x) const
     {
-      *oi = Wrapper{std::forward<T>(x)};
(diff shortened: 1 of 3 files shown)


Arrangement_on_surface_2/include/CGAL/Arr_curve_data_traits_2.h
```diff
@@ -233,8 +233,8 @@ class Arr_curve_data_traits_2 : public Traits_ {
         }
         // The current intersection object is an intersection point:
         // Copy it as is.
-        const Intersection_point* ip = boost::get<Intersection_point>(&item);
-        *oi++ = Intersection_result(*ip);
+        Intersection_result ip = *boost::get<Intersection_point>(&item);
+        *oi++ = ip;
```
theartful, line 237: Unfortunately can't do something like
```C++
*oi++ =*boost::get<Intersection_point>(&item);
```
Since output iterator can be of type `std::variant` or `CGAL::Object`, and there is no public constructor for `CGAL::Object` that would accept arbitrary type, so we have to construct the `Intersection_result` first (which is a variant).

theartful:
As far as I can tell this is a false positive. I reproduced the warning using this simple example
```C++
#include <CGAL/Cartesian.h>
#include <CGAL/Exact_rational.h>
#include <boost/variant.hpp>
#include <variant>
#include <stdlib.h>

using namespace CGAL;

template <typename Kernel>
class _Segment_cached_2 {
public:
  typedef typename Kernel::Line_2 Line_2;
  typedef typename Kernel::Segment_2 Segment_2;
  typedef typename Kernel::Point_2 Point_2;

  _Segment_cached_2();

protected:
  Line_2 m_l;
  Point_2 m_ps;
  Point_2 m_pt;
  bool m_is_directed_right;
  bool m_is_vert;
  bool m_is_degen;
};

template <typename Kernel>
_Segment_cached_2<Kernel>::_Segment_cached_2()
    : m_l(), m_ps(), m_pt(), m_is_directed_right(false), m_is_vert(false),
      m_is_degen(true) {}

typedef CGAL::Cartesian<CGAL::Exact_rational> Kernel;
typedef _Segment_cached_2<Kernel> X_monotone_curve_2;

int main() {
  std::vector<boost::variant<int, X_monotone_curve_2>> vec;
  auto oi = std::back_inserter(vec);
  int x = rand();
  *oi++ = boost::variant<int, X_monotone_curve_2>(x);
}
```
with compilation command
```
g++-10 -I /home/theartful/Downloads/boost_1_70_0/ -O3 -DCGAL_USE_CORE=1 -DCGAL_USE_GMPXX=1 -I/home/theartful/Programming/cgal/target/include -DCGAL_NDEBUG -Wmaybe-uninitialized -Wuninitialized -fPIC -std=gnu++14 example.cpp -o example -lgmp -lmpfr
```
(cgal and boost has to be included by -I not -isystem) and got warnings
```
In file included from /home/theartful/Programming/cgal/target/include/CGAL/Cartesian/Weighted_point_2.h:20,
                 from /home/theartful/Programming/cgal/target/include/CGAL/Cartesian/Cartesian_base.h:25,
                 from /home/theartful/Programming/cgal/target/include/CGAL/Cartesian.h:20,
                 from overlay.cpp:1:
/home/theartful/Programming/cgal/target/include/CGAL/Handle_for.h: In function ‘int main()’:
/home/theartful/Programming/cgal/target/include/CGAL/Handle_for.h:107:18: warning: ‘*((void*)(&<anonymous>)+16).CGAL::Handle_for<std::array<__gmp_expr<__mpq_struct [1], __mpq_struct [1]>, 2>, std::allocator<std::array<__gmp_expr<__mpq_struct [1], __mpq_struct [1]>, 2> > >::ptr_’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  107 |         ptr_ = h.ptr_;
```
The solution in the PR
```C++
Intersection_result ip = *boost::get<Intersection_point>(&item);
*oi++ = ip;
```
removes these specific warnings, but it is not robust, ugly, and slightly slower. The warnings don't occur with clang, newer versions of boost, or with "std::variant".

PR #5091, opened 2020-10-20
[Arrangement] Fix polycurve overlap intersection

sgiraudot:
## Summary of Changes

When computing the intersection of two polycurves, there is a specific case that creates a wrong output: when 2 polycurves oriented right-to-left overlap, the output is wrongly oriented left-to-right. This makes boolean operations fail if the polyline traits is used instead segment traits.

This PR fixes this problem by inverting the output overlap polycurve in both curves are detected as right-to-left oriented. A test is also added to check this.

## Release Management

* Affected package(s): Arrangement and derivated packages (Boolean Operations, etc.)
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/4998

PR #5050, opened 2020-10-02
2D Boolean Set Operations: Improve Dcoumentation of Free Function

efifogel:
## Summary of Changes

1. Added missing references to `oriented_side(Point_2, ....)`
2. Fixed references to all BSO_2 functions.
3. Fixed a bug in the `Arr_polycurve_traits_2::Intersect_2` functor. (This is unrelated to items (1) and (2) above, but it's only a one-liner.

It turns out that the `oriented_side()` overloaded functions that accept a point and a polygon were not documented.
This PR adds the missing functions. 
White at it, I've fixed the documentation of all other functions.
For example, we typically do not put too much emphasize of documenting a traits optional parameter.
I've decided to do it after all, because there are related preconditions that the user should be informed bout when passing a traits.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #4998 and more
* Feature/Small Feature (if any): [`Features/Small_Features/Improve_Dcoumentation_of_2D_Boolean_Set_Operations_Free_Functions`](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Improve_Dcoumentation_of_2D_Boolean_Set_Operations_Free_Functions)
* Pre-approved by @sloriot: 27 January 2021
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership:

PR #5094, opened 2020-10-20
[Small Feature] Boolean Operations: Use polyline-based GPS

sgiraudot:
## Rationale

The Boolean Operations in CGAL mainly rely on the `CGAL::General_polygon_2` and `CGAL::General_polygon_set_2` classes. Benchmarks show that using a decomposition of the polygons into a set of X-monotone polylines instead of using the raw segments (and templating these classes with `Arr_polyline_traits_2` instead of `Arr_segment_traits_2`) significantly speeds up the computation of these boolean operations (up to 80 %).

The only cases where doing so is counter-productive are cases where 2 X-monotone polylines intersect each other _many_ times. If we generate 2 polygons containing such polylines and compute the union, here are the timings depending on the number of time the 2 polylines intersect:

| # Intersections  | Segment traits (ms) | Polylines traits (ms) | Speed-up |
| ---------------- | ------------------- | --------------------- | -------- |
| 10               | 0.37                | 0.12                  | x 3.1    |
| 50               | 1.47                | 0.32                  | x 4.6    |
| 100              | 2.70                | 0.59                  | x 4.6    |
| 500              | 16.7                | 7.67                  | x 2.2    |
| 1000             | 32.7                | 23.4                  | x 1.4    |
| 5000             | 190                 | 639                   | x 0.3    |
| 10000            | 392                 | 2912                  | x 0.1    |

In practise, using polyline traits only becomes counter-productive if the number of intersections *between two X-monotone* polylines exceeds 1000, which seems unlikely to happen with real-world data sets.

Hence, this small feature makes polyline traits the default traits for free functions of the Boolean Operations package, while segment traits can be still be manually selected.

## Summary of API changes

- `General_polygon_2` is given a new constructor taking a `Traits::Curve_2` object as parameter: this constructor initializes the polygon by first subdividing the input curve into subcurves using a `Traits::Make_x_monotone_2` functor. This makes the initialization of a polyline-based `General_polygon_2` more user-friendly.

- All free functions of the package Boolean Operation which take as argument `Polygon_2` or `Polygon_with_holes_2` (or `InputIterator` to both these types) are given a new template parameter `UsePolylineTraits`. This parameter allows to select if the internal `General_polygon_2` objects generated for the BO computation will be templated by `Arr_polyline_traits_2` (if `UsePolylineTraits=CGAL::Tag_true`, default) or by `Arr_segment_traits_2` (if `UsePolylineTraits=CGAL::Tag_false`).

- Free functions of the package Boolean Operation which take `General_polygon_2` objects as argument already give users the choice of the traits to be used, so their are left unchanged

(no change)

## CHANGES.md

TODO

## Submission

(Links not valid yet, waiting for documentation to be compiled.)

* [__User Manual__](#)
* [__Reference Manual__](#)

## Status

* Depends on (and thus includes) https://github.com/CGAL/cgal/pull/5091
* Developed and locally tested (GNU/Linux)
* [Small Feature](#)


sgiraudot:
build:doc_v1

Boolean_set_operations_2/include/CGAL/General_polygon_2.h
```diff
@@ -43,9 +45,26 @@ class General_polygon_2
...
   General_polygon_2(){}
 
+  General_polygon_2 (const Curve_2& curve)
+  {
+    typename Traits_2::Make_x_monotone_2 make_x_monotone_2
+      = Traits_2().make_x_monotone_2_object();
+    make_x_monotone_2
+      (curve,
+       boost::make_function_output_iterator
+       ([&](const Object& obj)
```
maxGimeno, line 55: Object is not defined here

efifogel:
Hi,

I believe that it should be implemented somehow differently.
We have the main entry functions. They reside in Boolean_set_operations_2.h.
In general a binary op. accepts 2 operand, each of type

   1. Polygon_2
   2. Polygon_with_holes_2
   3. General_polygon_2, and
   4. General_polygon_with_holes_2

In addition there are functions that accept a Traits object.
There are overloads that construct a Traits and pass it on if not supplied
by the caller.

In order to make the following discussion more simple, let's assume that we
have only Polygon2_ and General_polygon_2. (The distinction between with
and without holes is irrelevant here).

Observations
1. The introduction of the overloads that accept Polygon_2 are merely
optimizations, because a Polygon_2 is just a specific type of
General_polygon_2.
These functions can deal with the points that define the polygon instead of
the more general functions that must deal with the curves that comprise the
boundaries of the general polygon.

2. It turns out that applying boolean operations on polygons can be
expedited by converting the polygons to general polygons bounded by
polylines, referred to as Polyline-polygons, and then calling a function
that accepts such general polygons, and these functions already exist!

Objectives
First, a user that would like to apply an operation, say op(P, Q) on 2
polygons, say P and Q, can simply convert P and Q to Polyline-polygons and
call the appropriate functions.
Secondly, we would like to change the semantics of existing calls, so that
op(P, Q) will actually convert P and Q to Polyline-polygons, etc.

Conclusions
Adding a Boolean tag that dispatch the appropriate calls is correct,
because of observation (1) above, as the entire purpose was expediting, so
we can enhance the API and expedite further.
However, this is the only change that should be applied.
In other words, the default traits should not be affected, and there is no
reason to add Gps_polyline_traits in Gps_default_traits.h
Laso, I would like to preserve the ability to pass a traits class
regardless of whether the input polygons remain polygons or converted to
Polyline-polygons.

This is what we have today (before the change in this branch)

template <class Pgn1, class Pgn2, class Traits>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2, Traits& tr)  {
  General_polygon_set_2<Traits> gps(tr);
  gps.insert(pgn1);
  return (gps.op(pgn2));
}

template <class Pgn1, class Pgn2>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2) {
  typename Gps_default_traits<Pgn1>::Traits tr;
  return _op(pgn1, pgn2, tr);
}

We simply need to duplicate each function above:

template <class Pgn1, class Pgn2, class Traits>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2, Traits& tr, Tag_true =
Tag_true())  {
  typedef Arr_polyline_traits_2<Traits>         Polyline_traits;
  Polyline_traits polyline_tr(&tr);
  General_polygon_set_2<Traits> gps(tr);
  gps.insert(convert_polygon(pgn1_);
             return (gps.op(convert_polygon(pgn2)));
}

template <class Pgn1, class Pgn2, class Traits>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2, Traits& tr, Tag_false)
 {
  General_polygon_set_2<Traits> gps(tr);
  gps.insert(pgn1);
  return (gps.op(pgn2));
}

template <class Pgn1, class Pgn2>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2, Tag_true t =
Tag_true()) {
  typename Gps_default_traits<Pgn1>::Traits tr;
  return _op(pgn1, pgn2, tr);
}

template <class Pgn1, class Pgn2>
inline bool _op(const Pgn1& pgn1, const Pgn2& pgn2, Tag_false t) {
  typename Gps_default_traits<Pgn1>::Traits tr;
  return _op(pgn1, pgn2, tr, t);
}

Other Things
1. Instantiating Polyline<Kernel> with a Kernel provided as a template
parameter should be replaced with  Polygon_2<Kernel, Container> where
Container should be added as an additional template parameter
2. Arr_polyline_traits_2 is oblivious to the direction of the curves.
However, when dealing with polygons (and polygons with holes) we know the
direction, and we can take advantage of it.
For example, in convert_polygon() we should create X-monotone polylines to
start with.
3. The documentation is wrong or missing (even for the current setup)
because the conversion is automatically applied only to Polygon_2 and not
to General_polygon_2.
After the change above it should be revamped anyway.

Efi

                         _/

sgiraudot:
I agree with your proposed changes. I ignored the traits parameter in the first time because it was undocumented, but as you indeed explained, it is going to be, so the proposed changed looks good to me.

> 1. Instantiating Polyline<Kernel> with a Kernel provided as a template
parameter should be replaced with  Polygon_2<Kernel, Container> where
Container should be added as an additional template parameter

Just to be sure, you're talking about the Polygon_2 created by the conversions functions? In that case, I agree indeed.

> 2. Arr_polyline_traits_2 is oblivious to the direction of the curves.
However, when dealing with polygons (and polygons with holes) we know the
direction, and we can take advantage of it.
For example, in convert_polygon() we should create X-monotone polylines to
start with.

`convert_polygon()` does create X-monotone polylines: it calls the newly added `General_polygon_2` constructor that takes a `Curve` as input and calls `Traits::Make_x_monotone_2` on it to create the polygon.

> 3. The documentation is wrong or missing (even for the current setup)
because the conversion is automatically applied only to Polygon_2 and not
to General_polygon_2.
After the change above it should be revamped anyway.

I'm not sure what you mean, I documented the tag for overloads that take `Polygon_2`, not for those using `General_polygon_2`, which is correct as I did not modify the functions that take `General_polygon_2`. But anyway, as you said, this will be changed.

efifogel:
On Mon, 9 Nov 2020 at 14:10, Simon Giraudot <notifications@github.com>
wrote:

> I agree with your proposed changes. I ignored the traits parameter in the
> first time because it was undocumented, but as you indeed explained, it is
> going to be, so the proposed changed looks good to me.
>
>
>    1. Instantiating Polyline with a Kernel provided as a template
>    parameter should be replaced with Polygon_2<Kernel, Container> where
>    Container should be added as an additional template parameter
>
> Just to be sure, you're talking about the Polygon_2 created by the
> conversions functions? In that case, I agree indeed.
>
>
>    1. Arr_polyline_traits_2 is oblivious to the direction of the curves.
>    However, when dealing with polygons (and polygons with holes) we know
>    the
>    direction, and we can take advantage of it.
>    For example, in convert_polygon() we should create X-monotone
>    polylines to
>    start with.
>
> convert_polygon() does create X-monotone polylines: it calls the newly
> added General_polygon_2 constructor that takes a Curve as input and calls
> Traits::Make_x_monotone_2 on it to create the polygon.
>

You are right. For some reason I thought that we can do better than calling
Traits::Make_x_monotone_2, but now I cannot think why, and i was probably
wrong.

>
>    1. The documentation is wrong or missing (even for the current setup)
>    because the conversion is automatically applied only to Polygon_2 and
>    not
>    to General_polygon_2.
>    After the change above it should be revamped anyway.
>
> I'm not sure what you mean, I documented the tag for overloads that take
> Polygon_2, not for those using General_polygon_2, which is correct as I
> did not modify the functions that take General_polygon_2. But anyway, as
> you said, this will be changed.
>
Yep, you are right again.

maxGimeno:

sgiraudot:
@efifogel what do you think is the status of this feature? Have you done all the modifications you wanted to do? Would you pre-approve it?

Also, please let me know if you want me to do the rebase and fix the conflicts or if you prefer to do it yourself.

efifogel:
It's not done yet.
I've decided to change few things, and I'm in the middle...
I rather wait for (#5050) to be integrated with master; then I'll merge
this one with master; then I'll finish it.
                         _/

efifogel:
Getting some progress.
I'll soon finish it.
                         _/

Arrangement_on_surface_2/include/CGAL/Arr_polyline_traits_2.h
```diff
@@ -437,11 +437,17 @@ class Arr_polyline_traits_2 : public Arr_polycurve_traits_2<SegmentTraits_2> {
...
       return operator()(boost::make_transform_iterator
-                        (make_zip_iterator(boost::make_tuple(begin, begin + 1)),
+                        (make_zip_iterator(boost::make_tuple(begin,
+                                                             begin_plus_one)),
                          point_pair_to_segment),
                         boost::make_transform_iterator
-                        (make_zip_iterator(boost::make_tuple(end - 1, end)),
+                        (make_zip_iterator(boost::make_tuple(end_minus_one,
+                                                             end)),
                          point_pair_to_segment));
```
lrineau: Hi @efifogel,

Next time you need that, think about [`std::next`](https://en.cppreference.com/w/cpp/iterator/next) and [`std::prev`](https://en.cppreference.com/w/cpp/iterator/prev), that are really useful for that sort of thing.

efifogel:
You are right; much better. Thanks!
                         _/

sgiraudot:
Thanks a lot for the updates Efi! I added the points you mentioned by mail in the PR's description + links to the newly compiled doc.

efifogel:
One more thing.
I've realized that the new constructor of General_polygon_2 from
ArrTraits::Curve_2 is conceptually erroneous and I've removed it, so you
need to remove it also from the API changes.
A traits in our context may have state, so it's illegal to simply construct
a traits on the fly as in:
Traits().make_x_monotone_object()
If at all, it is the job of the GpsTraits to construct General_polygon_2
objects (and also General_polygon_with_holes_2 objects).
So, if we really want this, we can add an operator to
Gps_traits_2::Construct_general_polygon_2 that accepts a curve of type
ArrTraits::Curve_2
This is analogous to Construct_x_monotone_curve_2 and Construct_curve_2
functors in ArrTraits.
                         _/

sgiraudot:
@efifogel about your last message: do you still want to change something in this PR (or do you want me to do it) or do you consider it good to go?

efifogel:
I cannot seem to find a way to edit the page, so just remove the item:

   - Added missing constructors to General_polygon_set_2 that accepts a
   traits in addition to a polygon; see
   http://www.cs.tau.ac.il/~efif/doc_output/Boolean_set_operations_2/classCGAL_1_1General__polygon__set__2.html

and it will be ready to go.
I will add a new small feature that suggests adding constructors to the Gps
traits (instead of to the General_polygon_set_2 class itself), but this is
independent of this feature.
                         _/

sgiraudot:
Done. Thanks a lot!

maxGimeno:
I postpone to 5.4 as it has still not been pre-accepted.

sloriot:
@efifogel said in an email that he approves the changes.

efifogel:
Yes, I do approved, and as far as I'm concerned the code can be integrated
                         _/

afabri:
In fact I can produce the same error with VC 2019    calling    `./build/Debug/test_agg_op.exe ./data/agg_op/test02.txt`, so let me go into the debugger now.

Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Bso_internal_functions.h
```diff
@@ -8,192 +8,522 @@
...
+  General_polygon_set_2<Traits> gps(*begin, traits);
+  gps.join(std::next(begin), end, k);
+  return gps.polygons_with_holes(oi);
+}
+
+// Without traits
+template <typename InputIterator, typename OutputIterator>
+inline OutputIterator r_join(InputIterator begin, InputIterator end,
+                             OutputIterator oi, unsigned int k=5) {
+  typedef typename std::iterator_traits<InputIterator>::value_type Pgn;
```
afabri, line 294: Here it needs the same `if (begin == end) return oi;` as in line 284, as otherwise the `*begin` in line 299 fails.

Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Bso_internal_functions.h
```diff
@@ -8,192 +8,522 @@
...
+// Without traits
+template <typename InputIterator1, typename InputIterator2,
+          typename OutputIterator>
+inline OutputIterator r_join(InputIterator1 begin1, InputIterator1 end1,
+                             InputIterator2 begin2, InputIterator2 end2,
+                             OutputIterator oi, unsigned int k=5) {
+  typedef typename std::iterator_traits<InputIterator1>::value_type Pgn;
+  typename Gps_polyline_traits<Pgn>::Traits traits;
+  const typename Gps_polyline_traits<Pgn>::Polyline_traits& ptraits(traits);
+  return r_join(convert_polygon_iterator(begin1, ptraits),
```
afabri, line 325: Here as well we have to do something special in case of  `begin == end`, but I do not yet what.

Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Bso_internal_functions.h
janetournois: I am working on the patch

janetournois: It seems your commit and my patch are in conflict.
Let me know if you need a hand for msvc 2017

afabri: So maybe just review my fix.   It now passes locally for me.

afabri: I should have confirmed Sebastien that I take the task.

janetournois:
Is there an issue with MPFR in our CI?

lrineau:
> Is there an issue with MPFR in our CI?

That is strange. In the same Github Action run, on job find MPFR and not the other one. There [was an incident with Github Actions, yesterday](https://www.githubstatus.com/incidents/rgkr7yjqv6gj). I re-launched the failed tests, to see if the result might improve.

sloriot:
Successfully tested in CGAL-5.4-Ic-62.

sloriot:
However, I noticed something strange. It seems that the space needed to compile changed a lot as I cannot locally compile test is my 2GB of `/tmp`. Can someone else give it a try?

sloriot:
OK it seems that I already have the problem with master so it is not related to this branch.

lrineau:
TODO: missing additions to `CHANGES.md`.

Note that reviewers should review the `CHANGES.md` before the feature is accepted.

sloriot:
@efifogel could you please approve the modification of `CHANGES.md` (see the top of the PR to get the rendering of the change) so that we can merge this branch?

Installation/CHANGES.md
```diff
@@ -97,6 +97,18 @@ Release date: December 2021
 -   Added new shapes to the Region Growing algorithm on a point set: circles in 2D, spheres in 3D,
     and cylinders in 3D.
 
+### [2D Regularized Boolean Set-Operations](https://doc.cgal.org/5.4/Manual/packages.html#PkgBooleanSetOperations2)
+
+-   Added an extra parameter (`UsePolylines`) to all free functions (
+    [`complement()`](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.4-Ic-62/output2/Boolean_set_operations_2/group__boolean__complement.html),
```
lrineau, line 103:
```suggestion
    [`complement()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__complement.html),
```

Installation/CHANGES.md
```diff
@@ -97,6 +97,18 @@ Release date: December 2021
...
+    [`do_intersect()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__do__intersect.html),
+    [`intersection()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__intersection.html),
+    [`join()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__join.html),
+    [`difference()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__difference.html),
+    [`symmetric_difference()`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__symmetric__difference.html),
+    and [`oriented_side`](https://doc.cgal.org/5.4/Boolean_set_operations_2/group__boolean__oriented__side.html))
+    to control whether to use `Arr_polyline_traits_2` as default traits. It is the new default as it provides better performances in general.
```
lrineau, line 110: Is that a breaking change?

lrineau: ... Probably not, because it does not seem that this PR changed any example file.

sloriot: no, it the last parameter and it has a default.

PR #5098, opened 2020-10-20
Allow Epec in Refine_edges_with_clusters

mhsaar:

## Summary of Changes

Allow instantiation of Refine_edges_with_clusters with Epec
.
Use CGAL::to_double before calling sqrt, this should be a null op with Epic, while avoiding cascaded constructions for Epec. 

## Release Management

* Affected package(s):

### DIFF
--- a/Mesh_2/include/CGAL/Mesh_2/Refine_edges_with_clusters.h
+++ b/Mesh_2/include/CGAL/Mesh_2/Refine_edges_with_clusters.h
@@ -273,8 +273,10 @@ private:
         const Point m = midpoint(a, b);
 
         typename Geom_traits::Vector_2 v = vector(a,m);
-        v = scaled_vector(v,CGAL_NTS sqrt(c.minimum_squared_length /
-                                      squared_distance(a,b)));
+        v = scaled_vector(
+            v, CGAL_NTS sqrt(CGAL::to_double(c.minimum_squared_length /
+                                             squared_distance(a, b))));
+
 
         Point i = translate(a,v), i2(i);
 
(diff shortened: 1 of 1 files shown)


Mesh_2/include/CGAL/Mesh_2/Refine_edges_with_clusters.h
```diff
@@ -274,8 +274,8 @@ class Refine_edges_base_with_clusters :
 
         typename Geom_traits::Vector_2 v = vector(a,m);
         v = scaled_vector(
-            v, CGAL_NTS sqrt(CGAL::to_double(c.minimum_squared_length /
-                                             squared_distance(a, b))));
+            v, CGAL_NTS sqrt(CGAL_NTS to_double(c.minimum_squared_length /
```
afabri, line 277: I may be wrong but the `CGAL_NTS` has become useless, as it is an empty `#define`

mhsaar: That's what I thought too, but since there seemed to be an error in the checks this seemed the most likely cause.

mhsaar: I looks more consistent now too. :)

lrineau (approved):

sloriot:
@lrineau does it make sense to use `approximate_sqrt()`? I mean in case someone uses a number type like `CORE::Expr` that implements `sqrt` the `to_double` might break something or on the contrary will avoid unneeded extra precision?

lrineau:
> @lrineau would it make sense to use `approximate_sqrt()` instead? I mean in case someone uses a number type like `CORE::Expr` that implements `sqrt` the `to_double` might break something or on the contrary will avoid unneeded extra precision?

Yes, why not. I do not think it makes sense to use an iterative meshing algorithm with an exact number type, anyway.

mhsaar:
Approximate sqrt seems good too.

The position of the point doesn't need to be exact. The exact number type
is needed so the point remains on the 3D plane it is computed in.

maxGimeno:

PR #5100, opened 2020-10-21
LCC: Fix examples cmake

maxGimeno:
## Summary of Changes
Only link the drawing example with CGAL_Qt5.
## Release Management

##TODO

- [ ] create a new macro that calls `add_executable` and then target_link_library to link with CGAL::CGAL, with an option like COMPONENTS ... to add links with CGAL::CGAL_Core, CGAL::CGAL_Qt5, CGAL::Eigen_support, etc.
 - [ ] replace `create_single_source_cgal_program` calls by the new macro where it is necessary


gdamiand (commented):
It was already the case before (CGAL_Qt5 was linked only for the drawing example).

Moreover, do you think it is better to have two find_package(CGAL) ? (the first one without component and the second one with component qt5 ?)

maxGimeno:
The macro `create_single_source_cgal_program()` uses the COMPONENTS part of the `find_package()` call to decide what to link. If you call CGAL_Qt5 in the beginning of the file, every executable will be linked with CGAL_Qt5. That's what is currently done in 5.0 and probably every release branch. 
I think the best fix would be to entirely remove this macro from CGAL and perform each call explicitely.
I will add fixes in this branches, maybe it will come to that, but it is easier and faster to just move the find_package COMPONENT call.

gdamiand:
> The macro `create_single_source_cgal_program()` uses the COMPONENTS part of the `find_package()` call to decide what to link. If you call CGAL_Qt5 in the beginning of the file, every executable will be linked with CGAL_Qt5. That's what is currently done in 5.0 and probably every release branch.
> I think the best fix would be to entirely remove this macro from CGAL and perform each call explicitely.
> I will add fixes in this branches, maybe it will come to that, but it is easier and faster to just move the find_package COMPONENT call.

Thanks for the explanations. I agree we can probably remove the use of this macro.
If there is a plan to solve this later, not sure we need to do soomething now.

maxGimeno:
You should be satisfied with this new version, @gdamiand

gdamiand:
> You should be satisfied with this new version, @gdamiand

Thanks ! Much better ;)
Maybe it can be an option of find_package(CGAL) ?

I think I already saw something like find_package(CGAL OPTIONAL_COMPONENTS Qt5) ?
Then you can remove the lines list( REMOVE_ITEM CGAL_REQUESTED_COMPONENTS Qt5) from the different cmakelists.

What do you think ?

maxGimeno:
OPTIONAL_COMPONENTS only specifies that if the component is not found, it should not fail. It won't change anything on the matter of `create_single_source_cgal_program`.

gdamiand:
> OPTIONAL_COMPONENTS only specifies that if the component is not found, it should not fail. It won't change anything on the matter of `create_single_source_cgal_program`.

It could be... If a component is optional, we can decide to not linked it automatically by  `create_single_source_cgal_program`.

sloriot:
Since we have `CGAL_EIGEN3_SUPPORT`, can't we have the same thing for QT?

gdamiand:
I think this is not the same: eigen is an external library, contrary to CGAL_Qt5 (and CGAL_Core, CGAL_ImageIO).

The problem comes from the fact that the macro create_single_source_cgal_program() automatically link with all the components.

Maybe the simplest solution is to change this behavior: now this macro only links with CGAL, other components need to be linked by hand.

maxGimeno:
> I think this is not the same: eigen is an external library, contrary to CGAL_Qt5 (and CGAL_Core, CGAL_ImageIO).
> 
> The problem comes from the fact that the macro create_single_source_cgal_program() automatically link with all the components.
> 
> Maybe the simplest solution is to change this behavior: now this macro only links with CGAL, other components need to be linked by hand.

I agree with that. I will do a 3rd pass.

Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
```diff
@@ -69,11 +69,6 @@ function(create_single_source_cgal_program firstfile )
     add_to_cached_list( CGAL_EXECUTABLE_TARGETS ${exe_name} )
 
     target_link_libraries(${exe_name} PRIVATE CGAL::CGAL)
-    foreach(comp ${CGAL_REQUESTED_COMPONENTS})
-      if(TARGET CGAL::CGAL_${comp})
-        target_link_libraries(${exe_name} PRIVATE CGAL::CGAL_${comp})
-      endif()
-    endforeach()
```
lrineau, line 76: That patch will break everything that relies on `create_single_source_cgal_program` to link with `CGAL_Core`, or `CGAL_ImageIO`.

maxGimeno: then it will be pointed out in the CI and I will be able to add the missing calls to `target_link_libraries()`, if it is actually needed.

lrineau: Actually no: a .cpp may be surrounded by tests of `CGAL_USE_CORE`. We need another way to detect which .cpp use CGAL_Core, or other libraries.

lrineau:
@maxGimeno This PR should be tested. It seems useful.

lrineau:
> @maxGimeno This PR should be tested. It seems useful.

ping @maxGimeno What is the future of this PR?

maxGimeno:
For now I am focusing on CGAL_IO. It will be tested after

lrineau:
This PR is a bug fix and CGAL I/O is for CGAL 5.3. You should ensure bug fixes are tested as soon as possible.

lrineau:
I have reassigned this PR to 5.3-beta, and added "Under discussion", because of the issues I talked about in https://github.com/CGAL/cgal/pull/5100#pullrequestreview-513705759.

gdamiand:
> I have reassigned this PR to 5.3-beta, and added "Under discussion", because of the issues I talked about in [#5100 (review)](https://github.com/CGAL/cgal/pull/5100#pullrequestreview-513705759).

@lrineau I think the main question is about the current behavior of macro create_single_source_cgal_program. For now, program is linked with all cgal components found in the find_package().

1) this is a bug, then changing this will necessarily break all the CMakeLists that use find_package; we need to do something in these cmakelist (do the link by hand, add options to macro create_single_source_cgal_program to link with components, other possibility...)

2) this is a feature, we can close this issue.

lrineau:
> > I have reassigned this PR to 5.3-beta, and added "Under discussion", because of the issues I talked about in [#5100 (review)](https://github.com/CGAL/cgal/pull/5100#pullrequestreview-513705759).
> 
> @lrineau I think the main question is about the current behavior of macro create_single_source_cgal_program. For now, program is linked with all cgal components found in the find_package(). I see two possibilities:
> 
>     1. this is a bug, then changing this will necessarily break all the CMakeLists that use macro create_single_source_cgal_program; we need to do something in these cmakelist (do the link by hand, add options to macro create_single_source_cgal_program to link with components, other possibility...)

I think that is a bug. `create_single_source_cgal_program` could stay as it is for compatibility, and then create a CMake function with a better name (did you know `create_single_source_cgal_program` can compose with multiple source files?) and with a new semantic. Probably `cgal_add_executable`, that would just call `add_executable` and then `target_link_library` to link with `CGAL::CGAL`. Maybe we add options like `COMPONENTS ...` to add links with `CGAL::CGAL_Core`, `CGAL::CGAL_Qt5`, `CGAL::Eigen_support`, and so on.

gdamiand:
> I think that is a bug. `create_single_source_cgal_program` could stay as it is for compatibility, and we should create a new CMake function with a better name (did you know `create_single_source_cgal_program` can compose with _multiple_ source files?) and with a new semantic. Probably `cgal_add_executable`, that would just call `add_executable` and then `target_link_library` to link with `CGAL::CGAL`. Maybe we add options like `COMPONENTS ...` to add links with `CGAL::CGAL_Core`, `CGAL::CGAL_Qt5`, `CGAL::Eigen_support`, and so on.

Ok, thanks ! I like the idea and the new macro.

And in a next step we can replace progressively all our use of `create_single_source_cgal_program` by the new macro.

sloriot:
What would we loose if we don't introduce a new macro and use documented cmake functions. My point is that it would make it easier for users to use CGAL in their own project if only standard cmake was used in our example cmake scripts (I'm not talking about the third party target macros).

gdamiand:
> What would we loose if we don't introduce a new macro and use documented cmake functions. My point is that it would make it easier for users to use CGAL in their own project if only standard cmake was used in our example cmake scripts (I'm not talking about the third party target macros).

Indeed, without the macro we need 2 lines (add_executable and target_link_library) of cmake instead of 1 (create_single_source_cgal_program).

afabri:
> 
> 
> What would we loose if we don't introduce a new macro and use documented cmake functions. My point is that it would make it easier for users to use CGAL in their own project if only standard cmake was used in our example cmake scripts (I'm not talking about the third party target macros).

I fully agree.

lrineau:
> What would we loose if we don't introduce a new macro and use documented cmake functions. My point is that it would make it easier for users to use CGAL in their own project if only standard cmake was used in our example cmake scripts (I'm not talking about the third party target macros).

_Side note on the vocabulary: we may want to create a CGAL-specific CMake "function", or use documented CMake "commands"._

I also agree, but not fully.

We a CGAL-specific CMake function, we would have:
```cmake
cgal_add_executable(mesh mesh.cpp)
```
using standard commands:
```cmake
add_executable(mesh mesh.cpp)
target_link_libraries(mesh PRIVATE CGAL::CGAL)
```

That is not a big deal, and I agree that would be profitable for the readability of our CMake scripts.

BUT... then what about our support for CTest?
1. Either we add a third line, for each target: `cgal_add_test(mesh)`.
2. Or we use nasty tricks, with introspection, to automatically detect the targets of the current directory, and call the necessary functions transparently.

gdamiand:
Indeed @lrineau , I forgot tests...

Thus we have 3 lines instead of 1, and we still need a macro ! (I don't like the idea of nasty tricks with introspection which is often complicated and can imply bugs).

Thus I vote for the new macro cgal_add_executable.

lrineau:
@sloriot @maxGimeno Please have a look at this PR, that is in the "Trash/Attic" milestone. It was closed by a technical error. You may want to close it anyway, but after a review.

sloriot:
We should rename the PR as the scope is broader. What was the initial issue? Examples being linked with not needed CGAL libraries?

maxGimeno:
This PR doesn't really make sense anymore. Initially, I realized that our cmake macro for target creation links vith pretty much everything, because a non graphical example was having problems linking with CGAL_Qt5 or something like that (LCC examples, from the title of the PR). But wafter I cleaned it up, we realized that the whole mechanism is wanted this way because it is convenient, and if we want to clean it we'll have to rewrite 90% of the tests and examples of CGAL, so we postponed. And now the whole work I did in this PR is to be entirely remade because every CMakeLists in CGAL has been modified by #5063, and 100% of the files that were edited by this PR are now conflicting. So I vote we sinply close this, and if the problem I initially raised is still considered an issue, let's open an issue instead.

PR #5102, opened 2020-10-21
Remove documentation warning  from Surface_mesh_approximation

albert-github:
When running the documentation generation we get the warning like:
```
warning: end of comment block while expecting command </tt>
```

This is due to the `.` (dot) inside the back-tick span (it is seen as the end of the brief description).
Also the `@a` has no meaning in this span.


albert-github:
Closed in favor of 5120.

PR #5106, opened 2020-10-22
CI : Message for the doc

maxGimeno:
Specify that the doc takes some time before being available.


lrineau:
Please do not use labels named "rm only:" at all. If you want me or Sébastien to act on a PR quickly, just `@mention` us.

PR #4978, opened 2020-09-04
Polyhedron demo - remesh a non-triangle mesh

janetournois:
## Summary of Changes

When a user wants to remesh a non-triangle mesh, the remeshing menu is available, but remeshing freezes the demo.
With this PR, a pop-up asks whether the demo should first triangulate faces, or cancel remeshing.

Actually, it is not needed that the input mesh is fully triangulated. Only the faces to be remeshed should be triangulated.

TODO :
- [x] add something similar in `isotropic_remeshing_of_several_polyhedra()`
- [x] add in `triangulate_facets_plugin` the possibility to triangulate a selection of facets (given by a `scene_selection_item`)

## Release Management

* **Depends on: #3268** 
* Affected package(s): Polyhedron demo, PMP
* License and copyright ownership: unchanged

PR #5108, opened 2020-10-23
Demo isotropic_remeshing_plugin - keep selection valid

janetournois:
## Summary of Changes

In the demo, the `isotropic_remeshing_plugin` was invalidating the set of selected edges (resp. vertices, facets) given as input via the selection plugin, because of the modifications made to the mesh associated with the call to `surface_mesh->collect_garbage()`.

This PR uses the visitor in `surface_mesh->collect_garbage(visitor)` to keep selection valid after remeshing and garbage collection.

This PR follows PR #4978 in the case we have a selection_item with selected edges

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Isotropic_remeshing_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Isotropic_remeshing_plugin.cpp
@@ -471,16 +471,11 @@ public Q_SLOTS:
         {
           selection_item->polyhedron_item()->setItemIsMulticolor(false);
         }
-
-
-        selection_item->polyhedron_item()->polyhedron()->collect_garbage(visitor);
-
-        selection_item->reset_numbers();
-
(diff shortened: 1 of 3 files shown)


janetournois:
There is still an issue though : `selection_item->polyhedron_item()->invalidateOpenGLBuffers();` empties the selection, though it is still valid.

maxGimeno:
The problem is that modifications to the poly_item made by other operations will still mess up the selection. 
But I think you can trick it by updating num_faces/edge/etc. in the Scene_polyhedron_selection_item's `d`.
You will need to add a function tu update them, and call it before calling invalidateOpenGLBuffers(). This way, the function will consider that no significant change has been performed, and won't clear the selection.

janetournois:
Selection is now valid after remeshing.

Note that using the visitor in `collect_garbage()` as done in this PR could probably be generalized in the demo to avoid clearing the selection when it's possible.

maxGimeno:

maxGimeno:

PR #5078, opened 2020-10-15
Use ctest in travis

maxGimeno:
## Summary of Changes

Use ctest in travis to avoid skipping tests when the architecture of the package differs from the usual (i.e BGL examples).

PR #5114, opened 2020-10-26
Fix CTest issues with CGAL_Qt5

lrineau:
## Summary of Changes

Fix CTest issues with CGAL_Qt5

Add a target `compilation_of__CGAL_Qt5_moc_and_resources`, and a corresponding CTest test, so that `CGAL_Qt5_moc_and_resources` is always compiled first, when needed. That will fix CTest race-conditions, where `CGAL_Qt5_moc_and_resources` was being compiled by several CTest test at the same time.

## Release Management

This PR is expected to fix the Travis issues that are currently in all branches, since we merged #5078.

### DIFF
--- a/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
+++ b/Installation/cmake/modules/CGAL_CreateSingleSourceCGALProgram.cmake
@@ -57,15 +57,6 @@ function(create_single_source_cgal_program firstfile )
       set(NO_TESTING TRUE)
     endif()
 
-    if(POLICY CMP0064)
-      # CMake 3.4 or later
-      if(NOT NO_TESTING)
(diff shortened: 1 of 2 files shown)


lrineau:
@maxGimeno Can you please review my patch (and test it on your machine as well)?

lrineau:
For a later work and PR (and only for master), I think we can fix the issues of the polyhedron demo the same way:
  - for all targets of the polyhedron demo, create a test named `compilation_of__${target}`,
  - then use dependencies between targets to infer dependencies between such test ``compilation_of__${target}`.

lrineau:
This PR is expected to fix the Travis issues that are currently in all branches, since we merged #5078.

maxGimeno (approved):

PR #5124, opened 2020-10-27
Add experimental non-documented function to simplify a 3D polyline

sloriot:
@lrineau if I put two lines in the `.cmd` does it run twice the exe?


lrineau:
> @lrineau if I put two lines in the `.cmd` does it run twice the exe?

That is the intended behavior. I hope it still works that way.

maxGimeno:
```
The following files have trailing whitespaces:
Polygon_mesh_processing/test/Polygon_mesh_processing/test_simplify_polylines_pmp.cpp
```

sloriot:
> > @lrineau if I put two lines in the `.cmd` does it run twice the exe?
> 
> That is the intended behavior. I hope it still works that way.

it does not work with ctest, let me check with `cgal_test_with_cmake`

sloriot:
Does not work neither, but I know it has always been like that, AFAIR

sloriot:
@maxGimeno you might want to create a plugin for the demo using the function.

Polyhedron/demo/Polyhedron/Plugins/IO/Polylines_io_plugin.cpp
```diff
@@ -253,6 +268,29 @@ void Polyhedron_demo_polylines_io_plugin::split()
   }
 }
 
+void Polyhedron_demo_polylines_io_plugin::simplify()
+{
+  Scene_polylines_item* item = qobject_cast<Scene_polylines_item*>(scene->item(scene->mainSelectionIndex()));
+  bool ok;
+  double err = QInputDialog::getDouble(mw, "Brechet Distance", "Enter the approximation error:", pow(0.01*item->diagonalBbox(),2),0,999,8,&ok);
```
sloriot, line 275: Who is Brechet? Squared Distance or do the square internally?

maxGimeno:

PR #5130, opened 2020-10-28
[Principal Component Analysis] Kernel compatibility, diagonalization precision and overload efficiency

sgiraudot:
## Summary of Changes

PR following the discussions with @palliez and @cportaneri.

It improves the _Principal Component Analysis_ package in the following ways:

- fix the computation of the centroid for sets of `Iso_cuboid_3` that were completely wrong, and also fixed the computation of PCA of `Iso_cuboid_3` in volumic variant which was wrong too. The PCA of `Iso_cuboid_3` in surface variant is broken too, but I couldn't fix it yet. I'm still working on it.

- for some reason, the matrices in `Eigen_diagonalize_traits` were filled using `static_cast<float>` operator on each value, which does not seem justified. Removing this cast drastically reduces errors between an estimated line (or plane) and points aligned on a line (or plane), from `1e-8` to `1e-15` in my experiments. This fix may also benefit to other packages which use `Eigen_diagonalize_traits` (point set processing for example)

- I also changed the implementation of `Eigen_diagonalize_traits` so that, internally, it does not use directly `FT` but a restriction of `FT` to `float` (is `CGAL::Simple_cartesian<float>` is used) or `double` (everywhere else), using `CGAL::to_double()` to convert `FT` to this restricted `FT`: this way, kernels such as `Epeck` can be used directly with the diagonalization traits

- I then made the few adaptations required in PCA to make it compatible with `Epeck` too (mainly replacing `std::sqrt` by `CGAL::approximate_sqrt`). I added a test to check compilation and execution of all variants with different kernels

- I added some assertions to check that the measures are not null (computing PCA on null measures may result in NaN, so this would help user detect where something went wrong)

- Finally, when computing the PCA on the subdimensions of an object (for example, on the edges of a triangle), a `std::list` of such subdimensional object was always created to call another variant on it: I replaced that with an adapted iterator that generates the subdimensional objects on the fly, which avoids the allocations of list objects and should improve timings

## Release Management

* Affected package(s): PCA, Solver Interface and packages relying on it (PSP, Classification, ...)

### DIFF
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -175,9 +175,9 @@ assemble_covariance_matrix_3(InputIterator first,
     transformation = volume * transformation * moment * transformation.transpose();
 
     // Translate the 2nd order moment to the minimum corner (x0,y0,z0) of the cuboid.
-    FT xav0 = (x1 - x0) / (4.0);
-    FT yav0 = (y1 - y0) / (4.0);
-    FT zav0 = (z1 - z0) / (4.0);
+    FT xav0 = (x1 - x0) / (2.0);
+    FT yav0 = (y1 - y0) / (2.0);
+    FT zav0 = (z1 - z0) / (2.0);
 
     // and add to covariance matrix
     covariance[0] += transformation(0,0) + volume * (2*x0*xav0 + x0*x0);
--- a/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
@@ -56,9 +56,9 @@ assemble_covariance_matrix_3(InputIterator first,
 
   // assemble 2nd order moment about the origin.
   Matrix moment;
-  moment << 1.0/12.0, 1.0/24.0, 1.0/24.0,
-            1.0/24.0, 1.0/12.0, 1.0/24.0,
-            1.0/24.0, 1.0/24.0, 1.0/12.0;
+  moment << FT(1.0/12.0), FT(1.0/24.0), FT(1.0/24.0),
+            FT(1.0/24.0), FT(1.0/12.0), FT(1.0/24.0),
+            FT(1.0/24.0), FT(1.0/24.0), FT(1.0/12.0);
 
   for(InputIterator it = first;
       it != beyond;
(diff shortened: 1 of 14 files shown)


Principal_component_analysis/include/CGAL/PCA_util_Eigen.h
```diff
@@ -215,6 +216,32 @@ assemble_covariance_matrix_3(InputIterator first,
...
+      return Triangle (c[indices[idx][0]], c[indices[idx][1]], c[indices[idx][2]]);
+    };
+
+  assemble_covariance_matrix_3
+    (make_subiterator<Triangle, 12> (first, converter),
+     make_subiterator<Triangle, 12> (beyond),
+     covariance, c, K(), (Triangle*)nullptr, CGAL::Dimension_tag<2>(),
+     Eigen_diagonalize_traits<FT, 3>());
+
+#else
```
lrineau, line 239: I think you should remove the old code, instead pf having `#if 1`. But you know that. Probably you have a reason.

sgiraudot: The reason is written as comment just after the `#else`: this is a workaround, not a fix. The old code is broken, but I want to keep it in case we want to come back to it later (it may be just an error in a formula, although I could not find one).

maxGimeno:
There seems to be some errors in this PR : 
```
In file included from /home/travis/build/CGAL/cgal/Solver_interface/examples/Solver_interface/diagonalize_matrix.cpp:3:
/home/travis/build/CGAL/cgal/Solver_interface/include/CGAL/Eigen_diagonalize_traits.h:78:18: error: cannot convert 'typename Real_embeddable_traits<double>::To_double::result_type' (aka 'CGAL::Null_tag') to 'CGAL::Eigen_diagonalize_traits<double, 3>::RFT' (aka 'double') without a conversion operator
        m(i,j) = static_cast<RFT>(CGAL::to_double(cov[(dim * i) + j - ((i * (i+1)) / 2)]));
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

sgiraudot:
Thanks, I'll look into it.

sgiraudot:
It should be fixed.

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-94/Principal_component_analysis/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

maxGimeno:
I'm suspicious about that : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-95/Principal_component_analysis/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz
May it come from this PR ?

sgiraudot:
It should be fixed. I hope I did not forget any…

maxGimeno:
warning: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-103/Principal_component_analysis/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

maxGimeno:

PR #5132, opened 2020-10-28
CI doc

maxGimeno:
## Summary of Changes
only clones the repo for the github pages after building the docs to reduce the interval in which there might be a conflict during simultaneous runs.

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -61,7 +61,6 @@ jobs:
         if: steps.get_round.outputs.result != 'stop'
         run: |
           set -ex
-          git clone https://CGAL:${{ secrets.PUSH_TO_CGAL_GITHUB_IO_TOKEN }}@github.com/CGAL/cgal.github.io.git --depth=5
           mkdir -p build_doc && cd build_doc && cmake ../Documentation/doc
 
       - name: Build and Upload Doc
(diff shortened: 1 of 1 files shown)


.github/workflows/build_doc.yml
```diff
@@ -76,15 +75,17 @@ jobs:
             LIST_OF_PKGS=$(git diff --name-only HEAD^1 HEAD |cut -s -d/ -f1 |sort -u | xargs -I {} ls -d {}/package_info 2>/dev/null  |cut -d/ -f1 |egrep -v Installation||true)
             if [ "$LIST_OF_PKGS" = "" ]; then
               exit 1
-            fi              
+            fi
             mkdir -p cgal.github.io/${PR_NUMBER}/$ROUND
             cd build_doc && make -j2 doc && make -j2 doc_with_postprocessing
+            cd ..
+            git clone https://CGAL:${{ secrets.PUSH_TO_CGAL_GITHUB_IO_TOKEN }}@github.com/CGAL/cgal.github.io.git
```
lrineau, line 82: Previously that clone was done using `--depth=5` but that option disappeared.

maxGimeno: yes, because now I keep the history to one single commit, so the option became useless

PR #5135, opened 2020-10-30
Ci test demo 

maxGimeno:
## Summary of Changes
Add a github worflow in 4 jobs to test the polyhedron demo compilation.

### DIFF
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -4,12 +4,9 @@ sudo apt-get update
 sudo apt-get install -y  libmpfr-dev \
  libeigen3-dev qtbase5-dev libqt5sql5-sqlite libqt5opengl5-dev qtscript5-dev \
  libqt5svg5-dev qttools5-dev qttools5-dev-tools libboost1.72-dev
-git clone --branch v3.18.4 --depth 1 git://github.com/Kitware/CMake.git
-cd ./CMake
-mkdir build
-cd build
-cmake ..
-make -j2
-sudo make install
-cd ../..
-rm -rf ./CMake
+#update cmake to 3.18.4
+sudo apt purge --auto-remove cmake
+cd /tmp
+wget https://cmake.org/files/v3.18/cmake-3.18.4-Linux-x86_64.sh
+sudo sh cmake-3.18.4-Linux-x86_64.sh --skip-license
+rm cmake-3.18.4-Linux-x86_64.sh
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -8,5 +8,5 @@ sudo apt-get install -y  libmpfr-dev \
 sudo apt purge --auto-remove cmake
 cd /tmp
 wget https://cmake.org/files/v3.18/cmake-3.18.4-Linux-x86_64.sh
-sudo sh cmake-3.18.4-Linux-x86_64.sh --skip-license
+sudo sh cmake-3.18.4-Linux-x86_64.sh --skip-license --prefix=/usr/local
 rm cmake-3.18.4-Linux-x86_64.sh
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+sudo add-apt-repository ppa:mikhailnov/pulseeffects -y
+sudo apt-get update
+sudo apt-get install -y  libmpfr-dev \
+ libeigen3-dev qtbase5-dev libqt5sql5-sqlite libqt5opengl5-dev qtscript5-dev \
+ libqt5svg5-dev qttools5-dev qttools5-dev-tools libboost1.72-dev
+git clone --branch v3.18.4 --depth 1 git://github.com/Kitware/CMake.git
+cd ./CMake
+mkdir build
+cd build
+cmake ..
+make -j2
+sudo make install
+cd ../..
+rm -rf ./CMake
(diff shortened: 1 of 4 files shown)


maxGimeno:
In the branch that uses PCH, there is :
```
if(${CMAKE_VERSION} VERSION_LESS "3.18.0")
  set(USE_PCH FALSE)
else()
```
So I took 3.18

.github/workflows/demo.yml
```diff
@@ -0,0 +1,77 @@
...
+        set -ex
+        cd Polyhedron/demo
+        LIST_OF_PLUGINS=$(for f in $(find ./Polyhedron/Plugins -iname "CMakeLists.txt" |egrep -v Three_examples); do egrep "polyhedron_demo_plugin" $f |cut -d\( -f2 | cut -d" " -f1|sort -u|egrep -v "partition_plugin" |egrep -v "basic_item_plugin"|egrep -v ||true; done)
+        PLUGINS_ARRAY=(${LIST_OF_PLUGINS});
+        NB_OF_PLUGINS=${#PLUGINS_ARRAY[@]}
+        DEL=$(($NB_OF_PLUGINS / 4))
+        mkdir build
+        cd build
+        /usr/local/bin/cmake -DCGAL_DIR=${{ github.workspace }} ../Polyhedron
+        make -j2 ${PLUGINS_ARRAY[@]:0:$DEL}
```
lrineau, line 23: That part should be in a script too.

.github/install.sh
```diff
@@ -0,0 +1,15 @@
...
+sudo apt-get install -y  libmpfr-dev \
+ libeigen3-dev qtbase5-dev libqt5sql5-sqlite libqt5opengl5-dev qtscript5-dev \
+ libqt5svg5-dev qttools5-dev qttools5-dev-tools libboost1.72-dev
+git clone --branch v3.18.4 --depth 1 git://github.com/Kitware/CMake.git
+cd ./CMake
+mkdir build
+cd build
+cmake ..
+make -j2
+sudo make install
```
lrineau, line 13: I think you should use the precompiled binaries from cmake.org. See https://askubuntu.com/a/865294, section "B. Using binary files" of the answer.

.github/test.sh
```diff
@@ -0,0 +1,14 @@
...
+LIST_OF_PLUGINS=$(for f in $(find ./Polyhedron/Plugins -iname "CMakeLists.txt" |egrep -v Three_examples); do\
+  egrep "polyhedron_demo_plugin" $f |cut -d\( -f2 | cut -d" " -f1|sort -u|egrep -v "partition_plugin"\
+ |egrep -v "basic_item_plugin"|egrep -v "vtk_plugin" |egrep -v "register_point_sets_plugin"| egrep -v "classification_plugin" | egrep -v "io_image_plugin" ||true; done)
+PLUGINS_ARRAY=(${LIST_OF_PLUGINS});
+NB_OF_PLUGINS=${#PLUGINS_ARRAY[@]}
+DEL=$(($NB_OF_PLUGINS / 4))
+mkdir build
+cd build
+/usr/local/bin/cmake -DCGAL_DIR=$2 ../Polyhedron
+make -j2 ${PLUGINS_ARRAY[@]:$(($FACOTR * $DEL)):$((($FACTOR + 1) * $DEL))}
```
lrineau, line 14: Nice trick.

But there seems to be a typo:
```suggestion
make -j2 ${PLUGINS_ARRAY[@]:$(($FACTOR * $DEL)):$((($FACTOR + 1) * $DEL))}
```

.github/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+FACTOR=$1
+set -ex
+cd Polyhedron/demo
+LIST_OF_PLUGINS=$(for f in $(find ./Polyhedron/Plugins -iname "CMakeLists.txt" |egrep -v Three_examples); do\
+  egrep "polyhedron_demo_plugin" $f |cut -d\( -f2 | cut -d" " -f1|sort -u|egrep -v "partition_plugin"\
+ |egrep -v "basic_item_plugin"|egrep -v "vtk_plugin" |egrep -v "register_point_sets_plugin"| egrep -v "classification_plugin" | egrep -v "io_image_plugin" ||true; done)
```
lrineau, line 7: Why do you remove those plugins:
  - `partition_plugin`,
  - `basic_item_plugin`,
  - `vtk_plugin`,
  - `register_point_sets_plugin`,
  - `classification_plugin`, and
  - `io_image_plugin`?

maxGimeno: Some of them need specific dependencies that I don't want to install on a CI, others come from the Three examples that I don't really care about and need a specific CMake option that I didn't add, so the targets do not exist and the script crashes trying to build them.

lrineau: OK, I understand. That is to remove plugins that were not configured.

For that I would use the output of `make help`:
```shell
LIST_OF_PLUGINS=$(cmake --build . -t help | egrep 'plugin$' |& cut -d\  -f2)
```

That way you get only the targets known by the build system.

maxGimeno: I didn't know it was possible to do that. Thanks for the tip

maxGimeno:
All green

lrineau:
I have clicked the "Merge" button with the option "Wait for successful checks": so it will be merged as soon as the CI is green.

Issue #4049, opened 2019-07-02
Use Abseil hash maps instead of std unordered containers for a hugde speedup


lrineau:
## Issue Details

Inspired by [one of the cppcon talks][*] I listen to during my free time, I have quickly tried the hash maps from [Abseil] instead of `std::unordered_map`.

[*]: https://youtu.be/fHNmRkzxHWs
[Abseil]: https://abseil.io/

My try was a modification of `copy_face_graph`:
```patch
diff --git a/BGL/include/CGAL/boost/graph/copy_face_graph.h b/BGL/include/CGAL/boost/graph/copy_face_graph.h
index 9a2cb628825..fd0b602f284 100644
--- a/BGL/include/CGAL/boost/graph/copy_face_graph.h
+++ b/BGL/include/CGAL/boost/graph/copy_face_graph.h
@@ -35,6 +35,7 @@
 #include <boost/unordered_map.hpp>
 #include <boost/utility/enable_if.hpp>
 #include <boost/function_output_iterator.hpp>
+#include <absl/container/flat_hash_map.h>
 
 namespace CGAL {
 
@@ -189,7 +190,7 @@ void copy_face_graph(const SourceMesh& sm, TargetMesh& tm,
   typedef typename boost::graph_traits<SourceMesh>::halfedge_descriptor sm_halfedge_descriptor;
   typedef typename boost::graph_traits<TargetMesh>::halfedge_descriptor tm_halfedge_descriptor;
 
-  boost::unordered_map<sm_halfedge_descriptor,
+  absl::flat_hash_map<sm_halfedge_descriptor,
                        tm_halfedge_descriptor> hash_map(num_halfedges(sm));
   copy_face_graph_impl(sm, tm,
                        boost::make_assoc_property_map(hash_map),
```

The results are stuning: now 3.2 seconds instead of 7.1 seconds

## Source Code

The .cpp code: https://gist.github.com/2e9ea602a7b1e46cd34773611acac3c6

Here is the local patch to `CMakeLists.txt` (a non-clean hack):
```patch
diff --git a/BGL/examples/BGL_polyhedron_3/CMakeLists.txt b/BGL/examples/BGL_polyhedron_3/CMakeLists.txt
index ac3bd5256f2..a821195ff9c 100644
--- a/BGL/examples/BGL_polyhedron_3/CMakeLists.txt
+++ b/BGL/examples/BGL_polyhedron_3/CMakeLists.txt
[...]

PR #5144, opened 2020-11-04
Proposed fix for unnecessary allocations in ID_support_handler. Issue #5143

### LINKED ISSUE #5143: Unnecessary allocations in ID_support_handler
## Issue Details

Recently I have been investigating the performance of binary operations on Nef Polyhedra. The performance analysis I made revealed that ID_support_handler is making lots of allocations, due to a default bucket size of 512 here
https://github.com/CGAL/cgal/blob/16fc8d1fe27ff2ea79340d5c7a892083acbb24fe/Hash_map/include/CGAL/Tools/chained_map.h#L276-L277

While this isn't unreasonable for a hash map, (anything smaller and there isn't much advantage to using hashes) in ID_support_handler this is used to make a hash map, of hash maps:

https://github.com/CGAL/cgal/blob/16fc8d1fe27ff2ea79340d5c7a892083acbb24fe/Nef_3/include/CGAL/Nef_3/ID_support_handler.h#L41-L42

Consequently, there are 262144 buckets allocated. Since this is used to store Halffacet_const_handle f1, and Halffacet_const_handle f2, i.e a pair of half facets, these are usually pairwise, so each of the sub hash tables seems to always contain only 1 entry, (wasting the other 512 buckets!)

## Proposed fix

I have a proposed fix, details of which I will publish below.

## Source Code

I have created the following benchmark source code:

https://gist.github.com/GilesBathgate/78f9df7b149ccfd021d63a1d84842c24

Which has the output as follows:

```
Total allocations 772
Total bytes 14333952
And now lets populate...
table_size: 512
number of entries: 168
fraction of entries in first position: 0.184524
fraction of empty lists: 0.0605469
table_size: 512
number of entries: 1
fraction of entries in first position: 1
fraction of empty lists: 0.00195312
table_size: 512
number of entries: 1
fraction of entries in first position: 1
fraction of empty lists: 0.00195312
... repeated 512 times ...
```

## Environment

Linux
* Compiler:
clang
* Release or debug mode:
Release
None
* CGAL version:
5.0.2
* Boost version:
1.71.0
None

GilesBathgate:
## Summary of Changes

Replace hash map of hash map with hash map of pairs.

## Release Management

* Affected package(s): Boolean Operations on Nef Polyhedra
* Issue(s) solved (if any): fix #5143
* License and copyright ownership: Returned to original CGAL authors


afabri:
Wouldn't it make sense to use [`boost::hash_combine()`](https://www.boost.org/doc/libs/1_74_0/doc/html/hash/combine.html)    instead of your [hand crafted code,](https://github.com/CGAL/cgal/pull/5144/files#diff-72f4c27e9b8013ec61e1d982fc52c76c2630b9efd1364c22e47ac6041d368285R49),  or are the magic constants from something like `boost::hash_combine()` ?

GilesBathgate:
@afabri I'll give `boost:hash_combine()` a look. The magic numbers approach came from here: https://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-overriding-gethashcode

GilesBathgate:
@afabri I modified the program in #5143 to do 1,000,000 iterations, and then compiled using the "hash giles" above, and the `boost:hash_combine()` as you suggested. I thought this would be a quick and simple way to compare performance as well as hash distribution:

```
time ./hash-giles 
Total allocations 1
Total bytes 18432
And now lets populate...
table_size: 512
number of entries: 168
fraction of entries in first position: 0.190476
fraction of empty lists: 0.0625

real	0m37.026s
user	0m37.026s
sys	0m0.000s
```

```
time ./hash-boost 
Total allocations 1
Total bytes 18432
And now lets populate...
table_size: 512
number of entries: 168
fraction of entries in first position: 0.845238
fraction of empty lists: 0.277344

real	0m37.367s
user	0m37.362s
sys	0m0.004s
```

The performance is about the same, but the has distribution (i.e the fraction of entries in the first position, and the fraction of empty lists) seems worse with `boost::hash_combine` of course this could be completely coincidental since the sample data is just random memory addresses. 

There might be other good reasons not to have magic numbers in the code base and use a library function instead. Both options seem to work well enough though. The code could do with a comment saying that the magic numbers are just randomly selected prime numbers that seem to work well.

GilesBathgate:
I tried the test using a simple xor hash and got similar numbers:

`return Handle_hash_function()(p.f1) ^ Handle_hash_function()(p.f2);`

```
time ./hash-xor 
Total allocations 1
Total bytes 18432
And now lets populate...
table_size: 512
number of entries: 166
fraction of entries in first position: 0.186747
fraction of empty lists: 0.0605469

real	0m37.186s
user	0m37.185s
sys	0m0.000s
```

xor is usually the worst option for hashing, but perhaps this is because the handle addresses started from are already quite random. Either way, I think perhaps hashing is more of an art than science to me :joy:

GilesBathgate:
The following paragraph appeals to me:

> ### Testing hash functions
>The most important test for a hash function is distribution on a sample of the expected input. No hash function is best for all possible inputs, which is why I introduced a large number of algorithms in this tutorial. Sometimes one of the good hash functions will work better than the others, and that one should be used. Other times they will all be fairly equal and you can choose at random. The key here is that a hash function should never be used blindly without testing it, no matter how good the author says it is. Often, hash tests are designed around a hash function, and that introduces a bias in favor of that function.

GilesBathgate:
@afabri I've realised that any hashing algo including `boost:hash_combine` doesn't provide collision free hashes needed for `Unique_hash_map`, this could be improved if we adopt the changes to use `robin_hood::unordered_map` suggested in #4049 which has comparable performance but allows duplicate keys.

afabri:
@maxGimeno In the testsuite  this PR has a conflict with the change of `set_index()`  in the PR #5498.  They must not put at the same time in integration.

GilesBathgate:
@afabri @maxGimeno This branch has both changes if needed for integration testing:

https://github.com/GilesBathgate/cgal/tree/Nef3-performance-id-support-handler

maxGimeno:
@GilesBathgate and the CI shows the same error as in the testsuite : https://github.com/GilesBathgate/cgal/runs/2084693398#step:4:345

GilesBathgate:
@maxGimeno Fixed _(**note to self**: more haste less speed :joy:)_

afabri (approved):

Nef_3/include/CGAL/Nef_3/ID_support_handler.h
```diff
@@ -18,7 +18,8 @@
 
 #include <CGAL/Nef_S2/ID_support_handler.h>
 #include <CGAL/Nef_3/SNC_indexed_items.h>
-#include <CGAL/Unique_hash_map.h>
+#include <boost/container_hash/hash.hpp>
```
afabri, line 21: I think this should be   `boost/functional/hash.hpp`  instead.   Note that we have a [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-95/Nef_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2015-Debug-64bits.gz)that fails.

afabri:
Replacing the include fixes the testsuite

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-98.shtml

lrineau:
Thanks again for your contribution, Giles. It was merged yesterday in `master`, in line for the 5.3-beta1 release (in about a month).

PR #3850, opened 2019-04-11
Triangulation_23: Enhance IO

maxGimeno:
## Summary of Changes

Add the possibility to use functions of the cell/face/vertex base, called read/write_data(), and that have access to the maps/vectors internal to the triangulation. This is used in the IO functions.

## Release Management

* Affected package(s):Triangulation_2, Triangulation_3
* Issue(s) solved (if any): fix #861, fix #2359
* Feature/Small Feature (if any):
* Link to compiled documentation (obligatory for small feature) 
* License and copyright ownership:

PR #5145, opened 2020-11-05
Triangulation_23: Add functions for extra data in IO

### LINKED ISSUE #861: I/O of Triangulation_2 or Triangulation_3, when the vertex or cell types have handles
(Cc @MoniqueTeillaud)

See the discussion in CGAL-develop:
* March 2015: https://sympa.inria.fr/sympa/arc/cgal-develop/2015-03/msg00189.html
* Followup in April 2015: https://sympa.inria.fr/sympa/arc/cgal-develop/2015-04/msg00011.html

Let's quote my mail:
> Let's take Triangulation_3 as an example. The operator<< of Triangulation_3 
> uses two maps:
>
>     // the "map" from indices to Vertex_handle
>     std::vector<Vertex_handle> TV;
>     // the map from Vertex_handle to indices
>     Unique_hash_map<Vertex_handle, std::size_t > V;
>
>The `operator<<` first displays:
> ````
>   <dimension>
>   <number of vertices>
> ````
> Then, for each vertices:
>  * it fills the `TV` and `V`
>  * and it calls the operator<< of the vertex
>
> Then:
>   * call `tds().print_cells(ostream, V);`
> 
> Then, for each cell, in dimension 3:
>   it calls the `operator<<` of the cell
> 
> Imagine that the vertex base has an extra field of type `Vertex_handle` (to 
> store for example the "parent vertex" of Steiner vertices, in Shewchuk 
> Delaunay refinement algorithm). To correctly output that triangulation, the 
> `operator<<` of the vertex should have access to the map `V`, but it does not have 
> access to it. What could be a solution for that?

maxGimeno:
## Summary of Changes
Adds functions to triangulations to be able to give access to the index/simplex and simplex/index maps to the `operator<</>>`  of the simplices.

It is the first part of #3850 
## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #861


maxGimeno:
Errors in c3t3_io

maxGimeno:

sloriot:
As the code cannot be used with Mesh_3 that requires a dedicated writing function (not at the TDS_3 level), we are closing this PR as it is not used.

PR #5146, opened 2020-11-05
Allow is_strongly_convex_3 with Surface_mesh.

mhsaar:

## Summary of Changes

Allow instantiation of CGAL::is_strongly_convex_3 with Surface_mesh.

## Release Management

* Affected package(s):

### DIFF
--- a/Convex_hull_3/include/CGAL/convexity_check_3.h
+++ b/Convex_hull_3/include/CGAL/convexity_check_3.h
@@ -14,12 +14,11 @@
 #ifndef CGAL_CONVEXITY_CHECK_3_H
 #define CGAL_CONVEXITY_CHECK_3_H
 
-#include <CGAL/license/Convex_hull_3.h>
-
-
-#include <CGAL/intersections.h>
-#include <CGAL/boost/graph/iterator.h>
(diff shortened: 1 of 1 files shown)


afabri:
Hello Michael, can you explain why this is needed. Is the problem not  that a   `const` is missing [here](https://github.com/CGAL/cgal/blob/de69ca0897ab9b11eaefb1026e5e99511be82624/Convex_hull_3/include/CGAL/convexity_check_3.h#L196).

Also the license check file should remain the first one to get `#include`. @sloriot right?

sloriot:
They are several issues that I'll fix asap.

mhsaar:
Agreed. I was certainly expecting that this won't be the final version.

afabri:
> 
> 
> Agreed. I was certainly expecting that this won't be the final version.

Thank you for pointing out !!!!

mhsaar:
Yes, but more because I was explicitly including Surface_mesh.

sloriot:
@mhsaar just to be clear the issues I mention are not in your patch but in the current code.

sloriot:
Replaced by #5158

PR #5151, opened 2020-11-09
C3t3 better io max gimeno

### LINKED ISSUE #2359: Mesh_3: Save/load c3t3 objects correctly
## Issue Details

For the moment, the I/O of C3t3 objects only save/load the information of the TDS_3, and the I/O of `C3t3::Triangulation::Vertex` and `C3t3::Triangulation::Cell` ensures that most of the information of the C3t3 is saved and restored.

But a C3t3 object has data that is not stored in the TDS. Some of it can be retrieved after a load by a call to the *undocumented* method `c3t3.rescan_after_load_of_triangulation()`, but the `c3t3.is_in_complex(Edge)` information cannot be retrieved. It should be stored in the file, and loaded from it.

This issue will require a solution to #861 first.

## Environment

* CGAL version: CGAL-4.11

Cc: @sloriot

maxGimeno:
## Summary of Changes

Save/load edges in complex in c3t3 operator<</>>
## Release Management

* Affected package(s):Mesh_3
* Issue(s) solved (if any): fix #2359


Mesh_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
```diff
@@ -547,7 +547,105 @@ class Mesh_complex_3_in_triangulation_3 :
...
+    {
+      if(is_ascii(os))
+      {
+        for(auto it = edges_.begin(); it != edges_.end(); ++it)
+        {
+          write(os, vertex_index_map[it->left]);
+          write(os, vertex_index_map[it->right]);
+          os << CGAL::oformat(it->info);
+        }
+      }
```
lrineau, line 578: The `if(is_ascii(os))` line 570 is in the `else` part of another `if(is_ascii(os))`.

And I suggest you regroup everything in one `if(is_ascii(os))/else`.

sloriot:
@lrineau should this PR be closed?

lrineau:
> @lrineau should this PR be closed?

No. Now that we have abandoned #5145, this PR seems the right thing to do.

Mesh_3/test/Mesh_3/test_c3t3_io.cpp
```diff
@@ -446,19 +481,135 @@ struct Test_c3t3_io {
...
+  v6->set_dimension(3);
+  v6->set_index(MD_1::get_sub_domain_index_1());
+
+  std::ios_base::openmode mode = std::ios_base::in | std::ios_base::out;
+  std::stringstream stream(mode);
+  CGAL::Mesh_3::save_binary_file(stream, c3t3, false);
+  assert(stream);
+  CGAL::Mesh_3::load_binary_file(stream, c3t3_2);
+  assert(stream);
+    std::cout << "Content of c3t3_1:\n"
```
sloriot, line 601: @lrineau dummy question: can we dump both c3t3 in `std::string` and assert that the strings are equal?

janetournois: IIRC it is what we do in the determinism test (but with the ASCII version, not with the binary version)

sloriot: So we can update the test and test this branch, right?

lrineau: Yes, I think so.

sloriot: @janetournois would you mind updating the test?

janetournois: Ok I'm doing it

janetournois:
@sloriot is this what you meant?
The assertion added in commit f11cc0f fails

janetournois:
It fails because `c3t3` (to be written) is output with its complex edges, so this is added to the end of the dump :
```
2
1 3 5
1 4 6
```
though `c3t3_2` does not have complex edges, they have not been loaded.

sloriot:
#7821 mentioned that it is similar. Shall we close this PR? @ange-clement @janetournois

janetournois:
they are similar and follow the same target, but for now they are complementary and should be "merged" (if I'm right, #7821 does not contain the "write" code for now)

PR #5162, opened 2020-11-16
CMake: Make it easier to find boost and qt on windows

### LINKED ISSUE #3529: Improvements to be done for CGAL CMake scripts

maxGimeno:
## Summary of Changes

Add paths to Qt5 and boost windows default installation directories to CMAKE_PREFIX_PATH in CMakeLists.

## Release Management

* Affected package(s):CGAL
* Issue(s) solved (if any): fix #3529


janetournois:
May I ask why you are using `BOOST_DIRS `?

[CMake documentation](https://cmake.org/cmake/help/latest/module/FindBoost.html) mentions `BOOST_ROOT` and [CGAL documentation](https://doc.cgal.org/latest/Manual/windows.html#title9) mentions `BOOST_LIBRARYDIR` and `BOOST_INCLUDEDIR`.

On my windows setup, I have the environment variable `BOOST_ROOT` which is enough to find everything.

maxGimeno:
@janetournois BOOST_DIRS is just a simple variable to store the list. The goal here is to detect if a directory is at the default location on windows so the user doesn't have to export BOOST_ROOT in a standard installation case.

CMakeLists.txt
```diff
@@ -57,6 +57,27 @@ if(BUILD_TESTING)
...
+  list(APPEND CMAKE_PREFIX_PATH "${boost_dir}")
+endif()#BOOST_DIRS
+
+file(GLOB QT_DIRS "C:/Qt/5")
+if(QT_DIRS)
+  list(GET QT_DIRS 0 qt_dir)
+  file(GLOB compilos "${qt_dir}/msvc*")
+  if(compilos)
+    list(GET compilos 0 compilo)
+    list(APPEND CMAKE_PREFIX_PATH "${compilo}/lib/cmake/Qt5")
```
lrineau, line 76: `/lib/cmake/Qt5` should not be part of `CMAKE_PREFIX_PATH`. The command `find_package` already knows where to search for in the prefix. For example, on certain 64 bits Linux systems, the path is `/lib64/cmake/Qt5`.

```suggestion
    list(APPEND CMAKE_PREFIX_PATH "${compilo}")
```

janetournois:
> 
> 
> @janetournois BOOST_DIRS is just a simple variable to store the list. The goal here is to detect if a directory is at the default location on windows so the user doesn't have to export BOOST_ROOT in a standard installation case.

Ok I understand.
Then, you should add `C:\Program Files (x86)` to the places to be checked.

CMakeLists.txt
afabri: Really "compilo" ????

afabri:
Did anybody of you install a pre-compiled version of boost?  The installer proposes `C:\local` as directory, and I would assume that at least some people accept what an installer proposes.
image

maxGimeno:
That's the information I was lookin for, but I only found Program Files...

maxGimeno:

PR #5166, opened 2020-11-17
Arrangement 2: Enhanced CCB Merging for Surface Sweep

sgiraudot:
_Release note: (Because 5.0 branch was deleted, [the previous PR](https://github.com/CGAL/cgal/pull/4875) was closed and we can't seem to open it again. Please refer to this PR for previous reviews.)_

## Summary of Changes

This PR replaces https://github.com/CGAL/cgal/pull/4851.

As said in the previous PR, _when merging 2 inner CCBs, all halfedges of one of the CCB are traversed again and updated, which can lead to an explosion of complexity, for example when computing the arrangement of a very large holes with many concavities_.

This new PR uses a similar approach than the previous one (using temporary redirections of inner CCBs towards a valid one and cleaning everything afterwards). The differences are:
- instead of union set, we simply keep, for each inner CCB, a pointer to another CCB: these pointers form chains to a valid CCB. These chains are shortened everytime we access an inner CCB. This is, in a way, a union find in-place. There is no need for a dependency to Union_find anymore
- we use a union of `Face*` / `Inner_ccb*` instead of the current face pointer in the inner CCB class, and we replace a boolean by an enum to determine if the inner CCB is valid (`Face*` is used) or not (it's just a garbage object that chains to a correct one via `Inner_ccb`). This mechanism makes the memory overhead = 0
- the valid inner CCB is always accessed through `Halfedge::inner_ccb()` without having to go through functions of AOS2, which makes the changes far less invasive and ensures we don't break anything (testsuite is fine)
- we use the construction visitor to make sure this mode is only used when sweeping (not when merging manually)

The resulting accelerations are the same (up to 40x faster in master's worst case scenario, 30% in common datasets).

## Release Management

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -483,10 +483,13 @@ public:
     if (out->is_valid())
       return out;
 
-    // else
-    out = const_cast<Inner_ccb*>(out)->reduce_path();
-    const_cast<Halfedge*>(this)->set_inner_ccb(out);
-    return out;
+    // else reduce path and get valid iccb
+    const Inner_ccb* valid = out->next();
+    while (!valid->is_valid())
+      valid = valid->next();
+    const_cast<Inner_ccb*>(out)->set_next(const_cast<Inner_ccb*>(valid));
+    const_cast<Halfedge*>(this)->set_inner_ccb(valid);
+    return valid;
   }
 
   /*! Get an incident inner CCB (non-const version).
@@ -500,10 +503,13 @@ public:
     if (out->is_valid())
       return out;
 
-    // else
-    out = out->reduce_path();
-    set_inner_ccb(out);
-    return out;
+    // else reduce path and get valid iccb
+    Inner_ccb* valid = out->next();
+    while (!valid->is_valid())
+      valid = valid->next();
+    out->set_next(valid);
+    set_inner_ccb(valid);
+    return valid;
   }
 
   Inner_ccb* inner_ccb_no_redirect()
@@ -798,9 +804,11 @@ private:
   Inner_ccb_iterator  iter;   // The inner CCB identifier.
   enum
   {
-    ITER_IS_SINGULAR,
-    ITER_IS_NOT_SINGULAR,
-    INVALID
+    ITER_IS_SINGULAR,     // singular = default iterator, not initialized
+    ITER_IS_NOT_SINGULAR, // not singular = iterator was assigned and is valid
+    INVALID               // invalid = the inner CCB is invalid and
+                          //           only links to another inner CCB
+                          //           in chain to valid CCB
   } status;
 
 public:
@@ -893,14 +901,6 @@ public:
     f_or_icc.icc = next;
   }
 
-  Arr_inner_ccb* reduce_path()
-  {
-    if (is_valid())
-      return this;
-    // else
-    f_or_icc.icc = f_or_icc.icc->reduce_path();
-    return f_or_icc.icc;
-  }
 };
 
 /*! \class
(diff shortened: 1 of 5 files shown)


maxGimeno:

lrineau:
@sgiraudot This PR is marked "not yet approved". Is it right?

sgiraudot:
Yes, @efifogel has not yet reviewed it (nor approved it).

maxGimeno:
This PR won't make it to 5.1 then.

sloriot:
@efifogel we need to get your review for this PR so that it can get integrated in the bugfix release. It's pretty short. Could you please have a quick look?

efifogel:
oh, sure, but it will take me a couple of days.
                         _/

efifogel:
I think that the change to the way inner-ccbs are handled that has been
applied already should also be applied to the way that outer ccbs are
handled. Moreover, it might be a good idea to make Arr_inner_ccb and
Arr_outer_ccb (defined in Arr_dcel_base.h) the same type.
In a planar arrangement there might be at most one outer-ccb, but this is
no longer the case in arrangements embedded on non-planar surfaces, such as
a sphere.
In the new Arrangement_on_surface_2 inner and outer ccbs are treated the
same, except for some rules (which I don't remember by heart) that
determine whether a ccb is inner or outer.
For example,
1. Consider a small circle on a sphere, which defines an arrangement of 2
faces. One face has an outer ccb, while the other has an inner ccb, but
which?
2. Consider an arrangement on a sphere that is divided into layers. (That
is, there are horizontal small circles that, sort of speaking, create
horizontal stripe-like faces.)
Here, it is unclear whether a boundary is inner or outer.
BW, the extra code that supports all this does not have a negative impact
on the performance of planer arrangement.
We have made sure of that.
On the other hand, I think it's possible to expedite arrangements of
segments.
For example, when sweeping arrangement of segments, the order of the
segments on the sweep-line swaps when several segments intersect in their
interiors.
As far as I remember, we have started to handle this case, but there is
still room to improve.

BW, I'm curious, what was the maximal gain that this optimization has
achieved?

                         _/

sgiraudot:
> I think that the change to the way inner-ccbs are handled that has been applied already should also be applied to the way that outer ccbs are handled. 

We detected a complexity issue with inner CCBs, not with outer CCBs, hence we created this new behavior specifically for inner CCBs.

> Moreover, it might be a good idea to make Arr_inner_ccb and Arr_outer_ccb (defined in Arr_dcel_base.h) the same type.

Possibly. However, that seems out of scope for this PR.

> BW, I'm curious, what was the maximal gain that this optimization has achieved?

As stated in the comment of this PR, in worst cases (or best, depending on you point of view), this PR makes code run 40x faster (4000% gain if you prefer). It changes a `O(n²)` algorithm to `O(n)` (`n` being the number of times 2 inner CCBs are merged during a sweep), so in some configurations it can be a tremendous change (such configurations are painfully slow  with the current CGAL releases).

afabri:
> so in some configurations

Maybe a word on what "slow" configurations are?   Does it depend on the size of the polygon, the number of holes,....

sgiraudot:
It's not easily described, but the obvious worst cases are cases with large holes that are heavily non-X-monotone. For example, imagine you have a large hole with integer point coordinates (points on a regular grid), then everytime you add a vertical segment, you have to merge 2 inner CCBs, which results in a `O(n²)` complexity, which results in very long computation times.

efifogel:
Applying the same changes to the outer ccb does seem to be in this scope.
I've started to write a long message, but come on---it can be completed in
a couple of hours.

                         _/

afabri:
A lot of things can be in a couple of hours, but let's do that in an independent PR.  If you agree that the PR as is improves arrangements then please just approve it.  If there was an API change I could understand that we wanted a single PR.

efifogel:
The Aos_2 major feature is in the making.
It has gone through some reviews already.
There, outer ccbs and inner ccbs are thought of as similar entities.
If this is not done now, I will be compelled to do it for this feature, but
it is extremely easier doing it now.
I don't understand the big deal.
I don't even mind doing it myself.
Now that it has been done for inner ccbs, it's really a matter of cut and
paste, and testing.
                         _/

efifogel:
If it's that important for you guys to close it now, then go ahead; I
approve.
It seems that I'll have time only next week.
                         _/

maxGimeno:

PR #5168, opened 2020-11-18
Fix use of uninitialized variable in File_scanner_OFF.h

GilesBathgate:
## Summary of Changes

Fix CWE-457: Use of Uninitialized Variable in File_scanner_OFF.h

If the function `get_color_from_line`, is called with the input stream containing only `#`, the while loop will exit at the break condition, the index will be 0 (i.e `index<2`) and `rgb[0]` will be dereferenced in the call to `get_indexed_color`. Since rgb is never initialised this could lead to undefined behaviour.

## Release Management

* Affected package(s): 3D Polyhedral Surface
* Issue(s) solved (if any):security/bugfix

### DIFF
--- a/Stream_support/include/CGAL/IO/File_scanner_OFF.h
+++ b/Stream_support/include/CGAL/IO/File_scanner_OFF.h
@@ -540,7 +540,7 @@ public:
         //split it into strings
         std::istringstream iss(col);
         //holds the rgb values
-        unsigned char rgb[3];
+        unsigned char rgb[3] {};
         int index =0;
         //split the string into numbers
         while(iss>>color_info){


Stream_support/include/CGAL/IO/File_scanner_OFF.h
```diff
@@ -540,7 +540,7 @@ class CGAL_EXPORT File_scanner_OFF : public File_header_OFF {
         //split it into strings
         std::istringstream iss(col);
         //holds the rgb values
-        unsigned char rgb[3];
+        unsigned char rgb[3] {0,0,0};
```
lrineau, line 543: I prefer the following value-initialization:
```suggestion
        unsigned char rgb[3] {};
```
otherwise there might be compilers that will warn about the conversion between `int` and `unsigned char`.

The effect is the same.

GilesBathgate: Fixed

lrineau (approved):

maxGimeno:
This is already fixed in #4255

GilesBathgate:
@maxGimeno Agreed, `File_scanner_OFF.h` was deleted entirely in that PR. Thank you.

maxGimeno:
it was actually moved to Stream_support/include/CGAL/IO/OFF/, but the `rgb` variable doesn't existe anymore.

GilesBathgate:
But it seems to be defined unintialised here:
https://github.com/maxGimeno/cgal/blob/CGAL_IO-maxGimeno/Stream_support/include/CGAL/IO/OFF/File_scanner_OFF.h#L538

GilesBathgate:
@maxGimeno Are you thinking of #5185 ? 

rgb seems to be  defined  uninitialised here:

https://github.com/maxGimeno/cgal/blob/CGAL_IO-maxGimeno/Stream_support/include/CGAL/IO/OFF/File_scanner_OFF.h#L538

maxGimeno:
Right, I must have missed my grep somehow...

GilesBathgate:
This is now merged into https://github.com/maxGimeno/cgal/tree/CGAL_IO-maxGimeno

PR #5170, opened 2020-11-18
Installation: Fix in Qt5 module

maxGimeno:
## Summary of Changes
Check if /demo exists before copying its content.

It might look like I'm only treating a symptom instead of the disease, but this error only seems to occur in the "library" tarball, from which it is impossible to build the demos because they are not included anyway. 
## Release Management

* Affected package(s):Installation, CGAL_Qt5

### DIFF
--- a/GraphicsView/src/CGAL_Qt5/CMakeLists.txt
+++ b/GraphicsView/src/CGAL_Qt5/CMakeLists.txt
@@ -42,14 +42,25 @@ install(
   DESTINATION "${CGAL_INSTALL_INC_DIR}/CGAL/Qt"
   COMPONENT CGAL_Qt5)
 if(CGAL_HEADER_ONLY)
-  install(
-    DIRECTORY "../../demo/resources/"
-    DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/resources"
-    COMPONENT CGAL_Qt5)
-  install(
-    DIRECTORY "../../demo/icons/"
-    DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/icons"
-    COMPONENT CGAL_Qt5)
+  if(EXISTS ${CGAL_MODULES_DIR}/demo/resources)
+    install(
+      DIRECTORY "${CGAL_MODULES_DIR}/demo/resources/"
+      DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/resources"
+      COMPONENT CGAL_Qt5)
+    install(
+      DIRECTORY "${CGAL_MODULES_DIR}/demo/icons/"
+      DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/icons"
+      COMPONENT CGAL_Qt5)
+  else()
+    install(
+      DIRECTORY "../../demo/resources/"
+      DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/resources"
+      COMPONENT CGAL_Qt5)
+    install(
+      DIRECTORY "../../demo/icons/"
+      DESTINATION "${CGAL_INSTALL_CMAKE_DIR}/demo/icons"
+      COMPONENT CGAL_Qt5)
+  endif()
 endif()
 
 message("libCGAL_Qt5 is configured")
--- a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
+++ b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
@@ -158,6 +158,8 @@ if(EXISTS ${GIT_REPO}/Maintenance/release_building/public_release_name)
   file(COPY "${GIT_REPO}/Maintenance/release_building/public_release_name"
     DESTINATION "${release_dir}/doc")
 endif()
+file(COPY ${GIT_REPO}/GraphicsView/demo/resources ${GIT_REPO}/GraphicsView/demo/icons
+  DESTINATION "${release_dir}/cmake/modules")
 
 #create VERSION
 file(WRITE ${release_dir}/VERSION "${CGAL_VERSION}")


lrineau:
`make install` does not work, from `CGAL-5.2-beta2-library.tar.xz`:
```
CMake Error at src/CGAL_Qt5/cmake_install.cmake:50 (file):
  file INSTALL cannot find
  "/home/lrineau/CGAL/Releases/CGAL-5.2-beta2/src/CGAL_Qt5/../../demo/resources":
  No such file or directory.
  src/cmake_install.cmake:62 (include)
  cmake_install.cmake:47 (include)
```

Please test the creation of a tarball with:
```
cmake -DPUBLIC=ON -DGENERATE_TARBALLS=ON -P Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```
and then test the tarball.

We should probably test the tarballs from Github Actions or Travis, to avoid that in the future.

lrineau (changes requested):
`make install` does not work, from `CGAL-5.2-beta2-library.tar.xz`:
```
CMake Error at src/CGAL_Qt5/cmake_install.cmake:50 (file):
  file INSTALL cannot find
  "/home/lrineau/CGAL/Releases/CGAL-5.2-beta2/src/CGAL_Qt5/../../demo/resources":
  No such file or directory.
  src/cmake_install.cmake:62 (include)
  cmake_install.cmake:47 (include)
```

Please test the creation of a tarball with:
```
cmake -DPUBLIC=ON -DGENERATE_TARBALLS=ON -P Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```
and then test the tarball.

We should probably test the tarballs from Github Actions or Travis, to avoid that in the future.

maxGimeno:
Do we want to be able to build our demos using this tar or not ?

lrineau:
> Do we want to be able to build our demos using this tarball or not ?

I do not know. We should ask their opinion to those who asked for that tarball. @afabri @sloriot, about the tarball of CGAL without the examples and demos, with only headers and CMake files, named `CGAL-5.2-beta1-library.tar.xz`, would you expect it to allow the build of a CGAL demo?

sloriot:
As a user I would not understand if I cannot build demo/examples of the same release just because they are from different archive.

maxGimeno:
Then I hope moving resources and icons from GraphicsView/demo to something else will be enough.

lrineau:
The code that has an issue, now, is the following one:
https://github.com/CGAL/cgal/blob/2f4d00af153408e0e66c12ac114b9332acb212af/Installation/cmake/modules/CGAL_Qt5_moc_and_resource_files.cmake#L23-L41

In the tarball `CGAL-5.2-beta1-library.tar.xz`, the `demo/` subdirectory has been removed, and thus the `else()` of the `if(...)` is evaluated. I suggest that, during the construction of the release tarball, we copy `GraphicsView/demo/resources/` and `GraphicsView/demo/icons/` into the directory `cmake/modules/` of the tarball. That way, the `else()` will always work:
  - with the big tarball where `demo/` exists, the "then-" part of the `if()`, that is more efficient, will be used,
  - and in the smaller tarball where `demo/` has been suppressed, then the files in `cmake/modules/` will be used, as if that was an installed CGAL version.

maxGimeno:
>I suggest that, during the construction of the release tarball, we copy GraphicsView/demo/resources/ and GraphicsView/demo/icons/ into the directory cmake/modules/ of the tarball. 

I agree. To me, that is the best (and easier) solution.

lrineau:
> > I suggest that, during the construction of the release tarball, we copy GraphicsView/demo/resources/ and GraphicsView/demo/icons/ into the directory cmake/modules/ of the tarball.
> 
> I agree. To me, that is the best (and easier) solution.

So, @sloriot, that would be that patch, right?
```patch
diff --git a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
index cfaa4eaaa8c..3f2cce6a01d 100644
--- a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
+++ b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
@@ -158,6 +158,8 @@ if(EXISTS ${GIT_REPO}/Maintenance/release_building/public_release_name)
   file(COPY "${GIT_REPO}/Maintenance/release_building/public_release_name"
     DESTINATION "${release_dir}/doc")
 endif()
+file(COPY ${GIT_REPO}/GraphicsView/demo/resources ${GIT_REPO}/GraphicsView/demo/icons
+  DESTINATION "${release_dir}/cmake/modules")
 
 #create VERSION
 file(WRITE ${release_dir}/VERSION "${CGAL_VERSION}")
```

lrineau:
> > > I suggest that, during the construction of the release tarball, we copy GraphicsView/demo/resources/ and GraphicsView/demo/icons/ into the directory cmake/modules/ of the tarball.
> > 
> > 
> > I agree. To me, that is the best (and easier) solution.
> 
> So, @sloriot, that would be that patch, right?
> 
> ```diff
> diff --git a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
> index cfaa4eaaa8c..3f2cce6a01d 100644
> --- a/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
> +++ b/Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
> @@ -158,6 +158,8 @@ if(EXISTS ${GIT_REPO}/Maintenance/release_building/public_release_name)
>    file(COPY "${GIT_REPO}/Maintenance/release_building/public_release_name"
>      DESTINATION "${release_dir}/doc")
>  endif()
> +file(COPY ${GIT_REPO}/GraphicsView/demo/resources ${GIT_REPO}/GraphicsView/demo/icons
> +  DESTINATION "${release_dir}/cmake/modules")
>  
>  #create VERSION
>  file(WRITE ${release_dir}/VERSION "${CGAL_VERSION}")
> ```

@maxGimeno We have to move on. Please test this patch. The command to generate release tarballs is:
```shell
cmake -DPUBLIC=ON -DGENERATE_TARBALLS=ON -P Scripts/developer_scripts/cgal_create_release_with_cmake.cmake
```

Once the patch is tested, replace the current PR by this patch, and we will merge it today, before the week-end.

sloriot:
I don't get why resources must be shipped with the lib and not with the demo.

lrineau:
> I don't get why resources must be shipped with the lib and not with the demo.

Because those files are parts of what we call "CGAL_Qt5". If a user has `CGAL-5.2-beta1-library.tar.xz` and the sources of a demo, then that user should be able to compile the demo, without any additional files.

Issue #5181, opened 2020-11-24
Uninitialized members - static analysis.


GilesBathgate:
## Issue Details

CWE-456: I use static analysis on my project provided by Synopsis Coverity Scan. Amongst other issues, there are approx 23 instances of class constructors not initialising all pointer, and scalar member variables. This is a trivial issue but one for which I have been able to create patches. I will submit these as pull requests.

## Source Code

Before applying patches my project has the following analysis results:

![coverity-defects](https://user-images.githubusercontent.com/444143/100069259-728bbb00-2e30-11eb-9952-3bf500a35b52.png)
 
After applying the patches the results are as follows:

![coverity-defects-partially-resolved](https://user-images.githubusercontent.com/444143/100070813-44a77600-2e32-11eb-9783-6794458d2c1f.png)

## Environment

* Operating system (Linux, 64 bits):
* Compiler: gcc
* Release or debug mode: release
* CGAL version: 5.0.2
* Boost version: 1.71.0
* Other libraries versions if used (Eigen, TBB, etc.): Qt 5.12.8

GilesBathgate:
The two `High` issues are fixed by #5169 and #5168

GilesBathgate:
One of the `Low` issues (Copying large values is inefficient, consider passing by reference; size thresholds for detection can be adjusted.) was fixed by #5143

lrineau:
https://github.com/CGAL/cgal/pull/5189 is only a partial fix.

PR #5187, opened 2020-11-24
Fix uninitialized member in Filtered_predicate.h

GilesBathgate:
## Summary of Changes

Fix uninitialized members (c2e, c2a) in Filtered_predicate.h

## Release Management

* Affected package(s): Filtered_kernel

### DIFF
--- a/Filtered_kernel/include/CGAL/Filtered_predicate.h
+++ b/Filtered_kernel/include/CGAL/Filtered_predicate.h
@@ -59,6 +59,7 @@ public:
   // AP::result_type must be convertible to EP::result_type.
 
   Filtered_predicate()
+    : c2e(), c2a()
   {}
 
   // These constructors are used for constructive predicates.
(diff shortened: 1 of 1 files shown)


Filtered_kernel/include/CGAL/Filtered_predicate.h
```diff
@@ -59,6 +59,7 @@ class Filtered_predicate
   // AP::result_type must be convertible to EP::result_type.
 
   Filtered_predicate()
+    : c2e{}, c2a{}
```
mglisse, line 62: Does it even complain about not initializing trivial empty types? :roll_eyes: 
I think this patch is ok, although I would have used parentheses for consistency with the other constructors. Don't you need to do the explicit constructor below as well?

GilesBathgate: My understanding is that trivial empty types (AKA class types) are always default initialised, so this does look like Coverity made a false positive, however, C2E, and C2A are template parameters, so I presume that the tool thinks you can pass pointer or scalar types for those parameters.  
e.g:
```c++
Filtered_predicate<EP,AP,C2E*,C2A*> pred;
```
That was my assumption.

PR #5189, opened 2020-11-24
Fix uninitialized member in Gmpz_type.h

GilesBathgate:
## Summary of Changes

Fix warning about uninitialized member (mpZ) in Gmpz_type.h

## Release Management

### DIFF
--- a/Number_types/include/CGAL/GMP/Gmpz_type.h
+++ b/Number_types/include/CGAL/GMP/Gmpz_type.h
@@ -43,7 +43,7 @@ struct Gmpz_rep
 // maybe the mpz_init_set* functions should move back to Gmpz_rep.
 // But then we should use the Storage_traits::construct/get...
 
-  mpz_t mpZ;
+  mpz_t mpZ; /* coverity[member_decl] */
 
   Gmpz_rep() {}
   ~Gmpz_rep() { mpz_clear(mpZ); }


Number_types/include/CGAL/GMP/Gmpz_type.h
```diff
@@ -45,7 +45,7 @@ struct Gmpz_rep
 
   mpz_t mpZ;
 
-  Gmpz_rep() {}
+  Gmpz_rep() : mpZ{} {}
```
mglisse, line 48: But
```c++
mpz_t z;
mpz_init(z);
```
is the documented way to use `mpz_t`, it is meant to be passed uninitialized to `mpz_init`...

GilesBathgate: @mglisse I agree, but the comment https://github.com/CGAL/cgal/blob/b9e320659e41c255d82642d03739150779f19575/Number_types/include/CGAL/GMP/Gmpz_type.h#L41-L44
Suggested this wasn't desirable.

The Coverity scan complains that:
```
2. uninit_member: Non-static class member field mpZ._mp_alloc is not initialized in this constructor nor in any functions that it calls.
4. uninit_member: Non-static class member field mpZ._mp_size is not initialized in this constructor nor in any functions that it calls.
6. uninit_member: Non-static class member field mpZ._mp_d is not initialized in this constructor nor in any functions that it calls.
```
so my fix was to default initialize the single-element array of struct (mpz_t).

mglisse: The comment suggests reorganizing the code so that mpz_init* is called in the constructor (that doesn't seem super convenient), not zeroing the mpz_t. IIRC GMP does not guarantee what happens if you call mpz_clear on a zero-initialized mpz_t, so this does not really answer the comment.

GilesBathgate: I'd be surprised if GMP guarantees what happens if you call mpz_clear on an uninitialized mpz_t too  :wink:

There is a `mpn_zero` function https://gmplib.org/manual/Low_002dlevel-Functions but I couldn't figure out how to use it in this context.

mglisse: mpn_zero would be useful to zero what the mpz_t points to, not the mpz_t itself, so it isn't relevant.

GilesBathgate: Added a coverity suppression comment instead as requested

mglisse:
ok with me.

maxGimeno:

PR #5191, opened 2020-11-24
Fix uninitialized member in Mpzf.h

GilesBathgate:
## Summary of Changes

Fix warnings about uninitialized members (data_, size, exp) in Mpzf.h

## Release Management

### DIFF
--- a/Number_types/include/CGAL/Mpzf.h
+++ b/Number_types/include/CGAL/Mpzf.h
@@ -253,15 +253,21 @@ struct Mpzf {
   typedef Mpzf_impl::no_pool<mp_limb_t*,Mpzf> pool;
 //#endif
 
-  mp_limb_t* data_; /* data_[0] is never 0 (except possibly for 0). */
+  /* data_[0] is never 0 (except possibly for 0). */
+  mp_limb_t* data_; /* coverity[member_decl] */
+
   inline mp_limb_t*& data() { return data_; }
   inline mp_limb_t const* data() const { return data_; }
 
 #ifdef CGAL_MPZF_USE_CACHE
   mp_limb_t cache[cache_size + 1];
 #endif
-  int size; /* Number of relevant limbs in data_. */
-  int exp; /* The number is data_ (an integer) * 2 ^ (64 * exp). */
+  /* Number of relevant limbs in data_. */
+  int size; /* coverity[member_decl] */
+
+  /* The number is data_ (an integer) * 2 ^ (64 * exp). */
+  int exp; /* coverity[member_decl] */
+
   typedef int Exponent_type;
   typedef int Size_type;
 
@@ -300,6 +306,8 @@ struct Mpzf {
     pool::push(data());
   }
 
+  // These constructors intentionally
+  // don't initialize any members
   Mpzf(noalloc){}
   Mpzf(allocate,int i) { init(i); }
 


mglisse:
Hello,
does your static analysis tool describe an execution path where those members could be used uninitialized? Or is this just some bogus coding guideline "initialize everything"? I don't like such guidelines because they can cause extra code generation, and most importantly they are counter-productive because they prevent better tools from detecting a real issue. Why would 0 be the right value if I somehow messed up and forgot to set the value somewhere?
If this is just to pacify a bad tool, I'd prefer to hide such code behind a macro that people using that tool can define.

GilesBathgate:
@mglisse Yes, the tool is very comprehensive, for example, it shows initialisation paths where scalar fields are neither initialised in the member-init-list or the constructor body.

 It can also detect execution paths where the member variable is used, but in the case of this PR it hasn't detected this as **use of uninitialized variable** that would be a **High** severity issue, whereas this one is **Medium** because it would have to be accompanied by accompanying user code which calls the constructor and then expects size,exp to be sensible values. Whether that is bogus or not I am not sure, it seems to be the implication of the core cpp guidelines https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-always which mentions that, this rule covers member variables.

I am happy to dismiss these in the static analysis tool, any of these PR's as false positive, or intentional if that is the consensus.

mglisse:
> Yes, the tool is very comprehensive, for example, it shows initialisation paths where scalar fields are neither initialised in the member-init-list or the constructor body.

Where can I see such a path?
The 2 constructors modified by this patch are private, so a user cannot call them directly. They are used internally, and each of those places supposedly initializes the fields (if I missed such a place, that would be good to know).

If we change something to follow the guidelines (and they mention several complicated techniques), I don't think we should put a bogus value like data_=nullptr (if you try to use or even destruct such a Mpzf, bad things may happen).

GilesBathgate:
> The 2 contructors modified by this patch are private, so a user cannot call them directly.

For what its worth @mglisse, I did suspect that these constructors were potential false positive, given that they require explicit struct `noalloc` to flag no allocation. I presume no allocation implies no initialisation either?

Anyway, regarding where you can see further details of these issues, you can either create a project and register with Synopsys Coverity Scan yourself (its provided free from open source projects) or I can add you as a member to my project.

https://scan.coverity.com/projects/gilesbathgate-rapcad

> I don't think we should put a bogus value like data_=nullptr (if you try to use or even destruct such a Mpzf, bad things may happen).

For sure dereferencing a null pointer is bad, is dereferencing an uninitialized pointer not equally as bad? If not a potential security flaw?

GilesBathgate:
@mglisse Is an acceptable solution to remove the noalloc constructor:

For example:
```cpp
static Mpzf aors(Mpzf const&a, Mpzf const&b, int bsize){
    if(bsize==0){
      Mpzf res(allocate(),std::abs(a.size))
      res.exp=a.exp;
      res.size=a.size;
      if(size!=0) mpn_copyi(res.data(),a.data(),size);
      return res;
    }
    int asize=a.size;
    if(asize==0){
      Mpzf res(allocate(),std::abs(bsize));
      res.exp=b.exp;
      res.size=bsize;
      mpn_copyi(res.data(),b.data(),size);
      return res;
    }
...
```

Potentially passing in exp too?

GilesBathgate:
@mglisse Like this? https://github.com/GilesBathgate/cgal/commit/736fd86dddd935d60d1d1202b3b059a9810f86d8

mglisse:
> @mglisse Like this? [GilesBathgate@736fd86](https://github.com/GilesBathgate/cgal/commit/736fd86dddd935d60d1d1202b3b059a9810f86d8)
> 
> Or see the branch: [master...GilesBathgate:fix-uninitialized-member-mpzf2](https://github.com/CGAL/cgal/compare/master...GilesBathgate:fix-uninitialized-member-mpzf2)

I think some compilers perform NRVO with the current code but not with your patch. You introduce an uninitialized `int exp;` which the tool will warn about :wink: . There are a few unnecessary writes of 0. I think this is a good try (thank you for trying this different version), but I am not convinced this makes the code better.

My comment probably belongs in a more general discussion, but to me uninitialized variables are a feature, and sanitizers and some static analysis tools can use it to find bugs (real ones, not style violations). Forcing an initialization to an arbitrary value prevents them from detecting bugs (and may generate slower code). Refactoring the code so that we always have a meaningful value to initialize variables with can make sense, but it should not be done to the detriment of generated code or code readability (some examples in the guidelines look pretty bad IMO).

Sometimes initializing a pointer to 0 can be used as a security measure (it increases the chances of a segfault over letting an attacker leave some well-crafted address where the code might use it), but it assumes that the code is broken to begin with, and should only be used as a last resort in the final binary, protected by some macro (IMO, apparently a number of people disagree :disappointed: ).

GilesBathgate:
>I think some compilers perform NRVO with the current code but not with your patch. 

Yes, it's very annoying that the cpp spec doesn't allow this to be explicitly specified (RVO/NRVO), and generate a compile error if it cannot be done.
> int exp; which the tool will warn about.

I don't think it does, because it doesn't require initialisation inline with the declaration, that is more of a style guideline. I think the tool is clever enough to see that on both the if and else, result in `exp` being initialised. Either way, this could potentially be fixed with the ternary operator

```
int exp = (aexp<bexp)? a.exp:b.exp;
if(aexp<bexp)...
```
at the cost of some code duplication. I will run this patch through Coverity and see if it flags up any errors. I can create benchmarks, to see if the performance impact is measureable.

GilesBathgate:
@mglisse Regarding NRVO, Is this example using it: https://godbolt.org/z/TbhGGr whereas some compilers only support it if it's written as https://godbolt.org/z/86eP9b

mglisse:
> Regarding NRVO, Is this example using it: https://godbolt.org/z/TbhGGr whereas some compilers only support it if it's written as https://godbolt.org/z/86eP9b

I find it easier to see if I add `Foo(){}Foo(Foo const&);` to the class, then I can just look for a call to the copy constructor in the generated code. And switch to gcc, which is not very good at NRVO (in the second version, just replacing `{` and `}` with `{{` and `}}` around the function body is enough for gcc to stop using NRVO.

GilesBathgate:
> I'd prefer to hide such code behind a macro that people using that tool can define.

@mglisse Reading up on the Coverity docs, I believe the error can be suppressed using

```c++
// coverity[UNINIT_CTOR]
Mpzf(noalloc)
```

If this is what you prefer?

I think size and exp are known for all cases with the `allocate()` constructor, would it be better to pass these in the constructor and remove the `struct allocate{}`?

mglisse:
> @mglisse Reading up on the Coverity docs, I believe the error can be suppressed using
> 
> ```c++
> // coverity[UNINIT_CTOR]
> Mpzf(noalloc)
> ```
> 
> If this is what you prefer?

I like this, yes, it does not affect other tools, and documents quite nicely that not initializing the fields is done on purpose in this particular constructor. Although I don't know if it applies recursively to functions calling this one or not, it would be better here if it didn't, so that it can still notice if a function calling this one forgets to initialize some fields.

> I think size and exp are known for all cases with the `allocate()` constructor, would it be better to pass these in the constructor and remove the `struct allocate{}`?

The first one I look at, `operator*`, only knows size at the very end. It seems easier to give it the same comment as the noalloc case (same for the Gmpz_type.h patch).

GilesBathgate:
@mglisse Like this: https://github.com/CGAL/cgal/compare/master...GilesBathgate:fix-uninitialized-member-mpzf3

mglisse:
As I said, I'd rather add the comment on both constructors and not change the code.

GilesBathgate:
Ok, if you prefer not to change the code, I'll just add the comments to both, but if you are willing to embrace change, I have one more approach for your consideration:

https://github.com/CGAL/cgal/compare/master...GilesBathgate:fix-uninitialized-member-mpzf4

Although there isn't a very compelling reason for this change, other than consistency.

I guess if this introduces a NRVO optimisation in `Mpzf_sqrt` it would be insignificant in comparison to computing the square root. :man_shrugging:

mglisse:
> @mglisse Ok, if you prefer not to change the code, I'll just add the comments to both, but if you are willing to embrace change, I have one more approach for your consideration:
> 
> [master...GilesBathgate:fix-uninitialized-member-mpzf4](https://github.com/CGAL/cgal/compare/master...GilesBathgate:fix-uninitialized-member-mpzf4)
> 
> Although there isn't a very compelling reason for this change, other than consistency.

I am not sure why you are so eager to change the code, but I guess this one is acceptable, yes.

GilesBathgate:
@mglisse I am just testing the supression comments It doesn't work quite as I expected:

>uninit_member - A class member or member field is uninitialized along this path in the constructor. The event occurs at the end of the path in the constructor. If a particular variable is always initialized before use (perhaps, outside the constructor), suppress this event to indicate that the declaration site is not a declaration of a potentially uninitialized member. Suppressing this event is the most severe suppression method. You will never receive an error if this member is uninitialized in any constructor for the class.
>member_decl - A class member has been uninitialized along a path in the constructor. If the particular path in question does contain an initialization, or the lack of initialization reliably deemed to be benign (perhaps, due to correctly initializing it before being used outside the constructor), the appropriate fix is to suppress this event.

GilesBathgate:
@mglisse I cannot seem to get these suppression comments to work. Either I am not understanding how to use them correctly, or there is an issue with coverity scan. I am trying one last thing, but if this fails, I propose the change that removes the `allocate()` constructor, which will at least eliminate one error.

GilesBathgate:
@mglisse Success!
```
Classification: | Unclassified | → | Intentional
Classification changed due to source code annotation or triage information passed in during commit.
```

mglisse:
It is a bit sad that suppression can only be done at the level of the member (all constructors may leave it uninitialized), and not at the level of the constructor (this constructor may leave some members uninitialized).
If there is consensus that we want to make coverity happy, this patch is ok with me.

lrineau:
>  If there is consensus that we want to make coverity happy, this patch is ok with me.

Is there a consensus... good question...

@GilesBathgate can you please give an URL to the documentation of Coverity where the suppression comments are documented?

GilesBathgate:
> @GilesBathgate can you please give an URL to the documentation of Coverity where the suppression comments are documented?

@lrineau I'd love to, unfortunately they are hidden behind a login screen! here is the URL none the less. https://scan4.coverity.com/doc/en/cov_checker_ref.html#codeannotate

PR #5194, opened 2020-11-24
Fix uninitialized member in Segment_2_Segment_2.h

GilesBathgate:
## Summary of Changes

Added NOT_COMPUTED_YET to enum (similar enum member is present in Line_2_Line_2.h)

Fix uninitialized member  (_result) in Segment_2_Segment_2.h

## Release Management

* Affected package(s): Intersections_2

### DIFF
--- a/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
+++ b/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
@@ -274,10 +274,10 @@ do_intersect(const typename K::Segment_2 &seg1,
 template <class K>
 class Segment_2_Segment_2_pair {
 public:
-    enum Intersection_results {NOT_COMPUTED_YET, NO_INTERSECTION, POINT, SEGMENT};
+    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT, UNKNOWN};
     Segment_2_Segment_2_pair(typename K::Segment_2 const *seg1,
                             typename K::Segment_2 const *seg2)
-            : _seg1(seg1), _seg2(seg2), _known(false), _result(NOT_COMPUTED_YET) {}
+            : _seg1(seg1), _seg2(seg2), _known(false), _result(UNKNOWN) {}
 
     Intersection_results intersection_type() const;
 
--- a/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
+++ b/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
@@ -274,10 +274,10 @@ do_intersect(const typename K::Segment_2 &seg1,
 template <class K>
 class Segment_2_Segment_2_pair {
 public:
-    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT};
+    enum Intersection_results {NOT_COMPUTED_YET, NO_INTERSECTION, POINT, SEGMENT};
     Segment_2_Segment_2_pair(typename K::Segment_2 const *seg1,
                             typename K::Segment_2 const *seg2)
-            : _seg1(seg1), _seg2(seg2), _known(false) {}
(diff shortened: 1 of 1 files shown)


mglisse:
The core guidelines say
> Using a value representing "uninitialized" is a symptom of a problem and not a solution

(in case we want to pay attention to them, but we probably don't)

GilesBathgate:
I agree. It's the same solution however as used in Line_2_Line_2.h . I am not adamant on following guidelines, or strictly adhering to requirements dictated by static analysis tools of any kind. But I also don't have the knowledge to say that all of these issues are intentional or false positives. My motivation was just to reduce my own projects defect density.

Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
```diff
@@ -274,10 +274,10 @@ do_intersect(const typename K::Segment_2 &seg1,
 template <class K>
 class Segment_2_Segment_2_pair {
 public:
-    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT};
+    enum Intersection_results {NOT_COMPUTED_YET, NO_INTERSECTION, POINT, SEGMENT};
```
sloriot, line 277:
```suggestion
    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT, UNKNOWN};
```

Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
```diff
@@ -274,10 +274,10 @@ do_intersect(const typename K::Segment_2 &seg1,
...
     Segment_2_Segment_2_pair(typename K::Segment_2 const *seg1,
                             typename K::Segment_2 const *seg2)
-            : _seg1(seg1), _seg2(seg2), _known(false) {}
+            : _seg1(seg1), _seg2(seg2), _known(false), _result(NOT_COMPUTED_YET) {}
```
sloriot, line 280:
```suggestion
            : _seg1(seg1), _seg2(seg2), _known(false), _result(UNKNOWN) {}
```

GilesBathgate:
@sloriot Regarding renaming the enum value to `UNKNOWN`, I was trying to use consistent naming as present in `Line_2_Line_2.h` (although, I can understand the change in order)

Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
```diff
@@ -274,10 +274,10 @@ do_intersect(const typename K::Segment_2 &seg1,
 template <class K>
 class Segment_2_Segment_2_pair {
 public:
-    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT};
+    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT, UNKNOWN};
     Segment_2_Segment_2_pair(typename K::Segment_2 const *seg1,
                             typename K::Segment_2 const *seg2)
-            : _seg1(seg1), _seg2(seg2), _known(false) {}
+            : _seg1(seg1), _seg2(seg2), _known(false), _result(UNKNOWN) {}
```
lrineau, line 280: It seems the code assumes:
```c++
assert(_known == (_result != UNKNOWN));
```
as an invariant. According to the [Don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, the Boolean variable `_known` should be removed.

GilesBathgate: @lrineau I agree. The pattern of using `_known` seems to be present in several other places and is always initialized, whereas `_result` isn't. I think the pattern of using an 'Unkown' enum value is simpler and more clearly understood. But it would be nice as part of this PR to make it consistent across all. I can do so if wanted.

sloriot: I'll take care of it.

PR #5202, opened 2020-11-24
Fix uninitialized members in Multiset.h

GilesBathgate:
## Summary of Changes

Moved dummy member initialisation into member-init-list
Fix uninitialized members (object, color) in Multiset.h

## Release Management

* Affected package(s): STL_extension

### DIFF
--- a/STL_Extension/include/CGAL/Multiset.h
+++ b/STL_Extension/include/CGAL/Multiset.h
@@ -102,8 +102,13 @@ protected:
     Node        *rightP;           // Points on the right child of the node.
     Node        *leftP;            // Points on the left child of the node.
 
-    /*! Default constructor. */
-    Node() :
+    /*!
+     * Constructor for dummy nodes
+     * \param _color The color of the node.
+     */
+    Node(Node_color _color) :
+      object{},
+      color(_color),
       parentP(nullptr),
       rightP(nullptr),
       leftP(nullptr)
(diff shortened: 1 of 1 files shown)


sloriot:
Thanks for all you PRs. It would be easier for us (and for you too maybe) if you group all your changes in a single PR. Would it be possible for you or shall I do it (by cherry-picking the commits for example)?

sloriot:
You don't have to list the packages modified when several are impacted.

GilesBathgate:
> Thanks for all you PRs. It would be easier for us (and for you too maybe) if you group all your changes in a single PR. Would it be possible for you or shall I do it (by cherry-picking the commits for example)?

I can merge these all into a single branch (octopus merge, or cherry-pick) I thought it would be better to keep these changes atomic, so if there needs to be any changes or discussion around them they can be treated individually.

sloriot:
Maybe @maxGimeno can comment as he is handling the testing but I would say it is simpler if all "similar" fixes goes into the same PR. We can comment in the diff anyway.

maxGimeno:
Actually I think it is better this way. There might be an impact on performance due to initialization of some of those variables, and we will have to review each of those PRs individually, so it will probably be easier if they are split.

GilesBathgate:
@maxGimeno I've not been able to measure any significant impact in performance but I would be willing to provide any before/after benchmarks if you require.

PR #5212, opened 2020-11-27
Force pyquery 1.4.1

maxGimeno:
## Summary of Changes
The last pyquery version to be compatible with python 2, which we still use in outr scripts, is 1.4.1. 
## Release Management

* Affected package(s):CI


lrineau:
What not there a plan to use Python 3.x?

PR #5214, opened 2020-11-27
Fix dead code warning in Mpzf.h

GilesBathgate:
## Summary of Changes

In the following code the `if mini <= cache_size` statement, where cache_size is defined as 8 the only way it can reach the code after the If body, is when mini is > 8, this means that the later check `if(mini < 2)` is always false.
 
https://github.com/CGAL/cgal/blob/97123fc3c725d2d9d489d7c70d88f3b1af5c572d/Number_types/include/CGAL/Mpzf.h#L272-L285

This commit just adds an  `#ifndef CGAL_MPZF_USE_CACHE` around the second if, since its only relevant when not using the cache. This wards off the error in static code analysis.

## Release Management

* Feature/Small Feature (if any): small/bugfix

### DIFF
--- a/Number_types/include/CGAL/Mpzf.h
+++ b/Number_types/include/CGAL/Mpzf.h
@@ -281,7 +281,9 @@ struct Mpzf {
       if(data()[-1] >= mini) return; // TODO: when mini==2, no need to check
       delete[] (data() - (pool::extra+1)); // too small, useless
     }
+#ifndef CGAL_MPZF_USE_CACHE
     if(mini<2) mini=2;
+#endif
     data() = (new mp_limb_t[mini+(pool::extra+1)]) + (pool::extra+1);
     data()[-1] = mini;
   }


mglisse:
I think I like the shorter code (without `#ifdef`) better, increasing mini if needed is the default behavior, and not checking for it when we know it isn't needed is an optimization (that compilers perform quite easily AFAIK). Static analysis tools probably handle things like that differently with templates (where the comparison can be redundant for some parameters but not others), but it isn't surprising that they do not know about macros.
Also, CGAL_MPZF_USE_CACHE is not the only condition that makes removing this code safe. If I reduce cache_size and increase the minimal allocation, the test can become relevant again even with CGAL_MPZF_USE_CACHE. Of course we can always remove this `#ifdef` if that ever happens.
I'll let someone else decide (I don't mind the patch, it doesn't clutter the file too badly).

maxGimeno:

lrineau:
@mglisse Is the current version of the code the one you agree with?

mglisse:
It is one I don't strongly object to.

GilesBathgate:
@mglisse If I knew of any other way to make it more palatable to you I would ;)

PR #5215, opened 2020-11-27
Fix dereference after null check in Nef_3/Vertex.h

GilesBathgate:
## Summary of Changes

In the following code `sncp_` is checked for a null pointer, implying that it can be null, however later on the code calls `sncp()` which dereferences `sncp_` without checking for null.

https://github.com/CGAL/cgal/blob/077937383555a61e5b23460fc21cf75f3f005018/Nef_3/include/CGAL/Nef_3/Vertex.h#L337-L346

Simply adding a check for `valid` to the if ensures no fault in this scenario.

## Release Management

* Affected package(s): Nef_3


lrineau (approved):

maxGimeno:

PR #5219, opened 2020-11-29
Double defined ID seg_coord_global

albert-github:
It is better to define the anchor in the detailed section of a method as when it is in the brief description it will be shown in the overview of the functions (top of the page) and in the detailed description.
This is due to the `REPEAT_BRIE=YES` setting.

See pages: Barycentric_coordinates_2/classCGAL_1_1Barycentric__coordinates_1_1Segment__coordinates__2.html and Barycentric_coordinates_2/index.html

### DIFF
--- a/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Segment_coordinates_2.h
+++ b/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Segment_coordinates_2.h
@@ -199,10 +199,10 @@ private:
 // Global functions
 
 /*!
-   \anchor seg_coord_global
  * \relates Segment_coordinates_2
  * This is a global function that takes both vertices of a segment and computes segment coordinates at a given query point with respect to these vertices.
 
+   \anchor seg_coord_global
 \tparam Traits must be a model of the concept `BarycentricTraits_2`.
 
 */


albert-github:
@danston What is the meaning of your comment?

lrineau:
> /build:v1

@albert-github We have a Github Actions bot, and that command triggers a build of the documentation.

danston:
I am just compiling the docs in order to see the difference between the standard docs and the docs with your fix.

albert-github:
@lrineau thanks
@danston Difference will be hard to spot, in the "old" version the jump from index.html will be to the brief description and in the "new" situation the jump will be to the detailed description.

danston:
@albert-github Ok, the new docs look fine. The fix does not break anything. It all looks good. However, I honestly do not see any reason for applying it since there is no detailed description in the funtion except for the `Traits` parameter. So, what does this fix give us? The only reason I can find for it is if in the future we extend the detailed description. But if we do, the whole thing can change completely so the fix will be useless as well.

albert-github:
On the outside the fix doesn't look spectacular but on the inside of the html file in the "old" situation there is an `id` (with value `seg_coord_global`) double defined i.e. in the brief section and in the detailed section (as in the detailed section the brief description is repeated). The double defining in one file is against the rules (consider it as defining one variable twice in one block of cpp code). Most HTML browsers have no problems with it as they use the first found `id` but as said it is not correct.

danston (approved):
Ok, that sounds reasonable. I approve of this fix. Thanks for the submission.

maxGimeno:

PR #5230, opened 2020-11-30
Add bounds checks in SNC_io_parser.h

GilesBathgate:
## Summary of Changes

CWE-20: Untrusted value as argument. 

In SNC_io_parser.h, the value for index is read from the file e.g `in >> index;`. These indexes are then used as `std::vector` index using the `[]` operator which doesn't perform any bounds checks and could lead to undefined behaviour. This PR simply adds bounds checks using `std::vector.at(...` which does perform bounds checks and throws an `out_of_range` exception. https://www.cplusplus.com/reference/vector/vector/at/

In addition `CGAL_assertion` checks are added to ensure the static analysis passes.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_io_parser.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_io_parser.h
@@ -1535,22 +1535,22 @@ read_vertex(Vertex_handle vh) {
   vh->sncp() = this->sncp();
 
   in >> index;
-  vh->svertices_begin() = (index >= 0 ? Edge_of[index] : this->svertices_end());
+  vh->svertices_begin() = (index >= 0 ? Edge_of.at(index) : this->svertices_end());
   in >> index;
-  vh->svertices_last()  = index >= 0 ? Edge_of[index] : this->svertices_end();
+  vh->svertices_last()  = index >= 0 ? Edge_of.at(index) : this->svertices_end();
   OK = OK && test_string(",");
   in >> index;
-  vh->shalfedges_begin() = index >= 0 ? SEdge_of[index] : this->shalfedges_end();
+  vh->shalfedges_begin() = index >= 0 ? SEdge_of.at(index) : this->shalfedges_end();
   in >> index;
-  vh->shalfedges_last()  = index >= 0 ? SEdge_of[index] : this->shalfedges_end();
+  vh->shalfedges_last()  = index >= 0 ? SEdge_of.at(index) : this->shalfedges_end();
   OK = OK && test_string(",");
   in >> index;
-  vh->sfaces_begin() = index >= 0 ? SFace_of[index] : this->sfaces_end();
+  vh->sfaces_begin() = index >= 0 ? SFace_of.at(index) : this->sfaces_end();
   in >> index;
-  vh->sfaces_last()  = index >= 0 ? SFace_of[index] : this->sfaces_end();
+  vh->sfaces_last()  = index >= 0 ? SFace_of.at(index) : this->sfaces_end();
   OK = OK && test_string(",");
   in >> index;
-  vh->shalfloop() = index >= 0 ? SLoop_of[index] : this->shalfloops_end();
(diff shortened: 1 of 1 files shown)


lrineau:
> An alternative approach could be to use `std::vector.at(...` which does perform bounds checks and throws an `out_of_range` exception. [cplusplus.com/reference/vector/vector/at](https://www.cplusplus.com/reference/vector/vector/at/) However this would then also require the index to also be declared as `size_t` since the bounds check doesn't check for negative values.

`vector.at(-1)` calls `vector.at(static_cast<size_type>(-1))`, and that is `vectot.at(SIZE_MAX)`. That is way above the size bound. :smile:

GilesBathgate:
@lrineau I think possibly not on all platforms. Here is the gnu libstdc++ implementation:

```c++
 /// Safety check used only from at().
      void
      _M_range_check(size_type __n) const
      {
	if (__n >= this->size())
	  __throw_out_of_range_fmt(__N("vector::_M_range_check: __n "
				       "(which is %zu) >= this->size() "
				       "(which is %zu)"),
				   __n, this->size());
      }

      reference
      at(size_type __n)
      {
	_M_range_check(__n);
	return (*this)[__n];
      }
```

GilesBathgate:
@lrineau Or are you saying the static cast is implicit?

lrineau:
> @lrineau Or are you saying the static cast is implicit?

Yes. If you pass an integer to `vector::at(size_type)`, then its value is converted to the argument type: `size_type`.

GilesBathgate:
@lrineau Do you have a preference for using `vector::at` then?

lrineau:
Yes, because the current patch cannot work: as soon as there is such an error, the code should exit from the function with an error. Otherwise the "read" function will create a data structure that is not correct, and that will probably segfault soon after the reading.

For example: at https://github.com/CGAL/cgal/pull/5230/files#diff-9136b254c9c709ee1cf35a2cf4ce8d09872ebb7f615a35dab8c7eae0a4b2232eR1607-R1608, if the index is wrong, then then `eh->twin()` will be left with whatever value there was (uninitialized, or zero-initialized?) and that cannot create a valid data structure.

GilesBathgate:
@lrineau Changed to use `vector::at`

maxGimeno:

lrineau:
Hi Giles, so far your contributions to CGAL were trivial enough that they did not require a contributor agreement. But this PR (and others that are not yet merged) is significant.

I do not know the details of the contributor agreements for the CGAL project. I will discuss that with my colleges.

GilesBathgate:
@lrineau I have to sign this and send it in the post? https://www.cgal.org/CLA/CLA-GF.html

lrineau:
> @lrineau I have to sign this and send it in the post? [cgal.org/CLA/CLA-GF.html](https://www.cgal.org/CLA/CLA-GF.html)

Yes, that is the document. But as your are employee of Blue Prism, there is something more to do. There is a CLA for legal entities.

GilesBathgate:
@lrineau If it makes any difference the work was done in personal time and is in no way associated with my employer. If you need me to sign a disclaimer with words to that effect I'd be happy to.

maxGimeno:
I think you won't have to bother with a disclaimer, I'm about ot replace this PR with one of mine, that doesn't fix the issue the same way, and also fixes similar issues in Nef_2 and Nef_S2.

maxGimeno:
Replaced by https://github.com/CGAL/cgal/pull/5371

GilesBathgate:
@maxGimeno Thanks, I'll try the patch against my project and see what the static analysis results are.

maxGimeno:
Thanks ! I hope I didn't miss something

GilesBathgate:
@maxGimeno After fixing the `returnl` issue in the patch, and replacing my patch with yours, I found that 0 more defects detected and 0 defects fixed meaning your patch has the same effect on fixing the static analysis defects, Thanks.

PR #5231, opened 2020-11-30
Rename 'D' to 'Dimension'

### LINKED ISSUE #5178: FuzzyQueryItem concept nested types
## Issue Details

The concept FuzzyQueryItem documents that a model must provide nested public types Point_d, FT and D. The first thing to notice is that Fuzzy_sphere does not provide D. Second, nothing uses any of those types in Kd_tree. It looks like they are present because the rest of the concept has a function that takes a Point_d and another one that takes a `Kd_tree_rectangle<FT,D>`, and because Kd_tree::search needs a way to say that the argument of contains() matches the type SearchTraits::Point_d.

Should there be a mention that those types are for "exposition only"? Or just remove them? Rename D to Dimension or vice versa?

## Environment

* CGAL version: 5.1

MaelRL:
## Summary of Changes

The type `D` was not a real requirement as none of the models provided by CGAL actually provided this typedef, but it's in the concept to explicit the dimension requirements between the KD-tree and the object.

Renamed to `Dimension` to match the actual code.

## Release Management

* Affected package(s): `Spatial_searching`
* Issue(s) solved (if any): fix #5178

### DIFF
--- a/Spatial_searching/doc/Spatial_searching/Concepts/FuzzyQueryItem.h
+++ b/Spatial_searching/doc/Spatial_searching/Concepts/FuzzyQueryItem.h
@@ -18,7 +18,7 @@ public:
 /*!
 Dimension Tag.
 */
-typedef unspecified_type D;
+typedef unspecified_type Dimension;
 
 /*!
 represents a `d`-dimensional point.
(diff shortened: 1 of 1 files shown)


maxGimeno:

PR #451, opened 2015-10-31
Introduces a CGAL_destructor_assertion macro.

GilesBathgate:
**The problem:**  Throwing an exception from a destructor can cause problems since the callee is unable to catch an exception raised while the stack is unwinding. The result is the application will terminate with a hard crash.

**Consequence:** Since the exception is uncatchable, the program will terminate, but also the real exception is hidden by the the destructor exception.

**Solution:** Its desirable to not throw exceptions from the destructor if and only if an existing exception has not been caught, and the stack is unwinding. For this purpose I propose  a new macro that can safely be called from a destructor, even when the stack is currently unwinding. The macro makes use of `std::uncaught_exception()` and results in a nop when this is the case.

Additional discussions about this issue and what problems it can cause for client applications (rapcad and openscad) is detailed here: https://github.com/openscad/openscad/issues/410, there was also a discussion about the issue here: http://cgal-discuss.949826.n4.nabble.com/Uncatchable-exception-converting-from-Nef-polyhedron-to-Polyhedron-3-td4658603.html

Another instance of the same problem present in the openscad project: https://github.com/openscad/openscad/issues/1455

PR #5233, opened 2020-11-30
Fix uncaught exceptions

GilesBathgate:
## Summary of Changes

In a previous issue ( #451 ) a `CGAL_destructor_assertion` macro was added to handle assertions made within destructors. When an exception is thrown from a destructor while the stack is being unwound, the existence of any uncaught exceptions cause the program to terminate immediately, without correct error handling. Sometimes it's safe to throw an exception even while std::uncaught_exception() == true. For example, if stack unwinding causes an object to be destructed, the destructor for that object could run code that throws an exception as long as the exception is caught by some catch block before escaping the destructor. That said coverity static analysis has found several places in the codebase where this is not the case, i.e the destructor calls some code which makes a `CGAL_assertion()` of some form or another, but doesn't catch the exception.

I propose a new macro:

```c++
#define CGAL_destructor_assertion_catch(CODE) try{ CODE } catch(...) { if(std::uncaught_exceptions() <= 0) throw; }
```

This preserves the assertion exception by rethrowing it if the stack is **not** being unwound and there are no uncaught exceptions. But in the case where we are recovering from an exception already, only that exception is preserved. A program which demonstrates this behaviour can be seen here: http://cpp.sh/6p7ka

This macro has then been applied to all places identified by static analysis as potentially throwing exceptions from CGAL_assertions. As before ( in #451 ) the `CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)` macro has to be applied, where needed.

An additional Macro `CGAL_destructor_warning` was also added, this replicates the behaviour of `CGAL_destructor_assertion`, except emits a `::CGAL::warning_fail`

## Release Management

* Affected package(s): several.
* Link to compiled documentation (obligatory for small feature). I will add documentation if in principle the idea is acceptable.

### DIFF
--- a/HalfedgeDS/include/CGAL/HalfedgeDS_list.h
+++ b/HalfedgeDS/include/CGAL/HalfedgeDS_list.h
@@ -395,7 +395,11 @@ public:
         // halfedges, and f faces. The reservation sizes are a hint for
         // optimizing storage allocation. They are not used here.
 
-    ~HalfedgeDS_list() { clear(); }
+    ~HalfedgeDS_list() noexcept {
+      try {
+        clear();
+      } catch (...) {}
+    }
 
     HalfedgeDS_list( const Self& hds)
     :  vertices( hds.vertices),
--- a/STL_Extension/include/CGAL/Handle_for.h
+++ b/STL_Extension/include/CGAL/Handle_for.h
@@ -149,12 +149,14 @@ public:
         return *this;
     }
 
-    ~Handle_for()
+    ~Handle_for() noexcept
     {
-      if (--(ptr_->count) == 0) {
-        Allocator_traits::destroy(allocator, ptr_);
-        allocator.deallocate( ptr_, 1);
-      }
+      try{
+        if (--(ptr_->count) == 0) {
+          Allocator_traits::destroy(allocator, ptr_);
+          allocator.deallocate( ptr_, 1);
+        }
+      } catch(...) {}
     }
 
     void
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -442,12 +442,15 @@ friend std::ostream& operator<<
 }
 
 
-~Node() {
+~Node() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+{
   CGAL_NEF_TRACEN("~Node: deleting node...");
-  if( !is_leaf()) {
-    delete left_node;
-    delete right_node;
-  }
+  CGAL_destructor_assertion_catch(
+    if( !is_leaf()) {
+      delete left_node;
+      delete right_node;
+    }
+  );
 }
 
 private:
@@ -1103,9 +1106,12 @@ bool update( Node* node,
   return (left_updated || right_updated);
 }
 
-~K3_tree() {
+~K3_tree() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+{
   CGAL_NEF_TRACEN("~K3_tree: deleting root...");
-  delete root;
+  CGAL_destructor_assertion_catch(
+    delete root;
+  );
 }
 
 private:
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -126,7 +126,8 @@ public:
 
   virtual void add_vertex(Vertex_handle) {}
 
-  virtual ~SNC_point_locator() {
+  virtual ~SNC_point_locator() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+  {
     CGAL_NEF_CLOG("");
     CGAL_NEF_CLOG("construction_time:  "<<ct_t.time());
     CGAL_NEF_CLOG("pointlocation_time: "<<pl_t.time());
@@ -422,8 +423,9 @@ public:
     return updated;
   }
 
-  virtual ~SNC_point_locator_by_spatial_subdivision() {
-    CGAL_warning(initialized ||
+  virtual ~SNC_point_locator_by_spatial_subdivision() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+  {
+    CGAL_destructor_warning(initialized ||
                  candidate_provider == 0); // required?
     if(initialized)
       delete candidate_provider;
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_map.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_map.h
@@ -229,7 +229,12 @@ public:
   Sphere_map(bool = false) : boundary_item_(boost::none),
     svertices_(), sedges_(), sfaces_(), shalfloop_() {}
 
-  ~Sphere_map() { clear(); }
+  ~Sphere_map() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+  {
+    CGAL_destructor_assertion_catch(
+      clear();
+    );
+  }
 
   Sphere_map(const Self& D) : boundary_item_(boost::none),
     svertices_(D.svertices_),
--- a/Nef_S2/include/CGAL/Nef_polyhedron_S2.h
+++ b/Nef_S2/include/CGAL/Nef_polyhedron_S2.h
@@ -73,7 +73,12 @@ class Nef_polyhedron_S2_rep {
 public:
   Nef_polyhedron_S2_rep() : sm_() {}
   Nef_polyhedron_S2_rep(const Self&) : sm_() {}
-  ~Nef_polyhedron_S2_rep() { sm_.clear(); }
+  ~Nef_polyhedron_S2_rep() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+  {
+    CGAL_destructor_assertion_catch(
+      sm_.clear();
+    );
+  }
 };
 
 /*{\Moptions print_title=yes }*/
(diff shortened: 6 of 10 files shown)


HalfedgeDS/include/CGAL/HalfedgeDS_list.h
```diff
@@ -385,7 +385,11 @@ class HalfedgeDS_list
         // halfedges, and f faces. The reservation sizes are a hint for
         // optimizing storage allocation. They are not used here.
 
-    ~HalfedgeDS_list() { clear(); }
+    ~HalfedgeDS_list() noexcept {
+      try {
+        clear();
+      } catch (...) {}
+    }
```
lrineau, line 392: How is it possible that a `clear()` can trigger? What does it do?!

lrineau: ... Oh... is it about assertions in the `clear()` functions, that can trigger?

GilesBathgate: precisely they are all from CGAL_assertion. I didn't think the overhead of this code in release builds was significant, to warrant using `CGAL_assertion_code`

maxGimeno:

lrineau:
This PR, and particularly 9f9e1e200d8b6e29e2e4277f591aafd9022601e4 is the one responsible for all the compilation errors about `boost::any` destructor, in the [last test results](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-90.shtml):
```
/usr/include/boost/any.hpp: In instantiation of 'class boost::any::holder<CGAL::Segment_2<CGAL::Cartesian<double> > >':
/usr/include/boost/any.hpp:253:17:   required from 'ValueType* boost::any_cast(boost::any*) [with ValueType = CGAL::Segment_2<CGAL::Cartesian<double> >]'
/mnt/testsuite/include/CGAL/Object.h:87:43:   required from 'bool CGAL::Object::assign(T&) const [with T = CGAL::Segment_2<CGAL::Cartesian<double> >]'
/mnt/testsuite/include/CGAL/Object.h:157:20:   required from 'bool CGAL::assign(T&, const CGAL::Object&) [with T = CGAL::Segment_2<CGAL::Cartesian<double> >]'
.../test/Triangulation_2/include/CGAL/_test_cls_delaunay_triangulation_2.h:379:24:   required from 'void _test_delaunay_duality(const Del&) [with Del = CGAL::Delaunay_triangulation_2<CGAL::Cartesian<double> >]'
.../test/Triangulation_2/include/CGAL/_test_cls_delaunay_triangulation_2.h:201:26:   required from 'void _test_cls_delaunay_triangulation_2(const Del&) [with Del = CGAL::Delaunay_triangulation_2<CGAL::Cartesian<double> >]'
.../test/Triangulation_2/test_delaunay_triangulation_2.cpp:70:46:   required from here
/usr/include/boost/any.hpp:169:15: error: looser exception specification on overriding virtual function 'virtual boost::any::holder<CGAL::Segment_2<CGAL::Cartesian<double> > >::~holder() noexcept (false)'
  169 |         class holder
      |               ^~~~~~
/usr/include/boost/any.hpp:156:21: note: overridden function is 'virtual boost::any::placeholder::~placeholder() noexcept'
  156 |             virtual ~placeholder()
      |                     ^
```

GilesBathgate:
> This PR, and particularly [9f9e1e2](https://github.com/CGAL/cgal/commit/9f9e1e200d8b6e29e2e4277f591aafd9022601e4) is the one responsible for all the compilation errors about `boost::any` destructor

Ah, yes. When the class inherits from another that has a noexcept destructor, it cannot use `CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)` in fact, the constructor must be noexecpt too, which means consequently that the destructor can never throw. Therefore the only sensible pattern that applies is:

```c++
~Handle_for() noexcept
    {
      try {
        if (--(ptr_->count) == 0) {
          Allocator_traits::destroy(allocator, ptr_);
          allocator.deallocate( ptr_, 1);
        }
      } catch(...) {}
    }
 ```
 
 This means you loose the exception in the destructor entirely, but its better than program termination which is what happens when you throw from a noexcept method.

**Comment by GilesBathgate** [comment]
@lrineau Fixed as I proposed above.

**Comment by maxGimeno** [comment]
@GilesBathgate Where can we see your fix ?

**Comment by GilesBathgate** [comment]
> @GilesBathgate Where can we see your fix ? The force push kind of hides the modifications
> EDIT: Nevermind, I found it.

https://github.com/CGAL/cgal/pull/5233/commits/48831abecee904e24303485b1b37653a6b4c98a7

**Comment by maxGimeno** [comment]
Thanks !

**Comment by maxGimeno** [comment]
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-98.shtml

File: STL_Extension/include/CGAL/Handle_for.h [hunk 2]
```diff
@@ -149,12 +149,14 @@ class Handle_for
         return *this;
     }
 
-    ~Handle_for()
+    ~Handle_for() noexcept
```
mglisse, line 152: That's useless, it is noexcept by default.

GilesBathgate: It's probably a hang-over from when instead I'd had `CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)` 
I've since found some more uncaught exception issues with the static analysis on the latest runs (where I upgraded from 5.0.3 to master) I will clean this up along with the others and put them in a new PR.

STL_Extension/include/CGAL/Handle_for.h
```diff
@@ -149,12 +149,14 @@ class Handle_for
...
-      if (--(ptr_->count) == 0) {
-        Allocator_traits::destroy(allocator, ptr_);
-        allocator.deallocate( ptr_, 1);
-      }
+      try{
+        if (--(ptr_->count) == 0) {
+          Allocator_traits::destroy(allocator, ptr_);
+          allocator.deallocate( ptr_, 1);
+        }
+      } catch(...) {}
```
mglisse, line 159: I am not going to fight it, but I want to mention that I am not fond of this new code
- it is ugly
- it goes against the default specified by the language

GilesBathgate: Many of the destructors in the CGAL codebase can throw because of a CGAL_assertion, so when it calls deallocate here the program will terminate. (presumably, that's not desirable, that's what the change intends to prevent)

 In most places, this is handled by a `CGAL_destructor_assertion_catch` which preserves the exception when possible. But here because the noexcept is enforced by a base class we can only swallow the exception.

However you might agree that the correct way to solve all these problems, is never to throw from a destructor in the first place. Then there wouldn't have to be workarounds for the default behaviour specified by the language.

mglisse: > Many of the destructors in the CGAL codebase can throw because of a CGAL_assertion, so when it calls destroy/deallocate here the program will terminate. (presumably, that's not desirable, that's what the change intends to prevent)

It isn't obvious that it isn't desirable. The standard didn't gratuitously introduce `std::terminate`, it was better in those circumstances than data corruption or security holes.

> In most places, this is handled by a `CGAL_destructor_assertion_catch` which preserves the exception when possible. But here because the noexcept is enforced by a base class we can only swallow the exception.
> 
> However you might agree that the correct way to solve all these problems, is never to throw from a destructor in the first place. Then there wouldn't have to be workarounds for the default behaviour specified by the language.

And assertions are not supposed to trigger, so that's good, no need to introduce workarounds for the case where they trigger :wink:
Presumably, we aren't checking preconditions on user input in those destructors. If we assert, it probably means a bug in CGAL, or a side effect of some memory corruption, anyway something bad. We can disable those checks, as your patch does, that makes your checker happy and possibly fails to detect an issue. Or we can have checks that abort if they cannot throw exceptions, which is what the language specifies.

I wonder if separating more clearly debug assertions from checks that can make sense for users would reduce the number of destructors affected.

Some assertions are meant to protect from users who don't know what they are doing, say a user constructing Interval_nt(3,2). CGAL never constructs an interval with a>b, but since CGAL and the user's code call the same constructor, they go through the same assertion, and it looks like the CGAL could throw although it won't. If this happens in a noexcept function (like a destructor), the compiler has to generate instructions on what to do with that non-existing exception, for instance ignore it or abort. The amount of code is essentially the same, there shouldn't be much impact on performance, and since the exception never happens, the behavior is also the same. One code is simpler (no need for try-catch), and on the off chance some memory corruption causes the assertion to trigger, it will safely abort instead of executing random instructions.

Anyway, I said I won't fight...

GilesBathgate: I appreciate your input @mglisse . It's always worthwhile having a discussion even if there are differences in opinion.

GilesBathgate: > Presumably, we aren't checking preconditions on user input in those destructors. 

Not a destructor, but here is a noexcept method with a `CGAL_precondition`

https://github.com/CGAL/cgal/blob/master/STL_Extension/include/CGAL/Handle.h#L70-L79

mglisse: > Not a destructor, but here is a noexcept method with a `CGAL_precondition`

Yes, I think I remember adding this noexcept and wondering about the precondition. In the end, I was ok with having the program terminate if a null pointer was found in debug mode. The precondition isn't there in release mode. Not testing in debug mode would be a regression. Making the function conditionally noexcept would have been an option, but it can change the path taken by the code, and it felt more useful to lie a bit and have debug mode test something closer to release mode, than to insist on propagating the error as an exception. Of course different opinions are always possible.

GilesBathgate: >Making the function conditionally noexcept 

In our usage of the CGAL library (RapCAD and OpenSCAD) we enable CGAL_precondition in release mode, so this would be more suitable for us I think.

sloriot: Can't you check the prerequisites in inputs before calling the functions? Or is it too expensive compared to do a try-catch?

GilesBathgate: > Can't you check the prerequisites in inputs before calling the functions?
Not sure I follow what you mean @sloriot

PR #5234, opened 2020-12-01
[Shape Detection] Fix/optimization + validity test

sgiraudot:
## Summary of Changes

- addition of a validity test that checks that Region Growing returns exactly the right number of planes and that RANSAC returns at least 75% of them
- fix the computation of the RANSAC stop probability which was wrong if something else than planes was detected
- add a small optimization that makes the RANSAC algorithm pick several shapes (up to 200) at each loop instead of 1 (reduces computation times by 80% in some rare cases, otherwise has no visible effect)
- fix the plugin which was broken (angle was mistakenly taken as a cosine)

## Release Management


Shape_detection/test/Shape_detection/CMakeLists.txt
```diff
@@ -42,6 +42,21 @@ if(EIGEN3_FOUND)
...
+    file(GLOB proto_src "${RANSAC_PROTO_DIR}/*.cpp")
+    file(GLOB proto_misc_src "${RANSAC_PROTO_DIR}/MiscLib/*.cpp")
+    add_library(libproto STATIC  ${proto_src} ${proto_misc_src})
+    add_executable(test_validity_sampled_data "test_validity_sampled_data.cpp")
+    target_link_libraries(test_validity_sampled_data libproto CGAL::CGAL CGAL::Eigen_support)
+  else()
+    add_executable(test_validity_sampled_data "test_validity_sampled_data.cpp")
+    target_link_libraries(test_validity_sampled_data CGAL::CGAL CGAL::Eigen_support)
+  endif()
+
```
lrineau, line 59: I think you need to add:
```suggestion
  endif()
  cgal_add_test(test_validity_sampled_data)
```
so that the CTest test is added.

maxGimeno:
There will be trouble with CGAL_IO because of the calls to `read_XXX_points()` in your tests.

maxGimeno:
errors: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-9/Shape_detection/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz

sgiraudot:
Fixed.

maxGimeno:
Now the test times out : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-10/Shape_detection/TestReport_lrineau_Debian-Stable.gz

Shape_detection/test/Shape_detection/test_validity_sampled_data.cpp
```diff
@@ -0,0 +1,244 @@
...
+  CGAL::cpp98::random_shuffle (points.begin(), points.end());
+
+  test_copied_point_cloud (points, 1);
+  test_copied_point_cloud (points, 2);
+  test_copied_point_cloud (points, 5);
+  test_copied_point_cloud (points, 10);
+#ifndef CGAL_TEST_SUITE // Disable tests too large for testsuite
+  test_copied_point_cloud (points, 20);
+  test_copied_point_cloud (points, 50);
+#endif
```
lrineau, line 70: On my machine, I had to kill the program `./test_validity_sampled_data` because it was consuming 24.5GB of memory. Are you sure that is expected? Maybe there is also a memory leak.

sgiraudot: Oh, that's weird, I'll look it up.

(Anyway, I think I still need to restrain the test on the testsuite because the tests can be long, even without considering a memory problem.)

maxGimeno:

lrineau:
I will not merge this PR because of https://github.com/CGAL/cgal/pull/5234#discussion_r535207671, waiting for @sgiraudot's answer.

sgiraudot:
Ok, so I made some additional tests:
- there does not seem to be any memory leak (confirmed using sanitizer)
- I'm mainly suspecting an infinite loop
- I think it's just that the algorithm takes a while and use a lot of memory with the high values that I deactivated for the testsuite (more than 1 hour on my computer and it did reach more than 15GB of memory)

So I think the PR can be integrated if tests are ok.

maxGimeno:
The tests that you left in the testsuite are fine.

PR #5237, opened 2020-12-02
Referencing non existing file

albert-github:
For a better quality also the close image was converted to an svg image, so the png does not exists anymore. This has to be reflected in this file as well.


lrineau:
The modification of this PR must be replicated in the four versions of the `resources/` directory.

albert-github:
@lrineau I don't think this is the case the close.svg only is a problem for the newer version i.e as of 1.8.19.

albert-github:
@MaelRL This is not an enhancement but an error (bug)

maxGimeno:

PR #5242, opened 2020-12-04
Kernel: Collinear_3: Static Analysis

### LINKED ISSUE #5207: Logically dead code in Collinear_3.h - static analysis
## Issue Details

In the following code the value of of `int_tmp_result` is either set to `1`, or `-1` or the function is exited early via the returns. Then `sign_of_determinant_return_value` (which is assigned from int_tmp_result) is checked to see if it is not equal to 0, which by the above code will always be true, so the function exits with `return false;`

https://github.com/CGAL/cgal/blob/16fc8d1fe27ff2ea79340d5c7a892083acbb24fe/Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h#L74-L104

The remaining code in the function is therefore never executed. Either this is an indication of a different bug, or the dead code should be removed.

## Environment

* Compiler: gcc
* Specific flags used (if any): None
* CGAL version: 5.0.2
* Boost version: 1.71.0

afabri:
## Summary of Changes

Make the changes as suggested by @mglisse to fix the issue.  I didn't address the further suggested improvements.

## Release Management

* Issue(s) solved (if any): fix #5207

### DIFF
--- a/Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h
+++ b/Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h
@@ -76,16 +76,11 @@ public:
       else if (max1 > upper_bound_1)
         upper_bound_1 = max1;
 
-      int int_tmp_result;
-
-      if ( (lower_bound_1 < 5.00368081960964635413e-147) || (upper_bound_1 > 1.67597599124282407923e+153) )
-        return Base::operator()(p, q, r);
-
-      double eps = (8.88720573725927976811e-16 * (max1 * max2));
-      if ( (double_tmp_result > eps) || (double_tmp_result < -eps) )
-        return false;
-
-
+      if ((lower_bound_1 >= 5.00368081960964635413e-147) && (upper_bound_1 <= 1.67597599124282407923e+153)) {
+        double eps = (8.88720573725927976811e-16 * (max1 * max2));
+        if ((double_tmp_result > eps) || (double_tmp_result < -eps))
+          return false;
+      }
       double dpz = (pz - rz);
       double dqz = (qz - rz);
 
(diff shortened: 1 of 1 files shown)


Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h
```diff
@@ -77,32 +77,18 @@ class Collinear_3
         upper_bound_1 = max1;
 
       int int_tmp_result;
-      if (lower_bound_1 < 5.00368081960964635413e-147)
+
+      if ( (lower_bound_1 < 5.00368081960964635413e-147) || (upper_bound_1 > 1.67597599124282407923e+153) )
         return Base::operator()(p, q, r);
```
mglisse, line 82: IIUC, if all the points have the same x coordinate, we end up here, it seems a bit too early to go to the exact computation, we could check the other projections. Maybe that falls under "further suggested improvements".

lrineau: I agree with Marc: the static filter should only end up on the exact computation if all the possible filtering have failed. The *raison d'être* of those static filters is to quickly find a reason to tell that the points are not collinear.

Points `p`, `q` and `r` and collinear if and only if the two vector `pq` and `pr` are collinear, that is if and only if their cross-product is null. The cross-product has three coordinates that are each 2×2 determinants. That gives exactly three opportunities for the static filters to prove that the three points are not collinear. The `return Base::operator()(p, q, r);` should be only at the end, when all three opportunities have failed.

afabri: > Maybe that falls under "further suggested improvements".

I meant what you suggest for `collinearC3()`.   Also I am wondering if  we could not profit from a function `is_determinant_zero()`  which would save us a subtraction.

mglisse: >  Also I am wondering if we could not profit from a function `is_determinant_zero()` which would save us a subtraction.

We have sign_of_determinant, although compilers may not be perfect at simplifying sign(double)==0.

afabri: In fact I was wondering about that for rational types.

Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h
```diff
@@ -76,33 +76,14 @@ class Collinear_3
...
-        if (double_tmp_result > eps)
-          int_tmp_result = 1;
-        else
-        {
-          if (double_tmp_result < -eps)
-            int_tmp_result = -1;
-          else
-            return Base::operator()(p, q, r);
-        }
+        if ((double_tmp_result > eps) || (double_tmp_result < -eps))
```
mglisse, line 81: you changed the next one to use abs()>eps but not this one? (not CGAL::abs so it would use sse2fabs on windows?)
we could move the computation of double_tmp_result just before this line, though it probably doesn't change anything, so not necessary.

Filtered_kernel/include/CGAL/internal/Static_filters/Collinear_3.h
mglisse: Indeed, and for rationals, cmp can be quite a bit more expensive than ==.

maxGimeno:

PR #5245, opened 2020-12-04
Allow -DNDEBUG and CGAL_nnn_assertions at the same time

### LINKED ISSUE #5240: Please allow -DNDEBUG and CGAL_nnn_assertions at the same time
## Issue Details

There seems to be some very confusing and inconsistent usage of the `NDEBUG` define within the CGAL library.

## Source Code

It seems the correct way to disable `CGAL_NEF_TRACE.*` is to define `NDEBUG` :
https://github.com/CGAL/cgal/blob/869833136d6e2b0487e8d083f55fae7afd151258/Nef_2/include/CGAL/Nef_2/debug.h#L44-L50
but then all the assertions generated from script cgal_create_assertions.sh, we need to not define `NDEBUG` or the assertions will be disabled:
https://github.com/CGAL/cgal/blob/869833136d6e2b0487e8d083f55fae7afd151258/STL_Extension/include/CGAL/triangulation_assertions.h#L30-L31

Furthermore in: 
https://github.com/CGAL/cgal/blob/869833136d6e2b0487e8d083f55fae7afd151258/Nef_3/include/CGAL/Nef_3/K3_tree.h#L527-L539
Defining `NDEBUG` would exclude some assertions that we potentially do want, meanwhile in: 
https://github.com/CGAL/cgal/blob/869833136d6e2b0487e8d083f55fae7afd151258/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h#L663-L688
Trace code is yet again included that isn't desired, in release builds. 

This leaves me wondering if I am not supposed to include `CGAL_assertions` in release builds at all, or is the developer expected to use `#include` hacks like this: 

https://github.com/openscad/openscad/commit/bb0ec94290733835df0716531c30460fc5b5210a

Or this:

https://github.com/GilesBathgate/RapCAD/commit/bedd95e3f12488f503c715e81c3a9d32fb1a3cf0

Would it not be better to have something like `CGAL_USE_TRACE` that can be defined if the trace code is desired?

Finally `CGAL_DEBUG` supposedly allows forcing CGAL assertions, even if `NDEBUG` but this rule doesn't seem to follow in the assertions generated by the script cgal_create_assertions.sh

## Environment

* Compiler: gcc
* Specific flags used (if any): NDEBUG
* CGAL version: 5.0.2
* Boost version: 1.71.0

GilesBathgate:
## Summary of Changes

Remove NDEBUG from various nnn_assertions.h files as it is already visible through `CGAL_NO_ASSERTIONS`, `CGAL_NO_PRECONDITIONS`, `CGAL_NO_POSTCONDITIONS`, `CGAL_NO_WARNINGS`

Include missing header files for files which reference the above defines.

## Release Management

* Affected package(s): Arrangement_on_surface_2, Convex_hull_2,Kernel_23,Partition_2,Point_set_processing_3,Poisson_surface_reconstruction_3,Polygon,Polytope_distance_d, STL_Extension, Scripts, SearchStructures, Straight_skeleton_2, Stream_lines_2

* Issue(s) solved (if any): fix #5240 
* Feature/Small Feature (if any): small feature

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
@@ -37,6 +37,8 @@
 
 #include <CGAL/Curved_kernel_via_analysis_2/Sweep_curves_adapter_2.h>
 
+#include <CGAL/kernel_assertions.h>
+
 namespace CGAL {
 
 namespace internal {
--- a/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Curved_kernel_via_analysis_2/Arc_2.h
@@ -1818,8 +1818,7 @@ protected:
      */
     void _check_arc_interior() const {
 
-#if !(defined(CGAL_KERNEL_NO_PRECONDITIONS) || defined(CGAL_NO_PRECONDITIONS) \
-        || defined(NDEBUG))
+#if !(defined(CGAL_KERNEL_NO_PRECONDITIONS) || defined(CGAL_NO_PRECONDITIONS))
 
         if(is_vertical()) {
             Coordinate_1 x0 = _minpoint().x();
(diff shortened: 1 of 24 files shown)


lrineau (approved):

maxGimeno:

PR #5263, opened 2020-12-10
Fix multiples calls to degenerate_edges() with an invalid input range

### LINKED ISSUE #5043: Infinite loop in remove_degenerate_edges
## Issue Details

Function `remove_degenerate_edges` in package `Polygon_mesh_processing` can get stuck in an infinite loop on some meshes. See attached zip file for an example mesh in OFF format along with a piece of C++ code getting stuck on this particular mesh.

The infinite loop in question is somewhere in the definition of function [`does_satisfy_link_condition`](https://doc.cgal.org/latest/BGL/group__PkgBGLEulerOperations.html#ga8af4fd79178c66fef7091614c0bc4a01). However what actually triggers it is that this function is called on an edge which doesn't exist anymore.

The responsible line can be found at [repair_degeneracies.h:1044](https://github.com/CGAL/cgal/blob/afc0abb82f341e5dadaae545dc3ac51d30a8f5dc/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h#L1044). Edge `e` comes from set `degenerate_edges_to_remove` which is itself built from `edge_range` using function `degenerate_edges` at the beginning of the while loop.

What happens is that when edge `e` is collapsed, it is removed from the mesh and from `degenerate_edges_to_remove`, but is still present in `edge_range`. Therefore at the next iteration of the while loop, edge `e` will again get included in `degenerate_edges_to_remove` although it doesn't exist in the mesh anymore. Function `does_satisfy_link_condition` will then get called on this non-existent edge, and will get stuck in a loop.

One easy fix, but which only works for `Surface_mesh` objects, is to add a condition in `is_degenerate_edge` in file [shape_predicates.h](https://github.com/CGAL/cgal/blob/afc0abb82f341e5dadaae545dc3ac51d30a8f5dc/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/shape_predicates.h#L70) to check if the edge actually belongs to the mesh : 

```cpp
template <typename PolygonMesh, typename NamedParameters>
bool is_degenerate_edge(typename boost::graph_traits<PolygonMesh>::edge_descriptor e,
                        const PolygonMesh& pm,
                        const NamedParameters& np)
{
  using parameters::get_parameter;
  using parameters::choose_parameter;

  typedef typename GetVertexPointMap<PolygonMesh, NamedParameters>::const_type VertexPointMap;
  VertexPointMap vpmap = choose_parameter(get_parameter(np, internal_np::vertex_point),
                                          get_const_property_map(vertex_point, pm));

  typedef typename GetGeomTraits<PolygonMesh, NamedParameters>::type Traits;
  Traits traits = choose_parameter<Traits>(get_parameter(np, internal_np::geom_traits));
  if (pm.is_removed(e)) return false; // NEW LINE HERE
  return traits.equal_3_object()(get(vpmap, source(e, pm)), get(vpmap, target(e, pm)));
}
```

But to my knowledge there is no general way to check if an edge descriptor is valid for a model of `HalfedgeGraph`, so it will only compiles with `Surface_mesh` types (but you can try with the OFF file in the attached zip, this single line will avoid the infinite loop).

We could simply copy `edge_range` into some container at the beginning of the `remove_degenerate_edges` function and update this container every time an edge is collapsed. Since I wasn't sure of the best way to fix this problem I opened an issue rather than doing a PR.

## Source Code

See test.cpp in attached zip file, and try it on the OFF file.

## Environment

* Operating system: Linux 5.8.10 x86_64
* Compiler: gcc 10.2.0
* Release or debug mode: both
* CGAL version: master/5.1
* Boost version: 1.72.0

## Example

[test.zip](https://github.com/CGAL/cgal/files/5308084/test.zip)

MaelRL:
## Summary of Changes

`remove_degenerate_edges()` iteratively calls `degenerate_edges(edge_range)`, but degenerate edges within `edge_range` will (should) disappear, resulting in a call over an invalid range.

## Release Management

* Issue(s) solved (if any): fix #5043 

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -1057,8 +1057,8 @@ bool remove_degenerate_edges(const EdgeRange& edge_range,
       if(CGAL::Euler::does_satisfy_link_condition(e, tmesh))
       {
         const halfedge_descriptor h = halfedge(e, tmesh);
-        degenerate_edges_to_remove.erase(it);
         local_edge_range.erase(*it);
+        degenerate_edges_to_remove.erase(it);
 
         // remove edges that could also be set for removal
         if(face(h, tmesh) != GT::null_face())
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -755,6 +755,7 @@ template <typename TriangleMesh, typename EdgeSet, typename FaceSet>
 typename boost::graph_traits<TriangleMesh>::vertex_descriptor
 remove_a_border_edge(typename boost::graph_traits<TriangleMesh>::edge_descriptor ed,
                      TriangleMesh& tm,
+                     EdgeSet& input_range,
                      EdgeSet& edge_set,
                      FaceSet& face_set)
 {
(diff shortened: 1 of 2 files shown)


maxGimeno:
runtime errors on windows : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-18/Polygon_mesh_processing/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz

maxGimeno:

PR #5266, opened 2020-12-11
PMP isotropic_remeshing() - improve edge flip step

### LINKED ISSUE #4144: Isotropic remeshing creates self intersections

janetournois:
## Summary of Changes

This PR improves the edge flipping step of `PMP::isotropic_remeshing()` by adding another criterion to trigger a flip : if one of the incident faces is badly shaped, the flip is tried, independently from the valence deviation improvement.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #4144 

this PR replaces #4192


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -238,6 +238,9 @@ self_intersections_impl(const FaceRange& face_range,
   // This loop is very cheap, so there is hardly anything to gain from parallelizing it
   for(face_descriptor f : face_range)
   {
+    if (f == boost::graph_traits<TriangleMesh>::null_face())
```
sloriot, line 241: how can this be possible? are you calling the function from a range of halfedge, some of them being on the border?

janetournois: this commit was part of the PR #4192
I am pretty sure I had met this case in the test case I was running at that time...
I cost of that check is low and prevents for misusing the function, so I think we can keep it

sloriot: @MaelRL what do you think? I'm not in favor of adding it and would prefer fixing the calling code.

MaelRL: I also would prefer if this function assumed a clean range.

lrineau: On the other hand, being robust to users' errors is also a nice property, no?

sloriot: Well, I usually agree but here we are talking about a NULL element in an input range.

lrineau: So the actual bug is in the calling code, from `demo/Polyhedron/`, right?

janetournois:
ok let's assume the user is not giving the null face in the input range

Polygon_mesh_processing/test/Polygon_mesh_processing/remeshing_test.cpp
```diff
@@ -171,74 +173,130 @@ Main(int argc, char* argv[])
...
+    t.stop();
+    std::cout << "Remeshing all took " << t.time() << std::endl;
+
+  }
+  else if (sharp_angle > 0)
+  {
+    typedef boost::property_map<Mesh, CGAL::edge_is_feature_t>::type EIFMap;
+    typedef boost::property_map<Mesh, CGAL::face_patch_id_t<int> >::type PIMap;
+    typedef boost::property_map<Mesh, CGAL::vertex_incident_patches_t<int> >::type VIMap;
+
```
sloriot, line 254:
```suggestion
    typedef typename  boost::property_map<Mesh, CGAL::edge_is_feature_t>::type EIFMap;
    typedef typename boost::property_map<Mesh, CGAL::face_patch_id_t<int> >::type PIMap;
    typedef typename boost::property_map<Mesh, CGAL::vertex_incident_patches_t<int> >::type VIMap;

```

maxGimeno:
Runtime errors : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-76/Surface_mesh_approximation/TestReport_lrineau_Debian-stable-Release.gz

sloriot:
@janetournois are you still targeting 5.3?

janetournois:
Yes. I am working on it

sloriot:
Should we disable that particular test while we debug what is wrong in the reported issue? It has nothing to do with the current patch of Jane.

maxGimeno:

maxGimeno:

PR #5268, opened 2020-12-12
Fix structurally dead code in SNC_simplify.h

GilesBathgate:
## Summary of Changes

In SNC_simplify.h the function  simplify_redundant_box_vertex unconditionally returns false.

https://github.com/CGAL/cgal/blob/a848e52552a9205124b7ae13c7bcd2b860eb4530/Nef_3/include/CGAL/Nef_3/SNC_simplify.h#L194-L202

This function should probably be removed as it is effectively not in use.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
@@ -191,16 +191,6 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
     return (sp1 == sp2.antipode());
   }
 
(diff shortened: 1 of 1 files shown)


lrineau (approved):
The commit that modified `simplify_redundant_box_vertex` to return `false` unconditionally is e722f28b5996215cbcaf26356dd6a15593b00ef9. I agree that code can be removed.

GilesBathgate:
> The commit that modified `simplify_redundant_box_vertex` to return `false` unconditionally is e722f28

@lrineau Yes, although I do wonder what that commit was ever intended to fix?!

maxGimeno:

PR #5269, opened 2020-12-12
Move assignment operators

GilesBathgate:
## Summary of Changes

The static analysis recommends, as a low priority, adding move assignment operators for various classes. I have implemented them using the swap idiom which is exception-safe. 

For Gmpq_type an assignment operator isn't needed because the class has no members, adding a defaulted move assignment operator suppresses the analysis warnings. However adding a move assignment operator when a copy assignment operator isn't explicitly declared deletes the copy assignment operator, which then causes compile errors. Likewise for the copy constructor, these therefore also have to be declared defaulted. I don't think declaring these defaulted member functions introduces any extra code generation.

## Release Management

* Affected package(s): Nef_3, STL_extension, Number_types

### DIFF
--- a/STL_Extension/include/CGAL/Handle.h
+++ b/STL_Extension/include/CGAL/Handle.h
@@ -68,6 +68,13 @@ class Handle
       return *this;
     }
 
+    Handle&
+    operator=(Handle&& x) noexcept
+    {
+      swap(*this,x);
+      return *this;
+    }
+
     friend void swap(Handle& a, Handle& b) noexcept { std::swap(a.PTR, b.PTR); }
 
     void reset()
--- a/STL_Extension/include/CGAL/Handle.h
+++ b/STL_Extension/include/CGAL/Handle.h
@@ -51,6 +51,12 @@ class Handle
       PTR->count++;
     }
 
+    Handle(Handle&& x) noexcept
+      : PTR(x.PTR)
+    {
+      x.PTR = static_cast<Rep*>(0);
+    }
+
     ~Handle()
     {
         if ( PTR && (--PTR->count == 0))
--- a/Nef_3/include/CGAL/Nef_3/Halfedge.h
+++ b/Nef_3/include/CGAL/Nef_3/Halfedge.h
@@ -100,6 +100,17 @@ class Halfedge_base
           return *this;
         }
 
+      Halfedge_base<Refs>& operator=(Halfedge_base<Refs>&& e) noexcept
+        { center_vertex_ = std::move(e.center_vertex_);
+          point_ = std::move(e.point_);
+          mark_ = std::move(e.mark_);
+          twin_ = std::move(e.twin_);
+          out_sedge_ = std::move(e.out_sedge_);
+          incident_sface_ = std::move(e.incident_sface_);
+          info_ = 0;
+          return *this;
+        }
+
       Vertex_handle& center_vertex() { return center_vertex_; }
       Vertex_const_handle center_vertex() const { return center_vertex_; }
 
--- a/Nef_3/include/CGAL/Nef_3/SHalfedge.h
+++ b/Nef_3/include/CGAL/Nef_3/SHalfedge.h
@@ -114,6 +114,22 @@ class SHalfedge_base  {
         return *this;
       }
 
+    SHalfedge_base<Refs>& operator=(SHalfedge_base<Refs>&& e) noexcept
+      {
+        source_ = std::move(e.source_);
+        sprev_ = std::move(e.sprev_);
+        snext_ = std::move(e.snext_);
+        incident_sface_ = std::move(e.incident_sface_);
+        twin_ = std::move(e.twin_);
+        prev_ = std::move(e.prev_);
+        next_ = std::move(e.next_);
+        facet_ = std::move(e.facet_);
+        info_ = 0;
+        mark_ = std::move(e.mark_);
+        circle_ = std::move(e.circle_);
+        return *this;
+      }
+
     Mark& mark() { return mark_; }
     const Mark& mark() const { return mark_; }
 
--- a/Nef_3/include/CGAL/Nef_3/SHalfloop.h
+++ b/Nef_3/include/CGAL/Nef_3/SHalfloop.h
@@ -75,6 +75,15 @@ class SHalfloop_base {
         return *this;
       }
 
+    SHalfloop_base<Refs>& operator=(SHalfloop_base<Refs>&& l) noexcept
+      { twin_ = std::move(l.twin_);
+        incident_sface_ = std::move(l.incident_sface_);
+        facet_ = std::move(l.facet_);
+        mark_ = std::move(l.mark_);
+        circle_ = std::move(l.circle_);
+        return *this;
+      }
+
     Mark& mark() { return mark_;}
     const Mark& mark() const { return mark_; }
 
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -369,8 +369,15 @@ protected:
   }
 
   Nef_polyhedron_3& operator=(const Nef_polyhedron_3<Kernel,Items, Mark>& N1) {
-    Base::operator=(N1);
+    Base::operator=(N1); // copy the handle
     set_snc(snc());
+    return (*this);
+  }
+
+  Nef_polyhedron_3& operator=(Nef_polyhedron_3<Kernel,Items, Mark>&& N1) noexcept {
+    N1.set_snc(snc()); // N1.set_snc sets N1.sncp_ not N1.snc_
+    Base::operator=(std::move(N1)); // swap the handles
+    set_snc(snc()); // snc() will return N1.snc_
     return (*this);
   }
 


Number_types/include/CGAL/GMP/Gmpq_type.h
```diff
@@ -240,6 +240,10 @@ class Gmpq
   }
 #endif
 
+  inline Gmpq(const Gmpq&) = default;
+  inline Gmpq& operator=(const Gmpq&) = default;
+  inline Gmpq& operator=(Gmpq&&) = default;
```
mglisse, line 245: I fail to see how that can be considered an improvement.

mglisse (commented):
(not familiar with Nef_3, so I won't comment much on that, my inclination would be to remove the assignment operator if the default works, but I don't know if that's the case)

STL_Extension/include/CGAL/Handle.h
```diff
@@ -68,6 +68,13 @@ class Handle
       return *this;
     }
 
+    Handle&
+    operator=(Handle&& x) noexcept
```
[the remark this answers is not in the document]
GilesBathgate: Added a move constructor.

GilesBathgate:
> (not familiar with Nef_3, so I won't comment much on that, my inclination would be to remove the assignment operator if the default works, but I don't know if that's the case)

The implementation of the copy assignment operator for Nef_3 does `set_snc(snc())` I don't know how an explicitly defaulted copy assignment operator could provide that same functionality, so I believe it cannot be removed.

mglisse:
> The implementation of the copy assignment operator for Nef_3 does `set_snc(snc())` I don't know how an explicitly defaulted copy assignment operator could provide that same functionality

`set_snc(snc())` looks like it copies a pointer from one base class to the other. If the object we are copying from has already done this operation, it might not matter whether we get the pointer from a copy of the first base, or from the original second base. But as I said, I am not familiar with Nef_3, so my comment may be completely off.

GilesBathgate:
@mglisse I'm inclined to leave it in place for this PR at least. If it would be better to remove it that can be addressed separately.

Number_types/include/CGAL/GMP/Gmpq_type.h
GilesBathgate: Well for reference I am basing my understanding on the top answer of this stack overflow post: https://stackoverflow.com/questions/11255027/why-user-defined-move-constructor-disables-the-implicit-copy-constructor/11255258#11255258

My understanding is that since C++11 implicit copy constructors are now "opt-in" so if you want them, (and the CGAL codebase does specifically require them) then they should be explicitly declared with `=default;` of course for backward compatibility they are generated even if you don't declare them (unless you declare a move constructor).

My opinion is that they serve the purpose of documenting that these functions have been considered and that due to the class being trivial, the default implementations are acceptable. In summary, it's an improvement to the software specification, rather than an improvement to the generated code.

mglisse: > My understanding is that since C++11 implicit copy constructors are now "opt-in" so if you want them, (and the CGAL codebase does specifically require them) then they should be explicitly declared with `=default;` of course for backward compatibility they are generated even if you don't declare them (unless you declare a move constructor).

The post you link says that the implicit copy/etc are deprecated **if you have a user-declared destructor**. But we don't have that here. Well, we do in a block protected by `#ifdef CGAL_PROFILE`, so you could add your stuff in that block. And even if the move constructor would be identical to the copy constructor, if we are trying to mimic the implicit stuff, I think we might as well add it with the other 3.

> My opinion is that they serve the purpose of documenting that these functions have been considered and that due to the class being trivial, the default implementations are acceptable. In summary, it's an improvement to the software specification, rather than an improvement to the generated code.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-zero
> C.20: If you can avoid defining default operations, do

GilesBathgate: @mglisse Thanks for your input. I will pass this on to Coverity to see if its a bug in their scanner. It does seem odd that it would suggest you need a move assignment operator when there is no copy assignment operator or copy constructor declared.

GilesBathgate: @mglisse Ah I think the issue is that for the static analysis I was backporting these changes to a version of CGAL which doesn't have commit https://github.com/CGAL/cgal/commit/6c65f5d92983801b566ef8540c2ae3978ed632b0 So the destructor always existed. I will see if adding `#ifdef CGAL_PROFILE` changes the analysis results.

GilesBathgate: @mglisse Yes the destructor was at was the problem (only in 5.0.3)

maxGimeno:
A git bisect marked the merge of this PR in integration as responsible for [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-71/Kernel_23/TestReport_lrineau_ArchLinux-CXX14.gz), but I couldn't reproduce the crash on this  branch alone. For that reason, I will stop testing it until the current batch is merged in master,  and from there we'll have to figure out what to do. Probably a fix in that branch then.

GilesBathgate:
@maxGimeno Shall I rebase this branch onto master?

maxGimeno:
@GilesBathgate No, it's not that. If you try pull `cgal-dev/testsuite-Thursday` and try to run `Filtered_cartesian` in Kernel_23/tests/Kernel_23, you will have a runtime error. This error doesn't occur in master, nore in your branch, nore in any other branch. It's the combination of an unidentified branch with yours that leads to that. As I don't know which, I'd rather wait for the other branches to get in to master and then let you or anyone that might help work on the crash, because I don't know what happens.

GilesBathgate:
@maxGimeno The nature of a move assignment or move constructor is that they will be used instead of a copy assignment or copy constructor when the compiler can optimise a copy to a move. So it's possible in this PR branch that these optimisations cannot be made, but in the other branch, they can. Ultimately its probably this PR that is wrong, but doesn't get executed/tested in this branch.

maxGimeno:
I guess it's the Handle that is used in another PR at some point in a Kernel class, it might be a missuse at that place. I'll keep looking with this new information, maybe I'll find specifically what goes wrong and all will become clear.

maxGimeno:

maxGimeno:
Actually, after a git bisect, it appears that this PR is responsible for the errors in Kernel_23 : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-89/Kernel_23/TestReport_lrineau_ArchLinux-CXX17-Release.gz.

GilesBathgate:
@maxGimeno Any chance you can narrow it down any further, and I will attempt to fix.

STL_Extension/include/CGAL/Handle.h
```diff
@@ -51,6 +51,12 @@ class Handle
       PTR->count++;
     }
 
+    Handle(Handle&& x) noexcept
+      : PTR(x.PTR)
+    {
+      x.PTR = static_cast<Rep*>(0);
+    }
+
```
maxGimeno, line 59: @GilesBathgate it is this block

GilesBathgate: Ah, I added this code as suggested by @mglisse 's comment:

>This violates the rule of 0/3/5, which is more common than your static analysis tool. If we add the move assignment, we should also add
>    Handle(Handle&& x) noexcept : PTR(x.PTR) { x.PTR = 0; }

I must admit, I don't know what the correct implementation should be for a move constructor here.

lrineau: That constructor is fine. If the git-bisection of Maxime is right, that new constructor has triggered a bug, but is only indirectly responsible for that bug.

mglisse:
> Kernel_23 : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-89/Kernel_23/TestReport_lrineau_ArchLinux-CXX17-Release.gz.

Is it expected that the CXXFLAGS used to compile Kernel_23 tests ('-Wall -frounding-math -DCGAL_NDEBUG -O3') have nothing to do with the "official" CXXFLAGS for this platform (Installation says '-Wall -O3 -msse4.1 -std=gnu++17 -DCGAL_NDEBUG -O3')?

lrineau:
With the debugger I found out that the issue is an uninitialized variable. See https://github.com/CGAL/cgal/pull/5363/files#r602446830. That is not related to this PR.

maxGimeno:
@lrineau the question still stands. The flags given to cmake don't match the ones actually logged. -std=c++17 and -msse4.1 don't seem to be used. The fact that the archlinux images are broken doesn't help. It is impossible to update, impossible to install a package inside, like a text editor, as the package repo is broken, etc.

lrineau:
> @lrineau the question still stands. The flags given to cmake don't match the ones actually logged. -std=c++17 and -msse4.1 don't seem to be used. The fact that the archlinux images are broken doesn't help. It is impossible to update, impossible to install a package inside, like a text editor, as the package repo is broken, etc.

I think that ArchLinux platform will be fixed with the commit https://github.com/CGAL/cgal-testsuite-dockerfiles/commit/7d24b315d2e10199c3b86a83198ff2aa40a08c21.

PR #5274, opened 2020-12-13
Allow FT that isn't floating point.

benlaurie:
## Summary of Changes

What it says on the tin.

## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): -

### DIFF
--- a/BGL/include/CGAL/boost/graph/generators.h
+++ b/BGL/include/CGAL/boost/graph/generators.h
@@ -485,13 +485,13 @@ make_regular_prism(typename boost::graph_traits<Graph>::vertices_size_type nb_ve
     put(vpmap, vertices[i],
         P(0.5*diameter * cos(i*precision*to_rad) + base_center.x(),
           height+base_center.y(),
-          -0.5*diameter * sin(i*precision*to_rad) + base_center.z()));
+          -0.5*diameter * sin(to_double((double)i*precision*to_rad)) + base_center.z()));
 
     put(vpmap,
         vertices[i+nb_vertices],
-        P(0.5*diameter * cos(i*precision*to_rad) + base_center.x(),
+        P(0.5*diameter * cos(to_double((double)i*precision*to_rad)) + base_center.x(),
           base_center.y(),
-          -0.5*diameter * sin(i*precision*to_rad) + base_center.z()));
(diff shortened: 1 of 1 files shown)


lrineau:
@MaelRL What do you think of the patch? It looks good to me.

maxGimeno:
I would like a test for that, but I'm not sure there is one already for the bgl generators

sloriot (commented):
I would remove the c-style cast that is harmful and useless IMHO

MaelRL:
> I would remove the c-style cast that is harmful and useless IMHO

That was the reason for the flag. At best it should be `FT(i)`, but it's still useless.

lrineau:
> > I would remove the c-style cast that is harmful and useless IMHO
> 
> That was the reason for the flag. At best it should be `FT(i)`, but it's still useless.

If the type `typename boost::graph_traits<Graph>::vertices_size_type` is a 64 bits integer, then compilers emit a warning about implicit casting to `double` loosing precision, because `double` has only 53 bits of mantissa. That is probably the reason of the explicit cast.

benlaurie:
Some kind of cast is actually needed - try it.

Agree I should've used a C++ cast - I'm a bit rusty on C++ these days!

benlaurie:
Without the cast, you get:

```
//home/ben/local/include/CGAL/boost/graph/generators.h:592:39: error: use of
      overloaded operator '*' is ambiguous (with operand types 'typename
      boost::graph_traits<Surface_mesh<Point_3<Epeck> > >::vertices_size_type'
      (aka 'unsigned int') and 'const FT' (aka 'const
      Lazy_exact_nt<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational,
      boost::multiprecision::et_on> >'))
        P(0.5*diameter*cos(to_double(i*precision*to_rad))+base_center.x(),
                                     ~^~~~~~~~~~
```

MaelRL:
Ah, you're right indeed, sorry.

Then `FT(i)` should be correct.

benlaurie:
Update coming soon.

maxGimeno:

lrineau:
@benlaurie Thanks for your contribution. It will be merged into `master` today.

I have checked your patch, and it *does not* require the signature of a contributor agreement with the CGAL project, because the patch is really small.

PR #5277, opened 2020-12-14
Surface_mesh: Fix sm_remove test

maxGimeno:
## Summary of Changes
Assertions sometimes appear in the test `sm_remove` . We concluded that sometimes, memory is reused and assertions  that test adresses are different fail. So I removed those assertions.

## Release Management


maxGimeno:

Issue #5280, opened 2020-12-14
Seg fault while invoking a shortest path query on a bounded Triangle Mesh.


YanisM42:
## Issue Details

Invoking basic shortest path queries on a bounded surface mesh raises 'segmentation fault'.
Invoking the same queries with closed mesh works fine.

While code can be compiled without issue, the problem comes from execution.

## Source Code

You may find a minimal source code, along with a simple mesh triggering the error. It is to be noted that provided CGAL examples also fail when executing with the given mesh but only when using **CGAL::Surface_mesh<K::Point_3>**. Indeed, it seems that using **CGAL::Polyhedron_3<Kernel, CGAL::Polyhedron_items_with_id_3>** to define the mesh works fine.
However, I am using the former for a project so it would be more suitable to find a solution with **CGAL::Surface_mesh<K::Point_3>**.
[code_and_mesh.zip](https://github.com/CGAL/cgal/files/5691774/code_and_mesh.zip)

## Environment

* Operating system: Linux, 64 bits
* Compiler: g++ - 7
* CGAL version: 5.1.1

Thank you for your support

sloriot:
Your input mesh has an isolated vertex. Calling [`remove_isolated_vertices()`]( https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__repairing__grp.html#ga91c02ef57e638faf2622eae93e7a25e2) fixes the issue. I'll see if I can easily ignore them directly in the code.

YanisM42:
Thank you very much!
Is there an exhaustive list of mesh imperfections that should be checked before invoking a shortest path query? I had used **CGAL::Surface_mesh::is_valid(bool verbose=true)** but it did not display anything and the output boolean was *true*.

PR #5283, opened 2020-12-15
Ignore isolated vertices

sloriot:
Related to #5280


maxGimeno:

PR #5284, opened 2020-12-15
[Small Feature] Unregularized do_intersect(Polygon, Polygon) compatible with inexact constructions

sgiraudot:
___Note:__ this small feature is still a proposal to be discussed and is not ready for review/testing/integration yet._

## Rationale

CGAL currently offers a function `do_intersect()` that, for simplicity, we can distinguish into 2 main variants:

- `CGAL::do_intersect()` provided by the packages `Intersection_2/3`: works on Kernel objects (points, segments, etc.), returns `true` if objects intersect even at their boundaries (e.g. 2 segments sharing a common vertex) compatible with inexact constructions

- `CGAL::do_intersect()` provided by `Boolean_set_operations_2`: works on polygon objects (`Polygon_2`, `Polygon_with_holes_2`, etc.), returns true only if objects overlap (only the "strict" intersection is tested, 2 polygons sharing only a vertex or an edge will return `false`), requires exact constructions

This second variants raises 2 issues: first, the behavior (open interval intersection VS closed interval interseciton) is counter-intuitive and differs from the general function. Second, requiring exact constructions for a function that is obviously a predicate is not satisfying.

After discussions with @efifogel, it is agreed that the difference of behavior is justified by the fact that the _Boolean Set Operations 2_ implements _regularized_ boolean operations, which implies that only the closure of the interior of the result is considered. So this behavior should be kept.

Nevertheless, it would still make sense to add a non-regularized variant of `do_intersect()` for polygons that doesn't require exact constructions.

## Proposal

1. Moving all free-function boolean operations of the `Boolean_set_operations_2` package to a dedicated namespace. My current implementation introduces the namespace `CGAL::Boolean_set_operations_2` as it is common practise to use the name of the package. This is debatable (see below).

2. Adding a non-strict `do_intersect()` test for polygons that does not require exact constructions. My current implementation relies on an exception mechanism that immediately exits the surface sweep as soon as a construction is required (which means that the return value is `true`). It's not the most satisfying way to do it, but in the current state of the Surface Sweep implementation, it seems difficult to do better (using a special visitor or helper) without altering deeply the code. Note that this early ending also means that the function should be much faster than the regularized variant.

This means that there will now be 2 separate functions for polygon `do_intersect()` test:

- `CGAL::do_intersect()` will work similarly as the general `do_intersect()` on kernel objects, same type of intersection (including boundaries) and compatible with Epick

- `CGAL::Boolean_set_operations_2::do_intersect()` will keep the current regularized exact-only test

## Open questions / Discussion

1. __Namespace:__ It would actually be nice to have the word "regularized" in the namespace to make it clear what it is about (`CGAL::Regularized_boolean_operations`?) but I'm not sure what's the best choice. @efifogel any opinion on this?

2. __Package:__ ideally, the unregularized inexact-compatible variant should go in `Intersection_2` and be documented [along with the other variants](https://doc.cgal.org/latest/Kernel_23/group__do__intersect__linear__grp.html#ga9514a218a097ac392d21775368d3dd0d). But this raises two problems:

   - this would technically mean that we should also handle intersections between polygons and other types (`do_intersect(Segment_2, Polygon_2)` for example). Although it would be a nice feature to have, it does seem out of scope

   - this would introduce, for `Intersection_2`, dependencies to `Arrangement_2`, `Surface_sweep`, etc., which is probably unreasonable.

So it's probably better to keep this variant in `Boolean_set_operations_2`, but then we have to find a nice way to document it.

## Summary of API changes

- The current `CGAL::do_intersect()` variant for polygons becomes `CGAL::Boolean_set_operations_2::do_intersect()`
- All other operations in this package are also placed in the namespace `Boolean_set_operations_2`
- A new variant of `CGAL::do_intersect()` is introduced, handling non-strict intersections without requiring exact constructions

(No change)

## CHANGES.md

TODO

## Submission

TODO

## Status

* Code developed and locally tested (GNU/Linux)
* API undecided yet


Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
```diff
@@ -295,30 +296,55 @@ class Gps_on_surface_base_2
...
     Gps_do_intersect_functor<Aos_2>  func;
-    overlay(*m_arr, *(other.m_arr), res_arr, func);
-    return func.found_reg_intersection();
+    try
+    {
+      overlay(*m_arr, *(other.m_arr), res_arr, func);
+    }
+    catch (typename Do_intersect_traits::Exception&)
+    {
+      std::cerr << "Caught exception" << std::endl;
```
sloriot, line 344: remove

PR #5286, opened 2020-12-15
PMP::isotropic_remeshing - fix is_corner(v)

janetournois:
## Summary of Changes

`is_corner(v)` should count both incoming and outgoing halfedges around `v`, because the `Halfedge_status` is not symmetric.

This PR fixes (at least) the case where `v` is part of the border of a facet selection, and an end-vertex of a constrained edge. One incident edge was being collapsed in the wrong direction because the corner-status of `v` was wrong.

## Release Management

* Affected package(s): PMP

@xiaoXiaoCam @lrineau this patch should fix the issue you met


sloriot (approved):

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-20.shtml

PR #5287, opened 2020-12-15
SSE2 version of square for intervals

mglisse:
## Summary of Changes

Add a SSE2 version of CGAL::square for Interval_nt. This gains a factor larger than 3 compared to the scalar code, and reduces the time to build a Delaunay_triangulation of points on a grid by more than 10%. Just replacing square(x) with x*x was enough to get most of the gain, but this version seems a bit faster.

## Release Management

### DIFF
--- a/Number_types/include/CGAL/Interval_nt.h
+++ b/Number_types/include/CGAL/Interval_nt.h
@@ -1159,8 +1159,13 @@ namespace INTERN_INTERVAL_NT {
   Interval_nt<Protected>
   square (const Interval_nt<Protected> & d)
   {
-    //TODO: SSE version, possibly using abs
     typename Interval_nt<Protected>::Internal_protector P;
+#ifdef CGAL_USE_SSE2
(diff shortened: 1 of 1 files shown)


lrineau:
Thanks for that improvement. What is so wrong in `square (const Interval_nt<Protected>&)` that replacing it with a call to `x*x` improves the speed?!

mglisse:
I guess that the unpredictable branch really hurts, although I didn't check an instruction-level profile. I was already very surprised to see `square` appear at all in the profile of a (dD) Delaunay triangulation (in the simulation of simplicity IIRC), and since I had left a helpful comment for my future self...

lrineau:
> Thanks for that improvement. What is so wrong in `square (const Interval_nt<Protected>&)` that replacing it with a call to `x*x` improves the speed?!

Self answer: `Interval_nt operator* (const Interval_nt &, const Interval_nt &)` is already implemented using SSE2 (or even SSE4 or AVX when possible). That explains why it is already faster than the `square` function overload.

maxGimeno:

PR #5288, opened 2020-12-15
Use a small_vector in Delaunay_triangulation

mglisse:
## Summary of Changes

Use a small_vector in Delaunay_triangulation. It is only used locally in a function, and its size is at most the dimension + 2.

Making the vector thread_local would work as well, but we already had a small_vector in Triangulation.h.

This saves about 7% on Delaunay for some specific point sets.

## Release Management


lrineau (approved):

maxGimeno:

PR #5290, opened 2020-12-16
Add "binary" expression in file open of registration_with_OpenGR example

maximecharriere:
## Summary of Changes
`registration_with_OpenGR` example can't be compilled without this small change:
```c++
std::ifstream input(fname_in1);
```
to
```c++
std::ifstream input(fname_in1, std::ios::binary);
```
`data/hippo1.ply` and `data/hippo2.ply` are in binary_little_endian format. So `read_ply_points` can't read them if they aren't opened in binary mode.

Also add a specific error if a file can't be open and a new argument to pass the output filename

## Release Management

* Affected package(s): Point_set_processing_3

### DIFF
--- a/Point_set_processing_3/examples/Point_set_processing_3/registration_with_OpenGR.cpp
+++ b/Point_set_processing_3/examples/Point_set_processing_3/registration_with_OpenGR.cpp
@@ -25,7 +25,7 @@ int main(int argc, const char** argv)
   const char* fname2 = (argc>2)?argv[2]:"data/hippo2.ply";
 
   std::vector<Pwn> pwns1, pwns2;
-  std::ifstream input(fname1);
+  std::ifstream input(fname1, std::ios::binary);
   if (!input ||
       !CGAL::read_ply_points(input, std::back_inserter(pwns1),
             CGAL::parameters::point_map (CGAL::First_of_pair_property_map<Pwn>()).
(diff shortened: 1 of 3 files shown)


sgiraudot (changes requested):
I agree with the proposed change (adding the `std::ios::binary` tag) but the diff is way too large for such a small change. Please rewrite your commit without reindenting the whole files you modify (and please use spaces and not tabs for consistency with the rest of the files).

maximecharriere:
I have revert my changes and only add the `binary` parameter.

MaelRL:
@maxGimeno If you change the milestone to 5.3, the PR loses its purpose since this was fixed in the IO PR already.

lrineau:
> @maxGimeno If you change the milestone to 5.3, the PR loses its purpose since this was fixed in the IO PR already.

As the base branch of the PR includes commits from `master` only (like ef13599d963), we cannot merge this PR to anything to `master`, anyway.

MaelRL:
https://github.com/CGAL/cgal/pull/5323 has the same diff, but targets 5.1. This PR can probably be closed.

sgiraudot:
Sorry, I completely forgot about this PR, it could also have been rebased…

MaelRL:
Replaced by #5323.

PR #5292, opened 2020-12-17
CTest: Fix some examples and tests

maxGimeno:
## Summary of Changes

Some examples and tests were not added as test(using cgal_add_test or cgal_create_source_file), and thus were ignored by ctest.


maxGimeno:

PR #5294, opened 2020-12-17
Fix unchecked return value warning in SM_point_locator.h

GilesBathgate:
## Summary of Changes

In SM_point_locator.h a call to `do_intersect_internally` doesn't require the return value to be checked. It's customary to cast the function to void to indicate  "Yes, I've decided to ignore the return value from this call"

## Release Management

* Affected package(s): Nef_S2
* Issue(s) solved (if any): bugfix / cleaning

### DIFF
--- a/Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h
+++ b/Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h
@@ -493,7 +493,7 @@ public:
           Sphere_segment first_half(p,p.antipode(),c);
           Sphere_segment second_part(p.antipode(),p,c);
           if(!do_intersect_internally(ei->circle(), first_half, p_res)) {
-            do_intersect_internally(ei->circle(), second_part, p_res);
+            (void)do_intersect_internally(ei->circle(), second_part, p_res);
           }
         }
 


lrineau:
@GilesBathgate Which compiler gives a warning on such code? We would like to reproduce.

GilesBathgate:
@lrineau I meant to say that this was a warning from static analysis. I think compilers will only emit this warning when the attribute `__attribute__((warn_unused_result));` is added to the function declaration.

lrineau:
@GilesBathgate Don't you think that work will be endless, if we start addressing that sort of warning in CGAL?

GilesBathgate:
> @GilesBathgate Don't you think that work will be endless, if we start addressing that sort of warning in CGAL?

@lrineau This was the only "Unchecked return value" that cropped up in the analysis. I think the analysis notes that the return value is checked in several other calls to that function in the same area of code. However I only use a subset of the packages in my project, so I can't say for sure whether it would likely find more.

GilesBathgate:
@lrineau I just found this "success story£, It gives a flavour of the static analysis output:

lrineau:
@sloriot I think you have worked on Nef_3 in the past. Do you see why that return value is ignored?
  - Is there a reason why the returned value is certainly `true`?

GilesBathgate:
@lrineau I think because the function has "side effects" which are needed, but it doesn't have to do anything else if it returns false. Maybe exception handling is a better design, but it was done for performance?

GilesBathgate:
@lrineau An alternative fix could be:
```
if(!do_intersect_internally(ei->circle(), first_half, p_res) &&
   !do_intersect_internally(ei->circle(), second_part, p_res)) 
{
}
```
But that then begs the question of why is there an empty If body :joy:

GilesBathgate:
@lrineau Actually I think its safe to add a `continue;` as the if body, and then at least its somewhat consistent with the rest of the code.

GilesBathgate:
@lrineau @sloriot Two unchecked calls were first introduced in this commit: https://github.com/CGAL/cgal/commit/1ec1a530f8e56c681f67783e14f75bc625e5c087

But one of them was later fixed in this commit: https://github.com/CGAL/cgal/commit/e0595a049024d24eada705352c457907598ce770

afabri:
Isn't the warning there because this function has no side effects but only tests something and does nothing based on the result? There was an assertion on the result. So maybe the function call should go in an `assertion_code()` statement?

lrineau:
Before we agree to "hide" the warning, are we sure that is intended to ignore that returned value?

Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h
```diff
@@ -492,9 +492,7 @@ class SM_point_locator : public SM_decorator {
         } else {
           Sphere_segment first_half(p,p.antipode(),c);
           Sphere_segment second_part(p.antipode(),p,c);
-          if(!do_intersect_internally(ei->circle(), first_half, p_res)) {
-            (void)do_intersect_internally(ei->circle(), second_part, p_res);
-          }
+          CGAL_assertion( (! do_intersect_internally(ei->circle(), first_half, p_res) ) && do_intersect_internally(ei->circle(), second_part, p_res));
```
lrineau, line 495: That looks wrong. The assertion should be like `do_intersect_internally || do_intersect_internally`.

GilesBathgate:
> Isn't the warning there because this function has no side effects 

I thought the function does have side effects since p is passed by reference:
https://github.com/CGAL/cgal/blob/74af3e7d33e7354e4090a1b17941a3f0c372ca8d/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h#L355-L369

Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h
GilesBathgate: yeah:

```c++
if(!a) throw;
if(b) throw;
```

is equivalent to:

```c++
if (!a || b) throw;
```

afabri:
You are right about the side effect.

afabri:
> 
> 
> You are right about the side effect.

I just undid my commit.

PR #5295, opened 2020-12-18
Add move ctor and move assign to Side_of_triangle_mesh

barcode:
## Summary of Changes

Add move ctor and move assign to Side_of_triangle_mesh.
These are only added, if `BOOST_NO_CXX11_RVALUE_REFERENCES` is defined

## Release Management

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
@@ -181,7 +181,6 @@ public:
     box = tree.bbox();
   }
 
-#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
   /**
   * Constructor moving an instance of Side_of_triangle_mesh to a new memory
   * location with minimal memory copy.
@@ -200,7 +199,6 @@ public:
   {
     other.own_tree = false;
   }
-#endif
 
   ~Side_of_triangle_mesh()
   {
@@ -213,7 +211,6 @@ public:
   }
 
 public:
-#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
   /**
   * Assign operator moving an instance of Side_of_triangle_mesh to this
   * location with minimal memory copy.
@@ -234,7 +231,6 @@ public:
     #endif
     return *this;
   }
-#endif
 
   /**
    * returns the location of a query point
--- a/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
@@ -181,6 +181,27 @@ public:
     box = tree.bbox();
   }
 
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+  /**
+  * Constructor moving an instance of Side_of_triangle_mesh to a new memory
+  * location with minimal memory copy.
+  * @param other The instance to be moved
+  */
+  Side_of_triangle_mesh(Side_of_triangle_mesh&& other) :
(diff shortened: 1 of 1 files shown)


lrineau:
Hi @barcode. Thanks for that contribution.

> These are only added, if `BOOST_NO_CXX11_RVALUE_REFERENCES` is defined

Please remove the checks for `BOOST_NO_CXX11_RVALUE_REFERENCES` because CGAL requires C++14 since CGAL-5.0.

sloriot:
@maxGimeno you could add the same thing in `Rigid_triangle_mesh_collision_detection`.

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -181,6 +181,27 @@ class Side_of_triangle_mesh
...
+  /**
+  * Constructor moving an instance of Side_of_triangle_mesh to a new memory
+  * location with minimal memory copy.
+  * @param other The instance to be moved
+  */
+  Side_of_triangle_mesh(Side_of_triangle_mesh&& other) :
+    tm_ptr{other.tm_ptr},
+    opt_vpm{std::move(other.opt_vpm)},
+    own_tree{other.own_tree},
+    box{other.box},
```
lrineau, line 194: I would prefer to see `std::move` called on all members. I agree with the current implementation a simple copy will be made anyway, but we might change the definition of some of those members, one day: for example `tm_ptr` could become a `std::unique_ptr`.

And I think, to be correct, the move constructor and move assignment operator must deal with all members, including `ray_functor` and `vector_functor`, so that the code is correct even if those functors have data members.

barcode:
> Please remove the checks for `BOOST_NO_CXX11_RVALUE_REFERENCES` because CGAL requires C++14 since CGAL-5.0.

Oh, I did not know that. (Fixed)

> I would prefer to see std::move called on all members.

Done

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
```diff
@@ -181,6 +181,25 @@ class Side_of_triangle_mesh
...
+    box{std::move(other.box)},
+    #ifdef CGAL_HAS_THREADS
+      atomic_tree_ptr{other.atomic_tree_ptr.load()}
+    #else
+      tree_ptr{std::move(other.tree_ptr)}
+    #endif
+  {
+    other.own_tree = false;
+  }
+
```
sloriot, line 202:
```suggestion
  Side_of_triangle_mesh(Side_of_triangle_mesh&& other)
  {
    *this = std::move(other);
  }
```

maxGimeno:

Polygon_mesh_processing/include/CGAL/Side_of_triangle_mesh.h
lrineau: @barcode Are you still willing to contribute this pull-request? Sébastien made a suggestion a week ago, but you have not yet answered.

This PR has been tested successfully in our `integration` branch, but I cannot merge it into CGAL if there is an ongoing suggestion.

barcode: I was busy with private life, so i did not look in here.
I don't see any problem with the suggestion (and it seems this is already marked as resolved).

lrineau: Right. Sébastien already did it in https://github.com/CGAL/cgal/pull/5295/commits/c30056b126f794d3fb2e9379a5a33162a6e042fd.

maxGimeno:

PR #5297, opened 2020-12-18
Polyhedron Demo: C3t3 colors enhancement

maxGimeno:
## Summary of Changes

If there is only one patch in a c3t3, de-synchronize the tets and item colors.


maxGimeno:

PR #5304, opened 2020-12-23
Add squared_distance from point3 to tetrhaedron3

jjcasmar:
This patch adds a squared_distance function from a point to a tetrahedron. A point inside a tetrahedron is considered to have a distance of 0. The distance from a point outside the tetrahedron to the tetrahedron is the smallest distance from the point to any of the 4 triangles defining the boundary of the tetrahedron.

### DIFF
--- a/Distance_3/include/CGAL/squared_distance_3_2.h
+++ b/Distance_3/include/CGAL/squared_distance_3_2.h
@@ -28,6 +28,7 @@
 #include <CGAL/Ray_3.h>
 #include <CGAL/Triangle_3.h>
 #include <CGAL/Plane_3.h>
+#include <CGAL/Tetrahedron_3.h>
 
 namespace CGAL {
 
@@ -387,6 +388,27 @@ typename K::FT
 squared_distance(const Triangle_3<K> & t,
                  const Point_3<K> & pt) {
   return internal::squared_distance(pt, t, K());
+}
+
+template <class K>
+inline
+typename K::FT
+squared_distance(const Tetrahedron_3<K> & t,
+                 const Point_3<K> & pt) {
+  if (t.has_on_boundary(pt) || t.has_on_bounded_side(pt))
+      return K::FT(0);
+
+  const Triangle_3<K> t0 = {t[0], t[1], t[2]};
+  const Triangle_3<K> t1 = {t[0], t[1], t[3]};
+  const Triangle_3<K> t2 = {t[1], t[2], t[3]};
+  const Triangle_3<K> t3 = {t[0], t[2], t[3]};
+
+  const typename K::FT d0 = CGAL::squared_distance(t0, pt);
+  const typename K::FT d1 = CGAL::squared_distance(t1, pt);
+  const typename K::FT d2 = CGAL::squared_distance(t2, pt);
+  const typename K::FT d3 = CGAL::squared_distance(t3, pt);
+
+  return std::min({d0, d1, d2, d3});
 }
 
 


Distance_3/include/CGAL/squared_distance_3_2.h
```diff
@@ -389,6 +390,27 @@ squared_distance(const Triangle_3<K> & t,
   return internal::squared_distance(pt, t, K());
 }
 
+template <class K>
+inline
+typename K::FT
+squared_distance(const Tetrahedron_3<K> & t,
+                 const Point_3<K> & pt) {
+  if (t.has_on_boundary(pt) || t.has_on_bounded_side(pt))
```
mglisse, line 398: `has_on_unbounded_side`?

Distance_3/include/CGAL/squared_distance_3_2.h
```diff
@@ -389,6 +390,27 @@ squared_distance(const Triangle_3<K> & t,
...
+
+  const Triangle_3<K> t0 = {t[0], t[1], t[2]};
+  const Triangle_3<K> t1 = {t[0], t[1], t[3]};
+  const Triangle_3<K> t2 = {t[1], t[2], t[3]};
+  const Triangle_3<K> t3 = {t[0], t[2], t[3]};
+
+  const typename K::FT d0 = CGAL::squared_distance(t0, pt);
+  const typename K::FT d1 = CGAL::squared_distance(t1, pt);
+  const typename K::FT d2 = CGAL::squared_distance(t2, pt);
+  const typename K::FT d3 = CGAL::squared_distance(t3, pt);
```
mglisse, line 409: This may involve some redundant computation. I am not saying we shouldn't do this, it can be better to have the functionality and only try to optimize later, but we should be aware that this is probably not optimal.

mglisse (commented):
If I understand the naming scheme correctly, this should go to a new file `squared_distance_3_3.h`.

Distance_3/include/CGAL/squared_distance_3_2.h
jjcasmar: That is the interior of the tetrahedron, right?

Distance_3/include/CGAL/squared_distance_3_2.h
jjcasmar: Could be. I dont really know how the computation between a point and a triangle is performed

Distance_3/include/CGAL/squared_distance_3_2.h
afabri: What Marc suggests is to replace the two tests by one  (and a negation).

afabri:
The new combination must also be documented [here ](https://github.com/CGAL/cgal/blob/master/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h#L2786).

afabri:
@MaelRL why do you consider this a small feature?

afabri:
@jjcasmar I've reimplemented it in another branch as it was another file, and somehow another code.  Thank you nevertheless for contributing to the CGAL project.

jjcasmar:
@afabri thanks. I didnt find time to update the patch, so thank you for taking care of polishing it!

MaelRL:
> @MaelRL why do you consider this a small feature?

That was just with respect to the doc addition you had requested, but I guess it's minor enough.

MaelRL:
Replaced by #5465.

PR #5324, opened 2021-01-05
PSP_3: Remove debug output

cjamin:
Just a minor patch: I think this output was for debug purpose.


maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-45.shtml

PR #5326, opened 2021-01-05
Optimize trivial predicates

mglisse:
## Summary of Changes

Some predicates do not perform any arithmetic operation and only do comparisons. For Epick_d, we can do them directly with `double`, no need for intervals. For Epeck_d, we do use intervals, but we do not need to set the rounding mode.

This last optimization might be a little bit risky. First, interval comparisons might at some point in the future require protection. Second, other operations might happen outside of the predicate, in particular the conversion to approximate (C2A), and some operations like constructing an interval from `int64_t` do require protection. I think that the restriction to Epeck_d (where we already have intervals) and to a small list of predicates makes this ok... :crossed_fingers:

## Release Management

### DIFF
--- a/Filtered_kernel/include/CGAL/Filtered_predicate.h
+++ b/Filtered_kernel/include/CGAL/Filtered_predicate.h
@@ -37,6 +37,10 @@ namespace CGAL {
 //   not, or we let all this up to the compiler optimizer to figure out ?
 // - Some caching could be done at the Point_2 level.
 
+// Protection is undocumented and currently always true, meaning that it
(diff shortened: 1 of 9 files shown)


mglisse:
By the way, does anyone ever run the testsuite with `-DCGAL_CHECK_EXPENSIVE`? There are checks in FPU.h that could help detect if we end up converting int64_t to Interval_nt (and thus doing arithmetic) with the wrong rounding mode, but they are CGAL_expensive_assertion and I am afraid noone will ever notice...

maxGimeno:

PR #5329, opened 2021-01-06
Remove mentions to python2

### LINKED ISSUE #5320: Documentation scripts should use Python 3
## Issue Details

Some scripts used to generate documentation (and maybe others, I haven't checked) still use Python 2 (for example [html_output_post_processing.py](https://github.com/CGAL/cgal/blob/master/Documentation/doc/scripts/html_output_post_processing.py)). The support for Python 2 ended last year and some recent OS (like the latest versions of Ubuntu and Linux Mint) stopped providing `pip` for Python 2, which makes compiling the doc tedious on such OS. This will become a common problem in the medium term.

I have no idea how hard it would be to translate these scripts to Python 3, but we'll have to do it in a near future.

sloriot:
Fixes #5320

### DIFF
--- a/Documentation/doc/scripts/pkglist_filter.py
+++ b/Documentation/doc/scripts/pkglist_filter.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 import codecs
 import re
--- a/Documentation/doc/CMakeLists.txt
+++ b/Documentation/doc/CMakeLists.txt
@@ -27,7 +27,7 @@ else()
 endif()
 
 find_package(Doxygen)
-find_package(PythonInterp 2.6.7)
+find_package(PythonInterp)
 
 if(NOT DOXYGEN_FOUND)
   message(WARNING "Cannot build the documentation without Doxygen!")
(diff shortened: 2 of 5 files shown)


Documentation/doc/scripts/pkglist_filter.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
mglisse, line 1: Note that some systems do not have anything called `python`, they only have `python3` (and possibly `python2`). I don't know if the reverse exists.

sloriot: I did not think about that. Maybe I should replace with python3 but it does not really matter as the python exe used is always the one found at cmake level.

sgiraudot: This line could also just be removed, I think. Unless I'm wrong, it's only used if someone tries to launch the script as an executable using `./script.py`, while using `python script.py` will work without it.

sloriot:
@MaelRL  what is the TODO?

maxGimeno:

PR #5335, opened 2021-01-07
Obsolete doxygen setting COLS_IN_ALPHA_INDEX

albert-github:
In the doxygen version 1.9.0 the setting `COLS_IN_ALPHA_INDEX` has been made obsolete and generates a warning during building.
As the default value is used it can also be removed (doxygen will automatically take the default value in the older versions).

### DIFF
--- a/Documentation/doc/resources/1.8.20/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.8.20/BaseDoxyfile.in
@@ -1188,13 +1188,6 @@ VERBATIM_HEADERS       = NO
 
 ALPHABETICAL_INDEX     = NO
 
-# The COLS_IN_ALPHA_INDEX tag can be used to specify the number of columns in
-# which the alphabetical index list will be split.
-# Minimum value: 1, maximum value: 20, default value: 5.
-# This tag requires that the tag ALPHABETICAL_INDEX is set to YES.
-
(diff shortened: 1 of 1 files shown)


sloriot:
There is nothing to test as it is the default value (I looked at the diff with the version generated with 1.8.20).

lrineau:
Is that on purpose that the removal of those lines was only for 1.8.20 but not 1.8.14 or earlier?

lrineau:
> Is that on purpose that the removal of those lines was only for 1.8.20 but not 1.8.14 or earlier?

Self-reply: yes, because the obsolescence is only since 1.9.0 (that is more recent than 1.8.20).

albert-github:
An idea might be to reduce the BaseDoxyfile.in to just those items the differ from the default setting (not forgetting that the setting ` EXTRACT_ALL_NO_DETAILED_IF_EMPTY` has to remain):
- doxygen will use for non specified settings the default value
- in case new obsolescence settings that had the default value  nothing has to be done
- it is now already the case that new settings are not in the file and for these the default value is taken

sloriot:
I agree that it would be nice if we are sure that the default values do not change.

albert-github:
@sloriot In respect to the default values doxygen keeps the constant (I cannot even remember a case where it was changed or even a default value was improved).

PR #5336, opened 2021-01-07
Graphics_view: Upgrade glsl version in basic_viewer_shaders

maxGimeno:
## Summary of Changes

Recent MacOS distributions do not support GLGL 120, so I upgraded to 150. 
I also added a function call in every `draw_XXX_example`, that should set up  the more recent context possible to use for the examples. 
That should fix the macOS bug that prevent the drawing of edges and points, but the edge size will probably be stuck at 1 as most drivers don't implement other sizes (cf https://stackoverflow.com/questions/34866964/opengl-gllinewidth-doesnt-change-size-of-lines). 

## Release Management

* Affected package(s):GraphicsView

### DIFF
--- a/AABB_tree/demo/AABB_tree/AABB_demo.cpp
+++ b/AABB_tree/demo/AABB_tree/AABB_demo.cpp
@@ -19,12 +19,12 @@
 #include <QApplication>
 #include <CGAL/Qt/resources.h>
 #include <QMimeData>
-#include <CGAL/Qt/Context_initialization.h>
+#include <CGAL/Qt/init_ogl_context.h>
 
 int main(int argc, char **argv)
 {
 
-  CGAL::init_ogl_context(4,3);
+  CGAL::Qt::init_ogl_context(4,3);
   QApplication app(argc, argv);
   app.setOrganizationDomain("inria.fr");
   app.setOrganizationName("INRIA");
--- a/AABB_tree/demo/AABB_tree/AABB_demo.cpp
+++ b/AABB_tree/demo/AABB_tree/AABB_demo.cpp
@@ -19,24 +19,12 @@
 #include <QApplication>
 #include <CGAL/Qt/resources.h>
 #include <QMimeData>
-
+#include <CGAL/Qt/Context_initialization.h>
 
 int main(int argc, char **argv)
 {
-  QSurfaceFormat fmt;
-#ifdef Q_OS_MAC
-  fmt.setVersion(4, 1);
-#else
-  fmt.setVersion(4, 3);
-#endif
-  fmt.setRenderableType(QSurfaceFormat::OpenGL);
-  fmt.setProfile(QSurfaceFormat::CoreProfile);
-  fmt.setOption(QSurfaceFormat::DebugContext);
-  QSurfaceFormat::setDefaultFormat(fmt);
-  //for windows
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 3, 0))
-  QApplication::setAttribute(Qt::AA_UseDesktopOpenGL);
-#endif
+
+  CGAL::init_ogl_context(4,3);
   QApplication app(argc, argv);
   app.setOrganizationDomain("inria.fr");
   app.setOrganizationName("INRIA");
--- a/AABB_tree/demo/AABB_tree/AABB_demo.cpp
+++ b/AABB_tree/demo/AABB_tree/AABB_demo.cpp
@@ -23,14 +23,24 @@
 
 int main(int argc, char **argv)
 {
+  QSurfaceFormat fmt;
+#ifdef Q_OS_MAC
+  fmt.setVersion(4, 1);
+#else
+  fmt.setVersion(4, 3);
+#endif
+  fmt.setRenderableType(QSurfaceFormat::OpenGL);
+  fmt.setProfile(QSurfaceFormat::CoreProfile);
+  fmt.setOption(QSurfaceFormat::DebugContext);
+  QSurfaceFormat::setDefaultFormat(fmt);
+  //for windows
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 3, 0))
+  QApplication::setAttribute(Qt::AA_UseDesktopOpenGL);
+#endif
   QApplication app(argc, argv);
   app.setOrganizationDomain("inria.fr");
   app.setOrganizationName("INRIA");
   app.setApplicationName("AABB tree demo");
-  //for windows
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 3, 0))
-  app.setAttribute(Qt::AA_UseDesktopOpenGL);
-#endif
 
   // Import resources from libCGALQt (Qt5).
   CGAL_QT_INIT_RESOURCES;
--- a/AABB_tree/demo/AABB_tree/Scene.cpp
+++ b/AABB_tree/demo/AABB_tree/Scene.cpp
@@ -156,11 +156,11 @@ void Scene::compile_shaders()
     const char tex_fragment_source[] =
     {
         "#version 150 \n"
-        "uniform sampler2D texture;\n"
+        "uniform sampler2D s_texture;\n"
         "in highp vec2 texc;\n"
         "out highp vec4 out_color; \n"
         "void main(void) { \n"
-        "out_color = texture2D(texture, texc.st);\n"
+        "out_color = vec4(texture(s_texture, texc));\n"
         "} \n"
         "\n"
     };
--- a/AABB_tree/demo/AABB_tree/Scene.cpp
+++ b/AABB_tree/demo/AABB_tree/Scene.cpp
@@ -91,8 +91,8 @@ void Scene::compile_shaders()
     //Vertex source code
     const char vertex_source[] =
     {
-        "#version 120 \n"
-        "attribute highp vec4 vertex;\n"
+        "#version 150 \n"
+        "in highp vec4 vertex;\n"
         "uniform highp mat4 mvp_matrix;\n"
         "uniform highp mat4 f_matrix;\n"
         "void main(void)\n"
@@ -103,10 +103,11 @@ void Scene::compile_shaders()
     //Vertex source code
     const char fragment_source[] =
     {
-        "#version 120 \n"
+        "#version 150 \n"
         "uniform highp vec4 color; \n"
+        "out highp vec4 out_color; \n"
         "void main(void) { \n"
-        "gl_FragColor = color; \n"
+        "out_color = color; \n"
         "} \n"
         "\n"
     };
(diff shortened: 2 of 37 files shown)


AABB_tree/demo/AABB_tree/Scene.cpp
```diff
@@ -1319,12 +1321,8 @@ void Scene::deactivate_cutting_plane()
 }
 void Scene::initGL()
 {
-    gl = new QOpenGLFunctions_2_1();
-   if(!gl->initializeOpenGLFunctions())
-    {
-        qFatal("ERROR : OpenGL Functions not initialized. Check your OpenGL Verison (should be >=3.3)");
-        exit(1);
-    }
```
lrineau, line 1327: Why have you removed the check for the return value of `initializeOpenGLFunctions`?

maxGimeno: because in QOpenGLFunctions() it is not a boolean anymore, only a void.

lrineau:
I have an issue with such a change. How can we test it correctly, with different OpenGL drivers?

maxGimeno:
You can setup an opengl context before starting the application. See https://github.com/CGAL/cgal/pull/5336/files#diff-b04e851d62aebec2b69e526ef98f8145132460dd275e97d387699b794ae8a584R11 for an example.

AABB_tree/demo/AABB_tree/AABB_demo.cpp
```diff
@@ -23,14 +23,24 @@
 
 int main(int argc, char **argv)
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 26: Why here you don't use code_to_call_before_creation_of_QCoreApplication ?

Alpha_shapes_3/demo/Alpha_shapes_3/Alpha_shape_3.cpp
```diff
@@ -8,16 +8,28 @@
 
 int main(int argc, char** argv)
 {
- QApplication application(argc,argv);
+  QSurfaceFormat fmt;
```
gdamiand, line 11: Same question: Why here you don't use code_to_call_before_creation_of_QCoreApplication ?

Circular_kernel_3/demo/Circular_kernel_3/Circular_kernel_3.cpp
```diff
@@ -3,15 +3,25 @@
 
 int main(int argc, char** argv)
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 6: Same question: why here you don't use code_to_call_before_creation_of_QCoreApplication ?

GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
```diff
@@ -54,21 +54,41 @@
 
 namespace CGAL
 {
+int& code_to_call_before_creation_of_QCoreApplication(int& i) {
```
gdamiand, line 57: name of the function: initialize_QCoreApplication ?
I don't understand the reason of i parameter. It is unused in this function, and only return. And why by reference ?

Linear_cell_complex/demo/Linear_cell_complex/Linear_cell_complex_3_demo.cpp
```diff
@@ -23,15 +23,15 @@ int main(int argc, char** argv)
   // std::cout<<"Size of dart: "<<sizeof(LCC::Dart)<<std::endl;
   CGAL::set_error_behaviour(CGAL::ABORT);
 
+  //for windows
```
gdamiand, line 26: Can you move this setAttribute into your new function ?

Linear_cell_complex/include/CGAL/draw_linear_cell_complex.h
```diff
@@ -406,7 +406,7 @@ void draw(const CGAL_LCC_TYPE& alcc,
   {
     int argc=1;
     const char* argv[2]={"lccviewer","\0"};
-    QApplication app(argc,const_cast<char**>(argv));
+    QApplication app(CGAL::code_to_call_before_creation_of_QCoreApplication(argc),const_cast<char**>(argv));
```
gdamiand, line 409: Do you really need to call the function in the constructor ? 
In other demo it is called before.
(same question for all the draw_XXX functions)

Periodic_3_triangulation_3/demo/Periodic_3_triangulation_3/periodic_3_triangulation_3_demo.cpp
```diff
@@ -4,6 +4,12 @@
 
 int main(int argc, char *argv[])
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 7: Use the new function.

Periodic_3_triangulation_3/demo/Periodic_Lloyd_3/Periodic_Lloyd_3.cpp
```diff
@@ -4,15 +4,20 @@
 
 int main(int argc, char** argv)
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 7: Use the new function.

Principal_component_analysis/demo/Principal_component_analysis/PCA_demo.cpp
```diff
@@ -21,6 +21,16 @@
 
 int main(int argc, char **argv)
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 24: Use the new function.

Triangulation_3/demo/Triangulation_3/T3_demo.cpp
```diff
@@ -16,15 +16,21 @@
 
 int main(int argc, char** argv)
 {
+  QSurfaceFormat fmt;
```
gdamiand, line 19: Use the new function.

AABB_tree/demo/AABB_tree/AABB_demo.cpp
maxGimeno: At some point I realized it was the same. The point is to set the default before the context is created, but not especially after the application is launched.

GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
maxGimeno: The function is to be called as `argc`, it is a trick to call a function before the initialization of a program, so it takes the app's `argc` and returns it, even it doesn't use it as the app will need them.

Linear_cell_complex/include/CGAL/draw_linear_cell_complex.h
maxGimeno: I thought so, maybe it has been the case in some older Qt version, but now it is not useful anymore. I started by copying what was done in the polyhedorn demo, and as I had to repete the same operations many times I started to think about it, and changed things  a little.

GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
lrineau: Actually, that function was only *needed* for a special case, involving a constructor, in [`Polyhedron_demo::Polyhedron_demo`](https://github.com/CGAL/cgal/blob/077937383555a61e5b23460fc21cf75f3f005018/Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp#L40-L48).

For any other usage, the initialization just has to be done early in the `main()` function.

lrineau: So, I agree with Guillaune: the name and the usage of it is wrong.

@maxGimeno Create a new function that does the job, and place it in a header that can be used by any of our 3D demo. Then, in `code_to_call_before_creation_of_QCoreApplication` from `Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp`, just call that demo.

And **never** copy-paste, please, when a function can be factorized in a header.

lrineau: And that was needed to keep the API of `Polyhedron_demo` simple, like that:
https://github.com/CGAL/cgal/blob/077937383555a61e5b23460fc21cf75f3f005018/Polyhedron/demo/Polyhedron/Mesh_3.cpp#L11-L18

So, everything had to be done in the constructor, and before the call to the base constructor `QApplication::QApplication`. That is why that function takes a `int&` and returns it untouched: that was a trick to insert the call to `code_to_call_before_creation_of_QCoreApplication` in the arguments of the constructor of `QApplication`.

lrineau: But that trick has useless, and looks strange, in the code of the other 3D demos, so, as I said in https://github.com/CGAL/cgal/pull/5336#discussion_r555579202:
>@maxGimeno Create a new function that does the job, and place it in a header that can be used by any of our 3D demo. Then, in `code_to_call_before_creation_of_QCoreApplication` from `Polyhedron/demo/Polyhedron/Polyhedron_demo.cpp`, just call that demo.

GraphicsView/include/CGAL/Qt/Context_initialization.h
```diff
@@ -0,0 +1,48 @@
...
+inline void init_ogl_context(int major, int minor) {
+  QSurfaceFormat fmt;
+#ifdef Q_OS_MAC
+  if(major == 4)
+  {
+    fmt.setVersion(4, 1);
+  }
+  else
+  {
+     fmt.setVersion(major, minor);o
```
lrineau, line 27:
```suggestion
     fmt.setVersion(major, minor);
```

GraphicsView/include/CGAL/Qt/Context_initialization.h
```diff
@@ -0,0 +1,48 @@
...
+  QCoreApplication::setAttribute(::Qt::AA_UseDesktopOpenGL);
+#endif
+
+  //We set the locale to avoid any trouble with VTK
+  setlocale(LC_ALL, "C");
+}
+
+}
+
+#endif // CGAL_QT_CONTEXT_INITIALIZATION_H
```
lrineau, line 48: As this header contains only one function, it should be named `CGAL/Qt/init_ogl_context.h`. And, usually, headers in `CGAL/Qt/` should add things to the `CGAL::Qt` namespace, and not `CGAL`.

lrineau (approved):
@maxGimeno @gdamiand I think I am done with remarks. I like the modifications. I hope they are fine at runtime on all platforms.

gdamiand:
> @maxGimeno @gdamiand I think I am done with remarks. I like the modifications. I hope they are fine at runtime on all platforms.

Same for me, much better. Just a last question: is it normal to use different context ? (sometimes 4.3 and other times 2.1 ?)

maxGimeno:
I keep 2.1 for demos that use QOpenGLFunctions_2_1, to reduce the diff. Code that use `Basic_viewer_qt` can support 4.3 easily.

maxGimeno:

PR #5337, opened 2021-01-07
Compile without MPFR and GMP

gaschler:
## Summary of Changes

For configurations with CGAL_NO_GMP and CGAL_NO_MPFR, includes
to MPFR, GMP, GMPXX are usually removed by #ifdefs
This change adds the missing #ifdefs to ensure that CGAL compiles
when MPFR or GMP do not exist.

### DIFF
--- a/Algebraic_kernel_d/include/CGAL/RS/dyadic.h
+++ b/Algebraic_kernel_d/include/CGAL/RS/dyadic.h
@@ -13,8 +13,12 @@
 
 #include <stdio.h>
 #include <math.h>
+#ifdef CGAL_USE_GMP
 #include <gmp.h>
+#endif
+#ifdef CGAL_USE_MPFR
 #include <mpfr.h>
+#endif
 #include <CGAL/assertions.h>
 
 // for c++, compile with -lgmpxx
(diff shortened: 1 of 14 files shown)


Algebraic_kernel_d/include/CGAL/RS/dyadic.h
```diff
@@ -13,8 +13,12 @@
 
 #include <stdio.h>
 #include <math.h>
+#ifdef CGAL_USE_GMP
 #include <gmp.h>
+#endif
+#ifdef CGAL_USE_MPFR
```
mglisse, line 19: This does not make much sense to me, the file uses MPFR unconditionally, removing the includes won't help if you try to include this file without MPFR.

mglisse:
Can you explain what you are trying to achieve? Most of those headers are files that you should not include if you do not have GMP/MPFR.

maxGimeno:
> Can you explain what you are trying to achieve? Most of those headers are files that you should not include if you do not have GMP/MPFR.

Which leads to compilation errors, like here : https://github.com/CGAL/cgal/pull/5337/checks?check_run_id=1663949505#step:4:4805

gaschler:
Thank you for the quick response.
Indeed, I found a cleaner solution how I include CGAL that does not require this upstream patch.
So, I'll close this pull request.

PR #5340, opened 2021-01-08
Clang warnings fixed

### LINKED ISSUE #5332: -Wrange-loop-analysis
## Issue Details

The [clang testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-34/Optimal_bounding_box/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz) warns for

`for(const  auto gnu : gnats){....} `

This seems a reasonable warning as the variable `gnu`  will get re-assigned values in the loop.
The warning suggests to write `const auto &`.     We can also drop the `const`, if it  is a small object.    

This should be fixed in the whole column of the testsuite.

## Environment

* Compiler: clang
* Specific flags used (if any):   probably due to the `-wextra`, but maybe also with `-wall`
* CGAL version: master

danston:
This PR fixes multiple warnings generated by clang in the test suite when using the options `-wall` and `-wextra`. Mostly these are the ones, which come for the range loop analysis that is `-Wrange-loop-analysis`. This PR is supposed to fix the warnings from the issue #5332.

## Release Management

* Affected package(s): `Arrangement_on_surface_2`, `Classification`, `HalfedgeDS`, `Optimal_bounding_box`, `Point_set_processing_3`, `Polyhedron`, `Spatial_searching`, `Surface_mesh_approximation`, `Surface_mesh_simplification`, and `Tetrahedral_remeshing`.
* Issue(s) solved (if any): fix #5332


lrineau (approved):
I am not found of adding references when the type is trivial to copy, but the performance impact is certainly undetectable.

danston:
I agree, indeed, copying simple types might be slightly faster than using the reference. But the compiler warning is still there. Another option is to turn off the `-wextra` flag in order to avoid these types of warnings.

maxGimeno:
Please merge cgal/master in the branch and fix the conflicts

maxGimeno:

PR #5299, opened 2020-12-22
CI: Add Checks in GitHub Actions

maxGimeno:
## Summary of Changes

Add a check for test_merge_of_branch and dependencies.

PR #5343, opened 2021-01-11
Deactivate geomview_demo

maxGimeno:
## Summary of Changes

Deactivate completely the Geomview demo.
## Release Management

* Affected package(s):Polyhedron_IO

### DIFF
--- a/Polyhedron_IO/demo/Polyhedron_IO/geomview_demo.cpp
+++ b/Polyhedron_IO/demo/Polyhedron_IO/geomview_demo.cpp
@@ -12,12 +12,12 @@
 
 #include <CGAL/basic.h>
 
-#ifdef  _MSC_VER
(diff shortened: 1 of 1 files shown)


sloriot:
What's the point of the PR? Shouldn't we completely remove Geomview support?

maxGimeno:
We did not really discuss it with anyone, so I figured it would be easier to put it back this way if someone disagrees one day.

lrineau:
> We did not really discuss it with anyone, so I figured it would be easier to put it back this way if someone disagrees one day.

Since we use Git, I have always said that old code should be removed instead of being commented out.

Here that is the same: the old code should be removed, instead of "deactivated".

lrineau:
Should be rebased on 5.1, because of #5299 that modified `CGAL/config.h`.

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-45.shtml

PR #5358, opened 2021-01-15
Fix macro name to be consistent with find_package (5.1)

sloriot:


Installation/cmake/modules/CGAL_Eigen3_support.cmake
```diff
@@ -0,0 +1,9 @@
+if(EIGEN3_FOUND AND NOT TARGET CGAL::Eigen_support)
+  if(NOT TARGET Threads::Threads)
+    find_package(Threads REQUIRED)
+  endif()
+  add_library(CGAL::Eigen_support INTERFACE IMPORTED)
```
lrineau, line 5: Should not the target be renamed as well?

PR #5360, opened 2021-01-15
Fix macro name to be consistent with find_package

sloriot:
master version of #5359


maxGimeno:

Installation/cmake/modules/CGAL_Eigen3_support.cmake
```diff
@@ -0,0 +1,9 @@
+if(EIGEN3_FOUND AND NOT TARGET CGAL::Eigen3_support)
+  if(NOT TARGET Threads::Threads)
+    find_package(Threads REQUIRED)
+  endif()
```
lrineau, line 4: @sloriot You search for the CMake support `Threads` in a file about Eigen. Was that a copy-paste mistake, or is there a need for that?

sloriot: AFAIR, I only did a copy of the existing file so I did not add anything. All come from 41519e30104f

PR #5366, opened 2021-01-18
Move non-documented experimental function in experimental namespace

sloriot:
cc @sgiraudot

### DIFF
--- a/BGL/include/CGAL/boost/graph/selection.h
+++ b/BGL/include/CGAL/boost/graph/selection.h
@@ -561,6 +561,9 @@ regularize_face_selection_borders(
 /// \endcond
 
 /// \cond SKIP_IN_MANUAL
+
+namespace experimental {
+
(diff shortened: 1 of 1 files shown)


sloriot:
I had an issue if the documented version is called with `float` instead of `double` as it tries to call the non-documented version instead of casting to double.

lrineau:
The undocumented version is the following:
```c++
// Variant of regularization without graphcut but with brut-force
// local expansions. Can be interesting in some cases but too
// experimental/messy so far to be officially integrated.
template <class FaceGraph, class IsSelectedMap, class VertexPointMap>
void
regularize_face_selection_borders(
  FaceGraph& fg,
  IsSelectedMap is_selected,
  VertexPointMap vertex_point_map)
```

As it does not do the same thing as the other overloads, it should actually have a different name.

maxGimeno:

Issue #5368, opened 2021-01-18
problems with Point_set_3 property maps


raphaelsulzer:
## Issue Details

I'm trying to use the Point_set_3 property maps but run into many issues as described in the last three lines in the code below.
What am I doing wrong?

## Source Code

```
    #include <CGAL/Point_set_3.h>
    typedef CGAL::Point_set_3<Point> Point_set;
    typedef std::array<unsigned char, 3> Color;
    typedef Point_set::Property_map<Color> Color_map;

    data.point_set.reserve(data.smesh.number_of_vertices());
    Color white = {{ 255, 255, 255 }};
    Color_map color;
    bool created;
    boost::tie(color, created) = data.point_set.add_property_map<Color>("color", white);
    cout << "color point map for point set created: " << created << endl;  // outputs true

    /// add some points to the point_set ///

   int j = 0;
    for(Point_set::const_iterator pit = data.point_set.begin(); pit !=data.point_set.end(); pit++){
        int c = get(index_map, j++);
        if(c == -1)
            continue; // leave unassigned points white
        color[*pit][0]=(unsigned char)color_palette[c%7].r();
        color[*pit][1]=(unsigned char)color_palette[c%7].g();
        color[*pit][2]=(unsigned char)color_palette[c%7].b();
    }

//// This works fine
//    Point_set::iterator it = data.point_set.begin();
//    while (it != data.point_set.end())
//      {
//        cout << (int)color[*it][0] <<  " " << (int)color[*it][1] << " " << (int)color[*it][2] << endl;
//        it++;
//      }

//// Here nothing works
    write_ply_point_set(outstream,data.point_set); // writes the points but doesn't write colors
    Color_map color = data.point_set.property_map("color").first; // no matching function error
    auto nt = data.point_set.properties_and_types(); // no matching function error in stl_pair.h
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Ubuntu 18.04
* Compiler: gcc 7.5.0
* Release or debug mode: Release
* CGAL version: 5.1.1
* Boost version: 1.6.5

sgiraudot:
[...]

PR #5369, opened 2021-01-18
[Point Set 3] Fix broken properties and types

sgiraudot:
## Summary of Changes

The function `Point_set_3::properties_and_types()` returned a pair of `std::string` and `std::type_info`, which lead to a compilation error as `std::type_info` is not copyable. This is fixed by using the C++11 [`std::type_index`](https://en.cppreference.com/w/cpp/types/type_index) class (which is specifically done for cases like this one).

## Release Management

* Affected package(s): Point Set 3
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/5368

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -17,6 +17,7 @@
 
 
 #include <stack>
+#include <typeindex>
 
 #include <CGAL/Surface_mesh/Properties.h>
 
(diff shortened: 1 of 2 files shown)


lrineau (approved):

lrineau:
1. You break the API...
2. But that is not important, because the current API cannot be compiled! and thus is unusable.

Am I right?

sgiraudot:
You are completely right.

I admit I took the freedom not do to a small feature for this, as it seems like a "trivial" fix (even though it does break the API, in practice it does realize - and fix - the intended API). My screw-up was that I never tested this in the first place…

maxGimeno:

Issue #5375, opened 2021-01-20
Mesh_3 compiles but crashes if compiled then run with `Parallel_tag` but without `CGAL_LINKED_WITH_TBB`


lrineau:
## Issue Details

See the result of examples/Mesh_3/ is that test results:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-I-42.shtml#Mesh_3_Examples

There are a lot of platforms were the example `mesh_implicit_domains.cpp` fails at runtime.

The reason is that, the tag `CGAL::Parallel_tag` was used for the concurrency tag of the triangulation **but** `CGAL_LINKED_WITH_TBB` was not defined. That did compile, but the code logic with wrong with that setting: the function `Refine_cells_3::remove_element_from_refinement_queue(Cell_handle, Parallel_tag)` is then implemented as a no-op, even if `CGAL_LINKED_WITH_TBB` is not defined.

In my opinion that is completely wrong: a usage of `CGAL::Parallel_tag` when `CGAL_LINKED_WITH_TBB` should lead to a compilation error.

That was already discussed in the small feature [`Patch_to_concurrency_in_CGAL`](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Patch_to_concurrency_in_CGAL) in 2016 and 2019, but I think the authors and reviewers focused on `Parallel_if_available_tag` and forgot to discuss about `Parallel_tag`, or made the wrong decision.

lrineau:
@afabri What was your opinion about the following?

```c++
struct Sequential_tag {};
struct Parallel_tag : public Sequential_tag {};
```

The implicit cast of `Parallel_tag` into `Sequential` is an error, in my opinion. I think you saw that during the review, at https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Patch_to_concurrency_in_CGAL#Discussion_3 but the feature moved forward anyway.

afabri:
@MaelRL and @sloriot I would like to discuss this next Monday.

MaelRL:
> @afabri What was your opinion about the following?
> 
> ```c++
> struct Sequential_tag {};
> struct Parallel_tag : public Sequential_tag {};
> ```
> 
[...]

PR #5376, opened 2021-01-20
Fix examples/Mesh_3/mesh_implicit_domains.cpp

lrineau:
## Summary of Changes

The error was actually due to an mismatch between `CGAL_LINKED_WITH_TBB` and `CGAL_CONCURRENT_MESH_3`. See #5375.

For CGAL-5.3 we should clean that up!

## Release Management

* Affected package(s): Mesh_3
* License and copyright ownership: N/A
The error was actually due to an mismatch between `CGAL_LINKED_WITH_TBB`
and `CGAL_CONCURRENT_MESH_3`.

For CGAL-5.3 we should clean that up!The error was actually due to an mismatch between `CGAL_LINKED_WITH_TBB`
and `CGAL_CONCURRENT_MESH_3`.

For CGAL-5.3 we should clean that up!


maxGimeno:

PR #5388, opened 2021-01-25
CGAL_Ipelets: Fix after merge in master.

maxGimeno:
## Summary of Changes

Restore missing lines in CMakelists that propagated the Ipe version to use to simple_triangulation
## Release Management

* Affected package(s):CGAL_ipelets

### DIFF
--- a/CGAL_ipelets/demo/CGAL_ipelets/CMakeLists.txt
+++ b/CGAL_ipelets/demo/CGAL_ipelets/CMakeLists.txt
@@ -140,6 +140,10 @@ if(IPE_FOUND AND IPE_VERSION)
   add_to_cached_list(CGAL_EXECUTABLE_TARGETS simple_triangulation)
   target_link_libraries(simple_triangulation CGAL::Eigen3_support
                         ${IPE_LIBRARIES})
+  target_include_directories(simple_triangulation BEFORE PRIVATE ${IPE_INCLUDE_DIR})
+  if (WITH_IPE_7)
+      target_compile_definitions(simple_triangulation PRIVATE CGAL_USE_IPE_7)
+    endif()
   cgal_add_compilation_test(simple_triangulation)
 
 else()
--- a/CGAL_ipelets/demo/CGAL_ipelets/CMakeLists.txt
+++ b/CGAL_ipelets/demo/CGAL_ipelets/CMakeLists.txt
@@ -32,15 +32,11 @@ if(NOT TARGET CGAL::Eigen3_support)
   return()
 endif()
 
-find_package(IPE 6)
+find_package(IPE 7)
 
 if(IPE_FOUND)
-  if (${IPE_VERSION} EQUAL "7")
-    set(WITH_IPE_7 ON)
-  elseif(${IPE_VERSION} EQUAL "6")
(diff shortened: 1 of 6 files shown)


CGAL_ipelets/demo/CGAL_ipelets/CMakeLists.txt
```diff
@@ -140,6 +140,10 @@ if(IPE_FOUND AND IPE_VERSION)
   add_to_cached_list(CGAL_EXECUTABLE_TARGETS simple_triangulation)
   target_link_libraries(simple_triangulation CGAL::Eigen3_support
                         ${IPE_LIBRARIES})
+  target_include_directories(simple_triangulation BEFORE PRIVATE ${IPE_INCLUDE_DIR})
+  if (WITH_IPE_7)
+      target_compile_definitions(simple_triangulation PRIVATE CGAL_USE_IPE_7)
+    endif()
```
lrineau, line 146: Actually, maybe it is time to drop support for Ipe V6. Ipe version 7.0.10 (first version of Ipe 7.0.x in Debian) was published in Dec 2009. Is not it old enough?

@sloriot @maxGimeno

sloriot: Sure . Shall I do it or @maxGimeno will take care of it?

maxGimeno: I can do it

lrineau: Since we agree we can drop the support for Ipe version 6. I think Maxime can:
  - modify the C++ code that depends on `CGAL_USE_IPE_7`, so that it works now only with Ipe 7, without the need for thast macro `CGAL_USE_IPE_7`,
  - then simplify the CMake code (remove the detection of version, and remove the CMake code that defines `CGAL_USE_IPE_7`).

That will clean the C++ and CMake code of CGAL ipelets, and remove a maintenance charge.

Installation/cmake/modules/FindIPE.cmake
```diff
@@ -26,7 +26,7 @@ else()
 
   if(IPE_INCLUDE_DIR)
     file(READ "${IPE_INCLUDE_DIR}/ipebase.h" IPEBASE_H)
-    string(REGEX MATCH "IPELIB_VERSION[ ]*=[ ]*([67])([0-9][0-9])([0-9][0-9]);" FOUND_IPE_VERSION "${IPEBASE_H}")
+    string(REGEX MATCH "IPELIB_VERSION[ ]*=[ ]*([7])([0-9][0-9])([0-9][0-9]);" FOUND_IPE_VERSION "${IPEBASE_H}")
```
lrineau, line 29:
```suggestion
    string(REGEX MATCH "IPELIB_VERSION[ ]*=[ ]*([6789])([0-9][0-9])([0-9][0-9]);" FOUND_IPE_VERSION "${IPEBASE_H}")
```
For Ipe 6, and future Ipe 8, I think we still want to fill the variables `IPE_VERSION` and `IPE_FULL_VERSION`, even if we only support version 7, for the moment.

maxGimeno:

PR #5352, opened 2021-01-14
Fixes after the PR "CGAL I/O

lrineau:
## Summary of Changes

Fix errors made in PR https://github.com/CGAL/cgal/pull/4255. A lot of deprecated functions were no longer tested, and wrongly implemented.

## Release Management

* Affected package(s): BGL Point_set_3 Point_set_processing_3 Stream_support  Surface_mesh

PR #5398, opened 2021-01-27
Polyhedron_demo: Fix off_plugin

maxGimeno:
## Summary of Changes
Restore the precision 17 in the writing of a surface mesh in off.

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
@@ -1505,10 +1505,9 @@ bool
 Scene_surface_mesh_item::save(std::ostream& out) const
 {
   QApplication::setOverrideCursor(Qt::WaitCursor);
-  out.precision(17);
-    out << *(d->smesh_);
-    QApplication::restoreOverrideCursor();
-    return (bool) out;
+  CGAL::write_OFF(out, *(d->smesh_), CGAL::parameters::stream_precision(17));
+  QApplication::restoreOverrideCursor();
+  return (bool) out;
 }
 
 bool


sloriot:
Is it the only IO plugin that can have this issue?

maxGimeno:
That seems to be the only plugin that set precision before writing

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -1505,10 +1505,9 @@ bool
 Scene_surface_mesh_item::save(std::ostream& out) const
 {
   QApplication::setOverrideCursor(Qt::WaitCursor);
-  out.precision(17);
-    out << *(d->smesh_);
-    QApplication::restoreOverrideCursor();
-    return (bool) out;
+  CGAL::write_OFF(out, *(d->smesh_), CGAL::parameters::stream_precision(17));
+  QApplication::restoreOverrideCursor();
+  return (bool) out;
```
lrineau, line 1510: I find strange that `write_OFF` does not use the precision of the stream. Where is it documented?

sloriot: I think it is because of the default value of the named parameter. It is documented as 6

MaelRL: I guess that when the target is a stream, the default value of the NP could have been the precision of the stream.

sloriot: it indeed seems more natural (`CGAL::parameters::is_default_parameter(np)` should be used to detect that)

maxGimeno:
Replaced by #5352

PR #5399, opened 2021-01-27
Fix NewKernel_d/Eigen compatibility

mglisse:
## Summary of Changes

Since it isn't tested, there was some bitrot in the code to use Eigen vectors in NewKernel_d (instead of std::vector/std::array). Tested locally with
```diff
--- a/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_base.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_base.h
@@ -34,7 +34,7 @@
 namespace CGAL {
 
 template < typename FT_, typename Dim_,
-#if 1
+#if 0
          typename Vec_=Mix_vector<Array_vector<FT_, Dim_>,
                                   Vector_vector<FT_, Dim_>,
                                   FT_, Dim_>,
```

## Release Management

* Affected package(s): Number_types, NewKernel_d

### DIFF
--- a/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_base.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Cartesian_LA_base.h
@@ -34,16 +34,18 @@
 namespace CGAL {
 
 template < typename FT_, typename Dim_,
-#if 1
+#ifndef CGAL_NEWKERNEL_D_USE_EIGEN_VECTOR
+# if 1
          typename Vec_=Mix_vector<Array_vector<FT_, Dim_>,
                                   Vector_vector<FT_, Dim_>,
                                   FT_, Dim_>,
-#elif 0
+# elif 0
          typename Vec_=Array_vector<FT_, Dim_>,
-#elif 0
+# else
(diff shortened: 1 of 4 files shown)


lrineau:
Instead of `#if 0`/`#if 1`, it should be something like `#if CGAL_NEWKERNEL_D_USE_EIGEN` and you should add test that `#define CGAL_NEWKERNEL_D_USE_EIGEN 1`.

mglisse:
I don't think people should use this, if only because it isn't safe to use before C++17. If the macro is not documented, it seems ok... I'll add it.

lrineau:
> I don't think people should use this, if only because it isn't safe to use before C++17. If the macro is not documented, it seems ok... I'll add it.

Actually, I was suggesting the macro so that you can add a test for it, to avoid the inevitable bitrot.

mglisse:
> Actually, I was suggesting the macro so that you can add a test for it, to avoid the inevitable bitrot.

I felt it might be a bit wasteful to test that version all the time, but it isn't that bad (and disabled on C++14, so that won't affect everyone).

maxGimeno:

PR #5401, opened 2021-01-27
Fix the Documentation jobs in the CI 

maxGimeno:
The index.html was not committed after the doc building.

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -92,7 +92,7 @@ jobs:
             egrep -v " ${PR_NUMBER}\." index.html > tmp.html || true
             echo "<li><a href=https://cgal.github.io/${PR_NUMBER}/$ROUND/Manual/index.html>Manual for PR ${PR_NUMBER} ($ROUND).</a></li>" >> ./tmp.html
             mv tmp.html index.html
-            git add ${PR_NUMBER}/$ROUND && git add index.html && git commit -q --amend -m "base commit" && git push -q -f -u origin master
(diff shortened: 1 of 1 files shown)


.github/workflows/build_doc.yml
```diff
@@ -92,7 +92,7 @@ jobs:
             egrep -v " ${PR_NUMBER}\." index.html > tmp.html || true
             echo "<li><a href=https://cgal.github.io/${PR_NUMBER}/$ROUND/Manual/index.html>Manual for PR ${PR_NUMBER} ($ROUND).</a></li>" >> ./tmp.html
             mv tmp.html index.html
-            git add ${PR_NUMBER}/$ROUND && git commit -q --amend -m "base commit" && git push -q -f -u origin master
+            git add ${PR_NUMBER}/$ROUND && git add index.html && git commit -q --amend -m "base commit" && git push -q -f -u origin master
```
lrineau, line 95: Note that you can also use one `git add` command:
```suggestion
            git add ${PR_NUMBER}/$ROUND index.html && git commit -q --amend -m "base commit" && git push -q -f -u origin master
```

lrineau:
Is a merge in `master` sufficient? Do we have the same CI job for release branches?

maxGimeno:
You are not supposed to change the doc in a bug fix, so we decided to only use it for master. I thought being based on master was a condition for the job but it is not. Should I add it ?

lrineau:
> You are not supposed to change the doc in a bug fix, so we decided to only use it for master. I thought being based on master was a condition for the job but it is not. Should I add it ?

No, I understand the rational.

PR #5402, opened 2021-01-27
Epeck thread safety

### LINKED ISSUE #2685: Read-safe multithreading with Lazy_exact_nt
I can't seem to find any discussion in [the documentation](https://doc.cgal.org/latest/Number_types/classCGAL_1_1Lazy__exact__nt.html) the thread safety (or lack thereof) when using `CGAL:: Lazy_exact_nt`, e.g., via `CGAL::Epeck`. I'm worried that this type uses non-atomic reference counting implying that even _reading_ a variable could result in a race condition and subsequent memory violation.

In my specific case, I have a giant list of points that should be reduced to constrained Delaunay triangulations (in my simplified code below, I put every 4 points into a CDT). _Conceptually_, my problem is [embarassingly parallel](https://en.wikipedia.org/wiki/Embarrassingly_parallel) with respect to the "points". However, my points are constructed via arithmetic involving common points, so I understand that the lazy evaluation makes this a non-trivial dependency graph (simulated below by a simple addition of the first point to all others). Still, my problem is completely parallel with respect to the outputs: no CDT depends on any other one.

Here's a minimal-ish reproducible example:

```c++
#define SERIAL

#include <vector>
#include <CGAL/Vector_2.h>
#include <CGAL/Point_2.h>
#include <cmath>
#include <cassert>
#include <thread>
#include <vector>
#include <algorithm>

template<
  typename Index, 
  typename FunctionType>
inline bool parallel_for(
  const Index loop_size, 
  const FunctionType & func)
{
  assert(loop_size>=0);
  if(loop_size==0) return false;
  // Estimate number of threads in the pool
  // http://ideone.com/Z7zldb
  const static size_t sthc = std::thread::hardware_concurrency();
  const size_t nthreads = (sthc==0?8:sthc);
  // Size of a slice for the range functions
  Index slice = 
    std::max(
      (Index)std::round((loop_size+1)/static_cast<double>(nthreads)),(Index)1);
  // [Helper] Inner loop
  const auto & range = [&func](const Index k1, const Index k2, const size_t t)
  {
    for(Index k = k1; k < k2; k++) func(k);
  };
  // Create pool and launch jobs
  std::vector<std::thread> pool;
  pool.reserve(nthreads);
  // Inner range extents
  Index i1 = 0;
  Index i2 = std::min(0 + slice, loop_size);
  {
    size_t t = 0;
    for (; t+1 < nthreads && i1 < loop_size; ++t)
    {
      pool.emplace_back(range, i1, i2, t);
      i1 = i2;
      i2 = std::min(i2 + slice, loop_size);
    }
    if (i1 < loop_size)
    {
      pool.emplace_back(range, i1, loop_size, t);
    }
  }
  // Wait for jobs to finish
  for (std::thread &t : pool) if (t.joinable()) t.join();
  return true;
}
 
{

  typedef CGAL::Epeck Kernel;
  typedef CGAL::Triangulation_vertex_base_2<Kernel>  TVB_2;
  typedef CGAL::Constrained_triangulation_face_base_2<Kernel> CTFB_2;
  typedef CGAL::Triangulation_data_structure_2<TVB_2,CTFB_2> TDS_2;
  typedef CGAL::Exact_intersections_tag Itag;
  typedef CGAL::Constrained_Delaunay_triangulation_2<Kernel,TDS_2,Itag> CDT_2;
  typedef CGAL::Constrained_triangulation_plus_2<CDT_2> CDT_plus_2;

  std::vector< CGAL::Point_2<CGAL::Epeck> > points(100000);
  points[0] = CGAL::Point_2<CGAL::Epeck>(0,1);
  for(int i = 1;i<points.size();i++)
  {
    points[i] = points[(i-1)%5]+(CGAL::Vector_2<CGAL::Epeck>(i%7,i%11))/CGAL::Epeck::FT(3);
  }

  std::vector< CDT_plus_2 > cdts(points.size()/4);
#ifdef SERIAL
  for(int i = 1;i<cdts.size();i++)
#else
  parallel_for(cdts.size(),[&](int i)
#endif
  {
    CDT_plus_2 cdt;
    cdt.insert_constraint(points[4*i+0],points[4*i+1]);
    cdt.insert_constraint(points[4*i+2],points[4*i+3]);
  }
#ifndef SERIAL
  );
#endif

}
```

Use `#define SERIAL` to control whether the code executes in serial (should exit cleanly) or in parallel (usually causes `EXC_BAD_ACCESS` fault).

_FWIW @qnzhou expressed a thought that perhaps this reference counting is easy to fix using atomic integers_

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Mac
* Compiler: Apple LLVM version 9.0.0 (clang-900.0.39.2)
* Release or debug mode: either
* CGAL version: 1040911000
* Boost version: 106400
* Other libraries versions if used (Eigen, TBB, etc.): (std::thread-based parallel_for borrowed from libigl)

### LINKED ISSUE #4340: Concurrently readable lazy-evaluated number type in CGAL
Hello, does CGAL provide a working (maybe experimental?) lazy-evaluated number type that allows concurrent read operations? I asked this question in https://github.com/libigl/libigl/issues/1339 and it seems that I'm not the only one with this requirement.

The program I'm working on could benefit from parallelization (potentially a 2x speedup), but when using Kernel<CGAL::Lazy_exact_nt<ExactNumberType>> or CGAL::Epeck, I've noticed that concurrent read operations cause data races and crash the program.

I've tried using the number type mpq_class (as mentioned in https://github.com/CGAL/cgal/issues/2685), which does not use lazy evaluation nor reference counting. As a result, the program runs without data races, but much slower. 

I've also tried cloning Laurent Rineau's branch implementing Thread_safe_epeck (https://github.com/lrineau/cgal/tree/Filtered_kernel-Epeck_reentrance-GF). The code does compile and it seems that for simple inputs everything works perfectly. However, when I tried to process slightly bigger inputs (~3000 vertices), the program will _sometimes_ crash. Is this a known issue? Maybe I'm doing something wrong - I could elaborate in another issue if that would be appropriate. Anyway, I want to thank Laurent for his effort in implementing a thread-safe epeck, I very much appreciate the work.

mglisse:
## Summary of Changes

A preview of a prototype of some changes to make Epeck safer in multi-threaded applications, in case some people want to play with it, look at the code, or discuss the topic. Alternative to #648, it makes more sense as a separate PR than a comment at the end of that PR.
It is surprising that the hardest part is cleaning up. If we were ok with not pruning the Lazy trees after update_exact (or if we had garbage collection?), other strategies would become possible. Anyway, call_once seems rather efficient and once_flag is small, at least on linux (it is bigger on windows), and it avoids reinventing the (spinning) wheel.
I think the worst overhead is for operations on Lazy_exact_nt, but people shouldn't do a lot of computations in that type.
One thing I hope is that the overhead will prove small enough that we don't need a thread-unsafe version for use in multi-threaded applications that do not share objects.
This change means that copying a Handle(_for) is now expensive, so work to reduce copies (for instance for mpq_rational) would also benefit reference-counted types (Gmpq, Lazy).

Some possible future steps:
* benchmarks (single-thread), also check memory use. Careful that glibc-2.32+ may have very different perf than the rest.
* write a test (for validity) and a benchmark with parallelism and a significant amount of contention
* benchmark also on windows (and mac?), things like call_once could have wildly different behavior
* test on non-x86 hardware (memory ordering on x86 is special, and thread sanitizer only goes so far)
* discuss configuration: how should this be enabled? (CMake has the imported target Threads::Threads defined in FindThreads, this is already used in CGAL_TBB and CGAL_Eigen3 support files, see also THREADS_PREFER_PTHREAD_FLAG) Should there be a CMake option defined before find_package(CGAL) to enable it? Or a CGAL::Threads component that users need to request?
* fix the FIXMEs, do the TODOs, etc

## Release Management

* Affected package(s): Filtered_kernel, STL_Extension, Installation, Number_types
* Issue(s) solved (if any): fix #4340 fix #2685


Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -230,59 +223,337 @@ struct Depth_base {
...
 
-  AT& approx()
+  const ET & exact() const
   {
-      return at;
+#if defined(CGAL_HAS_THREADS) && defined(__gnu_linux__) && __GNUC__ < 11 && !defined(_REENTRANT)
+    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55394
+    // -pthread or -lpthread is just needed for linking, but I can't test that here...
+    // We could take the lack of _REENTRANT on that platform as forcing CGAL_HAS_NO_THREADS
+    static_assert(false, "Please pass -pthread to the compiler");
```
lrineau, line 325: Please make that assertion type-dependent, so that it is only evaluated at the instantiation of `Lazy_rep<AT,ET,E2A>::exact()`.

See for example https://devblogs.microsoft.com/oldnewthing/20200311-00/?p=103553.


We will have to work at the CMake level, to properly add `-pthread` when it is required.

mglisse: > Please make that assertion type-dependent,

That makes sense (for now the static_assert was just a note to myself so I wouldn't repeatedly hit mysterious errors), but I am not sure at all we want to keep this static_assert. We could undefine CGAL_HAS_THREADS on some platforms if it is missing, or some other approach.

lrineau: For a start, we should try a test platform without `CGAL_HAS_THREAD`, in CGAL test suite. I am not sure that still compiles everywhere.

sloriot:
About not pruning the tree, I think it would be fine if we have an advanced mode where we disable the automatic pruning and run it manually when we know there is not several threads that could access the memory. I'm not sure it helps or if it makes more complicated to support both.

About the garbage collection, have you already thought about a possible implementation? To me it seems that it's going to be almost as hard as pruning while doing an update exact.

The most  worrying part in your message is that `Handle` is now expensive to copy. In such a case, I think we should try to simply get rid of it and try to fix "automatically" all places where a reference should be taken.

mglisse:
> About not pruning the tree, I think it would be fine if we have an advanced mode where we disable the automatic pruning and run it manually when we know there is not several threads that could access the memory. I'm not sure it helps or if it makes more complicated to support both.

If you mean that the user would call `x.prune()` on many variables `x`, then it would make things simpler to implement, but would be rather painful for the user. If you mean the user would call `prune_all_lazy_dags()`, then we would need a way to find all the Lazy currently in existence. Maybe register them somewhere at creation, maybe always allocate them from some pools... Anyway not something trivial.
`call_once` seems to perform quite well on linux and allows to write relatively simple code, so we may not need alternate strategies.

> About the garbage collection, have you already thought about a possible implementation? To me it seems that it's going to be almost as hard as pruning while doing an update exact.

That was mostly a remark about what we could do if CGAL was written in a different language, not a serious suggestion.

> The most worrying part in your message is that `Handle` is now expensive to copy. In such a case, I think we should try to simply get rid of it and try to fix "automatically" all places where a reference should be taken.

Expensive is relative. Compared to adding 2 intervals for Lazy_exact_nt, it is expensive. Compared to computing the circumcenter of 4 points, I think it is negligible.
Doing a lot of small operations (`+-*/`) on Lazy_exact_nt looks like a bad idea anyway, computations should be done in blocks with a single Lazy_exact_nt at the end. (expression templates could help do that automatically for some operations, but I am not sure it is worth the trouble)
The cost should be lower on non-x86 hardware.
Doing fewer copies would be good, I've been advocating that for a while. It can be a lot of work, but can be done incrementally.
Moving from Gmpq to mpq_class/mpq_rational already eliminates some uses of Handle_for.

Do you have an idea on how you would get rid of Handle in Lazy? We need to store a DAG. For cleanup purposes, we need a way to know when a node is not referenced anymore. That usually means some form of reference counting (the thing that is expensive with threads). In other languages it could be garbage collection, which basically checks the whole memory for a pointer to that object.

afabri:
@mglisse what exactly should @maxGimeno benchmark?    
 @lrineau in a private discussion with @sloriot and me you told that you put a review of this PR on your todo list.

mglisse:
> @mglisse what exactly should @maxGimeno benchmark?

Take a program that uses Epeck and where the performance is important for you. Run it on windows once without the branch, and twice with the branch: once with CGAL_HAS_THREADS (the default?) and once with CGAL_HAS_NO_THREADS. I am not that familiar with visual studio, it may require some flags like /MD or /MT for the version with threads.

The expectation is that the branch with CGAL_HAS_THREADS should be a bit slower than master, but by less than 10%, and without using noticably more memory. And CGAL_HAS_NO_THREADS should bring back the same performance as master.

maxGimeno:
Ok, so I chose `test_p3rt3_as_p3t3`  as it uses Epeck and take more than 15s to execute, which makes it easy to catch significant differences of performance, and here are the results:
```
 - WITH_THREADS :  Test #1457: execution___of__test_p3rt3_as_p3t3 ................   Passed   15.41 sec
 - WITHOUT THREADS :  Test #1457: execution___of__test_p3rt3_as_p3t3 ................   Passed   15.47 sec
 - master:  Test #1457: execution___of__test_p3rt3_as_p3t3 ................   Passed   14.95 sec
 ```

**Comment by mglisse** [comment]
Thank you. 15s seems surprisingly long, it only takes about 3s here on linux, and I don't think a recent laptop is sufficient to make such a huge difference, but ok, windows is always slower... Maybe I should just say "great, the overhead of my code is small enough, let's merge" :smiley: 
To reduce the noise a bit, I commented out the Epick call and kept just Epeck for the following test on linux. Master and branch without threads are around 1.8s, with the branch a little bit slower (the variance is large though). Branch with threads is more like 2.4s, that's a lot! And it doesn't match very well with the windows results, but maybe those extra 12s of work hide the difference?
Profiling the code, the first thing I notice is that this test spends half of its time in dynamic_cast (and in particular strcmp, maybe if we made those class names shorter... For master, gcc is 1.8s, but clang is 2.6s with libstdc++ and 1.2s with libc++, so random). Probably the optimization in Lazy_kernel to extract a point from a weighted point? The running time goes down to 1.2 or 1.3s if I disable it (or less than 1s with master). The degenerate part of _side_of_power_sphere is costly, first because it copies weighted points, and then because the comparison used in std::sort keeps creating points from weighted points. compare_xyz is the first place I would look for possible improvements there. It could be split into Equal_3 and Less_xyz_3, and it could be done without constructing points.

**Comment by maxGimeno** [comment]
> 15s seems surprisingly long, it only takes about 3s here on linux

If I look at the master results in the testsuite I see [576.376 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Debug-64bits.gz) sec in debug, [32.398 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz) in Release, [233.67 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_cgaltester_x86-64_Darwin-18.5_Apple-clang-default_Release.gz) on a mac, [7.88487 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_lrineau_ArchLinux-clang-CXX17-Release.gz) on an Archlinux with cxx17, but also [216.886 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_lrineau_Debian-Stable.gz) on a Debian Stable and [181.793 sec](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-132/Periodic_3_triangulation_3/TestReport_lrineau_Ubuntu-Latest.gz) on a Ubuntu. I don't know what causes those values but 15 sec doesn't seem that long to me, after that :/

**Comment by mglisse** [comment]
The slow mac and linux runs are unoptimized builds. For optimized builds, 8s on archlinux is better than 32s on windows. Anyway, that's for #5724 to look at.

**Comment by maxGimeno** [comment]
Ok, that makes sense

File: Installation/include/CGAL/config.h [hunk 2]
```diff
@@ -590,6 +590,16 @@ using std::max;
...
 
+#ifndef CGAL_HAS_THREADS
+inline bool is_currently_single_threaded(){ return true; }
+#elif __has_include(<sys/single_threaded.h>)
+#include <sys/single_threaded.h>
+inline bool is_currently_single_threaded(){ return ::__libc_single_threaded; }
+#else
+/* This is the conservative default */
+inline bool is_currently_single_threaded(){ return false; }
+#endif
```
lrineau, line 601:
```suggestion
namespace CGAL {
#ifndef CGAL_HAS_THREADS
  inline bool is_currently_single_threaded(){ return true; }
#elif __has_include(<sys/single_threaded.h>)
#  include <sys/single_threaded.h>
  inline bool is_currently_single_threaded(){ return ::__libc_single_threaded; }
#else
/* This is the conservative default */
  inline bool is_currently_single_threaded(){ return false; }
#endif
}
```

STL_Extension/include/CGAL/Handle.h
```diff
@@ -50,54 +50,77 @@ class Handle
...
+        auto c = PTR->count.load(std::memory_order_relaxed);
+        if (c == 1)
+          delete PTR;
+        else
+          PTR->count.store(c - 1, std::memory_order_relaxed);
+      } else {
+      // TSAN does not support fences :-(
+#if !defined __SANITIZE_THREAD__ && !__has_feature(thread_sanitizer)
+        if (PTR->count.load(std::memory_order_relaxed) == 1
+            || PTR->count.fetch_sub(1, std::memory_order_release) == 1) {
```
lrineau, line 102: @mglisse Can you explain why the `count.load(..) ||`?

STL_Extension/include/CGAL/Handle.h
```diff
@@ -50,54 +50,77 @@ class Handle
...
-        if (--PTR->count==0)
-          delete PTR;
+        decref();
         PTR=0;
       }
     }
 
     int
-    refs()  const noexcept { return PTR->count; }
+    refs()  const noexcept { return PTR->count.load(std::memory_order_relaxed); }
```
lrineau, line 123: We should put a warning, like a `[[deprecated]]` on that function `refs()`, to avoid incorrect usages.

STL_Extension/include/CGAL/Handle.h
mglisse: the left part is much cheaper than the right, so I only do the right part if I have no choice.

Installation/include/CGAL/config.h
mglisse: Ah, right, the namespace. I'll avoid including a system header from inside namespace CGAL though :wink:

STL_Extension/include/CGAL/Handle.h
mglisse: I'm not sure I'd consider it deprecated, it just requires care in presence of threads, like a number of other things.
Arrangement_on_surface_2 seems to be the only current user, their use is probably not very safe, but if the rest of the package is not thread-safe, it may not matter much.

STL_Extension/include/CGAL/Handle.h
mglisse: In particular, if you create and destroy a Handle without ever copying it, this avoids any "true" atomic operation.

lrineau: In `Handle`, the functions `incref()` and `decref()` are summarized as:
```c++
    void incref() const noexcept
    {
        PTR->count.fetch_add(1, std::memory_order_relaxed);
    }
    void decref()
    {

        if (PTR->count.load(std::memory_order_relaxed) == 1
            || PTR->count.fetch_sub(1, std::memory_order_release) == 1) {
          std::atomic_thread_fence(std::memory_order_acquire);
          delete PTR;
        }
    }
```

I initially had doubt about that modification of the ref-counting (`PTR->count.load(std::memory_order_relaxed) == 1`), because that relaxed load was not synchronized with any other modification of the counter, beside being atomic.

I think the reasoning is that if the counter is one, then only one thread is supposed to have a handle to the object, and it there was a copy to another thread in action, that would require a synchronization.

So, eventually, I judge that this seems safe. I wonder why we do not see that pattern more often.

lrineau (approved):

lrineau:
I have refreshed the branch, with a CMake machinery that pulls `Threads::Threads` as a dependency of CGAL. That will probably fix the tests (at least that fixed them on ma machine).

lrineau (approved):
I have reviewed this pull-request carefully.

- It adds atomic reference counting in `Handle` and `Handle_for` (but not the other variants).
- In `Lazy`:
  - the exact object is always through an atomic pointer,
  - there are three strategies that are employed for the approximate object (the `Interval_nt`):
    1. the approximate object is also updated through the same atomic pointer, with a copy of the original version,
    2. the two `double` in `Interval_nt` are assumed to be atomic on x86_64: that is unsafe according to C++ memory model, but it works and is efficient,
    3. the two `double` of the approximate object are stored as `std::atomic<double>`.

By default on x86_64, the second (unsafe) option is used, for efficiency. I would like that we conduct benchmarks in order to verify how slower is the third option (two `std::atomic<double>`) compared to the unsafe option.

STL_Extension/include/CGAL/Handle.h
mglisse: > I wonder why we do not see that pattern more often.

I might have "invented" it when writing the code, at least I don't remember seeing this pattern before, but afterwards I noticed it in a few other places, someone even proposed an advanced version checking 2 counters at once for libstdc++'s shared_ptr this year (I don't think it has been merged yet though).

Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake
```diff
@@ -144,6 +144,12 @@ function(CGAL_setup_CGAL_dependencies target)
     if ( RUNNING_CGAL_AUTO_TEST )
       target_compile_options(${target} INTERFACE "-Wall")
     endif()
+    if(NOT CGAL_HAS_NO_THREADS)
```
mglisse, line 147: I think using CGAL_HAS_NO_THREADS as a CMake thing is new. Should there be an `else` that defines CGAL_HAS_NO_THREADS as a macro?
We are going to need some doc...

mglisse:
> By default on x86_64, the second (unsafe) option is used, for efficiency. I would like that we conduct benchmarks in order to verify how slower is the third option (two `std::atomic<double>`) compared to the unsafe option.

IIRC the test I was looking at when I wrote that code was building a 3d Delaunay triangulation with Epeck, where option 2 was used for Point_3. I did not implement a version of option 3 for Point_3 (I only wrote it for FT), and the indirection in option 1 made it slower.

mglisse:
I also wonder if we should allow to disable this without promising that there won't be any threads at all, for people who currently do independent computations in several threads and don't want to pay the overhead of the atomic counters, but also don't want CGAL to take advantage of some no-thread promise to use global (not thread-local) objects.

Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake
lrineau: I agree with you:
  - `CGAL_HAS_NO_THREADS` as a preprocessor macro is currently documented in the [section "Thread Safety"](https://doc.cgal.org/5.2.2/Manual/preliminaries.html#title1) in the manual. The eponymous CMake option should be documented.
  - I agree the CMake option `CGAL_HAS_NO_THREADS`, if true, should trigger a `-DCGAL_HAS_NO_THREADS` in compiler options.

I will fix that...

lrineau: I have tried to fix that with commit https://github.com/CGAL/cgal/pull/5402/commits/530616534ad9c68954067b4722d65d6fba217ec2.

elfprince13:
@mglisse - I've thought about implementing something like this for a while as it is pretty closely related to my dissertation work. Could you talk a bit about the architectural decisions? I'm specifically curious about:

> It is surprising that the hardest part is cleaning up. If we were ok with not pruning the Lazy trees after update_exact (or if we had garbage collection?), other strategies would become possible. Anyway, call_once seems rather efficient and once_flag is small, at least on linux (it is bigger on windows), and it avoids reinventing the (spinning) wheel.
> _..._
> I think the worst overhead is for operations on Lazy_exact_nt, but people shouldn't do a lot of computations in that type.
One thing I hope is that the overhead will prove small enough that we don't need a thread-unsafe version for use in multi-threaded applications that do not share objects.
> This change means that copying a Handle(_for) is now expensive, so work to reduce copies (for instance for mpq_rational) would also benefit reference-counted types (Gmpq, Lazy).

Current memory reclamation is actually quite tricky, so I am not surprised to hear that clean-up was the hardest part; however, I'm curious what changes made copying more expensive: is it just locking/synchronization overhead, or is it the result of doing some deep copy that was previously safe to have as a shallow copy?

mglisse:
> Current memory reclamation is actually quite tricky, so I am not surprised to hear that clean-up was the hardest part;

Actually, it could have been hard, but with the current approach it is rather simple. The dependencies are never accessed directly (well, with one exception) but only through update_exact, which is called through call_once. This behaves like a lock and there is no risk of concurrent access.
At some point I considered having update_exact compute the new value without protection (it is ok if 2 threads compute the same thing) and atomically put it in place, but that caused reclamation issues and wasn't faster.

The exception mentioned above is when we extract a point from a weighted point built from a point and a weight, but I simply disable reclamation for such nodes. Since the input and output have different types, there cannot be a chain of such nodes, so it cannot keep many extra nodes alive.

> however, I'm curious what changes made copying more expensive: is it just locking/synchronization overhead, or is it the result of doing some deep copy that was previously safe to have as a shallow copy?

Almost all the cost is the reference count that became atomic. Well, if we have a true parallel use and there is contention, other bottlenecks could appear, but since that's a new type of use, we cannot compare. For a single thread, the main extra cost is this atomic counter. It doesn't help that copying used to be almost free (increment an integer) so people were happy copying objects around a lot.

If you are not on x86_64 or are using a type other than FT/Point_2/Point_3, there is also a cost related to the new indirection to access the approximate object, which is noticeable for "predicate-only" operations like building a Delaunay triangulation. With the current approach, handling of approximate objects is actually by far the most complicated part. It can't be a plain atomic operation like reference counting, and it isn't hidden behind call_once like update_exact, it really is something that can be written and read simultaneously. Some things help: the approximate object is updated at most once, so we can just store both and not bother with reclamation. It is ok to read a mix of old and new values as long as you don't get a double that was only half updated.

Some musings:
* it would be nice to have a version of call_once that can use an existing pointer instead of requiring to store an extra once_flag. It would also remove the current binary incompatibility between code compiled with / without threads. But I am not holding my breath.
* a lot of atomic increments and decrements happen for objects that will never leave their original thread and are thus wasted, but I haven't yet found a nice way to take advantage of that without wasting a lot of memory or cpu cycles or making these objects much less convenient to use.

sloriot:
@mglisse do you have time to fix the conflict so that we can start testing this PR tonight?

mglisse:
Note that `std::system_error` is the error you get at runtime with libstdc++ if you fail to link with libpthread.

maxGimeno:
The `std::system_error` is still there on some platforms. 
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-11/Triangulation/TestReport_gimeno_ArchLinux-clang-Release.gz
> if you fail to link with libpthread.
I guess it means there is a problem in the CMakeLists of those failing tests ?

mglisse:
> I guess it means there is a problem in the CMakeLists of those failing tests ?

CMake prints the same thing for gcc or clang, but for gcc it adds -lpthread while for clang it doesn't, I don't know why.
THREADS_PREFER_PTHREAD_FLAG could be relevant or not.

lrineau:
> > I guess it means there is a problem in the CMakeLists of those failing tests ?
> 
> CMake prints the same thing for gcc or clang, but for gcc it adds -lpthread while for clang it doesn't, I don't know why.
> THREADS_PREFER_PTHREAD_FLAG could be relevant or not.

I'll have a look...

lrineau:
> > > I guess it means there is a problem in the CMakeLists of those failing tests ?
> > 
> > 
> > CMake prints the same thing for gcc or clang, but for gcc it adds -lpthread while for clang it doesn't, I don't know why.
> > THREADS_PREFER_PTHREAD_FLAG could be relevant or not.
> 
> I'll have a look...

My bad. See https://github.com/CGAL/cgal/pull/5402/commits/f9444fa85d8aa3dcd708fc85eed9bfa8064b4fea. The dependency to `Threads::Threads` was in a block of CMake code specific to GNU compilers. I have relocated it to a section of code for all compilers and platforms.

mglisse:
Ah, good. By the way, what do you think of THREADS_PREFER_PTHREAD_FLAG? The documentation says "Use of both the imported target as well as this switch is highly recommended for new code." Or is it up to the user to define it?

lrineau:
> Ah, good. By the way, what do you think of THREADS_PREFER_PTHREAD_FLAG? The documentation says "Use of both the imported target as well as this switch is highly recommended for new code." Or is it up to the user to define it?

I do not really care. I find strange that something "highly recommended" by CMake is OFF by default. As it is, that means that `-pthread` is not used, and only `-lpthreads` is used at link time. Personnally I find it cleaner that way. Why would we want to use `-pthreads` instead?

lrineau:
On the perf side, I know @maxGimeno did a lot of benchmarks. Was there a summary of that? Who had a look at the benchs results?

mglisse:
> I find strange that something "highly recommended" by CMake is OFF by default.

True... So let's ignore this variable.

> As it is, that means that `-pthread` is not used, and only `-lpthreads` is used at link time. Personnally I find it cleaner that way. Why would we want to use `-pthreads` instead?

First, `-pthread` applies at both compile and link time. At compile time, it can define some macros like `_REENTRANT` which instruct the C library to declare some things differently. At link time, a "magic" flag like `-pthread` may for instance help the linker sort libraries into a good order (IIRC some platforms require(d) linking with the thread library before the C library).
The compiler knows what it is doing with `-pthread`, letting CMake second-guess it seems more likely to hurt than help.

maxGimeno:
It is still here : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-12/Minkowski_sum_2/TestReport_lrineau_Ubuntu-Latest.gz

mglisse:
Looking at https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-12/Arrangement_on_surface_2/TestReport_lrineau_Ubuntu-Latest-CXX11.gz (why does this still have -std=c++11?) I notice a std::system_error in spite of
> /usr/bin/c++  -Wall -frounding-math -std=c++11 -DTEST_NT=6 -DTEST_KERNEL=0 -DTEST_GEOM_TRAITS=11 -DTEST_TOPOL_TRAITS=2 -I../../include   CMakeFiles/test_sgm.dir/test_sgm.cpp.o  -o test_sgm  **-lpthread** /usr/lib/x86_64-linux-gnu/libgmpxx.so /usr/lib/x86_64-linux-gnu/libmpfr.so /usr/lib/x86_64-linux-gnu/libgmp.so

For some reason the resulting binary does not have libpthread.so as NEEDED. If I move `-lpthread` to the end of the line, the error disappears. If I replace `-lpthread` with `-pthread`, it also works. This seems to argue that CMake's Threads support is broken if you don't define the macro we were discussing yesterday.

lrineau:
> Looking at [cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-12/Arrangement_on_surface_2/TestReport_lrineau_Ubuntu-Latest-CXX11.gz](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-12/Arrangement_on_surface_2/TestReport_lrineau_Ubuntu-Latest-CXX11.gz) (why does this still have -std=c++11?) I notice a std::system_error in spite of
> 
> > /usr/bin/c++  -Wall -frounding-math -std=c++11 -DTEST_NT=6 -DTEST_KERNEL=0 -DTEST_GEOM_TRAITS=11 -DTEST_TOPOL_TRAITS=2 -I../../include   CMakeFiles/test_sgm.dir/test_sgm.cpp.o  -o test_sgm  **-lpthread** /usr/lib/x86_64-linux-gnu/libgmpxx.so /usr/lib/x86_64-linux-gnu/libmpfr.so /usr/lib/x86_64-linux-gnu/libgmp.so
> 
> For some reason the resulting binary does not have libpthread.so as NEEDED. If I move `-lpthread` to the end of the line, the error disappears. If I replace `-lpthread` with `-pthread`, it also works. This seems to argue that CMake's Threads support is broken if you don't define the macro we were discussing yesterday.

We'll try with the option `THREADS_PREFER_PTHREAD_FLAG` set to `ON`.

maxGimeno:
tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-15.shtml

lrineau:
Benchs of this PR seems good. I will merge it.

Supranaturaler:
@sloriot plz offer the right url to me, thanks!

sloriot:
Should be fine now.

Supranaturaler:
> Should be fine now.

Compile error. You should remove line 217 in Lazy_kernel.h
image

Anyway, it worked in multithread test. Thank you for all this.

PR #5403, opened 2021-01-27
Add an example do the union of a set of meshes in parallel

sloriot:


lrineau:
Don't you want to put that example in the user manual as well? Or do you want it only in the code, so that you can refer to it in answers to user questions?

sloriot:
Yep second option.

maxGimeno:
Error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-51/Polygon_mesh_processing_Examples/TestReport_lrineau_Debian-stable-Release.gz

sloriot:
No idea where those includes came from. I amended by last commit.

maxGimeno:

PR #5407, opened 2021-01-28
Triangulation_2 : Document the `Triangulation_2_projection_traits_3` class

### LINKED ISSUE #2915: Document Triangulation_2_projection_traits_3
The traits is now around for a while and is very useful. It should be documented.
Also an example in the Triangulation_2 package would be nice to have.

@lrineau as the author of the file I assigned you to issue (but @maxGimeno might want to do it).

maxGimeno:
## Summary of Changes

Add doc to the class `T2_projection_traits_3`.

TODO:
- [x] add an example
- [x] handle retro compatibility

## Release Management

* Affected package(s):Triangulation_2
* Issue(s) solved (if any): fix #2915
* Small feature: https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Triangulation_2_projection_traits_3
* link to the doc : https://cgal.github.io/5407/v6/Kernel_23/classCGAL_1_1Projection__traits__3.html


Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
+namespace CGAL{
+
+/*!
+ \ingroup PkgTriangulation2TriangulationClasses
+
+  The class `Triangulation_2_projection_traits_3` is an extension of the `Projection_traits_xy_3`,
+  `Projection_traits_xz_3` and `Projection_traits_yz_3` traits classes,that allows the use
+  of 2D algorithms on the projections of 3D data on any plane.
```
MaelRL, line 8:
```suggestion
  The class `Triangulation_2_projection_traits_3` is an extension of the `Projection_traits_xy_3`,
  `Projection_traits_xz_3` and `Projection_traits_yz_3` traits classes, that enables
  using 2D algorithms on the projections of 3D data on any plane.
```

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
...
+
+  \cgalHeading{Parameters}
+  The template parameter `K` has to
+  be instantiated by a model of the `K` concept.
+  `Triangulation_2_projection_traits_3` uses types
+  and predicates defined in `K`.
```
MaelRL, line 14:
```suggestion
  \tparam K must be a model of `Kernel`.
```

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
...
+  /// The functors provided by this class are those listed in the
+  /// concepts. The functors operate on the 2D projection of their
+  /// arguments. They come with preconditions that projections of the
+  /// arguments are non-degenerate, eg. a line segment does not project
+  /// on a single point, two points do not project on the same point, etc.
+  /// The following functor is an addition to the concepts.
+  /// @{
+
+  /*!
+  \brief Projects a 3D point on a plane oriented by the traits.
```
MaelRL, line 50:
```suggestion
  \brief projects a 3D point on a plane oriented by the traits.
```

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
...
+
+  A construction object.
+  Provides the constructor:
+
+  `Projection_to_plan(const Point_2& plane_point, const Triangulation_2_projection_traits_3& tr)`
+
+  and the operator
```
MaelRL, line 57:
```suggestion
  and the operator:
```

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
...
+
+ `Point_2 operator()(const Point_2& point) const`
+
+  which returns the projection of a `point` to a plane passing through
+  the point 'plane_point' and with orthogonal vector `normal`, the vector given to the traits constructor.
```
MaelRL, line 62:
```suggestion
 `Point_2 operator()(const Point_2& point) const`

  which returns the projection of the (3D) point `point` onto the plane passing through
  the point 'plane_point' and with orthogonal vector `normal`, the vector given to the traits constructor.
```

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
MaelRL: I guess this is copied from the other traits class, so it could be done for the others too.

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -0,0 +1,79 @@
...
+  ///@{
+
+  //!
+  typedef typename K::Point_3     Point_2;
+  //!
+  typedef typename K::Segment_3   Segment_2;
+  //!
+  typedef typename K::Triangle_3  Triangle_2;
+  //!
+  typedef typename K::Line_3      Line_2;
```
MaelRL, line 35: Don't know if there is much worth to documenting those: they are not used anywhere in the doc of this class (except `Point_2`).

Triangulation_2/doc/Triangulation_2/CGAL/Triangulation_2_projection_traits_3.h
MaelRL: Related to what @sgiraudot what talking about about "what type of Kernel is expected from a class", it seems needed here to address what is provided by this class depending on `K` since constructions are used to project points. Something like:

> \tparam K must be a model of `Kernel`. 
>
> \note Internal constructions (projections) are used in the predicate and construction functors of this class. If `K` is a model of `Kernel` providing exact constructions or if `K` is a `CGAL::Filtered_kernel`, which is detected when `K::Has_filtered_predicates` exists and is `true` (such as for `CGAL::Exact_predicates_inexact_constructions_kernel`), this class automatically provides exact predicates.

? It is not the easiest to follow...

Triangulation_2/doc/Triangulation_2/PackageDescription.txt
```diff
@@ -103,6 +103,8 @@ are described in Chapter \ref PkgTDS2Ref "2D Triangulation Data Structure".
 
 - `CGAL::Triangulation_cw_ccw_2`
 
+- `CGAL::Triangulation_2_projection_traits_3<Kernel>`
```
MaelRL, line 106:
```suggestion
- `CGAL::Triangulation_2_projection_traits_3<K>`
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <iostream>
+#include <vector>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel K1;
+typedef CGAL::Triangulation_2_projection_traits_3<K1> K;
+
+typedef CGAL::Exact_predicates_tag                               Itag;
+typedef CGAL::Constrained_Delaunay_triangulation_2<K, CGAL::Default, Itag> CDT;
+typedef CDT::Point Point;
+typedef CDT::Edge  Edge;
```
MaelRL, line 14:
```suggestion
typedef CGAL::Exact_predicates_inexact_constructions_kernel                EPICK;
typedef CGAL::Triangulation_2_projection_traits_3<EPICK>                   K;

typedef CGAL::Exact_predicates_tag                                         Itag;
typedef CGAL::Constrained_Delaunay_triangulation_2<K, CGAL::Default, Itag> CDT;
typedef CDT::Point                                                         Point;
typedef CDT::Edge                                                          Edge;
```

MaelRL (changes requested):

afabri:
@maxGimeno can you please make the changes and build the documentation again.

Also, unless I missed that a reviewer has spotted it as well,   `Project_to_plan`  is  wrong.   It should be `Project_to_plane`, but maybe even that is wrong as we have the concept [`ConstructProjectedPoint_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1ConstructProjectedPoint__3.html).

sloriot:
I don't think it needs to be documented.

afabri:
> 
> 
> I don't think it needs to be documented.

Just to be sure that I understand you right:  There is no need to document it, as none of the 2D triangulation classes will ever call it, as there is no need for it to project points directly?

sloriot:
it is not used and there is actually not plane since only a normal is requested.

afabri:
I do not understand why it is in the triangulation package and not just in the Kernel as the other projection traits classes.

sloriot:
Projection traits are models of `PolygonTraits_2`, `ConvexHullTraits_2`, `TriangulationTraits_2`, `DelaunayTriangulationTraits_2`, `ConstrainedTriangulationTraits_2`, `ConvexHullTraits_2` and `DelaunayMeshTraits_2`. If this class is only model of triangulation concept it makes sense to be here. If it is model of more then it can be move in kernel, renamed + headers and template alias added for backward compatibility.

maxGimeno:
@sloriot @afabri Please take note that I changed the licenses to LGPL as no dependency were GPL.

afabri:
ok for the license change

afabri:
In the kernel it should not be called `Triangulation_projection_traits`.

sloriot:
@maxGimeno In case you forgot please to not forget to handle the backward compatibility header

afabri:
I do not understand why the class in the neted namespace `internal`  must be renamed.  That's the sole goal of the namespace.

maxGimeno:
Somewhere in the Polyhedron demo, both headers are used, and my gcc got confused, throwing all kind of errors about those. It was probably due to a misinterpretation of the namespaces, and I concluded that the safest fix was to rename the internal class. It would probably have been possible to track all occurences of both classes and force the whole namespaces on them, but it wouldn't protect future user's code from hitting the same problem under certain conditions.

afabri:
I do not agree. My guess is that at some  point you have a `using internal` that you shouldn't.

Kernel_23/include/CGAL/Projection_traits_3.h
```diff
@@ -0,0 +1,50 @@
...
+//
+//
+// Author(s)     : Laurent Rineau
+
+
+#ifndef CGAL_PROJECTION_TRAITS_3_H
+#define CGAL_PROJECTION_TRAITS_3_H
+
+
+#include <CGAL/internal/Triangulation_2_filtered_projection_traits_3.h>
```
MaelRL, line 18: I see https://github.com/CGAL/cgal/pull/5407#issuecomment-802726241, but I don't see why you would want to keep mentions of `Triangulation_2` in the kernel. Is this for bw-compatibility issues? Shouldn't it then just be a typedef in the triangulation package?

Nef_3/include/CGAL/boost/graph/convert_nef_polyhedron_to_polygon_mesh.h
```diff
@@ -211,7 +211,7 @@ struct Shell_polygons_visitor
 
     // cases where a cdt is needed
     typedef typename Nef_polyhedron::Kernel Kernel;
-    typedef Triangulation_2_projection_traits_3<Kernel>            P_traits;
+    typedef Projection_traits_3<Kernel>            P_traits;
     typedef Triangulation_vertex_base_with_info_2<std::size_t, P_traits> Vb;
```
MaelRL, line 215:
```suggestion
    typedef Projection_traits_3<Kernel>                                  P_traits;
    typedef Triangulation_vertex_base_with_info_2<std::size_t, P_traits> Vb;
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/Visitor.h
```diff
@@ -389,7 +389,7 @@ class Surface_intersection_visitor_for_corefinement{
    typedef Intersection_nodes<TriangleMesh, VertexPointMap1, VertexPointMap2,
             Predicates_on_constructions_needed>                          INodes;
    typedef typename INodes::Exact_kernel                                     EK;
-    typedef Triangulation_2_projection_traits_3<EK>                  CDT_traits;
+    typedef Projection_traits_3<EK>                  CDT_traits;
     typedef Triangulation_vertex_base_with_info_2<Node_id,CDT_traits>        Vb;
```
MaelRL, line 393:
```suggestion
    typedef Projection_traits_3<EK>                                          CDT_traits;
    typedef Triangulation_vertex_base_with_info_2<Node_id,CDT_traits>        Vb;
```

Triangulation_2/doc/Triangulation_2/Triangulation_2.txt
```diff
@@ -487,7 +487,9 @@ See Section \ref Section_2D_Triangulations_Delaunay for an example.
 `Projection_traits_xz_3<R>` to
 deal with projections on the
 `yz` plane and `xz`-plane,
-respectively.
+respectively, as well as the geometric traits class
+`Projection_traits_3` that deals with projections on an
+arbitrary plane, defined by its orthogonal vector.
```
MaelRL, line 493:
```suggestion
respectively, as well as the geometric traits class
`Projection_traits_3<R>` that deals with projections onto an
arbitrary plane defined by its orthogonal vector.

```

Triangulation_2/include/CGAL/Triangulation_2_projection_traits_3.h
```diff
@@ -16,12 +16,16 @@
 
 #include <CGAL/license/Triangulation_2.h>
 
+#include <CGAL/disable_warnings.h>
```
MaelRL, line 19: Why? And as far as I see you don't ever re-enable warnings.

+ Needs CGAL_NO_DEPRECATED_CODE

Triangulation_2/doc/Triangulation_2/Concepts/ConstrainedTriangulationTraits_2.h
```diff
@@ -50,11 +50,6 @@ typedef unspecified_type Intersect_2;
 /// When the constrained triangulation is instantiated with the intersection tag `CGAL::Exact_predicates_tag`, the used algorithm needs to be able to compare some distances between points and lines and the following types are further required.
 /// @{
 
-/*!
-A number type supporting the comparison operator `<`.
-*/
-typedef unspecified_type RT;
```
MaelRL, line 56: `RT` is being used in the concept functor `Compute_squared_distance_2` below.

Triangulation_2/doc/Triangulation_2/Concepts/TriangulationTraits_2.h
```diff
@@ -187,29 +187,6 @@ typedef unspecified_type Construct_circumcenter_2;
...
-*/
-TriangulationTraits_2(TriangulationTraits_2 gtr);
-
-/*!
-Assignment operator.
-*/
-TriangulationTraits_2 operator=(TriangulationTraits_2 gtr);
-
-/// @}
-
```
MaelRL, line 212: If your new model doesn't fit requirements, the solution is not to strip the concept from the requirements your model doesn't honor... These are real requirements that are not to be removed.

MaelRL (changes requested):

Triangulation_2/doc/Triangulation_2/Concepts/TriangulationTraits_2.h
maxGimeno: No, when testing the new concept, we realized with @sloriot that all those requirements were unnecessary, and he asked me to remove them.

MaelRL: Pretty sure that (for example):
```
  Triangulation_2(const Geom_traits& geom_traits=Geom_traits());
```
(https://github.com/CGAL/cgal/blob/master/Triangulation_2/include/CGAL/Triangulation_2.h#L244) requires the geometric traits to have a default constructor. But you are removing that from the concept.

sloriot: Default constructor is not needed, even if the triangulation have that default value)

MaelRL: How is it not needed when you have `Geom_traits()` literally written in the constructor? Just try to compile `T2 tr;` with this projection traits and you will see that it does not compile...

lrineau: I think it is only required for user using the default value. And thus that is not strictly required.

MaelRL: Exaggerating, `Side_of_oriented_circle_2` is also only required if I insert points so it is not strictly required either.

Concepts should describe the requirements to use the complete code, not some arbitrary subset based on some random user code.

sloriot: I agree in general but not for the default constructor. I prefer having a compilation error than having some runtime bugs because I forgot to pass the constructed traits. If the user uses a traits that is not default constructible then she shall not use the triangulation default constructor.

MaelRL: Again, your model being only a partial fit of the concept is not a valid reason to strip the concept. And it strips of more than the default constructor.

You can leave your model not being compatible with some triangulation functionalities if you wish (although I think it would be cleaner to add whatever is missing and explicitly disable and document), just do not modify the concept.

lrineau: Concepts should be kept minimal. If, as I hope, CGAL manages to implement C++20 concepts, then concepts issues will become compilation errors.

Taking an example from the STL, in C++03, the requirements for `T` in `std::vector<T>` was:
> `T` must meet the requirements of `CopyAssignable` and `CopyConstructible`.

... but `std::vector<T> vec(20);` would also require that `T()` is valid, to use the default-value of `value` in that constructor:
```c++
explicit vector( size_type count,
                 const T& value = T(),
                 const Allocator& alloc = Allocator());
```

<details>
<summary>Why C++03?</summary>

I specify C++03, because in C++11 or later, the requirement have become
> The requirements that are imposed on the elements depend on the actual operations performed on the container. Generally, it is required that element type meets the requirements of `Erasable`, but many member functions impose stricter requirements. 

</details>

sloriot: The traits class has been here for years and worked with the triangulation. The concept is just too restrictive requesting things that are not needed.

MaelRL: A complete code of
```
struct Gt { };
int main() { Triangulation_2<Gt> tr; }
```
can also be put in CGAL and work for years, will that prove that all the concept `TriangulationTraits_2` is requesting things that are not needed? Partial traits can work with a partial usage of the triangulation, just like your partial model of projection traits have worked because it only uses a part of the triangulation.

The concept is not too restrictive, it covers requirements of all the functionalities. Feel free to compile all of Triangulation_2 with your projection traits (starting as I have already said in https://github.com/CGAL/cgal/pull/5407#discussion_r639477557 with the default constructor) and you will see why these requirements exist.

sloriot: I don't understand what's your problem here. Sure it will not compile as the traits has no default constructor. But that the same thing with `std::vector<T>::resize(n)`. It will not compile if `T` is not default constructible.

sloriot: Agreement: Do not modify the concept and kept the doc of the projection traits as is, even if it is not entirely true.

Installation/CHANGES.md
```diff
@@ -1,5 +1,13 @@
 Release History
 ===============
+[Release 5.4](https://github.com/CGAL/cgal/releases/tag/v5.4)
+-----------
+
+Release date:  December 2021
+
+### [2D and 3D Linear Geometry Kernel](https://doc.cgal.org/5.3/Manual/packages.html#PkgKernel23)
+
+-   Added documentation for the class `Projection_traits_3`, that enables the use of 2D algorithms on the projections of 3D data into an arbitrary plane.
```
MaelRL, line 10:
```suggestion
-   Added documentation for the class `Projection_traits_3`, which enables the use of 2D algorithms on the projections of 3D data onto an arbitrary plane.
```

Kernel_23/doc/Kernel_23/CGAL/Projection_traits_3.h
```diff
@@ -0,0 +1,57 @@
+namespace CGAL{
+
+/*!
+\ingroup kernel_classes
+
+The class `Projection_traits_3` works similarly to the `Projection_traits_xy_3`,
+`Projection_traits_xz_3`, and `Projection_traits_yz_3` traits classes, by enabling
```
MaelRL, line 7:
```suggestion
`Projection_traits_xz_3`, and `Projection_traits_yz_3` traits classes, enabling
```

Kernel_23/doc/Kernel_23/CGAL/Projection_traits_3.h
```diff
@@ -0,0 +1,57 @@
...
+
+  ///@}
+
+
+  /// \name Functors
+  /// The functors provided by this class are those listed in the
+  /// concepts. The functors operate on the 2D projections of their
+  /// arguments. They come with preconditions that projections of the
+  /// arguments are non-degenerate, e.g. a line segment does not project
+  /// on a single point, two points do not project on the same point, etc.
```
MaelRL, line 42:
```suggestion
  /// on a single point, two points do not project onto the same point, etc.
```

Kernel_23/include/CGAL/Projection_traits_3.h
```diff
@@ -0,0 +1,50 @@
...
+  : public Filtered_projection_traits_3<Kernel>
+{
+public:
+  explicit
+  Projection_traits_3(const typename Kernel::Vector_3& n_)
+    : Filtered_projection_traits_3<Kernel>(n_)
+  {}
+};
+
+} // end namespace CGAL
```
MaelRL, line 48:
```suggestion


#ifndef CGAL_PROJECTION_TRAITS_3_H
#define CGAL_PROJECTION_TRAITS_3_H

#include <CGAL/internal/Filtered_projection_traits_3.h>

namespace CGAL {

// This declaration is needed to break the cyclic dependency.
template < class Filtered_kernel >
class Filtered_projection_traits_3;

template <class Kernel, bool Has_filtered_predicates=Kernel::Has_filtered_predicates>
class Projection_traits_3
  : public Projection_traits_base_3<Kernel>
{
public:
  explicit Projection_traits_3(const typename Kernel::Vector_3& n_)
    : Projection_traits_base_3<Kernel>(n_)
  {}
};

template <class Kernel>
class Projection_traits_3<Kernel, true>
  : public Filtered_projection_traits_3<Kernel>
{
public:
  explicit Projection_traits_3(const typename Kernel::Vector_3& n_)
    : Filtered_projection_traits_3<Kernel>(n_)
  {}
};

} // namespace CGAL
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Constrained_Delaunay_triangulation_2.h>
+#include <CGAL/Projection_traits_3.h>
+
+#include <iostream>
+#include <vector>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel                K1;
+typedef CGAL::Projection_traits_3<K1>                      K;
```
MaelRL, line 9:
```suggestion
typedef CGAL::Projection_traits_3<K>                                       GT;
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  K gt{ { 0, 1, 1} };
+  CDT cdt(gt);
+  for(int i = 0; i< 4; ++i)
+    cdt.insert(ps[i]);
+
+  for (int i = 1; i < 3; ++i)
+    cdt.insert_constraint( ps[i], ps[i+1]);
+
+  for (CDT::Face_handle f : cdt.all_face_handles())
```
MaelRL, line 33:
```suggestion
  for(int i = 1; i < 3; ++i)
    cdt.insert_constraint(ps[i], ps[i+1]);

  for(CDT::Face_handle f : cdt.all_face_handles())
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
...
+  {
+   for(int i=0; i<3; ++i)
+     std::cout<<f->vertex(i)->point()<<" ";
+   std::cout<<std::endl;
```
MaelRL, line 37:
```suggestion
     std::cout << f->vertex(i)->point() << " ";
   std::cout << std::endl;
```

Triangulation_2/test/Triangulation_2/test_deprecated_projection_traits.cpp
```diff
@@ -0,0 +1,34 @@
...
+  ps[2] = Point(-1, 3, -3);
+  ps[3] = Point(1,0.5,-0.5);
+
+  K gt{ { 0, 1, 1} };
+  CDT cdt(gt);
+  for(int i = 0; i< 4; ++i)
+    cdt.insert(ps[i]);
+
+  for (int i = 1; i < 3; ++i)
+    cdt.insert_constraint( ps[i], ps[i+1]);
```
MaelRL, line 33:
```suggestion
  for(int i = 1; i < 3; ++i)
    cdt.insert_constraint(ps[i], ps[i+1]);
```

Triangulation_2/test/Triangulation_2/test_deprecated_projection_traits.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Constrained_Delaunay_triangulation_2.h>
+#include <CGAL/Triangulation_2_projection_traits_3.h>
+
+#include <iostream>
+#include <vector>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel                K1;
+typedef CGAL::Triangulation_2_projection_traits_3<K1>                      K;
```
MaelRL, line 11:
```suggestion
typedef CGAL::Exact_predicates_inexact_constructions_kernel                K;
typedef CGAL::Triangulation_2_projection_traits_3<K>                       GT;
```

Triangulation_2/test/Triangulation_2/test_deprecated_projection_traits.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+typedef CGAL::Exact_predicates_tag                                         Itag;
+typedef CGAL::Constrained_Delaunay_triangulation_2<K, CGAL::Default, Itag> CDT;
```
MaelRL, line 14:
```suggestion
typedef CGAL::Constrained_Delaunay_triangulation_2<GT, CGAL::Default, Itag> CDT;
```

Triangulation_2/test/Triangulation_2/test_deprecated_projection_traits.cpp
```diff
@@ -0,0 +1,34 @@
...
+int main()
+{
+  //4 points on the diagonal plane of a cube
+  std::vector<Point> ps(4);
+  ps[0] = Point(0,0,0);
+  ps[1] = Point(3,1,-1);
+  ps[2] = Point(-1, 3, -3);
+  ps[3] = Point(1,0.5,-0.5);
+
+  K gt{ { 0, 1, 1} };
```
MaelRL, line 27:
```suggestion
  GT gt{ { 0, 1, 1} };
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Constrained_Delaunay_triangulation_2.h>
+#include <CGAL/Projection_traits_3.h>
+
+#include <iostream>
+#include <vector>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel                K1;
```
MaelRL, line 8:
```suggestion
typedef CGAL::Exact_predicates_inexact_constructions_kernel                K;
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
...
+typedef CGAL::Projection_traits_3<K1>                      K;
+
+typedef CGAL::Exact_predicates_tag                                         Itag;
+typedef CGAL::Constrained_Delaunay_triangulation_2<K, CGAL::Default, Itag> CDT;
```
MaelRL, line 12:
```suggestion
typedef CGAL::Constrained_Delaunay_triangulation_2<GT, CGAL::Default, Itag> CDT;
```

Triangulation_2/examples/Triangulation_2/projection_traits.cpp
```diff
@@ -0,0 +1,39 @@
...
+int main()
+{
+  //4 points on the diagonal plane of a cube
+  std::vector<Point> ps(4);
+  ps[0] = Point(0,0,0);
+  ps[1] = Point(3,1,-1);
+  ps[2] = Point(-1, 3, -3);
+  ps[3] = Point(1,0.5,-0.5);
+
+  K gt{ { 0, 1, 1} };
```
MaelRL, line 25:
```suggestion
  GT gt{ { 0, 1, 1} };
```

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-40.shtml

PR #5409, opened 2021-01-28
Fix the move-semantic of Triangulation_hierarchy_3

### LINKED ISSUE #5396: Crash in Windows version of simple 3D Delaunay triangulation program
I have the following simple program that creates a 3D Delaunay triangulation:

```
#include <CGAL/Delaunay_triangulation_3.h>

typedef CGAL::Delaunay_triangulation_3<K, CGAL::Fast_location> Delaunay;
typedef Delaunay::Point Point;

Delaunay Triangulate()
{
  Delaunay tri;
  for (int i=0; i<10; i++)
        tri.insert(Point(i+1, i+2, i+3));

  return tri;
}

int main()
{
  std::cout << "About to triangulate..." << std::endl;
  const auto tri = Triangulate();
  std::cout << "Triangulated successfully!" << std::endl;
  return 0;
}
```

When I compile and run this program on Windows in the 64-bit Debug configuration, it crashes while executing the `return tri` line. The debugger shows a read access violation during the CGAL `~Triangulation_hierarchy_3()` deconstructor. I am using CGAL 5.1.1 with vcpkg, and building the application with Visual Studio 19.

The program behaves fine on Linux compiled with gcc. Any ideas what is causing the problem on Windows?

lrineau:
### What is fixed by this pull-request
This PR started as a fix for the issue https://github.com/CGAL/cgal/issues/5396.

Then I have discovered that, in the PR https://github.com/CGAL/cgal/pull/4496, I had not implemented a test-suite for the move-semantic.

I have also discovered that a moved-from triangulation, since #4496, is not a valid CGAL triangulation: it has an empty TDS, and thus no infinite vertex, and as such has:
  - `dimension()==2`,
  - `number_of_vertices() == -1` (or `number_of_vertices() + 1 == 0`, to avoid a warning).

### The solution

- Add a testsuite for the move-semantic:
  - check that a move-constructed triangulation is equivalent to its source,
  - check the strange state of move-from triangulation (`dimension()==2`),
  - check that a moved-from triangulation:
    - can be destroyed,
    - can be assigned,
    - and can be restored to the default-constructed (empty) triangulation with `clear()`.
- Re-implement the internals of `Triangulation_hierarchy_3` so that a default-constructed hierarchy does not allocate: the levels from 1 to `maxlevel-1` are stored as data-members (in a `std::array`). That allows to satisfy the new testsuite.

### The TODO-list
- [x] Add documentation about moved-from triangulations (*Edit:* done, because we decided not to document anything)
- [x] Implement the same for 2D triangulations

## Release Management

* Affected package(s): Triangulation_3, Triangulation_2
* Issue(s) solved (if any): fix #5396
* License and copyright ownership: maintenance by GeometryFactory, copyright transfered to Inria Sophia-Antipolis.

### DIFF
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -114,6 +114,7 @@ public:
     , random(std::move(other.random))
   {
     hierarchy[0] = this;
+    other.hierarchy[0] = nullptr;
     for(int i=1; i<maxlevel; ++i) {
       hierarchy[i] = other.hierarchy[i];
       other.hierarchy[i] = nullptr;
@@ -503,7 +504,8 @@ void
 Triangulation_hierarchy_3<Tr>::
 clear()
 {
-  for(int i=0;i<maxlevel;++i)
+  if(hierarchy[0] == nullptr) return; // moved-from object
+  for(int i=1;i<maxlevel;++i)
     hierarchy[i]->clear();
 }
 
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -92,7 +92,26 @@ public:
 
 private:
 
+  template <std::size_t...Is>
+  constexpr std::array<Tr_Base, sizeof...(Is)>
+  make_array_of_triangulations(const Geom_traits& traits, std::index_sequence<Is...>)
+  {
+    return {{(static_cast<void>(Is), traits)...}};
+  }
+  template <std::size_t N>
+  constexpr std::array<Tr_Base, N> create_array_of_triangulation(const Geom_traits& traits)
+  {
+    return make_array_of_triangulations(traits, std::make_index_sequence<N>());
+  }
+
+  void init_hierarchy() {
+    hierarchy[0] = this;
+    for(int i=1; i<maxlevel; ++i)
+      hierarchy[i] = &hierarchy_triangulations[i-1];
+  }
+
   // here is the stack of triangulations which form the hierarchy
+  std::array<Tr_Base,maxlevel-1> hierarchy_triangulations;
   std::array<Tr_Base*,maxlevel> hierarchy;
   boost::rand48  random;
 
@@ -111,25 +130,20 @@ public:
   Triangulation_hierarchy_3(Triangulation_hierarchy_3&& other)
     noexcept( noexcept(Tr_Base(std::move(other))) )
     : Tr_Base(std::move(other))
+    , hierarchy_triangulations(std::move(other.hierarchy_triangulations))
     , random(std::move(other.random))
   {
-    hierarchy[0] = this;
-    other.hierarchy[0] = nullptr;
-    for(int i=1; i<maxlevel; ++i) {
-      hierarchy[i] = other.hierarchy[i];
-      other.hierarchy[i] = nullptr;
-    }
+    init_hierarchy();
   }
 
   template < typename InputIterator >
   Triangulation_hierarchy_3(InputIterator first, InputIterator last,
                             const Geom_traits& traits = Geom_traits())
     : Tr_Base(traits)
+    , hierarchy_triangulations(create_array_of_triangulation<maxlevel-1>(traits))
   {
-      hierarchy[0] = this;
-      for(int i=1; i<maxlevel; ++i)
-          hierarchy[i] = new Tr_Base(traits);
-      insert(first, last);
+    init_hierarchy();
+    insert(first, last);
   }
 
   Triangulation_hierarchy_3 & operator=(const Triangulation_hierarchy_3& tr)
@@ -143,21 +157,12 @@ public:
     noexcept( noexcept(Triangulation_hierarchy_3(std::move(other))) )
   {
     static_cast<Tr_Base&>(*this) = std::move(other);
-    hierarchy[0] = this;
-    for(int i=1; i<maxlevel; ++i) {
-      hierarchy[i] = other.hierarchy[i];
-      other.hierarchy[i] = nullptr;
-    }
+    hierarchy_triangulations = std::move(hierarchy_triangulations);
+    init_hierarchy();
     return *this;
   }
 
-  ~Triangulation_hierarchy_3()
-  {
-    clear();
-    for(int i=1; i<maxlevel; ++i) {
-      delete hierarchy[i];
-    }
-  };
+  ~Triangulation_hierarchy_3() = default;
 
   void swap(Triangulation_hierarchy_3 &tr)
   {
@@ -461,10 +466,9 @@ template <class Tr >
 Triangulation_hierarchy_3<Tr>::
 Triangulation_hierarchy_3(const Geom_traits& traits)
   : Tr_Base(traits)
+  , hierarchy_triangulations(create_array_of_triangulation<maxlevel-1>(traits))
 {
-  hierarchy[0] = this;
-  for(int i=1;i<maxlevel;++i)
-    hierarchy[i] = new Tr_Base(traits);
+  init_hierarchy();
 }
 
 // copy constructor duplicates vertices and cells
@@ -472,10 +476,9 @@ template <class Tr>
 Triangulation_hierarchy_3<Tr>::
 Triangulation_hierarchy_3(const Triangulation_hierarchy_3<Tr> &tr)
     : Tr_Base(tr)
+    , hierarchy_triangulations(tr.hierarchy_triangulations)
 {
-  hierarchy[0] = this;
-  for(int i=1; i<maxlevel; ++i)
-    hierarchy[i] = new Tr_Base(*tr.hierarchy[i]);
+  init_hierarchy();
 
   // up and down have been copied in straightforward way
   // compute a map at lower level
@@ -504,8 +507,7 @@ void
 Triangulation_hierarchy_3<Tr>::
 clear()
 {
-  if(hierarchy[0] == nullptr) return; // moved-from object
-  for(int i=1;i<maxlevel;++i)
+  for(int i=0;i<maxlevel;++i)
     hierarchy[i]->clear();
 }
 
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -157,8 +157,7 @@ public:
     noexcept( noexcept(Triangulation_hierarchy_3(std::move(other))) )
   {
     static_cast<Tr_Base&>(*this) = std::move(other);
-    hierarchy_triangulations = std::move(hierarchy_triangulations);
-    init_hierarchy();
+    hierarchy_triangulations = std::move(other.hierarchy_triangulations);
     return *this;
   }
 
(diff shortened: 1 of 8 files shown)


Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
```diff
@@ -142,21 +157,12 @@ class Triangulation_hierarchy_3
     noexcept( noexcept(Triangulation_hierarchy_3(std::move(other))) )
   {
     static_cast<Tr_Base&>(*this) = std::move(other);
-    hierarchy[0] = this;
-    for(int i=1; i<maxlevel; ++i) {
-      hierarchy[i] = other.hierarchy[i];
-      other.hierarchy[i] = nullptr;
-    }
+    hierarchy_triangulations = std::move(hierarchy_triangulations);
```
MaelRL, line 160: `std::move(other.hierarchy_triangulations)` 
and
`other.init_hierarchy()` ?

lrineau:
@MaelRL @mglisse @GilesBathgate:
Before I implement the same solution for 2D triangulations (in particular for  `Triangulation_hierarchy_2`), can you please comment on the patch of this pull-request?

mglisse:
I am not familiar with that code and don't have anything interesting to add here. I guess the approach should work (after fixing the issues noticed by Mael). `hierarchy` is not really needed anymore, but it shouldn't really matter. I didn't remember that the constructor of Triangulation_3 from Traits was implicit, but indeed it is.

Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
lrineau: Oops, you are right, for `std::move(other.hierarchy_triangulations)`.

`this->init_hierarchy()` and `other.init_hierarchy()` are actually useless, because they were initialized correctly at the construction of the two objects, and the move operations do not change the pointers to sub-objects.

lrineau:
> `hierarchy` is not really needed anymore, but it shouldn't really matter.

Indeed. I kept it by convenience, to avoid rewriting too much of the code.

lrineau:
The testsuite for the move-semantic still do not test correctly the hierarchy of triangulation: it only tests `is_valid()`, but nothing more. The bug pointed out by @MaelRL at https://github.com/CGAL/cgal/pull/5409#pullrequestreview-579803646 was untested, and I still do not know how to test that code path correctly.

GilesBathgate:
Unless I am misunderstanding you now have:

```c++
 std::array<Tr_Base,maxlevel-1> hierarchy_triangulations;
 std::array<Tr_Base*,maxlevel> hierarchy;
  ```
  `hierarchy` positions 1..maxlevel are always set to `hierarchy_triangulations`
and `hierarchy[0]` is always set to `this`.
could this not simply be separated out into for example:

```c++
 std::array<Tr_Base,maxlevel-1> hierarchy_triangulations;
 Tr_Base* hierarchy_root;
 ```
 
 I couldn't see any 'view' or for loop that needs access to `hierarchy` from 0..maxlevel

**Comment by lrineau** [comment]
Hi Giles,

> I couldn't see any 'view' or for loop that needs access to `hierarchy` from 0..maxlevel

You did not search well enough. :smile: There are a lot:
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L520-L522
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L536-L540
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L701-L707
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L718-L725
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L736-L744
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L773-L782

The weirdest is in `locate(..)`:
https://github.com/lrineau/cgal/blob/f14ab371ae320732c1aad9b02436767fb4f4987a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h#L818-L824

**Comment by GilesBathgate** [comment]
@lrineau Oh silly me I just skimmed over the commit instead of the original file.

Several of these seem to want `heirachy[0]` treated separately anyway though e.g:

```c++
 for (int l = 0; l < maxlevel; ++l) {
    Vertex_handle u = v->up();
    if(l) hierarchy[l]->move_if_no_collision(v, p);
    else ans =
           hierarchy[l]->move_if_no_collision_and_give_new_cells(v, p, fit);
    if(ans != v) return ans;
    if (u == Vertex_handle())
      break;
    v = u;
  }
  ```
  
  could be:
  
  ```c++
   Vertex_handle u = v->up();
   ans = hierarchy_root->move_if_no_collision_and_give_new_cells(v, p, fit);
   for (int l = 0; l < maxlevel-1; ++l) {
     u = v->up();
     hierarchy_triangulations[l]->move_if_no_collision(v, p);
    if(ans != v) return ans;
    if (u == Vertex_handle())
      break;
    v = u;
  }
  ```
    
  But probably these changes are out of scope for your PR..

**Comment by MaelRL** [comment]
> The testsuite for the move-semantic still do not test correctly the hierarchy of triangulation: it only tests `is_valid()`, but nothing more. The bug pointed out by @MaelRL at [#5409 (review)](https://github.com/CGAL/cgal/pull/5409#pullrequestreview-579803646) was untested, and I still do not know how to test that code path correctly.

Using `Triangulate()` like in the issue,
```
  Delaunay tri;
  tri = Triangulate(); 
```
will go through this operator (`Tr& operator=(Tr &&)`).

**Comment by lrineau** [comment]
> > The testsuite for the move-semantic still do not test correctly the hierarchy of triangulation: it only tests `is_valid()`, but nothing more. The bug pointed out by @MaelRL at [#5409 (review)](https://github.com/CGAL/cgal/pull/5409#pullrequestreview-579803646) was untested, and I still do not know how to test that code path correctly.
> 
> Using `Triangulate()` like in the issue,
> 
> ```
>   Delaunay tri;
>   tri = Triangulate(); 
> ```
> 
> will go through this operator (`Tr& operator=(Tr &&)`).

Right. I have added it in https://github.com/CGAL/cgal/pull/5409/commits/e6fe1c203191fe76c10acb6fda8dd881d880d1e9.

**Comment by lrineau** [comment]
I think the patch is now ready to be implemented in 2D as well.

**Comment by lrineau** [comment]
@mglisse 

I have added the following test in the testsuite:
```c++
  // Test from issue https://github.com/CGAL/cgal/issues/5396
  {
    auto Triangulate = []() -> Triangulation
    {
      Triangulation tri;
      for (int i=0; i<10; i++)
        tri.insert(Point(i+1, i+2, i+3));

      return std::move(tri); // the move prevents the NRVO
    };
    Triangulation t = Triangulate();
  }
```
and gcc-10 warns that way:
```
../Triangulation_3/test/Triangulation_3/include/CGAL/_test_cls_delaunay_3.h:1212:27: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]
 1212 |       return std::move(tri); // the move prevents the NRVO
      |                           ^
```

Do you see a way to silence that warning, or trick the compiler sufficiently so that it cannot warn about what the code does?

mglisse:
Just `return tri;`? The compiler already moves automatically.

lrineau:
> Just `return tri;`? The compiler already moves automatically.

But I want to move, to test it, and prevent the copy-ellision/NRVO. With `return tri;`, the compiler is allowed to apply the NRVO instead of moving the newly create `tri`.

MaelRL:
Maybe adding some extra `{ }` like @mglisse proposed https://github.com/CGAL/cgal/issues/5396#issuecomment-768249689 ?

mglisse:
@MaelRL the extra `{}` only work with gcc, and only because gcc's NRVO sucks (hopefully it will be improved).

mglisse:
> But I want to move, to test it, and prevent the copy-ellision/NRVO.

`Triangulation tri2 = std::move(tri); return tri2;` ? Now there is explicitly a move construction...

GilesBathgate:
If nothing else works:

```c++
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wpessimizing-move"
#endif
return std::move(tri);
#ifdef __clang__
#pragma clang diagnostic pop
#endif
```

lrineau:
> > But I want to move, to test it, and prevent the copy-ellision/NRVO.
> 
> `Triangulation tri2 = std::move(tri); return tri2;` ? Now there is explicitly a move construction...

Oh, yes, of course! :smile: I was focused on preventing the NRVO, but all we want is to test an actual move-constructor!

lrineau:
@maxGimeno This PR is now ready to be tested. What is missing is a small documentation sentence about move-semantic in the manual of T_2 and T_3.

maxGimeno:
This PR breaks the whole MSVC 2015 column. 
I read that `Extended constexpr` and `Default member initializers for aggregates` are not supported by this compiler (source https://docs.microsoft.com/fr-fr/cpp/overview/visual-cpp-language-conformance?view=msvc-160).
Do we drop MSVC 2015 as it doesn't entirely support c++14 ?

lrineau:
> This PR breaks the whole MSVC 2015 column.
> I read that `Extended constexpr` and `Default member initializers for aggregates` are not supported by this compiler (source [docs.microsoft.com/fr-fr/cpp/overview/visual-cpp-language-conformance?view=msvc-160](https://docs.microsoft.com/fr-fr/cpp/overview/visual-cpp-language-conformance?view=msvc-160)).
> Do we drop MSVC 2015 as it doesn't entirely support c++14 ?

No, we decided that MSVC 2015, at the time of CGAL-5.0, was a compiler to support. That means that we can use C++14, but the intersection with what is supported by MSVC 2015.

lrineau:
> This PR breaks the whole MSVC 2015 column.
> I read that `Extended constexpr` and `Default member initializers for aggregates` are not supported by this compiler (source [docs.microsoft.com/fr-fr/cpp/overview/visual-cpp-language-conformance?view=msvc-160](https://docs.microsoft.com/fr-fr/cpp/overview/visual-cpp-language-conformance?view=msvc-160)).
> Do we drop MSVC 2015 as it doesn't entirely support c++14 ?

I think that is fixed in the last commit https://github.com/CGAL/cgal/pull/5409/commits/519870c4cbf6d04ee6931f0f8b8680940bd61e33.

maxGimeno:
https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-60.shtml
The 2015 column is now fully yellow and there is still some red squares

lrineau:
> [cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-60.shtml](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-60.shtml)
> The 2015 column is now fully yellow and there is still some red squares

The warning is fixed in 160118e7e9c. The compilation errors seems to be due to #4878.

maxGimeno:

PR #5417, opened 2021-02-01
Polyhedron demo: Fix qt5 include error 

maxGimeno:
## Summary of Changes

Add a dependency to CGAL_Qt5_moc_and_resources to try and force it to be compiled first. This is supposed to fix the demo compilation errors on magritte : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-52/Polyhedron_Demo/TestReport_magritte_arm64-apple_bigsur_clang1200-release-64bits.gz


Installation/test/Installation/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ else()
 endif()
 
 if(WITH_CGAL_Qt5)
-  find_package(Qt5 QUIET)
+  find_package(Qt5 QUIET COMPONENTS Core Widgets Xml OpenGL Gui)
```
lrineau, line 80: I wonder why we need so maybe components of Qt5. Should not `CGAL::CGAL_Qt5` take care of that by itself?

maxGimeno: Core, Widget, OpenGL and Gui are quite obvious, and Xml is for QDomElement that is required in some of the polyhedron demo. I think that specific test could do fine with only Core. I don't know if we can use CGAL_Qt5 without finding Qt5 though.

maxGimeno:

PR #5420, opened 2021-02-02
Kernel_d:   Deal with deprecated result_type

afabri:
Fix for the [`Kernel_d`](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-52/Kernel_d/TestReport_Blake_Windows_MSVCPreview-Release-64bits.gz)  testsuite with `/c++-latest` for VC++.

## Summary of Changes

Add the trivial functions in the namespace `CGAL::Kernel_d`.

## Release Management

* Affected package(s): Kernel_d

### DIFF
--- a/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
+++ b/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
@@ -37,7 +37,7 @@ namespace CGAL {
     struct equal_to {
       typedef bool result_type;
 
-      bool operator()(const T &lhs, const T &rhs) const
+      result_type operator()(const T &lhs, const T &rhs) const
       {
         return lhs == rhs;
       }
@@ -48,23 +48,13 @@ namespace CGAL {
     struct minus {
       typedef T result_type;
 
-      T operator()(const T &lhs, const T &rhs) const
+      result_type operator()(const T &lhs, const T &rhs) const
       {
         return lhs - rhs;
       }
 
     };
 
-    template <typename T>
-    struct multiplies {
-      typedef T result_type;
-
-      T operator()(const T &lhs, const T &rhs) const
-      {
-        return lhs * rhs;
-      }
-
-    };
   } // namespace Kernel_d
 
   template < typename Point_, typename Functor_ >
@@ -314,7 +304,7 @@ namespace CGAL {
               Iter;
       Iter b(ptr()->upper, ptr()->lower, Begin());
       Iter e(ptr()->upper, ptr()->lower, Cartesian_end());
-      return std::accumulate(b, e, RT(1), Kernel_d::multiplies<RT>());
+      return std::accumulate(b, e, RT(1), std::multiplies<RT>());
     }
 
     RT volume_denominator() const
--- a/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
+++ b/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
@@ -46,7 +46,7 @@ namespace CGAL {
 
     template <typename T>
     struct minus {
-      typedef bool result_type;
+      typedef T result_type;
 
       T operator()(const T &lhs, const T &rhs) const
       {
@@ -57,7 +57,7 @@ namespace CGAL {
 
     template <typename T>
     struct multiplies {
-      typedef bool result_type;
+      typedef T result_type;
 
       T operator()(const T &lhs, const T &rhs) const
       {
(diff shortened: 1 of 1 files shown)


lrineau:
Moving away from using `result_type` seems preferable. We now have `decltype` since C++11, and `result_type` was deprecated (in 2017) and removed (in 2020) because it is an artifact of the past.

afabri:
It is not documented and an implementation detail.

lrineau:
> It is not documented and an implementation detail.

That is a maintenance burden anyway. For example, in your patch the `result_type` of `minus` and `multiplies` is `bool`. That is false, but may be left unseen by the testsuite.

Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
```diff
@@ -277,11 +310,11 @@ namespace CGAL {
 
     RT volume_nominator() const
     {
-      typedef typename CIRT::template Iterator<Point_d,std::minus<RT> >::type
+      typedef typename CIRT::template Iterator<Point_d,Kernel_d::minus<RT> >::type
               Iter;
       Iter b(ptr()->upper, ptr()->lower, Begin());
       Iter e(ptr()->upper, ptr()->lower, Cartesian_end());
-      return std::accumulate(b, e, RT(1), std::multiplies<RT>());
+      return std::accumulate(b, e, RT(1), Kernel_d::multiplies<RT>());
```
mglisse, line 317: `std::accumulate` doesn't work with `std::multiplies`?

lrineau:
I think the following patch is working:
```patch
diff --git a/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h b/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
index 3794119da67..94c877f5e18 100644
--- a/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
+++ b/Kernel_d/include/CGAL/Kernel_d/Iso_box_d.h
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <numeric>
 #include <cstddef>
+#include <iterator>
 
 namespace CGAL {
 
@@ -41,7 +42,12 @@ namespace CGAL {
     typedef typename Point::Cartesian_const_iterator  Iterator;
     typedef Cartesian_iterator<Point,Functor>         Self;
 
-    typedef typename Functor::result_type  value_type;
+    typedef typename std::iterator_traits<Iterator>::value_type
+                                                      Coordinate_type;
+
+    typedef decltype(
+        std::declval<Functor>()(std::declval<Coordinate_type>(),
+                                std::declval<Coordinate_type>())) value_type;
     typedef value_type&                    reference;
     typedef value_type*                    pointer;
     typedef std::ptrdiff_t                 difference_type;

```
At least the testsuite passed with it.

mglisse:
Assuming that Functor always returns a prvalue... (which seems to be the case)
Isn't the typedef for `reference` bogus? It should be the return type of `operator*`.

lrineau:
> Assuming that Functor always returns a prvalue... (which seems to be the case)

I should have added a `std::decay_t<..>`, right?

> Isn't the typedef for `reference` bogus? It should be the return type of `operator*`.

.. and then equal to `value_type`.

mglisse:
> I should have added a `std::decay_t<..>`, right?

Or remove_cvref, not sure which one would fit best. But that's only needed if there is a risk Functor might return a reference.

mglisse:
@MaelRL I already commented above. This looks ok. Non-blocking comments:
* it would have been simpler to use `*declval<Iterator>()` without introducing Coordinate_type
* it would have been safer to decay/remove_cvref value_type, but it doesn't seem necessary for the 3 uses in this file
* the `reference` typedef is bogus, but that's an existing problem not made worse by this patch.

afabri:
>     * it would have been simpler to use `*declval<Iterator>()` without introducing Coordinate_type

@lrineau shall we simplify?

lrineau:
> @MaelRL I already commented above. This looks ok. Non-blocking comments:
> 
>     * it would have been simpler to use `*declval<Iterator>()` without introducing Coordinate_type

That is right. I prefer your way. I have changed that in https://github.com/CGAL/cgal/pull/5420/commits/21cda32aef14a14d9b34d1bd3546131859398efa.

>     * it would have been safer to decay/remove_cvref value_type, but it doesn't seem necessary for the 3 uses in this file

I am a bit lazy, because `std::remove_cvref` is C++20, and `std::remove_cv_t<std::remove_reference_t<T>>` is too long. As you said, that is not necessary here.

>     * the `reference` typedef is bogus, but that's an existing problem not made worse by this patch.

Fixed in https://github.com/CGAL/cgal/pull/5420/commits/21cda32aef14a14d9b34d1bd3546131859398efa as well.

I have also implemented the same for `Homogeneous_iterator` in https://github.com/CGAL/cgal/pull/5420/commits/ab5517e1acb494cba28496a07437be9a7af2a39d. Tests from `ctest -L Kernel_d` pass on my machine.

lrineau:
Side note: @maxGimeno The Github actions that compile the polyhedron demo should be triggered for all pull-requests, and not only the ones targetting `master`.

maxGimeno:

PR #5421, opened 2021-02-02
BGL, PMP: Fix a warning

maxGimeno:
## Summary of Changes

Fix a warning in Apple clang

## Release Management

* Affected package(s):BGL, PMP


lrineau:
As I usually say, since the subject of the PR is "fix *something*", should not it be applied to release branches?

maxGimeno:
I don't know for normal test but most of test_bgl_read_write was re-writen in CGAL_IO, so it would probably require several branches to fix a really minor warning. I don't think it is worth.

maxGimeno:

PR #5424, opened 2021-02-03
Do not emit warning as hole filling can fail

sloriot:
Hole filling is allowed to failed, not display a CGAL warning


maxGimeno:

PR #5152, opened 2020-11-10
Replace usages of boost::bind

sloriot:
Replace the usage of boost::bind by lambdas

Can replace #4614

PR #5433, opened 2021-02-05
Remove last bind

sloriot:
Follow up of #5152


maxGimeno:

PR #5438, opened 2021-02-07
Calculate indices only when required.

### LINKED ISSUE #5379: ~70% Performance speedup for polygon_mesh_to_nef_3
## Issue Details

## Environment

* Compiler: clang++ 
* CGAL version: 5.0.2
* Boost version: 1.71.0

GilesBathgate:
## Summary of Changes

Calculate Object_index hash tables as needed i.e when the assertion fails, or when trace message is built.

## Release Management

### DIFF
--- a/Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h
+++ b/Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h
@@ -275,7 +275,6 @@ where |M| is a sphere map and |F| is a sface.}*/
 
 private:
 std::string get_svertex_index(SVertex_const_handle) const;
-std::string get_shalfedge_index(SHalfedge_const_handle e) const;
 
 }; // SM_const_decorator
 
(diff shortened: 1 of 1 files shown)


Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h
```diff
@@ -273,36 +273,51 @@ The macros are then |CGAL_forall_svertices(v,M)|,
...
       ++iso_vert_num;
     } else {
-      CGAL_assertion_msg(first_out_edge(v) != SHalfedge_const_handle(),
-      VI(v).c_str());
-      CGAL_NEF_TRACEN(v->point()<<" "<<EI(first_out_edge(v)));
-      CGAL_assertion_msg(first_out_edge(v)->source() == v,
-                         VI(v).c_str());
+      CGAL_assertion_code(SHalfedge_const_handle e=first_out_edge(v));
+      CGAL_assertion_msg(e != SHalfedge_const_handle(), get_svertex_index(v).c_str());
+      CGAL_NEF_TRACEN(v->point()<<" "<<get_shalfedge_index(e));
```
GilesBathgate, line 319: A better approach here would be to keep EI in local scope but enclosed by `#ifdef CGAL_NEF_TRACE`, which is implemented in #5246

afabri (approved):
`VI`  could also be in a `CGAL_assertion_code()`, but as `EI` cannot because it is in a `CGAL_NEF_TRACEN()`  I approve your changes.

GilesBathgate:
@afabri Yes I did consider wrapping the method `get_svertex_index` in a `CGAL_assertion_code` but with `CGAL_NO_ASERTIONS`  it would just be an unused method which is ok?

sloriot (approved):

sloriot:
We are only waiting for the testsuite to be green to merge this branch.

maxGimeno:

PR #5441, opened 2021-02-08
Remove CMAKE_CXX_STANDARD commands in cmake scripts

### LINKED ISSUE #5311: CMAKE_CXX_STANDARD in our CMakeLists
## Issue Details
In some of our CMakeLists, we use 
`set(CMAKE_CXX_STANDARD 14)` or `set(CMAKE_CXX_STANDARD 11)`

It seems to me like an error. Instead we should test if the required features are present and disable the configuration if not. 
A more concrete side-effect is the warning D9025 that appears when using a flag like /std:c++latest:

`Command line warning D9025 : overriding '/std:c++latest' with '/std:c++14'`

It seems to me that we should remove these cmake commands, especially since we require c++14 since 5.0 (IIRC), 
but as we are currently ignoring those warning in our testsuite, I fear there were a reason to keep them.
@CGAL/geometryfactory any opinion ?

maxGimeno:
## Summary of Changes
CGAL now requires c++14, so the CMAKE_CXX_STANDARD instructions in our CMake scripts have become useless. 
This PR removes them.

## Release Management

* Issue(s) solved (if any): fix #5311

### DIFF
--- a/Arrangement_on_surface_2/demo/Arrangement_on_surface_2/CMakeLists.txt
+++ b/Arrangement_on_surface_2/demo/Arrangement_on_surface_2/CMakeLists.txt
@@ -3,9 +3,6 @@
 cmake_minimum_required(VERSION 3.1...3.15)
 project(Arrangement_on_surface_2_Demo)
 
-set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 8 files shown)


lrineau (approved):

maxGimeno:

PR #4729, opened 2020-05-15
PMP: Add local versions of `stitch_borders` and `stitch_boundary_cycles`

MaelRL:
## Summary of Changes

- Factorize implementations of `stitch_boundary_cycle()` and `stitch_borders()` (and thus enhance `stitch_boundary_cycles` to do more than zipping)
- Add a local version of stitch borders (useful to speed up offset and self-intersecting clipping)

## Release Management

* Affected package(s): `Polygon_mesh_processing`
* Issue(s) solved (if any): -
* Feature/Small Feature (if any): [Local stitch_borders](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Local_sitch_borders) -- **pre-approved on 2020/07/31**
* License and copyright ownership: GeometryFactory

PR #5442, opened 2021-02-08
Re-introduce some kind of zipping as a first step to boundary cycle stitching

MaelRL:
## Summary of Changes

Zipping was removed when local stitching was introduced, and a common interface between stitching-within-a-cycle and normal-stitching was introduced (https://github.com/CGAL/cgal/pull/4729). However, that common interface does not handle any non-manifold configuration, even if it's a folded cycle with obvious stitching (consecutive equal edges).

So, some kind of zipping is re-added to handle these cases again.

## Release Management

* Issue(s) solved (if any): -


MaelRL:
Runtime error on apple platforms, e.g. https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-66/Polygon_mesh_processing/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz

maxGimeno:

Issue #5431, opened 2021-02-05
How to convert CGAL::Surface_mesh to CGAL::Polyhedron_3?


Micalson:
Hello, I want to convert `CGAL::Surface_mesh`  to `CGAL::Polyhedron_3`  for the next work.  I checked all the docs on the official website, but I didn't find the result I wanted.  Can someone help me?  Looking forward to your reply.

Regard,

afabri:
You can use [`copy_face_graph()`](https://doc.cgal.org/latest/BGL/group__PkgBGLHelperFct.html#ga2d3eb044bfe42eab566779ce32216c11).   Just out of curiosity, what is your scenario that makes you use them both, instead of one of the two consistently?

Micalson:
Let me explaint it.
I could run the following code for surface reconstruction.
```
#define CGAL_EIGEN3_ENABLED

#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
#include <CGAL/poisson_surface_reconstruction.h>
#include <CGAL/Surface_mesh.h>
#include <CGAL/IO/read_xyz_points.h>
#include <vector>
#include <fstream>
// Types
typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
typedef Kernel::Point_3 Point;
typedef Kernel::Vector_3 Vector;
typedef std::pair<Point, Vector> Pwn;
typedef CGAL::Surface_mesh<Point> Mesh;

int main(void)
{
	std::vector<Pwn> points;
	std::ifstream stream("data/kitten.xyz");
	if (!stream ||
		!CGAL::read_xyz_points(
			stream,
			std::back_inserter(points),
			CGAL::parameters::point_map(CGAL::First_of_pair_property_map<Pwn>()).
			normal_map(CGAL::Second_of_pair_property_map<Pwn>())))
	{
		std::cerr << "Error: cannot read file data/kitten.xyz" << std::endl;
		return EXIT_FAILURE;
	}
	Mesh output_mesh;
	double average_spacing = CGAL::compute_average_spacing<CGAL::Sequential_tag>
		(points, 6, CGAL::parameters::point_map(CGAL::First_of_pair_property_map<Pwn>()));
	if (CGAL::poisson_surface_reconstruction_delaunay
	(points.begin(), points.end(),
		CGAL::First_of_pair_property_map<Pwn>(),
		CGAL::Second_of_pair_property_map<Pwn>(),
		output_mesh, average_spacing))
	{
[...]

PR #5444, opened 2021-02-09
[Poisson Reconstruction] Compatibility with Simple Cartesian

sgiraudot:
## Summary of Changes

Poisson currently only compiles with Epick (see https://github.com/CGAL/cgal/issues/5431). This PR makes it compile with simple cartesian (both using double and float). However, the code rarely works at runtime (because it is based on Delaunay triangulation which are not robust without exact predicates), so I couldn't add a test for this.

I have tried to make it compile with Epeck, but it required many changes (including changes in Surface Mesher) and ended up breaking the Epick version, so I figured it was not worth the trouble.

## Release Management

* Affected package(s): Poisson
* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/5431

### DIFF
--- a/Number_types/include/CGAL/NT_converter.h
+++ b/Number_types/include/CGAL/NT_converter.h
@@ -61,12 +61,34 @@ struct NT_converter < NT1, double >
     }
 };
 
+template < class NT1 >
+struct NT_converter < NT1, float >
+  : public CGAL::cpp98::unary_function< NT1, float >
+{
+    float
+    operator()(const NT1 &a) const
+    {
+        return static_cast<float>(to_double(a));
+    }
+};
+
 template <>
 struct NT_converter < double, double >
   : public CGAL::cpp98::unary_function< double, double >
 {
     const double &
     operator()(const double &a) const
+    {
+        return a;
+    }
+};
+
+template <>
+struct NT_converter < float, float >
+  : public CGAL::cpp98::unary_function< float, float >
+{
+    const float &
+    operator()(const float &a) const
     {
         return a;
     }
--- a/Point_set_processing_3/include/CGAL/compute_average_spacing.h
+++ b/Point_set_processing_3/include/CGAL/compute_average_spacing.h
@@ -81,7 +81,7 @@ compute_average_spacing(const typename NeighborQuery::Kernel::Point_3& query, //
      boost::make_function_output_iterator
      ([&](const Point& p)
       {
-        sum_distances += std::sqrt(CGAL::squared_distance (query,p));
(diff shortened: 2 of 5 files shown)


Number_types/include/CGAL/NT_converter.h
```diff
@@ -72,6 +83,17 @@ struct NT_converter < double, double >
     }
 };
 
+template <>
+struct NT_converter < float, float >
```
lrineau, line 87: Is not there already `template <class NT1> struct NT_converter<NT1, NT1>`?!

sgiraudot: I just had a look: you are correct, but for some reason there is also a specialization for `double/double` (which is why I introduced the `float/float` one). Both could be removed, right?

lrineau: Yes I think so. To be checked.

sgiraudot:
Alright, so I removed the two apparently useless `NT_converter<NT,NT>` specializations (the existing `double/double` and the `float/float` I introduced with this PR). I run the testsuite of `Poisson` and `Number_types`, both succeed, so I think it's safe to say they were really useless. Let's see if it passes the whole testsuite.

sloriot:
```
/home/runner/work/cgal/cgal/Cartesian_kernel/include/CGAL/Cartesian_converter.h:406:15: error: ambiguous template instantiation for ‘struct CGAL::NT_converter<double, double>’
     Converter c;
               ^
```

sloriot:
@sgiraudot  this one is getting urgent as it is one of the last targetting 5.1.3

sgiraudot:
Alright, I removed the commit that removed the overloads of `NT_converter<double,double>` and `NT_converter<float,float>`. It turns out these overloads are indeed useful because otherwise, compilers can't choose between:
- `NT_converter<NT, NT>`
- `NT_converter<NT, double>` (which also exists)

So these overloads _are_ indeed useful.

sloriot:
`disable_if/enable_if` is also an option but not necessarily useful in this context

lrineau:
> Alright, I removed the commit that removed the overloads of `NT_converter<double,double>` and `NT_converter<float,float>`. It turns out these overloads are indeed useful because otherwise, compilers can't choose between:
> 
>     * `NT_converter<NT, NT>`
> 
>     * `NT_converter<NT, double>` (which also exists)
> 
> 
> So these overloads _are_ indeed useful.

Then re-added them with a useful comment. :smile:

maxGimeno:

PR #5460, opened 2021-02-15
Allow to use CGAL as a submodule of a bigger CMake project

lrineau:
## Summary of Changes

Allow to use CGAL as a submodule of a bigger CMake project.

The trick is to use
  - `CGAL_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`, and
  - `CGAL_BINARY_DIR` instead of `CMAKE_BINARY_DIR`.

See
  - https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html
  - https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_BINARY_DIR.html
> This can be useful when add_subdirectory() is used to connect several projects.

## Release Management

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,14 +12,14 @@ set(CGAL_BRANCH_BUILD
     ON
     CACHE INTERNAL "Create CGAL from a Git branch" FORCE)
 
-include(${CMAKE_SOURCE_DIR}/CGALConfigVersion.cmake)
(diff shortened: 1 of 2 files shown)


lrineau:
I have verified that the CTest tests from `test/Installation` pass, in the branch build. We will have to check the release layout.

maxGimeno:

PR #5464, opened 2021-02-15
benchmark Delaunay 2d with NewKernel_d

mglisse:
## Summary of Changes

I put the benchmark in test/ so it gets compiled.
This is at the same time a private demo without type wrappers so that Point_2 is Eigen::Vector2d.

## Release Management

This has several other PRs merged in.


maxGimeno:

PR #5465, opened 2021-02-16
Distance_3: Add Point_3/Tetrahedron_3

afabri:
## Summary of Changes

Add squared distance computation for point and tetrahedron

This replaces the PR #5304.  The proposed code here makes fewer tests.

## Release Management

* Affected package(s): Distance_3

### DIFF
--- a/Distance_3/include/CGAL/squared_distance_3_3.h
+++ b/Distance_3/include/CGAL/squared_distance_3_3.h
@@ -95,7 +95,7 @@ squared_distance(const typename K::Tetrahedron_3 & t,
   }
 
   if(on_bounded_side){
-    return K::FT(0);
+    return typename K::FT(0);
   }
   return dmin;
 }
--- a/Distance_3/include/CGAL/squared_distance_3_3.h
+++ b/Distance_3/include/CGAL/squared_distance_3_3.h
@@ -30,34 +30,27 @@ namespace internal {
 template <class K>
 inline
 typename K::FT
-squared_distance(const typename K::Tetrahedron_3 & tet,
+squared_distance(const typename K::Tetrahedron_3 & t,
                  const typename K::Point_3 & pt,
                  const K& k)
 {
-  CGAL_assertion(orientation(tet[0],tet[1],tet[2],tet[3]) == POSITIVE);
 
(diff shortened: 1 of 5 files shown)


Distance_3/include/CGAL/squared_distance_3_3.h
```diff
@@ -0,0 +1,85 @@
...
+typename K::FT
+squared_distance(const typename K::Tetrahedron_3 & tet,
+                 const typename K::Point_3 & pt,
+                 const K& k)
+{
+  CGAL_assertion(orientation(tet[0],tet[1],tet[2],tet[3]) == POSITIVE);
+
+  if(orientation(tet[0],tet[1],tet[2],pt) != POSITIVE){
+    const Triangle_3<K> tr = {tet[0], tet[1], tet[2]};
+    return CGAL::squared_distance(tr, pt);
```
mglisse, line 41: I don't think this is true. If I take a triangle, and 2 points tet[3] and pt close to each other, far from the triangle, and on opposite sides of the plane containing the triangle, this returns a large distance.

afabri: you are right. I am ashamed.

afabri:
@mglisse can you confirm that I can stop early on a triangle of the tet, when the point lies in the extruded triangle (and outside of the tet)?   This extruded triangle test is done in the triangle/point distance function.

mglisse:
Yes, that makes more sense.
squared_distance_to_triangle could use a similar approach to compute at most 2 distances to segments, and even then you may end up computing the distance to the same segment twice (for its 2 adjacent triangles), but optimizing will be up to someone else.

Distance_3/include/CGAL/squared_distance_3_3.h
```diff
@@ -0,0 +1,126 @@
...
+    if(! dmin_initialized){
+      dmin = d;
+      dmin_initialized = true;
+    }else{
+      dmin = std::min(d,dmin);
+    }
+  }
+
+  if(on_bounded_side){
+    return K::FT(0);
```
maxGimeno, line 98: missing typename.

maxGimeno:

PR #5453, opened 2021-02-12
Remove outdated URL

afabri:
## Release Management

* Issue(s) solved (if any): fix #5448

PR #5467, opened 2021-02-17
Remove outdated URL

afabri:
## Summary of Changes

Replace PR  #5453


lrineau:
This patch is against the branch `5.2.x-branch` but it cannot be merged in it because CGAL non-header-only is still supported in CGAL-5.2.x. It is only from CGAL-5.3 that header-only will be mandatory.

afabri:
Well, just removing the URL was correct.

maxGimeno:

PR #5468, opened 2021-02-17
Mesh_3: various small bug-fixes

sloriot:


Mesh_3/include/CGAL/internal/Mesh_3/helpers.h
```diff
@@ -45,7 +45,7 @@ void dump_graph_edges(std::ostream& out, const Graph& g)
   typedef typename boost::graph_traits<Graph>::edge_descriptor edge_descriptor;
 
   out.precision(17);
-  for(edge_descriptor e : edges(g))
+  for(edge_descriptor e : CGAL::make_range(edges(g)))
```
afabri, line 48: Can you explain why this needed.  And why only here.

sloriot: `g` is a graph class from boost so edges is a `std::pair`. It is not tested by the testsuite and only activated with debug macros.

afabri (approved):

maxGimeno:

PR #5470, opened 2021-02-17
Explicitly call simplify after marking bounded volumes

### LINKED ISSUE #5379: ~70% Performance speedup for polygon_mesh_to_nef_3

GilesBathgate:
## Summary of Changes

Applying the single responsibility principle marking bounded volumes should neither `compute_external` nor `simplify`. Then `simplify()` can be called explicitly where needed (and only once)

More details in https://github.com/CGAL/cgal/issues/5379#issuecomment-780733581

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -349,6 +349,12 @@ protected:
     es.build_external_structure();
   }
 
+ private:
+  void mark_bounded_volumes() {
+    CGAL::Mark_bounded_volumes<Nef_polyhedron_3> mbv(true);
+    delegate(mbv, /*compute_external*/ false, /*simplify*/ false);
+  }
+
  public:
   /*{\Mcreation 3}*/
 
@@ -602,9 +608,8 @@ protected:
     polyhedron_3_to_nef_3
       <CGAL::Polyhedron_3<T1,T2,T3,T4>, SNC_structure>( P, snc());
     build_external_structure();
+    mark_bounded_volumes();
     simplify();
-    CGAL::Mark_bounded_volumes<Nef_polyhedron_3> mbv(true);
-    delegate(mbv);
     set_snc(snc());
   }
 
@@ -616,9 +621,8 @@ protected:
     initialize_infibox_vertices(EMPTY);
     polygon_mesh_to_nef_3<PolygonMesh, SNC_structure>(const_cast<PolygonMesh&>(pm), snc());
     build_external_structure();
+    mark_bounded_volumes();
     simplify();
-    CGAL::Mark_bounded_volumes<Nef_polyhedron_3> mbv(true);
-    delegate(mbv);
     set_snc(snc());
   }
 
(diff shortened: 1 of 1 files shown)


afabri:
@GilesBathgate did you read the code to convince yourself that there is no need to call `simplify()` twice, because it is needed before and after marking of volumes?

GilesBathgate:
> because it is needed before and after marking of volumes.

@afabri if that's the case then this PR can be closed. What would be an example of geometry that needs simplifying both before and after marking bounded volumes?

afabri:
I don''t know yet, and that's why I asked if you had a closer look.

GilesBathgate:
> I don''t know yet, and that's why I asked if you had a closer look.

@afabri Ah I need to fix my grammar parser (I misunderstood your comment) :joy: 

So the `simplify()` code does seem to require marks applied which led me to believe that it needs to be called after mark_bounded_volumes, and the git history shows that `simply` was present at first by itself, then later `Mark_bounded_volumes.h` was added but not called directly from the constructor. Later it was added to the constructor, and my assumption was that the second simplify was unintentional because `delegate` has a default parameter `do_simplify=true` and this might have been overlooked. (especially since the overload delegate below has `do_simplify=false` ) 

Either way, I experimentally removed the duplicate calls and couldn't find any apparent difference when doing tests, aside from the performance gain (which might be biasing my assumptions). 

I suppose what I am saying really is that unless there are test cases that prove otherwise, I can't see why it would need to be called twice.

GilesBathgate:
Links to various points in git history:

1. constructor with single simplify()

https://github.com/CGAL/cgal/blame/f048a210f648d3ba690946f002648e3ef2c34eed/Nef_3/include/CGAL/Nef_polyhedron_3.h#L412

2. Commit that introduced Mark_bounded_volumes.h

https://github.com/CGAL/cgal/commit/ae4623ca2bed3ec94d881fae22fda0c2364883cf

3. Commit where `do_simplify` parameter was added so that it can be "faster due to effective filtering"

https://github.com/CGAL/cgal/commit/dd0855cd7dfa5a0c558e8ba683154dc4b36b9941

afabri:
Thanks for your analysis.   While we are at it:  I grepped for `delegate()` calls and did not see it called once with other than the default parameters.   And the `Mark_bounded_volumes` is always constructed with `true`.  We might simplify.

GilesBathgate:
@afabri Yes I find these boolean default parameters rather confusing. A comment at the call can be used to make casual reading easier e.g: `Mark_bounded_volumes mbv(/* Mark */ true);`. Personally, I prefer a parameterless version of the constructor.

https://github.com/CGAL/cgal/blob/74af3e7d33e7354e4090a1b17941a3f0c372ca8d/Nef_3/include/CGAL/Nef_3/Mark_bounded_volumes.h#L65

GilesBathgate:
> I grepped for `delegate()` calls and did not see it called once with other than the default parameters.

`delegate' is (or at least was) part of the public interface though: https://doc.cgal.org/4.7/Nef_3/Nef_3_2polygon_construction_8cpp-example.html

afabri (approved):
The proposed changes make the code more uniform and readable.

maxGimeno:

PR #5474, opened 2021-02-19
Triangulation_3 and Mesh_3 Timeouts for parallel tests fix

danston:
When running the test suite on macOS with the TBB support, parallel tests from `Triangulation_3` and `Mesh_3` packages have timeouts. It happens because we call ctest with the -j4 option, which means that 4 tests are run in parallel. It may happen that 3 of them are themselves parallel tests using TBB so out of let's say 6 cores we have only 2 left for these tests, basically 1 per test so these tests fail due to not enough power. If we are lucky and 1 test is parallel and 1 is not then we may get 2 cores per this parallel test so it becomes faster and passes.

In this PR, I force all parallel tests from these packages to run on their own using all available cores. It seems to work pretty well locally on my machine according to my time profiler, but should be tested for the whole test suite.

## Release Management

* Affected package(s): `Triangulation_3`, `Mesh_3`
* Issue(s) solved (if any): timeouts
* Feature/Small Feature (if any): not a feature


Mesh_3/test/Mesh_3/CMakeLists.txt
```diff
@@ -121,6 +121,21 @@ if ( CGAL_FOUND )
...
+      set_property(TEST execution___of__test_meshing_implicit_function PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_meshing_3D_image PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_meshing_3D_gray_image PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_meshing_unit_tetrahedron PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_meshing_polyhedron PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_meshing_polyhedral_complex PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_mesh_capsule_var_distance_bound PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_mesh_3_issue_1554 PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_mesh_polyhedral_domain_with_features_deprecated PROPERTY RUN_SERIAL 1)
+      set_property(TEST execution___of__test_mesh_cell_base_3 PROPERTY RUN_SERIAL 1)
```
lrineau, line 137: That is just a matter of style, and not something I ask you to change: you can specify multiple test names in one command:
https://cmake.org/cmake/help/latest/command/set_property.html
> `TEST`
> Scope may name zero or more existing tests.

danston: You are right! I missed that.

maxGimeno:
I think this should be rebased onto 5.1

maxGimeno:
I rebased the PR on 5.1. @danston can you please check I didn't break anything ?

danston:
Oups, I think, I did something really wrong...

danston:
Now, it should all be ok.

maxGimeno:

PR #5476, opened 2021-02-19
PMP: Remove needles&caps fixes

sloriot:
## Summary of Changes

Fix a number of issues with `PMP::remove_almost_degenerate_faces()`:

- `Edge::collapse_edge` ignored the `ecm`
- A needle ignored due to having a too long collapsable edge was not considered as a cap
- Fix not purging the edge being currently considered from future sets (might have been inserted from an adjacency)
- Fix over-purging sets when using `remove_face()` in the case of a cap on the border
- Don't purge sets if `get_best_oriented()` is going to reject the collapse anyway (everything should still be in the future sets)
- Don't swap if we're going to return anyway

## Release Management

* Issue(s) solved (if any): -


maxGimeno:

PR #5480, opened 2021-02-19
Kernel_23:  Rename internal function and functor

### LINKED ISSUE #5471: Lazy Intersect_point_3
## Issue Details

I noticed some new code in Lazy.h, introduced in #5090, which looks suspicious. Why is it ok to use a global LazyPointRep for the returned object? See the testcase below for an issue this causes. I didn't see any kind of documentation or comment, so maybe there are restrictions on how this functor may be used? The code has a generic-looking name (Lazy_construction_optional) but actually hardcodes that it can only be used for Intersect_point_3, so I wouldn't be that surprised to discover other restrictions.

## Source Code

```c++
typedef CGAL::Epeck K;
int main(){
  K::Plane_3 a(rand(),rand(),rand(),rand());
  K::Plane_3 b(rand(),rand(),rand(),rand());
  K::Plane_3 c(rand(),rand(),rand(),rand());
  K::Plane_3 d(rand(),rand(),rand(),rand());
  K::Intersect_point_3 inter;
  auto i1 = inter(a,b,c);
  auto i2 = inter(a,b,d);
}
```
> free(): invalid pointer

or phrased differently
> attempting free on address which was not malloc()-ed

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux 64
* Compiler: gcc-10
* CGAL version: master

afabri:
## Summary of Changes

Change the name of undocumented functor and function so that it is clear that it is an implementation detail.

## Release Management

* Affected package(s): Kernel, PMP
* Issue(s) solved (if any): fix #5471

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -874,7 +874,7 @@ struct Lazy_construction_bbox
 
 
 template <typename LK, typename AC, typename EC>
-struct Lazy_construction_optional
+struct Lazy_construction_optional_for_polygonal_envelope
 {
   static const bool Protection = true;
   typedef typename LK::Approximate_kernel AK;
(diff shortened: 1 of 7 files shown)


maxGimeno:

PR #5484, opened 2021-02-20
Migration to faster hash maps - Handle_hash_map

### LINKED ISSUE #5379: ~70% Performance speedup for polygon_mesh_to_nef_3

GilesBathgate:
# Summary of Changes

This pull request adds a new `internal::Handle_hash_map` class that provides a drop-in replacement for `Unique_hash_map` but is backed by the single header `robin_hood::unordered_node_map`  This provides an easy means to migrate via an adaptor/wrapper which has then been applied to the Nef_2/Nef_S2 and Nef_3 packages.

Benchmarks for the Handle_hash_map are provided here: https://github.com/GilesBathgate/hashmap-benchmarks 
The main benefit is construction speed since Unique_hash_map suffers from a hard-coded minimum size of 512 buckets, which impacts the creation of small hashmaps. However there are measurable performance improvements on insert and update operations too, and being a robin hood hashing scheme it does not have the requirement for unique keys.

Some of the benefits were also discussed in #4049 . Other backends were discussed and this wrapper provides an easy way to migrate to another choice too. 

## Release Management

* Affected package(s): Nef_2/Nef_S2/Nef_3
* Link to compiled documentation: Undocumented internal use only.
    MIT license for additional code from https://github.com/martinus/robin-hood-hashing Copyright (c) 2018-2020 Martin Ankerl <http://martin.ankerl.com> 
   The adaptor/wrapper is authored by me but I return the copyright to CGAL.

### DIFF
--- a/Hash_map/include/CGAL/Handle_hash_map.h
+++ b/Hash_map/include/CGAL/Handle_hash_map.h
@@ -76,6 +76,11 @@ public:
         insert(key_begin,key_end,values_begin);
     }
 
+    void reserve(size_type n)
+    {
+        Base::reserve(n);
+    }
+
     void clear()
     {
         Base::clear();
--- a/Hash_map/include/CGAL/Handle_hash_map.h
+++ b/Hash_map/include/CGAL/Handle_hash_map.h
@@ -1,0 +1,122 @@
+// Copyright (c) 2021  GeometryFactory Sarl (France).
+// All rights reserved.
+//
+// Licensees holding a valid commercial license may use this file in
+// accordance with the commercial license agreement provided with the software.
+//
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: LGPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s)     : Giles Bathgate <giles.bathgate@gmail.com>
+
+#ifndef CGAL_HANDLE_HASH_MAP
+#define CGAL_HANDLE_HASH_MAP
+
+#include <CGAL/config.h>
+#include <CGAL/Handle_hash_function.h>
+#include <CGAL/Tools/robin_hood.h>
+#include <cstddef>
+#include <functional>
+
+namespace CGAL
+{
+namespace internal
+{
+
+template <class Key,
+          class T,
+          class Hash = Handle_hash_function,
+          class KeyEqual = std::equal_to<Key>,
+          class Allocator = void>
+class Handle_hash_map
+  : private robin_hood::unordered_node_map<Key,T,Hash,KeyEqual>
+{
+    typedef robin_hood::unordered_node_map<Key,T,Hash,KeyEqual> Base;
+
+public:
+    // STL compliance
+    typedef Key                           key_type;
+    typedef T                             mapped_type;
+    typedef std::size_t                   size_type;
+    typedef Hash                          hasher;
+    typedef KeyEqual                      key_equal;
+    typedef typename Base::const_iterator const_iterator;
+
+    // Backwards compatibility
+    typedef mapped_type  Data;
+
+private:
+    mapped_type _default;
+
+public:
+    Handle_hash_map()
+        : _default(mapped_type())
+    {
+    }
+
+    Handle_hash_map(const mapped_type& def)
+        : _default(def)
+    {
+    }
+
+    Handle_hash_map(const mapped_type& def,size_type table_size)
+        : _default(def)
+    {
+        Base::reserve(table_size);
+    }
+
+    Handle_hash_map(key_type key_begin,key_type key_end,mapped_type values_begin,
+                    const mapped_type& def,size_type table_size)
+        : _default(def)
+    {
+        Base::reserve(table_size);
+        insert(key_begin,key_end,values_begin);
+    }
+
+    void clear()
+    {
+        Base::clear();
+    }
+
+    void clear(const mapped_type& def)
+    {
+        clear();
+        _default = def;
+    }
+
+    bool is_defined(const key_type& key) const
+    {
+        return Base::contains(key);
+    }
+
+    const mapped_type& operator[](const key_type& key) const
+    {
+        const_iterator f = Base::find(key);
+        if(f != Base::end())
+            return f->second;
+        return _default;
+    }
+
+    mapped_type& operator[](const key_type& key)
+    {
+        return Base::try_emplace(key,_default).first->second;
+    }
+
+    mapped_type insert(key_type key_begin,key_type key_end,
+                       mapped_type values_begin)
+    {
+        for(; key_begin != key_end; (++key_begin,++values_begin)) {
+            Base::insert_or_assign(key_begin,values_begin);
+        }
+        return values_begin;
+    }
+};
+
+}
+}
+
+#endif // CGAL_HANDLE_HASH_MAP
+
--- a/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/Reflex_vertex_searcher.h
+++ b/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/Reflex_vertex_searcher.h
@@ -16,6 +16,7 @@
 
 
 #include<CGAL/Nef_3/SNC_decorator.h>
+#include<CGAL/Unique_hash_map.h>
 #include<CGAL/Convex_decomposition_3/is_reflex_sedge.h>
 #undef CGAL_NEF_DEBUG
 #define CGAL_NEF_DEBUG 229
--- a/Hash_map/include/CGAL/Tools/robin_hood.h
+++ b/Hash_map/include/CGAL/Tools/robin_hood.h
@@ -1,0 +1,2529 @@
+//                 ______  _____                 ______                _________
+//  ______________ ___  /_ ___(_)_______         ___  /_ ______ ______ ______  /
+//  __  ___/_  __ \__  __ \__  / __  __ \        __  __ \_  __ \_  __ \_  __  /
+//  _  /    / /_/ /_  /_/ /_  /  _  / / /        _  / / // /_/ // /_/ // /_/ /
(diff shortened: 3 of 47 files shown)


afabri:
Did you only run benchmarks for construction of Nef Polyhedra from Polyhedra, or did you also bench performing Boolean operations, which  is the main purpose of the package.
Concerning  integration of other code we have to take a careful look.  Is it multi purpose, maintained, maintainable, etc.

GilesBathgate:
> Did you only run benchmarks for the construction of Nef Polyhedra from Polyhedra, or did you also bench performing Boolean operations, which is the main purpose of the package.

@afabri Yes benchmarks I have been running locally are for Nef Polyhedron construction only. The benchmarks https://github.com/GilesBathgate/hashmap-benchmarks are a more general component-based comparison.

I can do some other benchmarks to see the impact when larger datasets are loaded into the hashmaps. So far things look promising in general, but I still thought it necessary to limit the scope to Nef at this stage.

> Concerning integration of other code we have to take a careful look. Is it multi-purpose, maintained, maintainable, etc.

I understand your concerns here. I wanted to make this PR to start a discussion around that. https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-01-overview/ makes interesting reading.

GilesBathgate:
@afabri I've tried a basic benchmark which is the boolean set union of two nef. both are the elephant, one of them is mirrored in the XZ plane. There is no performance degradation but also no real performance boost either. I think I would prefer to look for other more fruitful enhancements.

With robin_hood hashmaps:
```
14:45:17: Starting /home/giles/projects/rapcad-release/rapcad ~/cad/elephant.rcad -o null...
Importing '/home/giles/projects/cgal/Surface_mesh_segmentation/benchmark/Surface_mesh_segmentation/data/refined_elephant.off'
Importing '/home/giles/projects/cgal/Surface_mesh_segmentation/benchmark/Surface_mesh_segmentation/data/refined_elephant.off'
Total compiling time: 0h 1m 9s 748ms.
```

Using Unique_hash_map:
```
14:34:07: Starting /home/giles/projects/rapcad-release/rapcad ~/cad/elephant.rcad -o null...
Importing '/home/giles/projects/cgal/Surface_mesh_segmentation/benchmark/Surface_mesh_segmentation/data/refined_elephant.off'
Importing '/home/giles/projects/cgal/Surface_mesh_segmentation/benchmark/Surface_mesh_segmentation/data/refined_elephant.off'
Total compiling time: 0h 1m 14s 967ms.
Total export time: 0h 0m 0s 1ms.
```

mglisse:
> I still thought it necessary to limit the scope to Nef at this stage.

I agree that the right implementation has to be chosen on a case by case basis. I know that in Gudhi, the last time I tried switching to absl, I got absolutely horrible performance, with the program taking 4 times longer, all of it spent in function begin() (those "fast" implementations all consider iteration to be an unimportant operation, and optimize insert/erase/find to the detriment of this one).

I think we should try to make the interface of Unique_hash_map as close as possible to std::unordered_map (the main obstacles would be the default value and the const operator[], but not all maps use those), for instance rename `is_defined` to `contains`.

GilesBathgate:
>  I tried switching to absl, I got absolutely horrible performance, with the program taking 4 times longer

I got terrible performance with absl in my benchmarks, but I think those benchmarks might not be very authentic https://github.com/GilesBathgate/hashmap-benchmarks/tree/abseil-unordered-map


> I think we should try to make the interface of Unique_hash_map as close as possible to std::unordered_map (the main obstacles would be the default value and the const operator[], but not all maps use those), for instance rename `is_defined` to `contains`.

Good idea, Yes. But is Unique_hash_map, not a publicly documented interface? I thought creating an internal one `internal::Handle_hash_map` for which the implementation/interface can be gradually refactored, would be better.

lrineau:
> > I tried switching to absl, I got absolutely horrible performance, with the program taking 4 times longer
> 
> I got terrible performance with absl in my benchmarks, but I think those benchmarks might not be very authentic [https://github.com/GilesBathgate/hashmap-benchmarks/tree/abseil-unordered-map](https://github.com/GilesBathgate/hashmap-benchmarks/tree/abseil-unordered-map?rgh-link-date=2021-02-22T09%3A24%3A45Z)
> 
> > I think we should try to make the interface of Unique_hash_map as close as possible to std::unordered_map (the main obstacles would be the default value and the const operator[], but not all maps use those), for instance rename `is_defined` to `contains`.
> 
> Good idea, Yes. But is Unique_hash_map, not a publicly documented interface? I thought creating an internal one `internal::Handle_hash_map` for which the implementation/interface can be gradually refactored, would be better.

Please do not fork `Unique_hash_map`. You can add a function `contains` to it, and make `is_defined` deprecated, but we do not want to maintain two version of that class template in CGAL.

GilesBathgate:
@lrineau Perhaps `Handle_hash_map` could just be:

```c++
namespace CGAL { namespace internal {

template <class Key, class T, class Hash = Handle_hash_function, class KeyEqual = std::equal_to<Key> >
using Handle_hash_map = std::unordered_map<Key,T,Hash,KeyEqual>

}}
```
This has the advantage that other backends compatible with `std::unordered_map` can be easily tried out but more work to change code using Unique_hash_map.

afabri:
It could be interesting to load meshes like this one. 
image
The simplify will remove all edges between coplanar triangles.   So we get pretty complex faces. Also when you  intersect them with a cube or the same model slightly moved,  the algorithm has to compute much more intersections.   
I don't say that this kind of model is more representative than elephants, but it looks more CAD like.
Ideally a benchmark covered data of different nature.

GilesBathgate:
@afabri Well something like this takes 5min to compute using boolean operations: https://github.com/GilesBathgate/PrintWorks/blob/stl/holy-cube.stl

It only takes 4 seconds to construct from stl via polygon_mesh_to_nef_3

GilesBathgate:
> It could be interesting to load meshes like this one.

@afabri I have seen this mesh on the CGAL homepage, but where is the file, please?

afabri:
You can find it [here](https://github.com/sloriot/cgal/blob/CGAL_data-moving_files/Data/data/meshes/cheese.off).

GilesBathgate:
> Please do not fork `Unique_hash_map`. You can add a function `contains` to it, and make `is_defined` deprecated, but we do not want to maintain two version of that class template in CGAL.

@lrineau My thinking was more along the lines of deprecating Unique_hash_map, and creating Handle_hash_map that describes more clearly by its name what it is for. Because it's backed by robin_hood it doesn't need to be unique keys. However, because Unique_hash_map is a documented interface users may still want it until they can migrate their code away from it.

Another approach is to just make the changes directly to Unique_hash_map but this affects more packages, and could affect user code so is a high impact change.

I am still investigating replacing hashmaps directly in #5486 but this is also quite a large change.

afabri:
@GilesBathgate did you take into account that the `Unique_hash_map`  has a default value for items not yet in the map?

Hash_map/include/CGAL/Handle_hash_map.h
```diff
@@ -0,0 +1,122 @@
...
+    typedef std::size_t                   size_type;
+    typedef Hash                          hasher;
+    typedef KeyEqual                      key_equal;
+    typedef typename Base::const_iterator const_iterator;
+
+    // Backwards compatibility
+    typedef mapped_type  Data;
+
+private:
+    mapped_type _default;
```
GilesBathgate, line 52: @afabri Yes the wrapper handles the default value, and below the code uses try_emplace to utilise this default value.

afabri:
I will run benchmarks on some different data sets to see what is the potential performance gain.   Did you try to put an `std::unordered_map` instead of robin-hood inside the wrapper class?

GilesBathgate:
> @GilesBathgate did you take into account that the `Unique_hash_map` has a default value for items not yet in the map?

https://github.com/CGAL/cgal/issues/4049#issuecomment-775133131

afabri:
@GilesBathgate  did you investigate what happens if we change the hard coded minimal table size of 512?   It is even strange that the constructor of the Unique_hash_map has a parameter for the table size which is by default 1, and in the implementation this is ignored for  values less than 512.

GilesBathgate:
> @GilesBathgate did you investigate what happens if we change the hard coded minimal table size of 512?

Yes 32 is the minimum that can work, the LEDA implementation (chained_map) is pretty good, because it specialises on Unique keys. The robin_hood scheme is more general and allows duplicate keys, but because of the minimal variance of the addressing scheme it still outperforms. There are other compatible implementations such as https://github.com/Tessil/robin-map

afabri:
With "it outperforms" you mean that the `chained_map` is still better?

GilesBathgate:
> With "it outperforms" you mean that the `chained_map` is still better?

No, robin_hood seems to be better for **construction** **insert** and **update** in my benchmarks https://github.com/GilesBathgate/hashmap-benchmarks, and better in the tests I have run with this PR applied. 

I recently worked with @martinus to further improve the performance, of try_emplace, which is even more efficient now that it only does a single hash lookup.

GilesBathgate:
@afabri Removing double lookups is also the theme of #5486, although there is also plenty of benefit from using `reserve` in many  places

GilesBathgate:
@afabri The simplest implementation of handle_hash_map which implements the concept of a default value is as follows:

```c++
template <class key_type, class mapped_type>
  class Handle_hash_map
    : public robin_hood::unordered_node_map<key_type,mapped_type,Handle_hash_function>
  {
      typedef robin_hood::unordered_node_map<key_type,mapped_type,Handle_hash_function> Base;

    public:
        handle_hash_map(const mapped_type& def) : _default(def) {}

        mapped_type& operator[](const key_type& key)
        { return Base::try_emplace(key,_default).first->second; }

    private:
        mapped_type _default;
  };
  ```
  
  It would require some other refactoring to make this work, but maybe more palatable for those who raised concerns regarding forking Unique_hash_map.

GilesBathgate:
@afabri Would you also be so kind as to run your performance tests against this branch?

PR #5485, opened 2021-02-22
Use TBB captured exception in PMP

danston:
In the macOS test suite, the `self_intersections_example` is using TBB. When the TBB support is on, this example is aborted due to an exception thrown by TBB, however, this exception is not caught. As far as I see, the problem is that the TBB version provided by brew (which is what I think most often used on macOS) is compiled without C++11 support, which is required to handle this exception according to the posts [here](https://software.intel.com/content/www/us/en/develop/blogs/exact-exception-propagation-in-the-intel-threading-building-blocks-library-and-c11.html) and [here](https://www.threadingbuildingblocks.org/docs/help/reference/environment/feature_macros.html). In fact, only one CGAL file seems to use this type of exception and this is the `self_intersections` from PMP.

This PR proposes a simple fix that allows catching the `tbb:captured_exception` when checking for self-intersections in case TBB is on. Please let me know if there is a better or more flexible fix.

The exact error is
```
TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it
libc++abi.dylib: terminating with uncaught exception of type tbb::captured_exception: std::exception
Abort trap: 6
```

## Release Management

* Issue(s) solved (if any): no opened issues

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -47,6 +47,7 @@
 #include <exception>
 #include <sstream>
 #include <type_traits>
+#include <typeinfo>
 #include <vector>
 
 #ifdef DOXYGEN_RUNNING
@@ -508,9 +509,14 @@ bool does_self_intersect(const FaceRange& face_range,
     return true;
   }
   #if defined(CGAL_LINKED_WITH_TBB)
-  catch (tbb::captured_exception& e)
+  catch (const tbb::captured_exception& e)
   {
-    return true;
+    const char* ti1 = e.name();
+    const char* ti2 = typeid(const CGAL::internal::Throw_at_output_exception&).name();
+    const std::string tn1(ti1);
+    const std::string tn2(ti2);
+    if (tn1 == tn2) return true;
+    else throw;
   }
   #endif
   catch ( ... )
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -508,7 +508,7 @@ bool does_self_intersect(const FaceRange& face_range,
   {
     return true;
   }
-  #if defined(CGAL_LINKED_WITH_TBB) && defined(TBB_USE_CAPTURED_EXCEPTION)
+  #if defined(CGAL_LINKED_WITH_TBB) && TBB_USE_CAPTURED_EXCEPTION
   catch (const tbb::captured_exception& e)
   {
     const char* ti1 = e.name();
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -519,10 +519,6 @@ bool does_self_intersect(const FaceRange& face_range,
     else throw;
   }
   #endif
-  catch ( ... )
-  {
-    throw;
-  }
   return false;
 }
 
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -508,7 +508,7 @@ bool does_self_intersect(const FaceRange& face_range,
   {
     return true;
   }
-  #if defined(CGAL_LINKED_WITH_TBB)
+  #if defined(CGAL_LINKED_WITH_TBB) && !TBB_USE_CAPTURED_EXCEPTION
   catch (const tbb::captured_exception& e)
   {
     const char* ti1 = e.name();
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -503,7 +503,11 @@ bool does_self_intersect(const FaceRange& face_range,
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
   }
+  #if defined(CGAL_LINKED_WITH_TBB)
+  catch(tbb::captured_exception&)
+  #else
   catch(CGAL::internal::Throw_at_output_exception&)
+  #endif
   {
     return true;
   }
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -508,7 +508,7 @@ bool does_self_intersect(const FaceRange& face_range,
   {
     return true;
   }
-  #if defined(CGAL_LINKED_WITH_TBB) && !TBB_USE_CAPTURED_EXCEPTION
+  #if defined(CGAL_LINKED_WITH_TBB) && defined(TBB_USE_CAPTURED_EXCEPTION)
   catch (const tbb::captured_exception& e)
   {
     const char* ti1 = e.name();
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -498,11 +498,24 @@ bool does_self_intersect(const FaceRange& face_range,
 {
   CGAL_precondition(CGAL::is_triangle_mesh(tmesh));
 
-  try {
+  try
+  {
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
-  } catch(const std::exception&) {
+  }
+  catch (const CGAL::internal::Throw_at_output_exception&)
+  {
     return true;
+  }
+  #if defined(CGAL_LINKED_WITH_TBB)
+  catch (tbb::captured_exception& e)
+  {
+    return true;
+  }
+  #endif
+  catch ( ... )
+  {
+    throw;
   }
   return false;
 }
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
@@ -498,20 +498,12 @@ bool does_self_intersect(const FaceRange& face_range,
 {
   CGAL_precondition(CGAL::is_triangle_mesh(tmesh));
 
-  try
-  {
+  try {
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
-  }
-  #if defined(CGAL_LINKED_WITH_TBB)
-  catch(tbb::captured_exception&)
-  #else
-  catch(CGAL::internal::Throw_at_output_exception&)
-  #endif
-  {
+  } catch(const std::exception&) {
     return true;
   }
-
   return false;
 }
 


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -503,7 +503,11 @@ bool does_self_intersect(const FaceRange& face_range,
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
   }
+  #if defined(CGAL_LINKED_WITH_TBB)
+  catch(tbb::captured_exception&)
+  #else
   catch(CGAL::internal::Throw_at_output_exception&)
+  #endif
```
lrineau, line 510: Even if `CGAL_LINKED_WITH_TBB` is defined, then `does_self_intersect` can be instantiated with `Sequential_tag`. It seems more robust to catch both exceptions:
```suggestion
  #if defined(CGAL_LINKED_WITH_TBB)
  catch(tbb::captured_exception&) {
    return true;
  }
  #else
  catch(CGAL::internal::Throw_at_output_exception&)
  #endif
```
but even then, if would be a lot more robust to verify that the exception caught be TBB is of type `CGAL::internal::Throw_at_output_exception`, maybe with something like:
```c++
  #if defined(CGAL_LINKED_WITH_TBB)
  catch(tbb::captured_exception& e) {
    if(e.name() == typeid(CGAL::internal::Throw_at_output_exception).name())
      return true;
    else
      throw;
  }
  #else
```

danston: Actually, you are right. This is something I missed.

danston: In fact, since both `TBB` and `Throw_at_output_exception` are derived from the `std::exception` and we do not really need to handle them but rather return upon their arrival, I changed to `std::exception`. This seems cleaner and more general.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -498,16 +498,12 @@ bool does_self_intersect(const FaceRange& face_range,
...
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
-  }
-  catch(CGAL::internal::Throw_at_output_exception&)
-  {
+  } catch(const std::exception&) {
     return true;
   }
-
   return false;
 }
```
lrineau, line 508: That looks wrong to me. Now the function `does_self_intersect` will return `true` as soon as any exception deriving from `std::exception` is thrown. If there is a `std::bad_alloc` during the execution, then the function will catch it, and returns `true`, as if the surface self-intersected.

The function is supposed to return `true` only if an exception of the precise type `CGAL::internal::Throw_at_output_exception` is thrown, and for any other type of exception, then exception should not be caught.

lrineau (changes requested):

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
danston: I see.

danston: Ok, I now start from the most specific exception and then only handle TBB, however the comparison between the `typeid name` and ` exception name` do not work for some reason, always returning `false`. I will have a look at what can be done with that.

danston:
Ok, it should work now.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -508,9 +509,14 @@ bool does_self_intersect(const FaceRange& face_range,
...
   #if defined(CGAL_LINKED_WITH_TBB)
-  catch (tbb::captured_exception& e)
+  catch (const tbb::captured_exception& e)
   {
-    return true;
+    const char* ti1 = e.name();
+    const char* ti2 = typeid(const CGAL::internal::Throw_at_output_exception&).name();
+    const std::string tn1(ti1);
+    const std::string tn2(ti2);
+    if (tn1 == tn2) return true;
```
lrineau, line 518: Oh... yes, in my advice I was only comparing the C-strings with `==`  and that was wrong!

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -503,11 +504,25 @@ bool does_self_intersect(const FaceRange& face_range,
...
+    else throw;
+  }
+  #endif
+  catch ( ... )
+  {
+    throw;
+  }
```
lrineau, line 525: Remove that part:
```suggestion
```

lrineau (approved):

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -503,11 +504,21 @@ bool does_self_intersect(const FaceRange& face_range,
...
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
   }
-  catch(CGAL::internal::Throw_at_output_exception&)
+  catch (const CGAL::internal::Throw_at_output_exception&)
   {
     return true;
   }
-
+  #if defined(CGAL_LINKED_WITH_TBB)
+  catch (const tbb::captured_exception& e)
```
maxGimeno, line 512: /mnt/testsuite/include/CGAL/Polygon_mesh_processing/self_intersections.h:512:41: warning: 'captured_exception' is deprecated [-Wdeprecated-declarations]

maxGimeno:
The captured_exception is deprecated, you will have to find what has replaced it and condition it to the right tbb version if need be.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -503,11 +504,21 @@ bool does_self_intersect(const FaceRange& face_range,
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
   }
-  catch(CGAL::internal::Throw_at_output_exception&)
+  catch (const CGAL::internal::Throw_at_output_exception&)
   {
     return true;
   }
-
+  #if defined(CGAL_LINKED_WITH_TBB)
```
lrineau, line 511: My suggestion to remove the deprecation warning: only use that code when necessary (on MacOS with odd TBB versions):
```suggestion
  #if defined(CGAL_LINKED_WITH_TBB) && TBB_USE_CAPTURED_EXCEPTION
```

danston: In fact, they deprecated the whole tbb exception concept so it all will be gone in the next versions. Maybe it is already in version 2021 but brew provides only the formula for the 2020 version.

danston: I will check your suggestion. It sounds good.

danston:
Ok, it looks like the distribution of the brew TBB comes without the support for the exact exception propagation that is requested on the `x86_64` platforms and I get the warning:

```
TBB Warning: Exact exception propagation is requested by the application but the linked library is built without support for it.
```

that does not show up for the other platforms. Also, the name `TBB_USE_CAPTURED_EXCEPTION` is not really nice. In fact, according to the documentation [here](https://www.threadingbuildingblocks.org/docs/help/reference/environment/feature_macros.html) it means:

- Define `TBB_USE_CAPTURED_EXCEPTION=1` to make the library rethrow an approximation. This is useful for uniform behavior across platforms.
- Define `TBB_USE_CAPTURED_EXCEPTION=0` to request rethrow of the exact exception.

So, the change should be `&& !TBB_USE_CAPTURED_EXCEPTION` and it would be better if they named this macro `TBB_USE_APPROXIMATE_CAPTURED_EXCEPTION`.

maxGimeno:
remaining warnings :https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-70/Polygon_mesh_processing/TestReport_lrineau_Ubuntu-Latest-CXX11.gz

danston:
Ok, it looked like I had a syntax error when checking for the macro.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/self_intersections.h
```diff
@@ -503,11 +504,21 @@ bool does_self_intersect(const FaceRange& face_range,
     CGAL::Emptyset_iterator unused_out;
     internal::self_intersections_impl<ConcurrencyTag>(face_range, tmesh, unused_out, true /*throw*/, np);
   }
-  catch(CGAL::internal::Throw_at_output_exception&)
+  catch (const CGAL::internal::Throw_at_output_exception&)
   {
     return true;
   }
-
+  #if defined(CGAL_LINKED_WITH_TBB) && defined(TBB_USE_CAPTURED_EXCEPTION)
```
lrineau, line 511: No, that is wrong. Here is the correct expression:
```suggestion
  #if defined(CGAL_LINKED_WITH_TBB) && TBB_USE_CAPTURED_EXCEPTION
```
In my version of TBB (version 2020.2), `tbb_config.h` defines `TBB_USE_CAPTURED_EXCEPTION` as `0` (the false value) or `1` (the true value). Either way it is defined, and `defined(TBB_USE_CAPTURED_EXCEPTION)` will always evaluate to `1`.

danston: Damn. Ok, this is what is not working on my platform and if I use `!TBB_USE_CAPTURED_EXCEPTION` it works on macOS but gives warnings on yours.

danston: I tried to define other TBB macros and no luck either.

danston: I can try to use `!TBB_USE_CAPTURED_EXCEPTION` and `__clang__` to get rid of the warning maybe on your platform?

lrineau: It seems brew packages TBB version 2020.3. And the `tbb_config.h` of that version has the same logic as on my machine:
https://github.com/oneapi-src/oneTBB/blob/eca91f16d7490a8abfdee652dadf457ec820cc37/include/tbb/tbb_config.h#L488-L500.

What does not work with `#if defined(CGAL_LINKED_WITH_TBB) && TBB_USE_CAPTURED_EXCEPTION`?

danston:
```
TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it
libc++abi.dylib: terminating with uncaught exception of type tbb::captured_exception: std::exception
Abort trap: 6
```

lrineau: Because brew version of TBB seems to have been compiled with a compiler that is not the default compiler of the machine, you should add `-DTBB_USE_CAPTURED_EXCEPTION=1` to the `CMAKE_CXX_FLAGS` of that test platform, to fix that issue.

danston: Yes, this does work! Thanks a lot!

danston:
Ok, to sum up. The macOS apple clang still uses by default C++98. The TBB from brew (which is currently `2020_U3`) is compiled without the C++11 support and hence without the support for `std::exception_ptr` so it is compiled with the option `TBB_USE_CAPTURED_EXCEPTION=1`. To check the C++ default:
```
clang -E -dM -xc++ /dev/null | grep cplusplus
```
This is why PMP should also be compiled with `-DTBB_USE_CAPTURED_EXCEPTION=1` in the `CMAKE_CXX_FLAGS` then it works.

maxGimeno:

Issue #5479, opened 2021-02-19
CGAL_TEST_SUITE with CTest


mglisse:
## Issue Details

I followed the instructions from https://github.com/CGAL/cgal/wiki/Testing to run the testsuite with CTest:
```sh
mkdir build
cd build
cmake -DBUILD_TESTING=ON -DWITH_examples=ON -DWITH_tests=ON ..
nice ctest -j16
```
and I had to kill test_validity_sampled_data after 5h (it was using 30G of memory). This is a non-optimized build, so I expect programs to run slow, but the main issue is that the macro CGAL_TEST_SUITE was not defined, so it went for the expensive test. I don't know if it should be defined by default, or if the documentation should mention that we need to define this macro in CXXFLAGS, but the current behavior when following the documentation is not good.

(there are other expensive tests, but not the same order of magnitude, test_triply_periodic_minimal_surfaces 1800s, Triangle_coordinates_speed_test 1100s)

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux 64
* Compiler: gcc-10
* Release or debug mode: not optimized
* Specific flags used (if any):
* CGAL version: master

sloriot:
I'm using that often without `CGAL_TEST_SUITE` and don't have any problem. Were you using master or your branch? (with `CMAKE_BUILD_TYPE=Debug`)

mglisse:
It was my branch, but it doesn't touch any cmake stuff. CMAKE_BUILD_TYPE was empty. How long does test_validity_sampled_data take for you without CGAL_TEST_SUITE on a debug build?

sloriot:
I'm running it...

lrineau:
> I'm running it...

Then, three days after, what was the outcome?

sloriot:
I had to kill it because I had a meeting and it was taking too much memory. I restarted it this morning and it keeps filling my memory (22GB for now). It is in Shape_detection btw.

sloriot:
It is strange cause I never noticed it in the past. I wonder if it has always been like that of it is due to an update of the package. Maybe @sgiraudot @danston know more.

sloriot:
I only realize know that there is 
```
#ifndef CGAL_TEST_SUITE // Disable tests too large for testsuite
  test_copied_point_cloud (points, 20);
  test_copied_point_cloud (points, 50);
#endif
```
in the test file and which is probably why you mentioned `CGAL_TEST_SUITE`...

[...]

PR #5487, opened 2021-02-23
Shape Detection: Disable expensive tests by default

sgiraudot:
## Summary of Changes

Shape detection runs tests that can take a while in debug mode, so it's better to disable them completely unless users want to try them.

## Release Management

* Issue(s) solved (if any): fix https://github.com/CGAL/cgal/issues/5479
* (Based on 5.2 as guilty test was introduced by https://github.com/CGAL/cgal/pull/5234 integrated in 5.2)

### DIFF
--- a/Shape_detection/test/Shape_detection/test_validity_sampled_data.cpp
+++ b/Shape_detection/test/Shape_detection/test_validity_sampled_data.cpp
@@ -20,6 +20,9 @@
 
 #include <boost/iterator/function_output_iterator.hpp>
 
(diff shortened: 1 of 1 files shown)


lrineau (approved):

mglisse:
CGAL_TEST_SUITE is also used in this file to define some timeouts. Do you want to keep it as it is, or also switch it to the new macro? (this is less important than the part you already changed)

maxGimeno:

PR #5488, opened 2021-02-23
[Small Feature] Remove TensorFlow support

sgiraudot:
## Rationale

Over the years, TensorFlow's support for its own C++ API has gone from very slight to non-existent. This support has never really been documented: there used to be a `CMakeLists` deeply buried in the repository which did allow to compile the C++ library quite easily, but this has been removed and there currently doesn't seem to be a way to get this C++ library compiled.

I spent hours trying to get it working with many different TF versions, with no success (every version failing at some point of the several-hours Bazel build, not always the same). I posted [an issue on their tracker](https://github.com/tensorflow/tensorflow/issues/42492), which has not been given any answer in 6 months after maintainers kept reassigning the issue to one another for a while.

My understanding of the situation is that the TF C++ library is only there for internal uses and is not officially supported (even though they do document some of the functions). There does not seem to be plans to support it better in the future. As a consequence, our wrapper has not been tested for years and is most probably broken, we cannot provide support for our own users (https://github.com/CGAL/cgal/issues/4928) with no hope to improve the situation anytime soon.

As a precision, I'll also say that this wrapper has never been state-of-the-art for our package (ETHZ Random Forest has always been our best classifier) and was always there more as a "proof-of-concept" than something really used.

For all these reasons, I suggest we remove the neural network classifier based on _TensorFlow_ from the Classification package.

## Summary of API changes

Remove `CGAL::TensorFlow::Neural_network_classifier` and all files related to TF's support from CGAL.

GeometryFactory

## CHANGES.md

```markdown
### [Classification](https://doc.cgal.org/5.3/Manual/packages.html#PkgClassification)
- **Breaking change**: the support for TensorFlow was dropped, the
  classifier `CGAL::TensorFlow::Neural_network_classifier` was removed.
  ```

## Status

* [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Remove_TensorFlow_support), pre-approved by @lrineau 2021-02-23


danston:
I completely agree. I also tried to use TF on macOS, different versions, and I have never managed to compile the C++ version having the same issues with the build. Whatever I tried it always failed at some point sooner or later.

sloriot:
Are people usually using docker images to use the C++ version?

sgiraudot:
People usually don't use the C++ version as far as I understand, only the Python version is really used.

lrineau:
From what I heard in a [CppCast episode on Tensorflow](https://cppcast.com/andrew-selle-tensorflow/), during last Summer 2020, and if I remember well, TensorFlow networks are trained mostly using Python, and then to "run" the trained networks there is [TensorFlow Lite](https://www.tensorflow.org/lite), that is available in a lot of languages.

lrineau:
@sgiraudot This PR is ready to be tested, right?

maxGimeno:
There is a conflict with master

sgiraudot:
Rebased, it's now ready to be tested.

maxGimeno:

PR #5489, opened 2021-02-23
Fixed Polyhedron demo warnings in the macOS test suite

danston:
The macOS test suite reveals various little warnings when compiling the Polyhedron demo. This PR fixes these warnings.

## Release Management

* Issue(s) solved (if any): warnings

### DIFF
--- a/Polyhedron/demo/Polyhedron/Viewer.h
+++ b/Polyhedron/demo/Polyhedron/Viewer.h
@@ -92,7 +92,7 @@ public:
   //!Set total number of depth peeling passes.
    void setTotalPass(int);
    void resetFov();
-   const QVector3D& scaler() const;
+   const QVector3D& scaler() const override;
 Q_SIGNALS:
   void sendMessage(QString);
   void doneInitGL(CGAL::Three::Viewer_interface*);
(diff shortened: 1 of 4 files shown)


Polyhedron/demo/Polyhedron/Viewer.h
```diff
@@ -92,7 +92,7 @@ class VIEWER_EXPORT Viewer : public CGAL::Three::Viewer_interface {
   //!Set total number of depth peeling passes.
    void setTotalPass(int);
    void resetFov();
-   const QVector3D& scaler() const;
+   const QVector3D& scaler() const override;
```
maxGimeno, line 95: Q_DECL_OVERRIDE for consistency with all other functions

lrineau: > Q_DECL_OVERRIDE for consistency with all other functions

Well, no. We should move away from `Q_DECL_OVERRIDE` now that we require C++14, and use `override` everywhere. In Qt6 they no longer use `Q_DECL_OVERRIDE` either, even if the macro is still defined.

danston: I can change all `Q_DECL_OVERRIDE` to the `override` to be consistent and respect the standard.

lrineau: Let's do that in another pull-request. This one is fine.

PR #5498, opened 2021-02-27
Cleanup index adder in polygon_mesh_to_nef_3

GilesBathgate:
## Summary of Changes

This is mostly just cleaning the code in `Face_graph_index_adder` which is part of the code in polygon_mesh_to_nef_3.h

* Remove multiple vector lookups using index.
* Remove an additional property hash lookup.
* Add methods for explicitly creating a new index, and return it instead of set_index()/get_index()
* Rename variable member variable to `shalfedges` instead of `hash`. 

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
@@ -118,7 +118,7 @@ class SNC_indexed_items {
     }
     int new_index()
     { index = Index_generator::get_unique_index(); return index; }
-    void set_index(int idx = Index_generator::get_unique_index())
+    void set_index(int idx)
     { index = idx; }
     int get_index() const { return index; }
     Halffacet_const_handle get_index_facet() const {
@@ -154,7 +154,7 @@ class SNC_indexed_items {
     }
     int new_index()
     { index = index2 = Index_generator::get_unique_index(); return index; }
-    void set_index(int idx = Index_generator::get_unique_index())
+    void set_index(int idx)
     { index = index2 = idx; }
     int get_index() const {
       return index;
@@ -190,7 +190,7 @@ class SNC_indexed_items {
     }
     int new_index()
     { index = Index_generator::get_unique_index(); return index; }
-    void set_index(int idx = Index_generator::get_unique_index())
+    void set_index(int idx)
     { index = idx; }
     int get_index() const { return index; }
   };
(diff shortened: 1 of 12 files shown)


Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
```diff
@@ -116,7 +116,8 @@ class SNC_indexed_items {
       init_ifacet = sl.init_ifacet;
       return *this;
     }
-
+    int new_index()
+    { index = Index_generator::get_unique_index(); return index; }
     void set_index(int idx = Index_generator::get_unique_index())
```
afabri, line 121: Wouldn't it make sense to remove the default for the parameter `idx`?

Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
```diff
@@ -126,48 +126,44 @@ class Face_graph_index_adder<CGAL::SNC_indexed_items, PolygonMesh, SNC_structure
 
   typedef Halfedge_around_face_circulator<PolygonMesh>
     Halfedge_around_facet_const_circulator;
-    typedef std::vector<SHalfedge_handle> Hash;
+    typedef std::vector<SHalfedge_handle> SHalfedge_list;
```
afabri, line 129: It's only a local variable, but I am not sure that it is a good idea to call a vector list.

GilesBathgate: I can update this. Should it be:
* `SHalfedges`
* `SHalfedge_vector`
* `SHalfedgeVector`

afabri: I would go for `SHalfedges`

Nef_3/include/CGAL/Nef_3/SNC_indexed_items.h
GilesBathgate: Yes. I just didn't want to break any existing code.

afabri: If it is not documented, and not used then we can change it.

afabri (approved):

maxGimeno:
There is this error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-74/Nef_3/TestReport_lrineau_Ubuntu-Latest.gz
but it might be a problem with another branch. I'd say let's wait for the 39 branches currently in test to be integrated before further investigating.

GilesBathgate:
@maxGimeno No there was a mistake. Fixed.

maxGimeno:
Thank you

maxGimeno:

PR #5501, opened 2021-03-01
CMake Fixes when CGAL_Core is disabled (5.2)

maxGimeno:
## Summary of Changes

Don't build tests that needCGAL_Core when CGAL_Core is not found.
## Release Management
5.1 version of #5502

PR #5502, opened 2021-03-01
CMake Fixes when CGAL_Core is disabled (5.1)

maxGimeno:
## Summary of Changes
Don't build tests that need CGAL_Core when CGAL_Core is not found
## Release Management
5.1 version of #5501


Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
```diff
@@ -1424,3 +1424,9 @@ compile_and_run(test_io)
 compile_and_run(test_sgm)
 
 compile_and_run(test_polycurve_intersection)
+if(CGAL_DISABLE_GMP)
+  get_directory_property(LIST_OF_TESTS TESTS)
+  foreach(_test ${LIST_OF_TESTS})
+    set_tests_properties(${_test} PROPERTIES ENVIRONMENT CGAL_DISABLE_GMP=1)
```
lrineau, line 1430: Use `APPEND` to avoid overwriting any other env variable.
```suggestion
    set_property(TEST ${_test} APPEND PROPERTY ENVIRONMENT CGAL_DISABLE_GMP=1)
```

Issue #805, opened 2016-02-19
Isotropic meshing: make it output sensitive


sloriot:
Right now the function is not output sensitive as all halfedges in the mesh are marked in the function `tag_halfedges_status`. There might be other places breaking the output sensitiveness.

I'm not sure this is a critical issue.

cc @janetournois

sloriot:
In `tangential_relaxation()`, there is a loop on all vertices that should be only on the vertices in the input face range.

lrineau:
I flag this issue with the milestone `4.10.1`. Jane, once you start working on the issue, make sure your working branch is based on `releases/CGAL-4.10-branch`.

sloriot:
It's clearly not for the upcoming release. I changed the milestone.

lrineau:
I have reverted your change: 4.10.1 is the right milestone. It is not for 4.10, but in should be in the 4.10-branch

PR #5507, opened 2021-03-02
PMP:  Accelerate isotropic remeshing

afabri:
## Summary of Changes

vtuning in order to find small improvements to make.

### Todo
- [x] Do tests with smaller than average edge length
- [x] Do test with larger than average edge length

## Release Management

### DIFF
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -416,7 +416,7 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, const Tag_true&) const
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_true) const
       {
           return GeomTraits().do_intersect_3_object()
           (GeomTraits().construct_sphere_3_object()
@@ -424,7 +424,7 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, const Tag_false&) const
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_false) const
       {
           return GeomTraits().do_intersect_3_object()
           (GeomTraits().construct_sphere_3_object()
@@ -434,7 +434,7 @@ public:
 
       CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound) const
       {
-        return (*this)(p, bb, bound, CGAL::Boolean_tag<internal::Has_static_filters<GeomTraits>::value>());
+        return (*this)(p, bb, bound, Boolean_tag<internal::Has_static_filters<GeomTraits>::value>());
       }
 
       template <class Solid>
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -414,6 +414,14 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
+    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound) const
+      {
+          return GeomTraits().do_intersect_3_object()
+          (GeomTraits().construct_sphere_3_object()
+           (p, GeomTraits().compute_squared_distance_3_object()(p, bound)), bb,true)?
+          CGAL::SMALLER : CGAL::LARGER;
+      }
+
       template <class Solid>
       CGAL::Comparison_result operator()(const Point& p, const Solid& pr, const FT& sq_distance) const
       {
@@ -423,26 +431,7 @@ public:
           CGAL::SMALLER :
           CGAL::LARGER;
       }
-    /*
-    template <>
-      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& pr, const Point& bound) const
-      {
-          return GeomTraits().do_intersect_3_object()
-          (GeomTraits().construct_sphere_3_object()
-           (p, GeomTraits().compute_squared_distance_3_object()(p, bound)).bbox(), pr)?
-          CGAL::SMALLER : CGAL::LARGER;
-      }
 
-      template <>
-      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& pr, const FT& sq_distance) const
-      {
-        return GeomTraits().do_intersect_3_object()
-          (GeomTraits().construct_sphere_3_object()(p, sq_distance).bbox(),
-           pr) ?
-          CGAL::SMALLER :
-          CGAL::LARGER;
-      }
-    */
   };
 
   Closest_point closest_point_object() const {return Closest_point(*this);}
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -423,6 +423,26 @@ public:
           CGAL::SMALLER :
           CGAL::LARGER;
       }
+    /*
+    template <>
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& pr, const Point& bound) const
+      {
+          return GeomTraits().do_intersect_3_object()
+          (GeomTraits().construct_sphere_3_object()
+           (p, GeomTraits().compute_squared_distance_3_object()(p, bound)).bbox(), pr)?
+          CGAL::SMALLER : CGAL::LARGER;
+      }
+
+      template <>
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& pr, const FT& sq_distance) const
+      {
+        return GeomTraits().do_intersect_3_object()
+          (GeomTraits().construct_sphere_3_object()(p, sq_distance).bbox(),
+           pr) ?
+          CGAL::SMALLER :
+          CGAL::LARGER;
+      }
+    */
   };
 
   Closest_point closest_point_object() const {return Closest_point(*this);}
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -24,6 +24,8 @@
 #include <CGAL/internal/AABB_tree/Has_nested_type_Shared_data.h>
 #include <CGAL/internal/AABB_tree/Is_ray_intersection_geomtraits.h>
 #include <CGAL/internal/AABB_tree/Primitive_helper.h>
+#include <CGAL/internal/Has_boolean_tags.h>
+
 
 #include <boost/optional.hpp>
 
@@ -414,7 +416,7 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_true&) const
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, const Tag_true&) const
       {
           return GeomTraits().do_intersect_3_object()
           (GeomTraits().construct_sphere_3_object()
@@ -422,7 +424,7 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_false&) const
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, const Tag_false&) const
       {
           return GeomTraits().do_intersect_3_object()
           (GeomTraits().construct_sphere_3_object()
@@ -430,10 +432,22 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound) const
-    {
-      return (*this)(p, bb, bound, typename GeomTraits::Has_filtered_predicates_tag());
-    }
+      template <typename GT, bool HasTag = internal::Has_nested_type_Has_static_filters<GeomTraits>::value>
+      struct Use_conservative_static_filters
+      {
+        typedef CGAL::Tag_false type;
+      };
+
+      template <typename GT>
+      struct Use_conservative_static_filters<GT, true>
+      {
+        typedef CGAL::Boolean_tag<GT::Has_static_filters> type;
+      };
+
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound) const
+      {
+        return (*this)(p, bb, bound, typename Use_conservative_static_filters<GeomTraits>::type());
+      }
 
       template <class Solid>
       CGAL::Comparison_result operator()(const Point& p, const Solid& pr, const FT& sq_distance) const
--- a/AABB_tree/include/CGAL/AABB_traits.h
+++ b/AABB_tree/include/CGAL/AABB_traits.h
@@ -414,13 +414,26 @@ public:
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound) const
+    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_true&) const
       {
           return GeomTraits().do_intersect_3_object()
           (GeomTraits().construct_sphere_3_object()
            (p, GeomTraits().compute_squared_distance_3_object()(p, bound)), bb,true)?
           CGAL::SMALLER : CGAL::LARGER;
       }
+
+    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_false&) const
+      {
+          return GeomTraits().do_intersect_3_object()
+          (GeomTraits().construct_sphere_3_object()
+           (p, GeomTraits().compute_squared_distance_3_object()(p, bound)), bb)?
+          CGAL::SMALLER : CGAL::LARGER;
+      }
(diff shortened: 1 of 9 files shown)


afabri:
I also replaced some vector with small_vector but it slowed things down. 
When reading iphigenia.off   I see `std::getline()` and `CGAL::Euler::add_face()` relatively high. That is not in isotropic remeshing, but let's mention it here.

janetournois (approved):

afabri:
@janetournois when  I simplify (that is a target edge length larger than the average edge length)  iphigenia.off, we spend 16% of the time in an [exact collinearity test](https://github.com/CGAL/cgal/blob/master/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h#L1422).  Is that by accident or by construction?

BGL/include/CGAL/boost/graph/Euler_operations.h
```diff
@@ -737,9 +737,9 @@ add_face(const VertexRange& vr, Graph& g)
                       patch_start, patch_end;
   // cache for set_next and vertex' set_halfedge
   typedef std::pair<halfedge_descriptor, halfedge_descriptor> NextCacheEntry;
-  typedef std::vector<NextCacheEntry>    NextCache;
+  typedef boost::container::small_vector<NextCacheEntry,9>    NextCache;
   NextCache next_cache;
-  next_cache.reserve(3 * n);
+  //next_cache.reserve(3 * n);  Unfortunately small_vector has no reserve
```
lrineau, line 742: [`boost::small_vector`](https://www.boost.org/doc/libs/1_75_0/doc/html/boost/container/small_vector.html) inherits all the members from [`boost::vector`](https://www.boost.org/doc/libs/1_75_0/doc/html/boost/container/vector.html), including [`reserve(size_type)`](https://www.boost.org/doc/libs/1_75_0/doc/html/boost/container/vector.html#idm46216567859600-bb)

afabri:
@janetournois can we avoid this [vector of vectors](https://github.com/CGAL/cgal/blob/master/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h#L1883) ?  For example just a vector per patch with one normal. And when the next normal of the same patch is computed, we do the test right away, and replace the one stored normal.

Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Sphere_3_do_intersect.h
```diff
@@ -35,39 +35,59 @@ namespace internal {
...
                 Point center = sphere.center();
 
                 if(center.x() < (FT)bbox.xmin())
                 {
                         d = (FT)bbox.xmin() - center.x();
-                        distance += d * d;
+                        d = square(d);
+                        if(d > sr){
+                          return false;
+                        }
```
lrineau, line 48: I do not think that code can be trusted. `d` is a construction, and its comparison to `sr` is not exact. It can be wrong in both way (false-positive or false-negative).

lrineau: Actually, the issue is elsewhere: If `FT` is `Interval_nt`, then this comparison could throw the exception `Uncertain_conversion_exception` too early. There should be a `certainly`:
```suggestion
                        d = square(d);
                        if(certainly(d > sr)){
                          return false;
                        }
```
and the same for all the comparisons supposed to trigger an early `return false`.

sloriot: it is exact because we are using filtered predicates (i.e. interval or gmpq)

Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h
```diff
@@ -265,6 +264,7 @@ class Do_intersect_3
         if (double_tmp_result < -eps)
           return true;
         else {
+          // return true; // AF:  just to test conservative approach
```
afabri, line 267: @lrineau   Maybe you have an idea for the following problem:  We use this  sphere/bbox   test in the AABB tree for finding the closest triangle.  While vtuning code I noticed that there are filter failures that is we do the test with exact arithmetic.   I could well live with a `true` (that is overlap) in case of uncertainly.  The question is, if we have to add such a functor to the kernel, or if we can add somehow a parameter.

lrineau: That would be a sort of "possibly intersect" instead of "do intersect".

`CGAL::internal::Static_filters_predicates::Do_intersect_3` could have a third template parameter (a non-type template argument, I suggest a dedicated enum, to avoid [the curve of a Boolean parameter](https://softwareengineering.stackexchange.com/a/147983/118556)), with a default-value.

Then we can add a new functor name in the Kernel concept, and it would be implemented using either `Do_intersect_3` or `CGAL::internal::Static_filters_predicates::Do_intersect_3` with the special value as third template argument.


I might actually be more efficient to implement that `possibly_intersect(sphere, bbox)` by `do_overlap(sphere.bbox(), bbox)`.

sloriot: Note that I also mentioned the tag this morning with https://github.com/CGAL/cgal/pull/4654 in mind where we also brought the tag onto the table to allow/forbid degenerate input.

afabri: To use the bbox of the sphere makes it 25% slower.

afabri:
@lrineau do you agree that for a `Bbox_3`  there is no need to call `fit_in_double()` in [this ](https://github.com/CGAL/cgal/blob/master/Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h#L186)static filter?

lrineau:
> @lrineau do you agree that for a `Bbox_3` there is no need to call `fit_in_double()` in [this ](https://github.com/CGAL/cgal/blob/master/Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h?rgh-link-date=2021-03-10T10%3A53%3A47Z#L186)static filter?

That is right. The coordinates of `Bbox_3` are always `double` and never `Interval_nt`.

AABB_tree/include/CGAL/AABB_traits.h
```diff
@@ -414,26 +416,38 @@ class AABB_traits
           CGAL::SMALLER : CGAL::LARGER;
       }
 
-    CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_true&) const
+      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, const Tag_true&) const
```
lrineau, line 419:
```suggestion
      CGAL::Comparison_result operator()(const Point& p, const Bounding_box& bb, const Point& bound, Tag_true) const
```
Pass tags by value: they are empty objects. If you pass them by reference (to const or not), I do not think the compiler can optimize as easily.

Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp
```diff
@@ -1,6 +1,7 @@
+#define EARLY
```
sloriot, line 1: To be removed?

Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp
```diff
@@ -35,14 +37,20 @@ int main(int argc, char* argv[])
...
   {
     std::cerr << "Invalid input." << std::endl;
     return 1;
   }
 
-  double target_edge_length = 0.04;
+  std::cout << "Read: " << t.time() << " sec" << std::endl;
+  t.reset();
+
+  double target_edge_length = (argc > 2) ? std::stod(std::string(argv[2])) : 1.0;
```
sloriot, line 53: Any reason to change the previous default?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -825,6 +827,19 @@ namespace internal {
...
+      typedef typename boost::property_map<PM, CGAL::dynamic_vertex_property_t<int> >::type Vertex_degree;
+      Vertex_degree degree = get(CGAL::dynamic_vertex_property_t<int>(), mesh_);
+
+      for(vertex_descriptor v : vertices(mesh_)){
+        put(degree,v,0);
+      }
+      for(halfedge_descriptor h : halfedges(mesh_))
+      {
+        vertex_descriptor t = target(h, mesh_);
+        put(degree, t, get(degree,t)+1);
```
sloriot, line 840: @janetournois shall the degree depend only on the selected edges? This matters only for vertices on the boundary of a selection.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -859,6 +875,12 @@ namespace internal {
         vvb -= 1;
         vvc += 1;
         vvd += 1;
+#ifdef VALE
```
sloriot, line 878: Do you want to keep that ? you observe no gain when activating it?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
```diff
@@ -1879,8 +1913,10 @@ namespace internal {
     bool check_normals(const HalfedgeRange& hedges) const
     {
       std::size_t nb_patches = patch_id_to_index_map.size();
+      //std::vector<boost::optional<Vector_3> > normal_per_patch(nb_patches,boost::none);
```
sloriot, line 1916: remove?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
janetournois: The degree of a vertex that stands on the boundary of the selection (i.e. on a `PATCH_BORDER`) should count only internal edges.
Here all edges are counted, right?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
afabri: In fact THAT is the code we must take.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
afabri: Yes. I later replaced it with two vectors.   By the way I am not sure if `vector<bool>` is a good idea, and if it should not be `vector<int>`   instead.

Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp
afabri: By accident.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
sloriot: Note that in your code, IRC you were using `degree()` so Andreas' code does the same thing the original (to be checked)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
sloriot: why would it be bad?

Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp
afabri: No longer needed. It was for an `#ifdef EARLY` which I had in the code.

AABB_tree/include/CGAL/AABB_traits.h
afabri: So this is done ??

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
afabri: But in the original code we call the `degree()` function for the vertex which does not distinguish either.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
afabri: Because it is more time consuming to change a bit inside an int.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
janetournois: You are right.
`valence(v)` and `target_valence(v)` (which is 4 on borders and 6 elsewhere) both take into account only the "real" borders, not the patch borders.

It's not wrong, but maybe we can improve that, when a vertex stands at the limit between a patch to be remeshed and a patched to be left as is.
We can keep it this way for now.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
sloriot: but you checked with optional and `vector<bool>` was faster? I agree with int since nb_patches is not expected to be very large

afabri: I did not really see a difference in vtune but sticked to the second solution.  I don''t agree  on what you say on `nb_patches`.  If it is the number of patches of the data set and not only the patches incident to the vertex it may be arbitrarily  big.

janetournois: it is the total number of surface patches indeed, but still it's likely not to be arbitrarily big. A hundred would already be a strong data set...

lrineau: > Because it is more time consuming to change a bit inside an int.

That is very cheap, comparable to a memory load or a memory store from L1 cache. Both are about 3-4 CPU cycles. As soon as you have L1 caches misses, the memory load/stores are at more costly that that operation (setting a bit in a word).

That should be benched, but I doubt you will see any penalty for using `std::vector<bool>` instead of `std::vector<int>`

janetournois:
I made quite a few experiments with this branch, and all the results are good/similar to `master`.

About timings, the computation time (displayed by the demo) is on average 10 times faster than `master` on those experiments!

afabri:
I have not planned to do more work on this. Note also that some improvements are not in the PMP package, so even other packages will profit from this PR.

sloriot:
I think the next step with this function should be #805. The improvement is very nice! Congrats.

afabri:
> 
> 
> I think the next step with this function should be #805. The improvement is very nice! Congrats.

But in an independent PR I would suggest, so that we have this one quickly in the master branch.

sloriot:
> > I think the next step with this function should be #805. The improvement is very nice! Congrats.
> 
> But in an independent PR I would suggest, so that we have this one quickly in the master branch.

Of course !

Polygon_mesh_processing/examples/Polygon_mesh_processing/isotropic_remeshing_example.cpp
```diff
@@ -62,5 +73,13 @@ int main(int argc, char* argv[])
...
   std::cout << "Remeshing done." << std::endl;
 
+  std::cout << "took " << t.time() << " sec" << std::endl;
+  t.reset();
+
+  std::ofstream out("out.off");
+  out.precision(17);
+  //out << mesh << std::endl;
+  out.close();
+
```
afabri, line 83: @MaelRL just spotted this.   @janetournois shall we write a mesh or shall we just go back to the original file and also kick out the timer.

janetournois: Since it is the "main" example for isotropic remeshing, visible in the user manual, I would remove this and the timer too.

afabri: If you look at the other examples it seems that those who somehow modify the mesh or produce a mesh (union), at the end write the mesh to a file.   We might unify.   I also do not like the prefix or suffix `example`   in the file names.

sloriot (approved):

lrineau:
@maxGimeno Both Sébastien and Mael approved this PR. It can be tested now.

maxGimeno:

PR #5509, opened 2021-03-02
Robustify internal side-of-triangle-mesh calls

sloriot:
Use an exact AABB-tree using exact node values.

Not backported to 5.1 on purpose.


maxGimeno:

PR #5510, opened 2021-03-03
Round-to-nearest support for intervals

### LINKED ISSUE #2791: FPU: Support platforms without round-towards-infinity

pentacular:
## Summary of Changes

This adds support for using round-to-nearest for intervals, as an alternative to round-to-infinity, suitable for use with WASM.

Note: I'm a bit unclear on how this ought to be best organized for CGAL -- please let me know what would fit better.

## Release Management

* Issue(s) solved (if any): fix #2791


sloriot:
Thanks for the PR, this is something that was on my TODO list but I did not find time to do it.

We should add some cmake machinery so that the there is an option to allow to not change the rounding mode + do not add the `-frounding-math` for `g++` when ON.

Number_types/include/CGAL/FPU.h
```diff
@@ -347,16 +346,32 @@ inline double IA_bug_sqrt(double d)
 // With GCC, we can do slightly better : test with __builtin_constant_p()
 // that both arguments are constant before stopping one of them.
 // Use inline functions instead ?
-#define CGAL_IA_ADD(a,b) CGAL_IA_FORCE_TO_DOUBLE((a)+CGAL_IA_STOP_CPROP(b))
-#define CGAL_IA_SUB(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)-(b))
-#define CGAL_IA_MUL(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)*CGAL_IA_STOP_CPROP(b))
-#define CGAL_IA_DIV(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)/CGAL_IA_STOP_CPROP(b))
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
```
sloriot, line 349: IMO, if we are adding an extra case, we should simply assume that we do not know what the rounding mode is and not rely on to-nearest. I'm in favor of something like `CGAL_NO_FPU_ROUNDING_MODE_CHANGE`

Number_types/include/CGAL/FPU.h
```diff
@@ -503,12 +518,18 @@ FPU_get_and_set_cw (FPU_CW_t cw)
...
+#define CGAL_FE_PROTECTED CGAL_FE_UPWARD
+#endif
+
 template <bool Protected = true> struct Protect_FPU_rounding;
 
 template <>
 struct Protect_FPU_rounding<true>
 {
-  Protect_FPU_rounding(FPU_CW_t r = CGAL_FE_UPWARD)
+  Protect_FPU_rounding(FPU_CW_t r = CGAL_FE_PROTECTED)
```
sloriot, line 532: The protector should not do anything in the new mode. The partial specialization for the protected mode could be hidden behind the macro.

Number_types/include/CGAL/FPU.h
sloriot: I guess, this implies more changes but that should be the way to go.

Number_types/include/CGAL/Interval_nt.h
```diff
@@ -1141,16 +1141,20 @@ namespace INTERN_INTERVAL_NT {
...
       // dependency.
     }
-#else
+#else // no __AVX512F__
     // TODO: Alternative for computing CGAL_IA_SQRT_DOWN(d.inf()) exactly
     // without changing the rounding mode:
     // - compute x = CGAL_IA_SQRT(d.inf())
     // - compute y = CGAL_IA_SQUARE(x)
     // - if y==d.inf() use x, else use -CGAL_IA_SUB(CGAL_IA_MIN_DOUBLE,x)
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
```
mglisse, line 1150: `#elif` is usually nicer than nested `#if`.

Number_types/include/CGAL/FPU.h
```diff
@@ -347,16 +346,32 @@ inline double IA_bug_sqrt(double d)
 // With GCC, we can do slightly better : test with __builtin_constant_p()
 // that both arguments are constant before stopping one of them.
 // Use inline functions instead ?
-#define CGAL_IA_ADD(a,b) CGAL_IA_FORCE_TO_DOUBLE((a)+CGAL_IA_STOP_CPROP(b))
-#define CGAL_IA_SUB(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)-(b))
-#define CGAL_IA_MUL(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)*CGAL_IA_STOP_CPROP(b))
```
mglisse, line 352: You dropped CGAL_IA_FORCE_TO_DOUBLE for the normal case, that looks wrong.

Number_types/include/CGAL/FPU.h
pentacular: It's an interesting point.

I think it's difficult to predict correct behavior with all existing and future rounding modes.

So I'd avoid a claim like CGAL_NO_FPU_ROUNDING_MODE_CHANGE.

But something like CGAL_CONSERVATIVE_INTERVALS might describe what's actually happening in this change, and then you could decide if that covers whatever floating point modes you care about or not.

I also notice that there are systems other than interval which care about floating point mode, and it's not clear to me that there will be a generic fix for all of them.

In which case CGAL_ALWAYS_ROUND_TO_NEAREST would be the right kind of flag to tell them about the particular environment they're operating in -- maybe some would need a CGAL_ALWAYS_ROUND_UPWARD, etc

Number_types/include/CGAL/Interval_nt.h
pentacular: I'm happy to change it, but was concerned that the context of the comment about the alternative approach would become confused.

Number_types/include/CGAL/FPU.h
pentacular: Yes, I think you're right -- I didn't look deep enough into all of the cases it handles.

How about putting that into the other branch of CGAL_IA_UP?

Or is that getting too overloaded?

Number_types/include/CGAL/FPU.h
sloriot: I don't think so. I think as long as you use `nextafter()` for the upper and -lower bound of the interval you are safe side no matter how the value was rounded

pentacular: For intervals, certainly -- but CGAL_NO_FPU_ROUNDING_MODE_CHANGE seems to make a more universal claim that everything in CGAL will always work without changing rounding mode.

sloriot: AFAIK, this is the only place where is is needed.

Number_types/include/CGAL/FPU.h
mglisse: Putting it in the other branch of CGAL_IA_UP sounds fine.

Number_types/include/CGAL/FPU.h
mglisse: Static filters do require round-to-nearest AFAIK (the bounds would be worse with arbitrary rounding), not sure if that's what you are talking about.

Number_types/include/CGAL/FPU.h
```diff
@@ -347,16 +346,32 @@ inline double IA_bug_sqrt(double d)
...
-#define CGAL_IA_DIV(a,b) CGAL_IA_FORCE_TO_DOUBLE(CGAL_IA_STOP_CPROP(a)/CGAL_IA_STOP_CPROP(b))
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
+inline double CGAL_IA_UP(double d)
```
mglisse, line 350: Since it isn't a macro and it is already in namespace CGAL, you don't need to prefix it with CGAL and write it in all caps. I see functions called IA_opacify or IA_force_to_double nearby.

Number_types/include/CGAL/FPU.h
```diff
@@ -347,16 +346,32 @@ inline double IA_bug_sqrt(double d)
...
+#define CGAL_IA_MUL(a,b) CGAL_IA_UP(CGAL_IA_STOP_CPROP(a)*CGAL_IA_STOP_CPROP(b))
+#define CGAL_IA_DIV(a,b) CGAL_IA_UP(CGAL_IA_STOP_CPROP(a)/CGAL_IA_STOP_CPROP(b))
 inline double CGAL_IA_SQUARE(double a){
   double b = CGAL_IA_STOP_CPROP(a); // only once
-  return CGAL_IA_FORCE_TO_DOUBLE(b*b);
+  return CGAL_IA_UP(b*b);
 }
 #define CGAL_IA_SQRT(a) \
-        CGAL_IA_FORCE_TO_DOUBLE(CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(a)))
+        CGAL_IA_UP(CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(a)))
```
mglisse, line 374: It may be slightly confusing that in the new mode CGAL_IA_SQRT means up(sqrt) while in the old mode, it means sqrt with dynamic rounding, and it is actually used once with downward rounding (the other operations are only ever used with upward rounding).

Number_types/include/CGAL/FPU.h
pentacular: Hmm, I see a number of cases that aren't clear to me.

Bounding_volumes/include/CGAL/Approximate_min_ellipsoid_d.h -- toward-zero and upward
Apollonius_graph_2/include/CGAL/Apollonius_graph_2/compare_quadratic.h -- upward
Number_types/include/CGAL/leda_bigfloat.h -- upward
Mesh_3/include/CGAL/Mesh_3/Robust_intersection_traits_3.h -- to nearest
Skin_surface_3/include/CGAL/Skin_surface_base_3.h -- to nearest
Straight_skeleton_2/include/CGAL/Straight_skeleton_2/Straight_skeleton_builder_traits_2_aux.h -- to nearest
Polynomial/test/Polynomial/Polynomial_traits_d_Residue.cpp -- upward
Polynomial/test/Polynomial/Polynomial_using_core.cpp -- upward
Polynomial/test/Polynomial/resultant.cpp -- upward
Polynomial/test/Polynomial/Polynomial_using_leda.cpp -- upward
Convex_hull_3/include/CGAL/convex_hull_3.h

Some of these don't seem to be part of filters, e.g., Approximate_min_ellipsoid.

My suggestion is to require explicit opt-in to any forced rounding mode.

i.e. If we use CGAL_ALWAYS_ROUND_TO_NEAREST then we can have CGAL_IA_SETFPCW throw an exception (or a compile-time error) when something tries to use a different mode, which will require systems to explicitly opt into CGAL_ALWAYS_AROUND_TO_NEAREST.

This should minimize surprising behavior, and allow gradual adoption, and allow alternate implementation for particular modes.

On the other hand, CGAL_NO_FPU_ROUNDING_MODE_CHANGE seems like it might require code to dynamically dispatch on the current rounding mode in the worst case, since it would need to operate in any existing or future rounding mode.

In any case, I'm operating from a position of ignorance, so maybe I'm being overly conservative -- but please make sure you're confident that you've considered all of the possibilities. :)

Number_types/include/CGAL/FPU.h
pentacular: I think we should consider testing that we are using the expected rounding mode, and error otherwise.

Number_types/include/CGAL/FPU.h
pentacular: I've abstracted IA_sqrt_up and IA_sqrt_toward_zero.

It's a little awkward because of scoping.

Let me know what you think.

Number_types/include/CGAL/FPU.h
pentacular: Thanks -- I've changed to IA_up.

Please let me know if you have a better name.

Number_types/include/CGAL/Interval_nt.h
pentacular: I think this is obsoleted by the abstraction into IA_sqrt_toward_zero.

Number_types/include/CGAL/FPU.h
pentacular: Done

mglisse:
You mentioned benchmarks in the issue, could you post some benchmark results, explaining exactly what you tested? It would be interesting to know how this compares to current intervals, and how it compares to not using intervals at all.

pentacular:
I just did a simple test which produced a sphere, then subtracted a half size sphere incrementally around the equator at each 360 / 7 degrees. https://imgur.com/a/ucZu3cl

The time taken with this approach was about 65% of the time taken without intervals, which is a significant improvement, but not as much as hoped.

It looks like the overhead of supporting exceptions was the main reason that it wasn't significantly faster.

I looked into the possibility of replacing exceptions within CGAL, but it seems like it may be infeasible.

I was concerned about this, but it looks like llvm has added support for native wasm exceptions, and these should become available via emscripten in a few months -- hopefully with this it should become much faster.

I haven't benchmarked this change in x86, but can do so if you like -- under wasm I don't think we can produce a meaningful comparison.

mglisse:
Yes, I was thinking of a bench on x86_64. It would allow a comparison with the current code, and it would give some idea of what to expect compared to no-intervals once exceptions are properly supported on wasm.

pentacular:
I'm probably missing something completely obvious, but I don't see any obvious way to run a benchmark suite.
I do see per package benchmarks, but I don't see suites for those either.

Should I run some ad hoc benchmarks (if so, which packages do you think would be most useful)?
Or is there some benchmark documentation that I should have found? :)

mglisse:
Indeed, there is no benchmark suite. Some of the per package benchmarks are outdated and don't even compile...
I think anything that spends a lot of time in Interval_nt would do.

lrineau:
> Indeed, there is no benchmark suite. Some of the per package benchmarks are outdated and don't even compile...
> I think anything that spends a lot of time in Interval_nt would do.

I think in CGAL we should treat benchmarks directories as demos directories: 
  - compiled by the daily test suite,
  - but not run.

That would at least force us to update the code of the benchmarks in case of breaking changes.

(Cc @CGAL/geometryfactory)

pentacular:
Well, I'm having trouble finding a significant difference.

I couldn't find a reasonable benchmark, so I adapted corefinement_mesh_union to run

```
  for (int i = 0; i < 2000; i++) {
    Mesh working_mesh1(mesh1);
    Mesh working_mesh2(mesh2);
    Mesh out;
    valid_union = PMP::corefine_and_compute_union(working_mesh1, working_mesh2, out);
  }

```

time ./corefinement_mesh_union
Union was successfully computed

real    0m23.875s
user    0m23.869s
sys     0m0.002s

time ./corefinement_mesh_union_nearest
Union was successfully computed

real    0m23.757s
user    0m23.735s
sys     0m0.026s

time ./corefinement_mesh_union
Union was successfully computed

real    0m23.048s
user    0m23.009s
sys     0m0.036s

time ./corefinement_mesh_union_nearest
Union was successfully computed

real    0m23.030s
user    0m23.012s
sys     0m0.022s

I then went and broke FPU.h inside a CGAL_ALWAYS_ROUND_TO_NEAREST just to make sure corefinement_mesh_union_nearest was using it.

Let me know if you have a better idea for measuring it, but it doesn't seem like there's much of a hit.

mglisse:
> Well, I'm having trouble finding a significant difference.

I am not sure how you tested exactly, but by default, we have CGAL_USE_SSE2 which enables a different code path that avoids CGAL_IA_ADD, etc. By the way, the combination CGAL_USE_SSE2 and CGAL_ALWAYS_ROUND_TO_NEAREST is likely broken in your branch.

Anyway, running `Triangulation3/simple_2 100000` compiled with `-DCGAL_NO_STATIC_FILTERS -DCGAL_NO_STRUCTURAL_FILTERING` so the time is actually spent in intervals, I get .52s for normal CGAL, 1.24s if I disable CGAL_USE_SSE2, and 5.1s if I define CGAL_ALWAYS_ROUND_TO_NEAREST.

pentacular:
Ah, you're absolutely right.

time ./corefinement_mesh_union_nearest
Union was successfully computed

real    0m30.595s
user    0m30.512s
sys     0m0.038s

$ time ./corefinement_mesh_union

Union was successfully computed

real    0m24.609s
user    0m23.882s
sys     0m0.070s

$ time ./corefinement_mesh_union_nearest 
Union was successfully computed

real    0m32.082s
user    0m31.778s
sys     0m0.013s

$ time ./corefinement_mesh_union

Union was successfully computed

real    0m23.572s
user    0m23.348s
sys     0m0.023s

So, CGAL_ALWAYS_ROUND_TO_NEAREST seems to take about 130% as long for this.

That makes much more sense.

Thanks. :)

Number_types/include/CGAL/FPU.h
```diff
@@ -114,9 +114,13 @@ extern "C" {
 
 // Only define CGAL_USE_SSE2 for 64 bits where malloc has a suitable
 // alignment, 32 bits is too dangerous.
-#if defined(CGAL_HAS_SSE2) && (defined(__x86_64__) || defined(_M_X64))
+#if defined CGAL_HAS_SSE2
+#if defined __x86_64__ || defined _M_X64
+#if !defined CGAL_ALWAYS_ROUND_TO_NEAREST
```
mglisse, line 119: What's wrong with `&&`? You can still write it on several lines if you find it more readable (ending a line with `\`).

pentacular: Sounds good.

mglisse:
Later it may be possible to experiment with various things:
* replace nextafter with a function that takes a single argument and always rounds up (and forbid NaN input?), to reduce a bit the work it needs to do
* replace nextafter with something rougher, maybe multiplying with 1±Ɛ and adding something small (for denormals), making sure that it remains valid
* compute an error term (using `sum-max-min` for addition, `fma(a,b,-prod)` for multiplication, etc) and check the sign of the error before calling nextafter. This is extra computations, but nextafter would be called 2 times less often, and the resulting interval would be tighter.

pentacular:
Yes, I think it makes sense to check those out.
Do you mean prior to merging the PR or as follow up work?

I'm a little unclear on the review process here.
It looks as though reviews are resolved, and perhaps I am only blocked on CLA?

I've emailed @lrineau about the CLA, so hopefully we can sort it out shortly.
Please let me know if there is anything else I should be doing. :)

lrineau:
> Yes, I think it makes sense to check those out.
> Do you mean prior to merging the PR or as follow up work?

In my opinion it could be a follow up work. Let's not block this PR because of that. Marc will probably confirm he also think that as a follow up work in a later pull-request.

> I'm a little unclear on the review process here.
> It looks as though reviews are resolved, and perhaps I am only blocked on CLA?

Your pull-request is of good quality:
  - you followed our coding conventions, and the reviews from Marc Glisse,
  - and you have added two tests (and I appreciate that a lot).

> I've emailed @lrineau about the CLA, so hopefully we can sort it out shortly.

I have forwarded the CLA issue to @afabri. Hopefully this can be sorted our soon. Meanwhile your pull-request will be added in our daily testsuite, to verify that is does not add any regression on our testing plaforms (see for example https://cgal.geometryfactory.com/CGAL/testsuite/results-5.2.1-I-75.shtml).

mglisse:
>  Do you mean prior to merging the PR or as follow up work?

follow up

> It looks as though reviews are resolved,

I think you are still calling fesetround (or equivalent) unnecessarily.

mglisse:
Laurent, do you intend to add a platform with the macro defined in the daily testsuite?
The patch touches some CGAL_expensive_assertion, but I am not sure those are ever tested either, as mentioned in another PR.

lrineau:
> Laurent, do you intend to add a platform with the macro defined in the daily testsuite?

No, but Brian has smartly added two tests that defines that macro:
- https://github.com/CGAL/cgal/pull/5510/files#diff-7c22291ae8b98ac7cde6a5c86768e695c6c03bcbc0379c45f5c4e59902f5fca8
- https://github.com/CGAL/cgal/pull/5510/files#diff-59c59e495c4da3e060e8ed38154088eb6e343983cb99fa9494c25cb00e1498e3

> The patch touches some CGAL_expensive_assertion, but I am not sure those are ever tested either, as mentioned in another PR.

I do not want to test `CGAL_expensive_assertion` in the testsuite, even in one column, because there is no way to know how CPU time that would require. I think there are quadratic algorithms in those assertions, and they would be used with non-trivial sizes of inputs.

sloriot:
@lrineau the best way to know is to give it a try. Worst case there won't be any result on that platform. I can add it on Cezanne if you want.

lrineau:
> @lrineau the best way to know is to give it a try. Worst case there won't be any result on that platform. I can add it on Cezanne if you want.

You can try. I am pretty sure that will consume more than a day of CPU: those expansive assertions are nice to use on a specific input, for one .cpp, but for the general CGAL testsuite I am certain that will be too much.

sbrian23:
> > Do you mean prior to merging the PR or as follow up work?
> 
> follow up

Sounds good. :)

> 
> > It looks as though reviews are resolved,
> 
> I think you are still calling fesetround (or equivalent) unnecessarily.

That's a good point.

I've replaced the rounding mode operations with assertions.

Please let me know if that looks right.

Number_types/include/CGAL/FPU.h
```diff
@@ -486,9 +486,15 @@ inline
 FPU_CW_t
 FPU_get_cw (void)
 {
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
+    FPU_CW_t cw;
```
lrineau, line 490: To avoid warnings about unused variables when `CGAL_NDEBUG` is defined:
```suggestion
    CGAL_assertion_code(FPU_CW_t cw;)
```

sbrian23: Ah, cool -- I didn't notice CGAL_assertion_code.

Thanks. :)

afabri:
> I have forwarded the CLA issue to @afabri.

Hello @sbrian23 can you contact me by mail please andreas.fabri@geometryfactory.com

mglisse:
Do we want to document this somewhere? Auto-enable it in cmake if the target is wasm? Or keep it private for now?

lrineau:
> Do we want to document this somewhere? Auto-enable it in cmake if the target is wasm? Or keep it private for now?

Can CMake target wasm?

mglisse:
> Can CMake target wasm?

More or less: https://gitlab.kitware.com/cmake/cmake/-/issues/18423 https://emscripten.org/docs/compiling/Building-Projects.html?highlight=cmake

Number_types/test/Number_types/Interval_nt.cpp
```diff
@@ -139,13 +164,23 @@ bool overflow_test()
     DEBUG( std::cout << "f = " << f << std::endl; )
   }
 
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST_XX
```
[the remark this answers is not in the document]
sbrian23: Sorry, I forgot to remove that code after updating the tests.

Thanks for spotting it.

sbrian23:
Do we want to add cmake support at this point, or is that reasonable for a follow up?

I guess that's really about adding support for emscripten builds, so I'm not sure of the scoping.

mglisse:
It is ok to do it later, I guess we can keep the macro private for now, just for the lucky few who happen to notice its existence.

Number_types/include/CGAL/FPU.h
```diff
@@ -487,28 +503,43 @@ inline
 void
 FPU_set_cw (FPU_CW_t cw)
 {
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
+  CGAL_assertion(cw == CGAL_FE_TONEAREST);
```
mglisse, line 507: Should we make this a CGAL_expensive_assertion or just CGAL_assertion? Is it redundant with Checked_protect_FPU_rounding?

lrineau: If it is redundant, then remove the assertion. Otherwise, that is a regular expression. There is nothing expensive in comparing two words.

mglisse: I should have made this comment on the previous one, which has CGAL_IA_GETFPCW (in Checked_protect_FPU_rounding we use expensive for that).

sbrian23: Hmm.

My reasoning was (a) == is cheap, and (b) there are quite a lot of direct calls to FPU_set_cw().

For example, there's a whole bunch of calls in Apollonius_graph_2/compare_quadratic.h

With the assert we can be confident that only code that expects round-to-nearest will compile given CGAL_ALWAYS_ROUND_TO_NEAREST, which means that code that expects a different rounding mode should be easy to identify and correct incrementally.

sbrian23:
> It is ok to do it later, I guess we can keep the macro private for now, just for the lucky few who happen to notice its existence.

Sounds good -- emscripten support in cmake would be nicer than what I'm doing at the moment.

So I'll follow up on it. :)

Number_types/include/CGAL/FPU.h
mglisse: Thanks for the explanation.

mglisse (approved):

sbrian23:
Thanks for the review, @mglisse 

I'm wondering what the next steps are -- I guess we're waiting for @sloriot and @lrineau to complete their reviews?

I'm also wondering if this PR is correctly targeting CGAL:master, or if it should be targeting a different branch. :)

lrineau (approved):
I also approve this pull-request.

I am very glad CGAL eventually has a way to escape the need for setting the rounding-mode. That has been wanted... for a long time. Even before web assembly, CGAL was unable to run correctly under partial CPU emulation like Valgrind. With `CGAL_ALWAYS_ROUND_TO_NEAREST`, it will be possible.

Thanks a lot to @pentacular for the contribution, and @mglisse for the review.

lrineau:
> I'm wondering what the next steps are -- I guess we're waiting for @sloriot and @lrineau to complete their reviews?

This pull-request is being tested daily in [our testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-81.shtml) and it will be merged in 

> I'm also wondering if this PR is correctly targeting CGAL:master, or if it should be targeting a different branch. :)

`<CGAL/FPU.h>` and `<CGAL/Interval_nt.h>` are very central and sensitive headers files, for CGAL. I prefer to have them tested as much as possible, in particular in a beta release. CGAL-5.3-beta1 is due at the end of April 2021.

mglisse:
> I'm wondering what the next steps are

Unless some issue appears, you are done for this PR, you can just wait. But you are still free to work on a next PR in parallel, based on this one, that improves performance, or adds some documentation, or tweaks the build system, or whatever.

sbrian23:
Sounds good -- I just wanted to make sure I had the procedure correct. :)

afabri:
Is it possible that this PR is responsible for the failure of the [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-85/Number_types/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz )today ? Shall @danston have a look?

mglisse:
> Is it possible that this PR is responsible for the failure of the [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-85/Number_types/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz)today ?

Could be, since it touches those files, although it wasn't supposed to affect that target.

By the way, is there any non-x86_64 platform in the testsuite? At least one arm (32 or 64) or x86 (32) would be nice to check that the path without CGAL_USE_SSE2 is not broken.

lrineau:
> > Is it possible that this PR is responsible for the failure of the [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-85/Number_types/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz)today ?
> 
> Could be, since it touches those files, although it wasn't supposed to affect that target.
> 
> By the way, is there any non-x86_64 platform in the testsuite? At least one arm (32 or 64) or x86 (32) would be nice to check that the path without CGAL_USE_SSE2 is not broken.

We have a Linux Fedora running 32 bits, in the testsuite, but it seems to show up quite late in the day.

We also have a new Apple M1 machine (ARM64) running MacOS BigSur. But that one seems to be off for about a week.

I will check with @maxGimeno...

sbrian23:
I updated master from upstream and merged into a sub-branch to test.

On my machine, at least, both Interval_nt and Interval_nt_nearest pass.
So I guess it is either due to an interaction from another branch merged in for testing, or due to build options.
Or I've done something wrong in verification. :)

Would it be difficult to run the test suite with just this branch patched in?

danston:
I have tested this PR in isolation on my x86_64 Apple machine and I reproduce the error from the test suite. So it is not caused by merging with other branches. Both tests `bench_interval` and `Interval_nt` fail as in the test suite. Change of the build options does not help either. The error comes from the files `FPU.h` and `Interval_nt.h`. I am not familiar with the modifications, but the error seems to come from the part that removes these lines:

```
#define CGAL_IA_SQRT(a) \
        CGAL_IA_FORCE_TO_DOUBLE(CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(a)))
```

in the file `FPU.h`. By naively adding this macro back and undoing the corresponding modifications in the `Interval_nt.h`, I fixed the errors in both tests.

It is not a solution obviously but rather an idea of where to look to fix the actual error. Please let me know if more testing on my platform should be done.

Number_types/include/CGAL/FPU.h
```diff
@@ -579,6 +612,23 @@ inline void force_ieee_double_precision()
...
+inline double IA_sqrt_up(double a) {
+  return IA_up(CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(a)));
+}
+
+inline double IA_sqrt_toward_zero(double d) {
+#ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
+  return (d > 0.0) ? nextafter(std::sqrt(d), 0.) : 0.0;
+#else
+  FPU_set_cw(CGAL_FE_DOWNWARD);
+  double i = (d > 0.0) ? CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(d)) : 0.0;
```
mglisse, line 626: Did you lose CGAL_IA_FORCE_TO_DOUBLE somewhere?

sbrian23: Yes, I think so.

I went over the substitutions manually and have included them in the code as temporary notes.

I found one CGAL_IA_FORCE_TO_DOUBLE that looks like it was omitted.

sbrian23:
Thanks for isolating the issue.

I went over the code and I've put in a tentative fix, and added some temporary notes.

Please let me know how it goes on the x86_64 Apple target.

danston:
Ok, now, it seems to work locally on my x86_64 machine. Both tests succeed.

maxGimeno:

mglisse:
With -DCGAL_ALWAYS_ROUND_TO_NEAREST on master, the test Interval_nt_new.cpp fails because of _test_algebraic_structure.h `assert( CGAL_NTS square(  AS (23)) ==  AS (23*23) );`, and indeed I don't see how it could work with nextafter (adding `possibly` would help, but there are many other strict tests around there). Is that test supposed to be skipped? Or is CGAL_ALWAYS_ROUND_TO_NEAREST considered experimental enough that it does not need to pass the testsuite?

sbrian23:
I guess the test expectations should consider CGAL_ALWAYS_ROUND_TO_NEAREST?

lrineau:
> With -DCGAL_ALWAYS_ROUND_TO_NEAREST on master, the test Interval_nt_new.cpp fails because of _test_algebraic_structure.h `assert( CGAL_NTS square( AS (23)) == AS (23*23) );`, and indeed I don't see how it could work with nextafter (adding `possibly` would help, but there are many other strict tests around there). Is that test supposed to be skipped? Or is CGAL_ALWAYS_ROUND_TO_NEAREST considered experimental enough that it does not need to pass the testsuite?

I think we should move forward and try to make `CGAL_ALWAYS_ROUND_TO_NEAREST` pass the testsuite. @maxGimeno What could be the right platform to test that macro?

maxGimeno:
What about Fedora-rawhide ? It tests almost everything and it is usually all green.

PR #5513, opened 2021-03-05
Polyhedron_demo: Move LAS code from item to plugin

maxGimeno:
## Summary of Changes

 - There is a BOOL declaration in both lib3mf and LASTools, that conflict in 3mf_plugin. To prevent it, I moved the LAS code from the item to the LAS_plugin.
 - Some Qt::MidButton were added in Basic_viewer_qt in 5.3, change them in middlebutton to fix a warning

## Release Management
This is based on master because the CGAL_IO and the CMake modernization makes it really hard to have a single coherent fix base don an old release, and I don't think anyone uses lib3mf except me, but if asked I can do 2 other PRs for 5.1 and 5.2


maxGimeno:

PR #5516, opened 2021-03-08
Update doc for vcpkg use

maxGimeno:
## Summary of Changes
Since we now use gmp instead of mpir in vcpkg, we need to update the doc section about that, and also specify another step to be able to build gmp in 64 bits on windows (becaus eof a bug in vcpkg. See https://github.com/microsoft/vcpkg/issues/14703) 

## Release Management


maxGimeno:

PR #5486, opened 2021-02-22
Performance robin hood

GilesBathgate:
## Summary of Changes

This  pull request provides an alternative approach to bringing in `robin_hood` hashmaps without forking `Unique_hash_map`. As discussed in #5484 

Either `robin_hood::unordered_set` or `robin_hood:unordered_map` are used as needed. Functions from the STL compatible interfaces are used appropriately on a case by case basis.

Double hash lookups are avoided by using various techniques where appropriate.

## Release Management

* Affected package(s): Nef_3/Nef_S2
* Issue(s) solved (if any): fix #5379 
* Feature/Small Feature (if any): performance/speed
Link to compiled documentation: Undocumented internal use only.
License and copyright ownership:
MIT license for additional code from https://github.com/martinus/robin-hood-hashing Copyright (c) 2018-2020 Martin Ankerl http://martin.ankerl.com

PR #5522, opened 2021-03-10
Nef_3: Do not test has_on for a point where we know that it will be true

afabri:
## Summary of Changes

Given two segments of great circles we construct an intersection of the two great circles. In the test if the intersection point lies on the segments we first check if the point is on the great circle.  First we know that, and second this triggers an exact test.

## Release Management

* Affected package(s): Nef_S2, Nef_3

### DIFF
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -49,26 +49,24 @@ namespace CommonKernelFunctors {
 
     int operator()(const Vector_3& vec) const
     {
-      int dir = -1;
       if(certainly_not(is_zero(vec.x()))){
-        dir = 0;
+        return 0;
       } else if(certainly_not(is_zero(vec.y()))){
-        dir = 1;
+        return 1;
       }else if(certainly_not(is_zero(vec.y()))){
-        dir = 2;
+        return 2;
       }
 
-      if(dir == -1){
-        if(! is_zero(vec.x())){
-          return 0;
-        } else if(! is_zero(vec.y())){
-          return 1;
-        } else if(! is_zero(vec.z())){
-          return 2;
-        }
+      if(! is_zero(vec.x())){
+        return 0;
+      } else if(! is_zero(vec.y())){
+        return 1;
+      } else if(! is_zero(vec.z())){
+        return 2;
       }
-      return dir;
-    }
+
+      return -1;
+  }
   };
 
 
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -37,6 +37,41 @@ namespace CGAL {
 
 namespace CommonKernelFunctors {
 
+
+
+  template <typename K>
+  class Non_zero_dimension_3
+  {
+    typedef typename K::Vector_3 Vector_3;
+
+  public:
+    typedef int result_type;
+
+    int operator()(const Vector_3& vec) const
+    {
+      int dir = -1;
+      if(certainly_not(is_zero(vec.x()))){
+        dir = 0;
+      } else if(certainly_not(is_zero(vec.y()))){
+        dir = 1;
+      }else if(certainly_not(is_zero(vec.y()))){
+        dir = 2;
+      }
+
+      if(dir == -1){
+        if(! is_zero(vec.x())){
+          return 0;
+        } else if(! is_zero(vec.y())){
+          return 1;
+        } else if(! is_zero(vec.z())){
+          return 2;
+        }
+      }
+      return dir;
+    }
+  };
+
+
   template <typename K>
   class Are_ordered_along_line_2
   {
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -47,7 +47,7 @@ namespace CommonKernelFunctors {
   public:
     typedef int result_type;
 
-    int operator()(const Vector_3& vec) const
+    result_type operator()(const Vector_3& vec) const
     {
       if(certainly_not(is_zero(vec.x()))){
         return 0;
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -849,7 +849,24 @@ public:
           _CGAL_NEF_TRACEN("found on facet...");
           return make_object(f);
         }
-        if( is.does_intersect_internally(s,f,ip)) {
+
+        // We next check if v is a vertex on the face to avoid a geometric test
+        bool v_vertex_of_f = false;
+        Halffacet_cycle_iterator fci;
+        for(fci=f->facet_cycles_begin(); fci!=f->facet_cycles_end(); ++fci) {
+          if(fci.is_shalfedge()) {
+            SHalfedge_around_facet_circulator sfc(fci), send(sfc);
+            CGAL_For_all(sfc,send) {
+              if(sfc->source()->center_vertex() ==  v){
+                v_vertex_of_f = true;
+                break;
+              }
+            }
+          }
+        }
+
+
+        if( (! v_vertex_of_f) &&  is.does_intersect_internally(s,f,ip) ) {
           s = Segment_3(p, normalized(ip));
           result = make_object(f);
         }
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -853,7 +853,7 @@ public:
         // We next check if v is a vertex on the face to avoid a geometric test
         bool v_vertex_of_f = false;
         Halffacet_cycle_iterator fci;
-        for(fci=f->facet_cycles_begin(); fci!=f->facet_cycles_end(); ++fci) {
+        for(fci=f->facet_cycles_begin(); (! v_vertex_of_f) && (fci!=f->facet_cycles_end()); ++fci) {
           if(fci.is_shalfedge()) {
             SHalfedge_around_facet_circulator sfc(fci), send(sfc);
             CGAL_For_all(sfc,send) {
--- a/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
+++ b/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
@@ -73,12 +73,12 @@ Bounded_side bounded_side_3(IteratorForward first,
   auto apv = approx(plane.orthogonal_vector());
 
   int dir = 0;
-  auto max = CGAL::abs(apv.x());
-  if(CGAL::abs(apv.y()) > max){
+  double max = CGAL::abs(apv.x().sup());
+  if(CGAL::abs(apv.y().sup()) > max){
     dir = 1;
-    max = CGAL::abs(apv.y());
+    max = CGAL::abs(apv.y().sup());
   }
-  if(CGAL::abs(apv.z()) > max){
+  if(CGAL::abs(apv.z().sup()) > max){
     dir = 2;
   }
 
--- a/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
+++ b/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
@@ -99,8 +99,9 @@ Bounded_side bounded_side_3(IteratorForward first,
        we don't need to care about the plane orientation */
   }
 
-
-
+  typename R::Non_zero_dimension_3 non_zero_dimension_3;
+  int dir = non_zero_dimension_3(plane.orthogonal_vector());
+#if 0
   auto apv = approx(plane.orthogonal_vector());
 
   int dir = 0;
@@ -112,7 +113,7 @@ Bounded_side bounded_side_3(IteratorForward first,
   if(CGAL::abs(apv.z().sup()) > max){
     dir = 2;
   }
-
+#endif
 
   CGAL_assertion(!plane.is_degenerate());
   Point_2 (*t)(const Point_3&);
--- a/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
+++ b/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
@@ -82,7 +82,6 @@ Bounded_side bounded_side_3(IteratorForward first,
                             typename R::Plane_3 plane = typename R::Plane_3(0,0,0,0)) {
   typedef typename R::Point_2 Point_2;
   typedef typename R::Point_3 Point_3;
-  typedef typename R::Vector_3 Vector_3;
   typedef typename R::Plane_3 Plane_3;
 
   if(plane == Plane_3(0,0,0,0)) {
@@ -101,19 +100,7 @@ Bounded_side bounded_side_3(IteratorForward first,
 
   typename R::Non_zero_dimension_3 non_zero_dimension_3;
   int dir = non_zero_dimension_3(plane.orthogonal_vector());
-#if 0
-  auto apv = approx(plane.orthogonal_vector());
 
-  int dir = 0;
-  double max = CGAL::abs(apv.x().sup());
-  if(CGAL::abs(apv.y().sup()) > max){
-    dir = 1;
-    max = CGAL::abs(apv.y().sup());
-  }
-  if(CGAL::abs(apv.z().sup()) > max){
-    dir = 2;
-  }
-#endif
 
   CGAL_assertion(!plane.is_degenerate());
   Point_2 (*t)(const Point_3&);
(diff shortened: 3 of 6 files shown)


afabri:
@GilesBathgate I would be curious to see the performance impact for RapCAD and  Boolean operations on real world data (which you might share).  For an intersection of two triangulated spheres we gain ~40%.

GilesBathgate:
It's definitely an improvement but I only get 14% 

I use the following cad script:
($fn=128 specifies the number of segments in the great circle of the sphere)

```scad
intersection(){
  sphere(d=20,$fn=128);
  translate([5,5,5])sphere(d=20,$fn=128);
}
```

RapCAD Before: CGAL5.0.3
```
rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 14s 115ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 13s 940ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 14s 177ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 14s 69ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 14s 100ms.
Total export time: 0h 0m 0s 0ms.
```
RapCAD After: CGAL5.0.3-patched
```
rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 12s 391ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 12s 412ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 12s 365ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 12s 385ms.
Total export time: 0h 0m 0s 0ms.

rapcad cad/intersection.rcad -o null
Total compiling time: 0h 0m 12s 387ms.
Total export time: 0h 0m 0s 0ms.
```

((14-12)/14)*100 = 14%

afabri:
Do you do the translation with the `Exact_predicates_exact_constructions_kernel`?

GilesBathgate:
I try to translate the points before construction of the Nef when possible, but yes always using `Exact_predicates_exact_constructions_kernel`: https://github.com/GilesBathgate/RapCAD/blob/master/src/cgalprimitive.cpp#L640

GilesBathgate:
@afabri I forgot to mention, I construct the sphere twice, instead of constructing it once and then translate. I do have experimental cache support which takes the whole thing down to about 10s. but then this wouldn't work with the following:
```scad
intersection(){
  sphere(d=20,$fn=128);
  translate([5,5,5])sphere(d=30,$fn=128);
}
```
Since then it is two different sized speheres.

afabri:
I loaded two spheres as off files. Also for the moment I avoid that the vertices of one sphere are on an edge/face/vertex of the other sphere.

GilesBathgate:
@afabri I have to admit, most of these optimisations are way beyond my understanding of the code. But I am happy to test and feedback any results from profiling.

Nef_3/include/CGAL/Nef_3/bounded_side_3.h
```diff
@@ -70,21 +70,27 @@ Bounded_side bounded_side_3(IteratorForward first,
 
 
 
+  auto apv = approx(plane.orthogonal_vector());
+
+  int dir = 0;
```
GilesBathgate, line 75:
```c++
int dir = CGAL::abs(apv[0]) > CGAL::abs(apv[1]) ? 0 : 1;
    dir = CGAL::abs(apv[2]) > CGAL::abs(apv[dir]) ? 2 : dir;
```
Is this equivalent?

Nef_3/include/CGAL/Nef_3/bounded_side_3.h
```diff
@@ -30,17 +30,17 @@
 namespace CGAL {
 
 template <class Point_2, class Point_3>
-Point_2 point_3_get_x_y_point_2(Point_3 p) {
+Point_2 point_3_get_x_y_point_2(const Point_3& p) {
```
GilesBathgate, line 33: Would it be worthwhile to use `Projection_traits_xy_3` et al?

GilesBathgate:
@afabri I had an idea for an improvement in `SNC_k3_tree_traits.h`

```c++

if( CGAL::assign( v, o)) {
    if(!OnSideMap.is_defined(v))
      OnSizeMap[v] = (*this)(pop, v, depth);
    return OnSideMap[v];
  } else if( CGAL::assign( e, o)) {
    if(!OnSideMap.is_defined(e))
      OnSideMap[e] = (*this)(pop, e, depth);
    return OnSideMap[e];
  } else if( CGAL::assign( f, o)) {
    if(!OnSideMap.is_defined(f))
      OnSideMap[f] = (*this)(pop, f, depth);
    return OnSideMap[f];
    
 ```
  But I couldn't get this to work. Is it a bad idea?

**Comment by afabri** [comment]
What lines are you referring to with the above code?  Does it show up in a perfomance tool?

File: Nef_3/include/CGAL/Nef_3/bounded_side_3.h (hunk 1 above)
Comment (afabri) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h:75]: It seems to be equivalent, but what is the goal?  Elegance, compactness, performance?

File: Nef_3/include/CGAL/Nef_3/bounded_side_3.h (hunk 2 above)
Comment (afabri) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h:33]: It would be worthwhile not to use `hw()`  for Cartesian coordinates (as you did in another place.

File: Nef_3/include/CGAL/Nef_3/bounded_side_3.h (hunk 1 above)
Comment (GilesBathgate) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h:75]: Yeah, compactness. The performance is likely identical. it may even generate the same asm.

**Comment by GilesBathgate** [comment]
> What lines are you referring to with the above code? Does it show up in a perfomance tool?

Its here: 
https://github.com/CGAL/cgal/blob/74af3e7d33e7354e4090a1b17941a3f0c372ca8d/Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h#L352-L359

It's not very obvious in the performance analyser because its at the leaf of many K3_tree recursive function calls. But I noticed a performance boost when I applied a robin_hood hashmap to the OnSideMap code in this area.

**Comment by afabri** [comment]
> 
> 
> @afabri I have to admit, most of these optimisations are way beyond my understanding of the code. But I am happy to test and feedback any results from profiling.

Let me explain you, starting with the title of the PR, and even independantly from Nef polyhedra.  When Epeck constructs a  midpoint `mp` of a segment `seg`, it does that with interval arithmetic, and the lower and upper bound of these intervals have a good chance not to be identical. When we now make the test  `seg.has_on(mp)` we first perform this test with interval arithmetic, and if it is undecidable, an exact version of `mp` is computed using `Gmpq` ratioanal numbers.   We call that a filter failure.  What I mean with undecidable with interval arithmetic is the following:   Imagine you have an interval and want to know if it is positive or negative. If  the lower bound is positive, or the  upper bound is negative we know, otherwise we don't. 
In the Nef polygon we had several time such a situation, and instead of checking if a point is on a face (with a plane/point test), I check if the vertex  holding the point is as vertex on the face.

**Comment by GilesBathgate** [comment]
@afabri Yes regarding intervals I found this code:

https://github.com/CGAL/cgal/blob/74af3e7d33e7354e4090a1b17941a3f0c372ca8d/Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h#L67-L95

Which I thought might be faster performance with:

```c++
CGAL::Comparison_result operator()( const Point_3 p1, const Point_3 p2) {
    std::pair<double, double> i1;
    std::pair<double, double> i2;
    switch(coord) {
        case 0:
        {
            i1 = CGAL::to_interval(p1.x());
            i2 = CGAL::to_interval(p2.x());
            break;
        }
        case 1:
        {
            i1 = CGAL::to_interval(p1.y());
            i2 = CGAL::to_interval(p2.y());
            break;
        }
        case 2:
        {
            i1 = CGAL::to_interval(p1.z());
            i2 = CGAL::to_interval(p2.z());
            break;
        }
        default: CGAL_error();
    }
    if(i1.second < i2.first)
      return CGAL::SMALLER;
    else if(i2.second < i1.first)
      return CGAL::LARGER;
    return CGAL::EQUAL;
  }
  ```
  
  But then I found that this code isn't actually used with `Exact_predicates_exact_constructions_kernel` But then even by manually forcing the template specialisation this code isn't any better. That is where I struggle. Perhaps I should pay more attention to the profiler :joy:

File: Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h [hunk 3]
```diff
@@ -257,7 +257,7 @@ class SM_point_locator : public SM_decorator {
     Unique_hash_map<SHalfedge_handle,bool> visited(false);
```
GilesBathgate, line 257: I am wondering if this change: https://github.com/CGAL/cgal/pull/5486/commits/f72a006cd5cb328a740dd5e4bb920d86a9dffd77 makes a noticeable difference in your benchmarks.

Nef_3/include/CGAL/Nef_3/bounded_side_3.h
```diff
@@ -70,21 +70,27 @@ Bounded_side bounded_side_3(IteratorForward first,
...
 
+  auto apv = approx(plane.orthogonal_vector());
+
+  int dir = 0;
+  auto max = CGAL::abs(apv.x());
+  if(CGAL::abs(apv.y()) > max){
+    dir = 1;
+    max = CGAL::abs(apv.y());
+  }
+  if(CGAL::abs(apv.z()) > max){
```
GilesBathgate, line 81: @maxGimeno I get the exception: terminate called after throwing an instance of 'CGAL::Uncertain_conversion_exception' what(): Undecidable conversion of CGAL::Uncertain<T>

The stack trace shows the exception coming from line 81 of bounded_side_3.h

GilesBathgate: @afabri Would this approach have the performance benefit:

```c++
  CGAL_assertion(!plane.is_degenerate());
  Point_2 (*t)(const Point_3&);
  Vector_3 pv(plane.orthogonal_vector());
#if CGAL_USE_TRACE
  Vector_3 pxy(0,0,1), pyz(1,0,0), pzx(0,1,0);
#endif
  CGAL_NEF_TRACEN("pv*pxz: "<<pv*pzx);
  CGAL_NEF_TRACEN("pv*pyz: "<<pv*pyz);
  CGAL_NEF_TRACEN("pv*pxy: "<<pv*pxy);
  if( !CGAL_NTS is_zero(pv.y()) )
    /* the plane is not perpendicular to the ZX plane */
    t = &point_3_get_z_x_point_2< Point_2, Point_3>;
  else if( !CGAL_NTS is_zero(pv.x()) )
    /* the plane is not perpendicular to the YZ plane */
    t = &point_3_get_y_z_point_2< Point_2, Point_3>;
  else {
    CGAL_assertion( !CGAL_NTS is_zero(pv.z()) );
    /* the plane is not perpendicular to the XY plane */
    t = &point_3_get_x_y_point_2< Point_2, Point_3>;
  }
  ```

Comment (GilesBathgate) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h:81]: Hmm, although `is_zero` uses exact for the calculation what we actually want is `is_non_zero` which I think could use interval math?

Comment (GilesBathgate) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h:81]: I think using simply `pv.y() != 0`  instead of `!CGAL_NTS is_zero(pv.y())` achieves this for intervals.

Comment (afabri) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h]: My mistake was to compare intervals.   In case they overlap, that is for exampe   `a.inf() < b.inf() < a.sup()` the exception is thrown.

Comment (afabri) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h]: > I think using simply `pv.y() != 0` instead of `!CGAL_NTS is_zero(pv.y())` achieves this for intervals.

`is_zero()`   is a better choice. For example with a number type like `Gmpq`, it is pointless to constuct the quotient `0/1` and to do the crosswise multiplication with the denominators, but it is enough to test if the numerator is zero.

Comment (GilesBathgate) [reply Nef_3/include/CGAL/Nef_3/bounded_side_3.h]: Oh my solution doesn't take `approx(plane.orthogonal_vector())` Well do you want me to separate out the homogeneous/cartesian changes I did?

File: Nef_3/include/CGAL/Nef_3/bounded_side_3.h [hunk 5]
```diff
@@ -70,21 +70,27 @@ Bounded_side bounded_side_3(IteratorForward first,
 
 
 
+  auto apv = approx(plane.orthogonal_vector());
+
+  int dir = 0;
+  double max = CGAL::abs(apv.x().sup());
```
lrineau, line 76: Is that granted that the kernel used by this code is always a filtering kernel from CGAL (either `Epick` or `Epeck`)?

What will happen if the user chooses `CGAL::Simple_cartesian<CGAL::Gmpq>` as the kernel?

afabri: It is even wrong for `Epick` as the intervals of the three coordinates might all contain 0.   I'll fix it today.

GilesBathgate: Is it possible to make it a fuctor of the Kernel or is that overkill?

afabri: It is not an overkill but the only clean way to do it.   I  just added a functor that finds any dimension different from zero  (that's also what was done before),

GilesBathgate:
> why not return directly?

I agree wouldn't it just be:
```c++
      if(certainly_not(is_zero(vec.x()))){
        return 0;
      } else if(certainly_not(is_zero(vec.y()))){
        return 1;
      }else if(certainly_not(is_zero(vec.y()))){
        return 2;
      }

        if(! is_zero(vec.x())){
          return 0;
        } else if(! is_zero(vec.y())){
          return 1;
        } else if(! is_zero(vec.z())){
          return 2;
        }
      
     return -1;
  ```

File: Kernel_23/include/CGAL/Kernel/function_objects.h [hunk 6]
```diff
@@ -37,6 +37,39 @@ namespace CGAL {
...
+
+  template <typename K>
+  class Non_zero_dimension_3
+  {
+    typedef typename K::Vector_3 Vector_3;
+
+  public:
+    typedef int result_type;
+
+    int operator()(const Vector_3& vec) const
```
GilesBathgate, line 50: `result_type operator()(const Vector_3& vec) const`?

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -849,7 +849,24 @@ class SNC_point_locator_by_spatial_subdivision :
...
+        // We next check if v is a vertex on the face to avoid a geometric test
+        bool v_vertex_of_f = false;
+        Halffacet_cycle_iterator fci;
+        for(fci=f->facet_cycles_begin(); fci!=f->facet_cycles_end(); ++fci) {
+          if(fci.is_shalfedge()) {
+            SHalfedge_around_facet_circulator sfc(fci), send(sfc);
+            CGAL_For_all(sfc,send) {
+              if(sfc->source()->center_vertex() ==  v){
+                v_vertex_of_f = true;
+                break;
```
GilesBathgate, line 862: This will only break the inner loop. Perhaps if re-written as a function, a `return true` here would be better?

afabri: I appreciate your careful reading of the code.   I just pushed a fix.

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -849,7 +849,24 @@ class SNC_point_locator_by_spatial_subdivision :
...
+              }
+            }
+          }
+        }
+
+
+        if( (! v_vertex_of_f) &&  is.does_intersect_internally(s,f,ip) ) {
```
lrineau, line 869: The fact that `break` cannot exit more than one loop should give an hint that there should be a function there: `return` can exit any block of code!

```suggestion
        // We next check if v is a vertex on the face to avoid a geometric test
        auto v_vertex_of_f = [&]() {
          for(auto fci=f->facet_cycles_begin(); fci!=f->facet_cycles_end(); ++fci) {
            if(fci.is_shalfedge()) {
              SHalfedge_around_facet_circulator sfc(fci), send(sfc);
              CGAL_For_all(sfc,send) {
                if(sfc->source()->center_vertex() ==  v){
                  return true;
                }
              }
            }
          }
          return false;
        };

        if( (! v_vertex_of_f()) &&  is.does_intersect_internally(s,f,ip) ) {
```

GilesBathgate: I agree (not surprisingly :joy: )

afabri: Is what I wrote plain false, or do you consider the function  just more elegant and modern?

lrineau: The second: what you wrote is correct. The only difference is the extra test of `v_vertex_of_f` in the outer loop. I doubt a performance profiler can see the difference.

GilesBathgate: Some might even say that factoring into a separate function is a good way to break up long blocks of code and make the code self-documenting without the need for additional comments ;) Of course, If performance is the main concern then stylistic considerations like these may need to be ignored.

afabri: If you don't mind I would prefer to keep it as is.

Nef_3/include/CGAL/Nef_3/bounded_side_3.h
```diff
@@ -68,24 +99,31 @@ Bounded_side bounded_side_3(IteratorForward first,
...
+  if(CGAL::abs(apv.y().sup()) > max){
+    dir = 1;
+    max = CGAL::abs(apv.y().sup());
+  }
+  if(CGAL::abs(apv.z().sup()) > max){
+    dir = 2;
+  }
+#endif
```
maxGimeno, line 116: please remove this block

Nef_3/include/CGAL/Nef_3/bounded_side_3.h
```diff
@@ -68,24 +99,31 @@ Bounded_side bounded_side_3(IteratorForward first,
...
 
   CGAL_assertion(!plane.is_degenerate());
-  Point_2 (*t)(Point_3);
-  Vector_3 pv(plane.orthogonal_vector()), pxy(0,0,1), pyz(1,0,0), pzx(0,1,0);
```
maxGimeno, line 75: Removing this line makes the previous typedef useless, leading to these warnings : https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-90.shtml#Convex_decomposition_3_Examples

maxGimeno:

GilesBathgate:
@afabri I am wondering. There are several places in the code where it checks `!x.has_on(p)`. While I understand that determining that a point is on a plane/line etc, can it be an approximate calculation to determine that something is **not** on the plane/line. As such could it be added `x.not_has_on(p)` that would utilise the faster calculation?

Kernel_23/include/CGAL/Kernel/function_objects.h
```diff
@@ -37,6 +37,39 @@ namespace CGAL {
...
+  public:
+    typedef int result_type;
+
+    result_type operator()(const Vector_3& vec) const
+    {
+      if(certainly_not(is_zero(vec.x()))){
+        return 0;
+      } else if(certainly_not(is_zero(vec.y()))){
+        return 1;
+      }else if(certainly_not(is_zero(vec.y()))){
```
GilesBathgate, line 56: @afabri Yikes, this is a bug! should be `certainly_not(is_zero(vec.z()))`

afabri: Thank you for pointing out.

PR #5532, opened 2021-03-17
Remove operator/ from CGAL::Mpzf and implement Do_intersect_3 with RT only (without division)

lrineau:
## Summary of Changes

The operator `operator/(Mpfz,Mpfz)` actually implements integral division. With this commit, it is renamed `division((Mpfz,Mpfz)` unless the macro `CGAL_MPZF_DIVISION_OPERATOR` is defined.

I have commented the line

    #define CGAL_MPZF_DIVISION_OPERATOR 1

and that will breaks a lot of CGAL code in this branch.

The goal of this branch is to experiment to possibility to turn `Do_intersect_3` into a "functor using RT" (without any division). Let's see how far we can go with that.

## Release Management

* Affected package(s): Kernel_23, Number_types, Intersections_2, Intersections_3
* License and copyright ownership: maintenance b GeometryFactory


lrineau:
@mglisse I know that these two commits are for the moment completely broken. @afabri asked me if it was possible to detect overloads of `Do_intersect_3` that are implemented using divisions. This draft pull-request is my answer.

lrineau:
@mglisse Your `struct division` in the CGAL namespace in `NewKernel_d/include/CGAL/NewKernel_d/utils.h` does not help. It does not seem to be used anyway. If it it used anywhere, it seems to be replaceable by [`std::divides<void>`](https://en.cppreference.com/w/cpp/utility/functional/divides_void) (and `CGAL::multiplies` as well, replaceable by [`std::multiplies<void>`](https://en.cppreference.com/w/cpp/utility/functional/multiplies_void)).

mglisse:
> @mglisse Your `struct division` in the CGAL namespace in `NewKernel_d/include/CGAL/NewKernel_d/utils.h` does not help. It does not seem to be used anyway. If it it used anywhere, it seems to be replaceable by [`std::divides<void>`](https://en.cppreference.com/w/cpp/utility/functional/divides_void) (and `CGAL::multiplies` as well, replaceable by [`std::multiplies<void>`](https://en.cppreference.com/w/cpp/utility/functional/multiplies_void)).

It would have been nice if I had used a subnamespace. Well, both multiplies and division seem unused currently, I pre-approve if you want to remove them.

afabri:
I am struggling a bit with the `do_intersect(Plane,Plane,Plane)`.   The best page about how to do it is [this one](http://www.ambrsoft.com/TrigoCalc/Plan3D/3PlanesIntersection_.htm#ExIntersection3Planes).

Kernel_23/include/CGAL/rank.h
```diff
@@ -0,0 +1,164 @@
...
+  }
+  else
+  {
+    return 1 + rank_22(a0*b1-a1*b0, a0*c1-a1*c0, a0*b2-a2*b0, a0*c2-a2*c0);
+  }
+}
+
+
+template <class RT>
+int rank_23(const RT& a0, const RT& b0, const RT& c0,
```
mglisse, line 116: The rank is symmetric, shouldn't one of rank_23/rank_32 just call the other on the transposed matrix? Or is it an optimization because you know that for some uses it is better to reduce by lines or by columns?

sloriot: There is no particular reason, I start to add prefix because I did not want to mess up the way I pass parameters and avoid ambiguities.

afabri:
@lrineau I would like to stop here in order to get it into the next release.

Algebraic_foundations/include/CGAL/number_utils.h
```diff
@@ -324,6 +324,36 @@ NT approximate_sqrt(const NT& nt)
   return approximate_sqrt(nt, Sqrt());
 }
 
+template <class NT>
+Comparison_result
+compare(const NT& xnum, const NT& xden,
+        const NT& ynum, const NT& yden)
```
lrineau, line 330: I think you must rename that function template. Its behavior is not obvious from the name and the signature, and adding an overload to the overload set of `compare` will only degrade compilation performances, without any benefit. Maybe `compare_quotients`?

lrineau:
> @lrineau I would like to stop here in order to get it into the next release.

I agree. The set of changes is self-contained, and the performance gain is really nice for `do_intersect` with degenerate data.

lrineau:
@maxGimeno I suggest you test that PR first in the separate small testsuite.

Intersections_3/test/Intersections_3/do_intersect_3.cpp
```diff
@@ -0,0 +1,114 @@
...
+  Line_3 lin(p,q);
+  Triangle_3 t0(p,q,r), t1(p,q,r);
+  Tetrahedron_3 tet(p,q,r,s);
+  Plane_3 pla(p,q,r);
+  Iso_cuboid_3 ic(p,q);
+  Sphere_3 sp(p,1);
+  CGAL::Bbox_3 bb;
+
+  do_intersect(p,sp);
+  do_intersect(bb,t0);
```
MaelRL, line 28: - `bb` is not initialized
- there is already `test_intersection_3.cpp` that checks `do_intersect` and API calls, why a new file with only a subset of the supported combinations?

Number_types/include/CGAL/Mpzf.h
```diff
@@ -774,7 +776,11 @@ struct Mpzf {
     return res;
   }
 
+#ifndef CGAL_MPZF_DIVISION_OPERATOR
+  friend Mpzf division(Mpzf const&a, Mpzf const&b){
```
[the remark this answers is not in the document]
afabri: Wouldn't it make sense to have a base class without divsion (or the other way round)?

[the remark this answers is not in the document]
mglisse: You would need to redefine all the operations in the derived class so they return the right type, that seems ugly. And I don't clearly see how it would help.

maxGimeno:
here are the results for the small testsuite : 
https://cgal.geometryfactory.com/~mgimeno/test_suite/TESTRESULTS/results-0.0-Ic-23.shtml

Intersections_3/test/Intersections_3/do_intersect_3.cpp
afabri: bb is initialized to a bbox with [ + inf , - inf ] ^3
Is it only a subset?  I had started this file to add tests little by little, just to check that it compiles when divsion enabled. Shall I only keep the triple plane test here?

MaelRL: Just counting the lines, there are fewer tests than in https://github.com/CGAL/cgal/blob/master/Intersections_3/test/Intersections_3/test_intersections_3.cpp#L1232. 

Planex3 has its function: https://github.com/CGAL/cgal/blob/master/Intersections_3/test/Intersections_3/test_intersections_3.cpp#L223.

afabri: I just removed the additional file.

afabri:
Btw, shall we also replace all these comparisons with `0` by `is_zero()`  etc.   This should be better for almost any non built-in number type.

MaelRL:
@mglisse Wasn't there something recent about `square(x)` being much slower than `x*x` for... some case (I can't recall what)? Could this also affect other NT functors like `is_zero`?

mglisse:
> shall we also replace all these comparisons with `0` by `is_zero()` etc

`!is_negative`, etc, yes. (I should implement is_positive and is_negative for Mpzf...)

> Wasn't there something recent about square(x) being much slower than x*x for... some case (I can't recall what)?

That was only for intervals, and only because I had rewritten multiplication and not yet square, but that has been fixed, using `square` is always a good idea.

afabri:
> For types that don't specialize it, it is equivalent to `==FT(0)`, and if some type specializes it to something worse than that (how?), then that specialization should be fixed.

So even for `Gmpq`  to use is_zero()`  is not better?   I would have expected that.

mglisse:
> So even for `Gmpq` to use is_zero()` is not better?

Indeed, currently, for Gmpq, is_zero(q) is equivalent to `q==Gmpq(0)` (worse than writing `q==0`), because nobody has written an Is_zero in the traits for Gmpq. Someone should do that (and Sgn, Compare, Is_positive, etc), it isn't hard, mpq_class and mpq_rational have an Is_zero.
(this is from a quick look at the code, I didn't check generated code to be sure, so I may have missed something)

afabri:
I added data/blade.off which where you can see the performance gain when running the self intersection test.  While it takes 4sec with master it takes  0.6sec with your RT.   
image

Intersections_3/include/CGAL/Intersections_3/internal/intersection_3_1_impl.h
```diff
@@ -1651,29 +1681,131 @@ inline bool
...
 
 
 template <class R>
 inline bool
-do_intersect(const Iso_cuboid_3<R> &i, const Iso_cuboid_3<R> &j, const R&)
+do_intersect(const Iso_cuboid_3<R> &icub1, const Iso_cuboid_3<R> &icub2, const R&)
 {
-  return bool(CGAL::intersection(i, j));
+    typedef typename R::Point_3 Point_3;
+    typedef typename R::Iso_cuboid_3 Iso_cuboid_3;
```
mglisse, line 1721: -Wunused-local-typedefs

Intersections_3/include/CGAL/Intersections_3/internal/intersection_3_1_impl.h
```diff
@@ -971,16 +972,13 @@ do_intersect(const typename K::Plane_3 &plane,
 {
     typedef typename K::Point_3 Point_3;
 
-    typename Intersection_traits<K, typename K::Plane_3, typename K::Line_3>
-      ::result_type
-      line_intersection = internal::intersection(plane, ray.supporting_line(), k);
-
-    if(!line_intersection)
-        return false;
-    if(const Point_3 *isp = intersect_get<Point_3>(line_intersection))
```
mglisse, line 980: Looks like you can remove the typedef for Point_3 then.

mglisse:
I am a bit lost with the static filters, but I notice that Static_filters/Do_intersect_3.h does something completely different for do_intersect(tri1,tri2) and for do_intersect(tri,segment) (one is implemented as a regular predicate, and the other "outside" of the kernel). Is that difference still necessary? It is quite possible that it is, just checking.

lrineau:
> I am a bit lost with the static filters, but I notice that Static_filters/Do_intersect_3.h does something completely different for do_intersect(tri1,tri2) and for do_intersect(tri,segment) (one is implemented as a regular predicate, and the other "outside" of the kernel). Is that difference still necessary? It is quite possible that it is, just checking.

I cannot find the static filter for `Do_intersect_3(tri1,tri2)`. Where is it?

mglisse:
> I cannot find the static filter for `Do_intersect_3(tri1,tri2)`.

That's my point :wink:

afabri:
It is the result of an incremental approach, that is we observed do_intersect of segment/triangle  in vtune and optimized it.

Algebraic_foundations/include/CGAL/number_utils.h
afabri: done.

Algebraic_foundations/include/CGAL/number_utils.h
```diff
@@ -324,6 +324,36 @@ NT approximate_sqrt(const NT& nt)
   return approximate_sqrt(nt, Sqrt());
 }
 
+template <class NT>
+Comparison_result
```
mglisse, line 328: I wonder if it should return something a bit more generic, so that when NT is Interval_nt it would return `Uncertain<Comparison_result>`. The way it is used currently, it probably wouldn't matter much.

Intersections_3/include/CGAL/Intersections_3/internal/intersection_3_1_impl.h
```diff
@@ -1651,29 +1679,130 @@ inline bool
...
+
+  if(! is_zero(determinant(plane1.a(), plane1.b(), plane1.c(),
+                           plane2.a(), plane2.b(), plane2.c(),
+                           plane3.a(), plane3.b(), plane3.c()))){
+    return true;
+  }
+
+  int pcount = 0;
+  bool b12, b13,b23;
+  if(b12 = parallel(plane1,plane2)) pcount++;
```
maxGimeno, line 1692: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]

maxGimeno:
error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-99/Number_types/TestReport_lrineau_ArchLinux-CXX17-Release.gz

Algebraic_foundations/include/CGAL/number_utils.h
lrineau, line 328: Maybe:
```suggestion
 typename Same_uncertainty_nt<Comparison_result, NT>::type
```

along with 
```
#include <CGAL/Kernel/Same_uncertainty.h>
```
at the top of the file.

afabri: Can you just do it.

lrineau: Done in https://github.com/CGAL/cgal/pull/5532/commits/17b254639059bd247700d7c8c7e9fce3ff72dc42.

afabri: Didn't you write that the .h file must also be included?  And how is this change related to the fact that we want to get `Do_intersect_3` division-free?

lrineau: > Didn't you write that the .h file must also be included?

You are right: I forgot. I ran the CTest testsuite, though, so it is probably already included. I will add it anyway for completeness.

>  And how is this change related to the fact that we want to get `Do_intersect_3` division-free?

That change is only a reaction to Marc's comment: https://github.com/CGAL/cgal/pull/5532#discussion_r607908030

maxGimeno:
warning : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-110/Kernel_23/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz

maxGimeno:
error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-111/Intersections_3/TestReport_lrineau_Fedora-with-LEDA.gz

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-113/Intersections_3/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz

maxGimeno:

Number_types/include/CGAL/Mpzf.h
```diff
@@ -909,7 +917,11 @@ struct Mpzf {
   Mpzf& operator+=(Mpzf const&x){ *this=*this+x; return *this; }
   Mpzf& operator-=(Mpzf const&x){ *this=*this-x; return *this; }
   Mpzf& operator*=(Mpzf const&x){ *this=*this*x; return *this; }
+#ifdef CGAL_MPZF_DIVISION_OPERATOR
   Mpzf& operator/=(Mpzf const&x){ *this=*this/x; return *this; }
+#else // not CGAL_MPZF_DIVISION_OPERATOR
+  Mpzf& operator/=(Mpzf const&x){ *this=division(*this,x); return *this; }
```
mglisse, line 923: Shouldn't `operator/=` be disabled when we disable `operator/`?

lrineau: Hi Marc. You seem right. Would you mind write the pull-request?

PR #5533, opened 2021-03-17
Kernel_23: Add a version of Do_intersect_3 where we can use a RT

afabri:
## Summary of Changes

Doing a self intersection test of a data set with many neighboring coplanar triangles the intersection test is done with an exact number type.   As some `do_intersect()` functions perform divisions we cannot simply use the interface macro `Kernel_pred_RT`.  We have to get rid of the divisions first.  This PR is basically a work around and the functor added is an implementation detail and not  meant to be documented.  This faster test is only used in  `PMP::does_self_intersect()`.

## Release Management

* Affected package(s): Kernel_23, Intersections_3, PMP

### DIFF
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -3002,7 +3002,7 @@ namespace CommonKernelFunctors {
   public:
     typedef typename K::Boolean     result_type;
 
-    // There are x combinaisons, so I use a template.
+    // There are x combinations, so I use a template.
     template <class T1, class T2>
     result_type
     operator()(const T1& t1, const T2& t2) const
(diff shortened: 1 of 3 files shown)


lrineau:
[` PMP::does_self_intersect()`](https://doc.cgal.org/5.2/Polygon_mesh_processing/group__PMP__intersection__grp.html#gaf19c80ec12cbff7ebe9e69453f1d40b8]https://doc.cgal.org/5.2/Polygon_mesh_processing/classPMPSelfIntersectionTraits.html) documents that the `geom_traits` parameter must be a model of [`PMPSelfIntersectionTraits`](https://doc.cgal.org/5.2/Polygon_mesh_processing/classPMPSelfIntersectionTraits.html). We'll have to do something about that.

afabri:
I close this PR as #5533 provides a more elegant solution.

sloriot:
@afabri Wrong PR number for the replacing one. #5532?

PR #5547, opened 2021-03-23
Fix warning: std::pow(float, int) returns double instead of float in recent C++

sgiraudot:
## Summary of Changes

`std::pow(float,int)` used to return `float` but now returns `double`, which triggers warnings in our code when using `CGAL::Simple_cartesian<float>` where we assumed that the return type of `std::pow` was always the same as the argument type. This PR adds explicit casts where needed.

## Release Management

* Affected package(s): Shape Detection, PSP
* Based on 5.2 to avoid conflicts (incriminated lines have changed since 5.1)

### DIFF
--- a/Point_set_processing_3/include/CGAL/edge_aware_upsample_point_set.h
+++ b/Point_set_processing_3/include/CGAL/edge_aware_upsample_point_set.h
@@ -437,7 +437,7 @@ edge_aware_upsample_point_set(
   //
   FT cos_sigma = static_cast<FT>(std::cos(FT(CGAL::to_double(sharpness_angle))
                                           / FT(180) * FT(CGAL_PI)));
-  FT sharpness_bandwidth = std::pow((CGAL::max)((FT)1e-8, (FT)1.0 - cos_sigma), 2);
+  FT sharpness_bandwidth = static_cast<FT>(std::pow((CGAL::max)((FT)1e-8, (FT)1.0 - cos_sigma), 2));
 
   FT sum_density = 0.0;
   unsigned int count_density = 1;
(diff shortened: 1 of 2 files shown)


Point_set_processing_3/include/CGAL/edge_aware_upsample_point_set.h
```diff
@@ -437,7 +437,7 @@ edge_aware_upsample_point_set(
   //
   FT cos_sigma = static_cast<FT>(std::cos(FT(CGAL::to_double(sharpness_angle))
                                           / FT(180) * FT(CGAL_PI)));
-  FT sharpness_bandwidth = std::pow((CGAL::max)((FT)1e-8, (FT)1.0 - cos_sigma), 2);
+  FT sharpness_bandwidth = static_cast<FT>(std::pow((CGAL::max)((FT)1e-8, (FT)1.0 - cos_sigma), 2));
```
lrineau, line 440: I suggest to use `CGAL::square` instead of `std::pow(..., 2)`.

sgiraudot: Good point, I grepped the occurrences and replaced it everywhere.

afabri:
I check_headers not happy, because using `square()` introduces dependencies to add?

maxGimeno:
No, there is an actual error :

sgiraudot:
I'm on it.

maxGimeno:
error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-90/Surface_mesh_segmentation/TestReport_lrineau_Fedora-rawhide.gz

maxGimeno:

PR #5548, opened 2021-03-23
PCA: do not compile examples/tests without Eigen

sgiraudot:
## Summary of Changes

PCA depends on Eigen, but the CMakeLists still try to compile the examples and tests without it (leading to [warnings](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-88/Principal_component_analysis_Examples/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz) as without Eigen, the package fallbacks on an internal implementation that is deprecated). This PR fixes that.

## Release Management

* Affected package(s): PCA
* Based on 5.2 to avoid conflicts as the CMakeLists changed since 5.1


maxGimeno:

PR #5549, opened 2021-03-23
Fix Eigen in CMakeLists 5.1

maxGimeno:
## Summary of Changes
If Eigen is not found, Mesh_3 examples and some Tetrahedral_remeshing examples are broken instead of not being ran.

## Release Management

* Affected package(s):Mesh_3, Tetrahedral_remeshing


maxGimeno:
@lrineau This is the 5.1 version of #5550 which has been integrated yesterday

PR #5550, opened 2021-03-23
Fix Eigen in CMakeLists 5.2

maxGimeno:
## Summary of Changes

5.2 and master version for #5549.

If Eigen is not found, Mesh_3 examples and some Tetrahedral_remeshing examples are broken instead of not being ran.
## Release Management

* Affected package(s):Mesh_3, Tetrahedral_remeshing


Mesh_3/examples/Mesh_3/CMakeLists.txt
```diff
@@ -22,172 +22,175 @@ endif()
...
-find_package(VTK QUIET COMPONENTS vtkImagingGeneral vtkIOImage NO_MODULE)
-if(VTK_FOUND)
-  if(VTK_USE_FILE)
-    include(${VTK_USE_FILE})
-  endif()
-  if("${VTK_VERSION_MAJOR}" GREATER "5" OR VTK_VERSION VERSION_GREATER 5)
-    message(STATUS "VTK found")
-    if(TARGET VTK::IOImage)
-      set(VTK_LIBRARIES VTK::ImagingGeneral VTK::IOImage)
+if(TARGET CGAL::Eigen3_support)
```
lrineau, line 25: I disagree with the patch. The line `find_package(Eigen3 3.1.0 REQUIRED)` has `REQUIRED`. We do not want to add an extra `if(TARGET CGAL::Eigen3_support)` and then indent the whole file.

maxGimeno: good catch, I forgot to remove the requiered

lrineau:
I still do not understand. Before this PR, Eigen was `REQUIRED` for Mesh_3, and now it is required but differently. What do we gain?

maxGimeno:
We replace a red square by a blue one. We have test platforms that do not have Eigen installed, and that is OK as Eigen is documented as optional.

lrineau:
OK. I am not sure that is a good idea to have a test platform without Eigen (that is header-only). But let's assume that choice.

In that case, I would suggest the following (with `return()`):
```cmake
find_package(Eigen3 3.1.0 QUIET) #(3.1.0 or greater)
include(CGAL_Eigen3_support)
if(NOT TARGET CGAL::Eigen3_support
  message( STATUS "NOTICE: All Mesh_3 examples need the Eigen3 library, and will not be compiled." )
  return()
endif()
```
and then leave the rest of the `CMakeLists.txt` as it is.

There is even a chance that this small patch can be applied to both `master` and release branches.

Note: There is no underscore in the name `Eigen3`. It is [spelled that way](https://gitlab.com/libeigen/eigen/-/blob/master/eigen3.pc.in) in upstream code.

afabri:
What in Mesh_3 depends on Eigen?

lrineau:
> What in Mesh_3 depends on Eigen?

I think that is something in the optimizers, that relies on PCA, probably to reproject points on surfaces.

janetournois:
Right, at least PCA in the tangent plane approximation for optimization on surfaces.

maxGimeno:

PR #5556, opened 2021-03-26
Fix inconsistency in normal computation in robust projection code

MaelRL:
## Summary of Changes

Somewhat follow-up of https://github.com/CGAL/cgal/pull/1741.

If the triangle/plane is degenerate, the case is already handled separately.
But if we know the plane is not degenerate, the normal should not be recomputed in a different manner, as it might lead to a null normal and some issues down the line.

## Release Management

* Issue(s) solved (if any): -


maxGimeno:

PR #5557, opened 2021-03-26
Change the URL to internal releases

lrineau:
## Summary of Changes

Change the URL to internal releases. That will fix our documentation testsuite.

## Release Management


lrineau:
Note that the scripts handling the documentation testsuite was using `curl` without the `--location` option. That is why it did not follow the HTTP redirection from `/CGAL/Members/Releases` to `/CGAL/Releases`. I have also added that option.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.3-Ic-93/index.html

PR #5561, opened 2021-03-29
Kernel_23: Fix warnings in Circle_3

### LINKED ISSUE #5506: Cartesian_kernel: Warning in Circle_3
## Issue Details

There is a warning in a test of Kernel_23 : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-73/Kernel_23/TestReport_lrineau_ArchLinux-CXX17-Release.gz
The problem lies here: 
```cpp
 struct Rep
  {
    Sphere_3 first;
    Plane_3 second;
    Rep () : first(), second() { }
    Rep (const Sphere_3& s, const Plane_3& p) : first(s), second(p) { }
  };
  ```
Replacing the `Rep` by an std::pair fixes the warning, but is there a reason why `Rep` is a struct in the first place ? 

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): archlinux cxx14 and 17
* Compiler:c++ (GCC) 10.2.0
* Release or debug mode:Release
* Specific flags used (if any):-Wall -O3 -msse4.1 -std=gnu++17 -DCGAL_NDEBUG -O3
* CGAL version:5.1, 5.2, 5.3

### PR DESCRIPTION
**Pull request by maxGimeno**
## Summary of Changes

Replace the Rep type by a boost::tuple to silence a warning.
## Release Management

* Affected package(s):Kernel_23
* Issue(s) solved (if any): fix #5506

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/Circle_3.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Circle_3.h
@@ -17,6 +17,7 @@
 #define CGAL_CARTESIAN_CIRCLEC3_H
 
 #include <CGAL/Interval_nt.h>
+#include "boost/tuple/tuple.hpp"
 
 namespace CGAL {
 
@@ -29,13 +30,8 @@ class CircleC3 {
   typedef typename R_::Direction_3              Direction_3;
   typedef typename R_::FT                       FT;
 
-  struct Rep
-  {
-    Sphere_3 first;
-    Plane_3 second;
-    Rep () : first(), second() { }
-    Rep (const Sphere_3& s, const Plane_3& p) : first(s), second(p) { }
-  };
+  typedef boost::tuple<Sphere_3, Plane_3> Rep;
+
 
   typedef typename R_::template Handle<Rep>::type  Base;
   Base base;
@@ -125,7 +121,7 @@ public:
 
   const Plane_3& supporting_plane() const
   {
-    return get_pointee_or_identity(base).second;
+    return boost::get<1>(get_pointee_or_identity(base));
   }
 
   const Sphere_3& supporting_sphere() const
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mglisse** [comment]
Weren't we trying to replace boost::tuple with std::tuple everywhere at some point?

**Comment by maxGimeno** [comment]
@mglisse there are a lot of `boost::tuple` remaining, so I'd say no. I can always replace this one if you insist.

**Comment by mglisse** [comment]
I don't insist. It just seems that unless we have a good reason, it is a bit strange to keep using boost::tuple now that we can use C++14.

**Comment by lrineau** [comment]
I am a bit shocked that we had new `boost::tuple` in the CGAL code.

But for this one, there is a really good reason: https://github.com/CGAL/cgal/issues/5506#issuecomment-789552768. With libc++, a `std::pair` of a `std::tuple` cannot use incomplete types, and thus cannot be part of a CRTP. The implementation from `boost::tuple` (or from libstdc++) work.

**Comment by mglisse** [comment]
> But for this one, there is a really good reason: [#5506 (comment)](https://github.com/CGAL/cgal/issues/5506#issuecomment-789552768). With libc++, a `std::pair` of a `std::tuple` cannot use incomplete types, and thus cannot be part of a CRTP. The implementation from `boost::tuple` (or from libstdc++) work.

I hadn't seen that the comment also applied to tuple and not just pair. Of course, when there is a good reason, and preferably with a comment in the code, using the boost version is fine.

**Comment by maxGimeno** [comment]
I confirm, std::tuple won't work : 
```
 /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/type_traits:144: erreur : no member named 'value' in 'std::is_constructible<CGAL::Plane_3<CGAL::Epick>, const CGAL::Plane_3<CGAL::Epick> &>'
```

lrineau:
I agree a comment in the code, pointing to this discussion would be a plus.

maxGimeno:

Issue #5160, opened 2020-11-16
polygon_set_2_join.cpp doesn't always finish


mglisse:
## Issue Details

I compiled and ran polygon_set_2_join from benchmark/Boolean_set_operations_2 and ran it several times. Most of the time it runs fine in less than 10s, but it happened several times that it was still running after a minute. It had already printed
> Time for 402 grid crossing polygons 0.265669

so I assume the problem was with "1000 random polygons".

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux x86_64
* Compiler: gcc 10.2
* Release or debug mode: -O3 -DNDEBUG
* CGAL version: 5.1
* Boost version: 1.71

lrineau:
@sgiraudot I have assigned the issue to you, because you worked on Bool_op_2 recently.

sgiraudot:
I've managed to isolate the bug, it actually comes from the random polygon generator, from the internal function `make_simple_polygon` to be precise (used by the documented [`random_polygon_2()`](https://doc.cgal.org/5.1/Generator/group__PkgGeneratorsRef.html#gaa8cb58e4cc9ab9e225808799b1a61174) function). This function does not seem to ever exit its `while()` loop in some specific cases.

Here is a self contained executable to reproduce the bug:

```c++
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/Random_polygon_2_sweep.h>
#include <fstream>

using Kernel = CGAL::Exact_predicates_exact_constructions_kernel;
using FT = Kernel::FT;
using Point_2 = Kernel::Point_2;

int main (int argc, char** argv)
{
  std::vector<Point_2> points;
  std::ifstream ifile (argv[1]);

  FT x, y;
  while (ifile >> CGAL::iformat(x) >> CGAL::iformat(y))
    points.emplace_back (x, y);
  std::cerr << points.size() << " points read" << std::endl;

  CGAL::make_simple_polygon (points.begin(), points.end(), Kernel());

  return EXIT_SUCCESS;
}
```

And a data set that triggers it:
```
131770319/2147483648 404716703/2147483648
-7282203/1073741824 -503278921/2147483648
667025647/2147483648 -999173425/2147483648
-20908641/1073741824 -49933297/1073741824
-13031191/2147483648 -453624967/1073741824
-363419953/1073741824 -260292757/2147483648
33972111/134217728 -378506333/1073741824
-824649493/2147483648 115611005/1073741824
[...]

PR #5564, opened 2021-03-29
Handle "right" folds

### LINKED ISSUE #1445: Random polygon generation issue
The following examples shows an infinite loop while trying to generate a random polygon

```
/*  
There must be a bug to fix in the sweepline code if this package.
The following data set hangs with the vertices in this particular order.
As the first thing the random_polygon_2 functions does
is a random_shuffle, we have put it in an #ifndef  for this testcase
 */

#define CGAL_DONT_SHUFFLE_IN_RANDOM_POLYGON_2 1

#include <CGAL/Polygon_2.h>
#include <CGAL/random_polygon_2.h>
#include <CGAL/algorithm.h>



#include <fstream>
#include <list>

typedef K::Point_2                                          Point_2;
typedef std::list<Point_2>                                  Container;
typedef CGAL::Polygon_2<K, Container>                       Polygon_2;


int main( )
{
   Polygon_2            polygon;
   std::list<Point_2>   point_set;

   std::string input = "-5 -7 46 59 55 -50 27 -93 46 -62 55 17 -71 0 -27 -41 -86 -73 7 -85 -71 93 41 -47 78 0 -13 52 -73 -25 -17 -11 56 -61 71 -69 -44 59 -9 5 -64 -96 -66 -45 39 54 -93 -22 5 -21 -32 32 -26 77 -99 -2 62 -73 -35 7 -88 60 39 56 2 3 -5 -50 3 68 82 16 -64 -38 -2 -16";


   std::stringstream input_stream(input);

   std::istream_iterator< Point_2 > in(input_stream);
   CGAL::copy_n_unique(in, 38,
                       std::back_inserter(point_set));

   std::ostream_iterator< Point_2 >  out( std::cout, "\n " );

   CGAL::random_polygon_2(point_set.size(), std::back_inserter(polygon),
                          point_set.begin());

   std::cout << "done" << std::endl;
   return 0;
}
```

### LINKED ISSUE #2585: random_polygon_2 goes into infinite loop for some point input sequences
## Issue Details

CGAL::random_polygon_2  goes into an infinite loop for some point input sequences.
It occurs both with an exact kernel (Exact_predicates_exact_constructions_kernel), and with Simple_cartesian.

The infinite loop occurs in random_polygon_2.h in the call to:
```
make_simple_polygon(vertices.begin(), vertices.end(), traits);
```

### Sample input sequences triggering the bug:

Point_2D X Y:
```
0 1
2 2
2 1
3 1
```

```
0 1 
2 0
1 1
2 1
```

```
1 3 
1 0
1 1 
1 2
```

I believe it happens because there are coplanar points in the input sequence.

random_polygon_2 throwing an exception or failing with an empty output sequence would be acceptable, or at least preferable to the infinite loop.
It's desirable to be able to generate polygons with coplanar points though..

## Source Code

Gist with minimal sample code (-std=c++11):
https://gist.github.com/bobergj/7d13865080e923b7017f4761358828ce

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): macOS Sierra 10.12.6, 64 bits
* Compiler: Apple LLVM version 9.0.0 (clang-900.0.38)
* Release or debug mode: Both
* CGAL version: 4.11 built from source. 
* Boost version: 1.59.0_3 (macports)

MaelRL:
## Summary of Changes

The following "right fold" configurations of collinear points: `x-2 x x-1` and `x+2 x x+1` cannot be handled with the typical left swap because the interval has range `n-1` (so, it swaps a single vertex onto itself). 

Code is changed to manually swap like it is done when the fold is on the left (which can and is immediately detected at the beginning of the insertion event).

## Release Management

* Affected package(s): `Generator`, `Polygon`
* Issue(s) solved (if any): fix #2585, fix #1445


MaelRL:
Does _not_ fix https://github.com/CGAL/cgal/issues/5160, which is a completely different issue where the algorithm alternates between two polygons.

Polygon/include/CGAL/Polygon_2/Polygon_2_simplicity.h
```diff
@@ -74,6 +74,8 @@ struct Vertex_index {
     explicit Vertex_index(Index_t i): m_i(i) {}
     Index_t as_int() const {return m_i;}
     Vertex_index operator++() {++m_i; return *this; }
+    bool operator==(const Vertex_index& other) const = default;
```
lrineau, line 77: `operator==` cannot be `= default`. That is only allowed since C++20. Before C++20, only special member functions (cstor, dstor, copy, assign, and move-*) can be defaulted.
https://en.cppreference.com/w/cpp/language/member_functions#Special_member_functions

maxGimeno:

PR #5566, opened 2021-03-30
PMP - Polyhedral Envelope :  Make it faster by making combinatorial tests 

afabri:
## Summary of Changes

Do not test combinations of planes where we know that there is no intersection point.

## Release Management

* Affected package(s): Polygon Mesh Processing, Surface Mesh Simplification


maxGimeno:

PR #5569, opened 2021-03-31
[cleanup][surface_mesh] Remove unused and duplicate includes

timokoch:
## Summary of Changes

Removes unused boost includes and some duplicate STL includes.
The includes became unnecessary after a735e84.

## Release Management

### DIFF
--- a/Surface_mesh/examples/Surface_mesh/sm_do_intersect.cpp
+++ b/Surface_mesh/examples/Surface_mesh/sm_do_intersect.cpp
@@ -8,14 +8,6 @@
 #include <CGAL/box_intersection_d.h>
 #include <CGAL/Timer.h>
 
-#include <boost/bind.hpp>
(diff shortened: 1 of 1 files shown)


MaelRL:
Thanks for the patch. The STL headers are still required, though.

timokoch:
The STL header includes are duplicated of line 1-3

MaelRL:
Ah, I only checked below since the order is usually local -> global; sorry about that!

timokoch:
better like this?

lrineau:
Note: This contribution is small enough that it does not require a CLA.

afabri:
We might at the same time change  `Point_3 p = sm,point(..)`   to `const Point_3& p = sm.point()`  and change to a kernel with inexact constructions.

maxGimeno:

PR #5570, opened 2021-03-31
Speed up invalid polygon removal in repair_PS

MaelRL:
## Summary of Changes

Use `std::remove_if` to have `O(n)` complexity instead of `O(k*n)` (`k` being the number of removed polygons), which gets nasty when there are a lot of invalid polygons in the soup.

Based on the oldest release rather than master because it is dramatic enough.

## Release Management

* Issue(s) solved (if any): -


sloriot (approved):

maxGimeno:

PR #5571, opened 2021-03-31
Classification: Sanitize addresses

sgiraudot:
## Summary of Changes

The cluster example triggered [an error on _one_ platform](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-94/Classification_Examples/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz).

After running the address sanitizer (which did detect problems), I figured it came from the fact that I used a temporary variable when generating features. I made this variable defined at the right scope (as is already documented) and it fixed the problem. I'm surprised no more platforms actually crashed on this…

The address sanitizer also revealed a leak that I fixed (I used `std::unique_ptr<>::release()` instead of `std::unique_ptr<>::reset()`: _release_ does not free the memory but only gives the user the ownership of the pointer).

## Release Management

* Base on 5.2 to avoid conflicts in 5.1

### DIFF
--- a/Classification/examples/Classification/example_cluster_classification.cpp
+++ b/Classification/examples/Classification/example_cluster_classification.cpp
@@ -82,7 +82,12 @@ int main (int argc, char** argv)
 #endif
 
   generator.generate_point_based_features (pointwise_features);
-  generator.generate_normal_based_features (pointwise_features, pts.normal_map());
+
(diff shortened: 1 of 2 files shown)


maxGimeno:

PR #5573, opened 2021-04-01
Polyhedron Demo:  Trivial Warning

maxGimeno:
## Summary of Changes
Fix a trivial warning about braces in the scene
## Release Management
I did not find the issue in previous release so I only applied it in master.


maxGimeno:

PR #5577, opened 2021-04-02
Tetrahedral Remeshing - fix link error

janetournois:
## Summary of Changes

This PR fixes a "double definition" link error.
We `inline` the `indices(int, int)` function, which is was not template, nor inline and causing the error when compiling two cpp targets together.

## Release Management


maxGimeno:

PR #5583, opened 2021-04-03
Polyhedral_envelope: face epsilon map

### LINKED ISSUE #5581: Feature request: adaptive envelope
## Issue Details

I saw the nice addition of "Polyhedral Envelope Containment Check" for Polygon meshes and triangle soups in the 5.3 version. Do you think it would it be possible to add the ability of using it with a "per face eps. value" (that could be passed via a property map) ?
thanks.

sloriot:
Fixes #5581

**TODO**:
- [x] update np test
- [x] add it to triangle soup? Maybe using a map from `size_t` to `double`?
- [x] test API
- [x] can we avoid the extra degenerate test

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
+++ b/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
@@ -183,7 +183,7 @@ private:
 
 
   // The class  `Plane` is used for the 7-8 walls of a prism.
-  // We store at the same  time threee points and a plane.
+  // We store at the same time three points and a plane.
   // That is easier than retrieving the 3 points of a lazy plane.
   struct Plane {
     Plane()
--- a/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
+++ b/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
@@ -321,6 +321,11 @@ public:
    *     \cgalParamExtra{If this parameter is omitted, an internal property map for `CGAL::vertex_point_t`
    *                     must be available in `TriangleMesh`.}
    *   \cgalParamNEnd
+   *   \cgalParamNBegin{face_epsilon_map}
+   *     \cgalParamDescription{a property map associating to each face of `tm` a epsilon value}
+   *     \cgalParamType{a class model of `ReadablePropertyMap` with `boost::graph_traits<TriangleMesh>::%face_descriptor`
+   *                    as key type and `double` as value type}
+   *     \cgalParamDefault{Use `epsilon` for all faces}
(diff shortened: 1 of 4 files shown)


sloriot:
Since the class is new in 5.3 we might want to have that released at the same time.

Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
```diff
@@ -321,6 +321,11 @@ struct Polyhedral_envelope {
    *     \cgalParamExtra{If this parameter is omitted, an internal property map for `CGAL::vertex_point_t`
    *                     must be available in `TriangleMesh`.}
    *   \cgalParamNEnd
+   *   \cgalParamNBegin{face_epsilon_map}
+   *     \cgalParamDescription{a property map associating to each face of `tm` a epsilon value}
```
sloriot, line 325:
```suggestion
   *     \cgalParamDescription{a property map associating to each face of `tm` an epsilon value}
```

Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
```diff
@@ -390,6 +420,12 @@ struct Polyhedral_envelope {
...
```
sloriot, line 424:
```suggestion
   *     \cgalParamDescription{a property map associating to each face of `tm` an epsilon value}
```

Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
```diff
@@ -466,6 +527,10 @@ struct Polyhedral_envelope {
     *     \cgalParamType{a model of `ReadablePropertyMap` whose value type is `Point_3` and whose key
     *                    is the value type of `PointRange::const_iterator`}
     *     \cgalParamDefault{`CGAL::Identity_property_map`}
+    *   \cgalParamNBegin{face_epsilon_map}
+    *     \cgalParamDescription{a property map associating to each triangle a epsilon value}
```
sloriot, line 531:
```suggestion
    *     \cgalParamDescription{a property map associating to each triangle an epsilon value}
```

afabri (approved):
As discussed it would be good to add an example.  For example simplification of  a uniformly meshed bunny and a sizing field for the epsilon, for example graded by  `z`.

rjanvier:
https://github.com/CGAL/cgal/blob/d8cdbbdc28665634e913f1a5387c66c41a342e43/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h#L186
Extra space "same time" and extra "e" in "three"

Issue #5333, opened 2021-01-06
Default rational type


mglisse:
## Issue Details

This comes from #5153. In CGAL-5.2, with GMP, the default rational type was mpq_rational on linux and Gmpq on windows. #5153 has the side effect of making it mpq_rational on both. We need to decide if that is the right choice (I think so, even if it slows down some packages a bit), and announce the result in changes.md. That PR had the sugested wording

> On windows, the type used for Exact_rational, in Epick and indirectly (through Lazy_exact_nt) Epeck may now be boost::multiprecision::mpq_rational, as has been the case on other platforms for several releases. This depends on various options and is added to a list that includes mpq_class, CGAL::Gmpq, leda_rational and `CGAL::Quotient<CGAL::MP_Float>`.

PR #5584, opened 2021-04-04
Number_types: Add Is_zero functor for Gmpq

afabri:
## Summary of Changes

Is this  faster than a comparison with zero?    And in case we change code where we currently compare with zero, we have to find out if it becomes slower when we apply `is_zero()` to a double.  And what  about `> 0`  ?

## Release Management

* Affected package(s):

### DIFF
--- a/Number_types/include/CGAL/GMP/Gmpq_type.h
+++ b/Number_types/include/CGAL/GMP/Gmpq_type.h
@@ -228,12 +228,6 @@ public:
   double to_double() const noexcept;
   Sign sign() const noexcept;
 
-  bool is_zero() const
-  {
-    const __mpz_struct* zn = mpq_numref(mpq());
-    return mpn_zero_p(zn->_mp_d, zn->_mp_size);
-  }
-
   const mpq_t & mpq() const noexcept { return Ptr()->mpQ; }
   mpq_t & mpq() noexcept { return ptr()->mpQ; }
 
--- a/Number_types/include/CGAL/GMP/Gmpq_type.h
+++ b/Number_types/include/CGAL/GMP/Gmpq_type.h
@@ -228,6 +228,12 @@ public:
   double to_double() const noexcept;
   Sign sign() const noexcept;
 
+  bool is_zero() const
+  {
+    const __mpz_struct* zn = mpq_numref(mpq());
+    return mpn_zero_p(zn->_mp_d, zn->_mp_size);
+  }
+
   const mpq_t & mpq() const noexcept { return Ptr()->mpQ; }
   mpq_t & mpq() noexcept { return ptr()->mpQ; }
 
--- a/Number_types/include/CGAL/Gmpq.h
+++ b/Number_types/include/CGAL/Gmpq.h
@@ -27,11 +27,10 @@ template <> class Algebraic_structure_traits< Gmpq >
     typedef Tag_false            Is_numerical_sensitive;
 
     class Is_zero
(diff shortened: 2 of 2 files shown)


afabri:
@mglisse Do I also have to check the denominator?

Number_types/include/CGAL/GMP/Gmpq_type.h
```diff
@@ -228,6 +228,12 @@ class Gmpq
   double to_double() const noexcept;
   Sign sign() const noexcept;
 
+  bool is_zero() const
+  {
+    const __mpz_struct* zn = mpq_numref(mpq());
+    return mpn_zero_p(zn->_mp_d, zn->_mp_size);
```
mglisse, line 234: Why not use the same code as mpq_class, i.e. `mpq_sgn(mpq())==0`, or `mpz_size(mpq_numref(mpq()))==0`? Or if you want to use internals, you only need to check if `zn->_mp_size` is 0.

mglisse:
> Is this faster than a comparison with zero?

Yes! With the changes I suggested, all it does is test if one int is 0, which is definitely faster than anything else.

> And in case we change code where we currently compare with zero, we have to find out if it becomes slower when we apply `is_zero()` to a double.

That would mean that the compiler sucks, because all it does it replace `d==0.` with a call to an inlinable function that does the same thing.

> And what about `> 0` ?

We should to it as well, using `mpq_sgn(...)>0`.

> Do I also have to check the denominator?

No.

Side note: for `mpq_class` with GCC, `q==0` already does the right thing, but `q==mpq_class(0)` obviously doesn't.

afabri:
In fact with #5333 this becomes somehow irrelevant, or do I miss something?

afabri:
@mglisse do I have to add `Is_zero` to both `Algebraic_structure_traits`  and `Real_embeddable_traits`?

mglisse:
> @mglisse do I have to add `Is_zero` to both `Algebraic_structure_traits` and `Real_embeddable_traits`?

I see in number_utils.h

    // We take the Algebraic_structure_traits<>::Is_zero functor by default. If it
    //  is not available, we take the Real_embeddable_traits functor

so you probably only need one of them (Algebraic_structure_traits being safer), although it doesn't hurt if you put it in both.

mglisse:
> In fact with #5333 this becomes somehow irrelevant, or do I miss something?

It is still nice to have, for fairer comparisons, or if boost_mp is disabled, or if people use Gmpq directly. I don't think it has been settled yet whether we should use Gmpq or mpq_class/mpq_rational by default for the next release.

afabri:
It is not so much a question if it hurts or not, but about trying to understand the subtleties.  [Here](https://doc.cgal.org/latest/Algebraic_foundations/classAlgebraicStructureTraits.html#ad8eea5eb95c34851dbd4b454082947e3)  two concepts are mentioned but they seem identical.  Maybe Michael @mhsaar can explain.

afabri:
As you mention `boost::multiprecision::mpq_rational` do also have to add the functors for this type?

mglisse:
> As you mention `boost::multiprecision::mpq_rational` do also have to add the functors for this type?

I think it already has them?

afabri:
just found them.

mglisse:
> do also have to add the functors for this type?

One type that is missing Is_positive and Is_negative is Mpzf (they can probably be just `x.sign()>0`, etc).

maxGimeno:

PR #5588, opened 2021-04-06
Nef_3: Avoid a filter failure

afabri:
## Summary of Changes

Avoid a geometric test that will lead to a filter failure

## Release Management

* Affected package(s): Nef_3
* License and copyright ownership: maintenance

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/Nef_box.h
+++ b/Nef_3/include/CGAL/Nef_3/Nef_box.h
@@ -52,9 +52,9 @@ class Nef_box : public Box_intersection_d::Box_d< double, 3 > {
 
   void extend( const Point_3& p, const Tag_false& ) {
     std::pair<double, double> q[3];
-    q[0] = CGAL::to_interval( p.x() );
-    q[1] = CGAL::to_interval( p.y() );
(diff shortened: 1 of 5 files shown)


sloriot (approved):

afabri:
@GilesBathgate in case you give this a try let me know if it also makes things faster for you. I sometimes have the impression that depending on what exactly we test, we discover other aspects that can improve on.  For example I never did an affine transformation in my benchmarks.

GilesBathgate:
@afabri There does seem to be a performance increase in my tests but not as significant. Certainly in the profiler, this is a hotspot that your changes improve.

I still get overall better performance with my changes in #5486

GilesBathgate:
> Remove unused local variable (why did no compiler warn?)

Because the variable removed has a constructor that could have a side-effect
http://cpp.sh/4slyj

maxGimeno:
Error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-105/Convex_decomposition_3/TestReport_lrineau_Debian-stable-Release.gz

afabri:
Fixed and checked locally with the testsuite.

maxGimeno:

PR #5589, opened 2021-04-06
Fix approximate_sqrt return types

MaelRL:
## Summary of Changes

The default `CGAL::approximate_sqrt()` is implemented as:
```
template <typename NT>
NT approximate_sqrt(const NT& nt)
{
  return NT(sqrt(CGAL::to_double(nt)));
}
```
Since `__gmp_expr` uses template to keep a stack of operations, this will fail to compile if `NT` is not trivial as there will not be a constructor from `double` available.

<s>This PR adds an overload for `__gmp_expr` and uses `Coercion_traits` to find the base type.</s>
This PR changes the return types of `CGAL::approximate_sqrt()` to use `Coercion_traits` and `result_type` to use the correct return types.

**Remarks:**

- I am not sure if that would work for everything or just stuff for which we have overloaded `Coercion_traits`.

- With regard to `mpz_class`: I'm a bit confused why this class has a `Sqrt` in `mpz_class.h`:
  * this simply calls `::sqrt()`, so it is not exact (and its algebraic category rightfully does not advertise sqrt): `sqrt(mpz_class(3))` is `1`. As such, it is misleading.
  * `CGAL::approximate_sqrt(mpz_class)` is also broken is a somewhat similar way: `CGAL::approximate_sqrt()` will think a `Sqrt` operator is available: it is defined and not `Null_functor`. But it will fail to compile for something like: `K::FT l = CGAL::approximate_sqrt(mpz_class(0) + mpz_class(9)/mpz_class(3));` since it is not possible to build a complex `__gmp_expr` as output from the `::sqrt(x)` (which is just a `double`).
Should this sqrt be completely removed to just rely on the approximate sqrt properly introduced by this PR? 

## Release Management

* Affected package(s): `Number_types`
* Issue(s) solved (if any): -

### DIFF
--- a/Number_types/test/Number_types/utilities.cpp
+++ b/Number_types/test/Number_types/utilities.cpp
@@ -124,7 +124,7 @@ int main()
        // TEST Sqrt_extension
 #ifdef CGAL_USE_GMP
       typedef CGAL::Sqrt_extension<int,int> Ext_int;
-      TESTIT(Ext_int     , "CGAL::Sqrt_extension<CGAL::Gmpz,CGAL::Gmpz>");
+      TESTIT(Ext_int     , "CGAL::Sqrt_extension<int,int>");
       typedef CGAL::Sqrt_extension<CGAL::Gmpz,CGAL::Gmpz> Ext_int_int;
       TESTIT(Ext_int_int , "CGAL::Sqrt_extension<CGAL::Gmpz,CGAL::Gmpz>");
       typedef CGAL::Sqrt_extension<CGAL::Gmpq,CGAL::Gmpz> Ext_rat_int;
--- a/Number_types/test/Number_types/utilities.cpp
+++ b/Number_types/test/Number_types/utilities.cpp
@@ -1,10 +1,5 @@
 #include <CGAL/config.h>
 
-// TODO: solve conflict of CORE with GMPXX
-#ifdef CGAL_USE_CORE
-#undef CGAL_USE_GMPXX
-#endif
-
 #include <CGAL/Quotient.h>
 #include <CGAL/MP_Float.h>
 #include <CGAL/Lazy_exact_nt.h>
--- a/Number_types/test/Number_types/utilities.cpp
+++ b/Number_types/test/Number_types/utilities.cpp
@@ -72,7 +72,7 @@ int main()
   TESTIT(long double, "long double")
 
   // CGAL number types
-  //TESTIT(CGAL::MP_Float, "MP_Float")
+  //TESTIT(CGAL::MP_Float, "MP_Float") // CGAL::div(MP_Float, MP_Float) does not implement _integer_ division
   TESTIT(CGAL::Quotient<int>, "Quotient<int>")
   TESTIT(QMPF, "Quotient<MP_Float>")
   TESTIT(CGAL::Lazy_exact_nt<QMPF>, "Lazy_exact_nt<Quotient<MP_Float> >")
--- a/Algebraic_foundations/include/CGAL/number_utils.h
+++ b/Algebraic_foundations/include/CGAL/number_utils.h
@@ -302,19 +302,21 @@ to_interval( const Real_embeddable& x) {
 }
 
 template <typename NT>
-NT approximate_sqrt(const NT& nt, CGAL::Null_functor)
+typename Coercion_traits<double, NT>::Type
+approximate_sqrt(const NT& x, CGAL::Null_functor)
 {
-  return NT(sqrt(CGAL::to_double(nt)));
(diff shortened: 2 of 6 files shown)


mglisse:
> This PR adds an overload for `__gmp_expr` and uses Coercion_traits to find the base type.

Nice.

>     * I am not sure if that would work for everything or just stuff for which we have overloaded `Coercion_traits`.

That would probably indicate missing specializations of Coercion_traits.

>     * With regard to `mpz_class`: I'm a bit confused why this class has a `Sqrt` in `mpz_class.h`:

Probably because GMP provides mpz_sqrt, with semantics similar to division (mpz_sqrtrem also gives you the remainder).

>         Should this sqrt be completely removed to just rely on the approximate sqrt properly introduced by this PR?

I think that would make sense since CGAL probably never uses sqrt-with-remainder.

We probably have a very similar situation in boost_mp.h.

MaelRL:
> > This PR adds an overload for `__gmp_expr` and uses Coercion_traits to find the base type.
> 
> Nice.
> 
> > ```
> > * I am not sure if that would work for everything or just stuff for which we have overloaded `Coercion_traits`.
> > ```
> 
> That would probably indicate missing specializations of Coercion_traits.

Everything seemed fine so instead of doing it at the `__gmp_expr` level, I have put `Coercion_traits` directly in the return type of the generic `CGAL::approximate_sqrt` (and `Sqrt::result_type` when `Sqrt` is available).
> 
> > ```
> > * With regard to `mpz_class`: I'm a bit confused why this class has a `Sqrt` in `mpz_class.h`:
> > ```
> 
> Probably because GMP provides mpz_sqrt, with semantics similar to division (mpz_sqrtrem also gives you the remainder).
> 
> > ```
> >     Should this sqrt be completely removed to just rely on the approximate sqrt properly introduced by this PR?
> > ```
> 
> I think that would make sense since CGAL probably never uses sqrt-with-remainder.

Actually there were quite a few other `Sqrt` in integer types, so I'm just going to leave everything as is. With the change from `NT` to `Sqrt::result_type` for `CGAL::approximate_sqrt`, that fixes the case of a complex `mpz_class`.

> 
> We probably have a very similar situation in boost_mp.h.

Something like `CGAL::approximate_sqrt(zero + one)` will indeed return
```
CGAL/repair/Number_types/include/CGAL/boost_mp.h:104:20: error: no matching function for call to 'sqrt'
            return boost::multiprecision::sqrt(x);
```
Could you please have a look? Otherwise, I'll just simply test `CGAL::approx_sqrt(one)`

Number_types/test/Number_types/utilities.cpp
```diff
@@ -1,10 +1,5 @@
 #include <CGAL/config.h>
 
-// TODO: solve conflict of CORE with GMPXX
-#ifdef CGAL_USE_CORE
-#undef CGAL_USE_GMPXX
-#endif
```
MaelRL, line 6: This is an old (2007) commit and I did not get any error from removing this, does this ring a bell to anyone?

mglisse:
> Actually there were quite a few other `Sqrt` in integer types, so I'm just going to leave everything as is. With the change from `NT` to `Sqrt::result_type` for `CGAL::approximate_sqrt`, that fixes the case of a complex `mpz_class`.

Aren't you still using AST::Sqrt for integers (which we don't want)? At least that's what is causing the problem with mpz_integer.

MaelRL:
Yes it is using `AST::Sqrt`, but the problem with `mpz_class` was that it would not compile due to calling `NT(double)` with potentially `NT` being a `__gmp_exp< non_trivial_T >`, from e.g. calling `CGAL::approximate_sqrt(mpz_class(0) + mpz_class(1))`, not the fact that it was returning an integer.

The fact that `AST::Sqrt(NT)` with `NT` an integer returns an `NT` (and not e.g. a double) seems to be a feature : it's also the same for Core::BigInt, and others. Therefore, it seemed incorrect to me to modify them.

mglisse:
Ah, ok. The boost failure is https://github.com/boostorg/multiprecision/issues/319, I think we can work around it in boost_mp.h in Sqrt by making `operator()` take `Type const&` instead of a template type as argument.

maxGimeno:
Error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-108/Polygon_mesh_processing/TestReport_lrineau_ArchLinux.gz

maxGimeno:
Warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-110/Number_types/TestReport_Blake_Windows_MSVC-2019-Community-Release-64bits.gz

maxGimeno:

PR #5592, opened 2021-04-08
PMP Corefinement: Replace containers to make it faster

afabri:
## Summary of Changes

Running vtune for corefinement of meshes with coplanar faces showed that we can improve by replacing by a hash map.
Astonishingly it degraded performance when also doing it for another container [here](https://github.com/CGAL/cgal/blob/master/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/intersection_impl.h#L187)

## Release Management

* Affected package(s): PMP


sloriot (approved):

lrineau:
Compilation error: https://github.com/CGAL/cgal/pull/5592/checks?check_run_id=2296005481#step:4:670

That seems a serious cons against the unordered_map: it seems part of the code uses the ordering.

sloriot:
Successfully tested in CGAL-5.4-Ic-38

PR #5593, opened 2021-04-08
Fixed unassigned variable in the tests of Arrangement_on_surface_2

danston:
In the recent test suit on the mac mini arm architecture, we have several run-time errors in the `Arrangement_on_surface_2` package. It turns out that they are caused by the unassigned variable that is then used to call the constructor that further leads to errors in the code.

This little PR fixes this problem. The change is tested locally on `x86_64` and `arm` platforms.

## Release Management

* Affected package(s): `Arrangement_on_surface_2`
* Issue(s) solved (if any): run-time errors


danston:
There is still one error in the test suite now related to a different test. I will investigate it next time I have access to the arm machine.

maxGimeno:

PR #5595, opened 2021-04-08
Nef_3: Improve K3_tree

afabri:
## Summary of Changes

In vtune we can see that the kd tree has potential for improvement.  The first step is to replace the allocation on the heap with a `Compact_container`.   
The next step will be the sidedness test for split-planes  which are all axis aligned which currently seems not  to get exploited.   

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -695,7 +695,7 @@ class Objects_around_box {
     const Node_handle node;
 
   public:
-    Iterator() : node(0) {}
+    Iterator() : node(nullptr) {}
 
     Iterator( const Node_handle root, const Bounding_box_3& s) {
       S.push_front( Candidate( root, s));
@@ -712,7 +712,7 @@ class Objects_around_box {
     Self& operator++() {
 
       if(S.empty())
-        node = 0; // end of the iterator
+        node = nullptr; // end of the iterator
       else {
         while( !S.empty()) {
           const Node_handle n = S.front().first;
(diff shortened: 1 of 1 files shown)


lrineau:
Why `Compact_container` instead of something more mainstream like `std::vector` or `std::deque`? Does the code remove nodes inside the container?

afabri:
Good remark.  I will have a look.

afabri:
I only realize now by reading  #5598 that the split plane also can undergo affine transformations.

sgiraudot:
> Why Compact_container instead of something more mainstream like std::vector or std::deque? Does the code remove nodes inside the container?

For information, this is what is done in the Spatial Searching Kd Tree:

```c++
#if defined(CGAL_TBB_STRUCTURE_IN_KD_TREE)
  tbb::concurrent_vector<Internal_node> internal_nodes;
  tbb::concurrent_vector<Leaf_node> leaf_nodes;
#else
  boost::container::deque<Internal_node> internal_nodes;
  boost::container::deque<Leaf_node> leaf_nodes;
#endif
```

I don't remember why exactly the Boost deque is used instead of the STL one. `std::vector` is not used because the objects can be moved when allocation happens and we want to keep valid pointers all along.

lrineau:
> I don't remember why exactly the Boost deque is used instead of the STL one. `std::vector` is not used because the objects can be moved when allocation happens and we want to keep valid pointers all along.

That's the commit https://github.com/CGAL/cgal/commit/df6cd26804f532d951fa86d34396615f726bc26e, but that does not say much. Our [internal wiki says](https://cgal.geometryfactory.com/CGAL/Members/wiki/Staging_Area/Accelerate_the_kd_tree#Replaced_compact_container_with_boost::deque):
> It was a `std::deque` at first, but we found out that microsofts implementation of the `std::deque` has a really huge overhead (it was three times the size of a vector for some cases).

I wonder if that bad implementation of `std::deque` is still there in MS' STL.

maxGimeno:
warnings: 
 - https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-110/Nef_3_Examples/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz

afabri:
@maxGimeno you can test it again.

maxGimeno:
warning : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-111/Convex_decomposition_3/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz

afabri:
> 
> 
> warning : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-111/Convex_decomposition_3/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz

I fixed that in #5595

maxGimeno:

PR #5598, opened 2021-04-08
Speed up Plane_3 transformation using lazy initialization.

GilesBathgate:
## Summary of Changes

This provides an alternative implementation to #5590, instead of extending the API, the existing `Aff_transformationC3` is updated to lazy initialize the transposed_inverse matrix needed for transforming the orthogonal vector of the Plane_3.

Unfortunately, I don't know how to fix the issue in the build. Perhaps someone has some suggestions.

## Release Management

* Affected package(s): Nef_3/S2

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_3.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_3.h
@@ -18,6 +18,8 @@
 #define CGAL_CARTESIAN_AFF_TRANSFORMATION_3_H
 
 #include <cmath>
+#include <memory>
+#include <mutex>
 #include <CGAL/Handle_for_virtual.h>
 
 namespace CGAL {
@@ -146,10 +148,10 @@ public:
   {
     if (is_even())
       return Plane_3(transform(p.point()),
-                 transpose().inverse().transform(p.orthogonal_direction()));
+                 transposed_inverse().transform(p.orthogonal_direction()));
     else
       return Plane_3(transform(p.point()),
-               - transpose().inverse().transform(p.orthogonal_direction()));
+               - transposed_inverse().transform(p.orthogonal_direction()));
   }
 
   Plane_3
@@ -188,8 +190,20 @@ public:
 
 protected:
   Aff_transformation_3  transpose() const { return this->Ptr()->transpose(); }
-};
 
+private:
+  mutable std::unique_ptr<Aff_transformation_3> transposed_inverse_;
+  mutable std::mutex transposed_inverse_mutex_;
+
+  const Aff_transformation_3&
+  transposed_inverse() const
+  {
+      const std::lock_guard<std::mutex> lock(transposed_inverse_mutex_);
+      if(!transposed_inverse_)
+          transposed_inverse_ = std::make_unique<Aff_transformation_3>(transpose().inverse());
+      return *transposed_inverse_;
+  }
+};
 
 template < class R >
 std::ostream&


Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_3.h
```diff
@@ -188,8 +189,18 @@ class Aff_transformationC3
...
+  mutable std::unique_ptr<Aff_transformation_3> transposed_inverse_;
+
+  const Aff_transformation_3&
+  transposed_inverse() const
+  {
+      if(!transposed_inverse_)
+          transposed_inverse_ = std::make_unique<Aff_transformation_3>(transpose().inverse());
+      return *transposed_inverse_;
+  }
+};
```
lrineau, line 203: Unfortunately, laziness (or delayed initialization) means thread-safety issues as well. This code is not thread-safe: there could be two threads initializing `transposed_inverse_` at the same time.

lrineau (changes requested):

Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_3.h
GilesBathgate: @lrineau So we'd have to lock? Or shall I just abandon the idea?

lrineau: That's unfortunate but yes. Then you will have to check if that improves the speed anyway.

GilesBathgate: @lrineau I added the lock, but I still need to fix the build issue before I can test performance. The error is:

```
/usr/local/include/CGAL/Cartesian/Aff_transformation_rep_3.h:227: error: call to implicitly-deleted copy constructor of 'CGAL::Aff_transformation_repC3<CGAL::Epeck>::Aff_transformation_3' (aka 'Aff_transformation_3<CGAL::Epeck>')
In file included from ../rapcad/src/nodeevaluator.cpp:20:
In file included from ../rapcad/src/nodeevaluator.h:23:
In file included from ../rapcad/src/primitive.h:22:
In file included from ../rapcad/src/point.h:23:
In file included from ../rapcad/src/decimal.h:25:
In file included from ../rapcad/src/cgal.h:25:
In file included from /usr/local/include/CGAL/Exact_predicates_exact_constructions_kernel.h:20:
In file included from /usr/local/include/CGAL/Simple_cartesian.h:20:
In file included from /usr/local/include/CGAL/Cartesian/Cartesian_base.h:55:
In file included from /usr/local/include/CGAL/Cartesian/Aff_transformation_3.h:35:
/usr/local/include/CGAL/Cartesian/Aff_transformation_rep_3.h:227:10: error: call to implicitly-deleted copy constructor of 'CGAL::Aff_transformation_repC3<CGAL::Epeck>::Aff_transformation_3' (aka 'Aff_transformation_3<CGAL::Epeck>')
  return t.compose(*this);
         ^~~~~~~~~~~~~~~~
/usr/local/include/CGAL/Cartesian/Aff_transformation_rep_3.h:92:3: note: in instantiation of member function 'CGAL::Aff_transformation_repC3<CGAL::Epeck>::operator*' requested here
  Aff_transformation_repC3(
  ^
/usr/local/include/CGAL/Cartesian/Aff_transformation_3.h:113:23: note: in instantiation of member function 'CGAL::Aff_transformation_repC3<CGAL::Epeck>::Aff_transformation_repC3' requested here
      initialize_with(Aff_transformation_repC3<R>(m11/w, m12/w, m13/w, m14/w,
                      ^
/usr/local/include/CGAL/Aff_transformation_3.h:65:7: note: in instantiation of member function 'CGAL::Aff_transformationC3<CGAL::Epeck>::Aff_transformationC3' requested here
    : RAff_transformation_3(m11, m12, m13, m14,
      ^
../rapcad/src/nodeevaluator.cpp:336:27: note: in instantiation of member function 'CGAL::Aff_transformation_3<CGAL::Epeck>::Aff_transformation_3' requested here
        CGAL::AffTransformation3 t(
                                 ^
/usr/local/include/CGAL/Aff_transformation_3.h:31:30: note: copy constructor of 'Aff_transformation_3<CGAL::Epeck>' is implicitly deleted because base class 'Epeck::Kernel_base::Aff_transformation_3' (aka 'Aff_transformationC3<CGAL::Epeck>') has a deleted copy constructor
class Aff_transformation_3 : public R_::Kernel_base::Aff_transformation_3
                             ^
/usr/local/include/CGAL/Cartesian/Aff_transformation_3.h:195:49: note: copy constructor of 'Aff_transformationC3<CGAL::Epeck>' is implicitly deleted because field 'transposed_inverse_' has a deleted copy constructor
  mutable std::unique_ptr<Aff_transformation_3> transposed_inverse_;
                                                ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/unique_ptr.h:468:7: note: 'unique_ptr' has been explicitly marked deleted here
      unique_ptr(const unique_ptr&) = delete;
      ^
```

lrineau: As you manage memory resources by yourself, with the lazy cache, you have define the copy-constructor (and probably also  the other special member functions).

GilesBathgate: @lrineau I wonder whether the implementation should be in `Aff_transformation_repC3` since Aff_transformation_3 is a handle_for_virtual, we probably don't want to override the copy-constructors.

afabri:
I think yet another alternative would be to add the affine transformation to the `Lazy_kernel`.  Together with @sloriot we gave it a quick try but it is less trivial than we thought.  But  there should be a gain, as then the point and the direction of the plane would not get constructed in the exact constructions kernel.

GilesBathgate:
> I think yet another alternative would be to add the affine transformation to the `Lazy_kernel`. Together with @sloriot we gave it a quick try but it is less trivial than we thought.

That sounds like a great idea, I think it will benefit RapCAD. On the other hand, OpenSCAD doesn't use the Lazy Kernel due to the increaced memory usage, I think they could still benefit from my changes in #5590

PR #5596, opened 2021-04-08
Add doc for precompiled demos

sloriot:

PR #5599, opened 2021-04-09
CI Doc building: Accept modifications to the Documentation package only

maxGimeno:
## Summary of Changes
Add the package itself to the list, and not only the content of package_info/dependencies.
This way, when differencies only appear in Documentation, for exemple, the script will not exit with an empty list of modified packages.
## Release Management

* Affected package(s):CI
* Issue(s) solved (if any): bug in #5596


maxGimeno:

PR #5601, opened 2021-04-09
Polyhedron Demo: Fix edit bbox

maxGimeno:
## Summary of Changes

A missing call to viewer->makeCUrrent() was responsible for the bbox being teleported after each picking.


maxGimeno:

PR #5606, opened 2021-04-10
Fix uninitialized scalar variable in Lazy.h

GilesBathgate:
## Summary of Changes

Several places in Lazy.h return an uninitialised scalar variable. These show as new high impact errors in coverity static analysis. 

My speculation is that they were possibly introduced in 4248a8a9f79887ebc4ccaba95f035e6313e9160f where result_type was changed from `cpp11::result_of` which might have been auto initalised differently.

The proposed change is just to return initialised scalar variables, where needed, which seems more semantically correct.

## Release Management

* Affected package(s): Filtered_kernel
* Issue(s) solved (if any): bugfix/security

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -1664,14 +1664,14 @@ struct Lazy_construction_variant {
 
       // the approximate result requires the trait with types from the AK
       AT approx_v = lazy.approx();
-      // the result we build
-      result_type res;
 
       if(!approx_v) {
         // empty
-        return res;
+        return result_type();
       }
 
+      // the result we build
+      result_type res;
       // the static visitor fills the result_type with the correct unwrapped type
       internal::Fill_lazy_variant_visitor_2< result_type, AK, LK, EK, Lazy<AT, ET, E2A> > visitor(res, lazy);
       boost::apply_visitor(visitor, *approx_v);
@@ -1682,12 +1682,12 @@ struct Lazy_construction_variant {
       Protect_FPU_rounding<!Protection> P2(CGAL_FE_TONEAREST);
 
       ET exact_v = EC()(CGAL::exact(l1), CGAL::exact(l2));
-      result_type res;
 
       if(!exact_v) {
-        return res;
+        return result_type();
       }
 
+      result_type res;
       internal::Fill_lazy_variant_visitor_0<result_type, AK, LK, EK> visitor(res);
       boost::apply_visitor(visitor, *exact_v);
       return res;
@@ -1717,14 +1717,14 @@ struct Lazy_construction_variant {
 
       // the approximate result requires the trait with types from the AK
       AT approx_v = lazy.approx();
-      // the result we build
-      result_type res;
 
       if(!approx_v) {
         // empty
-        return res;
+        return result_type();
       }
 
+      // the result we build
+      result_type res;
       // the static visitor fills the result_type with the correct unwrapped type
       internal::Fill_lazy_variant_visitor_2< result_type, AK, LK, EK, Lazy<AT, ET, E2A> > visitor(res, lazy);
       boost::apply_visitor(visitor, *approx_v);
@@ -1735,12 +1735,12 @@ struct Lazy_construction_variant {
       Protect_FPU_rounding<!Protection> P2(CGAL_FE_TONEAREST);
 
       ET exact_v = EC()(CGAL::exact(l1), CGAL::exact(l2), CGAL::exact(l3));
-      result_type res;
 
       if(!exact_v) {
-        return res;
+        return result_type();
       }
 
+      result_type res;
       internal::Fill_lazy_variant_visitor_0< result_type, AK, LK, EK> visitor(res);
       boost::apply_visitor(visitor, *exact_v);
       return res;


mglisse:
Please explain exactly what is uninitialized and why. For most types, `result_type res;` calls the default constructor just as well as `result_type()`. What type is `result_type` when coverity complains? Or is it a style issue where they want us to spell it `result_type res{};` instead although it is equivalent for the types we use?

GilesBathgate:
@mglisse it says:

```
CID 369951 (#1 of 1): Uninitialized scalar variable (UNINIT)
3. uninit_use: Using uninitialized value res. Field res.m_storage is uninitialized.
```

mglisse:
(`result` is specialized with the macros below)
Telling you the type is really the static analyzer's job, clearly it knows it or it wouldn't know there is a field `m_storage` in it. I assume this is with MSVC, because I don't see `m_storage` in the other implementations of variant I have access to. `variant` has a default constructor, which is called by default initialization. Maybe there is a field `m_storage` in `variant` that that implementation manages not to write to in the constructor, on purpose, as an implementation detail... But then it should be the same with your version. Maybe the compiler manages to use (N)RVO in one case and not the other, which changes the opinion of the analyzer?

GilesBathgate:
@mglisse I think `m_storage` is from boost: https://www.boost.org/doc/libs/1_72_0/boost/optional/detail/optional_trivially_copyable_base.hpp

If res should be default initalised I will try with just adding `{}` then we should get (N)RVO back?

mglisse:
> @mglisse I think `m_storage` is from boost: https://www.boost.org/doc/libs/1_72_0/boost/optional/detail/optional_trivially_copyable_base.hpp

Boost.Optional does look likely (I had only checked variant). It could also be optional/optional.hpp, but let's go with your link.
Then maybe the analyzer has a problem with boost?
You mention "Using uninitialized value res", I assume the use is the move implied by `return`? Boost has a specific implementation for trivially copyable types (if only we knew what that type is...), possibly because it thinks that trivial copying of uninitialized data is safe, which is precisely what the analyzer is complaining about. In that case, a comment in boost disabling that diagnostic would make more sense.

> If res should be default initalised I will try with just adding `{}` then we should get (N)RVO back?

I don't know in what direction it changed if it did, it is more likely that your patch enabled RVO (with one compiler) if it quieted a warning about the copy... For NRVO, it would actually be better to move `result_type res;` earlier, in particular outside `try` and `catch`.
It would be an interesting piece of information to check if the warning is still there with `{}`, but I don't think that change would be right.

Since you are selling the patch as fixing an uninitialized use and not as an optimization enabling RVO, I'd like to understand if there is really an issue, not paper over it.

GilesBathgate:
@mglisse the debugger reports the type (in the original version) 
```
res	<uninitialized>	CGAL::Lazy_construction_variant<CGAL::Epeck, CGAL::CommonKernelFunctors::Intersect_3<CGAL::Simple_cartesian<CGAL::Interval_nt<false> > >, CGAL::CommonKernelFunctors::Intersect_3<CGAL::Simple_cartesian<CGAL::Gmpq> > >::result_type
```

GilesBathgate:
@mglisse I think this program demonstrates the uninitialised value:

http://cpp.sh/8wtvm

note the compiler warnings:
 In function 'result_type func()':
14:14: warning: 'res.Sresult_type::value' is used uninitialized in this function [-Wuninitialized]

mglisse:
> @mglisse the debugger reports the type

There is no `Lazy_construction_variant::result_type`... I wish it didn't try to be clever and instead printed the real type after following all the typedefs (something like `optional<variant<...>>`).
My current conjecture is still that boost is doing something "clever" and that the static analyzer does not understand it (when you have `optional<big>`, it is normal to want to avoid initializing the big buffer when the optional is not engaged, and when copying `optional<small_trivial>` it is normal to want to memcpy the structure instead of bothering with a branch). However, at first glance, it seems that the clever code should only be used for trivially copyable types, which is unlikely to be the case for something like `variant<Point_3, Line_3>`.

GilesBathgate:
@mglisse I made some experimental changes and got the compile error:
```
/usr/local/include/CGAL/Lazy.h:1672: error: cannot initialize return object of type 'boost::optional<boost::variant<CGAL::Point_2<CGAL::Epeck>, CGAL::Segment_2<CGAL::Epeck>>> ()' with an lvalue of type 'result_type ()' (aka 'optional<variant<Point_2<CGAL::Epeck>, Segment_2<CGAL::Epeck>>> ()')
```
So its an `optional<variant<...` ?

mglisse:
Thanks. Remains to check if m_storage is really from optional_trivially_copyable_base.hpp (and how it gets there when the type is not trivially copyable) or from optional.hpp (and where the warning comes from since the copy constructor checks if the optional is engaged before looking at m_storage).

GilesBathgate:
optional_trivially_copyable_base.hpp, is just what came up when I googled "boost m_storage", you are likely right its boost optional.hpp that Is used here which also has m_storage. I think the standard pattern is to return boost::none for an empty boost::optional as I've proposed in #5607. This makes more sense regardless of static analysis warnings.

GilesBathgate:
>  I think the standard pattern is to return boost::none

Well that doesn't work.
https://github.com/GilesBathgate/cgal/runs/2313559329?check_suite_focus=true

mglisse:
Default construction makes as much sense as boost::none and is more generic. I liked your original patch better than boost::none. If it works, what I suggested about moving the declaration of `res` earlier would be even nicer.
At this point, it could make sense to try and build an example using only boost and not CGAL that reproduces the warning, say with `optional<X>` for some non-trivial X. And assuming one manages to do that, report the issue to boost or to the analyzer.

GilesBathgate:
These errors turned out to be a symptom of the following issue in boost: https://github.com/boostorg/optional/issues/48

PR #5608, opened 2021-04-10
Fix uninitalized scalar in Cartesian_converter.h

GilesBathgate:
## Summary of Changes

A function in Cartesian_converter.h returns an uninitialised scalar variable. This shows as a new high impact error in coverity static analysis.

The proposed change is just to return boost::none

## Release Management

* Affected package(s): Cartesian_kernel
* Issue(s) solved (if any): bugfix/security


lrineau (changes requested):
I am surprised because the default constructor of `boost::optional<T>` initializes the object to an empty `boost::optional`.

Your patch prevents the [return value optimization](https://stackoverflow.com/a/12953129/1728537).

GilesBathgate:
These errors turned out to be a symptom of the following issue in boost: https://github.com/boostorg/optional/issues/48

PR #5616, opened 2021-04-13
PMP::isotropic_remeshing - keep corners incident to isolated constrained edges

janetournois:
## Summary of Changes

This PR introduces a new type of "halfedge status" of edges that are part of the input constrained edges property map, and are incident to a face to be remeshed only by a vertex.
These edges contribute to creating corner vertices (incident to >= 3 constrained edges), but should still not be considered as patch borders.

Without dealing with these "isolated constraints", the corner vertices that were incident to 2 patch borders and 1 of these edges were not considered as corners but as simple patch border vertices, hence they could be moved/deleted by a collapse or relaxation steps.

## Release Management

* Affected package(s): PMP


lrineau:
I think we were impacted by that bug in NURBS meshing, if I remember well.

janetournois:
I had already made a fix corresponding to the data @xiaoXiaoCam had sent me, I think it was another one :-)

maxGimeno:

PR #5617, opened 2021-04-13
Nef_3: Add Aff_transformation_3::is_translation()

afabri:
## Summary of Changes

The affine transformation internally has different implementations for translation, rotation, etc.    This PR adds a method to check if the internal representation is a translation.  This is done to accelerate the function `Nef_polyhedron_3::transform()`.  When we translate there is nothing to do for the sphere maps but changing their center.

This can be tested with translation.cpp in [this ](https://github.com/afabri/Nef_3_benchmark) repository.

## Release Management

* Affected package(s): Kernel_23, Nef_3
* Feature/Small Feature (if any):  [here](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_Aff_transformation_3::is_translation()) -- **pre-approval date: 2021/04/27**
* Link to compiled documentation: [`Aff_transformation_2`](https://cgal.github.io/5617/v2/Kernel_23/classCGAL_1_1Aff__transformation__2.html) and [ `Aff_transformation_3`](https://cgal.github.io/5617/v2/Kernel_23/classCGAL_1_1Aff__transformation__3.html) 
* License and copyright ownership:  evolutive maintenance


Nef_3/include/CGAL/Nef_polyhedron_3.h
```diff
@@ -1710,6 +1710,8 @@ class Nef_polyhedron_3 : public CGAL::Handle_for< Nef_polyhedron_3_rep<Kernel_,
     bool ninety = is_90degree_rotation(aff);
     bool scale = is_scaling(aff);
 
+    bool translate = aff.is_translation();
```
GilesBathgate, line 1713: I like this approach, and wonder if `is_scaling` might also find itself in the Aff_transformation_3 classes oneday?

afabri: That's what the _Todo_ is about.

GilesBathgate:
@afabri By the way, is there any chance that CGAL would support non-affine transformations such as a perspective transformation? I know that this has been requested before for OpenSCAD: https://github.com/openscad/openscad/issues/749

lrineau:
Instead of adding more and more virtual functions `is_translation`, `is_rotation`,...  then add just one `transformation_type()` that returns an enum. That would be a duplicate of the RTTI, but it seems that you want that anyway.

GilesBathgate:
> that returns an enum.

If it were a flag based enum (i.e values at 2^n) you can then check `if(aff.TransformationType == scaling || rotation)` etc?

afabri:
My first goal is to get it fast. In fact we maybe do not need the virtual function at all and just look at the matrix entries as we currently do for the scaling.  So no need for a code review yet.

afabri:
I continued in the spirit of rapid prototyping (which may need refactoring) and made the scaling of a Nef polyhedron faster.

Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_rep_3.h
```diff
@@ -127,6 +131,12 @@ class Aff_transformation_repC3
                        t31 * v.x() + t32 * v.y() + t33 * v.z());
   }
 
+  virtual Plane_3 transform(const Plane_3& p) const
```
GilesBathgate, line 134: `virtual Plane_3 transform(const Plane_3& p) const = 0;` ?

GilesBathgate: Oh sorry this is the subclass

afabri: In fact I think the code which is currently in Aff_transformation_3   for transforming the plane should go here.

afabri:
@GilesBathgate what I did here will accelerate things in case you construct the affine transformation with `CGAL::Scaling`  or `CGAL::Translation`.

GilesBathgate:
@afabri Yeah, but maybe I need a factory method that uses the old way of detection to apply these tags.

```c++
// Returns Aff_transformation_3(CGAL::Translation,Vector_3(x,y,z));
Aff_transformation_3::fromMatrix(1,0,0,x,
                                                      0,1,0,y,
                                                      0,0,1,z,
                                                                1);
`

**Comment by afabri** [comment]
Note that all what I did could as well be done by just looking at the entries of the full 4x3 matrix.

**Comment by maxGimeno** [comment]
errors: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-108/Nef_3_Examples/TestReport_lrineau_ArchLinux.gz

**Comment by lrineau** [comment]
> I continued in the spirit of rapid prototyping (which may need refactoring) and made the scaling of a Nef polyhedron faster.

@afabri @maxGimeno I think this PR should stay out of our testing as long as it is just a rapid prototyping.

**Comment by afabri** [comment]
@lrineau can you sketch how to do RTTI for a `Handle_for_virtual`.

**Comment by lrineau** [comment]
> @lrineau can you sketch how to do RTTI for a `Handle_for_virtual`.

By the use of `dynamic_cast`: check if `this->Ptr()` is a `Aff_transformation_repH3<R>` or `Aff_transformation_repC3<R>`. What is inconvenient is that there are two implementation for translations: one for Cartesian kernel and one for homogeneous kernel. If you do not want to use `dynamic_cast` the other possibility, as I suggested at https://github.com/CGAL/cgal/pull/5617#issuecomment-819523521, is to add one single virtual function `transformation_type()` that returns an enum.

**Comment by GilesBathgate** [comment]
FWIW `is_translation` and `is_scaling` do fit in with the rest of the API (such as `is_even`)

maybe you could use the tags

```c++\
CGAL::Aff_transformation_3<K> aff(CGAL::TRANSLATION,CGAL::Vector_3(10,10,0));
bool is_translation = (aff.getType() == CGAL::TRANSLATION);
...

It doesn't seem as neat.

**Comment by lrineau** [comment]
> FWIW `is_translation` and `is_scaling` do fit in with the rest of the API (such as `is_even`)
> 
> maybe you could use the tags
> 
> ```
> CGAL::Aff_transformation_3<K> aff(CGAL::TRANSLATION,CGAL::Vector_3(10,10,0));
> bool is_translation = (aff.getType() == CGAL::TRANSLATION);
> ```
> 
> It doesn't seem as neat.

Actually, I do not really care about affine transformations. If Andreas and you think the pull-request is no longer a rapid prototyping, then let's keep it as it is if that suit you both.

**Comment by afabri** [comment]
What is bad about my approach is that in case we add another implementation with a special treatment we have to add the `is_something()``  for all of them. It is better to test in just one place.   
However it is not in  `CGAL::Aff_transformation_3<K>`,  but in the class in Cartesian  [here](https://github.com/CGAL/cgal/blob/master/Cartesian_kernel/include/CGAL/Cartesian/Aff_transformation_3.h#L41).

**Comment by afabri** [comment]
In fact I am at the point that I say myself that it is not important to add methods like `is_translation()`  as it is not really expensive to check 12 values of a matrix.  The thing is that even for a translation we cannot just return `false` for `is_scaling()`, as in case the translation vector is the null vector, it IS a scaling  (unless we even distinguished between an identity transformation, and said that a scaling with the scale factor 1 was not a scaling). 
This PR is still important for Nef, as it adds the translation and scaling of `Plane_3`.

**Comment by afabri** [comment]
> Note that all what I did could as well be done by just looking at the entries of the full 4x3 matrix.

In fact it would be a bad idea to check the values of the full matrix every time we apply a general transformation. What we could do is to check it in the constructor which takes matrix values as input, but even there we could argue that when the users passes a matrix then the chances that it is a scaling or translation are rather low.

**Comment by GilesBathgate** [comment]
@afabri The translation and scaling overrides are good for Nef_3. Do you think a separate PR for those would be beneficial?

BTW I have updated my software to take advantage of CGAL::TRANSLATION and CGAL::SCALING : https://github.com/GilesBathgate/RapCAD/commit/f95cbe507259e629eb6d38bbcf3eea2d56c02c92

**Comment by afabri** [comment]
> Do you think a separate PR for those would be beneficial?

I had in mind to  reduce  this PR to a PR which only adds scaling and translation for a plane and throw the `is_...()`  out again.

**Comment by afabri** [comment]
I feel like bamboo in the wind. An alternative is to document that we provide functions to check if we have to do with the special representations as they are [documented](https://doc.cgal.org/latest/Kernel_23/classCGAL_1_1Aff__transformation__3.html). So this does not test if a matrix has special values.

**Comment by GilesBathgate** [comment]
> I feel like bamboo in the wind.

Not heard that one before. I presume you mean you are swaying between the alternatives. Which of the options delivers the most value with the smallest amount of change?

File: Cartesian_kernel/include/CGAL/Cartesian/Scaling_rep_3.h [hunk 3]
```diff
@@ -59,6 +60,12 @@ class Scaling_repC3 : public Aff_transformation_rep_baseC3<R>
     return d;
   }
 
+  virtual Plane_3  transform(const Plane_3 &p) const
+  {
+    // direction ( which is (p.a(), p.b(), p.c())) does not change
+    return Plane_3(p.a(),p.b(),p.c(), p.d()*scalefactor_);
```
GilesBathgate, line 66: As an aside, would it work if `scalefactor_` was a Vector_3 for non-uniform scaling?

afabri: I would say that when you non-uniform scale you change the normal of a plane.
image

GilesBathgate:
> /build:v0

@sloriot :thinking:

Cartesian_kernel/include/CGAL/Cartesian/Scaling_rep_3.h
GilesBathgate: Ah, thanks for the explanation, and picture.

afabri: I think we could do special treatment for mirroring and or 90 degree rotations, but for now I stop with the affine transformations.  We should return to the hash map, where I think that it's your turn to do the next steps (back).

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
```diff
@@ -199,6 +199,17 @@ returns `true`, if the transformation is reflecting.
 */
 bool is_odd() const;
 
+/*!
+returns `true`, if the transformation type is the specialized scaling.
```
sloriot, line 203: I would prefer something like: "return `true` is the object was constructed using the tag`CGAL::Translation`."

sloriot (commented):
Reviewing the API, I would recommend to add the 2D version.

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
afabri: I agree with the spirit, but note that it is also true for the composition of  translations.

afabri:
> 
> 
> Reviewing the API, I would recommend to add the 2D version.

The thing is that in 2D there are even more specialized operations. Also note that I try to be demand driven.

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
afabri: It is even more complex. Just like`CGAL::Translation` there is a `CGAL::Identity_transformation`.   However it creates a general matrix with 1 on the diagonal, not even a `CGAL::Scaling`  with 1.   So a compositon of a translation with an identity is not a translation.

afabri: So let me take the sentence you propose and add  "as well as the composition of such objects".

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
```diff
@@ -200,12 +200,12 @@ returns `true`, if the transformation is reflecting.
 bool is_odd() const;
 
 /*!
-returns `true`, if the transformation type is the specialized scaling.
+returns `true`, if the object was constructed using the tag `CGAL::Scaling`, or as a composition of such objects.
```
sloriot, line 203:
```suggestion
returns `true`, if the object was constructed using the tag `CGAL::Scaling`, or is the result of the composition of such objects only.
```

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
```diff
@@ -200,12 +200,12 @@ returns `true`, if the transformation is reflecting.
...
 */
 bool is_scaling() const;
 
 /*!
-returns `true`, if the transformation type is the specialized translation.
+returns `true`, if the object was constructed using the tag `CGAL::Translation`, or as a composition of such objects.
```
sloriot, line 208:
```suggestion
returns `true`, if the object was constructed using the tag `CGAL::Translation`, or is the result of the composition of such objects only.
```

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
GilesBathgate: not sure about 'only' on the end I'd put:
>returns `true`, if the object was constructed using the tag `CGAL::Scaling`, or is the result of the composition of only scaling transformations.

Kernel_23/doc/Kernel_23/CGAL/Aff_transformation_3.h
GilesBathgate: returns `true`, if the object was constructed using the tag `CGAL::Translation`, or is the result of the composition of only translation transformations.

Homogeneous_kernel/include/CGAL/Homogeneous/Aff_transformationH3.h
```diff
@@ -407,9 +411,9 @@ class Translation_repH3 : public Aff_transformation_rep_baseH3<R_>
   virtual  FT
            cartesian(int i, int j) const ;
 
-  Aff_transformation_3 compose(const Aff_transformation_rep_baseH3* aff) const
+  Aff_transformation_3 compose(const Base* aff) const
   {
-    const Translation_repH3* sr = dynamic_cast<const Translation_repH3*>(aff);
+    const Self* sr = dynamic_cast<const Self*>(aff);
```
GilesBathgate, line 416: Shouldn't you check for null in-case the user supplies the wrong type?

`CGAL_precondition(sr != 0)`

GilesBathgate:
> The documentation is built. 

@afabri sorry but the documentation still isn't quite right, you would either have:

> returns true, if the object was constructed using the tag CGAL::Scaling, or is the result of the composition of only such scaling transformation objects.

or 

> returns true, if the object was constructed using the tag CGAL::Scaling, or is the result of the composition of only such scaling transformations.

(same for CGAL::Translation) It's a little pedantic, but those would be correct English grammar.

afabri:
> Reviewing the API, I would recommend to add the 2D version.

Done.

lrineau:
https://github.com/CGAL/cgal/pull/5617#event-4651130670: @afabri You have re-added this PR to the milestone for 5.3-beta. Should I delay CGAL-5.3-beta1 so this PR can be in CGAL-5.3?

Homogeneous_kernel/include/CGAL/Homogeneous/Aff_transformationH3.h
```diff
@@ -82,11 +82,19 @@ class Aff_transformation_rep_baseH3 : public Ref_counted_virtual
...
+  virtual  bool
+           is_scaling() const { return false; }
+
   virtual  RT
            homogeneous(int i, int j) const = 0;
 
   virtual  FT
            cartesian(int i, int j) const = 0;
+
+  virtual  Aff_transformation_3 compose(const Aff_transformation_rep_baseH3* aff) const { return Aff_transformation_3(); }
```
maxGimeno, line 97: warning: `aff` is not used here

sloriot: if not used, shouldn't it be pure virtual?

GilesBathgate: I think it intentionally still has an implementation @sloriot

afabri:
@maxGimeno it is pre-approved and the wait period seems over.

maxGimeno:

PR #5621, opened 2021-04-14
Add support for protecting features in offset meshing

sloriot:
If one has some feature to protect, the offset mesher can respect them.
Screenshot from 2021-04-14 17-37-51


lrineau:
On the image, the features in red are smooth. Do you want to protect all feature curves, smooth or sharp?

sloriot:
Red polylines are input polylines. The user is responsible for providing them.

maxGimeno:

PR #5622, opened 2021-04-15
Update maximum cmake versions

maxGimeno:
## Summary of Changes

Announce new minimum cmake and update all scripts to use 3.20 at most.
## Release Management

* Affected package(s):All

### DIFF
--- a/AABB_tree/benchmark/AABB_tree/CMakeLists.txt
+++ b/AABB_tree/benchmark/AABB_tree/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Created by the script cgal_create_cmake_script
 # This is the CMake script for compiling a CGAL application.
 
-cmake_minimum_required(VERSION 3.1...3.14)
+cmake_minimum_required(VERSION 3.1...3.20)
 project(AABB_traits_benchmark)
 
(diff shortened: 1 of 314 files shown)


lrineau (approved):

sloriot:
Why is an upper bound needed?

lrineau:
[`cmake_minimum_required(VERSION min...max)`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings) internally calls [`cmake_policy(VERSION min...max)`](https://cmake.org/cmake/help/latest/command/cmake_policy.html) with the same arguments. All the CMake policies known to the running version of CMake and introduced in the range `min...max` will be set to the `NEW` behavior, and that disables all CMake warnings the corresponding policies. As CMake 3.20 is tested and runs fine in our testsuite, we can declare that by updating our `cmake_minimum_required`.

As for the minimal version... I am not sure what we should do: we do not tested with any CMake version lower than CMake 3.14, but it might still work. If we upgrade the `min` argument in `cmake_minimum_required`, that would stop the configuration if a user tries with a lower CMake version.

maxGimeno:

PR #5623, opened 2021-04-16
Polyhedron_demo : Fix a warning

maxGimeno:
## Summary of Changes
Replace a member of type QList<> with a pointer to silent a warning about QList<> not being a dll-interface and add a missing macro to an interface class.
## Release Management

* Affected package(s):Polyhedron_Demo, Three

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_group_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_group_item.cpp
@@ -10,18 +10,19 @@ Scene_group_item::Scene_group_item(QString name)
     expanded = true;
     already_drawn = false;
     scene = Three::scene();
+    children = new QList<Scene_interface::Item_id>();
 }
 
 bool Scene_group_item::isFinite() const
 {
-  Q_FOREACH(Scene_interface::Item_id id, children)
+  for(Scene_interface::Item_id id : *children)
     if(!getChild(id)->isFinite()){      return false;
     }
   return true;
 }
 
 bool Scene_group_item::isEmpty() const {
-  Q_FOREACH(Scene_interface::Item_id id, children)
+  for(Scene_interface::Item_id id : *children)
     if(!getChild(id)->isEmpty()){
       return false;
     }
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/Scene_group_item.cpp
```diff
@@ -10,18 +10,19 @@ Scene_group_item::Scene_group_item(QString name)
     expanded = true;
     already_drawn = false;
     scene = Three::scene();
+    children = new QList<Scene_interface::Item_id>();
```
lrineau, line 13: That should be a `std::unique_ptr`.

Three/include/CGAL/Three/Scene_group_item.h
```diff
@@ -196,7 +196,7 @@ public :
     //!
     //! Only returns children that have this item as a parent.
     //! Children of these children are not returned.
-    QList<Scene_interface::Item_id> getChildren() const {return children;}
+    QList<Scene_interface::Item_id> getChildren() const {return *children;}
```
lrineau, line 199: You will still have the warning about the DLL export stuff.

maxGimeno: The only warning I have left in 
```
C:\Cygwin\home\afabri\cgal\Three\include\CGAL/Three/Scene_group_item.h(263,63): warning C4251: 'CGAL::Three::Scene_group_item::children': class 'std::unique_ptr<QList<int>,std::default_delete<QList<int>>>' needs to have dll-interface to be used by clients of class 'CGAL::Three::Scene_group_item' [C:\Cygwin\home\afabri\cgal\Polyhedron\demo\build\scene
_c3t3_item.vcxproj]
```
So no, it should not be a unique_ptr, what I did was right.

lrineau: I do not understand why this internal data member is part of the DLL boundary.

maxGimeno:

PR #5624, opened 2021-04-16
Handle the update of points  for autorefinement

sloriot:
This makes sure all vertices that must be identical have the same point


sloriot:

PR #5630, opened 2021-04-19
Dependency inversion in Iterative_radon of segment_tree.h

GilesBathgate:
## Summary of Changes

The motivation for this PR was initially to suppress a static analysis warning. The error reported was that in the constructor initializer list that `this` is not initialized when reading `dist` since it depends on the member variable  `size` which in turn depends on `end` and `begin` However the initialization order is correct so it's not a problem. I did feel however that the approach here wasn't very conventional, what might be called a code smell, rather than an issue. 

Instead, this PR really is just a tidy up of the `Iterative_radon` class, by moving the dependencies of the generator out of the class it potentially allows different generators to be plugged in if dependency injection is desired, for example in unit tests and so on. It just happens to also suppress a false positive static analysis error.

## Release Management

* Affected package(s):  Box_intersection_d
* Issue(s) solved (if any): static analysis error. 
* Feature/Small Feature (if any): bugfix/cleaning

### DIFF
--- a/Box_intersection_d/include/CGAL/Box_intersection_d/segment_tree.h
+++ b/Box_intersection_d/include/CGAL/Box_intersection_d/segment_tree.h
@@ -207,24 +207,19 @@ median_of_three( RandomAccessIter a, RandomAccessIter b, RandomAccessIter c,
 }
 
 
-template< class RandomAccessIter, class Predicate_traits >
+template< class RandomAccessIter, class Predicate_traits, class Generator>
 class Iterative_radon {
 
   RandomAccessIter begin;
-  std::ptrdiff_t size;
   Predicate_traits traits;
   int dim;
-
-  boost::rand48 rng;
-  boost::uniform_int<std::ptrdiff_t> dist;
-  boost::variate_generator<boost::rand48&, boost::uniform_int<std::ptrdiff_t> > generator;
+  Generator& generator;
 
 public:
 
-Iterative_radon( RandomAccessIter begin, RandomAccessIter end,
-                 Predicate_traits traits, int dim, int /*num_levels*/ )
-  : begin(begin), size(end-begin), traits(traits), dim(dim),
-    rng(), dist(0,size-1), generator(rng,dist)
+  Iterative_radon( const RandomAccessIter& begin_, const Predicate_traits& traits_,
+                   int dim_, Generator& generator_)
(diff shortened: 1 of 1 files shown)


lrineau (changes requested):
The previous code was right in my opinion.

However, there is a important flaw in your version of the code: the iterator object returned by `iterative_radon` carries a (dangling) reference to the local variable `generator` that is destroyed once the function returns.

GilesBathgate:
> The previous code was right in my opinion.

I agree it wasn't wrong.
 
> However, there is a important flaw in your version of the code: the iterator object returned by `iterative_radon` carries a (dangling) reference to the local variable `generator` that is destroyed once the function returns.

It doesn't return an `Iterative_radon` its a call to `Iterative_radon::operator()` which returns an `RandomAccessIter` 
IR and Generator are both destroyed with the stack.

lrineau:
> > The previous code was right in my opinion.
> 
> I agree it wasn't wrong.
> 
> > However, there is a important flaw in your version of the code: the iterator object returned by `iterative_radon` carries a (dangling) reference to the local variable `generator` that is destroyed once the function returns.
> 
> It doesn't return an `Iterative_radon` its a call to `Iterative_radon::operator()(int)` which returns an `RandomAccessIter`
> IR and Generator are both destroyed with the stack.

I am sorry I misinterpreted the code. You are right.

afabri:
@lrineau as you "misinterpreted", do you now approve the PR, or is there still something to look into?

lrineau:
I do not object against this PR, even if I think the current version of code (without this PR) was correct. In my opinion, again the static analyzer Giles uses is wrong.

@afabri I let you review the new code, and approve or not. I do not now much about dD Bbox Intersections, and I cannot review the patch easily.

afabri:
Hi @GilesBathgate  I am wondering why we need this class at all, and why the code is not just right in the function.

GilesBathgate:
@afabri I think you introduced it here: https://github.com/CGAL/cgal/commit/d966755630f8d4293b311744bf42894a9c90154b It seems like a good pattern to save pushing extra invariant arguments onto the stack.

afabri:
When  I read it, I missed the recursion.

afabri (approved):

lrineau (approved):

sloriot:

PR #5631, opened 2021-04-21
Fix uncaught exceptions cleanup

GilesBathgate:
## Summary of Changes

This continues the previous changes done under #5233 and addresses further uncaught exceptions found in by static analysis since upgrading from CGAL version 5.0.3

* Make the changes less ugly
* Use the language features correctly (now that C++11 is minimum)
* Fix uncaught exceptions warnings in Handle.h Handle_for.h and Multiset.h

## Release Management

* Affected package(s): STL_extention
* Issue(s) solved (if any): static analysis/bugfix/cleaning

### DIFF
--- a/Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
+++ b/Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
@@ -389,7 +389,7 @@ namespace CGAL {
     {}
 
     /// Destructor.
-    ~CMap_non_basic_iterator() noexcept(!CGAL_assertions)
+    ~CMap_non_basic_iterator() noexcept(!CGAL_ASSERTIONS_ENABLED)
     {
       CGAL_destructor_assertion( this->mmark_number!=Map::INVALID_MARK );
       if (this->mmap->get_number_of_times_mark_reserved
--- a/Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
+++ b/Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
@@ -389,7 +389,7 @@ namespace CGAL {
     {}
 
     /// Destructor.
-    ~CMap_non_basic_iterator() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+    ~CMap_non_basic_iterator() noexcept(!CGAL_assertions)
     {
       CGAL_destructor_assertion( this->mmark_number!=Map::INVALID_MARK );
       if (this->mmap->get_number_of_times_mark_reserved
--- a/STL_Extension/include/CGAL/Handle.h
+++ b/STL_Extension/include/CGAL/Handle.h
@@ -46,8 +46,7 @@ class Handle
     Handle() noexcept
         : PTR(static_cast<Rep*>(0)) {}
 
-    // FIXME: if the precondition throws in a noexcept function, the program terminates
-    Handle(const Handle& x) noexcept
+    Handle(const Handle& x) noexcept(!(CGAL_preconditions || CGAL_assertions))
     {
       CGAL_precondition( x.PTR != static_cast<Rep*>(0) );
       PTR = x.PTR;
@@ -68,7 +67,7 @@ class Handle
     }
 
     Handle&
-    operator=(const Handle& x) noexcept
+    operator=(const Handle& x) noexcept(!CGAL_preconditions)
     {
       CGAL_precondition( x.PTR != static_cast<Rep*>(0) );
       x.PTR->count++;
--- a/STL_Extension/include/CGAL/Handle.h
+++ b/STL_Extension/include/CGAL/Handle.h
@@ -46,7 +46,7 @@ class Handle
     Handle() noexcept
         : PTR(static_cast<Rep*>(0)) {}
 
-    Handle(const Handle& x) noexcept(!(CGAL_preconditions || CGAL_assertions))
+    Handle(const Handle& x) noexcept(!(CGAL_PRECONDITIONS_ENABLED || CGAL_ASSERTIONS_ENABLED))
     {
       CGAL_precondition( x.PTR != static_cast<Rep*>(0) );
       PTR = x.PTR;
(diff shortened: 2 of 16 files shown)


GilesBathgate:
@mglisse These are the changes I am proposing, hopefully, they are less irksome?

Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
```diff
@@ -389,7 +389,7 @@ namespace CGAL {
     {}
 
     /// Destructor.
-    ~CMap_non_basic_iterator() CGAL_NOEXCEPT(CGAL_NO_ASSERTIONS_BOOL)
+    ~CMap_non_basic_iterator() noexcept(!CGAL_assertions)
```
mglisse, line 392: This essentially renames CGAL_NO_ASSERTIONS_BOOL. Did you not like the name?
The lower-case name makes the difference with CGAL_assertion not very clear.

STL_Extension/include/CGAL/Handle.h
```diff
@@ -46,8 +46,7 @@ class Handle
     Handle() noexcept
         : PTR(static_cast<Rep*>(0)) {}
 
-    // FIXME: if the precondition throws in a noexcept function, the program terminates
-    Handle(const Handle& x) noexcept
+    Handle(const Handle& x) noexcept(!(CGAL_preconditions || CGAL_assertions))
```
mglisse, line 49: Why is this test more restrictive than the one for `operator=`?

Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
GilesBathgate: I thought it was maybe less ugly. ALL CAPS seems more reminiscent of C89 based code.

You make a valid point about it being harder to distinguish between `CGAL_assertion` and `CGAL_assertions`. I also considered `CGAL_assertions_enabled`

STL_Extension/include/CGAL/Handle.h
GilesBathgate: This `CGAL_preconditions` is for the `CGAL_precondition`, and `CGAL_assertions` for the `CGAL_assume` which is a CGAL_assertion when assertions are enabled.

mglisse: Ah, right, github was misleadingly hiding the CGAL_assume, sorry.

mglisse:
This kind of assumes that `get_static_error_behaviour()` is always THROW_EXCEPTION. Well, no, but it "safely" pessimizes for that case. I guess that's the only "safe" way to handle it.

GilesBathgate:
> This kind of assumes that `get_static_error_behaviour()` is always THROW_EXCEPTION.

Hmm I guess it could do:
```c++
try { CODE } catch (...) { if(get_static_error_behaviour() == ... ) ; }
```

No that won't work. 

I don't think it matters. If it's ABORT, EXIT, EXIT_WITH_SUCCESS or CONTINUE it will never reach the try/catch.

Perhaps that's what you mean. I agree that it's extra code in that situation, but since it's configured at runtime, it has to be there anyway (if you and your team want it there of course)

Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
lrineau: The CGAL guidelines mandate that macros names must be uppercase. If the code looks old, that is because it uses a macro, and not because of the naming guidelines.

lrineau:
> This kind of assumes that `get_static_error_behaviour()` is always THROW_EXCEPTION. Well, no, but it "safely" pessimizes for that case. I guess that's the only "safe" way to handle it.

`noexecpt` is a compile-time property, and the value of `get_static_error_behaviour()` is a runtime property. The code has to be pessimistic.

Combinatorial_map/include/CGAL/Combinatorial_map_iterators_base.h
GilesBathgate: @lrineau Well I know, since I added CGAL_NO_ASSERTIONS_BOOL in https://github.com/CGAL/cgal/commit/f0ac2367139bc8190816754471e382e429d064f1

It's just that for the other assertion macros the code allows lowercase i.e `CGAL_assertion_msg` etc...
Anyway, is something like `CGAL_ASSERTIONS_ENABLED` preferred, or should I just revert back to the old code and add `CGAL_NO_PRECONDITONS_BOOL`

GilesBathgate:
> It's just that for the other assertion macros the code allows lowercase i.e `CGAL_assertion_msg` etc...
> Anyway, is something like `CGAL_ASSERTIONS_ENABLED` preferred, or should I just revert back to the old code and add `CGAL_NO_PRECONDITONS_BOOL`

@lrineau @mglisse Do you have any opinion on this, if you let me know I'll change the PR to your liking.

lrineau:
> > It's just that for the other assertion macros the code allows lowercase i.e `CGAL_assertion_msg` etc...
> > Anyway, is something like `CGAL_ASSERTIONS_ENABLED` preferred, or should I just revert back to the old code and add `CGAL_NO_PRECONDITONS_BOOL`
> 
> @lrineau @mglisse Do you have any opinion on this, if you let me know I'll change the PR to your liking.

In CGAL, macro names should be uppercased. When I saw your new version of the patch, I wondered where was defined that variable `CGAL_assertions`. I wanted to check if it was `inline constexpr`.

I know `CGAL_assertion` (and likes) is not following that naming rule, but that is a mistake, and it should not be followed.

GilesBathgate:
>In CGAL, macro names should be uppercased.

@lrineau Yes, that's fine I accept the CGAL standards. My question is whether to leave as it was or change to `CGAL_ASSERTIONS_ENABLED` because `NO` and `BOOL` seemed a little messy and confusing to me which is why I wanted to change it.

lrineau (approved):

lrineau:
This pull-request is a fix of https://github.com/CGAL/cgal/pull/5233, merged into CGAL-5.3-dev. It should go into CGAL-5.3 as well.

maxGimeno:

PR #5632, opened 2021-04-21
Linear Cell Complex: Check if stream is valid

### LINKED ISSUE #5576: LCC : uninitialized variable
## Issue Details

The [warning ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-96/Linear_cell_complex_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz) of the compiler that the variable `index`  may be uninitialized is correct, namely when the input file is not valid.
We could fix it by changing the return value from `void` to `bool` in `File_scanner_OFF.h`  `scan_facet_vertex_index()`and then do the right thing when the result is `false`.
This would not be a complete fix of dealing with invalid input but improve the situation a little bit.  

## Environment

* Compiler:
* Release or debug mode:
* CGAL version:
* Boost version:

afabri:
## Summary of Changes

Implement what @gdamiand suggested in Issue #5576, that is to check if the stream is still valid, in order to avoid a justified warning.

## Release Management

* Affected package(s): Linear cell complex
* Issue(s) solved (if any): fix #5576

### DIFF
--- a/Linear_cell_complex/include/CGAL/Linear_cell_complex_constructors.h
+++ b/Linear_cell_complex/include/CGAL/Linear_cell_complex_constructors.h
@@ -249,6 +249,9 @@ namespace CGAL {
       {
         std::size_t index;
         scanner.scan_facet_vertex_index(index, j+1, i);
+        if(! in){
+          return false;
+        }
         B.add_vertex_to_facet(index);
       }
       B.end_facet();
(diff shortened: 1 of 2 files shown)


gdamiand:
Thanks @afabri.

afabri:
I just addressed another [warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-110/Surface_mesh_topology/TestReport_afabri_x64_Cygwin-Windows10_MSVC2017-Release-64bits.gz).   It is in this new category of warnings but that is somehow misleading @sloriot.   
@gdamiand can you please check that my cast makes sense (And does this division of `std::size_t` make sense?).

gdamiand:
@afabri  It seems ok.

maxGimeno:

PR #5633, opened 2021-04-21
Add a filter in function to remove caps/needle

sloriot:
Addition to non-documented function

Depends on #5945


lrineau:
Do you need to sneak it into CGAL-5.3 or will it be enough to delay it to CGAL-5.4?

sloriot:
I'm not sure yet. I did something similar but not as good in another and I introduced new dependencies I would like to try to remove.

sloriot:
I could update named parameters to accept non-copyable object \o/. However, since objects being non-copyable are usually large, I wonder if it wasn't actually only a problem of documentation. That is, a function using a named parameter might simply indicate that it accepts `std::reference_wrapper`.

sloriot:

PR #5634, opened 2021-04-21
Simplify conditional code

sloriot:
Used only in a non-documented function


maxGimeno:

PR #5636, opened 2021-04-22
Polyhedron_demo: Animate plugin

maxGimeno:
## Summary of Changes

Change behavior when reading the previous frame, so that it restores the previous state, and does not simply read the previous frame, which might not do anything if the impacted points are not the same in both frames.
## Release Management

* Affected package(s):Polyhedron demo


sloriot:

PR #5640, opened 2021-04-24
Workaround doxygen bug

sloriot:
Proper error handling instead of a crash


sloriot:
Tested on `cgal.gf.com`

PR #5646, opened 2021-04-27
Documentation: Fix doxygen comparison scripts

maxGimeno:
## Summary of Changes

If the build of doxygen master fails, or if make fails to build the doc with it, the script will fail to upload the whole documentation testsuite. This PR modifies the scripts so that they can upload the testsuite and the log instead of nothing in such a situation.
## Release Management


lrineau:
I cannot review all the patch. Looks good from far. Let's try it in `integration`.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-116.shtml

PR #5527, opened 2021-03-12
Add squared_distance(Tri3, Tri3) + Distance_23 general improvements

MaelRL:
## Summary of Changes

Changes:
* Rewrite `squared_distance(Segment_3, Segment_3)` 
* Add `squared_distance(Triangle_3, Triangle_3)`
* Add more comprehensive tests for `Segment_3-Segment_3`, `Triangle_3-Triangle_3`
* Split `squared_distance_23_x.h` like in Intersections_3: `squared_distance(Object_1, Object_2)` is in `include/Distance_23/Object_1_Object_2.h`, see https://github.com/MaelRL/cgal/tree/Distance_3-Add_Tri_Tri-GF/Distance_3/include/CGAL.
* Add missing overloads + minor obvious improvements

TODO:
- [x] Add T3-T3 to the doc
- [x] Move stuff to Distance_2/X_Y to align with `Intersections_23`
- [x] `Fix global "squared_distance" function calling internal::implementation and not the kernel functor` OR `Fix all the bad kernel usage.`

Tentative TODO:
- [ ] Update all Segment-Ray-Line distances with the better implementation
- [ ] Properly test all squared distance combinations...

### Timers
CGAL::Timers, average of 5 runs (times are quite stable).

* Segment_3-Segment_3 (10 000 000 random segment pairs)
  * CGAL old (EPICK): ~0.52s
  * CGAL new (EPICK): ~0.47s
  * CGAL new (SC\<double\>): ~0.45s
  * GTE (double): ~0.42s
  * CGAL old (EPECK) ~72s
  * CGAL new (EPECK) ~8.5s (way fewer constructions, so pretty good speedup)

* Triangle_3-Triangle_3 (10 000 000 random triangle pairs)
  * GTE (double): ~15s
  * CGAL (EPICK): ~3.28s 
  * PQP (double): ~2.85s
  * CGAL (EPECK) ~230s

### Robustness
The new Seg_3-Seg_3 implementation is more robust for almost collinear inputs, e.g.
  `0.022196017875193794 1.183303893771295 1.2647888062495083`
  `1.1309939194349463 0.075232870508210398 0.55492123932674287`
  `0.22500616715372174 0.98062669775291222 1.1349469775329584`
  `0.24551823465713216 0.96012807704324754 1.1218148716453853`
yields:
* CGAL old: 0.0010707115283136903
* CGAL new (EPICK): 1.5407439555097887e-32
* GTE: 3.0814879110195774e-33

## Release Management

[...]

PR #5227, opened 2020-11-30
Fix nullptr issue in Hash_functor

jschwing:
## Summary of Changes

During the call to `SM_clone` it may happen that an initial
`SFace_const_handle` is used for a lookup into a `Unique_hash_map`,
which triggers UB.

This commit adds a template specialization for
`Inplace_list_const_iterator`, because not all values for `H` provider
`operator->` and `Inplace_list_const_iterator` seemed to be the only
type having an issue.
## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #0000, fix #0000,...
* Feature/Small Feature (if any): bugfix
* License and copyright ownership: Returned to CGAL authors.

PR #5647, opened 2021-04-27
Distance_3: Avoid needless orientation checks / distance computations

MaelRL:
## Summary of Changes

Accelerate `squared_distance_to_triangle()` by exiting early: if the orientation check says the projection is right of the edge, then we can just take the distance to the edge's segment. In addition, the old version caused a lot of filter failures in the `std::min(d1, d2, d3)` call as about half of the points in 3D space realize the min distance at a vertex of the triangle, resulting in that distance computation being expensive in e.g. nurbs meshing.

Use distance to the segment directly rather than checking if the query is right of two edges (and then the min would be at the common vertex) because the call to `on_left_of_triangle_edge()` is on average more expensive than a call to `squared_distance(Point_3, Segment_3)`.

## Release Management

Separate from #5227 and for 5.3 by request from @afabri.

There will be a conflict with [`b95c60f` (#5527)](https://github.com/CGAL/cgal/pull/5527/commits/b95c60fc9fe671eb42a6811e137ab48167657db2). Use #5527's version.

* Affected package(s): `Distance_3`
* Issue(s) solved (if any): -

### DIFF
--- a/Distance_3/include/CGAL/squared_distance_3_2.h
+++ b/Distance_3/include/CGAL/squared_distance_3_2.h
@@ -234,31 +234,37 @@ squared_distance_to_triangle(
   const Vector_3 oe3 = vector(t0, t2);
   const Vector_3 normal = wcross(e1, oe3, k);
 
-  if(normal != NULL_VECTOR
-     && on_left_of_triangle_edge(pt, normal, t0, t1, k)
-     && on_left_of_triangle_edge(pt, normal, t1, t2, k)
-     && on_left_of_triangle_edge(pt, normal, t2, t0, k))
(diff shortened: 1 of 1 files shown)


afabri:
Let me add the context:  This function is called by `do_intersect(Sphere,Triangle)`.   So for `Epick` the distance computation is done using interval arithmetic.

afabri (approved):

MaelRL:
Replaced by https://github.com/CGAL/cgal/pull/5680.

jschwing:
Thanks for fixing this and sorry for not being able to help more with this issue.

PR #5649, opened 2021-04-28
Replace internal use of boost shared_ptr

sloriot:
Triggered by some deprecation warnings from boost:

```
In file included from /usr/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 from /usr/include/boost/smart_ptr/detail/yield_k.hpp:23,
                 from /usr/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 from /usr/include/boost/smart_ptr/detail/spinlock.hpp:42,
                 from /usr/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 from /usr/include/boost/smart_ptr/shared_ptr.hpp:29,
                 from /usr/include/boost/shared_ptr.hpp:17,
                 from /mnt/testsuite/include/CGAL/Object.h:33,
                 from /mnt/testsuite/include/CGAL/kernel_basic.h:25,
                 from /mnt/testsuite/include/CGAL/basic.h:30,
                 from /mnt/testsuite/include/CGAL/Cartesian/Cartesian_base.h:20,
                 from /mnt/testsuite/include/CGAL/Simple_cartesian.h:20,
                 from /home/cgal_tester/build/src/cmake/platforms/ArchLinux-CXX17-Release/test/Classification_Examples/example_classification.cpp:6:
/usr/include/boost/detail/iterator.hpp:13:1: note: ‘#pragma message: This header is deprecated. Use <iterator> instead.’
   13 | BOOST_HEADER_DEPRECATED("<iterator>")
      | ^~~~~~~~~~~~~~~~~~~~~~~
```
for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-I-112/Classification_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz)

It might be that only the first commit is needed to fix it but then it might be triggered by next include directives.

I also have the change for shared_ptr used in public API that I'll put in a separate draft PR as I don't know how to handle it.

PR #5650, opened 2021-04-28
Replace boost::shared_ptr used in public API

sloriot:
Follow up of #5649 

Not sure how we should handle the deprecation of public API in that case so this PR


sloriot:
What do you think of doing the change and adding a macro so that user can switch back to boost::? It is a little work on our side but at least user friendly.

lrineau:
Which API is impacted?

sloriot:
Straight skeleton and classification packages.

sloriot:
Breaking the API won't bring anything so I'm closing this.

PR #5654, opened 2021-04-29
Polyhedron Demo: Popup for OFF

maxGimeno:
## Summary of Changes

Add a pop-up to ask if the user wants to export the internal properties in the off plugin.

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
@@ -1531,14 +1531,14 @@ Scene_surface_mesh_item::save(std::ostream& out) const
     }
     else if(s.compare("v:color") == 0)
     {
-      message.append(tr(" - Vertex Color\n"));
+      message.append(tr(" - Vertex Colors\n"));
     }
   }
   for(auto s : fprop)
   {
     if(s.compare("f:color") == 0)
     {
-      message.append(tr(" - Face Color\n"));
+      message.append(tr(" - Face Colors\n"));
     }
   }
   QMessageBox::StandardButton save_internal_properties =
--- a/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
@@ -1516,27 +1516,33 @@ Scene_surface_mesh_item::save(std::ostream& out) const
   std::vector<std::string> internal_properties;
   std::vector<std::string> vprop = d->smesh_->properties<vertex_descriptor>();
   std::vector<std::string> fprop = d->smesh_->properties<face_descriptor>();
+
+  QString message = tr("Do you want to save the following properties ? \n");
+
   for(auto s : vprop)
-     if (s.compare("v:normal") == 0
-         || s.compare("v:texcoord") == 0
-         || s.compare("v:color") == 0)
+  {
+    if (s.compare("v:normal") == 0)
     {
-       internal_properties.push_back(s);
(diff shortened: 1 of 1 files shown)


lrineau:
What are "internal properties"?

maxGimeno:
It should be normals and colors, but I just realized that colors don't work.

lrineau:
For the UX, that popup is not good at all: on most cases the surface mesh will not have any internal property anyway, and even if there are, the popup should name them, so that the user can understand and choose to save them or not.

As a user, I would freeze in front of that popup: what if I click on "yes"? what will happen. And what if I click on "no"? is there any chance that I lose something?

maxGimeno:
>  on most cases the surface mesh will not have any internal property anyway

The surface mesh will ALWAYS have internal properties for its normals, that is actually the real point of this PR.

> As a user, I would freeze in front of that popup: what if I click on "yes"? what will happen. And what if I click on "no"? is there any chance that I lose something?

I'll see what I can do about listing the concerned properties in the popup then.

maxGimeno:
New pop-up looks like this :
image

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -1513,11 +1513,42 @@ bool Scene_surface_mesh_item::hasPatchIds()
...
+    }
+  QString message = tr("Do you wish to export the following internal properties ? \n");
+  for(auto s : internal_properties)
+  {
+    message.append(tr(" -"));
+    message.append(s.c_str());
+    message.append(tr("\n"));
+  }
+  QMessageBox::StandardButton save_internal_properties =
+      QMessageBox::question(CGAL::Three::Three::mainWindow(), tr("Internal Properties"), message);
```
sloriot, line 1539:
```suggestion
      QMessageBox::question(CGAL::Three::Three::mainWindow(), tr("Save Properties?"), message);
```

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -1513,11 +1513,42 @@ bool Scene_surface_mesh_item::hasPatchIds()
...
 {
+  std::vector<std::string> internal_properties;
+  std::vector<std::string> vprop = d->smesh_->properties<vertex_descriptor>();
+  std::vector<std::string> fprop = d->smesh_->properties<face_descriptor>();
+  for(auto s : vprop)
+     if (s.compare("v:normal") == 0
+         || s.compare("v:texcoord") == 0
+         || s.compare("v:color") == 0)
+    {
+       internal_properties.push_back(s);
```
sloriot, line 1524: You can fill message directly here and use a user friendly text like "Vertex Normals"...

Polyhedron/demo/Polyhedron/Scene_surface_mesh_item.cpp
```diff
@@ -1513,11 +1513,42 @@ bool Scene_surface_mesh_item::hasPatchIds()
...
+    }
+  for(auto s : fprop)
+    if(s.compare("f:color") == 0)
+    {
+      internal_properties.push_back(s);
+    }
+  QString message = tr("Do you wish to export the following internal properties ? \n");
```
sloriot, line 1531:
```suggestion
  QString message = tr("Do you want to save the following properties? \n");
```

maxGimeno:
image

sloriot:
Either "Colors" or "Normal"  to be consistent :)

sloriot (approved):

sloriot:

PR #5655, opened 2021-04-29
[Small Feature] Support for libpointmatcher::GenericDescriptorOutlierFilter

### LINKED ISSUE #5648: Missing functionality in CGAL to libpointmatcher interface
## Issue Details

The libpointmatcher librairy give you the opportunity to do a point set registration by manually giving a weight to each point.
To do it, you have to set a `GenericDescriptorOutlierFilter` in the ICP algorithm.  

>#### GenericDescriptorOutlierFilter description :
>This filter weights matched points based on a 1D descriptor of either a single point cloud (either the reference or the reading). The descriptor values must be larger than zero.
>- source (default: reference) - Point cloud from which the descriptor will be used: reference or reading
>- descName (default: none) - Descriptor name used to weight paired points
>- useSoftThreshold (default: 0) - If set to 1 (true), uses the value of the descriptor as a weight. If set to 0 (false), uses the parameter 'threshold' to set binary weights. - min: 0 - max: 1
>- useLargerThan (default: 1) - If set to 1 (true), values over the 'threshold' will have a weight of one.  If set to 0 (false), values under the 'threshold' will have a weight of one. All other values will have a weight of zero. - min: 0 - max: 1
>- threshold (default: 0.1) - Value used to determine the binary weights - min: 0.0000001 - max: inf

The construction of the ICP algorithm works all fine with the CGAL to libpointmatcher interface (see function to construct the ICP [here](https://github.com/CGAL/cgal/blob/master/Point_set_processing_3/include/CGAL/pointmatcher/compute_registration_transformation.h#L70)).  

The problem occurs when running the ICP algorithm. libpointmatcher search for the `descName` descriptor in the `source` point cloud, but it can't find it because the CGAL interface doesn't transfer this information. We can see in the [`construct_PM_cloud`](https://github.com/CGAL/cgal/blob/master/Point_set_processing_3/include/CGAL/pointmatcher/compute_registration_transformation.h#L294) function that the position of each point and its normal is transferred to the libpointmatcher point cloud but not other informations like the weight of the point that interests us here.  

The interface should be improved so that it is possible to provide to the `compute_registration_transformation` function an additional optional PropertyMap containing the weight of each point, and that the CGAL interface adds the descriptor with the name provided during the ICP configuration to the pointmatcher point cloud.  

Would it be possible to see how to make the GenericDescriptorOutlierFilter work with CGAL? I am not experienced enough to provide you with a good PullRequest adding this feature. 
Maybe @pomerlef can help us on this one ?

danston:
This PR solves the issue #5648 and provides support for the `libpointmatcher::GenericDescriptorOutlierFilter` to the CGAL `libpointmatcher` interface. Now it is possible to provide a scalar map that maps a point from input range to a scalar value so-called weight assigned to this point. These weights are then stored as a matrix and submitted to the `libpointmatcher`. The default weights are all set to one and the corresponding `libpointmatcher` descriptor is added.

## Release Management

* Affected package(s): PSP
* Issue(s) solved (if any): fix #5648
* Feature/Small Feature (if any): [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Extended_Support_For_Pointmatcher)
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.geometryfactory.com/~danston/Point_set_processing_3/group__PkgPointSetProcessing3Algorithms.html#gaf75af5c1634fa83fa05a33e95570b127)

### DIFF
--- a/BGL/include/CGAL/boost/graph/named_params_helper.h
+++ b/BGL/include/CGAL/boost/graph/named_params_helper.h
@@ -435,30 +435,6 @@ CGAL_DEF_GET_INITIALIZED_INDEX_MAP(face, typename boost::graph_traits<Graph>::fa
         > ::type  type;
     };
 
-    template<typename PointRange, typename NamedParameters>
-    class GetScalarMap
-    {
-      struct DummyScalarMap
-      {
-        typedef typename std::iterator_traits<typename PointRange::iterator>::value_type key_type;
-        typedef typename GetK<PointRange, NamedParameters>::Kernel::FT value_type;
-        typedef value_type reference;
-        typedef boost::read_write_property_map_tag category;
-
-        typedef DummyScalarMap Self;
-        friend reference get(const Self&, const key_type&) { return value_type(1); }
-        friend void put(const Self&, const key_type&, const value_type&) { }
-      };
-
-    public:
-      typedef DummyScalarMap NoMap;
-      typedef typename internal_np::Lookup_named_param_def <
-        internal_np::scalar_t,
-        NamedParameters,
-        DummyScalarMap // default
-        > ::type  type;
-    };
-
     template<typename PlaneRange, typename NamedParameters>
     class GetPlaneMap
     {
--- a/BGL/include/CGAL/boost/graph/named_params_helper.h
+++ b/BGL/include/CGAL/boost/graph/named_params_helper.h
@@ -435,6 +435,30 @@ CGAL_DEF_GET_INITIALIZED_INDEX_MAP(face, typename boost::graph_traits<Graph>::fa
         > ::type  type;
     };
 
+    template<typename PointRange, typename NamedParameters>
+    class GetScalarMap
+    {
+      struct DummyScalarMap
+      {
+        typedef typename std::iterator_traits<typename PointRange::iterator>::value_type key_type;
+        typedef typename GetK<PointRange, NamedParameters>::Kernel::FT value_type;
+        typedef value_type reference;
+        typedef boost::read_write_property_map_tag category;
+
+        typedef DummyScalarMap Self;
+        friend reference get(const Self&, const key_type&) { return value_type(1); }
+        friend void put(const Self&, const key_type&, const value_type&) { }
+      };
+
+    public:
+      typedef DummyScalarMap NoMap;
+      typedef typename internal_np::Lookup_named_param_def <
+        internal_np::scalar_t,
+        NamedParameters,
+        DummyScalarMap // default
+        > ::type  type;
+    };
+
     template<typename PlaneRange, typename NamedParameters>
     class GetPlaneMap
     {
--- a/Point_set_processing_3/examples/Point_set_processing_3/registration_with_pointmatcher.cpp
+++ b/Point_set_processing_3/examples/Point_set_processing_3/registration_with_pointmatcher.cpp
@@ -21,6 +21,17 @@ typedef std::pair<Point_3, Vector_3> Pwn;
 typedef CGAL::First_of_pair_property_map<Pwn> Point_map;
 typedef CGAL::Second_of_pair_property_map<Pwn> Normal_map;
 
+struct Weight_map
+{
+  typedef Pwn key_type;
+  typedef typename K::FT value_type;
+  typedef value_type reference;
+  typedef boost::readable_property_map_tag category;
+
+  typedef Weight_map Self;
+  friend reference get(const Self&, const key_type&) { return value_type(1); }
+};
+
 namespace params = CGAL::parameters;
 
 int main(int argc, const char** argv)
@@ -63,12 +74,13 @@ int main(int argc, const char** argv)
   point_set_2_filters.push_back( ICP_config { /*.name=*/"MinDistDataPointsFilter"       , /*.params=*/{ {"minDist", "0.5" }}  } );
   point_set_2_filters.push_back( ICP_config { /*.name=*/"RandomSamplingDataPointsFilter", /*.params=*/{ {"prob"   , "0.05"}}  } );
 
-        // Prepare matcher function
+  // Prepare matcher function
   ICP_config matcher { /*.name=*/"KDTreeMatcher", /*.params=*/{ {"knn", "1"}, {"epsilon", "3.16"} } };
 
   // Prepare outlier filters
   std::vector<ICP_config> outlier_filters;
   outlier_filters.push_back( ICP_config { /*.name=*/"TrimmedDistOutlierFilter", /*.params=*/{ {"ratio", "0.75" }}  } );
+  outlier_filters.push_back( ICP_config { /*.name=*/"GenericDescriptorOutlierFilter", /*.params=*/{ {"descName", "weights" }}  } );
 
   // Prepare error minimizer
   ICP_config error_minimizer { /*.name=*/"PointToPointErrorMinimizer"};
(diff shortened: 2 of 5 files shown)


BGL/include/CGAL/boost/graph/named_params_helper.h
```diff
@@ -435,6 +435,30 @@ CGAL_DEF_GET_INITIALIZED_INDEX_MAP(face, typename boost::graph_traits<Graph>::fa
         > ::type  type;
     };
 
+    template<typename PointRange, typename NamedParameters>
+    class GetScalarMap
+    {
+      struct DummyScalarMap
```
sgiraudot, line 441: You can probably use `Constant_property_map` (see in `property_map.h` in the Property map package), provided you initialize it with the 1 value in the PSP code.

danston: Cool. This is what I was searching for. I will change. Thanks!

Point_set_processing_3/include/CGAL/pointmatcher/compute_registration_transformation.h
```diff
@@ -385,6 +401,12 @@ compute_registration_transformation(const PointRange1& range1, const PointRange2
                       of the iterator of `PointRange1` and whose value type is `geom_traits::Vector_3`}
      \cgalParamNEnd
 
+     \cgalParamNBegin{scalar_map}
+       \cgalParamDescription{a property map associating 1D values - scalars to the elements of the point set `point_set_1`}
+       \cgalParamType{a model of `ReadablePropertyMap` whose key type is the value type
+                      of the iterator of `PointRange1` and whose value type is `geom_traits::FT`}
```
MaelRL, line 407: Should indicate that the default is a constant pmap.

Point_set_processing_3/examples/Point_set_processing_3/registration_with_pointmatcher.cpp
```diff
@@ -63,15 +74,16 @@ int main(int argc, const char** argv)
...
   point_set_2_filters.push_back( ICP_config { /*.name=*/"RandomSamplingDataPointsFilter", /*.params=*/{ {"prob"   , "0.05"}}  } );
 
-        // Prepare matcher function
+  // Prepare matcher function
   ICP_config matcher { /*.name=*/"KDTreeMatcher", /*.params=*/{ {"knn", "1"}, {"epsilon", "3.16"} } };
 
   // Prepare outlier filters
   std::vector<ICP_config> outlier_filters;
   outlier_filters.push_back( ICP_config { /*.name=*/"TrimmedDistOutlierFilter", /*.params=*/{ {"ratio", "0.75" }}  } );
+  outlier_filters.push_back( ICP_config { /*.name=*/"GenericDescriptorOutlierFilter", /*.params=*/{ {"descName", "weights" }}  } );
```
maximecharriere, line 83: Should we warn in the doc that only "weights" can be given as "descName" and not any name?

maximecharriere:
In your example you're using a [`struct Weight_map{...}`](https://github.com/CGAL/cgal/blob/3efa4316d1882fd76fc3650096de5620de126259/Point_set_processing_3/examples/Point_set_processing_3/registration_with_pointmatcher.cpp#L24) as the property map.  
I don't realy understand how this struct work. Is it fine if I use the [add_property_map](https://doc.cgal.org/latest/Surface_mesh/classCGAL_1_1Surface__mesh.html#aa4c1cb3ee0a1e0530b5427554c6517b0) function with a `vertex_index` as the key and a `double` as the value ?

```c++
scanWeightPm= scan.add_property_map<Mesh::vertex_index, double>("v:weight", 1.0).first;
refWeightPm= ref.add_property_map<Mesh::vertex_index, double>("v:weight", 1.0).first;

std::pair<K::Aff_transformation_3, bool> res =
    CGAL::pointmatcher::compute_registration_transformation(
		scan.vertices(),
		ref.vertices(),
		CGAL::parameters::point_map(scan.points())
		                             .pointmatcher_config(&configFile)
		                             .scalar_map(scanWeightPm),
		CGAL::parameters::point_map(ref.points())
		                             .scalar_map(refWeightPm));
	});
```

`scan` and `ref` are `CGAL::Surface_mesh<Point_3>` type.

danston:
@maximecharriere  Since this example uses `std::vector` as a container whose element type is `std::pair`: point + normal, I use my own defined property map, which is `Weight_map`. Inside the code, the weight property map is used as
```
const auto& weight = get(scalar_map, p);
```
where `scalar_map` is the `Self` type of the `Weight_map` and `key_type` is `std::pair` that is it assigns `weight = 1` to each `point_with_normal` from the given vector and returns it inside the code.

Since you are using `Surface_mesh<Point_3>` yes, you can define the vertex map as you proposed. That should work. Let me know if it does not in whatever case. I will check what is wrong.

Last comment. Why do you use `Surface_mesh` with point clouds? In CGAL, there is a special data structure for that called `Point_set_3` - [see here](https://doc.cgal.org/latest/Point_set_3/classCGAL_1_1Point__set__3.html). Just in case you do not know that.

maximecharriere:
Thank you for this PR @danston !
It works realy fine. I tested not using `GenericDescriptorOutlierFilter`, using it on the reference, on the reading or on both and your modifications work in all cases. 👍   


Thanks also for your explanations, I now understand the use of the `Weight_map` struct.  
I did not know indeed the `Point_set_3` data structure. I will think if I modify it in my program or if I will need the Surface_mesh features in the future.  

Did I close my issue #5648  ?

danston:
@maximecharriere Great to hear that! It is always nice to get feedback and improve the code so I was glad to help. No, you can leave the issue open. We will close it once this PR is fully tested and integrated.

afabri (approved):

afabri:
This is a trivial PR as all what it does is making an underlying functionality of a 3rd party software available.  So I changed the milestone.

afabri (approved):

danston:
> This PR is not trivial:
> 
> * the patch is not trivial, and may break code,
> * there is new documentation, so it should be a small feature.

I transformed it into a small feature and added a link to the wiki and new docs.

danston:
To test this PR, the `libpointmatcher` (that also includes `libnabo` and possibly `opengr`, there is an example with `libpointmatcher` + `opengr`) must be installed on several platforms. I have found it on several linux machines. I have also installed it on mac `x86_64` and on the `arm` machine. @maxGimeno Could you please do it for at least one Windows machine? Thanks.

maxGimeno:
So apparently they broke again everything on windows. Couldn't we just admit that PointMatcher is not supported on windows machines ? They call that "partially supported" , I called that "untested".

danston:
> So apparently they broke again everything on windows. Couldn't we just admit that PointMatcher is not supported on windows machines ? They call that "partially supported" , I called that "untested".

In fact, it was not a smooth compilation on macOS as well. What is the exact error? I fixed several by setting `-Wno-c++11-narrowing`. I am not sure if it is the best solution, but it compiled and I succeeded with all related examples from PSP.

maxGimeno:
It's an error about their yaml-cpp-whatever file, then it's the libraries list that contains `optimized;` and `debug;`, whuch are interpreted as libraries instead of keywords.

danston:
> It's an error about their yaml-cpp-whatever file, then it's the libraries list that contains `optimized;` and `debug;`, whuch are interpreted as libraries instead of keywords.

Ok, then I would say we can ignore Windows for the moment.  Let's test on what we have and see if they fix this stuff until we decide to merge this PR.

maxGimeno:
They won't, they don't test on windows.

lrineau:
> They won't, they don't test on windows.

Should not you report a bug, so that they can know there is an issue?

maxGimeno:
It is done : https://github.com/ethz-asl/libpointmatcher/issues/469.

afabri:
In case it is trivial to fix we might make a PR.

maxGimeno:
@afabri I already did, and it has been integrated : https://github.com/ethz-asl/libpointmatcher/issues/469

danston:
@maxGimeno Can we also start testing this PR and remove TODO or is there anything else that should be done here? Thanks.

maxGimeno:

PR #5657, opened 2021-04-29
Fix uninitialized members in Edge_collapse.h

GilesBathgate:
## Summary of Changes

Fix uninitialized member (m_initial_edge_count, m_current_edge_count) in Edge_collapse.h

## Release Management

* Affected package(s): Surface Mesh Simplification
* Feature/Small Feature (if any): bugfix / static analysis

### DIFF
--- a/Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Edge_collapse.h
+++ b/Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Edge_collapse.h
@@ -425,7 +425,9 @@ EdgeCollapse(Triangle_mesh& tmesh,
     m_get_placement(get_placement),
     m_should_ignore(should_ignore),
     m_visitor(visitor),
-    m_has_border(!is_closed(tmesh))
+    m_has_border(!is_closed(tmesh)),
+    m_initial_edge_count(0),
+    m_current_edge_count(0)
 {
   m_max_dihedral_angle_squared_cos = CGAL::square(std::cos(1.0 * CGAL_PI / 180.0));
 


Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/internal/Edge_collapse.h
```diff
@@ -425,7 +425,9 @@ EdgeCollapse(Triangle_mesh& tmesh,
     m_get_placement(get_placement),
     m_should_ignore(should_ignore),
     m_visitor(visitor),
-    m_has_border(!is_closed(tmesh))
+    m_has_border(!is_closed(tmesh)),
+    m_initial_edge_count(0),
+    m_current_edge_count(0)
 {
   m_max_dihedral_angle_squared_cos = CGAL::square(std::cos(1.0 * CGAL_PI / 180.0));
```
afabri, line 432: Shouldn't this also go into the initializer list?

GilesBathgate: `m_max_dihedral_angle_squared_cos`? yes maybe, but at least it is initialized, and the focus of this PR was to initialize the members that aren't

GilesBathgate: Id probably do:
```c++
constexpr cos_1degree_squared = 0.99969541350954786500312172002196498;
//...
m_max_dihedral_angle_squared_cos(cos_1degree_squared)
```

MaelRL:
Probably should be initialized to their real values (`size_type(edges(m_tm).size())`) even though it is not used, just to reduce confusion.

Do you not get a warning about `m_edge_data`, which is also only initialized once the `run()` function is called?

GilesBathgate:
>Probably should be initialized to their real values (size_type(edges(m_tm).size())) even though it is not used, just to reduce

Sure, I will update it.

> Do you not get a warning about `m_edge_data`, which is also only initialized once the `run()` function is called?

Its a `boost::scoped_array` so it is default initialized.

lrineau:
@GilesBathgate Like with [any of the other pull-requests about uninitialized values](https://github.com/CGAL/cgal/pulls?q=author%3A%22GilesBathgate%22+is%3Aopen+uninitialized), we will not merge this pull-request. Uninitialized values are a very intersecting thing, because that allows static analyzers to reasons with. That is the *use* of uninitialized values that is interesting, and can reveal bugs.

What is the piece of code you were running static analyzers on? Can you show the logs of the analyzer? Because we certainly would need to fix that use of uninitialized value, instead of hiding the problem with a fake initialization with a fake value.

GilesBathgate:
@lrineau It just doesn't know that the class is only used internally, so it's assuming all members must be initialized. I'm just putting these PR's up so that they can be considered, but as you say in this case there may be no need to fix this.

PR #5660, opened 2021-04-30
Remove code duplication in Reflex_vertex_searcher.h

GilesBathgate:
## Summary of Changes

This simplifies the code in Reflex_vertex_searcher.h by utilising the right variables as an array (Object_handle instead of markedsf) so that a loop can be used instead of repeating the same code twice. There is a slightly different trace output format but I don't think that matters.

I don't think the existing code was a deliberate manual loop unrolling, and it should be up to the compiler to unroll loops for performance.

This also fixes a static analysis false positive where the analysis gets confused about the check against markedsf and thinks there is a dereference after null.

## Release Management

* Issue(s) solved (if any): bugfix / cleaning


afabri (approved):

maxGimeno:

PR #5662, opened 2021-04-30
Add a visitor to stop repair and track the current status

sloriot:


sloriot:
Successfully tested in CGAL-5.4-Ic-30

PR #5663, opened 2021-04-30
Add convenience function to snap all vertices of a mesh

sloriot:


maxGimeno:

PR #5665, opened 2021-04-30
Polyhedron_3 and TDS_3 compare handle nullptr

GilesBathgate:
## Summary of Changes

This uses the nullptr comparison operators of Compact_container to check for an empty handle instead of using an instance of an empty handle.  I think this is clearer semantically, and possibly slightly more performant.

The same operators are then added to In_place_list, and vector_iterator. Comparisons to nullptr utilised in the same way.

The change also suppresses some static analysis warnings about usage after explicit null.

## Release Management

* Affected package(s): Polyhedron_3 and TDS_3

### DIFF
--- a/Polyhedron/include/CGAL/Polyhedron_incremental_builder_3.h
+++ b/Polyhedron/include/CGAL/Polyhedron_incremental_builder_3.h
@@ -402,7 +402,7 @@ protected:
         CGAL_assertion( ! last_vertex);
         HalfedgeDS_items_decorator<HDS> decorator;
         Halfedge_handle e = get_vertex_to_edge_map( w);
-        if ( e != Halfedge_handle()) {
+        if ( e != nullptr) {
             CGAL_assertion( e->vertex() == index_to_vertex_map[w]);
             // check that the facet has no self intersections
             if ( current_face != Face_handle()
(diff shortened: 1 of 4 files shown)


lrineau (approved):

afabri:
@maxGimeno I guess that this PR is responsible for the red line, of the package `Polyhedron`, in [today's testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-119.shtml). Can you  try to fix it.

afabri:
@maxGimeno what are the warnings?

GilesBathgate:
@afabri Was this not the warnings in the test suite (that I think I've already fixed?)

maxGimeno:
@GilesBathgate yes it was. Thnaks for that

maxGimeno:
I guess what @afabri wants to know is why I added a flag, and the answer is in the description of the PR : 

> The change also suppresses some static analysis warnings about usage after explicit null.

GilesBathgate:
@afabri @maxGimeno I wonder if the CGAL developers have ever looked at using clang-analyser https://clang-analyzer.llvm.org/ it could provide some of these kinds of warnings in your testsuite...

maxGimeno:

PR #5671, opened 2021-05-03
CGAL_IO: Add missing fallbacks

maxGimeno:
## Summary of Changes

Add the remaining deprecated functions and fixes the result_of<> that "is not a part of the CGAL::cpp11 namespace"

## Release Management
* Affected package(s):CGAL_IO


Installation/include/CGAL/config.h
```diff
@@ -641,6 +641,7 @@ namespace CGAL {
     using std::is_enum;
     using std::unordered_set;
     using std::unordered_map;
+    using std::result_of;
```
sloriot, line 644: I think this will trigger a warning if c++17 is used

Installation/include/CGAL/config.h
```diff
@@ -623,6 +623,7 @@ using std::max;
 #  include <unordered_set>
 #  include <unordered_map>
 #  include <functional>
+#  include <CGAL/result_of.h>
```
lrineau, line 626: Why do you need that?! I disagree. That header will pull too much of Boost into all headers of CGAL.

Apollonius_graph_2/include/CGAL/Apollonius_graph_2/comparator_profiler.h
```diff
@@ -19,7 +19,6 @@
 
 
 #include <CGAL/Apollonius_graph_2/basic.h>
-#include <CGAL/atomic.h>
```
lrineau, line 22: I have a mixed feeling about that.

For the moment the code is using `CGAL::cpp11::atomic`, that is defined in `CGAL/config.h`, but it should be using `std::atomic`, from the header `<atomic>`.

Installation/include/CGAL/config.h
```diff
@@ -641,6 +644,32 @@ namespace CGAL {
...
+#if CGAL_CXX20 || __cpp_lib_is_invocable>=201703L
+    template<typename Signature> class result_of;
+    template<typename F, typename... Args>
+    class result_of<F(Args...)> : public std::invoke_result<F, Args...> { };
+#else
+    using std::result_of;
+#endif
+    using std::thread;
+
+    inline void sleep_for (double seconds)
```
lrineau, line 664: I think we want to remove that function `CGAL::cpp11::sleep_for`. It is used only twice in CGAL.

Apollonius_graph_2/include/CGAL/Apollonius_graph_2/comparator_profiler.h
lrineau: I am pretty sure CGAL code is not supposed to use something that is in `CGAL_NO_DEPRECATED_CODE`.

Usages of `<CGAL/atomic.h>` and `<CGAL/thread.h>` should be replaced by usages of proper classes and functions from C++11 `<atomic>` and `<thread>`.

lrineau:
By the way, why are we discussing about `CGAL::cpp11` in a PR that was supposed to be about deprecated I/O functions?

maxGimeno:
Because it was supposed to fix the result_of<> error, and then it became about threads and atomic too.

maxGimeno:
error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-125/AABB_tree/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz

maxGimeno:

lrineau:
@maxGimeno 
The file `<CGAL/result_of.h>` should still be tested, and working, even if it is deprecated. Now it gives an error:

```
.../STL_Extension/include/CGAL/result_of.h:59:9: error: redefinition of ‘class CGAL::cpp11::result_of<F(Args ...)>’
   59 |   class result_of<F(Args...)> : public std::invoke_result<F, Args...> { };
      |         ^~~~~~~~~~~~~~~~~~~~~

.../Installation/include/CGAL/config.h:672:11: note: previous definition of ‘class CGAL::cpp11::result_of<F(Args ...)>
```

PR #5673, opened 2021-05-04
Nef_3 performance chained map

GilesBathgate:
## Summary of Changes

This PR attempts to make the code in chained_map cleaner, with the main intention being to add a reserve method, and initialising the hash map on first access instead of during construction.

* Internalised chained_map_elem
* Put table/table_end/table_free into struct
* Rename parameters and variable names
* Init_table called from first call to access
* Delete table no longer called from access
* Delete old table after rehash, (removes need for old_table/old_table_end/old_table_free etc)
* Use constexpr for nullptr_key, and non_nullptr_key.
* Use constexpr for min_size, which is now 32 by default

## Release Management

* Affected package(s): Hash_map / Nef_3
* Issue(s) solved (if any): performance / cleaning

### DIFF
--- a/Hash_map/include/CGAL/Tools/chained_map.h
+++ b/Hash_map/include/CGAL/Tools/chained_map.h
@@ -125,7 +125,7 @@ public:
     return (it == &stop) ? nullptr : it;
   }
 
-  size_type number_of_entries()
+  size_type number_of_entries() const
   {
     size_type n = 0;
     if (!table.begin)
--- a/Hash_map/include/CGAL/Tools/chained_map.h
+++ b/Hash_map/include/CGAL/Tools/chained_map.h
@@ -23,362 +23,290 @@ namespace CGAL {
 
 namespace internal {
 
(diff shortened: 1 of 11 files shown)


GilesBathgate:
@afabri There does seem to be a slight performance boost in this PR which comes from using 32 as the min_size, but also using reserve when the number of rehashes needed exceeds 10. (I add some temporary code to rehash which counts the number of times called)

afabri:
I am wondering if the `Generic_handle_map`  is not only slower be cause the hash function does not divide by the size of the value type?

GilesBathgate:
@afabri It was an unexpected result from my testing. I imagined it could be a number of reasons, the layout of the memory for the handle addresses, the number of rehashes being reduced due to splitting out by handle type (because the buckets double each rehash) or maybe just because there are 128 (4x32) initial buckets across all maps.

GilesBathgate:
@afabri The separate commits were on this branch before I squashed them https://github.com/GilesBathgate/cgal/tree/Nef_3-performance_chained_map_refactor-GilesBathgate

PR #5674, opened 2021-05-04
parallel lloyd_optimize_mesh_2

yg42:
This PR is for debugging purposes.

## Summary of Changes

I am trying to modify the function lloyd_optimize_mesh_2 in order to have a parallel computation of all movements. 

When I simulate the threaded behavior (using a serie of different function calls), I have no problem and it seems to work. However, if I use actual threads, I get an error that occurs at Constrained_voronoi_diagram_2.h:360 (this line is polygon.push_back(m_cdt.circumcenter(face));)

This error is malloc(): unaligned tcache chunk detected


Mesh_2/examples/Mesh_2/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.1...3.20)
 project(Mesh_2_Examples)
 
 find_package(CGAL REQUIRED)
-
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pthread" )
```
maxGimeno, line 5: This shouldn't go in the file. Flags are supposed to be given at configuration time, as a cmake option. For example, calling `-pthread` will most certainly make the whole Mesh_2 examples unusable on Windows.

yg42: Thank you for this remark, I did not know this. Anyway, the example was added "just to be sure" I found a code that may segfault. The mesh_optimization raises the same error.

What should I do with this part of code ?

thank you again for your time

Yann

lrineau: - `-Wall` should not appear in the `CMakeLists.txt`. You can add it to your cmake user presets.
- As for `-pthread`, the CMake portable way to do it:
  ```cmake
  find_package(Threads)
  target_link_libraries(test_parallel_Mesh_global_optimizer PRIVATE Threads::Threads)
  ```
  and on Linux that will add `-pthread` and on other platform, that will modifies the compilation and linker flags with what is needed.

yg42: Thank you for this comment.

Is there a way to add this link directive within the foreach ? The solution I found was to explicitely make the linking :
```cmake
file(
  GLOB cppfiles
  RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
  ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
foreach(cppfile ${cppfiles})
  create_single_source_cgal_program("${cppfile}")
endforeach()

target_link_libraries(test_parallel_Mesh_global_optimizer PRIVATE Threads::Threads)
target_link_libraries(mesh_optimization PRIVATE Threads::Threads)
```

It may be enough for the debugging purpose.

Yann

lrineau: We do not want to add that in the `foreach`, to avoid linking all executables with `libpthread`. Only the two that require it are linked with threads.

yg42:
Hello all,
have anyone managed to reproduce this problem ?

best regards

yann

Mesh_2/include/CGAL/Mesh_2/Mesh_global_optimizer_2.h
```diff
@@ -221,6 +226,126 @@ class Mesh_global_optimizer_2
...
+	VERTEXSET_ITERATOR itBegin = std::next(moving_vertices.begin(), begin);
+	VERTEXSET_ITERATOR itEnd   = std::next(moving_vertices.begin(), end);
+	if (end == n_vertices)
+	  {
+	    itEnd = moving_vertices.end();
+	  }
+
+	// create and start thread
+	TYPE_THREAD th = std::async(&Mesh_global_optimizer_2::threaded_move, this,
+				    t, std::ref(itBegin), std::ref(itEnd),
```
janetournois, line 328: this line should be replaced by `t, itBegin, itEnd`

yg42: Do you mean this will solve the problem ?

janetournois: It should solve the problem, yes!

janetournois:
Also, I think `moving_vertices` should not be modified with `remove` in `compute_move()`, but after all moves have been computed, but maybe it works as is

Mesh_2/include/CGAL/Mesh_2/Mesh_global_optimizer_2.h
janetournois: However, the crash I was meeting was not the one you mention in the title message.

PR #5677, opened 2021-05-05
CGAL IO namespace

sloriot:
Move documented function/class in subdirectories `CGAL/IO` into the `CGAL::IO` namespace

Follow up of #4255


maxGimeno:
errors: 
 - https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-120/Mesh_2/TestReport_lrineau_Ubuntu-Latest.gz
 - https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-120/Mesh_2_Demo/TestReport_Blake_Windows_MSVC-2019-Community-Release-64bits.gz
 - https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-120/Triangulation_2_Demo/TestReport_lrineau_Fedora-rawhide.gz

maxGimeno:
There is also https://cgal.geometryfactory.com/~mgimeno/test_suite/TESTRESULTS/CGAL-0.0-Ic-36/BGL/TestReport_cgaltest_VS-19.gz
For this last one it is a little tricky, the overload of `read_OFF` has changed:
It's not the specialisation for Polyhedron (in Polyhedron_OFF_iostream.h) that is called anymore, which didn't know about normals at all, but the BGL overload.  
That means that the reading of a NOFF without normals was OK (we always discard normals in the polyhedron version), but the BGL version doesn't like it. 
I don't understand why the Polyhedron overload is not called anymore...

Surface_mesh/include/CGAL/Surface_mesh/IO/OFF.h
```diff
@@ -345,6 +345,48 @@ bool read_OFF(std::istream& is,
...
+}
+
+template <typename Point,
+          typename CGAL_BGL_NP_TEMPLATE_PARAMETERS>
+bool read_OFF(std::string& fname,
+              Surface_mesh<Point>& sm)
+{
+  return read_OFF(fname, sm, parameters::all_default());
+}
+
```
lrineau, line 389: There must be something really wrong with that API, if one needs to define so many overloads!

maxGimeno: WELL, normally BGL would take care of all the extra and only the overload that's already here is necessary, but with the CGAL::IO:: namespace and this ADL problem we discussed, it seems that now, ALL OF THE OVERLOADS need to be defined, so yes, there is something really wrong.

maxGimeno:
`write_vtu.h `can not exist in the same directory of `write_VTU.h` in a windows platform.

Surface_mesh/include/CGAL/Surface_mesh/IO/OFF.h
MaelRL: Is there a conclusion on the "really wrong" aspect?

lrineau: There is no conclusion. I am sorry if any of you two felt attacked by my comment. It was just a statement, without any proposed solution.

The issue seems to be a problem a [double-dispatch](https://en.wikipedia.org/wiki/Double_dispatch), not with dynamic polymorphism but with templates. The main key of that double-dispatch is the face-graph type. So maybe the facing function templates in `CGAL::IO::` could be generic for the first parameter:

```c++
template <typename Input, typename Face_graph>
bool read_OFF(Input&& input,
              Face_graph& graph);

template <typename Input, typename Point
bool read_OFF(Input&& input,
              Surface_mesh<Point>& graph);
              
// ...
```
and then, those function templates could dispatch again on the first parameter `Input input` calling functions in another namespace (`CGAL::IO::internal::`?)...

But I would recommend to refactor that way only for CGAL-5.4. That seems too late for CGAL-5.3. And anyway, the refactoring would be compatible with the API of CGAL-5.3.


... I am also surprised to see non-const references as parameter: `std::string& fname`. That seems to limit who users can call the function, and it should be fixed, in my opinion.

MaelRL: No, I was just confused how moving namespaces across all functions could introduce such ADL issues that require these new overloads. But I haven't looked into it in detail (and so I was just wondering if the problem was already solved).

The `string` parameter should indeed be a `const&`.

maxGimeno:

lrineau:
What about the backward compatibility?! :disappointed:

```
error: no member named 'is_ascii' in namespace 'CGAL'; did you mean 'CGAL::IO::is_ascii'?
      if( CGAL::is_ascii(is) )
          ^~~~~~~~~~~~~~
          CGAL::IO::is_ascii
/home/lrineau/Git/cgal/Stream_support/include/CGAL/IO/io.h:574:13: note: 'CGAL::IO::is_ascii' declared here
```

sloriot:
https://github.com/CGAL/cgal/blob/master/Stream_support/include/CGAL/IO/io.h#L713

lrineau:
> [`master`/Stream_support/include/CGAL/IO/io.h#L713](https://github.com/CGAL/cgal/blob/master/Stream_support/include/CGAL/IO/io.h?rgh-link-date=2023-01-31T13%3A21%3A07Z#L713)

Sorry. The code was a plugin for the 3D demo, in a user branch. And `demo/Polyhedron/CMakeLists.txt` has `add_definitions ( -DCGAL_NO_DEPRECATED_CODE )`. So everything worked as expected.

PR #5680, opened 2021-05-05
Intersections_3: Fix do_intersect(sphere/triangle) if FT has no division

afabri:
## Summary of Changes

The do_intersect function for triangle/sphere was never tested for a number type without division, and had bugs.
The test ` test_intersections_3.cpp`  passes.   I also added the fixes and the test for sphere/segment ,sphere/ray and sphere line which do pass.
 

## Release Management

NYA flag because there is some temporary code behind a macro, need to bench on Windows to take a final decision on "RT functions" being used in "FT functions".

* Affected package(s): Intersections_3
* Issue(s) solved (if any): 

### DIFF
--- a/Distance_3/include/CGAL/squared_distance_3_0.h
+++ b/Distance_3/include/CGAL/squared_distance_3_0.h
@@ -202,6 +202,23 @@ squared_distance_to_plane(const typename K::Vector_3 & normal,
   return FT(dot*dot) /
     FT(wmult((K*)0, squared_length, diff.hw(), diff.hw()));
 }
+  
+template <class K>
+void
+squared_distance_to_plane_RT(const typename K::Vector_3 & normal,
+                             const typename K::Vector_3 & diff,
+                             typename K::RT& num,
+                             typename K::RT& den,
+                             const K& k)
+{
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+  RT dot, squared_length;
+  dot = wdot(normal, diff, k);
+  squared_length = wdot(normal, normal, k);
+  num =  square(dot);
+  den = wmult((K*)0, squared_length, diff.hw(), diff.hw());
+}
 
 
 template <class K>
@@ -218,7 +235,55 @@ squared_distance_to_line(const typename K::Vector_3 & dir,
                               (K*)0, RT(wdot(dir, dir, k)), diff.hw(), diff.hw()));
 }
 
+template <class K>
+void
+squared_distance_to_line_RT(const typename K::Vector_3 & dir,
+                            const typename K::Vector_3 & diff,
+                            typename K::RT& num,
+                            typename K::RT& den,
+                            const K& k,
+                            Cartesian_tag)
+{
+  typedef typename K::Vector_3 Vector_3;
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+  Vector_3 wcr = wcross(dir, diff, k);
+  num = wcr*wcr;
+  den = wmult((K*)0, RT(wdot(dir, dir, k)), diff.hw(), diff.hw());
+}
 
+template <class K>
+void
+squared_distance_to_line_RT(const typename K::Vector_3 & dir,
+                            const typename K::Vector_3 & diff,
+                            typename K::RT& num,
+                            typename K::RT& den,
+                            const K& k,
+                            Homogeneous_tag)
+{
+  typedef typename K::Vector_3 Vector_3;
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+  Vector_3 wcr = wcross(dir, diff, k);
+  FT ft = FT(wcr*wcr)/FT(wmult(
+                               (K*)0, RT(wdot(dir, dir, k)), diff.hw(), diff.hw()));
+  num = Rational_traits<FT>().numerator(ft);
+  den = Rational_traits<FT>().denominator(ft);
+}
+
+template <class K>
+void
+squared_distance_to_line_RT(const typename K::Vector_3 & dir,
+                            const typename K::Vector_3 & diff,
+                            typename K::RT& num,
+                            typename K::RT& den,
+                            const K& k)
+{
+  typedef typename K::Kernel_tag Tag;
+  Tag tag;
+  squared_distance_to_line_RT(dir,diff,num,den,k,tag);
+}
+                                       
 template <class K>
 inline
 bool
--- a/Distance_3/include/CGAL/squared_distance_3_0.h
+++ b/Distance_3/include/CGAL/squared_distance_3_0.h
@@ -202,7 +202,7 @@ squared_distance_to_plane(const typename K::Vector_3 & normal,
   return FT(dot*dot) /
     FT(wmult((K*)0, squared_length, diff.hw(), diff.hw()));
 }
-  
+
 template <class K>
 void
 squared_distance_to_plane_RT(const typename K::Vector_3 & normal,
@@ -283,7 +283,7 @@ squared_distance_to_line_RT(const typename K::Vector_3 & dir,
   Tag tag;
   squared_distance_to_line_RT(dir,diff,num,den,k,tag);
 }
-                                       
+
 template <class K>
 inline
 bool
--- a/Distance_3/include/CGAL/squared_distance_3_0.h
+++ b/Distance_3/include/CGAL/squared_distance_3_0.h
@@ -188,22 +188,6 @@ squared_distance(const typename K::Point_3 & pt1,
   return k.compute_squared_distance_3_object()(pt1, pt2);
 }
 
-
-template <class K>
-typename K::FT
-squared_distance_to_plane(const typename K::Vector_3 & normal,
-                          const typename K::Vector_3 & diff,
-                          const K& k)
-{
-  typedef typename K::RT RT;
-  typedef typename K::FT FT;
-  RT dot, squared_length;
-  dot = wdot(normal, diff, k);
-  squared_length = wdot(normal, normal, k);
-  return FT(dot*dot) /
-    FT(wmult((K*)0, squared_length, diff.hw(), diff.hw()));
-}
-
 template <class K>
 void
 squared_distance_to_plane_RT(const typename K::Vector_3 & normal,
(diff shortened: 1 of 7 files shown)


Distance_3/include/CGAL/squared_distance_3_0.h
```diff
@@ -218,7 +235,55 @@ squared_distance_to_line(const typename K::Vector_3 & dir,
...
+                            Homogeneous_tag)
+{
+  typedef typename K::Vector_3 Vector_3;
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+  Vector_3 wcr = wcross(dir, diff, k);
+  FT ft = FT(wcr*wcr)/FT(wmult(
+                               (K*)0, RT(wdot(dir, dir, k)), diff.hw(), diff.hw()));
+  num = Rational_traits<FT>().numerator(ft);
+  den = Rational_traits<FT>().denominator(ft);
```
lrineau, line 271: I do not understand how that code can work with a number type without division. I see a division operator in the middle!

lrineau: ... Actually, I understand: that is the code for the homogeneous kernel, and it uses `K::FT`.

Distance_3/include/CGAL/squared_distance_3_0.h
```diff
@@ -218,7 +235,55 @@ squared_distance_to_line(const typename K::Vector_3 & dir,
...
+}
```
lrineau, line 272: I am a bit worried by the code duplication between the functions `squared_distance_to_something` and `squared_distance_to_something_RT`.

At least for the homogeneous kernel, one function can be reimplemented easily by the other:
```suggestion
template <class K>
void
squared_distance_to_line_RT(const typename K::Vector_3 & dir,
                            const typename K::Vector_3 & diff,
                            typename K::RT& num,
                            typename K::RT& den,
                            const K& k,
                            Homogeneous_tag)
{
  typedef typename K::FT FT;
  FT ft = squared_distance_to_line(dir, diff, k);
  num = Rational_traits<FT>().numerator(ft);
  den = Rational_traits<FT>().denominator(ft);
}
```

lrineau (approved):
You did a lot of job dealing with Cartesian and Homogeneous. I think the patch can be smaller if you use tag-dispatching in `Intersections_3/include/CGAL/Intersections_3/internal/Triangle_3_Sphere_3_do_intersect.h`, using the `K::Kernel_tag`, and then in the `Distance_3` package only add the `_rt` version for cartesian objects.

Plus the code duplication. Otherwise, I approve the patch.

Distance_3/include/CGAL/squared_distance_3_0.h
afabri: looks like a recursion.

Distance_3/include/CGAL/squared_distance_3_1.h
```diff
@@ -114,6 +194,42 @@ squared_distance(
...
+    typedef typename K::RT RT;
+    typedef typename K::FT FT;
+    // assert that the segment is valid (non zero length).
+    Vector_3 diff = construct_vector(seg.source(), pt);
+    Vector_3 segvec = construct_vector(seg.source(), seg.target());
+    RT d = wdot(diff,segvec, k);
+    if (d <= (RT)0){
+      FT ft = FT(diff*diff);
+      num = Rational_traits<FT>().numerator(ft);
+      den = Rational_traits<FT>().denominator(ft);
```
afabri, line 218: Here a `return`  is missing.

Distance_3/include/CGAL/squared_distance_3_2.h
```diff
@@ -261,6 +261,65 @@ squared_distance_to_triangle(
...
+  const Vector_3 normal = wcross(e1, oe3, k);
+
+  if(normal != NULL_VECTOR
+     && on_left_of_triangle_edge(pt, normal, t0, t1, k)
+     && on_left_of_triangle_edge(pt, normal, t1, t2, k)
+     && on_left_of_triangle_edge(pt, normal, t2, t0, k))
+      {
+        // the projection of pt is inside the triangle
+        inside = true;
+        return squared_distance_to_plane_RT(normal, vector(t0, pt), num, den, k);
```
afabri, line 289: nothing to return

afabri: In fact I think we do not even need this function.

Distance_3/include/CGAL/squared_distance_3_0.h
```diff
@@ -203,6 +204,23 @@ squared_distance_to_plane(const typename K::Vector_3 & normal,
...
+template <class K>
+void
+squared_distance_to_plane_RT(const typename K::Vector_3 & normal,
+                             const typename K::Vector_3 & diff,
+                             typename K::RT& num,
+                             typename K::RT& den,
+                             const K& k)
+{
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
```
MaelRL, line 216: unused typedef (same at other places)

Distance_3/include/CGAL/squared_distance_3_0.h
lrineau: `squared_distance_to_line_RT` with `Homogeneous_tag` as the last argument would call `squared_distance_to_line`. That is not a recursion, because the code of `squared_distance_to_line` is:
https://github.com/CGAL/cgal/blob/b7b2e1955556defdb60088db16ab7d7404c228c7/Distance_3/include/CGAL/squared_distance_3_0.h#L224-L236

MaelRL: Why not go the other way, with:
```
FT f(p...)
{
  	RT num, denum;
    f_RT(p, num, denum)
    return num/denum;
}
```
?

Distance_3/include/CGAL/squared_distance_3_1.h
```diff
@@ -44,6 +44,22 @@ squared_distance(
...
+    const typename K::Point_3 &pt,
+    const typename K::Line_3 &line,
+    typename K::RT& num,
+    typename K::RT& den,
+    const K& k)
+{
+  typedef typename K::Vector_3 Vector_3;
+  typename K::Construct_vector_3 construct_vector;
+  Vector_3 dir(line.direction().vector());
+  Vector_3 diff = construct_vector(line.point(), pt);
```
MaelRL, line 59: That's completely ignoring kernel concepts, but the whole package is afflicted... Might do it in the other PR if I have time.

lrineau: I think that is fine: the bodies of functions of the kernel are implementation details, and they are free to use whatever API they want.

Distance_3/include/CGAL/squared_distance_3_2.h
```diff
@@ -286,7 +286,7 @@ squared_distance_to_triangle_RT(
       {
         // the projection of pt is inside the triangle
         inside = true;
-        return squared_distance_to_plane_RT(normal, vector(t0, pt), num, den, k);
+        squared_distance_to_plane_RT(normal, vector(t0, pt), num, den, k);
```
lrineau, line 289: Now a `return` is missing.
```suggestion
        squared_distance_to_plane_RT(normal, vector(t0, pt), num, den, k);
        return;
```

lrineau: Well, with the `if`/`else`, that return is not mandatory.

Distance_3/include/CGAL/squared_distance_3_0.h
```diff
@@ -218,6 +236,54 @@ squared_distance_to_line(const typename K::Vector_3 & dir,
...
+                            typename K::RT& num,
+                            typename K::RT& den,
+                            const K& k,
+                            Cartesian_tag)
+{
+  typedef typename K::Vector_3 Vector_3;
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+  Vector_3 wcr = wcross(dir, diff, k);
+  num = wcr*wcr;
```
MaelRL, line 252: `wcr = wcross(...)` has removed the homogeneous (den) coefficients, so you can just use `num = wdot(wcr, wcr, k)` and merge both functions

Distance_3/include/CGAL/squared_distance_3_1.h
```diff
@@ -73,6 +89,70 @@ squared_distance(
...
+  typedef typename K::Vector_3 Vector_3;
+  typedef typename K::RT RT;
+  typedef typename K::FT FT;
+
+    Vector_3 diff = construct_vector(ray.source(), pt);
+    const Vector_3 &dir = ray.direction().vector();
+    if (!is_acute_angle(dir,diff, k) ){
+      FT ft = diff*diff;
+      num = Rational_traits<FT>().numerator(ft);
+      den = Rational_traits<FT>().denominator(ft);
```
MaelRL, line 112: Here you can just use:
```
num = wdot(diff, diff, k);
den = square(diff.hw());
```
and merge both functions

Distance_3/include/CGAL/squared_distance_3_1.h
MaelRL: If you look at it as the implementation of `Compute_squared_distance_3`, sure; but the global functions go straight for `internal::squared_distance(O1, O2, K())` and `K` is not necessarily a CGAL kernel.

I guess the error is that the global function should actually call `K().compute_squared_distance_3()(O1, O2)`, which THEN will call `internal::squared_distance(O1, O2)` in the case of CGAL kernels, and in the case of (very hypothetical) third party kernels it will go to their implementation.

lrineau: > If you look at it as the implementation of `Compute_squared_distance_3`, sure; but the global functions go straight for `internal::squared_distance(O1, O2, K())` and `K` is not necessarily a CGAL kernel.

As far as I know, global functions are not supposed to work for anything but CGAL kernels: in `CGAL::Point_3<K>`, the `K` must be one of the CGAL kernels. Maybe we want to change that fact, but for the moment that is hard-coded in a lot of places in the CGAL kernel code.

MaelRL: Right, I forgot that only the internal global function uses Kernel::Object, so it's indeed only CGAL kernels.

Distance_3/include/CGAL/squared_distance_3_0.h
mglisse: If this involves homogeneous where RT!=FT, num/den is probably Fraction_traits::Compose()(num,den) (I don't remember if there is a short notation for it).

maxGimeno:
In Distance_3 dependencies, there are now `Interval_support` and  `Modular_arithmetic`  but they are not referenced in the package_info.

afabri:
Is there still something to do?

MaelRL:
Yes, bench code behind the macro `CGAL_DISTANCE_3_USE_RT_FUNCTIONS_IN_FT_FUNCTIONS` vs the current one (i.e. the `else` of the macro).

maxGimeno:
I made a bench  : https://gist.github.com/maxGimeno/fbae0dae803182a2eb33bdc59097f5e2

Results from Sosno: 

macro defined
638 usable points.
Tests with epick took 0.038 s.
638 usable points.
Tests with epeck took 12.299 s.


macro not defined
638 usable points.
Tests with epick took 0.029 s.
638 usable points.
Tests with epeck took 9.63 s.

afabri:
@MaelRL let's discuss this on Tuesday.

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-134.shtml

PR #5685, opened 2021-05-06
Testsuite: Fix gmp_and_mpfr_dll test

maxGimeno:
## Summary of Changes
With the new versions of those dlls we use on some machines, this test fails because the names are not the same. 
This PR allows it to also pass with the new names. Unfortunately, those dlls have no VersionInfo, so the test will not fail if VersionInfo is missing anymore.
## Release Management


maxGimeno:

PR #5687, opened 2021-05-06
First search for TBB in config mode

maxGimeno:
## Summary of Changes
Add a first call to `find_package(TBB QUIET NO_MODULE)` in our `FindTBB.cmake`, to bypass it in case it is sufficient. This fixes the bug that occurs with oneAPI.
## Release Management

* Affected package(s):Installation
* Issue(s) solved (if any): fix #5676

PR #5692, opened 2021-05-07
CGAL:: Use std atomic and threads

maxGimeno:
## Summary of Changes
Deprecate `cpp11::atomic` and `cpp11::threads` and use std versions instead

## Release Management

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_vertex_base_3.h
+++ b/Mesh_3/include/CGAL/Mesh_vertex_base_3.h
@@ -66,6 +66,13 @@ template <>
 class Mesh_vertex_base_3_base<Parallel_tag>
 {
 public:
+  Mesh_vertex_base_3_base()
+  {}
+
+  Mesh_vertex_base_3_base( const Mesh_vertex_base_3_base& c)
+  {
+    m_erase_counter.store(c.erase_counter());
+  }
 
   // Erase counter (cf. Compact_container)
   unsigned int erase_counter() const
--- a/Mesh_3/include/CGAL/Mesh_vertex_base_3.h
+++ b/Mesh_3/include/CGAL/Mesh_vertex_base_3.h
@@ -28,6 +28,7 @@
 #include <CGAL/Mesh_3/io_signature.h>
 #include <CGAL/Has_timestamp.h>
 #include <CGAL/tags.h>
+#include <atomic>
 
 namespace CGAL {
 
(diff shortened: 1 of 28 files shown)


Mesh_3/include/CGAL/Mesh_vertex_base_3.h
```diff
@@ -81,7 +81,7 @@ class Mesh_vertex_base_3_base<Parallel_tag>
   }
 
 protected:
-  typedef tbb::atomic<unsigned int> Erase_counter_type;
+  typedef std::atomic<unsigned int> Erase_counter_type;
```
lrineau, line 84: Should not you include `<atomic>`?

Mesh_3/include/CGAL/Mesh_vertex_base_3.h
```diff
@@ -66,6 +66,13 @@ template <>
 class Mesh_vertex_base_3_base<Parallel_tag>
 {
 public:
+  Mesh_vertex_base_3_base()
+  {}
+
+  Mesh_vertex_base_3_base( const Mesh_vertex_base_3_base& c)
+  {
+    m_erase_counter.store(c.erase_counter());
+  }
```
lrineau, line 75: I see that this copy-constructor is needed by Mesh_3 global optimizers. Where as the `m_erase_counter` is actually only required by the core of the meshing process, in parallel.

I really wish we had dynamic properties, in CGAL triangulations, like in `CGAL::Surface_mesh`!

maxGimeno:
error : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-125/Polyhedron_Demo/TestReport_lrineau_Fedora.gz

maxGimeno:
errors: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-128/Classification_Examples/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz

lrineau:
#5687 makes CGAL able to find oneAPI-TBB (using the `CONFIG` mode of CMake `find_package`), but `tbb::atomic` does not exist, in oneAPI-TBB. So now this PR #5692 is mandatory to build CGAL using TBB, if both TBB-2020.1 and oneAPI-TBB-2021 are installed in parallel (like it is on my machine).

With this PR, that seems to pass.

maxGimeno:
Successfully tested on https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-129.shtml

PR #5695, opened 2021-05-08
Add with_marks parameter to shell_to_nef

GilesBathgate:
## Summary of Changes

Within my project, I often want to create a union of non-intersecting nef polyhedra. (I call this group, rather than union) Since there are no intersections performing a binary operation is expensive. Using the `delegate` API I create a combined nef by adding another nef using `CGAL::shell_to_nef_3` as can be seen here: https://github.com/GilesBathgate/RapCAD/blob/master/src/cgalprimitive.cpp#L348-L368

This works, but as a side effect, I lose the marks that were set during previous binary operations. This PR adds a new parameter to shell_to_nef_3 to allow the marks to be copied from the original.

## Release Management

* Affected package(s): Nef_3
* Feature/Small Feature (if any): small feature

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/shell_to_nef_3.h
+++ b/Nef_3/include/CGAL/Nef_3/shell_to_nef_3.h
@@ -62,9 +62,10 @@ class Shell_to_nef_3
 
  private:
   SNC_structure& S;
+  bool with_marks;
  public:
 
-  Shell_to_nef_3(SNC_structure& S_) : S(S_) {}
+  Shell_to_nef_3(SNC_structure& S_, bool m_) : S(S_), with_marks(m_) {}
 
   void visit(Vertex_const_handle ) {}
   void visit(Halfedge_const_handle ) {}
@@ -83,7 +84,7 @@ class Shell_to_nef_3
     Vertex_const_handle pv = sf->center_vertex();
     Vertex_handle nv = S.new_vertex();
     nv->point() = pv->point();
-    nv->mark() = true;
+    nv->mark() = with_marks ? pv->mark() : true;
 
     SM_decorator SM(&*nv);
     SHalfedge_around_sface_const_circulator
@@ -91,7 +92,7 @@ class Shell_to_nef_3
 
     SVertex_handle sv_0 =
       SM.new_svertex(pe->source()->point());
-    sv_0->mark() = true;
+    sv_0->mark() = with_marks ? pe->source()->mark() : true;
 #ifndef CGAL_NEF_NO_INDEXED_ITEMS
     sv_0->set_index(pe->source()->get_index());
 #endif
@@ -100,7 +101,7 @@ class Shell_to_nef_3
 
     do {
       SVertex_handle sv = SM.new_svertex(pe->source()->point());
-      sv->mark() = true;
+      sv->mark() = with_marks ? pe->source()->mark() : true;
 #ifndef CGAL_NEF_NO_INDEXED_ITEMS
       sv->set_index(pe->source()->get_index());
 #endif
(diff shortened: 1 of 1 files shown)


sloriot:
I think it would make sense to have a `append()` function in `Nef_polyhedron_3` class using your changes.

Nef_3/include/CGAL/Nef_3/shell_to_nef_3.h
```diff
@@ -62,9 +62,10 @@ class Shell_to_nef_3
 
  private:
   SNC_structure& S;
+  bool with_marks;
```
sloriot, line 65: what do you think about renaming it to `preserve_marks`?

GilesBathgate: Sure I can do that.

GilesBathgate:
> I think it would make sense to have a `append()` function in `Nef_polyhedron_3` class using your changes.

Ok. I have the code for this. Should I change the scope of this pull request, or create a new one?

GilesBathgate:
@sloriot The basic functionality for `append` is implemented now. Let me know what needs doing in terms of adding unit tests/documentation.

sloriot:
Thanks for the update Giles. There are a few things puzzling me that I'd like to check (mainly why the copy constructor does not need the new `bool` parameter to be `true` and why you need a call to `mark_bounded_volume()` while that should be the purpose of backporting the flag values). I'll try to have a look later today.

GilesBathgate:
> why you need a call to `mark_bounded_volume()` while that should be the purpose of backporting the flag values).

Is it because the marks that are copied with preserve_marks are on the unbounded volume? Anyway, I think I found I needed it through trial and error, so I don't have any good justification for it.

sloriot:
Successfully tested in CGAL-5.4-Ic-131 (still need to have a look at the new function)

GilesBathgate:
@sloriot Would it be helpful to post some screenshots with/without the patch?

sloriot:
After a quick look, here is how I think it works:
- `shell_to_nef()` simply import the combinatorics i.e. sphere maps and co. without any volume notion (mark are set to true but it is arbitrary).
- `build_external_structure()` relinks everything and creates volumes
- `mark_bounded_volumes()` flags them as in/out

So after calling your `CGALGroupModifier`, I'd be surprised if the Nef is valid as some maintenance is still needed to restore adjacencies. Did you try calling `is_valid()` after?

GilesBathgate:
@sloriot I've added:
```diff
diff --git a/src/cgalprimitive.cpp b/src/cgalprimitive.cpp
index 58766324..0f3bc142 100644
--- a/src/cgalprimitive.cpp
+++ b/src/cgalprimitive.cpp
@@ -367,6 +367,9 @@ Primitive* CGALPrimitive::group(Primitive* pr)
        CGAL::Mark_bounded_volumes<CGAL::NefPolyhedron3> mbv(true);
        nefPolyhedron->delegate(mbv,false,false);
 
+       if(!nefPolyhedron->is_valid())
+               throw;
+
        this->appendChild(that);
        return this;
```

But I can't get the the application to terminate. My test case is quite simple though, just the `group` of two cubes with cylinders differenced from them (holes)

GilesBathgate:
@sloriot In fact, I realise now that I don't need the Appender/delegate at all, because Nef_3 has direct access to `snc()` (That wasn't the case for my code which is what I ported it from)

lrineau:
@sloriot @afabri This PR was marked https://github.com/CGAL/cgal/labels/Tested but it has been modified since then with a new commit (untested).

And there is still the label https://github.com/CGAL/cgal/labels/under%20discussion. Can you go on with the discussion (and then reach an agreement if possible)?

GilesBathgate:
@sloriot I've added an append method that takes a range. As you can see this is beneficial because the external structure and mark bounded volumes only needs to be called once for N nef_polyhedra in the range.

PR #5697, opened 2021-05-10
Extra white space at bottom of page

albert-github:
Below the bottom blue block we see for the packages some extra white space. For the Manual page this white space is not present.
The reason is the placing of the "hack.js" code.

Note the problem was only observed with Firefox.

The old situation:

image

The corrected situation:

image


maxGimeno:

PR #5699, opened 2021-05-10
Remove duplicate code in SNC_const_decorator/SNC_decorator

GilesBathgate:
## Summary of Changes

This removes code duplication between `SNC_const_decorator` and `SNC_decorator`. The additional template parameter `Traits`, uses `SNC_decorator_traits` to provide both const and non-const handles such that the SNC_decorator version of the code can be replaced by a template instantiation. 

Other functions can potentially be de-duplicated in the same way.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_decorator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_decorator.h
@@ -824,8 +824,9 @@ class SNC_decorator : public SNC_const_decorator<Map> {
     return valid;
   }
 
-  template <typename Visitor>
-  void visit_shell_objects(SFace_handle f, Visitor& V) const;
+  template <typename Visitor, typename Traits = Self::Decorator_traits>
+  void visit_shell_objects(typename Traits::SFace_handle f, Visitor& V) const
+  { Base::template visit_shell_objects<Visitor, Traits>(f, V); }
 
   Vertex_iterator   vertices_begin()   { return sncp()->vertices_begin(); }
   Vertex_iterator   vertices_end()     { return sncp()->vertices_end(); }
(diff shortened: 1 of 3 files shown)


afabri:
@GilesBathgate is the change for the maps `Done`  needed for your code factorization, or did this just sneak in accidentally?

Nef_3/include/CGAL/Nef_3/SNC_decorator.h
```diff
@@ -851,146 +700,14 @@ class SNC_decorator : public SNC_const_decorator<Map> {
...
-template <typename EW>
-template <typename Visitor>
-void SNC_decorator<EW>::
-visit_shell_objects(SFace_handle f, Visitor& V) const
-{
-  typedef typename SM_decorator::SHalfedge_around_sface_circulator
-    SHalfedge_around_sface_circulator;
-  std::list<SFace_handle> SFaceCandidates;
-  std::list<Halffacet_handle> FacetCandidates;
-  CGAL::Generic_handle_map<bool> Done(false);
```
GilesBathgate, line 902: @afabri It's not a "sneak in", this is how it was implemented in SNC_decorator. It does however require the void cast in Generic_handle_map.

Nef_3/include/CGAL/Nef_3/SNC_decorator.h
```diff
@@ -824,8 +672,9 @@ class SNC_decorator : public SNC_const_decorator<Map> {
     return valid;
   }
 
-  template <typename Visitor>
-  void visit_shell_objects(SFace_handle f, Visitor& V) const;
+  template <typename Visitor, typename Traits = Self::Decorator_traits>
```
afabri, line 675: Is this `typename Traits = Self::Decorator_traits` the best way to get the const or not const version of the decorator traits class in the loop? I don't see that we take anything else than this default.

GilesBathgate: I can change this, but I am not sure of the syntax. Is the syntax:

```c++
template <typename Visitor, typename Traits>
void visit_shell_objects<Visitor,Self::Decorator_traits>(SFace_handle f, Visitor& V) const;
```

GilesBathgate: @afabri Oh I remember why I did this now in SNC_const_decorator it uses Self::Decorator_traits AKA `SNC_const_decorator::Decorator_traits`, but in SNC_decorator it uses `SNC_decorator::Decorator_traits` so it does take something other than the default, despite the name looking idential.

afabri: I think I just understood it.   In the call `Base::template visit_shell_objects<Visitor, Traits>(f, V);`  you call  the base class function but with the non-default traits.   So we might remove the `Traits`  in this overload in the non-const derived class, as for this overload we will never take something else.

GilesBathgate: Ok, so I think I could change it like this:

```diff
diff --git a/Nef_3/include/CGAL/Nef_3/SNC_decorator.h b/Nef_3/include/CGAL/Nef_3/SNC_decorator.h
index 1074672824..791cbdd7d6 100644
--- a/Nef_3/include/CGAL/Nef_3/SNC_decorator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_decorator.h
@@ -342,20 +342,17 @@ class SNC_decorator : public SNC_const_decorator<Map> {
     return( e1!=e2 && SD.cyclic_adj_succ(e2)==e1);
   }
 
-  template <typename Traits = Self::Decorator_traits>
-  typename Traits::Halffacet_handle get_visible_facet( const typename Traits::Vertex_handle v,
+  typename Self::Decorator_traits::Halffacet_handle get_visible_facet( const typename Self::Decorator_traits::Vertex_handle v,
                                                        const Ray_3& ray) const
-  { return Base::template get_visible_facet<Traits>(v, ray); }
+  { return Base::template get_visible_facet<typename Self::Decorator_traits>(v, ray); }
 
-  template <typename Traits = Self::Decorator_traits>
-  typename Traits::Halffacet_handle get_visible_facet( const typename Traits::Halfedge_handle e,
+  typename Self::Decorator_traits::Halffacet_handle get_visible_facet( const typename Self::Decorator_traits::Halfedge_handle e,
                                                        const Ray_3& ray) const
-  { return Base::template get_visible_facet<Traits>(e, ray); }
+  { return Base::template get_visible_facet<typename Self::Decorator_traits>(e, ray); }
 
-  template <typename Traits = Self::Decorator_traits>
-  typename Traits::Halffacet_handle get_visible_facet( const typename Traits::Halffacet_handle f,
+  typename Self::Decorator_traits::Halffacet_handle get_visible_facet( const typename Self::Decorator_traits::Halffacet_handle f,
                                                        const Ray_3& ray) const
-  { return Base::template get_visible_facet<Traits>(f, ray); }
+  { return Base::template get_visible_facet<typename Self::Decorator_traits>(f, ray); }
 
   Halffacet_handle get_visible_facet( const Vertex_handle v,
                                       const Segment_3& ray) const
@@ -672,9 +669,9 @@ class SNC_decorator : public SNC_const_decorator<Map> {
     return valid;
   }
 
-  template <typename Visitor, typename Traits = Self::Decorator_traits>
-  void visit_shell_objects(typename Traits::SFace_handle f, Visitor& V) const
-  { Base::template visit_shell_objects<Visitor, Traits>(f, V); }
+  template <typename Visitor>
+  void visit_shell_objects(typename Self::Decorator_traits::SFace_handle f, Visitor& V) const
+  { Base::template visit_shell_objects<Visitor,typename Self::Decorator_traits>(f, V); }
```

I do prefer it the way it is.

afabri: Maybe I am wrong, but instead of `return Base::template get_visible_facet<typename Self::Decorator_traits>(e, ray);` can't  you just write `return Base::template get_visible_facet<Decorator_traits>(e, ray);`?
And as the above functions are in the derived class, instead of `typename Traits::Halffacet_handle` just `Halffacet_handle`  should do it.

GilesBathgate: @afabri That's simpler ill give it a try. Thanks!

afabri: If you do it quickly it might make it  into the testsuite with 5.3 in mind.  You see that I work on the backlog.

Nef_3/include/CGAL/Nef_3/SNC_decorator.h
```diff
@@ -342,172 +342,20 @@ class SNC_decorator : public SNC_const_decorator<Map> {
...
-                                      const Ray_3& ray) const
-    /*{\Mop when one shoot a ray |ray| in order to find the facet below to
-      an object, and vertex |v| is hit, we need to choose one of the facets
-      in the adjacency list of |v| such that it could be 'seen' from the
-      piercing point of the |ray| on the sphere map on |v|.  We make it just
-      locating the sphere facet |sf| pierced by |ray| and taking the adjacent
-      facet to one of the sphere segments on the boundary of |sf|.
-      \precondition |ray| target is on |v| and the intersection between
-      |ray| and the 2-skeleton incident to v is empty. }*/ {
+  template <typename Traits = Self::Decorator_traits>
```
afabri: For this and the next functions it seems also that you can fully remove the template and hardcode the return type, and then call the function from `Base`  with `Decorator_traits`  without `Self::`

GilesBathgate: I think I've already made this change in: https://github.com/CGAL/cgal/pull/5699/commits/77a03f2084e9f6f78527f45144bc3d95ab2fb4c8

afabri (approved):

Nef_3/include/CGAL/Nef_3/SNC_decorator.h
```diff
@@ -343,171 +343,16 @@ class SNC_decorator : public SNC_const_decorator<Map> {
...
-    // We have to check which of the two halffacet is visible from
-    // the ray.
-
-    if(rv*res->plane().orthogonal_vector() > FT(0))
-      res = res->twin();
-
-    CGAL_NEF_TRACEN("return "<<res->plane());
-    return res; // never reached
-  }
+                                                       const Ray_3& ray) const
```
afabri, line 350: @GilesBathgate in case you  are motivated you could quickly fix the indentation.

maxGimeno:

PR #5701, opened 2021-05-11
Testsuite: Adapt ctest scripts for use with cygwin

maxGimeno:
## Summary of Changes
Change the ctest scripts to work on windows and  directly use MSBuild and its internal parallelism feature for compilation tests. 

## TODO
 - [x] Update collect_results_from_ctest to get the "t" for external warnings
 - [x] Find out why (and fix) the CXX FLAGS that do not appear in the testsuite page
## Release Management

* Affected package(s):Testsuite

### DIFF
--- a/Installation/cmake/modules/CGAL_add_test.cmake
+++ b/Installation/cmake/modules/CGAL_add_test.cmake
@@ -84,7 +84,9 @@ function(cgal_add_compilation_test exe_name)
       APPEND PROPERTY LABELS "${PROJECT_NAME}")
     set_property(TEST "compilation_of__${exe_name}"
       APPEND PROPERTY FIXTURES_REQUIRED "check_build_system_SetupFixture")
-  elseif(NOT TEST compilation_of__${PROJECT_NAME})#CMAKE_VS_MSBUILD_COMMAND
+  elseif(NOT TARGET compilation_of__${PROJECT_NAME})#CMAKE_VS_MSBUILD_COMMAND
+    #this target is just a flag, to deal with the scope problem with the tests
+    add_custom_target(compilation_of__${PROJECT_NAME})
     add_test(NAME "compilation_of__${PROJECT_NAME}"
       COMMAND ${TIME_COMMAND} "${CMAKE_VS_MSBUILD_COMMAND}" "${PROJECT_BINARY_DIR}/${PROJECT_NAME}.sln" "-m:$ENV{NUMBER_OF_PROCESSORS}" "/t:Build" "/p:Configuration=$<CONFIG>")
     set_property(TEST "compilation_of__${PROJECT_NAME}"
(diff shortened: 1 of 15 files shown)


Installation/cmake/modules/CGAL_add_test.cmake
```diff
@@ -32,6 +32,9 @@ if(CGAL_RUN_TESTS_THROUGH_SSH)
   find_program(scp_executable scp)
 endif()
 
+if(CGAL_WIN32_CMAKE_ON_CYGWIN)
+  find_program(MSBUILD_CMD NAMES msbuild.exe)
+endif()
```
lrineau, line 37: CMake-on-Cygwin is not the right way to test.

1. CYgwin cannot be mandatory to run CTest on Windows.
2. There are other compilers on Windows than `cl.exe`.

You should use the variable [`CMAKE_VS_MSBUILD_COMMAND`](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_MSBUILD_COMMAND.html). That variable is only defined by Visual Studio generators, and that is exactly for those generators we need another way to test.

Installation/cmake/modules/CGAL_add_test.cmake
```diff
@@ -74,13 +77,26 @@ function(expand_list_with_globbing list_name)
...
 
 function(cgal_add_compilation_test exe_name)
-  if(TEST compilation_of__${exe_name})
-    return()
-  endif()
-  add_test(NAME "compilation_of__${exe_name}"
-    COMMAND ${TIME_COMMAND} "${CMAKE_COMMAND}" --build "${CMAKE_BINARY_DIR}" --target "${exe_name}" --config "$<CONFIG>")
-  set_property(TEST "compilation_of__${exe_name}"
-    APPEND PROPERTY LABELS "${PROJECT_NAME}")
+  if(NOT CGAL_WIN32_CMAKE_ON_CYGWIN)
```
lrineau, line 80: I suggest to test the variable [`CMAKE_VS_MSBUILD_COMMAND`](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_MSBUILD_COMMAND.html) directly:
```suggestion
  if(NOT CMAKE_VS_MSBUILD_COMMAND)
```

lrineau (changes requested):

Polyhedron/demo/Polyhedron/Plugins/Camera_position/CMakeLists.txt
```diff
@@ -7,5 +7,6 @@ polyhedron_demo_plugin(camera_positions_plugin
     ${cameraUI_FILES}
     KEYWORDS Viewer)
 
+add_dependencies(camera_positions_plugin demo_framework)
 target_link_libraries( camera_positions_plugin PUBLIC demo_framework)
 target_link_libraries(camera_positions_plugin PUBLIC demo_framework)
```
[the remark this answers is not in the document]
maxGimeno: Linking a plugin to a target is unsufficient to be sure it will not be built before (or at the same time as) the dependency, It was one of the few fixes I put into place before encountering the `.tlog` bug. I figured it could not harm to leave them here.

maxGimeno:
The ctest testsuite now runs fine on Sosno, I think it is ready for integration.

lrineau:
Is there anything in it that should be tested in the regular testsuite?

maxGimeno:
I think so, to check I didn't break anything for the mac ctest suite

maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-138.shtml

PR #5702, opened 2021-05-12
Replace CGAL_NEF_DEBUG with CGAL_USE_TRACE

GilesBathgate:
## Summary of Changes

Following on from #5246 I've since discovered three places in the code where `#ifdef CGAL_NEF_DEBUG` is used to disable the compilation of various bits of code. This never worked because CGAL_NEF_DEBUG is defined at the top of the file.

Simply replacing these for the newly introduced `CGAL_USE_TRACE` define (introduced in #5246) solves the problem and ensures this extra code isn't included unless explicitly wanted.

## Release Management

* Affected package(s): Nef_2 / Nef_3

### DIFF
--- a/Nef_2/include/CGAL/Nef_2/Segment_overlay_traits.h
+++ b/Nef_2/include/CGAL/Nef_2/Segment_overlay_traits.h
@@ -39,7 +39,7 @@
 #include <sstream>
 
 namespace CGAL {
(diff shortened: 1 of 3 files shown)


afabri (approved):

maxGimeno:

PR #5705, opened 2021-05-12
Clean-up BOOST_VERSION and GCC_versions tests

### LINKED ISSUE #5682: Simplify tests of BOOST_VERSION
## Issue Details

In CGAL, there are a lot of preprocessor conditionals using `BOOST_VERSION`. Since CGAL-5.2, Boost>=1.66 is required, so we remove almost all conditionals.

```
% ack -h --output='$1' 'BOOST_VERSION *[<>=]+ *([0-9]+)' ~/Git/cgal-master --cpp | sort | uniq -c | sort -n 
      1 103500
      1 104600
      1 105100
      1 105700
      1 106000
      1 107000
      2 103900
      2 104100
      2 105800
      2 106300
      3 104400
      3 104700
      3 105500
      3 106200
      4 104300
      4 105300
      4 106600
      5 104000
      7 103600
      7 105000
    100 105600
```
Among those tests, only the single one about Boost 1.70 can be kept:
```
/home/lrineau/Git/cgal-master/Number_types/include/CGAL/boost_mp.h
24:    (!defined _MSC_VER || BOOST_VERSION >= 107000)
```

All the other ones can be removed because they tests Boost versions <=1.66.

### LINKED ISSUE #5683: Simplify tests of GCC version
## Issue Details

Since CGAL-5.0, gcc version 6 is required. There are a lot of preprocessor conditionals, in CGAL, about older versions of GCC:

```
% ack -h --output='$2' '(BOOST_GCC|GCC_VERSION|__GNUC__) *[<>=]+ *([0-9]+)' ~/Git/cgal-master --cpp | sort | uniq -c | sort -n 
      1 60000
      2 40600
      2 40800
      2 9
     11 4
     12 40900
     19 3
     95 40500
```

As CGAL now only supports gcc 6 or later, only the two tests about `__GNUC__` and version 9 should be kept:
```
/home/lrineau/Git/cgal-master/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
24:#if  __GNUC__ >= 9
41:#if __GNUC__ >= 9
```

maxGimeno:
## Summary of Changes
 - Removes all tests for BOOST_VERSION made useless now that we require boost >=1.66
 - Simplify the tests for BOOST_GCC , GCC_VERSION and __GNUC__
## Release Management

* Issue(s) solved (if any): fix #5682, fix #5683


lrineau:
Commit 516d921ea87016d057f5371b3352e43ce881aa3d has removed all of the `Apollonius_graph_2` package!

lrineau:
I have reapplied the part of e6c767d5c9869ab050102efe0dd4629ece65a447 that was in `Apollonius_graph_2`: commit https://github.com/CGAL/cgal/pull/5705/commits/45b73adf2c4b474449e80e7fc96a1422421f7959.

mglisse:
Note than clang defines `__GNUC__=4` and `__GNUC_MINOR__=2`, so testing for recent gcc was also implicitly testing that this was not clang, whereas `defined(__GNUC__)` accepts clang. Whether that's desirable may depend on each case, but it is a change.

lrineau:
> Note than clang defines `__GNUC__=4` and `__GNUC_MINOR__=2`, so testing for recent gcc was also implicitly testing that this was not clang, whereas `defined(__GNUC__)` accepts clang. Whether that's desirable may depend on each case, but it is a change.

Strange. We'll see what happens to the test suite.

mglisse:
I checked FPU.h, and there we are already in a #else of a check for llvm, so it looks safe.
In config.h though, this undermines the tests with `__has_feature`, which is was mostly a clang thing, where `__GNUC__` is always defined. Maybe all supported versions of clang support those features and we could simplify the tests further, but if not... For instance, Apple-clang was quite late in getting thread_local, I don't know if CGAL still supports versions without it.

maxGimeno:
all broken https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-11.shtml

sloriot:
Stream_support is all red. I guess it is a consequence of this PR.

maxGimeno:
The test was not tested in master, the include of boost/version.hpp was done too late, et BOOST_VERSION was unknown at the time it was tested. Removing that #if revealed the error. Polygon_2 doesn't have any `emplace_back` function....

maxGimeno:
tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-16.shtml

PR #5712, opened 2021-05-14
Remove redundant code in side_of_plane_3

GilesBathgate:
## Summary of Changes

This replaces #5708. Firstly this removes a check in the function side_of_plane_3 that tests `if(plane == Plane_3(0,0,0,0))`. 

>The check is needed so that when a parameter isn't given, an empty default parameter indicates that it should be populated from the points in the Iterator. However the function is never called without the plane parameter so the code is probably redundant, and the comment says it's not tested!
> The operator== for an empty plane isn't trivial. For cartesian kernels, it invokes equal_planeC3 which then in tern invokes equal_directionC3 which calculates the signs of 3 determinants...
>  Removing the redundant code doesn't give much of a performance boost (about 2% in my tests) but I think it's worthwhile nonetheless.

Secondly, the PR makes use of projection traits to simplify and remove the need for much of the code.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
+++ b/Nef_3/include/CGAL/Nef_3/bounded_side_3.h
@@ -18,109 +18,36 @@
 #include <CGAL/license/Nef_3.h>
 
 
-#include <CGAL/basic.h>
+#include <CGAL/assertions.h>
 #include <CGAL/Polygon_2_algorithms.h>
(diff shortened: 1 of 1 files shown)


afabri (approved):

maxGimeno:
Successfully tested on https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-129.shtml

PR #5714, opened 2021-05-15
Remove the OnSideMapRC when not needed

GilesBathgate:
## Summary of Changes

In SNC_k3_tree_traits a map is used to store the results of point comparisons. Two techniques are used, one of which is supposed to exploit reference counting (whatever that means) and is enabled by the define `CGAL_NEF_EXPLOIT_REFERENCE_COUNTING`, however regardless of whether the macro is defined the class member `OnSideMapRC` is present. (since its default constructed this has an overhead as discussed before with 512 elements)

This PR simply places OnSideMapRC into an `#ifdef` block.

I also noticed that p1 is not passed by reference, which is probably unintentional but the fix too small for its own PR.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
@@ -159,7 +159,6 @@ public:
 #ifdef CGAL_NEF_EXPLOIT_REFERENCE_COUNTING
   bool reference_counted;
 #endif
-  SNC_decorator D;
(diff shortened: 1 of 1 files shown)


afabri (approved):

afabri:
@GilesBathgate @maxGimeno I marked several Nef_3 pull requests as for 5.3 beta, but I am not completely sure that they are cumullatively mergeable, or if you Giles have to put an order.

GilesBathgate:
Just let me know which ones you want to combine and I can create a merged branch.

maxGimeno:

PR #5731, opened 2021-05-27
Hash value of handles use directly the address of the node

### LINKED ISSUE #5728: Undefined behaviour with taking the address of a dereferenced nullptr in HalfedgeDS_list operator=
## Issue Details

In HalfedgeDS_list::pointer_update(), h_map[Halfedge_const_iterator()] calls Unique_hash_map::operator\[\](), which calls HashFunctor::operator(), which calls the iterator's operator*() in:
`
      operator()(const H& h)
      {
        return std::size_t(&*h) /
          sizeof( typename std::iterator_traits<H>::value_type);
      }
`
The iterator is default-constructed, holding a nullptr. operator*() creates a null reference by de-referencing the nullptr, which is undefined behaviour, even if the null reference is never accessed (as in this case).

Sorry I don't know enough about the internals of the library to know the best way to fix it.

This was found with UBsan's -fsanitize=null option ("Use of a null pointer or creation of a null reference").

Reference: N4861 (https://timsong-cpp.github.io/cppwp/n4861/dcl.ref)
9.3.3.2:
> A reference shall be initialized to refer to a valid object or function. [ Note: In particular, a null reference cannot exist in a well-defined program, because the only way to create such a reference would be to bind it to the “object” obtained by indirection through a null pointer, which causes undefined behavior. As described in [class.bit], a reference cannot be bound directly to a bit-field. — end note ]

## Source Code

This is called, for example, in the implicit copy constructor of Polyhedron_3.

`
Polyhedron_3 a;
Polyhedron_3 b(a);
`
(run with the UBsan option above to repro)

## Environment

* Compiler: clang++
* Specific flags used (if any): -fsanitize=null
* CGAL version: d9b0e84d8d4d1299e14a8bdc78c2ad414552640c

sloriot:
Fixes #5728

### DIFF
--- a/Hash_map/include/CGAL/Handle_hash_function.h
+++ b/Hash_map/include/CGAL/Handle_hash_function.h
@@ -37,6 +37,16 @@ namespace internal{
           sizeof( typename std::iterator_traits<H>::value_type);
       }
     };
+
+    template <class H>
+    struct Hash_functor<H*>{
+      std::size_t
+      operator()(const H* h)
+      {
+        return std::size_t(h) /
+          sizeof(H);
+      }
+    };
   }
 }
 
--- a/Hash_map/include/CGAL/Handle_hash_function.h
+++ b/Hash_map/include/CGAL/Handle_hash_function.h
@@ -33,7 +33,7 @@ namespace internal{
       std::size_t
       operator()(const H& h)
       {
-        return std::size_t(&*h) /
+        return std::size_t(h.operator->()) /
           sizeof( typename std::iterator_traits<H>::value_type);
       }
     };
--- a/STL_Extension/include/CGAL/In_place_list.h
+++ b/STL_Extension/include/CGAL/In_place_list.h
@@ -179,7 +179,7 @@ namespace internal {
 template <class T, class Alloc>
   std::size_t hash_value(const In_place_list_iterator<T,Alloc>&  i)
   {
-    T* ptr = &*i;
+    T* ptr = i.operator->();
     return reinterpret_cast<std::size_t>(ptr)/ sizeof(T);
   }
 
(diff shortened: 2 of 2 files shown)


lrineau (approved):

GilesBathgate:
@sloriot Ah of course, thats a good solution for when H was a pointer.

afabri:
Is that not a bug fix which should be integrated earlier than 5.4???

GilesBathgate:
@afabri The "bug" was introduced on Jul 10, 2001 
in https://github.com/CGAL/cgal/commit/203e42a0f41fd3f5f4fad63f9d17475f2ac79fda#diff-9a4247c554808c2ea17f2d430f2288d2a11911636c7810630509bddef3f1cd9b

afabri:
So unless we have really good reasons I would still be in favor to  put it into the upcoming release.

lrineau:
> So unless we have really good reasons I would still be in favor to put it into the upcoming release.

I am OK with that. But it will not be a blocker for 5.3-beta1, that is already one month late.

Hash_map/include/CGAL/Handle_hash_function.h
```diff
@@ -33,10 +33,20 @@ namespace internal{
       std::size_t
       operator()(const H& h)
       {
-        return std::size_t(&*h) /
+        return std::size_t(h.operator->()) /
```
tkoeppe, line 36: If you'd like to avoid the functional-style cast here (which is equivalent to the overpowered C-style cast), you could use the appropriate `reinterpret_cast<std::size_t>` instead, which is idiomatic for reinterpreting pointers as integers and vice versa.

GilesBathgate: Maybe this can be applied more generally using clang-tidy: https://clang.llvm.org/extra/clang-tidy/checks/google-readability-casting.html

lrineau: I am in favor of using that clang rule. @maxGimeno Could you give it a try? First on a small set of CGAL, like a few CGAL plugins.

GilesBathgate:
It looks like this PR will fix #5227

maxGimeno:

PR #5737, opened 2021-05-28
[Feature] Weights New Package

danston:
This PR proposes a new Feature called `Weights` that implements a set of analytic and barycentric weights.

The weights implemented are:
- Wachspress weight
- authalic weight
- discrete harmonic weight
- cotangent weight
- mean value weight
- tangent weight
- three-point family weight
- uniform weight
- Shepard weight
- inverse distance weight
- uniform region weight
- triangular region weight
- barycentric region weight
- Voronoi region weight
- mixed Voronoi region weight

This package is used in the new version (PR #5738) of the `Barycentric_coordinates_2` package for computing weights. It will also be used later to update several CGAL packages with new weights: `Polygon mesh Processing`, several `Surface Mesh` related packages, and `Heat Method` to unify the weight usage among different CGAL packages.

## Release Management

**This PR depends on #5736.**

* Affected package(s): Feature, New Package
* Issue(s) solved (if any): no issues
* Feature/Small Feature (if any): [Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Weight_interface)
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.geometryfactory.com/~danston/Weights/index.html)
* License and copyright ownership: GF

PR #5736, opened 2021-05-28
Projection Traits With Centroid 2 and Determinant 2

danston:
This little PR extends the `Kernel 23 Projection Traits` with two objects:
- the one for computing `2D centroids`
- the one for computing `2D determinants`

This PR is required by the new package called `Weights` (PR #5737) where the two objects above make the part of the `Weight` concept. 

Tasks:
- [x] Check memory leaking.
- [x] Remove tmp modif in the docs.

## Release Management

* Affected package(s): Kernel_23
* Issue(s) solved (if any): no issues
* Feature/Small Feature (if any): [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Projection_Traits_With_Centroid_And_Determinant) pre-approved on June 25, 2021 
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.geometryfactory.com/~danston/Kernel_23-projection-traits/classCGAL_1_1Projection__traits__xy__3.html)
* License and copyright ownership: no change

PR #5738, opened 2021-05-28
[Feature] Barycentric Coordinates 2 Revision

danston:
This PR proposes a new updated and revised version of the `Barycentric_coordinates_2` package. Many things have been changed including the API that became much more flexible + the new harmonic coordinates are added. The docs and tests have been extended and improved, too.

## Release Management

* Affected package(s): `Barycentric_coordinates_2`
* Issue(s) solved (if any): no issues
* Feature/Small Feature (if any): Feature
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.geometryfactory.com/~danston/Barycentric_coordinates_2/index.html)
* License and copyright ownership: Inria, GF


danston:
A few points before testing:
- I tested it locally and it is ok.
- I checked it for memory leaks and it is ok.
- This PR should be tested together with the #5737 since it depends on the weights in that package.
- This PR should be tested together with the #5736 since several tests depend on the new modifications added in the projection traits.
- `Delaunay_domain_2.h` header depends on the `Mesh_2` package. Is it ok license-wise? @lrineau, @afabri Could you please check?

lrineau:
> A few points before testing:
>     * `Delaunay_domain_2.h` header depends on the `Mesh_2` package. Is it ok license-wise? @lrineau, @afabri Could you please check?

It is not really a question for me: licensewise, @afabri, would it be a problem if the 2D Delaunay Mesh Generator became a new dependency of barycentric coordinates? In my opinion it is an issue, because that will probably change a lot the price of the commercial license.

danston:
In fact, the `Delaunay_domain_2` is the only header that depends on `Mesh_2` and there is a concept called `DiscretizedDomain_2.h` so, in principle, if the users provide their own way to discretize a 2D polygon, they can avoid this dependency but computing harmonic coordinates without discretizing the interior part of the polygon is not possible that is why this class is introduced where the discretization is constructed using `Mesh_2`.

afabri:
The deprecated functions are in a `Deprecated` namespace, so it still just breaks code as explained [here](https://cgal.geometryfactory.com/~danston/Barycentric_coordinates_2/index.html#title12). Deprecated means that old code still compiles and only a warning is issued. After a discussion with Dmitry it seams that this can be achieved.

danston:
> The deprecated functions are in a `Deprecated` namespace, so it still just breaks code as explained [here](https://cgal.geometryfactory.com/~danston/Barycentric_coordinates_2/index.html#title12). Deprecated means that old code still compiles and only a warning is issued. After a discussion with Dmitry it seams that this can be achieved.

I have fixed that.

danston:
I have also checked that none of the headers includes the one that depends on `Mesh_2` now.

lrineau:
@danston Please mention the PR this one depends on, in the first message.

Barycentric_coordinates_2/test/Barycentric_coordinates_2/test_dh_deprecated_api.cpp
```diff
@@ -0,0 +1,74 @@
+#include <cmath>
+#include <cassert>
+#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
```
lrineau, line 3:
```suggestion
#include <CGAL/internal/disable_deprecation_warnings_and_errors.h>
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
```

All tests of deprecated functions or classes should include that header `<CGAL/internal/disable_deprecation_warnings_and_errors.h>`, to remove warnings or errors about deprecated code.

danston: Oo, I did not know that. Thanks. I will fix that.

danston: Fixed.

sloriot:
There are warnings in the testsuite. I already fixed the case issue in the cmake script (did it in the doc yesterday but did not think about cmake :( )

danston:
Actually, this error is super weird. I already fixed that many times and after pushing it gets back. Maybe it is my editor. I have no idea why this happens((

sloriot:
Last errors:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-30/Barycentric_coordinates_2/TestReport_lrineau_Fedora-32.gz

danston:
@sloriot In the last test suite lc-31, there are no errors, and the ones, you refer, do not seem to be caused by the BC package imho. Are they?

sloriot:
Ic-31 = master

danston:
@sloriot I think now it can be tested again. Thanks!

PR #5739, opened 2021-05-31
Polyhedron Demo: Fix Scene Recentering while Scaled

maxGimeno:
## Summary of Changes

If the user asks for a scene recentering while the scnee is scaled, the camera gets lost. This is a fix for that problem.
## Release Management
Base don 5.2 because the Scene Scaling appeared in 5.2


maxGimeno:
Successfully tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-138.shtml

PR #5741, opened 2021-05-31
[Small Feature] OSQP Support in Solver Interface

danston:
This PR adds the [OSQP solver](https://osqp.org/) support in CGAL. This solver is required in order to efficiently compute the convex Quadratic Programming (QP) problem, arising in the context of the new CGAL package called `Shape Regularization` (see #5742). This PR also better structures the `Solver_interface` package by splitting all solvers into several common categories. An additional example is added, too.

Tasks:
- [x] Check memory leaking.
- [x] Remove tmp modif in the docs.

## Release Management

* Affected package(s): `Solver_interface`
* Issue(s) solved (if any): no issues
* Feature/Small Feature (if any): [Small Feature](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/OSQP_Solver)  pre-approved on June 28, 2021
* Link to compiled documentation (obligatory for small feature): [docs](https://cgal.geometryfactory.com/~danston/Solver_interface/classCGAL_1_1OSQP__quadratic__program__traits.html)


afabri:
- Instead of `#if defined(CGAL_USE_OSQP)` deal with it in the CMakeLists.txt
- The brief documentation should start lower case 
- The sentence "It is also possible to derive new models from other quadratic programming libraries such as OOQP for example." is not really of interest.
- "Note that you should .. "        Use \note      Also "should" makes no sense in a spec. Do you mean "must"?  Or is it correct, but a waste of time?
- "all existing/previous entries".   Remove existing and previous and maybe even the entire sentence.
- So this solver supports any `FT` ?
- Does one use just the C OSQP, or the C++ wrapper?

danston:
@afabri Thank you for the review.

- `Instead of #if defined(CGAL_USE_OSQP) deal with it in the CMakeLists.txt` - And what about other examples? The mixed-integer example in this package has been written this way that is it is dealing with libraries inside the `cpp`. I just followed that style. Should I change them as well?

- `So this solver supports any FT ?` - As input yes, internally everything is converted to double because the OSQP requires `c_float` types.

- `Does one use just the C OSQP, or the C++ wrapper?` - It is the C version.

afabri:
It is hard to see which review items are done and which not.   Maybe copy and mark done as we do on the wiki.
Concerning the `FT` template parameter I am not sure that it makes sense if the package only works with `double`. @sloriot @lrineau your opinon?
@MaelRL had problems  with underconstrained systems.  Is that now fixed? Can it be underconstrained? If not then this merits maybe a `\warning`.

danston:
- Instead of #if defined(CGAL_USE_OSQP) deal with it in the CMakeLists.txt
  -  **Done.** I fixed it both for the osqp and mixed_integer examples.
- The brief documentation should start lower case
  - **Done.** I fixed all brief.
- The sentence "It is also possible to derive new models from other quadratic programming libraries such as OOQP for example." is not really of interest.
  - **Done.** I removed it.
- "Note that you should .. " Use \note Also "should" makes no sense in a spec. Do you mean "must"? Or is it correct, but a waste of time?
  - **Done.** I rephrased that.
- "all existing/previous entries". Remove existing and previous and maybe even the entire sentence.
  - **Done.** I removed that.
- So this solver supports any FT ?
  - No, it is C float.
- Does one use just the C OSQP, or the C++ wrapper?
  - It is the C version.

danston:
> It is hard to see which review items are done and which not. Maybe copy and mark done as we do on the wiki.
- Done. See above.
> Concerning the FT template parameter I am not sure that it makes sense if the package only works with double.
- I think it is nice because the user should not think about the provided type and any conversions in this case. E.g. if he has an exact input type, we do all the conversions internally and he should not bother about it.
> Mael had problems with underconstrained systems. Is that now fixed? Can it be underconstrained? If not then this merits maybe a \warning. 
- As far as I know, the new API is much more flexible now and Mael has succeeded to solve his problem with the OSQP but you are right, the warning about the underconstrained setting is definitely missing. I have added it.
> Is a c_float a float or a double?
- It is the C version of the library and internally it is using `c_float`, `c_int`, etc.

Solver_interface/doc/Solver_interface/Solver_interface.txt
```diff
@@ -23,9 +23,9 @@ For mixed integer programs (either constrained or unconstrained),
 we provide models using the \ref thirdpartySCIP and \ref thirdpartyGLPK
 libraries.
 
-For linear and quadratic programs, we provide a model using the built-in
-\ref PkgQPSolver "CGAL Linear and Quadratic Programming Solver" and a model using
-the external \ref thirdpartyOSQP library.
+For linear and quadratic programs, CGAl has the built-in
```
lrineau, line 26: `CGAl` (typo) should be replaced by `\cgal` or `CGAL`, or `%CGAL`. @sloriot What is recommended?

danston: Fixed. I first resolved the conflict in the browser editor and missed that. I am using `\cgal` now. I think it is commonly used in the docs.

danston:
- [ ] @maxGimeno Could you please install [OSQP](https://osqp.org/) solver on at least 1 Linux and 1 Windows tet suite machine. Thanks!

maxGimeno:
> * [x]  @maxGimeno Could you please install [OSQP](https://osqp.org/) solver on at least 1 Linux and 1 Windows tet suite machine. Thanks!

Done. It is now installed on Sosno and all Ubuntu docker images

danston:
Errors +. warnings on Windows: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-5/Solver_interface_Examples/TestReport_Sosno_MSVC-2022-Preview-Release.gz

maxGimeno:
tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-15.shtml

PR #5760, opened 2021-06-04
Range Loop Analysis Warnings in BGL

danston:
This little PR fixes several warnings, which show up in the `BGL` package, related to the flag `range-loop-analysis`.

## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): warnings
* Feature/Small Feature (if any): warning fix
* Documentation (not changed): [docs](https://cgal.geometryfactory.com/~danston/BGL/index.html)

### DIFF
--- a/BGL/include/CGAL/boost/graph/IO/GOCAD.h
+++ b/BGL/include/CGAL/boost/graph/IO/GOCAD.h
@@ -322,13 +322,13 @@ bool write_GOCAD(std::ostream& os,
   boost::container::flat_map<vertex_descriptor, vertices_size_type> reindex;
 
   vertices_size_type i = 0;
-  for(const vertex_descriptor v : vertices(g))
+  for(const vertex_descriptor& v : vertices(g))
   {
     os << "VRTX " << i << " " << get(vpm, v) << "\n";
     reindex[v] = i++;
   }
 
-  for(const face_descriptor f : faces(g))
+  for(const face_descriptor& f : faces(g))
   {
     halfedge_descriptor h = halfedge(f, g);
     os << "TRGL " << reindex[target(prev(h, g), g)] << " "
--- a/BGL/include/CGAL/boost/graph/IO/Generic_facegraph_printer.h
+++ b/BGL/include/CGAL/boost/graph/IO/Generic_facegraph_printer.h
@@ -143,7 +143,7 @@ public:
                           has_vertex_textures                 );
 
     vertices_size_type id = 0;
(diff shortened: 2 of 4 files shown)


danston:
@MaelRL Do you think it is better to keep like that here or remove const?

afabri:
Can you please  explain what the warning is about?

danston:
@afabri The warning is produced by the compiler flag `-range-loop-analysis` and it complains about a const item obtained from the loop by value rather than by reference. The compiler then suggests adding a reference. An alternative solution is to remove `const`.

afabri:
> An alternative solution is to remove `const`.

haha

afabri:
Can you  please put al link to the documentation. With what warning level do we get this?  I am not sure  we want to silent -Wall.

danston:
I have added the link but I did not change any docs. This warning is produced with `-Wall -Wextra`. I always use it to test my code and have it warning-free.

lrineau:
Can we see that warning in the testsuite results?

afabri:
> Can we see that warning in the testsuite results?

I think the question is: Do we have -Wall -Wextra in the testsuite, and is it really something we want?

danston:
We do not have it when compiling BGL in the test suite but similar warnings are produced in the PMP e.g.:
```
/Users/monet/Documents/test-suite/clang-default/CGAL-5.1.5-I-282/test/Polygon_mesh_processing/pmp_compute_normals_test.cpp:96:29: warning: loop variable 'f' of type 'const face_descriptor' (aka 'const CGAL::SM_Face_index') creates a copy from type 'const face_descriptor' [-Wrange-loop-analysis]
  for(const face_descriptor f : faces(mesh))
```
There is `-Wall -Wextra` set for my platform: `x86_64-apple_bigsur_clang1200-release-64bits`.

Whether we want to have them in the test suite, I would prefer having them since fewer warnings fewer possible errors to come later, I guess, but maybe I am wrong.

lrineau:
Actually, that version of Apple Clang seems buggy with that warning. @danston I suggest you do no use `-Wextra` or `-Wrange-loop-analysis`. Or use `-Wextra -Wno-range-loop-analysis`.

danston:
@lrineau So, we abandon these changes and I close this PR?

lrineau:
> @lrineau So, we abandon these changes and I close this PR?

Yes, I think so. The code seems better with the bare copies of handles. The handles are cheap to copy: just 64 bits words.

PR #5767, opened 2021-06-07
Filtered_kernel: Do_intersect(Tetrahedron_3,Bbox_3)

afabri:
## Summary of Changes

Currently the intersection test is directly done using interval arithmetic.
This PR does not really add a static filter but for the moment just tests if vertices of the tetrahedron are either all in the bbox or if some are inside and some outside to have an early result.   
The need for improvement comes from an AABB tree with tetrahedra as query objects. @cportaneri @palliez 

## Release Management

* Affected package(s):   Intersections_3

### DIFF
--- a/Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h
+++ b/Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h
@@ -43,6 +43,7 @@ class Do_intersect_3
   typedef typename K_base::Ray_3     Ray_3;
   typedef typename K_base::Segment_3 Segment_3;
   typedef typename K_base::Triangle_3 Triangle_3;
+  typedef typename K_base::Tetrahedron_3 Tetrahedron_3;
   typedef typename K_base::Sphere_3 Sphere_3;
   typedef typename K_base::Do_intersect_3 Base;
 
@@ -126,6 +127,101 @@ public:
     return Base::operator()(s,b);
   }
 
+
+
+  result_type
+  operator()(const Tetrahedron_3 &t, const Bbox_3& b) const
+  {
+    Get_approx<Point_3> get_approx;
+    bool in = false, out = false;
+    double px, py, pz;
+
+    {
+      const Point_3& p = t[0];
+      if (fit_in_double(get_approx(p).x(), px) && fit_in_double(get_approx(p).y(), py) &&
+          fit_in_double(get_approx(p).z(), pz) )
+        {
+          if( (px >= b.xmin() && px <= b.xmax()) && (py >= b.ymin() && py <= b.ymax())  && (pz >= b.zmin() && pz <= b.zmax()) ){
+            in = true;
+          }else{
+            out = true;
+          }
+        }else{
+        return Base::operator()(t,b);
+      }
+    }
+
+    {
+      const Point_3& p = t[1];
+      if (fit_in_double(get_approx(p).x(), px) && fit_in_double(get_approx(p).y(), py) &&
+          fit_in_double(get_approx(p).z(), pz) )
+        {
+          if( (px >= b.xmin() && px <= b.xmax()) && (py >= b.ymin() && py <= b.ymax())  && (pz >= b.zmin() && pz <= b.zmax()) ){
+            if(out){
+              return true;
+            }
+            in = true;
+          }else{
+            if(in){
+              return true;
+            }
+            out = true;
+          }
+        }else{
+        return Base::operator()(t,b);
+      }
+    }
+
+    {
+      const Point_3& p = t[2];
+      if (fit_in_double(get_approx(p).x(), px) && fit_in_double(get_approx(p).y(), py) &&
+          fit_in_double(get_approx(p).z(), pz) )
+        {
+          if( (px >= b.xmin() && px <= b.xmax()) && (py >= b.ymin() && py <= b.ymax())  && (pz >= b.zmin() && pz <= b.zmax()) ){
+            if(out){
+              return true;
+            }
+            in = true;
+          }else{
+            if(in){
+              return true;
+            }
+            out = true;
+          }
+        }else{
+        return Base::operator()(t,b);
+      }
+    }
+
+
+
+    {
+      const Point_3& p = t[3];
+      if (fit_in_double(get_approx(p).x(), px) && fit_in_double(get_approx(p).y(), py) &&
+          fit_in_double(get_approx(p).z(), pz) )
+        {
+          if( (px >= b.xmin() && px <= b.xmax()) && (py >= b.ymin() && py <= b.ymax())  && (pz >= b.zmin() && pz <= b.zmax()) ){
+            if(out){
+              return true;
+            }
+            in = true;
+          }else{
+            if(in){
+              return true;
+            }
+            out = true;
+          }
+        }else{
+        return Base::operator()(t,b);
+      }
+    }
+
+
+    if(in){
+      return true;
+    }
+    return Base::operator()(t,b);
+  }
 
 
   result_type


Filtered_kernel/include/CGAL/internal/Static_filters/Do_intersect_3.h
```diff
@@ -128,6 +129,68 @@ class Do_intersect_3
...
+    {
+      const Point_3& p = t[0];
+      if (fit_in_double(get_approx(p).x(), px) && fit_in_double(get_approx(p).y(), py) &&
+          fit_in_double(get_approx(p).z(), pz) )
+        {
+          if( (px >= b.xmin() && px <= b.xmax()) && (py >= b.ymin() && py <= b.ymax())  && (pz >= b.zmin() && pz <= b.zmax()) ){
+            return true;
+          }
+        }else{
+        return Base::operator()(t,b);
```
mglisse, line 147: Don't you want to check the next point before giving up? Or do you consider that if one point has ugly coordinates, it is likely that they all do?
(you could use a loop)

afabri: You are maybe right. And thanks for reading.   Before this commit the first block was slightly different. I also thought maybe a lambda.

afabri: On the other hand if for one point  the coordinates did not fit into a double  the chances are high that it is the same for the other points.

I was also wondering if it makes sense to check if a corner of the bbox is inside the tetrahedron.

afabri:
@sloriot you should maybe commit your benchmark concerning the bbox and the static filter, and report the numbers in the description of the PR.

maxGimeno:
I think this PR is responsible for all the red squares in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-27.shtml.

Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Triangle_3_do_intersect.h
```diff
@@ -179,50 +188,52 @@ namespace internal {
...
       // t_max >= t_min
-      Uncertain<bool> b =  do_axis_intersect_aux<K,AXE,SIDE>(k->x()-j->x(), k->z()-j->z(), sides) >= 0;
+      Uncertain<bool> b =  do_axis_intersect_aux_A1<FT,SIDE>(k[0]-j[0], k[2]-j[2], sides, do_axis_intersect_aux_impl) != NEGATIVE;
       if (is_indeterminate(b))
         return b;
       if(b) std::swap(j,k);
-      return  CGAL_AND((do_axis_intersect_aux<K,AXE,SIDE>(p_min.x()-j->x(), p_min.z()-j->z(), sides) <= 0),
-                       (do_axis_intersect_aux<K,AXE,SIDE>(p_max.x()-k->x(), p_max.z()-k->z(), sides) >= 0) );
+      return  CGAL_AND((do_axis_intersect_aux_A1<FT,SIDE>(p_min[0]-j[0], p_min[2]-j[2], sides, do_axis_intersect_aux_impl) != POSITIVE),
+                       (do_axis_intersect_aux_A1<FT,SIDE>(p_max[0]-k[0], p_max[2]-k[2], sides, do_axis_intersect_aux_impl) != NEGATIVE) );
```
afabri, line 227: Don't we have to write
````
b = Uncertain<bool>::indeterminate();
      try {
        b = CGAL_AND((do_axis_intersect_aux_A1<FT,SIDE>(p_min[0]-j[0], p_min[2]-j[2], sides, do_axis_intersect_aux_impl) != POSITIVE),
                       (do_axis_intersect_aux_A1<FT,SIDE>(p_max[0]-k[0], p_max[2]-k[2], sides, do_axis_intersect_aux_impl) != NEGATIVE) );
      }catch(const Uncertain_conversion_exception& ){
      }
      return b;
```

mglisse: Why?
(by the way, now that we require C++14, I think it would be a good time to rewrite CGAL_AND in Uncertain.h without statement expressions so visual studio can benefit as well)

sloriot: @mglisse started in #5904

sloriot: For the record, Andreas thought that a try/catch was missing because some parts of the predicate unprotect in the static version of the predicate and that was an issue on windows because `CGAL_AND` was converting the `Uncertain<bool>` to `bool`.

sloriot:
IMO we should undo 7d2321c

afabri:
> 
> 
> IMO we should undo [7d2321c](https://github.com/CGAL/cgal/commit/7d2321cecd791812de3142e70da2165959bf9b35)

I agree. Sorry for not having been patient.

Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Triangle_3_do_intersect.h
```diff
@@ -411,7 +413,8 @@ bool do_intersect_bbox_or_iso_cuboid(const typename K::Triangle_3& a_triangle,
                                        const FT& c_alpha,
                                        const FT& c_beta) -> Uncertain<Sign>
   {
-    return CGAL::sign(- c_alpha * alpha + c_beta * beta);
+    Sgn<FT> sign;
+    return sign(- c_alpha * alpha + c_beta * beta);
```
lrineau, line 417: For my curiosity, why does that change anything?

`CGAL::sign` is implemented that way:
https://github.com/CGAL/cgal/blob/8c9fc331cd0e523f897f830da480483b6fb70d5d/Algebraic_foundations/include/CGAL/number_utils.h#L240-L247
and `CGAL::Sgn` is implemented that way:
https://github.com/CGAL/cgal/blob/8c9fc331cd0e523f897f830da480483b6fb70d5d/Algebraic_foundations/include/CGAL/number_utils_classes.h#L58-L59

so in the end they both use `Real_embeddable_traits<NT>::Sgn`...

lrineau: Side note: we could reimplement `CGAL::Sgn` (and all the others) as template aliases, to avoid an extra template instantiation.

afabri: If I remember correctly for the testsuite with Leda numbers when not qualified it uses an overload coming from Leda which has a different return type.

mglisse: Might as well use `compare` instead of `sign(x-y)`.

sloriot: Bad grep

lrineau: > If I remember correctly for the testsuite with Leda numbers when not qualified it uses an overload coming from Leda which has a different return type.

Here the commit https://github.com/CGAL/cgal/pull/5767/commits/b67589e16b9db85d0a66bbeb6e2c4024e67a03bf replaced `CGAL::sign` with the use of `CGAL::Sgn`.

sloriot:

lrineau:
What is blocking this PR? I see that it is https://github.com/CGAL/cgal/labels/not%20yet%20approved.

Can you sum up what are the current points under discussion?

Filtered_kernel/include/CGAL/Filtered_kernel/internal/Do_intersect_3.h
```diff
@@ -128,6 +129,68 @@ class Do_intersect_3
...
 
 
+  result_type
+  operator()(const Tetrahedron_3 &t, const Bbox_3& b) const
+  {
+    Get_approx<Point_3> get_approx;
+    double px, py, pz;
+
+    {
+      const Point_3& p = t[0];
```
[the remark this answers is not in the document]
afabri: i<4??

Filtered_kernel/include/CGAL/Filtered_kernel/internal/Do_intersect_3.h
```diff
@@ -128,6 +129,68 @@ class Do_intersect_3
 
 
 
+  result_type
+  operator()(const Tetrahedron_3 &t, const Bbox_3& b) const
```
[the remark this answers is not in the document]
afabri: done.

PR #5770, opened 2021-06-08
Remove CGAL_NEF3_FACET_WITH_BOX

GilesBathgate:
## Summary of Changes

Compiling CGAL with the define CGAL_NEF3_FACET_WITH_BOX set doesn't work. There are a number of compile errors and the implementation is missing here:

https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h#L550-L560

As such I would assume that the code isn't tested or used publically, so I propose therefore that the code should be removed, and removing it correctly can be achieved using a tool called unifdef using the following: 

`grep CGAL_NEF3_FACET_WITH_BOX * -Rl | xargs unifdef -UCGAL_NEF3_FACET_WITH_BOX -m`

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/Binary_operation.h
+++ b/Nef_3/include/CGAL/Nef_3/Binary_operation.h
@@ -313,12 +313,6 @@ class Binary_operation : public CGAL::SNC_decorator<Map> {
     //    CGAL_NEF_SETDTHREAD(19*43*131);
     CGAL_NEF_TRACEN("=> binary operation");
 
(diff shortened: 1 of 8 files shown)


afabri (approved):
I agree with the cleanup as it was an experiment to use boxes for acceleration.

maxGimeno:

PR #5740, opened 2021-05-31
Remove CGAL_NEF3_TRIANGULATE_FACETS

GilesBathgate:
## Summary of Changes

Compiling CGAL with the define `CGAL_NEF3_TRIANGULATE_FACETS` set (which also requires CGAL_NEF_EXPLOIT_REFERENCE_COUNTING otherwise there is a compile error) doesn't work. There is a compile error here: https://github.com/CGAL/cgal/blob/68c44a162df2c5577a7f5f51bf45ac1d349ed143/Packages/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h#L250

Where `CT::Edge` should be preceded by a typename keyword. As such I would assume that the code isn't tested or used publically. I tested the code against the performance benchmark (spheregrid/shiftedspheregrid) and didn't find any performance benefit. I propose therefore that the code should be removed, and removing it correctly can be achieved using a tool called `unifdef` using the following: `grep CGAL_NEF3_TRIANGULATE_FACETS * -Rl | xargs unifdef -UCGAL_NEF3_TRIANGULATE_FACETS -m`

I have applied the command to the code for this PR. 

## Release Management

* Affected package(s): Nef_3
* Issue(s) solved (if any): cleaning
* License and copyright ownership: Returned to CGAL Authors.

PR #5772, opened 2021-06-08
18% performance boost in Nef_3 K3_tree.

GilesBathgate:
## Summary of Changes

Looking carefully at the code in K3_tree and running it through a profiler it can be seen that the code firstly creates an `Object_list` of vertices, edges, and facets. It then spends substantial amount of time unpicking the object list, and casting each object back into vertices/edges/facets to build the K3_tree. While the casts themselves are not very expensive, applying the `make_object` and `CGAL::assign` lavishly so many items does add up to significant overhead. By simply passing separate lists of Vertices, Edges, and Facets much of this casting is avoided and the performance can be boosted. Using the Nef_3_benchmark with spheregrid/shiftedspheregrid a performance gain of 10% can be observed. The Object lists are created at the point a K3_tree `Node` is created, so as to be backwards compatible with the existing code.

Another improvement is to select a splitting plane at the midpoint of the median and its predecessor, this ensures fewer points fall on the oriented boundary and gives a further performance boost of 8%

A few other minor tweaks and cleaning has been applied to the code, for which further details can be discussed if requested.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -150,7 +150,7 @@ public:
       return;
     }
 
-    Side_of_plane sop(splitting_plane.point(), depth);
+    Side_of_plane sop(splitting_plane.point(), depth%3);
     Oriented_side side = sop(f);
     if( side == ON_NEGATIVE_SIDE || side == ON_ORIENTED_BOUNDARY)
       left_node->add_facet(f, depth+1);
@@ -164,7 +164,7 @@ public:
       return;
     }
 
-    Side_of_plane sop(splitting_plane.point(), depth);
+    Side_of_plane sop(splitting_plane.point(), depth%3);
     Oriented_side side = sop(e);
     if( side == ON_NEGATIVE_SIDE || side == ON_ORIENTED_BOUNDARY)
       left_node->add_edge(e, depth+1);
@@ -178,7 +178,7 @@ public:
       return;
     }
 
-    Side_of_plane sop(splitting_plane.point(), depth);
+    Side_of_plane sop(splitting_plane.point(), depth%3);
     Oriented_side side = sop(v);
     if( side == ON_NEGATIVE_SIDE || side == ON_ORIENTED_BOUNDARY)
       left_node->add_vertex(v, depth+1);
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -464,6 +464,7 @@ public:
     reference_counted = (&(p1.hx()) == &(p2.hx()));
     CGAL_NEF_TRACEN("reference counted " << reference_counted);
 #endif
+    non_efective_splits=0;
     root = build_kdtree(vertices, edges, facets, 0);
   }
   const Object_list& objects_around_point( const Point_3& p) const {
@@ -610,8 +611,10 @@ std::string dump_object_list( const Object_list& O, int level = 0) {
 
 private:
 
+int non_efective_splits;
+
 Node_handle build_kdtree(Vertex_list& V, Halfedge_list& E, Halffacet_list& F,
-                         int depth, int non_efective_splits=0) {
+                         int depth) {
   CGAL_precondition( depth >= 0);
 
   if( !can_set_be_divided(depth, V.size())) {
@@ -661,8 +664,8 @@ Node_handle build_kdtree(Vertex_list& V, Halfedge_list& E, Halffacet_list& F,
     return &(nodes.back());
   }
 
-  Node_handle left_node = build_kdtree(V1, E1, F1, depth + 1, non_efective_splits);
-  Node_handle right_node = build_kdtree(V2, E2, F2, depth + 1, non_efective_splits);
+  Node_handle left_node = build_kdtree(V1, E1, F1, depth + 1);
+  Node_handle right_node = build_kdtree(V2, E2, F2, depth + 1);
   nodes.push_back(Node(left_node, right_node, construct_splitting_plane(point_on_plane, coord)));
   return &(nodes.back());
 }
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -632,8 +632,7 @@ Node_handle build_kdtree(Vertex_list& V, Halfedge_list& E, Halffacet_list& F,
   }
 
   int coord = depth%3;
-  Vertex_iterator median;
-  Point_3 point_on_plane = find_median_point(V, median, coord);
+  Point_3 point_on_plane = find_median_point(V, coord);
   CGAL_NEF_TRACEN("build_kdtree: plane: "<<partition_plane<< " " << point_on_plane);
 
 #ifdef CGAL_NEF_EXPLOIT_REFERENCE_COUNTING
@@ -643,8 +642,6 @@ Node_handle build_kdtree(Vertex_list& V, Halfedge_list& E, Halffacet_list& F,
 #endif
 
   Vertex_list V1,V2;
-  V1.reserve(std::distance(V.begin(), median));
-  V2.reserve(std::distance(median, V.end()));
   classify_objects(V, sop, V1, V2);
 
   Halfedge_list E1,E2;
@@ -711,14 +708,17 @@ static bool classify_objects(const List& l,Side_of_plane& sop,
   return (on_oriented_boundary != l.size());
 }
 
-static Point_3 find_median_point(Vertex_list& V,
-                          Vertex_iterator& median, int coord) {
+static Point_3 find_median_point(Vertex_list& V, int coord) {
   CGAL_assertion(V.size() > 1);
 
-  median = V.begin() + V.size()/2;
-  std::nth_element(V.begin(), median, V.end(), Smaller(coord));
+  Smaller smaller(coord);
+  Vertex_iterator begin = V.begin();
+  Vertex_iterator median = begin + V.size()/2;
+  std::nth_element(begin, median, V.end(), smaller);
+  Vertex_iterator prev = std::prev(median);
+  std::nth_element(begin, prev, median, smaller);
 
-  return (*median)->point();
+  return CGAL::midpoint((*median)->point(), (*prev)->point());
 }
 
 static Plane_3 construct_splitting_plane(const Point_3& pt, int coord)
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -716,9 +716,9 @@ static Point_3 find_median_point(Vertex_list& V, int coord) {
 static Plane_3 construct_splitting_plane(const Point_3& pt, int coord)
 {
   switch(coord) {
-  case 0: return Plane_3(1, 0, 0, -pt.x());
-  case 1: return Plane_3(0, 1, 0, -pt.y());
-  case 2: return Plane_3(0, 0, 1, -pt.z());
+  case 0: return Plane_3(1/pt.hw(), 0, 0, -pt.hx());
+  case 1: return Plane_3(0, 1/pt.hw(), 0, -pt.hy());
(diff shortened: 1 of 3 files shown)


Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
```diff
@@ -100,23 +101,13 @@ class ComparePoints<CGAL::Lazy_kernel<EK>, Coordinate> {
...
-public:
-  typedef typename SNC_decorator::SNC_structure SNC_structure;
+
   typedef typename SNC_decorator::Decorator_traits Decorator_traits;
 
   typedef typename Decorator_traits::Vertex_handle Vertex_handle;
   typedef typename Decorator_traits::Halfedge_handle Halfedge_handle;
   typedef typename Decorator_traits::Halffacet_handle Halffacet_handle;
 
-#ifdef CGAL_NEF3_TRIANGULATE_FACETS
```
afabri, line 111: Hi Giles, I guess you are not really happy that we are lagging behind with approving and merging your tsunami of pull requests.  We really appreciate them and we allocate resources where we find time. 
The problem with this PR is that it seems to combine several others in the diff. That makes it pretty hard to review.  I suggest that we deal with the PRs in order and when an older one  (e.g. the one on "triangulate faces") gets merged into mater this block should be no longer in the diff.

Nef_3/include/CGAL/Nef_3/K3_tree.h
```diff
@@ -381,45 +365,42 @@ typedef Smaller_than<
     return s;
   }
 
-  template<typename Depth>
```
afabri, line 384: I am puzzled to see `Depth`  as template parameter.  Wouldn't that need a constant known at compile time. @lrineau @mglisse

GilesBathgate: Its always an int, its specialised by a parameter `build_k3tree(O, O.size()+v_end, 0)`

Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
GilesBathgate: Yeah, I was finding it tricky to manage. If you are happy to hold off fully reviewing this until #5740 is merged, I can rebase and resolve conflicts.

Nef_3/include/CGAL/Nef_3/K3_tree.h
afabri: But isn't `O.size()+v_end`  only known at runtime?

GilesBathgate: No, it's the final parameter, i.e the `0` which is a literal int.

Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
```diff
@@ -56,15 +57,15 @@ class ComparePoints {
 
 
 template <typename Coordinate, typename EK>
-class ComparePoints<CGAL::Lazy_kernel<EK>, Coordinate> {
+class Compare_points<CGAL::Lazy_kernel<EK>, Coordinate> {
```
GilesBathgate, line 60: @afabri I think this template specialisation (and also for Smaller_than) should be removed because:

a) It doesn't seem to match against Epeck kernels.
b) I think CGAL::compare_x already does an approx comparison when possible.

Nef_3/include/CGAL/Nef_3/K3_tree.h
afabri: I misunderstood. I thought we have to do with for example `Depth-1`  being used in a recursion, but here it is just an integral type in the parameter list. So I agree with this change.

Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
afabri: I agree. I am even wondering if we should not add the 3D equivalent to  [`Compare_x_2`](https://github.com/CGAL/cgal/blob/master/Filtered_kernel/include/CGAL/internal/Static_filters/Compare_x_2.h).

afabri:
Is this still WIP or shall I review it?

GilesBathgate:
> Is this still WIP or shall I review it?

The conflicts regarding CGAL_NEF3_TRIANGULATE_FACETS are resolved so it should be easier to see the real changes now. 

The changes are all regarding `build_kdtree`, and I don't think I have any further improvements to this area, so it could be reviewed. Thanks.

Nef_3/include/CGAL/Nef_3/SNC_k3_tree_traits.h
afabri: I just convinced myself that there is no need for the "3D equivalent"   as there is a typedef in Static_filters.h

lrineau: I was wondering why a `typedef` as sufficient for `Compare_x_3` but a dedicated functor was required for `Compare_x_2`. The difference is:
  - the concept [`CompareX_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1CompareX__3.html) specifies only one overload, that only compares coordinates,
  - whereas the concept [`CompareX_2`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1CompareX__2.html) has overloads that require comparison of compute values.

GilesBathgate:
@afabri I've increased the scope of the benchmark to include Difference/Intersection/Union. The full results are now generated using the github actions CI. The latest results are here: https://github.com/afabri/Nef_3_benchmark/suites/3621074286/artifacts/87561967

In summary, I get 17% to 18% performance improvement. 

Perhaps the PR is only improving performance for this dataset (spheregrid/shiftedspheregrid), but I would find that very odd.

afabri:
I just edited my comment, where I wrote that master is sometimes faster.   This is only the case for the "(A - B)  union with intersection(A,B)".   This is  a rather artificial testcase where the operations end up to be the input object A,  But  that it is a little bit slower may also come from the fact that these are cascaded operations, that is the output of operations is the input of some more operations.  To combine both scenarios in a single test, was a bad idea.

afabri:
I wrote a just iterative testcase and it shows performance improvements:
```
In master :
$ ./build/Release/test_iterative.exe
47104
a1 : 47104  92160 21.908 sec.
intersection(a1,a2) : 86585  98402 12.817 sec.
a3 - a4 : 103600  120394 14.138 sec.
a5 - a6 : 96910  116952 13.202 sec.
union(a12, a34) : 203892  190133 25.606 sec.
a1234 - a56 : 274639  228382 38.42 sec.

$ ./build/Release/test_iterative.exe
47104
a1 : 47104  92160 17.897 sec.
intersection(a1,a2) : 86585  98402 10.294 sec.
a3 - a4 : 103600  120394 11.082 sec.
a5 - a6 : 96910  116952 10.653 sec.
union(a12, a34) : 203892  190133 20.267 sec.
a1234 - a56 : 274639  228382 31.08 sec.
```
As input I used several copies of the spheregrid, each rotated and shifted.  
image

afabri:
Maybe it is worth to still investigate the case where the branch is slower. What is particular about it is that two inputs share many vertices and faces as they got constructed by a previous Boolean operation.

afabri (approved):

afabri:
As far I can see it we can remove the label "Tests failing"

sloriot:
Still fully red:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-50/Nef_3/TestReport_Sosno_MSVC-2019-Community-Release.gz

GilesBathgate:
@sloriot Sorry, I assumed the patch by afabri would fix this so didn't test locally. I now fixed in 217ee10 and have the test result:

```
successful configuration
   successful compilation of nary
   successful execution   of nary
   successful compilation of nef_3_from_points
   successful execution   of nef_3_from_points
   successful compilation of Nef_3_problematic_construction
   successful execution   of Nef_3_problematic_construction
   successful compilation of nef_union_error_llvm
   successful execution   of nef_union_error_llvm
   successful compilation of test_nef_3_io_Cartesian
   successful execution   of test_nef_3_io_Cartesian
   successful compilation of test_nef_3_io_Cartesian_Lazy
   successful execution   of test_nef_3_io_Cartesian_Lazy
   successful compilation of test_nef_3_io_EPEC
   successful execution   of test_nef_3_io_EPEC
   successful compilation of test_nef_3_io_EPIC
   successful execution   of test_nef_3_io_EPIC
   successful compilation of test_nef_3_io_Homogeneous
   successful execution   of test_nef_3_io_Homogeneous
   successful compilation of test_nef_3_io_Homogenoeus_Lazy
   successful execution   of test_nef_3_io_Homogenoeus_Lazy
   successful compilation of test_nef_to_soup
   successful execution   of test_nef_to_soup
   successful compilation of test_with_extended_homogeneous
   successful execution   of test_with_extended_homogeneous
   successful compilation of test_with_homogeneous
   successful execution   of test_with_homogeneous
   successful compilation of test_with_lazy
   successful execution   of test_with_lazy
```

sloriot:
Full red line in [Minkowski_sum_3](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-52/Minkowski_sum_3/TestReport_lrineau_Debian-Stable.gz) and [Convex_decomposition](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-52/Convex_decomposition_3/TestReport_gimeno_ArchLinux-clang-Release.gz)

GilesBathgate:
@sloriot Sorry again. It just didn't occur to run the full test suite, locally. Reproduced and fixed in 1918ed3 

P.S the test suite has great coverage, and the assertion immediately pointed me to the problem. Really good.

GilesBathgate:
>And does the fix change anything for the performance tests?

@afabri There is no performance degradation as can be seen in these latest results: https://github.com/afabri/Nef_3_benchmark/suites/3786753639/artifacts/93351812

Nef_3/include/CGAL/Nef_3/K3_tree.h
```diff
@@ -664,163 +583,153 @@ std::string dump_object_list( const Object_list& O, int level = 0) {
...
+  return &(nodes.back());
 }
 
-template <typename Depth>
-bool can_set_be_divided(Object_iterator start, Object_iterator end, Depth depth) {
+bool can_set_be_divided(int depth, typename Vertex_list::size_type size) {
   if(depth >= max_depth)
     return false;
-  if(std::distance(start,end)<2)
+  if(size <= 2)
```
GilesBathgate, line 650: I changed the comparison here from `<2` to `<=2` based on the assumption that 1 vertex per tree cell is overkill. However maybe the code elsewhere assumes there can only be one candidate per cell when for example doing the point location?

sloriot:
Runtime errors for several platforms for `Convex_decomposition_3_Examples` in [CGAL-5.4-Ic-108](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.4-Ic-108.shtml).

Nef_3/include/CGAL/Nef_3/K3_tree.h
afabri: Is this change the culprit for the runtime errors signaled by @sloriot?    I had approved the PR, and it would make sense to not continue working on a PR  by adding further improvements, but to limit it to fixing bugs in the testsuite.

GilesBathgate: @afabri  The change was always in the PR that you approved, I just added the comment about it later in hindsight. That does raise the question of why this is showing up in the testsuite now where it previously passed. Anyway I am not intending to make any further improvements, like you say, limit this to fixing bugs, and make a new PR for anything else. 

When I have some time this weekend I'll investigate, I think I just need to run the testsuite locally, again, but like I say this didn't show up when I ran the testsuite before.

GilesBathgate:
@afabri does 7b8c3ab fix the failing test. If so that's great thanks.

afabri:
It should fix. We were only by minutes to late to put it in the testsuite. Will happen tonight.

sloriot:

PR #5773, opened 2021-06-09
Fix path to tmp.log

maxGimeno:
## Summary of Changes
There is a wrong path in the script that builds the doc, in case there is an error during the building process.
## Release Management

* Affected package(s):CI


maxGimeno:

PR #4213, opened 2019-09-09
Polyline_constraint_hierarchy_2.h: remove the useless point_ data member

lrineau:
## Summary of Changes

Small cleanup PR : remove the useless `point_` data member from `Polyline_constraint_hierarchy_2.h`. A *const* reference to the point can be retrieved as `vertex_->point()`.

The `id` public data seems also useless, but it is actually used in `Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h`.

## Release Management

* Affected package(s): Triangulation_2, Polyline_simplification_2
* License and copyright ownership: maintenance by GeometryFactory

PR #5774, opened 2021-06-09
Polyline_simplification: Bugfix in Points_in_constraint_iterator

afabri:
## Summary of Changes

When simplifying a polyline constraint in a triangulation the Points_in_constraint_iterator permits to traverse the input sequence of points. This is broken.  This draft only adds debug code and not a fix.

This PR undoes the PR #4213. The vertex handle in the Node class survives a simplification as it it is in a `Compact_container` but if after the simplification the user adds a new constraint that will recycle the vertex object and put another point in there.

The real bug though was spotted by Mael in the function that removes the points in the skip list for which the vertex was removed.

## Release Management

* Affected package(s): Polyline Simplification
* Issue(s) solved (if any):  [stackoverflow](https://stackoverflow.com/questions/67855845/cgal-polyline-simplification-results-in-self-intersection)
* License and copyright ownership: maintenance

### DIFF
--- a/Polyline_simplification_2/test/Polyline_simplification_2/CMakeLists.txt
+++ b/Polyline_simplification_2/test/Polyline_simplification_2/CMakeLists.txt
@@ -36,8 +36,7 @@ endif()
 # Creating entries for all .cpp/.C files with "main" routine
 # ##########################################################
 
+create_single_source_cgal_program( "issue-5774.cpp" )
 create_single_source_cgal_program( "simplify_polygon_test.cpp" )
 
 create_single_source_cgal_program( "simplify_polyline_with_duplicate_points.cpp" )
-
-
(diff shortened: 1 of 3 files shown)


GilesBathgate:
If fixed can the description be updated?

lrineau:
That is a bug-fix. It should be backported to `5.1.x-branch`.

maxGimeno:
errors : https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-148/Polyline_simplification_2/TestReport_Sosno_MSVC-2019-Community-Release.gz

The keyword Polyline is already used in a Windows file called `wingdi.h`, see [here](https://docs.microsoft.com/fr-fr/windows/win32/api/wingdi/nf-wingdi-polyline).

MaelRL:
And the test file should either be removed or actually test something.

lrineau:
> And the test file should either be removed or actually test something.

I agree the test should assert something on the result.

afabri:
> That is a bug-fix. It should be backported to `5.1.x-branch`.

done

maxGimeno:

PR #5775, opened 2021-06-09
fast dynamic_cast in Lazy_kernel::Construct_point_3

### LINKED ISSUE #5724: slow dynamic_cast in Lazy_kernel::Construct_point_3
With libstdc++ (and possibly even worse with visual studio), test_p3rt3_as_p3t3.cpp spends a lot of time in the dynamic_cast of Lazy_kernel used to recognize if a weighted point was constructed from a point and a weight, and thus if we can extract those without building a new lazy object. This was noticed in #5402.

Rewriting the test as
```c++
      auto&& t = typeid(*p.ptr());
      if(t == typeid(LR)){
        LR * lr = static_cast<LR*>(p.ptr());
        return std::get<1>(lr->l);
      }else{
        if(t == typeid(LRint)){
          LRint* lrint = static_cast<LRint*>(p.ptr());
          return std::get<1>(lrint->l);
        }
      }
```
without the dynamic_cast seems to speed things up nicely here (we could add `final` on Lazy_rep_n to confirm that this is equivalent). (it is possible that I got the test wrong, this needs checking)
Further replacing `t == typeid(X)` with `t.name() == typeid(X).name()` gives another nice gain on top of that (I think that's how libc++ implements it, but libstdc++ further checks `strcmp` if that comparison is false). It is supposedly unsafe when parts of the program are loaded with `dlopen` or similar, but since this is just an optimization and wrongly saying false still uses valid code, maybe that would be acceptable.
I don't know if getting rid of LRint (casting the `int` to FT and creating LR instead) would help or hurt more.
Another option may be adding a virtual function in Lazy_rep to detect these cases.

afabri:
## Summary of Changes

The Issue #5724 explains it all and the code was just copy pasted from the issue to the code. Thank you Marc.
I also reproduced the performance gain for the 3d periodic triangulation with a regular triangulation with constant weight.
I did not try to get rid of `LRint`.

## Release Management

* Issue(s) solved (if any): fix #5724

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy_kernel.h
+++ b/Filtered_kernel/include/CGAL/Lazy_kernel.h
@@ -413,12 +413,13 @@ public:
                          int
                          > LRint;
 
-      LR * lr = dynamic_cast<LR*>(p.ptr());
-      if(lr && (! lr->et)){
+      auto&& t = typeid(*p.ptr());
+      if(t.name() == typeid(LR).name()){
+        LR * lr = static_cast<LR*>(p.ptr());
         return std::get<1>(lr->l);
-      } else {
-        LRint* lrint = dynamic_cast<LRint*>(p.ptr());
-        if(lrint && (! lrint->et)){
+      }else{
+        if(t.name() == typeid(LRint).name()){
+          LRint* lrint = static_cast<LRint*>(p.ptr());
           return std::get<1>(lrint->l);
         }
       }
@@ -465,16 +466,17 @@ public:
                          int
                          > LRint;
 
-
-      LR * lr = dynamic_cast<LR*>(p.ptr());
-      if(lr && (! lr->et)){
+      auto&& t = typeid(*p.ptr());
+      if(t.name() == typeid(LR).name()){
+        LR * lr = static_cast<LR*>(p.ptr());
         return std::get<1>(lr->l);
       }else{
-        LRint* lrint = dynamic_cast<LRint*>(p.ptr());
-        if(lrint && (! lrint->et)){
+        if(t.name() == typeid(LRint).name()){
+          LRint* lrint = static_cast<LRint*>(p.ptr());
           return std::get<1>(lrint->l);
         }
       }
+
       return BaseClass().construct_point_3_object()(p);
     }
 
--- a/Filtered_kernel/include/CGAL/Lazy_kernel.h
+++ b/Filtered_kernel/include/CGAL/Lazy_kernel.h
@@ -346,15 +346,16 @@ public:
                          int
                          > LRint;
 
+      auto& obj = *p.ptr();
+      const char* tn = typeid(obj).name();
 
-      auto&& t = typeid(*p.ptr());
-      if(t.name() == typeid(LR).name()){
+      if(tn == typeid(LR).name()){
         LR * lr = static_cast<LR*>(p.ptr());
         if(lr->is_lazy()){
           return std::get<2>(lr->l);
         }
       }else{
-        if(t.name() == typeid(LRint).name()){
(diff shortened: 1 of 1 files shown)


mglisse:
In https://github.com/mglisse/cgal/commits/dynamic_cast-typeid I actually had one further commit https://github.com/mglisse/cgal/commit/edb574de75adbb13cff04a1a280043722947ed4c . If we restrict to the inexact comparison, that would mostly mean storing t.name() as a `const char*` instead of storing t.
In my patch, I was also changing the dynamic_cast in Compute_weight, it seems strange to keep it here.

afabri:
Hi Marc, I was not aware of that you committed the code. Do you want to turn it into a PR once you are done and I cancel this PR, or do you prefer that I take care of it and finalize it under your guidance?

mglisse:
It depends how soon you want it. My patch is on the branch for thread-safe Epeck (since I was already touching that code and wanted to avoid a conflict). Well, I pushed it temporarily under a different name since I wasn't sure which of the 2 alternatives to use, but it still contains all the atomicity stuff. If it can wait until that branch is integrated, then it may be easiest to drop this PR. If you want it earlier, then you can continue.

afabri:
[ERROR:      execution___of__test_regular_traits_3 ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-146/Triangulation_3/TestReport_Sosno_MSVC-2019-Community-Release.gz)

maxGimeno:
Errors:
 - https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-146.shtml#Interpolation
 - https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-146.shtml#Periodic_3_triangulation_3
 - https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-146.shtml#Skin_surface_3
 - https://cgal.geometryfactory.com/CGAL/testsuite/results-5.3-Ic-146.shtml#Triangulation_3

mglisse:
> [ERROR: execution___of__test_regular_traits_3 ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.3-Ic-146/Triangulation_3/TestReport_Sosno_MSVC-2019-Community-Release.gz)

Interestingly, if I test locally, it does indeed fail on your branch, but passes on mine...

mglisse:
Ah, yes, you seem to have dropped the is_lazy test.

mglisse:
I wondered about the Compute_weight LRint case. I hadn't added it because it was less clear if that was a win.
Without the patch, Compute_weight creates a new Lazy that references the WP and copies the interval, while with the patch it creates a new Lazy that contains the int and converts the int to an interval.
Without the patch, update_exact calls exact on the WP and copies the rational weight, while with the patch, it constructs a rational from the int.
The gain is less clear than with the LR case, which avoids constructing a new Lazy and can just increment the ref count. But now that we are using non-refcounted rationals, and with the atomic counters for Handle on their way, it may be that you do have a gain.

Unrelated, I'd still rather store `const char*tn=typeid(*p.ptr()).name();` and test `if(tn==typeid(LR).name())`, but that's a detail. Actually, it is even better to do it in 2 lines:
```
auto& obj = *p.ptr();
const char* tn = typeid(obj).name();
```
because that avoids a null pointer check (`typeid(*ptr)` is handled magically in the standard, not just as `*` followed by `typeid`).

afabri:
@mglisse I only did the "Unrelated.."   and kept the `Compute_weight` as I did not see a clear conclusion.

mglisse (approved):

maxGimeno:

PR #5778, opened 2021-06-14
License Generator One-Word Package Name Fix

danston:
This little PR proposes a fix to the `generate_files.cmake` script in the `Installation` package. The problem occurs when the package name consists of one word like `Weights` or `Classification` e.g. If this word is repeated in the second part of the package description in the file `gpl_package_list.txt`, then the internal function `REPLACE` substitutes the package name with the empty space that leads to the empty variable `PACKAGE_PRETTY_NAME` that is used in the file `gpl.h.in`. 

The proposed fix checks if the `PACKAGE_PRETTY_NAME` is empty and in case it is uses the content from the variable `PACKAGE_NAME`.

It fixes the case:
Package Package

What about the case
Package Package in 2D and 3D ?

In this case, it will generate:
`PACKAGE_NAME` = "Package"
`PACKAGE_PRETTY_NAME` = "in 2D and 3D"

## Release Management

* Affected package(s): `Installation`
* Issue(s) solved (if any): generating license


Installation/include/CGAL/license/generate_files.cmake
```diff
@@ -5,6 +5,10 @@ foreach(LINE ${package_list})
   string(REPLACE "/" "_" PACKAGE_CODE_NAME ${PACKAGE_NAME})
   string(TOUPPER ${PACKAGE_CODE_NAME} PACKAGE_CODE_NAME)
   string(REPLACE "${PACKAGE_NAME}" "" PACKAGE_PRETTY_NAME ${LINE})
+  string(COMPARE EQUAL ${PACKAGE_PRETTY_NAME} " " IS_EMPTY)
+  if(${IS_EMPTY})
+    set(PACKAGE_PRETTY_NAME ${PACKAGE_NAME})
+  endif()
```
lrineau, line 11: As an empty variable evaluates to false in conditional, I suggest this more idiomatic construction:
```suggestion
  if(NOT PACKAGE_PRETTY_NAME)
    set(PACKAGE_PRETTY_NAME ${PACKAGE_NAME})
  endif()
```
Please test if that works as well as the original patch.

Installation/include/CGAL/license/generate_files.cmake
```diff
@@ -6,6 +6,9 @@ foreach(LINE ${package_list})
   string(TOUPPER ${PACKAGE_CODE_NAME} PACKAGE_CODE_NAME)
   string(REPLACE "${PACKAGE_NAME}" "" PACKAGE_PRETTY_NAME ${LINE})
   string(STRIP "${PACKAGE_PRETTY_NAME}" PACKAGE_PRETTY_NAME)
+  if(NOT PACKAGE_PRETTY_NAME)
+    set(PACKAGE_PRETTY_NAME ${PACKAGE_NAME})
+  endif()
```
lrineau, line 11: @danston Have you check if that code has indeed the same behavior of the previous version?

Installation/include/CGAL/license/generate_files.cmake
danston: I slightly modified your suggestion because `PRETTY_PACKAGE_NAME` evaluates to "one whitespace", which does not evaluate to false. However, after stripping all trailing whitespaces it does work. I have also tested it locally and it seems to work well (the original patch I also tested and it is also valid).

lrineau:
I have merge this PR without testing, because that script is outside of any of our testing paths.

Installation/include/CGAL/license/generate_files.cmake
danston: Yes, it did not update any other license except for the following 4 packages: `Classification`, `Orthree`, `Three`, and `Triangulation_on_sphere_2`. The `Orthree` and `Triangulation_on_sphere_2` are modified even when running the original master version. The `Three` and `Classification`, the empty space in the license header is now substituted by the correct name.

danston: The question is: should I add those license modifications to this PR as well?

lrineau: > The question is: should I add those license modifications to this PR as well?

That is too late: I have already merged it. :smiley:

danston:
> I have merge this PR without testing, because that script is outside of any of our testing paths.

Should I create a PR with the license modifications after rerunning this script?

lrineau:
> > I have merge this PR without testing, because that script is outside of any of our testing paths.
> 
> Should I create a PR with the license modifications after rerunning this script?

If you do not mind, that would be a good addition. And that one will have to pass our testsuite, because it will modify a few headers.

PR #5781, opened 2021-06-17
PMP: Use deque instead of vector

afabri:
## Summary of Changes

As we have no idea what to `reserve()`  for a `vector` in which we collect duplicates, we better switch to a `deque`. This probably only plays a role in case of many duplicates.

## Release Management

* Affected package(s): PMP


sloriot:
Successfully tested in CGAL-5.4-Ic-28

PR #5790, opened 2021-06-21
PMP::Polyhedral_envelope:  Fix memory leak

afabri:
## Summary of Changes

In case `update_exact()` is called the rep stores a non null pointer in `et` which must be deleted before a re-assignment.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -904,6 +904,9 @@ struct Lazy_construction_optional_for_polygonal_envelope
       CGAL_STATIC_THREAD_LOCAL_VARIABLE_0(LazyPointRep, rep);
 
       const typename AK::Point_3 ap = *oap;
+      if(rep.et != nullptr){
+        delete rep.et;
+      }
       rep = LazyPointRep(2,ap, ec, l1, l2, l3);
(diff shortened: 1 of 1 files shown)


maxGimeno (approved):
I approve these changes, and so does my sanitizer :+1:

lrineau:
The *real* bug is that `CGAL::Lazy_rep` does not follow the rule-of-five: it has:
  - a copy-constructor (deleted)
  - a destructor,

and thus it should also define the other three special member functions:
  - a move-constructor,
  - a copy-assignment operator (probably as deleted),
  - and a move-assignment operator.

Here the bug comes from the lack of the move-assignment.

maxGimeno:
So the real bug would be to add a move-assignment operator that deletes the current `et` and then replace it by the target's `et`, right ?

lrineau:
> So the real bug would be to add a move-assignment operator that deletes the current `et` and then replace it by the target's `et`, right ?

Yes, the fix would be:
  - define a deleted copy-assignment,
  - define a move-assignment, that: delete the current `et`, replace it by the new one, *and* set `et` to `nullptr` in the other rep
  - define a move-constructor that does about the same as the move-assignment.

lrineau (approved):

maxGimeno:

PR #5794, opened 2021-06-22
WIP AABB_tree: workaround for unbalanced kd tree

afabri:
## Summary of Changes

For meshes with extremely dense clusters of triangles the kd tree which serves for acceleration of distance queries can be so unbalanced that it leads to a stack overflow  and at least to bad performance. 
This PR calls `grid_simplify()` with a rather adhoc chosen epsilion.  We also introduce a dependency on PMP.
It might also be a better idea to find a more balanced splitter in the kd tree (I should file an issue).

## Release Management

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -830,12 +830,12 @@ public:
   template<typename Tr>
   bool AABB_tree<Tr>::build_kd_tree()
   {
-		Bbox_3 bb;
+    Bbox_3 bb;
     for(const Primitive& p : m_primitives){
-			bb = bb + Helper::get_reference_point(p, m_traits).bbox();
-		}
-		double dmin = (std::min)((std::min)(bb.xmax()-bb.xmin(), bb.ymax()-bb.ymin()),bb.zmax()-bb.zmin());
-		dmin /= 30.0;
+      bb = bb + Helper::get_reference_point(p, m_traits).bbox();
+    }
+    double dmin = (std::min)((std::min)(bb.xmax()-bb.xmin(), bb.ymax()-bb.ymin()),bb.zmax()-bb.zmin());
+    dmin /= 30.0;
 
     // iterate over primitives to get reference points on them
     std::vector<Point_and_primitive_id> points;
@@ -843,11 +843,11 @@ public:
     for(const Primitive& p : m_primitives)
       points.push_back( Point_and_primitive_id( Helper::get_reference_point(p, m_traits), p.id() ) );
 
-		First_of_pair_property_map<Point_and_primitive_id> foppm;
-		typename std::vector<Point_and_primitive_id>::iterator
-			it = grid_simplify_point_set(points, dmin, CGAL::parameters::point_map(foppm));
+    First_of_pair_property_map<Point_and_primitive_id> foppm;
+    typename std::vector<Point_and_primitive_id>::iterator
+      it = grid_simplify_point_set(points, dmin, CGAL::parameters::point_map(foppm));
 
-		std::cout << "Insert " << std::distance(points.begin(), it)  << " out of " << points.size() << " in the kd tree" << std::endl;
+    std::cout << "Insert " << std::distance(points.begin(), it)  << " out of " << points.size() << " in the kd tree" << std::endl;
     // clears current KD tree
     return build_kd_tree(points.begin(), it);
   }
--- a/AABB_tree/include/CGAL/AABB_tree.h
+++ b/AABB_tree/include/CGAL/AABB_tree.h
@@ -832,10 +832,10 @@ public:
   {
     Bbox_3 bb;
     for(const Primitive& p : m_primitives){
-      bb = bb + Helper::get_reference_point(p, m_traits).bbox();
-    }
-    double dmin = (std::min)((std::min)(bb.xmax()-bb.xmin(), bb.ymax()-bb.ymin()),bb.zmax()-bb.zmin());
-    dmin /= 30.0;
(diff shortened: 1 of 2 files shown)


lrineau:
I think this change my degrade the performance of `AABB_tree::closest_point`. Before that patch, the kd-tree was quickly giving the closest *vertex* of the query point, and then a few rounds of decreasing the radius of the sphere probe would find the closest point. With this patch, on average only 1 vertex out of 900 will be in the kd-tree, and the rounds of decreasing the radius of the sphere will be more important. In my opinion, the code in `CGAL::AABB_traits::closest_point` is a lot more costly than the recursion in the kd-tree. In the end, the code will probably be slower.

AABB_tree/include/CGAL/AABB_tree.h
```diff
@@ -829,14 +830,26 @@ namespace CGAL {
   template<typename Tr>
   bool AABB_tree<Tr>::build_kd_tree()
   {
+    Bbox_3 bb;
+    for(const Primitive& p : m_primitives){
+      bb = bb + Helper::get_reference_point(p, m_traits).bbox();
+    }
+    double dmin = (std::min)((std::min)(bb.xmax()-bb.xmin(), bb.ymax()-bb.ymin()),bb.zmax()-bb.zmin());
+    dmin /= 30.0;
```
lrineau, line 838: This gives `dmin==0` if the point set is flat in one of the three axis, and then:
```
terminate called after throwing an instance of 'CGAL::Precondition_exception'
Expr: epsilon > 0
File: [...]]/cgal/Point_set_processing_3/include/CGAL/grid_simplify_point_set.h
Line: 228
```

afabri:
Surprised to see a 2D mesh.  And maybe it should be max and not min.   For the records, what we observed in one data set of the nurbs mesher was 
```
Number of items stored: 3182745
Number of nodes: 1475173
 Tree depth: 28458
That is the statistics for the kd-tree
The depth shows how unbalanced it is
The kd-tree offers several split functions, but the default is the best.
```

afabri:
Sorry, but the tabs are back. I'll fix my .emacs

afabri:
I just found this [publication](http://jcgt.org/published/0004/01/03/paper.pdf) entitled  _Building a Balancedk-d Tree inO(knlogn)Time_  which might be a good idea to look at.

danston:
Please have a look at my more detailed answer [here](https://github.com/CGAL/cgal/issues/5812#issuecomment-877246175). Regarding AABB tree, I would suggest the following:

1. Create a point set from the input mesh.
2. Clean it (remove duplicates, near-collinear points).
3. Provide these points as input to the function `accelerate_distance_queries()` (see [here](https://doc.cgal.org/latest/AABB_tree/classCGAL_1_1AABB__tree.html#a974e4c1bf69cfd46fd32631a80321a27) of the AABB tree) so that the internal Kd tree is built upon nice points. 

I hope that could be a quick fix to the problem.

sloriot:
I guess it should even be an option for the kd-tree to remove automatically duplicated points.

danston:
> I guess it should even be an option for the kd-tree to remove automatically duplicated points.

I agree. Might be a good idea.

afabri:
It should only be an option as the points may have associated information, the point  only being member of a struct.

lrineau:
> Please have a look at my more detailed answer [here](https://github.com/CGAL/cgal/issues/5812#issuecomment-877246175). Regarding AABB tree, I would suggest the following:
> 
> 1. Create a point set from the input mesh.
> 2. Clean it (remove duplicates, near-collinear points).
> 3. Provide these points as input to the function `accelerate_distance_queries()` (see [here](https://doc.cgal.org/latest/AABB_tree/classCGAL_1_1AABB__tree.html#a974e4c1bf69cfd46fd32631a80321a27) of the AABB tree) so that the internal Kd tree is built upon nice points.
> 
> I hope that could be a quick fix to the problem.

Can you explain why colinear points should be removed as well? Do they compromise the balancing of the kd-tree? Is that a performance issue?

danston:
I did not check for collinear points because it is hard to identify them in the given data set. But as far as I understood from the KD tree docs [here](https://doc.cgal.org/latest/Spatial_searching/index.html#Spatial_SearchingSplitting_rule_section):

> The tree can become linear for the median rules, if many points are collinear in a dimension which is not the cutting dimension.

So, it is not for all rules but I suppose it could be an issue that compromises the balancing of the kd tree in general since it is a known edge case.

afabri:
It compromises the balancing but I suggest not to spend 1 sec on identifying and removing them. Instread implement the paper as ithe algorithm is trivial.

danston:
But as far as I understand, the algorithm from the paper does not solve the main problem with duplicate points. Moreover, the algorithm is designed for integer coordinates x, y, z only, and in our case they are doubles. See also [here](https://github.com/CGAL/cgal/issues/5812#issuecomment-877246175). So we need to adapt the algorithm.

danston:
I did a quick test for the new algorithm using duplicate points and it seems like it is not that bad after all but my test data are rather small. I will check what I can do with doubles and make a quick sketch of the new splitter based on this paper.

afabri:
Dealing with duplicates is needed when searching an item. It might be a solution to make a copy of the first xyz index sequence because identical points are neighbors then and if one finds an item exactly one has to look for identical left and right neighbors.
I don’t see where the algorithm exploits that the nubers are integers. x:y:z isjust a triple of doubles and we sort them lexicographically.

danston:
You are right about duplicates. I did more tests (with larger inputs) with the new algorithm and I think we can handle that.

The paper has only integer examples and the code that comes with the paper supports only integers, but indeed it seems like it is not difficult to make it work with doubles.

danston:
This PR is replaced by PR #5854.

PR #4908, opened 2020-08-06
Polyhedron demo: Scene_triangulation_3_item

maxGimeno:
## Summary of Changes
Add an item and plugin for the Triangulation_3.

Attention: This PR is based on #4255 
## Release Management
fix #4819 
* Affected package(s):Polyhedron_demo

PR #5809, opened 2021-06-28
Polyhedron Demo: Filtering c3t3 item display

### LINKED ISSUE #5612: Demo Improvements for display of tetmeshes
Some features that are missing and that would be nice to have:

- add a way to filter what we see. For example, in the case of a complex, we might want to look at only one subdomain. There could be a menu with subdomain to display or even better it could directly be in the item list below `Intersection tetrahedra` (shouldn't it be `Intersected`? btw)
- Display tetra that are below some criteria. The most obvious case would be to see where are the slivers (with a bound on the dihedral angle provided by the user).
- show tetrahedra (there is an entry in the context menu but it is only for intersected tetrahedra AFAIS)

maxGimeno:
## Summary of Changes
This PR introduces a new plugin : the Mesh_3_filter_domain_plugin.
image

For each subdomain of the selected c3t3_item, it will create a colored checkable button, as you can see on the previous screenshot. The color of each button matches the color of the subdomain that it represents (which is not always the color of the surface_patch. Basically, the button will be of the same color  as the tetrahedras of the intersection of a subdomain, but not of the facets of the surface of that domain. On the screenshot, for example, the green button matches the green tets but the orange facets).
Checking a button will hide/show the corresponding subdomain. On the following screenshot, it is the same item in the same position but the green subdomain (255) is disabled.
image
This operation is performed without loading time, the filtering is done in the shaders (except for the tets od the intersection, which is always entirely recomputed when moving the plane, even in the current state (master)).

There is also now a new item type that allows to filter tetrahedras according to a criterion : 
image
You can use the two sliders to define an interval on that criterion, and all tetrahedra that don't fit in it will be discarder.
image

image

Limitations:
Filtering domains requires  converting the list of the visible domains into 4 integers, passing these to the shaders, and performing shifting on them. This is why this plugin is not available for :
 - users that do not have a recent enough version of OpenGL (4.0), because the shifting is only available since GLSL 1.30.
 - c3t3 items with more than 96 different subdomains, (4x24 bits). The operation require to convert the ulongs to floats to be passed to the shaders, from where converting them back uint drop the "fractional part of the float ([page 55 of the specs ](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.50.pdf)), which basically only keeps the 24 first bits."

## Release Management

* Issue(s) solved (if any): fix #5612 (partially)
* Depends on #4908


maxGimeno:
TODO: don't put all the buttons on the same line

Polyhedron/demo/Polyhedron/Scene_tetrahedra_item.cpp
```diff
@@ -0,0 +1,396 @@
...
+  d->extrema[2][1] =  0;
+  d->extrema[3][0] = 1.0;
+  d->extrema[3][1] = 0.0;
+  std::vector<double> tmp_min, tmp_max, tmp_rrr, tmp_v;
+  tmp_min.reserve(c3t3.number_of_cells_in_complex());
+  tmp_max.reserve(c3t3.number_of_cells_in_complex());
+  tmp_rrr.reserve(c3t3.number_of_cells_in_complex());
+  tmp_v.reserve(c3t3.number_of_cells_in_complex());
+
+  int keep = 0;
```
sloriot, line 172: variable unused (warning in the testsuite)

sloriot:

lrineau:
@sloriot I think you requested that feature. Have verified that the code does what is expected?

sloriot:
Yes

PR #5813, opened 2021-06-29
Basic viewer bugfix

gdamiand:

## Summary of Changes

Do not clear the pos buffer of clipping plane, when clearing the basic viewer.
Otherwise, the clipping plane is no more drawn after a clear.

Recompute the clipping plane in initialize_buffers, since the bounding box is maybe not the same and so the size of the plane.

Now this pr is ready for testing.

## Release Management

### DIFF
--- a/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
+++ b/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
@@ -962,7 +962,7 @@ protected:
       if(m_frame_plane)
       {
         for(int i=0; i< 16 ; i++)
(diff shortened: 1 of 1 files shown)


lrineau:
@maxGimeno Try to see if we can sneak that fix into CGAL-5.3...

sloriot:

lrineau:
This PR was initially planned as a bug-fix for CGAL-5.3.1. Do we backport it?

gdamiand:
> This PR was initially planned as a bug-fix for CGAL-5.3.1. Do we backport it?

We can, but the bug is light thus we can keep the PR only for 5.4, as you prefer.

PR #5821, opened 2021-07-01
Init pointer used in comparaison when adding curves on right during init

### LINKED ISSUE #5797: Uninitialized member in No_intersection_surface_sweep_2
Valgrind complains that [this conditional statment](https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/Surface_sweep_2/include/CGAL/Surface_sweep_2/Surface_sweep_2_impl.h#L436) can be reached with `m_currentEvent` still uninitialized.

I'm not familiar with the inner workings of the code, but from a quick glance, it seems that `m_currentEvent` is only set [here](https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/Surface_sweep_2/include/CGAL/Surface_sweep_2/No_intersection_surface_sweep_2_impl.h#L145), but, besides the fact that the loop may not execute at all, it seems that the uninitialized jump can also take place before this loop is reached at all (i.e. it's not just a matter of the case of `m_queue` being empty).

Anyway, I've tried the quick test of adding `m_currentEvent(NULL)` to the member initializer list of the constructors of `No_intersection_surface_sweep_2` and the errors go away (with the program still running fine), but I have no idea whether this is the correct fix.  If it is, and you want me to make a PR, of if you need a test case, let me know.

sloriot:
Fixes #5797

### DIFF
--- a/Surface_sweep_2/include/CGAL/Surface_sweep_2/No_intersection_surface_sweep_2_impl.h
+++ b/Surface_sweep_2/include/CGAL/Surface_sweep_2/No_intersection_surface_sweep_2_impl.h
@@ -33,6 +33,7 @@ No_intersection_surface_sweep_2<Vis>::
 No_intersection_surface_sweep_2(Visitor* visitor) :
   m_traits(new Traits_adaptor_2()),
   m_traitsOwner(true),
+  m_currentEvent(nullptr),
   m_statusLineCurveLess(m_traits, &m_currentEvent),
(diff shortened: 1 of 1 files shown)


lrineau:
@sloriot I just revived this PR, that was closed by accident when I closed the branch `5.1.x-branch`. I have changed its base branch to `5.2.x-branch`.

sloriot:

PR #5825, opened 2021-07-01
PMP::isotropic_remeshing() - improve split step

janetournois:
## Summary of Changes

The "split long edges" step of `PMP::isotropic_remeshing()` was not taking into account the edges created by splits. With this PR, newly created long edges are added to the range of long edges to be split.

## Release Management

* Affected package(s): PMP


lrineau:
Should it be integrated in release branches, as a bug fix? Or will it be only for new releases (CGAL-5.4 and later)?

PR #5838, opened 2021-07-07
Online Doc: Fix condition for push

maxGimeno:
## Summary of Changes
Apparently git diff actually returns 0 if there is a diff.


sloriot:
@maxGimeno this is not tested by the testsuite. I guess you already tested it and it can be merged, right?

maxGimeno:
yes

PR #5863, opened 2021-07-23
Fix index maps for non-const Face_filtered_graph

MaelRL:
## Summary of Changes

`Face_filtered_graph` defined its `property_map<CGAL::Face_filtered_graph<Graph, FIMap, VIMap, HIMap>, CGAL::XXX_index_t>::type` as its const type (a property binder to the underlying graph), which led to type conflicts in the `get()`. Also missing a `get(tag, graph)` for non-const filtered graphs.

This PR puts the index property map of a non-const FFG as a dynamic property map of the underlying graph (not some kind of unordered set despite a FFG being usually small because it feels better to spend memory than runtime).

## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): -

### DIFF
--- a/BGL/include/CGAL/boost/graph/Face_filtered_graph.h
+++ b/BGL/include/CGAL/boost/graph/Face_filtered_graph.h
@@ -1206,40 +1206,6 @@ CGAL_FFG_DYNAMIC_PMAP_SPEC(dynamic_face_property_t)
 
 #undef CGAL_FFG_DYNAMIC_PMAP_SPEC
 
-//specializations for indices
-template <class Graph,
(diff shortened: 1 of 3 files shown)


sloriot:
What is the minimal example to reproduce the issue?

MaelRL:
Something like:
```
PolygonMesh pmesh;
CGAL::Face_filtered_graph<PolygonMesh> ffg(pmesh);
auto vim = CGAL::get_initialized_vertex_index_map(ffg);
```

should be enough.

sloriot (approved):

PR #5865, opened 2021-07-23
Filter testsuite

maxGimeno:
## Summary of Changes
Adds files and change scripts to run the Filter testsuite from github.

## Release Management

* Affected package(s):Scripts, Maintenance, Testsuite


.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,100 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v3
+        id: check
+        with:
+         result-encoding: string
+         script: |
+           const userName = context.payload.comment.user.login
+           if(userName == 'maxGimeno') {
```
lrineau, line 17: We'll have to adapt that logic! :smile:

lrineau:
- [ ] add the flags to have the "normal" warnings

.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,105 @@
...
+             const body = context.payload.comment.body
+             if(body.includes("/testme")) {             
+               return 'OK'
+             }
+           }
+           return 'stop'
```
lrineau, line 23: This step should be replaced by a [`jobs.<job>.if`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif), because `comment.user.login` and `comment.body` are directly usable in a Github Action expression, and there exists the function [`contains`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains).

I will probably merge the PR anyway, to test it on https://github.com/CGAL/cgal, but my suggestion will speed-up the reaction time of our Github Actions: a `step` forks a container on a runner, and that is "long" (waiting in the wait queue, then about 15 seconds), whereas expressions seem to be evaluated even before entering the way queue (there is probably another wait queue that we do not see).

.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,105 @@
...
+        id: get_label
+        with:
+         result-encoding: string
+         script: |
+           //get branch name and username
+           const pr_url = context.payload.issue.pull_request.url
+           const pr_content = await github.request(pr_url)
+           const label = pr_content.data.head.label
+           console.log(label)
+           return label
```
lrineau, line 35: This step cannot be skipped, because a `comment` even does not have access to the full PR payload, and we need to request it using `github.request`.

.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,105 @@
...
+        id: get_base
+        with:
+         result-encoding: string
+         script: |
+           //get branch name and username   
+           const pr_url = context.payload.issue.pull_request.url
+           const pr_content = await github.request(pr_url)
+           const base = pr_content.data.base.ref
+           console.log(base)
+           return base
```
lrineau, line 47: This one can certainly be merged into the previous one.

.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,105 @@
...
+      - uses: actions/github-script@v3
+        if: steps.get_round.outputs.result != 'stop'
+        id: get_pr_number
+        with:
+         result-encoding: string
+         script: |
+           //get pullrequest url
+           const pr_number = context.payload.issue.number
+           return pr_number
```
lrineau, line 56: Can be removed. `issue.number` is directly a value in the payload. It can be accessed using the syntax `${{...}}}`.

.github/workflows/filter_testsuite.yml
```diff
@@ -0,0 +1,72 @@
...
+on:
+  issue_comment:
+      types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v3
+        if: github.event.comment.user.login == 'maxGimeno' && contains(github.event.comment.body, '/testme')
```
lrineau, line 12: Here you use [`jobs.<job_id>.steps[*].if`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif), and the conditional could be factorized to the job, using [`jobs.<job_id>.if`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif), I think. I have not tested. I just read the doc.

Scripts/developer_scripts/autotest_cgal
```diff
@@ -380,6 +380,8 @@ setup_dirs()
   CGAL_DIR=`readlink "${CGAL_ROOT}/CGAL-I"`
 
   CGAL_TEST_DIR=${CGAL_DIR}/test
+  CGAL_DATA_DIR=${CGAL_DIR}/data
+  export CGAL_DATA_DIR=$(echo "$CGAL_DATA_DIR" | sed -E 's/\/cygdrive\/([a-z])\//\U\1:\//')
```
lrineau, line 384: That seems in conflict with #4229. Is that on purpose?

maxGimeno: It is here to be able to test #4229, yes.

lrineau (approved):

lrineau:
I think we can merge this PR. The only files that could impact the CGAL testsuite are:
- `Testsuite/test/collect_cgal_testresults_from_cmake`,
- `Scripts/developer_scripts/autotest_cgal`, and
- `Testsuite/test/collect_cgal_testresults_from_cmake`
and none of them have been modified since the last time this PR was tested (see https://github.com/CGAL/cgal/pull/5865#event-5074189476).

lrineau:
I still see possible improvement, like the factorization of copy-pasted files in `Maintenance/test_handling/filter_testsuite/`, but we can do that later.

Do you agree @maxGimeno?

maxGimeno:
Yes, let's merge it

PR #5879, opened 2021-07-29
Update PLY.h

Cervidellus:
Aligned Edge property to PLY standard

_Please use the following template to help us managing pull requests._

## Summary of Changes

The PLY standard typically uses the following properties for edge elements:
property int vertex1  
property int vertex2

whereas PLY.h was using
property int v0
property int v1

This caused the resulting files to not be read properly by common software packages such as Blender or Meshlab when edge elements were written.

## Release Management

* Affected package(s):
* cgal\Surface_mesh\include\CGAL\Surface_mesh\IO\PLY.h
* Issue(s) solved (if any): fix #0000, fix #0000,...
* Feature/Small Feature (if any):
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership:

PR #5881, opened 2021-07-30
Aligned Edge property to PLY standard

sloriot:
replaces #5879


maxGimeno:

PR #5882, opened 2021-07-30
Update PLY files to valid format

sloriot:
Requires #5881


lrineau:
https://github.com/CGAL/cgal/pull/5881 target CGAL-5.2.x, and this one targets CGAL-5.3.x. Is that on purpose?

sloriot:
Yes those files don't exist in 5.2 (as well as the part of the header modified)

maxGimeno:

PR #5854, opened 2021-07-15
[Small Feature] Spatial Searching Balanced Kd Tree

danston:
This draft PR tries to modify the Kd tree class in order to support a new balanced Kd tree build from [this paper](http://jcgt.org/published/0004/01/03/).

## Release Management

* Affected package(s): `Spatial_searching`
* Issue(s) solved (if any): #5812 
* Feature/Small Feature (if any): Small Feature [not yet submitted]
* Link to compiled documentation (obligatory for small feature): docs [still to be added]
* License and copyright ownership: GF

Tasks:

- [ ] Add docs.
- [ ] Submit a small feature.
- [ ] Finish implementation.
- [ ] Test for various edge cases.
- [ ] Update changes.md.
- [ ] Check for memory leaks.
- [ ] Remove all experimental code (such as preprocess in Kd tree and modifs in AABB tree).

PR #5884, opened 2021-07-30
Filtered Kernel Construct Cartesian Iterator Return Type Fix

### LINKED ISSUE #5878: EPECK Construct Cartesian d Seg Fault
## Issue Details

I bumped into an issue when using `construct_cartesian_const_iterator_d_object()` from `Search_traits_3` parameterized by EPECK.

When compiling an example below, I first get a warning:

```bash
cgal/Filtered_kernel/include/CGAL/Filtered_kernel/Cartesian_coordinate_iterator_3.h:119:12: warning: returning reference to local temporary object [-Wreturn-stack-address]
    return *(*this + i);
           ^~~~~~~~~~~~
main.cpp:18:19: note: in instantiation of member function 'CGAL::Cartesian_coordinate_iterator_3<CGAL::Epeck>::operator[]' requested here
  const auto d = p[0] - q[0];
```

and then when running the code, it triggers an error:

```bash
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x0000000100014560 test_epeck`CGAL::Interval_nt<false>::operator-(this=0x00007ffeefbff078) const at Interval_nt.h:173:27
   170    IA operator-() const
   171    {
   172  #ifdef CGAL_USE_SSE2
-> 173      return IA (swap_m128d(val));
   174  #else
   175      return IA (-sup(), -inf());
   176  #endif
```

Is it correct that `operator[]()` returns a ref for EPECK?

## Source Code

I created a simple [minimal example](https://gist.github.com/danston/bf00bd7092d5943ebc2c189ae0465fa2) that illustrates the issue.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): macOS
* Compiler: clang 12.0.0
* Release or debug mode: both
* Specific flags used (if any): `-Wall -Wextra`
* CGAL version: master
* Boost version: 1.76.0
* Other libraries versions if used (Eigen, TBB, etc.): nope

danston:
This little PR proposes a fix for the issue #5878 by changing some return types in the `Construct_cartesian_iterator_2` and `Construct_cartesian_iterator_3` classes of the `Filtered_kernel`.

Tasks:
- [x] Check memory leaking.

## Release Management

* Affected package(s): `Filtered_kernel`
* Issue(s) solved (if any): fix #5878
* Feature/Small Feature (if any): bug

### DIFF
--- a/Filtered_kernel/include/CGAL/Filtered_kernel/Cartesian_coordinate_iterator_2.h
+++ b/Filtered_kernel/include/CGAL/Filtered_kernel/Cartesian_coordinate_iterator_2.h
@@ -42,6 +42,8 @@ public:
   typedef std::random_access_iterator_tag iterator_category;
   typedef FT                              value_type;
   typedef int                             difference_type;
+  typedef void                            pointer;
+  typedef value_type                      reference;
 
   Cartesian_coordinate_iterator_2()
     : var((const P*) nullptr), index(0) {}
@@ -53,7 +55,7 @@ public:
     : var(v), index(_index) {}
 
 
-  decltype(auto)
+  reference
   operator*() const {
     if (const P* const* p = boost::get<const P*>(&var))
       return (*p)->cartesian(index);
@@ -118,7 +120,7 @@ public:
     return index - x.index;
   }
 
-  decltype(auto)
+  reference
   operator[](difference_type i) const {
     return *(*this + i);
   }
--- a/Filtered_kernel/include/CGAL/Filtered_kernel/Cartesian_coordinate_iterator_2.h
+++ b/Filtered_kernel/include/CGAL/Filtered_kernel/Cartesian_coordinate_iterator_2.h
@@ -55,7 +55,7 @@ public:
     : var(v), index(_index) {}
 
 
-  const FT
+  decltype(auto)
   operator*() const {
     if (const P* const* p = boost::get<const P*>(&var))
       return (*p)->cartesian(index);
@@ -64,7 +64,8 @@ public:
     return (*v)->cartesian(index);
   }
 
-  Self&  operator++() {
+  Self&
+  operator++() {
     index++;
     return *this;
   }
@@ -91,7 +92,7 @@ public:
 
   Self&
   operator+=(difference_type i) {
-    index+=i;
+    index += i;
     return *this;
   }
 
@@ -107,7 +108,8 @@ public:
     return tmp += i;
   }
 
-  Self operator-(difference_type i) const {
+  Self
+  operator-(difference_type i) const {
     Self tmp=*this;
     return tmp -= i;
   }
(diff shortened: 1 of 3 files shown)


danston:
Should I maybe target this PR to an early release rather than master?

lrineau:
Indeed, that seems to be a bug-fix, and the bug is probably already in `5.2.x-branch`.

danston:
It is inside 5.2.x so I have changed.

mglisse:
It has been wrong since at least 2003 :wink:
And it has had a comment saying it should go away since 2008 :confused: 

This patch is the absolute minimum to avoid the issue you were hitting. It could be nice to take the chance to also improve the rest (maybe also return `decltype(auto)` from `operator*` (?), make the typedef `reference` consistent with what `*` and `[]` return, etc), but that's not a prerequisite for merging this patch.

danston:
No, in fact, I think it is a good idea. I could clean up a bit the class. I will.

danston:
I have changed the return type of the `operator*()` to `decltype(auto)`. The two classes 2D and 3D are basically the same now except for the point/vector type. I did not change the typedefs. They seem to define the things, which are expected from the name but they are not really used inside the classes except for `difference_type` that is `int`.

danston:
I use the `operator[]` extensively in the new algorithm for `Kd_tree` in #5854 and the fix seems to solve all my problems with EPECK. All my Kd tree tests pass.

mglisse:
The typedefs could be used if someone passes those iterators to some standard library algorithm for instance.
`reference` is the return type of `*` and `[]` while `pointer` is the return type of `->` (or `void` if `->` is not supported in recent standards). In modern C++, I think those typedefs are optional (so it is better not to provide them than to define them wrong), but not yet in C++14.

For your use in a Kd_tree, I haven't looked at the code, but my first instinct would be to avoid constructions (even trivial ones like `point.x()`) and use predicates as much as possible.

mglisse (approved):

danston:
I had another look and decided that it is better to use something as you suggested [here](https://github.com/CGAL/cgal/issues/5878#issuecomment-889219208). I made `pointer` a `void` and return `reference` that is `FT` from both `operator*` and `operator[]`. This way I managed to apply some algorithms from `std` on these iterators (e.g. `std::distance`) and it still fixes the problems I had in `Kd_tree`. 

Unfortunately, I do not see how I can avoid using these operators in `Kd_tree` because I need to iterate over the coordinates of a point and then apply some operations for a given index so I do not see how I could use there CGAL predicates, if I understood you correctly. 

I also checked which files are still using these iterators except for `Kd_tree` and it seems like `Mesh_3` benchmarks are using them as well.

mglisse:
We can talk about that again later, if the performance of Kd_tree with Epeck turns out to be too horrible.

sloriot:
Successfully tested in CGAL-5.4-Ic-38. (I tested several time the branch as I had a doubt about runtime errors in Periodic_3_mesh_3 but I could spotted the same error happening in CGAL-5.4-I-24).

PR #5894, opened 2021-08-09
Scripts: Fix and enhance Filter testsuite

maxGimeno:
## Summary of Changes
Remove the platform directory before running the filter testsuite if exists, and filter diff of files to ignore /doc files and every non .h or .cpp file.

## Release Management

* Affected package(s):Scripts

### DIFF
--- a/Scripts/developer_scripts/run_testsuite_from_branch_name.sh
+++ b/Scripts/developer_scripts/run_testsuite_from_branch_name.sh
@@ -63,6 +63,9 @@ cd ${CGAL_ROOT}
 
 if [ -L CGAL-I ]; then rm CGAL-I; fi
 ln -s $PWD/CGAL-TEST/$DEST CGAL-I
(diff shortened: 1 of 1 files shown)


Scripts/developer_scripts/run_testsuite_from_branch_name.sh
```diff
@@ -33,7 +33,7 @@ git fetch $USER_REPO
 git checkout $BRANCH_NAME
 git reset --hard $USER_REPO/$BRANCH_NAME
 #setup the list_test_packages
-TMP_LIST=$(git diff --name-only cgal/$BASE_NAME...HEAD |egrep -v /doc |egrep "*\.h"\|"*\.cpp" |cut -s -d/ -f1 |sort -u | xargs -I {} ls -d {}/package_info 2>/dev/null  |cut -d/ -f1 |egrep -v Installation||true)
```
lrineau: mac or not, a regexp starting with `*` is not correct.

PR #5904, opened 2021-08-13
Replace macros with inline functions and remove a config flag

sloriot:
re-implement CGAL_AND and CGAL_OR with lambdas to improve MSVC version

Also remove some no longer relevant config flags


STL_Extension/include/CGAL/Uncertain.h
```diff
@@ -520,6 +494,69 @@ Uncertain<T> make_uncertain(Uncertain<T> t)
...
+  return certainly(a) ? make_uncertain(true)
+                          : a | b;
+}
+
+inline
+Uncertain<bool> CGAL_OR(bool a, const Uncertain<bool>& b)
+{
+  return a ? make_uncertain(true)
+            : a | b;
+}
```
lrineau, line 542:
```suggestion
inline
Uncertain<bool> CGAL_OR(bool a, const Uncertain<bool>& b)
{
  return a ? make_uncertain(true)
            : b;
}
```

STL_Extension/include/CGAL/Uncertain.h
```diff
@@ -520,6 +494,69 @@ Uncertain<T> make_uncertain(Uncertain<T> t)
...
+  return certainly_not(a) ? make_uncertain(false)
+                          : a & b;
+}
+
+inline
+Uncertain<bool> CGAL_AND(bool a, const Uncertain<bool>& b)
+{
+  return !a ? make_uncertain(false)
+            : a & b;
+}
```
lrineau, line 515:
```suggestion
inline
Uncertain<bool> CGAL_AND(bool a, const Uncertain<bool>& b)
{
  return !a ? make_uncertain(false)
            : b;
}
```

STL_Extension/include/CGAL/Uncertain.h
afabri: because a  is `false   in the `else` case no need to 'or'  it.

STL_Extension/include/CGAL/Uncertain.h
afabri: because `a` is `true`  in the `else` case no need to 'and' it.

afabri:
Do we want `CGAL_AND`  to become `CGAL::AND`   or `CGAL::CGAL_AND`?

lrineau:
> Do we want `CGAL_AND` to become `CGAL::AND` or `CGAL::CGAL_AND`?

Should not we keep `CGAL_AND` and `CGAL_OR` as macros, and implement then using inline functions? That would preserve the compatibility with third-party code. That is even part of the [documented API of `Uncertain`](https://doc.cgal.org/5.3/STL_Extension/classCGAL_1_1Uncertain.html).

sloriot:
> > Do we want `CGAL_AND` to become `CGAL::AND` or `CGAL::CGAL_AND`?
> 
> Should not we keep `CGAL_AND` and `CGAL_OR` as macros, and implement then using inline functions? That would preserve the compatibility with third-party code. That is even part of the [documented API of `Uncertain`](https://doc.cgal.org/5.3/STL_Extension/classCGAL_1_1Uncertain.html).

I agree

STL_Extension/include/CGAL/Uncertain.h
```diff
@@ -520,6 +494,69 @@ Uncertain<T> make_uncertain(Uncertain<T> t)
...
+// they lack the "short-circuiting" property.
+// We provide macros CGAL_AND and CGAL_OR, which attempt to emulate their behavior.
+// Key things : do not evaluate expressions twice, and evaluate the right hand side
+// expression only when needed.
+
+inline
+Uncertain<bool> CGAL_AND(const Uncertain<bool>& a, const Uncertain<bool>& b)
+{
+  return certainly_not(a) ? make_uncertain(false)
+                          : a & b;
```
mglisse, line 507: That looks equivalent to just `a&b` and doesn't provide the short-circuiting behavior that was the whole reason for using statement expressions. You could use this to replace the `&&` fallback for CGAL_CFG_NO_STATEMENT_EXPRESSIONS, but not for gcc/clang.
As indicated in the old comment, you could try using lambdas, like (completely untested)
```c++
#define CGAL_AND(A,B) plouf((A),[&](){return (B);})
template<class BF>
auto plouf(Uncertain<bool> a, BF&&bf){
  if(certainly_not(a))return make_uncertain(false);
  return a & make_uncertain(bf());
}
```

sloriot: I don't understand why the code I wrote is not equivalent to the old one. I'm probably missing something but to me the goal was to evaluate `a&b` only if a is not certainly false.

mglisse: Currently, `CGAL_AND(false, f())` does not call `f` at all with gcc/clang. And `f` may be some rather costly predicate... Evaluating `a&b` is cheap if `a` and `b` are already known, the point is to avoid computing b, just like `false&&f()` does not call `f`. In the code you wrote, you always compute both `a` and `b` to pass them to the function. In the macro code, the computation of `b` is in a branch that is not always taken.

sloriot: oh I see, b might be the result of a function directly passed to the macro.

lrineau:
> /testme

@maxGimeno is still, as far as I know, the only one that can trigger that action. Please Maxime, transfer that power to Sébastien and me. We will have to be careful with that, because the action behavior is undefined if it is triggered concurrently.

maxGimeno:
done in #5942

lrineau:
/testme

It should work, now.

... We have to be careful not to quote any message with those special words, because the quotation itself would trigger the action!

lrineau:
The testsuite results are at https://cgal.geometryfactory.com/~mgimeno/test_suite/TESTRESULTS/results-5.4-Ic-5904.shtml. Looks good.

sloriot:

PR #5916, opened 2021-08-18
Mesh 3 - mesh polyhedra with self intersections

janetournois:
## Summary of Changes

This PR introduces a new visitor for facet criteria to be able to deal with self-intersecting polyhedral inputs.

In the current version (c4142fa), there are 2 breaking changes :
- `Mesh_domain::Intersection` is changed from `std::tuple<Point_3, Index, int>` to `std::tuple<Point_3, Index, int, bool>`
- a new field `std::array<bool, 4> self_intersections_table_` along with the associated `set_facet_on_self_intersection()` and `is_facet_on_self_intersection()` has been added to `Compact_mesh_cell_base`

## Release Management

* Affected package(s): Mesh_3
* Feature/Small Feature (if any): todo


lrineau:
Would you mind using `std::bitset<4>` instead of `std::array<bool, 4>`? I think that will save memory space.

janetournois:
Sure!
I would like to avoid as much as possible breaking changes, and to discuss that with you

Mesh_3/include/CGAL/Mesh_3/mesh_standard_facet_criteria.h
```diff
@@ -626,7 +626,53 @@ class Facet_criterion_visitor
...
+    : Base(tr, f)
+    , has_self_intersections_(false)
+  {
+    if (f.first->is_facet_on_self_intersection(f.second))
+      has_self_intersections_ = true;
+    else
+    {
+      const Facet& mf = tr.mirror_facet(f);
+      if(mf.first->is_facet_on_self_intersection(mf.second))
+        has_self_intersections_ = true;
```
lrineau, line 654: Is that possible that a facet and its mirror facet disagree of the "self intersection" property?

janetournois: they should not

lrineau: Then I think you can just check `f.first->is_facet_on_self_intersection(f.second)`.

lrineau:
I have reviewed this PR. The functionality is certainly something we want to have in CGAL. However, as Jane said, that is a breaking change, where two concepts are amended: `MeshDomain_3` and `MeshCellBase_3`.

Even if only polyhedral mesh domains can benefit from the feature, for the moment all the users of Mesh_3 will be impacted by the new additions to the two concepts. Maybe we can find out a way to require the new fields only conditionally.

janetournois:
I agree
We willl discuss that next week!

janetournois:
I have the feeling that now that we have `facet_min_size`, this PR becomes useless. Both approaches tend to ignore problematic regions.
@lrineau @mael I move this PR to "stalled" for now. Do you think we should close it?

lrineau:
I agree with you that the `facet_min_size` feature is a more general solution to the issue. We can close this PR.

PR #5931, opened 2021-08-24
Interaction between `HAVE_DOT` setting and inheritance graph

albert-github:
In the BGL package we get a warning like:
```
/home/cgal-testsuite/cgal_doc_build/CGAL-5.4-Ic-33/doc/BGL/PackageDescription.txt:29: warning: ignoring \dot command because HAVE_DOT is not set
```
The BGL package uses the `\dot` and has now to use the setting `HAVE_DOT=YES` to see the dot image.
Till now this went OK as the `\dot` command didn't look at the setting of the `HAVE_DOT` setting in the Doxyfile, though by the issue https://github.com/doxygen/doxygen/issues/7273 and the pull request https://github.com/doxygen/doxygen/pull/8663 it is now required that `HAVE_DOT` is set.
The idea of setting `HAVE_DOT` for the BGL project removed the warning (and shows the figure again) though has also as side effect that the inheritance graphs are shown. Until now CGAL shows for the inheritance just the text version.

Doxygen now has for the `CLASS_GRAPH` setting the possibility `TEXT` that does exactly what is required.

Due to the general setting of `HAVE_DOT` also the not needed `graph_legend` was generated and some `INCLUDE_GRAPH`s, by setting this to `NO ` these are now suppressed as well (as they were in the past).


sloriot:
If this setting in introduced with the next release of doxygen, il will break 1.8.20 -> 1.9.2, right?

albert-github:
Yes this will break some things, as "TEXT" is not known it will be reverted back to the default "YES, so no big breaks.
The current CGAL documentation with the doxygen master does not give the correct results, hence the fix.
doxygen I.8.20 is as far as I know not used by CGAL anymore it was used as at the moment of introduction of the 1.8.20 directory this was the master.
There is a small inconsistency between 1.9.2 and newer and the older versions that has been fixed in master and will be fixed in the next release 1.9.3.
Seen a crash and this inconsistency I actually expect that the new release 1.9.3 will be out reasonably soon (I've not discussed a date with Dimitri though, so it is my own guess).

sloriot:
Sure, what I mean is that I could put those changes in resources for 1.9.3 (just checked and the version number is already updated in the master branch of doxygen).

albert-github:
I think that it might even be good to create a new directory for 1.9.3 (or rename 1.8.20) or maybe even better to define a directory doxy_master.
Also It might be good to do a `doxygen -u` on some files as the description texts are sometimes a bit outdated (or maybe even use `doxygen -u -s`).

Documentation/doc/resources/1.9.3/footer.html
```diff
@@ -0,0 +1,21 @@
+<!-- HTML footer for doxygen 1.8.20-->
+<!-- start footer part -->
```
albert-github, line 2: @sloriot 

I think some updating would be in place ...
Probably also valid for other files.

albert-github (commented):
Shouldn't the build procedure not be involved in one way or another too (so the 1.9.3 version will be picked up).

sloriot:
About the cmake machinery, since 1.9.3 is not yet out, the default should still be 1.8.20 but 1.9.3 should correctly be picked up.

Documentation/doc/resources/1.9.3/footer.html
sloriot: I only did a copy/paste for now. I can do the update once doxygen is officially out (to avoid redoing it again).

sloriot:
Generated doc of CGAL-5.4-Ic-36 is fine.

PR #5937, opened 2021-08-27
Robust Quotient with cpp_int

### LINKED ISSUE #5928: How to robustify to_interval(Quotient)
## Issue Details

[Construction of an interval](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/Quotient.h#L686) from a quotient fails if the numerator or denominator is a large integer which does not fit into a `double`.       `Gmpq`  uses  `MPFR`  and we seek for a solution for `boost::multiprecision::cpp_int`.

Any idea is welcome.   I am wondering if a division with remainder as intermediate step can help.   This leaves us with remainder/denominator, that is there is no improvement if the denominator does not fit into a double.  At the same time  we know that this is between 0 and 1 which is a still a bound, even if not very tight.   We now can do a binary search. Can we get it tighter cheaply by looking at the number of limbs?

danston:
This PR provides a benchmark that compares different exact number types available in CGAL and robustifies the new option that is `Quotient<boost::multiprecision::cpp_int>` by providing a better/faster version of the `to_interval()` function with tighter intervals.

The results of the benchmark are available [here](https://cgal.geometryfactory.com/CGAL/Members/wiki/Staging_Area/Number_Types_Benchmark).

Currently, we have three different versions:
- one based on gmpzf (passes tests but not NEF benches)
- one based on boost (passes tests but not NEF benches)
- one based on cpp_rational (stable, works for all benches / tests)

TODO:
- [x] Remove changes from this commit: 6e41cd5  ---> DONE in  [4bd7049](https://github.com/CGAL/cgal/pull/5937/commits/4bd7049f2f88e89e657e95358157bf2d003185f4)
- [x] Remove benchmark data?

## Release Management

* Affected package(s): `Number_types`
* Issue(s) solved (if any): fix #5928
* Feature/Small Feature (if any): Small Feature
* Link to compiled documentation (obligatory for small feature): docs


afabri:
Should the 3 implementations of `to_interval()` return the same result (modulo not being infimum == supremum).

mglisse:
> Should the 3 implementations of `to_interval()` return the same result (modulo not being infimum == supremum).

How tight the interval is may be the most important difference between the implementations. IIRC 2 is the only one that is tight, and 1, 3 may or may not return the same interval depending on the input (maybe 3 always matches either 1 or 2).

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          l = boost::multiprecision::detail::do_cast<double>(p);
+          l = std::ldexp(l, -shift);
+          u = boost::multiprecision::detail::do_cast<double>(q);
+          u = std::ldexp(u, -shift);
+
+          if (change_sign) {
+            const double t = l;
+            l = -u;
+            u = -t;
+          }
```
afabri, line 843: I added this
```

	  Type lb(l), ub(u);
	  assert(lb <= ub);
	  assert(lb <= x);
	  assert(ub >= x);
```   
and it asserts.

afabri:
Hi @mglisse we have put a self contained gist [here](https://gist.github.com/danston/44aa470c849565cb1e25f74d450ddaea) and it must have a bug as the assert shows that the lower bound is not smaller than the value.  We would be glad about a code review, as we can't spot the mistake.

Number_types/include/CGAL/Quotient.h
danston: I simplified the code/input, added these assertions, and now another assertion fails. See the [gist](https://gist.github.com/danston/44aa470c849565cb1e25f74d450ddaea) below.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          double l = 0.0, u = 0.0;
+          if (x.num == 0) { // return [0.0, 0.0]
+            return std::make_pair(l, u);
+          }
+          CGAL_assertion(x.num != 0);
+          CGAL_assertion(x.den != 0);
+
+          // Handle signs.
+          bool change_sign = false;
+          if (x.num < 0 && x.den < 0) {
```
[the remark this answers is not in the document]
danston: If I remember correctly, at least when I create a quotient with a negative denominator, the minus sign shows up in the later computations using this quotient. That is why we check the sign when converting quotient to cpp_rational in the 3rd implementation:
```
  boost::multiprecision::cpp_rational rat;
  CGAL_assertion(x.den != 0);
  if (x.den < 0) {
    rat = boost::multiprecision::cpp_rational(-x.num, -x.den);
  } else {
    rat = boost::multiprecision::cpp_rational( x.num,  x.den);
  }
```

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          } else if (x.num < 0 && x.den > 0) {
+            change_sign = true;
+            x.num = -x.num;
+          } else if (x.num > 0 && x.den < 0) {
+            change_sign = true;
+            x.den = -x.den;
+          }
+          CGAL_assertion(x.num > 0 && x.den > 0);
+
+          Protect_FPU_rounding<true> P(CGAL_FE_TONEAREST);
```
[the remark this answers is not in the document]
danston: I think it does not affect the results in this case but I left it because we figured out that when converting cpp_int to double with boost using `x.convert_to<double>()` we need to set the rounding mode to nearest. I think it can be removed here. I will check.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+                return std::make_pair(-1.0, 0.0);
+              }
+
+            #else // tight bounds using cpp_rational
+
+              // TODO: Is it fast enough? Seems so because this conversion happens
+              // only a few times during the run, at least for NEF.
+              boost::multiprecision::cpp_rational rat;
+              CGAL_assertion(x.den != 0);
+              if (x.den < 0) { // we do it due to the bug in boost
```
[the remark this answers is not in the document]
danston: Wrong comment, it is not a bug.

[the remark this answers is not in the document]
danston: https://github.com/boostorg/rational/issues/27

afabri:
> 
> 
> Hi @mglisse we have put a self contained gist [here](https://gist.github.com/danston/44aa470c849565cb1e25f74d450ddaea) and it must have a bug as the assert shows that the lower bound is not smaller than the value. We would be glad about a code review, as we can't spot the mistake.

In fact we work on it, fixed things, but didn't push yet.

mglisse:
> Hi @mglisse we have put a self contained gist [here](https://gist.github.com/danston/44aa470c849565cb1e25f74d450ddaea) and it must have a bug as the assert shows that the lower bound is not smaller than the value.

Probably an off-by-1 error in how I interpret what msb means.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
       : public CGAL::cpp98::unary_function< Type, std::pair< double, double > > {
       public:
+
+        // =========================================================================== //
+
+        // CURRENT VERSION
+
+        // =========================================================================== //
+
+        std::pair< std::pair<double, double>, long > get_interval_exp( NT& x ) const {
```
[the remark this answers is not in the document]
danston: Fixed both, using `Interval_nt<>` and `int64_t` on return.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+            x.den = -x.den;
+          }
+          CGAL_assertion(x.num > 0 && x.den > 0);
+
+          Protect_FPU_rounding<true> P(CGAL_FE_TONEAREST);
+          const auto n = get_interval_exp(x.num);
+          const auto d = get_interval_exp(x.den);
+
+          CGAL_assertion_msg(
+            CGAL::abs(double(n.second) - double(d.second)) < (1<<30)*2.0,
```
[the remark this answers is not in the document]
danston: This is right. Moreover, the `ldexp` overload for intervals already handles these cases as far as I understand.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          CGAL_assertion(x >= 0);
+          const unsigned n = msb(x);
+          const unsigned num_dbl_digits = std::numeric_limits<double>::digits;
+          if (n > num_dbl_digits) {
+
+            const unsigned d = n - num_dbl_digits;
+            x = x >> d;
+
+            const double l = boost::multiprecision::detail::do_cast<double>(x);
+            const double u = boost::multiprecision::detail::do_cast<double>(x + 1);
```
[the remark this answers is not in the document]
danston: I think using what you proposed is better. I have fixed that.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+
+          } else {
+
+            const double l = boost::multiprecision::detail::do_cast<double>(x);
+            const double u = l;
+            return std::make_pair( std::make_pair(l, u), 0 );
+          }
+        }
+
+        std::pair<double, double> get_interval_as_gmpzf( Type& x ) const {
```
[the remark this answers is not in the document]
danston: This is not anymore the case so no warning is required.

Number_types/include/CGAL/Quotient.h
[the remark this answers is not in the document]
danston: I have removed that.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+            x.den <<= boost::multiprecision::detail::unsigned_abs(shift);
+          }
+          CGAL_assertion(int(msb(x.den) - msb(x.num)) == -53);
+
+          decltype(x.num) p, q, r;
+          boost::multiprecision::divide_qr(x.num, x.den, q, r);
+
+          if (r != 0) {
+            const int q_bits = msb(q);
+            if (q_bits == num_dbl_digits + 1) {
```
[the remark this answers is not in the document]
danston: This part is completely changed and simplified now so it is fixed.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+              ++q;
+            }
+          } else {
+            p = q;
+          }
+
+          l = boost::multiprecision::detail::do_cast<double>(p);
+          l = std::ldexp(l, -shift);
+          u = boost::multiprecision::detail::do_cast<double>(q);
+          u = std::ldexp(u, -shift);
```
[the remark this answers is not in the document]
danston: You are right. I have fixed that.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+            change_sign = true;
+            x.num = -x.num;
+          } else if (x.num > 0 && x.den < 0) {
+            change_sign = true;
+            x.den = -x.den;
+          }
+          CGAL_assertion(x.num > 0 && x.den > 0);
+
+          const int num_dbl_digits = std::numeric_limits<double>::digits;
+          const int shift = num_dbl_digits + (msb(x.den) - msb(x.num));
```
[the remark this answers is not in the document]
danston: True. Fixed.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,234 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+            u = l;
+          }
+          return std::make_pair( Interval_nt<>(l, u), d );
+        }
+
+        std::pair<double, double> get_interval_as_gmpzf( Type x ) const {
+
+          CGAL_assertion_code(const Type input = x);
+          double l = 0.0, u = 0.0;
+          if (x.num == 0) { // return [0.0, 0.0]
```
afabri, line 756: In the code of [`rational_adaptor`](https://github.com/boostorg/multiprecision/blob/rational_adaptor_standalone/include/boost/multiprecision/rational_adaptor.hpp#L33)  they have constants `zero` and `one`.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,234 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          if (r != 0) {
+            p = q;
+            ++q;
+          } else {
+            p = q;
+          }
+
+          CGAL_assertion(p >= 0 && q >= 0);
+          const uint64_t pp = static_cast<uint64_t>(p);
+          const uint64_t qq = static_cast<uint64_t>(q);
```
afabri, line 849: In case this is expensive we can avoid doing it twice in case `p == q`.

afabri:
At the end of arithmetic operations we call `simplify_quotient()`  which divides by the gcd.     In boost they call gcd several times, to simplify intermediate expressions. Does anybody know if that pays off for all sizes of integers, or if in our use case they tend to be small and one gcd call at the end was better?  We might just bench it.

Number_types/include/CGAL/Quotient.h
mglisse: which are just 0 and 1 in type `signed char` or `int` depending on the type. I'd rather use CGAL's is_zero and centralize the choice of how to do that in one place, the specialization of Is_zero.

mglisse:
> At the end of arithmetic operations we call `simplify_quotient()` which divides by the gcd. In boost they call gcd several times, to simplify intermediate expressions.

GMP does the same.

> Does anybody know if that pays off for all sizes of integers, or if in our use case they tend to be small and one gcd call at the end was better? We might just bench it.

It indeed probably depends on many factors (including the size of the inputs and the gcd, how fast gcd and division are for this type, etc), so benchmarking seems right, although the typical size may not be the same for all packages.

Number_types/include/CGAL/Quotient.h
danston: I fixed it using `iz_zero`, `is_negative`, and `is_positive`.

Number_types/include/CGAL/Quotient.h
danston: True. Fixed.

Number_types/include/CGAL/Quotient.h
```diff
@@ -731,13 +732,13 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
 
         std::pair< Interval_nt<>, int64_t > get_interval_exp( NT& x ) const {
 
-          CGAL_assertion(x >= zero);
+          CGAL_assertion(x >= 0);
           int64_t d = 0;
           double l = 0.0, u = 0.0;
           const int64_t n = static_cast<int64_t>(msb(x));
           const int64_t num_dbl_digits = std::numeric_limits<double>::digits - 1;
```
lrineau, line 816: I think that `-1` should be removed. The real value should be 53, and not 52.

danston: It fails with 53 but does not with 52, similar to the other implementation that we discussed together.

danston: I think it is related to the fact that `msb` is an index starting from 0 while `digits` is the size.

lrineau: Here in `get_interval_exp`, I understand, because `num_dbl_digits` is compared, or substracted to `msb(x)`. I would actually write add `+ 1` to the msb value, to avoid confusion.
```c++
          const int64_t n = static_cast<int64_t>(msb(x) + 1);
          const int64_t num_dbl_digits = std::numeric_limits<double>::digits;
```

danston: I see.

Number_types/include/CGAL/Quotient.h
```diff
@@ -685,12 +783,239 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+            CGAL_assertion(msb_diff > num_dbl_digits);
+            x.den <<= boost::multiprecision::detail::unsigned_abs(shift);
+          }
+          CGAL_assertion(num_dbl_digits ==
+            static_cast<int64_t>(msb(x.num)) - static_cast<int64_t>(msb(x.den)));
+
+          NT q, r;
+          boost::multiprecision::divide_qr(x.num, x.den, q, r);
+          const int64_t q_bits = static_cast<int64_t>(msb(q));
+          CGAL_assertion(q_bits == num_dbl_digits || r != 0 /* when q_bit = num_dbl_digits - 1 */ );
```
lrineau, line 921: In that code, `num_dbl_digits` is always compare to a difference of msb, and I think the value could be `53`, with a small change: add `+ 1` to `q_bits`:
```c++
const int64_t q_bits = static_cast<int64_t>(msb(q)) + 1;
```

Number_types/include/CGAL/Quotient.h
danston: I am now using `msb + 1` and I also tested that. It works.

Number_types/include/CGAL/Quotient.h
danston: In fact, `q_bits` is not even used. I put it inside `CGAL_assertion_code` because it is used only inside assertions. So changing it does not change anything. However, if I change 52 to 53, we end up with the zero remainder and then everything goes wrong. This will also require changing the conditions for setting `p` and `q` that is what we tried to do previously and it never worked.

Number_types/include/CGAL/Quotient.h
```diff
@@ -44,7 +46,28 @@ namespace CGAL {
...
+simplify_quotient(boost::multiprecision::cpp_int & a, boost::multiprecision::cpp_int & b) {
+
+  // TODO:
+  // - move it to the boost_mp.h
+  // - can we use gcd only sometimes to save time?
+  const boost::multiprecision::cpp_int r = boost::multiprecision::gcd(a, b);
+  // std::cout << "r: " << r << std::endl;
+  // std::cout << "a: " << a << std::endl;
+  // std::cout << "b: " << b << std::endl;
+  a = a / r;
```
mglisse, line 63: It is a good habit to write `a /= r;` instead, so the compiler and library don't need to go out of their way to notice that it is the same `a`. Same with `>>=` elsewhere.

Number_types/include/CGAL/Quotient.h
```diff
@@ -685,12 +785,239 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+
       public:
+
+        bool are_correct_bounds( const double l, const double u, const Type& x ) const {
+
+          if (
+            CGAL::abs(l) == std::numeric_limits<double>::infinity() ||
+            CGAL::abs(u) == std::numeric_limits<double>::infinity() ||
+            CGAL::abs(l) == 0.0 ||
+            CGAL::abs(u) == 0.0) {
```
mglisse, line 797: Those conditions are strange...

Number_types/include/CGAL/Quotient.h
```diff
@@ -685,12 +785,239 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          CGAL_assertion(CGAL::abs(l) != std::numeric_limits<double>::infinity());
+          CGAL_assertion(CGAL::abs(u) != std::numeric_limits<double>::infinity());
+          CGAL_assertion(CGAL::abs(l) != 0.0);
+          CGAL_assertion(CGAL::abs(u) != 0.0);
+
+          const Type lb(l), ub(u);
+          CGAL_assertion(lb <= ub);
+          CGAL_assertion(lb <= x);
+          CGAL_assertion(ub >= x);
+          return lb <= ub && lb <= x && ub >= x;
```
mglisse, line 809: The first test is redundant with the other 2.

Number_types/include/CGAL/Quotient.h
```diff
@@ -685,12 +785,239 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          CGAL_assertion(x >= 0);
+          int64_t d = 0;
+          double l = 0.0, u = 0.0;
+          const int64_t n = static_cast<int64_t>(msb(x)) + 1;
+          const int64_t num_dbl_digits = std::numeric_limits<double>::digits;
+
+          if (n > num_dbl_digits) {
+            d = n - num_dbl_digits;
+            x = x >> d;
+            const NT y = x + 1;
```
mglisse, line 823: The +1 is still done in type NT, not in type `int64_t` or similar, and there are 2 casts from NT to double. I don't know if it would help, but since you wrote "I think using what you proposed is better. I have fixed that."...

Number_types/include/CGAL/Quotient.h
```diff
@@ -685,12 +785,239 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          const Type lb(l), ub(u);
+          CGAL_assertion(lb <= ub);
+          CGAL_assertion(lb <= x);
+          CGAL_assertion(ub >= x);
+          return lb <= ub && lb <= x && ub >= x;
+        }
+
+        std::pair< Interval_nt<>, int64_t > get_interval_exp( NT& x ) const {
+
+          CGAL_assertion(x >= 0);
```
mglisse, line 814: I think `x>0` (well, `CGAL::is_positive(x)`), since the following fails for 0.

Number_types/include/CGAL/Quotient.h
danston: True. I have fixed that.

Number_types/include/CGAL/Quotient.h
danston: The condition with `inf` comes from the fact that when we convert `l` and `u` to `Type`, which is is `Quotient<cpp_int>`, and `l` or `u` are `inf`, this `Type` construction fails because `cpp_int` does not accept the `inf`. The condition with 0 is similar. In some cases, when `l` is 0, `u` is a small value like `min_denorm` and again `Type` construction fails on this value, so I skip all cases with 0 to avoid this.

Number_types/include/CGAL/Quotient.h
danston: Removed.

Number_types/include/CGAL/Quotient.h
danston: I did that in the other implementation but did not in this one. Sorry. I think now it should be ok.

Number_types/include/CGAL/Quotient.h
danston: Fixed.

Number_types/include/CGAL/Quotient.h
mglisse: I see, so l=-inf or u=+inf is ok and we don't want to convert in that case. I am not convinced for 0 though. We should be able to convert from denorm_min, or if we cannot then we may need to add some special code to handle that case.

danston: For `denorm_min` it fails [here](https://github.com/danston/cgal/blob/2e976c4048101036cf59c0235dde36b5c129bef8/Number_types/include/CGAL/double.h#L206).

mglisse: I opened #5982 about that.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+
+        std::pair<double, double> get_interval_using_cpp_rational_stable( const Type& x ) const {
+
+          // std::cout << "n: " << x.numerator()   << std::endl;
+          // std::cout << "d: " << x.denominator() << std::endl;
+
+          const double inf = std::numeric_limits<double>::infinity();
+          const double xn = x.num.template convert_to<double>();
+          const double xd = x.den.template convert_to<double>();
+          if (xn == inf || xd == inf) {
```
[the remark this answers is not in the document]
danston: Yes, that makes sense. I think I do it better now.

Number_types/include/CGAL/Quotient.h
```diff
@@ -686,11 +707,547 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+              // TODO: Is it fast enough? Seems so because this conversion happens
+              // only a few times during the run, at least for NEF.
+              boost::multiprecision::cpp_rational rat;
+              CGAL_assertion(x.den != 0);
+              if (x.den < 0) { // we do it due to the bug in boost
+                rat = boost::multiprecision::cpp_rational(-x.num, -x.den);
+              } else {
+                rat = boost::multiprecision::cpp_rational( x.num,  x.den);
+              }
+
```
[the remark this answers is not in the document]
danston: True. Fixed.

Number_types/include/CGAL/Quotient.h
```diff
@@ -989,48 +989,54 @@ template < class NT > class Real_embeddable_traits_quotient_base< Quotient<NT> >
...
+          } else {
+            CGAL_assertion(CGAL::is_positive(x.den));
+            rat = boost::multiprecision::cpp_rational( x.num,  x.den);
+          }
+
+          double l, u;
+          std::tie(l, u) = to_interval(rat);
+          const double inf = std::numeric_limits<double>::infinity();
+
+          if (l == +inf) {
```
mglisse, line 1009: Uh? It is a bug in to_interval (and should be fixed there) if it returns +inf as a lower bound.

danston: What I do currently is: I call `to_interval()` on `Type` that can be either `cpp_int` or `cpp_rational`. Inside `to_interval()`, I try to convert this `Type` number to double using `static_cast<double>(x)` and if it returns `+-inf`, I quit early and return `std::make_pair(+-inf, +-inf)`. This way I can check if I am in the limit case and return the proper interval. If I do not do that, then the call to `const int cmp = x.compare(i);` in the next lines of `to_interval()` fails because `i` is `+-inf`. I am not sure if this is a valid way to check it though.

mglisse: to_interval is an official interface, and (+inf,+inf) is not a valid return value for it. If casting to double gives +inf, the correct return value is (dbl_max, inf). You seem to do that anyway, just not in the right place.

Number_types/include/CGAL/Quotient.h
```diff
@@ -44,7 +46,26 @@ namespace CGAL {
 // This function is not documented as a number type requirement for now.
 template < typename NT >
 inline void
-simplify_quotient(NT &, NT &) {}
+simplify_quotient(NT & a, NT & b) {
+
+// TODO:
+// - move it to the boost_mp.h
+// - can we use gcd only sometimes to save time?
+#if defined(CGAL_USE_CPP_INT) && true
```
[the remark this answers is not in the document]
danston: I changed that now.

Number_types/include/CGAL/Quotient.h
danston: This part of the code is changed quite a lot now and it seems to guarantee the tight interval so I think it is fixed.

Number_types/include/CGAL/Quotient.h
danston: True. I changed this.

Number_types/include/CGAL/Exact_integer.h
```diff
@@ -50,30 +50,22 @@ typedef unspecified_type Exact_integer;
...
+#endif
+#elif defined(CGAL_USE_LEDA)
 typedef leda_integer Exact_integer;
-
-#elif CGAL_USE_CORE
-
+#elif defined(CGAL_USE_BOOST_MP)
+typedef BOOST_cpp_arithmetic_kernel::Integer Exact_integer;
+#elif defined(CGAL_USE_CORE)
 typedef CORE::BigInt Exact_integer;
```
mglisse, line 66: As someone mentioned recently, this case is probably dead code (cannot use Core without GMP), but cleaning that up is unrelated to this patch. I wonder if we should add `#else #error "cannot determine a bigint type"` or if continuing is on purpose, in case we don't actually use Exact_integer later.

Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
```diff
@@ -54,19 +54,24 @@ namespace CGAL { namespace internal {
...
+#if defined(CGAL_USE_BOOST_MP)
+{ typedef BOOST_gmp_arithmetic_kernel::Rational Type; };
+#else
 { typedef Gmpq Type; };
-# endif
+#endif
 #elif defined(CGAL_USE_LEDA)
 { typedef leda_rational Type; };
-#elif 0 && defined(CGAL_USE_BOOST_MP)
+#elif defined(CGAL_USE_BOOST_MP)
```
mglisse, line 68: So we enable it even for old boost? It wasn't really supported, so I guess that's ok.

Number_types/include/CGAL/boost_mp.h
```diff
@@ -223,7 +242,163 @@ struct RET_boost_mp <NT, boost::mpl::int_<boost::multiprecision::number_kind_int
...
+
+          const Type lb(l), ub(u);
+          CGAL_assertion(lb <= x);
+          CGAL_assertion(ub >= x);
+          const bool are_bounds_respected = (lb <= x && x <= ub);
+          return are_bounds_tight && are_bounds_respected;
+        }
+
+        Interval_nt<false>
+        my_ldexp( const Interval_nt<false>& intv, const int e ) const {
```
mglisse, line 276: Why are we duplicating this function? (several times!)

Number_types/include/CGAL/boost_mp.h
```diff
@@ -607,6 +782,325 @@ namespace internal {
...
+          if (CGAL::is_zero(x.num)) { // return [0.0, 0.0]
+            CGAL_assertion(are_bounds_correct(l, u, input));
+            return std::make_pair(l, u);
+          }
+          CGAL_assertion(!CGAL::is_zero(x.num));
+
+          // Handle signs.
+          bool change_sign = false;
+          const bool is_num_pos = CGAL::is_positive(x.num);
+          const bool is_den_pos = CGAL::is_positive(x.den);
```
mglisse, line 1028: I think you mentioned that a negative denominator can only appear here because of a bug in a specific version of boost? If so, a comment and link to the bug would be helpful, I'd like to replace this with an assertion at some point in the future, if possible.

Number_types/include/CGAL/boost_mp.h
```diff
@@ -607,6 +782,325 @@ namespace internal {
 #endif
 } // namespace internal
 
+#ifdef CGAL_USE_BOOST_MP
+
+template < >
+inline void
+simplify_quotient(boost::multiprecision::cpp_int & a, boost::multiprecision::cpp_int & b) {
```
mglisse, line 789: It is more common to overload than specialize, but here it seems ok. On the other hand, it would be possible to make it a bit more general so it also works with `number<cpp_int_backend<512>>` for instance.

Number_types/include/CGAL/boost_mp.h
```diff
@@ -192,24 +194,41 @@ struct RET_boost_mp_base
 
     struct To_interval
         : public CGAL::cpp98::unary_function< Type, std::pair< double, double > > {
+
         std::pair<double, double>
         operator()(const Type& x) const {
```
mglisse, line 199: Do we get an efficient `to_interval(cpp_int)` (not relying on `convert_to<double>`+`compare`+`nextafter`) with this patch? You have probably written one somewhere, but I don't see it. On the other hand, this is less important than rationals.

mglisse (commented):
(I haven't read the patch yet, these are just quick comments)

Number_types/include/CGAL/Exact_integer.h
danston: I have added `#else #error ....`.

Number_types/include/CGAL/boost_mp.h
danston: True. I refactored the duplicated code.

Number_types/include/CGAL/boost_mp.h
danston: Actually, it was not here. It was when constructing a cpp rational. Here, we handle the sign of a general rational number that can be either negative or positive.

danston: https://github.com/CGAL/cgal/pull/5937#discussion_r703599730

Number_types/include/CGAL/boost_mp.h
danston: I changed it to the overload version.

Number_types/include/CGAL/boost_mp.h
danston: Actually no. I have it only for rational types: `cpp_rational` and `Quotient<cpp_int>`.

Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
mglisse: Some future version of boost will have a fast cpp_rational, and it makes perfect sense to enable it there. For users who only have an older version of boost, cpp_rational is rather slow, and it isn't clear if that is a better choice than the next option: `Quotient<MP_Float>` (bugs can also be a factor). But since this wasn't a supported case before, I am ok with your simple code that just always prefers boost.

Number_types/include/CGAL/boost_mp.h
mglisse: We already have a function `ldexp` in Interval_nt.h. If you don't use it, you need to add a comment before `my_ldexp` explaining why this alternate function is useful, how it differs from the original function. Does it use positivity to skip some computations and become faster? Does it avoid setting the rounding mode? Does it fix some bug in the original function (why not fix it there then?)?
my_ldexp contains `double` multiplications without protection. Are we sure these multiplications always do the right thing whatever the rounding mode? Or that the rounding mode will always be the right one (an assertion would help document that)?

Number_types/include/CGAL/boost_mp.h
```diff
@@ -223,7 +396,19 @@ struct RET_boost_mp <NT, boost::mpl::int_<boost::multiprecision::number_kind_int
...
+    : RET_boost_mp_base <NT> {
+    typedef NT Type;
+    struct To_interval
+        : public CGAL::cpp98::unary_function< Type, std::pair< double, double > > {
+
+        std::pair<double, double> operator()( const Type& x ) const {
+
+          const auto& xnum = boost::multiprecision::numerator(x);
+          const auto& xden = boost::multiprecision::denominator(x);
+          return Boost_MP_internal::to_interval(x, xnum, xden);
```
mglisse, line 408: Since `Boost_MP_internal::to_interval` takes its last 2 arguments by value, it would be better to write this body as a single line (`to_interval(x,numerator(x),denominator(x))`), for types where `numerator` returns a value and not a reference.

Number_types/include/CGAL/boost_mp.h
mglisse: My question was whether boost has an invariant that `denominator` always returns a positive number. But it doesn't seem documented, although some backends do provide this invariant, so I guess it would be ok for a rational backend not to provide it, and you are right to check the sign.

Number_types/include/CGAL/boost_mp.h
mglisse: `get_interval_exp` is probably just one `ldexp` away, if you feel like adding it.

NewKernel_d/test/NewKernel_d/Epick_d.cpp
```diff
@@ -1,6 +1,7 @@
 #include <boost/config.hpp>
 
 //#define BOOST_RESULT_OF_USE_DECLTYPE 1
+#define CGAL_DO_NOT_USE_BOOST_MP // this test does not compile with cpp_rational
```
danston, line 4: @mglisse Could you please have a look at this [compilation error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-83/NewKernel_d/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz) if you have time? It is happenning only when `cpp_rational` type from `boost mp` is selected. For other types including `mpq_rational` it works. I tried to fix it but I cannot find a good solution because I do not fully understand what exactly causes this error and what is missing inside `cpp_rational` that is not missing inside `mpq_rational`. There is only one error of this type and it is inside `NewKernel_d`. To continue testing this PR and avoiding errors in the github workflow, I switch off `Epick_d.cpp` test for the `cpp_rational` type, but it is an ugly solution and I would like to avoid it. Thank you in advance!

mglisse: I'll take a look, but I can't make any promise on how soon...
Random try, since this involves Eigen, does `#include <boost/multiprecision/eigen.hpp>` have any effect?

danston: Just tried. Nope. I will also continue searching for better solutions.

mglisse: https://github.com/boostorg/multiprecision/issues/393
(it is a bit strange to see `-DCGAL_USE_GMPXX=1` in the compiler arguments for something that ends up using boost multiprecision)
This error only seems to appear with very recent versions of Eigen.

danston: Thanks for having a look into that! Regarding compiler arguments, this is a temporary hack that I will remove. It is there in order to test this boost type and `to_interval` implementation in the regular CGAL test suite. Once tested, I will clean it up. This is the unchecked task in the description of this PR.

lrineau: @mglisse:
> (it is a bit strange to see `-DCGAL_USE_GMPXX=1` in the compiler arguments for something that ends up using boost multiprecision)

The definition of `CGAL_USE_GMPXX` is triggered by our CMake scripts. So far, this PR has not touched anything in CMake, as far as I know: you patch tweaks `Number_types/include/CGAL/Exact_integer.h`, reversing the logic so that `CGAL_USE_BOOST_MP` has a precedence over `CGAL_USE_GMPXX`.

Let us remember that `CGAL_USE_BOOST_MP` is defined that way:
```c++
#if !defined CGAL_DO_NOT_USE_BOOST_MP && \
    (!defined _MSC_VER || BOOST_VERSION >= 107000)
#define CGAL_USE_BOOST_MP 1
```
https://github.com/danston/cgal/blob/d8da3502bc8feb9a4bfbcf3bc369bfd77f53e8b0/Number_types/include/CGAL/boost_mp.h#L17-L24

and then in this PR you have add in `Number_types/include/CGAL/Exact_integer.h`:
```c++
#if BOOST_VERSION > 107700 && defined(CGAL_USE_BOOST_MP)
// TODO: That is used for testing, it must be removed when merging into master.
{ typedef BOOST_cpp_arithmetic_kernel::Rational Type; };
#else // BOOST_VERSION <= 107700
#ifdef CGAL_USE_GMPXX
{ typedef mpq_class Type; };
#elif defined(CGAL_USE_GMP)
```
https://github.com/danston/cgal/blob/b2f6c0c5df43032b8c8ea0363048ce257e14d0c0/Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h#L58-L64

So with the current patch, `CGAL_USE_GMPXX` is only considered if `CGAL_DO_NOT_USE_BOOST_MP` was defined by the user, or `BOOST_VERSION <= 107700`.

That shows the big issue with the naming of preprocessor macros `CGAL_USE_<something>`: they are used by CMake to instruct the C++ code that `<something>` was found by CMake and *can be used*. The naming is not right. And we are not consistent.
  - for TBB, that macro is named `CGAL_LINKED_WITH_TBB`,
  - for Eigen, it is named `CGAL_EIGEN3_ENABLED`,
  - and for GMPXX, it is `CGAL_USE_GMPXX`...
  
We should choose something and stick with it.

sloriot:
Test failing most probably due to this branch:
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Barycentric_coordinates_2/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Barycentric_coordinates_2/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Triangulation/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Triangulation/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Triangulation_Examples/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Triangulation_Examples/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz
- https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.4-Ic-109/Polygon_mesh_processing_Examples/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz

NewKernel_d/test/NewKernel_d/Epick_d.cpp
sloriot: This line generated a warning [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-109/NewKernel_d/TestReport_cgaltester_x86-64_Darwin-18.5_Apple-clang-default_Release.gz)

sloriot:
While trying to debug a failing case of the testsuite, we hit the following case with Andreas:
https://gist.github.com/sloriot/e0b97c3b465ab826d982ee16cda20527

It runs fine on my machine (`g++ (Debian 10.2.1-6) 10.2.1 20210110`) but not on Andreas' (MSVC something).
The issue is that the interval is not tight (inf and sup are separated by more than one ulp).

We are in the case of an integral division than is not exact after the shift so we call `get_1ulp_interval()` to get the output interval.
This function calls the `ldexp` function of `Interval_nt` with `e=-1714`. on my machine `std::ldexp(1.0, e)=4.9406564584124654e-324` (because rounding mode is set to `UPWARD`, it is 0 with `TO_NEAREST`). On Andreas' machine it is 0 even with `UPWARD`.

```
template < bool b >
Interval_nt<b>
ldexp(const Interval_nt<b> &i, int e)
{
  double d = 1.0;
  double scale = std::ldexp(d, e);
  Interval_nt<b> scale_interval (
                      CGAL_NTS is_finite(scale) ? scale : CGAL_IA_MAX_DOUBLE,
                      scale == 0 ? CGAL_IA_MIN_DOUBLE : scale);
  return i * scale_interval;
}
  ```
What happens when `scale=0` is that the `scale_interval` is no longer a singleton and since `i` is not either  (since we don't have an exact division) then the interval produced cannot be of length 1 ulp.

I have two questions:
- if `scale` is 0, we could call `std::ldexp` twice on `-i.inf()` and `i.sup()` and we would increase the change of getting a tight interval but it would still be without guarantees, right?
- I have a doubt about the fact that we set the rounding mode to `UPWARD` before calling `std::ldexp`. To me it seems that we would need two calls with different rounding modes for getting valid bounds for inf/sup. What am I missing?

@lrineau @mglisse any opinion?

**Comment by mglisse** [comment]
> on my machine `std::ldexp(1.0, e)=4.9406564584124654e-324` (because rounding mode is set to `UPWARD`, it is 0 with `TO_NEAREST`). On Andreas' machine it is 0 even with `UPWARD`.

I see people complaining about msvc not respecting the rounding mode in ldexp 10 years ago (https://stackoverflow.com/q/32150888/1918193), I see one related (?) neglected bug report (https://developercommunity.visualstudio.com/t/ldexp-vs-muliplication-by-a-power-of-2/465720) and when I try it with vs2019 (16.11.9) `ldexp(1.,-10000)` indeed gives 0 even when rounding up. So we should consider ldexp broken on windows in all cases where the result is not exact. On the other hand, what we are doing on linux seems wrong as well, if scale is rounded up, then it isn't a valid lower bound for scale_interval (back to #6004).

> * if `scale` is 0, we could call `std::ldexp` twice on `-i.inf()` and `i.sup()` and we would increase the change of getting a tight interval but it would still be without guarantees, right?

What cases are you thinking of? I don't immediately see a case where a tight interval becomes not tight.

> * I have a doubt about the fact that we set the rounding mode to `UPWARD` before calling `std::ldexp`. To me it seems that we would need two calls with different rounding modes for getting valid bounds for inf/sup. What am I missing?

#6004 , the function is broken (there is a reference to that issue somewhere in a comment in the patch).

**Comment by afabri** [comment]
> * I have a doubt about the fact that we set the rounding mode to `UPWARD` before calling `std::ldexp`. To me it seems that we would need two calls with different rounding modes for getting valid bounds for inf/sup. What am I missing?

Aren't you missing that `Interval_nt` stores `( -inf, sup)` ?

File: Number_types/include/CGAL/boost_mp.h [hunk 33]
```diff
@@ -139,6 +141,258 @@ struct Algebraic_structure_traits<boost::multiprecision::detail::expression<T1,T
...
+    CGAL_assertion(intv.inf() > 0.0);
+    CGAL_assertion(intv.sup() > 0.0);
+    typedef std::numeric_limits<double> limits;
+
+    if (e < std::numeric_limits<double>::min_exponent)
+    {
+      if ( e+std::numeric_limits<double>::digits < std::numeric_limits<double>::min_exponent)
+        return CGAL::Interval_nt<false>(0, (limits::min)());
+      // following calls to ldexp call are exact (e+digits is larger or equal to min_exponent) is less than min_exponent
+      return CGAL::Interval_nt<false>(std::ldexp(intv.inf(), e), std::ldexp(intv.sup(), e));
```
lrineau, line 157:
```suggestion
      // following calls to ldexp call are exact (e+digits is larger or equal to min_exponent)
      return CGAL::Interval_nt<false>(std::ldexp(intv.inf(), e), std::ldexp(intv.sup(), e));
```

NewKernel_d/test/NewKernel_d/Epick_d.cpp
lrineau: @sloriot I think this conversion can be closed.

sloriot:
There is an error popping on some plateform only in the `Triangulation` package:
```
Triangulation_Examples/TestReport_gimeno_ArchLinux-clang.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
Triangulation_Examples/TestReport_gimeno_ArchLinux-clang-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
Triangulation_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:10: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:30:187: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation_Examples/TestReport_lrineau_ArchLinux.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:10: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation_Examples/TestReport_lrineau_ArchLinux.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:30:187: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation_Examples/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz:/usr/local/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
Triangulation/TestReport_gimeno_ArchLinux-clang.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
Triangulation/TestReport_gimeno_ArchLinux-clang-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
Triangulation/TestReport_lrineau_ArchLinux-CXX17-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:10: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation/TestReport_lrineau_ArchLinux-CXX17-Release.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:30:187: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation/TestReport_lrineau_ArchLinux.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:29:10: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation/TestReport_lrineau_ArchLinux.gz:/usr/include/boost/multiprecision/traits/is_byte_container.hpp:30:187: error: no type named ‘value_type’ in ‘struct std::iterator_traits<void>’
Triangulation/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz:/usr/local/include/boost/multiprecision/traits/is_byte_container.hpp:29:116: error: no type named 'value_type' in 'std::iterator_traits<void>'
```

sloriot:
As I don't have it locally, maybe it has to do with the version of boost on the failing plateforms.

mglisse:
That error message rings a bell. Does it match with platforms that have an older version of boost?

sloriot:
@mglisse, that's also my guess. Jane is currently updating it on one platform to check.

lrineau:
> That error message rings a bell. Does it match with platforms that have an older version of boost?
> [boostorg/multiprecision#393](https://github.com/boostorg/multiprecision/issues/393) maybe?

By "older" you mean any Boost release. That issue boostorg/multiprecision#393 (reported [Nov 20, 2021](https://github.com/boostorg/multiprecision/issues/393#issue-1059106258)) was fixed by the PR boostorg/multiprecision#398, merged [Dec 5, 2021](https://github.com/boostorg/multiprecision/pull/398#event-5716540111). All that is really new, and the fix will only be in next release Boost-1.79, planned mid-April 2022.

Number_types/include/CGAL/boost_mp.h
```diff
@@ -139,6 +141,258 @@ struct Algebraic_structure_traits<boost::multiprecision::detail::expression<T1,T
 
 // Real_embeddable_traits
 
+namespace Boost_MP_internal {
+
+  inline
+  Interval_nt<false> shift_positive_interval( const Interval_nt<false>& intv, const int e ) {
```
mglisse, line 147: Could you document the preconditions for this function? If we only assume that the interval is positive, the code seems wrong. With more assumptions maybe... :man_shrugging:

Number_types/include/CGAL/boost_mp.h
```diff
@@ -139,6 +141,259 @@ struct Algebraic_structure_traits<boost::multiprecision::detail::expression<T1,T
...
+    CGAL_assertion(intv.inf() > 0.0);
+    CGAL_assertion(intv.sup() > 0.0);
+    typedef std::numeric_limits<double> limits;
+
+    if (e < std::numeric_limits<double>::min_exponent)
+    {
+      if ( e+std::numeric_limits<double>::digits < std::numeric_limits<double>::min_exponent)
+        return CGAL::Interval_nt<false>(0, (limits::min)());
+      // following calls to ldexp call are exact (e+digits is larger or equal to min_exponent)
+      return CGAL::Interval_nt<false>(std::ldexp(intv.inf(), e), std::ldexp(intv.sup(), e));
```
mglisse, line 158: If those are exact, then so is `ldexp(1,e)`, why do you need this special case?

Number_types/include/CGAL/boost_mp.h
```diff
@@ -139,6 +141,259 @@ struct Algebraic_structure_traits<boost::multiprecision::detail::expression<T1,T
...
+namespace Boost_MP_internal {
+
+  // here we know that `intv` contains int64 numbers such that their msb is std::numeric_limits<double>::digits-1
+  inline
+  Interval_nt<false> shift_positive_interval( const Interval_nt<false>& intv, const int e ) {
+    CGAL_assertion(intv.inf() > 0.0);
+    CGAL_assertion(intv.sup() > 0.0);
+    typedef std::numeric_limits<double> limits;
+
+    if (e < std::numeric_limits<double>::min_exponent)
```
mglisse, line 153:
```suggestion
    if (e < limits::min_exponent)
```

Number_types/include/CGAL/boost_mp.h
sloriot: not if `e < std::numeric_limits<double>::min_exponent`

Number_types/include/CGAL/boost_mp.h
```diff
@@ -139,6 +141,259 @@ struct Algebraic_structure_traits<boost::multiprecision::detail::expression<T1,T
...
+    {
+      if ( e+std::numeric_limits<double>::digits < std::numeric_limits<double>::min_exponent)
+        return CGAL::Interval_nt<false>(0, (limits::min)());
```
mglisse, line 156: I thought you had an assertion somewhere that intervals are tight. Is that assertion gone?

Number_types/include/CGAL/boost_mp.h
mglisse: Er, there are denormal values below that, we can use them...

mglisse: But mostly, the comment says that `ldexp(intv.inf(),e)` is exact without having looked at `intv`, so the comment really seems to imply that `ldexp(1,e)` is exact, since 1 is a possible value for `intv.inf()`. No?

sloriot: The thing is I was suspecting that `ldexp` is not exact as soon as you switch to denormals, especially with MSVC. 
About `intv.inf()`, it cannot be 1 as `msb()` is 52. However, it might be that `msb()` of `intv.sup()` is 53 since it is `inf()+1` in case the division is not exact.

Number_types/include/CGAL/boost_mp.h
sloriot: It is in `are_bounds_correct()` and it is tight only if we did not go into the special cases around zero or at infinity.

Number_types/include/CGAL/boost_mp.h
mglisse: I thought the problem with MSVC was the rounding mode, I would expect it to handle denormals correctly in the exact case, but I didn't try.

sloriot:
I wanted to test the latest optimization (47af8d4491f2c4b86a05d42bd0d28e26d0d654e9) we did for the case when input are double. So I created a set of random point from `Simple_cartesian<double>` and converted them to `EPECK` and built a 3D Delaunay triangulation. There was actually no runtime difference as the `to_interval()` is not called in such cases. Indeed interval points are constructed directly from the input double. I don't see any situation where we would have a `to_interval` numbers constructed from double so I've disabled it for now. I also realized that the `to_interval()` function will mainly be called only in the case of cascaded constructions when `exact()` is called and the interval approximation will get refined from the exact value.

mglisse:
I tried the program that Efi sent on 2019-1-22, which IIRC was sensitive to having to_interval return points, but it doesn't spend much time in to_interval.
Maybe `Filtered_kernel<Simple_cartesian<Exact_rational>>` (a fairly natural kernel for people doing parallelism without all the atomic stuff, and I know GF spent quite a bit of time on a variant that caches the interval #4495) would be a good way to spend a lot of time in to_interval. For instance, if I replace Epeck with this kernel in Efi's program (what I had handy, but I guess building a triangulation should work as well), 40% of the time is spent in to_interval, and the optimization reduces the running time from 1.5 to 1.28 (< 30% of the running time in to_interval).

sloriot:
There is a runtime error on Fedora-32 [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-65/Number_types/TestReport_lrineau_Fedora-32-Release.gz)

mglisse:
> There is a runtime error on Fedora-32 [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-65/Number_types/TestReport_lrineau_Fedora-32-Release.gz)

I think we said we were going to try the strict test for this one, but if it didn't work we would go back to: d>=DBL_MAX on line 811, and r.first <= d on the next line. Does that match your memory?
I have no idea how the 2 platforms where this failed differ from the others, but I don't really want to look into it unless there is a bigger issue.

Number_types/test/Number_types/to_interval_test_boost.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+  #endif
+
+  std::cout << "---SUCCESS ALL---" << std::endl;
+  std::cout << std::endl;
+}
+
+void test_shift_positive() {
+  {
+    double d = (1L << 53) - 1;
```
mglisse, line 799:
```suggestion
    double d = (1LL << 53) - 1;
```
(same for the other tests below)
On platforms with 32-bit long it won't fit otherwise.

Number_types/test/Number_types/to_interval_test_boost.cpp
```diff
@@ -796,7 +796,7 @@ void test_to_interval_tight_integer() {
 
 void test_shift_positive() {
   {
-    double d = (1L << 53) - 1;
+    double d = (1LL << 53) - 1;
```
mglisse, line 799: You also need to change the others below... (there are warnings if you compile with -m32)

sloriot:

If I look at the table, it seems that boost 1.79 is not tested, only 1.80 is one 3 plateforms.

PR #5944, opened 2021-08-31
Add polygon set drawing functionality.

dpapavas:
## Summary of Changes

This is not necessarily a pull request; I was wondering why there's no support for drawing polygon sets, as it seems to be as useful as it would be easy to implement.  This PR is an (unfinished and unpolished) implementation attempt I made which seems to work fine (_seems_ being the operative word here, as I don't pretend to have studied in detail how the underlying polygon-with-holes drawing functionality works, but I'm fairly confident this extension should work).

Let me know if you'd like me to finish and polish this into a proper PR, or feel free to do it yourself.

## Release Management

* Affected package(s): Polygon
* Feature/Small Feature (if any): Polygon set drawing functionality


Polygon/include/CGAL/draw_polygon_set_2.h
```diff
@@ -0,0 +1,61 @@
+#ifndef CGAL_DRAW_POLYGON_SET_2_H
```
afabri, line 1: Can you copy paste the license header from [here ](https://github.com/CGAL/cgal/blob/908f5811c0f1b06630716eddfcd219b5d4f1b75c/Polygon/include/CGAL/draw_polygon_with_holes_2.h#L1)

dpapavas: Done.  I've kept the original author's name, as virtually none of the code is mine (I only did some reshuffling).  Also note (as you're probably already aware) that more stuff is missing (at least the documentation-related stuff is).

afabri: I guess it also only works for the case that the edges of the polygons are straight line segments,

gdamiand:
Thanks @dpapavas for this contribution. Indeed this viewer is simple to write based on the viewer of polygon with holes. 
The reason why it is not yet developed is just a matter of time... Thus you are welcome to finish this PR!
You can also add a small example (like [this one](https://github.com/CGAL/cgal/blob/master/Polygon/examples/Polygon/draw_polygon.cpp)).

Thanks again!

Polygon/include/CGAL/draw_polygon_set_2.h
gdamiand: @afabri The same restriction applies for polygon and polygon with holes viewers. But does CGAL polygons allow non straight line segments?

Polygon/include/CGAL/draw_polygon_with_holes_2.h
```diff
@@ -122,7 +129,7 @@ class SimplePolygonWithHoles2ViewerQt : public Basic_viewer_qt
   }
 
 protected:
-  const P2& p2;
+  ;
```
gdamiand, line 132: remove

Polygon/include/CGAL/draw_polygon_set_2.h
```diff
@@ -0,0 +1,77 @@
...
+public:
+  SimplePolygonSet2ViewerQt(QWidget* parent, const PS2& aps2,
+                            const char* title="Basic Polygon_set_2 Viewer") :
+    Base(parent, title)
+  {
+    const int n = aps2.number_of_polygons_with_holes();
+    typename PS2::Polygon_with_holes_2 polygons[n];
+    aps2.polygons_with_holes(polygons);
+
+    for (typename PS2::Polygon_with_holes_2 P: polygons) {
```
gdamiand, line 43: Question: do we need to use typename PS2::Polygon_with_holes_2& P?
I am not sure if typename PS2::Polygon_with_holes_2 P will copy the polygons or not.

Polygon/include/CGAL/draw_polygon_set_2.h
dpapavas: >But does CGAL polygons allow non straight line segments?

I probably misunderstand the question, but CGAL supports polygons bounded by various weakly x-monotone curves (circles, conics, beziers, etc.).  Indeed, the current functionality only supports drawing "regular" polygons with linear edges and indeed being able to draw curved polygons would be useful, but this would require approximating the curves with polylines, which is a bit more involved.

I have working code to do that for circle-segment polygons and could try to adapt it into a viewer, if I can find some time and if you think it'd be useful.

Polygon/include/CGAL/draw_polygon_set_2.h
dpapavas: You're right of course; this should be a reference, as it will copy the polygn needlessly as it is.

Polygon/include/CGAL/draw_polygon_with_holes_2.h
dpapavas: Sorry about that.  I probably moved the declaration to the argument list of `compute_elements` and left the rest there. :grin:

dpapavas:
I've added an example, which basically draws a pair of nested rectangles, so that there's a polygon with a hole, one without and a polygon set.  The result looks like this.

image

I've also tried to add some documentation, so that there's at least a reference to drawing polygon sets in the reference manual.  I added it to the `Boolean_set_operations_2` package, where I've also moved the drawing code, so that it's in the same place as `Polygon_set_2`, but I'm having trouble getting the description of `void draw(const PS& aps)`, which is incuded in the source code, as it is for the other draw functions, to show up in the relevant section of the manual.  I only get `#include <CGAL/draw_polygon_set_2.h>` there, from `PackageDescription.txt`.

The other thing I'm not very happy with, is that the new section "Draw a 2D Polygon Set" ends up right at the top, although it's certainly not that important and would be better placed at the end, after the section on "Polygon_set_2".

I haven't studied the whole setup in detail, as this is a pretty complex codebase, but I've tried to read through the relevant sections in the wiki pages on adding and building documentation.  I'm going to have another look, but, in the meantime, if you notice anything obvious (to you) that I'm missing, or if you have any advice, please let me know.

dpapavas:
I got the documentation for the draw function to show up (it was an obvious error in specifying the proper group) and I've hopefully fixed the check I was failing (I had forgotten to `git add` a header).  The documentation is now as complete as for the other draw functions, and although it still show up at the top, it's the same for the other draw functions too, so I'll leave it as is for now.

Let me know if you have a problem with it, or with any other aspect of the PR.

sloriot:
Warnings:
```
/srv/CGAL/www/Members/testsuite/CGAL-5.4-Ic-58/Boolean_set_operations_2_Examples/TestReport_Sosno_MSVC-2022-Preview-Release.gz:    26>C:\cgal_root\CGAL-5.4-Ic-58\include\CGAL/draw_polygon_set_2.h(56,1): warning C4267: 'initializing': conversion from 'size_t' to 'const int', possible loss of data [C:\cgal_root\CGAL-5.4-Ic-58\cmake\platforms\MSVC-2022-Preview-Release\test\Boolean_set_operations_2_Examples\draw_polygon_set.vcxproj]
/srv/CGAL/www/Members/testsuite/CGAL-5.4-Ic-58/Boolean_set_operations_2_Examples/TestReport_Sosno_MSVC-2022-Preview-Release.gz:    26>C:\cgal_root\CGAL-5.4-Ic-58\include\CGAL/draw_polygon_set_2.h(56,1): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data [C:\cgal_root\CGAL-5.4-Ic-58\cmake\platforms\MSVC-2022-Preview-Release\test\Boolean_set_operations_2_Examples\draw_polygon_set.vcxproj]
/srv/CGAL/www/Members/testsuite/CGAL-5.4-Ic-58/Boolean_set_operations_2_Examples/TestReport_Sosno_MSVC-2022-Preview-Release.gz:         C:\cgal_root\CGAL-5.4-Ic-58\include\CGAL/draw_polygon_set_2.h(56,1): warning C4267: 'initializing': conversion from 'size_t' to 'const int', possible loss of data [C:\cgal_root\CGAL-5.4-Ic-58\cmake\platforms\MSVC-2022-Preview-Release\test\Boolean_set_operations_2_Examples\draw_polygon_set.vcxproj]
/srv/CGAL/www/Members/testsuite/CGAL-5.4-Ic-58/Boolean_set_operations_2_Examples/TestReport_Sosno_MSVC-2022-Preview-Release.gz:         C:\cgal_root\CGAL-5.4-Ic-58\include\CGAL/draw_polygon_set_2.h(56,1): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data [C:\cgal_root\CGAL-5.4-Ic-58\cmake\platforms\MSVC-2022-Preview-Release\test\Boolean_set_operations_2_Examples\draw_polygon_set.vcxproj]
```

Boolean_set_operations_2/include/CGAL/draw_polygon_set_2.h
```diff
@@ -0,0 +1,94 @@
...
+  public SimplePolygonWithHoles2ViewerQt<typename PS2::Polygon_with_holes_2>
+{
+  typedef SimplePolygonWithHoles2ViewerQt<typename PS2::Polygon_with_holes_2> Base;
+
+public:
+  SimplePolygonSet2ViewerQt(QWidget* parent, const PS2& aps2,
+                            const char* title="Basic Polygon_set_2 Viewer") :
+    Base(parent, title)
+  {
+    const int n = aps2.number_of_polygons_with_holes();
```
sloriot, line 56:
```suggestion
    const std::size_t n = aps2.number_of_polygons_with_holes();
```

Boolean_set_operations_2/include/CGAL/draw_polygon_set_2.h
```diff
@@ -0,0 +1,94 @@
...
+    typename PS2::Polygon_with_holes_2 polygons[n];
```
sloriot, line 57: Except if I missed something n is not known at compile time, so you should replace the array with a `std::vector`.

```
         C:\cgal_root\CGAL-5.4-Ic-58\include\CGAL/draw_polygon_set_2.h(57,49): error C2131: expression did not evaluate to a constant [C:\cgal_root\CGAL-5.4-Ic-58\cmake\platforms\MSVC-2022-Preview-Release\test\Boolean_set_operations_2_Examples\draw_polygon_set.vcxproj]
```

dpapavas: No, you're not missing anything.  I naively assumed that variable length arrays (a feature introduced to C in C99) were also supported in C++.  They are not, but GCC was being helpful in allowing their use.  I've instructed him to be sufficiently pedantic so as to avoid such problems in the future.

On the other hand GCC does not seem to mind the implicit `size_t` to `int` conversion even with `-Wextra` and `-Wpedantic`, but I've switched to using a `vector`, so the offending line was removed anyway.

lrineau: The variable length arrays is common known as a mistake that should have never been in C99. It has never entered the C++ norm. See for example that StackOverflow answer, for the reasons:
https://stackoverflow.com/a/21519062/1728537

sloriot:

PR #5945, opened 2021-08-31
Improvement for named parameters

sloriot:
Allow non-copyable object to be passed as parameters

### DIFF
--- a/BGL/include/CGAL/boost/graph/Named_function_parameters.h
+++ b/BGL/include/CGAL/boost/graph/Named_function_parameters.h
@@ -16,6 +16,7 @@
 
 #include <boost/type_traits/is_same.hpp>
 #include <boost/mpl/if.hpp>
+#include <type_traits>
 
 #define CGAL_BGL_NP_TEMPLATE_PARAMETERS T, typename Tag, typename Base
 #define CGAL_BGL_NP_CLASS CGAL::Named_function_parameters<T,Tag,Base>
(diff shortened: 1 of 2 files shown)


BGL/test/BGL/test_cgal_bgl_named_params.cpp
```diff
@@ -46,7 +46,7 @@ template<class NamedParameters>
 void test_no_copyable(const NamedParameters& np)
 {
   typedef typename inp::Get_param<typename NamedParameters::base,inp::visitor_t>::type NP_type;
-  static_assert( boost::is_same<NP_type,std::reference_wrapper<const B> > ::value );
```
lrineau: ## Note

See https://en.cppreference.com/w/cpp/language/static_assert:
 - `static_assert(bool, message)` is C++11,
 - `static_assert(bool)` without the message is C++17.

BGL/include/CGAL/boost/graph/Named_function_parameters.h
```diff
@@ -316,7 +316,7 @@ const D& choose_parameter(const internal_np::Param_not_found&, const D& d)
 template <typename D>
 D choose_parameter(const internal_np::Param_not_found&, D&& d)
 {
-  return d;
+  return std::move(d);
```
lrineau, line 319:
```suggestion
  return std::forward(d);
```

sloriot:

PR #5850, opened 2021-07-13
Port Mesh_3 to oneAPI-TBB

lrineau:
## Summary of Changes

The work was already mostly done in PR #4892, but there was still two issues:
  - `tbb::parallel_do` has been removed, and `tbb::parall_for_each` was to be used instead,
  - the support for `tbb_hasher` has been removed.

This time, I have tested in a container were intel-oneapi-tbb-common-devel-2021.3.0-2021.3.0-511.noarch was installed, but not tbb version 2020.

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): fix #5845
* License and copyright ownership: maintenance by GeometryFactory

PR #5732, opened 2021-05-27
FindTBB.cmake: only search TBB in Config mode if version >= 2019 update 5

janetournois:
## Summary of Changes

Avoid the following error when TBB was already found a few lines earlier, by adding a condition `NOT TBB_FOUND` before `find_package(TBB)` :
```
CMake Error at C:/dev/tbb2019_20181203oss_win/tbb2019_20181203oss/cmake/TBBConfig.cmake:83 (add_library):
  add_library cannot create imported target "TBB::tbb" because another target
  with the same name already exists.
```

## Release Management

* Affected package(s): Polyhedron (demo)
* License and copyright ownership: unchanged

Issue #5676, opened 2021-05-05
Concurrency-in-CGAL CMakeLists.txt Error _ CMake TBB support for parallel excution not working


AnasFX:
## Issue Details
* Trying the help file in this link ( https://github.com/CGAL/cgal/wiki/Concurrency-in-CGAL ).
* Here is a part of the CMakeLists.txt file that I did : 
```cmake
# CGAL and its components
find_package(CGAL REQUIRED)
# include helper file
include(${CGAL_USE_FILE})
# Activate concurrency?
option(ACTIVATE_CONCURRENCY
       "Enable concurrency"
       ON)

if( ACTIVATE_CONCURRENCY )
  find_package( TBB REQUIRED )
  if( TBB_FOUND )
    include(${TBB_USE_FILE})
    list(APPEND CGAL_3RD_PARTY_LIBRARIES ${TBB_LIBRARIES})
  endif()
endif()
```
-------------------------------------
* Trying to CMake the example ( Polygonal_surface_reconstruction ) with TBB support gives this error :
```
CMake Error at C:/dev/CGAL-5.2/cmake/modules/FindTBB.cmake:405 (file):
  file STRINGS file "C:/dev/oneapi-tbb-2021.2.0/include/tbb/tbb_stddef.h"
  cannot be read.
Call Stack (most recent call first):
  CMakeLists.txt:20 (find_package)


CMake Deprecation Warning at C:/dev/CGAL-5.2/cmake/modules/UseTBB.cmake:8 (message):
  This file UseTBB.cmake is deprecated, and the imported target
  `CGAL::TBB_support` from CGAL_TBB_support.cmake should be used instead.
Call Stack (most recent call first):
  CMakeLists.txt:22 (include)
````
-------------------------------------
* Although I added these environment paths variables, the CMake didn't find them:
```
TBB_DLL_X64_DIR=C:\dev\oneapi-tbb-2021.2.0\redist\intel64\vc14\
TBB_DLL_X86_DIR=C:\dev\oneapi-tbb-2021.2.0\redist\ia32\vc14\
TBB_INCLUDE_DIR=C:\dev\oneapi-tbb-2021.2.0\include\
TBB_LIB_X64_DIR=C:\dev\oneapi-tbb-2021.2.0\lib\intel64\vc14\
TBB_LIB_X86_DIR=C:\dev\oneapi-tbb-2021.2.0\lib\ia32\vc14\
[...]

PR #4892, opened 2020-07-31
Mesh_3: now uses tbb:task_group instead of tbb::task (deprecated)

lrineau:
## Summary of Changes

Mesh_3, now uses `tbb:task_group` instead of `tbb::task` (deprecated). Fixes #4872.

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): fix #4872
* License and copyright ownership: maintenance by GF

PR #5948, opened 2021-08-31
FindTBB is no longer compatible with oneTBB 2021.1.1

maximecharriere:
Since oneTBB 2021.1.1, `/tbb/tbb_stddef.h` no longer exists.  
The version can be found in [/oneapi/tbb/version.h](https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/version.h)

### DIFF
--- a/Installation/cmake/modules/FindTBB.cmake
+++ b/Installation/cmake/modules/FindTBB.cmake
@@ -416,7 +416,7 @@ if(NOT TBB_VERSION)
 
  #only read the start of the file
  file(STRINGS
-      "${TBB_INCLUDE_DIR}/tbb/tbb_stddef.h"
+      "${TBB_INCLUDE_DIR}/oneapi/tbb/version.h"
       TBB_VERSION_CONTENTS
       REGEX "VERSION")
 


maximecharriere:
<tbb/tbb_config.h> no longer exists either, which causes me to get errors when building the Polyhedron demo. 

image

Due to this line :

https://github.com/CGAL/cgal/blob/6f7fd79a1e714937c9a29fd3b4e3d6de225e1922/STL_Extension/include/CGAL/thread.h#L37

maxGimeno:
I find it strange that you have this kind of problems. Since #5732 oneTBB should be found with the cmake stuff they provide. The rest of the file only serves for 2019 versions or something. 

Anyway, applying your patch would probably break compatibility with every version of TBB except oneTBB 2021

maxGimeno:
After discussions with @lrineau, it appears that CGAL is not yet compatible with oneTBB 2021, in the code. But an easy workaround for your bug is to specify the right TBB_DIR to cmake when configuring the Polyhedron Demo.
for example, on my testing machine, it is `-DTBB_DIR=C:/dev/OneAPI/tbb/2021.2.0/lib/cmake/tbb"`

lrineau:
> After discussions with @lrineau, it appears that CGAL is not yet compatible with oneTBB 2021, in the code. But an easy workaround for your bug is to specify the right TBB_DIR to cmake when configuring the Polyhedron Demo.
> for example, on my testing machine, it is `-DTBB_DIR=C:/dev/OneAPI/tbb/2021.2.0/lib/cmake/tbb"`

We are working on making the next bug-fix release CGAL-5.3.1 (not yet released) compatible with TBB. See for example PRs https://github.com/CGAL/cgal/pull/4892 and https://github.com/CGAL/cgal/pull/5850.

But we have not yet modified our documentation. TBB from oneAPI should be found by CMake using the config mode of `find_package`.


... I have a deja-vu impression actually... We already discussed that with you, @maximecharriere, in https://github.com/CGAL/cgal/issues/5676! See my answer at https://github.com/CGAL/cgal/issues/5676#issuecomment-832833724.

lrineau (changes requested):
This change would prevent anybody from using TBB-2019 or TBB 2020.

sloriot:
Replaced by #5953

PR #5958, opened 2021-09-03
Call is-termimal functor only on degree 2 nodes

sloriot:
I was surprised by the former implementation as from the doc (and my memory) the functor should be called only on degree 2 nodes. Degree 0 nodes have to be terminal.

### DIFF
--- a/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
+++ b/BGL/include/CGAL/boost/graph/split_graph_into_polylines.h
@@ -153,7 +153,7 @@ void duplicate_terminal_vertices(Graph& graph,
   {
     typename boost::graph_traits<OrigGraph>::vertex_descriptor orig_v = graph[v];
     typename boost::graph_traits<Graph>::degree_size_type deg = degree(v, graph);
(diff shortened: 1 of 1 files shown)


lrineau (approved):
[The documentation](https://doc.cgal.org/5.3/BGL/group__PkgBGLPartition.html#ga99ea6bf193f1194db5e523e713c82fb9) says:
> splits into polylines the graph g at vertices of degree greater than 2 and at vertices for which `is_terminal(v,graph)==true`. 

Without looking at the code, my understanding of this documentation is something like:
```c++
bool terminal = degree(v, graph) > 2 || is_terminal(v,graph);
```

... but I am wrong, that condition is the condition to "split". For vertices of degree<2, the vertex is already a terminal vertex.

So eventually I agree with you.

sloriot:
Errors: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-50/Polygon_mesh_processing/TestReport_lrineau_ArchLinux-CXX17-Release.gz

sloriot:
Successfully tested in CGAL-5.4-Ic-62

PR #5970, opened 2021-09-08
Minor automated message improvements

MaelRL:
## Release Management

* Affected package(s): - 
* Issue(s) solved (if any): - 
* License and copyright ownership: -

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -109,7 +109,7 @@ jobs:
             mv tmp.html index.html
             git add ${PR_NUMBER}/$ROUND index.html && git commit -q --amend -m "base commit" && git push -q -f -u origin master
           else
-            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build"
+            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build."
             exit 1
           fi
 
@@ -121,7 +121,7 @@ jobs:
             const tmp_round = "${{ steps.get_round.outputs.result }}";
             const id = tmp_round.indexOf(":");
             const round = tmp_round.substring(0,id);
-            const address = "The documentation is built. It will be available, after a few minutes, here : https://cgal.github.io/${{ steps.get_pr_number.outputs.result }}/"+round+"/Manual/index.html"
+            const address = "The documentation is built. It will be available, after a few minutes, here: https://cgal.github.io/${{ steps.get_pr_number.outputs.result }}/"+round+"/Manual/index.html"
             github.issues.createComment({
               owner: "CGAL",
               repo: "cgal",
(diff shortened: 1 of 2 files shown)


.github/workflows/filter_testsuite.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         uses: actions/github-script@v3
         with:
           script: |
-            const address = "The testsuite is lauched. results will be found, after it is done, here : https://cgal.geometryfactory.com/~cgaltest/test_suite/TESTRESULTS/index.shtml "
+            const address = "Testsuite launched. Result page: https://cgal.geometryfactory.com/~cgaltest/test_suite/TESTRESULTS/index.shtml "
             github.issues.createComment({
```
lrineau, line 65: Actually, we want to convey that the results make take a while to appear.

.github/workflows/filter_testsuite.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         uses: actions/github-script@v3
         with:
           script: |
-            const address = "The testsuite is lauched. results will be found, after it is done, here : https://cgal.geometryfactory.com/~cgaltest/test_suite/TESTRESULTS/index.shtml "
+            const address = "Testsuite launched. Result page: https://cgal.geometryfactory.com/~cgaltest/test_suite/TESTRESULTS/index.shtml "
```
MaelRL, line 64:
```suggestion
            const address = "Testsuite launched. Results will appear on the following page: https://cgal.geometryfactory.com/~cgaltest/test_suite/TESTRESULTS/index.shtml "
```

.github/workflows/build_doc.yml
```diff
@@ -109,7 +109,7 @@ jobs:
             mv tmp.html index.html
             git add ${PR_NUMBER}/$ROUND index.html && git commit -q --amend -m "base commit" && git push -q -f -u origin master
           else
-            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build"
+            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build."
```
lrineau, line 112: We can even display the full command, with the `:round` part:
```suggestion
            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build:${ROUND}."
```

MaelRL: I think it'd be nice if these "build_doc" commands actually contained "doc" in the name too.

lrineau: I agree. It should be build-doc:v1 and force-build-doc:v1.

I cannot add the `/` in my comment, otherwise our stupid action bot would trigger the action. I would be safer if the action bot only reacted when the command is alone in the message, or at least alone in the first line of the message.

PR #5971, opened 2021-09-08
Make sure the rounding mode is restored when doing exact computations (5.2)

sloriot:
Thoses changes should have been done when the code was modified for
CGAL_CFG_FPU_ROUNDING_MODE_UNWINDING_VC_BUG

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -803,6 +803,7 @@ struct Lazy_construction_bbox
     }
     CGAL_BRANCH_PROFILER_BRANCH(tmp);
     Protect_FPU_rounding<!Protection> P2(CGAL_FE_TONEAREST);
+    CGAL_expensive_assertion(FPU_get_cw() == CGAL_FE_TONEAREST);
     return ec(CGAL::exact(l1));
   }
 };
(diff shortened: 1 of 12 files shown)


mglisse:
Should we use Checked_protect_FPU_rounding for the TO_NEAREST before exact computations? (I didn't think about it much)

sloriot:
It is actually a good idea as we recently notice it is important with `boost::cpp_int` and `boost::cpp_rattional`. I'll add that.

Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -1594,7 +1602,7 @@ struct Lazy_construction_variant {
     }
     CGAL_BRANCH_PROFILER_BRANCH(tmp);
     Protect_FPU_rounding<!Protection> P2(CGAL_FE_TONEAREST);
-
+    CGAL_expensive_assertion(FPU_get_cw() == CGAL_FE_TONEAREST);
```
mglisse: Why not use Checked_protect_FPU_rounding (as in Interval_nt) which is supposed to do the same as those 2 lines? (although its last assertion seems to be bogus)

lrineau:
@sloriot What is the bug, and when was it introduced?

sloriot:
Years ago, when we discovered the non-restoration of the rounding mode by MSVC when going out of a try block to the catch block. The workaround has been correctly implemented for Filtered_predicates for example but it was not correct at some places. The PRs implements the workaround consistently. We can discuss IRL if you need more info

sloriot:
Successfully tested in CGAL-5.4-Ic-53

PR #5973, opened 2021-09-08
master version of #5971 (Make sure the rounding mode is restored when doing exact computations)

sloriot:
see #5971


sloriot:

PR #5992, opened 2021-09-15
Deal with memset, memcpy, fscanf...    and their safe equivalents

### LINKED ISSUE #5991: Use of unsafe functions
## Issue Details

CGAL uses functions like `memset()` which should be replace with `memset_s`. 
Occurrences in files in `include/CGAL/`  _must_ be replace, while occurrences in examples demos or developer scripts _may_ be replaced. 

The fix must be done for the 5.3 branch.

Note that this also requires to be documented because the users must `#define`  a macro in order to activate the runtime check (see the [`memset()` manual page](https://en.cppreference.com/w/c/string/byte/memset) on cppreference).

afabri:
## Summary of Changes

As pointed out by @lrineau most of the usages ARE safe, in the sense that the sizes of source and target memory are constants known at compile time.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #5991

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
@@ -885,7 +885,7 @@ public:
 
   /*! Clean all operation counters */
   void clear_counters()
-  { memset(m_counters, 0, sizeof(m_counters)); }
+  { memset_s(m_counters, sizeof(m_counters), 0, sizeof(m_counters)); }
 
 private:
   /*! The operation counters */
--- a/Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
@@ -27,6 +27,7 @@
 #include <iostream>
 #include <string.h>
 #include <atomic>
+#include <array>
 
 #include <CGAL/basic.h>
 #include <CGAL/Arr_enums.h>
@@ -885,11 +886,11 @@ public:
 
   /*! Clean all operation counters */
   void clear_counters()
-  { memset_s(m_counters, sizeof(m_counters), 0, sizeof(m_counters)); }
+  { m_counters = {}; }
 
 private:
   /*! The operation counters */
-  mutable size_t m_counters[NUMBER_OF_OPERATIONS];
+  mutable std::array<size_t, NUMBER_OF_OPERATIONS> m_counters;
 };
 
 template <typename Out_stream, class Base_traits>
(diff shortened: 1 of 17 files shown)


Arrangement_on_surface_2/include/CGAL/Arr_counting_traits_2.h
```diff
@@ -966,7 +966,7 @@ class Arr_counting_traits_2 : public Base_traits {
 
   /*! Clean all operation counters */
   void clear_counters()
-  { memset(m_counters, 0, sizeof(m_counters)); }
+  { memset_s(m_counters, sizeof(m_counters), 0, sizeof(m_counters)); }
```
afabri, line 969: @efifogel do you agree that this is not time critical and we can replace with  `std::fill()` ?

lrineau: In addition, you can modernize the code, and replace `m_counters` with a `std::array`.

CGAL_Core/include/CGAL/CORE/CoreAux_impl.h
```diff
@@ -198,10 +177,8 @@ void core_error(std::string msg, std::string file, int lineno, bool err) {
   outFile.close();
   if (err) {
      char buf[65];
```
lrineau, line 179: `buf` is now useless. It should be removed to avoid a warning about an unused variable.

Arrangement_on_surface_2/include/CGAL/IO/Fig_stream.h
```diff
@@ -1716,9 +1715,11 @@ class Fig_stream
...
       else
       {
-        // Convert the current character to an octal string and write it.
-        sprintf (oct, "\\%03o", text[i]);
-        _ofile << oct;
+        // Convert the current character to an octal string and write
+        // it.
+        std::stringstream out;
+        out << "\\" << std::setfill('0') << std::setw(3) << std::oct << text[i];
+        _ofile << out.str();
```
lrineau, line 1722: That code should stream to `_ofile` directly, instead of creating that temporary `std::stringstream`. :smile:

afabri: I did not do that as `std::hex`  and `std::oct` change all consecutive ìnt`.

Stream_support/include/CGAL/IO/trace.h
```diff
@@ -26,7 +26,7 @@ inline void CGAL_print_stderr(const char *fmt, ...)
 {
   va_list argp;
   va_start(argp, fmt);
-  vfprintf(stderr, fmt, argp);
+  vfprintf_s(stderr, fmt, argp);
```
lrineau: That will compile only with MSVC.

afabri:
@lrineau do we still need this [`Input_rep`](https://github.com/CGAL/cgal/blob/master/Stream_support/include/CGAL/IO/io.h#L245), that is do we still support `_MSC_VER < 1910`

sloriot:

sloriot:

PR #5993, opened 2021-09-15
CORE:  Fix memory leak

afabri:
## Summary of Changes

In case there is a read error we have to relase the memory.

## Release Management

* Affected package(s): CGAL_CORE


afabri:
Note that in #5992 I removed the `new` calls as I use `std::string` instead of arrays of `char`.

sloriot:
Successfully tested in CGAL-5.4-Ic-55

PR #5996, opened 2021-09-17
Restore sub-directories accidently removed in 4009

sloriot:


sloriot:
Successfully tested in CGAL-5.4-Ic-55

PR #6010, opened 2021-09-23
Avoid extra bool by relying on enum

sloriot:
Replaces #5194 and #5656

### DIFF
--- a/Intersections_2/include/CGAL/Intersections_2/Iso_rectangle_2_Line_2.h
+++ b/Intersections_2/include/CGAL/Intersections_2/Iso_rectangle_2_Line_2.h
@@ -36,11 +36,10 @@ namespace internal {
 template <class K>
 class Line_2_Iso_rectangle_2_pair {
 public:
-    enum Intersection_results {NO_INTERSECTION, POINT, SEGMENT};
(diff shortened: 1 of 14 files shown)


GilesBathgate:
@sloriot The original `NOT_COMPUTED_YET` was in Line_2_Line_2.h : https://github.com/CGAL/cgal/blob/master/Intersections_2/include/CGAL/Intersections_2/Line_2_Line_2.h
Should this be renamed as UNKNOWN?

sloriot:
Just did but maybe somebody will tell me do rename the other way round

lrineau:
> @sloriot What's weird is that `_known` was only ever initlaised to false, when did it ever take on a value of true in the old code?

Search for `_know = true` in the diff. For example at https://github.com/CGAL/cgal/pull/6010/files#diff-1e09ec32b96e20ff1ad8936bcafc56ab390a12bed70d84c98043262eb535bc81L90 (on the left side, in red because the line disappears with the patch).

GilesBathgate:
@lrineau Sorry, I need to get my eyes checked :eyes: . I was certain it wasn't there.

sloriot:

PR #6011, opened 2021-09-24
Mesh_3 - warning message after bad initialization

### LINKED ISSUE #5300: precondition/warning if no facet in complex are present in a c3t3 passed to refine_mesh_3
If the c3t3 is manually initialized, one has then to call `refine_mesh_3()` to run the generation of the mesh. However, if the initialization is not sufficient (no facet is in the complex) then nothing is done. Adding a return type, a warning or an assertion would help the user understand what is going wrong. Maybe the initialization function of `make_mesh_3()` could also be called in such a case.

janetournois:
## Summary of Changes

This PR adds a warning message that is displayed without crashing when initialization of Mesh_3 has failed and refinement cannot start

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): fix #5300

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
+++ b/Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
@@ -45,6 +45,7 @@
 #include <CGAL/Mesher_level_visitors.h>
 #include <CGAL/Kernel_traits.h>
 #include <CGAL/point_generators_3.h>
+#include <CGAL/assertions_impl.h>
 
 #ifdef CGAL_MESH_3_USE_OLD_SURFACE_RESTRICTED_DELAUNAY_UPDATE
 #include <CGAL/Surface_mesher/Surface_mesher_visitor.h>
@@ -719,6 +720,20 @@ initialize()
   else
 #endif // CGAL_LINKED_WITH_TBB
   {
+
+    if (r_c3t3_.number_of_facets() == 0)
+    {
+      CGAL::warning_fail("c3t3.number_of_facets() == 0",
+        "CGAL/Mesh_3/Mesher_3.h",
+        725,
+        "Warning : The mesh refinement process can't start.\n"
+        "When calling refine_mesh_3(), the input `c3t3` should have been initialized and have "
+        "at least one facet in the complex. Try to solve this issue using :\n"
+        "\t- The automatic initialization provided by make_mesh_3()\n"
+        "\t- Adding more and better chosen points on the input surface\n"
+        "\t- defining the compilation macro CGAL_SEQUENTIAL_MESH_3_ADD_OUTSIDE_POINTS_ON_A_FAR_SPHERE\n");
+    }
+
     if (r_c3t3_.number_of_far_points() == 0 &&
         r_c3t3_.number_of_facets() == 0 &&
         (r_c3t3_.triangulation().dimension() < 3
--- a/STL_Extension/include/CGAL/assertions_impl.h
+++ b/STL_Extension/include/CGAL/assertions_impl.h
@@ -14,6 +14,9 @@
 //
 // Author(s)     : Geert-Jan Giezeman and Sven Schönherr
 
+#ifndef CGAL_ASSERTIONS_IMPL_H
+#define CGAL_ASSERTIONS_IMPL_H
+
 #ifdef CGAL_HEADER_ONLY
 #define CGAL_INLINE_FUNCTION inline
 #else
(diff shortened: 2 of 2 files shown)


Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
```diff
@@ -45,6 +45,7 @@
 #include <CGAL/Mesher_level_visitors.h>
 #include <CGAL/Kernel_traits.h>
 #include <CGAL/point_generators_3.h>
+#include <CGAL/assertions_impl.h>
```
lrineau, line 48: The `*_impl.h` headers are implementation details (for header-only) and should never be included explicitly: they are included by the non-impl headers.
```suggestion
#include <CGAL/assertions.h>
```

Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
```diff
@@ -719,6 +720,20 @@ initialize()
   else
 #endif // CGAL_LINKED_WITH_TBB
   {
+
+    if (r_c3t3_.number_of_facets() == 0)
+    {
+      CGAL::warning_fail("c3t3.number_of_facets() == 0",
+        "CGAL/Mesh_3/Mesher_3.h",
+        725,
+        "Warning : The mesh refinement process can't start.\n"
```
lrineau, line 729: We do not want to hard-code the file and line, but use the macros `__FILE__` and `__LINE__`.
```suggestion
      CGAL::warning_fail("c3t3.number_of_facets() == 0",
        __FILE__,
        __LINE__,
        "Warning : The mesh refinement process can't start.\n"
```
Those macros are [ones of the few officially documented in the C++ norm](https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros).

Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
```diff
@@ -719,6 +720,20 @@ initialize()
...
+        "When calling refine_mesh_3(), the input `c3t3` should have been initialized and have "
+        "at least one facet in the complex. Try to solve this issue using :\n"
+        "\t- The automatic initialization provided by make_mesh_3()\n"
+        "\t- Adding more and better chosen points on the input surface\n"
+        "\t- defining the compilation macro CGAL_SEQUENTIAL_MESH_3_ADD_OUTSIDE_POINTS_ON_A_FAR_SPHERE\n");
```
lrineau, line 734: Are you sure `CGAL_SEQUENTIAL_MESH_3_ADD_OUTSIDE_POINTS_ON_A_FAR_SPHERE` will solve anything?

Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
```diff
@@ -719,6 +720,20 @@ initialize()
   else
 #endif // CGAL_LINKED_WITH_TBB
   {
+
+    if (r_c3t3_.number_of_facets() == 0)
+    {
+      CGAL::warning_fail("c3t3.number_of_facets() == 0",
+        "CGAL/Mesh_3/Mesher_3.h",
```
lrineau, line 727: @CGAL/geometryfactory Here Jane and I have decided to call `CGAL::warning_fail` directly, because we wanted a warning that is not disable in Release mode. Is that a good idea? The advantage of `CGAL::warning_fail` over a plain display to `std::cerr` is that the behavior can be [tweaked by the user/developer, using behaviors and handlers](https://doc.cgal.org/latest/STL_Extension/group__PkgSTLExtensionAssertions.html).

Should it be a macro, so that it can be disabled anyway with a dedicated macro definition?

Mesh_3/include/CGAL/Mesh_3/Mesher_3.h
janetournois: I'm not sure, but it's a hint...
Do you think it's there only for benchmarking/performance reasons?

lrineau: We have added `CGAL_SEQUENTIAL_MESH_3_ADD_OUTSIDE_POINTS_ON_A_FAR_SPHERE` for robustness issue, to avoid to have facets of the c3t3 on the convex hull. I do not know if that can help the initialization of the triangulation.

janetournois: ok I'll remove it

sloriot:
Successfully tested in CGAL-5.4-Ic-65

PR #6019, opened 2021-09-28
Fix warnings about arithmetic operations on enum

sloriot:
-Wdeprecated-enum-enum-conversion


sloriot:
Successfully tested in CGAL-5.4-Ic-61

PR #5563, opened 2021-03-29
[Small Feature] Region Growing Revision

danston:
This PR is a revision of the `Region_growing` component from the `Shape_detection` package. It includes multiple small and big changes (including a few breaking changes!):

- added named parameters (that changed the APIs of many class constructors)
- now the update function from the `RegionType` concept returns a boolean flag that enables to detect if the input conditions for the new region are satisfied (it is also a breaking change)
- old fitting classes are now using better versions of the region conditions, more precise and faster, including the correct normal orientations
- added new classes for getting linear regions in a set of 2D and 3D segments
- added the `Polyline_graph` class for extracting a set of polylines from a face graph, which splits this graph into a set of user-defined regions (this is necessary for the decimation algorithm in the PMP)
- added new classes for getting linear regions on 2D and 3D polylines
- added several internal functions, which are not yet public, since I am not sure whether we really want to document them
- extended docs with new examples
- better and stricter tests for the test suite
- a few other fixes and internal changes
- fixed the Polyhedron demo using new API

To be yet done:
- this version should be later merged with the PR #5808 that extends the region growing algorithm with detecting circles, spheres, and cylinders

**The doc files changed:**
- the concept [RegionType.h](https://cgal.geometryfactory.com/~danston/Shape_detection/classRegionType.html)
- [Shape_detection.txt](https://cgal.geometryfactory.com/~danston/Shape_detection/index.html#Shape_detection_RegionGrowing) (sections [Segment Set](https://cgal.geometryfactory.com/~danston/Shape_detection/index.html#Shape_detection_RegionGrowingSegments),  [Polyline](https://cgal.geometryfactory.com/~danston/Shape_detection/index.html#Shape_detection_RegionGrowingPolyline), [Deprecated](https://cgal.geometryfactory.com/~danston/Shape_detection/index.html#Shape_detection_DeprecatedComponents))
- named parameters descriptions for all headers in the ref manual for [Region Growing](https://cgal.geometryfactory.com/~danston/Shape_detection/group__PkgShapeDetectionRG.html)
- [Segment Set](https://cgal.geometryfactory.com/~danston/Shape_detection/group__PkgShapeDetectionRGOnSegments.html) and [Polyline](https://cgal.geometryfactory.com/~danston/Shape_detection/group__PkgShapeDetectionRGOnPolyline.html) classes

**TODO:**
- [ ] shall `Polygon_mesh` sub-namespace/headers be renamed `Triangle_mesh`? is not a good name
[...]

PR #5461, opened 2021-02-15
Add coplanar decimation

sloriot:
## Summary of Changes

Add two functions in PMP that generate from a triangle mesh a new version with patches of (almost) coplanar faces being retriangulated while removing interior vertices and collinear boundary vertices:

- [remesh_planar_patches](https://cgal.github.io/5461/v1/Polygon_mesh_processing/group__PMP__meshing__grp.html#ga8a0dab9d4b6fd799672f9b37ddc9481f)
- [remesh_almost_coplanar_patches](https://cgal.github.io/5461/v0/Polygon_mesh_processing/group__PMP__meshing__grp.html#ga9c589830c74a19d5f513ff408967b343)

## TODO:
- [x] split the example in 2, one part should go into test (together with the data provided)
- [x] decide license group
- [x] small feature
- [x] fix dependencies
- [x] update PMP user manual
- [x] demo plugins
- [x] Update changes.md
- [x] add an option to to triangulate faces
- [x] debug bracket

## Release Management

* Affected package(s): PMP,
* Feature/Small Feature (if any): [here](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/PMP-decimate), pre-approved 20 March 2023
* Link to compiled documentation (obligatory for small feature) [here](https://cgal.github.io/5461/v0/Polygon_mesh_processing)
* License and copyright ownership: GF

PR #6024, opened 2021-09-29
Surface Mesh Segmentation Using Region Growing

danston:
This PR is a draft version of the planar decimation feature from PMP using Region Growing when segmenting an input mesh. We use Region Growing first to detect planar shapes. We then exploit these shapes to build a graph of polylines on the mesh and apply Region Growing again on top of this graph in order to detect linear regions. When we know all linear regions, we can easily detect unique corners of the mesh.

This PR depends on PR #5563 and PR #5461.

Files to check:
- `PMP decimation_planar_patches.cpp`
- `PMP planar_segmentation.h`
- `SMS plane_mesh_segmentation.h `

## Release Management

* Affected package(s): `Polygon_mesh_processing`, `Surface_mesh_segmentation`
* Issue(s) solved (if any): no fixes
* Feature/Small Feature (if any): to be discussed
* Link to compiled documentation (obligatory for small feature): no docs, only internal changes


sloriot:
We decided not to add a new API to the segmentation package.

PR #6035, opened 2021-10-07
CORE: Use boost::multiprecision::number API

afabri:
## Summary of Changes

Currently CGAL_Core uses the low level API of GMP for `BigInt` and `BigRat`.  In this PR  I replace this with the `boost::multiprecision::number` API, so that we can easily switch between the GMP and the cpp backends.

## Release Management

* Affected package(s): CGAL_Core

TODO:
 - [x] document Boost 1.70 is required.
 - [x] Announce (CHANGES.md) the boost version bump


CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -260,147 +292,165 @@ class CGAL_CORE_EXPORT BigInt : public RCBigInt {
...
   a.makeCopy();
-  mpz_neg(a.get_mp(), a.get_mp());
+  a.get_mp() = - a.get_mp();
 }
+
 /// cmpabs
 inline int cmpabs(const BigInt& a, const BigInt& b) {
-  return mpz_cmpabs(a.get_mp(), b.get_mp());
+  assert(false);
+  // return mpz_cmpabs(a.get_mp(), b.get_mp());   AF: todo
```
mglisse, line 452:
```suggestion
   return cmp(abs(a), abs(b));
```

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -260,147 +292,165 @@ class CGAL_CORE_EXPORT BigInt : public RCBigInt {
...
 inline void negate(BigInt& a) {
   a.makeCopy();
-  mpz_neg(a.get_mp(), a.get_mp());
+  a.get_mp() = - a.get_mp();
 }
+
 /// cmpabs
 inline int cmpabs(const BigInt& a, const BigInt& b) {
-  return mpz_cmpabs(a.get_mp(), b.get_mp());
+  assert(false);
```
mglisse, line 451:
```suggestion
```

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -409,73 +459,99 @@ inline int cmpabs(const BigInt& a, const BigInt& b) {
...
+        m.get_mp() = q.get_mp();
+        ++e;
+    }
 }
 
 /// divisible(x,y) = "x | y"
 inline bool isDivisible(const BigInt& x, const BigInt& y) {
-  return mpz_divisible_p(x.get_mp(), y.get_mp()) != 0;
+    BigInt z, r;
+    divide_qr(x.get_mp(), y.get_mp(), z.get_mp(), r.get_mp());
```
mglisse, line 524: integer_modulus?

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -409,73 +459,99 @@ inline int cmpabs(const BigInt& a, const BigInt& b) {
...
 inline bool isDivisible(long x, long y) {
   return x % y == 0;
+
 }
 /// exact div
 inline void divexact(BigInt& z, const BigInt& x, const BigInt& y) {
   z.makeCopy();
-  mpz_divexact(z.get_mp(), x.get_mp(), y.get_mp());
+  BigInt r;
+  divide_qr(x.get_mp(), y.get_mp(), z.get_mp(), r.get_mp() );  // was void mpz_divexact (mpz_t q, const mpz_t n, const mpz_t d)   Is this faster?
```
mglisse, line 540: assert r.is_zero()

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -24,86 +24,116 @@
...
+  BigIntRep()
+    : mp()
+    {}
+
+
+  // Note : should the copy-ctor be allowed at all ? [Sylvain Pion]
+    BigIntRep(const BigIntRep& z) : RCRepImpl<BigIntRep >(), mp(z.mp)
+    {}
+
+  BigIntRep(signed char c)
```
mglisse, line 54: Maybe a template `BigIntRep(T t):mp(t){}` would be possible?

mglisse (commented):
With Sébastien

afabri:
The problem in the testsuite comes from, I believe, a bug in boost multiprecision.  When reading "5,600" from a stream into a multiprecision rational the comma is not recognized as a delimiter.  I filed an [issue ](https://github.com/boostorg/multiprecision/issues/468).

CGAL_Core/include/CGAL/CORE/BigInt.h
mglisse: Ah, indeed, `std::enable_if<boost::multiprecision::detail::is_integral<Integer>::value`. I think it would be worth opening an issue so they clarify their doc. The only hint there is the fact that the argument is passed by value...

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -24,86 +24,80 @@
 #ifndef _CORE_BIGINT_H_
 #define _CORE_BIGINT_H_
 
+#include <CGAL/boost_mp.h>
```
sloriot, line 27: boost mp is available only in boost 1.70. What is the plan, require 1.70 if Core is used?

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -24,86 +24,80 @@
...
 #include <CGAL/CORE/Gmp.h>
 #include <CGAL/CORE/RefCount.h>
 #include <CGAL/CORE/MemoryPool.h>
 #include <string>
 
 namespace CORE {
 
+#ifdef CGAL_CORE_USE_GMP_BACKEND
```
sloriot, line 35: Similar question here. We recommend to use at least boost 1.79 for boost backend. Should we take that into account for defining `CGAL_CORE_USE_GMP_BACKEND`?

CGAL_Core/include/CGAL/CORE/BigInt.h
afabri: What do we require today? And is that because it is what is in Linux distributions?

afabri:
> The problem in the testsuite comes from, I believe, a bug in boost multiprecision. When reading "5,600" from a stream into a multiprecision rational the comma is not recognized as a delimiter. I filed an [issue ](https://github.com/boostorg/multiprecision/issues/468).

According to @jzmaddock  [the fix](https://github.com/boostorg/multiprecision/issues/468#issuecomment-1146788711) is now in the boost develop branch . @lrineau @sloriot do we have a testsuite that uses HEAD of the boost develop branch?

lrineau:
> > The problem in the testsuite comes from, I believe, a bug in boost multiprecision. When reading "5,600" from a stream into a multiprecision rational the comma is not recognized as a delimiter. I filed an [issue ](https://github.com/boostorg/multiprecision/issues/468).
> 
> According to @jzmaddock [the fix](https://github.com/boostorg/multiprecision/issues/468#issuecomment-1146788711) is now in the boost develop branch . @lrineau @sloriot do we have a testsuite that uses HEAD of the boost develop branch?

For the moment we have two: they are the ones showing version `1.80.0` for Boost:
  - [`ArchLinux-clang-CXX20-Release.gz`](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4.1-I-161/Installation/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz), and
  - [`Ubuntu-GCC_master_CXX20-Release`](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4.1-I-161/Installation/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz).

They coincide with the platform using C++20.

I thought that we should have at least one Windows platform, using Boost `develop` branch.

lrineau:
> > > The problem in the testsuite comes from, I believe, a bug in boost multiprecision. When reading "5,600" from a stream into a multiprecision rational the comma is not recognized as a delimiter. I filed an [issue ](https://github.com/boostorg/multiprecision/issues/468).
> > 
> > 
> > According to @jzmaddock [the fix](https://github.com/boostorg/multiprecision/issues/468#issuecomment-1146788711) is now in the boost develop branch . @lrineau @sloriot do we have a testsuite that uses HEAD of the boost develop branch?
> 
> For the moment we have two: they are the ones showing version `1.80.0` for Boost:
> 
>     * [`ArchLinux-clang-CXX20-Release.gz`](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4.1-I-161/Installation/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz), and
> 
>     * [`Ubuntu-GCC_master_CXX20-Release`](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4.1-I-161/Installation/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz).
> 
> 
> They coincide with the platforms using C++20.
> 
> I thought that we should have at least one Windows platform, using Boost `develop` branch.

The platform `MSVC-2022-Preview-Release` will soon use Boost `develop` as well. There was a permission error on a `git-fetch`.

CGAL_Core/include/CGAL/CORE/BigInt.h
lrineau: With Andreas we have decided to document that Boost 1.70 is required, starting with CGAL 5.6.

lrineau:
With Andreas we have decided to document that Boost 1.70 is required, starting with CGAL 5.6.

lrineau:
Yes. All the platforms announcing boost 1.80 are using the develop branch of it.

lrineau:
> Yes. All the platforms announcing boost 1.80 are using the develop branch of it. 

Then that means that platform is still stuck with an older version. I will have a look.

lrineau:
> > Yes. All the platforms announcing boost 1.80 are using the develop branch of it.
> 
> Then that means that platform is still stuck with an older version. I will have a look.

On the machine `sosno`, several versions (1.78, 1.79, 1.80 aka `develop`) are installed in the directory `C:/dev/boost/boost-git/install_dir/` and CMake picked the first one. I have removed (actually renamed) the `BoostConfig.cmake` file for 1.78 and 1.79. :crossed_fingers:

sloriot:
CGAL-5.6-Ic-6 is still not good. In particular,there are some issues on platforms with old boost version.

sloriot:
@afabri we can revive this PR now.

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -46,54 +46,18 @@ namespace CORE {
...
-    {}
-
-  BigIntRep(signed char c)
-    : mp(c)
+  BigIntRep(const BigIntRep& z) : RCRepImpl<BigIntRep >(), mp(z.mp)
     {}
 
-  BigIntRep(unsigned char c)
+  template <class T>
+  BigIntRep(T c)
```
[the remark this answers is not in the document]
sloriot: @lrineau explicit should be safe enough, right?

CGAL_Core/include/CGAL/CORE/BigRat.h
```diff
@@ -261,36 +261,37 @@ class BigRat : public RCBigRat {
   /// Canonicalize
   void canonicalize() {
     makeCopy();
-    mpq_canonicalize(get_mp());
+    assert(false);
+    // AF todo    mpq_canonicalize(get_mp());
```
sloriot, line 265: Shouldn't we simply remove that function that does not seem to be used?

CGAL_Core/include/CGAL/CORE/BigInt.h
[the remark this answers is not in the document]
lrineau: > @lrineau explicit should be safe enough, right?

No, `explicit` will not change anything. But, on the second look, I think there is no ambiguity. The following constructors take a reference to const:
```c++
  BigIntRep(const std::string& s, int base=0);
  BigIntRep(const std::string& s, int base=0)
```
where as the constructor template takes the parameter by value:
```c++
  template <class T> BigIntRep(T c);
```

lrineau:
In my opinion, this pull-request should be a top priority for CGAL-5.6. It would allow to announce the possibility to use CGAL completely without GMP. As we have already removed the need for precompiled Boost libraries, that would make CGAL a header-only library (without binary dependencies).

CGAL_Core/include/CGAL/CORE/BigInt.h
```diff
@@ -24,499 +24,134 @@
 #ifndef _CORE_BIGINT_H_
 #define _CORE_BIGINT_H_
 
+#include <CGAL/boost_mp_type.h>
 #include <CGAL/CORE/Gmp.h>
```
sloriot, line 28: I'm not sure the header `CGAL/CORE/Gmp.h` is needed anymore.

lrineau:
There are compilation errors in the CI: see the first error at https://github.com/CGAL/cgal/actions/runs/4113759064/jobs/7100276679#step:4:5109.

Number_types/include/CGAL/boost_mp_type.h
```diff
@@ -0,0 +1,1012 @@
...
+    class Decompose {
+    public:
+        typedef Type first_argument_type;
+        typedef Numerator_type& second_argument_type;
+        typedef Denominator_type& third_argument_type;
+        void operator () (
+                const Type& rat,
+                Numerator_type& num,
+                Denominator_type& den) {
+            num = numerator(rat);
```
afabri, line 701: @sloriot don't we have a problem here, namely that we do not get a reference of the numerator and denominator?

sloriot: No, a copy is made and put into the reference passed as input.

CGAL_Core/include/CGAL/CORE/BigFloat.h
```diff
@@ -614,11 +614,23 @@ inline BigFloat gcd(const BigFloat& a, const BigFloat& b) {
   //mpz_tdiv_qr(q.get_mp(), r.get_mp(), a.get_mp(), b.get_mp());
 //}//
 
-
+/* AF
```
afabri: We need a construction of a `BigRat` from a `BigFloat`  [here ](https://github.com/CGAL/cgal/blob/d979e070bee8f49344125f58e298be51eba84838/Arrangement_on_surface_2/include/CGAL/Arr_rat_arc/Algebraic_point_2.h#L229)in  the class `Algebraic_point_2`

Algebraic_kernel_d/include/CGAL/Algebraic_kernel_d/Bitstream_descartes_rndl_tree_traits.h
```diff
@@ -28,7 +28,7 @@
 #include <vector>
 
 #if CGAL_USE_CORE
-namespace CORE { class BigInt; }
+#include <CGAL/CORE/BigInt.h>
```
lrineau, line 31: That include makes CGAL_Core a new dependency of all CGAL packages.

Why not keep the forward declaration?

afabri: Because it is only a `typedef` to the boost mp rational type.

Number_types/include/CGAL/Root_of_traits.h
```diff
@@ -76,7 +76,9 @@ compute_roots_of_2(const NT &a_, const NT &b_, const NT &c_, OutputIterator oit)
     }
   }
   else {
-    *oit++ = -c/b; return oit;
+      assert(false); 
+    // *oit++ = - c / b;  // AF: HELP
```
afabri, line 80: @sloriot can you uncomment and investigate.  If I write `NT(-c/b)`  it won''t work for `Quotient<MP_Float>` and without it has a problem with expression templates of boost::mp.

sloriot: It is strange that line 66 is not an issue then.

sloriot:
Test failing in CGAL-5.6-Ic-191

Envelope_3/include/CGAL/Env_sphere_traits_3.h
```diff
@@ -1080,7 +1080,7 @@ class Env_sphere_traits_3 : public ConicTraits_2
     Algebraic x_diff = x1 - a, y_diff = y1 - b;
     // the coefficients are:
     Algebraic A = 1,
-              B = -2*c,
+              B = Rational(-2*c),
```
afabri, line 1083: To write `Rational` to fix CGAL code is not really a problem, but it worries me for user code.
I have put an [issue ](https://github.com/boostorg/multiprecision/issues/541) in the boost multiprecision project on github.

lrineau: https://github.com/boostorg/multiprecision/issues/541 is another reason to require C++17 for CGAL-5.6.

lrineau: Actually, that is the fix https://github.com/boostorg/multiprecision/pull/543 that will require C++17, in Boost 1.82.

jzmaddock:
> Actually, that is the fix https://github.com/boostorg/multiprecision/pull/543 that will require C++17, in Boost 1.82.

If anyone can figure out a way to make this operator https://github.com/boostorg/multiprecision/blob/02e753d60364d2318827a98f7b12d98cc60244a4/include/boost/multiprecision/detail/number_base.hpp#L960 work in C++11 (or 14) so that both  these two tests compile: https://github.com/boostorg/multiprecision/blob/develop/test/bug11922.cpp and https://github.com/boostorg/multiprecision/blob/develop/test/git_issue_541.cpp I'd love to hear about it.  I've tried every enable_if configuration I can think of but one test or the other always seems to fail except post C++17.

afabri:
In the Number_types testsuite we have [`test_eigen.cpp`](https://github.com/CGAL/cgal/blob/master/Number_types/test/Number_types/test_eigen.cpp) which fails for VC2017.   I boiled it down to a [self contained example](https://gist.github.com/afabri/acc723427888dc49c60099bdb6220c53) without any CGAL
and put it on gist.github.com

Envelope_3/include/CGAL/Env_sphere_traits_3.h
afabri: It is one thing to require C++17, but a completely different thing  to require Boost 1.82 as minimal version for CGAL. 
The fix is not some code like a traits class or a free function that could be packaged separately to fix older versions of Boost. 
So maybe the idea to just `typedef  boost::mp::cpp_int BigInt`   is not viable and we still need  a wrapper class around it.   Which is a pity as we would not profit from expression templates.

mglisse (commented):
The cpp_float approach should work (I didn't check the details).
It is hard to predict what performance will be like, but I would expect something like this to be good enough to be useful (i.e. faster than cpp_rational to evaluate a polynomial predicate exactly from double input).
The name `cpp_float` might step a bit on Boost's toes.

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,432 @@
...
+    rat += other.rat;
+    int shift = exp - other.exp;
+    if(shift > 0){
+      man <<= shift;
+      man += other.man;
+      exp = other.exp;
+    }else if(shift < 0){
+      boost::multiprecision::cpp_int cpy(other.man);
+      cpy <<= -shift;
+      man += cpy;
```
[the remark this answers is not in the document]
afabri: When I compute a Delaunay triangulation with a lot of degeneracies to force the usage of `Mpzf/cpp_float` the execution is two times slower with `cpp_float`, most of the time spent in resizing allocated memory when computing a product. Can we know the maximum size of the mantissa of all predicates?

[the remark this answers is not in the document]
mglisse: I don't think you want the maximum (DBL_MIN^4+DBL_MAX^4 type of operation quickly gives a large mantissa), just something that covers most cases in practice. It looks like I used 512 bits for Mpzf, by trial and error, picking the lowest value that seemed to have ok performance. I was suggesting trying `number<cpp_int_backend<1024>>` in my old email, though the number 1024 was likely random.

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,487 @@
...
+#endif
+    man *= other.man;
+    exp += other.exp;
+    return *this;
+  }
+
+
+  friend
+  cpp_float operator*(const cpp_float& a, const cpp_float&b){
+    return cpp_float(a.man*b.man, a.exp+b.exp);
```
mglisse, line 193: It could be useful to have a constructor for cpp_float that accepts an expression as first argument so the product can be evaluated directly in the new cpp_float instead of first in a cpp_int and then copied. Or at least one that accepts a cpp_int rvalue so it can be moved instead of copied.

Number_types/include/CGAL/cpp_float.h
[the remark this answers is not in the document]
afabri: But where in Mpzf do you deal with when you need more than 512 bits?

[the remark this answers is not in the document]
mglisse: In the init() function, I allocate with new if the cache is too small. Same as cpp_int_backend does if you leave the template parameter MaxBits to 0.

Number_types/include/CGAL/cpp_float.h
afabri: It in fact IS useful.

Number_types/include/CGAL/cpp_float.h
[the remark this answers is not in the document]
afabri: I only realized now that `cpp_int_backend` allocates more if needed. cool.  This together with the constructor for expression templates my testcase of Delaunay_3 with the ocean data set takes 24 sec for `Mpfz` and 32 sec for `cpp_float`, so the gap starts closing.

afabri:
With ` Mpzf` as well as with `cpp_float`   I observed this in vtune.
image

I am wondering if we make copies that could be replaced with a ` std::move()` ?

mglisse:
> I am wondering if we make copies that could be replaced with a ` std::move()` ?

Yes we do (not really a new problem, it also affects Epeck). First thing could be dropping the use of make_array, `Rep{x,y,z}` works just as well, and that's one function less to update. And then you would want to have move versions (or perfect forwarding) of a number of functions like point constructors, etc.

Number_types/include/CGAL/cpp_float.h
```diff
@@ -169,7 +169,7 @@ class cpp_float {
     }
     // std::cout << "m = " << m << " * 2^" << exp  << std::endl;
     // fmt(m);
-    man = Mantissa(m);
+    man = std::move(m);
```
lrineau, line 172: As `m` is `uint64_t m;`, the operator for a move of `m` is a simple copy of it.

afabri:
The ` operator()`  of some partial specializations of ` NT_converter` returns const references. That makes it maybe more complicated.

Number_types/include/CGAL/cpp_float.h
afabri: I thought this avoids the construction of a ` Mantissa`  object that in the assignment gets copied.

mglisse: s/copied/moved/
`man = m;` should be fine.

afabri:
> 



> 



> > I am wondering if we make copies that could be replaced with a ` std::move()` ?
> 
> Yes we do (not really a new problem, it also affects Epeck). First thing could be dropping the use of make_array, `Rep{x,y,z}` works just as well, and that's one function less to update. And then you would want to have move versions (or perfect forwarding) of a number of functions like point constructors, etc.

Should maybe become an issue.

afabri:
Why does the  `Mpfz` move constructor have to call `mpn_copy_i()` [here](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/Mpzf.h#L350)?

mglisse:
> Why does the `Mpfz` move constructor have to call `mpn_copy_i()` [here](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/Mpzf.h#L350)?

If the number is small and stored directly in the object (as opposed to some memory allocated with `new`), you don't have much choice, you can't make the new object point to the data in the old object. The good thing is that this copy cannot be more than 512 bits (or whatever size was chosen).

sloriot:
It is not reasonable to have this branch in 5.6 since we don't have yet a green testsuite and no bench on memory and runtime. I postpone it to 5.7

Kernel_23/include/CGAL/Vector_3.h
```diff
@@ -91,6 +91,13 @@ class Vector_3 : public R_::Kernel_base::Vector_3
   Vector_3(const T1 &x, const T2 &y, const T3 &z)
     : Rep(typename R::Construct_vector_3()(Return_base_tag(), x, y, z)) {}
 
+  template < typename T1, typename T2, typename T3 >
+  Vector_3(T1&& x, T2&& y, T3&& z)
```
mglisse, line 95: You can remove the previous constructor, which becomes redundant.

Cartesian_kernel/include/CGAL/Cartesian/Point_3.h
```diff
@@ -45,6 +45,11 @@ class PointC3
   PointC3(const FT &x, const FT &y, const FT &z)
     : base(x, y, z) {}
 
+  PointC3(FT &&x, FT &&y, FT &&z)
+    : base(std::forward<FT>(x), std::forward<FT>(y), std::forward<FT>(z))
```
mglisse, line 49: It is a bit confusing to use std::forward here, std::move would be more natural. Or you could make the types of the 3 arguments template parameters and then using std::forward would be good.

afabri:
@sloriot the CI shows errors for  `#if CGAL_USE_CORE`    If I get it right it  seems not to be defined. Should it now always be defined, or do we have to replace the `#if` by a `#ifdef`  ?

afabri:
In the[ testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-189.shtml#AABB_tree)   we get error messages for the platform _x86-64_Darwin-18.5_Apple-clang-default_Release_.  This is because in `Installation/cmake/modules/CGAL_SetupCGALDependencies.cmake`  we switch it off:  " -- Boost MP is turned off for all Apple Clang versions below 11.0.3!".

afabri:
For the platform  `macOS-Apple_M2-boost_backend-arm64e`  the version of boost is recent enough for using `boost::MP` in `Core`. 

In the compilation line in the [AABB testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-189/AABB_tree/TestReport_Taylor_macOS-Apple_M2-boost_backend-arm64e.gz) where we use `Epick` we see `-DCGAL_DISABLE_GMP=1`.

Why do we include the Core header if we only use `Epick`? 

In `Core/Bigint.h` line 33 we write
```
#if !(defined(CGAL_CORE_USE_BOOST_BACKEND) && BOOST_VERSION > 107900 && defined(CGAL_USE_BOOST_MP))
    return mp;
#define CGAL_CORE_USE_GMP_BACKEND 1
  }
#endif
```
Why do we test for `CGAL_CORE_USE_BOOST_BACKEND`?  Is that not hard-wired by this PR ???

sloriot:
> Bigint
`BOOST_BACKEND`  by opposition to `GMP_BACKEND` that is the default

sloriot:
Successfully tested in CGAL-6.0-Ic-200

lrineau:
Is this pull-request (marked https://github.com/CGAL/cgal/labels/Tested) still a work-in-progress, or can it be merged?

sloriot:
It can be merged now.

PR #6038, opened 2021-10-08
Various fixes for property maps (5.2)

MaelRL:
## Summary of Changes

- Some bug fixes (Dereference_pmap was broken)
- Use `value_type` when the returned type of `get(map, key)` is not a reference
- Enforce `reference = value_type` if the returned type is not a reference (to avoid `typename PM::reference r = get(m, k)`, which will take a ref to a temporary if the `reference` typedef is an actualy reference)
- Do not use `put_get_helper` if the property map is not a `lvalue` **mutable** property map: the `put()` is `map[k] = v`, which is broken if `operator[]` does not return a reference
- The concept `boost::lvalue_property_map_tag` requires `operator[](key)`, not a reference in `get(map, key)`. You can have a readable property map returning a reference through its `get(map, key)`, but if there is no `operator[]`, it's just a `boost::readable_property_map_tag`
- Some const correctness to avoid copying maps with state or heavy keys in `get(map, key)` / `put(map, key, value)`
- Base the category of a wrapping pmap on what it offers instead of just forwarding the base property map's category
- Tried to do something like mutable lvalue pmap: 
  - `value_type& operator[](key&)` 
  - `ref get(map, const key&)`
  - `put(map, const key&, const value_type&)`
  and non-mutable lvalue pmap:
  - `const value_type& operator[](const key&)`
  - `ref get(map, const key&)`
  but not everything fits properly...

## Release Management
- For `master`: #6046
- For `5.3.x-branch`: #6047
- For `5.2.x-branch`: #6038

* Affected package(s): Various
* Issue(s) solved (if any): -

### DIFF
--- a/Property_map/include/CGAL/property_map.h
+++ b/Property_map/include/CGAL/property_map.h
@@ -206,7 +206,7 @@ struct Identity_property_map
 
   /// Access a property map element.
   /// @param k a key which is returned as mapped value.
-  value_type& operator[](key_type& k) const { return k; }
+  const value_type& operator[](const key_type& k) const { return k; }
 
   /// \name Put/get free functions
   /// @{
@@ -260,7 +260,7 @@ struct First_of_pair_property_map
 
   /// Access a property map element.
   /// @param pair a key whose first item is accessed
-  value_type& operator[](key_type& pair) const { return pair.first; }
+  const value_type& operator[](const key_type& pair) const { return pair.first; }
 
   /// \name Put/get free functions
   /// @{
@@ -300,7 +300,7 @@ struct Second_of_pair_property_map
 
   /// Access a property map element.
   /// @param pair a key whose second item is accessed
-  value_type& operator[](key_type& pair) const { return pair.second; }
+  const value_type& operator[](const key_type& pair) const { return pair.second; }
 
   /// \name Put/get free functions
   /// @{
@@ -343,7 +343,7 @@ struct Nth_of_tuple_property_map
 
   /// Access a property map element.
   /// @param tuple a key whose Nth item is accessed
-  value_type& operator[](key_type& tuple) const { return tuple.template get<N>(); }
+  const value_type& operator[](const key_type& tuple) const { return tuple.template get<N>(); }
 
   /// \name Put/get free functions
   /// @{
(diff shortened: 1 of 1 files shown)


lrineau:
I think we use C++ concepts, in our testsuite, to do real concept-checking in the testsuite, for the new columns that actually compile with C++20.

> Use `value_type` when the returned type of `get(map, key)` is not a reference

In my understanding, you use that as a way to read the code better, so that `value_type` appear as the return type of the `get` method. But the concepts of Boost are clear: [Readable Property Map](https://www.boost.org/doc/libs/1_77_0/libs/property_map/doc/ReadablePropertyMap.html) says that the return type of `get(pmap, key)` is `reference`. That means that, if `get(pmap, key)` does not return a reference type, then `reference` must be equal to `value_type` in the `boost::property_traits<PMap>`.

lrineau:
> The "valid expression" § says `reference`, but the full doc actually says: `The return type of the get() function is either the value_type of the property map or a (const or non-const) reference to that type.` (first line of the page).

Both are true. That means that the return type of the `get()` function, *that is named `reference`*  is either the value_type of the property map or a (const or non-const) reference to that type.



> The concept stack is a bit of a mess, it gets even worse in the lvalue pmaps with `operator[]` requiring `value_type&` or `const value_type&`, which is just confusing when the reference must be `const` or not depending on the mutability of the pmap.

I do not read the same as what you say. [`LvaluePropertyMap` (from the doc of Boost 1.77](https://www.boost.org/doc/libs/1_77_0/libs/property_map/doc/LvaluePropertyMap.html):
> The `LvaluePropertyMap` provides `operator[]` and function `get()` for accessing a reference to a value object. The return type refines that of function `get()` in `ReadablePropertyMap`; it can only be a reference (for a mutable `LvaluePropertyMap`) or a const reference (for a non-mutable `LvaluePropertyMap`). 

(I do not like the wording "const reference", because all references are const: they cannot be rebound to a different object. The correct wording is a "reference to const".)

Any, here the concept for lvalue pmap says that the `reference` type from the traits class must be a real reference, to a const or not const object (for mutable of not). The table in "Valid Expressions" says the same. Maybe we could discuss that IRL.

I do not know if Boost concepts have managed to converge to a correct support of C++20 concepts. But I think C++20 concepts could really benefit CGAL, for the testsuite.

> Anyway, I only put `value_type` when it is is identical to `reference`; it is just for readability.

I understand. I like that. But I would appreciate that our testsuite has a way to check that `reference` is equal to `value_type`, in that case. Hence the C++20 concept checking...

PR #6053, opened 2021-10-14
AABB_Tree: Use unique reference points for each primitive

MaelRL:
## Summary of Changes

The AABB tree uses the primitives' reference points at two different steps: in the primitive splitting and for the internal KD tree. In the case of face graphs, the reference point is a vertex of the descriptor, which might not be optimal: elements will often share the same reference point and so the splitting of the trees will not necessarily reflect the correct ordering of elements while sweeping in a given direction, and thus a non-optimal tree.

The changes in this PR are just to illustrate and not final. On some code with heavy use of the AABB tree, it yields up to 20% speed-ups. However, it can also cause a noticeable slow down if the usage of the AABB tree is light, especially if the reference point is not cached (e.g. uncached PMP::slicer).

Another important aspect is that when a reference point is taken as a vertex, it is an exact value. On the other hand, the reference points here are constructions, which are not so important for the primitive splitting (at worst, a non-optimal AABB is used) but are important for the internal kd tree. Maybe these two reference points could be distinguished.

@CGAL/geometryfactory 

## Release Management

* Affected package(s): `AABB_tree`
* Issue(s) solved (if any): -

### DIFF
--- a/AABB_tree/include/CGAL/AABB_primitive.h
+++ b/AABB_tree/include/CGAL/AABB_primitive.h
@@ -157,7 +157,7 @@ public:
   typename Base::Datum_reference
   datum() const { return get(m_obj_pmap,this->m_id); }
 
-  typename Base::Point_reference
+  decltype(auto)
   reference_point() const { return get(m_pt_pmap,this->m_id); }
 };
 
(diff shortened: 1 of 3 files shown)


sloriot:
@MaelRL I worked around the issue of duplicated points by adding a new function [here](https://github.com/CGAL/cgal/pull/9029/files#diff-e030ef6601b49232fdd55c18da209127e39416dab084942c6178713ca6070e52R471). You can give the range of vertices of the mesh vertices together with the vpm.

PR #6069, opened 2021-10-20
Add random generator np for sampling

### LINKED ISSUE #6067: Problems with pmp::sample_triangle_mesh()
## Issue Details

1. The documentation [1] lists the named parameters sample_vertices and sample_edges, but in fact they are called do_sample_vertices and do_sample_edges. (perhaps also for other parameters, e.g., sample_faces).
BW, it's nice that you can fold/unfold the description of a named parameter in the HTML reference manual of the function, but it is extremely annoying that you cannot cut & paste the string name of the named parameter.

2. I need uniform sampling at random, but I need to be able to produce the same output. So, I need to pass the seed (or something like that).
BW, I see that Random_points_in_triangle_mesh_3 is used. I think it's worth mentioning this; see below.

3. I suggest rephrasing:
If true is passed, points are generated in a random and uniform way on the surface of tm, and/or on edges of tm.
First, "uniformly generated at random" is more precise.
Second, I assume that, in the case of the faces, for example, (A) faces are selected first, and then (B) samples are selected on the selected closed faces.
This is not equivalent to "uniformly generated at random on the surface of tm, because each edge is counted twice.(Similar for edges: each vertex is counted more than once.)
Third, and/or is wrong,  so I suggest splitting into 2 phrases, for example:
```

If true is passed, then
    If sample_faces is true, points are generated uniformly at random on faces of tm (selected with probability proportional to their areas).
    If sample_edges is true, points are generated uniformly at random on edges of tm (selected with probability proportional to their length).
    See Random_points_in_triangle_mesh_3 for the precise sampling method.
```
[[1]](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__distance__grp.html#ga506b11b02a3b108ba97cccbf5d78a513)

## Environment

* Operating system: irrelevant
* Compiler: irrelevant
* Release or debug mode: both
* CGAL version: latest
* Boost version: irrelevant

sloriot:
Fixes #6067

### DIFF
--- a/BGL/include/CGAL/boost/graph/parameters_interface.h
+++ b/BGL/include/CGAL/boost/graph/parameters_interface.h
@@ -128,7 +128,6 @@ CGAL_add_named_parameter(match_faces_t, match_faces, match_faces)
 CGAL_add_named_parameter(face_epsilon_map_t, face_epsilon_map, face_epsilon_map)
 CGAL_add_named_parameter(maximum_number_t, maximum_number, maximum_number)
 CGAL_add_named_parameter(use_one_sided_hausdorff_t, use_one_sided_hausdorff, use_one_sided_hausdorff)
-CGAL_add_named_parameter(random_generator_f, random_generator, random_generator)
 
 // List of named parameters that we use in the package 'Surface Mesh Simplification'
 CGAL_add_named_parameter(get_cost_policy_t, get_cost_policy, get_cost)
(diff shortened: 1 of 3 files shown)


lrineau:
I think the feature should be tested in at least one .cpp file.

MaelRL:
Why not use the existing named parameter `random_seed` since this is a CGAL::Random in all cases?

sloriot:
Successfully tested in CGAL-5.4-Ic-131

PR #6096, opened 2021-10-29
Mention mt safety

sloriot:
We might want to put it at the top as it's a big announcement.

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -5,6 +5,11 @@ Release History
 
 Release date: December 2021
 
+### [General changes](https://doc.cgal.org/5.4/Manual/general_intro.html)
+-   The kernel providing exact constructions and exact predicates (`Exact_predicates_exact_constructions_kernel`)
+    has been made thread-safe. See changes in 2D and 3D Linear Geometry Kernel
+    for more details.
+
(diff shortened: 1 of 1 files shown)


Installation/CHANGES.md
```diff
@@ -41,6 +41,14 @@ Release date: December 2021
 
 -   Added documentation for the class `Projection_traits_3`, which enables the use of 2D algorithms on the projections of 3D data onto an arbitrary plane.
 
+-   Most operations on `Exact_predicates_exact_constructions_kernel` objects are now thread-safe
+    if `CGAL::Exact_rational` is `mpq_class` (from `GMPXX`), `boost::multiprecision::mpq_rational` or
+    `CGAL::Quotient<CGAL::MP_Float>`. The objects are not atomic though, so the usual restrictions
+    on avoiding race conditions apply. For users who do not use threads, this can be disabled with `CGAL_HAS_NO_THREADS`.
```
lrineau, line 47: This paragraph will probably be obscure for most users. We should restrict to the default for `Exact_predicates_exact_constructions_kernel`. Users do not really know or care what is the number type that is used to implement the kernel.

sloriot: This could be left as is and a note added in `General` referring to here for more details

Issue #6073, opened 2021-10-20
Ambiguous operator overload when dividing Gmpq with unsigned int


dpapavas:
Not sure if this is a bug, or simply unsupported, but I suspect the former, so I'm reporting it.  The following program fails to compile for me (against the current master branch):
```C++
#include <CGAL/Gmpq.h>

int main() {
    CGAL::Gmpq g;
    unsigned int n;
    g / n;
}
```

The compiler's output is the following:

```
/home/dimitris/source/sandbox/cgal_test/test.cpp: In function ‘int main()’:
/home/dimitris/source/sandbox/cgal_test/test.cpp:6:7: error: ambiguous overload for ‘operator/’ (operand types are ‘CGAL::Gmpq’ and ‘unsigned int’)
    6 |     g / n;
      |     ~ ^ ~
      |     |   |
      |     |   unsigned int
      |     CGAL::Gmpq
In file included from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/Gmp_coercion_traits.h:25,
                 from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/Gmpz.h:25,
                 from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/Gmpq.h:17,
                 from /home/dimitris/source/sandbox/cgal_test/test.cpp:1:
/home/dimitris/source/gc/cgal/Number_types/include/CGAL/GMP/Gmpq_type.h:368:1: note: candidate: ‘CGAL::Gmpq CGAL::operator/(const CGAL::Gmpq&, const CGAL::Gmpq&)’
  368 | operator/(const Gmpq &x, const Gmpq &y)
      | ^~~~~~~~
In file included from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/Interval_nt.h:46,
                 from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/long_double.h:27,
                 from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/number_type_basic.h:52,
                 from /home/dimitris/source/gc/cgal/Number_types/include/CGAL/Gmpq.h:16,
                 from /home/dimitris/source/sandbox/cgal_test/test.cpp:1:
/usr/include/boost/operators.hpp:269:1: note: candidate: ‘CGAL::Gmpq boost::operators_impl::operator/(const CGAL::Gmpq&, const int&)’
[...]

PR #6100, opened 2021-10-30
Fix Doo Sabin subdivision for Gmpq-based kernels

dpapavas:
This PR introduces a simple fix for Doo-Sabin subdivision, with regard to the issue described in #6073.  The problem is circumvented by simply switching a couple of `size_t` integers in the mask calculations for corner nodes to plain old (signed) `int`s.

I have tested the code with a `CGAL::Cartesian<CGAL::Gmpq>` kernel and it seems to be working fine.


sloriot:
Successfully tested in CGAL-5.4-Ic-87

PR #6102, opened 2021-10-31
Kernel_23: Add NonZeroDimension_3

afabri:
## Summary of Changes

Add a function object to the concept `Kernel` that enables to find any dimension of a 3D vector that is different from zero. 

- [x] Document the model

## Release Management

* Feature/Small Feature (if any): [link](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/NonZeroDimension_3)
* Link to compiled documentation:  [link](https://cgal.github.io/6102/v3/Kernel_23/classKernel_1_1NonZeroDimension__3.html)
* License and copyright ownership: not changed

### DIFF
--- a/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
+++ b/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
@@ -9138,7 +9138,7 @@ public:
 
   \cgalRefines `AdaptableFunctor` (with one arguments)
 */
-class NonZeroDimension_3
+class NonZeroCoordinateIndex_3
 {
 public:
 
--- a/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
+++ b/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
@@ -9147,10 +9147,10 @@ public:
   /// @{
 
   /*!
-    returns any of `0`, `1`, or `2`, if this dimension of vector `v` is not
-    equal to zero, and `-1` if `v` equals the null vector.
+    returns any of `0`, `1`, or `2` if the corresponding coordinate of the vector `v` is not
+    equal to zero, and `-1` if `v` is the null vector.
   */
-  int operator(const Kernel::Vector_3&v);
+  int operator(const Kernel::Vector_3& v);
 
   /// @}
 
--- a/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
+++ b/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
@@ -9139,7 +9139,7 @@ public:
   \cgalRefines `AdaptableFunctor` (with one arguments)
 */
 class NonZeroDimension_3
-
+{
   /// \name Operations
   /// A model of this concept must provide:
   /// @{
--- a/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
+++ b/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
@@ -9130,6 +9130,31 @@ public:
 }; /* end Kernel::LessZ_3 */
 
 
+
+
+/*!
+  \ingroup PkgKernel23ConceptsFunctionObjects
+  \cgalConcept
+
+  \cgalRefines `AdaptableFunctor` (with one arguments)
+*/
+class NonZeroDimension_3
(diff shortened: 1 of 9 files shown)


afabri:
@sloriot why is the added concept not on [this page](https://cgal.github.io/6102/v1/Kernel_23/group__PkgKernel23ConceptsFunctionObjects.html)?

Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
```diff
@@ -9130,6 +9130,31 @@ class LessZ_3 {
...
+
+
+/*!
+  \ingroup PkgKernel23ConceptsFunctionObjects
+  \cgalConcept
+
+  \cgalRefines `AdaptableFunctor` (with one arguments)
+*/
+class NonZeroDimension_3
+
```
sloriot, line 9142:
```suggestion
{
```

MaelRL:
Shouldn't this be called `Non_zero_coordinate_3` rather? Using "dimension" is abusing nomenclature, and it's confusing given that there exists `Vector_3::dimension() = 3`

lrineau:
> Shouldn't this be called `Non_zero_coordinate_3` rather? Using "dimension" is abusing nomenclature, and it's confusing given that there exists `Vector_3::dimension() = 3`

I agree. Here the input of the functor is a `Vector_3`, and `Non_zero_coordinate_3` seems a better naming.

I would understand "non zero dimension" if the input was something like a `Bbox_3`.

afabri:
Note that it does not return a coordinate but the index of the dimension.    Concerning the name I am wondering it it should be prefixed with `Any_`.

afabri:
What about  `Any_non_zero_coordinate_index_3`?

sloriot:
I don't like the any prefix.

afabri:
So, do we converge to `Non_zero_coordinate_index_3`?

sloriot:
fine with me

lrineau:
> So, do we converge to `Non_zero_coordinate_index_3`?

It seems a good name.

Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
```diff
@@ -9130,6 +9130,33 @@ class LessZ_3 {
...
+  /// \name Operations
+  /// A model of this concept must provide:
+  /// @{
+
+  /*!
+    returns any of `0`, `1`, or `2`, if this dimension of vector `v` is not
+    equal to zero, and `-1` if `v` equals the null vector.
+  */
+  int operator(const Kernel::Vector_3&v);
```
MaelRL, line 9153:
```suggestion
  /*!
    returns any of `0`, `1`, or `2` if the corresponding coordinate of the vector `v` is not
    equal to zero, and `-1` if `v` is the null vector.
  */
  int operator(const Kernel::Vector_3& v);
```

lrineau:
Interdependently from its documentation, about the implementation of `Non_zero_coordinate_index_3`, should not it return the index of the axis for which the vector has the biggest absolute value? That could increase the performance of the code using that index.

Let's consider we have a set of 3D points that are almost vertical, in the XZ plane. The fitting plane of that point set would have the orthogonal vector `v={small_x, big_y, small_z}`, where `small_x` and `small_z` have small, but non-zero, absolute values. Then, with the current implementation, `Non_zero_coordinate_index_3()(v)` might return `0`, because the value is small, but big enough that `certainly_not(is_zero(v.x()))` is true. In that case, the `bounded_side_2` would be called with a projection on the YZ plane, where the point set is almost degenerated, instead of the plane XZ. If the implementation of `Non_zero_coordinate_index_3` tries to return the index of the  largest coordinate, it would be `1`, instead of `0`.


Also, as I said in a private conversion with Andreas, the code should use `hx()`/`hy()`/`hz()`, instead of `x()`/`y()`/`z()`, so that it is as efficient in the homogeneous case as in the cartesian case.

afabri:
The main gain comes from the fact that we avoid filter failures.  I won't explore searching for the biggest for now. I will do the `hx()` etc.

lrineau:
This small feature is not yet approved, and yet the release 5.4-beta1 was due two weeks ago. I think this small feature is really late, and should not go into CGAL-5.4.

In particular, this PR conflicts with the preparation of `CHANGES.md`: https://github.com/CGAL/cgal/pull/6122.

sloriot:

PR #6114, opened 2021-11-09
Clean cmake scripts of examples and tests

sloriot:


Barycentric_coordinates_2/examples/Barycentric_coordinates_2/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 
 project(Barycentric_coordinates_2_Examples)
 
-cmake_minimum_required(VERSION 3.1...3.15)
+cmake_minimum_required(VERSION 3.1...3.20)
```
lrineau, line 6: We could "upgrade" to `3.1...3.22`, because we test with CMake 3.22 in our testsuite.

lrineau (approved):

Segment_Delaunay_graph_2/test/Segment_Delaunay_graph_2/CMakeLists.txt
```diff
@@ -1,14 +1,11 @@
...
 
-cmake_minimum_required(VERSION 3.1...3.20)
+cmake_minimum_required(VERSION 3.1...3.22)
 project(Segment_Delaunay_graph_2_Tests)
 
 find_package(CGAL REQUIRED COMPONENTS Core)
 
-include(${CGAL_USE_FILE})
-include_directories(BEFORE "include")
-
```
afabri, line 11: This leads to an error in the testsuite.

Arrangement_on_surface_2/test/Arrangement_on_surface_2/cgal_test.cmake
```diff
@@ -151,6 +151,7 @@ function(cgal_arr_2_add_target exe_name source_file)
     set(name ${exe_name}_${suffix})
   endif()
   add_executable(${name} ${source_file})
+  target_link_libraries(${name} CGAL::CGAL)
```
lrineau, line 154: I think you need to link with `CGAL::CGAL_Core` as well, because the `CMakeLists.txt` used:
```cmake
find_package(CGAL REQUIRED COMPONENTS Core)

include(${CGAL_USE_FILE})
```

Arrangement_on_surface_2/test/Arrangement_on_surface_2/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ enable_testing()
 
 find_package(CGAL REQUIRED COMPONENTS Core)
 
+include (${CGAL_USE_FILE})
```
lrineau, line 11: Have you found out why that is necessary?

Installation/test/Installation/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ endmacro()
 
 find_package(CGAL REQUIRED COMPONENTS Core)
 
+include(${CGAL_USE_FILE})
```
lrineau: Same here: have you found out why that is necessary?

Arrangement_on_surface_2/test/Arrangement_on_surface_2/CMakeLists.txt
sloriot: No I gave up. As I wrote in the commit message, I suspect it has to do with the `flags` variable that is not passed to cmake if the use-file is not included (maybe a scope issue?). Note that it is an issue for non cmake testsuite only. As it was in a test and not an example I considered that it could be fixed later (arrangement is a particular case anyway)

Installation/test/Installation/CMakeLists.txt
sloriot: After having spent so much time on arrangement I gave up on this one too (for now)

sloriot:

PR #6117, opened 2021-11-09
Remove -old option for Polyhedron_3 draw

### LINKED ISSUE #6099: Incorrect rendering of polyhedrons in CGAL::draw
I have an issue that is similar (probably the same) to the one described in #5691.  The problem seems to be caused by the `--old` argument passed [here](https://github.com/CGAL/cgal/blob/9543deb304f4a6adb1271b1fbdf7bbbfbda87ce2/Polyhedron/include/CGAL/draw_polyhedron.h#L50).

From what I could glean from the message and comments of the commit that introduced it (bace15874b12ef86f96e3f442e263283b823497c), this seems to have the intent of forcing a legacy GL context, but it's not what it actually does, as far as I can see.  From [here](https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/GraphicsView/include/CGAL/Qt/qglviewer_impl.h#L205-L216), it seems that it simply overrides the detection of the actually generated context and forces the use of shaders for an older context.  The result in my case, is that a 4.3 context is still created, but older shader are used.  This used to result in a bunch of shader compilation errors until recently when, after a driver upgrade or something of the kind, I simply get a white screen (which I suspect may be a rendering of a white polyhedron on a white background).

Removing the `--old` argument from `argv` and setting `argc` to 1 fixes the rendering for me, but since this is probably highly platform and hardware dependent, you might want to run a few tests of your own.  Let me know if you'd like me to create a PR for you to try out.

Edit:  forgot to add that this hack is not used for the `Nef_polyhedron_3` and `Surface_mesh` drawing code, which worked fine for me, so, unless you have reason to believe that Nef and mesh drawing may not be working for others, this would be an indication that we could simply do away with the `--old` flag altogether.

gdamiand:

## Summary of Changes

Remove -old parameter for Polyhedron_3 draw.

## Release Management

* Affected package(s): Polyhedron_3
* Issue(s) solved (if any): fix #6099

### DIFF
--- a/Polyhedron/include/CGAL/draw_polyhedron.h
+++ b/Polyhedron/include/CGAL/draw_polyhedron.h
@@ -46,8 +46,8 @@ void draw(const CGAL_POLY_TYPE& apoly,
   if (!cgal_test_suite)
   {
     CGAL::Qt::init_ogl_context(4,3);
-    int argc=2;
-    const char* argv[2]={"polyhedron_viewer","--old"};
+    int argc=1;
+    const char* argv[2]={"polyhedron_viewer","\0"};
     QApplication app(argc,const_cast<char**>(argv));
     SimpleFaceGraphViewerQt
       mainwindow(app.activeWindow(), apoly, title, nofill);
--- a/Polyhedron/include/CGAL/draw_polyhedron.h
+++ b/Polyhedron/include/CGAL/draw_polyhedron.h
@@ -47,7 +47,7 @@ void draw(const CGAL_POLY_TYPE& apoly,
   {
     CGAL::Qt::init_ogl_context(4,3);
     int argc=1;
-    const char* argv[2]={"polyhedron_viewer","\0"};
+    const char* argv[1]={"polyhedron_viewer"};
     QApplication app(argc,const_cast<char**>(argv));
     SimpleFaceGraphViewerQt
       mainwindow(app.activeWindow(), apoly, title, nofill);
--- a/Polyhedron/include/CGAL/draw_polyhedron.h
+++ b/Polyhedron/include/CGAL/draw_polyhedron.h
@@ -47,7 +47,7 @@ void draw(const CGAL_POLY_TYPE& apoly,
   {
     CGAL::Qt::init_ogl_context(4,3);
     int argc=1;
-    const char* argv[1]={"polyhedron_viewer"};
+    const char* argv[2]={"polyhedron_viewer", nullptr};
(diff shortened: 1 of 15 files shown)


Polyhedron/include/CGAL/draw_polyhedron.h
```diff
@@ -46,8 +46,8 @@ void draw(const CGAL_POLY_TYPE& apoly,
   if (!cgal_test_suite)
   {
     CGAL::Qt::init_ogl_context(4,3);
-    int argc=2;
-    const char* argv[2]={"polyhedron_viewer","--old"};
+    int argc=1;
+    const char* argv[2]={"polyhedron_viewer","\0"};
```
lrineau, line 50:
```suggestion
    const char* argv[2]={"polyhedron_viewer",""};
```

`""` is already a (null-terminated) string literal. See https://en.cppreference.com/w/cpp/language/string_literal.

lrineau: What is more: if `argc` is 1, then `argv` should be an array of size 1, and not 2.
```suggestion
    const char* argv[1]={"polyhedron_viewer"};
```

gdamiand: Thanks @lrineau ; changed in all viewers cf. 
ee11c85 (I don't remember where this code comes from).

(BTW it is really nice to be able to run an example without needing to add a link to the data directory ! Good job @sloriot )

lrineau:
Actually, I was partially wrong. That string literal `"\0"` was wrong, but now the code is wrong also. `argv` must be an array of type `argc+1`, and its last pointer must be `nullptr` (and not a pointer to a null string).

https://en.cppreference.com/w/cpp/language/main_function
> **`argv`:** Pointer to the first element of an array of `argc + 1` pointers, of which the last one is null and the previous ones, if any, point to null-terminated multibyte strings that represent the arguments passed to the program from the execution environment. If `argv[0]` is not a null pointer (or, equivalently, if` argc > 0`), it points to a string that represents the name used to invoke the program, or to an empty string.

gdamiand:
> Actually, I was partially wrong. That string literal `"\0"` was wrong, but now the code is wrong also. `argv` must be an array of size `argc+1`, and its last pointer must be `nullptr` (and not a pointer to a null string).

Will do asap.

lrineau (approved):

sloriot:

PR #6119, opened 2021-11-10
Add a triangle AABB tree which can use multiple primitives to cover a datum

MaelRL:
## Summary of Changes

Big triangles can really mess with the balance of a tree, and result in awkward traversals. This PR adds a new AABB tree (for triangles, but it could be something more generic) where multiple primitives cover a single datum: their bboxes are smaller than the bbox of the datum and the union of the bboxes of the primitives representing a datum contains the bbox of the datum. The datum is the same for all these primitives, but specific traversal traits are provided to avoid doing superfluous "intersection(query, datum)" calls while traversing. See the test for usage.

The splitting criterion is currently an edge length bound. It could be something different; a better understanding of when a covering of a datum is useful (the anisotropy of a triangle seems more important than the size of the triangle, for example).

## Release Management

* Affected package(s): `AABB_tree`
* Issue(s) solved (if any): -
* Feature/Small Feature (if any): undocumented

### DIFF
--- a/AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
+++ b/AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
@@ -1,0 +1,495 @@
+// Copyright (c) 2021 INRIA Sophia-Antipolis (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org).
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s) : Mael Rouxel-Labbé
+
+#ifndef CGAL_AABB_TREE_INTERNAL_TRIANGLE_DATUM_COVERING_H
+#define CGAL_AABB_TREE_INTERNAL_TRIANGLE_DATUM_COVERING_H
+
+#include <CGAL/license/AABB_tree.h>
+
+#include <CGAL/Simple_cartesian.h>
+#include <CGAL/Cartesian_converter.h>
+
+#include <CGAL/AABB_tree/internal/AABB_traversal_traits.h>
+#include <CGAL/AABB_primitive.h>
+#include <CGAL/AABB_traits.h>
+#include <CGAL/AABB_tree.h>
+#include <CGAL/array.h>
+#include <CGAL/Bbox_3.h>
+#include <CGAL/Container_helper.h>
+#include <CGAL/property_map.h>
+
+#include <queue>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+namespace CGAL {
+namespace AABB_trees {
+namespace internal {
+
+// std::vector to property map
+// Not using Pointer_property_map because the underlying range is empty initially and can change
+template <typename T>
+struct Vector_property_map
+{
+  using Range = std::vector<T>;
+
+  using key_type = std::size_t;
+  using value_type = T;
+  using reference = value_type&;
+  using category = boost::read_write_property_map_tag;
+
+  Vector_property_map() : m_range_ptr(nullptr) { } // for compilation only
+  Vector_property_map(Range* range_ptr) : m_range_ptr(range_ptr) { }
+
+  inline friend void put(const Vector_property_map& map, const key_type& k, const value_type& v)
+  {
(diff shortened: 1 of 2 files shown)


AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
```diff
@@ -0,0 +1,495 @@
...
+//                    << "Point: (" << m_ref_points.back() << ") "
+//                    << "Datum: [" << m_data.back() << "]" << std::endl;
+
+      Primitive p(id/*, m_dpm, m_rppm*/); // pmaps are external, shared data
+      this->insert(p);
+      return;
+    }
+
+    K2AK k2ak;
+    AK2K ak2k;
```
sloriot, line 396: `/srv/CGAL/www/Members/testsuite/CGAL-5.4-Ic-118/AABB_tree/TestReport_lrineau_Ubuntu-Latest-CXX11.gz:/mnt/testsuite/include/CGAL/AABB_tree/internal/triangle_datum_covering.h:396:10: warning: unused variable 'ak2k' [-Wunused-variable]`

sloriot:
One warning in `AABB_tree` of CGAL-5.4-Ic-118

sloriot:

AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
```diff
@@ -0,0 +1,493 @@
...
+                                         CGAL::Tag_false /*no caching*/>;
+
+  using AABB_geom_traits = Kernel;
+  using AABB_traits = CGAL::AABB_traits<AABB_geom_traits, Primitive, BPM>;
+  using AABB_tree = CGAL::AABB_tree<AABB_traits>;
+};
+
+template <typename Kernel,
+          typename Point = typename Kernel::Point_3>
+struct AABB_covered_triangle_tree
```
sloriot, line 264: I don't like inheritance. In principle all could be done in a free function with the extra data being stored as shared data in the traits. This is just a suggestion, not an implementation request.

sloriot: Also, we can add a new `insert()` function to the AABB-tree than would not blindly take the input range but process it before doing the real insert. This could be use to augment that range or to reduce it (think about removing degenerate faces for example or filtering out some elements)

AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
```diff
@@ -0,0 +1,493 @@
...
+    Projection_traits(const Point& hint,
+                      const typename Primitive::Id& hint_primitive,
+                      const AABBTraits& traits)
+      : Base(hint, hint_primitive, traits)
+    { }
+
+    void intersection(const Point& query, const Primitive& primitive)
+    {
+      // check a datum only once
+      auto is_insert_successful = visited_data.insert(primitive.id().second/*unique input face ID*/);
```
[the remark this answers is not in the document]
MaelRL: See e.g. https://github.com/CGAL/cgal/blob/master/AABB_tree/include/CGAL/AABB_tree.h#L968, but there is indeed a risk.

sloriot:
Compilation error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-139/AABB_tree/TestReport_Sosno_MSVC-2022-Preview-Release.gz)

sloriot:

sloriot:
@MaelRL , you have a TODO that will prevent this PR from being merged.

lrineau:
> @MaelRL , you have a TODO that will prevent this PR from being merged.

Mael, do you still consider this PR as a work in progress?

MaelRL:
Yes, the current implementation is bugged. I have the fix in another branch, and I need to port it here.

sloriot:
Successfully tested in CGAL-5.5-Ic-52

AABB_tree/include/CGAL/AABB_tree/internal/triangle_datum_covering.h
lrineau: This discussion is not marked as resolved. @MaelRL @sloriot is that still something you want to discuss?

PR #6126, opened 2021-11-12
Mesh distance: use spatially sorted sample points

dthul:
## Summary of Changes

The Hausdorff distance code makes a copy of the provided sample points and sorts them spatially such that the hints provided to the AABB tree will be good. There is an apparent mistake in the code though and instead of the spatially sorted sample points, the original (unsorted) ones are given to the AABB tree. The sorted ones are in fact never used.
This PR changes the code such that the sorted points are used.

## Release Management


sloriot:

PR #6132, opened 2021-11-16
Create normal map only if normal estimation is done

sloriot:
Create normal map only if normal are estimated.

@lrineau `MainWindow::filterOperations(bool)` should be called once the normals are estimated as normal orientation should then become applicable. But it is not accessible from the plugin. If I manually unselect and select the item,  `MainWindow::filterOperations(bool)` is called but a call to `Scene::setSelectedItem(int)` does not help. Any idea or do I need to modify an interface?

### DIFF
--- a/Polyhedron/demo/Polyhedron/MainWindow.cpp
+++ b/Polyhedron/demo/Polyhedron/MainWindow.cpp
@@ -228,6 +228,9 @@ MainWindow::MainWindow(const QStringList &keywords, bool verbose, QWidget* paren
           this, SLOT(updateInfo()));
 
   connect(scene, SIGNAL(dataChanged(const QModelIndex &, const QModelIndex & )),
+          this, SLOT(filterOperations()));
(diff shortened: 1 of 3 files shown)


lrineau:
I think there are three options:
  - the MainWindow class could be modified: connect the SIGNAL `scene, SIGNAL(dataChanged(const QModelIndex &, const QModelIndex & )` to `filterOperations`,
  - or, in the plugin that needs access to the main window, cast the pointer `mw`, to the class `MainWindow` to call the function directly.
  - or, as you said, as a new function to an existing interface.

I prefer the first option, if that does not impact the performance of the demo.

sloriot:
Successfully tested in CGAL-5.4-Ic-100

sloriot:
But I still need to implement the signal connection.

sloriot:
Successfully tested in CGAL-5.4-Ic-102

PR #6144, opened 2021-11-22
Fix conversion error in METIS wrappers

### LINKED ISSUE #5410: METIS partition graph type conversion error
# Issue Details
In [partition_graph()](https://doc.cgal.org/latest/BGL/group__PkgBGLPartition.html#gae7e37b6101e63f4c5959f1ae6c6fe7d2) function that we can found in `CGAL-5.1.1/include/CGAL/boost/graph/METIS/partition_graph.h` the `nparts` parameter is an `int`.
Inside partition_graph(), the METIS_PartMeshNodal() function (from `metis.h`) is called, and it take `nparts` as parameter.

The problem is that in METIS_PartMeshNodal(), `nparts` is of type `idx_t*` and not `int*`.
`idx_t` is a typedef for `long long` so it doesn't seem to me that conversion should be a problem, but VisualStudio throw me this error: 

```
'int METIS_PartMeshNodal(idx_t *,idx_t *,idx_t *,idx_t *,idx_t *,idx_t *,idx_t *,real_t *,idx_t *,idx_t *,idx_t *,idx_t *)': cannot convert argument 7 from 'int *' to 'idx_t *'
```

Somebody know why he don't want to convert an `int*` to an `long long *` ?

# Environment
Operating system: Windows 10, 64 bits
Compiler: MSVC 14.2
Release or debug mode: Release
CGAL version: 5.1.1
Boost version: 1.74.0
Eigen version: 3.3.9
Metis version: 5.1.0

MaelRL:
## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): fix #5410

### DIFF
--- a/BGL/include/CGAL/boost/graph/METIS/partition_dual_graph.h
+++ b/BGL/include/CGAL/boost/graph/METIS/partition_dual_graph.h
@@ -79,6 +79,7 @@ void partition_dual_graph(const TriangleMesh& tm,
 
   // a dual edge between elements exists if they share 'nparts' vertices
   idx_t ncommon = 2;
+  idx_t nparts_as_idx_t = nparts;
 
   // either the edgecut or the total communication volume of the dual graph’s partitioning
   idx_t objval;
@@ -98,7 +99,7 @@ void partition_dual_graph(const TriangleMesh& tm,
   CGAL_assertion_code(int ret =)
     METIS_PartMeshDual(&ne, &nn, eptr, eind,
                        nullptr /* elements weights*/, nullptr /*elements sizes*/,
-                       &ncommon, static_cast<idx_t*>(&nparts),
+                       &ncommon, &nparts_as_idx_t,
                        nullptr /* partitions weights */,
                        *options,
                        &objval, epart, npart);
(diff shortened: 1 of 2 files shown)


BGL/include/CGAL/boost/graph/METIS/partition_dual_graph.h
```diff
@@ -98,7 +98,7 @@ void partition_dual_graph(const TriangleMesh& tm,
   CGAL_assertion_code(int ret =)
     METIS_PartMeshDual(&ne, &nn, eptr, eind,
                        nullptr /* elements weights*/, nullptr /*elements sizes*/,
-                       &ncommon, &nparts,
+                       &ncommon, static_cast<idx_t*>(&nparts),
```
lrineau, line 101: That cannot work. The variable `nparts` is a parameter of the function, of type `int`, on the stack. If `idx_t` and `int` have different size (and they seem so), then reading `nparts` through a point to `long long` will add 32 bits to the type, and whatever is in memory in those 32 bits (4 octets) will be added to the value as the 32 most significant bits.

The right solution is to create a real variable of the right type, with the right value:
```c++
idx_t nparts_as_idx_t = nparts;
// [...]
METIS_PartMeshDual(&ne, &nn, eptr, eind,
                       nullptr /* elements weights*/, nullptr /*elements sizes*/,
                       &ncommon, &nparts_as_idx_t,
                       nullptr /* partitions weights */,
                       *options,
                       &objval, epart, npart);
```

lrineau (approved):

BGL/include/CGAL/boost/graph/METIS/partition_dual_graph.h
MaelRL: Right, thank you.

sloriot:
Successfully tested in CGAL-5.4-Ic-100

PR #6145, opened 2021-11-23
Fix collapse_edge bug when both edges to remove are constrained edges

sloriot:
also added a test for the function testing all possible branches.


sloriot:
Successfully tested in CGAL-5.4-Ic-102

PR #6149, opened 2021-11-25
Remove some unintended artifacts from the documentation

albert-github:
Some parts were due to to the markdown handling of 4 spaces meaning code converted to code and this was not the intention.


sloriot:
Successfully in CGAL-5.4-Ic-103. I manually looked at the output and did observed the fixes.

PR #6152, opened 2021-11-29
Fix not clearing mesh after failing to read

MaelRL:
## Summary of Changes

`CGAL::IO::read_polygon_mesh()` does not clear the mesh before writing, and does not clear the mesh if it fails to read.

In the demo, we try to read via `CGAL::IO::read_polygon_mesh()` first, and if that fails through the `read_polygon_soup -> repair -> orient -> soup_to_mesh` pipeline. However, the mesh needs to be cleared before calling that pipeline.

## Release Management

* Issue(s) solved (if any): -


sloriot:
Shouldn't it target 5.3?

sloriot:
Successfully tested in CGAL-5.4-Ic-106

PR #6159, opened 2021-12-07
Apply the operations per patches

sloriot:
This will workaround issue with boundary non-manifold vertices


sloriot:
5.4 milestone is fine (no need to be in the beta).

sloriot:

PR #6166, opened 2021-12-10
Convex_decomposition_3: Address warning issued by clang

### LINKED ISSUE #6165: Convex_decomposition: puzzled about a warning
## Issue Details

In [this testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-I-107/Convex_decomposition_3/TestReport_gimeno_ArchLinux-clang-CXX17-Release.gz) the compiler warns that `M`  is not used. I later see `M *= D;`  so why do we get the warning?  Can it be that the `gcd()`function is called somewhere with a constant as argument and the compiler then knows that there is an early `return`? 

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): archlinux
* Compiler: clang
* Release or debug mode:
* CGAL version:  CGAL-5.4-I-107
* Boost version:

afabri:
## Summary of Changes

Protect code that gets only used when tracing is switched on.

## Release Management

* Issue(s) solved (if any): fix #0000, fix #6165


sloriot:

PR #6188, opened 2021-12-17
Remove useless moc include

sloriot:
needed only if Q_OBJECT or other macros are in the file


sloriot:

PR #6193, opened 2021-12-22
Fix shortcut composition

sloriot:
Fix bug introduced in #6019 (cherry-picked in 5.3.x)

Seems that I did not understand correctly the doc here:
https://doc.qt.io/qt-5/qkeysequence.html#QKeySequence-2


sloriot:

PR #6195, opened 2021-12-23
Add visitor function call for tracking faces split

sloriot:
modification on an internal function


sloriot:

PR #6207, opened 2021-12-27
Fix set but unused variable warning

sloriot:


sloriot:
Successfully tested in CGAL-5.4-Ic-125 (+manually checked that the warnings are gone)

PR #6209, opened 2021-12-27
Use features available in std instead of boost

sloriot:
I was chasing a warning but then realized that it cannot be fixed but here is some of the changes that still could be interesting to keep.

I wonder if there is anything that prevents us from always using `std::unordered` instead of `boost::unordered`.

**TODO:**
- [x] Understand why 2cc2fa fixes runtime issues. Bug in MSVC's STL? (see https://github.com/CGAL/cgal/pull/6348)

### DIFF
--- a/Heat_method_3/include/CGAL/Heat_method_3/internal/Intrinsic_Delaunay_triangulation_3.h
+++ b/Heat_method_3/include/CGAL/Heat_method_3/internal/Intrinsic_Delaunay_triangulation_3.h
@@ -32,8 +32,8 @@
 
 
 #include <boost/iterator/transform_iterator.hpp>
-#include <boost/unordered_map.hpp>
 
+#include <unordered_map>
 #include <set>
 #include <stack>
 #include <cmath>
@@ -485,7 +485,7 @@ private:
   std::vector<int> mark_edges;
 public:
 
-  boost::unordered_map<vertex_descriptor,vertex_descriptor> v2v, vtov;
+  std::unordered_map<vertex_descriptor,vertex_descriptor> v2v, vtov;
 };
 
 } // namespace Heat_method_3
--- a/Heat_method_3/include/CGAL/Heat_method_3/internal/Intrinsic_Delaunay_triangulation_3.h
+++ b/Heat_method_3/include/CGAL/Heat_method_3/internal/Intrinsic_Delaunay_triangulation_3.h
@@ -37,6 +37,7 @@
 #include <set>
 #include <stack>
 #include <cmath>
+#include <list>
 
 #ifndef DOXYGEN_RUNNING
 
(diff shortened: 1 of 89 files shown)


afabri:
> I wonder if there is anything that prevents us from always using `std::unordered` instead of `boost::unordered`.

Isn't it just legacy? At a time it was not avaible in the standard template library.

janetournois:
There are some key types for which `std::unordered` does not implement the hash function, and `boost::unordered` does

sloriot:
Successfully tested in  CGAL-5.4-Ic-131

Heat_method_3/include/CGAL/Heat_method_3/internal/Intrinsic_Delaunay_triangulation_3.h
```diff
@@ -37,6 +37,7 @@
 #include <set>
 #include <stack>
 #include <cmath>
+#include <list>
```
lrineau, line 40: That is used by:
```
  typedef typename std::stack<edge_descriptor, std::list<edge_descriptor> > edge_stack;
```
Should not that stack use `std::deque` instead of a list?!

sloriot: The question is for @janetournois

janetournois: `std::deque` is the default, so I wonder why we would have chosen something else in the first place...

janetournois: It's been there for a while, and using `git blame` does not help much

janetournois:
Changing this map back to `boost::unordered_map` fixes the test on my platform - though we still need to figure out why

sloriot:
Successfully tested in CGAL-5.5-Ic-25

lrineau:
@sloriot You marked this PR with both https://github.com/CGAL/cgal/labels/TODO and https://github.com/CGAL/cgal/labels/under%20discussion. Do you now consider the pull-request to be mergable?

sloriot:
I think we should understand why reverting https://github.com/CGAL/cgal/pull/6209/commits/2cc2fab4a28a81ce971f983b6d89a13f1412fee6 would break the code. It might be a real issue for other packages but we simply  did not notice it. Maybe @afabri can help me here to understand why it breaks.

afabri:
I can reproduce it.  It fails with `test_04.off`.    My guess is that when we iterate over a map you get the elements in another order.

sloriot:
The autoref code was trying to do more than it should. I updated the code to does not try to fix what it can't in #6348

lrineau:
> The autoref code was trying to do more than it should. I updated the code to does not try to fix what it can't in #6348

Should not this PR (#6209) and #6348 be merged? Or at least, #6348 should be merged into this one.

sloriot:
They are in the same batch of test, they will be merged together

sloriot:
Some `boost::unordered` are still present in T3 based packages (Mesh_3, TDS_3, Periodic, ...) but I don't have time to fix them. So except if someone wants to do it, I'd like to merge this PR as is.

sloriot:
Successfully tested in CGAL-5.5-Ic-27

PR #6210, opened 2021-12-28
CTest Testsuite: resilience to non-UTF8 characters

lrineau:
## Summary of Changes

CTest Testsuite: resilience to non-UTF8 characters, use the method `backslashreplace` to deal with non-UTF8 characters.

See https://docs.python.org/3/library/codecs.html#error-handlers

## Release Management

* Affected package(s): Testsuite


sloriot:
Succesfully tested in CGAL-5.4-Ic-125 (fixes MSVC-2022-Community-Release | Sosno)

PR #6213, opened 2022-01-03
fix \cond \endcond mismatches

sloriot:


sloriot:
Successfully tested in CGAL-5.4-Ic-129 (warnings with doxygen master are gone)

PR #6219, opened 2022-01-05
Allow for more situations

sloriot:


sloriot:
@joachim-reichel @lrineau while at it, where goes the data directory when installing the example/demo package on Debian and Fedora?

joachim-reichel:
In the Debian packages there are three archives at
```
/usr/share/doc/libcgal-dev/data.tar.gz
/usr/share/doc/libcgal-dev/demo.tar.gz
/usr/share/doc/libcgal-dev/examples.tar.gz
```
that contain the content of these subdirectories from the tarball. The Debian policy is to use archives for larger conent (IIRC to avoid cluttering the file system with thousands of small files) -- the users have to copy/extract them anyway since they are read-only in that location. There is also a short README.Debian in that directory that explains how to build the examples/demos.

sloriot:
Successfully tested in CGAL-5.4-Ic-131

PR #6222, opened 2022-01-07
fix warnings

sloriot:
@lrineau please review changes in Surface_mesher.

### DIFF
--- a/Surface_mesher/include/CGAL/Surface_mesher/Surface_mesher.h
+++ b/Surface_mesher/include/CGAL/Surface_mesher/Surface_mesher.h
@@ -298,6 +298,7 @@ namespace CGAL {
 
           for(bool exit = false; ; exit = true)
           {
+            CGAL_USE(exit);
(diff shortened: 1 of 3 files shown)


Surface_mesher/include/CGAL/Surface_mesher/Surface_mesher.h
```diff
@@ -298,6 +298,7 @@ namespace CGAL {
 
           for(bool exit = false; ; exit = true)
           {
+            CGAL_USE(exit);
```
lrineau, line 301: That loop `for(bool exit = false; ; exit = true)` was a bad trick to execute the body exactly twice. But the `return false` line 324 is no longer conditional, so the loop will be executed only once. I agree with the patch, because I do not want to do refactoring in Surface_mesher.

lrineau (approved):

sloriot:
The warnings I was trying to fix are gone in CGAL-5.4-Ic-136. Successfully tested.

PR #6225, opened 2022-01-07
Fix infinite loop

sloriot:
Fix infinite loop in case no proxy could be added  error threshold is not yet reached


sloriot:
Successfully tested in CGAL-5.4-Ic-136

PR #6243, opened 2022-01-17
Tet remeshing - fix warning

janetournois:
## Summary of Changes

Warning seen in the testsuite
```
/mnt/testsuite/include/CGAL/Tetrahedral_remeshing/internal/compute_c3t3_statistics.h:50:10: warning: variable 'total_angle' set but not used [-Wunused-but-set-variable]
```

computing the average angle does not make much sense so let's just remove it

## Release Management

* Affected package(s): Tetrahedral remeshing


sloriot:
Warning is gone in CGAL-5.5-Ic-1

PR #6244, opened 2022-01-17
Remove conditional code as the flag has disappeared

sloriot:
Flag removed in #5904


sloriot:
Successfully tested in CGAL-5.5-Ic-1

PR #6246, opened 2022-01-18
Voroni testsuite

afabri:
## Summary of Changes

As the errors are related to IO my suspicion is that the simultaneous execution of two tests uses the same file. 
So I add a random number to the file name.  Or should we better determine the PID?

## Release Management

* Affected package(s): Voronoi testsuite


lrineau:
> As the errors are related to IO my suspicion is that the simultaneous execution of two tests uses the same file.

Note that it can also be solved at the CMake level, like here:
https://github.com/CGAL/cgal/blob/ad08f020b671295e95807ccc7cfc787236beb03b/Mesh_3/test/Mesh_3/CMakeLists.txt#L147-L152

Tests sharing the same non-empty value for [`RESOURCE_LOCK`](https://cmake.org/cmake/help/latest/prop_test/RESOURCE_LOCK.html) will not be executed concurrently.

lrineau:
Replaced by #6252.

PR #6247, opened 2022-01-18
Installation:  Fix issues for CGAL_WARNING

afabri:
## Summary of Changes

The warning must be put in quotes.
The check for gmp.h does not work for VC2015 so we have to disable it.

This PR fixes [this warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.4-Ic-141/AABB_tree_Demo/TestReport_Christo_MSVC2015-Release-64bits.gz)

## Release Management

* Affected package(s): Installation   and the warning shows up in AABB Tree

### DIFF
--- a/Installation/include/CGAL/Installation/internal/enable_third_party_libraries.h
+++ b/Installation/include/CGAL/Installation/internal/enable_third_party_libraries.h
@@ -25,13 +25,13 @@
 #  undef CGAL_USE_GMP
 #endif
 
-#if defined(__has_include)
+#if defined(__has_include) && ( ! defined _MSC_VER || _MSC_VER > 1900)
 #  if CGAL_USE_GMP && ! __has_include(<gmp.h>)
-#    pragma CGAL_WARNING(<gmp.h> cannot be found. Less efficient number types will be used instead. Define CGAL_NO_GMP=1 if that is on purpose.)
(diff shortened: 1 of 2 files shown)


sloriot (approved):

Installation/include/CGAL/Installation/internal/enable_third_party_libraries.h
```diff
@@ -25,13 +25,13 @@
 #  undef CGAL_USE_GMP
 #endif
 
-#if defined(__has_include)
+#if defined(__has_include) && ( ! defined _MSC_VER || _MSC_VER > 1900)
```
lrineau, line 28: What is special about MSVC 2015? It support `#pragma message` as well, according to the documentation:
https://docs.microsoft.com/en-us/cpp/preprocessor/message?view=msvc-140

sloriot: We tested that with @janetournois and the macro does not return what it should.

sloriot: on MSVC2015 only

lrineau: What is also shocking is that `defined(__has_include)` should evaluate to false, according to [this page](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-140), because `__has_include` was only supported since VS 2017 15.3.

sloriot:
Warning is gone in CGAL-5.5-Ic-1

PR #6258, opened 2022-01-19
fix post condition

sloriot:


lrineau:
You did not say: I imagine the code was first introduced in 5.5.x?

sloriot:
No but it does not really matter as it is only used internally

sloriot:

PR #6259, opened 2022-01-19
Mesh_3: Fixes for VC++ with option /permissive-

afabri:
## Summary of Changes

With the option `/permissive-` some of the tests of `Mesh_3` do not compile. 
I could boil it down to this [gist](https://gist.github.com/afabri/0416bebec1c32fb4efd6632446698972).
The fix is to move the counter in the base class and add a function to increment it.

## Release Management

* Affected package(s): Mesh_3

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/Mesh_global_optimizer.h
+++ b/Mesh_3/include/CGAL/Mesh_3/Mesh_global_optimizer.h
@@ -106,7 +106,16 @@ protected:
   Lock_data_structure *get_lock_data_structure() { return 0; }
   void unlock_all_elements() {}
 
+
+  // Workaround for problem with VC and /permissive
+  // See: https://gist.github.com/afabri/0416bebec1c32fb4efd6632446698972
+  void increment_frozen_points() const
+  {
+    ++nb_frozen_points_;
+  }
+
 protected:
+  mutable unsigned int nb_frozen_points_;
   std::size_t big_moves_size_;
   std::multiset<FT> big_moves_;
 };
@@ -116,6 +125,7 @@ protected:
 template <typename Tr>
 class Mesh_global_optimizer_base<Tr, Parallel_tag>
 {
+
 protected:
   typedef typename Tr::Geom_traits                          Gt;
   typedef typename Gt::FT                                   FT;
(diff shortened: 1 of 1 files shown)


afabri:
Fails in some [testsuites](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-2/Mesh_3/TestReport_lrineau_Ubuntu-Latest-CXX11.gz)
with the error message
```
Ubuntu-Latest-CXX11/test/Mesh_3/test_mesh_3_issue_1554.cpp:69:81:   required from here
/mnt/testsuite/include/CGAL/Mesh_3/Mesh_global_optimizer.h:203:57: error: use of deleted function 'std::atomic<unsigned int>::atomic(const std::atomic<unsigned int>&)'
  203 |   mutable std::atomic<unsigned int> nb_frozen_points_ = 0
```

janetournois (approved):

Mesh_3/include/CGAL/Mesh_3/Mesh_global_optimizer.h
```diff
@@ -200,7 +200,7 @@ class Mesh_global_optimizer_base<Tr, Parallel_tag>
 public:
 
 protected:
-  mutable std::atomic<unsigned int> nb_frozen_points_ = 0;
```
lrineau: A note for a next time, and for education. The original code was:

```c++
std::atomic<unsigned int> nb_frozen_points_ = 0;
```
That code would be accepted in C++17, with the copy-elision mandated by C++17. In C++17, that code is equivalent to:
```c++
std::atomic<unsigned int> temporary_var{0}; // direct initialization of a temporary variable
std::atomic<unsigned int> nb_frozen_points_{temporary_var}; // direct initialization using the copy-constructor
```
and that does not compile because `std::atomic` has no copy-constructor.

As you can see, you can also initialize it directly, with either parenthesis:
```c++
mutable std::atomic<unsigned int> nb_frozen_points_(0);
```
or braces:
```c++
mutable std::atomic<unsigned int> nb_frozen_points_{0u};
```

sloriot:
Errors are gone in CGAL-5.5-Ic-3

PR #6261, opened 2022-01-20
Intersections_3 :   Use Coercion_traits

afabri:
## Summary of Changes

This PR fixes a problem with the GMP backend in `boost::multiprecision`. 

The failure of the [Intersections_3 testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-2/Intersections_3/TestReport_Sosno_MSVC-2022-Preview-Release.gz) for VC++ comes from the error in  `operator-()`.    The type `CGAL::Exact_rational` is `boost::multiprecision::number<struct boost::multiprecision::backends::gmp_rational,1>`  on this platform and the boost version is 1_78.       When I use `CGAL::Gmpq` directly or if I use `boost::multiprecision::cpp_rational` there is no problem.  

The `operator-()`  is used [here](https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Segment_3_do_intersect.h#L197).    

In the intersection code we  now  construct a `CFT` from the `double`  before performing the subtraction. 

I isolated this into a testcase in the  `Number_types` package, but didn't put it into the `CMakeLists.txt` as it would break the testsuite.  I made  a [bug report](https://github.com/boostorg/multiprecision/issues/415) to boost.org. 

## Release Management

* Affected package(s):  Intersections_3

### DIFF
--- a/Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Segment_3_do_intersect.h
+++ b/Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Segment_3_do_intersect.h
@@ -173,11 +173,11 @@ do_intersect_bbox_segment_aux(const FT& px, const FT& py, const FT& pz,
     }
     else
     {
-      tmax = bxmax - px;
+      tmax = CFT(bxmax) - px;
       dmax = qx - px;
     }
 
-    tmin = bxmin - px;
+    tmin = CFT(bxmin) - px;
     dmin = qx - px;
   }
   else
@@ -194,11 +194,11 @@ do_intersect_bbox_segment_aux(const FT& px, const FT& py, const FT& pz,
     }
     else
     {
-      tmax = px - bxmin;
+      tmax = px - CFT(bxmin);
       dmax = px - qx;
     }
 
-    tmin = px - bxmax;
+    tmin = px - CFT(bxmax);
     dmin = px - qx;
   }
 
@@ -246,11 +246,11 @@ do_intersect_bbox_segment_aux(const FT& px, const FT& py, const FT& pz,
     }
     else
     {
-      tymax = bymax - py;
+      tymax = CFT(bymax) - py;
       dymax = qy - py;
     }
 
-    tymin = bymin - py;
+    tymin = CFT(bymin) - py;
     dymin = qy - py;
   }
   else
@@ -267,11 +267,11 @@ do_intersect_bbox_segment_aux(const FT& px, const FT& py, const FT& pz,
     }
     else
     {
-      tymax = py - bymin;
+      tymax = py - CFT(bymin);
       dymax = py - qy;
     }
 
-    tymin = py - bymax;
+    tymin = py - CFT(bymax);
     dymin = py - qy;
   }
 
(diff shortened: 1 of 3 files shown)


sloriot:

sloriot:
@afabri do you want to address Marc's comment?

afabri:
If I understood @MaelRL correctly the change I made is the right one which he made in other intersection functions as well,

mglisse:
> If I understood @MaelRL correctly the change I made is the right one which he made in other intersection functions as well,

I see the function starts with a test `px >= bxmin` so we already assume that FT and BFT are ImplicitInteroperable, meaning that we are allowed to compute `bxmax - px` without a cast. It is not clear where the `double` in `Coercion_traits<double,FT>` comes from, is `BFT` always `double`?

(I am not at all trying to block the PR)

lrineau:
I have merged the pull-request anyway. We have time before CGAL-5.5 to improve the code if necessary.

afabri:
> I see the function starts with a test `px >= bxmin`

I must admit that it also puzzled me that  the comparison just works fine.  Do you think we should create the FTs from the doubles?

MaelRL:
The BFT template exists because it can be double for a `CGAL::Bbox_3` or FT for an `Isocuboid_3`, e.g. from the `do_intersect` for a ray and an isocuboid:
```
template <class K>
bool do_intersect(const typename K::Ray_3& ray,
                  const typename K::Iso_cuboid_3& ic,
                  const K&)
{
  typedef typename K::FT FT;
  typedef typename K::Point_3 Point_3;

  const Point_3& source = ray.source();
  const Point_3& point_on_ray = ray.second_point();

  return do_intersect_bbox_segment_aux
      <FT,FT,
      true,  // bounded at t=0
      false,  // bounded at t=1
      false> // do not use static filters
      (
        source.x(), source.y(), source.z(),
        point_on_ray.x(), point_on_ray.y(), point_on_ray.z(),
        (ic.min)().x(), (ic.min)().y(), (ic.min)().z(),
        (ic.max)().x(), (ic.max)().y(), (ic.max)().z()
        );
}
```

It seems wrong to have `double` explicit, and it should be traits between `BFT` and `FT`, like done there: https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Line_3_do_intersect.h#L45.

mglisse:
> I must admit that it also puzzled me that the comparison just works fine. Do you think we should create the FTs from the doubles?

I don't know what requirement is documented. If FT and BFT are ImplicitInteroperable (do we have any field type that is not ImplicitInteroperable with double in CGAL?), comparisons and `bxmax - px` are fine, changing anything is a pessimization. If FT and BFT are only ExplicitInteroperable, I guess we do need to cast even for the comparison. Note that we should probably cast both operands before subtracting, not just one. Spelling the cast `Type(x)` has the disadvantage that it always copies when FT==BFT. Using `Coercion_traits::Cast` would be nicer because I think it could be changed to return a const reference. There is also the possibility of relying on "A and B are valid argument types for all binary functors in [CGAL::Algebraic_structure_traits](https://doc.cgal.org/latest/Algebraic_foundations/classCGAL_1_1Algebraic__structure__traits.html)<Type> and [CGAL::Real_embeddable_traits](https://doc.cgal.org/latest/Algebraic_foundations/classCGAL_1_1Real__embeddable__traits.html)<Type>. This is also the case for the respective global functions." which is what [Bbox_3_Line_3_do_intersect.h](https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Bbox_3_Line_3_do_intersect.h) does by using `compare`.

PR #6263, opened 2022-01-21
Warning when building Periodic_2_triangulation_2 documentation

albert-github:
When building the Periodic_2_triangulation_2 documentation we get the warning:
```
/home/cgal-testsuite/cgal_doc_build/CGAL-5.5-Ic-3/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt:202: warning: End of list marker found without any preceding list items
```

This has been corrected.

### DIFF
--- a/Periodic_2_triangulation_2/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt
+++ b/Periodic_2_triangulation_2/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt
@@ -199,7 +199,6 @@ parameters:
   `Periodic_2TriangulationTraits_2` (for `Periodic_2_triangulation_2`)
   and `Periodic_2DelaunayTriangulationTraits_2` (for
   `Periodic_2_Delaunay_triangulation_2`) in the reference manual and
-.
 - the <B>triangulation data structure</B> class, which stores the
   combinatorial structure, described in Section \ref
   P2Triangulation2sectds and in more detail in Chapter \ref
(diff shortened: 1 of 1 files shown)


Periodic_2_triangulation_2/doc/Periodic_2_triangulation_2/Periodic_2_triangulation_2.txt
```diff
@@ -199,7 +199,6 @@ parameters:
   `Periodic_2TriangulationTraits_2` (for `Periodic_2_triangulation_2`)
   and `Periodic_2DelaunayTriangulationTraits_2` (for
   `Periodic_2_Delaunay_triangulation_2`) in the reference manual and
```
MaelRL, line 201:
```suggestion
  `Periodic_2_Delaunay_triangulation_2`).
```

albert-github: I agree, though it is not up to me to decide to change the texts (unless it is obviously wrong).
Against the change is that when printing (and it is a general text so this might happen) there is no direct link to the reference manual (although I expect that the later is valid on many more places).

MaelRL: I don't think it is very likely that someone will ever print the P2T2 user manual, and concepts and classes are used without reminding every time the reader that they are part of the reference manual.

Also I am a maintainer of the periodic triangulations, so it is ok to apply the suggestion and thanks for the patch.

sloriot:
Successfully tested in CGAL-5.5-Ic-8 (warning is gone)

PR #6266, opened 2022-01-24
CGAL Examples: CGAL_assertion ->  assert

afabri:
## Summary of Changes

Do not use `CGAL_assertion()`  as this is reserved to being used inside CGAL code. 

## Release Management

* Affected package(s): many

### DIFF
--- a/Alpha_shapes_2/examples/Alpha_shapes_2/ex_alpha_shapes_2.cpp
+++ b/Alpha_shapes_2/examples/Alpha_shapes_2/ex_alpha_shapes_2.cpp
@@ -6,7 +6,6 @@
 #include <CGAL/Delaunay_triangulation_2.h>
 
 #include <CGAL/algorithm.h>
-#include <CGAL/assertions.h>
 
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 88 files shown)


afabri:
@mglisse I didn't change this in the package `Triangulation`   as I am not sure if it should be replaced to `assert` to illustrate the users what they can expect to be true after doing operations, or  if it is in reality a testsuite that only happens to be in the wrong directory.

afabri:
@gdamiand @efifogel what I just wrote to Marc also holds for your packages.   Feel free to tell me what you expect me to do, unless you do the changes yourself.

mglisse:
In Triangulation examples, `CGAL_assertion` seems to be used to document what the user can expect. `assert` is indeed probably better suited for that.
Of course any example could also be used as a test, but it doesn't seem to be their primary use here.

gdamiand:
> 

@afabri I think you have more experience than me to decide when it is better to use assert or cgal assertion.
If it only concerns examples and tests, feel free to do what you want.

afabri:
@gdamiand it is in fact more than changing to `assert()`.   For example  I would remove an `assert(lcc..is_valid()` if there is no reason not to be valid, for example when you have constucted it with `CGAL::make_hexahedron(lcc)`.   I would only assert for example that a file could be opened  and the file name comes as command line argument.

gdamiand:
> @gdamiand it is in fact more than changing to `assert()`. For example I would remove an `assert(lcc.is_valid()` if there is no reason not to be valid, for example when you have constucted it with `CGAL::make_hexahedron(lcc)`. I would only assert for example that a file could be opened and the file name comes as command line argument.

Ok I understand. I the example you gave, I agree we can remove the assert.

sloriot:

PR #6268, opened 2022-01-24
Set precision of the ostringstream

afabri:
## Summary of Changes

Fix  for an assertion in the [Polynomial testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-I-4/Polynomial/TestReport_Sosno_MSVC-2019-Community-Release.gz) which is there since ever.

For testing I/O we write to an `ostringstream` and read back from it, but we forgot to call `precision(17)`.   
No idea why we only have this problem with VC++.

## Release Management

* Affected package(s): Polynomial


sloriot (approved):

sloriot:

PR #6274, opened 2022-01-26
Distance_3: Construct a FT from non integral double

afabri:
## Summary of Changes

When constructing a homogeneous point from 2 `int` and a double we call the constructor for 3 `Gmpz`  which warns about doubles that are not integral.  When we use `FT`  we construct quotients. 

## Release Management

* Affected package(s): Distance_3
* Issue(s) solved (if any): See this [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-6/Distance_3/TestReport_Christo_MSVC2017-Debug-64bits.gz) with a warning

### DIFF
--- a/Distance_3/test/Distance_3/test_distance_3.cpp
+++ b/Distance_3/test/Distance_3/test_distance_3.cpp
@@ -86,7 +86,7 @@ private:
 
   P random_point() const
   {
-    return P(r.get_double(m, M), r.get_double(m, M), r.get_double(m, M));
+    return P(FT(r.get_double(m, M)), FT(r.get_double(m, M)), FT(r.get_double(m, M)));
   }
 
   Pl pl(int a, int b, int c, int d)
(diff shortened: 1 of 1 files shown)


mglisse:
> When constructing a homogeneous point from 2 `int` and a double we call the constructor for 3 `Gmpz` which warns about doubles that are not integral.

Did we ever consider / try marking the constructor `Gmpz(double)` with `explicit`?

afabri:
> > When constructing a homogeneous point from 2 `int` and a double we call the constructor for 3 `Gmpz` which warns about doubles that are not integral.
> 
> Did we ever consider / try marking the constructor `Gmpz(double)` with `explicit`?

Mael made the same suggestion. Note that this does not solve the problem.

mglisse:
> Note that this does not solve the problem.

Could you clarify why? A quick test shows that the runtime warning disappears and I don't see any new compiler warning about a double being converted to an int. But I didn't think about this much.

afabri:
Ah, you mean that when the point constructor is called with, say `(0, 1, 2.1)`   the overload with three parameters of type `Gmpz` no longer matches and the constructor with three `FT` will be used instead.

mglisse:
> What about other arbitrary precision integral types?

The constructor is explicit in boost's cpp_int. Sadly, the constructor is implicit in gmpxx. I don't know about leda.

sloriot:
Successfully tested in CGAL-5.5-Ic-15 (warning is gone)

PR #6276, opened 2022-01-26
Polynomial: Workaround for VC bug when cout << double  while rounding up

### LINKED ISSUE #6272: Windows 10: Setting the rounding mode influences output of double
## Issue Details

In the Polynomial [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-5/Polynomial/TestReport_Sosno_MSVC-2019-Community-Release.gz )for some versions of `VC++` we get an assertion that `3 != 3.0000000000001`

I boiled that down to this [gist](https://gist.github.com/afabri/db27a95b2818f63889e6a928d30f6c2e)

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):  Windows
* Compiler:  VC 2017 and 2019 
* CGAL version:  afabri/Polynomial-precision-GF  because otherwise we get another assert earlier

afabri:
## Summary of Changes

Locally in the IO test function set the  rounding mode to nearest to avoid a known VC++ bug.
So we can keep the intended check that when rounding up the Modular Arithmetic for Polynomials works correctly.

## Release Management

* Affected package(s): Polynomial
* Issue(s) solved (if any): fix #6272

### DIFF
--- a/Polynomial/test/Polynomial/test_polynomial.h
+++ b/Polynomial/test/Polynomial/test_polynomial.h
@@ -285,6 +285,13 @@ void division(CGAL::Integral_domain_tag) {
 
 template <class NT>
 void io() {
+
+    // https://github.com/CGAL/cgal/issues/6272#issuecomment-1022005703
+    // VCbug:  Setting the rounding mode influences output of double
+    // The rounding mode is set to CGAL_FE_UPWARD in the test ccp files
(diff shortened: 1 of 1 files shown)


lrineau (approved):

sloriot:

PR #6239, opened 2022-01-15
Fix uninitialised members in Convex_hull_vertex_base_2.h

GilesBathgate:
Fix uninitialized members (_info,_p) in Convex_hull_vertex_base_2.h

## Summary of Changes

This is another one of these 'contentious' uninitialised member issues, thrown by the static analysis for this relatively new class. I've created this pull request just to record the issue and potential fix.

## Release Management

* Affected package(s): Convex_hull
* Issue(s) solved (if any): static analysis
* Feature/Small Feature (if any): small / bugfix
* License and copyright ownership: Returned to CGAL authors.

PR #6280, opened 2022-01-26
Fix info type and set default value

sloriot:
Replaces #6239


sloriot:

PR #6284, opened 2022-01-27
use BGL for OFF writing

### LINKED ISSUE #6278: Triangulation_2:  move triangulation_print_off.h into official IO headers
## Issue Details

Do we still need the file [`triangulation_print_OFF.h`](https://github.com/CGAL/cgal/blob/master/Triangulation_2/examples/Triangulation_2/triangulation_print_OFF.h) in an example directory?

sloriot:
Fixes #6278


sloriot:
I forgot to mention that I checked that the output before and after my change are identical.

Triangulation_2/examples/Triangulation_2/terr_trian.cpp
```diff
@@ -5,17 +5,19 @@
 
 #include <CGAL/IO/Verbose_ostream.h>
 #include <CGAL/IO/OFF.h>
+#include <CGAL/boost/graph/IO/OFF.h>
 #include <CGAL/Projection_traits_xy_3.h>
 #include <CGAL/Triangulation_2.h>
 #include <CGAL/Delaunay_triangulation_2.h>
+#include <CGAL/boost/graph/graph_traits_Delaunay_triangulation_2.h>
+#include <CGAL/boost/graph/graph_traits_Triangulation_2.h>
```
afabri, line 13: I would expect the Delaunay one to include the Triangulation one.

afabri:
As we are at it, what about replacing `int* indices = new int[ scanner.size_of_vertices()];`

Triangulation_2/examples/Triangulation_2/terr_trian.cpp
afabri: Looking at the file I now see that we also have a variable of type `Triangulation_2` which explains the inclusion. So let's keep it.

afabri (approved):

sloriot:

sloriot:

PR #6288, opened 2022-01-31
Simplify initialisation in Nef_polyhedron_3

GilesBathgate:
## Summary of Changes

In Nef_polyhedron_3, the following pattern seems to be used in constructors:

```c++
SNC_structure rsnc;
*this = Nef_polyhedron_3(rsnc, new SNC_point_locator_default, false);
```

This pattern is not familiar, but looks like a way of delegating constructors pre C++11. The issue with this method though is that the base is double initialised, as demonstrated by this minimal example: http://cpp.sh/2xgim This isn't too bad in itself, however, because of SNC_structure also being initialised twice, this leads to four Unique_hash_maps, (that allocate 512 buckets)

To clean the code this pull request introduces a private constructor that can be using constructor delegation. The constructor only initialises the point locator and sets the snc pointer. Because of the overload clash with the public constructor that has a single default argument, a `struct Private_tag{};` is used. The constructor replaces calls to empty_rep which echos the same pattern, and also within the construction of `res` objects in the boolean operation functions.

The performance measurements show this change to be an improvement, however, it's unlikely to have a big impact since generally the constructors call other methods (such as `simplify()`) which take the majority of the time. That said simply constructing a Nef_3 from a Point_3 seems to be 60% faster.

## Release Management

* Affected package(s): Nef_3
* Issue(s) solved (if any): cleaning / performance

### DIFF
--- a/Nef_3/include/CGAL/Nef_polyhedron_3.h
+++ b/Nef_3/include/CGAL/Nef_polyhedron_3.h
@@ -117,7 +117,7 @@ class Nef_polyhedron_3_rep
   SNC_point_locator* pl_;
 
  public:
-  Nef_polyhedron_3_rep() : snc_(), pl_() {}
+  Nef_polyhedron_3_rep() : snc_(), pl_(nullptr) {}
   ~Nef_polyhedron_3_rep() {
     CGAL_NEF_TRACEN( "Nef_polyhedron_3_rep: destroying SNC structure "<<&snc_<<
             ", point locator "<<pl_);
@@ -355,6 +355,12 @@ protected:
     delegate(mbv, /*compute_external*/ false, /*simplify*/ false);
   }
 
+  struct Private_tag {};
+  Nef_polyhedron_3(Private_tag) {
+    pl() = new SNC_point_locator_default;
+    set_snc(snc());
+  }
+
  public:
   /*{\Mcreation 3}*/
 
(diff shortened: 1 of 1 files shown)


GilesBathgate:
An excerpt from the benchmark results:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_construct                             -0.4771         -0.4772             0             0             0             0
BM_construct                             -0.4688         -0.4688             0             0             0             0
BM_construct                             -0.4806         -0.4806             0             0             0             0
BM_construct                             -0.4732         -0.4732             0             0             0             0
BM_construct                             -0.4786         -0.4786             0             0             0             0
BM_construct                             -0.4669         -0.4669             0             0             0             0
BM_construct                             -0.4643         -0.4643             0             0             0             0
BM_construct                             -0.4752         -0.4753             0             0             0             0
BM_construct                             -0.4720         -0.4720             0             0             0             0
BM_construct                             -0.4734         -0.4734             0             0             0             0
BM_construct_pvalue                       0.0002          0.0002      U Test, Repetitions: 10 vs 10
BM_construct_mean                        -0.4730         -0.4730             0             0             0             0
BM_construct_median                      -0.4740         -0.4740             0             0             0             0
BM_construct_stddev                      +0.1205         +0.1190             0             0             0             0
```

and the code:

```c++
static void BM_construct(benchmark::State& state) {
  for (auto _ : state) {
      Nef_polyhedron nefA(Point_3(20.1,12.0,18.333));
      benchmark::DoNotOptimize(nefA);
  }
}
```

afabri:
@GilesBathgate : Is it the same in `Nef_S2`  and `Nef_2`? As a `Nef_S2`  is used at every vertex of a `Nef_3` this might be interesting to investigate.
@lrineau and @mglisse  : Do you approve this redesign  from a C++ point of view?

lrineau (approved):
This looks good. Let's check the change with the testsuite.

GilesBathgate:
> Is it the same in `Nef_S2` and `Nef_2`? As a `Nef_S2` is used at every vertex of a `Nef_3` this might be interesting to investigate.

@afabri There does not seem to be a similar pattern applied to `Nef_2` or `Nef_S2`

sloriot:
Successfully tested in CGAL-5.5-Ic-18

PR #6289, opened 2022-02-01
Hash_map: Restructure the benchmark code 

afabri:
## Summary of Changes

In @GilesBathgate's PR #5673 the `CGAL::Unique_hash_map` undegoes changes that we have to benchmark in isolation.
We already have a benchmark subdirectory, and I restructured the test of filling maps and querying maps for the vertex descriptors of a `CGAL::Polyhedron_3`.

@GilesBathgate and @sloriot can you have a look if this test has no design flaw and makes sense. We then can cherry pick it  to the branch of #5673 and compare performances of the two branches on several platforms.  With VC2019, I have the impression that the proposed changes make it slower.


afabri:
It might be that a call to `reserve()`  completely changes the picture.  This is not offered by the `CGAL::Unique_hash_map` in the master branch.

GilesBathgate:
@afabri The code does make sense to me, and I think using some degree of randomness in the input data should stop the effect of branch prediction in the CPU, to get a true picture of the performance. I do think that micro-benchmark frameworks offer features that can make the tests more accurate.

> It might be that a call to `reserve()` completely changes the picture. This is not offered by the `CGAL::Unique_hash_map` in the master branch.

Yeah, the consequence of a small table size for a large data set is that it will have to rehash more times. (actually 4 extra times to get from 32 to 512 buckets) So while the small table size improves the performance of small data sets (maybe hash tables shouldn't be being used for these in the first place) `reserve()` should be called when the total element count is known in advance.

afabri:
@GilesBathgate can you cherrypick the benchmark and then perform it in this branch and in your branch, to confirm that your changes do not improve performance.

GilesBathgate:
> Is that random enough, and if not what do you propose.

It was a positive affirmation. I noted that you do randomise, and this is a good idea.

GilesBathgate:
> confirm that your changes do not improve performance.

That sounds like loaded bias :(

Anyway like my results were showing in the [Nef_benchmark](https://github.com/afabri/Nef_3_benchmark/actions/runs/1780771074) I am only getting 8% at most. 

I added the following to the test:
```diff

+template <typename Map>
+void reserve(Map& sm, int ii){
+    sm.reserve(ii);
+}
+
+template <typename X, typename Y>
+void reserve(std::map<X,Y>&, int){
+}
 
 template <typename Map, typename VPM>
 double fct(int ii, int jj, const std::vector<vertex_descriptor>& V, const VPM& vpm, const std::string& s)
@@ -35,6 +43,7 @@ double fct(int ii, int jj, const std::vector<vertex_descriptor>& V, const VPM& v
   construct.start();
   for(int j=0; j <jj; j++){
     Map sm;
+    reserve(sm,ii);
     for(vertex_descriptor vh : V){
       sm[vh] = get(vpm,vh);
     }
```

Then I had to make further changes to include the contender in the same run (because of the variance across runs caused by std::shuffle) These changes can be seen here: https://github.com/GilesBathgate/cgal/tree/Hash_map-benchmark-GilesBathgate

The results are attached:
[results.txt](https://github.com/CGAL/cgal/files/7985300/results.txt)

I think it's an improvement, but it's possible I also made a mistake.

GilesBathgate:
@afabri This maybe a better way to use the reserve method, only when available.

```c++
template <typename Map>
auto reserve(Map& sm, int ii) -> decltype(sm.reserve(ii), void()){
    sm.reserve(ii);
}
```

lrineau:
> @afabri This maybe a better way to use the reserve method, only when available.
> 
> ```c++
> template <typename Map>
> auto reserve(Map& sm, int ii) -> decltype(sm.reserve(ii), void()){
>     sm.reserve(ii);
> }
> ```

I agree.

Explanation: using `auto` with a trailing return type allows to SFINAE-reject the method if the expression `sm.reserve(ii)` does not compile.

I also find dubious the usage of `int`, here. The parameter of `reserve` is `Container::size_type` and it is documented to be unsigned. You can use `typename Map::size_type` instead of `int`, or `std::size_t`.

GilesBathgate:
@afabri The benchmark appears to only check for keys that do exist, it doesn't check call `sm[unknown_key]` (both const and non-const versions of `operator[]`)

afabri:
Just to check that I got it right.   You created the branch  [Hash_map-benchmark-GilesBathgate](https://github.com/GilesBathgate/cgal/compare/Hash_map-benchmark-GilesBathgate)    from this branch and then you added your modified ` Unique_hash_map` from the Nef branch so that we can compare the two Unique hash map versions in one run.   Good idea.

afabri:
> @afabri The benchmark appears to only query keys that do exist, it doesn't call `sm[unknown_key]` (both const and non-const versions of `operator[]`)

Do not all the insertions are an unknown key when we insert them?   This even triggers the question if when we bench the construction phase, if  we do not have to use an ` insert()`  function.

GilesBathgate:
> Do not all the insertions are an unknown key when we insert them

Yes that's true, perhaps I need to better explain. There is `const Data& operator[](K) const` and `Data& operator[K]` when the non-const one is called, if the value doesn't exist its inserted (chained_map.access) This is what is done in the construction phase. However when the const one is called it looks up the value (chained_map.lookup), and if it doesn't exist it returns the default value. Thus a const query that looks for a non-existent item doesn't have the overhead of inserting the value. I thought this would make a good addition to the benchmark.

afabri:
> I thought this would make a good addition to the benchmark.

I would be glad if you added it.

GilesBathgate:
> What I do not like about your changes of the Unique hash map implementation is that it is a huge monolithic diff.

I think I have the full set of changes, but I squashed them. I will see if I can find them, then perhaps cherry-pick the key changes to another branch and see if it still performs the same.

Hash_map/benchmark/Hash_map/polyhedron.cpp
```diff
@@ -18,125 +16,134 @@
...
-  typedef boost::property_map<Mesh,CGAL::vertex_point_t>::type  VPM;
-  VPM vpm = get(CGAL::vertex_point,mesh);
+template <typename X, typename Y>
+Point_3 lookup(CGAL::Unique_hash_map<X,Y>& sm,vertex_descriptor vh){
+    const CGAL::Unique_hash_map<X,Y>& const_sm = sm;
+    return const_sm[vh];
+}
 
+template <typename Map>
+double fct(int ii, int jj, const Vertex_list& V, const Vertex_list& V2, const VPM& vpm, const std::string& s)
```
GilesBathgate, line 56: I don't know why `fct` returns a `double`, when `int x` and `int res`/`int temp`

afabri: Should be just int.
I do this adding to make sur that the compiler does not optimize loops away that do nothing.
And also to check that for all the map types the result is the same. 

Btw, we should maybe only work in your branch GilesBathgate/Hash_map-benchmark-GilesBathgate 
instead of this one.

GilesBathgate:
> Btw, we should maybe only work in your branch GilesBathgate/Hash_map-benchmark-GilesBathgate
> instead of this one.

I don't mind how we do the github workflow. I intended my branch as 'throw away' because it has changes to include `Unique_hash_map_contender` and `contender::chained_map` which aren't intended for CGAL:master

afabri:
We can also use this branch. I just invited you yo become collaborator, so that you can push the contenders and/or improve the benchmark code.   Please don't force push though.

GilesBathgate:
> Please don't force push though.

Thanks for the invite! I won't force push.

GilesBathgate:
@afabri Is the intention for this branch still to be tested/merged?

afabri:
> @afabri Is the intention for this branch still to be tested/merged?

Do you want this branch to be in master so that we have the same tests in future Nef-related branches to come?

GilesBathgate:
> Do you want this branch to be in master so that we have the same tests in future Nef-related branches to come?

I think that's a good approach.

afabri:
> > Do you want this branch to be in master so that we have the same tests in future Nef-related branches to come?
> 
> I think that's a good approach.

@lrineau @sloriot can you please merge this PR

lrineau:
I eventually understood that this pull-request can be, and must be, merged without being tested, because it modifies only one file that is not part of what is tested by any of our test processes.

I just merged it.

Issue #6269, opened 2022-01-24
Linear_cell_complex:   ICE of GNU 12.0.1


afabri:
## Issue Details

All testsuites using the Linear_cell_complex fail, for example [this  one](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-I-4/Linear_cell_complex/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz).

```
CGAL/Linear_cell_complex_base.h:279:40: internal compiler error: in cp_parser_template_id, at cp/parser.cc:18367
  279 |     { return this->template attribute<0>(adart); }
```

## Environment

* Compiler: GNU 12.0.1
* maybe due to option `-std=c++1z`

sloriot:
More precisely
-- USING GCC_VERSION = '12.0.1 20220123 (experimental)'

sloriot:
While it semms fine with: '12.0.0 20220116 (experimental)'

gdamiand:
It looks like a compiler bug ? (since it compiles with 12.0.0 20220116 but not with 12.0.1 20220123)

sloriot:
It is, Andreas forgot to mention that it is an ICE.

lrineau:
Yes. For information, ICE stands for *internal compiler error*. ICEs are usually caused by strange or incorrect C++ code. But sometimes, that is just a plain bug, where the compiler crashes on a valid C++ code.

Anyway, ICEs are usually reported, because even if the code is invalid, the compiler should report an error message, instead of crashing.

mglisse:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104235

afabri:
Thank you Marc. I am glad that the CGAL Project can serve as regression test for the GNU Project ==:>--

lrineau:
The issue was fixed upstream:
  - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104235 is marked as fixed,
  - by the commit https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=00d8321124123daf41f7c51526355a5a610cdeb8
  - and the testsuite is now OK: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-I-9/Linear_cell_complex/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz

sloriot:
It is stil here:
[...]

PR #6295, opened 2022-02-01
try to workaround ICE

sloriot:
Try to workaround [this ICE](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-10/BGL/TestReport_lrineau_Fedora-rawhide.gz)

```
/mnt/testsuite/include/CGAL/Linear_cell_complex_base.h:279:40: internal compiler error: in cp_parser_template_id, at cp/parser.cc:18367
  279 |     { return this->template attribute<0>(adart); }
```

### DIFF
--- a/Linear_cell_complex/include/CGAL/Linear_cell_complex_base.h
+++ b/Linear_cell_complex/include/CGAL/Linear_cell_complex_base.h
@@ -262,11 +262,11 @@ namespace CGAL {
 
     /// @return the Vertex_attribute_range for all vertex_attributes.
     Vertex_attribute_range& vertex_attributes()
-    { return this->template attributes<0>(); }
+    { return Base::template attributes<0>(); }
 
     /// @return the Vertex_attribute_const_range for all vertex_attributes.
     Vertex_attribute_const_range& vertex_attributes() const
-    { return this->template attributes<0>(); }
+    { return Base::template attributes<0>(); }
 
     /// @return the size of the vertex_attribute container.
     typename Base::size_type number_of_vertex_attributes() const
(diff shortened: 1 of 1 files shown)


sloriot:
It seems to be working as the error is now:
```
/mnt/testsuite/include/CGAL/Combinatorial_map.h:798:38: internal compiler error: in cp_parser_template_id, at cp/parser.cc:18367
  798 |       if ( this->template attribute<i>(dh)==ah ) return;
```

@gdamiand , shall we remplace by `Base::template` everywhere?

gdamiand:
> @gdamiand , shall we remplace by Base::template everywhere?

If this solves the ereror, yes. Do you want me to do the modifications ?

sloriot:
If you have some time that would be nice :)

gdamiand:
Done [gdamiand:LCC-ICE_workaround](https://github.com/gdamiand/cgal/tree/LCC-ICE_workaround) and tested locally.

sloriot:
Note that the error reported is now in `Combinatorial_map.h`. I've sent you an invite to be  a collaborator. You should be able to push directly in my remote to update this PR now.

lrineau:
@sloriot @gdamiand As far as I know, you can working around a bug that is already fixed.

It was already reported in the issue #6269, and reported upstream by Marc Glisse at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104235. And that bug was fixed by a commit to gcc `master`: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=00d8321124123daf41f7c51526355a5a610cdeb8

What is confusing is that the bug was solved Wed, 26 Jan 2022, and our gcc-master image was correctly [rebuild and pushed][1] Jan 30th, but the [error log in CGAL-5.5-Ic-10][2] was Tuesday, 1st February, with the newly compiled gcc-master.

@mglisse What do you think? If that is not the same ICE, then it should be reported as well.

[1]: https://github.com/CGAL/cgal-testsuite-dockerfiles/runs/4994351993?check_suite_focus=true
[2]: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-10/BGL/TestReport_lrineau_Fedora-rawhide.gz

mglisse:
I just tried compiling test_clear.cpp with a gcc built today, and it doesn't ICE.
Looking at the logs of Installation for the platform you mention
> c++ (GCC) 12.0.1 20220125 (Red Hat 12.0.1-0)

lrineau:
Ah, OK. Last version of gcc in Fedora-rawhide was a build from Jan 25th. The column to follow is the ones named "Ubuntu-latest-GCC6"... Even if it is incorrectly named "GCC6", that compiler is gcc-master, recompiled each Sundays.

So, now compare:
- https://cgal.geometryfactory.com/CGAL/testsuite/results-5.5-I-4.shtml#BGL
- https://cgal.geometryfactory.com/CGAL/testsuite/results-5.5-I-9.shtml#BGL

The bug disappeared from the columns "GCC6", and it will disappear from Fedora-rawhide as well with the new Fedora build of gcc. Fedora rawhide [got a new build of gcc](https://koji.fedoraproject.org/koji/buildinfo?buildID=1909942), started Sat, 29th Jan, and finished Sun, 30th Jan. It was too late to be in our renewal of Docker containers, but it will be next week.

So, in my opinion, @sloriot @gdamiand you are working around a bug that is already fixed.

sloriot:
OK let's close it and see if we still have it next week.

PR #6298, opened 2022-02-02
Triangulation_2: Quiet a warning

afabri:
## Summary of Changes

We only default initialize a point as we are sure that it will be set by the intersection function before it is used, but the compiler does not know and warns for several test platforms.  So let's just initialize it.

## Release Management


Triangulation_2/include/CGAL/Constrained_triangulation_plus_2.h
```diff
@@ -1186,7 +1186,8 @@ intersect(Face_handle f, int i,
             << " , #" << vd->time_stamp() << "= " << vd->point()
             << " , Exact_intersections_tag)\n";
 #endif // CGAL_CDT_2_DEBUG_INTERSECTIONS
-  Point pi;
+  Point pi(ORIGIN); // initialize although we are sure that it will be
+                    // set by the intersection, but to quit a warning
```
MaelRL, line 1190:
```suggestion
                    // set by the intersection, but to quiet a warning
```

sloriot:
Successfully tested in CGAL-5.5-Ic-15 (warning is gone)

PR #6304, opened 2022-02-03
Intersection_3: Use Coercion_traits

afabri:
## Summary of Changes

Use the correct types when mixing double and coordinate types which may be not double.  This was discussed in PR #6261

## Release Management

* Affected package(s): Intersections_3


sloriot:
Successfully tested in CGAL-5.5-Ic-15

PR #5201, opened 2020-11-24
Fix uninitialized member in chained_map.h

GilesBathgate:
## Summary of Changes

Change chained_map_elem to struct
Fix uninitialized members (STOP, old_table_end, old_free, old_table_size, old_table_size_1, old_index) in chained_map.h

## Release Management

* Affected package(s): Hash_map
* Issue(s) solved (if any): fix #5181
* License and copyright ownership: Returned to CGAL authors

PR #6306, opened 2022-02-03
Reserve method for Chained Map / Unique Hash Map 

GilesBathgate:
## Summary of Changes

This replaces #5673 and implements just the minimal changes required to add a reserve method and initialise the hash map on first access instead of during construction.

* The table is now initialized on first use instead of at construction
* Keep the `default_size` of 512
* Add a `min_size` of 32
* Passing a table_size value less than 512 in the constructor works.
* Calling `reserve` can set the table_size before (and only before) table first use.

---
Changes among Nef_3 classes.

* Utilise the `reserve` method in places where the hash table is created ahead of knowing how many items it would contain
* Utilise the table_size  parameter in the constructor where the hash table is created and the number of items is known.

## Release Management

* Affected package(s): Nef_3 / Hash_map
* Issue(s) solved (if any): performance / speed
* License and copyright ownership: Returned to CGAL Authors.

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
@@ -545,7 +545,6 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
       Unique_hash_map< SFace_handle, UFH_sface>& hash,
       Union_find< SFace_handle>& uf ) {
     Unique_hash_map< SHalfedge_handle, bool> linked(false, this->sncp()->number_of_shalfedges());
-    this->sncp()->reserve_sm_boundary_items(this->sncp()->number_of_sfaces());
 
     SNC_decorator D(*this->sncp());
     SHalfedge_iterator e;
--- a/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
@@ -316,9 +316,9 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
     CGAL_NEF_TRACEN(">>> simplifying");
     SNC_decorator D(*this->sncp());
 
-    Unique_hash_map< Volume_handle, UFH_volume> hash_volume;
-    Unique_hash_map< Halffacet_handle, UFH_facet> hash_facet;
-    Unique_hash_map< SFace_handle, UFH_sface> hash_sface;
+    Unique_hash_map< Volume_handle, UFH_volume> hash_volume(UFH_volume(), this->sncp()->number_of_volumes());
+    Unique_hash_map< Halffacet_handle, UFH_facet> hash_facet(UFH_facet(), this->sncp()->number_of_halffacets());
+    Unique_hash_map< SFace_handle, UFH_sface> hash_sface(UFH_sface(), this->sncp()->number_of_sfaces());
     Union_find< Volume_handle> uf_volume;
     Union_find< Halffacet_handle> uf_facet;
     Union_find< SFace_handle> uf_sface;
@@ -544,7 +544,7 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
   void create_boundary_links_forall_sfaces(
       Unique_hash_map< SFace_handle, UFH_sface>& hash,
       Union_find< SFace_handle>& uf ) {
-    Unique_hash_map< SHalfedge_handle, bool> linked(false);
+    Unique_hash_map< SHalfedge_handle, bool> linked(false, this->sncp()->number_of_shalfedges());
     SNC_decorator D(*this->sncp());
     SHalfedge_iterator e;
     CGAL_forall_shalfedges(e, *this->sncp()) {
@@ -597,7 +597,7 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
   void create_boundary_links_forall_facets(
       Unique_hash_map< Halffacet_handle, UFH_facet>& hash,
       Union_find< Halffacet_handle>& uf) {
-    Unique_hash_map< SHalfedge_handle, bool> linked(false);
+    Unique_hash_map< SHalfedge_handle, bool> linked(false, this->sncp()->number_of_shalfedges());
     SNC_decorator D(*this->sncp());
     SHalfedge_iterator u;
     CGAL_forall_shalfedges(u, *this->sncp()) {
--- a/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
@@ -545,6 +545,8 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
       Unique_hash_map< SFace_handle, UFH_sface>& hash,
       Union_find< SFace_handle>& uf ) {
     Unique_hash_map< SHalfedge_handle, bool> linked(false, this->sncp()->number_of_shalfedges());
+    this->sncp()->reserve_sm_boundary_items(this->sncp()->number_of_sfaces());
+
     SNC_decorator D(*this->sncp());
     SHalfedge_iterator e;
     CGAL_forall_shalfedges(e, *this->sncp()) {
--- a/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_simplify.h
@@ -651,6 +651,7 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
 
     SNC_decorator D(*this->sncp());
     Volume_setter setter(D);
+    setter.reserve(this->sncp()->number_of_sfaces());
 
     SFace_iterator sf;
     Volume_handle c;
--- a/Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
+++ b/Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
@@ -205,7 +205,6 @@ void polygon_mesh_to_nef_3(PolygonMesh& P, SNC_structure& S, FaceIndexMap fimap,
   Face_graph_index_adder<typename SNC_structure::Items,
                  PolygonMesh, SNC_structure,HalfedgeIndexMap> index_adder(P,himap);
 
-  S.reserve_sm_boundary_items(num_vertices(P));
   for(vertex_descriptor pv : vertices(P) ) {
 
     typename boost::property_traits<PMap>::reference npv = get(pmap,pv);
--- a/Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
+++ b/Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
@@ -205,7 +205,7 @@ void polygon_mesh_to_nef_3(PolygonMesh& P, SNC_structure& S, FaceIndexMap fimap,
   Face_graph_index_adder<typename SNC_structure::Items,
                  PolygonMesh, SNC_structure,HalfedgeIndexMap> index_adder(P,himap);
 
-
+  S.reserve_sm_boundary_items(num_vertices(P));
   for(vertex_descriptor pv : vertices(P) ) {
 
     typename boost::property_traits<PMap>::reference npv = get(pmap,pv);
--- a/Hash_map/include/CGAL/Tools/chained_map.h
+++ b/Hash_map/include/CGAL/Tools/chained_map.h
@@ -83,13 +83,15 @@ private:
    }
 
 public:
+   static constexpr std::size_t min_size = 32;
+   static constexpr std::size_t default_size = 512;
    typedef chained_map_elem<T>*  chained_map_item;
    typedef chained_map_item item;
 
    std::size_t index(chained_map_item it) const { return it->k; }
    T&            inf(chained_map_item it) const { return it->i; }
 
-   chained_map(std::size_t n = 1);
+   chained_map(std::size_t n = default_size);
    chained_map(const chained_map<T, Allocator>& D);
    chained_map& operator=(const chained_map<T, Allocator>& D);
 
@@ -273,8 +275,8 @@ template <typename T, typename Allocator>
 chained_map<T, Allocator>::chained_map(std::size_t n) :
   nullptrKEY(0), NONnullptrKEY(1), old_table(0)
 {
-  if (n < 512)
-    init_table(512);
+  if (n < min_size)
+    init_table(min_size);
   else {
     std::size_t ts = 1;
     while (ts < n) ts <<= 1;
@@ -336,7 +338,7 @@ void chained_map<T, Allocator>::clear()
     destroy(item);
   alloc.deallocate(table, table_end - table);
 
-  init_table(512);
+  init_table(min_size);
 }
 
 template <typename T, typename Allocator>
--- a/Hash_map/include/CGAL/Tools/chained_map.h
+++ b/Hash_map/include/CGAL/Tools/chained_map.h
@@ -95,7 +95,7 @@ public:
    chained_map(const chained_map<T, Allocator>& D);
    chained_map& operator=(const chained_map<T, Allocator>& D);
 
-
+   void reserve(std::size_t n);
    void clear_entries();
    void clear();
    ~chained_map()
@@ -322,6 +322,13 @@ chained_map<T, Allocator>& chained_map<T, Allocator>::operator=(const chained_ma
     }
   }
   return *this;
+}
+
+template <typename T, typename Allocator>
+void chained_map<T, Allocator>::reserve(std::size_t n)
+{
+  CGAL_assertion(!table);
+  reserved_size = n;
 }
 
 template <typename T, typename Allocator>
--- a/Hash_map/include/CGAL/Tools/chained_map.h
+++ b/Hash_map/include/CGAL/Tools/chained_map.h
@@ -118,6 +118,7 @@ public:
    chained_map_item lookup(std::size_t x) const;
    chained_map_item first_item() const;
    chained_map_item next_item(chained_map_item it) const;
+   std::size_t size() const;
    void statistics() const;
 };
 
@@ -384,6 +385,10 @@ chained_map<T, Allocator>::next_item(chained_map_item it) const
   do it++; while (it < table + table_size && it->k == nullptrKEY);
   return (it < free ? it : 0);
 }
+
+template <typename T, typename Allocator>
+std::size_t chained_map<T, Allocator>::size() const
+{ return table_size; }
 
 template <typename T, typename Allocator>
 void chained_map<T, Allocator>::statistics() const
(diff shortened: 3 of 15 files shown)


afabri:
How important is the change to a `min_size` of 32?  Can that make  things slower for use cases with only very few entries in the hash map?

GilesBathgate:
@afabri It's the other way around. The main benefit is when there are few items (lets say 10) because you only have to allocate 32 buckets for the 10 items, instead of 512 buckets for 10 items. 

```
10 items and queries (repeated 1000000 times)
Name			| Version	| Construction	| Queries	| Lookups
CGAL::Unique_hash_map	| master	| 0.754515 sec.	| 0.013408 sec.	| 0.017352 sec.
CGAL::Unique_hash_map	| a2ebade min	| 0.122417 sec.	| 0.034371 sec.	| 0.015394 sec.
```

For larger hash maps if you insert more items than there are buckets, it has to double the table size and rehash all the items. If you know ahead of time how many items you are going to add you can call `reserve()`, which saves having to do the rehashing.

GilesBathgate:
@afabri perhaps as a compromise you want to be able to reserve less than 512 (with a `min_size` of 32) but you'd like the `default_size` to be 512?

afabri:
> @afabri perhaps as a compromise you want to be able to reserve less than 512 (with a `min_size` of 32) but you'd like the `default_size` to be 512?

When I do Boolean operations with the grid of spheres it is in fact faster.  I only do not know if other CGAL packages that use the Unique_hash map profit as well or become slower.

afabri:
Hello @efifogel 
I see a heavy usage in [Gaussian_map,h](https://github.com/CGAL/cgal/blob/master/Minkowski_sum_3/include/CGAL/Minkowski_sum_3/Gaussian_map.h).    Can you tell us if this code profits or suffers from whether the `Unique_hash_map` starts with a table size of 512.

GilesBathgate:
>  Can you tell us if this code profits or suffers from whether the `Unique_hash_map` starts with a table size of 512.
@afabri , @efifogel To be clear 512 is the default value before this pull request was proposed. The question is would the default value benefit from being smaller than 512, i.e 32.

afabri:
Thinking about it again, when the table is at the end bigger than ~1000 but started with size 32 the additional rehash for going from 32 to 64, 128, 256, and 512 should get negligible (in an amortized sense).

lrineau:
@GilesBathgate In the same pull-request you change two things:
  - the table is now initialized on first use (and thus every access has to be protected by the code `if(!table) init_table(reserved_size);`
  - you changed the default size from 512 to 32.

Are those changes related? Do you have an idea of the impact on performance for those two changes, independently?

GilesBathgate:
> Are those changes related? Do you have an idea of the impact on performance for those two changes, independently?

They are interrelated because if the default size is 32, you might want to avoid the compound rehashing 32 to 64, 128, 256, and 512 as @afabri mentioned by calling `reserve`  but, I've reverted the change from 512 to 32. in commit https://github.com/CGAL/cgal/pull/6306/commits/d697f5b9a54028b0fa0ca6a5a08ec1f56cdc8b7a because I understand that smaller changes are easier to manage in terms of performance and regression testing.

lrineau:
> > Are those changes related? Do you have an idea of the impact on performance for those two changes, independently?
> 
> They are interrelated because if the default size is 32, you might want to avoid the compound rehashing 32 to 64, 128, 256, and 512 as @afabri mentioned by calling `reserve` but, I've reverted the change from 512 to 32. in commit [d697f5b](https://github.com/CGAL/cgal/commit/d697f5b9a54028b0fa0ca6a5a08ec1f56cdc8b7a) because I understand that smaller changes are easier to manage in terms of performance and regression testing.

But what's the relationship between `reserve`, and the lazy initialization of the hashtable? 

I am also not so sure in which scenario there could be a use for `reserve(n)` where `n` would not be known at the construction of the map itself.

Nef_3/include/CGAL/Nef_3/polygon_mesh_to_nef_3.h
```diff
@@ -205,7 +205,7 @@ void polygon_mesh_to_nef_3(PolygonMesh& P, SNC_structure& S, FaceIndexMap fimap,
   Face_graph_index_adder<typename SNC_structure::Items,
                  PolygonMesh, SNC_structure,HalfedgeIndexMap> index_adder(P,himap);
 
-
+  S.reserve_sm_boundary_items(num_vertices(P));
```
GilesBathgate, line 208: >I am also not so sure in which scenario there could be a use for reserve(n) where n would not be known at the construction of the map itself.

@lrineau So for example in the above, (maybe this can be moved further up the call stack?). Hovever I guess its mostly for developer convenience, which is why I think all the other hash_maps compatible with std::unordered_map  implement a `reserve` method.

afabri:
@GilesBathgate  can you instead of `reserve(N)` pass `N` to the constructor where possible in your code. That has the same effect, is less code, and leads to a smaller diff.

GilesBathgate:
@afabri If you don't mind, I will open a new PR for the suggested alternative approach.

afabri:
> @afabri If you don't mind, I will open a new PR for the suggested alternative approach.

As I said I came to the conclusion that 32 is a better choice than 512.   And to have a `reserve()` function is a good idea.  I only think that     `UHM<K,V>  uhm;     uhm.reserve(10);`    is less elegant than  `UHM<K,V>  uhm(V(),10)`.  But I probably misunderstand something.

GilesBathgate:
> I only think that `UHM<K,V> uhm; uhm.reserve(10);` is less elegant than `UHM<K,V> uhm(V(),10)`

Yes, I agree too, which is why I want to create a new PR because I think the commits:

[Add reserve for select hashmaps](https://github.com/CGAL/cgal/pull/6306/commits/6aa7b48acc1ba4e42dc0a9937c4729d1a87d704b)
[Multiple hash maps perform better than single generic hash map](https://github.com/CGAL/cgal/pull/6306/commits/00f8a7de5bc99485eed5cb4da97497685c31de88)
[Add reserve to several uses of Unique_hash_map](https://github.com/CGAL/cgal/pull/6306/commits/1b39a79ad220cd0a12cf7956cf6d51b2bf540c3b)
[Add const to is_boundary_object and is_sm_boundary_object](https://github.com/CGAL/cgal/pull/6306/commits/b5bc0beb61e6da2f353bb4be2b3f35c2a83295b7)
[Add reserve for some other large hashmaps](https://github.com/CGAL/cgal/pull/6306/commits/48bc54ffae2348e08f9c34f1429b66d795419461)

Would be redundant. I can revisit the `reserve` later. But I might have to force push 😉

GilesBathgate:
@afabri I started to implement it but this change in particular has got me thinking: https://github.com/GilesBathgate/cgal/commit/e5c166ac7b007ebe2df98e29bf8d88d5137fb320/Nef_3/include/CGAL/Nef_3/Halffacet.h
I don't think my additions using a hashtable size of `1`,`2` and `facet->number_of_vertices` in the `add_vertex`,`add_edge`,`add_facet` methods is very sensible. Furthermore in `Nef_polyhedron_3` a method for updating the K3_tree was available via `CGAL_NEF3_UPDATE_K3TREE_AFTER_SIMPLIFICATION` that used those three add methods. Perhaps this was a potential optimisation that was dismissed because the hash table overhead.  Maybe a specialisation of `Side_of_plane` without the `OnSideMap` would work better? Do you think its worth an experiment?

GilesBathgate:
@afabri It seemed to make sense to make the changes to this PR in the end.

afabri:
As 32 was a good choice at many places in Nef, do I get it right that there is now another PR where you call the constructor or the reserve function with a value smaller than 512?

GilesBathgate:
> As 32 was a good choice at many places in Nef, do I get it right that there is now another PR where you call the constructor or the reserve function with a value smaller than 512?

It's a good idea. I have some commits for it somewhere, but not for all cases. I can add the changes here, another PR, or whatever you prefer. (Smaller incremental changes might be better.)

afabri (approved):

afabri:
I think this branch has a merge conflict with   Nef_3-performance_chained_map-GilesBathgate.   
Does it make sense to have them both, or will only one of them go into master?   
If they  both go into master, can you please resolve conflicts so that Sebastien can put them at the same time (and together with the other Nef_3 related PRs into the testsuite.

GilesBathgate:
@afabri I've already closed #5673 ,as it doesn't make sense to have both, at this time

GilesBathgate:
@afabri Can you put testing on hold, I've noticed performance regression in minkowski.

GilesBathgate:
> @afabri , @sloriot Can you put testing on hold, I've noticed performance regression in minkowski.

Ok, I managed to track down the changes causing the issue and revert them.

I've attached the results I was seeing, for anyone interested:

[good.txt](https://github.com/CGAL/cgal/files/8291494/good.txt)
[bad.txt](https://github.com/CGAL/cgal/files/8291520/bad.txt)

afabri:
Are you sure that [this assertion](https://github.com/CGAL/cgal/blob/d9d85fbe3cd576479c2d2e28a61d17aa0ccf7c9c/Hash_map/include/CGAL/Tools/chained_map.h#L330) is justified ?  It means that one cannot call reserve once the table got allocated.   
I think that is the runtime error we see in the [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-43/Nef_3_Examples/TestReport_cgaltester_x86-64_Darwin-18.5_Apple-clang-default_Release.gz).

GilesBathgate:
@afabri Basically calling reserve after the table got allocated would make the call redundant, so I wanted to trap usages where that was the case. What is the code that triggers the assertion is it `topological_operations.cpp`? was reserve added to this test?

afabri:
I will run `topological_operations.cpp` in the debugger. But it does not seem logic to me that I am not allowed to call `reserve()` several times, I can also do that with a `std::vector`.

GilesBathgate:
> I am not allowed to call `reserve()` several times, I can also do that with a `std::vector`.

You should be able to reserve as many times as you like, so long as you do so before you add an entry to the hash_table, this is a limitation of the current implementation.

Nef_3/include/CGAL/Nef_3/SNC_simplify.h
```diff
@@ -544,7 +544,9 @@ class SNC_simplify_base : public SNC_decorator<SNC_structure> {
   void create_boundary_links_forall_sfaces(
       Unique_hash_map< SFace_handle, UFH_sface>& hash,
       Union_find< SFace_handle>& uf ) {
-    Unique_hash_map< SHalfedge_handle, bool> linked(false);
+    Unique_hash_map< SHalfedge_handle, bool> linked(false, this->sncp()->number_of_shalfedges());
+    this->sncp()->reserve_sm_boundary_items(this->sncp()->number_of_sfaces());
```
GilesBathgate, line 548: This is the offending line. I should have picked this up earlier, but when performance testing I disable cgal assertions, and forgot to re-enable. This line should be removed as it cannot work currently.

GilesBathgate:
Maybe it could be modified simply:

```
template <typename T, typename Allocator>
void chained_map<T, Allocator>::reserve(std::size_t n)
{
   if(table)
      rehash(n);
  reserved_size = n;
}
```
 It would need changes to rehash though.

Nef_3/include/CGAL/Nef_3/SNC_simplify.h
afabri: So unless you rewrite the rehash immediately, please remove the offending line.

GilesBathgate:
@afabri ok, I'll take a look at these.

GilesBathgate:
@afabri I could not reproduce the llvm segfault, but I don't have a mac. The other issue you mentioned is fixed though.

sloriot:

afabri:
@GilesBathgate can you read the diff once again, because we suspect that this is the one that introduces a run time error for clang++ on Apple.   It fails while constructing a Nef_polyhedron from a Polyhedron_3 and is in a recursion in chained_map with the two functions  access and del_old_table.  I will look into it tomorrow as the Apple is in the office, but if you reread your code, maybe you see what happens.

sloriot:
Note that I could reproduce the crash with 7d0d6a06bed91501bd1c5d6d95f2e36659e49919 alone.

GilesBathgate:
@sloriot What happens if you change min_size to various different values 64, 128, 256...?

sloriot:
64 still crashes but 128 seems fine.

lrineau:
> 64 still crashes but 128 seems fine.

That sort of symptoms cries the need for a run with AddressSanitizer. There is probably somewhere a use of uninitialized memory, or use-after-free.

GilesBathgate:
@lrineau https://github.com/CGAL/cgal/pull/5201

sloriot:
We already did that, and it no longer crashes.

sloriot:
Actually, I gave it a try myself (somebody else told me it was no longer crashing) and here is what I have:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==71394==ERROR: AddressSanitizer: stack-overflow on address 0x7ffee85effa0 (pc 0x00010856d8fa bp 0x7ffee85f07d0 sp 0x7ffee85effa0 T0)
    #0 0x10856d8fa in wrap__Znwm+0x1a (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x508fa)
    #1 0x106e5b618 in std::__1::__libcpp_allocate(unsigned long, unsigned long) new:253
    #2 0x106eb0836 in std::__1::allocator<CGAL::internal::chained_map_elem<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::allocate(unsigned long) memory:1664
    #3 0x106eb0281 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::init_table(unsigned long) chained_map.h:147
    #4 0x1070c0ab2 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::rehash() chained_map.h:189
    #5 0x1070c0239 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::access(CGAL::internal::chained_map_elem<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > >*, unsigned long) chained_map.h:255
    #6 0x1070bf3e0 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::access(unsigned long) chained_map.h:138
    #7 0x1070bfad0 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::del_old_table() chained_map.h:227
    #8 0x1070bf212 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::access(unsigned long) chained_map.h:126
    #9 0x1070bfea0 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::del_old_table() chained_map.h:238
    #10 0x1070bf212 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> >, std::__1::allocator<boost::optional<std::__1::__list_iterator<CGAL::Object, void*> > > >::access(unsigned long) chained_map.h:126
    #11 0x1070bfea0 in CGAL::internal::chained_map<boost::optional<std::__1::__list_iterator<CGAL::Object, ... last two lines repeats up to #255
```

GilesBathgate:
@sloriot If I reduce min_size to 16, I have the exact same stack trace on linux. Looks like it gets stuck in infinite recursion?

GilesBathgate:
@sloriot This fixes it for me:

```diff
@@ -228,20 +228,22 @@ void chained_map<T, Allocator>::del_old_table()
   table = old_table;
   table_end = old_table_end;
   table_size = old_table_size;
   table_size_1 = old_table_size_1;
   free = old_free;
 
   old_table = 0;
 
   T p = access(old_index);
 
+  old_table = 0;
+
   for (chained_map_item item = table ; item != table_end ; ++item)
     destroy(item);
   alloc.deallocate(table, table_end - table);
```

GilesBathgate:
@afabri @sloriot with the above change, it works successfully with min_size = 1

afabri:
Does it get set to a value different from zero by `rehash()`?   In this case, shouldn't it then be reset to zero before returning from `rehash`  instread of doung it where you suggest to do it.

afabri:
That solves it on the Apple.  Why do we have the non-determinism?

GilesBathgate:
> Shouldn't it then be reset to zero before returning from `rehash` instead of doing it where you suggest to do it.

Yes, this was just the outcome of a 'hotfix' as part of the investigation.

> That solves it on the Apple. Why do we have the non-determinism?

My guess is when min_size is 128 an overflow is writing 0 into old_table, but when it's 32 the overflow doesn't happen. Something to do with padding, and the location in memory of old_table with respect to table

GilesBathgate:
`old_table = nullptr;` as the last statement in `rehash()` also fixed the issue

afabri:
The last two comments are a little bit cryptic

GilesBathgate:
About the overflow? Let's just say I am not sure what causes it, but I think my changes in https://github.com/CGAL/cgal/commit/7d0d6a06bed91501bd1c5d6d95f2e36659e49919 alone, possibly expose an already existing non-deterministic memory bug. Its worth further investigation...

afabri:
> The last two comments are a little bit cryptic

Stupid me, you just confirmed what I wrote, namely that it should be reset in `hash()`.    
But I do not understand what you write about overflow and writing in some memory. 

I think the workflow in this PR is too complex.

GilesBathgate:
>I think the workflow in this PR is too complex.

Would you prefer if I create a PR for "reset to zero before returning from rehash", which is at least a working fix, or do you prefer to keep investigating to make sure it's the correct fix?

GilesBathgate:
@sloriot @afabri This seems to be an alternative fix for just https://github.com/CGAL/cgal/commit/7d0d6a06bed91501bd1c5d6d95f2e36659e49919 which I write only to shed light on the investigation:

```diff
@@ -336,9 +336,10 @@ void chained_map<T, Allocator>::clear()
 
   for (chained_map_item item = table ; item != table_end ; ++item)
     destroy(item);
-  alloc.deallocate(table, table_end - table);
+  size_t current_size = table_end - table;
+  alloc.deallocate(table, current_size);
 
-  init_table(min_size);
+  init_table(current_size);
 }
```

PR #6312, opened 2022-02-09
Remove some wide 'PMP' namespace declarations

MaelRL:
## Summary of Changes

There are a few `namespace PMP =` living at namespace level within the headers of package `Polygon_mesh_processing` (`CGAL::PMP` from isotropic remeshing for example). It's never within the global namespace so it's not particularly dangerous, but from time to time I wonder how did my compiler recognize the `PMP` I forgot to declare. It's also quite specific to this package to have these aliases.

I also removed those in `CGAL::Polygon_mesh_processing::Corefinement::PMP` for the same reason. Look-up can find the function without even putting the namespace, but I can put an alias within the function.

If these aliases are ok to most people, I can just cancel the PR.

## Release Management

* Issue(s) solved (if any): -


sloriot:
Successfully tested in CGAL-5.5-Ic-18

PR #6316, opened 2022-02-11
Use non-deprecated API for boost progress

sloriot:


sloriot:
Successfully tested in CGAL-5.5-Ic-25

PR #6323, opened 2022-02-15
[Intersections_2] Make the crossing point computation deterministic

sloriot:
Always use the same ordered segments

### DIFF
--- a/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
+++ b/Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
@@ -34,6 +34,22 @@ namespace Intersections {
 
 namespace internal {
 
+// indices of lexicographically smallest endpoints
+// depending on config parameter of S2S2_inter_info.
+// {s1_id0,s1_id1,s2_id0,s2_id1}
+constexpr int s2s2_id[8][4] =
+{
+//  seg1   seg2
+  { 0,1,   0,1 },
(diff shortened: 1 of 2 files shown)


Intersections_2/include/CGAL/Intersections_2/Segment_2_Segment_2.h
```diff
@@ -215,8 +242,8 @@ do_intersect_with_info(const typename K::Segment_2 &seg1,
     typename K::Compare_xy_2 compare_xy;
 
   // first try to filter using the bbox of the segments
-    if (less_xy(A2,B1)
-     || less_xy(B2,A1))
+    if (certainly(less_xy(A2,B1))
```
sloriot, line 245: @afabri this change is not related to the patch but I fixed it while I was in this file. Do you agree that it does not make sense to have a filtered failure in that case?

afabri: Latest the `switch(make_certain(..))` will lead to the exception being thown, so we can throw it here as well.

lrineau:
@sloriot During the rebase of this PR, I could not rebase it on `5.3.x-branch`, because of important conflicts. I have rebased it on `5.4.x-branch`. I let you rebase it on `5.3.x-branch` if you think that is important.

sloriot:
Successfully tested in CGAL-5.5-Ic-25 (5.4 is fine)

PR #6265, opened 2022-01-21
Segment_delaunay_graph: fstream -> sstream

afabri:
## Summary of Changes

Do not write to files to avoid collisions during parallel execution of tests.

## Release Management

* Affected package(s):  SDG_2 and SDG_Linf_2

PR #6329, opened 2022-02-17
PSP: ofstream -> sstream

afabri:
## Summary of Changes

Just as in PR #6265 we have two test executable which may dp I/O with the same file, which leads to assertions as in this [teststuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-22/Point_set_processing_3/TestReport_Sosno_MSVC-2019-Community-Release.gz). 

## Release Management

* Affected package(s): PSP


sloriot:

PR #6340, opened 2022-02-21
Added missing copy constructor && assignment

efifogel:
## Summary of Changes

Added missing copy constructors and copy assignments to several classes that store pointers to objects that are either passed in via user-defined constructors or allocated. The default implementations caused segmentation faults.

The pointers and the corresponding flags that indicate ownerships should be eventually replaced with a smart pointers. Meanwhile, the copy constructors and copy assignments prevent double deletion. Notice that once a copy constructor (assignment) is present, the move constructor (assignment) is implicitly not generated anyway.

## Release Management

* Affected package(s): Minkowski_sum_2
* Issue(s) solved (if any): 
* Link to compiled documentation (obligatory for small feature)


sloriot:

PR #6344, opened 2022-02-22
Add missing static keyword

### LINKED ISSUE #6343: Inappropriate use of non-static constexpr variables
## Issue Details

I have found inappropriate usages of non-static `constexpr` variables in the following files:
- `Intersections_3/include/CGAL/Intersections_3/internal/Iso_cuboid_3_Plane_3_intersection.h`
- `Intersections_3/include/CGAL/Intersections_3/internal/Plane_3_Tetrahedron_3_intersection.h`
- `Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/clip.h`

The `constexpr` specifier, for a variable, does not imply that a variable is initialized at compile-time. It only specifies that the value of the variable *can be* evaluated at compile time. And `constexpr` does not imply any storage duration.

```c++
constexpr int f() { return 0; }

int g() {
  constexpr int i = f();
  return i;
}
```

Here the variable `i` has *automatic storage duration*. That means that is a variable living on the stack and, depending on the compiler choice, it can be initialized at each call of `g()`.

If instead the variable was `static constexpr`:
```c++
int g() {
  static constexpr int i = f();
  return i;
}
```
then the variable would be guaranteed to be initialized only once. And, as it is `constexpr` the compiler can avoid the penalty of the hidden variable guard (that makes the construction of the static variable thread-safe).

### References
- https://stackoverflow.com/q/13865842/1728537
- https://www.youtube.com/watch?v=4pKtPWcl1Go

## Release Management

The bug has been there since https://github.com/CGAL/cgal/pull/5435 (merged in CGAL versions 5.1.3, 5.2.1, and 5.3).

sloriot:
Fixes  #6343

### DIFF
--- a/Intersections_3/include/CGAL/Intersections_3/internal/Iso_cuboid_3_Plane_3_intersection.h
+++ b/Intersections_3/include/CGAL/Intersections_3/internal/Iso_cuboid_3_Plane_3_intersection.h
@@ -60,7 +60,7 @@ intersection(
     } };
 
   // description of faces of the bbox
-  constexpr std::array<int, 24> face_indices
+  static constexpr std::array<int, 24> face_indices
     { { 0, 1, 2, 3,
         2, 1, 5, 6,
         3, 2, 6, 7,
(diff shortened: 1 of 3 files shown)


sloriot:

PR #6348, opened 2022-02-23
Never try to fix when a polyline does not split a patch

sloriot:
Relates to issue reveled by  #6209


sloriot:
Successfully tested in CGAL-5.5-Ic-27

PR #6352, opened 2022-02-24
ignore CDT2 if normal is null

sloriot:
should also fix [this warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-27/Polygon_mesh_processing_Examples/TestReport_lrineau_ArchLinux-CXX17-Release.gz)


sloriot:

sloriot:
bonus point: the warning is gone.

PR #6356, opened 2022-02-25
pedantic changes

ffteja:
## Summary of Changes
Hi,

I compile with GCC 10.3 with pedantic error checking https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-pedantic-1
As CGAL is header based, I need either to make it compliant (for the few CGAL files I depend) or disable the pedantic check for the objects that depend on CGAL.

In order to contribute, here has few easy fixes.

## Release Management

* Affected package(s): 5.4

Please do compilation checks on your side of course.


lrineau (approved):

sloriot:

PR #6359, opened 2022-02-25
Fix squared_distance(Point_3, Triangle_3)

MaelRL:
## Summary of Changes

I blundered in commit https://github.com/CGAL/cgal/pull/5527/commits/b95c60fc9fe671eb42a6811e137ab48167657db2: one orientation is of course not sufficient to determine which segment realizes the min.

I'm a little perplexed that nothing exploded, but I think it's because e.g. during an AABB Tree computation, if the computation is wrong for an edge, it still gave the correct result for the same edge seen from the neighboring face.

Reverting to the previous implementation.

## Release Management

* Affected package(s): `Distance_3`
* Issue(s) solved (if any): -

### DIFF
--- a/Distance_3/include/CGAL/Distance_3/Point_3_Triangle_3.h
+++ b/Distance_3/include/CGAL/Distance_3/Point_3_Triangle_3.h
@@ -68,8 +68,8 @@ squared_distance_to_triangle_RT(const typename K::Point_3& pt,
 
   if(normal == NULL_VECTOR)
   {
-    // The case normal==NULL_VECTOR covers the case when the triangle
-    // is colinear, or even more degenerate. In that case, we can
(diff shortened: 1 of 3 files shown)


lrineau:
Fix #5527, merged in 5.4-beta1.

sloriot:

lrineau:
Mael, why the label https://github.com/CGAL/cgal/labels/not%20yet%20approved?

MaelRL:
The PR could use another set of eyes on a review of the algorithm, maybe there's a better (simple) approach.

lrineau:
On the other hand, it seems this PR fixes a bug. Maybe there could be a better code, but we have the urgency to fix the bug, and rollback to the previous implementation. I will merge it.

PR #6361, opened 2022-02-25
Add CGAL::Color operator< and return values

maximecharriere:
## Summary of Changes

This PR add 2 features to the `CGAL::Color` class.  

### 1. Add `CGAL::Color::operator<`

I'm personnaly using `Color` as a key of a `std::map`. The map therefore needs a [comparator](https://en.cppreference.com/w/cpp/named_req/Compare). 
Instead of implementing a comparator separately in my code, I think it might be useful to add the `operator<` in the class.   
To do so, I compared the two `m_data` arrays that call internally the [`std::lexicographical_compare`](https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare).

### 2. Add return value

It could be useful to return the color when calling `set_hsv()` and `set_rgb()`, and not `void`.
For example in my code I'm doing this:
```c++
CGAL::Color newColor;
newColor.set_hsv(angle, 100, 100);
put(colorPm, vertex, newColor);
```
But with a return value I could be written on one line.
```c++
put(colorPm, vertex, CGAL::Color().set_hsv(angle, 100, 100));
```

## Release Management

* Affected package(s): [IO Streams](https://doc.cgal.org/latest/Stream_support/index.html)
* Small Feature
* Link to compiled documentation (obligatory for small feature) [CGAL::IO::Color Class Reference](https://doc.cgal.org/latest/Stream_support/classCGAL_1_1IO_1_1Color.html)

### DIFF
--- a/Stream_support/include/CGAL/IO/Color.h
+++ b/Stream_support/include/CGAL/IO/Color.h
@@ -132,6 +132,11 @@ public:
     return !( (*this) == c);
   }
 
+  bool operator<(const Color& c) const
+  {
+      return m_data < c.to_rgba();
+  }
+
   unsigned char r() const { return red(); }
   unsigned char g() const { return green(); }
   unsigned char b() const { return blue(); }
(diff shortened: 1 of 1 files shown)


lrineau (approved):

mglisse:
> I'm personnaly using `Color` as a key of a `std::map`. The map therefore needs a [comparator](https://en.cppreference.com/w/cpp/named_req/Compare).

Out of curiosity, is `hash` supported, in case you later decide to change your map to an unordered_map?

PR #5639, opened 2021-04-23
Global clean-up of Named Parameters

maxGimeno:
## Summary of Changes

- replaces various overloads of functions using named parameters with a default: as a consequence, it is now obvious in the documention of all functions that the named parameters is optional.
- add and document default type `CGAL::parameters::Default_named_parameters`
- add and document default function `CGAL::parameters::use_default_values()` (replaces `CGAL::parameters::all_default()` that is now deprecated)
- update the named parameter mechanism of `Point_set_processing_3`: in order to support `CGAL::Point_set_3`, `CGAL::parameters::Point_set_processing::all_default(ps)` has been introduced (but not documented). The code has been update to also use  `CGAL::parameters::all_default()` and delegate the former mechanism to an internal helper class used by all functions (does not increase the codebase, even decreases it)
- Move doxygen macro definition for named parameters in doxygen (no longer need to do special hacks per package)

## Release Management

* Affected package(s):PMP/BGL/STL_Extension/SMS/...

## TODO
- [x] use default values and remove useless overloads + update dependencies
- [x] move into STL_extension + update path
- [X] rename PR

**Before**
Screenshot from 2022-01-06 19-09-42

**After**
Screenshot from 2022-01-06 19-08-10

PR #6365, opened 2022-02-28
Non-inlined doc NP update

sloriot:
Follow up of #5639


sloriot:

PR #6366, opened 2022-02-28
Consistent count in the doc

mglisse:
We already say 5 a few lines below.


sloriot:

PR #6368, opened 2022-03-01
remove deprecated recommendation

### LINKED ISSUE #5752: Hard to reach link (behind login) in documentation
## Issue Details

In the file `Documentation/doc/Documentation/Developer_manual/Chapter_code_format.txt` there is the link to:
```
https://sympa.inria.fr/sympa/arc/cgal-develop/2010-10/msg00118.html
```
this is behind a login and it looks like one can get access through a login process though it is not 100% clear what the restrictions / necessities are.
This is a bit user unfriendly (in my opinion) and would also be time consuming.

Might be better to include the mentioned file verbatim into the documentation.

sloriot:
fix #5752

thanks @mglisse


sloriot:

PR #6370, opened 2022-03-01
Fix not using / forwarding traits

### LINKED ISSUE #6369: remove_degenerate_edges() fails to pass named parameters to degenerate_edges()
## Issue Details

Hello all, 

I am making use of the remove_degenerate_edges() function found in repair_degeneraces.h, in an effort to clean up degenerate edges/faces found within my triangulated mesh. The triangles that I am attempting to clean up contain 2 vertices which are _almost_ the same exact point. An example of the two _almost_ same vertices are {75.78643271209685, 75.78643271209685, 92.50000000000003} and {75.78643271209688, 75.78643271209685, 92.50000000000003}. Due to remove_degenerate_edges() not viewing these as the same point, and thus not removing them, I am passing in a GeomTraits object which has an Equal_3 object custom defined. To my surprise, this still did not remove the _almost_ duplicate points. Upon investigating, the named parameters that are passed into remove_degenerate_edges() are not passed along to the degenerate_edges() call. As a result, degenerate_edges() is using the CGAL defined geometry traits and no "degenerate" edges are being picked up. I was able to fix this issue by simply passing along the named parameters, but I am unsure if this leads to any unintended consequences. From a bit of investigation it doesn't seem that there are any unintended consequences, but I can not say that for sure. I understand that this is an undocumented function, so if this is not priority for the CGAL team right now it should not be a problem for me to make a PR with such a fix. Just let me know.

Thanks, 
Chase

## Source Code

Original code:
```
    std::set<edge_descriptor> degenerate_edges_to_remove;
    degenerate_edges(local_edge_range, tmesh, std::inserter(degenerate_edges_to_remove,
                                                            degenerate_edges_to_remove.end()));

```

Fix: 
```
    std::set<edge_descriptor> degenerate_edges_to_remove;
    degenerate_edges(local_edge_range, tmesh, std::inserter(degenerate_edges_to_remove,
                                                            degenerate_edges_to_remove.end()), np);
```

## Environment

* CGAL version: 5.3, but issue appears to be present in 5.4 as well

MaelRL:
## Release Management

* Issue(s) solved (if any): fix #6369

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -186,6 +186,7 @@ bool is_collapse_geometrically_valid(typename boost::graph_traits<TriangleMesh>:
 }
 */
 
+// @todo handle boundary edges
 template <class TriangleMesh, typename VPM, typename Traits>
 boost::optional<typename Traits::FT>
 get_collapse_volume(typename boost::graph_traits<TriangleMesh>::halfedge_descriptor h,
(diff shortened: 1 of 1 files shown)


sloriot:

PR #6373, opened 2022-03-02
Fix make_icosahedron() having wrong radius

### LINKED ISSUE #6064: CGAL::make_icosahedron creates mesh of wrong scale
According to the documentation, the function `CGAL::make_icosahedron`, takes the circumradius of the icosahedron, but as it is now, the argument is used as half the edge length, rather than radius.  To make it the radius, the mesh would have to be scaled by `sqrt(φ + 2)`, but this would make all components of the vertices approximate.  Perhaps simply changing the documentation would be simpler.

MaelRL:
## Summary of Changes

I chose to correct the coordinate computations and leave the doc untouched because it is more intuitive for a user to give the center and radius of the circumscribing sphere than to give the edge length (or half of the edge length, or etc.).

## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): fix #6064

### DIFF
--- a/BGL/include/CGAL/boost/graph/generators.h
+++ b/BGL/include/CGAL/boost/graph/generators.h
@@ -19,6 +19,10 @@
 #include <CGAL/Random.h>
 #include <CGAL/function_objects.h>
 
+#include <array>
+#include <iterator>
+#include <vector>
+
 namespace CGAL {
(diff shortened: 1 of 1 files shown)


sloriot:

PR #6377, opened 2022-03-03
Triangulation 2: Example for Boolean operations with CDT_2

afabri:
## Summary of Changes

This example does twice the flooding and stores that it is in input polygon A and B. 
We do not extract boundary polygons yet.  This necessitates that A and B are both simple polygons.

## Release Management

* Affected package(s): Trinagulation_2


sloriot:
Shall we now postpone to 5.7?

sloriot:
Replaced by https://github.com/CGAL/cgal/pull/8510

PR #6378, opened 2022-03-03
Nef_3: Sphere circle constructors.

GilesBathgate:
## Summary of Changes

In the constructor of `Sphere_circle` a geometric test is made `h.d() != 0`
https://github.com/CGAL/cgal/blob/077937383555a61e5b23460fc21cf75f3f005018/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h#L81-L87
This is necessary because in the general case it isn't known whether the plane passed to the sphere_circle passes through the origin.

> I didn't really know much about plane equations before looking at this code but as I understand in the equation `ax+by+cz+d=0` the x,y,z intercepts are given by `x = -d/a` , `y = -d/b`, `z = -d/c` if we set d=0 then all the intercepts are zero meaning the plane passes through the origin.

However, in the case of `polygon_mesh_to_nef_3.h` the sphere_circle is constructed from a `Plane_3(CGAL::ORIGIN, normals[I])` that passes through the origin, and in the case of `SNC_constructor.h`, the sphere_circle becomes the circle containing the origin and parallel to the plane.

This pull request adds two new constructors to facilitate these usages, and avoid the geometric test.

Finally, in the function opposite, the call to `Base::opposite` implicitly invokes the constructor too. Since in this case, all that is needed is a copy of a handle, a private constructor added just for this case.

Unfortunately, the performance improvement is quite small in this pull request, only about 3% faster Nef_3_benchmark, but I thought the change was simple enough to propose.

There is also some cleanup of the existing code calling of base functions

## Release Management

* Issue(s) solved (if any): cleaning,speed

### DIFF
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -66,8 +66,7 @@ create any great circle that contains $p$ and $q$.}*/
 {
   if ( p != q.antipode() ) {
     Point_3 po = CGAL::ORIGIN + Base::orthogonal_vector();
-    //TODO replace Point_3(CGAL::ORIGIN), with CGAL::ORIGIN below
-    if (R_().orientation_3_object()(Point_3(CGAL::ORIGIN), Point_3(p),
+    if (R_().orientation_3_object()(CGAL::ORIGIN, Point_3(p),
                                     Point_3(q), po) != CGAL::POSITIVE )
       *this = opposite();
   } else {
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -85,16 +85,17 @@ Sphere_circle(const Plane_3& h) : Base(h)
 |h|. If |h| does not contain the origin, then |\Mvar| becomes the
 circle parallel to |h| containing the origin.}*/
 {
-  if(h.d() != 0)
+  if(!is_zero(h.d()))
     *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
 }
 
 Sphere_circle(const Origin& o, const Vector_3& v) : Base(o,v) {}
 
-Sphere_circle(const Origin&, const Plane_3& h) : Base(h.a(),h.b(),h.c(),RT(0)) {}
+Sphere_circle(const Origin&, const Plane_3& h) : Base(h.a(),h.b(),h.c(),RT(0))
+{/* Even if |h| does not contain the origin, the circle will contain the origin
+and be parallel to |h| */}
 
 Sphere_circle(const RT& x, const RT& y, const RT& z) : Base(x,y,z,0) {}
-
 /*{\Mcreate creates the circle orthogonal to the vector $(x,y,z)$.}*/
 
 Sphere_circle(Sphere_circle<R> c, const Sphere_point<R>& p)
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -17,6 +17,7 @@
 
 
 #include <CGAL/basic.h>
+#include <CGAL/Nef_S2/Sphere_point.h>
 
 namespace CGAL {
 
@@ -47,6 +48,7 @@ typedef std::pair< Sphere_segment<R>,Sphere_segment<R> >
 typedef typename R_::Plane_3 Plane_3;
 typedef typename R_::Line_3 Line_3;
 typedef typename R_::Point_3 Point_3;
+typedef typename R_::Vector_3 Vector_3;
 typedef Sphere_circle<R_> Self;
 typedef typename R_::Plane_3 Base;
 
@@ -78,15 +80,20 @@ create any great circle that contains $p$ and $q$.}*/
   }
 }
 
- Sphere_circle(const Plane_3& h) : Base(h)
+Sphere_circle(const Plane_3& h) : Base(h)
 /*{\Mcreate creates the circle of $S_2$ corresponding to the plane
 |h|. If |h| does not contain the origin, then |\Mvar| becomes the
 circle parallel to |h| containing the origin.}*/
 {
-  if(h.d() != 0) *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
+  if(h.d() != 0)
+    *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
 }
 
-Sphere_circle(const RT& x, const RT& y, const RT& z): Base(x,y,z,0) {}
+Sphere_circle(const Origin& o, const Vector_3& v) : Base(o,v) {}
+
+Sphere_circle(const Origin&, const Plane_3& h) : Base(h.a(),h.b(),h.c(),RT(0)) {}
+
+Sphere_circle(const RT& x, const RT& y, const RT& z) : Base(x,y,z,0) {}
 
 /*{\Mcreate creates the circle orthogonal to the vector $(x,y,z)$.}*/
 
@@ -102,10 +109,9 @@ Sphere_circle(Sphere_circle<R> c, const Sphere_point<R>& p)
 }
 
 /*{\Moperations 4 2}*/
-
 Sphere_circle<R> opposite() const
 /*{\Mop returns the opposite of |\Mvar|.}*/
-{ return Base::opposite(); }
+{ return Sphere_circle<R>(Base::opposite(),Private_tag{}); }
 
 bool has_on(const Sphere_point<R>& p) const
 /*{\Mop returns true iff |\Mvar| contains |p|.}*/
@@ -117,7 +123,7 @@ Plane_3 plane() const { return Base(*this); }
 Plane_3 plane_through(const Point_3& p) const
 /*{\Mop returns the plane parallel to |\Mvar| that
 contains point |p|.}*/
-{ return Plane_3(p,((Base*) this)->orthogonal_direction()); }
+{ return Plane_3(p,((Base*) this)->orthogonal_vector()); }
 
 Sphere_point<R> orthogonal_pole() const
 /*{\Mop returns the point that is the pole of the
@@ -133,6 +139,10 @@ Sphere_segment_pair split_at_xy_plane() const;
 of splitting |\Mvar| at the $x$-$y$-coordinate plane if |\Mvar|
 is not part of it. Otherwise |\Mvar| is split at the
 $x$-$z$-coordinate plane.}*/
+
+private:
+struct Private_tag{};
+Sphere_circle(const Plane_3& h,Private_tag) : Base(h){}
 
 }; // Sphere_circle<R>
 
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -112,7 +112,7 @@ Sphere_circle(Sphere_circle<R> c, const Sphere_point<R>& p)
 /*{\Moperations 4 2}*/
 Sphere_circle<R> opposite() const
 /*{\Mop returns the opposite of |\Mvar|.}*/
-{ return Sphere_circle<R>(Base::opposite(),Private_tag{}); }
+{ return Sphere_circle<R>(Base::opposite(),Assume_d_equal_0{}); }
 
 bool has_on(const Sphere_point<R>& p) const
 /*{\Mop returns true iff |\Mvar| contains |p|.}*/
@@ -142,8 +142,8 @@ is not part of it. Otherwise |\Mvar| is split at the
 $x$-$z$-coordinate plane.}*/
 
 private:
-struct Private_tag{};
-Sphere_circle(const Plane_3& h,Private_tag) : Base(h){}
+struct Assume_d_equal_0{};
+Sphere_circle(const Plane_3& h, Assume_d_equal_0) : Base(h){}
 
 }; // Sphere_circle<R>
 
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -63,7 +63,7 @@ $q$ are not antipodal on $S_2$, then this circle is unique and oriented
 such that a walk along |\Mvar| meets $p$ just before the shorter segment
 between $p$ and $q$. If $p$ and $q$ are antipodal of each other then we
 create any great circle that contains $p$ and $q$.}*/
-{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + ((Base*) this)->orthogonal_vector();
+{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + Base::orthogonal_vector();
   if ( p != q.antipode() ) {
     if (R_().orientation_3_object()(p1,Point_3(p),
                                     Point_3(q), p4) != CGAL::POSITIVE )
@@ -123,12 +123,12 @@ Plane_3 plane() const { return Base(*this); }
 Plane_3 plane_through(const Point_3& p) const
 /*{\Mop returns the plane parallel to |\Mvar| that
 contains point |p|.}*/
-{ return Plane_3(p,((Base*) this)->orthogonal_vector()); }
+{ return Plane_3(p,Base::orthogonal_vector()); }
 
 Sphere_point<R> orthogonal_pole() const
 /*{\Mop returns the point that is the pole of the
 hemisphere left of |\Mvar|.}*/
-{ return CGAL::ORIGIN+((Base*) this)->orthogonal_vector(); }
+{ return CGAL::ORIGIN+Base::orthogonal_vector(); }
 
 Sphere_segment_pair split_at(const Sphere_point<R>& p) const;
 /*{\Mop returns the pair of circle segments that is the result
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -63,15 +63,16 @@ $q$ are not antipodal on $S_2$, then this circle is unique and oriented
 such that a walk along |\Mvar| meets $p$ just before the shorter segment
 between $p$ and $q$. If $p$ and $q$ are antipodal of each other then we
 create any great circle that contains $p$ and $q$.}*/
-{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + Base::orthogonal_vector();
+{
   if ( p != q.antipode() ) {
-    if (R_().orientation_3_object()(p1,Point_3(p),
-                                    Point_3(q), p4) != CGAL::POSITIVE )
-      *this = Self(opposite());
+    Point_3 po = CGAL::ORIGIN + Base::orthogonal_vector();
+    //TODO replace Point_3(CGAL::ORIGIN), with CGAL::ORIGIN below
+    if (R_().orientation_3_object()(Point_3(CGAL::ORIGIN), Point_3(p),
+                                    Point_3(q), po) != CGAL::POSITIVE )
+      *this = opposite();
   } else {
-    /* previous method was: *this = Self(Plane_3(p1,q-p));
-       but p, q don't belong to he plane ((0,0,0), q-p) */
-
+    /* previous method was: *this = Self(Plane_3((0,0,0),q-p));
+       but p, q don't belong to the plane ((0,0,0), q-p) */
     if(!Line_3(p,q).has_on(Point_3(1,0,0)))
       *this = Self(Plane_3(p,q,Point_3(1,0,0)));
     else
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
@@ -230,14 +230,14 @@ Point_3 orthogonal_pole() const
 { return CGAL::ORIGIN + sphere_circle().orthogonal_vector(); }
 
 CGAL::Orientation source_orientation(const CGAL::Sphere_point<R>& p) const
-{ return orientation(Point_3(CGAL::ORIGIN),
+{ return orientation(CGAL::ORIGIN,
                      orthogonal_pole(),
                      source(),
                      p);
 }
 
 CGAL::Orientation target_orientation(const CGAL::Sphere_point<R>& p) const
-{ return orientation(Point_3(CGAL::ORIGIN),
+{ return orientation(CGAL::ORIGIN,
                      target(),
                      orthogonal_pole(),
                      p);
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
@@ -20,6 +20,7 @@
 #include <CGAL/Handle_for.h>
 #include <CGAL/Nef_S2/Sphere_point.h>
 #include <CGAL/Nef_S2/Sphere_circle.h>
+#include <CGAL/Kernel/global_functions.h>
 #include <vector>
 
 namespace CGAL {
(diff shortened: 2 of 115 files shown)


Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -117,12 +123,12 @@ Plane_3 plane() const { return Base(*this); }
 Plane_3 plane_through(const Point_3& p) const
 /*{\Mop returns the plane parallel to |\Mvar| that
 contains point |p|.}*/
-{ return Plane_3(p,((Base*) this)->orthogonal_direction()); }
+{ return Plane_3(p,Base::orthogonal_vector()); }
```
afabri, line 126: Is this change just a cleanup?

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -78,15 +80,20 @@ create any great circle that contains $p$ and $q$.}*/
...
 {
-  if(h.d() != 0) *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
+  if(h.d() != 0)
+    *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
 }
 
-Sphere_circle(const RT& x, const RT& y, const RT& z): Base(x,y,z,0) {}
+Sphere_circle(const Origin& o, const Vector_3& v) : Base(o,v) {}
+
+Sphere_circle(const Origin&, const Plane_3& h) : Base(h.a(),h.b(),h.c(),RT(0)) {}
```
afabri, line 94: We should add an assertion that the plane goes through the origin or at least put that in a comment.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -61,7 +63,7 @@ Sphere_circle(const Sphere_point<R>& p, const Sphere_point<R>&q)
 such that a walk along |\Mvar| meets $p$ just before the shorter segment
 between $p$ and $q$. If $p$ and $q$ are antipodal of each other then we
 create any great circle that contains $p$ and $q$.}*/
-{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + ((Base*) this)->orthogonal_vector();
+{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + Base::orthogonal_vector();
   if ( p != q.antipode() ) {
     if (R_().orientation_3_object()(p1,Point_3(p),
```
afabri, line 68: No idea if this line shows up often in vtune, but maybe it makes sense to have an orientation predicate for the case that one point is the origin.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -134,6 +140,10 @@ of splitting |\Mvar| at the $x$-$y$-coordinate plane if |\Mvar|
 is not part of it. Otherwise |\Mvar| is split at the
 $x$-$z$-coordinate plane.}*/
 
+private:
+struct Private_tag{};
```
afabri, line 144: The name must be improved or we must even come up with a better design than such a tag class. @sloriot any idea? 
If we keep it, it should have a name telling us what we want to achieve.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -78,15 +80,20 @@ create any great circle that contains $p$ and $q$.}*/
...
 }
 
- Sphere_circle(const Plane_3& h) : Base(h)
+Sphere_circle(const Plane_3& h) : Base(h)
 /*{\Mcreate creates the circle of $S_2$ corresponding to the plane
 |h|. If |h| does not contain the origin, then |\Mvar| becomes the
 circle parallel to |h| containing the origin.}*/
 {
-  if(h.d() != 0) *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
+  if(h.d() != 0)
```
afabri, line 88: While we are at it, maybe use `is_zero()`here.

afabri:
> Unfortunately, the performance improvement is quite small in this pull request, only about 3% faster Nef_3_benchmark, but I thought the change was simple enough to propose.

3% globally on a Boolean operation?  I find that quite impressive.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: Yes, I've added a comment to the change summary.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: In this constructor, it assumes that the given plane does not pass through the origin, and sets d explicitly to 0. So I think a comment is appropriate. I will add one.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: I initially investigated `!is_zero()` I think they ultimately use the same filtering. But I will add to better show intention.

GilesBathgate:
> 3% globally on a Boolean operation? I find that quite impressive.

I guess what I meant was Its unfortunate for my usage within RapCAD, I'm wanting to find much larger gains. I've recently added concurrency using QtConcurrent (dependant on 5.4 multiple thread support). This does also help, but not as much as I expected.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: Oh, wait. I changed `orthogonal_direction` to `orthogonal_vector`. I don't think it makes a difference, down the call chain it calls v.to_direction() anyway.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
lrineau: That would be the same as the `Orientation_3::operator()(Vector_3, Vector_3, Vector_3)`, but with points instead of vectors.

I think `Sphere_point<K>` should actually derive from `K::Vector_3`, and not `K::Direction_3` and not `K::Point_3`, and that would make the code simpler.

GilesBathgate: @lrineau That makes sense, a point on the sphere is actually a vector from the origin of the sphere towards the point on the surface of the sphere.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: It seems to be a name used elsewhere: https://github.com/CGAL/cgal/blob/b575892a164d83e6d47a960839f86ae24dca4114/STL_Extension/include/CGAL/Object.h#L135

GilesBathgate: @lrineau I wasn't intending to imply a 'member access modifier' by the name 'Private' in the name of the tag. Rather, it's just an implementation detail really, so perhaps `CGAL::internal_tag` might be more appropriate. Like I say I think I chose `Private_tag` because it was used elsewhere in the CGAL codebase. 
@afabri I don't like the design either, but it is a neat workaround for what I consider a limitation of C++ when you need two copy constructors that take the same parameter type, but with different behaviour.

lrineau: Anyway, a name like `Skip_the_test` or `Assume_d_equal_0` is more informative than something related to "private" or "internal".

GilesBathgate: @lrineau I've changed it to `Assume_d_equal_0` for now. I mulled over `Assume_origin_intercept`, or just `Intercepts_origin`...

Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
```diff
@@ -37,7 +38,7 @@ Sphere_segment_rep() { ps_ = pt_ = Point(); c_ = Circle(); }
 
 Sphere_segment_rep(const Point& p1, const Point& p2,
                    bool shorter_arc=true) :
-  ps_(p1), pt_(p2), c_(Plane_3(p1,p2,Point_3(CGAL::ORIGIN)))
+  ps_(p1), pt_(p2), c_(CGAL::ORIGIN,CGAL::orthogonal_vector(p1,p2,Point_3(CGAL::ORIGIN)))
```
afabri, line 41: This should not be a global function, but a functor of the kernel.   And, @lrineau, correct me if I am wrong, is that not the crossproduct of `Vector_3(p1)` and `Vector_3(p2)` as the third point is the origin?

GilesBathgate: I can change this to:
```c++
c_(CGAL::ORIGIN,Plane_3(p1,p2,Point_3(CGAL::ORIGIN)).orthogonal_vector())
```
I just thought constructing a plane just for its orthogonal vector seemed unnecessary.

afabri: But in case I am right with cross product it would even be simpler, wouldn't it?
```
c_(CGAL::ORIGIN, R_().cross_product_3_object()(Vector_3(p1), Vector_3(p2)));
```

GilesBathgate: Ah I see. I was thinking It would still have to use the global function `CGAL::cross_product(Vector_3(p1), Vector_3(p2))` I misunderstood your suggestion.

GilesBathgate: > correct me if I am wrong, is that not the crossproduct of

@afabri Assuming using the crossproduct is correct I think it would be:
```c++
R_().construct_cross_product_vector_3_object()(p1-CGAL::ORIGIN, p2-CGAL::ORIGIN)
```

I wonder if it's worth cleaning up this too:
```c++
 Plane_3 h(Point_3(CGAL::ORIGIN),(p1-CGAL::ORIGIN));
    c_ = Sphere_circle<R_>(Plane_3(Point_3(CGAL::ORIGIN),h.base1()));
 ```

Comment (afabri) [reply Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h:41]: `Vector_3(p)` was more a symbolic writing to convey the idea  with the crossproduct function.  I am only wondering how much we will pay for    `p1 - ORIGIN`   as this constructs  nodes in the DAG of the Epeck objects. But maybe just benchmarking is better than meditating about this  kind of problem.

Comment (afabri) [reply Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h:41]: Which line do you refer to  when you write "I am wondering ...." ?

Comment (GilesBathgate) [reply Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h:41]: I think the assertion that cross_product == orthogonal_vector of three points when the third is the origin is correct

```c++
#include <CGAL/Kernel/global_functions.h>

using K = CGAL::Exact_predicates_exact_constructions_kernel;
using Point = CGAL::Point_3<K>;

int main(int, char*[])
{
	CGAL::Random_points_on_sphere_3<Point> rnd;

	for(int i=0; i<1000; ++i)
	{
		Point p1=*rnd++;
		Point p2=*rnd++;
		if(CGAL::cross_product(p1-CGAL::ORIGIN,p2-CGAL::ORIGIN) != CGAL::orthogonal_vector(p1,p2,Point(CGAL::ORIGIN)))
		{
			throw 1;
		}
	}
	return 0;
}
```

afabri:
Can you provide a self contained cpp on gist.github.com where you observe the gain of 3% so that I can try to reproduce.

Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
afabri: > is it possible to add to the `Construct_base_vector_3` functor?
> 
> ```c++
> result_type operator()( const Vector_3& v, int index ) const { ... }
> ```

We can add any documented kernel functors  as the requirements of the traits class are not documented.

lrineau: > I think the assertion that cross_product == orthogonal_vector of three points when the third is the origin is correct

It is correct, because mathematically that is correct.

The whole situation here is because those sphere "points" should actually be vectors, or best: directions.

GilesBathgate:
@afabri The benchmark results are here: https://github.com/afabri/Nef_3_benchmark/runs/5412199023?check_suite_focus=true

```
Comparing ./baseline to ./contender
Benchmark                                   Time             CPU      Time Old      Time New       CPU Old       CPU New
------------------------------------------------------------------------------------------------------------------------
BM_difference_mean                       -0.0247         -0.0248         29070         28352         29067         28346
BM_difference_median                     -0.0302         -0.0303         29086         28209         29083         28203
BM_intersection_mean                     -0.0223         -0.0224         28262         27631         28259         27627
BM_intersection_median                   -0.0226         -0.0227         28302         27662         28300         27659
...
```

GilesBathgate:
@afabri Here is a quick gist: https://gist.github.com/GilesBathgate/6a50e72772fdf7c36f90cb7ad5305fd1 

```
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_Plane        18203 ns        18203 ns        36931
BM_NoPlane        579 ns          579 ns      1120330
```

Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
GilesBathgate: > The whole situation here is because those sphere "points" should actually be vectors, or best: directions.

I agree, whilst looking at the code your suggestion became more compelling. But I think it should be a separate Issue/Pull request and if someone wants to implement it ;)

GilesBathgate: @afabri No what I really had in mind was just this:

```c++
 Plane_3 h(CGAL::ORIGIN,(p1-CGAL::ORIGIN));
 c_ = Sphere_circle<R_>(CGAL::ORIGIN,h.base1());
```

But your other comments got me wondering...

afabri:
@CGAL/developers    I would like to add an overload with 3 points `p, q, r` to[ `Kernel::Orientation_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1Orientation__3.html)  which applies the operator to `p - ORIGIN, q - ORIGIN, r - ORIGIN`.   Similar for the crossproduct of two vectors.      I want to do this as `p - ORIGIN`  is almost a cast for any kernel, but it is expensive for `Epeck` as this constructs a lazy object.  

I know that we have the distinction between point/vector/direction since CGAL 0.9, so the question is if such an overload is acceptable  or  a contradiction to principles.

GilesBathgate:
>  I would like to add an overload with 3 points `p, q, r` to `Kernel::Orientation_3` which applies the operator to `p - ORIGIN, q - ORIGIN, r - ORIGIN`. Similar for the crossproduct of two vectors. I want to do this as `p - ORIGIN` is almost a cast for any kernel, but it is expensive for `Epeck` as this constructs a lazy object.

Out of interest, how does this impact my proposed PR, would it replace this, or should parts of it still be needed?

GilesBathgate:
> But if you prefer we can do that in another PR,

I think everyone prefers smaller changesets 😉

mglisse:
> I would like to add an overload with 3 points `p, q, r` to[ `Kernel::Orientation_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1Orientation__3.html) which applies the operator to `p - ORIGIN, q - ORIGIN, r - ORIGIN`.

It could also be an overload that takes (Origin, Point, Point, Point), which would automatically be provided by the overload for 4 points when Point can be constructed implicitly from Origin, but could still be overloaded as an optimization. Or it could have a different name (orientation_with_origin, orientation_as_vectors, etc), we already have coplanar_orientation that computes "the orientation" of 3 points in 3D. Just listing possibilities.

afabri:
> > I would like to add an overload with 3 points `p, q, r` to[ `Kernel::Orientation_3`](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1Orientation__3.html) which applies the operator to `p - ORIGIN, q - ORIGIN, r - ORIGIN`.
> 
> It could also be an overload that takes (Origin, Point, Point, Point), which would automatically be provided by the overload for 4 points when Point can be constructed implicitly from Origin, but could still be overloaded as an optimization. Or it could have a different name (orientation_with_origin, orientation_as_vectors, etc), we already have coplanar_orientation that computes "the orientation" of 3 points in 3D. Just listing possibilities.

You are absolutely right with the overload with Origin as first parameter. Let me make a PR and check if there is a gain.

Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
```diff
@@ -37,14 +37,15 @@ Sphere_segment_rep() { ps_ = pt_ = Point(); c_ = Circle(); }
 
 Sphere_segment_rep(const Point& p1, const Point& p2,
                    bool shorter_arc=true) :
-  ps_(p1), pt_(p2), c_(Plane_3(p1,p2,Point_3(CGAL::ORIGIN)))
+  ps_(p1), pt_(p2),
+  c_(CGAL::ORIGIN,R_().construct_orthogonal_vector_3_object()(p1,p2,Point_3(CGAL::ORIGIN)))
```
afabri, line 41: Is it correct that the three parameters could be in any order?

afabri, line 41: Is it correct that the three parameters could be in any order?

lrineau: If the orientation ("sign" along the normal line) of the orthogonal vector is important, then no. The direction of the orthogonal vector will flip depending on the parity of the permutation. You can change to the order `ORIGIN, p1, p2` and it will not change the orthogonal vector.

GilesBathgate: I don't think so. If p1 and p2 have reversed the vector points in the opposite direction? 
Either way, I didn't change the order.

GilesBathgate: By the way, i found `R_().construct_orthogonal_vector_3_object()(p1,p2,Point_3(CGAL::ORIGIN))` to be marginally faster than `R_().construct_cross_product_vector_3_object()(p1-CGAL::ORIGIN, p2-CGAL::ORIGIN)`

afabri: I want to change, as I only want to add the overload with ` Origin` as first parameter. Thanks for the explanation about the parity rule.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -78,16 +81,22 @@ create any great circle that contains $p$ and $q$.}*/
...
 |h|. If |h| does not contain the origin, then |\Mvar| becomes the
 circle parallel to |h| containing the origin.}*/
 {
-  if(h.d() != 0) *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
+  if(!is_zero(h.d()))
+    *this = Plane_3(h.a(),h.b(),h.c(),RT(0));
 }
 
-Sphere_circle(const RT& x, const RT& y, const RT& z): Base(x,y,z,0) {}
+Sphere_circle(const Origin& o, const Vector_3& v) : Base(o,v) {}
```
GilesBathgate, line 93: @mglisse I should probably do the same in my own PR then. i.e `Origin o` instead of `const Origin&` o?

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
```diff
@@ -61,15 +63,16 @@ Sphere_circle(const Sphere_point<R>& p, const Sphere_point<R>&q)
...
 between $p$ and $q$. If $p$ and $q$ are antipodal of each other then we
 create any great circle that contains $p$ and $q$.}*/
-{ Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + ((Base*) this)->orthogonal_vector();
+{
   if ( p != q.antipode() ) {
-    if (R_().orientation_3_object()(p1,Point_3(p),
-                                    Point_3(q), p4) != CGAL::POSITIVE )
-      *this = Self(opposite());
+    Point_3 po = CGAL::ORIGIN + Base::orthogonal_vector();
+    //TODO replace Point_3(CGAL::ORIGIN), with CGAL::ORIGIN below
```
afabri, line 69: I think you can just change it as `Point_3` has a constructor from `Origin`.  The same for the other todo further below.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
mglisse: Indeed. If the function is inlined, both are equivalent. And if it isn't, the cost is negligible. Passing tag types by value is a nicer default when you have the choice. But note that we often end up passing tag types by reference, because they go through generic forwarders, and that's ok, not worth complicating the code.

GilesBathgate: I guess I just wanted the code to be consistent, but I see `const Origin&` elsewhere, so I guess it doesn't matter.

Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
GilesBathgate: It must be explicit, because I had a compile error until I merge your change.

afabri (approved):

afabri:
@sloriot this could go in the testsuite.   Note that @GilesBathgate has merged #6399

GilesBathgate:
@afabri Regarding the test-suite, do you use coverage analysis? I use [codecov.io](https://about.codecov.io/)

afabri:
We can change in [this ](https://github.com/CGAL/cgal/blob/42cff9ee28b416a37a054a183c6758c07ab918be/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h#L232) and the next function as well, right?

Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h
```diff
@@ -187,15 +188,15 @@ void split_halfcircle(Sphere_segment<R>& s1,
 bool is_short() const
 /*{\Mop a segment is short iff it is shorter than a halfcircle.}*/
 {
-  return R().orientation_3_object()(Point_3(CGAL::ORIGIN),
+  return R().orientation_3_object()(CGAL::ORIGIN,
```
GilesBathgate, line 191: >We can change in [this ](https://github.com/CGAL/cgal/blob/42cff9ee28b416a37a054a183c6758c07ab918be/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h#L232)and the next function as well, right?

@afabri Already done.

GilesBathgate:
>e can change in [this ](https://github.com/CGAL/cgal/blob/42cff9ee28b416a37a054a183c6758c07ab918be/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h#L232)and the next function as well, right?

Actually you are right I missed those two.

GilesBathgate:
@afabri I know you want to pause changes to consolidate. But the following code:

https://github.com/CGAL/cgal/blob/1c4b191c71b189c719003bdbe98a5954d1f2ed40/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h#L170-L181

Is it mathematically equivalent to:
```c++
template <typename R>
Sphere_point<R> intersection(const Sphere_circle<R>& c1,
                             const Sphere_circle<R>& c2)
/*{\Mfunc returns one of the two intersection points of
|c1| and |c2|. \precond |c1 != c2| as sets.}*/
{
  CGAL_assertion(!equal_as_sets(c1,c2));

  typedef typename R::Plane_3 Plane_3;
  return CGAL::ORIGIN + R().construct_cross_product_vector_3_object()(
        static_cast<Plane_3>(c1).orthogonal_vector(),
        static_cast<Plane_3>(c2).orthogonal_vector());
}
```

afabri:
That seems correct. With consolidation I mean the open PRs, so  let's do this one in a next round right after integration into master.

sloriot:

PR #6380, opened 2022-03-04
PMP hole filling: Add an option to only use Delaunay based versions

sloriot:
Add a way to avoid the cubic search space to be used, early exit without hole filled.

It is useful for very large holes


sloriot:
what about `do_not_use_cubic_algorithm` -> `do_not_use_cubic_search_space`?

afabri:
> what about `do_not_use_cubic_algorithm` -> `do_not_use_cubic_search_space`?

What about `use_only_delaunay_triangulation`

sloriot:
I think it is ambiguous with which version can be used, 2D, 3D, both.

sloriot:

PR #6383, opened 2022-03-04
Performance enhancements for SNC_point_locator ray shoot

GilesBathgate:
## Summary of Changes

In #5522 code was added to avoid a geometric test for a point which is a vertex of the facet. This was in the SNC_point_locator `locate` function. It seems the same approach is relevant for the `shoot` function, and also introduces a small performance enhancement.

## Release Management

* Affected package(s): Nef_3
* Issue(s) solved (if any): speed

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -216,6 +216,26 @@ public:
       delete candidate_provider;
   }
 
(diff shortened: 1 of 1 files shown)


GilesBathgate:
This sometimes causes the error:
```File: /usr/include/CGAL/Nef_3/SNC_const_decorator.h
Line: 276
Explanation: it is not possible to decide which one is a visible facet (if any)
```

PR #6388, opened 2022-03-07
Triangulation_2:  Add Example for Starring a Conflict Zone

afabri:
## Summary of Changes

Add an example that computes a conflict zone and then stars it.

Fix at the same time the documentation of  a function, as one parameter has a default.

## Release Management

### DIFF
--- a/Triangulation_2/examples/Triangulation_2/star_conflict_zone.cpp
+++ b/Triangulation_2/examples/Triangulation_2/star_conflict_zone.cpp
@@ -22,7 +22,7 @@ int main( )
   dt2.insert(Point_2(10,0));
   dt2.insert(Point_2(0,10));
 
-  std::array<Point_2,3> points = { Point_2(2,2), Point_2(1,0), Point_2(9,9) };
+  std::array<Point_2,3> points = { Point_2(2,2), Point_2(1,0), Point_2(2,2) };
 
   CGAL::spatial_sort(points.begin(), points.end());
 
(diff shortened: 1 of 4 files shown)


Triangulation_2/examples/Triangulation_2/star_conflict_zone.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+int main( )
+{
+  Dt2 dt2;
+
+  dt2.insert(Point_2(0,0));
+  dt2.insert(Point_2(10,0));
+  dt2.insert(Point_2(0,10));
+
+  std::array<Point_2,3> points = { Point_2(2,2), Point_2(1,0), Point_2(2,2) };
```
sloriot, line 25: why using a degenerate case in an example?

Triangulation_2/examples/Triangulation_2/star_conflict_zone.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  CGAL::spatial_sort(points.begin(), points.end());
+
+  Face_handle hint;
+
+  std::vector<Face_handle> faces;
+  std::vector<Edge> edges;
+
+  assert(dt2.dimension() == 2); // precondition of get_conflicts_and_bounda
```
sloriot, line 34:
```suggestion
  assert(dt2.dimension() == 2); // precondition of get_conflicts_and_boundary
```

sloriot:

PR #6392, opened 2022-03-08
better reference Kernel intersection functions

sloriot:
When searching "intersection" in the search bar one has a hard time to find the function in the 2D Linear Kernel as we have functions and documentation groups in Boolean Set Operations as well as in Polygon Mesh Processing.

This PR tries to improve the situation.


sloriot:

PR #6395, opened 2022-03-09
TDS: Initialize the size of Unique_hash_map

afabri:
## Summary of Changes

In functions like `copy_tds()`, and hence in the copy constructor of triangulation classes, as well as when writing to a stream we use `Unique_hash_map`.    When we know that all vertices will be inserted in a loop we better construct the map with a table as big as the number of vertices.

I stumbled in this while checking whether there is a drawback to start with table size 512 or 32 as suggested  by @GilesBathgate  in PR #6306.       I opted for an independent PR. 

We have to check more packages for the same improvement to make.

## Release Management

* Affected package(s):  TDS_2/3, Periodic_2/3

### DIFF
--- a/Combinatorial_map/include/CGAL/Combinatorial_map.h
+++ b/Combinatorial_map/include/CGAL/Combinatorial_map.h
@@ -3376,7 +3376,7 @@ namespace CGAL {
       CGAL_assertion( is_without_boundary(dimension) );
 
       CGAL::Unique_hash_map< Dart_handle, Dart_handle,
-        typename Self::Hash_function > dual;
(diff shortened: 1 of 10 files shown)


GilesBathgate:
Yes when it's possible to determine the size in the construction of the hash map, this is a better option. I could revise some of my changes where it's possible.

afabri:
@gdamiand I started making changes in your packages where the `Unique_hash_map` is used. When I know how many elements will be put into it, I construct right away a table of the appropriate size.  Maybe you can see if there are further occurences.

sloriot:

PR #6398, opened 2022-03-10
Replaced boost::tuple with std::tuple

GYuvanShankar:
## Summary of Changes

Replaced boost::tuple with std::tuple where there is no difference in use of the two.

## Release Management

* Affected package(s): Surface_mesh_simplification, AABB_tree


afabri (approved):

sloriot:

PR #6399, opened 2022-03-10
Kernel_23: Add Orientation_3::operator()(Origin,Point_3,Point_3,Point_3)

afabri:
## Summary of Changes

Adding the operator in `CGAL/Cartesian/function_objects.h` required some additions in converters, and in Origin itself. 
I did not add an overload for the homogeneous kernel and I did not add it for the global function. 

The added test case can also be used for benchmarking. For example with ` data/points/sphere_20k.xyz`. 
The main interest is for the exact constructions kernel.

This PR is triggered by a discussion in #6378, and @GilesBathgate might be interested to merge it. 

## Release Management

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -3033,7 +3033,7 @@ namespace CartesianKernelFunctors {
     }
 
     Vector_3
-    operator()( Origin o , const Point_3& q, const Point_3& r ) const
+    operator()( Origin, const Point_3& q, const Point_3& r ) const
     {
       // Cross product oq * or
       FT vx = q.y()*r.z() - r.y()*q.z();
@@ -4320,7 +4320,7 @@ namespace CartesianKernelFunctors {
     }
 
     result_type
-    operator()( Origin o, const Point_3& u,
+    operator()( Origin, const Point_3& u,
                 const Point_3& v, const Point_3& w) const
     {
       return orientationC3(u.x(), u.y(), u.z(),
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -3035,16 +3035,10 @@ namespace CartesianKernelFunctors {
     Vector_3
     operator()( Origin o , const Point_3& q, const Point_3& r ) const
     {
-      FT orx = r.x();
-      FT ory = r.y();
-      FT orz = r.z();
-      FT oqx = q.x();
-      FT oqy = q.y();
-      FT oqz = q.z();
       // Cross product oq * or
-      FT vx = oqy*orz - oqz*ory;
-      FT vy = oqz*orx - oqx*orz;
-      FT vz = oqx*ory - oqy*orx;
+      FT vx = q.y()*r.z() - r.y()*q.z();
+      FT vy = q.z()*r.x() - r.z()*q.x();
+      FT vz = q.x()*r.y() - r.x()*q.y();
       typename K::Construct_vector_3 construct_vector;
 
       return construct_vector(vx, vy, vz);
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -3031,6 +3031,24 @@ namespace CartesianKernelFunctors {
 
       return construct_vector(vx, vy, vz);
     }
+
+    Vector_3
+    operator()( Origin o , const Point_3& q, const Point_3& r ) const
+    {
+      FT orx = r.x();
+      FT ory = r.y();
+      FT orz = r.z();
+      FT oqx = q.x();
+      FT oqy = q.y();
+      FT oqz = q.z();
+      // Cross product oq * or
+      FT vx = oqy*orz - oqz*ory;
+      FT vy = oqz*orx - oqx*orz;
+      FT vz = oqx*ory - oqy*orx;
+      typename K::Construct_vector_3 construct_vector;
+
+      return construct_vector(vx, vy, vz);
+    }
   };
 
   template <typename K>
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -4308,6 +4308,15 @@ namespace CartesianKernelFunctors {
     }
 
     result_type
+    operator()( Origin o, const Point_3& u,
+                const Point_3& v, const Point_3& w) const
+    {
+      return orientationC3(u.x(), u.y(), u.z(),
+                           v.x(), v.y(), v.z(),
+                           w.x(), w.y(), w.z());
+    }
+
+    result_type
     operator()( const Tetrahedron_3& t) const
     {
       return t.rep().orientation();
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -799,7 +799,7 @@ struct Approx_converter
   template < typename T >
   decltype(auto)
   operator()(const T&t) const
-  { return t.approx(); }
+  { return approx(t); }
 
   const Null_vector&
   operator()(const Null_vector& n) const
@@ -824,7 +824,7 @@ struct Exact_converter
   template < typename T >
   decltype(auto)
   operator()(const T&t) const
-  { return t.exact(); }
+  { return exact(t); }
 
   const Null_vector&
   operator()(const Null_vector& n) const
--- a/Filtered_kernel/include/CGAL/Static_filtered_predicate.h
+++ b/Filtered_kernel/include/CGAL/Static_filtered_predicate.h
@@ -28,9 +28,7 @@ public:
   result_type operator()(const A1& a1) const
   {
     CGAL::Epic_converter<AK> convert;
-    typedef typename Kernel_traits<A1>::type EK;
-    typedef typename Type_mapper<A1,EK,Exact_predicates_inexact_constructions_kernel>::type T1;
-    std::pair<T1,bool> aa1 = convert(a1.approx());
+    auto aa1 = convert(approx(a1));
     if(! aa1.second){
       return fp(a1);
     }
@@ -42,9 +40,7 @@ public:
   result_type operator()(const A1& a1, const Null_vector& v) const
   {
     CGAL::Epic_converter<AK> convert;
-    typedef typename Kernel_traits<A1>::type EK;
-    typedef typename Type_mapper<A1,EK,Exact_predicates_inexact_constructions_kernel>::type T1;
-    std::pair<T1,bool> aa1 = convert(a1.approx());
+    auto aa1 = convert(approx(a1));
     if(! aa1.second){
       return fp(a1, v);
     }
@@ -56,14 +52,11 @@ public:
   result_type operator()(const A1& a1, const A2& a2) const
   {
     CGAL::Epic_converter<AK> convert;
-    typedef typename Kernel_traits<A1>::type EK;
-    typedef typename Type_mapper<A1,EK, Exact_predicates_inexact_constructions_kernel>::type T1;
-    std::pair<T1,bool> aa1 = convert(approx(a1));
+    auto aa1 = convert(approx(a1));
     if(! aa1.second){
       return fp(a1, a2);
     }
-    typedef typename Type_mapper<A2,EK, Exact_predicates_inexact_constructions_kernel>::type T2;
-    std::pair<T2,bool> aa2 = convert(approx(a2));
+    auto aa2 = convert(approx(a2));
     if(! aa2.second){
       return fp(a1, a2);
     }
@@ -76,9 +69,7 @@ public:
   result_type operator()(const Bbox_2& bb, const A2& a2) const
   {
     CGAL::Epic_converter<AK> convert;
-    typedef typename Kernel_traits<A2>::type EK;
-    typedef typename Type_mapper<A2,EK, Exact_predicates_inexact_constructions_kernel>::type T2;
-    std::pair<T2,bool> aa2 = convert(approx(a2));
+    auto aa2 = convert(approx(a2));
     if(! aa2.second){
       return fp(bb, a2);
     }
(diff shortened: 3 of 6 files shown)


Filtered_kernel/include/CGAL/Static_filtered_predicate.h
```diff
@@ -127,23 +126,29 @@ class Static_filtered_predicate {
   result_type operator()(const A1& a1, const A2& a2, const A3& a3, const A4& a4) const
```
GilesBathgate, line 126: Could you alternatively create a static filtered predicate overload that takes a `Origin`, there is an overload that takes `Bbox_3` for example:

https://github.com/CGAL/cgal/blob/a8e87526348d2d1faa2ea893747314853ee6868d/Filtered_kernel/include/CGAL/Static_filtered_predicate.h#L90

This would then save having to implement approx, and exact on class `Origin`. Maybe your intention was to keep the code more generic.

Kernel_23/include/CGAL/Origin.h
```diff
@@ -23,7 +23,18 @@
...
+public:
+  const Origin& approx() const
+  {
+    return *this;
+  }
+
+  const Origin& exact() const
+  {
+    return *this;
+  }
```
mglisse, line 36: Lazy.h already contains
```c++
// For tag classes: Return_base_tag, Homogeneous_tag, Null_vector, Origin
template<class T>inline std::enable_if_t<std::is_empty<T>::value, T> exact(T){return {};}
```
Would it make sense, instead of adding this function, to change some `a1.approx()` to `CGAL::approx(a1)` where you were getting errors?

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
```diff
@@ -127,23 +126,29 @@ class Static_filtered_predicate {
...
   {
     CGAL::Epic_converter<AK> convert;
-    typedef typename Kernel_traits<A1>::type EK;
-    typedef typename Type_mapper<A1,EK,Exact_predicates_inexact_constructions_kernel>::type T1;
+    typedef typename Kernel_traits<A1>::type EK1;
+    typedef typename Type_mapper<A1,EK1,Exact_predicates_inexact_constructions_kernel>::type T1;
     std::pair<T1,bool> aa1 = convert(a1.approx());
```
mglisse, line 131: Could we just use `auto` here and skip the complications?

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
GilesBathgate: @mglisse 's suggestion seems better.

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
afabri: The problem is that we only look into this code from time to time and have to understand what happens, and said that I do not find `auto` a good idea.  On the other hand we could try to come  up with a better name for `T1`.

GilesBathgate: can you even use `std::tie`?

GilesBathgate (approved):

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
mglisse:
```c++
// std::pair<Epick::XXX, bool>
auto aa1 = convert(CGAL::approx(a1));
```
seems clear enough to me, but as you say, you are the one who will have to reread this, so your choice.
(I didn't remember this use of `std::pair` in place of `boost::optional`)

GilesBathgate:
```c++
using Pair1 = Some_mapper<A1>::type;
Pair1 aa1 = convert(CGAL::approx(a1));
```

GilesBathgate: possible implementation:

```c++
template <typename A>
struct Some_mapper
{
    typedef typename Kernel_traits<A>::type EK;
    typedef typename Type_mapper<A,EK,Exact_predicates_inexact_constructions_kernel>::type T;
    typedef typename T type;
}
```

afabri: When running the Kernel testsuite I saw that we have a problem with `EK4`  when the fourth argument is a number type as the kernel traits of a number type is the  ` Dummy_kernel`.  That is the same for ` Origin` and its ` EK1` but without consequence.   So we must find the real kernel of one of the argument types ` Ai`.

lrineau: > When running the Kernel testsuite I saw that we have a problem with `EK4` when the fourth argument is a number type as the kernel traits of a number type is the `Dummy_kernel`. That is the same for `Origin` and its `EK1` but without consequence. So we must find the real kernel of one of the argument types `Ai`.

I agree with Marc:
```c++
auto aa1 = convert(CGAL::approx(a1));
```
That has two advantages:
  - use `CGAL::approx` instead of a member function `a1.approx()`,
  - no need to do complex meta-programming to implement what you suggest.

afabri: I came up with [this ](https://gist.github.com/afabri/9aa601c5a469e362caa1e65f9c851deb) to find a kernel different from ` Dummy_kernel`.    I guess that can be simplified to make ` Kernel_traits`  itself recursive.

afabri: But I agree with you that we only need the exact kernel type for a return type of something we immediately pass as an argument to another function.   More out of curiosity:  Can one write  `std::pair<auto,bool> a  = convert(,,);`?

mglisse: In more recent C++ you can write `std::pair a = convert(...);` (no explicit template parameters), but I don't think you can partially specify them the way you did. I could have missed some extensions though.

lrineau: > In more recent C++ you can write `std::pair a = convert(...);` (no explicit template parameters), but I don't think you can partially specify them the way you did. I could have missed some extensions though.

No, you are right (`auto` cannot be used that way).

As for using class template argument deduction (`std::pair a = convert(...);`), we cannot use C++17 in CGAL, for now.

afabri: I could use   `auto a; bool b;`  and `std::tie(a,b)`  but just `auto` is fine as well.

lrineau: > I could use `auto a; bool b;` and `std::tie(a,b)` but just `auto` is fine as well.

No you cannot. `auto a;` will not compile because there is no initialization, and the compiler cannot infer the type of `a`.

In C++17, there would be the [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) (a powerful replacement of `tie` in the core language of C++17):
```c++
auto [aa1, conversion_ok] = convert(CGAL::approx(a1));
```
but we cannot use C++17, for now.

afabri: The future is always brighter!!!

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
```diff
@@ -3031,6 +3031,24 @@ namespace CartesianKernelFunctors {
 
       return construct_vector(vx, vy, vz);
     }
+
+    Vector_3
+    operator()( Origin o , const Point_3& q, const Point_3& r ) const
```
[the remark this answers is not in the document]
mglisse: It is better to pass it by value, as Andreas did.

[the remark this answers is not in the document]
lrineau: > Could you use `const Origin& o` here instead of `Origin o` or doesn't it matter?

That would be a pessimization. `Origin` is an empty class. Passing it by reference to const forces the compiler to pass the address of `CGAL::ORIGIN` on the stack, whereas passing it by value can be optimized out.

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
```diff
@@ -57,65 +52,31 @@ class Static_filtered_predicate {
...
     if(! aa2.second){
       return fp(a1, a2);
     }
     return epicp(aa1.first, aa2.first);
   }
 
-    // We need these two specializations as in general we determine
-    // the kernel for the template argument A1, and this does not work for Bbox_2 and Bbox_3
-  template <typename A2>
-  result_type operator()(const Bbox_2& bb, const A2& a2) const
```
mglisse, line 77: You can probably also drop the overload with Null_vector, though it doesn't hurt to keep it.

Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -799,7 +799,7 @@ struct Approx_converter
   template < typename T >
   decltype(auto)
   operator()(const T&t) const
-  { return t.approx(); }
+  { return approx(t); }
```
mglisse, line 802: Nice, that probably makes all the overloads in this class useless :smiley:

Filtered_kernel/include/CGAL/Static_filtered_predicate.h
mglisse: > `auto [aa1, conversion_ok] = convert(CGAL::approx(a1));`

Indeed Andreas' message reminded me of exactly this. But in C++17, I would expect we stop using `pair<T,bool>` and use `optional<T>` instead as return type of `convert`.

Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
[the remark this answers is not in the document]
GilesBathgate: @lrineau I think I already knew this. Typically, pass by reference when the object you are passing is much larger than the size of a pointer. However, I've seen it in other codebases (such as Qt), where it's more a case of consistency than performance. I prefer performance ;)

Kernel_23/test/Kernel_23/origin_3.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for(int i = 0; i < N; ++i){
+      if(orientation(CGAL::ORIGIN, points[i], points[i+1], points[i+2]) == CGAL::POSITIVE){
+        --positive;
+     }
+    }
+  }
+  t.stop();
+
+
+  if(positive =! 0){
```
GilesBathgate, line 60: this is a typo
should be `if(positive != 0)`

Kernel_23/test/Kernel_23/origin_3.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for (int k = 0; k < 100; ++k)
+      for(int i = 0; i < N; ++i){
+        Vector_3 v = construct_orthogonal_vector(CGAL::ORIGIN, points[i], points[i+1]);
+        sumx2 += CGAL::to_double(v.approx().x());
+      }
+  }
+
+  t.stop();
+
+  if(sumx1 == sumx2){
```
GilesBathgate, line 100: it says Not the same results if they are equal? should it be:

`if(sumx1 != sumx2)`

GilesBathgate (commented):
Maybe in the test you want to remove the dependency on an input file using `CGAL::Random_points_on_sphere_3<Point_3> rnd;` like this:

```c++
//#define CGAL_PROFILE
#include <CGAL/Timer.h>
#include <vector>
#include <iostream>
typedef CGAL::Exact_predicates_exact_constructions_kernel K;
typedef K::Point_3 Point_3;
typedef K::Vector_3 Vector_3;
typedef CGAL::Timer Timer;

int main(int argc, char* argv[] )
{
  CGAL::Random_points_on_sphere_3<Point_3> rnd;

  std::vector<Point_3> points;
  Point_3 p;

  for(int i = 0; i < 100'000; ++i)
    points.push_back(*rnd);

  std::cout << "Orientation_3" << std::endl;
  Timer t;
  //...
  ```
  
  I see the following results when using the above:
  
  ```
Orientation_3
overload with 4 points
5.34537 sec.
overload with origin and 3 points
3.57668 sec.
Construct_orthogonal_vector_3
overload with 3 points
4.23489 sec.
overload with origin and 2 points
0.399501 sec.

```

Kernel_23/test/Kernel_23/origin_3.cpp
afabri: Your "should it be" states what I have written.  And I expect it to be zero as I count down.

GilesBathgate: @afabri you have written `positive =   !0` with different whitespace.

afabri: =!    just see it. thanks.

sloriot:

lrineau:
The PR is tested, but this PR is incomplete. It missed the equivalent overload for the global function, and the equivalent overload for the homogeneous kernel.

As `CGAL::Origin` implicitly converts to `CGAL::Point_3`, I wonder if the lack of documentation is an issue. That might be considered as an implementation detail for `Orientation_3`.

GilesBathgate:
> As `CGAL::Origin` implicitly converts to `CGAL::Point_3`

I am not sure this is true, I found that `CGAL::Origin` could not be substituted for `Point_3(CGAL::Origin)` in those functors until after I had merged @afabri 's changes. (although I am unsure why looking at the code for Point_3)

lrineau:
> > As `CGAL::Origin` implicitly converts to `CGAL::Point_3`
> 
> I am not sure this is true, I found that `CGAL::Origin` could not be substituted for `Point_3(CGAL::Origin)` in those functors until after I had merged @afabri 's changes. (although I am unsure why looking at the code for Point_3)

That is partially true. The following code compiles in `master` (without this PR):
```c++

using Point = K::Point_2;

int main() {
  [[maybe_unused]] auto o = K().orientation_2_object()(CGAL::ORIGIN, Point(), Point());
  //[[maybe_unused]] auto o2 = CGAL::orientation(CGAL::ORIGIN, Point(), Point());
}
```
but if one uncomments the line with `CGAL::orientation(CGAL::ORIGIN, Point(), Point());`, that does not compile.

I agree that we have to be cautious. A `CGAL::Point_3` is implicitly constructible from `CGAL::Origin`, but that does not mean that it always works well with function templates overloads.

afabri:
I did not want to engage that global functions work with `ORIGIN` passed as argument.  I also do not see the need to add it systematically to Homogeneous.

GilesBathgate:
>> It missed the equivalent overload for the global function, and the equivalent overload for the homogeneous kernel.
> I did not want to engage that global functions work with ORIGIN passed as argument. I also do not see the need to add it systematically to Homogeneous.

It makes sense to me, if its proposed to be internal undocumented optimisation, then no need to expose as a global function, or implement for Homogeneous. End users can use  `CGAL::orientation(Vector(),Vector(),Vector());` which you mentioned before @lrineau

PR #6402, opened 2022-03-11
Mesh_3 - add tunable angle to `Angle_tester` for `polylines_to_protect()`

janetournois:
## Summary of Changes

`polylines_to_protect(features, existing_features_begin, existing_features_end)` was able to detect acute angles in input polylines. It can now be tuned to angles smaller than a given bound.

When the parameter is left to its default value, the code is unchanged and checks for acute angles as before.

This function is undocumented for now.

## Release Management

* Affected package(s): Mesh_3

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
+++ b/Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
@@ -301,6 +301,14 @@ struct Polyline_visitor
 template <typename Kernel>
 struct Angle_tester
 {
+  const double m_angle_rad;//in radian to avoid extra computations
+  const bool m_test_angle;
+
+  Angle_tester(const double angle_deg)//angle given in degrees for readability
+    : m_angle_rad(angle_deg * CGAL_PI /180.)//stored in radian
+    , m_test_angle(angle_deg != 90.)//under 90 we check for acute, with exact computation
+  {}
+
   template <typename vertex_descriptor, typename Graph>
   bool operator()(vertex_descriptor& v, const Graph& g) const
   {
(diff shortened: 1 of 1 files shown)


sloriot:

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -301,6 +301,19 @@ struct Polyline_visitor
...
+
+  Angle_tester()
+    : m_angle_rad(0.5 * CGAL_PI)//stored in radian
+    , m_test_angle(false)//under 90 we check for acute, with exact computation
+  {}
+
+  Angle_tester(const double angle_deg)//angle given in degrees for readability
+    : m_angle_rad(angle_deg * CGAL_PI /180.)//stored in radian
+    , m_test_angle(angle_deg != 90.)//under 90 we check for acute, with exact computation
+  {}
```
lrineau, line 315: We should document, and ensure, that the angle must be bigger than 90°: if the angle between [p, p1] and [p p2] is smaller than 90°, then Mesh_3 requires that the vertex is a terminal vertex.

Then, I think than, instead of storing the actual angle, we should store the cosine of the angle, to avoid a call to `std::acos` for each call to `operator()`.

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -320,15 +333,21 @@ struct Angle_tester
...
-        //                          / CGAL::sqrt(e2*e2))
-        //                * 180 / CGAL_PI ) << std::endl;
-        return true;
+      if (!m_test_angle)
+      {
+        if (CGAL::angle(p1, p, p2) == CGAL::ACUTE)
+          return true;
+      }
+      else
+      {
```
lrineau, line 342: I suggest the following, to ensure that an acute angle will always `return true`:
```suggestion
      if (CGAL::angle(p1, p, p2) == CGAL::ACUTE)
        return true;
      if (m_test_angle)
      {
```

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
lrineau: Actually, as we know that the angle is > 90°, then the sign of the cosine is negative, so let's store the squared cosine directly:
```suggestion
  Angle_tester(const double angle_deg)//angle given in degrees for readability
    : m_angle_sq_cosine(CGAL::square(std::cos((std::min)(90, angle_deg) * CGAL_PI /180.))) // squared cosine of `std:min(90, angle_deg)`
    , m_test_angle(angle_deg != 90.)//under 90 we check for acute, with exact computation
  {}
```

Mesh_3/include/CGAL/Mesh_3/polylines_to_protect.h
```diff
@@ -320,15 +333,21 @@ struct Angle_tester
...
+        const typename Kernel::Vector_3 e1 = p1 - p;
+        const typename Kernel::Vector_3 e2 = p2 - p;
+        const double a_rad = std::acos(e1 * e2
+                                      / CGAL::sqrt(e1*e1)
+                                      / CGAL::sqrt(e2*e2));
+        // std::cerr << "At point " << p << ": the angle is " << a << std::endl;
+        if(a_rad < m_angle_rad)
+          return true;
```
lrineau, line 350: Here I suggest to use a comparison with the cosine of the angle. Something like:
```c++
  const auto scalar_product = e1 * e2;
  if(CGAL::is_positive(scalar_product)) return true;
  const auto sq_scalar_product = CGAL::square(scalar_product);
  if(sq_scalar_product <= m_angle_sq_cosine * (e1*e1) * (e2*e2)) return true;
```

lrineau (changes requested):

PR #6407, opened 2022-03-13
Remove redundant HTML type of anchor

albert-github:
Remove redundant HTML type of anchor, the anchor is not used and night give problems in the future (especially with https://github.com/doxygen/doxygen/pull/9204)


sloriot:

PR #6411, opened 2022-03-14
Point_set_processing_3: Missing end command for cgalParam

albert-github:
A `\cgalParamEnd` was missing, resulting in an incorrect sequence of HTML tags (in the output just a small extra white space was shown)

**Original**
image

**New**
image


sloriot:

PR #6412, opened 2022-03-14
Triangulation_2: Add CGAL::mark_domain_in_triangulation

### LINKED ISSUE #6393: Including `mark_domains` in the library
Hello, 

  The [mark_domains](https://doc.cgal.org/latest/Triangulation_2/Triangulation_2_2polygon_triangulation_8cpp-example.html) functions are very important for the constrained 2D Delaunay triangulation. Is there any reason for not including them in the library?

face

afabri:
## Summary of Changes
Make the function which is currently just in an example official.   In order to make it generic add a property map for the "in domain". 

Todo: 
- [x] Document it.    See    [Reference Manual](https://cgal.github.io/6412/v4/Triangulation_2/group__PkgTriangulation2Ref.html#gabe989734ed8d720d4385900d50b3c9c4) 
- [x] Add an example in the `Mesh_2` package.  See   [User Manual](https://cgal.github.io/6412/v4/Mesh_2/index.html#title14)
- [x] Add overloads for `write_VTU()`.   See [Reference Manual](https://cgal.github.io/6412/v1/Mesh_2/group__PkgMesh2IO.html#ga400b8da0d91f032e90c9b709d4996f72)
- [x] Select the property map automatically for `Delaunay_mesh_face_base_2`. 

## Release Management

* Affected package(s): Triangulation_2, Mesh_2
* Issue(s) solved (if any): fix #6393
* Feature/Small Feature (if any):  [link](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/mark_domain_in_triangulation_2)


Triangulation_2/examples/Triangulation_2/polygon_triangulation.cpp
```diff
@@ -2,22 +2,16 @@
 #include <CGAL/Constrained_Delaunay_triangulation_2.h>
 #include <CGAL/Triangulation_face_base_with_info_2.h>
 #include <CGAL/Polygon_2.h>
-#include <iostream>
-
-struct FaceInfo2
-{
-  FaceInfo2(){}
-  int nesting_level;
+#include <CGAL/mark_domains.h>
```
sloriot, line 5: The name of the header seems a bit generic

Triangulation_2/include/CGAL/mark_domains.h
```diff
@@ -0,0 +1,96 @@
...
+//to constrained edges bounding the former set and increase the nesting level by 1.
+//Facets in the domain are those with an odd nesting level.
+template <typename CDT, typename InDomainPmap>
+void
+mark_domains(CDT& cdt, InDomainPmap ipm)
+{
+  typedef typename CDT::Face_handle Face_handle;
+  typedef typename CDT::Edge Edge;
+
+  typename Unique_hash_map<Face_handle,int> nesting_level(-1, cdt.number_of_faces());
```
sloriot, line 76:
```suggestion
  Unique_hash_map<Face_handle,int> nesting_level(-1, cdt.number_of_faces());
```

Triangulation_2/examples/Triangulation_2/polygon_triangulation.cpp
afabri: Another possibility would be to add it as member function to the `Constrained_triangulation_2` class.

Triangulation_2/include/CGAL/mark_domain_in_triangulation.h
```diff
@@ -0,0 +1,112 @@
...
+#include <boost/property_map/property_map.hpp>
+
+#include <CGAL/Triangulation_2/internal/In_domain.h>
+#include <CGAL/Unique_hash_map.h>
+
+namespace CGAL {
+
+namespace internal {
+
+template <typename CDT, typename InDomainPmap>
```
GilesBathgate, line 29: I think the algorithm applies to any constrained triangulation, not a Delunlay one so should this be CT, instead of CDT?

Triangulation_2/examples/Triangulation_2/polygon_triangulation.cpp
GilesBathgate: I prefer @afabri 's suggestion but alternatively, it could be `CGAL/constrained_triangulation_mark_domains.h`

afabri: The name needs probably some discussion. Should there be a `_2` somewhere?   Is it domain or domains, the former having potentially several connected components?  etc.

GilesBathgate: What actually is a domain geometrically speaking? As far as I understood this is a way of finding holes, but then there could be an island inside the hole. There must be a geometric term for islands within islands. (At least I would have thought so with geo meaning earth in latin :laughing: )

afabri: For the Delaunay mesh generator the domain is the area where the refinement shall happen.  It can be controlled by seed points, which can be placed in holes as well as in non-holes.   The function I wrote considers the unbounded area as outside and then alternates between inside and outside the domain.  That's why we speak about nesting level and parity.

afabri: See  [here](https://github.com/CGAL/cgal/pull/6412/files#diff-8a166ef6d855401ed695deab613fefe7164c2832a17f3fa5e104f0bd42503719R8)

GilesBathgate: It seems like `domains`, `bounded domains`, `regions`, `bounded regions` or even `islands` (used in https://doc.cgal.org/latest/Polyline_simplification_2/index.html#Chapter_2D_Polyline_simplification) would be acceptable terms. I am not sure I understand how a non-plural term such as `domain` would apply when there can be more than one of them.

GilesBathgate: Oh, would you say the `island` inside a lake inside an `island`, all belong to the same domain? In that case, I understand why you would use a non-plural.

GilesBathgate: In geography, the lithosphere is a domain concerning land. If that makes for a good analogy?

afabri: Reading [this ](https://en.wikipedia.org/wiki/Domain_(mathematical_analysis)) wikipedia page we probably should use plural.

afabri: But I will stick to singular, because in the [User Manual](https://doc.cgal.org/latest/Mesh_2/index.html#title5) we define domain as something that can have several connected components.

GilesBathgate: Maybe the best way to explain it in the documentation would be an image, for example, the orange coloured parts belong to the same domain:
image

GilesBathgate: Oh like this image:
image

GilesBathgate: It seems a bit odd to say that the green triangle is "connected" to the other green parts, but ok.

Installation/CHANGES.md
```diff
@@ -7,6 +7,10 @@ Release History
 
 Release date: June 2022
 
+### [2D Triangulations](https://doc.cgal.org/5.5/Manual/packages.html#PkgTriangulation2)
+
+-   Added the function `mark_domians_in_triangulation()`
```
lrineau, line 12:
```suggestion
-   Added the function `mark_domain_in_triangulation()`
```

afabri:
@gdamiand can you please have a look at the modified draw function in the [manual](https://cgal.github.io/6412/v0/Triangulation_2/group__PkgDrawTriangulation2.html).

gdamiand:
Suggestion: replace "opens a new window and draws a constrained triangulation with faces marked as being part of a domain." by "opens a new window and draws a constrained triangulation (only faces marked as being part of a domain being drawn)."

To precise what is done for the faces in the domain (which I found unclear in the initial formulation).

afabri:
It draws all triangles but draws the triangles in the domain in a fill color.
image

afabri:
> Ok; so just says it in the doc (like it is done for the other draw function).

opens a new window and draws a constrained triangulation. If the triangulation
has constraints they are drawn.  The faces inside and outside of the
domain, based on the property map, are drawn in different colors.

lrineau:
> Not needed as draw_triangulation_2.h includes the other one.

You are right. I reverted that change.

There was a problem with the two examples using `CGAL::draw`, because:
  - the detection of `face->is_in_domain()` was wrong,
  - `CGAL::refine_Delaunay_mesh` is documented to override the existing `is_in_domain()` marks, using seeds.

I had to patch the pull-request that way: https://github.com/CGAL/cgal/pull/6412/files/4975b1a0f70daf1d00bfc315574243d29591f35a..7a869e898118f2614362b362cc8171e97c7c08d6.

Now, the main issue is that the example `Mesh_2/examples/Mesh_2/mesh_marked_domain.cpp`, added to the user manual of Mesh_2, uses an undocumented feature of `CGAL::refine_Delaunay_mesh`, to disable its default behavior with marks and seeds.

lrineau:
@afabri @sloriot Before I document it, can you please review this proposal for a new API:
```c++
CGAL::refine_Delaunay_mesh_2(cdt,
                             CGAL::Use_existing_in_domain_marks(),
                             Criteria());
```
from commit [`44317ac` (#6412)](https://github.com/CGAL/cgal/pull/6412/commits/44317acb53a5ad199c47dcfb5a8a51fde8814f6a)?

I want to use a dedicated tag class to avoid using a `bool` parameter (that is a known anti-pattern).

lrineau:
> > I want to use a dedicated tag class to avoid using a `bool` parameter (that is a known anti-pattern).
> 
> Can you not have a different function name instead of a tag? `refine_Delaunay_mesh_using_existing_in_domain_marks()`

I thought about that. And I was not able to find a "good" name. @sloriot @afabri would you prefer the following?
```c++
CGAL::refine_Delaunay_mesh_using_existing_in_domain_marks(cdt, Criteria());
```

GilesBathgate:
> I thought about that. And I was not able to find a "good" name. @sloriot @afabri would you prefer the following?
> 
> ```c++
> CGAL::refine_Delaunay_mesh_using_existing_in_domain_marks(cdt, Criteria());
> ```

Well a better name I agree, but otherwise the callee would have to write:

```c++
 CGAL::refine_Delaunay_mesh_2(cdt, Use_existing_in_domain_marks(), Criteria());
 ```
 
 which isn't much different ;)

**Comment by sloriot** [comment]
> @afabri @sloriot Before I document it, can you please review this proposal for a new API:
> 
> ```c++
> CGAL::refine_Delaunay_mesh_2(cdt,
>                              CGAL::Use_existing_in_domain_marks(),
>                              Criteria());
> ```
> 
> from commit [`44317ac` (#6412)](https://github.com/CGAL/cgal/pull/6412/commits/44317acb53a5ad199c47dcfb5a8a51fde8814f6a)?
> 
> I want to use a dedicated tag class to avoid using a `bool` parameter (that is a known anti-pattern).

If you ask me, I would go for using function named parameters. You would need to use the macros so as to differentiate a function named from a criteria to not break backward compatibility.

**Comment by lrineau** [comment]
> > @afabri @sloriot Before I document it, can you please review this proposal for a new API:
> > ```c++
> > CGAL::refine_Delaunay_mesh_2(cdt,
> >                              CGAL::Use_existing_in_domain_marks(),
> >                              Criteria());
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > from commit [`44317ac` (#6412)](https://github.com/CGAL/cgal/pull/6412/commits/44317acb53a5ad199c47dcfb5a8a51fde8814f6a)?
> > I want to use a dedicated tag class to avoid using a `bool` parameter (that is a known anti-pattern).
> 
> If you ask me, I would go for using function named parameters. You would need to use the macros so as to differentiate a function named from a criteria to not break backward compatibility (`CGAL_NP_TEMPLATE_PARAMETERS`/`CGAL_NP_CLASS`). You'd be also able to handle that `mark = true` at the same time.

I thought about that possibility too, but:
  - I thought pull the named parameters machinery for just *one* Boolean parameter was a bit of overkill,
  - and I am not well practiced in the details of named parameters. In particular, I have no clear idea of what you are talking about, and why it might be required.

I suggest one you two (@sloriot or @afabri) tries a possible implementation for that.

**Comment by sloriot** [comment]
For booleans it is easy and now it is part of `STL_extension` :)

**Comment by sloriot** [comment]
FYI, I'm doing a prototype implementation.

**Comment by lrineau** [comment]
> FYI, I'm doing a prototype implementation.

I like the proposal from commit https://github.com/CGAL/cgal/pull/6412/commits/38852cf8f8514951c57dbe40bf52eb4cd7a43b6c. The API seems really nice, that way.

**Comment by afabri** [comment]
As we are on it, it would be glad if @janetournois could fix or tell something about the commented `#define`  in the Lloyd optimization example.

**Comment by janetournois** [comment]
* `#define CGAL_MESH_2_OPTIMIZERS_DEBUG` can simply be removed from the example, it's a debugging helper
* `#define CGAL_MESH_2_SIZING_FIELD_USE_BARYCENTRIC_COORDINATES` enables the use of barycentric coordinates to compute the sizing function at `p` inside a finite face. It is not used by default because it is expensive. It can be useful to be more precise computing the sizing function. The sizing function is there to keep the same density "locally everywhere" after optimization

**Comment by afabri** [comment]
> * `#define CGAL_MESH_2_OPTIMIZERS_DEBUG` can simply be removed from the example, it's a debugging helper
> 
>     * `#define CGAL_MESH_2_SIZING_FIELD_USE_BARYCENTRIC_COORDINATES` enables the use of barycentric coordinates to compute the sizing function at `p` inside a finite face. It is not used by default because it is expensive. It can be useful to be more precise computing the sizing function. The sizing function is there to keep the same density "locally everywhere" after optimization

Should this become a names parameter, or does it also introduce a dependency on other packages?

**Comment by janetournois** [comment]
It could become a NP, but the Lloyd function already uses the old type of named parameters, see e.g.
`CGAL::lloyd_optimize_mesh_2(cdt, CGAL::parameters::max_iteration_number = 10);`

**Comment by afabri** [comment]
I just discussed this with Jane and then with Laurent.    It all started  with the usage of an undocumented parameter. Sebastien proposed named parameters which are elegant. Then I saw that we already have named parameters but for `lloyd_optimize()`, BUT in the boost named parameters way.   We do not mix in the same package.   I then thought that we can have at the same time the two named parameter APIs  and just deprecate the  boost one, but Laurent pointed out that we will have a problem when using the same `CGAL::parameters::XX"  for the two.

**Comment by lrineau** [comment]
> I saw that we already have named parameters but for `lloyd_optimize()`, BUT in the boost named parameters way. We do not mix in the same package. I then thought that we can have at the same time the two named parameter APIs and just deprecate the boost one, but Laurent pointed out that we will have a problem when using the same `CGAL::parameters::XX" for the two.

I had a quick look at that issue.

# Compatibility of CGAL parameters and Boost Parameter Library

## Part 1: compatibility of `CGAL::parameters::XX`

### Boost Parameter
With parameters from the Boost Parameter Library, a parameter is:
  - a tag class, let's call it `Tag`,
  - and a keyword object of type `boost::parameter::keyword<Tag>`.

So, for example, the name `CGAL::parameters::max_iteration_number` is a reference to const of type:
```c++
const boost::parameter::keyword<CGAL::parameters::tag::max_iteration_number_>&
```

### CGAL named parameters, copied and adapted from BGL
With the CGAL named parameters inherited from the BGL, a parameter is:
  - a tag class `Tag`,
  - a function template with one argument of type `K`, returning a `CGAL::parameters::Named_function_parameters<K, Tag>`,
  - and a member function of that template class `CGAL::parameters::Named_function_parameters`, that allows the chaining of parameters using a `.`.

For example, `CGAL::parameters::do_not_use_cubic_algorithm<bool>`, is a function of type:
```c++
CGAL::Named_function_parameters<bool, Tag, CGAL::internal_np::No_property>(const bool&)
```
where `Tag` is `CGAL::internal_np::do_not_use_cubic_algorithm_t`.

### How to make those two compatible?

Maybe we could derive (or copy-paste) our own template class `CGAL::parameters::keyword<tag>` from `boost::parameter::keyword<Tag>`, to add a `template <typename T> operator()(const T&)` to it, returning a `CGAL::parameters::Named_function_parameters<K, Tag>`. That would allow to use the named parameters defined by the Boost Parameter Library in the framework of the CGAL parameters. So, example, `CGAL::parameters::do_not_use_cubic_algorithm` would no longer be a function template, but a function object with a template function member `operator()`.

## Part 2: compatibility of parameter packs

I have not really looked at that part.

In CGAL parameters, the parameter pack is the class template `CGAL::Named_function_parameters<Type, Tag, Base` itself, where `Base` is the tail of a recursive tuple-like implementation.

In Boost Parameter, [`ArgumentPack`] is a concept, and its concrete implementation is very complex. Its particularity is its `operator,` that allows to gather several arguments in a pack.

[`ArgumentPack`]: https://www.boost.org/doc/libs/1_78_0/libs/parameter/doc/html/reference.html#argumentpack

lrineau:
> > I saw that we already have named parameters but for `lloyd_optimize()`, BUT in the boost named parameters way. We do not mix in the same package. I then thought that we can have at the same time the two named parameter APIs and just deprecate the boost one, but Laurent pointed out that we will have a problem when using the same `CGAL::parameters::XX" for the two.
> 
> I had a quick look at that issue.
> # Compatibility of CGAL parameters and Boost Parameter Library

I have transformed my comment into an issue (#6468), so it can be easily found. Please answer to it, if you want to discuss about the general idea to make Compatibility of CGAL parameters and Boost Parameter Library more compatible.

Continue here if you want to keep on discussing on the consequences of adding [CGAL::mark_domain_in_triangulation](https://cgal.github.io/6412/v1/Triangulation_2/group__PkgTriangulation2Ref.html#gabe989734ed8d720d4385900d50b3c9c4).

sloriot:

sloriot:
Note that the API issue with lloyd function is addressed in https://github.com/CGAL/cgal/pull/6642

sloriot:
@lrineau do you now approve the small feature?

lrineau (approved):
I will pre-approve the small feature.

lrineau:
This pull-request is a lot more than the addition of `CGAL::mark_domain_in_triangulation_2`: there is also the addition of named parameters to Mesh_2, and the deprecated of the older functions. It should be properly documented in `CHANGES.md`.

lrineau:
> I will pre-approve the small feature.

I have reviewed all the modifications, and particularly the modification of the documentation:
  - [`refine_Delaunay_mesh_2` with named parameters, and deprecation the other two overloads](https://cgal.github.io/6412/v4/Mesh_2/group__PkgMesh2Functions.html#gad0e4753d46c0203cb3d45d102840e483),
  - [two overloads of `mark_domain_in_triangulation`](https://cgal.github.io/6412/v4/Triangulation_2/group__PkgTriangulation2Ref.html#gabe989734ed8d720d4385900d50b3c9c4). I note that those two overloads are in no documentation group. Maybe they should go to "Miscellaneous".
  - [a new overload of `write_VTU`](https://cgal.github.io/6412/v4/Mesh_2/group__PkgMesh2IO.html#ga938b5d46b0dcff1b56f7342926e9f564),

and the addition of two examples.

This should be documented in the `CHANGES.md`.

sloriot:

@afabri : You have some todos that prevent Laurent from pre-approving the small-feature...

Installation/CHANGES.md
```diff
@@ -1,6 +1,14 @@
 Release History
 ===============
 
+### [2D Triangulations](https://doc.cgal.org/5.6/Manual/packages.html#PkgTriangulation2)
+
+-   Add function `mark_domina_in_triangulation()` to mark faces connected with non constrained edges as inside of the domain based on the nesting level.
```
sloriot, line 6:
```suggestion
-   Added function `mark_domains_in_triangulation()` to mark faces connected with non constrained edges as inside of the domain based on the nesting level.
```

Installation/CHANGES.md
```diff
@@ -1,6 +1,14 @@
...
+
+### [2D Conforming Triangulations and Meshes](https://doc.cgal.org/5.6/Manual/packages.html#PkgMesh22)
+
+-   Deprecate two overloads of Function `refine_Delaunay_mesh()` and replace them with versions with named parameteters.
```
sloriot, line 10:
```suggestion
-   Deprecated two overloads of Function `refine_Delaunay_mesh()` and replaced them with versions using function named parameters.
```

lrineau (approved):
I have reviewed all the modifications, and particularly the modification of the documentation:
  - [`refine_Delaunay_mesh_2` with named parameters, and deprecation the other two overloads](https://cgal.github.io/6412/v5/Mesh_2/group__PkgMesh2Functions.html#gad0e4753d46c0203cb3d45d102840e483),
  - [two overloads of `mark_domain_in_triangulation`](https://cgal.github.io/6412/v5/Triangulation_2/group__PkgTriangulation2Miscellaneous.html). They are now in the group "Miscellaneous".
  - [a new overload of `write_VTU`](https://cgal.github.io/6412/v5/Mesh_2/group__PkgMesh2IO.html#ga938b5d46b0dcff1b56f7342926e9f564),

and the addition of two examples.

**I pre-approve**.

sloriot:

PR #6413, opened 2022-03-15
Make write_STL (and write_polygon_mesh) compile with sqrt-less kernels

MaelRL:
## Summary of Changes

`unit_normal()` is called in STL writing of soup and mesh. Unfortunately, it requires a square root, and will thus not compile with EPECK. This gets especially unpleasant when calling non-STL writing that still requires STL writing compilation e.g., `IO::write_polygon_mesh("out.off", mesh)`.

This PR proposes to use an approximate square root call instead.

## Release Management

* Affected package(s): `Stream_support`, `BGL`
* Issue(s) solved (if any): -

### DIFF
--- a/BGL/include/CGAL/boost/graph/IO/STL.h
+++ b/BGL/include/CGAL/boost/graph/IO/STL.h
@@ -271,9 +271,9 @@ bool write_STL(std::ostream& os,
     for(const face_descriptor f : faces(g))
     {
       const halfedge_descriptor h = halfedge(f, g);
-      const Point_ref p = get(vpm, target(h, g));
-      const Point_ref q = get(vpm, target(next(h, g), g));
-      const Point_ref r = get(vpm, source(h, g));
+      Point_ref p = get(vpm, target(h, g));
+      Point_ref q = get(vpm, target(next(h, g), g));
+      Point_ref r = get(vpm, source(h, g));
 
       const Vector n = internal::construct_normal_of_STL_face(p, q, r, k);
 
@@ -296,9 +296,9 @@ bool write_STL(std::ostream& os,
     for(const face_descriptor f : faces(g))
     {
       const halfedge_descriptor h = halfedge(f, g);
-      const Point_ref p = get(vpm, target(h, g));
-      const Point_ref q = get(vpm, target(next(h, g), g));
-      const Point_ref r = get(vpm, source(h, g));
+      Point_ref p = get(vpm, target(h, g));
+      Point_ref q = get(vpm, target(next(h, g), g));
+      Point_ref r = get(vpm, source(h, g));
       const Vector n = internal::construct_normal_of_STL_face(p, q, r, k);
 
       os << "facet normal " << n << "\nouter loop"<< "\n";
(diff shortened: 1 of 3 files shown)


BGL/include/CGAL/boost/graph/IO/STL.h
```diff
@@ -267,11 +271,11 @@ bool write_STL(std::ostream& os,
     for(const face_descriptor f : faces(g))
     {
       const halfedge_descriptor h = halfedge(f, g);
-      Point_ref p = get(vpm, target(h, g));
-      Point_ref q = get(vpm, target(next(h, g), g));
-      Point_ref r = get(vpm, source(h, g));
+      const Point_ref p = get(vpm, target(h, g));
```
sloriot, line 274: `vpm's reference might already provide a const type. This will lead to compilation issues in such a case.

Stream_support/include/CGAL/IO/STL.h
```diff
@@ -222,6 +222,31 @@ bool read_STL(const std::string& fname,
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Write
 
+namespace internal {
+
+template <typename K>
+typename K::Vector_3 construct_normal_of_STL_face(const typename K::Point_3& p,
```
afabri, line 228: What about calling it `CGAL::PMP::approximate_normal()`   or similar. Or does that open a can of worms?

MaelRL: We do have `PMP::internal::normalize()`, which uses `approximate_sqrt()`, but that would make `Stream_support` depend on `PMP`.

GilesBathgate: It would be really handy to have an `approximate_normal()` function globally, (or outside PMP)

Stream_support/include/CGAL/IO/STL.h
```diff
@@ -222,6 +222,31 @@ bool read_STL(const std::string& fname,
...
+{
+  typedef typename K::FT FT;
+  typedef typename K::Vector_3 Vector;
+
+  if(k.collinear_3_object()(p, q, r))
+    return k.construct_vector_3_object()(1, 0, 0);
+
+  const Vector pq = k.construct_vector_3_object()(p, q);
+  const Vector pr = k.construct_vector_3_object()(p, r);
+  Vector res = k.construct_cross_product_vector_3_object()(pq, pr);
```
GilesBathgate, line 241: Couldn't you just do:
`Vector res = k.construct_orthogonal_vector_3_object()(p, q, r);`

lrineau (approved):
I do not know why you added https://github.com/CGAL/cgal/labels/under%20discussion. If there is a discussion, you current proposal looks good.

Stream_support/include/CGAL/IO/STL.h
MaelRL: `Kernel_3::Compute_approximate_angle_3` has already introduced inexact kernel functors (due to the trigonometry), so this could be an idea.

GilesBathgate:
@MaelRL by the way, typically you would have a name after `solid ` in the ascii STL representation. See: https://github.com/CGAL/cgal/blob/fd950dad362c2638d80f1eaaf45b0d63a145b415/BGL/include/CGAL/boost/graph/IO/STL.h#L295
It can be omitted but @kintel  and I  found that buggy software expects a space after the solid keyword. I wish I could track down the original STL spec, but the [Wikipedia article](https://en.wikipedia.org/wiki/STL_(file_format)) mentions the space (because I added it to the Wikipedia article years ago)  

For the sake of compatibility it may be worth creating a bug for either having an additional space, or preferably adding a name.

PR #6415, opened 2022-03-15
Nef_3: Proof of concept for no-hashmap approach.

GilesBathgate:
## Summary of Changes

In Nef_3 hashmaps are used to associate information with objects at runtime, for example, to flag that a HalfEdge has been visited there might be a `Unique_hash_map<Halfedge_handle,bool> visited;`. If the class HalfEdge had a member variable called `visited` this could be used instead to indicate whether the edge had been visited (although it would have to be reset afterwards) However this would mean adding lots of extraneous data to the classes depending on the method or algorithm that is processing the list of objects.

There is a generalised `info` container on all the classes which could be used. However, a mechanism is needed to make sure that the existing info isn't overwritten, and is restored to whatever it was before the processing was done.

## Release Management

* Affected package(s): Nef_3
* Issue(s) solved (if any): draft/discussion only

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/Halfedge.h
+++ b/Nef_3/include/CGAL/Nef_3/Halfedge.h
@@ -64,52 +64,55 @@ class Halfedge_base
   SFace_handle       incident_sface_;
   GenPtr             info_;
   Sphere_point       point_;
+  mutable bool       visited_;
 
  public:
 
-  Halfedge_base() : center_vertex_(), mark_(), twin_(),
-    out_sedge_(), incident_sface_(),
-    info_(), point_() {}
+      Halfedge_base() : center_vertex_(), mark_(), twin_(),
+        out_sedge_(), incident_sface_(),
+        info_(), point_(), visited_() {}
 
-    Halfedge_base(Mark m) :  center_vertex_(), mark_(m), twin_(),
-      out_sedge_(), incident_sface_(),
-      info_(), point_() {}
+      Halfedge_base(Mark m) :  center_vertex_(), mark_(m), twin_(),
+        out_sedge_(), incident_sface_(),
+        info_(), point_(), visited_() {}
 
       ~Halfedge_base() {
         CGAL_NEF_TRACEN("  destroying Halfedge item "<<&*this);
       }
 
-      Halfedge_base(const Halfedge_base<Refs>& e)
-        { center_vertex_ = e.center_vertex_;
-          point_ = e.point_;
-          mark_ = e.mark_;
-          twin_ = e.twin_;
-          out_sedge_ = e.out_sedge_;
-          incident_sface_ = e.incident_sface_;
-          info_ = 0;
-        }
+      Halfedge_base(const Halfedge_base<Refs>& e) :
+        center_vertex_(e.center_vertex_),
+        point_(e.point_),
+        mark_(e.mark_),
+        twin_(e.twin_),
+        out_sedge_(e.out_sedge_),
+        incident_sface_(e.incident_sface_),
+        info_(0),
+        visited_(false) {}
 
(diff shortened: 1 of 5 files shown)


GilesBathgate:
What I don't like about the PoC is it has to wrap the original info in a value type, and copying that value type is expensive. You have to any_cast just to check whether the flag is set. So I came up with a possible alternative that could work, but needs to slightly alter the API:

```c++
#include <CGAL/Handle_for.h>
#include <CGAL/Timer.h>
#include <CGAL/Unique_hash_map.h>
#include <boost/any.hpp>
#include <iostream>
#include <list>
#include <stack>

namespace CGAL {

const size_t itterations = 10'000'000;

using Info = boost::any;

class Context;

class Provider : public std::stack<Info>
{
public:
	Provider(){ push(Info()); }
	bool has_context(Context* c) { return context_ == c; }
	void set_context(Context* c){ context_ = c; };
	void reset() { pop(); context_=nullptr; }
private:
	Context* context_;
};

class Facet
{
	friend class Facet_handle;
	Provider provider_;
};

class Facet_handle : public Handle_for<Facet>
{
public:
	Info& info() { return ptr()->provider_.top(); }
	Provider& provider() { return ptr()->provider_; }
};

class Context
{
	std::list<Provider*> providers_;
	Info default_value;
public:
	Context() = default;

	template <typename T>
	Context(const T& d) : default_value(d) {}

	~Context() {
		for(auto& p: providers_)
			p->reset();
	}

	bool has_info(Provider& p) {
		return p.has_context(this);
	}

	template <typename T>
	T get_info(Provider& p) {
		if(!has_info(p))
			set_info(p,default_value);
		return boost::any_cast<T>(p.top());
	}

	void set_info(Provider& p,const Info& i) {
		if(has_info(p)) {
			p.top() = i;
		} else {
			p.set_context(this);
			providers_.push_back(&p);
			p.push(i);
		}
	}
};

void count_cycles(std::list<Facet_handle>& facets) {
	Context FacetCycles(-1);
	for(auto& f: facets) {
		if(!FacetCycles.has_info(f.provider())) {
			FacetCycles.set_info(f.provider(),0);
		} else {
			FacetCycles.set_info(f.provider(), FacetCycles.get_info<int>(f.provider())+1);
			std::cout << FacetCycles.get_info<int>(f.provider()) << std::endl;
		}
	}
}

void visit1(std::list<Facet_handle>& facets) {
	Context Done;
	for(auto& f: facets) {
		if(!Done.has_info(f.provider())) {
			Done.set_info(f.provider(),true);
			std::cout << "visited" << std::endl;
		}
	}
}

void visit2(std::list<Facet_handle>& facets) {
	Unique_hash_map<Facet*,bool> Done(false);
	for(auto& f: facets) {
		auto* h = const_cast<Facet*>(f.Ptr());
		if(!Done[h]) {
			Done[h]=true;
			std::cout << "visited" << std::endl;
		}
	}
}

template <typename Function>
void test(Function visit)
{
	Facet_handle f1;
	Facet_handle f2;
	Facet_handle f3;
	Facet_handle f4;

	Context c;
	c.set_info(f1.provider(),"Hello World");

	f2.info() = 1.0f;

	std::list<Facet_handle> facets;
	facets.push_back(f1);
	facets.push_back(f1);
	facets.push_back(f1);
	facets.push_back(f2);
	for(size_t i=0; i< itterations; ++i)
		facets.push_back(f3);
	for(size_t i=0; i< itterations; ++i)
		facets.push_back(f4);

	Timer t;
	t.start();
	visit(facets);
	t.stop();
	std::cout << t.time() << std::endl;

	std::cout << boost::any_cast<const char*>(f1.info()) << std::endl;
}

}

int main()
{
	CGAL::test(CGAL::visit1);
	CGAL::test(CGAL::visit2);
	return 0;
}
```

PR #6421, opened 2022-03-16
Nef_3: Use has_smaller_distance_to_point()

afabri:
## Summary of Changes

Avoids constructing exact numbers. We might write the static filter so that it uses Epick if all intervals have `inf==sup`

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -349,10 +349,8 @@ public:
 
     CGAL_assertion( initialized);
     _CGAL_NEF_TRACEN( "locate "<<p);
-    typename SNC_structure::FT min_distance;
-    typename SNC_structure::FT tmp_distance;
     Object_handle result;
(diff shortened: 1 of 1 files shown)


lrineau:
> We might write the static filter so that it uses Epick if all intervals have `inf==sup`

`CGAL::has_smaller_distance_to_point` calls `k.less_distance_to_point_2_object()(p, r, q);`, in the end. So that uses the functor `Less_distance_to_point_2` from the kernel.

If the kernel is `CGAL::Epeck`, then all predicates are a `CGAL::Static_filtered_predicate` from the file `Filtered_kernel/include/CGAL/Static_filtered_predicate.h`. That means that the logic "use the `Epick` predicates if the parameters are convertible to `Epick` without loss" is enabled, unless the macros `CGAL_DONT_USE_LAZY_KERNEL` or `CGAL_NO_STATIC_FILTERS_FOR_LAZY_KERNEL` are defined (and they are not by default).

sloriot:

PR #6428, opened 2022-03-17
Various fixes for Modifiable_priority_queue

MaelRL:
## Summary of Changes

Some fixes and enhancements for the `CGAL_BOOST_PAIRING_HEAP` specialization of `Modifiable_priority_queue`.

- Fix `pop()` not resetting the handle to null
- Disable copy and assignment since the handles become invalid when the queue is copied
- `top()` can and should return a const ref
- Make `push()` return a `handle_type` rather than `void`
- Add preconditions and assertions
- Add missing `clear()` and `size()`
- Explicitly initialize the handles in the constructor

## Release Management

* Affected package(s): `STL_Extension`
* Issue(s) solved (if any): -


sloriot (approved):

sloriot:

PR #6431, opened 2022-03-18
Nef_3:  Avoid Filter Failures

afabri:
## Summary of Changes

Check if the vertex of the ray is a vertex of an edge instead of comparing their points.

There are more filter failures that merit a closer look

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
@@ -80,7 +80,7 @@ public:
 
   virtual Object_handle locate(const Point_3& p) const = 0;
 
-  virtual Object_handle shoot(const Ray_3& s, int mask=255) const = 0;
+  virtual Object_handle shoot(const Ray_3& s, Vertex_handle ray_source_vertex, int mask=255) const = 0;
 
   virtual void intersect_with_edges( Halfedge_handle edge,
                                      const Intersection_call_back& call_back)
@@ -216,7 +216,7 @@ public:
       delete candidate_provider;
   }
 
-  virtual Object_handle shoot(const Ray_3& ray, int mask=255) const {
+  virtual Object_handle shoot(const Ray_3& ray, Vertex_handle ray_source_vertex, int mask=255) const {
     CGAL_NEF_TIMER(rs_t.start());
     CGAL_assertion( initialized);
     _CGAL_NEF_TRACEN( "shooting: "<<ray);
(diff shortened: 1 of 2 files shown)


GilesBathgate:
Do you think an overload function again here might help compatibility? (open-closed principle)

```c++
virtual Object_handle shoot(const Ray_3& s, int mask=255) const = 0;
virtual Object_handle shoot(const Ray_3& s, Vertex_handle ray_source_vertex) const = 0;

...
Object_handle shoot(const Ray_3& s, int mask=255) { ...
Object_handle shoot(const Ray_3& s, Vertex_handle ray_source_vertex) { ...
private:
Object_handle shoot(const Ray_3& s, Vertex_handle ray_source_vertex, int mask=255) { ...
```

GilesBathgate:
... or just apply the same fix here to Ray_hit_generator2.h, not sure why a duplicate file that is mostly identical exists...

https://github.com/CGAL/cgal/blob/bfffbf48843710428a6177b46a815980c677a530/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/Ray_hit_generator2.h#L80

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -250,22 +257,25 @@ class SNC_point_locator_by_spatial_subdivision :
...
-              continue;
-            _CGAL_NEF_TRACEN("is the intersection point on the current cell? "<<
-                    candidate_provider->is_point_on_cell( q, objects_iterator));
-            if( !candidate_provider->is_point_on_cell( q, objects_iterator))
-                continue;
-            eor = q;
-            result = make_object(e);
-            hit = true;
-            _CGAL_NEF_TRACEN("the edge becomes the new hit object");
+          if ((ray_source_vertex != Vertex_handle()) && (ray_source_vertex != e->source()) && (ray_source_vertex != e->twin()->source())) {
```
GilesBathgate, line 260: You check for a null_handle here but not above?

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -234,7 +241,7 @@ class SNC_point_locator_by_spatial_subdivision :
       CGAL_for_each( o, candidates) {
         if( CGAL::assign( v, *o) && ((mask&1) != 0)) {
           _CGAL_NEF_TRACEN("trying vertex on "<<v->point());
-          if( ray.source() != v->point() && ray.has_on(v->point())) {
+          if( (ray_source_vertex != v) && (ray.source() != v->point()) && ray.has_on(v->point())) {
```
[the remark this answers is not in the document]
GilesBathgate: Ah, I thought you added `ray.source() != v->point()` but it was already there. Actually thinking about it, a ray only has a source, so strictly ordered along line cannot apply.

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
GilesBathgate: Yep, that's it.

In addition, I was wondering why it does it like this on line 260 but it is not the same on line 244.

GilesBathgate: > In addition, I was wondering why it does it like this on line 260 but it is not the same on line 244.


Oh you've reverted commit [211cb75](https://github.com/CGAL/cgal/pull/6431/commits/211cb75dd34cad6a07d3ea2ed18446c5fb36879a) Well that explains the second question at least ;)

GilesBathgate:
@afabri  Do you think these additional changes would be of benefit:

https://github.com/CGAL/cgal/commit/479fe0dcac65cbd4c8d14bb6d7b065821de8521b#diff-8316acfe25c13d7862fec508cb0f539e6a0a8c08ee0097a3c48120fd8a1708feR301-R316

sloriot:

PR #6441, opened 2022-03-21
PMP repair: Avoid large DAG when using Lazy

sloriot:
Avoid issues with EPECK


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair.h
```diff
@@ -269,26 +265,16 @@ std::size_t remove_connected_components_of_negligible_size(TriangleMesh& tmesh,
     typename GT::Compute_volume_3 cv3 = traits.compute_volume_3_object();
     Point_3 origin(0, 0, 0);
```
afabri, line 266: you no longer need the functor and the origin.

PR #6447, opened 2022-03-22
Tetrahedral remeshing - avoid useless dumps

janetournois:
## Summary of Changes

Protect the dump of statistics and bad cells with the _DEBUG macro instead of the _VERBOSE macro (which is used in the demo)

## Release Management

* Affected package(s): Tetrahedral remeshing


sloriot:
Successfully tested in CGAL-5.5-Ic-55

PR #6454, opened 2022-03-28
Stream_support: Make line a data member and call clear()

afabri:
## Summary of Changes

Instead of creating again and again an empty string for a call to   `getline()` we turn the local variable `line` into a data member and call `std::string::clear()`. 

This PR does it for reading OFF files, but it might apply to other readers as well,
It would be good if somebody else could confirm that there is a (small) performance gain.

## Release Management

### DIFF
--- a/Stream_support/include/CGAL/IO/OFF/File_scanner_OFF.h
+++ b/Stream_support/include/CGAL/IO/OFF/File_scanner_OFF.h
@@ -37,6 +37,7 @@ class File_scanner_OFF
   std::vector<double> entries;
   std::size_t color_entries;
   std::size_t first_color_index;
+  std::string line;
   std::istream& m_in;
   bool normals_read;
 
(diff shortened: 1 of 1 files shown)


lrineau (approved):

lrineau:
Note: [`std::getline`](https://en.cppreference.com/w/cpp/string/basic_string/getline) already clears the string before the operation, but I find clearer to have an explicit `line.clear()` anyway.

sloriot:
Successfully tested in CGAL-5.5-Ic-52

Stream_support/include/CGAL/IO/OFF/File_scanner_OFF.h
```diff
@@ -78,7 +79,7 @@ class File_scanner_OFF
     else
     {
       skip_comment();
-      std::string line;
+      line.clear();
       std::getline(m_in, line);
       // First remove the comment if there is one
       std::size_t pos = line.find('#');
```
GilesBathgate, line 85: I just had a thought. Would it not be beneficial to do the same with the 
`std::istringstream issline(line);`

i.e move it to a class member and call `issline.clear()`

GilesBathgate: you can set the content with `issline.str(line)`

PR #6465, opened 2022-03-31
Fix traits adapters

sloriot:
Fixes #6464


afabri:
We should add it to the testsuite.

lrineau:
> We should add it to the testsuite.

I think Sébastien already did. He has added a few lines in an example: https://github.com/CGAL/cgal/pull/6465/files#diff-dd0f5b062d6007947a70393ae86b512c61557195db4039ca83f11aafe99c8e0dR34.

PR #6469, opened 2022-04-02
Nef_3: Use cross product instead of Plane_Plane intersection

GilesBathgate:
## Summary of Changes

Whilst implementing #6378 I noticed another minor optimisation. The Plane_Plane intersection of two sphere circles can be simplified to the cross product of the two planes orthogonal vectors. 

## Release Management

### DIFF
--- a/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
+++ b/Nef_S2/include/CGAL/Nef_S2/Sphere_circle.h
@@ -173,11 +173,9 @@ Sphere_point<R> intersection(const Sphere_circle<R>& c1,
 |c1| and |c2|. \precond |c1 != c2| as sets.}*/
 {
   CGAL_assertion(!equal_as_sets(c1,c2));
-  typename R::Line_3 lres;
   CGAL_NEF_TRACEN("circle_intersection "<<c1<<" "<<c2);
-  CGAL::Object o = CGAL::intersection(c1.plane(),c2.plane());
-  if ( !CGAL::assign(lres,o) ) CGAL_error();
-  return CGAL::ORIGIN + lres.direction().vector();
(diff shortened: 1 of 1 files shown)


GilesBathgate:
When I benchmarked with the spheregrid intersection it was 0.5% faster.

afabri:
I approve this PR and just want to add something to the discussion. 
The construction of an orthogonal vector for a sphere circle shows up in vtune which is somehow a pity, as a `Sphere_circle` derives from `Plane_3` with the particularity that `d` is always zero, that is we basically use the plane for storing the orthogonal vector.

afabri (approved):

GilesBathgate:
> We basically use the plane for storing the orthogonal vector.
It's an interesting observation. I still think changing the code to leverage this fact might not be worth the effort.

GilesBathgate:
> If the profiler shows that the construction of that orthogonal vector takes 5% or more of the total time, for example, then that can be interesting to optimize

I didn't intend to imply I was ruling it out, rather that it could be considered in another PR.

GilesBathgate:
Would we ultimately find a sphere point is a vector, a sphere circle is a vector, and a sphere segment is two vectors 😁

GilesBathgate:
@lrineau In particular id be interested if this can be simplified. Instead of checking twice the side of plane (including the orthogonal) pole maybe checking orientation `s` , `p`, `e` (where s and e are the segment start and end sphere points) is coplanar, and then whether its the long or short segment.

GilesBathgate:
> @lrineau In particular id be interested if this can be simplified.

Forgot to include the link:
https://github.com/CGAL/cgal/blob/1c4b191c71b189c719003bdbe98a5954d1f2ed40/Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h#L348-L362

PR #6470, opened 2022-04-03
Nef_3: Avoid more filter failures in ray shoot

GilesBathgate:
## Summary of Changes

Whilst looking at #6431 I notices the same optimisation can be made for facets, (which is already done in locate). The code was moved into a static function to avoid duplication.

In addition, the call to shoot, via `determine_volume` passes in `MinimalSFace[ShellSf[f]]`, where `determine_volume` already does: `MinimalSFace[ShellSf[sf]]` , making the compound operation equivalent to `MinimalSFace[ShellSf[MinimalSFace[ShellSf[f]]]]` which is redundant.

## Release Management

* Affected package(s): Nef_3


GilesBathgate:
When I benchmarked with the spheregrid intersection it was 1.82% faster.

afabri (approved):

sloriot:
Sucessfully tested in CGAL-5.5-Ic-95

PR #6471, opened 2022-04-03
Nef_3: Performance std::map -> std::unordered_map

GilesBathgate:
## Summary of Changes

There are a couple of places in Nef_3 where `std::map<int, int>` is used even though the order isn't important and the number of elements that will be placed in the map is known ahead of use. `std::unordered_map<int, int>` is a better choice when this is the case.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
+++ b/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
@@ -56,10 +56,14 @@ class ID_support_handler<SNC_indexed_items, Decorator> {
     }
   };
   std::unordered_map<Halffacet_pair, int, Handle_pair_hash_function> f2m;
-  std::map<int, int> hash;
+  std::unordered_map<int, int> hash;
 
  public:
   ID_support_handler() {}
+
+  void reserve(std::size_t n) {
+    hash.reserve(n);
+  }
 
   int get_hash(int i) {
     int root(i);
@@ -78,12 +82,13 @@ class ID_support_handler<SNC_indexed_items, Decorator> {
     hash[get_hash(i)] = parent;
   }
 
+  void initialize_hash(int i) {
+    hash[i] = i;
+  }
+
   template<typename Handle>
   void initialize_hash(Handle h) {
-    hash[h->get_index()] = h->get_index();
-  }
-  void initialize_hash(int i) {
-        hash[i] = i;
+    initialize_hash(h->get_index());
   }
 
   void hash_facet_pair(SVertex_handle sv,
--- a/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
+++ b/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
@@ -65,16 +65,18 @@ class ID_support_handler<SNC_indexed_items, Decorator> {
     hash.reserve(n);
   }
 
-  // The method get_hash implements a
-  // two-pass union find algorithm
-  //    __    __           _______
-  //   /  \  /  \         /  _____\
-  //  _|__|__|__|_      _/__/__/__|_
-  // | |  v  |  v |    | |  |  |  v |
-  // | O  O  O  O | => | O  O  O  O |
-  // |____|__^____|    |____________|
-  //      |  | root
-  //      \_/
+  /*
+     The method get_hash implements a
+     two-pass union find algorithm
+        __    __           _______
+       /  \  /  \         /  _____\
+      _|__|__|__|_      _/__/__/__|_
+     | |  v  |  v |    | |  |  |  v |
+     | O  O  O  O | => | O  O  O  O |
+     |____|__^____|    |____________|
+          |  | root
+          \_/
+  */
   int get_hash(int i) {
     int root(i);
     while(hash[root] != root)
--- a/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
+++ b/Nef_3/include/CGAL/Nef_3/ID_support_handler.h
@@ -65,6 +65,16 @@ class ID_support_handler<SNC_indexed_items, Decorator> {
     hash.reserve(n);
   }
 
+  // The method get_hash implements a
+  // two-pass union find algorithm
+  //    __    __           _______
+  //   /  \  /  \         /  _____\
+  //  _|__|__|__|_      _/__/__/__|_
+  // | |  v  |  v |    | |  |  |  v |
+  // | O  O  O  O | => | O  O  O  O |
+  // |____|__^____|    |____________|
+  //      |  | root
+  //      \_/
   int get_hash(int i) {
     int root(i);
     while(hash[root] != root)
--- a/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
@@ -30,6 +30,7 @@
 #include <CGAL/Nef_3/SNC_simplify.h>
 #include <map>
 #include <list>
+#include <unordered_map>
 
 #undef CGAL_NEF_DEBUG
 #define CGAL_NEF_DEBUG 43
--- a/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
@@ -1308,8 +1308,10 @@ public:
      //     O0.print();
     link_shalfedges_to_facet_cycles();
 
-    std::map<int, int> hash;
-    CGAL::Unique_hash_map<SHalfedge_handle, bool> done(false, this->sncp()->number_of_shalfedges());
+    std::size_t num_shalfedges = this->sncp()->number_of_shalfedges();
+    std::unordered_map<int, int> hash;
+    hash.reserve(num_shalfedges);
+    CGAL::Unique_hash_map<SHalfedge_handle, bool> done(false, num_shalfedges);
 
     SHalfedge_iterator sei;
     CGAL_forall_shalfedges(sei, *this->sncp()) {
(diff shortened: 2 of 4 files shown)


GilesBathgate:
When I benchmarked with the spheregrid intersection it was 3.17% faster.

afabri:
Why could you get rid of the `while` loops in `get_hash()`?

GilesBathgate:
> Why could you get rid of the `while` loops in `get_hash()`?

I did intentionally remove it because I think the code is redundant, It's quite a complex explanation...

GilesBathgate:
The first while loop can be unrolled... After the first iteration check can never be true.

```diff
   int get_hash(int i) {
     int root(i);
-    while(hash[root] != root)
+    if(hash[root] != root)
       root = hash[root];
+    //if(hash[root] != root)
+
     while(hash[i] != i) {
       int tmp = hash[i];
       hash[i] = root;
```

Since i doesn't change i can be used instead of root

```diff
   int get_hash(int i) {
     int root(i);
-    if(hash[root] != root)
-      root = hash[root];
-    //if(hash[root] != root)
+    if(hash[i] != root)
+      root = hash[i];
+    //if(hash[i] != root)
 
     while(hash[i] != i) {
       int tmp = hash[i];
```

Unrolling the second while. Since hash[i] was assigned to root, and i was assigned to what hash[I] was before the loop, the if is the same check as made in the first if.

```diff
       root = hash[i];
     //if(hash[i] != root)
 
-    while(hash[i] != i) {
+    int hash_i = hash[i];
+    if(hash[i] != i) {
       int tmp = hash[i];
       hash[i] = root;
       i = tmp;
     }
+    //if(root != hash_i)
+
     return root;
   }
```

Since i isn't used beyond the if, i=tmp is redundant, so remove tmp too

```diff
     int hash_i = hash[i];
     if(hash[i] != i) {
-      int tmp = hash[i];
       hash[i] = root;
-      i = tmp;
     }
     //if(root != hash_i)
```

Sub in i for root

```diff
   int get_hash(int i) {
     int root(i);
-    if(hash[i] != root)
+    if(hash[i] != i)
       root = hash[i];
     //if(hash[i] != root)
```

Combine the if's since the check is identical

```diff
  int get_hash(int i) {
     int root(i);
-    if(hash[i] != i)
-      root = hash[i];
     //if(hash[i] != root)
 
     int hash_i = hash[i];
     if(hash[i] != i) {
+      root = hash[i];
       hash[i] = root;
     }
     //if(root != hash_i)
```

 hash[i] = root is the same as hash[i] = hash[i] because of the line above

```diff
     int hash_i = hash[i];
     if(hash[i] != i) {
       root = hash[i];
-      hash[i] = root;
+      //hash[i] = hash[i];
     }
     //if(root != hash_i)
```

Cleanup

```diff
   int get_hash(int i) {
     int root(i);
-    //if(hash[i] != root)
 
-    int hash_i = hash[i];
     if(hash[i] != i) {
       root = hash[i];
-      //hash[i] = hash[i];
     }
-    //if(root != hash_i)
 
     return root;
   }
```

 Sub in root for I. Now it can be seen that the condition ensures root = hash[root]
 
 ```diff
    int get_hash(int i) {
     int root(i);
 
-    if(hash[i] != i) {
-      root = hash[i];
-    }
+    if(hash[root] != root) {
+      root = hash[root];
+    } //else root == hash[root]
 
     return root;
   }
```

 If can be replaced with an assignment
 
 ```diff
    int get_hash(int i) {
     int root(i);
 
-    if(hash[root] != root) {
-      root = hash[root];
-    } //else root == hash[root]
+    root = hash[root];
 
     return root;
   }
```

Sub in i, remove redundant initalisation

```diff
   int get_hash(int i) {
-    int root(i);
+    int root;
 
-    root = hash[root];
+    root = hash[i];
 
     return root;
   }
```

Remove redundant local variable

```diff
   int get_hash(int i) {
-    int root;
-
-    root = hash[i];
-
-    return root;
+    return hash[i];
   }
```

GilesBathgate:
Actually, I think my explanation is wrong, but I don't understand what the code is intended to do, nor why it works without it.

GilesBathgate:
I'll leave the `get_hash` code as it was, since changing it is not needed for the performance improvement.

Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
```diff
@@ -1308,8 +1308,10 @@ class SNC_external_structure<SNC_indexed_items, SNC_structure_>
      //     O0.print();
     link_shalfedges_to_facet_cycles();
 
-    std::map<int, int> hash;
-    CGAL::Unique_hash_map<SHalfedge_handle, bool> done(false, this->sncp()->number_of_shalfedges());
+    std::size_t num_shalfedges = this->sncp()->number_of_shalfedges();
+    std::unordered_map<int, int> hash;
+    hash.reserve(num_shalfedges);
```
mglisse, line 1313: There is a constructor for that.

GilesBathgate: Thanks

afabri:
image

Nef_3/include/CGAL/Nef_3/ID_support_handler.h
```diff
@@ -78,12 +82,13 @@ class ID_support_handler<SNC_indexed_items, Decorator> {
     hash[get_hash(i)] = parent;
   }
 
+  void initialize_hash(int i) {
```
afabri, line 85: Did you intentionally move this overload before the function template?  If not, let's move it back in order to have a smaller diff.

afabri:
> When I benchmarked with the spheregrid intersection it was 3.17% faster.

I can reproduce that it is slightly faster.

Nef_3/include/CGAL/Nef_3/ID_support_handler.h
GilesBathgate: I guess the order of declaration doesn't matter in a C++ class, but I thought from a reader's perspective it's better to have the dependant method first, and I made the template method depend on this one.

GilesBathgate:
> The function ` get_hash` does the following (just like in union find) image

Ah: https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Finding_set_representatives
```
function Find(x) is
    root := x
    while root.parent ≠ root do
        root := root.parent
    end while

    while x.parent ≠ root do
        parent := x.parent
        x.parent := root
        x := parent
    end while

    return root
end function
```

Can I add a comment to the code here, to explain that its the two pass union find algorithm?

afabri:
> Can I add a comment to the code here, to explain that its the two pass union find algorithm?

sure. Then the next generation does not have to reverse engineer it, as we just did.

Nef_3/include/CGAL/Nef_3/ID_support_handler.h
afabri: Fine with me, as t is intentional.

GilesBathgate:
@afabri A comment like this? or just the words and no ascii-art?:

```c++
  // The following implements a two-pass
  // union find algorithm
  //    __    __            _______
  //   /  \  /  \          /  _____\
  //  _|__|__|__|__      _/__/__/__|_
  // | |  v  |  v  |    | |  |  |  v |
  // | O  O  O  O  | => | O  O  O  O |
  // |____|__^_____|    |____________|
  //      |  |  root
  //      \_/
  int get_hash(int i) {
    int root(i);
    while(hash[root] != root)
      root = hash[root];
    while(hash[i] != i) {
      int tmp = hash[i];
      hash[i] = root;
      i = tmp;
    }
    return root;
  }
  ```

**Comment by afabri** [comment]
> @afabri A comment like this? or just the words and no ascii-art?:
> 
> ```c++
>   // The following implements a two-pass
>   // union find algorithm
>   //    __    __            _______
>   //   /  \  /  \          /  _____\
>   //  _|__|__|__|__      _/__/__/__|_
>   // | |  v  |  v  |    | |  |  |  v |
>   // | O  O  O  O  | => | O  O  O  O |
>   // |____|__^_____|    |____________|
>   //      |  |  root
>   //      \_/
>   int get_hash(int i) {
>     int root(i);
>     while(hash[root] != root)
>       root = hash[root];
>     while(hash[i] != i) {
>       int tmp = hash[i];
>       hash[i] = root;
>       i = tmp;
>     }
>     return root;
>   }
> ```

The original aurhors liked literate programming and will love ascii art.

**Review by afabri — APPROVED** [review]

**Comment by GilesBathgate** [comment]
```diff
diff --git a/Nef_S2/include/CGAL/Nef_S2/ID_support_handler.h b/Nef_S2/include/CGAL/Nef_S2/ID_support_handler.h
index 9a9b3bcfcd1..6704462d201 100644
--- a/Nef_S2/include/CGAL/Nef_S2/ID_support_handler.h
+++ b/Nef_S2/include/CGAL/Nef_S2/ID_support_handler.h
@@ -38,6 +38,7 @@ class ID_support_handler {
  public:
   ID_support_handler() {}
 
+  void reserve(std::size_t n) {}
   int get_hash(int) { return 0; }
   template<typename Handle> void initialize_hash(Handle /*h*/) {}
   void initialize_hash(int /*i*/) {}
```

afabri:
I just added exactly that and run the testsuite locally before pushing.     As best practice we both should run ctest on the Nef related packages as described in the [wiki](https://github.com/CGAL/cgal/wiki/Testing#using-ctest)

afabri:
We had a warning in this [testsuite ](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-63/Nef_3_Examples/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz)

Nef_S2/include/CGAL/Nef_S2/ID_support_handler.h
```diff
@@ -38,6 +38,8 @@ class ID_support_handler {
  public:
   ID_support_handler() {}
 
+  void reserve(std::size_t n) {}
```
sloriot, line 41:
```suggestion
  void reserve(std::size_t) {}
```

sloriot:
Successfully tested in CGAL-5.4-Ic-86

PR #6476, opened 2022-04-06
Simplify code and make it read-only thread safe

sloriot:
I guess it should replace #6474 

@afabri, with and without the tss should be benched.

Each commit contains a unique change.

### DIFF
--- a/Hash_map/include/CGAL/Hash_map/internal/chained_map.h
+++ b/Hash_map/include/CGAL/Hash_map/internal/chained_map.h
@@ -139,7 +139,7 @@ void chained_map<T, Allocator>::init_table(std::size_t n)
   free = table + t;
   table_end = table + t + t/2;
 
-  for (Item p = table; p < free; p++)
+  for (Item p = table; p < free; ++p)
   { p->succ = nullptr;
     p->k = nullkey;
   }
@@ -173,7 +173,7 @@ void chained_map<T, Allocator>::rehash()
 
   Item p;
 
-  for(p = old_table; p < old_table_mid; p++)
+  for(p = old_table; p < old_table_mid; ++p)
   { std::size_t x = p->k;
     if ( x != nullkey ) // list p is non-empty
     { Item q = HASH(x);
@@ -185,7 +185,7 @@ void chained_map<T, Allocator>::rehash()
   while (p < old_table_end)
   { std::size_t x = p->k;
     insert(x,p->i);
-    p++;
+    ++p;
   }
 
   for (Item item = old_table ; item != old_table_end ; ++item)
@@ -238,7 +238,7 @@ chained_map<T, Allocator>::chained_map(const chained_map<T, Allocator>& D)
 {
   init_table(D.table_size);
 
-  for(Item p = D.table; p < D.free; p++)
+  for(Item p = D.table; p < D.free; ++p)
   { if (p->k != nullkey || p >= D.table + D.table_size)
     { insert(p->k,p->i);
       //D.copy_inf(p->i);  // see chapter Implementation
@@ -253,7 +253,7 @@ chained_map<T, Allocator>& chained_map<T, Allocator>::operator=(const chained_ma
 
   init_table(D.table_size);
 
-  for(Item p = D.table; p < D.free; p++)
+  for(Item p = D.table; p < D.free; ++p)
   { if (p->k != nullkey || p >= D.table + D.table_size)
     { insert(p->k,p->i);
       //copy_inf(p->i);    // see chapter Implementation
@@ -299,8 +299,8 @@ template <typename T, typename Allocator>
 void chained_map<T, Allocator>::statistics() const
 { std::cout << "table_size: " << table_size <<"\n";
   std::size_t n = 0;
-  for (Item p = table; p < table + table_size; p++)
-     if (p ->k != nullkey) n++;
+  for (Item p = table; p < table + table_size; ++p)
+     if (p ->k != nullkey) ++n;
   std::size_t used_in_overflow = free - (table + table_size );
   n += used_in_overflow;
   std::cout << "number of entries: " << n << "\n";
--- a/Hash_map/include/CGAL/Hash_map/internal/chained_map.h
+++ b/Hash_map/include/CGAL/Hash_map/internal/chained_map.h
@@ -13,8 +13,8 @@
 //
 //
 // Author(s)     : Courtesy of LEDA
-#ifndef CGAL_CHAINED_MAP_H
-#define CGAL_CHAINED_MAP_H
+#ifndef CGAL_HASH_MAP_INTERNAL_CHAINED_MAP_H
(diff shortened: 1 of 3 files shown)


sloriot:
@GilesBathgate you might want to have a look at the changes.

GilesBathgate:
@sloriot Yeah, cool. Removing the old_table entirely was something I experimented with [here ](https://github.com/GilesBathgate/cgal/commit/af104977d895ff92e79ce7a1b1ac49046cf53ac5)

GilesBathgate:
@sloriot Is the `T p = access(old_index)...` entirely redundant, it looks like you removed it.

Hash_map/include/CGAL/Tools/chained_map.h
```diff
@@ -233,8 +239,8 @@ T& chained_map<T, Allocator>::access(chained_map_item p, std::size_t x)
 
 
 template <typename T, typename Allocator>
-chained_map<T, Allocator>::chained_map(std::size_t n)
-  : table(nullptr), reserved_size(n)
+chained_map<T, Allocator>::chained_map(std::size_t n, const T& d)
+  : table(nullptr), reserved_size(n), def(d)
```
GilesBathgate, line 233: I think warnings get emitted when the initialisation order here doesn't match the member declaration order.

sloriot:
> @sloriot Is the `T p = access(old_index)...` entirely redundant, it looks like you removed it.
> 
> (it works with my basic tests, so probably only needed when the code delays deletion of the old_table, I could never figure out why that was ever necessary)

We could not find any reason either why this would be needed. As this code is inherited from LEDA, my guess is that maybe it was useful for something that did not make it in CGAL.

Hash_map/include/CGAL/Tools/chained_map.h
```diff
@@ -187,19 +165,17 @@ inline void chained_map<T, Allocator>::insert(std::size_t x, T y)
 template <typename T, typename Allocator>
 void chained_map<T, Allocator>::rehash()
 {
-  old_table = table;
-  old_table_end = table_end;
-  old_table_size = table_size;
-  old_table_size_1 = table_size_1;
-  old_free = free;
+  chained_map_elem<T>* old_table = table;
```
GilesBathgate, line 168: I don't know if you want to simplify the typedefs:
```c++
typedef chained_map_elem<T>*  chained_map_item;
typedef chained_map_item item;
```
and then just use one of them consistently?

GilesBathgate:
@sloriot @afabri There are definitely some performance gains from this PR too. I can publish some results later, but you can also verify the results of `Hash_map/benchmark/Hash_map/polyhedron`.

Hash_map/include/CGAL/Tools/chained_map.h
```diff
@@ -37,42 +37,33 @@ class chained_map_elem
 template <typename T, typename Allocator>
 class chained_map
 {
-   const std::size_t nullptrKEY;
-   const std::size_t NONnullptrKEY;
-
-   chained_map_elem<T> STOP;
+   static constexpr std::size_t nullptrKEY = (std::numeric_limits<std::size_t>::max)();
```
lrineau, line 40: I think you should modify that variable name. It was a victim of the transformation `NULL`->`nullptr`.

It should be renamed `NULLKEY` (its original name), or `nullkey` (that is better because it satisfies our naming guildelines).

Hash_map/include/CGAL/Hash_map/internal/chained_map.h
```diff
@@ -161,18 +139,17 @@ void chained_map<T, Allocator>::init_table(std::size_t n)
   free = table + t;
   table_end = table + t + t/2;
 
-  for (chained_map_item p = table; p < free; p++)
-  { p->succ = &STOP;
-    p->k = nullptrKEY;
+  for (Item p = table; p < free; p++)
```
afabri, line 142: `++p`

afabri:
Should we use something like [`_BitScanReverse`](https://stackoverflow.com/questions/466204/rounding-up-to-next-power-of-2) to save the loop in the beginning of `init_table()`?  Probably an overkill.

GilesBathgate:
@sloriot Hi I was wondering if you could help me with something... I want to add an `erase()` method to the chained map, so far I have:

```c++
template <typename T, typename Allocator>
void chained_map<T, Allocator>::erase(std::size_t x)
{
  if(!table)
    return;

  Item p,q = HASH(x);

  if (q && q->k == x) {
    Item tmp=q->succ;
    if(tmp) {
      // move succ to head
      q->i = tmp->i;
      q->k = tmp->k;
      q->succ = tmp->succ;
      destroy(tmp);
    } else {
      // create new head
      destroy(q);
      construct(q);
      q->k = nullkey;
      q->succ = nullptr;
    }
    return;
  }

  while (q && q->k != x) {
    p = q;
    q = q->succ;
  }

  if (q == nullptr)
    return;

  p->succ = q->succ;

  destroy(q);
}
```

I think it's right but I am not really sure how to handle the case when it needs to remove the head from the chain.

Hash_map/include/CGAL/Hash_map/internal/chained_map.h
```diff
@@ -37,42 +37,33 @@ class chained_map_elem
 template <typename T, typename Allocator>
 class chained_map
 {
-   const std::size_t nullptrKEY;
-   const std::size_t NONnullptrKEY;
-
-   chained_map_elem<T> STOP;
+   static constexpr std::size_t nullkey = (std::numeric_limits<std::size_t>::max)();
```
GilesBathgate, line 40: I get some warning about this line:
```
chained_map.h:40:49: error: implicit instantiation of undefined template 'std::numeric_limits<unsigned long>'
max_size_type.h:53:10: note: template is declared here
```
If I add `#include <limits>` I then get:
```
chained_map.h:41:33: error: constexpr variable 'nullkey' must be initialized by a constant expression
limits:1274:57: note: value 18446744073709551614 is outside the range of representable values of type 'long long'
chained_map.h:41:43: note: in call to 'max()'
```

Wasn't sure how to fix it but 
```c++
static constexpr std::size_t nullkey = SIZE_MAX;
```
Seems to work.

sloriot:
The issue with erase is that you would need to update the free list. 
Indeed, first element in the chained map are within the first `t`
elements of the array. Then successor are in the t/2 elements after
the `t`'th. And free elements must be consecutive.
It's doable but requires some swap to restore those three properties.

Best,

Sebastien.

GilesBathgate:
> The issue with erase is that you would need to update the free list. Indeed, first element in the chained map are within the first `t` elements of the array. Then successor are in the t/2 elements after the `t`'th. And free elements must be consecutive. It's doable but requires some swap to restore those three properties. Best, Sebastien.

I think I just need to move the predecessor of free in the vacancy and decrement free. There would need to be special handling when free is at the start. I'll give it a try.

PR #6478, opened 2022-04-06
Nef_3: Allocate a big enough vector for the nodes of the K3_tree

afabri:
## Summary of Changes

Replace the `boost::container::deque<Node> nodes;`  by an `std::vector<Node>`  and reserve (too much) .

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -22,13 +22,11 @@
 #include <CGAL/Nef_3/quotient_coordinates_to_homogeneous_point.h>
 #include <CGAL/Lazy_kernel.h>
 #include <CGAL/Cartesian.h>
-#include <boost/container/deque.hpp>
 
-
-#include <deque>
 #include <sstream>
 #include <string>
 #include <map>
+#include <vector>
 
 #undef CGAL_NEF_DEBUG
 #define CGAL_NEF_DEBUG 503
@@ -212,7 +210,7 @@ private:
   Object_list object_list;
 };
 
-  typedef boost::container::deque<Node> Node_range;
+  typedef std::vector<Node> Node_range;
   typedef Node* Node_handle;
 
 
@@ -439,6 +437,7 @@ public:
     CGAL_NEF_TRACEN("reference counted " << reference_counted);
 #endif
     non_efective_splits=0;
+    nodes.reserve(2 * (vertices.size() + edges.size() + facets.size()));
     root = build_kdtree(vertices, edges, facets, 0);
   }
   const Object_list& objects_around_point( const Point_3& p) const {


GilesBathgate:
When you say "(too much)", is it worst case `O(n)` or something else? (I use Big-Oh to refer to memory not speed)

Do you think it could be made more accurate by "presorting the data in each of k dimensions prior to building the tree", as per: https://jcgt.org/published/0004/01/03/

afabri:
I made it of size 2 * number of elements to store in case that each leaf node stores only one element, in order to be sure that no reallocation happens.

GilesBathgate:
> I made it of size 2 * number of elements to store in case that each leaf node stores only one element, in order to be sure that no reallocation happens.

That sounds reasonable.Previously I made this change:
https://github.com/CGAL/cgal/blob/68a1fb4aad6741a56cc25d045abd3dd620dacdc8/Nef_3/include/CGAL/Nef_3/K3_tree.h#L650
To ensure that each leaf contains at least two vertices. So changing this back or changing the `2 *` might be worth investigation?

afabri:
> To ensure that each leaf contains at least two vertices. So changing this back or changing the `2 *` might be worth investigation?

I missed that. You are right that the `2 *` is then not needed.

Nef_S2/include/CGAL/Nef_S2/Sphere_direction.h
```diff
@@ -57,8 +57,8 @@ the great circle through $p$ and $q$ (oriented such that the segment
 $pq$ is the shorter one of the two possible ones. \precond $p$ and $q$
 are not opposite on $S_2$.}*/
 { CGAL_assertion(p!=q.opposite());
-  Point_3 p1(0,0,0), p4 = CGAL::ORIGIN + ((Base*) this)->orthogonal_vector();
-  if ( CGAL::orientation(p1,p,q,p4) != CGAL::POSITIVE )
+  Point_3 p4 = CGAL::ORIGIN + ((Base*) this)->orthogonal_vector();
+  if ( CGAL::orientation(CGAL::ORIGIN,p,q,p4) != CGAL::POSITIVE )
```
lrineau, line 61: If I remember well, in PR #6399 you decided not to add the optimized overload to the global function. See https://github.com/CGAL/cgal/pull/6399#issuecomment-1071565003.

GilesBathgate:
> I missed that. You are right that the `2 *` is then not needed.

I think I should explain why I made that change. The K3_tree is mostly intended for looking up objects. Since an edge always contains 2 points, it doesn't make sense to me to split smaller than 2 points. Obviously, in the case where it is finding a vertex it now has at least 2 to check, but I didn't think this made a significant difference.

Nef_S2/include/CGAL/Nef_S2/Sphere_direction.h
GilesBathgate: That's well spotted! I also think this change is orthogonal to the allocation of a vector in K3_tree.

afabri: Seems the function is not called. I had to fix exacty that in line 105 below.

afabri: The file contains some more places where we should replace `Point_3(0,0,0)`.  Is it ok if we deal with this in this PR?

GilesBathgate: If the commits are atomic, and the PR title/description matches the change scope, this workflow works for me. ;)

GilesBathgate:
This commit [26571bf](https://github.com/CGAL/cgal/pull/6478/commits/26571bf3549aa553e7febae5c819c18d6dfc710a) causes a crash in my tests, I will provide an example to re-produce.

GilesBathgate:
```c++
#include <iostream>

using K = CGAL::Exact_predicates_exact_constructions_kernel;
using P = CGAL::Polyhedron_3<K>;
using N = CGAL::Nef_polyhedron_3<K>;
using V = CGAL::Vector_3<K>;

{
	std::ifstream cube("cube.off"); // cube is attached
	P p;
	cube >> p;

	N result;
	for(int i=0; i<=33; i+=11) {
		N n(p);
		CGAL::AffTransformation3 t(CGAL::TRANSLATION,V(0,0,i));
		n.transform(t);
		result += n;
	}
}
```

[cube.zip](https://github.com/CGAL/cgal/files/8639117/cube.zip)

sloriot:
@afabri could you please check the code shared by Giles?

afabri:
I can reproduce the problem and investigate now.

GilesBathgate:
@afabri Hello, and Happy New Year :confetti_ball: :tada: 

The number of nodes in a kd-tree depends on the number of points you are inserting and the structure of the tree. It is not possible to determine the number of nodes in the tree without more information.

If the points are uniformly distributed, the tree is likely to have a balanced structure, with roughly the same number of points in each leaf node. On the other hand, if the points are concentrated in a small region of the space, the tree is likely to be more imbalanced, with some leaf nodes having many points and others having few points.

Furthermore, if the `std::vector` is under-allocated when it grows the referential integrity of the vector is invalidated, which is what causes the crash in [my comment](https://github.com/CGAL/cgal/pull/6478#issuecomment-1119445347)

Therefore the best approach is to revert back to using `std::deque`, which does not invalidate pointers, references, or iterators when it grows. However, I realise initially the whole purpose of the change was to use a container which could be preallocated, even though I don't think this can be done, you made some other changes which I still think are worthwhile.


```diff
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -210,7 +210,7 @@ private:
   Object_list object_list;
 };
 
-  typedef std::vector<Node> Node_range;
+  typedef std::deque<Node> Node_range;
   typedef Node* Node_handle;
 
 
@@ -437,7 +437,6 @@ public:
     CGAL_NEF_TRACEN("reference counted " << reference_counted);
 #endif
     non_efective_splits=0;
-    nodes.reserve(vertices.size() + edges.size() + facets.size());
     root = build_kdtree(vertices, edges, facets, 0);
   }
   const Object_list& objects_around_point( const Point_3& p) const {
```

afabri:
Hello @GilesBathgate,
All the best for the new year. And please never give up on reminders.  They are really appreciated, and you see that it also works.
Best, Andreas

Nef_S2/include/CGAL/Nef_S2/Sphere_direction.h
```diff
@@ -95,18 +95,20 @@ bool strictly_ordered_ccw_at(const Sphere_point<R>& p,
   const Sphere_direction<R>& d2,
   const Sphere_direction<R>& d3)
 { CGAL_assertion(d1.has_on(p) && d2.has_on(p) && d3.has_on(p));
-  typename R::Point_3 p0(0,0,0);
-  typename R::Point_3 p1(CGAL::ORIGIN + d1.orthogonal_vector());
-  typename R::Point_3 p2(CGAL::ORIGIN + d2.orthogonal_vector());
-  typename R::Point_3 p3(CGAL::ORIGIN + d3.orthogonal_vector());
+  typedef typename R::Vector_3 Vector_3;
+  Vector_3 v0 = p - CGAL::ORIGIN;
```
GilesBathgate, line 99: Might there be any advantage to making `v0` through to `v3` declared as  `const Vector_3&` even if Vector_3 is cheap to copy?

afabri:
@sloriot I think we can remove the "Not  yet approved".

sloriot:
Successfully tested in CGAL-5.6-Ic-187

PR #6480, opened 2022-04-06
Nef_3: Use separate lists K3_tree node to avoid object casts

GilesBathgate:
## Summary of Changes

It is believed that the object casts in K3_tree/SNC_point_locator creates an overhead. This pull request aims to reduce this overhead by keeping lists of vertices, edges and facets separately in the Node.

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -394,7 +394,7 @@ public:
     non_efective_splits=0;
     root = build_kdtree(vertices, edges, facets, 0);
   }
-  Node_handle locate_node_containing( const Point_3 p) const {
+  Node_handle locate_node_containing( const Point_3& p) const {
     return locate_node_containing( p, root);
   }
   Node_list nodes_along_ray( const Ray_3& r) const {
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -193,8 +193,8 @@ private:
   Node_handle right_node;
   Plane_3 splitting_plane;
   Vertex_list vertex_list;
-  Halffacet_list facet_list;
   Halfedge_list edge_list;
+  Halffacet_list facet_list;
 };
 
   typedef boost::container::deque<Node> Node_range;
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -402,12 +402,12 @@ public:
     return nodes_around_segment(s);
   }
   Node_list nodes_around_segment( const Segment_3& s) const {
-    Node_list nodes;
+    Node_list result;
     Objects_around_segment objects( *this, s);
     for(typename Objects_around_segment::Iterator oas = objects.begin(); oas != objects.end(); ++oas) {
-      nodes.push_back(oas.get_node());
+      result.push_back(oas.get_node());
     }
-    return nodes;
+    return result;
   }
   bool is_point_in_node( const Point_3& p, const Node_handle target) const {
     return is_point_in_node( p, target, root);
--- a/Nef_3/include/CGAL/Nef_3/K3_tree.h
+++ b/Nef_3/include/CGAL/Nef_3/K3_tree.h
@@ -108,15 +108,8 @@ typedef Smaller_than<
 public:
     typedef Node* Node_handle;
   Node(const Vertex_list& V, const Halfedge_list& E, const Halffacet_list& F) :
-    left_node(nullptr), right_node(nullptr)
+    left_node(nullptr), right_node(nullptr), vertex_list(V), edge_list(E), facet_list(F)
   {
-      object_list.reserve(V.size()+E.size()+F.size());
-      for(Vertex_const_iterator vi=V.begin(); vi!=V.end(); ++vi)
-          object_list.push_back(make_object(*vi));
-      for(Halfedge_const_iterator ei=E.begin(); ei!=E.end(); ++ei)
-          object_list.push_back(make_object(*ei));
-      for(Halffacet_const_iterator fi=F.begin(); fi!=F.end(); ++fi)
-          object_list.push_back(make_object(*fi));
   }
 
   Node(Node_handle l, Node_handle r, const Plane_3& pl) :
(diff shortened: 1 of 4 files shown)


GilesBathgate:
There is a further possible improvement which is that in `SM_point_locator` instead of storing `result` in an object it uses `            v_res = vi; solution = is_vertex_;` etc https://github.com/CGAL/cgal/blob/af913fd00feb4444ecf471d1ad0ae2486dea646f/Nef_S2/include/CGAL/Nef_S2/SM_point_locator.h#L249
The same can be done for `SNC_point_locator`

afabri:
Seems shorter, but a big diff.   Do you see a performance gain?

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -223,220 +230,203 @@ class SNC_point_locator_by_spatial_subdivision :
     return this->shoot(ray, null_handle, mask);
   }
 
+  enum SOLUTION { is_vertex_, is_edge_, is_facet_ };
```
afabri, line 233: We have a [`Feature_dimension` ](https://doc.cgal.org/latest/Kernel_23/structCGAL_1_1Feature__dimension.html) class.   Even if it is maybe not the perfect fit here, it might lead to  a better name than `SOLUTION`.  And note  that all uppercase is for constants. Types are just capitalized.

GilesBathgate: Ok I'll look into it. It was copied from `SM_point_locator`. I like to make things consistently inconsistent :joy:

GilesBathgate:
> Seems shorter, but a big diff. Do you see a performance gain?

I am measuring 1.95% in my benchmark. I am using the following as I thought it might reflect my 'real world' usage better:
```bash
#!/bin/bash

for i in {1..30}; do
./rapcad-baseline ~/cad/complex_intersection.rcad -o null 2>&1 | grep "Total compiling time:" | tee -a baseline.txt;
./rapcad-contender ~/cad/complex_intersection.rcad -o null 2>&1 | grep "Total compiling time:" | tee -a contender.txt;
done
```

complex_intersection.rcad is the spheregrid

afabri:
We use both the word spheregrid, but it might well be that we do not use the same data sets.  I generated two by rotating and shifting it, as the initial grid has many vertices with the same x and y coordinate and just a variation on z, and the same for the other directions. I feared that this has an impact on the usage of the K3_tree. It's more a gut-feeling argument, and maybe one should compare this.

GilesBathgate:
complex_intersection.rcad
```openscad
import </home/giles/projects/Nef_3_benchmark/data/spheregrid.off> as a;
import </home/giles/projects/Nef_3_benchmark/data/shiftedspheregrid.off> as b;

intersection(){
a();
b();
}
```
If you push different datasets to the Nef_3_benchmark, I can run the tests again.

afabri:
I get a segfault when I run it in this branch.

afabri:
> I get a segfault when I run it in this branch.

It's the third commit which makes it segfault (VC++ in Release mode).

Nef_3/include/CGAL/Nef_3/K3_tree.h
```diff
@@ -441,53 +394,23 @@ void divide_segment_by_plane( Segment_3 s, Plane_3 pl,
...
-    return Objects_along_ray( *this, r);
+  Node_list nodes_along_ray( const Ray_3& r) const {
+    Segment_3 s = ray_to_segment(r);
+    return nodes_around_segment(s);
   }
-  Object_list objects_around_segment( const Segment_3& s) const {
-    Object_list O;
-
+  Node_list nodes_around_segment( const Segment_3& s) const {
+    Node_list nodes;
```
afabri, line 405: Note that there is also a data member called `nodes`.

GilesBathgate:
> > I get a segfault when I run it in this branch.
> 
> It's the third commit which makes it segfault (VC++ in Release mode). Reading the code in the third commit it seems correct.

What test/program were you running when it segfaults, I'll try to reproduce.

GilesBathgate:
>[initialize solution](https://github.com/CGAL/cgal/pull/6480/commits/3aea3f4691f8fbaff1f0c35e9cb01a00bea1615e)

Thank you @afabri I had a feeling it might be something like that.

GilesBathgate:
> Seems shorter, but a big diff.

@afabri I could split this up into smaller changes as before. In its current state, it won't cleanly merge with #6470 So maybe it's better to leave this out of the test suite until the other changes are merged?

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -328,8 +328,8 @@ class SNC_point_locator_by_spatial_subdivision :
       case is_vertex_: return make_object(v_res);
       case is_edge_: return make_object(e_res);
       case is_facet_: return make_object(f_res);
+      case is_none_ : return Object_handle();;
```
GilesBathgate, line 331: You wrote double `;;` 
```c++
case is_none_ : return Object_handle();;
```

Also, I think something similar should be done in `locate()` which also uses the solution enum. It might be good to initialise to `_is_none` in locate.  It uses an If/else If , though so I expect no warning there.

afabri: Can you take care of these fixes. Thanks.

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
```diff
@@ -223,220 +230,203 @@ class SNC_point_locator_by_spatial_subdivision :
...
-          CGAL_error_msg( "wrong handle");
       }
       if(!hit)
-        ++objects_iterator;
+        ++nodes_iterator;
     }
+
     CGAL_NEF_TIMER(rs_t.stop());
-    return result;
+    switch (solution) {
```
afabri, line 327: Yesterday I added the fourth `case`   as we got a warning not to have tested all cases. That now leads to a warning that not all control paths return a value.  Shall I put a `break` in the last case and the `return` statement after the switch?   @lrineau @sloriot @mglisse

mglisse: The usual would be `default: CGAL_UNREACHABLE();`, or possibly another `CGAL_UNREACHABLE();` after `}`, depending on what the compiler is complaining about. But I am a bit surprised by the warning you mention about not all control paths returning a value.

afabri: The enum is in line 233 above and the switch lists them all.  And several compileres warn about it.

mglisse: > The enum is in line 233 above and the switch lists them all.

In C++, an enum can take values other than just those that were listed in its declaration. On the other hand, 4 values means exactly all values with 2 bits, so that should be the whole range.

GilesBathgate: `default: break;` could also work. It could also be a series of `if/else if` ....

Nef_3/include/CGAL/Nef_3/SNC_point_locator.h
GilesBathgate: I'll leave this suggestion for potential refactoring in the future.

PR #6485, opened 2022-04-08
do not use default constructed kernel

sloriot:
Follow-up of #6465


sloriot:
Successfully tested in CGAL-5.5-Ic-60

Issue #6479, opened 2022-04-06
Bug in "Split selected polyhedra"


palliez:
## Issue Details

in the polyhedron demo - I load a mesh (ply binary format) with many connected components - I would say few hundreds. 
alas the model is confidential (frigate) - I can only show a video privately. 

I then select the menu "Split selected polyhedra"

I get a crash.

the menu "color per component" works.

## Environment

* Operating system : Windows 10 64 bits
* Compiler: I use the precompiled version of the demo 5.3-I-162

palliez:
PS: the following code does work, hence it seems like this is an issue with rendering, VBOs buffers, etc

#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>

#include <CGAL/Surface_mesh.h>
#include <CGAL/Polygon_mesh_processing/connected_components.h>
#include <CGAL/Polygon_mesh_processing/IO/polygon_mesh_io.h>

#include <CGAL/Timer.h>
#include <CGAL/bounding_box.h>




// kernel
using Kernel = CGAL::Exact_predicates_inexact_constructions_kernel;
using FT = Kernel::FT;
using Point_3 = Kernel::Point_3;
using Plane_3 = Kernel::Plane_3;
using Vector_3 = Kernel::Vector_3;
using Segment_3 = Kernel::Segment_3;
using Triangle_3 = Kernel::Triangle_3;
using Iso_cuboid_3 = Kernel::Iso_cuboid_3;


// surface mesh
using Surface_mesh = CGAL::Surface_mesh<Point_3>;



// usage
void usage()
{
    std::cout << "Split mesh into components\n"
        "Usage: split -i input_mesh" <<
        "Reads OFF/STL/OBJ/PLY/TS/VTP files only. \n";
}


int main(int argc, char** argv)
{
    const int argc_check = argc - 1;
    char* mesh_file_name_ptr = nullptr;

    // parse parameters
    for (int i = 1; i < argc; ++i)
    {
        if (!strcmp("-h", argv[i]))
        {
            usage();
            return EXIT_FAILURE;
        }
        else if (!strcmp("-i", argv[i]) && i < argc_check) 
            mesh_file_name_ptr = argv[++i];
    }


    // read input mesh
    Surface_mesh input_mesh;
    std::string mesh_file_name(mesh_file_name_ptr);
    std::cout << "read mesh...";
[...]

PR #6489, opened 2022-04-13
Polyhedron_3: Faster split in the demo

afabri:
## Summary of Changes
Make the splitting of connected components faster.
First step: do not use a vector but a list, which avoids reallocation and swapping.

There are more copies made that we might want to avoid.    The surface mesh item gets constructed with a reference and makes a copy on the heap.  How to make the split function generate the mesh directly on the heap and then give the pointer to the surface mesh item.  Can one transfer ownership.  Also having a `swap()`  member function for the Surface_mesh would be helpful.

## Release Management

* Affected package(s):  Polyhedron
* Issue(s) solved (if any): improves #6479

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/PMP/Join_and_split_polyhedra_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/PMP/Join_and_split_polyhedra_plugin.cpp
@@ -144,11 +144,11 @@ void Polyhedron_demo_join_and_split_polyhedra_plugin::on_actionSplitPolyhedra_tr
     if(item)
     {
       QApplication::setOverrideCursor(Qt::WaitCursor);
-      std::vector<FaceGraph> new_polyhedra;
+      std::list<FaceGraph> new_polyhedra;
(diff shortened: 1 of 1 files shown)


lrineau:
The solution is a bit strange.

- Why `std:list` and not `std::deque`? Does the algorithm insert or erase inside the container?
- `Surface_mesh` has a copy-constructor. So it should also have a move-constructor, to avoid such issues with reallocations in a `std::vector`.

afabri:
I don't know what is strange about a list.  We put a whole mesh in there so the overhead of a a list is certainly not even measurable.  Moreover it offers a sort function that swaps list items and not entire meshes. The move constructor is for later, and maybe just do it.

lrineau:
> I don't know what is strange about a list. We put a whole mesh in there so the overhead of a a list is certainly not even measurable. Moreover it offers a sort function that swaps list items and not entire meshes. The move constructor is for later, and maybe just do it.

The `sort` member function is certainly a good reason for `std::list`. I had not realized that.

lrineau:
Why have you closed this PR?

afabri:
closed by accident.

sloriot:

PR #6492, opened 2022-04-15
Fix C3T3_facets_to_TM conversion for periodic meshes

MaelRL:
## Summary of Changes

The function `facets_in_complex_3_to_triangle_mesh()` uses a vertex handle-based unordered map to convert triangulation vertices to triangle mesh vertices. In the case of periodic meshes, triangulation vertices might correspond to different geometric positions depending on the periodic instance.

This PR changes the unordered map to be handle-based to point-based (in periodic triangulations, `Triangulation::point(Cell_handle, int)` returns a sensible position.

Some minor improvements while there (avoid the `to_double()`).

## Release Management

* Affected package(s): `Mesh_3`
* Issue(s) solved (if any): -


lrineau (approved):

sloriot:

Issue #5799, opened 2021-06-23
PMP Feature request: Add longest_edge()


afabri:
## Issue Details

Add a function that returns a halfedge descriptor for the longest edge of a mesh.   I am also surprised that we have a function computing the edge length and not the squared edge length.  We should not call sqrt for all edges for finding the longest one.

sloriot:
almost a oneliner with `std::max_element()` + but I agree it is nice to have.

PR #6496, opened 2022-04-18
Add PMP::longest_edge()

hoskillua:
## Summary of Changes
I have implemented the `longest_edge()` function requested in this [issue](https://github.com/CGAL/cgal/issues/5799). I have tested it on simple input and it works fine.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): implemented #5799


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -211,6 +211,54 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
   return squared_edge_length(halfedge(e, pmesh), pmesh, np);
 }
 
+/**
+  * \ingroup PMP_measure_grp
+  *
+  * Finds the longest edge of a given polygon mesh.
+  *
+  * @tparam PolygonMesh a model of `HalfedgeGraph`
```
sloriot, line 219:
```suggestion
  * @tparam PolygonMesh a model of `HalfedgeListGraph`
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -211,6 +211,54 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
...
+  */
+
+template<typename PolygonMesh,
+         typename NamedParameters = parameters::Default_named_parameters>
+inline typename boost::graph_traits<PolygonMesh>::edge_descriptor
+longest_edge(const PolygonMesh& pmesh, const NamedParameters& np = parameters::default_values()) {
+
+    auto edge_range = edges(pmesh);
+
+    CGAL_assertion(edge_range.size() > 0);
```
sloriot, line 254: `boost::graph_traits<PolygonMesh>::null_edge()` could be returned if the mesh is empty (testing `edge_range.begin()==edge_range.end()`)

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -211,6 +211,54 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
   return squared_edge_length(halfedge(e, pmesh), pmesh, np);
 }
 
+/**
+  * \ingroup PMP_measure_grp
+  *
+  * Finds the longest edge of a given polygon mesh.
```
sloriot, line 217:
```suggestion
  * returns the longest edge of a given polygon mesh.
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -211,6 +211,54 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
...
+  *
+  * @tparam PolygonMesh a model of `HalfedgeGraph`
+  * @tparam NamedParameters a sequence of \ref bgl_namedparameters "Named Parameters"
+  *
+  * @param pmesh the polygon mesh to which `h` belongs
```
sloriot, line 222: there is no `h` mentioned above. Copy-paste issue?

hoskillua: Indeed. Thank you for the correction.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
hoskillua: @sloriot  Thank you for the feedback. I am a bit confused. I assume you mean that I should replace the assertion with a condition `(edge_range.begin()==edge_range.end())` and return `boost::graph_traits<PolygonMesh>::null_edge()`. Is that right?
I also added a check on the returned variable from the `max_element()` before dereferencing it for safety.  I added these updates and your suggestions to new commit.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -211,6 +211,63 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
...
+longest_edge(const PolygonMesh& pmesh, const NamedParameters& np = parameters::default_values()) {
+
+    auto edge_range = edges(pmesh);
+
+    // if mesh has no edges
+    if (edge_range.begin() == edge_range.end())
+        return boost::graph_traits<PolygonMesh>::null_edge();
+
+    auto edge_reference = std::max_element(edge_range.begin(), edge_range.end(), [&pmesh, &np](auto l, auto r) {
+        return squared_edge_length(l, pmesh, np)
```
afabri, line 259: This should use [CompareDistance_3](https://doc.cgal.org/latest/Kernel_23/classKernel_1_1CompareDistance__3.html#a2ee64f4b6e1ac24eab6935efcf9d22ca)

afabri:
What is missing:   
A line in the Reference Manual [index page](https://github.com/CGAL/cgal/blob/master/Polygon_mesh_processing/doc/Polygon_mesh_processing/PackageDescription.txt#L222)
A line in the [change log](https://github.com/CGAL/cgal/blob/master/Installation/CHANGES.md)

hoskillua:
@afabri, I updated the Reference Manual and change log files. about testing, I added calls to the function in [measure_test.cpp](https://github.com/CGAL/cgal/pull/6496/files#diff-63e7d31cdc21d74d4d80bb85cece9a712a52f2f05afb4a6732299eaeaae6d486) and it ran as expected on these tests. But I can also make a testing program for this function if required.

sloriot:
There is a test failing [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-87/Polygon_mesh_processing/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz). Basically it says that the edge iterator of Polyhedron is not detected as a `ForwardIterator`. It might be a bug on our side with the definition of the edge iterator type.

janetournois:
Hello, what is the status of this PR?
I can't find the associated small feature.

sloriot:
@janetournois from what I remember there is a technical issue that I wasn't able to easily workaround at the time I looked at it. Let me have a fresh look.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -217,6 +217,75 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
...
+    using parameters::choose_parameter;
+    using parameters::get_parameter;
+
+    typename GetVertexPointMap<PolygonMesh, NamedParameters>::const_type
+        vpm = choose_parameter(get_parameter(np, internal_np::vertex_point),
+            get_const_property_map(CGAL::vertex_point, pmesh));
+
+    Geom_traits gt = choose_parameter<Geom_traits>(get_parameter(np, internal_np::geom_traits));
+
+    auto edge_range = edges(pmesh);
```
afabri, line 267: This should be `const&`  as a reference to an array is potentially returned.
Or we should  test   `edges(pmesh).empty()`

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -217,6 +217,75 @@ squared_edge_length(typename boost::graph_traits<PolygonMesh>::edge_descriptor e
...
+    auto edge_reference = std::max_element(edge_range.begin(), edge_range.end(), [&, vpm, pmesh](auto l, auto r) {
+        auto res = gt.compare_squared_distance_3_object()(
+            get(vpm, source((l), pmesh)),
+            get(vpm, target((l), pmesh)),
+            get(vpm, source((r), pmesh)),
+            get(vpm, target((r), pmesh)));
+        return res == SMALLER;
+    });
+
+    // if edge_reference is not derefrenceble
```
afabri, line 282: I  think that cannot happen as the range is not empty.

sloriot:

sloriot:
@MaelRL @afabri considering that https://github.com/CGAL/cgal/pull/8676 add minmax_dihedral_angle, shouldn't we change this PR to add `minmax_edge_length()`? (`minmax` comes from `std::minmax_element`)

PR #6501, opened 2022-04-20
Nef_3: Use sizeof(void*) in Generic handle map

GilesBathgate:
## Summary of Changes

I believe this is correct, no two addresses can be closer than `sizeof(void*)` apart, and a handle usually is that size, but if it's bigger then dividing by a smaller size is ok. This helps pack the handles into the hash map more tightly.

## Release Management


afabri:
I basically agree but I am wondering if we can go further.  Usually we divide by the size of the pointee. @sloriot any idea how to achieve that.

afabri:
Also, although it does not really matter technically, it is not so much a question of tightness but about a better distribution in the hash map, I think.

GilesBathgate:
> I basically agree but I am wondering if we can go further. Usually we divide by the size of the pointee. @sloriot any idea how to achieve that.

Because the map contains generic handles, I think the best you can do is divide by the smallest or you'll have overlapping addresses, but maybe that doesn't matter as it's just a collision? Perhaps the Generic hash map needs to have a `Types...` parameter and you can find the min/max sizeof at runtime? If you can divide by variable sizes, you can probably just pass that through to the hash functor.


> Also, although it does not really matter technically, it is not so much a question of tightness but about a better distribution in the hash map, I think.

Yeah, that's basically what I mean, the table index ( k % table_size) when table_size is 32, will never be odd, if the k is even, meaning only half the buckets will get filled.

I think I understand now why prime numbers are good for the table size.

GilesBathgate:
@afabri Here is a way:

```c++
struct Size_hash_function {
    std::size_t operator() (std::size_t h) const {
        return h;
    }
};

template <class I>
class Generic_handle_map : public
  Unique_hash_map<size_t,I,Size_hash_function>
{ typedef Unique_hash_map<size_t,I,Size_hash_function> Base;
public:
  Generic_handle_map() : Base() {}
  Generic_handle_map(I i) : Base(i) {}

  template <class H>
  const I& operator[](H h) const
  { return Base::operator[](std::size_t((void*)&*h)/sizeof(h)); }

  template <class H>
  I& operator[](H h)
  { return Base::operator[](std::size_t((void*)&*h)/sizeof(h)); }

};
```

GilesBathgate:
`Size_hash_function` can just be `std::identity`

GilesBathgate:
@afabri The trouble with this approach I think is that the hashmap doesn't check the entries for equality, only the "hashes" of the entries. This works for handle addresses because they are unique (hence Unique_hash_map) but it doesn't work if you divide by the handle size, when the handle sizes differ, because you can get overlapping values.

sloriot:
Sucessfully tested in CGAL-5.5-Ic-95

sloriot:
> I basically agree but I am wondering if we can go further. Usually we divide by the size of the pointee. @sloriot any idea how to achieve that.

So do you approve this PR and it can be integrated as is?

sloriot:
ah I see that https://github.com/CGAL/cgal/pull/6560 replaces it. @GilesBathgate  shall we close this PR ?

GilesBathgate:
@sloriot I am unsure about the right approach. I can close this draft if you want but I would rather leave it until the correct solution is merged.

PR #6502, opened 2022-04-21
prevent remove_self_intersections from emitting a warning 

sloriot:
(5.5 only as the refactor is introduced in 5.5)


sloriot:

PR #6461, opened 2022-03-29
Fast Box Intersection:  Add progress tracking

afabri:
## Summary of Changes

Introduce an API for tracking progress.  We report the progress in the traversal of the primary tree, that is we report 50% once we come back from the left subtree of the root of the segment tree.

The API is not yet documented, that is for the moment you have to look at the callback class in the example `progress.cpp`.

This PR also adds progress tracking to the Boolean operations based on corefinement.

## Release Management

* Affected package(s): Box_intersection_d, Polygon_mesh_processing
* Feature/Small Feature (if any):  tbd.
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership:  unchanged

PR #6506, opened 2022-04-21
New implementation for Has_member_*

lrineau:
## Summary of Changes

Followup to https://github.com/CGAL/cgal/pull/6461/commits/6260d5a4e47f891deb6cd8eed47d4b73a88ae884. See also https://github.com/CGAL/cgal/pull/6461#discussion_r839550495 and following comments.

## Release Management

### DIFF
--- a/BGL/test/BGL/test_Has_member_clear.cpp
+++ b/BGL/test/BGL/test_Has_member_clear.cpp
@@ -11,7 +11,7 @@ struct with_clear_but_args {
   void clear(int) {}
 };
 
-struct with_clear_but_const {
+struct with_clear_const {
   void clear() const {}
 };
 
@@ -25,7 +25,7 @@ int main()
 
   CGAL_static_assertion(!Has_member_clear<with_clear_but_args>::value);
 
-  CGAL_static_assertion(!Has_member_clear<with_clear_but_const>::value);
+  CGAL_static_assertion(Has_member_clear<with_clear_const>::value);
 
   return 0;
 }
--- a/BGL/include/CGAL/boost/graph/internal/Has_member_clear.h
+++ b/BGL/include/CGAL/boost/graph/internal/Has_member_clear.h
@@ -19,11 +19,8 @@ template<class T>
 class Has_member_clear
 {
 private:
-  template<class U, U>
-  class check {};
-
-  template<class C>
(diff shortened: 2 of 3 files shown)


MaelRL:
Couldn't https://github.com/CGAL/cgal/blob/master/STL_Extension/include/CGAL/Has_member.h be updated, and both of these use `CGAL_GENERATE_MEMBER_DETECTOR` ?

lrineau:
`CGAL_GENERATE_MEMBER_DETECTOR(X)` only detects that, for a class `C`, then there exists a `C::X` member, being whatever type of member (data member, member function). It cannot use expression-SFINAE, and cannot be made more robust.

`Has_member_clear`, and `Has_member_id` check not only that `::clear` and `::id` exist, but also that they are member functions, or data member functors, with a compatible signature.

MaelRL:
I meant that we only use `CGAL_GENERATE_MEMBER_DETECTOR` to find specific member functions whose signature is known (increment_erase_counter, size, resize, reserve, ...) so if this newer version is better, it could possibly replace the macro?

sloriot:

BGL/test/BGL/test_Has_member_clear.cpp
```diff
@@ -25,7 +25,7 @@ int main()
 
   CGAL_static_assertion(!Has_member_clear<with_clear_but_args>::value);
 
-  CGAL_static_assertion(!Has_member_clear<with_clear_but_const>::value);
```
sloriot, line 28: actually I don't understand that change

lrineau: The previous implementation of `Has_member_clean` was testing a specific signature: `void(C::*)(void)` (and not `void(C::*)(void) const`). The new implementation checks if the expression `c.clear()` is valid, and it can detect a non-const or a const member function. That is a difference in behavior. I hope that does not really matter.

... a method `void clear() const` would be strange, anyway: it is supposed to modify the object (by clearing it).

sloriot: Then, it does not really make sense to keep the line as this is not something that we want to enforce.

lrineau: As you want. In my opinion it does not really matter.

PR #6507, opened 2022-04-21
Ridges_3: Added definitions for boost program options

oesi82:
## Summary of Changes

Enable program options if they are found.

## Release Management

* Affected package(s): Ridges_3


sloriot:

PR #6510, opened 2022-04-22
Fixed degenerated case of plane fitting to a linear set of points.

### LINKED ISSUE #6146: linear_least_squares_fitting_3 may return NAN
## Issue Details

I use the cgal function "linear_least_squares_fitting_3" to fit a plane, I found the return value may be "NAN";

My code is like this:
```
Plane plane;
FT conf = linear_least_squares_fitting_3(points.begin(), points.end(), plane, CGAL::Dimension_tag<0>());
```
the value "conf" can be "NaN" or "2", whereas it should be within `[0,1]`.

Please take a look at the source code:

https://github.com/CGAL/cgal/blob/4321e776d0eb2ac6b2581ff06445aebca6e06a81/Principal_component_analysis/include/CGAL/PCA_util.h#L712-L727

If the `eigen_values[0]`, `eigen_values[1]`  and `eigen_values[2]` are all `0`, it will return zero; otherwise it will return `1 - eigen_values[0] / eigen_values[1]`. But there is a case, `eigen_values[1]` is `0`, while one of the others are not; in this case, the function can return "nan'. 

and also, `eigen_values[0] / eigen_values[1]` could `-1`, thus the return value will be `2`.

I think some degenerate cases are not taken into consideration, such as we fitting a plane for just two points.

Please fix it

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux
* Compiler: gcc
* CGAL version: 5.3

oesi82:
## Summary of Changes

Added a check for degenerated input data in fitting_plane_3 used by linear_least_squares_fitting_3 to avoid NaN as a return value.

## Release Management

* Affected package(s): Principal Component Analysis
* Issue(s) solved (if any): fix #6146

### DIFF
--- a/Principal_component_analysis/include/CGAL/PCA_util.h
+++ b/Principal_component_analysis/include/CGAL/PCA_util.h
@@ -723,7 +723,11 @@ fitting_plane_3(typename DiagonalizeTraits::Covariance_matrix& covariance, // co
                   eigen_vectors[1],
                   eigen_vectors[2]);
     plane = Plane(c,normal);
-    return FT(1) - eigen_values[0] / eigen_values[1];
+
+    if (eigen_values[1] == 0)
+        return FT(0); // line case
+    else
+        return FT(1) - eigen_values[0] / eigen_values[1]; // regular case
   } // end regular case
 }
 


sloriot:
This could probably go into 5.3

PR #6518, opened 2022-04-25
Correcting incorrect link in bibliography file

albert-github:
In the closed issue #3819 there WAS already A suggestions to fix this problem (https://github.com/CGAL/cgal/issues/3819#issuecomment-479892766), but it had not been implemented.
It is possible to use https://haltools.inria.fr/Public/afficheRequetePubli.php?struct=prisme&affi_exp=inria&format_export=xml&langue=Anglais&CB_accent_latex=oui&CB_auteur=oui&CB_titre=oui&CB_article=oui&Fen=Rech but some further searching is necessary (e.g. for `Olivier Devillers, Giuseppe Liotta,` ) so better to use https://dx.doi.org/10.1016/S0925-7721(98)00039-X

Other references in the bibliography were corrected in a similar way.


sloriot:

PR #6525, opened 2022-04-26
CGAL::Mesh_3::polylines_to_protect with angle bound - bug fix

janetournois:
## Summary of Changes

`min(90, angle_bound)` should be `max(90, angle_bound)`.
90 is a conservative lower bound. This means that each vertex corresponding to an angle smaller than 90 should be set as terminal by `CGAL::Mesh_3::polylines_to_protect()`, even if `angle_bound < 90`.
If `angle_bound > 90`, we keep `angle_bound` and collect/set more terminal vertices.

This PR fixes PR #6402

## Release Management

* Affected package(s): Mesh_3


sloriot:

lrineau (approved):

PR #6532, opened 2022-04-27
Workaround deprecation warnings in Barycentric Coordinates

afabri:
## Summary of Changes

We have two platforms which give us unexpected deprecated warnings.  Workaround them...


sloriot:
Successfully tested in CGAL-5.5-Ic-83 (warnings are gone)

PR #6533, opened 2022-04-28
Barycentric coordinates: bug fix: Increment output iterator 

afabri:
## Summary of Changes

After the assignment to an output iterator it  must be incremented. 

## Release Management

* Affected package(s): Barycentric_coordinates


afabri:
I am also wondering if [here ](https://github.com/CGAL/cgal/blob/master/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Discrete_harmonic_2.h#L128) we better returned `boost::none`.  @sloriot any idea?

sloriot:
> I am also wondering if [here ](https://github.com/CGAL/cgal/blob/master/Barycentric_coordinates_2/include/CGAL/Barycentric_coordinates_2/Discrete_harmonic_2.h#L128) we better returned `boost::none`. @sloriot any idea?
That is identical and won't fix the maybe unintialized warning if this is what you have in mind.

PR #6535, opened 2022-04-28
remove mentions to  CMAKE_GENERATOR in test scripts

sloriot:
Fix recent issue in the testsuite


sloriot:

PR #6553, opened 2022-05-04
Do not set cxx standard in the final target cmake script

sloriot:
Fix recent errors due to a bug in GCC

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105436
[error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-75/Weights/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz)


sloriot:

PR #6558, opened 2022-05-05
CMake script updates: version of CMake and /wd4503 is only for MSVC 2015

lrineau:
## Summary of Changes

- Condition `/wd4503` to MSVC version 2015 only
- Simplify the CMake script by assuming CMake>=3.11 (3.14 is documented), and the maximal supported version is CMake 3.23

## Release Management


lrineau:
I think this PR should either be in 5.5-beta1, or postponed to 5.6-beta.

lrineau:
@sloriot Note that commit https://github.com/CGAL/cgal/pull/6558/commits/22882254483d6a8e43138d2b9c4d4363d92647e3 is a massive update to:
```cmake
cmake_minimum_required(VERSION 3.1...3.23)
```
for all CMake files. Previously it was most commonly
```cmake
cmake_minimum_required(VERSION 3.1...3.22)
```
and sometimes other versions.

sloriot:
Successfully tested in CGAL-5.5-Ic-83

PR #6562, opened 2022-05-06
Fixed Eigen3 3.4.90 deprecation warnings

soesau:
## Summary of Changes

Moved parameters from compute method to class template.

## Release Management

* Affected package(s): Surface_mesh_deformation

### DIFF
--- a/Solver_interface/include/CGAL/Eigen_svd.h
+++ b/Solver_interface/include/CGAL/Eigen_svd.h
@@ -51,7 +51,11 @@ public:
   /// \return the condition number of \f$ M\f$
   static FT solve(const Matrix& M, Vector& B)
   {
-    Eigen::JacobiSVD<Matrix::EigenType> jacobiSvd(M.eigen_object(),::Eigen::ComputeThinU | ::Eigen::ComputeThinV);
+#if EIGEN_VERSION_AT_LEAST(3,4,90)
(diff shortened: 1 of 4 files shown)


sloriot:
Did you check that the change works with the first supported version of Eigen? 
See [here](https://doc.cgal.org/latest/Manual/thirdparty.html#thirdpartyEigen). If needed the minimal version could be updated but then cmake scripts will need an update too.

I guess this branch should be based on cgal/5.4-x-branch. See [here](https://github.com/CGAL/cgal/wiki/Source-Code-Management-with-Git#rebase-an-existing-bug-fix-branch) for how to do it.

soesau:
I compiled the examples and tests with both versions, i.e., 3.4.0 and 3.4.90.

sloriot:
Actually 3.3.4 is the minimal version that we are testing. We should update the requirements but then the PR can only target 5.5. @lrineau do you agree?

soesau:
I made a mistake. I compiled the tests, but I did not execute them.
The options for solving the SVD were moved into the template parameter with version 3.4.90. This only works for 3.4.90, but not for 3.4.0 and before. Thus, we can suppress the warnings by using
#ifdef EIGEN_VERSION_AT_LEAST(3,4,90)...

sloriot:

PR #6570, opened 2022-05-09
Convex_hull_3: Use K::Boolean and not bool

afabri:
## Summary of Changes

As the predicates are used with a kernel that may return an `Uncertain`  we must use `K::Boolean` instead of `bool`.
This fixes the 'N' we can see in the VC2015 [Convex_hull_3 ](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.5-I-79.shtml#Convex_hull_3) testsuite. 

## Release Management


sloriot:
Nice!

sloriot:
Successfully tested in CGAL-5.5-Ic-83 (error is gone)

PR #6575, opened 2022-05-12
PLY: Support writing per-face/vertex property lists

nh2:
## Summary of Changes

Hi, I'd like to contribute the writing of PLY property lists.

This allows to directly write multi-texture PLY files with CGAL that can be opened directly in e.g. MeshLab, as well as other use cases that require `property list`.

## Release Management

* Affected package(s):

### DIFF
--- a/Stream_support/include/CGAL/IO/PLY/PLY_writer.h
+++ b/Stream_support/include/CGAL/IO/PLY/PLY_writer.h
@@ -298,6 +298,42 @@ public:
   }
 };
 
+
+template <typename Index,
+          typename PropertyMap,
+          typename VectorType = typename boost::property_traits<PropertyMap>::value_type,
+          typename ElementType = typename VectorType::value_type>
+class Simple_property_vector_printer
+  : public Abstract_property_printer<Index>
+{
+  PropertyMap m_pmap;
+public:
+  Simple_property_vector_printer(const PropertyMap& pmap) : m_pmap(pmap) { }
+
+  virtual void print(std::ostream& stream, const Index& index)
+  {
+    const VectorType& vec = get(m_pmap, index);
+    if(get_mode(stream) == CGAL::IO::ASCII)
+    {
+      stream << vec.size();
+      for(const ElementType& v : vec)
+      {
+        stream << " " << v;
+      }
+    }
+    else
+    {
+      unsigned char size = (unsigned char)(vec.size());
+      stream.write(reinterpret_cast<char*>(&size), sizeof(size));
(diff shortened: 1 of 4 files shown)


Stream_support/include/CGAL/IO/PLY/PLY_writer.h
```diff
@@ -298,6 +298,42 @@ class Char_property_printer
...
+    {
+      stream << vec.size();
+      for(const ElementType& v : vec)
+      {
+        stream << " " << v;
+      }
+    }
+    else
+    {
+      unsigned char size = (unsigned char)(vec.size());
```
nh2, line 326: I don't like this potentially-overflowing cast.

I copied it from existing code that uses the same logic:

https://github.com/CGAL/cgal/blob/acbd1ee182211c6309f355925927802f036fbe13/Surface_mesh/include/CGAL/Surface_mesh/IO/PLY.h#L1019

Should this be done better in both places, e.g. throwing an exception if the vector size does not fit `uchar`?

lrineau:
There should be a test .cpp file using the feature.

nh2:
@lrineau Could you point me at a similar test to get some inspiration from, and how it is added to the test suite? Thanks1

afabri:
Hello @nh2, 
Each package has a subdirectory `test`. And this one has a subdirectory `data`.   So it should go [here](https://github.com/CGAL/cgal/tree/master/Stream_support/test/Stream_support).

Stream_support/include/CGAL/IO/PLY/PLY_writer.h
sloriot: There should indeed be a check that `polygon.size()<=std::numeric_limits<unsigned char>::max())`, skipping this property if that's the case.

sloriot:

lrineau:
What is the status of this pull-request. As far as I can see, it seems waiting for tests. See https://github.com/CGAL/cgal/pull/6575#issuecomment-1125875273 and following message.

sloriot:
Exactly, I did some refactoring after the initial patch but the main remaining point is testing the new feature added.

Stream_support/include/CGAL/IO/PLY/PLY_writer.h
sloriot: There is actually no "overflow" problem here as the size will simply be modulo `sizeof(uchar)`. Which means that less elements will be put.

sloriot:

PR #6577, opened 2022-05-12
add an example to simplify a mesh in parallel

sloriot:
instead of having it somewhere locally, it is better shipped with CGAL...


afabri:
You should add it also in  examples.txt in the documentation

sloriot:

PR #6586, opened 2022-05-15
Arrangement_on_surface_2: unknown attribute compact specified

albert-github:
The `<OL>` tag has no attribute `compact`.
- only the HTML tag `<menu>` has the `compact` attribute according to https://www.w3schools.com/tags/att_menu_compact.asp but it is not supported in any major browser.
- according to https://www.geeksforgeeks.org/html-ol-compact-attribute/ the `<OL>` tag can have a `compact` attribute but it is not supported by any any major browser.


sloriot:

PR #6588, opened 2022-05-16
PMP: Fast clipping with a plane

afabri:
## Summary of Changes

When we clip with a plane no need for fast box intersection.

This is work in progress as there is no clean API.   I even break the code, as we encode the plane as a mesh with just one triangle. 

## Release Management

* Affected package(s): PMP


afabri:
@sloriot the coref plugin which fails in the CI compiles fine with VC++

PR #6609, opened 2022-05-24
Polygon: Add move semantics

afabri:
## Summary of Changes

Use move semantics for `Polygon_2`  and `Polygon_with_holes_2`.   
The latter gets passed a polygon as outer boundary and an iterator range of polygons for the holes.  

## Release Management

* Affected package(s):  Polygon_2

### DIFF
--- a/Polygon/include/CGAL/General_polygon_with_holes_2.h
+++ b/Polygon/include/CGAL/General_polygon_with_holes_2.h
@@ -127,10 +127,10 @@ protected:
 //                          operator<<
 //-----------------------------------------------------------------------//
 /*!
-This operator exports a General_polygon_with_holes_2 to the output stream `out`.
+This operator exports a `General_polygon_with_holes_2` to the output stream `os`.
 
 An \ascii and a binary format exist. The format can be selected with
-the \cgal modifiers for streams, `set_ascii_mode(0` and `set_binary_mode()`
+the \cgal modifiers for streams, `set_ascii_mode()` and `set_binary_mode()`,
 respectively. The modifier `set_pretty_mode()` can be used to allow for (a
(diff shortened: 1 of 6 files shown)


pentacular:
I think that you cannot assume that the holes can be std::move'd on the basis of how the boundary was passed, since std::move will destroy the holes in the caller, and these may have different lifetimes.

afabri:
> I think that you cannot assume that the holes can be std::move'd on the basis of how the boundary was passed, since std::move will destroy the holes in the caller, and these may have different lifetimes.

What are the options.
- Document that this is what we do. If you don't want to move holes, simply do not move the outer boundary and everything gets copied.
- Add overloads which move the boundary one but not the others, and even not the boundary but the holes. 
- Move would not work with a  `const`  boundary polygon, We could move the holes in case they are not const.  That could lead to nasty to debug user code as it would be rather implicit.

pentacular:
> > I think that you cannot assume that the holes can be std::move'd on the basis of how the boundary was passed, since std::move will destroy the holes in the caller, and these may have different lifetimes.
> 
> What are the options.
> 
> * Document that this is what we do. If you don't want to move holes, simply do not move the outer boundary and everything gets copied.
> * Add overloads which move the boundary one but not the others, and even not the boundary but the holes.
> * Move would not work with a  `const`  boundary polygon, We could move the holes in case they are not const.  That could lead to nasty to debug user code as it would be rather implicit.

Well, moving the holes requires a move_iterator -- how about only supporting that where the boundary is also moved?

sloriot:
The bug of the doc comes from the post-processing. I'll fix it.

sloriot:

sloriot:

lrineau (approved):

sloriot:

PR #6611, opened 2022-05-24
AW3: Fix overzealous filtering during traversal (+ misc enhancements)

MaelRL:
## Summary of Changes

* Bug fix: AW3 uses outside/inside info to avoid some do_intersect calls during traversal. This can be done between datum and query, because we know an outside cell does not intersect any datum, but cannot be done between a query and a node's bounding box because the cell - even if outside - can have intersections with the node's bounding box that signal that a lower level of the tree must be investigated. Bug could be easily reproduced with sparse point sets.
* Add segments/points to the demo + possibility to select which simplex type to wrap
* Simplify (internal) API for oracles: no need to template the oracle by the range type.
* More minor stuff

## Release Management

* Issue(s) solved (if any): -


sloriot:
New warnings in 5.5-Ic-91

sloriot:

PR #6612, opened 2022-05-24
Please intel 2019 compiler

sloriot:
I could add a `#ifdef __INTEL_COMPILER` but it does not hurt as is.


sloriot:

PR #6616, opened 2022-05-25
Workaround issue with pyquery

sloriot:
Fix issue like [here](https://doc.cgal.org/latest/Polygon/classCGAL_1_1Polygon__2.html#aacf55ebc4ccb144496b1aae6b867ef5e)


sloriot:
Successfully tested in CGAL-5.5-Ic-92 (I checked and the pb is gone)

PR #6622, opened 2022-05-27
initialize some stack variables

kda:
## Summary of Changes

Initialize a few stack variables.  Using clang memory sanitizer, they were detected as uninitialized in some cases.

These changes should result in no functional change (NFC).

### DIFF
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -538,7 +538,7 @@ insert(const Point &p, Cell_handle start)
 {
   int vertex_level = random_level();
   Locate_type lt;
-  int i, j;
+  int i = 0, j = 0;
   // locate using hierarchy
   locs positions[maxlevel];
   locate(p, lt, i, j, positions, start);
@@ -578,7 +578,7 @@ insert_and_give_new_cells(const Point &p, OutputItCells fit, Cell_handle start)
   Locate_type lt;
   int i, j;
   // locate using hierarchy
-  locs positions[maxlevel];
+  locs positions[maxlevel] = {0};
   locate(p, lt, i, j, positions, start);
   // insert at level 0
   Vertex_handle vertex = hierarchy[0]->insert_and_give_new_cells(p,
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -538,7 +538,7 @@ insert(const Point &p, Cell_handle start)
 {
   int vertex_level = random_level();
   Locate_type lt;
-  int i = 0, j = 0;
+  int i = -1, j = -1;
   // locate using hierarchy
   locs positions[maxlevel];
   locate(p, lt, i, j, positions, start);
@@ -578,7 +578,7 @@ insert_and_give_new_cells(const Point &p, OutputItCells fit, Cell_handle start)
   Locate_type lt;
   int i, j;
   // locate using hierarchy
-  locs positions[maxlevel] = {0};
+  locs positions[maxlevel] = {-1};
   locate(p, lt, i, j, positions, start);
   // insert at level 0
   Vertex_handle vertex = hierarchy[0]->insert_and_give_new_cells(p,
--- a/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
@@ -439,7 +439,7 @@ protected:
 
   struct locs {
       Cell_handle pos;
-      int li, lj;
+      int li = -1, lj = -1;
       Locate_type lt;
   };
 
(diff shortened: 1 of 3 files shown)


mglisse:
Can you explain in which case the variables are used uninitialized, and assuming such a case exists, why using the arbitrary value 0 would be right? Is the *use* simply the copy that happens when passing `li` by value to `insert` (and passing it by const reference would silence the warning)? `valgrind` is clever enough to wait until a "real" use to warn, I don't know if the memory sanitizer has such a feature.
If there is a bug, I am all for fixing it. But if there isn't, an uninitialized variable is preferable to one initialized with an arbitrary value, it gives good tools a better chance to detect bugs and should not be sacrificed for bad tools.

GilesBathgate:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-always

kda:
Marc,

clang sanitiize individually tracks each location and does not through
until the variable is read (when not initialized).
No known bug, and I believe that initializing is safer than not.  If a
default value other than zero is better, I'm happy to replace.

Giles,

Thank you for the support.

kda:
spelling: through -> throw

mglisse:
(github lets you edit your own comments if you have a typo, click on "..." in the top right corner of the comment)

> clang sanitize individually tracks each location and does not throw until the variable is read (when not initialized).

Yes, but I am asking what it considers a *read*. As I said, valgrind has a rather sophisticated notion so that a plain copy does not count. https://github.com/google/sanitizers/wiki/MemorySanitizer seems to claim that MemorySanitizer does as well, so it would be useful to study the full report (with `-fsanitize-memory-track-origins=2`) to check if there is actually a real issue, that's the whole point of this sanitizer after all. Only then does it make sense to discuss patches.

> No known bug,

Good :smiley: 

> and I believe that initializing is safer than not.

In what way? If there is no bug currently, the initialization just adds some redundant operations. The main consequence is that if we ever modify the code in a broken way that starts using uninitialized values, with the current code good tools can tell us about it, while with initialization to 0 they have no chance to notice, and we get hard to debug wrong results.

The supposed *safety* of initializing everything is usually stated in the context where someone would try to hack your computer (for instance by giving you a specially crafted set of points to triangulate), that exploits a bug in your code, and then leaving some variables uninitialized may give the hacker more options on how to make the code jump to a place they control. So initializing can be a good mitigation for the final production build, but still shouldn't be done most of the time because detecting (and fixing) the bug is even better. So this useless initialization is best done automatically by the compiler when some option is used (https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Optimize-Options.html#index-ftrivial-auto-var-init), or if the compiler doesn't offer this, based on some macro like CGAL_INITIALIZE_ALL so we can easily disable it.

In our particular case, we call locate, which takes li and lj by reference, and the guideline that got cited says this counts as an initialization, for weak static analysis purposes. The compiler (after inlining) or the runtime can see that li and lj are not always initialized, but doesn't go as far as noticing that they are only used in cases where they are initialized :disappointed: . If we initialize `li` to a value not between 0 and 3, then at least we have a chance to detect a potential wrong use in debug mode with https://github.com/CGAL/cgal/blob/ebf9fb025a5e1b3c6922071a8000ea456826e81c/TDS_3/include/CGAL/Triangulation_ds_cell_base_3.h#L69 , although we might still miss it if we only use it as `(li+1)%4` or some other code that doesn't just access the vertex.

kda:
Alright.  It seems the way forward here is for me to figure out how to run
the CGAL unit tests and turn on the sanitizer and see what shows up.
It may be necessary to add a few tests to demonstrate the problem I am
seeing.

Unfortunately, the cases where these are popping up are within a large
external code base.

mglisse:
@kda were you also using the new `-fsanitize-memory-param-retval`?

lrineau:
Hi @kda, we discussed among GeometryFactory's developers and decided that we can allow the initialization of those variable to the value `-1` (that is invalid and will trigger assertions latter) instead of `0`. Please modify your pull-request accordingly.

Triangulation_3/include/CGAL/Triangulation_hierarchy_3.h
```diff
@@ -578,7 +578,7 @@ insert_and_give_new_cells(const Point &p, OutputItCells fit, Cell_handle start)
   Locate_type lt;
   int i, j;
   // locate using hierarchy
-  locs positions[maxlevel];
+  locs positions[maxlevel] = {-1};
```
mglisse, line 581: Note that this initializes the first element to -1, and the others to 0, IIRC.

kda: Thank you for allowing this to roll forward.

-1 does not work anyway, as 'locs' is a struct. (line 440).
The alternative here would be to default the values in the struct.

I did a quick test for the failure in my codebase and found that initializing the members li and lj with -1, was sufficient to address the issue.

GilesBathgate: I think `locs positions[maxlevel] = {-1,-1,-1,-1};` would have worked (Struct initialization), but the  solution you proposed is better.

afabri:
@lrineau @sloriot, and @MaelRL , do you agree that we can put this in the testsuite, and even rebase for 5.4, together with PR #6631  where the initialization should also be changed to -1.

lrineau (approved):

lrineau:
> @lrineau @sloriot, and @MaelRL , do you agree that we can put this in the testsuite, and even rebase for 5.4, together with PR #6631 where the initialization should also be changed to -1.

Yes I agree.

sloriot:

kda:
yes.

Issue #6579, opened 2022-05-13
Linear_cell_complex:  Failure in a testsuite


afabri:
## Issue Details
In this [VC2015 testuite ](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.5-Ic-83.shtml#Linear_cell_complex) we get an assertion.  
```
------------------------------------------------------------------
- Linear_cell_complex/ProgramOutput.execution___of__Linear_cell_complex_copy_test
------------------------------------------------------------------

CGAL::Random()::get_seed() = 1652389513
Linear cell complex copy test (v1).Assertion failed: false, file C:\CGAL_ROOT\CGAL-5.5-Ic-83\test\Linear_cell_complex\Linear_cell_complex_copy_test.cpp, line 83
```
I could not reproduce it locally (in the master branch), also not by inserting 
```
CGAL::get_default_random() = CGAL::Random(1652389513);
```
as first line in `main()`.    This is maybe not enough as at the same time the function `rand()` is used. 

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 64bit
* Compiler:  VC2015
* Release or debug mode: Debug
* Specific flags used (if any):
* CGAL version:
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

sloriot:
Also [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-22/Linear_cell_complex/TestReport_Sosno_MSVC-2022-Community-Release.gz) and [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5-Ic-32/Linear_cell_complex/TestReport_Sosno_MSVC-2022-Community-Release.gz).

gdamiand:
Thanks. I will look.

afabri:
The first step might be to replace the call of `rand()`   and wait for the next failure,

PR #6627, opened 2022-05-31
Linear_cell_complex: Replace call to rand() in one testcase

afabri:
## Summary of Changes

Make the testsuite reproducible by replacing the call to `rand()`.

## Release Management

* Affected package(s): Linear_cell_complex
* Issue(s) solved (if any): related to #6579

### DIFF
--- a/Linear_cell_complex/test/Linear_cell_complex/Linear_cell_complex_copy_test.cpp
+++ b/Linear_cell_complex/test/Linear_cell_complex/Linear_cell_complex_copy_test.cpp
@@ -93,7 +93,7 @@ public:
 
 struct MonInfo
 {
-  MonInfo(long long int i=0) : mnb(i==0?rand():i),
+  MonInfo(long long int i=0) : mnb(i==0?CGAL::get_default_random().get_int(0,RAND_MAX):i),
                                ptr(reinterpret_cast<char*>(this))
   {}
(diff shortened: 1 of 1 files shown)


gdamiand (commented):
Thanks @afabri, sorry for the delay to solve this issue.
My plan is to solve this problem with the new branch cmap with index since there are many modifications.

lrineau:
How does this PR fixes the issue? I do not see why `CGAL::get_default_random().get_int(0,RAND_MAX)` is better than `rand()`.

afabri:
It is to make it seedable so the next time we see when it crashes we can reproduce.

gdamiand:
I plan to solve the issue later, with the branch cmap index.

sloriot:
I think you can see the error in [CGAL-5.6-Ic-1](https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-1/Linear_cell_complex/TestReport_Christo_MSVC2015-Release-64bits.gz)

sloriot:
Replaced by #6688. @gdamiand please confirm that you have handled this issue.

gdamiand:
> Replaced by #6688. @gdamiand please confirm that you have handled this issue.

I think so, but we will see the results of the test suite.

Issue #6582, opened 2022-05-15
AABB_tree: Incorrect documentation


albert-github:
Doing some extra tests on the CGAL documentation (setting `GENERATE_XML=YES`), the following warning appeared:
```
.../AABB_tree/include/CGAL/AABB_tree.h:565: warning: argument 'last' of command @param is not found in the argument list of CGAL::AABB_tree< AABBTraits >::expand(Node &node, ConstPrimitiveIterator first, ConstPrimitiveIterator beyond, const std::size_t range, const ComputeBbox &compute_bbox, const SplitPrimitives &split_primitives, const AABBTraits &)
.../AABB_tree/include/CGAL/AABB_tree.h:565: warning: The following parameters of CGAL::AABB_tree::expand(Node &node, ConstPrimitiveIterator first, ConstPrimitiveIterator beyond, const std::size_t range, const ComputeBbox &compute_bbox, const SplitPrimitives &split_primitives, const AABBTraits &) are not documented:
  parameter 'node'
  parameter 'beyond'
  parameter 'compute_bbox'
  parameter 'split_primitives'
```

Even though the function is not part of the official documentation the documentation of the function should be correct.

albert-github:
To be honest I wouldn't as I think **all* functions should be documented properly to show their intention for future maintainers.
The documentation in the code is not just for the outside world but also for maintainers / people who want to extend code.

afabri:
I added the parameters in PR #6629, but just state that it are two functors.  While it is true that documenting also all private functions would be great this is currently not a realistic option.

MaelRL:
Fixed by https://github.com/CGAL/cgal/pull/6629.

PR #6629, opened 2022-05-31
AABB_tree: Document parameters

afabri:
## Summary of Changes

Document the parameter `node` as well as two parameters as functor (without saying more) and remove the parameter `traits` as it is not used.

## Release Management

* Issues:  #6582


MaelRL:
`node` is still undocumented

sloriot:

PR #6631, opened 2022-05-31
Fix a potential uninitialized memory error

sgayda2:
## Summary of Changes

Fixes an msan error

==1142==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f378950d748 in CGAL::Alpha_wraps_3::internal::Alpha_wrap_3<CGAL::Alpha_wraps_3::internal::Triangle_mesh_oracle<CGAL::Surface_mesh<CGAL::Point_3<CGAL::Epick> >, CGAL::Named_function_parameters<bool, CGAL::internal_np::all_default_t, CGAL::internal_np::No_property>, true, int>, CGAL::Surface_mesh<CGAL::Point_3<CGAL::Epick> >, CGAL::Surface_mesh<CGAL::Point_3<CGAL::Epick> >::Property_map<CGAL::SM_Vertex_index, CGAL::Point_3<CGAL::Epick> >, CGAL::Epick>::alpha_flood_fill()

## Release Management

* Affected package(s): Alpha_wrap_3

### DIFF
--- a/Alpha_wrap_3/include/CGAL/Alpha_wrap_3/internal/Alpha_wrap_3.h
+++ b/Alpha_wrap_3/include/CGAL/Alpha_wrap_3/internal/Alpha_wrap_3.h
@@ -1050,7 +1050,7 @@ private:
         CGAL_assertion(CGAL::abs(CGAL::approximate_sqrt(m_oracle.squared_distance(steiner_point)) - m_offset) <= 1e-2 * m_offset);
 
         // locate cells that are going to be destroyed and remove their facet from the queue
-        int li, lj = 0;
+        int li = 0, lj = 0;
         Locate_type lt;
         const Cell_handle conflict_cell = m_dt.locate(steiner_point, lt, li, lj, neighbor);
         CGAL_assertion(lt != Dt::VERTEX);


afabri:
@sgayda2 can you provide a minimal example. When I compile with clang  with -fsanitize=memory in release mode and execute the examples we distribute in CGAL I don't get the message you report.

mglisse:
> This was using -fsanitize-memory-param-retval with the in progress release version of clang 15.

Ah, so you were explicitly passing an option that asks to give an error even if the uninitialized value is never actually used. Is there somewhere online a good discussion about why this option was created? It seems to defeat the cleverness that was put in -fsanitize=memory :disappointed: I guess some places have a policy never to pass uninitialized values (by value) to functions, that's not absurd, but it goes against the design of locate+insert in CGAL, and if we wanted to satisfy this policy, the right fix would be to change their interface...
I am still in favor of not changing the code (so the clever `-fsanitize=memory` can properly reason about it), but now I know about -fsanitize-memory-param-retval, I won't oppose initializing with a clearly invalid value (and possibly adding an assertion that we don't use that invalid value, unless it is redundant because we are already doing an array access that can be checked in debug mode).

Alpha_wrap_3/include/CGAL/Alpha_wrap_3/internal/Alpha_wrap_3.h
```diff
@@ -1050,7 +1050,7 @@ class Alpha_wrap_3
         CGAL_assertion(CGAL::abs(CGAL::approximate_sqrt(m_oracle.squared_distance(steiner_point)) - m_offset) <= 1e-2 * m_offset);
 
         // locate cells that are going to be destroyed and remove their facet from the queue
-        int li, lj = 0;
+        int li = 0, lj = 0;
```
lrineau, line 1053: Please use `-1`:
```suggestion
        int li = -1, lj = -1;
```

PR #6638, opened 2022-06-02
Disable boost pool allocator

### LINKED ISSUE #6293: Degraded performance of CGAL::intersection between a 32 bit and 64 bit application while multi-threading
## Issue Details

We have recently changed our application from 32 bit to 64 bit and have noticed a degradation in performance when using CGAL::intersection in multiple worker threads. Each thread is using a distinct set of shapes, and so do not interfere with each other.

## Source Code

Here is a simple example I put together to illustrate the issue:

```
#include <iostream>
#include <vector>
#include <thread>

#include <CGAL/Boolean_set_operations_2.h>
#include <CGAL/Gps_circle_segment_traits_2.h>

typedef CGAL::Exact_predicates_exact_constructions_kernel   Kernel;
typedef CGAL::Gps_circle_segment_traits_2<Kernel>           Traits_2;

typedef Kernel::Point_2                                     Point_2;
typedef Traits_2::General_polygon_2                         Polygon_2;
typedef Traits_2::General_polygon_with_holes_2              Polygon_with_holes_2;
typedef Traits_2::X_monotone_curve_2                        X_monotone_curve_2;

const bool multiThreading = true;
const int threadCount = 7;
const int attempts = 100;
const int overallAttempts = 10;

Polygon_2 construct_polygon(const Point_2& p1, const Point_2& p2, const Point_2& p3, const Point_2& p4, const Point_2& p5)
{
    Polygon_2 pgn;
    X_monotone_curve_2 s1(p1, p2);    pgn.push_back(s1);
    X_monotone_curve_2 s2(p2, p3);    pgn.push_back(s2);
    X_monotone_curve_2 s3(p3, p4);    pgn.push_back(s3);
    X_monotone_curve_2 s4(p4, p5);    pgn.push_back(s4);
    X_monotone_curve_2 s5(p5, p1);    pgn.push_back(s5);
    return pgn;
}

void do_intersection()
{
    Polygon_2 outerPolygon = construct_polygon(Point_2(0, 0), Point_2(300, 0), Point_2(310, 150), Point_2(300, 300), Point_2(0, 300));

    int numHoles = 100;
    std::vector<Polygon_2> holes(numHoles);
    for (int i = 0; i < numHoles; i++)
    {
        holes[i] = construct_polygon(Point_2(2 * i + 1, 2 * i + 1), Point_2(2 * i + 1, 2 * i + 1.5), Point_2(2 * i + 2, 2 * i + 2), Point_2(2 * i + 5, 2 * i + 2), Point_2(2 * i + 3, 2 * i + 1));
    }

    Polygon_with_holes_2 p(outerPolygon, holes.begin(), holes.end());

    Polygon_2 poly1 = construct_polygon(Point_2(1, 0), Point_2(50, 1), Point_2(12, 13), Point_2(25, 50), Point_2(12, 100));

    int intersectAttempts = multiThreading ? attempts / threadCount : attempts;
    for (int i = 0; i < intersectAttempts; i++)
    {
        std::list<Polygon_with_holes_2> intersect;
        CGAL::intersection(p, poly1, std::back_inserter(intersect));
    }
}

{
    long long averageTime = 0;
    for (int x = 0; x < overallAttempts; x++)
    {
        auto startTime = std::chrono::high_resolution_clock::now();

        if (multiThreading)
        {
            std::thread threads[threadCount];
            for (int i = 0; i < threadCount; i++)
            {
                threads[i] = std::thread(do_intersection);
            }

            for (int i = 0; i < threadCount; i++)
            {
                threads[i].join();
            }
        }
        else
        {
            do_intersection();
        }

        auto endTime = std::chrono::high_resolution_clock::now();
        auto diffTime = endTime - startTime;
        averageTime = averageTime + diffTime.count();
        std::cout << "Total cost: " << diffTime.count() * 1e-9 << std::endl;
    }
    std::cout << std::endl;
    std::cout << "Average cost: " << (averageTime * 1e-9)/overallAttempts << std::endl;
    
    return 0;
}
```
When using a single thread, this example is roughly 15% faster in x64 than in x86. However when using 7 worker threads (my machine has 8 logical processors), this example is roughly 50% slower in x64 than x86.

We are unsure why this is considerably slower, but i think it may be related to using CGAL::intersection on a polygon with a lot of holes. Any helpful suggestions would be greatly appreciated.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 10 64 bit
* Compiler: MSVC
* Release or debug mode: Both
* CGAL version: 5.3
* Boost version: 1.77

sloriot:
Some regressions are observed in multithreaded env

Should fix #6293

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_dcel_base.h
@@ -34,8 +34,6 @@
 #include <CGAL/Arrangement_2/Arrangement_2_iterators.h>
 #include <CGAL/assertions.h>
 
-#include <boost/pool/pool_alloc.hpp>
-
 namespace CGAL {
 
 inline void* _clean_pointer(const void* p)
@@ -964,7 +962,7 @@ public:
  * The arrangement DCEL class.
  */
 template <class V, class H, class F,
-          class Allocator = boost::fast_pool_allocator<int> >
+          class Allocator = CGAL_ALLOCATOR(int) >
 class Arr_dcel_base {
 public:
   // Define the vertex, halfedge and face types.
--- a/Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arrangement_on_surface_2.h
@@ -46,8 +46,6 @@
 #include <CGAL/Iterator_project.h>
 #include <CGAL/Iterator_transform.h>
 
-#include <boost/pool/pool_alloc.hpp>
-
 namespace CGAL {
 
 /*! \class Arrangement_on_surface_2
@@ -66,7 +64,7 @@ class Arrangement_on_surface_2 {
 public:
   typedef GeomTraits_                                     Geometry_traits_2;
   typedef TopTraits_                                      Topology_traits;
-  typedef boost::fast_pool_allocator<int>                 Allocator;
+  typedef CGAL_ALLOCATOR(int)                             Allocator;
 
   // first define adaptor ...
   typedef Arr_traits_basic_adaptor_2<Geometry_traits_2>   Traits_adaptor_2;


sloriot:

PR #6640, opened 2022-06-04
Qualify do_intersect to avoid ambiguous dispatch

### LINKED ISSUE #6600: `CGAL::do_intersect` on `Segment_3` is Ambiguous in v5.4
## Issue Details

Up to v5.3.1, calling `CGAL::do_intersect()` on two `Segment_3`s works as expected (the following program prints `3D segments do intersect`). In v5.4, the following minimum (non-)working example does not compile; CGAL's internal call to `do_intersect` is ambiguous. The two possible candidates between which the compiler cannot distinguish are https://github.com/CGAL/cgal/blob/c58ac97e93c838ebfb1e8adaf23ff4fd185dc8e4/Intersections_3/include/CGAL/Intersections_3/internal/Line_3_Line_3_do_intersect.h#L20-L24 and https://github.com/CGAL/cgal/blob/c58ac97e93c838ebfb1e8adaf23ff4fd185dc8e4/Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/do_intersect.h#L237-L240 .

## Source Code

```

using CGAL_Kernel = CGAL::Exact_predicates_exact_constructions_kernel;
using Point_3 = CGAL::Point_3<CGAL_Kernel>;
using Segment_3 = CGAL_Kernel::Segment_3;

int main(int argc, char* argv[]) {
        Segment_3 seg1(Point_3(1., 1.5, 1.), Point_3(1., 1.5, 10.5));           // Vertical in the Z axis
        Segment_3 seg2(Point_3(-3.5, 1.5, 2.2), Point_3(42., 1.5, 2.2));        // Horizontal in X axis, intersecting
        const bool intersects = CGAL::do_intersect(seg1, seg2);
        printf("3D segments do%s intersect\n", intersects ? "" : " not");
        return 0;
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Ubuntu 20.04 LTS
* Compiler: clang version 12.0.0-3ubuntu1~20.04.5
* Specific flags used (if any): `-O3 -DNDEBUG`
* CGAL version: v5.4
* Boost version: 1.78
* Other libraries versions if used (Eigen, TBB, etc.): n/a

pentacular:
## Summary of Changes

Add qualifying namespace to resolve ambiguity in do_intersect dispatch.

## Release Management

* Issue(s) solved (if any): fix #6600

### DIFF
--- a/Intersections_3/include/CGAL/Intersections_3/internal/Segment_3_Segment_3_do_intersect.h
+++ b/Intersections_3/include/CGAL/Intersections_3/internal/Segment_3_Segment_3_do_intersect.h
@@ -30,7 +30,7 @@ do_intersect(const typename K::Segment_3& s1,
 {
   CGAL_precondition(!s1.is_degenerate() && !s2.is_degenerate());
 
-  bool b = do_intersect(s1.supporting_line(), s2.supporting_line(), k);
+  bool b = internal::do_intersect(s1.supporting_line(), s2.supporting_line(), k);
   if(b)
   {
     // supporting_line intersects: points are coplanar


lrineau (approved):

MaelRL:
Won't the same issue show up in https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Line_3_Segment_3_do_intersect.h#L34=, https://github.com/CGAL/cgal/blob/master/Intersections_3/include/CGAL/Intersections_3/internal/Tetrahedron_3_Bounded_3_do_intersect.h#L55=, and likely many more?

pentacular:
Probably the better solution is to constrain the InputIterator template which accepts any three argument case where the first two arguments have the same time.

pentacular:
I think we can drop this one as there is a more principled solution in https://github.com/CGAL/cgal/pull/6650 :)

lrineau:
I disagree. This PR helps the compiler, that does not have to look at so many overloads.

sloriot:

Intersections_3/include/CGAL/Intersections_3/internal/Segment_3_Segment_3_do_intersect.h
```diff
@@ -30,7 +30,7 @@ do_intersect(const typename K::Segment_3& s1,
 {
   CGAL_precondition(!s1.is_degenerate() && !s2.is_degenerate());
 
-  bool b = do_intersect(s1.supporting_line(), s2.supporting_line(), k);
+  bool b = internal::do_intersect(s1.supporting_line(), s2.supporting_line(), k);
```
lrineau, line 33: @afabri We discussed recently and I claimed that we should never call functions unqualified, to avoid ADL.

I just found that item in our coding *rules*:
>  Do not call global functions unqualified, that is, always specify explicitly the namespace where the function is defined. 

See https://doc.cgal.org/latest/Manual/devman_code_format.html#secprogramming_conventions

PR #6646, opened 2022-06-07
Use ASAN optionally

lrineau:
## Summary of Changes

Enable the AddressSanitizer in CGAL is the CMake or env variable `CGAL_USE_ASAN` is set.

## Release Management


afabri:
> Successfully tested in CGAL-5.6-Ic-1 (it does not break the testsuite)

Does it mean that one column no uses the sanitizer?  Which one is it?

sloriot:
Successfully tested in CGAL-5.6-Ic-15 (no error induced by the changes on other platforms)

PR #6648, opened 2022-06-07
Add an option to a remove_almost_degenerate_faces

sloriot:
Add an option to a non-documented function prevent flipping triangles with a height larger than a given input threshold.


lrineau:
That new option does not seem to be tested. @sloriot?

sloriot:

PR #6669, opened 2022-06-15
Better doc + early quit

sloriot:
improve doc + add an early quit to avoid building the AABB-tree when we reached the last level of the hierarchy


afabri:
Can you provide some explanation as you not only change the documentation but the code.

sloriot:

PR #6673, opened 2022-06-16
STL_extension: Fix memory leak in Skiplist

afabri:
## Summary of Changes

While `push_back()` creates elements on the heap and `clear()` deletes them, `pop_back()` does not, which led to a memory leak.
The class `Skiplist` is used by packages such as 2D Triangulations.

## Release Management


lrineau (approved):
I confirm that patch fixes the memory leak issue.

sloriot:

PR #6674, opened 2022-06-16
Polyhedron_3:  Accelerate removal of many items

afabri:
## Summary of Changes

To erase many items takes too long.  This PR will accelerate this.  The first commit does not yet bring the desired result.

## Release Management

* Affected package(s):  Polyhedron demo

### DIFF
--- a/Polyhedron/demo/Polyhedron/Scene.cpp
+++ b/Polyhedron/demo/Polyhedron/Scene.cpp
@@ -27,6 +27,7 @@
 #include <QMimeData>
 #include <QOpenGLFramebufferObject>
 
+#include <unordered_set>
 
 Scene::Scene(QObject* parent)
     : QStandardItemModel(parent),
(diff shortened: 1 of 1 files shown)


lrineau (approved):

afabri:
It is the `delete`   in  [`Vao()`](https://github.com/CGAL/cgal/blob/master/Three/include/CGAL/Three/Buffer_objects.h#L72) which takes all the time.

sloriot:
Interestingly, Andreas is using a Nvidia card on windows and I'm using a Nvidia card on Linux. We both observe the slow removal issue. Laurent did not observe that issue with his Intel graphic card on Linux and neither do I with my old laptop that is also using a Intel card. Would be interesting to test this with an AMD graphic card.

lrineau:
> Interestingly, Andreas is using a Nvidia card on windows and I'm using a Nvidia card on Linux. We both observe the slow removal issue. Laurent did not observe that issue with his Intel graphic card on Linux and neither do I with my old laptop that is also using a Intel card. Would be interesting to test this with an AMD graphic card.

That probably depends on the version of the OpenGL API that is used: on my laptop, if I use the `--old` option (as in `./Polyhedron_3 --old`) then I experience the bug of slow destruction of a thousand of items.

sloriot:

PR #6688, opened 2022-06-21
CMap/GMap/LCC with index

gdamiand:

## Summary of Changes

Add a version of Combinatorial map, Generalized map and Linear cell complex that use indices instead of handles.

## Release Management

* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Cmap_with_index
* Link to compiled documentation (obligatory for small feature) see the small feature web page

**TODO:**
- [x] Check in the testsuite that https://github.com/CGAL/cgal/issues/6579 is fixed by this branch


Combinatorial_map/doc/Combinatorial_map/Combinatorial_map.txt
```diff
@@ -168,7 +168,11 @@ Considering these different advantages and drawbacks, you can choose to use gene
...
 
-The darts and attributes are accessed through <I>handles</I>. A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `boost::unordered_map`.
+\cgalModifBegin
+
+The darts and attributes are accessed through <I>descriptors</I> (either Indices or Handles). A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `boost::unordered_map`. An index is a model of the `Index` concept, which is mainly an integer which is convertible from and to std::size_t. As the indices are contiguous, they can be used as index into vectors which store properties.
```
sloriot, line 173:
```suggestion
The darts and attributes are accessed through <I>descriptors</I> (either Indices or Handles). A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `std::unordered_map`. An index is a model of the `Index` concept, which is mainly an integer which is convertible from and to std::size_t. As the indices are contiguous, they can be used as index into vectors which store properties.
```

Combinatorial_map/doc/Combinatorial_map/Combinatorial_map.txt
```diff
@@ -228,9 +232,36 @@ typedef CGAL::Combinatorial_map<3, Example_items_3> Example_custom_cmap3;
 
 \endcode
 
+\cgalModifBegin
+
+\subsection ssecCMapIndicesHandles Indices or Handles
+
+By default, descriptors used to access darts and attributes are handles, and the darts and attributes are stored in a `Compact_container`. To use the index version, you should define the type `Use_index` equal to `CGAL::Tag_true` in the item class like in the code below. You can also define the type `Index_type` used to store indices (std::uint32_t by default when this type is not defined).
```
sloriot, line 239:
```suggestion
By default, descriptors used to access darts and attributes are handles, and the darts and attributes are stored in a `Compact_container`. To use the index version, you should define the type `Use_index` to `CGAL::Tag_true` in the item class like in the code below. You can also define the type `Index_type` used to store indices (`std::uint32_t` by default when this type is not defined).
```

Combinatorial_map/doc/Combinatorial_map/Concepts/CombinatorialMap.h
```diff
@@ -15,72 +15,72 @@ class CombinatorialMap {
...
+     and `d`\f$ \in\f$ \link GenericMap::darts `darts()`\endlink.
 
 */
 template<int i, int j>
-Dart_const_handle beta(Dart_const_handle dh) const;
+Dart_const_descriptor beta(Dart_const_descriptor d) const;
 
 /*!
-Returns a handle to a dart belonging to the same edge than dart `*dh`, and not to the same vertex. `nullptr` if such a dart does not exist.
+Returns a descriptor to a dart belonging to the same edge than dart `d`, and not to the same vertex. `null_descriptor` if such a dart does not exist.
```
sloriot, line 76: Should it be `null_dart_decriptor` here? If yes, please check all of them.

Combinatorial_map/doc/Combinatorial_map/Concepts/CombinatorialMap.h
```diff
@@ -108,67 +108,67 @@ bool is_valid() const;
...
-\pre \link CombinatorialMap::is_sewable `is_sewable<i>(dh1,dh2)`\endlink.
+If \link GenericMap::are_attributes_automatically_managed `are_attributes_automatically_managed()`\endlink`==true`, when necessary, non void attributes are updated to ensure the validity of the generic map: for each <I>j</I>-cells <I>c1</I> and <I>c2</I> which are merged into one <I>j</I>-cell during the sew, the two associated attributes <I>attr1</I> and <I>attr2</I> are considered. If one attribute is `null_descriptor` and the other not, the non `null_descriptor` attribute is associated to all the darts of the resulting cell. When the two attributes are non `null_descriptor`, functor \link CellAttribute::On_merge `Attribute_type<i>::type::On_merge`\endlink is called on the two attributes <I>attr1</I> and <I>attr2</I>. If set, the dynamic onmerge function of <i>i</i>-attributes is also called on <I>attr1</I> and <I>attr2</I>. Then, the attribute <I>attr1</I> is associated to all darts of the resulting <I>j</I>-cell. Finally, attribute <I>attr2</I> is removed from the generic map.
```
sloriot, line 149: For attributes, shouldn't you have `null_attribute_descriptor`?

Combinatorial_map/doc/Combinatorial_map/Concepts/GenericMapItems.h
```diff
@@ -35,6 +36,19 @@ The following examples show two possible models of the `GenericMapItems` concept
 */
 class GenericMapItems {
 public:
+  /*!
+    \cgalModifBegin
+    If Use_index is equal to CGAL::Tag_true, use indices as descriptors. Otherwise (if this type is not defined or different than CGAL::Tag_true), use handles.
```
sloriot, line 41:
```suggestion
    If `Use_index` is  `CGAL::Tag_true`, uses indices as descriptors. Otherwise (if this type is not defined or different from `CGAL::Tag_true`), use handles.
```

Combinatorial_map/doc/Combinatorial_map/Concepts/GenericMapItems.h
```diff
@@ -35,6 +36,19 @@ The following examples show two possible models of the `GenericMapItems` concept
...
+    \cgalModifEnd
+  */
+  typedef unspecified_type Use_index;
+
+  /*!
+    \cgalModifBegin
+    Number type used for indices when Use_index is CGAL::Tag_true. By default, use std::uint32_t.
```
sloriot, line 48:
```suggestion
    Number type used for indices when Use_index is `CGAL::Tag_true`. By default, use `std::uint32_t`.
```

Combinatorial_map/include/CGAL/Combinatorial_map.h
```diff
@@ -198,14 +200,14 @@ namespace CGAL {
...
-      init_dart(null_dart_handle);
+      init_dart(null_dart_descriptor);
 
       CGAL_assertion(number_of_darts()==0);
     }
 
     /** Copy the given combinatorial map 'amap' into *this.
      *  Note that both CMap can have different dimensions and/or non void attributes.
-     *  Here CMap2 is necessarily non const; while Dart_handle_2 can be a const or non const handle.
+     *  Here CMap2 is necessarily non const; while Dart_descriptor_2 can be a const or non const handle.
```
sloriot, line 210:
```suggestion
     *  Here CMap2 is necessarily non const; while Dart_descriptor_2 can be a const or non const descriptor.
```

Combinatorial_map/include/CGAL/Combinatorial_map.h
```diff
@@ -759,84 +752,84 @@ namespace CGAL {
...
+    int highest_nonfree_dimension(Dart_const_descriptor dh) const
     {
       for (int i=(int)dimension; i>=0; --i)
       { if ( !is_free(dh, i) ) return i; }
       return -1;
     }
 
     /** Return a dart belonging to the same edge and to the second vertex
-     * of the current edge (nullptr if such a dart does not exist).
+     * of the current edge (null_descriptor if such a dart does not exist).
```
sloriot, line 763:
```suggestion
     * of the current edge (`null_dart_descriptor` if such a dart does not exist).
```

sloriot: same change should be done everywhere, right?

Generalized_map/doc/Generalized_map/CGAL/Generalized_map.h
```diff
@@ -6,7 +6,9 @@ namespace CGAL {
 
 The class `Generalized_map` represents a <I>d</I>D generalized map.
 
-Darts and non void attributes are stored in memory using `Compact_container`, using `Alloc` as allocator.
+\cgalModifBegin
+Two versions exist: one where Darts and non void attributes are stored in memory using `Compact_container`, using `Alloc` as allocator, and use handles as descriptors; a second one where Darts and non void attributes are stored in an internal std::vector like data-structure, and use indices as descriptors. The choice between the two versions is done through the item class.
```
sloriot, line 10:
```suggestion
Two versions exist: one where darts and non void attributes are stored in memory using `Compact_container`, using `Alloc` as allocator, and use handles as descriptors; a second one where darts and non void attributes are stored in an internal `std::vector` like data-structure, and use indices as descriptors. The choice between the two versions is done through the item class.
```

Generalized_map/doc/Generalized_map/Generalized_map.txt
```diff
@@ -167,7 +167,11 @@ Considering these different advantages and drawbacks, you can choose to use gene
...
 
-The darts and attributes are accessed through <I>handles</I>. A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `boost::unordered_map`.
+\cgalModifBegin
+
+The darts and attributes are accessed through <I>descriptors</I> (either Indices or Handles). A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `boost::unordered_map`. An index is a model of the `Index` concept, which is mainly an integer which is convertible from and to std::size_t. As the indices are contiguous, they can be used as index into vectors which store properties.
```
sloriot, line 172:
```suggestion
The darts and attributes are accessed through <I>descriptors</I> (either Indices or Handles). A handle is a model of the `Handle` concept, thus supporting the two dereference operators `operator*` and `operator->`. All handles are model of `LessThanComparable` and `Hashable`, that is they can be used as keys in containers such as `std::map` and `std::unordered_map`. An index is a model of the `Index` concept, which is mainly an integer which is convertible from and to `std::size_t`. As the indices are contiguous, they can be used as index into vectors which store properties.
```

Generalized_map/doc/Generalized_map/Generalized_map.txt
```diff
@@ -227,9 +231,36 @@ typedef CGAL::Generalized_map<3, Example_items_3> Example_custom_gmap3;
...
+An important operation in generalized maps consists in iterating over specific subsets of darts or over attributes. For that, several <I>ranges</I> are offered (see Section \ref ssecrange "Iterating over Orbits, Cells, and Attributes"). A range is a model of the `Range` concept, thus supporting the two methods `begin()` and `end()` allowing to iterate over all the elements in the range. Several functions allow to create specific configurations of darts into a generalized map (see Section \ref ssecconstructiongmap "Construction Operations"). Darts can be marked during operations, for example when performing a breadth-first search traversal, thanks to Boolean marks (see Sections \ref ssecadvmarksgmap "Boolean Marks"). \tred{In the following, we denote by `d0`, `d1`, `d2` for dart descriptors, and identify in explanations the descriptor and the object it describes.}
```
sloriot, line 263:
```suggestion
An important operation in generalized maps consists in iterating over specific subsets of darts or over attributes. For that, several <I>ranges</I> are offered (see Section \ref ssecrange "Iterating over Orbits, Cells, and Attributes"). A range is a model of the `Range` concept, thus supporting the two methods `begin()` and `end()` allowing to iterate over all the elements in the range. Several functions allow to create specific configurations of darts into a generalized map (see Section \ref ssecconstructiongmap "Construction Operations"). Darts can be marked during operations, for example when performing a breadth-first search traversal, thanks to Boolean marks (see Sections \ref ssecadvmarksgmap "Boolean Marks"). \tred{In the following, we denote by `d0`, `d1`, `d2` for dart descriptors, and identify in the explanations the descriptor and the object it describes.}
```

Linear_cell_complex/doc/Linear_cell_complex/CGAL/Linear_cell_complex_for_combinatorial_map.h
```diff
@@ -5,6 +5,9 @@ namespace CGAL {
 \ingroup PkgLinearCellComplexClasses
 
 The class `Linear_cell_complex_for_combinatorial_map` represents a linear cell complex in dimension `d`, in an ambient space of dimension `d2`, using a combinatorial map as underlying combinatorial data-structure.
+\cgalModifBegin
+Like for `CombinatorialMap`, two versions exist: one where Darts and non void attributes are stored in memory using `Compact_container`, using `Alloc` as allocator, and use handles as descriptors; a second one where Darts and non void attributes are stored in an internal std::vector like data-structure, and use indices as descriptors. The choice between the two versions is done through the item class.
```
sloriot, line 9:
```suggestion
Like for `Combinatorial_map`, two versions exist: one where Darts and non void attributes are stored in memory using `Compact_container`, using `Alloc` as allocator, and use handles as descriptors; a second one where Darts and non void attributes are stored in an internal std::vector like data-structure, and use indices as descriptors. The choice between the two versions is done through the item class.
```

sloriot (commented):
One possible technical solution instead of having both `Use_Index` and `Index_type` would be have only `Index_type`. If it is defined, then you use indices and not then you use handles. Perfectly backward compatible (+ I guess you already have something internally checking if `Use_index` is here or not).

Do you have the notion of garbage when using indices? I did not see any mention of it.

gdamiand:
> One possible technical solution instead of having both `Use_Index` and `Index_type` would be have only `Index_type`. If it is defined, then you use indices and not then you use handles. Perfectly backward compatible (+ I guess you already have something internally checking if `Use_index` is here or not).

Why not; but I am not sure that your proposal is better. Indeed, users will probably mainly use only `Use_Index` and keep the default type for Index_type. This second type will be used only for some special cases. Thus the current solution is probably better for most "classical cases".

> Do you have the notion of garbage when using indices? I did not see any mention of it.

What do you mean? The solution for free list is exactly the same for indices than for handles; erased elements are linked and reused for the next allocations.

Combinatorial_map/doc/Combinatorial_map/Concepts/CombinatorialMap.h
gdamiand: No this is null_descriptor (equivalent to nullptr for indices).

null_dart_descriptor is a descriptor to the special dart used to represent borders.

Combinatorial_map/doc/Combinatorial_map/Concepts/CombinatorialMap.h
gdamiand: No (there is no a special attribute to represent border like for darts).

sloriot:
> > One possible technical solution instead of having both `Use_Index` and `Index_type` would be have only `Index_type`. If it is defined, then you use indices and not then you use handles. Perfectly backward compatible (+ I guess you already have something internally checking if `Use_index` is here or not).
> 
> Why not; but I am not sure that your proposal is better. Indeed, users will probably mainly use only `Use_Index` and keep the default type for Index_type. This second type will be used only for some special cases. Thus the current solution is probably better for most "classical cases".

Oh I see, you are right. `using Index_type = Default` would also be an option. I'd prefer having only one but I don't mind if you prefer to keep your solution.

> 
> > Do you have the notion of garbage when using indices? I did not see any mention of it.
> 
> What do you mean? The solution for free list is exactly the same for indices than for handles; erased elements are linked and reused for the next allocations.

Do you have the equivalent to "collect_garbage()` for example? Say it the users want to have contiguous indices. Some info about how to get the max index would also be intersecting.

Combinatorial_map/include/CGAL/Combinatorial_map.h
gdamiand: No (cf answer above)

gdamiand:
> > > Do you have the notion of garbage when using indices? I did not see any mention of it.
> > 
> > 
> > What do you mean? The solution for free list is exactly the same for indices than for handles; erased elements are linked and reused for the next allocations.
> 
> Do you have the equivalent to `collect_garbage()` for example? Say it the users want to have contiguous indices. Some info about how to get the max index would also be intersecting.

No collect_garbage; like for handle, it is enough to copy a cmap into a new one to remove all freed elements. I can try to add a sentence somewhere to say that.

About max index, there is a method upper_bound() in the compact container with index. But this class is internal (undocumented).

sloriot:
What about adding a free function like `num_darts()` ?

sloriot:
> > > > Do you have the notion of garbage when using indices? I did not see any mention of it.
> > > 
> > > 
> > > What do you mean? The solution for free list is exactly the same for indices than for handles; erased elements are linked and reused for the next allocations.
> > 
> > 
> > Do you have the equivalent to `collect_garbage()` for example? Say it the users want to have contiguous indices. Some info about how to get the max index would also be intersecting.
> 
> No collect_garbage; like for handle, it is enough to copy a cmap into a new one to remove all freed elements. I can try to add a sentence somewhere to say that.

Note that the swap trick is c++98. Since c++11 there is `shrink_to_fit()` in `std::vector`.

> 
> About max index, there is a method upper_bound() in the compact container with index. But this class is internal (undocumented).

gdamiand:
> What about adding a free function like `num_darts()` ?

I don't like the name num_xxx to get an upper bound ! I think this is really confusing.

sloriot:
> > What about adding a free function like `num_darts()` ?
> 
> I don't like the name num_xxx to get an upper bound ! I think this is really confusing.

I don't really care about the name, I just need a function to exist and to be documented (I really use it a lot in `Surface_mesh` so will the user of CMap)

sloriot:
> > > > What about adding a free function like `num_darts()` ?
> > > 
> > > 
> > > I don't like the name num_xxx to get an upper bound ! I think this is really confusing.
> > 
> > 
> > I don't really care about the name, I just need a function to exist and to be documented (I really use it a lot in `Surface_mesh` so will the user of CMap)
> 
> Instead of free functions, I can add methods in CMap class? But the question is what to return when we use handles? 0?
Sure. Something like "returns an upper bound of the id of descriptor if indices are used or 0 otherwise". Note that you also need to add something about the non-contiguity of ids.

gdamiand:
> should be `returns ...`

I need to update all others methods; probably better to stay homogeneous.

sloriot:
I could not see any example using indices. Did I miss them?

sloriot:
I think `upper_bound_dart_ids()` should be either `upper_bound_on_dart_ids()` or `dart_id_upper_bound()` (but since you have `number_of_xxx` you should pick  `upper_bound_on_dart_ids()`.

gdamiand:
> I think `upper_bound_dart_ids()` should be either `upper_bound_on_dart_ids()` or `dart_id_upper_bound()` (but since you have `number_of_xxx` you should pick `upper_bound_on_dart_ids()`.

done in https://github.com/CGAL/cgal/pull/6688/commits/cbc038fff856acca0454f8c96684d8d06b806e5b

sloriot:
Last question :) What about renaming `Use_index` to `Use_indices`? The comment you wrote in the example seems to confirm it would be a better choice.

sloriot:
You need to update CHANGES.md

gdamiand:
> You need to update CHANGES.md

1) Will do
2) plus add the example in the cmap user manual (please confirm the example is ok)
3) ok for you if I keep Use_index ?

sloriot:
I pre-approved the small feature so I'm fine with the current version.

Installation/CHANGES.md
```diff
@@ -1,6 +1,14 @@
...
 ===============
 
+[Release 5.6](https://github.com/CGAL/cgal/releases/tag/v5.6)
+-----------
+
+Release date: XXX
+
+### [Combinatorial Maps](https://doc.cgal.org/5.6/Manual/packages.html#PkgCombinatorialMaps)
+
+- Added a version that uses indices instead of handles as dart and attribute descriptors. As the indices are integers convertible from and to std::size_t, they can be used as index into vectors which store properties. To use the index version, you should define the type Use_index equal to CGAL::Tag_true in the item class.
```
sloriot, line 11:
```suggestion
- Added a version that uses indices instead of handles as dart and attribute descriptors. As the indices are integers convertible from and to `std::size_t`, they can be used as index into vectors which store properties. To use the index version,  `Use_index` must be defined and be equal to `CGAL::Tag_true` in the item class.
```

sloriot:
Several errors and warnings in CGAL-5.6-Ic-28:
- Generalized_map
- Hash_map
- Combinatorial_map
- Linear_cell_complex

Combinatorial_map/include/CGAL/Compact_container_with_index.h
```diff
@@ -0,0 +1,1022 @@
...
+// (1) Case with a deque for the freelist and a vector for Booleans.
+template<typename CC_with_index>
+class Free_list_management<CC_with_index, CGAL::Tag_true, CGAL::Tag_true>
+{
+  using Self=Free_list_management<CC_with_index, CGAL::Tag_true, CGAL::Tag_true>;
+  using T=typename CC_with_index::value_type;
+  using size_type=typename CC_with_index::size_type;
+
+public:
+  static const size_type null_descriptor=std::numeric_limits<size_type>::max();
```
sloriot, line 100:
```suggestion
  static const size_type null_descriptor=(std::numeric_limits<size_type>::max)();
```

Combinatorial_map/include/CGAL/Compact_container_with_index.h
```diff
@@ -0,0 +1,1022 @@
...
+template<typename CC_with_index>
+class Free_list_management<CC_with_index, CGAL::Tag_false, CGAL::Tag_true>
+{
+  using Self=Free_list_management<CC_with_index, CGAL::Tag_false, CGAL::Tag_true>;
+  using T=typename CC_with_index::value_type;
+  using size_type=typename CC_with_index::size_type;
+  using Traits=Compact_container_with_index_traits <T, size_type>;
+
+public:
+  static const size_type null_descriptor=std::numeric_limits<size_type>::max();
```
sloriot, line 203:
```suggestion
  static const size_type null_descriptor=(std::numeric_limits<size_type>::max)();
```

Combinatorial_map/include/CGAL/Compact_container_with_index.h
```diff
@@ -0,0 +1,1022 @@
...
+template<typename CC_with_index>
+class Free_list_management<CC_with_index, CGAL::Tag_false, CGAL::Tag_false>
+{
+  using Self=Free_list_management<CC_with_index, CGAL::Tag_false, CGAL::Tag_false>;
+  using T=typename CC_with_index::value_type;
+  using size_type=typename CC_with_index::size_type;
+  using Traits=Compact_container_with_index_traits <T, size_type>;
+
+public:
+  static const size_type null_descriptor=std::numeric_limits<size_type>::max()/2;
```
sloriot, line 313:
```suggestion
  static const size_type null_descriptor=(std::numeric_limits<size_type>::max)()/2;
```

gdamiand:
> [This one](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-28/Ridges_3_Examples/TestReport_lrineau_Ubuntu-Latest-CXX11.gz) is strange too:
> 
> ```
> Ridges_Umbilics_LCC.cpp:(.text._ZN4CGAL8halfedgeILj2ELj3ENS_26Linear_cell_complex_traitsILj3ENS_16Simple_cartesianIdEEEENS_33Linear_cell_complex_bgl_min_itemsESaIiENS_22Combinatorial_map_baseENS_34CMap_linear_cell_complex_storage_1ILj2ELj3ES4_S5_S6_EEEESt4pairIN5boost12graph_traitsINS_41Linear_cell_complex_for_combinatorial_mapIXT_EXT0_ET1_T2_T3_T4_T5_EEE19halfedge_descriptorEbENSK_17vertex_descriptorESN_RKSJ_[_ZN4CGAL8halfedgeILj2ELj3ENS_26Linear_cell_complex_traitsILj3ENS_16Simple_cartesianIdEEEENS_33Linear_cell_complex_bgl_min_itemsESaIiENS_22Combinatorial_map_baseENS_34CMap_linear_cell_complex_storage_1ILj2ELj3ES4_S5_S6_EEEESt4pairIN5boost12graph_traitsINS_41Linear_cell_complex_for_combinatorial_mapIXT_EXT0_ET1_T2_T3_T4_T5_EEE19halfedge_descriptorEbENSK_17vertex_descriptorESN_RKSJ_]+0x19c): undefined reference to `CGAL::CMap_linear_cell_complex_storage_1<2u, 3u, CGAL::Linear_cell_complex_traits<3u, CGAL::Simple_cartesian<double> >, CGAL::Linear_cell_complex_bgl_min_items, std::allocator<int> >::null_descriptor'
> ```

null_descriptor is defined in the class CMap_linear_cell_complex_storage_1 by

<code>static constexpr Null_descriptor_type null_descriptor=nullptr;</code>

I think with c++11, inline initialization is enough to define the variable?

gdamiand:
> I would have opted for `Foobar() = default`;

Does default constructor zero initialize built-in types?

gdamiand:
> [This one](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-28/Ridges_3_Examples/TestReport_lrineau_Ubuntu-Latest-CXX11.gz) is strange too:
> 
> ```
> Ridges_Umbilics_LCC.cpp:(.text._ZN4CGAL8halfedgeILj2ELj3ENS_26Linear_cell_complex_traitsILj3ENS_16Simple_cartesianIdEEEENS_33Linear_cell_complex_bgl_min_itemsESaIiENS_22Combinatorial_map_baseENS_34CMap_linear_cell_complex_storage_1ILj2ELj3ES4_S5_S6_EEEESt4pairIN5boost12graph_traitsINS_41Linear_cell_complex_for_combinatorial_mapIXT_EXT0_ET1_T2_T3_T4_T5_EEE19halfedge_descriptorEbENSK_17vertex_descriptorESN_RKSJ_[_ZN4CGAL8halfedgeILj2ELj3ENS_26Linear_cell_complex_traitsILj3ENS_16Simple_cartesianIdEEEENS_33Linear_cell_complex_bgl_min_itemsESaIiENS_22Combinatorial_map_baseENS_34CMap_linear_cell_complex_storage_1ILj2ELj3ES4_S5_S6_EEEESt4pairIN5boost12graph_traitsINS_41Linear_cell_complex_for_combinatorial_mapIXT_EXT0_ET1_T2_T3_T4_T5_EEE19halfedge_descriptorEbENSK_17vertex_descriptorESN_RKSJ_]+0x19c): undefined reference to `CGAL::CMap_linear_cell_complex_storage_1<2u, 3u, CGAL::Linear_cell_complex_traits<3u, CGAL::Simple_cartesian<double> >, CGAL::Linear_cell_complex_bgl_min_items, std::allocator<int> >::null_descriptor'
> ```

Maybe it was related to the previous incorrect use of max? We can give it a try in the next test suite.

sloriot:
> > I would have opted for `Foobar() = default`;
> 
> Does default constructor zero initialize built-in types?

From [here](https://en.cppreference.com/w/cpp/language/value_initialization)
*2) if T is a class type with a default constructor that is not user-declared (until C++11)neither user-provided nor deleted (since C++11) (that is, it may be a class with an implicitly-defined or defaulted default constructor), the object is [zero-initialized](https://en.cppreference.com/w/cpp/language/zero_initialization) and the semantic constraints for default-initialization are checked, and if T has a non-trivial default constructor, the object is [default-initialized](https://en.cppreference.com/w/cpp/language/default_initialization);*

gdamiand:
> > > I would have opted for `Foobar() = default`;
> > 
> > 
> > Does default constructor zero initialize built-in types?
> 
> From [here](https://en.cppreference.com/w/cpp/language/value_initialization) _2) if T is a class type with a default constructor that is not user-declared (until C++11)neither user-provided nor deleted (since C++11) (that is, it may be a class with an implicitly-defined or defaulted default constructor), the object is [zero-initialized](https://en.cppreference.com/w/cpp/language/zero_initialization) and the semantic constraints for default-initialization are checked, and if T has a non-trivial default constructor, the object is [default-initialized](https://en.cppreference.com/w/cpp/language/default_initialization);_

Thanks @sloriot  !

sloriot:
Compilation issues:
https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-29/BGL/TestReport_lrineau_Ubuntu-Latest-CXX11.gz
https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-29/BGL/TestReport_lrineau_Debian-Stable.gz
https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-29/BGL/TestReport_gimeno_ArchLinux-clang.gz
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-29/Generalized_map_Examples/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-29/Ridges_3_Examples/TestReport_lrineau_Ubuntu-Latest-CXX11.gz
Warnings:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-29/Combinatorial_map/TestReport_Christo_MSVC2017-Debug-64bits.gz
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-29/Linear_cell_complex/TestReport_monet_x86_64-apple_bigsur_clang1200-release-64bits.gz
Link error:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-29/Polygon_mesh_processing/TestReport_lrineau_Ubuntu-Latest-CXX11.gz

Combinatorial_map/doc/Combinatorial_map/Concepts/GenericMapItems.h
```diff
@@ -2,16 +2,15 @@
...
 
 \cgalHasModel \link CGAL::Generic_map_min_items `CGAL::Generic_map_min_items`\endlink
 
 \sa `GenericMap`
 
 \cgalHeading{Example}
 
-The following examples show two possible models of the `GenericMapItems` concept: the first one for a generic map without dart information, nor enabled attributes, the second one for a generic map with a `double` associated with each dart, and edge attributes enabled, and associated with a \link CGAL::Cell_attribute `Cell_attribute`\endlink containing an `int`.
+The following examples show two possible models of the `GenericMapItems` concept: the first one for a generic map without dart information, nor enabled attributes, the second one for a generic map \tred{that uses std::uint16_t as indices,} with a `double` associated with each dart, and edge attributes enabled, and associated with a \link CGAL::Cell_attribute `Cell_attribute`\endlink containing an `int`.
```
albert-github, line 13: Seen the warning in the overnight build (see: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/ e.g. https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-29/index.html and https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-29/logs_master/Combinatorial_map.log, note that the line numbers are not 100% accurate) it looks like that the command `\tred` is not defined in the `ALIASES` or maybe something else is meant. When a command for the `ALIASES` is meant it is probably better to rename it to something like `cgalTred`.

gdamiand:
> New warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-31/Linear_cell_complex/TestReport_Christo_MSVC2015-Release-64bits.gz

I am not able to solve this warning:  'argument': conversion from 'size_t' to 'unsigned int', possible loss of data

I don't know which variable is converted. @sloriot any advice?

sloriot:
> For the error https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-31/Linear_cell_complex_Demo/TestReport_lrineau_ArchLinux-clang-CXX20-Release.gz
> 
> /mnt/testsuite/include/CGAL/CMap_linear_cell_complex_storages_with_index.h:312:33: error: no matching member function for call to 'emplace'
> 
> this is maybe a compiler bug?
> 
> I compile without error the same prog on my os, with clang version 13.0.0-2, in release and with -std=c++20.

What is the type of the class?

sloriot:
> > New warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-31/Linear_cell_complex/TestReport_Christo_MSVC2015-Release-64bits.gz
> 
> I am not able to solve this warning: 'argument': conversion from 'size_t' to 'unsigned int', possible loss of data
> 
> I don't know which variable is converted. @sloriot any advice?

A guess: maybe you use `Compact_container_with_index` somewhere where you did not pass the `IndexType` so the default is used while another type is expected.

gdamiand:
> > > New warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-31/Linear_cell_complex/TestReport_Christo_MSVC2015-Release-64bits.gz
> > 
> > 
> > I am not able to solve this warning: 'argument': conversion from 'size_t' to 'unsigned int', possible loss of data
> > I don't know which variable is converted. @sloriot any advice?
> 
> A guess: maybe you use `Compact_container_with_index` somewhere where you did not pass the `IndexType` so the default is used while another type is expected.

Thanks @sloriot  for the advice; but I think I found the problem and it is probably solved by https://github.com/CGAL/cgal/pull/6688/commits/195b0c1f48c5c68d78a12c89bc0154e10ccd2ad4

whatseven:
I searched about the combinatorial map and saw this PR occasionally. Thanks for all your effort about index-based CMap.

I have another small question about the upcoming feature. Would it be possible to use it in a multi-thread context? After digging into the current implementation of CMap, I find it is not compatible with parallel processing. Especially for the marking system. Each mark operation will trigger a global varieble update in the whole map.

gdamiand:
> I searched about the combinatorial map and saw this PR occasionally. Thanks for all your effort about index-based CMap.
> 
> I have another small question about the upcoming feature. Would it be possible to use it in a multi-thread context? After digging into the current implementation of CMap, I find it is not compatible with parallel processing. Especially for the marking system. Each mark operation will trigger a global varieble update in the whole map.

No, it can not be used in a multi-thread context. This is one of my todo; the problem is not only the global variable but also that bitset is not thread safe.

sloriot:
Still some link errors for example [here](https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-33/Polygon_mesh_processing_Examples/TestReport_lrineau_Debian-Stable.gz) and [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-33/Polygon_mesh_processing/TestReport_lrineau_Ubuntu-Latest-CXX11.gz). Do you want me to see if I can reproduce them locally?

gdamiand:
> Still some link errors for example [here](https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-33/Polygon_mesh_processing_Examples/TestReport_lrineau_Debian-Stable.gz) and [there](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-33/Polygon_mesh_processing/TestReport_lrineau_Ubuntu-Latest-CXX11.gz). Do you want me to see if I can reproduce them locally?

Yes please Sebastien. 
It is strange that this linker error occurs for PMP and not for the examples and tests in LCC!

sloriot:
That should be fixed. Actually it was noticed only in the tests/examples using BGL API because it seems to be the only cases where it is used and the standard says that a definition out of the class is needed (previous to c++17) if the variable is [odr-used](https://en.cppreference.com/w/cpp/language/definition#ODR-use).

gdamiand:
Thank you very much @sloriot !

gdamiand:
Tests seem ok in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-Ic-35.shtml

sloriot:

@gdamiand Before we can merge in master, the TODO must be addressed and please confirm that no macro for the review is present in the doc.

gdamiand:
> Successfully tested in CGAL-5.6-Ic-36.
> 
> @gdamiand Before we can merge in master, the TODO must be addressed and please confirm that no macro for the review is present in the doc.

The macro were removed in https://github.com/CGAL/cgal/pull/6688/commits/233f75714855f03dc863158b94ddaf88ba51b3a7

gdamiand:
> Successfully tested in CGAL-5.6-Ic-36.
> 
> @gdamiand Before we can merge in master, the TODO must be addressed and please confirm that no macro for the review is present in the doc.

In https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-Ic-35.shtml and in https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-Ic-36.shtml there is no assertion failed; thus https://github.com/CGAL/cgal/issues/6579 is maybe fixed.

PR #6691, opened 2022-06-21
Orthtree: Fix memory leak in test code

afabri:
## Summary of Changes

The helper class generates nodes that allocate memory without freeing it. 

## Release Management

* Affected package(s):  Orthtree

### DIFF
--- a/Orthtree/include/CGAL/Orthtree/Node.h
+++ b/Orthtree/include/CGAL/Orthtree/Node.h
@@ -45,7 +45,7 @@ struct Node_access
   template <typename Node>
   static void split(Node node) { return node.split(); }
 
-  template <int D, typename Node>
+  template <typename Node>
   static void free(Node node)
   {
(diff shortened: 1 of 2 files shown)


Orthtree/include/CGAL/Orthtree/Node.h
```diff
@@ -44,6 +44,22 @@ struct Node_access
 
   template <typename Node>
   static void split(Node node) { return node.split(); }
+
+  template <typename Node, int D>
```
sloriot, line 48:
```suggestion
  template <int D, typename Node>
```

`Node` can be deduced

sloriot:

lrineau:
The memory leak is not over. Here is a gist of the log of ASAN, in Debug:
https://gist.github.com/lrineau/5e3ade0bde620d718ed5c345efadc074

PR #6692, opened 2022-06-21
Cleanup CMake scripts

lrineau:
## Summary of Changes

1. Starting with CMake 3.18, use `cmake_language(DEFER CALL..)` instead of `variable_watch`. That was a really bad hack!
2. Create fixture `demo_framework_SetupFixture`.
   That avoids that all targets of CGAL depend on `demo_framework`. Now only the plugins depend on it.
3.  Fix the dependencies of CTest tests on `compilation_of__CGAL_Qt5_moc_and_resources`.

## Release Management

* Affected package(s): Installation, Polyhedron demo


sloriot:

PR #6695, opened 2022-06-22
Surface_mesh_segmentation: fix the memory leak

afabri:
## Summary of Changes

As pointed out by @lrineau the assignment of an empty graph, does not clear le lhs of the assignment.   Instead of writing an assignment operator for `graph` this PR adds a method    `graph::clear()".   
## Release Management

* Affected package(s): Surface_mesh_segmentation

### DIFF
--- a/Surface_mesh_segmentation/include/CGAL/Surface_mesh_segmentation/internal/auxiliary/graph.h
+++ b/Surface_mesh_segmentation/include/CGAL/Surface_mesh_segmentation/internal/auxiliary/graph.h
@@ -520,6 +520,8 @@ public:
   /* Destructor */
   ~Graph();
 
+  void clear();
+
   /* Adds a node to the graph */
   node_id add_node();
 
@@ -701,7 +703,7 @@ inline Graph::Graph(void (*err_function)(const char *))
   flow = 0;
 }
 
-inline Graph::~Graph()
+inline void Graph::clear()
 {
   while (node_block_first) {
     node_block *next = node_block_first -> next;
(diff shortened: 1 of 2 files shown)


afabri:
This code does not have a leak:
```
 MaxFlow::Graph g; 

    for (int i = 0; i < 5000; ++i) {
        auto u = g.add_node();
        auto v = g.add_node();
        g.add_edge(u,v,1, 1);
    }
    ```

**Comment by lrineau** [comment]
See https://gist.github.com/762d130066a2aaa71d9ffaecc0714432 for the full log of the run of `test/Surface_mesh_segmentation/mesh_segmentation_test` in Debug, with the AddressSanitizer. It seems th leak comes from MaxFlow.

**Comment by lrineau** [comment]
I think I have a clue about the issue. in `Surface_mesh_segmentation/include/CGAL/Surface_mesh_segmentation/internal/auxiliary/graph.h`, the class `Graph` allocated nodes with `new`, and its destructor `Graph::~Graph()` does deallocate them with `delete`. But there is a function:
```c++
  void clear_graph()
  {
    graph = MaxFlow::Graph();
  }
```
in `Surface_mesh_segmentation/include/CGAL/boost/graph/Alpha_expansion_MaxFlow_tag.h`. The class `Graph` does not have `operator=`, and thus the graph is reassigned without releasing the memory.

afabri:
I added a `clear()` function with the code of the destructor. We can then declare assignment and copy constructor private or deleted. The graph class is a data member in an undocumented class, and objects of this class are not assigned or copied.

lrineau:
@afabri You added this PR in the milestone for CGLA-5.5, but your branch is based on `master`.

I have rebased on `5.4.x-branch`, because that is a bug fix.

sloriot:
Successfully tested in CGAL-5.6-Ic-18 ([this error is gone](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-14/Surface_mesh_segmentation/TestReport_lrineau_Fedora-Release.gz))

PR #6701, opened 2022-06-23
Fix exterior offset creation for a polygon with holes

sloriot:
This PR is both a fix and a breaking change of a behavior that was IMO buggy in the first place (thus not backported to 5.4).

Fix the function creating exterior offsets to also consider holes + update functions dumping exterior offsets as polygons with holes to ignore the offset of the outer frame.

### DIFF
--- a/Straight_skeleton_2/examples/Straight_skeleton_2/exterior_offset_of_multiple_polygons_with_holes.cpp
+++ b/Straight_skeleton_2/examples/Straight_skeleton_2/exterior_offset_of_multiple_polygons_with_holes.cpp
@@ -14,11 +14,11 @@
 typedef CGAL::Exact_predicates_inexact_constructions_kernel K ;
 
 typedef K::Point_2                    Point ;
-typedef CGAL::Polygon_2<K>            Polygon ;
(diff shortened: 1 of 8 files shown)


Straight_skeleton_2/examples/Straight_skeleton_2/exterior_offset_of_multiple_polygons_with_holes.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <CGAL/draw_polygon_with_holes_2.h>
+
+
+#include <vector>
+#include <cassert>
+
+typedef CGAL::Exact_predicates_inexact_constructions_kernel K ;
+
+typedef K::Point_2                    Point ;
+typedef CGAL::Polygon_2<K>            Polygon ;
```
afabri, line 17: Elsewhere we have `typedef CGAL::Polygon_2<K>  Polygon_2 ;`  and I think this is because the token `Polygon`  is already used by some Windows header file,.  This is probalbly the reason for the failing VC testsuite of today.

PR #6704, opened 2022-06-23
Add an overload of `repair_polygon_soup()` for ranges of arrays

### LINKED ISSUE #6693: Problems using `std::array` as the value type of a `PolygonRange` in polygon soup related functionality
The documentation of CGAL 5.4.1 contains the following note for `PMP::polygon_mesh_to_polygon_soup()`:

> PolygonRange can also be a model of the concepts [RandomAccessContainer](https://doc.cgal.org/latest/Manual/classRandomAccessContainer.html) and [BackInsertionSequence](https://doc.cgal.org/latest/Manual/classBackInsertionSequence.html) whose value type is an array, but it is the user's responsability to ensure that all faces have the same number of vertices, and that this number is equal to the size of the array.

This is true and works as expected for `PMP::polygon_mesh_to_polygon_soup()`. However, this useful property in practice is not very useful, since other functions that operate on a polygon soup have stricter requirements on the value type of a `PolygonRange`. For example, `PMP::repair_polygon_soup()` requires that it is possible to construct a container with a pair of iterators, which `std::array` as an almost-but-not-quite-`SequenceContainer` does not support:

https://github.com/CGAL/cgal/blob/7ba81ca1d5d519d8f3691eeb9051f93933dd969d/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_polygon_soup.h#L275

I am wondering whether it would be possible for the functions that operate on a polygon soup to loosen their restrictions on the `PolygonRange` type such that `std::array` can be used consistently as a value type of a `PolygonRange` within this domain?

MaelRL:
## Summary of Changes

This overload does not call functions that are meaningless for arrays (changing the size and splitting facets).

## Release Management

* Affected package(s):  `PMP`
* Issue(s) solved (if any): fix #6693 

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_polygon_soup.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_polygon_soup.h
@@ -17,6 +17,7 @@
 #include <CGAL/Named_function_parameters.h>
 #include <CGAL/Polygon_mesh_processing/internal/named_params_helper.h>
 
+#include <CGAL/Container_helper.h>
 #include <CGAL/iterator.h>
 #include <CGAL/Kernel_traits.h>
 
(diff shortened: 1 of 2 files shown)


sloriot:

PR #6707, opened 2022-06-24
Fix forward declaration living in the wrong namespace

### LINKED ISSUE #6705: Bad Interval_nt forward declaration in 5.4.x-branch
https://github.com/CGAL/cgal/blob/017614c1f09020d90aac5a7b9837df41c7ae507a/Number_types/include/CGAL/NT_converter.h#L20-L22

Interval_nt is forward declared, but in the wrong namespace. This was fixed in #6120. Without backporting the whole PR, I wonder if we could backport just this tiny piece to 5.4.x-branch, moving the declaration a few lines below.

```c++
#ifndef NOBUG
#include <CGAL/NT_converter.h>
#endif
#include <CGAL/Epeck_d.h>

int main(){
  boost::multiprecision::mpq_rational q = 42;
  CGAL::NT_converter<boost::multiprecision::mpq_rational,CGAL::Interval_nt<>> c;
  c(q);
}
```
fails to compile with 5.4 and would be fixed by such a change.

sloriot:
Fix #6705

### DIFF
--- a/Bounding_volumes/test/Bounding_volumes/min_sphere_test.cpp
+++ b/Bounding_volumes/test/Bounding_volumes/min_sphere_test.cpp
@@ -33,10 +33,12 @@
 #include <CGAL/Random.h>
 #include <CGAL/Cartesian_d.h>
 #include <CGAL/Homogeneous_d.h>
-#include <sstream>
-#include <cassert>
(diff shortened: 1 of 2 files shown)


sloriot:

PR #6716, opened 2022-06-27
Parameterization: Make Iterative_authalic_parameterization faster (WIP)

### LINKED ISSUE #6694: Parallelization with SMP: Iterative Authalic Parameterization
Hello!
I'd like to know if it's possible to benefit from parallelization when running the Iterative_authalic_parameterizer_3 algorithm. I'm running it on a huge mesh (59M faces, 2.4G size) and it's taking a very long time (weeks), even on an HPC environment. Memory is not an issue, but running single-thread is. I have TBB support compiled-in, but I see no way of telling the algorithm to use it.
Any pointer in this direction would be appreciated.
Thanks!

afabri:
## Summary of Changes

I hard coded the copy of a sparse matrix.   We have to add this properly to a Solver API.
I removed debug output that writes the mesh after each iteration.

## Release Management

* Affected package(s): Surface_mesh_parameterization
* Issue(s) solved (if any): fix #6694

### DIFF
--- a/Solver_interface/doc/Solver_interface/Concepts/SparseLinearAlgebraTraits_d.h
+++ b/Solver_interface/doc/Solver_interface/Concepts/SparseLinearAlgebraTraits_d.h
@@ -127,6 +127,7 @@ NT& operator[](Index row);
 
 `SparseLinearAlgebraTraits_d::Matrix` is a concept of a sparse matrix class.
 
+\cgalRefines Assignable
 \cgalRefines DefaultConstructible
 
 \cgalHasModel `CGAL::Eigen_sparse_matrix<T>`
@@ -220,10 +221,6 @@ operator*(const NT& c, const Matrix& M);
 /// Sum of two matrices.
 friend Matrix
 operator+(const Matrix& M0, const Matrix& M1);
-
-/// Assignment operator
-friend Matrix
-operator=(const Matrix& M);
 
 /// @}
 
--- a/Solver_interface/doc/Solver_interface/Concepts/SparseLinearAlgebraTraits_d.h
+++ b/Solver_interface/doc/Solver_interface/Concepts/SparseLinearAlgebraTraits_d.h
@@ -221,6 +221,10 @@ operator*(const NT& c, const Matrix& M);
 friend Matrix
 operator+(const Matrix& M0, const Matrix& M1);
 
+/// Assignment operator
+friend Matrix
+operator=(const Matrix& M);
+
 /// @}
 
 }; /* end Matrix */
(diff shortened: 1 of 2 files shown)


afabri:
This PR requires that we add copying a matrix as operation, unless we think that matrices are `Assignable`.  Any opinion on this @sloriot ?

sloriot:
We can either update the concept or add a new one. Provided that Eigen is the only current model we have, checking if Eigen is fine with this and updating the concept is fine to me. MKL being in oneAPI now, we might want to re-add it as a model at some point.

sloriot:

lrineau:
This PR is still a draft. What should be done before it can be marked as ready?

afabri:
> This PR is still a draft. What should be done before it can be marked as ready?

We have to discuss if it makes sense to add to some solver package concepts that the matrix type is assignable.   @soesau can you  please take over as you are the maintainer of the solver package.

soesau:
I added an assignment operator to the concept SparseLinearAlgebraTraits_d. It seems that besides Surface_mesh_parameterization_3 also the package Heat_method_3 rely on the assignment operator of CGAL::Eigen_solver_traits.

There is another concept SparseLinearAlgebraWithFactorTraits_d which also uses Vector and Matrix types without specifying them. I guess we should add them both to the concept and including the assignment operator?

sloriot:
`SparseLinearAlgebraWithFactorTraits_d` refines `SparseLinearAlgebraTraits_d` so they are somehow "inherited"

Solver_interface/doc/Solver_interface/Concepts/SparseLinearAlgebraTraits_d.h
```diff
@@ -221,6 +221,10 @@ operator*(const NT& c, const Matrix& M);
 friend Matrix
 operator+(const Matrix& M0, const Matrix& M1);
 
+/// Assignment operator
```
afabri, line 224: I though we only wanted to write that the concept ` Matrix`  refines ` Assignable`.

soesau: I see, makes sense.
And the assignment operator alone is actually not sufficient as `Assignable` refines `CopyAssignable` and `CopyConstructible`.

sloriot:

@afabri you approve the changes?

afabri:
As it is my PR I cannot approve.

Surface_mesh_parameterization/include/CGAL/Surface_mesh_parameterization/Iterative_authalic_parameterizer_3.h
```diff
@@ -50,7 +50,7 @@
 #include <iomanip>
 #include <unordered_set>
 
-#define DEBUG_L0 1 // @fixme
+#define DEBUG_L0 0 // @fixme
```
lrineau, line 53: That macro name should be prefixed by `CGAL_`.

PR #6729, opened 2022-06-30
Update basic viewer to set camera mode earlier (following Efi issue)

### LINKED ISSUE #6722: Wrong distance computation of the camera caused by Basic_viewer_qt::draw()
## Issue Details

The problem is in the `draw()` function of the `Basic_viewer_qt`.
When using the `Basic_viewer_qt` construct to draw a 2D data structure, an orthogonal camera should be used.
The type of the camera is set in lines 1007 (orthogonal) and 1023 (perspective) as part of the `draw()` function.
However, this happens too late in the game, because the call `showEntireScene()` in line 1391 from the init() function of `Basic_viewer_qt` has already been occurred. This call triggers the call `camera->showEntireScene()` on line 3574 of `gqlviwer.h`, which in turn triggers the call `fitSphere(sceneCenter(), sceneRadius())` on line 907 (in `Camera.h`), which computes the distance based on the camera type. As the default type is perspective, the wrong distance is obtained (line 951).

The net effect is that the interesting area of the window is smaller than it should be. You can think of it has having too large margins.

The solution is simple, but I believe that the entire handling of the camera (the compound statement that starts with `if (is_two_dimensional())` in line 1005 of `Basic_viewer_qt.h`)  should not be part of the `draw()` function, because it should not be executed for every frame.

While at it, it will be nice to add a function to `Basic_viewer_qt` that the user can be used to obtain the width and height of the near plane. The rational for this request follows:

Consider rendering non-linear curves in the plane.
I need to approximate a given curve by a polyline in order to render it.
The approximation depends on two things, namely
(i) the window width and height in pixels, and
(ii) the near plane width and height.
I can obtain the former, no problem, with the `resizeGL()` callback.
Naturally, I can compute the bounding box of my scene.
What I'm missing is the margin of the bounding box used to define the camera.
Observe that I do not have the final camera parameters at hand when I generate the polyline approximation (well, these are based on the resulting approximated polylines...).

## Source Code

Polygon/examples/Polygon/draw_polygon.cpp

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): irrelevant
* Compiler: irrelevant
* Release or debug mode: both
* Specific flags used (if any): irrelevant
* CGAL version: latest (5.5)
* Boost version: irrelevant
* Other libraries versions if used (Eigen, TBB, etc.): irrelevant

gdamiand:

## Summary of Changes

Update the basic viewer to set camera mode in 2D earlier

TODO: please @efifogel could you test and tell me if this solves your bug?

## Release Management

* Issue(s) solved (if any): maybe fix #6722

### DIFF
--- a/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
+++ b/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
@@ -54,8 +54,12 @@
 #include <CGAL/Random.h>
 #include <CGAL/assertions.h>
 
+#define CGAL_BASIC_VIEWER_INIT_SIZE_X 500
+#define CGAL_BASIC_VIEWER_INIT_SIZE_Y 450
+
 namespace CGAL
 {
+
 //------------------------------------------------------------------------------
 inline CGAL::IO::Color get_random_color(CGAL::Random& random)
 {
@@ -202,7 +206,7 @@ public:
     else
       setWindowTitle(title);
 
-    resize(500, 450);
+    resize(CGAL_BASIC_VIEWER_INIT_SIZE_X, CGAL_BASIC_VIEWER_INIT_SIZE_X);
 
     if (inverse_normal)
     { negate_all_normals(); }
--- a/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
+++ b/GraphicsView/include/CGAL/Qt/Basic_viewer_qt.h
@@ -605,6 +605,7 @@ protected:
 
   void initialize_buffers()
   {
+    set_camera_mode();
     rendering_program_p_l.bind();
 
     // 1) POINT SHADER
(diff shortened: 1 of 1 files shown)


sloriot:
@gdamiand if you want this to end up in 5.5, you shoudl rebase the branch onto `5.5.x-branch`

gdamiand:
> @gdamiand if you want this to end up in 5.5, you shoudl rebase the branch onto `5.5.x-branch`

No matter. 5.5 or 5.6 is ok for me; choose the simpler for you.

sloriot:

efifogel:
Looks good. As far as I'm concerned the fix can be integrated.

I have a small request: Can you please use either enums or constants for the initial window width and height used in the new version of Basic_viewer_qt.h, Line 205: 
  resize(500, 450);
I need to mimic the computation of the pixel/GL-unit ratio before the window is constructed.
If you do, make sure to make them public.

gdamiand:
> Looks good. As far as I'm concerned the fix can be integrated.
> 
> I have a small request: Can you please use either enums or constants for the initial window width and height used in the new version of Basic_viewer_qt.h, Line 205: resize(500, 450); I need to mimic the computation of the pixel/GL-unit ratio before the window is constructed. If you do, make sure to make them public.

Done in https://github.com/CGAL/cgal/pull/6729/commits/e92373740e15e2cc2b71e311131d82f70c38a71b

sloriot:

PR #6733, opened 2022-07-01
sed -i -e 's/egrep/grep -E/g'

### LINKED ISSUE #6710: egrep -> grep -E
https://github.com/CGAL/cgal/blob/9c9139ba50daa19bab71ea4a6e07b1ec11754807/Scripts/scripts/cgal_create_cmake_script#L62

(picking one random occurence)
https://lists.gnu.org/archive/html/grep-commit/2021-08/msg00050.html
`egrep` and `fgrep` now produce a warning, it would be a good idea to replace them with `grep -E` / `grep -F`.

lrineau:
## Summary of Changes

Change `egrep` to `grep -E`. Fix #6710.

## Release Management

* Affected package(s): Scripts, Testsuite
* Issue(s) solved (if any): fix #6710


sloriot:

PR #6735, opened 2022-07-02
Fix eddy/bykat convex hulls sometimes returning extra points on the hull

### LINKED ISSUE #6723: Failed postcondition in ch_bykat
I am getting the following postcondition failure when hulling two circles (forming a sort of pill shape) via `convex_hull_2`:

```
/home/dimitris/source/gc/cgal/Convex_hull_2/include/CGAL/Convex_hull_2/ch_bykat_impl.h: 106: CGAL postcondition violation 'is_ccw_strongly_convex_2( res.output_so_far_begin(), res.output_so_far_end(), ch_traits)'
```

The failure occurred after a change in computation produced slightly different coordinates for the same geometry (the difference being that in the failing case, the points lie exactly on the circles, while in the former, succeeding case, they lie on circles up to floating-point accuracy).  It seems that, due to the nature of my input iterator, `convex_hull_2` delegates to `ch_bykat` which is the failing algorithm.  By contrast, using `std::vector`s as input, runs via the `ch_akl_toussaint` algorithm, which works fine with both sets of numbers.

I separated both failing and succeeding datasets into [a test case](https://github.com/CGAL/cgal/files/8999886/test.zip).  You can simply comment out the definition of `FAILING_CASE`, to switch between the "good" and "bad" point sets.

MaelRL:
## Summary of Changes

Fix eddy & bykat methods sometimes returning collinear vertices on the hull.
    
In these two algorithms, points farthest from a line must be found. In case of multiple points at the same distance, an "extremity" of this set of points must be returned.  This was properly documented in the Less_signed_distance_to_line_2 functor of the ConvexHullTraits_2 concept:

```
The predicate must provide a total order compatible with convexity, i.E., for any line segment \f$ s\f$ one of the endpoints of \f$ s\f$ is the smallest point among the points on \f$ s\f$, with respect to the order given by `Less_signed_distance_to_line_2`.
```

And specific models did respect this requirement. However, the kernel is supposed to also be a model of ConvexHullTraits_2, but Kernel_23::LessSignedDistanceToLine_2 does not require this, and the models (EPICK, Simple_Cartesian, etc.) do not implement it. Hence, if a kernel was used as geometric traits for 2D Convex Hull algorithms, one might not get the expected extreme point when there are multiple points at the same distance to a line at some point in the algorithm. The algorithm recovered, but this created extra points on the hull.

Fixing the functor in the kernel would create an inconsistency with `Kernel_23::Compare_signed_distance_to_line_2`, so instead this commit changes the concept and the code to require providing `Compare_signed_distance_to_line_2` and then uses `Less_xy` in the code in case of `EQUAL` return (that is actually how it was done in the `Convex_hull_2`-specific versions of the `ConvexHullTraits_2` model).

Also rework and complete all CH2 tests + some minor doc improvements.

## Release Management

* Affected package(s): `Convex_hull_23`
* Issue(s) solved (if any): fix #6723


sloriot:
@afabri if you know what was changed could you please update CHANGES.md?

sloriot:

PR #6737, opened 2022-07-04
Fix a bug in weight assignation?

sloriot:


sloriot:
@janetournois @lrineau do you think that there is no bug and we can close this PR?

lrineau:
I believe Jane has been testing the weighted images for a while without this patch.

janetournois:
To be sure of this decision, I just compared meshing with the same parameters, using
* master (dbf244c4db8a7ebb437f096ba5a49e25789e94b8) results in a mesh composed of 143503 cells, with surface looking like that
image

* master + this PR (cherry-pick aaba9ce59c276b0f96f9c9cf59cbdf6cd234e6ad) results in a mesh composed of 314905 cells, with surface looking like that (a lot smoother)
image

So I close this PR. Thank you @sloriot for the suggestion

PR #6752, opened 2022-07-13
Use boost mp by default

sloriot:
As we are early in the release cycle, we switch to boost mp NT by default.

### DIFF
--- a/Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
+++ b/Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
@@ -55,12 +55,11 @@ namespace CGAL { namespace internal {
 template < typename >
 struct Exact_field_selector
 
-#if ( (defined(CGAL_TEST_SUITE) && CGAL_VERSION_NR == 1050500900) || defined(CGAL_FORCE_USE_BOOST_MP))\
-    && BOOST_VERSION > 107800 && defined(CGAL_USE_BOOST_MP)
-// use boost-mp by default in the testsuite until 5.5-beta is out
+#if BOOST_VERSION > 107900 && defined(CGAL_USE_BOOST_MP)
+// use boost-mp by default
 // Boost
 { typedef BOOST_cpp_arithmetic_kernel::Rational Type; };
-#else // BOOST_VERSION > 107800
+#else // BOOST_VERSION > 107900
 #ifdef CGAL_USE_GMPXX
 { typedef mpq_class Type; };
 #elif defined(CGAL_USE_GMP)


sloriot:
The testsuite looks good but only one platform has a recent enough version of boost. All windows are using boost 1.66
I does not make sense to make this PR as tested if no windows test it IMO.

sloriot:
Note that despite @janetournois updates, only 2 platforms are testing this PR (`Ubuntu-GCC_master_CXX20-Release` and `MSVC-2022-Preview-Release` CGAL-5.6-Ic-70). Boost 1.80 is required.

Number_types/include/CGAL/Exact_integer.h
```diff
@@ -50,11 +50,10 @@ typedef unspecified_type Exact_integer;
 
 #else // not DOXYGEN_RUNNING
 
-#if ( (defined(CGAL_TEST_SUITE) && CGAL_VERSION_NR == 1050500900) || defined(CGAL_FORCE_USE_BOOST_MP))\
-    && BOOST_VERSION > 107800 && defined(CGAL_USE_BOOST_MP)
+#if BOOST_VERSION > 107900 && defined(CGAL_USE_BOOST_MP)
 // use boost-mp by default in the testsuite until 5.5-beta is out
```
MaelRL, line 54:
```suggestion
// use boost-mp by default
```

MaelRL:
The doc in `SDG_2` says: `If the GMP package is installed with CGAL, the template parameter EK has the default value: EK = Simple_cartesian<Gmpq>, otherwise its default value is EK = Simple_cartesian<Quotient<MP_Float> >.`, and I imagine there are other similar places.

Should `Exact_field_selector` be documented?

lrineau:
> Successfully tested in CGAL-5.6-Ic-79 but we are missing a test on MacOS (monet being down)

Let's wait a bit for more testing, then.

sloriot:
Successfully tested in CGAL-5.6-Ic-84 on 9 platforms (GNU/Linux, MS Windows and macOS)

lrineau:
Is there a way to check if ones local setup is using Boost MP?

lrineau:
# How to Use Boost Multiprecision with CGAL-5.6 (current `master`)

This PR "Use boost mp by default" has been merged October 2022. It requires Boost 1.80 or later (Boost 1.80 was published late August 2022, and Boost 1.81 mid-December 2022).

I did some experiment...

> Is there a way to check if ones local setup is using Boost MP?

I have installed 1.81 using Homebrew on Linux (because Fedora Linux is lagging with Boost 1.79 and, and so is vcpkg).

## `Display_type`
I have compiled a program with:
```c++
template<typename T>
struct Display_type;

Display_type<CGAL::Epick::Exact_nt> display_type;
```

so that the compiler displays an error message indicating the type `CGAL::Epick::Exact_nt`.

## Without any option

Without anything special, I get:

```
issue_129.cpp:28:37: error: implicit instantiation of undefined template 'Display_type<__gmp_expr<mpq_t, mpq_t>>'
Display_type<CGAL::Epick::Exact_nt> display_type;
                                    ^
```

## With `#undef CGAL_USE_GMPXX`

If I add
```c++
#undef CGAL_USE_GMPXX
```
to the program I get:
```
issue_129.cpp:28:37: error: implicit instantiation of undefined template 'Display_type<boost::multiprecision::number<boost::multiprecision::backends::rational_adaptor<boost::multiprecision::backends::cpp_int_backend<0, 0, boost::multiprecision::signed_magnitude, boost::multiprecision::unchecked, std::allocator<unsigned long long>>>, boost::multiprecision::et_on>>'
Display_type<CGAL::Epick::Exact_nt> display_type;
                                    ^
```

## Conclusion

Should not we add a new pull-request so that `CGAL_USE_GMPXX` is no longer defined by default (if `gmpxx.h` is detected)? That way, Boost MP will really be by default.

PR #6753, opened 2022-07-13
Add an example for default vector normal map

sloriot:
cc @hoskillua


sloriot:

PR #6762, opened 2022-07-19
licence -> license

sloriot:
Even if correct it is not consistent with paths and the rest of the lib.


sloriot:

PR #6767, opened 2022-07-21
Confusing line numbers in case of warnings

albert-github:
Based on the problems with the line numbers in some warnings, see review of https://github.com/CGAL/cgal/pull/6688 one of the problems was the `ALIASES` for figures, here we have the code `\latexonly ^^ \endlatexonly` though the code in the `\latexonly` part should be LaTeX commands. Here the "attempted" newline is not necessary at all so the `^^` can be removed leaving an empty `\latexonly` part that can be removed as well.

There are a number of other numbering problems found in https://github.com/CGAL/cgal/pull/6688 but they are on the doxygen side as a proposed pull request https://github.com/doxygen/doxygen/pull/9465 has been submitted.


afabri:
Is there a need to do the same for the  several other `Documentation/doc/resources/1.8.N/BaseDoxyfile.in`  ?

albert-github:
It definitely won't hurt. For the warning determination the 1.9.3 version of the `BaseDoxyfile.in` is the most important one.

albert-github:
Is there something to be done before setting this to "Under Testing"?

albert-github:
@MaelRL You labeled it "TODO", what should be do in your opinion?

PR #6773, opened 2022-07-28
PMP::tangential_relaxation() - avoid using global function barycenter()

janetournois:
## Summary of Changes

Use `construct_barycenter_3_object()` instead of `CGAL::barycenter()` and avoid including `CGAL/Kernel/global_functions_3.h`.

The old code (that was in `isotropic_remeshing`) had a compilation error with CGAL 5.4.1 and gcc 10.3, because the compiler could not differentiate between `CGAL::midpoint()` for `Point_2` and `CGAL::midpoint()` for `Point_3`.

This PR is related to #6774 and implements the "same" fix for the code modified for 5.5.

## Release Management

* Affected package(s): PMP


sloriot:

PR #6774, opened 2022-07-28
PMP::isotropic_remeshing() - avoid using global function midpoint()

janetournois:
## Summary of Changes

Use `construct_midpoint_3_object()` instead of `CGAL::midpoint()` and avoid including `CGAL/Kernel/global_functions_3.h`.

The old code had a compilation error with CGAL 5.4.1 and gcc 10.3, because the compiler could not differentiate between `CGAL::midpoint()` for `Point_2` and `CGAL::midpoint()` for `Point_3`.

Thanks @Simon-Lopez for the report.

## Release Management

* Affected package(s): PMP


lrineau:
@sloriot This PR cannot be tested in `integration`, because the patched file has been removed.

Should we merge the PR without testing and see the results on the release branch during the week-end?

sloriot:
I agree

PR #6782, opened 2022-08-02
Fix package license

sloriot:
The package is documented as GPL and even include the license checker but SPDX and package info was invalid.

https://github.com/CGAL/cgal/pull/3984 indicates that the package was intended to be GPL.


sloriot:
CI is happy with the change. Merging it.

PR #6796, opened 2022-08-11
Mesh_3 speedup (filter failures in `Compare_distance_3(Point_3, Segment_3, Point_3)` )

### LINKED ISSUE #6792: Performance degrade with `Compare_distance_3` (seen in Mesh_3)
## Issue Details

## Source Code

## Environment

janetournois:
There are more than 90% filter failures in calls to
`Compare_distance_3()(Point_3, Segment_3, Point_3)`, while called by the protecting balls placement algorithm.

This commit divides by 10 the initialization (protecting ball placement) time, at least on polyhedral surfaces input.

## Summary of Changes

I temporarily replaced the explicit kernel-calling code by a simple call to global functions, before we have a proper fix.

## Release Management

The full Mesh_3 `ctest` testsuite (examples and tests) passes, in Release and Debug mode, on my machine

* Affected package(s): Mesh_3 (and Kernel)

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_domain_with_polyline_features_3.h
+++ b/Mesh_3/include/CGAL/Mesh_domain_with_polyline_features_3.h
@@ -443,8 +443,9 @@ private:
                               const T1& obj1,
                               const T2& obj2) const
   {
-    return CGAL::compare(CGAL::squared_distance(p, obj1),
-                         CGAL::squared_distance(p, obj2));
+    typename Kernel::Compare_distance_3 compare_distance =
+      Kernel().compare_distance_3_object();
+    return compare_distance(p,obj1,obj2);
   }
 
 public:


MaelRL:
Also in temporary fix, could this define a filtered Compare_distance_3 predicate as it would be in the macro, but using the RT kernel? Since you know that the overload you call is division-free, it'll be fine. The advantage is that it'd remain exact whereas here it is no longer.

To do it systematically is more annoying... A way could be to template predicates by both K and K_RT, and then the operator() overload uses the correct kernel depending if there are divisions in the operator()?

sloriot:
There are not so many calls to `compare_distance` in `Mesh_domain_with_polyline_features_3.h`. It would be interested to understand which one are producing the filter failures and also see if it is fine to replace those calls with an approximate version.

janetournois:
The predicate that has all there filter failures is `Compare_distance_3(point, segment, point)`, with 54.435 / 59.184  failures/calls while running `mesh_polyhedral_domain_with_features_sm.cpp`.
Just note that the filter failures were already there before the commit ba2f2c1c20bd4812c128ac08e082b70f55ef6f18, though the meshing speed was a lot higher

lrineau:
With commit https://github.com/CGAL/cgal/pull/6796/commits/e7ffd0b943d79c377801ee9c32dcb909290a7ebe I have pushed a new patch, that keeps the calls to `Compare_distance_3`, but removes the source of filter failures.

<details>
Before my patch, the piece of code having the filter failures was in `Mesh_3/include/CGAL/Mesh_domain_with_polyline_features_3.h` in the member function `const_iterator locate(const Point_3& p, bool end_point_first=false) const` of the nested class `Polyline<K>`. Given a 3D polylines (that define on 1D-feature of the mesh domain), and a query point, the function returns the iterator to the source point of the segment nearest to the query point. If the nearest thing is one vertex of the polyline, it returns the iterator to the corresponding point. The filter failures were because, when current closest feature of the polyline is the target of the current segment, then the code lead to comparing `distance(query_point, segment)` with `distance(query_point, segment_target)`, that are equal. Commit https://github.com/CGAL/cgal/commit/ba2f2c1c20bd4812c128ac08e082b70f55ef6f18 adds an extra class to `Angle_3` to check if that is the case, to avoid the call to `Compare_distance_3` that would fail.
</details>

sloriot:

PR #6801, opened 2022-08-17
Divider between treeview and text

albert-github:
In doxygen the setting `HTML_COLORSTYLE` has been introduced with as default `AUTO_LIGHT` but due to the current strategy of CGAL in respect to HTML stylesheets the divider between the treeview and the text is not properly  visible anymore. Due to the same strategy the setting to `AUTO_LIGHT` does not make sense for CGAL.
When setting `HTML_COLORSTYLE=LIGHT` the divider is properly visible again.

### DIFF
--- a/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
@@ -1308,6 +1308,23 @@ HTML_STYLESHEET        = ${CGAL_DOC_RESOURCE_DIR}/stylesheet.css
 
 HTML_EXTRA_STYLESHEET  =
 
+# The HTML_COLORSTYLE tag can be used to specify if the generated HTML output
+# should be rendered with a dark or light theme. Default setting AUTO_LIGHT
+# enables light output unless the user preference is dark output. Other options
+# are DARK to always use dark mode, LIGHT to always use light mode, AUTO_DARK to
+# default to dark mode unless the user prefers light mode, and TOGGLE to let the
+# user toggle between dark and light mode via a button.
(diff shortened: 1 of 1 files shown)


afabri:
As this PR was in [this testsuite](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGALGAL--5.6-Ic-53/) you can check if the result looks like you expect and then mark here "Succcesfully tested .in CGALGAL--5.6-Ic-53".

albert-github:
@afabri was your remark directed at me or at @sloriot ?
The link you provided doesn't work, but I know a bit around in the results of the test suite though.
The result of the test in CGALGAL--5.6-Ic-53 look OK to me (but I already tested locally before submitting the PR).
I would be a bit strange for me to say it is tested successfully as I submitted the PR and this would be like in the proverb "a bit like having a butcher inspecting the quality of his own meat."

Note: PR might also be necessary to be backported to the older distributions that ysed the doxygen master (see e.g. CGAL-5.4.3-I-173)

afabri:
If your visual inspection shows that it IS fixed I would say you can report.   I personally have no idea what glitch to look for.
Compared to a local test, the doxygen testsuite shows what to expect in the official documentation.

albert-github:
@afabri you see the difference when you compare e.g. the CGALGAL--5.6-Ic-53 and the CGAL-5.4.3-I-173 testsuite (any file) and look at the divider between the treeview and the text.
**CGAL-5.4.3-I-173**

image

Succcesfully tested in CGALGAL--5.6-Ic-53

**CGALGAL--5.6-Ic-53**

image

sloriot:
Following Albert recommendation: Successfully tested in CGAL-5.6-Ic-53

PR #6802, opened 2022-08-17
PSP_3: Fix test_read_write_point_set

lrineau:
## Summary of Changes

The LAS format, even binary, is not an exact representation. The coordinates are encoded using and offset and a scaling, for each coordinate.

In `write_LAS_with_properties()`, Simon has hard-coded:
```c++
  LASheader header;
  header.x_scale_factor = 1e-9 * (bbox.xmax() - bbox.xmin());
  header.y_scale_factor = 1e-9 * (bbox.ymax() - bbox.ymin());
  header.z_scale_factor = 1e-9 * (bbox.zmax() - bbox.zmin());
  header.x_offset = bbox.xmin();
  header.y_offset = bbox.ymin();
  header.z_offset = bbox.zmin();
```
So, the approximate comparison of coordinates, for the I/O test, should compare using about the same precision.

## Release Management

* Affected package(s): Point_set_processing_3

Cc: @soesau @afabri 

That should fix the error of `Point_set_processing_3/test/Point_set_processing_3/test_read_write_point_set.cpp` here: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-42/Polyhedron_Demo/TestReport_gimeno_ArchLinux-clang.gz

The bug is in the test file, and not in CGAL.

### DIFF
--- a/Point_set_processing_3/test/Point_set_processing_3/test_read_write_point_set.cpp
+++ b/Point_set_processing_3/test/Point_set_processing_3/test_read_write_point_set.cpp
@@ -19,24 +19,12 @@ typedef CGAL::Simple_cartesian<double> Kernel;
 typedef Kernel::Point_3 Point_3;
 typedef Kernel::Vector_3 Vector_3;
 
-const double epsilon = 5e-4;
+const double precision = 1e-8; // relates to 1e9 in write_LAS_with_properties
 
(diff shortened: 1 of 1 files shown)


sloriot:
Successfully tested in CGAL-5.6-Ic-53

PR #6825, opened 2022-09-02
CGAL kernel: Introduce CGAL_Kernel_pred_RT_or_FT

### LINKED ISSUE #6792: Performance degrade with `Compare_distance_3` (seen in Mesh_3)

lrineau:
## Summary of Changes

This pull-request introduces a new kind of predicate in `<CGAL/Kernel/interface_macros.h>`. In addition to
  - `CGAL_kernel_pred` for predicates,
  - `CGAL_Kernel_pred_RT` for predicates that can be implemented using aring-type,

now there is also:
  - `CGAL_Kernel_pred_RT_or_FT` for predicates with multiple overloads of `operator()`, some needing a field type and other needing a ring type (without the division operator).

The C++ code can discriminate between the two cases with a special wrapper for the return type: `CGAL::Needs_FT<result_type` instead of `result_type` (defined in `<CGAL/tags.h>`.

In `<CGAL/Filtered_predicate.h>`, in addition to the usual class template `Filtered_predicate`, there is now also `Filtered_predicate_RT_FT` that takes three predicates as template parameters instead of two:
  - the exact predicate with an ring-type,
  - the exact predicate with a field-type,
  - the approximate predicate (with `Interval_nt` as number-type).

For the moment, only `Compare_distance_3` in `<CGAL/Cartesian/function_objects.h>` is using the new `Filtered_predicate_RT_FT`.

Before this commit, the file `Kernel_23/test/Kernel_23/include/CGAL/_test_new_3.h` was testing `Compare_distance_3` only with three points or for points. This commit adds:
  - a test with `Point_3, Point_3, Segment_3`, and
  - a test with `Line_3, Point_3, Point_3`, that actually needs a field type with its current implementation.

In the test `Kernel_23/test/Kernel_23/Filtered_cartesian.cpp`, the macro `CGAL_NO_MPZF_DIVISION_OPERATOR` is defined, to remove the division operator from `CGAL::Mpzf`. `CGAL::Mpzf` is a ring-type, even with its `operator/` (because that `operator/` can only compute exact divisions), but with `CGAL_NO_MPZF_DIVISION_OPERATOR` defined, that is now checked by the compiler.

## Release Management

**Please do not merge as it is: I have used C++17 `if constexpr` on purpose during my experimentation, and will revert to C++14 with tag dispatching before the pull-request can be merged.**

* Affected package(s): CGAL kernels
* Feature/Small Feature (if any): no, because that is an undocumented implementation detail of CGAL kernels

### DIFF
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -2506,7 +2506,7 @@ namespace CartesianKernelFunctors {
       FT rsy = psz*qsx-psx*qsz;
       FT rsz = psx*qsy-psy*qsx;
 
-      // The following determinants can be developped and simplified.
+      // The following determinants can be developed and simplified.
       //
       // FT num_x = determinant(psy,psz,ps2,
       //                              qsy,qsz,qs2,
--- a/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
+++ b/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h
@@ -17,6 +17,7 @@
 #ifndef CGAL_CARTESIAN_FUNCTION_OBJECTS_H
 #define CGAL_CARTESIAN_FUNCTION_OBJECTS_H
 
+#include <CGAL/tags.h>
 #include <CGAL/Kernel/function_objects.h>
 #include <CGAL/predicates/kernel_ftC2.h>
 #include <CGAL/predicates/kernel_ftC3.h>
@@ -591,14 +592,14 @@ namespace CartesianKernelFunctors {
     }
 
     template <class T1, class T2, class T3>
-    result_type
+    Needs_FT<result_type>
     operator()(const T1& p, const T2& q, const T3& r) const
     {
       return CGAL::compare(squared_distance(p, q), squared_distance(p, r));
     }
 
     template <class T1, class T2, class T3, class T4>
-    result_type
+    Needs_FT<result_type>
     operator()(const T1& p, const T2& q, const T3& r, const T4& s) const
     {
       return CGAL::compare(squared_distance(p, q), squared_distance(r, s));
--- a/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
+++ b/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
@@ -754,7 +754,7 @@ power_side_of_bounded_power_sphereC3(
 }
 
 // return the sign of the power test of weighted point (rx,ry,rz,rw)
- // with respect to the smallest sphere orthogoanal to
+ // with respect to the smallest sphere orthogonal to
 // p,q
 template< class FT >
 typename Same_uncertainty_nt<Bounded_side, FT>::type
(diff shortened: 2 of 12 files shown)


lrineau:
@mglisse In particular, I am worried if the return type of a new overloads of `CGAL::CartesianKernelFunctors::Compare_distance_3::operator()` being `Needs_FT<Comparison_result>` instead of `Comparison_result`, if that can confuse thinks like:
  - expression templates in gmpxx (but I think not),
  - or implementation details of `CGAL::Lazy_kernel` (and for the later... I do not know).

lrineau:
> or implementation details of `CGAL::Lazy_kernel` (and for the later... I do not know).

I think that is OK, because in `Lazy_kernel_base`, predicates are instances of `Filtered_predicate`, and its `operator()` hardcodes `return_type` as the return type (and not `auto` or `decltype(auto)`, with that definition:
```c++
  typedef typename EP::result_type  result_type;
```

So all `Needs_FT<T>` are forcibly casted back to `T`.

And same for `Static_filtered_predicate` used as the predicate class of `Lazy_kernel_generic_base`.

MaelRL:
* Could all three Filtered_predicate  classes / macros be merged into the generic version? We would potentially carry useless RT kernels if it is using FT, but that's less code and it reduces the number of things to change (e.g. I have a RT-requiring functor using `CGAL_Kernel_pred_RT`, I add a new FT-compatible overload, I have to also remember to change the macro).

* Maybe a tag as parameter instead of wrapping the result type such that the result type is the actual result type and doesn't require removing the `Needs_FT` (which should show up if you use the decltype() and not `result_type`) ? Something like
```
struct Predicate
{
  K::Boolean operator()(...parameters..., CGAL_kernel_tag::FT_required); // possibly optional
  K::Boolean operator()(...parameters..., CGAL_kernel_tag::RT_sufficient);
};
```
and then `Filtered_predicate` extracts the type of the last argument in the package (and if it doesn't find it, then it's `FT_required`).

* Shouldn't it require FT by default and RT only when specified? That'd be fewer changes, and less code.

lrineau:
## Side note ##

Please do not merge as it is: I have used C++17 `if constexpr` on purpose during my experimentation, and will revert to C++14 with tag dispatching before the pull-request can be merged.

*I have modified the top message of this pull-request to add that information.*

lrineau:
Dear reviewers (I expect @sloriot, @MaelRL, @mglisse to pay attention). I have refreshed this pull-request with three commits:
  - one about spelling typo, unimportant (https://github.com/CGAL/cgal/pull/6825/commits/b114789abf5ede7053b3fcf2f3795253504d159b),
  - one that refactors the condition of the `if constexpr` that switch between RT and FT (https://github.com/CGAL/cgal/pull/6825/commits/81410701f788a648dd379941536faa6ac062ba47),
  - and another one adding an important comment for future developers wanting to modify `Filtered_predicate_RT_FT` (https://github.com/CGAL/cgal/pull/6825/commits/1e485113e8a90685519d09d5dbeece4ccb7b4e86).

I have not modified the proposal, in this pull-request.

I have added a second pull-request draft that addresses parts of [Mael's comments](https://github.com/CGAL/cgal/pull/6825#issuecomment-1235353390): the new pull-request is https://github.com/CGAL/cgal/pull/6836.

lrineau:
If you want to read the diff between the two proposal (and not the diff with `master`), the difference between the two pull-requests, so far, is only the commit https://github.com/CGAL/cgal/pull/6836/commits/4b660d9ec9f6a2f7634ff24a0f873a0548af2248.

MaelRL:
Replaced by #6866.

PR #6827, opened 2022-09-03
AW3 Plugin: Show both absolute and relative values for alpha & offset

### LINKED ISSUE #6805: Polyhdron demo: alpha wrapping plugin
## Issue Details

Additionally to the ratio of the bbox one should show what the alpha value will be, and the alpha should also be changeable (and then the ratio value being updated). Same for the offset.   Espacially as the name of the item carries the alpha value.

MaelRL:
## Summary of Changes

The main change is the change from one box (value) + one check box (relative or not) to always showing absolute and relative values.

Here is the (proposed) new window:

Screenshot from 2022-09-03 14-09-35

Some other improvements along the way.

## Release Management

* Affected package(s): `Alpha_wrap_3`, `Polyhedron`

* Issue(s) solved (if any): fix #6805


sloriot:

Issue #6783, opened 2022-08-02
PMP:   compute_face_normal() of degenerate faces


afabri:
## Issue Details

In [this testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-38/Polygon_mesh_processing/TestReport_gimeno_ArchLinux-clang-Release.gz) we get an assertion that the normal of a degenerate face is not the null vector.
It seems that the algorithm does not check if a face is degenerate but computes something.

At the same time the platform has several errors for other packages, comparing numerical results.  

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux 
* Compiler: Clang 14.0.6
* Release or debug mode:
* Specific flags used (if any):
* CGAL version:  master
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

janetournois:
IIRC we had discussed with @MaelRL that it makes no sense to return NULL_VECTOR as the normal of a degenerate face.
But the test seems to think the other way around...

afabri:
But what would be the better choice?  Also not only that the test thinks it, but  it is only one platform where this triggers an assertion.

soesau:
The problem is that for certain doubles a * b - b * a does not equal 0.
The cross product does exactly that in the case of identical vectors.
As Laurent pointed out, intermediate results may be stored as a double or in a FPU register with higher precision. Storing both products explicitly in a double resolves the problem.

PR #6831, opened 2022-09-05
Skipping degenerate tris in normal computation

soesau:
## Summary of Changes

Added a check for co-linear vertices before calculating normals.

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): #6783

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/compute_normal.h
@@ -102,14 +102,18 @@ void sum_normals(const PM& pmesh,
     const Point_ref pvn = get(vpmap, the);
     const Point_ref pvnn = get(vpmap, tnhe);
 
-    const Vector n = internal::triangle_normal(pv, pvn, pvnn, traits);
+    // Check whether the three vertices are collinear and skip normal computation due to numerical stability.
+    if (!traits.collinear_3_object()(pv, pvn, pvnn))
+    {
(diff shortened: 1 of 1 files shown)


lrineau:
@MaelRL @sloriot @soesau I wonder what will be the runtime impact.

- For a mesh with degenerate triangles, this pull-request will add filter failures.
- Even for a mesh without degenerate triangles, the code will be a bit slower, but how much?

I have the feeling that the current code (without this pull-request) is actually correct (and faster), and that the assertions should be modified.

soesau:
I agree, it will certainly slow down the usually rather light computation.

After calculating the normal for each face the result is compared to the null vector and normalized afterwards. Thus, the very short normal is rescaled.

sloriot:
If the bug you are trying to fix is the division by 0 when doing the normalization, then we can simply check that the normalization is not done using 0.

MaelRL:
If you were to call `internal::triangle_normal()` (rather than `PMP::compute_face_normal()`), you'd currently still get the same issue. So, if the idea is to add a call to `Collinear_3` to ensure that a null vector is returned if the face is degenerate, that check should probably get down a level into the internal function.

On whether this should be done or not, this is the usual (almost philosophical) problem of whether constructions should be written to be as fast as possible, or as precise as possible when using a kernel not providing exact constructions. My opinion is that the kernel should be as fast as possible, and it's for the user to add extra checks if he decides to use an inexact construction kernel while still caring about having null normals for his degenerate faces.

soesau:
Created new pull request which restricts certain tests to epeck #6880

PR #6833, opened 2022-09-06
Python is mandatory not optional to build the doc

sloriot:


sloriot:

PR #6835, opened 2022-09-06
Some CMakeLists.txts cleaning

MaelRL:
## Summary of Changes

- Make it so `message()` calls are more consistent: if it's non-essential, it's a `STATUS`; if it's important, it's a `NOTICE`.
- Structure the usage of `QUIET` and `REQUIRED`, following this:
	- CGAL: always `REQUIRED`
	- CGAL component: `COMPONENT` / `OPTIONAL_COMPONENT` (and never `QUIET`)
	- *required* 3rd party: `QUIET` + `message(NOTICE "")` on NOT_FOUND
	- *optional* 3rd party: `QUIET` + `message(STATUS "")` on NOT_FOUND, except for Polyhedron/demo (not `QUIET`)
	- 3rd party components: `COMPONENT` / `OPTIONAL_COMPONENT` and `QUIET` depending on required or optional 3rd party
- Various cleaning + tiny bug fixes (see commits)

## Release Management

* Affected package(s): All
* Issue(s) solved (if any): - 


MaelRL:
@lrineau Could you please clarify what's the best practice for `PRIVATE`/`PUBLIC` on linking / compilation macros definition? Reading at https://cmake.org/pipermail/cmake/2016-May/063400.html, I don't understand why some 3rd party such as `TBB`, `METIS`, `Eigen`, ... are always `PUBLIC` whereas the rest is `PRIVATE`.

Documentation/doc/Documentation/Developer_manual/create_and_use_a_cmakelist.txt
```diff
@@ -14,7 +14,7 @@ target_link_libraries(my_executable CGAL::CGAL)
 Other \cgal libraries are linked similarly. For example, with `CGAL_Core`:
 
 \code
-find_package(CGAL REQUIRED COMPONENTS Core)
+find_package(CGAL COMPONENTS Core)
```
sloriot, line 17: why that change?

lrineau:
> @lrineau Could you please clarify what's the best practice for `PRIVATE`/`PUBLIC` on linking / compilation macros definition? Reading [cmake.org/pipermail/cmake/2016-May/063400.html](https://cmake.org/pipermail/cmake/2016-May/063400.html), I don't understand why some 3rd party such as `TBB`, `METIS`, `Eigen`, ... are always `PUBLIC` whereas the rest is `PRIVATE`.

## For CGAL
As in CGAL we have only executables (and not libraries), that does not matter at all. I recommend not to change anything about that, because that really does not matter. Please avoir a change like https://github.com/CGAL/cgal/pull/6835/commits/e55d9274fa57d817d1dd7e75c4e284989eec1fb7#diff-87475bd5e5c9b63205f73198f0d5e1dc46c55ccf2e1ec0b279e229bc1eb35f66R102 where you used both `PUBLIC` and `PRIVATE` for the same target.

## The rational
For a library, the rational is:
  - `PRIVATE` for the dependencies of the `.cpp` files of the library, or private header files,
  - `PUBLIC` for the dependencies of the public header files of the library.

For example, if a library `libfoo` has a `.cpp` that uses Boost headers, or Boost libraries, then the `PRIVATE` linking with Boost headers is sufficient, but if the same library has a public header `<foo.h>`, part of its documented API, that includes `<boost/variant.hpp>`, then the `PUBLIC` linking with Boost headers is required, so that consumers of `libfoo` are also linked with Boost headers automatically.

## For the 3D demo in `demo/Polyhedron`

For the "polyhedron demo", we have to be careful because we actually have libraries that are defined in the CMake files.

Documentation/doc/Documentation/Developer_manual/create_and_use_a_cmakelist.txt
MaelRL: It's documentation on using CGAL in an external program so it is that external program's responsability to use whatever flag they want. And the other `find_package(CGAL)` just a couple lines above did that too by not having `REQUIRED`.

lrineau: I agree with Mael. See https://doc.cgal.org/5.5/Manual/devman_create_and_use_a_cmakelist.html:
  - either we remove `REQUIRED` in the second occurrence,
  - or we add it as well in the first occurrence.

The current situation can confuse readers and lead them to think that `REQUIRED` and `COMPONENTS` are linked.

lrineau: Actually, I would add `REQUIRED` everywhere in that documentation page. There is a third occurrence a bit further in the page, and it has `REQUIRED`.

sloriot:
`Point_set_3`, `Point_set_processing_3` tests+examples, `Solver_interface_Examples`, `Surface_mesh`, `T3_examples` are having a lot of `r` in CGAL-5.6-Ic-66

[This one](https://cgal.geometryfactory.com/CGAL/Members/testsuite/CGAL-5.6-Ic-66/Core_Examples/TestReport_lrineau_Fedora-with-LEDA.gz) is new too.

MaelRL:
Some missing "missing 3rd party dependencies" messages have been added (e.g. in T3), so this is expected, but I will check them individually.

sloriot:
I guess this PR is responsible for warnings in `Polygonal_surface_reconstruction_Examples` in CGAL-5.6-Ic-69

Arithmetic_kernel/test/Arithmetic_kernel/CMakeLists.txt
```diff
@@ -27,16 +27,13 @@ if(GMP_FOUND)
     # version needs GMP>=4.2, so we require this dependency only here and
     # not in FindMPFI.cmake
     if(_IS_GMP_VERSION_TO_LOW)
-      message(
-        STATUS
-          "MPFI tests need GMP>=4.2, some of the tests will not be compiled")
+      message(STATUS "NOTICE: MPFI tests need GMP>=4.2, some of the tests will not be compiled")
```
albert-github, line 30: According to the description of the PR:
>Make it so message() calls are more consistent: if it's non-essential, it's a STATUS; if it's important, it's a NOTICE.

But what is this `STATUS` type with text `NOTICE`?

Algebraic_kernel_d/examples/Algebraic_kernel_d/CMakeLists.txt
```diff
@@ -4,15 +4,13 @@ project(Algebraic_kernel_d_Examples)
...
   include(${MPFI_USE_FILE})
-  include(CGAL_VersionUtils)
   create_single_source_cgal_program("Compare_1.cpp")
   create_single_source_cgal_program("Construct_algebraic_real_1.cpp")
   create_single_source_cgal_program("Isolate_1.cpp")
   create_single_source_cgal_program("Sign_at_1.cpp")
   create_single_source_cgal_program("Solve_1.cpp")
 else()
-    message(STATUS "This program requires the CGAL, CGAL_Core and MPFI libraries, and will not be compiled.")
+  message("NOTICE: This project requires the MPFI library and GMP support, and will not be compiled.")
```
albert-github, line 15: According to the CMake documentation (https://cmake.org/cmake/help/latest/command/message.html):

>(none) or NOTICE
>
>    Important message printed to stderr to attract user's attention.

So isn't this a bit double?

albert-github (commented):
Didn't do a full check but just 2 things that caught my eye.

Algebraic_kernel_d/examples/Algebraic_kernel_d/CMakeLists.txt
MaelRL: `NOTICE` is not yet available in the oldest CMake version still supported, hence no `NOTICE`. I have put "NOTICE:" like this so this is easier to update these "none" occurrences once the minimal version is upgraded.

The other point is that it prints something in capital letters, in case your stderr is not showing up in red.

Arithmetic_kernel/test/Arithmetic_kernel/CMakeLists.txt
MaelRL: It indicates that a third party dependency is missing, but this dependency is only required for *some* of the targets. The `NOTICE:` is there to make it more visible since CMake can sometimes dump a lot of info.

Algebraic_kernel_d/examples/Algebraic_kernel_d/CMakeLists.txt
albert-github: Valid argument, in the documentation is says (as reference):
> New in version 3.15: Added the NOTICE, VERBOSE, DEBUG, and TRACE levels.

sloriot:
This PR was marked as tested at the beginning of Decembre

> @lrineau why "Not yet approved"?

@lrineau Can this PR be tested again or are you still blocking its integration?

lrineau (approved):
I already discussed the specific of the choices made in this, with PR.

Before it can be merged, I wanted to have a look at the "results". I compared the CMake log of a full configuration of CGAL (with demos, examples, and tests). The results are pretty good. The CMake log is no longer cluttered with messages about required libraries (Boost and Eigen) found. For example:

```patch
@@ -60,17 +60,17 @@
 -- Configuring examples in examples/Arrangement_on_surface_2
 
 -- Configuring examples in examples/BGL_LCC
--- Found Boost: /home/linuxbrew/.linuxbrew/lib/cmake/Boost-1.81.0/BoostConfig.cmake (found version "1.81.0")  
 
 -- Configuring examples in examples/BGL_OpenMesh
--- NOTICE: These examples require OpenMesh and will not be compiled.
+-- Could NOT find OpenMesh (missing: OPENMESH_INCLUDE_DIR OPENMESH_LIBRARIES) 
+NOTICE: This project requires OpenMesh and will not be compiled.
 
 -- Configuring examples in examples/BGL_arrangement_2
 
 -- Configuring examples in examples/BGL_graphcut
 
 -- Configuring examples in examples/BGL_polyhedron_3
--- Examples that use OpenMesh will not be compiled.
+-- NOTICE: The example 'copy_polyhedron' will not use OpenMesh.
 
 -- Configuring examples in examples/BGL_surface_mesh
```

```patch
@@ -139,7 +138,6 @@
 -- Configuring examples in examples/HalfedgeDS
 
 -- Configuring examples in examples/Heat_method_3
--- Found Eigen3: /usr/include/eigen3 (found suitable version "3.4.0", minimum required is "3.3.0") 
 
 -- Configuring examples in examples/Hyperbolic_triangulation_2
 
@@ -150,7 +148,6 @@
 -- Configuring examples in examples/Interval_skip_list
 
 -- Configuring examples in examples/Jet_fitting_3
--- Found Eigen3: /usr/include/eigen3 (found suitable version "3.4.0", minimum required is "3.1.0") 
 
 -- Configuring examples in examples/Kernel_23
```

The downside is that maybe we should have somewhere in CGAL where Eigen is search non-quietly, so that the testsuite can show which version is used.

Anyway, I approved this PR, and would like to see it tested again in `integration`, before it can be merged.

sloriot:

PR #6836, opened 2022-09-07
CGAL kernel: second proposal for CGAL_Kernel_pred_RT_or_FT

### LINKED ISSUE #6792: Performance degrade with `Compare_distance_3` (seen in Mesh_3)

lrineau:
## Summary of Changes

This pull-request introduces a new kind of predicate in `<CGAL/Kernel/interface_macros.h>`. In addition to
  - `CGAL_kernel_pred` for predicates,
  - `CGAL_Kernel_pred_RT` for predicates that can be implemented using aring-type,

now there is also:
  - `CGAL_Kernel_pred_RT_or_FT` for predicates with multiple overloads of `operator()`, some needing a field type and other needing a ring type (without the division operator).

The C++ code can discriminate between the two cases with a tag as the last argument: if the last argument of the call operator is `CGAL::RT_sufficient`, then the exact predicate call operator will be compiled with a ring-type, otherwise it will be compiled using a field-type.

In `<CGAL/Filtered_predicate.h>`, in addition to the usual class template `Filtered_predicate`, there is now also `Filtered_predicate_RT_FT` that takes three predicates as template parameters instead of two:
  - the exact predicate with an ring-type,
  - the exact predicate with a field-type,
  - the approximate predicate (with `Interval_nt` as number-type).

For the moment, only `Compare_distance_3` in `<CGAL/Cartesian/function_objects.h>` is using the new `Filtered_predicate_RT_FT`.

Before this commit, the file `Kernel_23/test/Kernel_23/include/CGAL/_test_new_3.h` was testing `Compare_distance_3` only with three points or for points. This commit adds:
  - a test with `Point_3, Point_3, Segment_3`, and
  - a test with `Line_3, Point_3, Point_3`, that actually needs a field type with its current implementation.

In the test `Kernel_23/test/Kernel_23/Filtered_cartesian.cpp`, the macro `CGAL_NO_MPZF_DIVISION_OPERATOR` is defined, to remove the division operator from `CGAL::Mpzf`. `CGAL::Mpzf` is a ring-type, even with its `operator/` (because that `operator/` can only compute exact divisions), but with `CGAL_NO_MPZF_DIVISION_OPERATOR` defined, that is now checked by the compiler.

## Release Management

**Please do not merge as it is: I have used C++17 `if constexpr` on purpose during my experimentation, and will revert to C++14 with tag dispatching before the pull-request can be merged.**

* Affected package(s): CGAL kernels
* Feature/Small Feature (if any): no, because that is an undocumented implementation detail of CGAL kernels

### DIFF
--- a/Filtered_kernel/include/CGAL/Filtered_predicate.h
+++ b/Filtered_kernel/include/CGAL/Filtered_predicate.h
@@ -19,6 +19,8 @@
 #include <CGAL/Uncertain.h>
 #include <CGAL/Profile_counter.h>
 
+#include <type_traits>
+
 namespace CGAL {
 
 // This template class is a wrapper that implements the filtering for any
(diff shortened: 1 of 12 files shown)


Filtered_kernel/include/CGAL/Filtered_predicate.h
```diff
@@ -111,6 +113,77 @@ Filtered_predicate<EP,AP,C2E,C2A,Protection>::
...
+
+  // ## Important note
+  //
+  // If you want to remove of rename that member function template `needs_ft`,
+  // please also change the lines with
+  // `CGAL_GENERATE_MEMBER_DETECTOR(needs_ft);`
+  // or `has_needs_ft<typename R::Compare_distance_3>` in
+  // the file `Kernel_23/test/Kernel_23/include/CGAL/_test_new_3.h`.
+  template <typename... Args>
+  bool needs_ft(const Args&...) const {
```
MaelRL, line 156: constexpr?

MaelRL:
I don't have any other remark on the core (except for that QOL point about merging the `Filtered_predicate_*` macros as long as it's not a slowdown)

Filtered_kernel/include/CGAL/Filtered_predicate.h
lrineau: We can add `constexpr`, and I will... but I do not think that will be useful, or used.

That `constexpr` qualifier would be useful for something like:
```c++
  constexpr b = compare_distance_functor.needs_ft(arg1, arg2, arg3);
```
**if** both `compare_distance_functor` and all three arguments are constant expressions.

I could also make that function `static`, to remove the implicit `this` argument, but I would need another version of `has_needs_ft` for the test in `Kernel_23/test/Kernel_23/include/CGAL/_test_new_3.h`. Given that this function template `needs_ft` is only used for a few statements in a test, I do not want to work more on it.

lrineau: Done in https://github.com/CGAL/cgal/pull/6836/commits/8521c44c6f7ec9bb0096b3b74af4547ef41bf1c4.

MaelRL:
Replaced by #6866.

PR #6844, opened 2022-09-13
Complete array-based polygon soup repair

MaelRL:
## Summary of Changes

Follow-up of #6704: in the original pipeline, the "split pinched" + "remove consecutive" will make it so invalid faces have size `< 3`. For array-based soups, we cannot apply these functions since they modify the size, but we still need to remove those invalid faces.

## Release Management

* Issue(s) solved (if any): -


MaelRL:
> /build:v1

There's no documentation change, it's internal (`//` not `///`) documentation. It made me notice some errors on them though, so thanks!

PR #6845, opened 2022-09-13
PMP corefinement: Handle inconsistency with open mesh as input

sloriot:
Return false for such a case:
Screenshot from 2022-09-08 19-05-31


Polygon_mesh_processing/test/Polygon_mesh_processing/test_corefinement_bool_op.cmd
```diff
@@ -1 +1,2 @@
 ${CGAL_DATA_DIR}/meshes/elephant.off ${CGAL_DATA_DIR}/meshes/sphere.off ALL 1 1 1 1
+${CGAL_DATA_DIR}/meshes/open_cube.off data-coref/incompatible_with_open_cube.off ALL 0 0 0 0
```
lrineau, line 2: Does that work correctly to have multiple lines in `.cmd` files? Even with ctest?

sloriot: the test handle them correctly (there is a for loop taking argc by tuples)

lrineau: I see that ctest is launching:
> .../test/Polygon_mesh_processing/test_corefinement_bool_op "/home/lrineau/Git/cgal/Data/data/meshes/elephant.off" "/home/lrineau/Git/cgal/Data/data/meshes/sphere.off" "ALL" "1" "1" "1" "1" "/home/lrineau/Git/cgal/Data/data/meshes/open_cube.off" "data-coref/incompatible_with_open_cube.off" "ALL" "0" "0" "0" "0"

but it does not seems that the `main` of the test can handle 16 arguments.

sloriot: You are right, this mechanism is implemented only in test_corefine.

sloriot:

PR #6854, opened 2022-09-19
fix typo and update documentation

zfb132:
## Summary of Changes

Fix typo in `Documentation/doc/Documentation/advanced/Configuration_variables.txt`

### DIFF
--- a/Documentation/doc/Documentation/advanced/Configuration_variables.txt
+++ b/Documentation/doc/Documentation/advanced/Configuration_variables.txt
@@ -72,12 +72,12 @@ and will serverly limit performances.
 
 | Variable              | Description                                                           | Type        | %Default Value |
 | :-                    | :-                    | :-   |  :-            |
-| `CMAKE_BUILD_TYPE`    | Indicates type of build. Possible values are 'Debug' or 'Release'     | CMake       |  |
-| `CMAKE_INSTALL_PREFIX`| Installation directory path                                           | CMake       | Debug |
+| `CMAKE_BUILD_TYPE`    | Indicates type of build. Possible values are 'Debug' or 'Release'     | CMake       | none |
+| `CMAKE_INSTALL_PREFIX`| Installation directory path                                           | CMake       | platform-dependent |
 | `CMAKE_C_COMPILER`    | Full-path to the executable corresponding to the C compiler to use.   | CMake       | platform-dependent |
 | `CMAKE_CXX_COMPILER`  | Full-path to the executable corresponding to the C++ compiler to use. | CMake       | platform-dependent |
 | `CXX`                 | Idem                                                                  | Environment | Idem |
-| `BUILD_SHARED_LIBS`   | Whether to build shared or static libraries.                          | CMake       | TRUE |
+| `BUILD_SHARED_LIBS`   | Whether to build shared or static libraries.                          | CMake       | ON |
 
 \subsection installation_variables_building Variables Used Only When Building Programs (Such as Demos or Examples)
 
--- a/Documentation/doc/Documentation/advanced/Installation.txt
+++ b/Documentation/doc/Documentation/advanced/Installation.txt
@@ -138,8 +138,8 @@ all entries are grayed. You are now ready to press *Generate*. Once this is
 done, you can quit `cmake-gui`.
 
 Without an explicit `CMAKE_BUILD_TYPE`, `CMake` does not add any additional build flags.
-This will result in a default build without any optimizations and without any debug symbols.
-You should set the variable `CMAKE_BUILD_TYPE` to `Release` or `Debug` to enable optimizations or debug symbols.
+This will result in a default build without any debug symbols, and without any optimizations.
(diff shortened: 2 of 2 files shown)


Documentation/doc/Documentation/advanced/Configuration_variables.txt
```diff
@@ -72,12 +72,12 @@ and will serverly limit performances.
...
 | :-                    | :-                    | :-   |  :-            |
-| `CMAKE_BUILD_TYPE`    | Indicates type of build. Possible values are 'Debug' or 'Release'     | CMake       |  |
-| `CMAKE_INSTALL_PREFIX`| Installation directory path                                           | CMake       | Debug |
+| `CMAKE_BUILD_TYPE`    | Indicates type of build. Possible values are 'Debug' or 'Release'     | CMake       | Release |
+| `CMAKE_INSTALL_PREFIX`| Installation directory path                                           | CMake       | `/usr/local` |
 | `CMAKE_C_COMPILER`    | Full-path to the executable corresponding to the C compiler to use.   | CMake       | platform-dependent |
 | `CMAKE_CXX_COMPILER`  | Full-path to the executable corresponding to the C++ compiler to use. | CMake       | platform-dependent |
 | `CXX`                 | Idem                                                                  | Environment | Idem |
-| `BUILD_SHARED_LIBS`   | Whether to build shared or static libraries.                          | CMake       | TRUE |
+| `BUILD_SHARED_LIBS`   | Whether to build shared or static libraries.                          | CMake       | ON |
```
lrineau, line 80: - As far as I know, the default of `CMAKE_BUILD_TYPE` is actually empty. See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html.
- `CMAKE_INSTALL_PREFIX` has a different default on Unix and Windows. Anyway, you ar right is default *is not* `Debug`.
- As for `BUILD_SHARED_LIBS`. `ON` and `TRUE` are equivalent for CMake. Why do want to change `TRUE` to `ON`?

lrineau (changes requested):

Documentation/doc/Documentation/advanced/Configuration_variables.txt
zfb132: * change it from `Release` to `none`
* change `/usr/local` to `platform-dependent`
* You're right, but most projects use `ON/OFF` to set `BUILD_SHARED_LIBS`, and the other commands in this file use `ON/OFF` too.

So, the sentence in this file also needs to be updated.
https://github.com/CGAL/cgal/blob/e1d4a96401e30cb2a3035e97c8bda44c15696cc9/Documentation/doc/Documentation/advanced/Installation.txt#L81

lrineau: > Thank you. Should I do the following?
> 
>     * change it from `Release` to `none`

Yes.

>     * change `/usr/local` to `platform-dependent`

That's appropriate.

>     * You're right, but most projects use `ON/OFF` to set `BUILD_SHARED_LIBS`, and the other commands in this file use `ON/OFF` too.

I understand. So let's switch to `ON` as you proposed in your patch.

> So, the sentence in this file also needs to be updated.
> 
> https://github.com/CGAL/cgal/blob/e1d4a96401e30cb2a3035e97c8bda44c15696cc9/Documentation/doc/Documentation/advanced/Installation.txt#L81

Oh, right! Thanks for pointing that out. Would you willing to propose a patch for that sentence as well?

zfb132: @lrineau Yes, I update the code.

Documentation/doc/Documentation/advanced/Installation.txt
```diff
@@ -137,7 +137,9 @@ Providing information and pressing *Configure* goes on until
 all entries are grayed. You are now ready to press *Generate*. Once this is
 done, you can quit `cmake-gui`.
 
-If you do not need to debug, you should set the variable `CMAKE_BUILD_TYPE` to `Release`.
+Without an explicit `CMAKE_BUILD_TYPE`, `CMake` does not add any additional build flags.
+This will result in a default build without any optimizations and without any debug symbols.
+You should set the variable `CMAKE_BUILD_TYPE` to `Release` or `Debug` to enable optimizations or debug symbols.
```
lrineau, line 142: @zfb132 Thanks you for your contribution, and in particular for that added paragraph, that is really helpful. :pray:

Your pull-request will be merged, as soon as our internal test process reports that it is OK. It might take a few days.

lrineau (approved):

Documentation/doc/Documentation/advanced/Installation.txt
MaelRL, line 142:
```suggestion
This will result in a default build without any debug symbols, and without any optimizations.
You should set the variable `CMAKE_BUILD_TYPE` to `Debug` to compile with debug symbols, or `Release` to enable optimizations.
```

PR #6862, opened 2022-09-20
PMP: Add function that trims triangle mesh at a close polyline

afabri:
## Summary of Changes

We add a function that trims a triangle mesh at a given polyline that is close to the surface. 
The current implementation projects the points and then creates a clipping mesh by extruding 
the projected polyline by local normals. 

## Todo
* [ ] The extrusion distance is still a constant
* [ ] Is the function name well chosen since we keep both parts 
* [ ] Could this be implemented with the shortest paths via projections points (intrinsic triangulation ?).

## Release Management

* Issue(s) solved (if any): -
* Feature/Small Feature (if any):  tbd
* Link to compiled documentation:  [link](https://cgal.github.io/6862/v0/Polygon_mesh_processing/group__PMP__corefinement__grp.html#ga5024bb7302c9a7159c7f1dc1b5e3b0b8)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/trim.h
```diff
@@ -0,0 +1,136 @@
...
+  typedef CGAL::AABB_face_graph_triangle_primitive<TriangleMesh>          AABB_face_graph_primitive;
+  typedef CGAL::AABB_traits<Geom_traits, AABB_face_graph_primitive>       AABB_face_graph_traits;
+  typedef CGAL::AABB_tree<AABB_face_graph_traits>                         Tree;
+  typedef Face_location<TriangleMesh, double>                        Face_location;
+
+  Tree tree;
+  build_AABB_tree(tm, tree);
+
+  int n = std::distance(pr.begin(), pr.end());
+  const double extrusion_distance = 0.1;
```
sloriot, line 71: this looks arbitrary to me

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/trim.h
```diff
@@ -0,0 +1,136 @@
...
+
+  // We could have used a modulo for closing
+  {
+    std::array<int,3> a = {2*(n-1)-2, 2*(n-1)-1, 0};
+    faces.push_back(a);
+  }
+  {
+    std::array<int,3> a = {2*(n-1)-1, 0, 1};
+    faces.push_back(a);
+  }
```
sloriot, line 102:
```suggestion
    faces.push_back({2*(n-1)-2, 2*(n-1)-1, 0});
    faces.push_back({2*(n-1)-1, 0, 1});
```

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/trim.h
```diff
@@ -0,0 +1,136 @@
...
+
+  // The extruded polyline may have problems
+  bool orientable = orient_polygon_soup(points,faces);
+  if(! orientable){
+    std::cerr << "Provide better curve" << std::endl;
+    return false;
+  }
```
sloriot, line 109: shouldn't you check for self-intersections too?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/trim.h
```diff
@@ -0,0 +1,136 @@
...
+
+  // CGAL::IO::write_polygon_mesh("splitter.off", splitter);
+
+  split(tm, splitter);
+
+  std::vector<TriangleMesh> cctm;
+
+  split_connected_components(tm, cctm);
+
+  if(cctm.size() != 2){
```
sloriot, line 122: why? it should be one component more than the input

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/trim.h
sloriot: What you can do is keep the normal direction per vertex and increase the extrusion while the trimer mesh does not intersect entirely the mesh. You can check that for each "quad" corresponding to a segment of the polyline

MaelRL:
After offline discussion, it was decided to postpone the integration of this new function as the current implementation is too tailored for specific inputs (single CC, closed mesh, nice polyline, etc.) compared to the genericity of the name.

Also the projection of the polyline onto the mesh should likely be geodesics-based instead.

sloriot:
Could it be a demo plugin?

PR #6864, opened 2022-09-20
Mesh 3 - fix construction of weights image for labeled image input

janetournois:
## Summary of Changes

Computation of weights for labeled image input was broken. Using `itkDiscreteGaussianImageFilter` with voxel-number standard deviation (instead of `itkRecursiveGaussianImageFilter` with absolute standard deviation) fixes it.

## Release Management

* Affected package(s): Mesh_3

Todo
- [ ] make screenshots before/after
- [x] check that user code has the same improvement
- [x] improve documentation about sigma

### DIFF
--- a/Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
+++ b/Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
@@ -1,3 +1,5 @@
+#define CGAL_MESH_3_WEIGHTED_IMAGES_DEBUG
+
 #include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
 
 #include <CGAL/Mesh_triangulation_3.h>
@@ -39,7 +41,7 @@ int main(int argc, char* argv[])
   /// [Loads image]
 
   /// [Domain creation]
-  const float sigma = 10.f;
+  const float sigma = (std::max)(image.vx(), (std::max)(image.vy(), image.vz()));
   CGAL::Image_3 img_weights =
     CGAL::Mesh_3::generate_label_weights(image, sigma);
 
--- a/Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
+++ b/Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
@@ -1,5 +1,3 @@
-#define CGAL_MESH_3_WEIGHTED_IMAGES_DEBUG
-
 #include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
 
 #include <CGAL/Mesh_triangulation_3.h>
--- a/Mesh_3/include/CGAL/Mesh_3/generate_label_weights.h
+++ b/Mesh_3/include/CGAL/Mesh_3/generate_label_weights.h
@@ -108,6 +108,39 @@ SIGN get_sign()
 //    SGN_UNKNOWN
 }
 
+template<typename Image_word_type>
+void convert_itk_to_image_3(itk::Image<Image_word_type, 3>* const itk_img,
+                            const char* filename = "")
+{
+  auto t = itk_img->GetOrigin();
+  auto v = itk_img->GetSpacing();
+  auto region = itk_img->GetRequestedRegion();
+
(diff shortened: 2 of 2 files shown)


Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
```diff
@@ -1,3 +1,5 @@
+#define CGAL_MESH_3_WEIGHTED_IMAGES_DEBUG
+
```
lrineau, line 2: I know the PR is a draft, but I want to check with you if you want to let that macro `CGAL_MESH_3_WEIGHTED_IMAGES_DEBUG` in the example code. Maybe that is intended. Or maybe you plan to remove it... let's leave this conversion unresolved until it is solved.

lrineau (changes requested):

Mesh_3/examples/Mesh_3/mesh_3D_weighted_image.cpp
janetournois: I will remove it before making the PR "ready for review"

Mesh_3/include/CGAL/Mesh_3/generate_label_weights.h
```diff
@@ -235,17 +295,15 @@ CGAL::Image_3 generate_label_weights_with_known_word_type(const CGAL::Image_3& i
 *
 * @param image the input labeled image from which the weights image is computed.
 *   Both will then be used to construct a `Labeled_mesh_domain_3`.
-* @param sigma the standard deviation parameter of the internal Gaussian filter
```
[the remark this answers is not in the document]
janetournois: I'm making experiments and maybe we will decide to keep it eventually...

janetournois:
Let's reintroduce `sigma` to have this PR as a real bug-fix, and we'll see later if we remove it

lrineau:
@sloriot This pull-request is ready to be tested, code-wise, even if there are still TODO items about subjects external to Github (communication and checks in an external user-code).

janetournois:
Making more experiments on some user data, I found out that there is a trade-off in the choice of parameter `sigma` :
- if too small we get the "stair-artefact" back
- if too large the fine volumes (volumes made by 2 close surfaces) are lost and not represented in the output mesh

So let's keep the parameter for now, though it would be easier to set it as a number of voxels than a real-world distance

lrineau:
> Making more experiments on some user data, I found out that there is a trade-off in the choice of parameter `sigma` :
> 
>     * if too small we get the "stair-artefact" back
> 
>     * if too large the fine volumes (volumes made by 2 close surfaces) are lost and not represented in the output mesh
> 
> 
> So let's keep the parameter for now, though it would be easier to use if it were a number of voxels than a real-world distance

What is the right value, according to your experiments? About the width of one voxel?

janetournois:
About the size of one voxel yes. But experiments did not really bring conclusions on weather we should use the minimal dimension, or maximal dimension, for the default value (?)

Mesh_3/include/CGAL/Mesh_3/generate_label_weights.h
```diff
@@ -137,7 +137,7 @@ void convert_itk_to_image_3(itk::Image<Image_word_type, 3>* const itk_img,
             itk_img->GetBufferPointer() + size,
             img_ptr);
 
-  if(filename != "")
+  if(std::strlen(filename)!=0)
```
lrineau, line 140: [`std::strlen` is unsafe](https://stackoverflow.com/q/5935413/1728537), and `filename` is a user-provided string. I suggest using `std::string` instead:
```suggestion
  if(!std::string(filename).empty())
```

Actually, I do not really understand why `filename` has default empty value, and why there is a test for `filename` before the call to `_writeImage`. Maybe the right patch is removing that line (and fix the indentation):
```suggestion
```

sloriot:

lrineau:
@janetournois You said in the TODO list in the first message (https://github.com/CGAL/cgal/pull/6864#issue-1379697365):
> make screenshots before/after

PR #6866, opened 2022-09-21
CGAL kernel: Use a division-less exact kernel by default in `Filtered_predicate`

### LINKED ISSUE #6841: Kernel predicate with RT or FT
In PR https://github.com/CGAL/cgal/pull/6796 we have removed the calls to predicates causing the filter failures. But the original issue (see https://github.com/CGAL/cgal/issues/6792#issuecomment-1209411342) is still there: given a kernel predicate having multiple overloads of `operator()`, some needing only `RT` and other needing `FT` (with division), could it be possible for `Filtered_predicate` to wrap those different `operator()` differently.

I have several ideas:
  - split the class into two: `Compare_distance_3_RT` and `Compare_distance_3_FT`, and have `Compare_distance_3` inherit from both sets of `operator()`. Then with a special `Filtered_predicate` wrap all accordingly.
  - have a special return type like `CGAL::Needs_FT<typename Kernel::Boolean>` for overloads using `FT`, and have a simple meta-programming in `Filtered_predicate` to wrap differently. `CGAL::Need_FT<Boolean>` would be have the trivial `operator Boolean()`.

What do you think of that? Maybe another idea?

_Originally posted by @lrineau in https://github.com/CGAL/cgal/issues/6792#issuecomment-1221975723_

MaelRL:
## Summary of Changes

This PR enhances `CGAL::Filtered_predicate` to use a division-less exact kernel (e.g. `Simple_cartesian<Mpzf>` + macro) for predicates and only using a division-providing exact kernel if needed.

See:
- https://github.com/CGAL/cgal/issues/6841
- https://github.com/CGAL/cgal/pull/6825 (replaces)
- https://github.com/CGAL/cgal/pull/6836 (replaces)
- https://github.com/CGAL/cgal/issues/6792

Todo:
- [x] Switch from `RT_sufficient` to `FT_required`
- [x] Rewrite the Filtered_predicate to detect `FT_required` and not `RT_sufficient`
- [x] Flag all kernel operators that require `FT` with `FT_necessary`
- [x] Remove C++17
- [x] Set up the testsuite properly (compilation messages are from `c++ (GCC) 11.3.1 20220421 (Red Hat 11.3.1-2)`)

Todo (optional):
- [ ] Somehow put a clear error when `FT_required` is not present but the filtered predicate does *not* compile with division-less `RT`.
- [ ] Check which non-kernel usages of `Filtered_predicate` could use `Filtered_predicate_RT_FT`

## Release Management

* Affected package(s): `Kernel_23`, `Cartesian_kernel`
* Issue(s) solved (if any): fix #6841, #6792


lrineau (commented):
*I have removed this review (wrong PR).*

lrineau:
> The documentation is fine, but I am not sure about the impact of parts of the implementation.

Wrong PR. Sorry.

MaelRL:
I didn't notice any slow down in compilation/runtime when using `Filtered_predicate_RT_FT` for predicates that are FT-only (I ran a large compilation and execution of `Compare_distance_2`).

[Here](https://gist.github.com/MaelRL/2332e0ff4d2d29231f3ccb845cd53181) is a log of what happens if you try to use `Filtered_predicate_RT_FT` with an `operator()` that requires `FT` but is not marked with the `FT_necessary` tag (and `Exact_RT` is a division-less `RT`).

MaelRL:
I have one issue left, related to templated operators, e.g. `Do_intersect_2`: with some arguments, it might be `RT_sufficient` while with some others, it might need `FT_necessary`. As such, whichever tag I put, the test will complain and I do not see a way to detect that this is the same (templated) `operator()` being called as there might be other non-templated `operator()`s with the same arity.

I will also leave the potential improvements of existing non-kernel `CGAL::Filtered_predicate` usages (`SDG_2`, `P23T23`, `AG2`, ...) as future work.

Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
```diff
@@ -764,20 +764,19 @@ power_side_of_bounded_power_sphereC3(
...
-  FT drw = rw - (pw + qw)/FT2;
+  FT drx = FT2 * rx - (px + qx);
+  FT dry = FT2 * ry - (py + qy);
+  FT drz = FT2 * rz - (pz + qz);
+  FT drw = FT2 * rw - (pw + qw);
   FT dr2 = CGAL_NTS square(drx) + CGAL_NTS square(dry) + CGAL_NTS square(drz);
   FT dpr = dpx*drx + dpy*dry +dpz*drz;
   return enum_cast<Bounded_side>(
-    - CGAL_NTS sign (dr2 - dp2/FT4 + dpr*dpw/dp2 - drw ));
+    - CGAL_NTS sign (dr2*dp2 - dp2*dp2 + FT2*dpr*dpw - FT2*drw*dp2 ));
```
lrineau, line 779: To save few operations, I suggest:
```suggestion
    - CGAL_NTS sign ((dr2-dp2)*dp2 + FT2*(dpr*dpw - drw*dp2) ));
```

Filtered_kernel/include/CGAL/Filtered_predicate.h
```diff
@@ -111,6 +113,78 @@ Filtered_predicate<EP,AP,C2E,C2A,Protection>::
...
+          Ares res = ap(c2a(args)...);
+          if (is_certain(res))
+            return get_certain(res);
+        }
+      catch (Uncertain_conversion_exception&) {}
+    }
+    CGAL_BRANCH_PROFILER_BRANCH(tmp);
+    Protect_FPU_rounding<!Protection> p(CGAL_FE_TONEAREST);
+    CGAL_expensive_assertion(FPU_get_cw() == CGAL_FE_TONEAREST);
+    if constexpr (Call_operator_needs_FT<Args...>::value)
```
sloriot, line 180: `C:\CGAL_ROOT\CGAL-5.6-Ic-80\include\CGAL/Filtered_predicate.h(180): error C4984: 'if constexpr' is a C++17 language extension [C:\CGAL_ROOT\CGAL-5.6-Ic-80\cmake\platforms\MSVC2017-Debug-64bits\test\Generator\generic_random_test.vcxproj]`

MaelRL:
[Here](https://gist.github.com/MaelRL/d6976402cb2fe94fc1311768b24436f2) is the log for the new test. There are a couple of errors, but it's expected: they are all within templated `operator()`s, where some combinations are RT-sufficient, and some require FT. In this configuration, the templated `operator()` is flagged with `Needs_FT` and thus some combinations complain that they could do away with the `Needs_FT`.

I also noticed that I was pruning too aggressively in the previous iterations, so it is slower than it was. In fact, I think it's likely too slow to be put in tests. I propose that this test is instead put in benchmarks, and running it once is added to the checklist before release.

lrineau:
@sloriot @MaelRL 
  - I have removed the last use of C++17 (commit 2b41ebaaaa4),
  - and I have remove the test from the set of programs that are configured (by not defining `CGAL_KERNEL_23_TEST_RT_FT_PREDICATE_FLAGS`, see commit 21f60772a1b).

I think this PR can now be tested in `integration`.

sloriot:
Intersections_2 is fully red in CGAL-5.6-Ic-113
`/mnt/testsuite/include/CGAL/Intersections_2/Bbox_2_Point_2.h:35:35: error: cannot convert 'CGAL::Needs_FT<CGAL::Uncertain<bool> >' to 'bool' in return`

sloriot:
Most probably responsible for [this](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-120/Intersections_3/TestReport_Christo_MSVC2017-Debug-64bits.gz)

sloriot:
And the red line of Kernel_d in https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-120/

MaelRL:
> Most probably responsible for [this](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-120/Intersections_3/TestReport_Christo_MSVC2017-Debug-64bits.gz)

I can reproduce this with master.

STL_Extension/include/CGAL/Handle.h
```diff
@@ -122,7 +122,7 @@ class Handle
     int
     refs()  const noexcept { return PTR->count.load(std::memory_order_relaxed); }
 
-    Id_type id() const noexcept { return PTR - static_cast<Rep*>(0); }
+    Id_type id() const noexcept { return static_cast<Id_type>(reinterpret_cast<std::intptr_t>(static_cast<void*>(PTR)) / sizeof(Rep)); }
```
sloriot, line 125: This conflicts with another fix:
```
commit 4d797b55c494f48dc1ffcfe558cd17dd3a29e9ef
Author: Sébastien Loriot <sebastien.loriot@cgal.org>
Date:   Thu Nov 17 19:03:30 2022 +0100

    try working around a warning

diff --git a/STL_Extension/include/CGAL/Handle.h b/STL_Extension/include/CGAL/Handle.h
index 4fa3a17286b..07fb7748e21 100644
--- a/STL_Extension/include/CGAL/Handle.h
+++ b/STL_Extension/include/CGAL/Handle.h
@@ -122,7 +122,7 @@ class Handle
     int
     refs()  const noexcept { return PTR->count.load(std::memory_order_relaxed); }
 
-    Id_type id() const noexcept { return PTR - static_cast<Rep*>(0); }
+    Id_type id() const noexcept { return std::distance(static_cast<Rep*>(0), PTR); }
 
     bool identical(const Handle& h) const noexcept { return PTR == h.PTR; }
``` 
I guess you should merge `master` and fix the conflict.

sloriot: I'm not sure to understand why you divide by `sizeof(Rep)`, this will change the current behavior, right?

lrineau: Pointer arithmetic is a nasty thing. In cpprer

lrineau: [In cppreference.com, we can read](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Additive_operators):
> - If the pointer `P` points to the ith element of an array, and the pointer `Q` points at the jth element of the same array, the expression `P - Q` has the value `i - j`, if the value fits in [std::ptrdiff_t](https://en.cppreference.com/w/cpp/types/ptrdiff_t). Both operands must point to the elements of the same array (or one past the end), otherwise the behavior is undefined. If the result does not fit in [std::ptrdiff_t](https://en.cppreference.com/w/cpp/types/ptrdiff_t), the behavior is undefined.

That paragraph sums up all the issue:
  - `PTR - nullptr` is undefined behavior, because `PTR` and `nullptr` do not point to elements of an array.
  - the use of `std::distance` is the same undefined behavior,
  - `P - Q` has the value `i - j`, and that computation involves a division by `sizeof(Rep)`. For example, if `Q = P + 1`, points to a the element just next to the element pointed by `P`, in array of `int`, then
    - `Q - P  == 1`, but
    - `reinterpret_cast<std::intptr_t>(Q) - reinterpret_cast<std::intptr_t>(P) == sizeof(int)`.

PR #6867, opened 2022-09-21
Add default copy constructor and operator=

sloriot:
They were automatically removed when we added the move operator


sloriot:

PR #6868, opened 2022-09-22
Number_types: Fix in testsuite code concerning min() and VC++

afabri:
## Summary of Changes

Fix 'N' in this [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-74/Number_types/TestReport_Christo_MSVC-2019-Community-Release.gz)

Note that the bug is in the testsuite code, and not in CGAL itself.
## Release Management


sloriot:
Successfully tested in CGAL-5.6-Ic-79 (error is gone)

PR #6874, opened 2022-09-24
Documentation error in Arrangement_on_surface_2

albert-github:
The construct:
```
...\ref lst_paz "code
example" ...
```
leads to the fact that the link is shown as:
```
 codeexample"
```
Note the joined words and the `"` as doxygen doesn't like the return in the string here.

doc_output/Arrangement_on_surface_2/index.html

**Old result**

image

**New result**

image


afabri:
Does the doxygen manual state that the string should be in a single line ?

albert-github:
The syntax of the `\ref` command is:
> \ref <name> ["(text)"]

and the definition of the round braces is:
> If (round) braces are used the argument extends until the end of the line on which the command was found. 

here the end of line is a bit deceptive as the `"` is mandatory as well  in case the text is given but the answer, in my opinion, is yes.

sloriot:

PR #6877, opened 2022-09-24
Documentation error in Manual

albert-github:
The construct:
```
In the previous section, we wrote that "Any model of the CGAL concept
Kernel provides what is required by the concept `ConvexHullTraits_2`.
```
leads to the fact that the entire paragraph is badly formatted as the double quote is not properly closed. Closing the double quote would lead to the fact that the `` `ConvexHullTraits_2` `` is not resolved properly

File:
- doc_output/Manual/tutorial_hello_world.html#intro_concept

**Old result**

image

**New result**

image

### DIFF
--- a/Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
+++ b/Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
@@ -265,7 +265,7 @@ given by a direction, which is hardwired in the class
 
 \section intro_concept Concepts and Models
 
-In the previous section, we wrote that "Any model of the CGAL concept
+In the previous section, we wrote that: Any model of the CGAL concept
 Kernel provides what is required by the concept `ConvexHullTraits_2`.
 
 A \em concept is a set of requirements on a type, namely that it has
--- a/Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
+++ b/Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
@@ -241,8 +241,8 @@ ch_graham_andrew( InputIterator  first,
 There are two obvious questions: What can be used as argument for
 this template parameter? And why do we have template parameters at all?
 
-To answer the first question, any model of the \cgal concept `Kernel` provides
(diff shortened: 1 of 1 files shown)


Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
```diff
@@ -265,7 +265,7 @@ given by a direction, which is hardwired in the class
 
 \section intro_concept Concepts and Models
 
-In the previous section, we wrote that "Any model of the CGAL concept
+In the previous section, we wrote that: Any model of the CGAL concept
 Kernel provides what is required by the concept `ConvexHullTraits_2`.
```
MaelRL, line 269:
```suggestion
In the previous section, we wrote: any \em model of the CGAL \em concept
`Kernel` provides what is required by the concept `ConvexHullTraits_2`.
```

albert-github: I have chosen for a slightly different formulation, making both sentences 100% identical.

I also added the `\em` for the `model` and `concept` but I've seen that we have here a bigger inconsistency problem in the form that not everywhere where `model` or `concept` is used this is used the `\em` is used. Probably we should walk through everything though this is a very very big task, I just did a quick estimate of the number  of times the word model / concept appeared without the `\em`:
- model: 6311
- models: 925
- concept: 3938
- concepts: 1019

not all of these might refer to the CGAL model / concept word and need a `\em` (as model and concept are also regular words), but still ...

MaelRL: I had put `\em` here simply because definitions were following up shortly; it was not intended to be something done systematically.

sloriot:

Documentation/doc/Documentation/Tutorials/Tutorial_hello_world.txt
lrineau: @MaelRL, is this discussion solved?

MaelRL (approved):

PR #6879, opened 2022-09-25
Some obvious missing quotes

albert-github:
Some obvious missing quotes  (last set)


sloriot:

PR #6880, opened 2022-09-26
Pmp tests only for epeck

### LINKED ISSUE #6783: PMP:   compute_face_normal() of degenerate faces

soesau:
## Summary of Changes

pmp_compute_normals_test: Computation of null vector as normal for degenerate faces is only verified for epeck
test_pmp_locate: Several predicates are only verified for epeck as they can fail with non-exact kernels

Should fix test on [this plateform](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-76/Polygon_mesh_processing/TestReport_gimeno_ArchLinux-clang-Release.gz).

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): fix #6783

### DIFF
--- a/Polygon_mesh_processing/test/Polygon_mesh_processing/pmp_compute_normals_test.cpp
+++ b/Polygon_mesh_processing/test/Polygon_mesh_processing/pmp_compute_normals_test.cpp
@@ -98,8 +98,11 @@ void test(const Mesh& mesh,
     // tests on non triangular meshes are @todo
     if(CGAL::is_triangle(halfedge(f, mesh), mesh))
     {
-      if(PMP::is_degenerate_triangle_face(f, mesh))
-        assert(get(fnormals, f) == CGAL::NULL_VECTOR);
+      if (PMP::is_degenerate_triangle_face(f, mesh))
+      {
+        if (std::is_same<K, EPECK>())
+          assert(get(fnormals, f) == CGAL::NULL_VECTOR);
+      }
       else
         assert(get(fnormals, f) != CGAL::NULL_VECTOR);
     }
--- a/Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
+++ b/Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
@@ -514,8 +514,10 @@ void test_locate_in_face(const G& g,
     PMP::locate_in_adjacent_face(loc, neigh_f, g);
     assert(PMP::locate_in_common_face<FT>(loc, neigh_loc, g));
 
-    assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K())));
-    assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K()), 1e-7));
+    if (std::is_same<K, EPECK>()) {
+      assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K())));
+      assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K()), 1e-7));
+    }
   }
 }
 
@@ -588,13 +590,19 @@ struct Locate_with_AABB_tree_Tester // 2D case
     assert(is_equal(CGAL::squared_distance(to_p3(
       PMP::construct_point(loc, g, CGAL::parameters::vertex_point_map(vpm))), p3_a), FT(0)));
 
-    assert(PMP::is_in_face(loc, g));
+    if (std::is_same<K, EPECK>()) {
+      assert(PMP::is_in_face(loc, g));
+    }
 
     loc = PMP::locate_with_AABB_tree(CGAL::ORIGIN, tree_b, g, CGAL::parameters::vertex_point_map(vpm_b));
-    assert(PMP::is_in_face(loc, g));
+    if (std::is_same<K, EPECK>()) {
+      assert(PMP::is_in_face(loc, g));
+    }
 
     loc = PMP::locate(CGAL::ORIGIN, g, CGAL::parameters::vertex_point_map(vpm_b));
-    assert(PMP::is_in_face(loc, g));
+    if (std::is_same<K, EPECK>()) {
+      assert(PMP::is_in_face(loc, g));
+    }
 
     // ---------------------------------------------------------------------------
     Ray_2 r2 = random_2D_ray<CGAL::AABB_tree<AABB_face_graph_traits> >(tree_a, rnd);
@@ -894,7 +902,7 @@ void test(CGAL::Random& rnd)
 {
   test_2D_triangulation<K>("data/stair.xy", rnd);
 //  test_2D_surface_mesh<K>("data/blobby_2D.off", rnd); // temporarily disabled, until Surface_mesh's IO is "fixed"
-  test_surface_mesh_3D<K>(CGAL::data_file_path("meshes/mech-holes-shark.off"), rnd);
+  test_surface_mesh_3D<K>("meshes/mech-holes-shark.off", rnd);
   test_surface_mesh_projection<K>("data/unit-grid.off", rnd);
   test_polyhedron<K>("data-coref/elephant_split_2.off", rnd);
 }
--- a/Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
+++ b/Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
@@ -172,15 +172,25 @@ void test_constructions(const G& g,
 
   // ---------------------------------------------------------------------------
   bar = PMP::barycentric_coordinates(p, q, r, p, K());
-  assert(is_equal(bar[0], FT(1)) && is_equal(bar[1], FT(0)) && is_equal(bar[2], FT(0)));
+  if (std::is_same<K, EPECK>()) {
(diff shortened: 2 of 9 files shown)


afabri:
I am wondering if the tests in [Spatial_searching](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-I-67.shtml#Spatial_searching), [Kernel_23](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-67/Kernel_23/TestReport_gimeno_ArchLinux-clang-Release.gz),  and [Intersections_3](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-I-67.shtml#Intersections_3) have the same problem.

MaelRL:
Some packages (like `Intersections_3`) use a variation of `approx_equal_nt` (https://github.com/CGAL/cgal/blob/master/Intersections_3/test/Intersections_3/intersection_test_helper.h#L115) to handle inexact constructions.

Polygon_mesh_processing/test/Polygon_mesh_processing/pmp_compute_normals_test.cpp
```diff
@@ -98,8 +98,11 @@ void test(const Mesh& mesh,
     // tests on non triangular meshes are @todo
     if(CGAL::is_triangle(halfedge(f, mesh), mesh))
     {
-      if(PMP::is_degenerate_triangle_face(f, mesh))
-        assert(get(fnormals, f) == CGAL::NULL_VECTOR);
+      if (PMP::is_degenerate_triangle_face(f, mesh))
+      {
+        if (std::is_same<K, EPECK>())
```
lrineau, line 103: Just a note...

That is the first time is see a use of the call operator ofor a type traits: `std::is_same<K, EPECK>()`. So far I had not know that existed. [According to the documentation](https://en.cppreference.com/w/cpp/types/is_same), it is available from C++14 and is the same as:
  - `std::is_same<K, EPECK>::value`
  - or `std::is_same_v<K, EPECK>` (C++17).

The call operator `()` is `constexpr`, so the usages are equivalent.

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
```diff
@@ -201,22 +213,31 @@ void test_constructions(const G& g,
...
+  if (std::is_same<K, EPECK>()) {
+    assert(hd);
+  }
 
   loc = std::make_pair(f, CGAL::make_array(FT(1), FT(0), FT(0)));
-  assert(PMP::is_on_vertex(loc, source(halfedge(f, g), g), g));
+  if (std::is_same<K, EPECK>()) {
+    assert(PMP::is_on_vertex(loc, source(halfedge(f, g), g), g));
+  }
+
```
MaelRL, line 240: These are not necessary (these checks are exact inputs and combinatorial checks)

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
```diff
@@ -249,24 +270,33 @@ void test_random_entities(const G& g, CGAL::Random& rnd)
...
+      assert(loc.second[0] >= FT(0) && loc.second[0] <= FT(1) &&
+        loc.second[1] >= FT(0) && loc.second[1] <= FT(1) &&
+        loc.second[2] >= FT(0) && loc.second[2] <= FT(1));
+    }
     int h_id = CGAL::halfedge_index_in_face(h, g);
-    assert(loc.second[(h_id+2)%3] == FT(0));
+
+    if (std::is_same<K, EPECK>()) {
+      assert(loc.second[(h_id + 2) % 3] == FT(0));
+    }
```
MaelRL, line 299: This isn't necessary (it sets up a coordinate to 0 by construction)

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
```diff
@@ -307,14 +337,18 @@ void test_helpers(const G& g, CGAL::Random& rnd)
...
   std::vector<face_descriptor> vec;
   PMP::internal::incident_faces(loc, g, std::back_inserter(vec));
-  assert(PMP::is_on_vertex(loc, source(h, g), g) ||
-         PMP::is_on_vertex(loc, target(h, g), g) ||
-         vec.size() == 2);
+  if (std::is_same<K, EPECK>()) {
+    assert(PMP::is_on_vertex(loc, source(h, g), g) ||
+      PMP::is_on_vertex(loc, target(h, g), g) ||
+      vec.size() == 2);
+  }
```
MaelRL, line 351: Not required (either it's precise enough, or the combinatorial part, `s.size()`, will be correct)

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
```diff
@@ -475,11 +520,12 @@ void test_locate_in_face(const G& g,
...
+    if (std::is_same<K, EPECK>()) {
+      assert(PMP::locate_in_common_face<FT>(loc, neigh_loc, g));
 
-    assert(PMP::locate_in_common_face<FT>(loc, neigh_loc, g));
-
-    assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K())));
-    assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K()), 1e-7));
+      assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K())));
+      assert(PMP::locate_in_common_face<FT>(loc, p, neigh_loc, g, CGAL::parameters::vertex_point_map(vpm).geom_traits(K()), 1e-7));
+    }
```
MaelRL, line 528: Everything above (since ~ L. 430) is construction-less and doesn't need to be EPECK only

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
```diff
@@ -537,33 +583,44 @@ struct Locate_with_AABB_tree_Tester // 2D case
     // sanitize otherwise some test platforms fail
     PMP::internal::snap_location_to_border(loc, g, FT(1e-7));
```
MaelRL, line 584: From below, there is snapping, which should take care of the approximations that come with an inexact construction kernel, and so I would not expect the EPECK filtering to be required. Did you in practice observe issues with those calls?

soesau:
Changed all the cases @MaelRL mentioned, the test still passes.

But as you mentioned, there are other exact kernels in CGAL, although they are not used in this test. Maybe a CGAL::is_exact template could be useful instead of using std::is_same<Kernel, Epeck>?

sloriot:

lrineau:
> Successfully tested in CGAL-5.6-Ic-79

There is still a TODO label, and open discussions.

soesau:
Can be tested.
The `approx_equal_nt` in Intersections_3 is actually not used. It would require some adaptations to work with more types, e.g. Point_3.
In Kernel_23, there were already several comparisons marked as exact only and this PR adds more comparisons. This makes the test pass, but it is no general solution (like implementing an `approx_equal_nt`). Thus, other comparisons may fail in the future for a different compiler version.

sloriot:
Successfully tested in CGAL-5.6-Ic-105. However three conversations (one marked as outdated) are still opened. @soesau please check the status of them.

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
soesau: I tested it with EPICK and there was at least one exception. Due to the use of random numbers, it is not that easy to reproduce.
The assert in line 526 failed with the following seeds & models:
1667563258 data-coref/elephant_split_2.off
1667562794 & 1667563434 meshes/mech-holes-shark.off

soesau: On Archlinux-clang-release

soesau:
@MaelRL I reverted the exclusion of some tests. But I kept some tests EPECK exclusive as there were still failing with EPICK in some cases.

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
MaelRL: @soesau Did this also give exceptions when reverted?

soesau: All PMP::is_in_face in Locate_with_AABB_tree_Tester::test fail.
The first asserts after snap_location_to_border do not fail with Epick.

sloriot: @MaelRL resolved?

Polygon_mesh_processing/test/Polygon_mesh_processing/test_pmp_locate.cpp
sloriot: @MaelRL do you consider this as resolved?

sloriot:

PR #6883, opened 2022-09-26
Various fixes for Hyperbolic_Delaunay_triangulation_2

### LINKED ISSUE #6869: Reading Hyperbolic Delaunay Triangulations
## Issue Details

Hi everyone, thanks for taking the time to look into my issue. I'm trying to use CGAL to compute hyperbolic Delaunay triangulations and am under the impression that I'm doing something wrong as it seems like one edge of the triangulation is not returned.

Here's the [instance](https://github.com/CGAL/cgal/files/9623294/Poincare.Disk.pdf) that I'm interested in. Points A-J are the sites and the remaining points are intersections that I added when drawing the relevant parts of the Voronoi diagram.

To me it looks like E and H are in neighboring Voronoi cells and should therefore be connected by an edge in the Delaunay triangulation. However, the following code yields the edges of the triangulation that I would expect, except for E-H. I tried to follow the CGAL example code for computing hyperbolic Delaunay triangulations in [ht2_example.cpp](https://github.com/CGAL/cgal/blob/master/Hyperbolic_triangulation_2/examples/Hyperbolic_triangulation_2/ht2_example.cpp) and wonder whether I'm missing something when reading the edges of the triangulation.

## Source Code

[Source Code](https://gist.github.com/maxkatzmann/3d0e6d25f8eab4785d799d274e112f7a)
[CMake File](https://gist.github.com/maxkatzmann/2b18f96159693a6429e03d365b188acd)

Here's the [output](https://gist.github.com/maxkatzmann/462579ce21bf77a54ddd6b0a32828e7b) of
```
cmake -DCMAKE_BUILD_TYPE=Release .
make
./cgal_util
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): macOS Big Sur Version 11.6.4 (64 bits)
* Compiler: AppleClang 13.0.0.13000029
* CGAL version: stable 5.5 (installed via homebrew)
* Boost version: stable 1.79.0 (installed via homebrew)

I hope that the report lists what is expected. If more information is needed, I'm happy to provide it. Thanks again for taking the time!

MaelRL:
## Summary of Changes

Various fixes for the `HT2` package and its demo; commits are atomic and self-explanatory.

There are some doc changes, but I would say these are fixes:
- The previous code tries to handle finite/infinite and hyperbolic/non-hyperbolic with a single flag, which caused errors to get stuff like finite but non-hyperbolic edges (see e.g. #6869). In addition, it used `tds_data()`, which is something that now also exists in `CGAL::triangulation_ds_face_base_2` and could create some conflicts. Hence, I completely re-implemented the way it was done, and added a new `tds_data`-like class in the face base called `Hyperbolic_face_data`, which adds new requirements in the face data class.
- I have removed `HDT2::insert(Point, Locate_type, int)`: The 'lt' is documented to be `HDT2::Locate_type`, which is wrong, the function's signature uses `HDT2::DT2::Locate_type`, and that's for a precise reason: `HDT2::locate()` returns location based on hyperbolic simplices, and not the underlying DT2 triangulation. So, one might retrieve `OUTSIDE_CONVEX_HULL` if a query is on a non-hyperbolic Delaunay face, but on a finite underlying DT2 face. In that configuration, it'd break the triangulation to do an insertion using `lt = OUTSIDE_CONVEX_HULL`. The documentation could be fixed, but it is too dangerous so it's best to just remove it. (Maybe it should also be simply removed from the triangulation class.)

TODO:
- [x] Announce breaking changes (doc)

## Release Management

* Affected package(s): `Hyperbolic_triangulation_2`
* Issue(s) solved (if any): fix #6869

### DIFF
--- a/Hyperbolic_triangulation_2/demo/Hyperbolic_triangulation_2/HDT2.cpp
+++ b/Hyperbolic_triangulation_2/demo/Hyperbolic_triangulation_2/HDT2.cpp
@@ -1,7 +1,3 @@
-#include <fstream>
-
-// CGAL headers
-
 #define USE_CORE_EXPR_KERNEL
 
 #ifndef USE_CORE_EXPR_KERNEL
@@ -17,14 +13,6 @@
 #include <CGAL/Hyperbolic_Delaunay_triangulation_2.h>
 #include <CGAL/point_generators_2.h>
 
-// Qt headers
-#include <QtGui>
-#include <QString>
-#include <QActionGroup>
-#include <QFileDialog>
-#include <QInputDialog>
-#include <QGraphicsEllipseItem>
-
 // GraphicsView items and event filters (input classes)
 #include <internal/Qt/TriangulationCircumcircle.h>
 #include <internal/Qt/TriangulationConflictZone.h>
@@ -38,8 +26,19 @@
 // the two base classes
 #include <CGAL/Qt/DemosMainWindow.h>
 
+// Qt headers
+#include <QtGui>
+#include <QString>
+#include <QActionGroup>
+#include <QFileDialog>
+#include <QInputDialog>
+#include <QGraphicsEllipseItem>
+
 #include "ui_HDT2.h"
 
+#include <fstream>
+#include <limits>
+#include <vector>
 
 #ifndef USE_CORE_EXPR_KERNEL
   typedef CGAL::Hyperbolic_Delaunay_triangulation_CK_traits_2<> K;
@@ -78,7 +77,7 @@ private:
 public:
   MainWindow();
 
-public slots:
+public Q_SLOTS:
 
   void processInput(CGAL::Object o);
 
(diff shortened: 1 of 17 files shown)


sloriot:
@MaelRL Could you implement the todo so that we can integrate this PR?

lrineau:
@MaelRL You [added](https://github.com/CGAL/cgal/pull/6883#event-7467457580) the label https://github.com/CGAL/cgal/labels/CHANGES.md%20not%20updated, two weeks ago. Can it be removed now, so that I can merge the PR?

MaelRL:
I have modified the log and removed the label a couple of days ago. The "not yet approved" flag is that I would have liked someone (@MoniqueTeillaud or @sloriot) to review the doc changes.

sloriot (approved):

PR #6889, opened 2022-09-27
STL_extension: doc improvements

afabri:
## Summary of Changes

Add `()` to functions.   Backtick.

## Release Management

* Affected package(s):  STL_extension

### DIFF
--- a/STL_Extension/doc/STL_Extension/CGAL/algorithm.h
+++ b/STL_Extension/doc/STL_Extension/CGAL/algorithm.h
@@ -125,7 +125,6 @@ the result of `operator++()` on a forward iterator.
 
 \sa `CGAL::predecessor()`
 
-\returns `it`.
 */
 template <class ForwardIterator>
 ForwardIterator successor(ForwardIterator it);
--- a/STL_Extension/doc/STL_Extension/CGAL/algorithm.h
+++ b/STL_Extension/doc/STL_Extension/CGAL/algorithm.h
@@ -7,7 +7,7 @@ namespace CGAL {
 /*!
 \ingroup STLAlgos
 
-\deprecated This function is deprecated, std::copy_n should be
+\deprecated This function is deprecated, `std::copy_n()` should be
 used instead.
 
 Copies the first `n` items from `first` to `result`.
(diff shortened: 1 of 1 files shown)


STL_Extension/doc/STL_Extension/CGAL/algorithm.h
```diff
@@ -7,7 +7,7 @@ namespace CGAL {
 /*!
 \ingroup STLAlgos
 
-\deprecated This function is deprecated, std::copy_n should be
+\deprecated This function is deprecated, `std::copy_n()` should be
```
lrineau, line 10: I question that choice, and all the similar changes. In C++ `std::copy_n` designs something: that is an overload set, documented: https://en.cppreference.com/w/cpp/algorithm/copy_n

But `std::copy_y()` does not design anything. In my opinion that is an error.

afabri: Don't we have the documentation convention to put `()`  behind functions in the manual?  [link](https://github.com/CGAL/cgal/wiki/Writing-Documentation#function-documentation).   It is well possible that I wrote this.

lrineau: It might be a consequence of the behavior of Doxygen with auto-links. Doxygen will [generate a link (see doc)](https://doxygen.nl/manual/autolink.html) for:
  - `convex_hull_2()`, or
  - `CGAL::convex_hull_2`,

but not for `convex_hull_2` without `()` or `CGAL::`. Maybe at the time we did not want to add `CGAL::` everywhere, and decided to use something like `convex_hull_2()`, even if that is not correct.

As a matter of fact, in [our wiki](https://github.com/CGAL/cgal/wiki/Writing-Documentation#function-documentation) I even would like to change ``The function `convex_hull_2()` ....`` to ``The function template `CGAL::convex_hull_2` ....`` to identify `CGAL::convex_hull_2` as a function **template**.

STL_Extension/doc/STL_Extension/CGAL/algorithm.h
```diff
@@ -115,17 +115,17 @@ namespace CGAL {
...
-Returns the next iterator, i.e.
-the result of `operator++` on a forward iterator.
+Returns the next iterator, i.e.,
+the result of `operator++()` on a forward iterator.
 
 
 \sa `CGAL::predecessor()`
 
-\returns `++it`.
+\returns `it`.
```
MaelRL, line 128: Not sure about that change; maybe just remove the `\returns` and simply write "calls `operator++()` on `it` and returns `it` afterward"?

In any case, `predecessor()` should also have these changes.

afabri: Or maybe returns `++it`  ?

MaelRL: but that's what it was before...?

sloriot:

STL_Extension/doc/STL_Extension/CGAL/algorithm.h
sloriot: It returns the successor of `it` without incrementing `it`

afabri: predecessor

MaelRL: > It returns the successor of `it` without incrementing `it`

The point was that "returns `it`" is wrong.

Maybe the simplest would be to just say this functions is `std::next(it)`?

sloriot, line 128:
```suggestion
```

sloriot:
Note that all those functions have been deprecated for a while so we can simply remove the doc.

PR #6902, opened 2022-09-28
Polynomial: Spurious backslashes

albert-github:
In the file Polynomial/doc/Polynomial/CGAL/Polynomial.h we see some spurious backslashes resulting in the file Polynomial/classCGAL_1_1Polynomial.html in:

```
The template argument Coeff must be at least a model of IntegralDomainWithoutDivision. For all operations naturally involving division, an IntegralDomain is required. Polynomial offers a full set of algebraic operators, i.e.\ binary +, -, *, / as well as +=, -=, *=, /=; not only for polynomials but also for a polynomial and a number of the coefficient type. (The / operator must only be used for integral divisions, i.e.\ those with remainder zero.) The operations are implemented naively: + and - need a number of Coeff operations which is linear in the degree while * is quadratic. Unary + and - and (in)equality ==, != are provided as well.
```

### DIFF
--- a/Polynomial/doc/Polynomial/CGAL/Polynomial.h
+++ b/Polynomial/doc/Polynomial/CGAL/Polynomial.h
@@ -12,11 +12,11 @@ The template argument `Coeff` must be at
 least a model of `IntegralDomainWithoutDivision`.
 For all operations naturally involving division, an `IntegralDomain`
 is required.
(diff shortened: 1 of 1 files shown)


afabri (approved):

sloriot:

Issue #6891, opened 2022-09-27
Kernel_23: inconsistent documentation


albert-github:
In the file Kernel_23/classCGAL_1_1Sphere__3.html we see:

![image](https://user-images.githubusercontent.com/5223533/192515313-c87b96fa-aded-4fbf-9753-1b457e1c58d8.png)

and

![image](https://user-images.githubusercontent.com/5223533/192515462-e78e5155-3b1a-4e89-945b-eb77be34deae.png)

So in one case a mathematical symbol and in the other case a description.
(Happens on more places)

In my opinion both should either be with a mathematical symbol or as a description.

albert-github:
Which version should it be:
- mathematical symbol (I think best also seen the use of the greater equal)
- the descriptive version

sloriot:
I prefer the c++ version as it does not required mathjax

albert-github:
@sloriot, neither version uses MathJax one is just text and the other is just `=`. The MathJax version is used necessary for things like unequal and greater equal (or do you want to have here `!=` and `>=` as text.

albert-github:
Made a small test:
```
- in backticks `==` and plain text ==
- in backticks `!=` and plain text !=
- in backticks `>=` and plain text >=
- in backticks `<=` and plain text <=
- `\f$ == \f$` result \f$ == \f$
- `\f$ != \f$` result \f$ != \f$
- `\f$ >= \f$` result \f$ >= \f$
- `\f$ <= \f$` result \f$ <= \f$
- HTML symbol `&Equal;&Equal;` result &Equal;&Equal;
- HTML symbol `&ne;` result &ne;
- HTML symbol `&le;` result &le;
- HTML symbol `&ge;` result &ge;
```
result:

![image](https://user-images.githubusercontent.com/5223533/192569560-6d07d27f-5879-46c5-b1c6-6ba4a327ca9e.png)

afabri:
In a mathematical formula I prefer mathematical symbols, but  to test if a point equals the origin I prefer `==`.   And I would not  even consider `x == 0` a formula.

albert-github:
And what about `1 <= x <= 100` formula or test so:And how about ranges:

![image](https://user-images.githubusercontent.com/5223533/192572255-42ea950d-4c48-4b96-b724-f845159d53ea.png)

afabri:
Why not just backticks  `1 <= x <= 100`

albert-github:
Might be the best readable in the code.

[...]

PR #6916, opened 2022-09-29
Kernel_23: inconsistent documentation

albert-github:
Made preconditions in the C++ form.

### DIFF
--- a/Kernel_23/doc/Kernel_23/CGAL/Direction_2.h
+++ b/Kernel_23/doc/Kernel_23/CGAL/Direction_2.h
@@ -60,7 +60,7 @@ Direction_2(const Kernel::RT &x, const Kernel::RT &y);
 
 /*!
 returns values, such that `d``== Direction_2<Kernel>(delta(0),delta(1))`.
-\pre \f$ 0 \leq i \leq1\f$.
+\pre `0 <= i <= 1`.
 */
 Kernel::RT delta(int i) const;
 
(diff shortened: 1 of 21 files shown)


Kernel_23/doc/Kernel_23/CGAL/Direction_2.h
```diff
@@ -60,7 +60,7 @@ Direction_2(const Kernel::RT &x, const Kernel::RT &y);
 
 /*!
 returns values, such that `d``== Direction_2<Kernel>(delta(0),delta(1))`.
-\pre \f$ 0 \leq i \leq1\f$.
+\pre `0 <= i <= 1`.
```
lrineau, line 63: Do we want to have incorrect C++ expressions, in preconditions? The alternative could be:
```suggestion
\pre `0 <= i` and `i <= 1`.
```

sloriot:
Successfully tested in CGAL-5.6-Ic-84

lrineau:
This pull-request seems still under discussion. @MaelRL, can you tell us if the discussion is resolved or not?

albert-github:
@lrineau in the comment https://github.com/CGAL/cgal/pull/6916#discussion_r984595658 I did some suggestions about alternative solutions, of which I think only the first one is really in the C++ style (see also #6891), but there has been no further discussion so far.
In my opinion the chosen version (`` \pre `0 <= i <= 1` ``) is OK , alternatively it would be possible to change this type of construct on a number of places into `` i == 0 || i == 1 `` (to keep the idea of the C++ style as mentioned in #6891).

afabri:
I think that `0 <= i <=1`  is ok in the documentation.  For just two possible values I would prefer "`i == 0`  or `i == 1`".  But all that does not really matter, as what is important is that we do not use `f$`.

PR #6920, opened 2022-09-30
Use `<details>` in the `cgalNParam`

albert-github:
Since doxygen 1.9.3 doxygen supports the HTML `<details>` tag, so no difficult constructs necessary with `\htmlonly`.

`cgalNParam` is used in 122 files.
Example `etrahedral_remeshing/group__PkgTetrahedralRemeshingRef.html#gadeec4a90f1b4f8c9bca76eaee9181235`

**Old**

image

**New**

image

### DIFF
--- a/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
@@ -367,12 +367,12 @@ ALIASES                = "cgal=%CGAL" \
                          "cgalNamedParamsBegin=<dl class=\"params\"><dt>Optional Named Parameters</dt><dd> <table class=\"params\">" \
                          "cgalNamedParamsBegin{1}=<dl class=\"params\"><dt>\1</dt><dd> <table class=\"params\">" \
                          "cgalNamedParamsEnd=</table> </dd> </dl>" \
-                         "cgalParamNBegin{1}=<tr><td class> \htmlonly[block] <button class=\"collapsible\">\endhtmlonly <b>\1</b> \htmlonly[block]</button> <div class=\"content\">\endhtmlonly<ul>" \
+                         "cgalParamNBegin{1}=<tr><td class> <details><summary><b>\1</b></summary><ul>" \
                          "cgalParamDescription{1}=<li>\1</li>" \
                          "cgalParamType{1}=<li><b>Type: </b>\1</li>" \
                          "cgalParamDefault{1}=<li><b>%Default: </b>\1</li>" \
                          "cgalParamExtra{1}=<li><b>Extra: </b>\1</li>" \
-                         "cgalParamNEnd=</ul> \htmlonly[block] </div> \endhtmlonly </td><td></td></tr>"
+                         "cgalParamNEnd=</ul> </details></td><td></td></tr>"
 
 # Set the OPTIMIZE_OUTPUT_FOR_C tag to YES if your project consists of C sources
 # only. Doxygen will then generate output that is more tailored for C. For


sloriot:
I don't really like the change. Comparing [this (proposed change)](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-82/master/Polygon_mesh_processing/group__PMP__corefinement__grp.html#ga95f618914f8d54d840065204b029492f) to [this (current version)](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-82/output2/Polygon_mesh_processing/group__PMP__corefinement__grp.html#ga95f618914f8d54d840065204b029492f).

There is an extra empty line and the close/open action was smooth while it would be "non-continuous" with the current proposal. The arrow style is not great too. Maybe this is something we can deal with by looking at the options of `details` in the doc.

albert-github:
In the original version the arrows are quite intrusive and don't look like those in the tree view.
This can however be overcome quite easily in the current version by changing in the `cgal_styleshee.css` for
```
.collapsible:after {
```
the lines:
```
  content: '\25B6';
  color: #7A93C5;
  font-weight: bold;
```
into
```
  content: '\25BA';
  font-size: 80%;
```
(in the past doxygen fiddled around a bit with the the right pointing arrow, shee for the differences between the 25B6 and the 25BA https://www.codetable.net/decimal/9654 and https://www.codetable.net/decimal/9660:

image

image

so just a small difference in width and height)

In the new version there are indeed 2 things:
- extra empty whitespace0
- the jump versus smooth scroll 

I'll see if I can find something, easy, for these 2 problems as I think it is favorable / more consistent to use the `<details>` version compared to have something with a `\htmlonly` block / workaround.

MaelRL:
I don't think the change in arrow shape in the tree view -and a fortiori in the NPs- is actually a good thing.

Also I liked that it had the same blue color, but I understand that this is personal pref.

And nitpicking, but (on the line) the text is too close to the arrow. Even if it's to have symmetry with the tree view, it's a wider space there.

sloriot:
> I don't think the change in arrow shape in the tree view -and a fortiori in the NPs- is actually a good thing.
> 
> Also I liked that it had the same blue color, but I understand that this is personal pref.
> 
> And nitpicking, but (on the line) the text is too close to the arrow. Even if it's to have symmetry with the tree view, it's a wider space there.

I agree I don't see why when it's expended it's a equilateral triangle and when it's collapsed it's an isosceles triangle. My personal opinion would be to update the treeview to use only equilateral triangles.

albert-github:
> I don't think the change in arrow shape in the tree view -and a fortiori in the NPs- is actually a good thing.

This is a personal preference, and this would, probably, mean just a small change in the relevant css file for the color and the weight.
I think it is good that both shapes are on one line.

> Also I liked that it had the same blue color, but I understand that this is personal pref.

Indeed a personal preference, I like it when it has the same color as the (optional) parameter it refers to as in that case it indicates that there is something for that parameter.

> And nitpicking, but (on the line) the text is too close to the arrow. Even if it's to have symmetry with the tree view, it's a wider space there.

This can easily be adjusted currently it is at 4px, but this can easily be changed

albert-github:
https://github.com/CGAL/cgal/pull/6920#issuecomment-1266638117

As far as I remember well the different browsers had some different opinions about the arrows and this combination was the best we could find.

albert-github:
Small update:
>    extra empty whitespace

this looks like a doxygen problem, as there is a paragraph befor the `<summary>` in doxygen, we are looking into this problem.

>    the jump versus smooth scroll

In the current code a small `hack,js` function is made to get the smooth scroll in of the text, looking into the problem to obtain this also with the `<details>` tag (question at stack exchange, no solution yet).

PR #6925, opened 2022-10-03
Do not use `void` as default for the geometric traits if Eigen is not present

MaelRL:
## Summary of Changes

This change is caused by the improvements of named parameters which added support for reference named parameters: https://github.com/CGAL/cgal/commit/aaa3947e4001487e5b2c0c0a14e8459a9c20d7bb.
Despite not being used in OBB, compilers will try to compile the `::reference` type in `Lookup_named_param_def` if Eigen is not present, and then proceed to complain about the type being `Default& == void&`.

This PR changes the default to be `CGAL::Default`, which solves that, and also aligns with what is done in other packages such as `Surface_mesh_parameterization`.

Bug reported by @fdrmrc (thank you).

## Release Management

* Affected package(s): `Optimal_bounding_box`
* Issue(s) solved (if any): -


sloriot:
Successfully tested in CGAL-5.6-Ic-84

PR #6931, opened 2022-10-04
Purge to-be-border halfedges from edges-to-collapse sets

MaelRL:
## Summary of Changes

In the function `PMP::remove_almost_degenerate_faces`, we need to maintain a set of to-collapse edges that correspond to badly shaped faces. These edges describe faces and thus shouldn't be border halfedges.

When an edge is collapsed, the sets need to be purged from deleted edges, but also need to be purged from non-deleted edges that will become boundary edges as these become meaningless since they do not describe a face anymore.

Here is an illustration of the configuration, `prev(h, g)` is constrained, so `rm_h` is `next(h, g)`, and `edge(next(h, g), g)` is a boundary edge.
illustration

## Release Management

* Issue(s) solved (if any): -


sloriot:

PR #6934, opened 2022-10-05
Number_types:  Cleanup concerning Quotient

afabri:
## Summary of Changes

Move the implementation history of  `Quotient`  to the subsection in the User Manual 

## Release Management

* Affected package(s):  Number_types


afabri:
As I was on a page with footnotes, the question to others, if it would not make sense to give up footnotes?   @MaelRL what do you think?

sloriot:
Some packages heavily use footnotes. I agree that is breaks the reading but depending on the size of what's in the footnote it can also be cumbersome to have in between parenthesis.

albert-github:
Besides the reason from @sloriot I think it is also possible that the same footnote is used on multiple places.

PR #6942, opened 2022-10-06
Clean up in headers

sloriot:
- Move/add include guards
- Remove useless internal header


sloriot:

PR #6985, opened 2022-10-25
fix indentation

sloriot:


sloriot:
Successfully tested in CGAL-5.6-Ic-105

PR #6992, opened 2022-10-27
Tet remeshing - `cell_is_selected_map` should be a property map

janetournois:
## Summary of Changes

The doc and name say that `cell_is_selected_map` is a property map, and it was not!

## Release Management

Todo
- [x] ~~make backward compatible with operator()~~ the existing functors were actually not working, so it's better to make sure they are not used at all
- [x] make sure the pmap is updated after each atomic operation, similarly to subdomain indices
- [ ] this PR is a bug fix so it will not appear in `changes.md` but the users should be warned when the release is published. The message could be
> Breaking change in Tetrahedral remeshing : the named parameter `cell_is_selected_map` is now consistently a property map in documentation and code.


sloriot:
Runtime errors for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-113/Tetrahedral_remeshing_Examples/TestReport_lrineau_Ubuntu-no-gmp-no-leda.gz)
Compilation error: [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-113/Tetrahedral_remeshing/TestReport_Christo_MSVC2017-Debug-64bits.gz)

sloriot:
Successfully tested in CGAL-5.6-Ic-126

PR #6998, opened 2022-11-01
Rename traits class to avoid confusion

sloriot:
`Alpha_wrap_AABB_traits` is a model of [AABBGeomTraits](https://doc.cgal.org/latest/AABB_tree/classAABBGeomTraits.html) and not [AABBTraits](https://doc.cgal.org/latest/AABB_tree/classAABBTraits.html).

Rename it to `Alpha_wrap_AABB_geom_traits`.


sloriot:
Successfully tested in CGAL-5.6-Ic-105

PR #7003, opened 2022-11-03
Add a const version of the operator

sloriot:


lrineau:
```c++
#if defined(CGAL_USE_OM_POINTS)
  typedef boost::lvalue_property_map_tag       category;
  typedef typename OpenMesh::Point             value_type;
  typedef const typename OpenMesh::Point&      reference;
#else
  typedef boost::read_write_property_map_tag category;
  typedef P value_type;
  typedef P reference;
#endif
```

I think you can remove the non-const version of the operator (that would mean that the PR sums up to adding `const` to the existing operator), because the type `reference` is:
  - either a an actual reference (and then `const reference` is the same type, because [references are always `const`](https://stackoverflow.com/a/50124580/1728537))
  - or an rvalue (returned by copy), and in that case adding `const` to it is useless (and will even trigger a warning).

sloriot:
Successfully tested in CGAL-5.6-Ic-106

PR #7006, opened 2022-11-03
Add Workflow to remove the 'Tested' label

SaillantNicolas:
## Summary of Changes

Added a workflow that removes the `Tested` labels during a new commit in a pull request

## Release Management

* Affected package(s): -
* Issue(s) solved (if any): -

### DIFF
--- a/.github/workflows/Remove_labels.yml
+++ b/.github/workflows/Remove_labels.yml
@@ -1,0 +1,32 @@
+name: remove_labels
+on:
+  pull_request_target:
+    types: [synchronize]
+jobs:
+  remove_label:
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'Tested')
+    name: remove label
+    steps:
+      - name: removelabel
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.rest.issues.removeLabel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              name: "Tested",
+            });
+      - name: Post address
+        uses: actions/github-script@v6
+        if: ${{ success() && github.event_name == 'pull_request_target' }}
+        with:
+          script: |
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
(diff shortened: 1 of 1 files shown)


.github/workflows/Remove_labels.yml
```diff
@@ -0,0 +1,32 @@
...
+          script: |
+            github.rest.issues.removeLabel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              name: "Tested",
+            });
+      - name: Post address
+        uses: actions/github-script@v6
+        if: ${{ success() && github.event_name == 'pull_request_target' }}
```
lrineau, line 24: Why the test `github.event_name == 'pull_request_target'`?

lrineau (approved):

sloriot:
"Prevously tested" and "Ready to be tested" would indeed make sense.

MaelRL:
"Ready to be tested" should denote a stable state that isn't guaranteed to be just because new stuff has been pushed.

If we were to automate labels a lot -- which I think is not worth the investment and maintenance time --, then it could be that it is at least passing all the automated checks (but it would still not be a good translation).

albert-github:
I'm not sure what @sloriot  meant by setting the already existing label `Ready to be tested` but maybe he meant, at least I would interpret it as such: the "simple user" can set in one way or another this label.

@MaelRL you might probably be right about "which I think is not worth the investment and maintenance time", my comment was also just an idea so that people start to think about the possibilities (when not done yet) or start a brainstorm about it.

albert-github:
Interesting point here is also that the "Tested" label is gone when the PR is synchronized with master (see e.g. https://github.com/CGAL/cgal/pull/6866#issuecomment-1335919831).

MaelRL:
> I'm not sure what @sloriot meant by setting the already existing label Ready to be tested but maybe he meant, at least I would interpret it as such: the "simple user" can set in one way or another this label.

I don't know if it is very important if non-CGAL devs (I assume this is what you meant by "simple user") are able to set flags or not; in that case of "simple user", the "Ready to be tested" serves as a clear "some CGAL dev looked at it", which is useful.

> Interesting point here is also that the "Tested" label is gone when the PR is synchronized with master (see e.g. https://github.com/CGAL/cgal/pull/6866#issuecomment-1335919831).

Which is good, the merge commit is not necessarily trivial, especially if there's conflict resolution.

Issue #6964, opened 2022-10-13
Github action workflow .github/workflows/build_doc.yml: set-output is deprecated


lrineau:
## Issue Details

`.github/workflows/build_doc.yml` uses `set-output`, that is [deprecated by Github](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). We will have to update the workflow.

## Source Code

The workflow sets the output parameter `DoxygenError` using `set-output`:

```shellsession
% ack set-output .github 
.github/workflows/build_doc.yml
93:              echo "::set-output name=DoxygenError::No package affected."
100:              echo "::set-output name=DoxygenError::$(cat tmp.log)"
120:            echo "::set-output name=DoxygenError::This round already exists. Overwrite it with /force-build."
```

PR #7007, opened 2022-11-03
Update build_doc Workflow to add reaction

SaillantNicolas:
## Summary of Changes

Added a workflow that adds a reaction during the build command.  `/build:`

Fix also set-output which has become deprecated.

## Release Management

* Issue(s) solved (if any):  #6964

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -51,12 +51,22 @@ jobs:
         name: "checkout branch"
         if: steps.get_round.outputs.result != 'stop'
         with:
-          repository: ${{ github.repository }}
-          ref: refs/pull/${{ steps.get_pr_number.outputs.result }}/merge
-          token: ${{ secrets.PUSH_TO_CGAL_GITHUB_IO_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_doc.yml
```diff
@@ -90,14 +100,14 @@ jobs:
...
-              echo "::set-output name=DoxygenError::No package affected."
+              echo "DoxygenError=No package affected." >> $GITHUB_OUTPUT
               exit 1
             fi
             cd build_doc && make -j2 doc
             make -j2 doc_with_postprocessing 2>tmp.log
             if [ -s tmp.log ]; then
               content=`cat ./tmp.log`
-              echo "::set-output name=DoxygenError::$(cat tmp.log)"
+              echo "DoxygenError=$(cat tmp.log)" >> $GITHUB_OUTPUT
```
lrineau, line 110: That will not work, because the content of `tmp.log` is a multilines string. Have a look at the documentation: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

PR #7020, opened 2022-11-07
Polyhedron demo - meshing from images of any word type

janetournois:
## Summary of Changes

The Mesh_3 plugin does not have restrictions on the input word type anymore, thanks to the various `CGAL::create_**_mesh_domain()` constructor functions that are now available in `Labeled_mesh_domain_3.h`

In the past, the word type was hardcoded in the mesh domain in the demo. It's not the case anymore.

## Release Management

* Affected package(s): Polyhedron_demo, and the mesh_3_plugin


janetournois:
@lrineau I'm tempted to add the "bug" label, but it's only an unnecessary restriction that is removed by this PR

PR #7028, opened 2022-11-08
Orthtree:  Fix memory leak

### LINKED ISSUE #6951: Memory leak in Orthtree
## Issue Details

The Node in the Orthtree package has a member variable called m_data that is initialized in the constructor with _new_. However, there is no destructor that frees the memory. There is a member function called _free_ that does the job but when the member function _unsplit_ is called, there is no call to _free()_.

I recommend replacing the c-style pointer with a shared pointer.

## Environment

* CGAL version: 5.5

afabri:
## Summary of Changes

Fix a memory leak by freeing all nodes, right as the destructor would do, but keeping the root node.

## Release Management

* Affected package(s): Orthtree
* Issue(s) solved (if any): fix #6951

### DIFF
--- a/Orthtree/include/CGAL/Orthtree.h
+++ b/Orthtree/include/CGAL/Orthtree.h
@@ -320,8 +320,21 @@ public:
   void refine(const Split_predicate& split_predicate) {
 
     // If the tree has already been refined, reset it
-    if (!m_root.is_leaf())
+    if (!m_root.is_leaf()){
+      std::queue<Node> nodes;
+      for (std::size_t i = 0; i < Degree::value; ++ i)
+            nodes.push (m_root[i]);
+      while (!nodes.empty())
+      {
+        Node node = nodes.front();
+        nodes.pop();
+        if (!node.is_leaf())
+          for (std::size_t i = 0; i < Degree::value; ++ i)
+            nodes.push (node[i]);
+        node.free();
+      }
       m_root.unsplit();
+    }
 
     // Reset the side length map, too
     m_side_per_depth.resize(1);
--- a/Orthtree/include/CGAL/Orthtree/Node.h
+++ b/Orthtree/include/CGAL/Orthtree/Node.h
@@ -345,7 +345,7 @@ public:
   }
 
   /*!
-    \brief returns the nth child fo this node.
+    \brief returns the nth child of this node.
 
     \pre `!is_null()`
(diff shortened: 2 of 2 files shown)


DanielZint (approved):
looks good

lrineau:
Would it be possible to backport this pull-request to `5.4.x-branch`? It looks like a bug-fix.

afabri:
> Would it be possible to backport this pull-request to `5.4.x-branch`? It looks like a bug-fix.

Sure. I'll do it.

afabri:
It is rebased.

afabri:
@DanielZint the memory leak only showed up when calling `Orthree::refine()` more than once.  Can you imagine other scenarios leading to memory leaks?   I was also wondering if instead of allocation of `Data` on the heap  it might be in a `std::deque`  per tree. As it is now, it would be easy to  unhook a node and move it with its children*  in a new tree.   Is such a functionality a desirable feature (and hence the heap is the better choice)?

DanielZint:
I don't think that the feature of moving nodes is necessary. At least I cannot think of an application. More important would be that nodes can be dynamically added and deleted. This would not be possible with a `std::deque`. Actually, I would recommend to change the c-style pointers to `std::shared_ptr`. That would allow for future changes to Orthtree without the necessity of considering possible memory leaks.

JacksonCampolattaro:
@DanielZint I agree that moving nodes between trees is unlikely to be useful; because Orthtree nodes have strict size & location requirements depending on where they appear in the tree, you would need to be very careful to avoid producing an invalid structure.

JacksonCampolattaro:
I think an earlier version of the Orthtree (back when it was just an Octree) used a unique_ptr array ([like this](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_at)) rather than c-style pointers to the children. This meant that the default destructor for the Node type automatically deallocated its children, so deleting a node would recursively free its entire subtree. Maybe something like that could be brought back, though it does give less flexibility.

You can also get a similar result with fewer changes by overriding the default destructor so that it invokes unsplit() before freeing, I think that's how some of the earliest prototypes of the Octree worked.

JacksonCampolattaro (approved):
This looks like a good solution, with minimal changes to the existing implementation and avoiding recursion.

sloriot:
Successfully tested in CGAL-5.6-Ic-111

PR #7032, opened 2022-11-09
Shortest Path: Add a default parameter so that the test does something

afabri:
## Summary of Changes

TestMesh.cpp uses boost program options, and currently has no default for a mesh.

## Release Management

* Affected package(s): Surface_mesh_shortest_path


afabri:
> Full red line in 5.6-Ic-110

That is a step forward.

sloriot:
Most probably this:
```
  if (vm.count("randomseed"))
  {
    programInstance.randomizer = new CGAL::Random(vm["randomSeed"].as<unsigned int>());
  }
```
as there is no delete.

sloriot:
Successfully tested in CGAL-5.6-Ic-119

PR #7056, opened 2022-11-17
Fix implicit conversion from `std::size_t` to `bool` creating ambiguous calls

MaelRL:
## Summary of Changes

`is_selected` is documented to be a `face_descriptor` -> `bool` property map.

## Release Management

* Affected package(s): `BGL`
* Issue(s) solved (if any): -


sloriot:
Successfully tested in CGAL-5.6-Ic-119

PR #4795, opened 2020-06-17
Mesh_3 - fix parallel optimisers

janetournois:
## Summary of Changes

In Mesh_3, the parallel versions of global optimizers, and of sliver perturber, were missing several locks.

This PR
- adds the missing locks,
- introduces `adjacent_vertices_threadsafe()` in TDS_3,
- fixes `incident_edges_threadsafe()` in TDS_3.

## Release Management

* Affected package(s): Mesh_3, TDS_3, Triangulation_3
* License and copyright ownership: unchanged

Issue #4469, opened 2020-01-14
Mesh_3: Meshing is not deterministic in parallel mode with one thread


maxGimeno:
## Issue Details
The test for determinism in mesh_3 is not linked with TBB in the CMakeLists, which hides that it is not deterministic in parrallel.

## Environment

* CGAL version: 4.14

lrineau:
The bug was introduced by #3368.

sloriot:
Isn't it fixed by https://github.com/CGAL/cgal/pull/4795?

HollowSun0:
I am still generating different numbers of points/number of tetrahedrons for every identical execution as of commit SHA 3840bc47eed9b0de79c8dfdf9195f345bf0153f8

janetournois:
This issue was fixed by commit efcebf2294fb0f9c1af3c25f349d7ea58b3476d2 in PR #7098

PR #7098, opened 2022-11-30
Mesh_3: Add test for determinism when not checking for features

### LINKED ISSUE #6958: Mesh_3 - add a test for determinism without feature protection
## Issue Details

The test `test_meshing_determinism` tests Mesh_3 on a polyhedral domain with features. Because of feature protection (and the fact that it succeeded to create a non-empty C3T3), the initialization with randomly chosen points is not challenged.
We should add a test that tests determinism with the initialization without features.

afabri:
## Summary of Changes

Add a testcase as suggested in the issue. 

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): fix #6958

### DIFF
--- a/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
+++ b/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
@@ -3,7 +3,7 @@
 #include <CGAL/Mesh_complex_3_in_triangulation_3.h>
 #include <CGAL/Mesh_criteria_3.h>
 
-#include <CGAL/Polyhedral_mesh_domain_with_features_3.h>
+#include <CGAL/Polyhedral_mesh_domain_3.h>
 #include <CGAL/make_mesh_3.h>
 #include <CGAL/lloyd_optimize_mesh_3.h>
 #include <CGAL/odt_optimize_mesh_3.h>
@@ -36,7 +36,7 @@ void test()
   // Domain
   typedef CGAL::Exact_predicates_inexact_constructions_kernel K;
   typedef CGAL::Mesh_polyhedron_3<K>::type Polyhedron;
-  typedef CGAL::Polyhedral_mesh_domain_with_features_3<K> Mesh_domain;
+  typedef CGAL::Polyhedral_mesh_domain_3<K> Mesh_domain;
 
   // Triangulation
   typedef typename CGAL::Mesh_triangulation_3<Mesh_domain, K, Concurrency_tag>::type Tr;
--- a/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
+++ b/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
@@ -1,0 +1,157 @@
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Mesh_triangulation_3.h>
+#include <CGAL/Mesh_complex_3_in_triangulation_3.h>
+#include <CGAL/Mesh_criteria_3.h>
+
+#include <CGAL/Polyhedral_mesh_domain_with_features_3.h>
+#include <CGAL/make_mesh_3.h>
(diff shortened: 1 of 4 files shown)


afabri:
@janetournois I kept the domain and the cube, but I only switched off the detection of features. 
@sloriot I am also wondering about the [CMakeLists.txt](https://github.com/CGAL/cgal/blob/master/Mesh_3/test/Mesh_3/CMakeLists.txt#L88).   Why does the original test `test_meshing_determinism.cpp` not need to link against TBB.   Or is the whole block of linking no longer necessary?

sloriot:
If you don't link with TBB then the sequential version of the mesher will be used. If the test/example you mentioned is not in the list of what is linked with TBB it has most probably be forgotten.

janetournois:
It is not linked with TBB because the parallel version is known to be non-deterministic. We want to test the sequential version

Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
```diff
@@ -0,0 +1,157 @@
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Mesh_triangulation_3.h>
+#include <CGAL/Mesh_complex_3_in_triangulation_3.h>
+#include <CGAL/Mesh_criteria_3.h>
+
+#include <CGAL/Polyhedral_mesh_domain_with_features_3.h>
```
janetournois, line 6: This can be replaced by `Polyhedra_mesh_domain_3.h` to be entirely out of the "with features" framework

janetournois, line 6:
```suggestion
#include <CGAL/Polyhedral_mesh_domain_3.h>
```

Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
```diff
@@ -0,0 +1,157 @@
...
+  std::size_t nb_runs   = 2;
+  unsigned int nb_lloyd = 2;
+  unsigned int nb_odt   = 2;
+  double perturb_bound  = 10.;
+  double exude_bound    = 15.;
+
+  // Domain
+  typedef CGAL::Exact_predicates_inexact_constructions_kernel K;
+  typedef CGAL::Mesh_polyhedron_3<K>::type Polyhedron;
+  typedef CGAL::Polyhedral_mesh_domain_with_features_3<K> Mesh_domain;
```
janetournois, line 39:
```suggestion
  typedef CGAL::Polyhedral_mesh_domain_3<K> Mesh_domain;
```

janetournois (changes requested):

janetournois:
Right, I had forgotten that point. Then, both determinism tests should be added to the list to be linked with TBB

sloriot:
Test failing on several platforms like [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-128/Mesh_3/TestReport_lrineau_Ubuntu-Latest.gz)

afabri:
> Test failing on several platforms like [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-128/Mesh_3/TestReport_lrineau_Ubuntu-Latest.gz)

@janetournois it is your turn now.

janetournois:
Actually this already is a known issue, see #4469

janetournois:
PR #3368 and #4795 have improved but not fully fixed the determinism in parallel mode with one thread problem.
I'm working on it

lrineau:
> Insertion of far points for parallel code was not deterministic. It's fixed now.
> 
> @lrineau I don't think that here, we want to keep the ability to change the seed (as we did in the mesh domains for initialization), so I hardcoded `CGAL::Random(0)` as generator for `Random_points_on_sphere_3`. Do you agree with that?

Yes, that is fine.

janetournois (approved):

sloriot:

PR #7100, opened 2022-11-30
CORE: Use istream.eof()

### LINKED ISSUE #6790: char type variable compare with EOF(-1) error in build type -Werror=type-limits

## Issue Details
compilation error:
     char default as unsigned char in ARM platform, cause char type variable compare with EOF(-1) error in build type -Werror=type-limits

### example:
```
 char c;
  ....
 if (c == EOF)
    printf("CoreIO::read_from_file(), unexpected end of file.");
```

### error log:
```
CGAL/CORE/CoreIO_impl.h:106:9: error: comparison is always false due to limited range of data type [-Werror=type-limits]
   if (c == EOF)

CGAL/CORE/CoreIO_impl.h:194:60: error: comparison is always true due to limited range of data type [-Werror=type-limits]
   for (int i=0; (i<size)&&((c=skip_backslash_new_line(in)) != EOF ); i++) {
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):  Linux, 64bit
* Compiler: aarch64-linux-gnu-g++, aarch64-linux-gnu-gcc
* Specific flags used (if any):  -Werror=type-limits
* CGAL version: 5.0
* Boost version: libboost1.65-dev:amd64

afabri:
## Summary of Changes

Do not compare a `char` with `EOF`

## Release Management

* Issue(s) solved (if any): fix #6790

### DIFF
--- a/CGAL_Core/include/CGAL/CORE/CoreIO_impl.h
+++ b/CGAL_Core/include/CGAL/CORE/CoreIO_impl.h
@@ -103,7 +103,7 @@ int skip_comment_line (std::istream & in) {
     }
   } while (c == ' ' || c == '\t' || c == '\n');
 
-  if (c == EOF)
+  if (in.eof())
     core_io_error_handler("CoreIO::read_from_file()","unexpected end of file.");
 
   in.putback(c);
@@ -191,7 +191,11 @@ void read_base_number(std::istream& in, BigInt& m, long length, long maxBits) {
 
   buffer = new char[size+2];
   // read digits
-  for (int i=0; (i<size)&&((c=skip_backslash_new_line(in)) != EOF ); i++) {
+  for (int i=0; i<size; i++) {
+    c=skip_backslash_new_line(in);
+    if(in.eof()){
+      break;
+    }
     if (c != ' ' && c != '\t' && c != '\n')
       append_char(buffer, size, pos++, c);
   }


albert-github:
Wouldn't the following files also be candidates?
```
Arrangement_on_surface_2/include/CGAL/IO/Arr_text_formatter.h:    while (((c = m_in->get()) != EOF) && ((c == ' ') || (c == '\n'))) {};
Arrangement_on_surface_2/include/CGAL/IO/Arr_text_formatter.h:    while (c != EOF && c == '#') {
Arrangement_on_surface_2/include/CGAL/IO/Arr_with_history_text_formatter.h:    while ((c = this->in().get()) != EOF && c == '#')
```

lrineau:
@afabri Your fix is correct. And the code of `Arrangement_on_surface_2/include/CGAL/IO/` is correct as well.

- In `CGAL_Core/include/CGAL/CORE/CoreIO_impl.h`: the used API is `std::istream::get(char& c)`, and `c` will never be set to the value of the macro `EOF`: instead the function sets `failbit` and `eofbit`. So your test of `in.eof()`, in this PR, is correct.
- `In Arrangement_on_surface_2/include/CGAL/IO` the API that is used is `int std::istream::get()` and the returned value can be set of `EOF` (that is a negative value, usually `-1`). So the code is correct.

sloriot:
Successfully tested in 5.6-Ic-129

PR #7101, opened 2022-11-30
Close files before writing

sloriot:


afabri:
Was this a trial to fix the documentation error when building it from within a PR with the build command? Seems not to be the solution.

sloriot:
It was indeed a test and it does not work. The fix is still valid.

sloriot:
For reference, I submitted this issue to pyquery https://github.com/gawel/pyquery/issues/237

sloriot:
Seems not to be sufficient :( I have to check the difference with my local settings...

sloriot:
It seems that my workflow update is ignored in the PR. It requires to be merged in master I guess. @lrineau @SaillantNicolas ?

lrineau:
> It seems that my workflow update is ignored in the PR. It requires to be merged in master I guess. @lrineau @SaillantNicolas ?

Yes, that is because of the trigger: https://github.com/sloriot/cgal/blob/be58448e639522ede424163ad908dfba0d43d306/.github/workflows/build_doc.yml#L3-L6

With the event [`issue_comment`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment), the context is the default branch of the repo (`master`).

sloriot:
It works now:
https://cgal.github.io/7076/v0/Manual/index.html

PR #7103, opened 2022-12-01
Do not apply smoothing if the CC has some degenerate faces

sloriot:


sloriot:
Successfully tested in 5.6-Ic-129

PR #7108, opened 2022-12-07
PMP: Fix for a -Wmaybe-uninitialized

afabri:
## Summary of Changes

Initialize a variable. This is not really needed, because we know that the inner body of the nested for loop will be executed, but the compiler doesn't because it has never read the paper. See this [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-129/Polygon_mesh_processing/TestReport_lrineau_Debian-stable-Release.gz)

## Release Management

* Affected package(s): PMP:  Polyhedral_envelope


sloriot:
Successfully tested in CGAL-5.6-Ic-133 (warning is gone)

sloriot (approved):

PR #7109, opened 2022-12-07
Three, and Polyhedron Demo: Add an #include and remove deprecated code

afabri:
## Summary of Changes

When compiling the Polyhedron demo with  `add_definitions ( -DCGAL_NO_DEPRECATED_CODE )`    there is an error message for `std::array` being not defined.

I would also consider it as a good idea to add the above mention `#define`  in the CMakeLists.txt flle of the demo. What do you think @sloriot ?

## Release Management

* Affected package(s): Polyhedron_3


sloriot:
Successfully tested in CGAL-5.6-Ic-133

PR #7110, opened 2022-12-07
Mesh_3, dump_c3t3: remove verbose flag

sloriot:


Mesh_3/include/CGAL/Mesh_3/Dump_c3t3.h
```diff
@@ -123,11 +124,11 @@ void dump_c3t3_edges(const C3t3& c3t3, std::string prefix)
   }
 }
 template <typename C3t3>
-void dump_c3t3(const C3t3& c3t3, std::string prefix)
+void dump_c3t3(const C3t3& c3t3, std::string prefix, bool verbose = false)
```
janetournois, line 127: what about

```suggestion
#ifdef CGAL_MESH_3_VERBOSE
void dump_c3t3(const C3t3& c3t3, std::string prefix, bool verbose = true)
#else
void dump_c3t3(const C3t3& c3t3, std::string prefix, bool verbose = false)
#endif
```

sloriot:
Successfully tested in CGAL-5.6-Ic-133

PR #7115, opened 2022-12-12
(Partial) Cleaning of Triangulation_3 tests

MaelRL:
## Summary of Changes

- Add include guards to a file
- Fix the large amount of duplicate content in regular T3 tests
- Tests some more kernels

## Release Management

* Affected package(s): `Triangulation_3`
* Issue(s) solved (if any): -


sloriot:
Test failing: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-139/Triangulation_3/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz

sloriot:
Actually I'm not sure the crash comes from this branch in Ic-141 it segfaults on another platform: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-141/Triangulation_3/TestReport_Christo_MSVC2015-Release-64bits.gz

MaelRL:
On https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.5.2-I-132/Triangulation_3/TestReport_lrineau_Fedora-Release.gz, asan complains about memory leaks for `test_delaunay_3` and `test_regular_3`.

sloriot:

sloriot:
This PR reveals the following error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-184/Triangulation_3/TestReport_gimeno_ArchLinux-clang.gz and it seems related to the parallel insertion with a pb with a ref counted object. I cannot reproduce it locally. @mglisse does that ring a bell?

mglisse:
I can reproduce, maybe once every 5 runs, with the branch, with both gcc-12 and clang-15, on debian testing x86_64. It does seem related to threading. `-fsanitize=thread` reports many issues, but if you ignore the ones in TBB, some seem to come from CGAL trying to read and write in the same place in different threads. It happens to be detected in the ref-counting code because that's one of the few places with an assertion, but most likely there is a race condition in the triangulation code (of course I can't 100% rule out a bug in the ref-counting, but that wouldn't be my first guess for a culprit).

sloriot:
Successfully tested in CGAL-5.6-Ic-187

lrineau:
> I can reproduce, maybe once every 5 runs, with the branch, with both gcc-12 and clang-15, on debian testing x86_64. It does seem related to threading. `-fsanitize=thread` reports many issues, but if you ignore the ones in TBB, some seem to come from CGAL trying to read and write in the same place in different threads. It happens to be detected in the ref-counting code because that's one of the few places with an assertion, but most likely there is a race condition in the triangulation code (of course I can't 100% rule out a bug in the ref-counting, but that wouldn't be my first guess for a culprit).

It should be checked, before the PR is merged. Andreas assigned that to me.

mglisse:
> Successfully tested in CGAL-5.6-Ic-187

Just checking: most likely there is a bug in master in the parallelism code and the improved test detects it with a low probability (about 20% here) when assertions are enabled. So the branch can be merged but an issue should be opened to track the bug. Is that right?

sloriot:
> > Successfully tested in CGAL-5.6-Ic-187
> 
> Just checking: most likely there is a bug in master in the parallelism code and the improved test detects it with a low probability (about 20% here) when assertions are enabled. So the branch can be merged but an issue should be opened to track the bug. Is that right?

Yes, the bug is in master, not this PR.

PR #7118, opened 2022-12-13
Add special case for intersection of a vertical segment with an horizontal segment

sloriot:
Only applied with a Cartesian kernel with a FT being floating point.


sloriot:

lrineau:
@MaelRL @sloriot: This PR is marked https://github.com/CGAL/cgal/labels/work%20in%20progress and https://github.com/CGAL/cgal/labels/Tested at the same time. Should I merge it as it is?

sloriot:
During a live discussion, I think you've raised the question of with it was not also done with an exact NT and with the homogeneous kernel.

sloriot:
I'm not planning to update that PR, so except if someone is blocking it, it can be merged.

PR #7126, opened 2022-12-19
Make smooth an option that is OFF by default

sloriot:
While providing very good results in some case, it lead to issues in some reported cases. Get back to the behavior in 5.4.


sloriot:

PR #7129, opened 2022-12-20
Leopard is retired for quite some time now

sloriot:
`CGAL_APPLE_LEOPARD` is not used anywhere.


sloriot:

PR #7133, opened 2022-12-21
Scripts, run_testsuite_with_ctest: CONFIG_TYPE maybe used on other platforms

lrineau:
## Summary of Changes

So far, the config type was only used if Cygwin was detected, because implicitly it was for the Visual Studio CMake generator. But the config type maybe used on other platforms. On MacOS, with the XCode generator, the config type is required.

So, the change is that `CONFIG_TYPE` is used if and only if it is defined. Simple.

## Release Management

I have based the change on 5.4 because we need the change for the testsuite on `monet`, a MacOS test machine, and we still test 5.4.x and 5.5.x.

@sloriot This PR should be in the current batch of PRs to be tested.

* Affected package(s): Scripts, testsuite


sloriot:

PR #7141, opened 2022-12-22
Add an example for spatial searching with projection

ipadjen:
## Summary of Changes

Added iso_rectangle_2_query example adapted for 2D spatial searching using 3D projected points. Follow-up on a discussion https://github.com/CGAL/cgal/discussions/7135

## Release Management

* Affected package(s): dD Spatial Searching
* Issue(s) solved (if any): 

### DIFF
--- a/Spatial_searching/doc/Spatial_searching/examples.txt
+++ b/Spatial_searching/doc/Spatial_searching/examples.txt
@@ -6,6 +6,7 @@
 \example Spatial_searching/fuzzy_range_query.cpp
 \example Spatial_searching/general_neighbor_searching.cpp
 \example Spatial_searching/iso_rectangle_2_query.cpp
+\example Spatial_searching/iso_rectangle_2_query_projection.cpp
 \example Spatial_searching/nearest_neighbor_searching.cpp
 \example Spatial_searching/searching_with_circular_query.cpp
 \example Spatial_searching/searching_surface_mesh_vertices.cpp
--- a/Spatial_searching/examples/Spatial_searching/CMakeLists.txt
+++ b/Spatial_searching/examples/Spatial_searching/CMakeLists.txt
@@ -28,6 +28,8 @@ create_single_source_cgal_program("distance_browsing.cpp")
 
 create_single_source_cgal_program("iso_rectangle_2_query.cpp")
 
+create_single_source_cgal_program("iso_rectangle_2_query_projection.cpp")
+
 create_single_source_cgal_program("nearest_neighbor_searching.cpp")
 
 create_single_source_cgal_program("searching_with_circular_query.cpp")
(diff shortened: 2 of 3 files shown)


ipadjen:
I wrote a quick test here: https://github.com/ipadjen/cgal/blob/Spatial_searching-example_projection_test/Spatial_searching/test/Spatial_searching/iso_rectangle_2_query_projection.cpp

It seems the projection is performing considerably worse than directly using the 3D spatial searching (with dummy values for the z axis). I'm not sure if I made any mistakes with the test.

sloriot:
> I wrote a quick test here: https://github.com/ipadjen/cgal/blob/Spatial_searching-example_projection_test/Spatial_searching/test/Spatial_searching/iso_rectangle_2_query_projection.cpp
> 
> It seems the projection is performing considerably worse than directly using the 3D spatial searching (with dummy values for the z axis). I'm not sure if I made any mistakes with the test.

Your code seems valid and I also observe version with the adaptor to be  ~10 time slower. I think we already observe such a slow down even with indices. Something with caching of points. @cjamin @afabri does that ring a bell?

sloriot:
Even more worring,  when setting `EnablePointsCache=CGAL::Tag_true` we have even worse performances (while it should be the converse)...

afabri:
The source of the slowdown is [here](https://github.com/CGAL/cgal/blob/master/Spatial_searching/include/CGAL/Search_traits_adapter.h#L147)

sloriot:
https://github.com/CGAL/cgal/pull/7144 fixes the slow down. 2D is now faster than 3D. Nice team work :+1:

afabri:
This [property map ](https://github.com/CGAL/cgal/pull/7141/files#diff-54611b6610a132021a6b67eb7048f6fed954c1b12335bc9660c56ff89770ee07R11) class in the example should become something generic.

sloriot:
> This [property map ](https://github.com/CGAL/cgal/pull/7141/files#diff-54611b6610a132021a6b67eb7048f6fed954c1b12335bc9660c56ff89770ee07R11) class in the example should become something generic.

It could indeed be added in `Property_map/include/CGAL/property_map.h`.

afabri:
> > This [property map ](https://github.com/CGAL/cgal/pull/7141/files#diff-54611b6610a132021a6b67eb7048f6fed954c1b12335bc9660c56ff89770ee07R11) class in the example should become something generic.
> 
> It could indeed be added in `Property_map/include/CGAL/property_map.h`.

Can we decompose that in a functor that projects from 2D to 3D (xy, xz, yz) and in a property map templated with a functor?

sloriot:
We should use the machinery in `CGAL/Kernel_23/internal/Projection_traits_3.h`. I mean `Projector` and in particular `x_index`, `Y_index`. Then you can have a generic property_map with an int as template and 3 template aliases.

sloriot:

PR #7144, opened 2022-12-23
Replace shared_ptr with a plain copy

sloriot:
It is 10 time faster than way when building the kd-tree. Only apply to property maps return points as lvalue.


sloriot:

PR #7152, opened 2023-01-02
Convex_hull_2: Use of 2D Delaunay

### LINKED ISSUE #7113: convex_hull_2 doesn't include collinear points.
## Issue Details

Hi, I am new to CGAL, so this may be either a documentation problem, or a missing feature. 
I am using `Exact_predicates_exact_constructions_kernel` and the `convex_hull_2` function to create a convex hull from a set of points, but the output doesn't include points that are on top of the convex hull edges (collinear with 2 other points on the convex hull). 
It would be optimal in terms of speed and ease of use if there was a way to tell CGAL to include those points, instead of adding them after manually. So my question is, is there a way to do that?
I've seen there is Traits::Left_turn_2, I am not sure if this is relevant to the question, or how to modify it to be true for collinear points too.

## Example

```
#include <CGAL/convex_hull_2.h>

typedef K::Point_2 Point;
typedef std::vector<Point> Points;
```
Then I have the set of points: (0,0), (1,1), (2,2), (0,2)
But the convex hull is: (0,0), (2,2), (0,2), but I want it to include (1,1).

```
  Points set;
  Points convex_hull;
  CGAL::convex_hull_2(set.begin(), set.end(), std::back_inserter(convex_hull_v));
```

## Environment

* Operating system: Linux 64bits
* Compiler: GCC
* CGAL version: 5.5

afabri:
## Summary of Changes

Add an example where we extract the not necessarily strong convex hull from a triangulation.

## Release Management

* Affected package(s): CH2 
* Issue(s) solved (if any): fix #7113

### DIFF
--- a/Convex_hull_2/doc/Convex_hull_2/Convex_hull_2.txt
+++ b/Convex_hull_2/doc/Convex_hull_2/Convex_hull_2.txt
@@ -143,6 +143,10 @@ check whether a given sequence of 2D points forms a (counter)clockwise strongly
 convex polygon. These are used in postcondition
 testing of the two-dimensional convex hull functions.
 
+In case you want to keep collinear points you can use the 2D Delaunay triangulation as
+in the following example. This sequence is then <em>not</em> strongly convex.
+
+\cgalExample{Convex_hull_2/ch_delaunay_2.cpp}
+
 */
 } /* namespace CGAL */
-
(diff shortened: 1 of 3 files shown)


mglisse:
Is it documented somewhere that all the vertices on the convex hull (including degenerate ones) end up connected to the infinite vertex? I was afraid simulation of simplicity might randomly put some inside the hull.

afabri:
> Is it documented somewhere that all the vertices on the convex hull (including degenerate ones) end up connected to the infinite vertex? I was afraid simulation of simplicity might randomly put some inside the hull.

If  the point was put "inside the hull", wouldn't that mean that we then have a degenerate face of the triangulation?

mglisse:
> If the point was put "inside the hull", wouldn't that mean that we then have a degenerate face of the triangulation?

Yes. I had trouble finding a guarantee in the doc that we never produce degenerate faces (on the hull).

sloriot:

PR #7171, opened 2023-01-09
Projection_traits_xy_3: Enable structural filtering

afabri:
## Summary of Changes

We can enable `Triangulation_2::inexact_locate()` out of the box for the projection in the xy-plane. 

We should have a look if it is also trivial for the projections in the xz and yz plane. 

## Release Management

* Affected package(s): Kernel_23, Triangulation_2

### DIFF
--- a/Kernel_23/include/CGAL/Projection_traits_xy_3.h
+++ b/Kernel_23/include/CGAL/Projection_traits_xy_3.h
@@ -14,6 +14,7 @@
 #define CGAL_PROJECTION_TRAITS_XY_3_H
 
 #include <CGAL/Kernel_23/internal/Projection_traits_3.h>
+#include <CGAL/Triangulation_structural_filtering_traits.h>
 
 namespace CGAL {
 
@@ -21,6 +22,11 @@ template < class R >
 class Projection_traits_xy_3
   : public internal::Projection_traits_3<R,2>
 {};
+
+template < class R >
+struct Triangulation_structural_filtering_traits<Projection_traits_xy_3<R> > {
+  typedef typename Triangulation_structural_filtering_traits<R>::Use_structural_filtering_tag  Use_structural_filtering_tag;
+};
 
 } //namespace CGAL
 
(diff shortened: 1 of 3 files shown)


afabri:
Just an idea (and maybe a bad one), why not having a function `approximate_orientation()` in the kernel instead of [this code](https://github.com/CGAL/cgal/blob/master/Triangulation_2/include/CGAL/Triangulation_2.h#L3161) inside `Triangulation_2`?

MaelRL:
Doesn't this `approx_orientation_2` already exist as `Simple_Cartesian<double>::Orientation_2`?

Hence this code of T2 could be just a call to `Cartesian_converter<K, SC<double> >` on the 3 points + a call to `SC<double>::Orientation_2`?

afabri:
Note that I want to use `Projection_traits_yz::Compute_x_2`.

afabri:
> Would it be ok to add `Compute_x_2` and `Compute_y_2` to the [`TriangulationTraits_2`](https://doc.cgal.org/latest/Triangulation_2/classTriangulationTraits__2.html) concept? This would enable to address the xz and yz plane projections.

As pointed out by @MaelRL `TriangulationTraits_2` refines `SpatialSortingTraits_2` which requires `Compute_x_2`.  It compiles without as the triangulation uses a policy that does not use `Compute_x_`.

mglisse:
> why not having a function `approximate_orientation()` in the kernel

That goes in the direction of something we used to discuss years ago (with Laurent? or Sylvain?), like `Functor<K, Orientation, Approximate>`.
(and disable structural filtering if approximate_orientation is not provided?)

> Hence this code of T2 could be just a call to `Cartesian_converter<K, SC<double> >` on the 3 points + a call to `SC<double>::Orientation_2`

The conversion to double may not be exactly the same, but it probably could. However, that seems less general than Andreas' approach. But I have no idea if Triangulation_2 can really benefit from the extra genericity...

sloriot:

lrineau:
I have merged this PR, even if the discussion was still not finished. I estimated that the patch was a good contribution anyway.

PR #7172, opened 2023-01-09
Use deduced type for source/target of a segment

sloriot:
EPECK does not return a ref for example.


sloriot:

Issue #7120, opened 2022-12-13
Tutorial runtime error


sloriot:
For example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-86/Classification_Examples/TestReport_Christo_MSVC-2019-Community-Release.gz)

lrineau:
@SaillantNicolas The error code is `0xc0000135`. That is the error code for DLLs that are not found at runtime. It is probably the DLLs of Boost serialization and Boost iostream.

lrineau:
> For example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-86/Classification_Examples/TestReport_Christo_MSVC-2019-Community-Release.gz)

As reported by Nicolas (IRL) the test results for CGAL-5.6-I-86 are two months old. That issue has been fixed since then. See for example [in CGAL-5.6-I-132 on the same platform](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-132/Classification_Examples/TestReport_Christo_MSVC-2019-Community-Release.gz).

sloriot:
The error code that I actually meant to report is `Exit code 0xc0000409` like in CGAL-5.6-Ic-121, CGAL-5.6-Ic-126, CGAL-5.6-Ic-133, CGAL-5.6-Ic-134, ...

sloriot:
Note that this is still an issue. See [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-153/Classification_Examples/TestReport_Christo_MSVC-2022-Preview-Release.gz)

sloriot:
@soesau @afabri maybe one of you could try to see if they can reproduce the error?

afabri:
I just noticed that we have several executables that write to `std::ofstream out("classification.ply")`.   When run at the same time they may conflict. I will make a PR with different names.

afabri:
> I just noticed that we have several executable that write to `std::ofstream out("classification.ply")`. When run at the same time they may conflict. I will make a PR with different names.

In fact the executable that has a problem has not the same file name :<

afabri:
To seed default random also did not help to reproduce. Maybe it needs more trace output to narrow it down.

soesau:
Same here, can't reproduce.

afabri:
I noticed that the fairing step of the hole filling is not deterministic.

afabri:
[...]

PR #7177, opened 2023-01-11
Classification: Write to different files and change degree of fair()

afabri:
## Summary of Changes

We have several examples that write the result in the same file. In case they are executed in parallel in a testsuite that may be a problem.

Reading the diff while creating this pull request I notice that the first two names were equal, but the third one (where we have the issue #7120) was not. So this won't fix the issue, but the PR makes nevertheless sense.  

It finally turned out that the real problem is that the fairing step done by filling a hole of a terrain led to a mesh that is no longer a terrain. We had to reduce the degree of the fairing continuity. 

## Release Management


sloriot:

PR #7178, opened 2023-01-11
Force html output (needed with lxml greater than 4.6.5)

sloriot:
Follow up of #7101

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -47,7 +47,7 @@ jobs:
            //get pullrequest url
            const pr_number = context.payload.issue.number
            return pr_number
-           
+
       - name: Emoji-comment
         uses: actions/github-script@v6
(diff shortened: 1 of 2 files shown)


sloriot:

sloriot:
(need to be double checked once in master with a call to build:doc)

lrineau:
@sloriot This PR has been merged, and has a https://github.com/CGAL/cgal/labels/TODO%20after%20merge.

lrineau:
Do you need help on something? Where should we look at?

sloriot:
I need to look again locally and see if I can reproduce. I only looked at one package the last time so I might have missed something.

afabri:
I guess this PR is to avoid the artefacts in the navtree of the documentation built on github. If so it would make sense to write it in the description of  a PR.

sloriot:
Actually it is working, but of course if we build the doc of a branch, that branch needs the patches from this branch.

PR #7184, opened 2023-01-12
PMP: Make longest_border() extract_boundary_cycles() coherent

afabri:
## Summary of Changes

Factorize code in order to guarantee that the halfedge returned by  `longest_border()` is among those reported by `extract_boundary_cycles()`.   We might think about adding an overload of `longest_border()` that takes the range produced by `extract_boundary_cycles()`, because otherwise one traverses all halfedges twice.

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
@@ -346,14 +346,14 @@ longest_border(const PolygonMesh& pmesh,
   {
     FT len = 0;
     for(halfedge_descriptor haf : halfedges_around_face(h, pmesh))
-      {
-        len += edge_length(haf, pmesh, np);
-      }
+    {
+      len += edge_length(haf, pmesh, np);
+    }
     if(result_len < len)
-      {
-        result_len = len;
-        result_halfedge = h;
-      }
+    {
+      result_len = len;
+      result_halfedge = h;
+    }
   }
   return std::make_pair(result_halfedge, result_len);
 }
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
@@ -24,6 +24,7 @@
 #include <CGAL/Named_function_parameters.h>
 #include <CGAL/boost/graph/named_params_helper.h>
 
+#include <CGAL/Polygon_mesh_processing/border.h>
 
 #include <CGAL/Lazy.h> // needed for CGAL::exact(FT)/CGAL::exact(Lazy_exact_nt<T>)
 
@@ -31,6 +32,7 @@
 #include <boost/graph/graph_traits.hpp>
 #include <boost/dynamic_bitset.hpp>
 
+#include <deque>
 #include <utility>
 #include <algorithm>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -308,6 +310,7 @@ face_border_length(typename boost::graph_traits<PolygonMesh>::halfedge_descripto
   *   - `first`: a halfedge on the longest border.
   *     The return type `halfedge_descriptor` is a halfedge descriptor. It is
   *     deduced from the graph traits corresponding to the type `PolygonMesh`.
+  *     `first` and it is among the halfedges reported by `extract_boundary_cycles()`.
```
albert-github, line 313: A bit a strange grammar.

afabri: The "and"  is too much.  Without, do you find the sentence still strange?

albert-github: I think that would be better.

albert-github: The fix with `and it` removed will also do.

afabri: Not really, because the first half of the sentence is about the type and not about the variable.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -334,29 +338,22 @@ longest_border(const PolygonMesh& pmesh,
             typename property_map_value<PolygonMesh, CGAL::vertex_point_t>::type>::Kernel::FT  FT;
   typedef typename boost::graph_traits<PolygonMesh>::halfedge_descriptor                       halfedge_descriptor;
 
-  std::unordered_set<halfedge_descriptor> visited;
+  std::deque<halfedge_descriptor> boundary_cycles;
```
MaelRL, line 341: why use a deque?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
```diff
@@ -334,28 +338,21 @@ longest_border(const PolygonMesh& pmesh,
...
-        }
-
-        if(result_len < len)
-        {
-          result_len = len;
-          result_halfedge = h;
-        }
-      }
+      len += edge_length(haf, pmesh, np);
+    }
```
MaelRL, line 351: Just call `face_border_length(haf, pmesh, np)`?

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
afabri: to avoid copying when reallocations happen.

MaelRL: Wouldn't this require an absurd number of boundary CCs for this to start mattering?

sloriot:

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
afabri: you are right, but on the other hand what is wrong about a deque.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
afabri: done. thanks for pointing out.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/measure.h
MaelRL: When I see a deque rather than a vector, I expect the deque's specific toolbox to come into play, and I would wonder if I missed something if it is not used. 

It's not very important.

afabri:
What is to do?

sloriot:

PR #7189, opened 2023-01-14
Nef_3: Reduce re-calculation of is_long and orthogonal_pole

GilesBathgate:
## Summary of Changes

Within `Nef_S2/include/CGAL/Nef_S2/Sphere_segment.h` the function `do_intersect_internally` requires a calculation of the `is_long` predicate which uses an orientation test against the source and target of the segment against its orthogonal pole. The orientation test is a relatively slow geometric test. In the scenario where the point doesn't intersect the antipode of the point has to also be checked, at this point the code recalculates `is_long`

The solution is to introduce some function overloads and store the result of is_long in a boolean variable such that the second check doesn't need to recalculate it.

Note that due to the short-circuiting, the is_long for s2 isn't necessarily needed, so it isn't given the same optimisation. There may be a way to lazy evaluate it, but this is yet to be explored.

## Release Management

* Affected package(s):


GilesBathgate:
By "Lazy evaluate" I mean something like this:

```c++
template <typename R>
bool do_intersect_internally(const Sphere_segment<R>& s1,
                             const Sphere_segment<R>& s2,
                             Sphere_point<R>& p)
{
  if ( equal_as_sets(s1.sphere_circle(),s2.sphere_circle()) )
    return false;
  p = CGAL::intersection(s1.sphere_circle(),s2.sphere_circle());

  typedef typename Sphere_segment<R>::Point_3 Point_3;
  const Point_3& s1_pole = s1.orthogonal_pole();
  bool s1_long = s1.is_long(s1_pole);

  std::unique_ptr<Point_3> s2_pole;
  bool s2_long;

  if ( s1.has_in_relative_interior(p, s1_pole, s1_long, false))
  {
       s2_pole = std::make_unique<Point_3>(s2.orthogonal_pole());
       s2_long = s2.is_long(*s2_pole);
       if( s2.has_in_relative_interior(p, *s2_pole, s2_long, false))
         return true;
  }

  p = p.antipode();

  if ( s1.has_in_relative_interior(p, s1_pole, s1_long, false))
  {
       if (!s2_pole) {
         s2_pole = std::make_unique<Point_3>(s2.orthogonal_pole());
         s2_long = s2.is_long(*s2_pole);
       }
       if( s2.has_in_relative_interior(p, *s2_pole, s2_long, false))
         return true;
  }

  return false;
}
```

But the code is a little asymmetric.

afabri:
Does it have a measurable impact?

GilesBathgate:
> Does it have a measurable impact?

I was sure it did but now I cannot reproduce it.

sloriot:

lrineau:
> Successfully tested in CGAL-5.6-Ic-184

But the PR is still a draft. @GilesBathgate?

lrineau:
> > Successfully tested in CGAL-5.6-Ic-184
> 
> But the PR is still a draft. @GilesBathgate?

@afabri, @MaelRL, @sloriot, what should be do with this PR? See also comments above (starting https://github.com/CGAL/cgal/pull/7189#issuecomment-1403443907).

GilesBathgate:
@lrineau It didn't seem to make a significant performance impact. So I put it as a draft, thinking this would suppress progress.

PR #7190, opened 2023-01-15
Nef_3: Performance SNC_external_structure Visit shell objects only once

GilesBathgate:
## Summary of Changes

Within `Nef_3/include/CGAL/Nef_3/SNC_external_structure.h`, the code makes use of the function `visit_shell_objects` four times. Firstly the minimal sface of each shell is calculated using the `Shell_explorer`. Then for all of the outer shells, all connected sfaces and facets for the shell are set to a new volume. Likewise, for all of the inner shells, these are set to a different volume, either directly or via the `determine_volume` function. Visitors are instantiated and the visit_shell_objects function is called each time to find sfaces and facets that are connected, via a closed walk.

Calling visit_shell_objects multiple times is unnecessary (and slow) since in the first pass all of the connected sfaces and facets can be stored in lists. 

A struct `Shell_entry` has been introduced, which contains a minimal sface, and the lists of connected sfaces and facets. There is one shell_entry for each shell, which themselves are stored in a `std::vector<Shell_entry>`  and indexed by the shell_number which was the way the shells were indexed previously.

The shell entries are populated during the first visit using the Shell_explorer visitor. Instead of calling the functions `link_as_inner_shell`, and `link_as_outer_shell`, the lists collected during the first visit are used to set the volumes eliminating the need to call `visit_shell_objects` the additional times. 

## Release Management

* Affected package(s): Nef_3

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
@@ -255,7 +255,12 @@ public:
   typedef CGAL::Unique_hash_map<SFace_const_handle,bool> SFace_visited_hash;
   typedef CGAL::Unique_hash_map<SFace_const_handle,bool> Shell_closed_hash;
 
+  typedef std::vector<SFace_handle> SFace_list;
+  typedef std::vector<Halffacet_handle> Halffacet_list;
+
   using SNC_decorator::visit_shell_objects;
+  using SNC_decorator::store_boundary_object;
+  using SNC_decorator::set_volume;
   using SNC_decorator::link_as_inner_shell;
   using SNC_decorator::link_as_outer_shell;
   using SNC_decorator::link_as_prev_next_pair;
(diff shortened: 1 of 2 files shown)


afabri:
Can you report a gain in performance?

GilesBathgate:
> Can you report a gain in performance?

Here are my results:

version | time (ms) | version | time (ms) |  
-- | -- | -- | -- | --
baseline: | 12468 | contender: | 11635 |  
baseline: | 12328 | contender: | 11975 |  
baseline: | 12568 | contender: | 12839 |  
baseline: | 12689 | contender: | 12862 |  
baseline: | 12716 | contender: | 12882 |  
baseline: | 13435 | contender: | 12839 |  
baseline: | 13393 | contender: | 12835 |  
baseline: | 13684 | contender: | 12120 |  
baseline: | 13694 | contender: | 12249 |  
baseline: | 13738 | contender: | 12207 |  
  | 13071.3 |   | 12444.3 | 4.80%

Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
```diff
@@ -255,34 +255,50 @@ class SNC_external_structure_base : public SNC_decorator<SNC_structure_>
...
+  };
+
+  typedef std::vector<Shell_entry> Shell_entry_list;
+
   struct Shell_explorer {
-    const SNC_decorator& D;
-    Sface_shell_hash&  ShellSf;
-    Face_shell_hash&   ShellF;
-    //    Shell_closed_hash& Closed;
+    Sface_shell_hash&   shell_lookup;
```
afabri, line 281: Hello @GilesBathgate, can you please the next time avoid to put a simple renaming in the same commit. This makes it harder to see the *real*  improvement.

GilesBathgate: Yes in future I will do that. I am not sure why I didn't this time 🤷

Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
```diff
@@ -255,34 +255,50 @@ class SNC_external_structure_base : public SNC_decorator<SNC_structure_>
...
   using SNC_decorator::make_twins;
 
+  struct Shell_entry {
+    Shell_entry(SFace_handle f) : minimal_sface(f) {}
+    SFace_handle minimal_sface;
+    SFace_list connected_sfaces;
+    Halffacet_list connected_halffacets;
+  };
+
+  typedef std::vector<Shell_entry> Shell_entry_list;
```
afabri, line 278: I would rename it to `Shell_entries`.

afabri:
Can we assume that this does not really increase the memory consumption?

Nef_3/include/CGAL/Nef_3/SNC_external_structure.h
```diff
@@ -255,34 +255,50 @@ class SNC_external_structure_base : public SNC_decorator<SNC_structure_>
   typedef CGAL::Unique_hash_map<SFace_const_handle,bool> SFace_visited_hash;
   typedef CGAL::Unique_hash_map<SFace_const_handle,bool> Shell_closed_hash;
 
+  typedef std::vector<SFace_handle> SFace_list;
+  typedef std::vector<Halffacet_handle> Halffacet_list;
```
afabri, line 259: I would not call a vector a list.    Maybe `SFace_handles` just does it?

GilesBathgate: Yes, and I also realise these two typedefs can be moved to the Shell_entry struct. Should I also make the new implementation details private?

GilesBathgate:
> Can we assume that this does not really increase the memory consumption?

I can measure that, but considering they are vectors of handles, I am not expecting anything significant.

sloriot:
Runtime errors on all plateforms. See for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-182/Nef_3/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz).

GilesBathgate:
@sloriot I don't know what causes the fail. When I debug I get that this line (1072 test_Nef_3.h) fails:

```c++
assert(does_nef3_equals_file(N2,"synthesis.nef3.SH"));
```

afabri:
> @sloriot I don't know what causes the fail. When I debug I get that this line (1072 test_Nef_3.h) fails:
> 
> ```c++
> assert(does_nef3_equals_file(N2,"synthesis.nef3.SH"));
> ```

This tests against a verified pre-computed result.

GilesBathgate:
> This tests against a verified pre-computed result.

Given that it tests against a verified pre-computed result, I still don't know why my changes produce an outcome that deviates from the per-computed result.

afabri:
When I compare the written file and the precomputed file, the difference is  a line for a shalfedge `1 { 1 8 16 16 40 48 } 1` with two times the number `16`.

afabri:
@GilesBathgate it goes wrong here in [`SNC_io_parser.h`](https://github.com/CGAL/cgal/blob/master/Nef_3/include/CGAL/Nef_3/SNC_io_parser.h#L1836) that one `SFaceHandle` too much gets written. So that means that one SFaceHandle too much makes it into the `Shell_entry`.  Any idea how that can happen?  I was reading the diff several times and I am in the debugger, but didn't find the problem yet.

afabri:
In fact the `SFaceHandles` do not come from a `ShellEntry`  but from an `ObjectList` inside a `Volume`.  When I add an assertion in [`store_boundary_object()`](https://github.com/CGAL/cgal/blob/master/Nef_3/include/CGAL/Nef_3/SNC_decorator.h) that the entry is not yet in there it fails.     And this function gets called by your new code.

PR #7191, opened 2023-01-15
Add an alternative of Mpzf using boost cpp_int

sloriot:
## TODO:
- [x] branch size
- [x] boost backend should not be the default in 5.6


Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
```diff
@@ -93,8 +93,12 @@ struct Exact_ring_selector : Exact_field_selector < T > { };
 
 template <>
 struct Exact_ring_selector<double>
-#ifdef CGAL_HAS_MPZF
+#if BOOST_VERSION > 107900 && defined(CGAL_USE_BOOST_MP)
+{ typedef boost::multiprecision::cpp_int Type; };
+#elif CGAL_HAS_MPZF
 { typedef Mpzf Type; };
+ #elif defined(CGAL_USE_BOOST_MP)
+{ typedef boost::multiprecision::cpp_int Type; };
```
albert-github, line 101: The space before `#elif` should not be there (consistency.

Maybe should in the complete `#if ... #endif` construct all lines not starting wit `#` be indented (readability), although this is a bit "contrary" to normal `struct` definitions (see the definition a bit lower for `Exact_field_selector`) that start with `{` at the beginning of the line.

sloriot:
Tests failing [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-158/Intersections_3/TestReport_lrineau_Fedora-32.gz) for example.

Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
```diff
@@ -93,8 +93,12 @@ struct Exact_ring_selector : Exact_field_selector < T > { };
 
 template <>
 struct Exact_ring_selector<double>
-#ifdef CGAL_HAS_MPZF
+#if BOOST_VERSION > 107900 && defined(CGAL_USE_BOOST_MP)
+{ typedef boost::multiprecision::cpp_int Type; };
```
mglisse, line 97: What are you trying to achieve? `Exact_ring_selector<double>` should provide a ring type **that contains `double`**, which is not the case of `cpp_int`.

sloriot: ah right I forget that point. Since Mpzf depends on GMP I was trying to not use it in case boost mp built-in was the "default" exact NT.

mglisse: If you want to centralize the logic in this file, we could make Exact_integer an alias for `Exact_ring_selector<int>` and move all the ifdefs from Exact_integer.h to Exact_type_selector.h. But that doesn't seem to be the goal here.
MP_Float is our only option without GMP, for `Exact_ring_selector<double>`, otherwise it will fall back to a rational type. But if you don't want GMP, why not tell it to CGAL with CGAL_NO_GMP?

sloriot:
We should try to bench a DT3 without static filtering with `CGAL_NO_GMP` and compare it with the current version without static filtering to see the impact of a filter failure when using `MP_Float`

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,496 @@
...
+
+  bool is_zero () const {
+    return man==0 && exp == 0;
+  }
+
+
+  bool is_one () const {
+    assert(false);
+    return true;
+    //  return exp==0 && size==1 && data()[0]==1;
```
afabri, line 370: We might have a shifted 1, e,g. 8 / 2^3

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,496 @@
...
+  }
+
+  bool positive() const
+  {
+    return is_positive(man);
+  }
+
+
+  friend bool operator<(const cpp_float& a, const cpp_float& b)
+  {
```
afabri, line 276: It makes sense to first compare  just the sign.   And then the sign of the exponent. And then only something expensive.

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,496 @@
...
+
+  CGAL::Sign sign () const
+  {
+    return CGAL::sign(man);
+  }
+
+};
+
+
+  template <> struct Algebraic_structure_traits< cpp_float >
```
afabri, line 382: @mglisse I copied this from `Mpzf`, but I am wondering if we should only keep what is required by `Integral_domain_without_division_tag`.

mglisse:
> Constructing a 3D Delaunay triangulation for the `ocean_r.xyz` data set is 50% slower with `cpp_float` than with `Mpzf` but has the same performance with data sets generated with `ocean.cpp`.

Isn't ocean_r.xyz more degenerate than what ocean.cpp produces? Is this benchmark with `Simple_cartesian<cpp_float>`, Epick, Epick with structural / static / dynamic filtering disabled, other?

> For VC++ with the precompiled GMP/MPFR distributed with CGAL `cpp_float` is even 2 times faster than `Mpzf`.

The version where Mpzf was faster was with a GMP compiled for this specific machine?

> I am wondering if that is due to the trailing zeros in `ocean_r.xyz`.

The decimal representation is not particularly meaningful. Trailing zeros in binary representation mean that the mantissa can be small, which should be good for cpp_float. Mpzf may benefit a little less, but in the worst case it can only use one more limb than cpp_float I think. Without a profiler to say where the time is going, it is hard to say more.

afabri:
> > Constructing a 3D Delaunay triangulation for the `ocean_r.xyz` data set is 50% slower with `cpp_float` than with `Mpzf` but has the same performance with data sets generated with `ocean.cpp`.
> 
> Isn't ocean_r.xyz more degenerate than what ocean.cpp produces? Is this benchmark with `Simple_cartesian<cpp_float>`, Epick, Epick with structural / static / dynamic filtering disabled, other?

It seemed to me that ocean is  vertical columns placed randomly and in each column points at random height.
That is what I then coded  in ocean.cpp.  Maybe I missed something.

I use `Epick` and I in the file  `Exact_type_selector.h`   I change the `typedef` for the ringtype. And I just use `simple.cpp` in the benchmark folder. 
> 
> > For VC++ with the precompiled GMP/MPFR distributed with CGAL `cpp_float` is even 2 times faster than `Mpzf`.
> 
> The version where Mpzf was faster was with a GMP compiled for this specific machine?
It is the GMP/MPFR that cgal.org distributes. Compiled with mingw if I remember correctly.
> 
> > I am wondering if that is due to the trailing zeros in `ocean_r.xyz`.
> 
> The decimal representation is not particularly meaningful. Trailing zeros in binary representation mean that the mantissa can be small, which should be good for cpp_float. Mpzf may benefit a little less, but in the worst case it can only use one more limb than cpp_float I think. Without a profiler to say where the time is going, it is hard to say more.

Thanks for your explanations.

mglisse:
> It seemed to me that ocean is vertical columns placed randomly and in each column points at random height. That is what I then coded in ocean.cpp. Maybe I missed something.

To be interesting, it needs to have many (almost) degenerate configurations, i.e. cospherical sets of 5 points, which does not seem like an obvious property for ocean.cpp. I expect that the randomly generated dataset has few filter failures, and thus the performance of Mpzf/cpp_float should not matter much. I could be wrong of course, we would need to check if the profiler says that enough time is spent in cpp_float to be sure.

> It is the GMP/MPFR that cgal.org distributes. Compiled with mingw if I remember correctly.

"For VC++ with the precompiled GMP/MPFR" seemed to imply that you were comparing to something else. Was that something else linux? Or windows with a different GMP? Or clang on windows?

afabri:
You are right that there are only very few predicates executed with `cpp_float` for what gets generated with ocean.cpp.

sloriot:
Configuration issue with cezanne:
```
-- Apple Clang version 10.0.1.10010046 compiler detected
-- Boost MP is turned off for all Apple Clang versions below 11.0.3!
```

sloriot:
New warnings in CGAL-5.6-Ic-222. For example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-222/AABB_tree/TestReport_gimeno_ArchLinux-clang-CXX17-Release.gz). Very yellow testsuite

sloriot:
The mechanism for exact backend  needs to be fixed for leda, see [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-222/AABB_tree_Examples/TestReport_lrineau_Fedora-with-LEDA.gz).

afabri:
> New warnings in CGAL-5.6-Ic-222. For example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-222/AABB_tree/TestReport_gimeno_ArchLinux-clang-CXX17-Release.gz). Very yellow testsuite

Was already done.

afabri:
The [` Segment_delaunay_graph_2`  testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-Ic-222.shtml#Segment_Delaunay_graph_2)  wants to do IO with the ringtype.   So it seems that we have to do more than just offering some undocumented low level functionality for predicated without division.

mglisse:
> The [` Segment_delaunay_graph_2` testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/results-5.6-Ic-222.shtml#Segment_Delaunay_graph_2) wants to do IO with the ringtype.

Can't you just remove (part of) that test, if it tests for undesired functionality?

afabri:
You are right that [this test](https://github.com/CGAL/cgal/blob/master/Segment_Delaunay_graph_2/test/Segment_Delaunay_graph_2/test_sdg_2_et.cpp#L11) uses a type from `CGAL::internal::`.   On the other hand it is not uninteresting to test that no divisions are made.

afabri:
We should also fix the documentation about the default parameter for `EK` [here](https://doc.cgal.org/latest/Segment_Delaunay_graph_2/structCGAL_1_1Segment__Delaunay__graph__filtered__traits__without__intersections__2.html)

afabri:
Adding `operator>()` solves the problem for the SDG_2 testsuite.

sloriot:
New warnings:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-228/AABB_tree/TestReport_Christo_MSVC-2022-Community-Release.gz

sloriot:
And no tests on arm64-apple_bigsur_clang1200-release-64bits in 5.6-Ic-228 (r everywhere)

sloriot:
+ https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-228/Number_types/TestReport_cgaltester_x86-64_Darwin-18.5_Apple-clang-default_Release.gz

afabri:
> New warnings: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-228/AABB_tree/TestReport_Christo_MSVC-2022-Community-Release.gz

I am surprised not to be able to reproduce, but it is clear that `int msb(man)`   is wrong as the [documentation of `number`](https://www.boost.org/doc/libs/1_82_0/libs/multiprecision/doc/html/boost_multiprecision/ref/number.html)  shows `unsigned msb()`.

Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
```diff
@@ -49,58 +50,137 @@ class Expr;
...
+template < typename >
+struct Exact_field_selector
+{
+  using Type = typename Exact_NT_backend<Default_exact_nt_backend>::Rational;
+};
+
+template < typename >
+struct Exact_ring_selector
+{
+  using Type = typename Exact_NT_backend<Default_exact_nt_backend>::Integer;
```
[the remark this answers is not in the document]
sloriot: This is only the default, specialization are at the end of the file.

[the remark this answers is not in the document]
sloriot: should be good now.

sloriot:
About branch size, it is 125K with `Data/data/points_3/ocean_r.xyz` and 31K without.

Number_types/include/CGAL/cpp_float.h
```diff
@@ -0,0 +1,571 @@
...
+    man(i),exp(0)
+  {}
+
+  cpp_float(long i)
+    :
+#ifdef CGAL_CPPF
+    rat(i),
+#endif
+    man(i),exp(0)
+  {}
```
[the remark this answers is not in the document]
afabri: It is not done for integers in `Gmpzf`, or do I miss something.

Number_types/include/CGAL/Number_types/internal/Exact_type_selector.h
```diff
@@ -49,57 +50,129 @@ class Expr;
...
+template <ENT_backend_choice>
+struct Exact_NT_backend;
+
+#ifdef CGAL_USE_GMP
+template <>
+struct Exact_NT_backend<GMP_BACKEND>
+  : public GMP_arithmetic_kernel
+{
+#ifdef CGAL_HAS_MPZF
+  typedef Mpzf Ring_for_float;
```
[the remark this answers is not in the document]
sloriot: I see what you mean. Let's leave it for now and we'll see later if we think it should be move. At least inheriting from Arithmetic_kernel was a must. Thanks!

sloriot:
Successfully tested in CGAL-5.6-Ic-254

PR #7199, opened 2023-01-18
pkg_filter use detected python version

sloriot:


sloriot:
Doc is fine in  CGAL-5.6-Ic-160, can be merged.

PR #7200, opened 2023-01-18
Nef_3: Fix cyclic header include dependencies.

GilesBathgate:
## Summary of Changes

When viewing the Nef_3 code in the QtCreator IDE the code model parser highlighted that the code has recursive cycles in the header includes. On further investigation, the following cyclic dependency knot was observed:

cyclic-depedency

While this isn't a practical issue, it can often be a sign of structural issues in the code, and indeed by simply removing some unneeded headers and moving some code into a common file the dependencies were fixed so that they are so:

acyclic-depedency

One forward declaration was required for SNC_io_parser, which was added to SNC_structure, where some other forward declarations are located.  A couple of other dual-references were fixed by removing unneeded headers and adding the required ones.

NOTE: The cpp core guidelines promote the idea of avoiding cyclic dependencies among source files: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf9-avoid-cyclic-dependencies-among-source-files

## Release Management

* Affected package(s): Nef_3, Convex_decomposition_3
* Issue(s) solved (if any): Fix cyclic header includes
* License and copyright ownership: CGAL authors.

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/Infimaximal_box.h
+++ b/Nef_3/include/CGAL/Nef_3/Infimaximal_box.h
@@ -24,6 +24,7 @@
 #include <CGAL/Extended_homogeneous.h>
 
 #include <CGAL/Nef_3/SNC_intersection.h>
+#include <CGAL/Nef_3/SNC_iteration.h>
 
 namespace CGAL {
 


Nef_3/include/CGAL/Nef_3/Halfedge.h
```diff
@@ -21,6 +21,7 @@
 
 #include <string>
 #include <sstream>
+#include <CGAL/IO/io.h>
```
afabri, line 24: If you have to add it here, should it not have disappeared on the left side of the diff.    Or does it just already included indirectly? Then even adding it now makes the code cleaner.

Nef_3/include/CGAL/Nef_3/Nef_box.h
```diff
@@ -16,7 +16,11 @@
 
 #include <CGAL/box_intersection_d.h>
 #include <CGAL/Box_intersection_d/box_limits.h>
+#include <CGAL/Extended_cartesian.h>
```
afabri, line 19: Same question. Why must this be added.

Nef_3/include/CGAL/Nef_3/SNC_constructor.h
```diff
@@ -24,12 +24,13 @@
 #include <CGAL/Nef_S2/Normalizing.h>
 #include <CGAL/Nef_3/bounded_side_3.h>
 #include <CGAL/Nef_3/Pluecker_line_3.h>
-#include <CGAL/Nef_3/SNC_decorator.h>
+#include <CGAL/Nef_3/SNC_const_decorator.h>
```
afabri, line 27: So here we only need the const version?

Nef_3/include/CGAL/Nef_3/Nef_box.h
GilesBathgate: These two are for the `Is_extended_kernel<Kernel>::value_type()`

afabri: So again, they get indirectly included.

afabri (approved):

Nef_3/include/CGAL/Nef_3/Nef_box.h
GilesBathgate: Yes, I don't think unnecessary includes are a good idea. (if a.h includes b.h, which includes c.h, then a.h need not include c.h directly) but if I need to include both a.h and g.h for a single translation unit to compile I think something is not quite right.

I think my changes reflect this paradigm.

Nef_3/include/CGAL/Nef_3/SNC_constructor.h
GilesBathgate: Yes since `SNC_decorator` includes `SNC_constructor`  (and SNC_constructor included SNC_decorator) I removed it to break the dual-cyclic include. Instead, the SNC_decorator is forward declared by SNC_structure which is included.

SNC_const_decorator can be included because it is self-contained.  I created an include graph for the whole Nef_3 package using: https://github.com/Leedehai/C-include-2-dot I can include it in the PR if needed.

sloriot:

PR #7212, opened 2023-01-25
Nef_S2: Skip running assertion code when runtime assertions are disabled

GilesBathgate:
## Summary of Changes

Since CGAL_assertion_code ignores the state of `CGAL::get_use_assertions()` some assertion code in SM_const_decorator.h gets unnecessarily run. The only sensible thing to me seems to be to add some additional CGAL_assertion_code to skip everything when runtime assertions are disabled.

## Release Management

* Affected package(s): Nef_3/Nef_S2
* Issue(s) solved (if any): performance/bugfix

### DIFF
--- a/Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h
+++ b/Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h
@@ -297,6 +297,8 @@ check_integrity_and_topological_planarity(bool faces) const
 #ifdef CGAL_USE_TRACE
   Object_index<SHalfedge_const_iterator>
     EI(shalfedges_begin(),shalfedges_end(),'e');
+#else
+  CGAL_assertion_code( if( !CGAL::get_use_assertions()) return; )
 #endif
   SVertex_const_handle v;
   int iso_vert_num=0;


GilesBathgate:
@afabri I also request for your comment regarding: https://github.com/CGAL/cgal/commit/9b357c6afff4cb7e6175e9096ea49e8585dc9681#r97801479

afabri:
In case it only concerns `check_integrity_and_topological_planarity(bool faces)`  would it not make sense to put that one in a CGAL_assertion(..)`   and to replace the assertions in the function itself?

GilesBathgate:
> would it not make sense to put that one in a CGAL_assertion(..)` and to replace the assertions in the function itself?

That would be an appropriate alternative. I also did it this way for brevity, and to keep the code active when CGAL_NEF_TRACEN/CGAL_USE_TRACE is active.

afabri:
This PR is also the right moment to fix [this warning](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-167/Nef_3/TestReport_gimeno_ArchLinux-clang-CXX17-Release.gz)    concerning [this line](https://github.com/CGAL/cgal/blob/d1ee393d788c4a90aac88d84a5915f3eee39104a/Nef_S2/include/CGAL/Nef_S2/SM_const_decorator.h#L304) (as well as line 310)  where we have to put  `CGAL_assertion_code()` around.

GilesBathgate:
> This PR is also the right moment to fix this warning concerning this line (as well as line 310) where we have to put `CGAL_assertion_code()` around.

Fixed in #7260

PR #7232, opened 2023-01-31
CGAL 3D demo: Fix current context in the destroy of the Scene

lrineau:
## Summary of Changes

Fix current context in the destroy of the Scene. Fixes the error:
> QOpenGLVertexArrayObject::destroy() failed to restore current context

that we can see at the exit of the demo.

## Release Management


sloriot:

PR #7238, opened 2023-02-02
Making description of author(s)  of a package consistent

albert-github:
- use `\cgalPckAuthor` in case of one author
- use of `\cgalPckAuthors` in case of multiple authors
- using in case of multiple authors always `, and `
- in the 1.9.3 `BaseDoxyfile.in` let the `\cgalPckAuthors` point to `\cgalPckAuthor` to get consistent output (not done for other versions as in the past the `ALIASES` could not call one another).

### DIFF
--- a/AABB_tree/doc/AABB_tree/PackageDescription.txt
+++ b/AABB_tree/doc/AABB_tree/PackageDescription.txt
@@ -8,7 +8,7 @@
 \cgalPkgDescriptionBegin{3D Fast Intersection and Distance Computation,PkgAABBTree}
 \cgalPkgPicture{aabb-teaser-thumb.png}
 \cgalPkgSummaryBegin
-\cgalPkgAuthors{Pierre Alliez, Stéphane Tayeb, Camille Wormser}
+\cgalPkgAuthors{Pierre Alliez, Stéphane Tayeb, and Camille Wormser}
 \cgalPkgDesc{The AABB (axis-aligned bounding box) tree component offers a static data structure and algorithms to perform efficient intersection and distance queries on sets of finite 3D geometric objects.}
 \cgalPkgManuals{Chapter_3D_Fast_Intersection_and_Distance_Computation,PkgAABBTreeRef}
 \cgalPkgSummaryEnd
--- a/Advancing_front_surface_reconstruction/doc/Advancing_front_surface_reconstruction/PackageDescription.txt
+++ b/Advancing_front_surface_reconstruction/doc/Advancing_front_surface_reconstruction/PackageDescription.txt
@@ -9,7 +9,7 @@
 \cgalPkgDescriptionBegin{Advancing Front Surface Reconstruction,PkgAdvancingFrontSurfaceReconstruction}
 \cgalPkgPicture{afsr-detail.png}
 \cgalPkgSummaryBegin
-\cgalPkgAuthors{Tran Kai Frank Da, David Cohen-Steiner}
+\cgalPkgAuthors{Tran Kai Frank Da, and David Cohen-Steiner}
 \cgalPkgDesc{This package provides a greedy algorithm for surface reconstruction from an
 unorganized point set. Starting from a seed facet, a piecewise linear
 surface is grown by adding Delaunay triangles one by one. The most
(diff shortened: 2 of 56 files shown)


MaelRL:
Nitpicking, but couldn't the "and" be automatically added to the last author? It's a little weird writing a macro with parameters "X" "Y" "and Z".

albert-github:
@MaelRL 
This would be possible but in that case we would need multiple macros like:
```
ALIASES += "cgalPckAuthor{1}=<div class=\"PkgAuthors\">\1</div>"
ALIASES += "cgalPckAuthors{2}=\cgalPckAutor{\1, and \2}"
ALIASES += "cgalPckAuthors{3}=\cgalPckAutor{\1, \2, and \3}"
...
ALIASES += "cgalPckAuthors{n}=\cgalPckAutor{\1, \2,... \n-1 and \n}"
```
And when when we would have  n+1 authors a new `ALIASES` has to be defined.

afabri:
Note that we have in average two new packages per year, so this has little value.

albert-github:
@afabri do I understand you correctly that my example with adding the extra `ALIASES` would not make sense (like I also suggested, a bit hidden, in "And when when we would have n+1 authors a new ALIASES has to be defined.".

afabri:
What I wanted to say is that we/you will spend time to make something automatic, while we rarely add new packages or modify the author list. So the "return on this investment" will come in a far future.

albert-github:
@afabri Well the work has already be done in this PR and didn't cost that much time (less than 1hr to implement it for me, thinking did take a bit of time too but this was also done during thinking about #7236.) It is not only about the time invested but also about giving a consistent appearance in the output.

albert-github:
@afabri I now think I actually a bit misread / misinterpreted your comment and that making `\cgalPckAuthor{1}` ... `cgalPckAuthor{n}` would not pay off and that the current pull request is sufficient.
Is this correct?

afabri:
Yes, just manually fixing is, seems enough for me.
Also I think if you only have two authors, then there should not be a comma before the "and". See [here](https://languagetool.org/insights/post/comma-before-and/).

albert-github:
@afabri 
Regarding the `,` before the `and` this is indeed an issue. We should probably take the same approach as the bibliography does.
(from geom.bib):
```
@inproceedings{bbp-iayed-98scg
, author =      "Herv{\'e} Br{\"o}nnimann and Christoph Burnikel and Sylvain Pion"
, title =       "Interval Arithmetic Yields Efficient Dynamic Filters for Computational Geometry"
, booktitle =   "Proc. 14th Annu. ACM Sympos. Comput. Geom."
, year =        1998
, pages =       "165--174"
, update =      "99.07 devillers, 98.07 tamassia"
}
```
and
```
@incollection{yd-ecp-95
, author =      "C. K. Yap and T. Dub{\'e}"
, title =       "The exact computation paradigm"
, editor =      "D.-Z. Du and F. K. Hwang"
, booktitle =   "Computing in Euclidean Geometry"
, series =      "Lecture Notes Series on Computing"
, volume =      4
, edition =     "2nd"
, publisher =   "World Scientific"
, address =     "Singapore"
, year =        1995
, pages =       "452--492"
, keywords =    "survey paper, numeral computing, exact compuation, fixed-precision, multiprecision number packages"
, update =      "98.07 icking+vismara, 97.11 icking, 97.03 devillers+pocchiola"
, abstract =    "A survey of the approaches to non-robustness of geometric algorithms, and especially the exact computation approach. Also surveys available big-number packages."
}
```

Which results in:
image
and
image

So yes I think we indeed should remove the `,` in case of just 2 authors.

MaelRL:
> in the 1.9.6 BaseDoxyfile.in let the \cgalPckAuthors point to \cgalPckAuthor to get consistent output (not done for other versions as in the past the ALIASES could not call one another).

We have 1.9.3, is it still fine?

albert-github:
@MaelRL The directory name at `Documentation/doc/resources` is called `1.9.3` but the files there are used for the build with the current master (and maintained as such). I don't think that the directories `1.8.14` and `1.8.20` are not used anymore.

(Unfortunately CGAL still uses, as far as I know, the old versions 1.8.4 for comparison purposes and the 1.8.13, from December 2016, for their release.)

sloriot:

PR #7246, opened 2023-02-03
Triangulation_3.h: Rewrite/factorize remove functions

lrineau:
## Summary of Changes

`CGAL/Triangulation_3.h`: Rewrite/factorize remove functions. There is **a lot** of copy-pasting in that file, 620 lines were removed and I could have gone on.

New private sub-functions used to factorize code.
  - `create_hole_outer_map`,
  - `create_triangulation_inner_map`.
  - `copy_triangulation_into_hole`, and
  - `fill_auxiliary_triangulation_with_vertices_around_v`.

## Release Management

* License and copyright ownership: maintenance by GeometryFactory, copyrights unchanged

### DIFF
--- a/Triangulation_3/include/CGAL/Triangulation_3.h
+++ b/Triangulation_3/include/CGAL/Triangulation_3.h
@@ -1659,8 +1659,8 @@ private:
   typedef typename Base::template Vertex_handle_unique_hash_map_generator<
   Vertex_handle>::type Vertex_handle_unique_hash_map;
 
-  Vertex_triple make_vertex_triple(const Facet& f) const;
-  void make_canonical_oriented_triple(Vertex_triple& t) const;
+  static Vertex_triple make_vertex_triple(const Facet& f);
+  static void make_canonical_oriented_triple(Vertex_triple& t);
 
   template < class VertexRemover >
   VertexRemover& make_hole_2D(Vertex_handle v, std::list<Edge_2D>& hole,
@@ -4359,7 +4359,7 @@ Triangulation_3<GT,Tds,Lds>::insert_and_give_new_cells(const Point& p,
 template < class Gt, class Tds, class Lds >
 typename Triangulation_3<Gt,Tds,Lds>::Vertex_triple
 Triangulation_3<Gt,Tds,Lds>::
-make_vertex_triple(const Facet& f) const
+make_vertex_triple(const Facet& f)
 {
   Cell_handle ch = f.first;
   int i = f.second;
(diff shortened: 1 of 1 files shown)


Triangulation_3/include/CGAL/Triangulation_3.h
```diff
@@ -6594,7 +6068,7 @@ template < class Gt, class Tds, class Lds >
 void
 Triangulation_3<Gt,Tds,Lds>::
 _make_big_hole_3D(Vertex_handle v,
-                  std::map<Vertex_triple,Facet>& outer_map,
+                  Vertex_triple_Facet_map& outer_map,
                   std::vector<Cell_handle>& hole,
                   std::vector<Vertex_handle>& vertices,
                   std::map<Vertex_handle, REMOVE_VERTEX_STATE>& vstates)
```
afabri, line 6074: Are these maps ever big, and we could gain from using unordered maps.

Triangulation_3/include/CGAL/Triangulation_3.h
```diff
@@ -1673,13 +1673,32 @@ class Triangulation_3
...
-                    Vertex_triple_Facet_map& outer_map);
+  Vertex_triple_Facet_map
+  create_hole_outer_map(Vertex_handle v,
+                        const std::vector<Cell_handle>& hole);
+
+  template < class Triangulation >
+  static Vertex_triple_Facet_map
+  create_triangulation_inner_map(const Triangulation& t,
+                                 const Vertex_handle_unique_hash_map& vmap,
+                                 bool inf);
```
afabri, line 1684: I always thought it was a bad idea to return a map, but you probably will tell me that there happens a move on return.

afabri:
@MaelRL could periodic T3 profit from this?

Triangulation_3/include/CGAL/Triangulation_3.h
lrineau: The worst case would be move, but the usual case is the [Return value optimization](https://sigcpp.github.io/2020/06/08/return-value-optimization): the object is created directly in the stack of the caller instead of the stack of the callee. That optimization is mandatory since C++17, but has been a common optimization of C++ compilers even since the age of C++03.

Triangulation_3/include/CGAL/Triangulation_3.h
lrineau: Small or big, I think a flat map would be a better implementation than `std::map`. But I did not want to mix too many different modifications in this PR.

sloriot:

PR #7253, opened 2023-02-05
STL_Extension/Triangulation_2: improve CDT_2 debug messages

lrineau:
## Summary of Changes

This pull-request has three commits:
  - Introduce `Base_with_time_stamp`, that introduces a real vertex base template class, instead of various ad-hoc solutions in various places.
  - Use `Output_rep` to display debug info, that introduces a nice way to display the debug information about vertices in CDT_2 debug messages.
  - `Circulator_from_container<const Container>` now works... because otherwise that was a hamper against adding `const`.

## Release Management

* Affected package(s): STL_Extension, Circulator (that should be merged into `STL_Extension`), and `Triangulation_2`
* License and copyright ownership: maintenance by GeometryFactory, ownership unchanged.

### DIFF
--- a/Circulator/include/CGAL/circulator.h
+++ b/Circulator/include/CGAL/circulator.h
@@ -701,7 +701,13 @@ typedef Iterator_from_circulator< C, const_reference, const_pointer>
 template <class Container>
 class Circulator_from_container {
     typedef Circulator_from_container<Container>      Self;
-    typedef typename Container::iterator              iterator;
+    typedef typename Container::iterator              container_iterator;
+    typedef typename Container::const_iterator        container_const_iterator;
+    typedef std::conditional_t<
+        std::is_const<Container>::value,
+        container_const_iterator,
+        container_iterator
+        >                                             iterator;
     typedef std::iterator_traits<iterator>            iterator_traits;
 public:
     typedef typename iterator_traits::value_type      value_type;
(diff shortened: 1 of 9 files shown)


afabri:
Would it make sense to document the time stamping and show examples. When other developers see it, they might adopt it for their needs.

lrineau:
> Would it make sense to document the time stamping and show examples. When other developers see it, they might adopt it for their needs.

For the moment I am not satisfied the way it is implemented. because that uses an undefined behavior in `Compact_container`. I would prefer to document it only once that implementation is improved and reliable. For the moment, that can only be used as a debugging tool.

afabri:
Error in this [VC testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-176/Triangulation_2/TestReport_Christo_MSVC2015-Release-64bits.gz)

lrineau:
> Error in this [VC testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-176/Triangulation_2/TestReport_Christo_MSVC2015-Release-64bits.gz)

If is probably fixed with newest commit 166ff0fdc71e9287a9c77e36e047f21b45fce529.

afabri:
> > Error in this [VC testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-176/Triangulation_2/TestReport_Christo_MSVC2015-Release-64bits.gz)
> 
> If is probably fixed with newest commit [166ff0f](https://github.com/CGAL/cgal/commit/166ff0fdc71e9287a9c77e36e047f21b45fce529).

The error is gone.

sloriot:

PR #7254, opened 2023-02-06
Fix CMAKE_NO_SYSTEM_FROM_IMPORTED

lrineau:
## Summary of Changes

With the current `master`, if one configure all CGAL with `-DWITH_examples=ON`, `-DWITH_tests=ON`... then the compiler warnings are disable for tests (but not for examples).

This pull-request fixes the situation, until I can find a better workaround.

## Release Management

This PR is a bug-fix... but it has conflicts with CGAL-5.4.x and 5.5.x. I have decided to push it to `master` only, given that the impact is probably quite limited to myself only.


sloriot:

PR #7262, opened 2023-02-13
Surface Mesh Simplification: Improve Examples

afabri:
## Summary of Changes

The first improvement is to collapse half of the edges and not just one, as consequence of a typo.

We should also switch from `operator>>()`   to functions that can read formats different from `.off`.

## Release Management

* Affected package(s): Surface_mesh_simplification


sloriot:

PR #7273, opened 2023-02-15
Polyhedron demo, Mesh_3 plugin: modify the criptic debug message "USE SIZING FIELD!"

lrineau:
## Summary of Changes

When using the Mesh_3 plugin, one might see that message:

> USE SIZING FIELD!

It is now replaced by:

> Note: Mesh_3 is using a sizing field based on AABB tree.

## Release Management


sloriot:

Issue #7271, opened 2023-02-14
Nef_3: it is not possible to decide which one is a visible facet (if any) assertion error


GilesBathgate:
## Issue Details

Following on the investigation in #7205 this is an example with a different exception. The polyhedron is:

```c++
Polyhedron polyhedron = {.points = {{0,0,0},{0,-10,0},{10,0,0},{20,0,0},{10,10,0},{5,0,0}}, .faces = {{0,1,2},{4,3,5}}};
```

```
CGAL error: assertion violation!
Expression : 
File       : /usr/include/CGAL/Nef_3/SNC_const_decorator.h
Line       : 276
Explanation: it is not possible to decide which one is a visible facet (if any)
Refer to the bug-reporting instructions at https://www.cgal.org/bug_report.html
terminate called after throwing an instance of 'CGAL::Assertion_exception'
  what():  CGAL ERROR: assertion violation!
File: /usr/include/CGAL/Nef_3/SNC_const_decorator.h
Line: 276
Explanation: it is not possible to decide which one is a visible facet (if any)
```

## Source Code

```c++
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/Modifier_base.h>
#include <CGAL/Nef_polyhedron_3.h>
#include <CGAL/Polyhedron_incremental_builder_3.h>
#include <vector>
#include <tuple>

using K=CGAL::Exact_predicates_exact_constructions_kernel;
using HalfedgeDS=CGAL::Polyhedron_3<K>::HalfedgeDS;

using Points = std::vector<CGAL::Point_3<K>>;
using Facet = std::vector<Points::size_type>;
using Facets = std::vector<Facet>;
struct Polyhedron { Points points; Facets faces; };

struct Builder : public CGAL::Modifier_base<HalfedgeDS>
{
	const Polyhedron& polyhedron;
	Builder(const Polyhedron& p) : polyhedron(p) {}

	void operator()(HalfedgeDS& hds) override {

		CGAL::Polyhedron_incremental_builder_3<HalfedgeDS> builder(hds,true);
		const auto& points= polyhedron.points;
		const auto& polygons = polyhedron.faces;

		builder.begin_surface(points.size(),polygons.size());
		for(const auto& p: points)
			builder.add_vertex(p);

		for(const auto& pg: polygons) {
			if(builder.test_facet(pg.begin(),pg.end()))
				builder.add_facet(pg.begin(),pg.end());
		}
		builder.end_surface();
[...]

PR #7274, opened 2023-02-15
Nef_3: Use the skipVEL parameter to ensure result is always an SFace

GilesBathgate:
## Summary of Changes

This sets the skipVEL (Skip SVertices SEdges and SLoops) parameter to true, when calling the locate method in `get_visible_facet` of SNC_const_decorator.  Since it's trying to locate an Sface, it makes no sense for locate to match SVertices, SEdges or SLoops, and if it does the assertion in get_visible_facet will fail.

## Release Management

* Affected package(s): Nef_3
* Issue(s) solved (if any): #7271 

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_const_decorator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_const_decorator.h
@@ -269,7 +269,7 @@ public:
     sp = Infi_box::simplify(sp);
     CGAL_NEF_TRACEN( "Locating "<<sp <<" in "<<v->point());
     SM_point_locator L(&*v);
-    Object_handle o = L.locate(sp);
+    Object_handle o = L.locate(sp, true);
 
     SFace_handle sf;
     if(!CGAL::assign(sf,o)) {


GilesBathgate:
As was previously stated:

> In commit https://github.com/CGAL/cgal/commit/ebd5e00ba8e5041e64b415eb9421f89045591ac9 the value of true was passed to L.locate() which skips needless tests.  
> Applying the same change to the get_visible_facet method of SNC_const_decorator makes the assertion exception never occur.
> @afabri However I am not sure if it would be correct to apply the same change here or not.

GilesBathgate:
> @GilesBathgate your reasoning makes sense. Did you close the PR out of no-reply-fatigue ?

I am unsure, but I will re-open.

afabri:
Going up in the call stack we find [this comment](https://github.com/CGAL/cgal/blob/9601e967245681fa5344488cbceebd955050e087/Nef_3/include/CGAL/Nef_3/SNC_external_structure.h#L869).  Reading it, just skipping is maybe not the right solution.

GilesBathgate:
@afabri It always returns into `f_below` which is a `Halffacet_handle`, so either the comment is misleading or there is also a type error?

PR #7278, opened 2023-02-20
Mesh_3: Add back missing consts

sloriot:


lrineau:
We should be careful. That change may break user-code, with domain or criteria classes with non-const call operators.

sloriot:
The constness was removed from 5.5. The impl functions uses `const&`

sloriot:
So I somehow restore something that was mistakenly removed

lrineau:
> The constness was removed from 5.5. The impl functions uses `const&`

According to git blame, the culprit PR was https://github.com/CGAL/cgal/pull/6642, merged in `master` for 5.6, not 5.5.

sloriot:

PR #7279, opened 2023-02-20
Polyhedron demo - fix the bbox of `scene_c3t3_item`

janetournois:
## Summary of Changes

Even if the c3t3 has no complex cells, there may be facets in complex, edges in complex, vertices in complex, or even just vertices of the underlying triangulation, possibly with weights.
Hence, the bbox should not be the default when `cells_in_complex()` is empty. This PR adds a fallback to the bbox of the underlying triangulation vertices.

## Release Management

* Affected package(s): Demo


sloriot:

PR #7293, opened 2023-03-02
Straight_skeleton_2: File Format Description Correction

bvessels:
## Corrected Clockwise-ness intellisense for >> operator overload

The >> operator overload is incorrect. The following is an example of the contents of a .txt file in the correct format for use with the CGAL show_straight_skeleton.cpp and show_offset_polygon.cpp in the [CGAL 5.5.2 examples.](https://github.com/CGAL/cgal/releases/download/v5.5.2/CGAL-5.5.2-examples.zip)
```
4
0 0
1 0
1 1
0 1
1
4
0.125 0.125
0.875 0.125
0.875 0.875
0.125 0.875
```

## Release Management

* Affected package(s): CGAL 5.5.2 Examples


MaelRL:
Thanks for the patch, but the hole being clockwise is correct.

It is https://github.com/CGAL/cgal/commit/875f8236e20c67a2287ec6850c11a7cc23c97905 that is bad. In `Show_straight_skeleton.cpp`, it flips the orientation of the hole if it is bad:
```
        if (it->is_counterclockwise_oriented())
          it->reverse_orientation();
```

and in `Show_offset_polygon.cpp`, it doesn't have any default argument so you can't see it would break, see https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-188/Straight_skeleton_2_Examples/TestReport_Christo_MSVC-2019-Community-Release.gz:
```
- Straight_skeleton_2_Examples/ProgramOutput.execution___of__Show_offset_polygon

Computes the interior offset of a polygon with holes and draws the result in an EPS file.

Usage: show_offset_polygon <intput_file> [offset_distance] [output_eps_file]

       intput_file  Text file describing the input polygon with holes.
         (See inputfile_format.txt for details)
       offset_distance [default=0.25].
       output_file     [default='innput_file.offset.eps']
```

The proper fix is to fix the assertions of https://github.com/CGAL/cgal/commit/875f8236e20c67a2287ec6850c11a7cc23c97905 and give a default argument (input data) to both these executables. Let me know if you would like to do it, otherwise I will fix it.

bvessels:
Ok, I see that now. I've reverted the Intellisense changes for the >> operator and instead have added clarity to the default output for the two examples.  I've also added the input_file_example.txt.  I didn't add it as a default argument because to do so would remove the option for a user to just run the application to see the "help" explanation. 

Thanks for your help!

MaelRL:
Thanks for the changes!

The assertion is now correct, but the old addition of the assertion in https://github.com/CGAL/cgal/commit/875f8236e20c67a2287ec6850c11a7cc23c97905 in `Show_straight_skeleton.cpp` was bad because the code could fix the hole's orientation if it were wrong. So, you should either remove this assertion or remove the now-obsolete code that attempts to reverts the hole orientation. I'd say it is simpler to keep the assertion and not tolerate bad inputs.

sloriot:
Successfully tested in CGAL-5.6-Ic-198

PR #7316, opened 2023-03-09
Use reserve(size+1) before to copy attribute

gdamiand:
This solves a bug in one of my prog, which occurs only for CMap with index.
The problem comes from the copy of existing attributes. The new attribute is allocated through a compact container with index, using an existing attribute as parameter. But the reference becomes invalid when the underlying vector is reallocated.

A simple solution is to use reserve(size+1) before the allocation (thanks @lrineau for the advice).

## Summary of Changes

Add reserve before attribute creation.

## Release Management

* Issue(s) solved (if any): fix a local bug.


sloriot:
Do you have an example showing the bug?

gdamiand:
> Do you have an example showing the bug?

No sorry. I can try to write one.

sloriot:
I'd be interested in being able to reproduce,  like you I'm a but surprised that the patch fixes a bug.

gdamiand:
I added one example showing a bug for index and not for handle; not sure if this is the same bug. This one is not solved by adding cell attribute copy constructor.
The bug occurs for attribute 63, thus when the cc with index is full and reallocated.
But I am not (yet) able to understand why.

gdamiand:
This is a bug; but cmap with index is for now only in master branch.

sloriot:
Successfully tested in CGAL-5.6-Ic-198

PR #7342, opened 2023-03-20
Add example of printing vertex coordinates

oleg-alexandrov:
## Summary of Changes

This adds a very basic example how how to access the coordinates of a vertex. 

## Release Management

* Affected package(s):

Latest source code.

* Issue(s) solved (if any): https://github.com/CGAL/cgal/discussions/7341
* Feature/Small Feature (if any):
* License and copyright ownership: Freely released in the public domain.

PR #7343, opened 2023-03-21
Surface_mesh: Add example for accessing points

afabri:
## Summary of Changes

As pointed out by @oleg-alexandrov the functions to access the coordinates of a vertex are rather hidden.
This PR at the same time adds the missing documentation for `Vertex_around_face_circulator`.

This PR does not replace #7342 but complements it.

## Release Management

* Issue(s) solved (if any): fix #7341

### DIFF
--- a/BGL/include/CGAL/boost/graph/iterator.h
+++ b/BGL/include/CGAL/boost/graph/iterator.h
@@ -980,6 +980,12 @@ faces_around_face(typename boost::graph_traits<Graph>::halfedge_descriptor h, co
   return make_range(I(h,g), I(h,g,1));
 }
 
+/**
+ * \ingroup PkgBGLIterators
+ * A bidirectional circulator with value type `boost::graph_traits<Graph>::%vertex_descriptor` over all vertices incident to the same face or border.
+ * \tparam Graph must be a model of the concept `HalfedgeGraph`
+ * \cgalModels `BidirectionalIterator`
+ */
 template <typename Graph>
 class Vertex_around_face_circulator
 #ifndef DOXYGEN_RUNNING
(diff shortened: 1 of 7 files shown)


sloriot:
Successfully tested in 5.6-Ic-208

BGL/include/CGAL/boost/graph/iterator.h
```diff
@@ -980,6 +980,12 @@ faces_around_face(typename boost::graph_traits<Graph>::halfedge_descriptor h, co
...
 }
 
+/**
+ * \ingroup PkgBGLIterators
+ * A bidirectional circulator with value type `boost::graph_traits<Graph>::%vertex_descriptor` over all vertices incident to the same face or border.
+ * \tparam Graph must be a model of the concept `HalfedgeGraph`
+ * \cgalModels `BidirectionalIterator`
+ */
 template <typename Graph>
 class Vertex_around_face_circulator
```
lrineau, line 990: This documentation of `Vertex_around_face_circulator` is useless because it does not explain how to construct such a circulator. I will merge the PR anyway.

It seems to be a common curve of all circulators and iterators in the BGL: only "model of" and the `value_type` are documented. That is fine for a type that is the return type of a function, because in that case the function is the creator, but `Vertex_around_face_circulator` is not used or produced by any function of the CGAL BGL.

afabri: I had added it because it is mentioned in the [User Manual of `Surface_mesh`.](https://doc.cgal.org/latest/Surface_mesh/index.html#sectionSurfaceMesh_circulators)

PR #7352, opened 2023-03-28
Do not test twice for self-intersections if there is none

sloriot:


sloriot:
Successfully tested in CGAL-5.6-Ic-216

PR #7382, opened 2023-04-12
Define NOMINMAX to work around problem in vtk

afabri:
## Summary of Changes

As in the testsuite we `#include <windows.h>`  this triggers a compilation problem in a vtk header file.
The workaround is to `#define NOMINMAX` in two  CMakeLists.txt files.

## Release Management

* Affected package(s): Polyhedron_demo,  BGL.


lrineau (approved):

janetournois:
I wonder if this macro definition should be added only under the condition of being on a windows platform?

lrineau:
The macro is almost certainly harmless anyway on other platforms.

lrineau (approved):

sloriot:
Successfully tested in CGAL-5.6-Ic-223

PR #7280, opened 2023-02-21
Mesh_3 - improve initialization when features are inserted

janetournois:
## Summary of Changes

When the c3t3 contains no cells after initialization by features protection, it may be needed to add more points for initialization, even when the nb of facets in complex is > 0.
It is the case for example when facets cover a very small part, or too few connected components of the domain.

Note this PR does not fully solve the problem. If, after initialization by features protection, there are facets and cells in one connected component though there are lots, they might not be detected during the refinement process.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #7261
* License and copyright ownership: unchanged

PR #7390, opened 2023-04-17
Mesh 3 - fix initialization

janetournois:
## Summary of Changes

PR #7280 highlighted the fact that the mesh refinement `one_step()` loop may crash if the C3T3 had been corrupted during the initialization step.
Now both `refine_mesh()` and the `one_step()` loop have the exact same behaviour.

This crash was first met in the demo :
- open cube.off
- "Tetrahedral mesh generation -> Generate surface mesh"
- keep all default parameters
- OK

## Release Management

* Affected package(s): Mesh_3

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/C3T3_helpers.h
+++ b/Mesh_3/include/CGAL/Mesh_3/C3T3_helpers.h
@@ -1167,7 +1167,7 @@ private:
      * @brief Updates cell \c ch in c3t3
      * @param ch the cell to update
      * @param update if set to \c false, checking only is done
-     * @return true if \c ch is in c3t3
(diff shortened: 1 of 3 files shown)


sloriot:
Error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-228/Mesh_3/TestReport_Christo_MSVC2017-Debug-64bits.gz)

janetournois:
I can't reproduce the error, even with the seed

janetournois:
I gave it another try with the same options as in the testsuite but still cannot reproduce the error...

sloriot:
Successfully tested in CGAL-5.6-Ic-241

PR #7403, opened 2023-04-23
Replace boost by std

sloriot:
## Summary of Changes

Replace features in boost by features introduces in c++11/c++14


sloriot:
> It seems you need to check the included headers.

I was using `std::empty` and it is available only in c++17

Polyline_simplification_2/include/CGAL/Polyline_simplification_2/simplify.h
```diff
@@ -154,11 +155,11 @@ class Polyline_simplification_2
       (*it)->set_removable(false);
       ++it;
       for(; it != ite; ++it){
-        if((std::next(it) != ite) && (std::prev(it)== std::next(it))){
+        if((boost::next(it) != ite) && (boost::prior(it)==boost::prior(it))){
```
lrineau, line 158: Any idea why the behavior is different?

sloriot: The iterator implementation is most probably buggy (some platforms complains that it is not bidirectional and there are infinite loops on other platforms).

sloriot:
Successfully tested in `CGAL-6.0-Ic-15`

PR #7411, opened 2023-04-26
CMake: allow spaces in git working directory

lrineau:
## Summary of Changes

Imagine you have cloned CGAL in a local directory named `C:\Users\Laurent Rineau\cgal`, before this patch CMake would fail..., because of the space in the full path. Now it works.

## Release Management

* Affected package(s): CGAL (CMake)
* License and copyright ownership: na


lrineau:
@sloriot The patch is almost trivial. I think it can be added to `integration` without fear.

sloriot:
Successfully tested in CGAL-5.6-Ic-240

Issue #7099, opened 2022-11-30
SCP API is deprecated in libssh 0.10


sloriot:
Starting with libssh 0.10 the compilation of `Polyhedron/demo/Polyhedron/Use_ssh.cpp` will trigger depreciation warnings as as the SCP API is deprecated. The recommended replacement is to use the SFTP API.

The online doc is not up-to-date and point to the old version. However if you download libssh from [here](https://www.libssh.org/files/0.10/), you can configure it with cmake and build the target `docs` and will have a local version of the doc in `doc/html`.

I could not quickly find a tutorial about the migration from scp API to sftp API but there should be something somewhere as we are not the only impacted.

sloriot:
@SaillantNicolas is it something that you could have a look at?

PR #7414, opened 2023-04-28
Fix: Use SFTP API instead of SCP API

SaillantNicolas:
## Summary of Changes

SCP API is deprecated in libssh 0.10 and trigger depreciation warnings. So I replaced SCP API with SFTP API

## Release Management

* Issue(s) solved (if any): #7099


sloriot:
Successfully tested in CGAL-5.6-Ic-264 (+manually tested on 2  different machines: one that showed deprecation warning and one that did not)

PR #7417, opened 2023-04-30
Polyhedron Demo: Add reading NIFTI files (*.nii)

afabri:
## Summary of Changes

Add the possibility to read NIFTI files for the Tet mesh generator in case VTK is installed.

I do not understand why I have to choose the plugin when selecting a *nii file in the demo.

## Release Management

### DIFF
--- a/CGAL_ImageIO/include/CGAL/Image_3.h
+++ b/CGAL_ImageIO/include/CGAL/Image_3.h
@@ -150,6 +150,10 @@ public:
   float ty() const { return image_ptr->ty; }
   float tz() const { return image_ptr->tz; }
 
+  float& tx(){ return image_ptr->tx; }
+  float& ty(){ return image_ptr->ty; }
+  float& tz(){ return image_ptr->tz; }
+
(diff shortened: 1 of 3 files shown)


CGAL_ImageIO/include/CGAL/Image_3.h
```diff
@@ -150,6 +150,10 @@ class CGAL_IMAGEIO_EXPORT Image_3
   float ty() const { return image_ptr->ty; }
   float tz() const { return image_ptr->tz; }
 
+  float& tx(){ return image_ptr->tx; }
+  float& ty(){ return image_ptr->ty; }
+  float& tz(){ return image_ptr->tz; }
+
```
lrineau, line 156: That modification is not consistent with the rest of the file.

If that is not a hard requirement, it should be reverted.

afabri: We need a way to override the origin in case it is not in the file.   Alternatively we might add parameters to the constructor.

sloriot:

PR #7419, opened 2023-05-03
Surface_mesh: Fix links to functions replacing deprecated ones

afabri:
## Summary of Changes

[Here ](https://cgal.geometryfactory.com/CGAL/doc/master/Surface_mesh/group__PkgSurfaceMeshIOFuncDeprecated.html#ga34db73ce1c76b70e97bcef45daf63468) the links for  `read_polygon_mesh()` and `write_polygon_mesh()` are missing.  Hopefully it works when we add the parameters. 

## Release Management

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/IO.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/IO.h
@@ -30,7 +30,9 @@ namespace CGAL {
 
 /*!
   \ingroup PkgSurfaceMeshIOFuncDeprecated
-  \deprecated This function is deprecated since \cgal 5.3, `CGAL::IO::read_polygon_mesh()` should be used instead.
+  \deprecated This function is deprecated since \cgal 5.3,
+    `CGAL::IO::read_polygon_mesh(std::istream& is, Graph& g, const NamedParameters& np)`
+    should be used instead.
 */
 template <typename K>
 CGAL_DEPRECATED bool read_mesh(Surface_mesh<K>& sm, const std::string& filename)
@@ -40,7 +42,9 @@ CGAL_DEPRECATED bool read_mesh(Surface_mesh<K>& sm, const std::string& filename)
 
 /*!
   \ingroup PkgSurfaceMeshIOFuncDeprecated
-  \deprecated This function is deprecated since \cgal 5.3, `CGAL::IO::write_polygon_mesh()` should be used instead.
+  \deprecated This function is deprecated since \cgal 5.3,
+    `CGAL::IO::write_polygon_mesh(const std::string& fname, Graph& g,  const NamedParameters& np)`
+    should be used instead.
 */
 template <typename K>
 CGAL_DEPRECATED bool write_mesh(const Surface_mesh<K>& mesh, const std::string& filename)


afabri:
It still does not create a link.  @albert-github any idea what we have to write to get the link right?

albert-github:
When using the master version of doxygen the link is present but not in the 1.8.13 version.
I would have to dive into the problem, but I'm very (understatement) reluctant to do this as the 1.8.13 version is from December 29, 2016 and is long overdue to be updated.
About the update this issue has been brought up already to the CGAL people (mentioned this  even this morning in an email with some CGAL people) but the problem is that there is no time / capacity for this at CGAL.

Maybe a version without the optional parameter works (see in the documentation a few lines below https://cgal.github.io/7419/v0/Surface_mesh/group__PkgSurfaceMeshIOFuncDeprecated.html#gaa741f5463c01ec92df3d7dd61f569997where `read_off` / `read_OFF` is mentioned).
Please have a go at it.

afabri:
You are right that the latest version of doxygen produces c[orrect results](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-240/master/Surface_mesh/group__PkgSurfaceMeshIOFuncDeprecated.html) in the documentation testsuite.  
Are there any known problems that make us reluctant to just switch?

sloriot:
> You are right that the latest version of doxygen produces c[orrect results](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-240/master/Surface_mesh/group__PkgSurfaceMeshIOFuncDeprecated.html) in the documentation testsuite. Are there any known problems that make us reluctant to just switch?

Time. But it is on the roadmap for the n+1 release.

afabri:
Why don't we build the [documentation of master](https://cgal.geometryfactory.com/CGAL/doc/master/Manual/index.html)  with the latest doxygen .  That is what I usually look at.  And if there are imperfections we will see them. And even if users see them this is not really a problem.

albert-github:
@afabri Problem probably is that the user will also see the imperfections with the releases.

@sloriot 
> But it is on the roadmap for the n+1 release.

`n` being a moving target ?

sloriot:
> @afabri Problem probably is that the user will also see the imperfections with the releases.
> 
> @sloriot
> 
> > But it is on the roadmap for the n+1 release.
> 
> `n` being a moving target ?

Yes, 5.6 will keep using 1.8.13 and we will migrate with 6.0.

albert-github:
I just did a quick test and my first suggestion (leaving out the parameter with the default) doesn't look like to work.
I think the solution:
```
\deprecated This function is deprecated since \cgal 5.3, \link CGAL::IO::read_polygon_mesh() `CGAL::IO::read_polygon_mesh()` \endlink should be used instead.
```
does work (at least in my quick tests with the doxygen versions 1.8.13 and the current master working on CGAL master).

sloriot:
The fix does not fix the link issue with 1.8.13 and 1.9.6 does not have any issue. Let's drop this PR and wait for the update of our doc pipeline.

PR #7424, opened 2023-05-04
fix warnings 

lrineau:
fix 2 warnings I got with gcc-13


sloriot:

PR #7427, opened 2023-05-05
Defining a new resource directory to be used with doxygen master

albert-github:
- fix problem in search functionality
- use default doxygen stylesheet and define differences in the extra strylesheet
- remove non used file (compared to the 1.93 resource directory)


sloriot:
Are those changes only for the upcoming 1.9.7 or are they also valid for 1.9.6?

sloriot:
I'm working on your branch right now to move all to 1.9.6. There is a bug that I'm currently tracking.

albert-github:
What kind of bug do you see?

albert-github:
Do you mean that you get an unexpected page and an extra item `Polygon Mesh Processing Reference` in the generated tree?
(file: Polygon_mesh_processing/group__PkgPolygonMeshProcessingRef.html from the overnight build: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-249/master/Polygon_mesh_processing/group__PkgPolygonMeshProcessingRef.html) ?

The mentioned overnight build ([CGAL-5.6-Ic-249](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-249/index.html)) gave:

image

- Should I dive into it?

sloriot:
I'm on it, It is still not clear if the pb is on doxygen side or CGAL (since doc built with CGAL master is fine).

albert-github:
- let me know if I should dive into it.

sloriot:
As far I can see this comes from your modified versions of `header.html` and `header_package.html`. As this goes out of the scope of what I want to achieve today, I'll push my resources for 1.9.6 into another branch and won't mess up with yours.

albert-github:
I think I found the problem the line
```
<script src="modules.js" type="text/javascript"></script>
```
is missing.

sloriot:
I actually was confused by the title of the PR and thought that 1.9.6 was the latest release but it is actually 1.9.7. I think I fixed the issues you were mentioning in our [1.9.6](https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-5.6-Ic-258/output2/Manual/index.html) (even though there might still be a tiny issue with the style of the search bar). Would you nevertheless recommend to switch to 1.9.7 right away? (there are issues in the testsuite that I just fixed caused by a bad merge).

albert-github:
The doxygen version 1.9.7 was released May 18, 2023 (last Thursday) so a bit of confusing is understandable.
I think that it would be wise to switch straight away to 1.9.7 (at least when there are regression problems between 1.9.6 and 1.9.7).

sloriot:
The resources files from 1.9.6 in the master branch seems to work fine with 1.9.7 (timestamp is deactivated for now). Shall I update this branch to move your changes into 1.9.6 so that we look at the diff and discuss it?

albert-github:
Not sure if you should move it into the 1.9.6 directory as there are some changes is some settings (e.g the mentioned timestamp) I think it would be better to have the 1.9.6 just for the 1.9.6 version of doxygen and the 1.9.7 for the master version.  As soon as you are confident with the 1.9.7 one could remove the 1.9.6.

sloriot:
Replaced by #7758

PR #7437, opened 2023-05-10
Activate CDT2 for face triangulation

sloriot:
(master only)


sloriot:
Successfully tested in CGAL-5.6-Ic-248

PR #7439, opened 2023-05-12
Introduce remove_all_elements(Graph)

MaelRL:
## Summary of Changes

Using clear() removes mesh properties, which we might not want to do. The code will still remove everything, but iteratively.

TODO:
- [x] Small feature

## Release Management

* Affected package(s): `BGL`, `Polygon_mesh_processing`
* Issue(s) solved (if any): -
* Documentation: [link](https://cgal.github.io/7439/v0/BGL/group__PkgBGLHelperFct.html#gac7e199820c95ed1fc6ab536750749358)
* Feature/Small Feature (if any): [link](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/remove_all_elements) **pre-approved on 2024/04/04**


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -1861,12 +1861,6 @@ bool remove_degenerate_faces(const FaceRange& face_range,
   if(degenerate_face_set.empty())
     return true;
 
-  if(degenerate_face_set.size() == faces_size)
```
sloriot, line 1864: Thinking about it, I think I would keep that and update the clear function for Surface_mesh to call the version that is keeping properties. as `clear()` is meant to be used by CGAL developers within there algo, it makes senses not to mess up with user properties.

BGL/include/CGAL/boost/graph/helpers.h
```diff
@@ -960,24 +941,83 @@ void swap_edges(const typename boost::graph_traits<FaceGraph>::halfedge_descript
...
+ * up to the user to e.g. clean garbage or remove internal property maps (if relevant, and desired).
+ *
+ * @tparam FaceGraph model of `MutableHalfedgeGraph` and `MutableFaceGraph`
+ *
+ * @param g the graph whose elements will be removed
+ *
+ * @sa `CGAL::clear()`
+ **/
+template<typename FaceGraph>
+void remove_all_elements(FaceGraph& g)
```
sloriot, line 958: I would update `Surface_mesh` package to have its own overload.

sloriot: Polyhedron could also.

sloriot:

BGL/include/CGAL/boost/graph/helpers.h
MaelRL: > I would update `Surface_mesh` package to have its own overload.

Like [146f9ec](https://github.com/CGAL/cgal/pull/7439/commits/146f9ecb21db4815354058f6e061ccffaf93df9d)? Not sure that I like specializing BGL functions that aren't trivial, it defeats the point of the common API.

sloriot: Yes. I get your point but for dynamic property map we also have something taking advantage of the index based of surface mesh so it is somehow similar. In both cases it is done for efficiency reasons.

sloriot:

sloriot:
Successfully tested in CGAL-6.0-Ic-64

sloriot:
Ready to be merged

PR #7443, opened 2023-05-15
Polyhedron demo - fix dihedral angles histogram

janetournois:
## Summary of Changes

Before this PR, the histogram of dihedral angles displayed in the demo was the histogram of all angles of all tetrahedra.
With this PR, only the angles of tetrahedra with a non-zero Subdomain_index will be computed and used to build the histogram.
Note that the tetrahedra with Subdomain_index zero
* are not displayed at all, so it is made consistent
* are not treated (on purpose) by Delaunay refinement, so they usually are of bad quality

## Release Management


sloriot:
Successfully tested in CGAL-5.6-Ic-254

PR #7449, opened 2023-05-17
Update settings in respect to `TIMESTAMP`

albert-github:
Due to the PR https://github.com/doxygen/doxygen/pull/10045 the `HTML_TIMESTAMP` and `LATEX_TIMESTAMP` settings have been replaced by the setting `TIMESTAMP`. Adjusting the file used to test CGAL against the master version of doxygen.

### DIFF
--- a/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.9.3/BaseDoxyfile.in
@@ -583,6 +583,14 @@ LOOKUP_CACHE_SIZE      = 0
 
 NUM_PROC_THREADS       = 1
 
+# If the TIMESTAMP tag is set different from NO then each generated page will
+# contain the date or date and time when the page was generated. Setting this to
+# NO can help when comparing the output of multiple runs.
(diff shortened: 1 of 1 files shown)


sloriot:
I don't think this PR matters anymore since timestep is kept as default an is no longer in the CGAL config files.

albert-github:
Indeed this does not really matter anymore as in the 1.9.6 BaseDoxyfile.in the `*_TIMESTAMP` is removed.
(In the past the idea I got from some discussions was the idea to keep the Doxyfile as close as possible, regarding the fields to the standard Doxyfile. Theer are a lot of other fields as well that could be removed as they just contain default values or are not used at all).

sloriot:
> Indeed this does not really matter anymore as in the 1.9.6 BaseDoxyfile.in the `*_TIMESTAMP` is removed. (In the past the idea I got from some discussions was the idea to keep the Doxyfile as close as possible, regarding the fields to the standard Doxyfile. Theer are a lot of other fields as well that could be removed as they just contain default values or are not used at all).

I did pay attention to keep only values that are different from the default. If you think I did a mistake please let me know.

PR #7455, opened 2023-05-18
Remove usage of boldsymbol in formulas

albert-github:
Only on a few places the `\boldsymbol` is used in formulas, this has been removed to make it consistent with other packages.

### DIFF
--- a/Barycentric_coordinates_2/doc/Barycentric_coordinates_2/Barycentric_coordinates_2.txt
+++ b/Barycentric_coordinates_2/doc/Barycentric_coordinates_2/Barycentric_coordinates_2.txt
@@ -584,7 +584,7 @@ one should be cautious when using the unnormalized mean value weights. In that c
 The harmonic coordinates are computed by solving the Laplace equation
 
 <center>
-\f$\Delta \boldsymbol{b} = \boldsymbol{0}\f$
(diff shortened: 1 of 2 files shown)


afabri (approved):

afabri:
When I saw `\mathbb`  [here](https://github.com/CGAL/cgal/blob/master/Mesh_3/doc/Mesh_3/Concepts/BisectionGeometricTraits_3.h#L15) I looked up this PR, but I guess `\mathbb` is ok. Or do we have a better maco for the Real Number letter?

albert-github:
@afabri 
A quick mockup (in Barycentric_coordinates_2/index.html#title19):
image

the first version is with `\boldsymbol`
the second version is as proposed in this PR
the third version is with `\mathbb`

A bit further down the page we have:
image

which is, originally, already like the version in this PR.

albert-github:
I checked the repository and as far as I can see there is none for it defined. There are some macros defined (a.o. some lowercase bold face and some uppercase roman.
A quick mockup:

image

first line `\mathbb`
second line `\mathrm`
third line `\mathbf`

So probably there will be different versions used throughout CGAL.

sloriot:
Successfully tested in CGAL-5.6-Ic-257

PR #7456, opened 2023-05-19
Upgrade FindTBB.cmake

### LINKED ISSUE #7349: CGAL's FindTBB.cmake creates target TBB::tbb
## Issue Details

In another program that uses CGAL we have the issue that in cmake, `if(TARGET TBB::tbb)` returns true even though TBB is not present on this system, causing the build to fail. That seems wrong.

We can work around it by `find_package(TBB CONFIG)` (we don't mind not supporting old TBB versions). Is that considered safe with respect to CGAL's use of TBB?

## Source Code
CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.8)
project(Plouf)
find_package(TBB)
if (TARGET TBB::tbb)
  message("yes")
else()
  message("no")
endif()
```
```sh
cmake . -DCMAKE_MODULE_PATH=/usr/local/lib64/cmake/CGAL
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): linux 64
* Compiler: gcc
* CGAL version: 5.4.1
* Other libraries versions if used (Eigen, TBB, etc.): CMake 3.16.2, no TBB

lrineau:
## Summary of Changes

Upgrade `FindTBB.cmake` with a copy of https://github.com/NVIDIA/thrust/blob/9f1cddc62b1bebbaeb8d07d4476c285be5aa0adc/thrust/cmake/FindTBB.cmake.

Fixes #7349.

## Release Management

### DIFF
--- a/Installation/cmake/modules/FindTBB.cmake
+++ b/Installation/cmake/modules/FindTBB.cmake
@@ -195,6 +195,25 @@ endmacro()
 # Get path, convert backslashes as ${ENV_${var}}
 getenv_path(TBB_ROOT)
 
+#start with looking for TBB_DIR and TBB_ROOT
+if((TBB_ROOT OR "$ENV{TBB_ROOT}" OR "$ENV{TBB_DIR}" ) AND NOT TBB_FOUND)
+  find_package(TBB QUIET NO_MODULE NO_CMAKE_SYSTEM_PATH NO_SYSTEM_ENVIRONMENT_PATH)
+endif()
+if(TBB_FOUND)
+  return()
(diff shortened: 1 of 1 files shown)


sloriot:
TBB config error here:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-257/Mesh_3_Examples/TestReport_lrineau_Fedora-rawhide.gz
What surprises me is that it is mentioning a FindTBB from the system and not CGAL.

sloriot:

PR #7461, opened 2023-05-22
Inconsistent return type in Kernel_d concepts

mglisse:
## Summary of Changes

Reported by Michael Kerber: the concept https://doc.cgal.org/latest/Kernel_d/classKernel__d_1_1Contained__in__simplex__d.html documents a return type of Bounded_side but all implementations return bool instead. This is particularly confusing because Bounded_side could have made sense, and you can easily write code that compiles and gives nonsensical results.

## Release Management

* Affected package(s): Kernel_d


sloriot:
Successfully tested in CGAL-5.6-Ic-261

PR #7463, opened 2023-05-23
Polygonal Surface Reconstruction: adapted polygonal surface reconstruction to updated api of region growing

soesau:
## Summary of Changes

adapted polygonal surface reconstruction to updated api of region growing

## Release Management

* Affected package(s): Polygonal Surface Reconstruction


sloriot:
Successfully tested in CGAL-5.6-Ic-261

PR #7467, opened 2023-05-24
Stream support:  Upgrade to recent version of lib3MF

afabri:
## Summary of Changes

Upgrade to the latest version.

## Release Management


sloriot:
Maybe we should have a reader that does not depend on lib3MF. Extracting mesh information should not be that hard. Maybe we can assign this task to @SaillantNicolas ?

PR #7480, opened 2023-05-30
Add missing AppendedData closing tag, so vtu binary file contains valid xml

rdengate:
## Summary of Changes

Currently the binary vtu file written out by CGAL doesn't contain a closing XML tag for AppendedData. Meshio is unable to parse the vtu file, failing instead with an xml.etree.ElementTree.ParseError. Once the missing tag is added, meshio is able to parse the file correctly.

This bug doesn't affect the ascii vtu, only the binary version.

### DIFF
--- a/SMDS_3/include/CGAL/IO/output_to_vtu.h
+++ b/SMDS_3/include/CGAL/IO/output_to_vtu.h
@@ -333,7 +333,7 @@ void output_to_vtu_with_attributes(std::ostream& os,
     os << "<AppendedData encoding=\"raw\">\n_";
     write_c3t3_points(os,tr,V); // fills V if the mode is BINARY
     write_cells(os,c3t3,V);
-    for(std::size_t i = 0; i< attributes.size(); ++i)
+    for(std::size_t i = 0; i< attributes.size(); ++i) {
       switch(attributes[i].second.which()){
       case 0:
         write_attributes(os, *boost::get<const std::vector<double>* >(attributes[i].second));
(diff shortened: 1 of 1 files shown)


afabri:
Thank you for the bug report and proposed fix.   This should go into the branch `cgal/5.5.x-branch` instead of `master`.

afabri:
We should also add a testcase.

rdengate:
Thanks. I will fix this on Thursday.

sloriot:
If find the extra `_` fishy line 333:
```
os << "<AppendedData encoding=\"raw\">\n_";
```

sloriot:
> Thank you for the bug report and proposed fix. This should go into the branch `cgal/5.5.x-branch` instead of `master`.

The file was moved into a new package so backporting is not "trivial". We can keep targetting master IMHO.

rdengate:
> I find the extra `_` fishy line 333:
> 
> ```
> os << "<AppendedData encoding=\"raw\">\n_";
> ```

It is a bit fishy! However, also vitally important:

meshio/_vtu.py lines 176-178:

```
# The appended data always begins with a (meaningless) underscore.
if appended_data[0] != "_":
    raise ReadError()
```

and then it's used several more times. I don't know if programs other than
meshio use the underscore, but I wouldn't rule it out.

rdengate:
I'm working on a test case, but it'll probably not be finished before next Tues, and it won't be exhaustive. Is that okay? I'm testing whether the XML tags are all there for both the ascii and binary versions, and ignoring the question of whether the binary blob is valid. There are no existing test cases that cover vtu writing that I could find.

afabri:
Do you get the exception in a vtk library function when reading an incorrect file?   You are free to use such a function in the testsuite.

rdengate:
Thanks, I'll take a look at that one.

PR #7488, opened 2023-06-01
Corefinement: remove incorrect early exit

sloriot:
Always use classification on all patch polylines as an impossible operation flag could be set.


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Corefinement/Face_graph_output_builder.h
```diff
@@ -1052,16 +1052,18 @@ class Face_graph_output_builder
...
-                !patch_status_was_not_already_set[3])
-              continue; // all patches were already classified, no need to redo it
+             if (!patch_status_was_not_already_set[0] &&
+                 !patch_status_was_not_already_set[1] &&
+                 !patch_status_was_not_already_set[2] &&
+                 !patch_status_was_not_already_set[3])
+               continue; // all patches were already classified, no need to redo it
           }
-#endif
+*/
```
albert-github, line 1066: Would be clearer to use here `#if 0` / `#endif`

sloriot:

PR #7490, opened 2023-06-01
Alpha Wrapping: minor cleanup

afabri:
## Summary of Changes

No need to set the precision or define a namespace alias.

## Release Management

* Affected package(s): Alpha_wrap_3


sloriot:

PR #7448, opened 2023-05-16
Triangulation_3: fix a race-condition in `Regular_triangulation_3::insert(begin, end)`.

lrineau:
## Summary of Changes

Fix a race-condition in `Regular_triangulation_3::insert(begin, end)`.

The race condition is in the batch insertion of a range of points, with or without info.

<details><summary><h3>Details...</h3> (click to open/close)</summary>

In the code of `insert(begin, end)`, after the spatial sorting, the points are inserted using a `tbb::parallel_for`. The insertion object used by the paraller-for inserts a sub-range of points, and maintain a thread-local `hint` of type `Vertex_handle`.

The problem is that one needs to call `hint->point()` **before** anything can be locked, and it is possible that the `hint` vertex is being modified by another thread.

The solution is to store the `Vertex_handle` *and its points* in the thread-local cache.

</details> 

## Release Management

* Affected package(s): Triangulation_3 
* Issue(s) solved (if any): fix #7295.
* License and copyright ownership: maintenance by GeometryFactory

PR #7496, opened 2023-06-05
Regular_triangulation_3: fix another race-condition

### LINKED ISSUE #7295: Potential race-condition in Triangulation_3
Reopened issue: see https://github.com/CGAL/cgal/issues/7295#issuecomment-1573284097.

---

This PR reveals the following error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-Ic-184/Triangulation_3/TestReport_gimeno_ArchLinux-clang.gz and it seems related to the parallel insertion with a pb with a ref counted object. I cannot reproduce it locally. @mglisse does that ring a bell?

_Originally posted by @sloriot in https://github.com/CGAL/cgal/issues/7115#issuecomment-1442913374_

lrineau:
(Followup to the PR #7448.)

## Summary of Changes

During the insert or removal of points, in a regular triangulation, vertices can be destroy in the operation. During the removal of a range of vertices, that is obvious, but because of hidden points, even the insertion of new points can trigger the removal of vertices.

If the code keeps a `Vertex_handle hint` during the algorithm, that vertex can be destroyed, by another thread, and in particular its points can be destroyed. As the locking uses points, one have store store the points alongside the `Vertex_handle`, because the code `this->try_lock_point(hint->point())` accesses the point via an indirection *before* the lock.

With `Epick`, that was not detected, because the destruction of a point from `Epick` does nothing (the coordinates stay in memory). But with `Epeck`, a destruction of a point means that the `Handle` is assigned to a null pointer.

I have fixed one source of race-condition, and I have executed more that 10k runs without any assertion. But I have the impression that the code is still not correct, around those lines: 

https://github.com/lrineau/cgal/blob/8cd8bc7b0604d03511bb318ab2dc4c050e214562/Triangulation_3/include/CGAL/Regular_triangulation_3.h#L2622-L2626

The call to `finite_vertices_begin()` is not thread-safe: the first vertex might be under destruction by another thread. Or maybe destroyed between that call and the access to `hint->point()`. @MaelRL @sloriot I would like to discuss the issue.

## Release Management

* Issue(s) solved (if any): fix #7295

### DIFF
--- a/STL_Extension/include/CGAL/Spatial_lock_grid_3.h
+++ b/STL_Extension/include/CGAL/Spatial_lock_grid_3.h
@@ -209,7 +209,7 @@ public:
     int index_x = static_cast<int>( (CGAL::to_double(point.x()) - m_bbox.xmin()) * m_resolution_x);
     //index_x = std::max( 0, std::min(index_x, m_num_grid_cells_per_axis - 1) );
     index_x =
-      (index_x < 0 ?
+      (index_x < 0 ? /// @TODO: use std::clamp
         0
         : (index_x >= m_num_grid_cells_per_axis ?
             m_num_grid_cells_per_axis - 1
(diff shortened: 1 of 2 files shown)


sloriot:

Issue #7421, opened 2023-05-04
WLOP algorithm preserving original normal, colour and intensity map


ElHouas:
## Issue Details

Hello,
Thanks in advance for the help :)

I am processing a point cloud stored in a std::vector<PNCI>, the fields are defined as:

```
// Type declarations.
using Kernel = CGAL::Exact_predicates_inexact_constructions_kernel;
using FT       = typename Kernel::FT;
using Point_3  = typename Kernel::Point_3;
using Vector_3 = typename Kernel::Vector_3;
using Color    = typename std::array<unsigned char, 4>;
using Direction_3 = typename Kernel::Direction_3;

// Point with normal, color and intensity
using PNCI = std::tuple<Point_3, Vector_3, Color, int> ;
using Point_map = CGAL::Nth_of_tuple_property_map<0, PNCI> ;
using Normal_map = CGAL::Nth_of_tuple_property_map<1, PNCI> ;
using Color_map = CGAL::Nth_of_tuple_property_map<2, PNCI> ;
using Intensity_map = CGAL::Nth_of_tuple_property_map<3, PNCI> ;
```

After processing the point cloud, in the final step I want to perform WLOP and preserve the original normal, colour and intensity values. I map back again using std::transform, but the properties are not mapped correctly as I see in the pointcloud colour values that not correspond to the original points. 

I already tried adding normal_map, color_map and intensity_map as values in CGAL::parameters. It did not work...

Here is a snippet of the code I am using for wlop:

## Source Code

```
std::vector<Point_3> output;
  const double retain_percentage = 2;   // percentage of points to retain.
  const double neighbor_radius = 0.4;   // neighbors size.
  CGAL::wlop_simplify_and_regularize_point_set<Concurrency_tag>
  (point_cloud, std::back_inserter(output), CGAL::parameters::point_map(Point_map()).select_percentage(retain_percentage).neighbor_radius(neighbor_radius));

  std::vector<PNCI> point_cloud_filtered;
  point_cloud_filtered.reserve(output.size());
  std::transform(output.begin(), output.end(), point_cloud.begin(),
                std::back_inserter(point_cloud_filtered),
                [](const Point_3& point, const PNCI& pnci) -> PNCI {
                    const Vector_3& normal = get<1>(pnci);
                    const Color& color = get<2>(pnci);
                    int I = get<3>(pnci);

                    return std::make_tuple(point,normal,color,I);
                });
```

## Environment

[...]

PR #7497, opened 2023-06-05
PSP: wlop_simplify() performance improvement

afabri:
## Summary of Changes

While reading code so that I could answer #7421 I noticed that we should call `vector<T>::reserve()` and `vector<T>swap()`.

## Release Management


sloriot:

PR #7498, opened 2023-06-05
Fix link issue with pointmatcher

sloriot:
It's a bug on libpointmatcher side IMO as it should be part of the variables providing deps to link with. But let's test it first before opening a PR.


sloriot:

PR #4389, opened 2019-12-03
Fix  DT3 parallel performance loss (issue #4388)

lrineau:
## Summary of Changes

Fix #4388.

This fixes the performance loss.

With the macro `CGAL_CONCURRENT_COMPACT_CONTAINER_APPROXIMATE_SIZE` defined to `1`, then the `Concurrent_compact_container` implements a function `approximate_size()`, but at a cost of 3% of performance loss, even if the function is not used (to maintain the atomic values).

## Release Management

* Affected package(s): STL_Extension, Triangulation_[23]
* Issue(s) solved (if any): fix #4388 
* License and copyright ownership: maintenance by GF

**PR #4392 is the same patch, but for the branch of 4.14.x.**

PR #7500, opened 2023-06-07
Add stop button for AW3 demo

MaelRL:
## Release Management

* Affected package(s): `Polyhedron`, `Alpha_wrap_3`
* Issue(s) solved (if any): -

### DIFF
--- a/Polyhedron/demo/Polyhedron/Plugins/Alpha_wrap_3/Alpha_wrap_3_plugin.cpp
+++ b/Polyhedron/demo/Polyhedron/Plugins/Alpha_wrap_3/Alpha_wrap_3_plugin.cpp
@@ -485,7 +485,7 @@ public Q_SLOTS:
   // This is for the message box and thread interruption
   void wrapping_done(Wrapper_thread* wrapper_thread)
   {
-    Scene_surface_mesh_item* wrap_item = new Scene_surface_mesh_item(wrapper_thread->wrap);
+    Scene_surface_mesh_item* wrap_item = new Scene_surface_mesh_item(std::move(wrapper_thread->wrap));
     wrap_item->setName(tr("Wrap with alpha %2 offset %3").arg(wrapper_thread->alpha)
                                                          .arg(wrapper_thread->offset));
     wrap_item->setColor(Qt::gray);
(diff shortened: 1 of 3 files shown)


Polyhedron/demo/Polyhedron/Plugins/Alpha_wrap_3/Alpha_wrap_3_plugin.cpp
```diff
@@ -127,85 +169,143 @@ struct Iterative_AW3_visualization_visitor
...
+
+    std::cout << "Wrapping took " << elapsed_timer.elapsed() / 1000. << "s" << std::endl;
   }
+
+public Q_SLOTS:
+  void emit_status()
+  {
+    Q_EMIT status_report(QString("%1 vertices").arg(wrapper.triangulation().number_of_vertices()));
+  }
+
```
lrineau, line 304: That is the only place where I might see an issue. But the Mesh_3 plugin has the same: the timer is created and started in the main thread, so this slot will be called by the main thread. The call `wrapper.triangulation().number_of_vertices()` is not thread-safe. However, the only consequence I can see is a possible incoherent number of vertices during one iteration of the slot.

Polyhedron/demo/Polyhedron/Plugins/Alpha_wrap_3/Alpha_wrap_3_plugin.cpp
```diff
@@ -275,35 +386,138 @@ class Polyhedron_demo_alpha_wrap_3_plugin
...
+  void reset_iterative_wrap_item()
+  {
+    m_iterative_wrap_item = nullptr;
+  }
+
+  // This is for the message box and thread interruption
+  void wrapping_done(Wrapper_thread* wrapper_thread)
+  {
+    Scene_surface_mesh_item* wrap_item = new Scene_surface_mesh_item(wrapper_thread->wrap);
```
lrineau, line 488: To avoid a copy: move the `wrap` surface mesh, that is useless after that call:
```suggestion
    Scene_surface_mesh_item* wrap_item = new Scene_surface_mesh_item(std::move(wrapper_thread->wrap));
```

lrineau (commented):
I have compile and tried the plugin, and reviewed as much code as possible. It seems fine. I have two minor comments.

Polyhedron/demo/Polyhedron/Plugins/Alpha_wrap_3/Alpha_wrap_3_plugin.cpp
lrineau: That is related to the macro `CGAL_CONCURRENT_COMPACT_CONTAINER_APPROXIMATE_SIZE` from PR #4389.

sloriot:

sloriot:
Successfully tested in CGAL-6.0-Ic-15

PR #7508, opened 2023-06-09
Offset Meshing Plugin Improvements

MaelRL:
## Summary of Changes

- fixed broken formula for surface mesh and polygon soup items creating inverted meshes and sometimes triggering Mesh_3 assertions
 - fixed not even actually working for polygon soups...
- fixed applicable() allowing wrong item combinations
- fixed a few crashes (like crashing for non-triangulated inputs)
- fixed some memory leaks
- fixed UI file names (used PMP::iso_remeshing names)
- fixed inflate() needlessly normalizing already normalized normals
- removed useless sqrts in offset meshing implicit function: we just need a label, we don't care about the actual distance
- avoid recomputing bbox way too many times
- cleaned UI
- some more stuff

## Release Management

* Issue(s) solved (if any): -


sloriot:

PR #7512, opened 2023-06-13
create a new CMake function CGAL_setup_CGAL_flags

lrineau:
That new CMake function will ease the work with the Conan CGAL package.


sloriot:

PR #6967, opened 2022-10-14
GSoC'22: New CGAL's design to handle basic viewers 

gdamiand:
Fixes https://github.com/CGAL/cgal/issues/6943
Replace PR #6921

Mostafa Ashraf's GSoC 2022 Submission

Mentor: @gdamiand
GitHub Page: https://github.com/Mostafa-ashraf19
LinkedIn: [Mostafa Ashraf](https://www.linkedin.com/in/mostafa-ashraf-a62807142/)

Link to the [Reference Manual](https://cgal.github.io/6967/v1/Basic_viewer/group__PkgBasicViewerRef.html)

## TODOs:
@gdamiand

- [x] Add Custom KeyPressEvent SimpleSurfaceMeshWithSmallFacesViewerQt, draw_surface_mesh_small_faces.h.
- [x]  Continue Documenting the project.
- [x] Add something to `CHANGES.md`
- [x] check branch size (for @sloriot)
## Review
File to review:

- [x] Boolean_set_operations_2 draw_polygon_set_2.h
- [x] BGL draw_face_graph.h
- [x] GraphicsView Buffer_for_vao.h 
- [x] GraphicsView Drawing_functor.h 
- [x] GraphicsView Graphic_buffer.h 
- [x] GraphicsView Basic_viewer_qt.h
- [x] Linear_cell_complex Viewer.h
- [x] Linear_cell_complex Viewer.cpp
- [x] Linear_cell_complex typedefs.h
- [x] Linear_cell_complex draw_linear_cell_complex.h
- [x] Nef_3 draw_nef_3.h
- [x] Periodic_2_triangulation_2 draw_periodic_2_triangulation_2.h
- [x] Point_set_3 draw_point_set_3.h
- [x] Polygon draw_polygon_2.h
- [x] Polygon draw_polygon_with_holes_2.h
- [x] Polyhedron draw_polyhedron.h
- [x] Straight_skeleton_2 draw_straight_skeleton_2.h
- [x] Surface_mesh/examples/Surface_mesh/draw_surface_mesh_small_faces.h
- [x] Surface_mesh draw_surface_mesh.h
- [x] Surface_mesh_topology/examples/Surface_mesh_topology/draw_facewidth.h
- [x] Surface_mesh_topology/examples/Surface_mesh_topology/unsew_edgewidth_repeatedly.cpp
- [x] Surface_mesh_topology draw_face_graph_with_paths.h
- [x] Triangulation_2 draw_triangulation_2.h
[...]

PR #7523, opened 2023-06-14
Fix typos in the user manual for the surface mesh segmentation pkg

nmnobre:
Hi @sloriot,

New user and new contributor here, so I hope I'm doing this right. :)

## Summary of Changes

The description for the second term of the energy function used for hard clustering should, I believe, say its contribution increases for facets sharing a sharp and concave edge when they belong to _different_ clusters.
The remaining tweaks are just for improved readability.

## Release Management

* Affected package(s): Triangulated Surface Mesh Segmentation


sloriot:
One small comment: if you want this to be released with the next release, it should be based on the branch `5.6.x-branch`.

See: https://github.com/CGAL/cgal/wiki/Source-Code-Management-with-Git#rebase-an-existing-bug-fix-branch

nmnobre:
> One small comment: if you want this to be released with the next release, it should be based on the branch `5.6.x-branch`.

Done.

Surface_mesh_segmentation/doc/Surface_mesh_segmentation/Surface_Mesh_Segmentation.txt
```diff
@@ -216,7 +217,7 @@ When using a `Surface_mesh`, you can use the built-in property mechanism.
 <!-- \subsection SMSRuntime Runtime of the functions sdf_values() and segmentation_from_sdf_values() -->
 
 The following tables provide the runtime of the functions `sdf_values()` and `segmentation_from_sdf_values()`.
-The results were produced with the release 4.4 of \cgal, on an Intel i7 3.2 Ghz laptop with 8 GB RAM, compiled by Visual C++ 2010 with /O2 option.
+The results were produced with the release 4.4 of \cgal, on an Intel i7 3.2 GHz laptop with 8 GB RAM, compiled by Visual C++ 2010 with /O2 option.
```
albert-github, line 220: I see a similar problem in Spatial_searching/doc/Spatial_searching/Spatial_searching.txt

nmnobre: Done.

nmnobre:
Hi @sloriot,

Slightly unrelated, but I thought I could avoid opening a new issue.
How do you usually visualise the results of a surface mesh segmentation?
In other words, how did you generate the nice pictures on the user manual? :)

-Nuno

sloriot:
We are using the Polyhedron demo. It is located in `demo/Polyhedron` in a flat release or `Polyhedron/demo/Polyhedron` in a git checkout. You can get a precompiled version for windows [here](https://www.cgal.org/demo/5.6/polyhedron_3.zip). Dlls to be put in the same directory as `Polyehdron_3.exe` can be downloaded [here](https://www.cgal.org/demo/5.6/CGAL-demoDLLs.zip).

afabri:
@gdamiand does the `draw()`  function allow to specify per face colors with a property map of a `Surface_mesh`?

afabri:
So we should update the [documentation](https://cgal.geometryfactory.com/CGAL/doc/master/Surface_mesh/group__PkgDrawSurfaceMesh.html)   and maybe add an example that uses the `draw()` function.   @MaelRL just pointed out that we also have a function to produce a color map for a range of indices. However that is  in the Polyhedron demo and not a documented function.

gdamiand:
Indeed @afabri But I suggest to update the documentation after the switch to the new basic viewer api  cf #6967  (I hope submit the small feature soon...).

afabri (approved):

nmnobre:
Hey,

Sorry for taking advantage of the open PR for adding some extra commits while it wasn't approved by a maintainer.
I'll stop now. I just have a final question, in the following manual, do we want to say `Facet_handle` instead?

https://github.com/CGAL/cgal/blob/4673df281a195fa7977b242d8e6fe57f8e241b67/BGL/doc/BGL/graph_traits.txt#L62

I know `Face_handle` exists, but it doesn't seem to be documented?

sloriot (approved):

sloriot:
Successfully tested in CGAL-6.0-Ic-6

PR #7416, opened 2023-04-28
CGAL 6.0 requires c++17 

sloriot:

PR #7526, opened 2023-06-15
Using c++17 standard library

sloriot:
use std::shared_ptr, std::optional, std::variant, ...

These are important changes since they appear in the CGAL API and thus are breaking changes.

includes #7416


Stream_support/include/CGAL/IO/io.h
```diff
@@ -186,6 +188,30 @@ class Output_rep
...
+{
+   const std::optional<T>& t;
+
+public:
+  Output_rep( const std::optional<T>& tt) : t(tt) {}
+  std::ostream& operator()( std::ostream& os) const {
+    if (t==std::nullopt) return (os << "--");
+    return (os << t.value()); }
+};
+
```
lrineau, line 202: Identation:
```suggestion
template <class T, class F>
class Output_rep<std::optional<T>, F>
{
  const std::optional<T>& t;

public:
  Output_rep( const std::optional<T>& tt) : t(tt) {}
  std::ostream& operator()( std::ostream& os) const {
    if (t==std::nullopt) return (os << "--");
    return (os << t.value()); }
};

```

Stream_support/include/CGAL/IO/io.h
```diff
@@ -186,6 +188,30 @@ class Output_rep
...
+class Output_rep<std::variant<T...>, F>
+{
+   const std::variant<T...>& t;
+
+public:
+  Output_rep( const std::variant<T...>& tt) : t(tt) {}
+  std::ostream& operator()( std::ostream& os) const {
+    return (os << "--");
+  }
+};
```
lrineau, line 213: Why print `--` and not the contained value? With `std::visit` and a generic lambda.

And indentation again:
```suggestion
template <class ... T, class F>
class Output_rep<std::variant<T...>, F>
{
  const std::variant<T...>& t;

public:
  Output_rep( const std::variant<T...>& tt) : t(tt) {}
  std::ostream& operator()( std::ostream& os) const {
    return std::visit([](auto&& v) {
                        return (os << v);
                      }, t);
  }
};
```

Note that I have not tested that code.

sloriot: That exactly the old behavior with boost. See https://github.com/boostorg/optional/blob/develop/include/boost/optional/optional_io.hpp

lrineau: > Note that I have not tested that code.

... at least the lambda has to catch `os` by reference.

Stream_support/include/CGAL/IO/io.h
```diff
@@ -267,6 +267,19 @@ class Input_rep
...
+{
+  std::optional<T>& t;
+
+public:
+  //! initialize with a reference to \a t.
+  Input_rep( std::optional<T>& tt) : t(tt) {}
+
+  //! perform the input, calls \c operator\>\> by default.
+  std::istream& operator()( std::istream& is) const { return (is >> t.value()); }
+};
```
lrineau, line 285: According to the [documentation of `optional::value()`](https://en.cppreference.com/w/cpp/utility/optional/value):
> If `*this` contains a value, returns a reference to the contained value.
>  Otherwise, throws a [`std::bad_optional_access`].(https://en.cppreference.com/w/cpp/utility/optional/bad_optional_access) exception.

If the `optional` is empty, then `is >> t.value()` will throw.

I suggest:
```suggestion
template <class T>
class Input_rep<std::optional<T>>
{
  std::optional<T>& t;

public:
  //! initialize with a reference to \a t.
  Input_rep( std::optional<T>& tt) : t(tt) {}

  //! perform the input, calls \c operator\>\> by default.
  std::istream& operator()( std::istream& is) const {
    T v;
    if(is >> v) t = v;
    return is;
  }
};
```

Arrangement_on_surface_2/doc/Arrangement_on_surface_2/CGAL/Arr_point_location_result.h
```diff
@@ -1,25 +1,4 @@
 /*!
```
albert-github, line 1: Probably forgot to remove this line.
(see: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-6.0-Ic-21/logs_master/Arrangement_on_surface_2.log)

sloriot:
Tests failing in CGAL-6.0-Ic-21

sloriot:
Successfully tested in CGAL-6.0-Ic-43

PR #7519, opened 2023-06-13
disable MSVC 2015 with CGAL 6.0

sloriot:

PR #7538, opened 2023-06-22
Disable MSVC 2015 starting with CGAL 6.0

sloriot:
Follow up of #7519 that was not done on the right file


lrineau (approved):

sloriot:
Seems to be working now! \o/
https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-18.shtml

Issue #6690, opened 2022-06-21
CGAL Demo bugs


MaelRL:
## Issue Details

- [ ] Inflate mesh with multiple selected items = crash
- [ ] Selection item's move point regularly crashes
- [ ] "Edge Splitted"
- [ ] Join polyhedra -> no visibility arrows
- [ ] Affine transformation on deleted item = crash
- [ ] Heat Method GUI resets the menu when you selected sources

MaelRL:
1. Generate spheres #1 & #2 with the demo's generator. Respective radii 1 and 0.5, 4 iterations of subdivision 
2. Mesh_3 with no sharp features, facet and cell size 0.03
3. segfault

afabri:
After a corefnement Boolean operation we should no longer visualize the two input meshes, or as wireframe.

afabri:
When remeshing a surface the result should be drawn double sided.

afabri:
Concerning the inflate I changed it to only work if a single polyhedron item is selected.  It should be easy to put a loop around.    As an inflated mesh has quickly self intersections I am wondering if the proposed inflate distance should not be a fraction of the average edge length and not a fraction of the bbox diagonal.  @sloriot any opinion?

afabri:
> 1. Generate 2 spheres with the demo's generator. Respective radii 1 and 0.5, 4 iterations of subdivision
> 
>     2. Mesh_3 with no sharp features, facet and cell size 0.03
> 
>     3. segfault

Sqrt3 subdivision?

afabri:
The average angle computation is fixed by #7161

PR #7539, opened 2023-06-22
Display property / Heat method plugins improvements

MaelRL:
## Summary of Changes

Full pass on the display property plugin.

- separated property display and Heat method into two independent plugins
- redesigned UI widgets
- added "Largest Angle per face" and "face areas" to the display properties
- enable using a selection item to select heat method's source vertices
- fixed smallest angle computation
- fixed storing plain wrong display values
- fixed applicable() being too permissive
- fixed zoom not working for property maps other than the plugin's
- fixed zoom not working for simplices other than faces
- fixed zoom not working for point sets
- fixed color map ignoring the user's color choice
- fixed property combobox resetting when they should not
- fixed not resetting colors when items are modified
- fixed broken connections in the heat method plugin
- fixed not actually emitting anything when a list selection is changed
- more fixes / improvements

## Release Management

* Issue(s) solved (if any): partial fix of #6690


sloriot:
```
/home/cgal_tester/build/src/cmake/platforms/Fedora/test/Polyhedron_Demo/Plugins/Display/Heat_method_plugin.cpp:140:10: warning: variable ‘fnormals’ set but not used [-Wunused-but-set-variable]
```

sloriot:
Successfully tested in CGAL-6.0-Ic-22

PR #7552, opened 2023-06-26
Fix initialisation issues in Mesh_3

MaelRL:
## Summary of Changes

- Issue 1 (fbecb18): bad comparison in minimal weight clamping in 2D case, caused protection balls to become hidden.
- Issue 2 (39c0a9f): robust intersection traits returned no intersection for a ray spawned within a triangle, so the `init_c3t3()` could not insert initial points.
- Issue 3 (49bb14b): in the 2D case, the code was exiting too early while adjusting ball sizes before a new point insertion.
- Issue 4 (1f258bc): after 39c0a9f, initialization did not work anymore if the source is on the domain.

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): -
* Feature/Small Feature (if any):  -

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h
+++ b/Mesh_3/include/CGAL/Mesh_3/Protect_edges_sizing_field.h
@@ -633,6 +633,13 @@ insert_point(const Bare_point& p, const Weight& w, int dim, const Index& index,
 {
   using CGAL::Mesh_3::internal::weight_modifier;
 
+#if CGAL_MESH_3_PROTECTION_DEBUG & 1
+  std::cerr << "insert_point( (" << p
+            << "), w=" << w
+            << ", dim=" << dim
+            << ", index=" << CGAL::IO::oformat(index) << ")\n";
+#endif
+
   // Convert the dimension if it was set to a negative value (marker for special balls).
   if(dim < 0)
     dim = -1 - dim;
@@ -709,6 +716,7 @@ smart_insert_point(const Bare_point& p, Weight w, int dim, const Index& index,
             << "), w=" << w
             << ", dim=" << dim
             << ", index=" << CGAL::IO::oformat(index) << ")\n";
+  std::cerr << "triangulation dimension is " << c3t3_.triangulation().dimension() << std::endl;
 #endif
   const Tr& tr = c3t3_.triangulation();
 
@@ -877,10 +885,12 @@ smart_insert_point(const Bare_point& p, Weight w, int dim, const Index& index,
       }
     }
 
+    // Change w in order to be sure that no existing point will be included in (p,w)
     FT min_sq_d = w;
+#if CGAL_MESH_3_PROTECTION_DEBUG & 1
     typename Tr::Point nearest_point;
-    // Change w in order to be sure that no existing point will be included
-    // in (p,w)
+#endif
+
     for ( typename Tr::Finite_vertices_iterator it = tr.finite_vertices_begin(),
          end = tr.finite_vertices_end() ; it != end ; ++it )
     {
@@ -888,7 +898,9 @@ smart_insert_point(const Bare_point& p, Weight w, int dim, const Index& index,
       FT sq_d = tr.min_squared_distance(p, cp(it_wp));
       if(sq_d < min_sq_d) {
         min_sq_d = sq_d;
+#if CGAL_MESH_3_PROTECTION_DEBUG & 1
         nearest_point = c3t3_.triangulation().point(it);
+#endif
       }
     }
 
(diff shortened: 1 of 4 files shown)


sloriot:
Runtime errors in CGAL-6.0-Ic-13 in Generator package (at least)

```
Built target random_points_on_tetrahedral_mesh_3
Expr: c3t3.triangulation().dimension() >= 2
File: /home/sloriot/CGAL/git/integration/Mesh_3/include/CGAL/make_mesh_3.h
Line: 464
Aborted
```

MaelRL:
@janetournois @lrineau  That's actually the configuration that we were discussing: the source of the ray being shot in the C3T3 initialization is a point exactly on the domain. Beforehand, it would find another intersection; now, source of the ray is returned whatever the ray direction, and the initialization only spawns a single point.

lrineau:
You merged `master` but the base branch was `5.4.x-branch`, and the milestone as well.

MaelRL:
> You merged `master` but the base branch was `5.4.x-branch`, and the milestone as well.

Thanks

MaelRL:
@lrineau @janetournois 

[c0d6f13](https://github.com/CGAL/cgal/pull/7552/commits/c0d6f13bd3c1300a12259acffb6de0b1ed1b7d75) is some wip algorithm for polyhedral mesh domain initialization. It replaces ray shooting from the center by a grid sampling, projection onto the surface, and simplification.

Here are some comparisons for blobby and the cheese:

image

Some advantages:
- Every AABB tree call is useful. For ray shooting, you can make a mesh surface / bbox volume ratio as high as you want with needle-like meshes, requiring arbitrarily high amounts of ray shots.
- Better distribution of the initial points (for a simple program like mesh_polyhedral_domain.cpp that does not optimize much at the end and blobby / cheese inputs, we get better angles).
- No visibility issues (especially visible on the cheese for example, where the source of the ray is within a cavity)
- No more issue if the ray's source is on the domain

Some open points:
- Points being on a grid is not so great from a Delaunay point of view. They could be perturbed a bit around the grid vertices.
- The projection tends to create points onto the corner / creases; that's rather nice, but if we have protected, these points will be rejected (but then, we should have enough points already?).
- The algorithmic complexity is bad (does that matter?).

MaelRL:
> Don't we get undesired points in the interior, in case we want as big as possible tetrahedra inside? Ot are they not inserted but only used as source of the ray.

There are no points in the interior for either method. This new approach also do not use any ray shooting, it projects (slightly perturbed) grid points onto the surface using AABB_tree::closest_point(). Below is some image illustrating the pipeline:
- Generate `10*n` grid points (red) + perturb them (slightly less red)
- Project them on the surface (orange)
- Extract `n` sampling points

image

MaelRL:
For the standard 8 points:

image

The initialization of the pruning process I guess got lucky taking a point that is at the tip, this is related to the point above ("Related to above, we could even intentionally keep corners during the simplification process."): maybe we could give priority to points that are far AND that are interesting (on a corner, on a crease). Maybe that doesn't really matter because there are just a few initialization points, so for big meshes it is not significant if you manage to insert a handful of these interesting points.

MaelRL:
> That looks good. In my opinion, once tested, it could be merged into `master` for CGAL-6.0. But not for release branches. It should be only given to given users as an experimental feature, until it is tested for at least one release cycle.

Initialization of 2D polyhedral domain is still broken with my bug fixes and with the ray shooting, hence why I put this change in this PR.

I could do a quick fix for release branches of changing the source of the ray shooting if the sampled point is the source of the ray, and move the work on the newer version to a different PR. Is that OK?

lrineau:
> I could do a quick fix for release branches of changing the source of the ray shooting if the sampled point is the source of the ray, and move the work on the newer version to a different PR. Is that OK?

:+1:

MaelRL:
> I guess we still need something special if we have many small connected components, say a cloud of soap bubbles.

If there aren't too many bubbles, then the projection could find them all. It all depends on the number of initialization points, which give you a lower bound between the distance between two grid points, and thus the distance to separate two connected components.

We could also initialize for each CCs, but that's an ill-posed problem: if you have a bad mesh with many CCs, then you don't want to put points on all of them. And in between the soap bubbles and the broken meshes, cases where it's not obvious what you want to do.

MaelRL:
This PR should be finished now.

Other issues are related to edge protection and will be handled in a different pull request.

lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-27.shtml

lrineau:
This PR was merged into 5.5.x-branch (too late for 5.5.3 though).

Issue #7577, opened 2023-07-05
Compilation errors when using PMP::remove_almost_degenerate_faces with Exact_predicates_inexact_constructions_kernel


wojciech-k-volue:
## Issue Details

Attempting to use the `CGAL::Polygon_mesh_processing::remove_almost_degenerate_faces` function from 5.6-beta1 with a `CGAL::Exact_predicates_inexact_constructions_kernel` kernel results in the following error while compiling with Visual Studio 17.6.1:
```
error C2664: 'int CGAL::Polygon_mesh_processing::internal::Is_cap_angle_over_threshold_impl<CGAL::Simple_cartesian<boost::multiprecision::number<boost::multiprecision::backends::gmp_int,boost::multiprecision::et_on>>>::operator ()(const CGAL::Point_3<Kernel_> &,const CGAL::Point_3<Kernel_> &,const CGAL::Point_3<Kernel_> &,const boost::multiprecision::number<boost::multiprecision::backends::gmp_int,boost::multiprecision::et_on>) const': cannot convert argument 4 from 'T' to 'const boost::multiprecision::number<boost::multiprecision::backends::gmp_int,boost::multiprecision::et_on>'
 with
[         
 Kernel_=CGAL::Simple_cartesian<boost::multiprecision::number<boost::multiprecision::backends::gmp_int,boost::multiprecision::et_on>>
]
and
[
  T=double
]
```

The `CGAL::Polygon_mesh_processing::experimental::remove_almost_degenerate_faces` function from previous versions didn't display this behavior.

## Source Code

Simple example which reproduces the issue:
```cpp
#include <CGAL/Surface_mesh.h>
#include <CGAL/Polygon_mesh_processing/repair_degeneracies.h>
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>

using Kernel = CGAL::Exact_predicates_exact_constructions_kernel;
using Point_3 = Kernel::Point_3;
using Mesh = CGAL::Surface_mesh<Point_3>;

int main()
{
  Mesh mesh;
  CGAL::Polygon_mesh_processing::remove_almost_degenerate_faces(mesh);
}
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 11
* Compiler: Visual Studio 17.6.1
* Release or debug mode: both
* Specific flags used (if any):
* CGAL version: 5.6-beta1
* Boost version: 1.75 and 1.80
* Other libraries versions if used (Eigen, TBB, etc.):

[...]

PR #7588, opened 2023-07-07
add missing partial specialization

sloriot:
Add missing partial specialization for boost's gmp wrapper.

Fixes https://github.com/CGAL/cgal/issues/7577


sloriot:
MP_Float is fine, it has its specialization. Actually, the default should be removed. I don't see any reason to keep it.

sloriot:
You are right MP_FLoat is missing one. I'll fix it.

lrineau:
The bug was reported on 5.6-beta1. Should it be fixed in 5.6?

sloriot:
github is buggy and I cannot change the base for now.

sloriot:
Successfully tested in CGAL-6.0-Ic-24.

PR #7594, opened 2023-07-11
(Really) Disable MSVC 2015 starting with CGAL 6.0 

sloriot:
#7538 followup

one more time 🎵 🎵 🎵 🎵


sloriot:
log of Christo in 6.0-Ic-21 seems to indicate that this really works now. Let's do another round...

sloriot:
6.0-Ic-22 is also good.

PR #7595, opened 2023-07-12
Generalize documentation of Face_filtered_graph

### LINKED ISSUE #7570: BGL - add `Face_filtered_graph` constructor from boolean property map
## Issue Details

Would it be difficult/would it make sense to add a constructor `Face_filtered_graph(const Graph& g, FaceIsSelectedPmap f_bool_pmap)` that constructs the filtered graph wrt selected-or-not status of every facet, similarly to the patch-id based constructor?

cc @sloriot @ipadjen

MaelRL:
## Summary of Changes

The current doc says that the selection patch IDs need to be of type faces_size_type, but this requirement doesn't exist in the code.

The doc is generalized, and the actual requirement (the type must be `Hashable`) is added.

TODO:
- [x] Add a few words in the description of the FFG class

## Release Management

* Affected package(s): `BGL`, `Stream_support`
* Issue(s) solved (if any): fix #7570

### DIFF
--- a/BGL/include/CGAL/boost/graph/Face_filtered_graph.h
+++ b/BGL/include/CGAL/boost/graph/Face_filtered_graph.h
@@ -49,9 +49,11 @@ namespace CGAL {
    * the adapted graph must define a manifold mesh. In order to check that this condition is verified, you can
    * use the function `is_selection_valid()`.
    *
-   * There are two different ways to initialize this class. You can directly provide the set of faces selected, or
-   * if you have a face patch map, select the patches of faces. The latter option is convenient if you want to access
-   * some connected components of a graph after having called `CGAL::Polygon_mesh_processing::connected_components()`.
+   * There are two different ways to initialize this class: you can directly provide a set of selected faces,
+   * or provide a set of patch identifiers as well as a map between faces and patch identifiers.
+   * The latter option is convenient if you want to access some connected components of a graph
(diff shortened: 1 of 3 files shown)


sloriot:
Compilation error: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-22/BGL/TestReport_lrineau_Ubuntu-Latest-NO_DEPRECATED_CODE.gz

MaelRL:
@janetournois What was the use case where you needed a FFG with Booleans? We could add an example as there are currently 3 examples with FFG, but they are all using a std::size_t value type.

sloriot:
Successfully tested in CGAL-6.0-Ic-24.

janetournois:
> @janetournois What was the use case where you needed a FFG with Booleans? We could add an example as there are currently 3 examples with FFG, but they are all using a std::size_t value type.

It was for isotropic remeshing of a selection of faces. @ipadjen do you remember where exactly you needed that?

ipadjen:
It was one of ideas to use it for curvature calculation of a mesh selection. Currently we have the face descriptor for ffg: https://github.com/janetournois/cgal/blob/2bb1e97847204d2ce5acc1025487201f81d7a4dc/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/Adaptive_sizing_field.h#L65-L72

PR #7605, opened 2023-07-19
Fix connect infinite loop in heat method plugin

MaelRL:
## Summary of Changes

The proper fixes would be to prevent the "reset colors" and the creation of a selection item from emitting "item changed": in neither case the item changed. Future work...

## Release Management

* Issue(s) solved (if any): -


sloriot:

PR #7606, opened 2023-07-19
Improve initialization of polyhedral mesh domains

### LINKED ISSUE #921: Mesh_3: improve initialization for polyhedral mesh domains
See the discussion in CGAL-discuss:
* in Nabble: http://cgal-discuss.949826.n4.nabble.com/CGAL-3D-mesh-generation-prerequisite-condition-td4661693.html
* in the Sympa archives: https://sympa.inria.fr/sympa/arc/cgal-discuss/2016-03/msg00035.html

When there are no 1D features to protect, the initialization of the mesh from a polyhedral domain might not be good enough.

One possible solution, proposed by @afabri
--------------------------------------------------------
Andreas proposed one way to ensure that the initialization points are correctly spread over the whole surface of the domain:
https://gist.github.com/afabri/1bbf46c6fd4f742bd982

Another possibility: check the result
---------------------------------------------
There should be a way to:
  - check if the output if good enough (Haussdorf distance?)
  - then possibly re-insert new points, and re-run the meshing process

MaelRL:
## Summary of Changes

The current method uses ray shooting from a single point, which has a number of limitations:
- it can only initializes with what is visible (the first intersection is used)
- many rays may be required if the mesh has a small area

This PR uses projection to initialize uniformly over the domain. The algorithm is:
- Generate `10*n` grid + perturb them (red below)
- Project them on the surface (orange)
- Extract `n` sampling points (green)

image

Some advantages:
- Every AABB tree call is useful. For ray shooting, you can make a mesh surface / bbox volume ratio as high as you want with needle-like meshes, requiring arbitrarily high amounts of ray 
- Better distribution of the initial points (for a simple program like mesh_polyhedral_domain.cpp that does not optimize much at the end and blobby / cheese inputs, we get better angles).
- No visibility issues (especially visible on the cheese for example, where the source of the ray is within a cavity).
- No more issue if the ray's source is on the domain.

Some open questions:
- The projection tends to create points onto the corner / creases; that's rather nice, but if we have protected, these points will be rejected (but then, we should have enough points already?).
- Related to above, we could even intentionally keep corners during the simplification process.
- The algorithmic complexity is bad (does that matter?).

Below are some comparisons for master (red) / branch (green):

image

See also discussions in https://github.com/CGAL/cgal/pull/7552.

TODO:
- [x] first finalize PR #7798 
- [ ] adapt this PR code to other types of domains. The same algo should work for all of them
- [ ] better tests for Mesh_3 init (esp. Polyhedral_complex_mesh_domain_3)

## Release Management

* Affected package(s): `Mesh_3`
* Issue(s) solved (if any): fix #921 ?

### DIFF
--- a/Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
+++ b/Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
@@ -747,7 +747,7 @@ Construct_initial_points::operator()(OutputIterator pts,
     projected_points.emplace(cp_and_p.first, index);
   }
 
-  CGAL_warning(projected_points.size() > n);
+  CGAL_warning(projected_points.size() > static_cast<std::size_t>(n));
 
   // n farthest points
   std::vector<Point_with_index> initial_points;
--- a/Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
+++ b/Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
@@ -697,6 +697,8 @@ Construct_initial_points::operator()(OutputIterator pts,
 
   const Bounding_box bbox = r_domain_.tree_.bbox();
 
+  CGAL::Random& rng = *(r_domain_.p_rng_ != 0 ? r_domain_.p_rng_ : new Random(0));
+
   // The initialization proceeds as follows:
   // - Generate a grid of n' points with n' > n, to get good candidates
   // - Project them onto the surface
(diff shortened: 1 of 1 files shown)


MaelRL:
Relevant is the issue: https://github.com/CGAL/cgal/issues/921, which has some gists where the initialization takes a subset of vertices and grid-simplifies them.

sloriot:
New warnings in CGAL-6.0-Ic-42

- `/mnt/testsuite/include/CGAL/Polyhedral_mesh_domain_3.h:803:40: warning: comparison of integer expressions of different signedness: 'std::set<std::pair<CGAL::Point_3<CGAL::Epick>, int>, std::less<std::pair<CGAL::Point_3<CGAL::Epick>, int> >, std::allocator<std::pair<CGAL::Point_3<CGAL::Epick>, int> > >::size_type' {aka 'long unsigned int'} and 'const int' [-Wsign-compare]`

Windows does not seem to be happy with some modifications too.

Mesh_3/include/CGAL/Polyhedral_mesh_domain_3.h
```diff
@@ -687,64 +687,124 @@ Polyhedral_mesh_domain_3<P_,IGT_,TA,Tag,E_tag_>::
...
+  // @todo trivial to parallelize
+  for(const Point_3& gp : grid_points)
+  {
+    typename AABB_tree_::Point_and_primitive_id cp_and_p = r_domain_.tree_.closest_point_and_primitive(gp);
+    auto index = r_domain_.index_from_surface_patch_index(
+                   r_domain_.make_surface_index(cp_and_p.second /*primitive ID*/));
+    projected_points.emplace(cp_and_p.first, index);
+  }
+
+  CGAL_warning(projected_points.size() > n);
```
sloriot, line 750:
```suggestion
  CGAL_warning(projected_points.size() > static_cast<std::size_t>(n));
```

sloriot:

sloriot:
@MaelRL do you want to merge the branch as is or shall you discuss it with @janetournois and @lrineau ?

MaelRL:
Maybe it makes sense to wait for Ange's work on this topic (or even that this PR would be merged into his branch)?

lrineau:
@ange-clement Can you please have a look at this pull-request, and have the new code be integrated in your pull-request. This pull-request should be merged into yours.

janetournois:
Since this would be the new default behavior for polyhedral domains, I would keep it independent from @ange-clement 's PR. 
Both PRs are about initialization, but they are somehow orthogonal.

janetournois:
I just updated the TODO list for this PR

afabri:
Just wondering if this [Poisson Disk Sampling](http://www.cemyuksel.com/cyCodeBase/soln/poisson_disk_sampling.html) code may be helpful.

lrineau:
This PR has a good heuristic to initialize a polyhedral mesh domain. Now that the small feature of the PR https://github.com/CGAL/cgal/pull/7798 is merged, it must be converted to a custom initializer.

PR #7608, opened 2023-07-19
Polygon repair

kenohori:
## Summary of Changes

New Polygon_repair package, which is under development for GSoC 2023. It includes:
* The main Polygon_repair class with functions to repair Polygon_2, Polygon_with_holes_2 and Multipolygon_with_holes_2.
* Underlying classes to build a constrained triangulation with odd-even counting of constraints and to store the necessary repair info in every face.
* Validation functions for Polygon_2, Polygon_with_holes_2 and Multipolygon_with_holes_2 (not comprehensive)
As well as some additions to the Polygon package:
* Concept and class to represent multipolygons with holes (Multipolygon_with_holes_2)
* Specialisation of CGAL::draw for multipolygons with holes.

## Release Management

* Affected package(s): Polygon_repair, Polygon
* Feature: [Polygon repair](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Polygon_repair), accepted 5 March 2024
* Wiki page: [link](https://cgal.geometryfactory.com/CGAL/Members/wiki/GSoC2023/Polygon_repair)
* Built docs: [Polygon_repair](https://cgal.github.io/7608/v2/Polygon_repair/index.html) and [Multipolygon_with_holes](https://cgal.github.io/7608/v2/Polygon/classCGAL_1_1Multipolygon__with__holes__2.html)


Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -82,6 +82,47 @@ class Polygon_repair_2 {
   typedef CGAL::Constrained_Delaunay_triangulation_2<Kernel, Triangulation_data_structure, Tag> Constrained_Delaunay_triangulation;
   typedef Triangulation_with_odd_even_constraints_2<Constrained_Delaunay_triangulation> Triangulation;
 
+  struct Polygon_compare {
```
afabri, line 86: Reading the code this is a Less and not a Compare.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -82,6 +82,47 @@ class Polygon_repair_2 {
...
+      typename Polygon_2<Kernel, PolygonContainer>::Vertex_iterator vb = pb.outer_boundary().vertices_begin();
+      while (va != pa.outer_boundary().vertices_end() && vb != pb.outer_boundary().vertices_end()) {
+        if (*va != *vb) return *va < *vb;
+        ++va;
+        ++vb;
+      } if (vb != pb.outer_boundary().vertices_end()) return true;
+      else if (va != pa.outer_boundary().vertices_end()) return false;
+      typename Polygon_with_holes_2<Kernel, PolygonContainer>::Hole_const_iterator ha = pa.holes_begin();
+      typename Polygon_with_holes_2<Kernel, PolygonContainer>::Hole_const_iterator hb = pb.holes_begin();
+      while (ha != pa.holes_end() && hb != pb.holes_end()) {
```
afabri, line 111: Apply the operator of `Polygon_compare`  instead of duplicating code.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -82,6 +82,47 @@ class Polygon_repair_2 {
   typedef CGAL::Constrained_Delaunay_triangulation_2<Kernel, Triangulation_data_structure, Tag> Constrained_Delaunay_triangulation;
   typedef Triangulation_with_odd_even_constraints_2<Constrained_Delaunay_triangulation> Triangulation;
 
+  struct Polygon_compare {
+    bool operator()(const Polygon_2<Kernel, PolygonContainer>& pa, const Polygon_2<Kernel, PolygonContainer>& pb) const {
+      typename Polygon_2<Kernel, PolygonContainer>::Vertex_iterator va = pa.vertices_begin();
```
afabri, line 88: Make a typedef instead of repeating complex types for reasons of readability.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
kenohori: Sure, I can do it, but it would be slower since I'd need to do the comparison twice (a < b and b < a) in some cases

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
lrineau: > Make a typedef instead of repeating complex types for reasons of readability.

In modern code, [C++ Code Guidelines recommend](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t43-prefer-using-over-typedef-for-defining-aliases) type aliases (`using Type = ...`) instead of typedefs. As Polygon repair is a new package, all typedefs could be converted to aliases.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -0,0 +1,346 @@
...
+    for (typename std::list<typename Kernel::Point_2>::iterator current_vertex = ring.begin();
+         current_vertex != ring.end(); ++current_vertex) {
+      if (*current_vertex < *smallest_vertex) smallest_vertex = current_vertex;
+    } if (ring.front() != *smallest_vertex) {
+      ring.splice(ring.begin(), ring, smallest_vertex, ring.end());
+    }
+  }
+
+  void compute_hole_nesting() {
+    for (int i = 0; i < number_of_holes; ++i) {
```
afabri, line 237: Why is this loop needed?

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -0,0 +1,346 @@
...
+      nesting.emplace_back();
+    } for (auto const &face: t.finite_face_handles()) {
```
afabri, line 239: Do not start with the `for` after the closing `}` but put a newline for readability.  It is ok for a `do { ..} while(..);`

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -0,0 +1,346 @@
...
+    // int hole_label = -2;
+    // for (auto const &hole: nesting) {
+    //   std::cout << "Hole " << hole_label-- << " contained in polygon(s): ";
+    //   for (auto const &polygon: hole) {
+    //     std::cout << polygon << " ";
+    //   } std::cout << std::endl;
+    // }
+  }
+
+  // Reconstruct multipolygon based on the triangles labelled as inside the polygon
```
afabri, line 256: labelled (UK)  ->  labeled (US)

Polygon_repair_2/examples/Polygon_repair_2/repair_polygon_2.cpp
```diff
@@ -0,0 +1,53 @@
...
+using Kernel = CGAL::Exact_predicates_inexact_constructions_kernel;
+using Point_2 = Kernel::Point_2;
+using Polygon_2 = CGAL::Polygon_2<Kernel>;
+using Polygon_with_holes_2 = CGAL::Polygon_with_holes_2<Kernel>;
+using Multipolygon_with_holes_2 = CGAL::Multipolygon_with_holes_2<Kernel>;
+using Polygon_repair_2 = CGAL::Polygon_repair_2::Polygon_repair_2<Kernel>;
+
+int main(int argc, char* argv[]) {
+
+  for (const auto& file: std::__fs::filesystem::directory_iterator("../../test/Polygon_repair_2/data/in")) {
```
afabri, line 22: Move the code to the `test` directory and write a new small example.

Polygon_repair_2/examples/Polygon_repair_2/repair_polygon_2.cpp
```diff
@@ -0,0 +1,53 @@
...
+    std::cout << "Reading " << file.path().filename() << "..." << std::endl;
+
+    // if (file.path().filename() != "hole-outside.wkt") continue;
+
+    std::string in;
+    std::getline(std::ifstream(file.path()), in);
+    std::istringstream iss(in);
+    std::size_t is_polygon = in.find("POLYGON");
+    std::size_t is_multipolygon = in.find("MULTIPOLYGON");
```
afabri, line 31: Is it intentional that when it is `MULTIPOLYGON` als `is_polygon` will be `true`?

afabri: Ah, it is not a  `bool` but the position.  I was mislead by the `is_` prefix. Maybe swith to `bool`  and do the `==0`  right away.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -0,0 +1,346 @@
...
+            } to_check.front()->processed() = true;
+            to_check.pop_front();
+          }
+
+          break;
+        }
+      }
+    }
+
+    // Create polygons with holes and put in multipolygon
```
afabri, line 305: I think globally there is a lot of copying polygons going on.  In the optimisation phase of the GSoC  we should check where we can use move semantics.

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
kenohori: It creates as many nesting entries as there are holes. They can't be created on the fly since mostly likely we won't find them in order.

kenohori: Agreed in meeting to use resize for this one and other similar operations

Polygon_repair_2/examples/Polygon_repair_2/clipart.cpp
```diff
@@ -0,0 +1,151 @@
+#include <iostream>
```
sloriot, line 1: This should go in test IMO

Polygon_repair_2/examples/Polygon_repair_2/write_labeled_triangulation.cpp
```diff
@@ -0,0 +1,83 @@
...
+  std::ofstream ofs("/Users/ken/Downloads/triangulation.geojson");
+
+  std::string in;
+  std::getline(ifs, in);
+  std::istringstream iss(in);
+  Polygon_repair_2 pr;
+
+  if (in.find("POLYGON") == 0) {
+    Polygon_with_holes_2 p;
+    if (in != "POLYGON()") { // maybe should be checked in WKT reader
```
sloriot, line 35: Do you mean it might be a bug in CGAL or boost?

Polygon_repair_2/include/CGAL/Polygon_repair_2/Multipolygon_with_holes_2.h
```diff
@@ -0,0 +1,143 @@
...
+  void clear() { m_polygons.clear(); }
+
+  Size number_of_polygons() const { return static_cast<Size>(m_polygons.size()); }
+
+protected:
+  Polygons_container m_polygons;
+};
+
+/*!
+This operator exports a multipolygon with holes to the output stream `os`.
```
sloriot, line 92:
```suggestion
exports a multipolygon with holes to the output stream `os`.
```

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
```diff
@@ -0,0 +1,470 @@
...
+//
+// Author(s)     : Ken Arroyo Ohori
+//                 Guillaume Damiand
+
+#ifndef CGAL_POLYGON_REPAIR_2_H
+#define CGAL_POLYGON_REPAIR_2_H
+
+#include <CGAL/license/Polygon_repair_2.h>
+
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
```
sloriot, line 18: do you really need that include?

Polygon_repair_2/test/Polygon_repair_2/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# Created by the script cgal_create_cmake_script
+# This is the CMake script for compiling a CGAL application.
+
+cmake_minimum_required(VERSION 3.1...3.23)
+project(Polygon_repair_2_Tests)
+
+set(CMAKE_CXX_STANDARD 17)
```
sloriot, line 7: this is not needed

kenohori: what should I do to compile with C++17 then?

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
kenohori: Only if we want to have that kernel as a default template parameter, I think?

Polygon_repair_2/examples/Polygon_repair_2/write_labeled_triangulation.cpp
kenohori: It's an unhandled case of invalid input I found while doing tests, but maybe the check shouldn't be there at all. Arguably, it's more of a syntax error than an invalid geometry since the polygon has no outer ring.

What is a bit weird IMO is that `MULTIPOLYGON()` is handled fine but `POLYGON()` isn't.

Polygon_repair_2/test/Polygon_repair_2/CMakeLists.txt
sloriot: If you merge master, then it will be set ON by default

Polygon_repair_2/include/CGAL/Polygon_repair_2/Polygon_repair_2.h
sloriot: we should get it from the polygon, shouldn't we?

Polygon_repair_2/examples/Polygon_repair_2/clipart.cpp
kenohori: What about the clipart data for this?

sloriot: Since we cannot include it, maybe the best is to move the .cpp in benchmark, and add a comment about where to get the data.

Polygon_repair/include/CGAL/Polygon_repair/Polygon_repair.h
```diff
@@ -0,0 +1,715 @@
...
+/// \ingroup PkgPolygonRepairFunctions
+/// repairs polygon `p` using the given rule
+/// \tparam Kernel parameter of the input and output polygons
+/// \tparam Container parameter of the input and output polygons
+///  \tparam Rule must be `Even_odd_rule`
+template <class Kernel, class Container, class Rule>
+Multipolygon_with_holes_2<Kernel, Container> repair(const Polygon_2<Kernel, Container>& p , Rule rule) {
+    CGAL_assertion(false); // rule not implemented
+    return Multipolygon_with_holes_2<Kernel, Container>();
+  }
```
sloriot, line 44:
```suggestion
Multipolygon_with_holes_2<Kernel, Container> repair(const Polygon_2<Kernel, Container>& p , Rule rule);
```

sloriot:
Errors and warnings in CGAL-6.0-Ic-187

afabri:
@kenohori do you find the time to investigate the assertion in the above mentioned testsuite.  I fixed the warning.  The runtime error does not happen on all platforms.

kenohori:
Haven't had the chance to do it, sorry. Things are quite hectic with teaching this quarter.

Debugging this is a bit messy since the output that fails the check is not being printed before the assertion fail. Any ideas to ensure that it is printed? There's a cout before the assertion, see:
https://github.com/kenohori/cgal/blob/gsoc2023-polygon_repair-kenohori/Polygon_repair/test/Polygon_repair/repair_polygon_2_test.cpp#L62

Otherwise, I can try to reproduce it in a VM here, but it will take me some time to set it up.

lrineau:
@sloriot You added https://github.com/CGAL/cgal/labels/TODO, [two weeks ago](https://github.com/CGAL/cgal/pull/7608#event-12026308990). Is there still something to do? The feature itself is accepted.

sloriot:
I reduced the size, you can merge it now (`git diff kenohori/gsoc2023-polygon_repair-kenohori..HEAD` was empty)

MaelRL:
Will there be a news entry?

PR #5693, opened 2021-05-07
SMDS_3: Mesh_data_structure_3 - new package for C3T3

janetournois:
## Summary of Changes

This PR introduces a new package "Tetrahedral Mesh Data Structure 3" (`TMDS_3` for short) that compiles existing code and concepts about the `Mesh_complex_3_in_triangulation_3`.

With this change, the package `Tetrahedral_remeshing` does not depend on `Mesh_3` anymore, but only on `TMDS_3`.

## Release Management

* Affected package(s): Mesh_3, TMDS_3
* Issue(s) solved (if any): fix #4810, #4125, #5789
* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_new_MDS_3_package
* Link to compiled documentation : [SMDS_3](https://cgal.github.io/5693/v11/SMDS_3/index.html)
* License and copyright ownership: unchanged (Inria)

## ToDo :
- [x] keep the `Surface_patch_index`'s from c3t3 to surface mesh in `facets_in_complex_3_to_triangle_mesh()` 
- [x] add figures in the user manual
- [x] reduce branch size (@sloriot)
- [ ] handle rewrite_rules after the release of the doc (the after merge ToDo is to add a ToDo in the release issue)
- [x] fix #6098

Issue #7395, opened 2023-04-18
Improvement of layout of model relations.


albert-github:
After successful implementation of the adjusted documentation approach for refines (see #7236 and #7231), it would also be good to have a look at other possible candidates, especially the `\cgalHasModel` and `\cgalModel`. like in Algebraic_foundations/classFieldNumberType.html:

![image](https://user-images.githubusercontent.com/5223533/232797555-9edb6c91-4d5d-4afd-a646-c2920126a700.png)

and in the "treeview" the part:
![image](https://user-images.githubusercontent.com/5223533/232798752-c20a9246-8f36-4db6-b6ae-11b10dc786f5.png)

The questions that arise:
1. would this be useful, in my opinion YES
2. in general types in "Has Model" should appear in "code" format (so between backticks) should "simple" types like "float" and "double" also appear here in "code" format
3. should `Has Model Relationships` appear in the "treeview"
4. should `Is Model Relationships` appear in the "treeview"

albert-github:
Any comments?

sloriot:
I 100% agree to do the same thing, it is useless in my opinion to have such pages and even more in the treeview.

albert-github:
As a start I've just pushed a proposed patch, pull request #7576 for the `AABB_tree` so the results can be checked on "problems" before implementing it with the other packages.

PR #7610, opened 2023-07-20
Fixes for vertex/cell classes of SMDS_3 and Tetrahedral_remeshing

MaelRL:
## Summary of Changes

The main point of this PR is to restore the API of `Remeshing_cell_base_3` to the initial state (since CGAL 5.1), which was changed with the introduction of SMDS_3 in CGAL 5.6 (https://github.com/CGAL/cgal/pull/5693).

In CGAL 5.5, the class `Remeshing_cell_base_3` is templated by a geom traits and a cell base.
In CGAL 5.6, the class became templated by only two index types, removing the possibility to derive from custom (TDS) base cell types.

Also:
- Decompactify Simplicial_cell_base_3, allowing a cell base to be used
- Changed the API of Remeshing_vertex_base_3 to match that of Remeshing_cell_base_3 (and other CGAL triangulations): a Gt and a Vb
- Proper rebind mechanism for VB/CB of SMDS_3 / Tetrahedral remeshing
- Various fixes of the SMDS_3 / Tetrahedral_remeshing doc.

TODO:
- [x] Announce the breaking changes in CHANGES.md
- [x] Introduce Compact_simplicial_cell_base_3?

## Release Management

* Affected package(s): `SMDS_3`, `Tetrahedral_remeshing`
* Issue(s) solved (if any): -


SMDS_3/include/CGAL/Simplicial_mesh_cell_base_3.h
```diff
@@ -32,187 +32,48 @@
...
-    return V[i];
   }
 
-  bool has_vertex(Vertex_handle v) const
-  {
-    return (V[0] == v) || (V[1] == v) || (V[2]== v) || (V[3]== v);
-  }
+  Simplicial_mesh_cell_3(Vertex_handle v0, Vertex_handle v1,
+                         Vertex_handle v2, Vertex_handle v3)
+    : Cb(v0, v1, v2, v3)
```
MaelRL, line 68: Should these other constructors also initialize the timestamp?

SMDS_3/include/CGAL/Simplicial_mesh_cell_base_3.h
```diff
@@ -32,187 +32,48 @@
...
-    , V(rhs.V)
-    , time_stamp_(rhs.time_stamp_)
-    , subdomain_index_(rhs.subdomain_index_)
+    : Cb(static_cast<const Cb&>(rhs)),
+      time_stamp_(rhs.time_stamp_),
+      subdomain_index_(rhs.subdomain_index_)
   {
-    for(int i=0; i <4; i++){
+    for(int i=0; i <4; ++i)
       surface_index_table_[i] = rhs.surface_index_table_[i];
```
albert-github, line 63: Even though it is just one assignment wouldn't it be better to use `{` / `}` as well?

albert-github: Note:
I see that the declarations for `surface_index_table_`  are all like:
```
std::array<Surface_patch_index, 4> surface_index_table_ = {};
```
and
```
Surface_patch_index surface_index_table_[4];
```
also the construct
```
CGAL_precondition(facet>=0 && facet<4);
```
is used in this context.

Wouldn't it make sense to have a `#define` (or alike) that defines this value of `4`?

MaelRL: Thanks for the remarks.

>  Even though it is just one assignment wouldn't it be better to use { / } as well?
This code is old, and such updates would have to be mirrored into other packages. I don't want to extend the scope of this pull request.

> Wouldn't it make sense to have a #define (or alike) that defines this value of 4?
No, it's a 3D tetrahedral cell, there is no context in which it'll ever be something else than 4. I think it would be needlessly verbose to write:
```cpp
std::array<Surface_patch_index, NUMBER_OF_FACES_IN_A_TETRAHEDRON> surface_index_table_ = {};
```

albert-github: > > Even though it is just one assignment wouldn't it be better to use `{` / `}` as well?

> This code is old, and such updates would have to be mirrored into other packages. I don't want to extend the scope of this pull request.

It is a decisions that has to be taken on a higher level as I fully understand your reasoning. When doing it one place it should be consistent on all places (this is all a question of coding standards).

Indeed a tetrahedron has 4 faces and this will not change and the `NUMBER_OF_FACES_IN_A_TETRAHEDRON` would indeed be a bit long, From the name `surface_index_table_` it is not clear that this is a list of faces of a tetrahedron (might be my lack of knowledge of the CGAL package).

sloriot:
Tests seems to be failing in CGAL-6.0-Ic-37, for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-37/SMDS_3_Examples/TestReport_lrineau_Ubuntu-Latest.gz)

SMDS_3/include/CGAL/Simplicial_mesh_cell_base_3.h
janetournois: is it initialized in the `Cb` constructor?

MaelRL: no, the timestamp is a member of this class (Simplicial_mesh_cell_3).

MaelRL: @janetournois noticed that the initialization of the cell's timestamp to `-1` is actually performed on every new allocation of compact container new block (`initialize_time_stamp()`), so the initialization to `-1` in the cell bases is never required. Ideally it sould be cleaned from all the cell bases that do it (some in Mesh_3, Alpha_wrap_3, etc.) and maybe some comments in the code.

Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/Remeshing_cell_base_3.h
```diff
@@ -15,36 +15,63 @@
...
+It must be a model of the `SimplicialMeshCellBase_3` concept.
 
-template<typename Subdomain_index,
-         typename Surface_patch_index,
-         typename TDS>
-class Remeshing_cell_3
-  : public CGAL::Simplicial_mesh_cell_3<Subdomain_index,
-                                        Surface_patch_index,
-                                        TDS>
+\cgalModels `RemeshingCellBase_3`
```
albert-github, line 38: Due to #7395 this has changed and has to be:
```
\cgalModels{RemeshingCellBase_3}
```

albert-github (changes requested):

PR #7611, opened 2023-07-20
Triangulation_3: the testsuite of the simplex traverser needs C++17

lrineau:
## Summary of Changes

The files `test_simplex_iterator_3.cpp` and `test_segment_simplex_traverser_3.cpp` require C++17. This pull-request fixes the CMake logic so that they are only compiled if the compiler is able to use C++17.

Fix Triangulation_3 test suite on a few compilers:
https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-5.6-I-277/Triangulation_3/TestReport_Christo_MSVC2017-Debug-64bits.gz

## Release Management

Not a blocker in my option.

* License and copyright ownership: n/a


lrineau:
Tested in https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-27.shtml

PR #7613, opened 2023-07-21
Add an example on combining 3D Alpha Wrap and Tetrahedral Remeshing

MaelRL:
## Summary of Changes

Alpha wrap is a surface extracted from an ambient 3D triangulation.

Its volume mesh can also be useful, but since we don't care about it during carving, its quality is poor. However, we can use tetrahedral remeshing to remesh it. This PR adds an example on how to do it.

The example is undocumented because it is using undocumented features of Alpha_wrap_3.

image
input (left), wrap (middle), remeshed volumetric mesh (right)

### TODO
- [ ] It isn't clear in the doc of tetrahedral remeshing that the subdomain index 0 is not going to be remeshed.
- [ ] Can tetrahedral remeshing truly respect the surfaces given in input? (at least don't relocate new vertices?)
- [ ] Maybe we can avoid copying the mesh.

## Release Management

Depends on https://github.com/CGAL/cgal/pull/7610.

* Issue(s) solved (if any): -


janetournois:
Tetrahedral remeshing can exactly respect the surfaces given as input when using `CGAL::parameters::remesh_boundaries(false)`.
However `facet_is_constrained_map` is used only to add facets to the surfaces taken into account by `remesh_boundaries` (which is not really correct imo)

janetournois:
About subdomain 0 : the default of `cell_is_selected_map` is "a default property map where all cells of the domain (i.e. with a non-zero `Subdomain_index`) are selected.", so it is already mentioned.
Do you think we should mention it earlier in the function description? (if so, we must make sure it keeps compatible with `cell_is_selected_map`)

MaelRL:
>  Do you think we should mention it earlier in the function description? (if so, we must make sure it keeps compatible with cell_is_selected_map)

I think so. I guessed `0` would be considered outside and not remeshed, but I searched for zero and could not find it (blame whoever came up with collapsible named parameters documentation :>).

I would in fact put a line about it in the user manual. And I would also move the functions generating the input triangulations directly into the examples rather than hiding them away [*], and put a comment above the subdomain index setting:
```cpp
    /// A subodmain index 0 is considered outside and is not remeshed by default
    for (typename Tr::Cell_handle c : tr.finite_cell_handles())
      c->set_subdomain_index(1);
```

[*] the example code is small, and it's just a couple of lines.

What do you think?

janetournois:
> And I would also move the functions generating the input triangulations directly into the examples rather than hiding them away [*]

I'm pretty sure that's what I had done in the first place, so I agree :-)

I will make the modifications you suggest.

janetournois:
I could move
```cpp
  template<typename Tr>
  void generate_input_one_subdomain(const std::size_t nbv, Tr& tr)
  {...}
```

inside `tetrahedral_remeshing_example.cpp`, before the `main` definition.

@lrineau I think you were against it in the past, to make the example focused on remeshing. @MaelRL 's argument is to make clearer that subdomain 0 will not be remeshed.
What do you think?

lrineau:
> I could move
> 
> ```c++
>   template<typename Tr>
>   void generate_input_one_subdomain(const std::size_t nbv, Tr& tr)
>   {...}
> ```
> 
> (that inserts random points in a triangulation and sets all subdomain indices of finite cells to 1) inside `tetrahedral_remeshing_example.cpp`, before the `main` definition.
> 
> @lrineau I think you were against it in the past, to make the example focused on remeshing. @MaelRL 's argument is to make clearer that subdomain 0 will not be remeshed. What do you think?

I do not really remember.

sloriot:
@MaelRL please let us know if the TODOs are for later and you want to open an issue instead of blocking the merge. I'm mostly worried of the interaction with https://github.com/CGAL/cgal/pull/7610

MaelRL:
https://github.com/CGAL/cgal/pull/7610 should be merged first and this can be ignored until then.

sloriot:
Successfully tested in CGAL-6.0-Ic-64.

@MaelRL  there are some todos remaining. Are you planning to address them in this PR or shall you open a dedicated issue?

MaelRL:
I think @janetournois said she would take care of that one:

>  It isn't clear in the doc of tetrahedral remeshing that the subdomain index 0 is not going to be remeshed.

The other two, I leave for later.

janetournois:
If the tests are successful I suggest to merge this PR and open a new issue about doc improvements

PR #7628, opened 2023-08-01
cgal_create_release_with_cmake.cmake: only call Git one

lrineau:
## Summary of Changes

`cgal_create_release_with_cmake.cmake` was terribly slow (in particular on Windows). In order to fix that:

- call Git only once, to get the Git hash of the repository, and
- Use `$Id: ${f} ${GIT_HASH} $` to replace `$Id$` (all files have the same hash).

That is sufficient to identify without any doubt the version is the original file.

@sloriot: do you agree with those changes?

## Release Management

Cc: @janetournois


sloriot:
Successfully tested in [CGAL-6.0-Ic-46](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-46.shtml) (provided it is used by scripts)

PR #7629, opened 2023-08-02
AW3: improve manifoldness enforcing heuristics

### LINKED ISSUE #7625: Alpha Wrapping Strange Features

## Issue Details
I've apha-wrapped ~400 stl format meshes using a slightly modified version of the alpha wrapping example script and, on four out of the 400, strange triangles appeared connecting different components. I have a [link](https://mega.nz/folder/aVpF3BAZ#pID2KeyYR2ww1Uk2pi8U5w) to the original meshes in question and various alpha wrapping attempts. The meshes are all named according to this format: <originalName>_<relativeAlpha>_<relativeOffset>.off

The ones originally noticed with this issue have a relativeAlpha of 1000 and a relativeOffset of 4000 however (on the same mesh) offsets similar to 4000 can also exhibit the same behavior. 

This isn't pressing as I can just use the ones with a relativeOffset of 4050 however I would like to understand the issue, if possible, and make sure that this is known about if it is unintended.

## Source Code

This is the source code that I used which is a minimally changed version of the alphaWrappingDemo code

#include <CGAL/alpha_wrap_3.h>
#include <CGAL/Polygon_mesh_processing/bbox.h>
#include <CGAL/Polygon_mesh_processing/IO/polygon_mesh_io.h>
#include <CGAL/Real_timer.h>
#include <iostream>
#include <string>
namespace AW3 = CGAL::Alpha_wraps_3;
using Point_3 = K::Point_3;
using Mesh = CGAL::Surface_mesh<Point_3>;
{
  std::cout.precision(17);
  // Read the input
  const std::string filename = (argc > 1) ? argv[1] : CGAL::data_file_path("meshes/armadillo.off");
  std::cout << "Reading " << filename << "..." << std::endl;
  Mesh mesh;
  if(!PMP::IO::read_polygon_mesh(filename, mesh) || is_empty(mesh) || !is_triangle_mesh(mesh))
  {
    std::cerr << "Invalid input." << std::endl;
    return EXIT_FAILURE;
  }
  std::cout << "Input: " << num_vertices(mesh) << " vertices, " << num_faces(mesh) << " faces" << std::endl;
  // Compute the alpha and offset values
  const double relative_alpha = (argc > 2) ? std::stod(argv[2]) : 20.;
  const double relative_offset = (argc > 3) ? std::stod(argv[3]) : 600.;
  CGAL::Bbox_3 bbox = CGAL::Polygon_mesh_processing::bbox(mesh);
  const double diag_length = std::sqrt(CGAL::square(bbox.xmax() - bbox.xmin()) +
                                       CGAL::square(bbox.ymax() - bbox.ymin()) +
                                       CGAL::square(bbox.zmax() - bbox.zmin()));
  const double alpha = diag_length / relative_alpha;
  const double offset = diag_length / relative_offset;
  // Construct the wrap
  CGAL::Real_timer t;
  t.start();
  Mesh wrap;
  CGAL::alpha_wrap_3(mesh, alpha, offset, wrap);
  t.stop();
  std::cout << "Result: " << num_vertices(wrap) << " vertices, " << num_faces(wrap) << " faces" << std::endl;
  std::cout << "Took " << t.time() << " s." << std::endl;
  // Save the result
  std::string input_name = std::string(filename);
  input_name = input_name.substr(input_name.find_last_of("/") + 1, input_name.length() - 1);
  input_name = input_name.substr(0, input_name.find_last_of("."));
  std::string output_name = input_name
                            + "_" + std::to_string(static_cast<int>(relative_alpha))
                            + "_" + std::to_string(static_cast<int>(relative_offset)) + ".off";
  std::cout << "Writing to " << output_name << std::endl;
  CGAL::IO::write_polygon_mesh(output_name, wrap, CGAL::parameters::stream_precision(17));
  return EXIT_SUCCESS;
}

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): MacOS with M2
* Compiler: g++
* Specific flags used (if any): -DCMAKE_BUILD_TYPE=Release
* CGAL version: 5.5.2
* Boost version: 1.82.0_1
* Other libraries versions if used (Eigen, TBB, etc.): None that I know of

MaelRL:
## Summary of Changes

There is, at the end of the alpha wrap process, a post-processing step that re-adds material to enforce geometrical manifoldness.

This step is driven by a heuristics that prioritize cells around non-manifold vertices. The current heuristic prefers:
- cells that are not incident to artificial vertices (bbox and cave initilization points)
- cells that are incident to a lot of facets
- cells that are adding a small volume

Turns out the 2nd criterion is not so great: because it ignores volume, it can add very large tetrahedra. Furthermore, it can cascade into adding absurd amounts of tetrahedra. Its purpose was to reduce the amount of tetrahedra added, hoping that by adding few tetrahedra (regardless of their volume), it would end up adding less volume than adding small tetrahedra. In practice, it seems like it doesn't work that well, so just remove it and let the process be driven entirely by tetrahedron size.

Here is a comparison for a typical case reported in https://github.com/CGAL/cgal/issues/7625:

image

TODO:
- [x] do not resort at every step since we don't use the #boundary facet anymore
- [x] retest manually post [bc8351f](https://github.com/CGAL/cgal/pull/7629/commits/bc8351f15678b90955a9d5c2a5cc4b0389b49642)

## Release Management

* Issue(s) solved (if any): fix #7625 


Alpha_wrap_3/include/CGAL/Alpha_wrap_3/internal/Alpha_wrap_3.h
```diff
@@ -1398,6 +1405,9 @@ class Alpha_wrap_3
 //      return geom_traits().compute_squared_distance_3_object()(m_dt.point(c, 0), cc);
 //    };
 
+    // the reasonning behind using longest edge rather than volume is that we want to avoid
```
albert-github, line 1408: reasonning -> reasoning

lrineau:
Is this PR a change that could be backported to 5.5.x or 5.6.x?

MaelRL:
It's on the fence between bug fix and changing the behavior. I picked the latter for the (low) probability that it would be an unwanted change for some users.

lrineau:
> It's on the fence between bug fix and changing the behavior. I picked the latter for the (low) probability that it would be an unwanted change for some users.

Then, maybe rebase the branch onto a release branch, but let the PR be based on `master`. That will ease a possible merge of the changes by a user (by a merge of your branch).

Alpha_wrap_3/include/CGAL/Alpha_wrap_3/internal/Alpha_wrap_3.h
```diff
@@ -1402,6 +1409,9 @@ class Alpha_wrap_3
 //      return geom_traits().compute_squared_distance_3_object()(m_dt.point(c, 0), cc);
 //    };
 
+    // the reasoning behind using longest edge rather than volume is that we want to avoid
+    // spikes (which would have a small volume), and can often happen since we do not spend
+    // any care on the quality of tetrahedra.
     auto sq_longest_edge = [&](Cell_handle c) -> FT
```
sloriot, line 1415: This fails in CGAL-6.0-Ic-37

MaelRL:
Merged into #7805.

PR #7633, opened 2023-08-02
reproject barycenter to avoid boundary shrinking

### LINKED ISSUE #7631: Border smoothing in isotropic remeshing
## Issue Details

The combination of NP ```relax_constraints(true)``` and a larger number of tangential smoothing iterations can cause unexpected results along the mesh border:

vs the same case but with ```relax_constraints(false)```

To recreate the results from above, I used Polyhedron demo, nefertiti.off mesh and the following setup:

Screenshot 2023-08-02 at 11 12 22

No feature edges were used.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Mac
* Compiler: Apple Clang
* CGAL version: master (2023-08-02)
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.): QT5

sloriot:
Fixes #7631

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/tangential_relaxation.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/tangential_relaxation.h
@@ -218,6 +218,7 @@ void tangential_relaxation(const VertexRange& vertices,
     typedef std::tuple<vertex_descriptor, Vector_3, Point_3> VNP;
     std::vector< VNP > barycenters;
     auto gt_barycenter = gt.construct_barycenter_3_object();
+    auto gt_project = gt.construct_projected_point_3_object();
 
     // at each vertex, compute vertex normal
     std::unordered_map<vertex_descriptor, Vector_3> vnormals;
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/tangential_relaxation.h
```diff
@@ -269,8 +270,19 @@ void tangential_relaxation(const VertexRange& vertices,
...
-              gt_barycenter(get(vpm, ph0), 0.25, get(vpm, ph1), 0.25, get(vpm, v), 0.5));
+          {
+            typename GT::Point_3 bary = gt_barycenter(get(vpm, ph0), 0.25, get(vpm, ph1), 0.25, get(vpm, v), 0.5);
+            // to avoid shrinking of borders, we project back onto the incident segments
+            typename GT::Segment_3 s1(get(vpm, ph0), get(vpm,v)),
+                                   s2(get(vpm, ph1), get(vpm,v));
+
+            typename GT::Point_3 p1 = gt_project(s1, bary), p2 = gt_project(s2, bary);
+
+            bary = squared_distance(p1, bary)<squared_distance(p2,bary)? p1:p2;
```
MaelRL, line 281: `compare_squared_distance(p1, p2, bary) == CGAL::SMALLER`?

sloriot: I agree but it is more expensive and I don't think it really matters here.

sloriot:

PR #7650, opened 2023-08-16
Be more permissive in Compare_dihedral_angle_3

sloriot:
Do not throw an error but return SMALLER
The collinearity check being already done in the predicate it would be more expensive to do it also before calling the predicate

### DIFF
--- a/Kernel_23/include/CGAL/Kernel/function_objects.h
+++ b/Kernel_23/include/CGAL/Kernel/function_objects.h
@@ -246,6 +246,8 @@ namespace CommonKernelFunctors {
       const Vector_3 abad1 = xproduct(ab1, ad1);
       const FT sc_prod_1 = abac1 * abad1;
 
+      if (abac1==NULL_VECTOR || abad1==NULL_VECTOR) return SMALLER;
+
       CGAL_kernel_assertion_msg( abac1 != NULL_VECTOR,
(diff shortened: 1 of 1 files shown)


Kernel_23/include/CGAL/Kernel/function_objects.h
```diff
@@ -246,6 +246,8 @@ namespace CommonKernelFunctors {
       const Vector_3 abad1 = xproduct(ab1, ad1);
       const FT sc_prod_1 = abac1 * abad1;
 
+      if (abac1==NULL_VECTOR || abad1==NULL_VECTOR) return SMALLER;
```
afabri, line 249: Why don't you do this test before the multiplication?

sloriot: more expensive

afabri, line 249:
```suggestion
      if (abac1==NULL_VECTOR || abad1==NULL_VECTOR) return SMALLER;
      const FT sc_prod_1 = abac1 * abad1;
```

sloriot:
Successfully tested in [CGAL-6.0-Ic-46](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-46.shtml)

PR #7651, opened 2023-08-17
Fix collapse test in snap

sloriot:
don't do collinear test for a triangle to be collapsed


sloriot:
Successfully tested in [CGAL-6.0-Ic-46](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-46.shtml)

PR #7660, opened 2023-08-22
Arrangement: small doc fixes

afabri:
## Summary of Changes

Throw in `%` to avoid links.

## Release Management

* Affected package(s): Arrangements


albert-github:
I think good corrections , though I think there are more places in the CGAL software where these problems occur (probably also the word `Cartesian` a bit earlier in the same file.
I think a `grep` / `grep -v` might be useful here
- check Cartesian (and also cartesian)
- check orientation()
- check Homogeneous (and also homogeneous)

afabri:
No need for the lower case versions.

afabri:
So far I did only fixes in  `Pkg/doc/Pkg`

afabri:
> Regarding lower case conversions I wrote this as a reminder to check whether or not the case of the words is correct or not.

ok.  In fact I changed several "Homogeneous" to "homogeneous".

sloriot:
Successfully tested in CGAL-6.0-Ic-56

PR #7672, opened 2023-08-28
Orthtree generalization

JacksonCampolattaro:
## Summary of Changes

Orthtree is no longer restricted to only subdividing a `Point_set_3`, the contents of the tree are now a customization point of its Traits type.

Orthtree nodes are now represented as IDs. All node data is now stored using a new property system which allows users to add custom data to nodes. The new property system is included in `Property_map`.   It initially  also replaced the property system used by `Surface_mesh` and `Point_set_3`, but as this introduced problems we postpone that. 

## Release Management

* Affected package(s): `Orthtree`
* Issue(s) solved (if any): N/A
* Feature/Small Feature (if any): [link](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Orthtree_Generalization#Summary_of_API_changes) **pre-approved on 24/03/06**
* Link to compiled documentation   [link](https://cgal.github.io/7672/v0/Orthtree/index.html)
* License and copyright ownership: INRIA

## TODO

- [x] Polish the Property system
  - [x] Give `Properties.h` a more descriptive name
  - [x] Fix issue with `emplace_group` on MSVC.
  - [x] <s>Fix iteration over deleted elements.</s> unsplit is not implemented so no issue for now.
  - [x] Delete old property system from `Surface_mesh`
  - ~~Expand documentation of `Property_array`, `Property_container`, etc.~~ These will be left undocumented for now, as we expect the API to change in the future.
  - ~~Consider preferring the use of `Property_array_handle` and eliminating direct access to `Property_array`; mixing the two is a potential source of confusion.~~ This doesn't need to be handled until the new property system is documented, but I'd like to come back to this if there's time.
- [x] Polish the generalized Orthtree
  - [x] Merge `Orthtree_traits_point_*` classes; only the typedefs are different, so the functionality should be de-duplicated.
    - [x] Use template machinery to choose the appropriate base class based on the dimension tag.
  - [x] Ensure shared corners of adjacent nodes have identical coordinates (Sven's solution should resolve this)
    - [x] This should have a unit test
    - Seems to be working after switching to Iso_cubioid. I might just need to find the right numerical edge case though.
  - [x] Remove `enlarge_ratio` feature
    - ~~Call `bbox.dilate(1)` instead; this behavior is more appropriate in almost all cases.~~ This does not seem to be necessary now that Iso_cuboid is used.
    - [x] <s>Add an example that extends a Traits class to provide the enlargement feature using a modified `root_node_bbox` functor.</s>
  - [x] Replace Array type in Traits.
    - ~~Remove integer offsets. `Simple_cartesian_d<int>::Point` may work as an alternative, but I'll need generic kernel conversion to integrate it.~~ An array type is still used internally, but never exposed to the user.
  - [x] Replace Bbox with Iso_cuboid
  - [x] Change the new Traits methods to use the functor object pattern. (Is there a page I can link to in my documentation that explains why this pattern is necessary?)
    - [x] Document functor objects ~~(may require some template magic, along with the macros for `unspecified_type`)~~ Functor objects only need to be documented in the concept itself, because the explanation will appear in model documentation too.
  - [x] Nearest neighbors functionality only makes sense for some traits, and only has examples which work with point sets. It should become an external function. This may also let us remove element access from the traits class.
  - [x] Comb through documentation to make sure everything matches the new interface.
    - [x] Update user manual text
    - [x] Some features have been moved to Traits, their explanations should be moved.
    - [x] Orthtree documentation should not specifically mention points or point sets, now that it has been generalized.
  - [x] Update list of models for orthtree traits
  - [x] Expand and clarify `root_node_contents_object()` documentation.
  - [x] Replace all remaining instances of `typedef` with `using` for consistency.
    - [x] Does `using Dimension = unspecified_type;` break doxygen? If not, Concepts should also use `using`.
  - [x] Run benchmarks to measure performance change vs original Orthtree implementation.
    - [x] Benchmark results should be mentioned in the history section.
    - [x] Benchmark plots must be re-created for the manual.
  - [x] <s>Implement `unsplit(n)`. This features isn't actually used anywhere now, but it could be useful in the future.</s>
  - [x] The new orthtree requires `boost::span`; the boost requirement will either need to be bumped to 1.78, or a span can be added to `Stl_extension`.
- [x] Adapt `Efficient_RANSAC` package to account for the new interface
- [x] Adapt `Polyhedron` package to account for non-nullable property handles
- [x] Double check license headers in all newly added files. Dates may need to be updated in some other files.
- [x] Fix bbox computation (doc box overlaps?)
- [x] Update changes.md
- [x] Check branch size(todo for @sloriot)


Orthtree/include/CGAL/Orthtree_traits_point_3.h
```diff
@@ -107,46 +108,55 @@ struct Orthtree_traits_point_3 : public Orthtree_traits_3_base<GeomTraits> {
...
+    };
   }
 
-  typename Self::Point_d get_element(const Node_data_element& index) const {
-    return get(m_point_map, index);
+  auto get_element_object() const {
+    return [&](const Node_data_element& index) -> typename Self::Point_d {
+      return get(m_point_map, index);
+    };
   }
```
JacksonCampolattaro, line 160: @sloriot I'm testing out switching over the new Traits functionality to the functor object approach used by `construct_bbox_d_object`. For now I've used lambda captures to avoid defining functors which need constructors, since they depend on State in the Traits class. 

This works fine as a temporary solution, but I will need to add more boilerplate in order to document the functor object itself. That should be pretty straightforward to add, but I would prefer not to go through with this change, because it makes defining a custom Traits class much more difficult for the user. Is there a document which explains why the functor+_object function idiom is necessary?

soesau:
There seems to be a bug in Surface Mesh due to the new properties.
When I delete an edge, the halfedges of that edge are not marked as deleted.

And it seems that deleted elements still appear in the iterators, i.e., halfedges_begin()/faces_begin()/...

JacksonCampolattaro:
> There seems to be a bug in Surface Mesh due to the new properties. When I delete an edge, the halfedges of that edge are not marked as deleted.
> 
> And it seems that deleted elements still appear in the iterators, i.e., halfedges_begin()/faces_begin()/...

Hmm, I can take a look this weekend, is there an existing unit test which catches this?

afabri:
[This](https://gist.github.com/afabri/3d1f85feee53e4b5973e50b2c9d43a7f) should be an example.  Only one point should be written as output.

JacksonCampolattaro:
> [This](https://gist.github.com/afabri/3d1f85feee53e4b5973e50b2c9d43a7f) should be an example. Only one point should be written as output.

JacksonCampolattaro:
The old version had a baroque system of memory management which ensured edges & faces were placed contiguously. This came at the cost of additional indirection when accessing the neighbors of an edge, because it needed to maintain a table of indices which was re-mapped when they were moved around in memory. The new property system provides its own memory management which keeps indices stable for theoretically better performance, at the cost of allowing gaps after deletion. The best short term solution for retaining the same behavior might be to revert the Surface_mesh package to do memory management the old way, but still use the new property system. The built-in memory management of the new property system adds minimal overhead, and when that feature is ignored the interface is similar to the old property system.

In the long term, I don't think this is a good solution. Eventually I think it would be worth doing one of the following:

- Add an iterator which skips empty indices in the property array. This loses random-access iterator semantics, but maintains performance for most important tasks.
- Keep the existing behavior, including deleted points in the iterator. This preserves random-access semantics (`begin() + edge_index` behaves as expected ), but makes forward iteration confusing if points have been deleted. I don't think this is a good solution.
- Condense the property arrays after each deletion, updating indices to account for the offsets. This could be the responsibility of the garbage collection function. This may be an expensive operation, but it's a cost that can be moved around. 
  - For example, we could condense the arrays before returning iterators, or we could add a precondition that `has_garbage() == false` before the user iterates over properties.
- Different users of the property container system don't have the exact same needs. We could add property containers which perform memory management in different ways while providing the same interface. `Property_container` would then be a concept, implemented by versions which do no memory management, skip-list memory management, contiguous memory management, etc. This was my original hope for the property container system, but for the purposes of this PR I only created the simplest version.

In any case, these probably belong in their own PR. Re-working the entire property system as part of the orthtree update was definitely ambitious. By reverting to the old behavior, we can keep the new underlying property system as an implementation detail, and defer these choices until later.

afabri:
Currently elements get deleted they are marked as deleted only.  The deleted elements are "linked"  to form a free-list, so that when adding elements instead of adding them at the end of the storage we first reuse slots from the free-list.   Is that what you call "baroque" (I do not feel offended, only want o understand what you try to replace).

JacksonCampolattaro:
> Currently elements get deleted they are marked as deleted only. The deleted elements are "linked" to form a free-list, so that when adding elements instead of adding them at the end of the storage we first reuse slots from the free-list. Is that what you call "baroque" (I do not feel offended, only want o understand what you try to replace).

Sorry, I didn't mean baroque in a negative way. I must have misunderstood the old system or gotten it mixed up with another package, because it looks like `Index_iterator` already does what I suggest in my first bullet point. In that case the change in behavior is actually a bug (I'm still using `Index_iterator`, but something isn't working as expected). The only functional difference between the two is then replacing the linked free-list with an `std::vector<bool>` active-list provided by the property container. I suspect the vector system is still faster, because while `std::find_first_of` is technically O(n), it should be extremely fast on the packed bit representation of `vector<bool>`, and it has nicer access patterns than traversing a linked list. I haven't confirmed any of this with benchmarking though, so the old system could definitely be faster on real data where N is extremely large and you only need to allocate 1 element at a time instead of searching for a large-enough block.

JacksonCampolattaro:
Switching back to the old implementation is probably the best course of action for now. The main purpose of the changes to `Surface_mesh` was to figure out the interface the new property container system should provide. Eventually there should be a property container implementation which does swap-with-last deletion for packages which don't require stable indices (like `Point_set_3`) and another which manages a free-list like `Surface_mesh`. That way more packages can have user-added properties in the future without the need to implement custom iterators and memory management. 

When I came to the conference we decided to create the new property system in part because of licensing issues with the existing one. We started with `vector<bool>` garbage management because it's simple to implement.

afabri:
Will the new orthtree work with the old property mechanism?   I would expect that.

JacksonCampolattaro:
I think it would work fine, aside from the licensing issues, which I think @sloriot knows more about. 

The new property system also adds some features useful for an efficient implementation of the Orthtree (like an `emplace_group` method) and some quality-of-life changes like returning `boost::optionals` where appropriate instead of property maps in an invalid "null" state. These differences could all be worked around if necessary, it would just take some time.

afabri:
@JacksonCampolattaro can you check that your new employer agrees that what you contribute to this PR from now on is under the copyright of Inria (@palliez).

JacksonCampolattaro:
I was a little bit concerned about TU Delft's ancillary activities requirements initially, so I've already spoken with my advisor. Because this work isn't paid and I'm doing it on my own time, I don't need to report it officially and there shouldn't be any licensing issues.

afabri:
I would  be glad to discuss the changes made in `Surface_mesh`, but it is annoying that there are also distracting changes of indentation.

JacksonCampolattaro:
Sorry about that, I'm testing out a roll back to the previous memory management scheme, but I'm running into the same issue.

afabri:
> Sorry about that, I'm testing out a roll back to the previous memory management scheme, but I'm running into the same issue.

Hello, I would be glad to join the today's orthtree meeting to understand things better and work on Surface_mesh myself.

JacksonCampolattaro:
That sounds like a good idea. We typically meet at 13:15, perhaps you can join the call in Element.

Orthtree/doc/Orthtree/Orthtree.txt
```diff
@@ -5,7 +5,7 @@ namespace CGAL {
 \anchor Chapter_Orthtree
 
 \cgalAutoToc
-\authors Jackson Campolattaro, Simon Giraudot, Cédric Portaneri, Tong Zhao, Pierre Alliez
+\authors Jackson Campolattaro, Simon Giraudot, Cedric Portaneri, Tong Zhao, Pierre Alliez
```
lrineau, line 8: I do not get that. Why remove the accent from Cédric's name?!

soesau: Ah, that was me. Visual Studio didn't showed the accent, but some weird characters. I'll fix it.

soesau:
\build:v0

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
```diff
@@ -0,0 +1,152 @@
...
+    };
+  }
+
+  auto construct_root_node_contents_object() {
+    return [&]() -> Node_data {
+      return {faces(m_pm).begin(), faces(m_pm).end()};
+    };
+  }
+
+  auto distribute_node_contents_object() {
```
afabri, line 95: This seems to be written for ``TriangleMesh` and not `PolygonMesh`.  What do we want?

STL_Extension/include/CGAL/span.h
```diff
@@ -0,0 +1,405 @@
...
+#include <array>
+#include <iterator>
+#include <type_traits>
+
+namespace boost {
+
+constexpr std::size_t dynamic_extent = static_cast<std::size_t>(-1);
+
+template<class T, std::size_t E = dynamic_extent>
+class span;
```
afabri, line 27: Why do we need this fork?

QP_solver/include/CGAL/QP_solver/basic.h
```diff
@@ -23,6 +23,7 @@
 #include <CGAL/Quotient.h>
 #include <CGAL/QP_solver/assertions.h>
 #include <CGAL/QP_solver/debug.h>
+#define _HAS_AUTO_PTR_ETC False // temporarility necessary for boost's compatibility with Clang 15
```
afabri, line 26: What is this about?

Shape_detection/include/CGAL/Shape_detection/Efficient_RANSAC/Efficient_RANSAC.h
```diff
@@ -1086,7 +1086,7 @@ class Efficient_RANSAC {
       Cell cell = stack.top();
       stack.pop();
 
-      FT width = octree->width() / (1 << (cell.depth()));
+      FT width = octree->width() / (1 << (octree->depth(cell)));
 
       FT diag = CGAL::sqrt(FT(3) * width * width) + epsilon;
```
afabri, line 1091: What is the `epsilon` about?  Must it be documented?

STL_Extension/include/CGAL/span.h
sloriot: std::span is c++20

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
```diff
@@ -18,45 +17,109 @@ class OrthtreeTraits
...
+   * Must be able to take on values ranging from 0 to the number of faces of the (hyper)rectangle type, equivalent to 2 * D.
+   */
+  using Adjacency = unspecified_type; ///< Specify the adjacency directions
 
   /*!
-    Functor with an operator to construct a `Bbox_d` from two `Array` objects (coordinates of minimum and maximum points).
-  */
-  typedef unspecified_type Construct_bbox_d;
+   * \brief Functor with an operator to create the bounding box of the root node.
+   *
```
afabri, line 45: Here the signatures of all the functors must be added.

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
```diff
@@ -18,45 +17,109 @@ class OrthtreeTraits
...
+   * It should take no arguments, and return an instance of `Node_data`.
+   *
+   * Typically, the `Node_data` of the root node contains all the elements in the tree.
+   * For a tree in which each node contains an `std::span()` this function would return the span containing all items.
+   *
+   */
+  using Construct_root_node_contents = unspecified_type;
+
+  /*!
+   * \brief Functor to locate in which halfspace a number of type `FT` is located with respect to another number of type `FT`.
```
afabri, line 67: Looking at the implementation in `Orthtree_traits_base_for_dimension`  this looks just like a `Less`  functor (which we do not have in `RealEmbeddableTraits`, but where we have a `Compare`).

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -43,258 +48,224 @@
...
 
   /*!
    * \brief A predicate that determines whether a node must be split when refining a tree.
    */
-  typedef std::function<bool(Node)> Split_predicate;
+  using Split_predicate = std::function<bool(Node_index, const Self&)>;
 
   /*!
-   * \brief A model of `ConstRange` whose value type is `Node`.
+   * \brief A model of `ConstRange` whose value type is `Node_index` and its iterator type is `ForwardIterator`.
```
afabri, line 140: Is it really only `ForwardIterator` ?

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
```diff
@@ -0,0 +1,152 @@
...
+#include <boost/graph/graph_traits.hpp>
+#include <CGAL/Dimension.h>
+#include <CGAL/Bbox_3.h>
+
+namespace CGAL {
+
+/*!
+  \ingroup PkgOrthtreeTraits
+
+  The class `Orthtree_traits_face_graph` can be used as a template parameter of
```
afabri, line 29: Where will faces end up in the tree?  In each node where the bbox intersects the face?  Is that evident, or must that be documented.

Orthtree/include/CGAL/Quadtree.h
```diff
@@ -15,24 +15,24 @@
...
+  \ingroup PkgOrthtreeRef
 
   \brief Alias that specializes the `Orthtree` class to a 2D quadtree.
 
   These two types are exactly equivalent:
   - `Quadtree<GeomTraits, PointRange, PointMap>`
-  - `Orthtree<Orthtree_traits_2<GeomTraits>, PointRange, PointMap>`.
+  - `Orthtree<Orthtree_traits_point<GeomTraits, PointRange, PointMap, Dimension_tag<2>>>`.
 
   \warning This is a not a real class but an alias, please refer to
```
afabri, line 31: I do not understand the warning.

Orthtree/include/CGAL/Orthtree_traits_base_for_dimension.h
```diff
@@ -0,0 +1,198 @@
...
+  /// \name Operations
+  /// @{
+  auto construct_point_d_object() const {
+    return [](auto... Args) -> Point_d {
+      std::initializer_list<FT> args_list{Args...};
+      return Point_d{static_cast<int>(args_list.size()), args_list.begin(), args_list.end()};
+    };
+  }
+
+  auto locate_halfspace_object() const {
```
afabri, line 102: Note that the `typedef` for the functor is missing.

QP_solver/include/CGAL/QP_solver/basic.h
afabri, line 26:
```suggestion
```

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
```diff
@@ -0,0 +1,166 @@
...
+          if (do_intersect(t, bbox))
+            child_data.push_back(f);
+        }
+      }
+      };
+  }
+
+  /// @}
+
+  /// Recommanded split predicate to pass to `Orthtree::refine()` function so
```
afabri, line 127:
```suggestion
  /// Recommended split predicate to pass to `Orthtree::refine()` function so
```

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
```diff
@@ -0,0 +1,166 @@
...
+  /// that the octree is refined until a node is either empty or has an extend
```
afabri, line 128:
```suggestion
  /// that the octree is refined until a node is either empty or has an extent
```

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -43,258 +48,222 @@
...
-    m_bbox_min = construct_point_d_from_array(bbox_min);
-    m_side_per_depth.push_back(bbox_max[0] - bbox_min[0]);
-    m_root.points() = {point_range.begin(), point_range.end()};
+    m_side_per_depth.push_back(size);
+    data(root()) = m_traits.construct_root_node_contents_object()();
   }
 
   /// @}
 
-  /// \cond SKIP_IN_MANUAL
```
albert-github, line 254: Looks like the corresponding `\endcond` is still present (line 269), either both should be present or both should be absent.
(See also: https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-6.0-Ic-158/logs_master/Orthtree.log)

(Note: around line 160 there looks like to be an empty `\cond` / `\endcond` block).

albert-github (changes requested):

afabri:
The last commit was done to fix a bug revealed by the shape detection testsuite.  When an orthtree is constructed from a range of points,  followed by a call to `Orthtree::locate(Point p)`  for these points, this may assert that point `p` is not in the bbox of the root.   This is due to the fact that the bbox of a node is a construction. Internally one stores `Epeck::FT`   for the  size of the nodes. This needs a deeper look.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -468,14 +465,16 @@ class Orthtree {
...
+    std::size_t node_depth = depth(n);
+    Bbox_dimensions size = m_side_per_depth[node_depth];
+    const std::size_t last_coord = std::pow(2,node_depth)-1;
     for (int i = 0; i < Dimension::value; i++) {
-      min_corner[i] = approx(m_bbox_min[i] + int(global_coordinates(n)[i]) * size[i]).inf();
-      max_corner[i] = approx(m_bbox_min[i] + int(global_coordinates(n)[i] + 1) * size[i]).sup();
+      min_corner[i] = (m_bbox.min)()[i] + int(global_coordinates(n)[i]) * size[i];
+      max_corner[i] = std::size_t(global_coordinates(n)[i]) == last_coord
+                    ? (m_bbox.max)()[i]
+                    : (m_bbox.min)()[i] + int(global_coordinates(n)[i] + 1) * size[i];
```
soesau, line 477: I don't think that this will guarantee that neighboring nodes of different depth in the orthtree will have an exact shared coplanar side.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -459,77 +395,162 @@ class Orthtree
...
+    \param n node to generate a bounding box for
+
+    \return the bounding box of the node n
+   */
+  Bbox bbox(Node_index n) const {
+    using Cartesian_coordinate = std::array<FT, Dimension::value>;
+    Cartesian_coordinate min_corner, max_corner;
+    std::size_t node_depth = depth(n);
+    Bbox_dimensions size = m_side_per_depth[node_depth];
+    const std::size_t last_coord = std::pow(2,node_depth)-1;
```
afabri, line 472:
```suggestion
    const std::size_t last_coord = ( 1 << node_depth)-1;
```
not only that `pow()` triggers a conversion warning, it is slower than shifting.

Orthtree/doc/Orthtree/PackageDescription.txt
```diff
@@ -40,16 +42,17 @@
...
 - `CGAL::Octree<GeomTraits, PointRange, PointMap>`
-- `CGAL::Orthtree<Traits, PointRange, PointMap>`
+- `CGAL::Orthtree<GeomTraits, PointRange, PointMap>`
 
 \cgalCRPSection{Traits}
-- `CGAL::Orthtree_traits_2<GeomTraits>`
-- `CGAL::Orthtree_traits_3<GeomTraits>`
-- `CGAL::Orthtree_traits_d<GeomTraits, Dimension>`
+- `CGAL::Orthtree_traits_point<GeomTraits>`
+- `CGAL::Orthtree_traits_base_for_dimension<GeomTraits>`
```
afabri, line 54:
```suggestion
- `CGAL::Orthtree_traits_base<GeomTraits>`
```

Orthtree/include/CGAL/Orthtree_traits_base.h
```diff
@@ -0,0 +1,167 @@
...
+  The base class is extended by `CGAL::Orthtree_traits_point<GeomTraits, PointRange, PointMap, dimension>` and by `CGAL::Orthtree_traits_face_graph<PolygonMesh, VertexPointMap>`.
+
+  \tparam K a model of `Kernel`.
+  \tparam dim dimension of the ambient Euclidean space.
+
+  \sa `CGAL::Orthtree_traits_point<GeomTraits, PointRange, PointMap, dim>`
+  \sa `CGAL::Orthtree_traits_face_graph<PolygonMesh, VertexPointMap>`
+*/
+
+template <typename K, int dim>
```
afabri, line 37: Shall we rename `K`  to `GeomTraits` ?

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -43,338 +46,258 @@
...
-  \tparam Traits_ must be a model of `OrthtreeTraits`
-  \tparam PointRange_ must be a model of range whose value type is the key type of `PointMap_`
-  \tparam PointMap_ must be a model of `ReadablePropertyMap` whose value type is `Traits_::Point_d`
+  \tparam GeomTraits must be a model of `OrthtreeTraits`
  */
-template<typename Traits_, typename PointRange_,
-         typename PointMap_ = Identity_property_map<typename Traits_::Point_d> >
-class Orthtree
-{
+template <typename GeomTraits>
```
afabri, line 66: Did we want to rename `GeomTraits` to `OrthtreeTraits`, similar to how it is done for [`AABB_tree`](https://doc.cgal.org/latest/AABB_tree/classCGAL_1_1AABB__tree.html)?

soesau: Sounds good to me, GeomTraits is very generic.

STL_Extension/include/CGAL/Pair_optional_adaptor.h
```diff
@@ -0,0 +1,60 @@
...
+// Author(s)     : Sven Oesau
+
+#ifndef CGAL_PAIR_OPTIONAL_ADAPTOR
+#define CGAL_PAIR_OPTIONAL_ADAPTOR
+
+namespace CGAL {
+
+// T is supposed to be a handle
+template<typename T>
+class Pair_optional_adaptor : public std::optional<T> {
```
sloriot, line 20: I'm not sure you need to inherit from optional. I think I'd simply go for storing first and second and have conversion operators. Remember that both first and second will be lightweigth to you should not bother with references, union, ...

soesau: And I don't now how to really do without references. The type of first is T, if the optional I get is nullopt, I still need to provide an instance of T. It can be invalid, though. Thus, I would need to require T to be default constructible or use a union to avoid that.

sloriot: It is already default constructible since that's how it is currently implemented using a pair.

soesau: It is fully working now. I can still switch to simply storing first and second if you prefer or think it looks cleaner.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -459,77 +387,174 @@ class Orthtree
...
+
+    \tparam T the type of the property to add
+
+    \param name the name of the new property
+    \param default_value the default value assigned to nodes for this property
+
+    \return pair of the property map and a Boolean which is `true` if the property needed to be created
+   */
+  template <typename T>
+  Pair_optional_adaptor<Property_map<T>>
```
sloriot, line 514: here you should use plain `std::optional`

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -459,77 +387,174 @@ class Orthtree
...
+    \brief gets a property of the nodes if it exists.
+
+    \tparam T the type of the property to retrieve
+
+    \param name the name of the property
+
+    \return an optional containing the property map if it exists
+   */
+  template <typename T>
+  Pair_optional_adaptor<Property_map<T>>
```
sloriot, line 530: same here

STL_Extension/include/CGAL/Pair_optional_adaptor.h
```diff
@@ -0,0 +1,49 @@
...
+  Pair_optional_adaptor(std::optional<T>& obj) : std::optional<T>(obj), second(obj.has_value()), first(t_storage.t) {
+    if (obj.has_value())
+      t_storage.t = *obj;
+  }
+
+  Pair_optional_adaptor(const std::nullopt_t& obj) : std::optional<T>(std::nullopt), second(false), first(t_storage.t) {}
+
+  Pair_optional_adaptor(std::pair<T, bool>& p) : std::optional<T>(p.second ? p.first : std::optional<T>()), first(t_storage.t), second(p.second), t_storage(p.first) {}
+
+  operator std::pair<T, bool>() {
```
sloriot, line 31: deprecation warning is missing

STL_Extension/include/CGAL/Pair_optional_adaptor.h
```diff
@@ -0,0 +1,49 @@
...
+    T t;
+    int i;
+    T_value() : i(0) {}
+    T_value(T t) : t(t) {}
+  } t_storage;
+};
+
+} // CGAL
+
+#endif
```
sloriot, line 49: missing empty line at the end of the file

sloriot (commented):
I thought we said that Surface_mesh usage of std::pair instead of optional should be deprecated but I don't see any change in Surface_mesh (should also include a line in CHANGES.md + a plain std::optional in case CGAL_NO_DEPRECATED_CODE is defined)

soesau:
We decided after the meeting to do this in two separate PRs, the Orthtree and the changes to Surface_mesh.
But you are right, the adaptor should only be used in the Surface_mesh and not in the Orthtree.

sloriot:
> We decided after the meeting to do this in two separate PRs, the Orthtree and the changes to Surface_mesh. But you are right, the adaptor should only be used in the Surface_mesh and not in the Orthtree.

Then the adaptor should also be in the other PR?

sloriot:

Orthtree/include/CGAL/Orthtree.h
soesau: Yes, based on a lambda "next" function. See Orthtree::traverse() for details.

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
soesau: For now the implementation is restricted to triangles. As there is no use case so far, extending to polygons does not seem useful now.

STL_Extension/include/CGAL/span.h
soesau: I cannot see where this is used in the Orthtree.

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
soesau: removed

STL_Extension/include/CGAL/span.h
sloriot: range of points set AFAIR

Orthtree/include/CGAL/Orthtree.h
afabri: What about  https://doc.cgal.org/latest/Manual/classForwardRange.html

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
soesau: Doc is now more specific.

Orthtree/include/CGAL/Orthtree_traits_base_for_dimension.h
soesau: removed

STL_Extension/include/CGAL/span.h
soesau: Node_data in Orthtree_traits_point.h is `boost::iterator_range<typename PointRange::iterator>`.
span.h was included in Orthtree, but can be removed and everything still compiles.

The word span is used a few times in the doc in OrthtreeTraitsWithData. Is there any difference between a span and a random access range?

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
soesau: done

Orthtree/include/CGAL/Orthtree.h
soesau: done

Orthtree/doc/Orthtree/Concepts/CollectionPartitioningOrthtreeTraits.h
```diff
@@ -0,0 +1,62 @@
...
+
+  /*!
+   * \brief The data type contained by each node; must be a model of `ForwardRange`.
+   */
+  using Node_data = unspecified_type;
+
+  /*!
+   * \brief An element of the `Node_data` list-like type.
+   *
+   * Must be constructible from the value type of a `Node_data` iterator.
```
sloriot, line 38: so it's simply Node_data, no?

Orthtree/doc/Orthtree/Concepts/OrthtreeTraitsWithData.h
```diff
@@ -0,0 +1,76 @@
...
+  /*!
+   * \brief Functor which fills the contents of the nodes children.
+   *
+   * Provides the operator:
+   * `void operator()(typename Tree::Node_index, Tree&, const Point_d&)`
+   *
+   * The functor is called during refinement of the `Orthtree` on a node after it has been split. The purpose of the functor is to
+   * distribute the `Node_data`, accessible via `tree.data()`, to the data of the nodes children, accessible via `tree.children()`.
+   * The first parameter is the `Node_index` of the node. The second parameter provides the instance of the `Orthtree`
+   * and the last parameter is the barycenter of the node which delimits the internal boundaries of the children.
```
sloriot, line 52:
```suggestion
   * and the last parameter is the barycenter of the node which will be used as shared corner amongst the children of the node.
```

Orthtree/doc/Orthtree/Orthtree.txt
```diff
@@ -55,29 +57,27 @@ the existing ones do not match users' needs.
 
 \subsection Section_Orthtree_Quadtree Building a Quadtree
 
-The `Orthtree` class may be templated with `Orthtree_traits_2` and thus
-behave as a %quadtree. For convenience, the alias `Quadtree` is provided.
+The `Orthtree` class may be templated with `Orthtree_traits_point<>`
+while specifying a 2d ambient space and thus behave as a %quadtree.
```
sloriot, line 61: why did you remove the s at behave (actually you remove it line 75 but the question is valid here too)?

Orthtree/include/CGAL/Orthtree_traits_point.h
```diff
@@ -0,0 +1,181 @@
...
+  \cgalModels{CollectionPartitioningOrthtreeTraits}
+  \sa `CGAL::Octree`
+  \sa `CGAL::Quadtree`
+  \sa `CGAL::Orthtree_traits_base<GeomTraits, dimension>`
+*/
+template <
+  typename GeomTraits,
+  typename PointRange,
+  typename PointMap = Identity_property_map<typename std::iterator_traits<typename PointRange::iterator>::value_type>,
+  bool cubic = false,
```
sloriot, line 81: cubic -> hypercubic_nodes as we are in dD?

Orthtree/include/CGAL/Orthtree_traits_point.h
```diff
@@ -0,0 +1,181 @@
...
+        const typename Self::Point_d& point = get(m_point_map, p);
+        std::size_t i = 0;
+        for (const typename Self::FT& x: cartesian_range(point)) {
+          bbox_min[i] = (std::min)(x, bbox_min[i]);
+          bbox_max[i] = (std::max)(x, bbox_max[i]);
+          ++i;
+        }
+      }
+
+      if constexpr (cubic) {
```
sloriot, line 133: Nice!

Orthtree/doc/Orthtree/Concepts/CollectionPartitioningOrthtreeTraits.h
soesau: No, Node_data is a range and Node_data_element is boost::range_value<Node_data>.

sloriot: so maybe directly write Node_data::iterator?

Orthtree/doc/Orthtree/Orthtree.txt
soesau: This is from Jackson and I think behave is the infinitive here and not the 3rd person.

Orthtree/doc/Orthtree/Concepts/CollectionPartitioningOrthtreeTraits.h
```diff
@@ -0,0 +1,101 @@
...
+
+  /// @}
+
+  /// \name Operations
+  /// @{
+
+  /*!
+   * constructs an object of type `ConstructSphere_3`.
+   */
+  Construct_sphere_3 get_construct_sphere_3_object() const;
```
sloriot, line 83: no get_

afabri:
ICE with [VC2017](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-178/Orthtree/TestReport_Christo_MSVC2017-Release-64bits.gz)

Khalaq04:
Hi, I am interested in contributing to this project and would like to work on this issue. Can you please guide me on how to and from where to start working on this project and this issue? Thank you.

sloriot (commented):
Maybe it does not make sense now to have the free function since we have a way to disable the member function in case it cannot be called.

afabri:
> Maybe it does not make sense now to have the free function since we have a way to disable the member function in case it cannot be called.

When I suggested to re-add the member function I only had in mind that users have to change less code.  That's why I would not document the member function, but only the free function.

sloriot:
@soesau with dffac51 I could restore the former behavior but we are loosing the possibility to do Octree({points, map}) as it becomes ambiguous. What do you think?

Orthtree/include/CGAL/Orthtree_traits_face_graph.h
```diff
@@ -0,0 +1,169 @@
...
+namespace CGAL {
+
+/*!
+\ingroup PkgOrthtreeTraits
+
+Traits class for the `Orthtree` class to be used to construct a 3D octree around
+a triangulated surface mesh. Each node of the octree will store all the faces of the
+mesh intersected by its bounding box. The subdivision of the octree is controlled
+by the nested class `Orthtree_traits_face_graph::Split_predicate_node_min_extent`
+to which the minimal extend of a node should be provided.
```
afabri, line 33:
```suggestion
to which the minimal extent of a node should be provided.
```

Orthtree/include/CGAL/Orthtree_traits_base.h
```diff
@@ -0,0 +1,167 @@
...
+  using Point_d = typename GeomTraits::Point_d;
+  using Bbox_d = typename GeomTraits::Iso_box_d;
+  using Sphere_d = typename GeomTraits::Sphere_d;
+  using Cartesian_const_iterator_d = typename GeomTraits::Cartesian_const_iterator_d;
+  /*!
+   * Adjacency type.
+   *
+   * \note This type is used to identify adjacency directions with
+   * easily understandable keywords (left, right, up, down, ...) and is thus
+   * mainly useful in 2d and 3d. In
```
afabri, line 54:
```suggestion
   * mainly useful in 2D and 3D. In
```

Orthtree/include/CGAL/Orthtree_traits_base.h
```diff
@@ -0,0 +1,167 @@
...
+   * higher dimensions, such keywords do not exist and this type is
+   * simply an integer. Conversions from this integer to bitsets still
+   * work but do not provide any user-friendly API for adjacency selection.
+   *
+   * Two directions along each axis in %Cartesian space, relative to a node.
+   *
+   * Directions are mapped to numbers as 3-bit integers in the 3d case or as 2-bit integers in the 2d case.
```
afabri, line 61:
```suggestion
   * Directions are mapped to numbers as 3-bit integers in the 3D case or as 2-bit integers in the 2D case.
```

Orthtree/include/CGAL/Orthtree_traits_base.h
```diff
@@ -0,0 +1,167 @@
...
+   * In the 3d case the numbers 6 and 7 are not used because there are only 6 different directions.
+   *
+   * The first two bits indicate the axis (00 = x, 01 = y, 10 = z),
+   * the third bit indicates the direction along that axis (0 = -, 1 = +).
+   *
+   * The following diagram and table showing the 3d case may be a useful reference (2d case is identical with one dimension less):
```
afabri, line 67:
```suggestion
   * The following diagram and table showing the 3D case may be a useful reference (2D case is identical with one dimension less):
```

Orthtree/doc/Orthtree/PackageDescription.txt
```diff
@@ -1,20 +1,22 @@
...
-/// \ingroup PkgOrthtreeClasses
+/// \ingroup PkgOrthtreeRef
 
 /// \defgroup PkgOrthtreeTraversal Traversal
-/// \ingroup PkgOrthtreeClasses
+/// \ingroup PkgOrthtreeRef
 
-/// \defgroup PkgOrthtreeConcepts Concepts
+/// \defgroup PkgOrthtreeNeighbors Neighbor Search Functions
 /// \ingroup PkgOrthtreeRef
```
afabri, line 17: This group is now empty, so we can remove it, right?
```suggestion
```

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -39,342 +41,332 @@
...
   contain \f$2^{dim}\f$ other nodes which further subdivide the
   region.
 
   \sa `CGAL::Quadtree`
   \sa `CGAL::Octree`
 
-  \tparam Traits_ must be a model of `OrthtreeTraits`
-  \tparam PointRange_ must be a model of range whose value type is the key type of `PointMap_`
-  \tparam PointMap_ must be a model of `ReadablePropertyMap` whose value type is `Traits_::Point_d`
+  \tparam GeomTraits must be a model of `OrthtreeTraits` or `OrthtreeTraitswithData`.
```
afabri, line 114:
```suggestion
  \tparam GeomTraits must be a model of `OrthtreeTraits` or `OrthtreeTraitsWithData`.
```

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -39,342 +41,332 @@
...
-  typedef typename Traits::Dimension Dimension; ///< Dimension of the tree
-  typedef typename Traits::FT FT; ///< Number type.
-  typedef typename Traits::Point_d Point; ///< Point type.
-  typedef typename Traits::Bbox_d Bbox; ///< Bounding box type.
-  typedef typename Traits::Sphere_d Sphere; ///< Sphere type.
+#ifndef DOXYGEN_RUNNING
+  static inline constexpr bool has_data = Orthtree_impl::has_Node_data<GeomTraits>::value;
+  static inline constexpr bool supports_neighbor_search = true;// Orthtree_impl::has_Squared_distance_of_element<GeomTraits>::value;
+#else
+  static inline constexpr bool has_data = bool_value; ///< `true` if `GeomTraits` is a model of `OrthtreeTraitswithData` and `false` otherwise.
```
afabri, line 130:
```suggestion
  static inline constexpr bool has_data = bool_value; ///< `true` if `GeomTraits` is a model of `OrthtreeTraitsWithData` and `false` otherwise.
```

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -637,288 +785,629 @@ class Orthtree
...
+
+  /*!
+    \brief finds the next sibling in the parent of the node specified by the index `n`.
+
+    Traverses the tree in increasing order of local index (e.g., 000, 001, 010, etc.)
+
+    \param n the index of the node to find the sibling of
+
+    \return the index of the next sibling of n
+    if n is not the last node in its parent, otherwise nothing.
```
afabri, line 955:
```suggestion
    if n is not the last node in its parent, otherwise `std::nullopt`.
```

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -637,288 +785,629 @@ class Orthtree
...
+  Node_index descendant(Node_index node, Indices... indices) {
+    return recursive_descendant(node, indices...);
+  }
+
+  /*!
+    \brief convenience function for retrieving arbitrary nodes.
+
+    Equivalent to `tree.descendant(tree.root(), indices...)`.
+
+    \param indices the integer indices specifying the descent to perform, starting from root
```
afabri, line 938:
```suggestion
    \param indices the integer indices specifying the descent to perform, starting from the root
```

Orthtree/include/CGAL/Quadtree.h
```diff
@@ -15,34 +15,28 @@
...
-
-  \warning This is a not a real class but an alias, please refer to
-  the documentation of `Orthtree`.
+  \brief Alias that specializes the `Orthtree` class to a 2D quadtree storing 2D points.
 
   \tparam GeomTraits must be a model of `Kernel`
-  \tparam PointRange_ must be a model of range whose value type is the key type of `PointMap`
+  \tparam PointRange must be a model of `Range` whose value type is the key type of `PointMap`
   \tparam PointMap must be a model of `ReadablePropertyMap` whose value type is `GeomTraits::Point_2`
+  \tparam square_nodes boolean to enforce square nodes
```
afabri, line 30:
```suggestion
  \tparam square_nodes Boolean to enforce square nodes
```

Orthtree/include/CGAL/Octree.h
```diff
@@ -15,34 +15,27 @@
...
-  \warning This is a not a real class but an alias, please refer to
-  the documentation of `Orthtree`.
-
-  \tparam GeomTraits must be a model of `Kernel`
-  \tparam PointRange_ must be a model of range whose value type is the key type of `PointMap`
-  \tparam PointMap must be a model of `ReadablePropertyMap` whose value type is `GeomTraits::Point_3`
+  \tparam GeomTraits a model of `Kernel`
+  \tparam PointRange a model of `Range` whose value type is the key type of `PointMap`
+  \tparam PointMap a model of `ReadablePropertyMap` whose value type is `GeomTraits::Point_3`
+  \tparam cubic_nodes boolean to enforce cubic nodes
```
afabri, line 30:
```suggestion
  \tparam cubic_nodes Boolean to enforce cubic nodes
```

Orthtree/test/Orthtree/test_octree_bbox.cpp
```diff
@@ -1,17 +1,16 @@
 
 #include <iostream>
```
afabri, line 2: Please move  in all test files all  standard includes at the end, and move the includes of headers of the package to test to the front, so check that we do not miss any dependencies.

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -538,77 +629,131 @@ class Orthtree
...
+  \tparam OutputIterator a model of `OutputIterator` that accepts `GeomTraits::Node_data_element` objects.
+
+  \param query query sphere
+  \param output output iterator
+
+  \warning Nearest neighbor searches requires `GeomTraits` to be a model of `CollectionPartitioningOrthtreeTraits`.
+ */
   template<typename OutputIterator>
-  OutputIterator nearest_neighbors (const Sphere& query, OutputIterator output) const {
+  auto nearest_neighbors(const Sphere& query, OutputIterator output) const -> std::enable_if_t<supports_neighbor_search, OutputIterator> {
```
afabri, line 708: This seems not to be the right name for what the function does.

Orthtree/doc/Orthtree/Concepts/OrthtreeTraits.h
```diff
@@ -17,46 +18,62 @@ class OrthtreeTraits
...
-  typedef unspecified_type Dimension; ///< Dimension type (see `CGAL::Dimension_tag`).
-  typedef unspecified_type Bbox_d; ///< Bounding box type.
-  typedef unspecified_type FT; ///< The number type of the %Cartesian coordinates of types `Point_d`
-  typedef unspecified_type Point_d; ///< Point type.
-  typedef unspecified_type Sphere_d; ///< The sphere type for neighbor queries.
+  using Node_index = unspecified_type; ///< An integer type for nodes
+  constexpr int dimension; ///< Dimension.
+  using FT = unspecified_type; ///< The number type of the %Cartesian coordinates of types `Point_d`
+  using Point_d = unspecified_type; ///< Point type.
+  using Bbox_d = unspecified_type; ///< Bounding box type. Must be constructible from a pair of `Point_d` types.
```
afabri, line 25:
```suggestion
  using Bbox_d = unspecified_type; ///< Bounding box type. Must be constructible from a pair of `Point_d` objects.
```

Orthtree/doc/Orthtree/Concepts/OrthtreeTraitsWithData.h
```diff
@@ -0,0 +1,76 @@
...
+   * For a tree in which each node contains a span (such as `std::span()`) this function would return the span containing all items.
+   *
+   */
+  using Construct_root_node_contents = unspecified_type;
+
+  /*!
+   * \brief Functor which fills the contents of the nodes children.
+   *
+   * Provides the operator:
+   * `void operator()(typename Tree::Node_index, Tree&, const Point_d&)`
```
afabri, line 47: How can the type `Tree` be known here?

Orthtree/examples/Orthtree/orthtree_build.cpp
```diff
@@ -1,27 +1,27 @@
...
+using Traits = CGAL::Orthtree_traits_point<Kernel, Point_vector>;
+using Orthtree = CGAL::Orthtree<Traits>;
 
 int main()
 {
   CGAL::Random r;
 
   Point_vector points_dd;
-  for (std::size_t i = 0; i < 5; ++ i)
+  for (std::size_t i = 0; i < 500; ++ i)
```
afabri, line 22: You raised it from 5 to 500, but that are too many to see them written to `cout`.  Would 20  do it to illustrate something?

Orthtree/doc/Orthtree/Concepts/OrthtreeTraitsWithData.h
soesau: The type of the tree is Orthtree<Self> for the traits. The traits include the Orthtree.h

soesau: Tree::Node_index is just an alias for OrthtreeTraits::Node_index

I am not sure how to write the type of the tree correctly. How to reference the traits themselves?
void operator()(Node_index, Orthtree<Self>&, const Point_d&)

afabri: So it is then just `Node_index`.    But what about `Tree`?

afabri: I think now I understand the problem. We do not speak about a member function of the traits class, but about the operator of a functor.

soesau: Sorry, yes. A functor the traits have to provide that is called by the orthtree. As this functor needs to access nodes and their associated data, it needs an instance of the orthtree to operate on.

sloriot: It refines `OrthtreeTraits` so `Node_index` is defined. For the concept description you can use `Orthtree<Traits,b>`

Orthtree/doc/Orthtree/Orthtree.txt
```diff
@@ -256,15 +284,78 @@ purposes.
...
+\code{.cpp}
+typedef CGAL::Point_set_3<Kernel::Point_3> Point_set;
+typedef CGAL::Octree<Kernel, Point_set, Point_map, false> Octree;
+...
+Octree octree(points, points.point_map());
+\endcode
+
+The node class does not exist anymore and has been replaced by the lightweight type `Node_index`. All information formerly contained in the node class is now accessible via the `Orthtree` interface.
+
+Former node access was as follows:
```
sloriot, line 309: It would be helpful IMO to show how to access the points of a node in both versions

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -538,77 +629,131 @@ class Orthtree
...
+  \tparam OutputIterator must be a model of `OutputIterator` that accepts `GeomTraits::Node_data_element`
+
+  \param query the region to search within
+  \param k the number of elements to find
+  \param output the output iterator to add the found elements to (in order of increasing distance)
+
+  \warning Nearest neighbor searches requires `GeomTraits` to be a model of `CollectionPartitioningOrthtreeTraits`.
+ */
+  template <typename OutputIterator>
+  auto nearest_k_neighbors_in_radius(
```
afabri, line 730: Note that the function for points  has  no `_k_`  in the name. Shall we unify (adding or dropping)?

Orthtree/include/CGAL/Orthtree.h
```diff
@@ -538,77 +629,131 @@ class Orthtree
...
+  \tparam OutputIterator a model of `OutputIterator` that accepts `GeomTraits::Node_data_element` objects.
+
+  \param query query sphere
+  \param output output iterator
+
+  \warning Nearest neighbor searches requires `GeomTraits` to be a model of `CollectionPartitioningOrthtreeTraits`.
+ */
   template<typename OutputIterator>
-  OutputIterator nearest_neighbors (const Sphere& query, OutputIterator output) const {
+  auto neighbors_in_radius(const Sphere& query, OutputIterator output) const -> std::enable_if_t<supports_neighbor_search, OutputIterator> {
```
sloriot, line 708: In Kd-tree this function is called [`search()`](https://doc.cgal.org/latest/Spatial_searching/classCGAL_1_1Kd__tree.html#a4e760af1ae3da968dec45b378db207fc) and they also have `search_any()`

soesau: I think search could be confused with locate? I would just add _k_ to the nearest neighbor function without a radius to make it clear.

Orthtree/include/CGAL/Orthtree.h
soesau: I added `_k_`.

Orthtree/doc/Orthtree/Orthtree.txt
sloriot: @soesau did you see that comment? What do you think?

sloriot:
Successfully tested in 6.0-Ic-197

PR #7685, opened 2023-09-04
FIX:Change the return value to tree.root()->num_items().

xonx4l:
issue-:#7676

PR #7686, opened 2023-09-04
Correct size() in case some points were removed

### LINKED ISSUE #7676: The tree size of a kd tree didn't update after removal of points.
## Issue Details

The tree size of a kd tree didn't update after the removal of points. The tree size returned from [interface](https://github.com/CGAL/cgal/blob/v5.5.2/Spatial_searching/include/CGAL/Kd_tree.h#L685) is the length of the points vector stored in the tree, which does not be updated within the [remove function](https://github.com/CGAL/cgal/blob/v5.5.2/Spatial_searching/include/CGAL/Kd_tree.h#L499).   

Change the return value to `tree.root()->num_items()` resolves the problem.

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): centos
* Compiler: clang++
* Release or debug mode: debug
* Specific flags used (if any): tbb parallel tag
* CGAL version: latest
* Boost version: latest
* Other libraries versions if used (Eigen, TBB, etc.): tbb

sloriot:
Fixes  #7676

alternative fix to #7685

With this fix, `size()` stays `O(1)`

### DIFF
--- a/Spatial_searching/include/CGAL/Kd_tree.h
+++ b/Spatial_searching/include/CGAL/Kd_tree.h
@@ -137,7 +137,7 @@ private:
   mutable CGAL_MUTEX building_mutex;//mutex used to protect const calls inducing build()
   #endif
   bool built_;
-  bool removed_;
+  std::size_t removed_=0;
 
   // protected copy constructor
   Kd_tree(const Tree& tree)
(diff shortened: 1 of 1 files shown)


sloriot:
Successfully tested in CGAL-6.0-Ic-56

PR #7700, opened 2023-09-07
Incorrect closing HTML comment

albert-github:
Newer versions of doxygen are more restrictive (and following the rules) in respect to the end comment of a HTML type of comment, i.e. `<!-- ... -->`, so parts of the paragraphs were missing (Triangulation/index.html and Triangulation/group___pkg_triangulations_ref.html)

See also https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-comments

### DIFF
--- a/Triangulation/doc/Triangulation/PackageDescription.txt
+++ b/Triangulation/doc/Triangulation/PackageDescription.txt
@@ -67,7 +67,7 @@ of the underlying Euclidean space \f$ \mathbb{R}^d\f$.
 The neighbors of a full cell are also
 indexed  in such a way that the neighbor indexed by \f$ i\f$
 is opposite to the vertex with the same index.
---->
+-->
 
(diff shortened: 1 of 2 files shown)


afabri (approved):

afabri:
Don't we run a html-checker?

albert-github:
> Don't we run an html-checker?

I don't think a HTML checker would be of any use here as the mentioned code is interpreted by doxygen and doxygen follows, now, the rule about the `-->`. So doxygen already filters the input and the offending code does not land in the HTML output at all.
(might be an idea that doxygen would give a warning when it encounters the forbidden construct `--->`)

albert-github:
Create a proposed pull request for doxygen: https://github.com/doxygen/doxygen/pull/10298

sloriot:
Successfully tested in CGAL-6.0-Ic-61

PR #7704, opened 2023-09-07
Fix Is_in_domain potential infinite loop

sloriot:
if p is on a face then Is_in_domain will loop infinitely.

Side point: `CGAL_USE_ALL_INTERSECTIONS` I'm not sure it makes sense to keep both implementation. It is even not usable as you need to modify the header to use it.

If we want an exact solution, we should use `Compare_distance()(p, t1, t2)`.

### DIFF
--- a/Mesh_3/include/CGAL/Polyhedral_complex_mesh_domain_3.h
+++ b/Mesh_3/include/CGAL/Polyhedral_complex_mesh_domain_3.h
@@ -523,8 +523,8 @@ public:
       return r_domain_.bounding_aabb_tree_ptr()->
         first_intersected_primitive(ray);
     }
-
-#if USE_ALL_INTERSECTIONS
+#define CGAL_USE_ALL_INTERSECTIONS 0
+#if CGAL_USE_ALL_INTERSECTIONS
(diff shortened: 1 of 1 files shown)


Mesh_3/include/CGAL/Polyhedral_complex_mesh_domain_3.h
```diff
@@ -523,8 +523,8 @@ class Polyhedral_complex_mesh_domain_3
       return r_domain_.bounding_aabb_tree_ptr()->
         first_intersected_primitive(ray);
     }
-
-#if USE_ALL_INTERSECTIONS
+#define CGAL_USE_ALL_INTERSECTIONS 0
+#if CGAL_USE_ALL_INTERSECTIONS
     boost::optional<AABB_primitive_id> shoot_a_ray_2(const Ray_3 ray) const {
```
afabri, line 528:
```suggestion
    boost::optional<AABB_primitive_id> shoot_a_ray_2(const Ray_3& ray) const {
```

sloriot:
Successfully tested in CGAL-6.0-Ic-51

PR #7710, opened 2023-09-11
Fix concept doc

sloriot:
edge and face edges were exchanged.

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
@@ -58,15 +58,15 @@ public:
   /// \param i_id the id of the intersection point, starting at 0. Ids are consecutive.
   /// \param sdim indicates the dimension of the simplex part of the face that is intersected by the edge
   ///             (0 for a vertex, 1 for an edge, 2 for the interior of the face)
+  /// \param tm_e mesh containing `h_e`
+  /// \param tm_f mesh containing `h_f`
+  /// \param h_e a halfedge from `tm_e`
   /// \param h_f a halfedge from `tm_f` indicating the simplex intersected:
   ///            if `sdim==0` the target of `h_f` is the intersection point,
   ///            if `sdim==1` the edge of `h_f` contains the intersection point in its interior,
   ///            if `sdim==2` the face of `h_f` contains the intersection point in its interior.
-  /// \param h_e a halfedge from `tm_e`
   /// \param is_target_coplanar `true` iff the target of `h_e` is the intersection point
   /// \param is_source_coplanar `true` iff the source of `h_e` is the intersection point
-  /// \param tm_f mesh containing `h_f`
-  /// \param tm_e mesh containing `h_e`
   void intersection_point_detected(std::size_t i_id,
                                    int sdim,
                                    halfedge_descriptor h_e,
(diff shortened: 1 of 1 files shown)


Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
```diff
@@ -69,10 +69,10 @@ class PMPCorefinementVisitor{
...
   void intersection_point_detected(std::size_t i_id,
                                    int sdim,
-                                   halfedge_descriptor h_f,
                                    halfedge_descriptor h_e,
-                                   const Triangle_mesh& tm_f,
+                                   halfedge_descriptor h_f,
                                    const Triangle_mesh& tm_e,
+                                   const Triangle_mesh& tm_f,
                                    bool is_target_coplanar,
                                    bool is_source_coplanar);
```
albert-github, line 77: - Wouldn't it be better to also change the order of the `\param` commands so they represent the order of the arguments as well.
- doesn't this indicate, again, that it is not a good idea to have the used source and the documentation in separate files.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
```diff
@@ -58,15 +58,15 @@ class PMPCorefinementVisitor{
   /// \param i_id the id of the intersection point, starting at 0. Ids are consecutive.
   /// \param sdim indicates the dimension of the simplex part of the face that is intersected by the edge
   ///             (0 for a vertex, 1 for an edge, 2 for the interior of the face)
+  /// \param tm_e mesh containing `h_e`
+  /// \param tm_f mesh containing `h_f`
+  /// \param h_e a halfedge from `tm_e`
```
albert-github, line 63: Order is still a bit strange first description of `tm_*` and afterwards `h_*` whilst in the argument lists it is the other way around.

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
albert-github: > It's a concept so I don't see how it could be in the code.

I didn't look at the "surrounding" but my comment was based on the fact that the prototype is now at 2 places.

sloriot: Why 2 places? Are you talking of the default model?

albert-github: What I mean, as far as I understand it:
- the part used in the documentation: doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h at line 72
- the part used in the code: include/CGAL/Polygon_mesh_processing/internal/Corefinement/face_graph_utils.h at line 455

sloriot: That's what I said. These are 2 different things. One is a concept, the other one is the default model.

albert-github: Probably my lack of knowledge about concepts.

PR #7712, opened 2023-09-13
Adding a default to dynamic property maps

sloriot:
## TODO
- [x] think about `get(tag, property_map, value)` that may be an issue if coded not strictly enough for some models
- [x] find where the doc should be added


sloriot:

PR #7725, opened 2023-09-19
Bounding_volumes: Use Eigen library in demo

afabri:
## Summary of Changes

The minimal enclosing ellipse represented as center and major and minor axis vectors was not constructed correctly.
I rewrote the function using Eigen.

## Release Management

* Affected package(s):  Bounding Volumes


afabri:
@HoffmannMichael as the demo also computes the p-center I noticed an assertion on one of the data sets that I committed with your algorithm.

sloriot:

PR #7726, opened 2023-09-19
Added type cast to preprocess_weights() to avoid compilation error

Kuprin2000:
## Summary of Changes

A faced with an compilation error while using CGAL::extrude_skeleton() with float kernel. The error was caused by std::max() function getting arguments of different types. My improvement should fix this problem.

## Release Management

* Affected package(s): Straight Skeleton

### DIFF
--- a/Straight_skeleton_extrusion_2/include/CGAL/extrude_skeleton.h
+++ b/Straight_skeleton_extrusion_2/include/CGAL/extrude_skeleton.h
@@ -955,7 +955,7 @@ preprocess_weights(WeightRange& weights)
   // Since the max value might not be very close to 90°, take the max between of the large-% weight
   // and the weight corresponding to an angle of 89.9999999°
   const FT weight_of_89d9999999 = 572957787.3425436; // tan(89.9999999°)
-  const FT scaled_max = (std::max)(weight_of_89d9999999, 1e3 * max_value);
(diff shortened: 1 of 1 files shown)


MaelRL:
Thank you for the patch.

Which kernel are you using, that you have `float` as FT?

Kuprin2000:
> Thank you for the patch.
> 
> Which kernel are you using, that you have `float` as FT?

I was using CGAL::Simple_cartesian<float> kernel.

MaelRL:
The code for skeleton extrusion might work for this kernel, but if you want robustness you need exact predicates (i.e., a kernel such as `CGAL::Exact_predicates_inexact_constructions_kernel`). If the `float` rather than `double` aspect is important to you, you can have a look at the (currently undocumented) kernel added in https://github.com/CGAL/cgal/pull/6950.

lrineau:
I agree with you, Mael, but is not the patch right anyway? The `double` literal `1e3` should be converted to `FT`.

MaelRL:
Sorry if this wasn't clear: the patch is OK to be tested & merged.

The remark about the kernel is just a heads up to @Kuprin2000.

Kuprin2000:
> Sorry if this wasn't clear: the patch is OK to be tested & merged.
> 
> The remark about the kernel is just a heads up to @Kuprin2000.

Thank you very much! I will try to use kernel you suggested.

sloriot:

Issue #7730, opened 2023-09-20
Bounding volumes: assertion in rectangular_p_center


afabri:
## Issue Details

tbd 

## Source Code

_If your issue arises by using CGAL in your own source code, please provide a minimalist example that we can compile easily to reproduce the bug. If your issue arises from using a CGAL program (demo, example, etc.), please let us know which one. Helping you solving an issue is much easier and efficient if we can reproduce it._

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):
* Compiler:
* Release or debug mode:
* Specific flags used (if any):
* CGAL version:
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

PR #7731, opened 2023-09-20
Bounding_volumes: Add issue7730.cpp

afabri:
## Summary of Changes

Add the failing test case for Issue #7730

## Release Management

* Affected package(s): Bounding_volumes


afabri:
@sloriot note that this is only the proof of the bug and not the fix.

PR #7732, opened 2023-09-20
Reformulate a cross product to increase precision

MaelRL:
## Summary of Changes

The Lindstrom-Turk simplification method can suffer from significant numerical errors when the point coordinates are large (or there is a big difference of magnitude between coordinates). The main source was one cross product. 

A few fixes were tried before settling on a reformulation of the cross product based on the assumption that points will be nearby and have similar coordinates. It's not the best method for stability, but the others are too expensive.

I'm leaving the code in case this pops up again in the future.

Note: Garland-Heckbert simplification has no such issues.

## Release Management

* Affected package(s): `Surface_mesh_simplification`
* Issue(s) solved (if any):  n/a


Surface_mesh_simplification/test/Surface_mesh_simplification/test_edge_collapse_stability.cpp
```diff
@@ -0,0 +1,71 @@
...
+    return EXIT_FAILURE;
+  }
+
+  std::cout << "input has " << num_vertices(mesh) << " vertices." << std::endl;
+
+  CGAL::Iso_cuboid_3<Kernel> bbox(CGAL::Polygon_mesh_processing::bbox(mesh));
+
+  // scale a bit the bounding box bbox, because the kernel is SC
+  bbox = { bbox.min() - 0.01 * (bbox.max() - bbox.min()),
+           bbox.max() + 0.01 * (bbox.max() - bbox.min()) };
```
afabri, line 48:
```suggestion
  bbox = { (bbox.min)() - 0.01 * ((bbox.max)() - (bbox.min)()),
           (bbox.max)() + 0.01 * ((bbox.max)() - (bbox.min)()) };
```

afabri: Or maybe use [`Bbox_2.dilate()`](https://doc.cgal.org/latest/Kernel_23/classCGAL_1_1Bbox__2.html#a61646263f99a472f3a52a308075b9fe5)

sloriot:

Surface_mesh_simplification/examples/Surface_mesh_simplification/edge_collapse_surface_mesh.cpp
```diff
@@ -16,10 +16,12 @@ namespace SMS = CGAL::Surface_mesh_simplification;
 
 int main(int argc, char** argv)
 {
+  std::cout.precision(17);
```
sloriot, line 19: leftover?

lrineau: I will merge without those two added lines.

PR #7742, opened 2023-09-25
Add missing header

lrineau:
CGAL-5.5.x latest test results are completely broken.

https://cgal.geometryfactory.com/CGAL/testsuite/results-5.5.4-I-166.shtml

That is not due to any code change (there was none since the previous testsuite for 5.5.x). The error are due to Boost 1.84. Probably something change in the includes of Boost. The error is about `boost::prior`.  The header `<boost/next_prior.hpp>` must be included.


mglisse:
As part of the modernization effort, in master, could we use std::next and std::prev instead? Or do those boost versions have some special behavior?

lrineau:
> As part of the modernization effort, in master, could we use std::next and std::prev instead? Or do those boost versions have some special behavior?

The patch is for 5.5.x only. CGAL-5.6.x and `master` do not have the issue.

sloriot:

PR #7752, opened 2023-09-29
Add missing functions capacity and upper bound in dart_range for cmap and lcc with index

gdamiand:

## Summary of Changes

Add 2 missing methods in dart_range in CMap and LCC_cmap with index.
Without these two methods, we cannot use a dart_range like a compact container with index which is a bug.

## Release Management

* Affected package(s): CMap and LCC


sloriot:

PR #7753, opened 2023-10-02
Fix warning

sloriot:
The compiler complains about potential out of bound access of the array


lrineau (approved):

sloriot:

PR #7766, opened 2023-10-09
Correcting stylesheet name

albert-github:
The name `stylesheet.css` is replaced by a better name `cgal_stylesheet.css` a while ago, but apparently these 2 files were not updated.

### DIFF
--- a/Documentation/doc/scripts/generate_how_to_cite.py
+++ b/Documentation/doc/scripts/generate_how_to_cite.py
@@ -85,7 +85,7 @@ pre_html=r"""<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "htt
 <link rel="icon" type="image/png" href="../Manual/g-196x196-doc.png"/>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
 <meta http-equiv="X-UA-Compatible" content="IE=9"/>
-<link href="stylesheet.css" rel="stylesheet" type="text/css" />
+<link href="cgal_stylesheet.css" rel="stylesheet" type="text/css" />
 <title>CGAL ${CGAL_CREATED_VERSION_NUM} - Manual: Acknowledging CGAL</title>
 </head>
 <body>
--- a/Documentation/doc/scripts/html_output_post_processing.py
+++ b/Documentation/doc/scripts/html_output_post_processing.py
@@ -70,7 +70,7 @@ def clean_doc():
     duplicate_files=list(package_glob('./*/jquery.js'))
     duplicate_files.extend(package_glob('./*/dynsections.js'))
     duplicate_files.extend(package_glob('./*/resize.js'))
-    duplicate_files.extend(package_glob('./*/stylesheet.css'))
+    duplicate_files.extend(package_glob('./*/cgal_stylesheet.css'))
     # kill _all_, including the one in CGAL tabs.css files
     duplicate_files.extend(glob.glob('./*/tabs.css'))
     # left-over by doxygen?


lrineau:
I am surprised that https://doc.cgal.org/latest/Manual/how_to_cite_cgal.html does use `cgal_stylesheet.css`. Do you know why?

albert-github:
As far as I can quickly see there is only one place where the stylsheet is used i.e. `<h1>`, but I could imagine that is would be used at other places as well (the `<a>`, the `<pre>`)

At the moment the `how_to_cite.html` (not the `how_to_cite_cgal` is the file we are talking about) is just a "stand alone" page that is called from `gow_to_cit_cgal.html` through the `bib` links without the "CGAL header" and without the tree view. I think it would also be worthwhile to adjust this (but that would be something for a different enhancement issue / pull request).

At the moment though the usage of `stylesheet,css` is not correct and should either be removed or replaced as suggested in this proposed PR.

Documentation/doc/scripts/generate_how_to_cite.py
```diff
@@ -85,7 +85,7 @@
 <link rel="icon" type="image/png" href="../Manual/g-196x196-doc.png"/>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
 <meta http-equiv="X-UA-Compatible" content="IE=9"/>
-<link href="stylesheet.css" rel="stylesheet" type="text/css" />
+<link href="cgal_stylesheet.css" rel="stylesheet" type="text/css" />
```
lrineau, line 88: That change is wrong. The file `stylesheet.css` is used and should be deduplicated as well. Here is the right patch:

```suggestion
<link href="stylesheet.css" rel="stylesheet" type="text/css" />
<link href="cgal_stylesheet.css" rel="stylesheet" type="text/css" />
```

lrineau (changes requested):

Documentation/doc/scripts/generate_how_to_cite.py
lrineau: Oops! I am wrong.

Documentation/doc/scripts/html_output_post_processing.py
```diff
@@ -70,7 +70,7 @@ def clean_doc():
     duplicate_files=list(package_glob('./*/jquery.js'))
     duplicate_files.extend(package_glob('./*/dynsections.js'))
     duplicate_files.extend(package_glob('./*/resize.js'))
-    duplicate_files.extend(package_glob('./*/stylesheet.css'))
+    duplicate_files.extend(package_glob('./*/cgal_stylesheet.css'))
```
lrineau, line 73:
```suggestion
    duplicate_files.extend(package_glob('./*/stylesheet.css'))
    duplicate_files.extend(package_glob('./*/cgal_stylesheet.css'))
```

lrineau (changes requested):

Documentation/doc/scripts/generate_how_to_cite.py
lrineau: That change is okay. I tested it.

Documentation/doc/scripts/html_output_post_processing.py
albert-github: Why as `stylesheet.css` does not exist anymore (as far as I can see in the 1.8.13 ... versions)?

lrineau: You are right. Your pull-request is already fine with recent versions of CGAL. I searched in an older version I had on my computer.

lrineau (approved):

sloriot:
Successfully tested in CGAL-6.0-Ic-86

Issue #7765, opened 2023-10-09
Problem with Boost 1.54 and a non reachable ticket


albert-github:
In the code we see in Point_set_processing_3/include/CGAL/mst_orient_normals.h (output: Point_set_processing_3/group___pkg_point_set_processing3_algorithms.html):
```
   \warning This function may fail when Boost version 1.54 is used,
   because of the following bug: https://svn.boost.org/trac/boost/ticket/9012
```
though https://svn.boost.org/trac/boost/ticket/9012 is not reachable

- is this warning still valid?
- when valid what is the correct reference?

lrineau:
What is more, Boost 1.54 is no longer supported by CGAL (but only versions >= 1.66 as of CGAL-5.6) . That warning should be removed entirely.

albert-github:
I've just pushed a proposed patch, pull request #7777

albert-github:
**Note** in the code of Point_set_processing_3/include/CGAL/mst_orient_normals.h we also see:
```
#include <CGAL/boost/graph/dijkstra_shortest_paths.h> // work around a
                                                      // bug in boost
                                                      // 1.54
```
the question here is what to do with this / what can also be removed (I cannot do this as I cannot compile CGAL / test it).

PR #7777, opened 2023-10-10
issue #7765 Problem with Boost 1.54 and a non reachable ticket

albert-github:
CGAL as of version 5.6 requires Boost 1.66 or higher so the warning is not valid anymore


sloriot:

PR #7783, opened 2023-10-10
Ipelet for k-nearest neighbor graph

dfunke:
## Summary of Changes

This PR adds an Ipelet to draw the nearest-neighbor graph. It is based on the EMST Ipelet.

## Release Management

* Affected package(s): CGAL_Ipelets
* Issue(s) solved (if any): 
* Link to compiled documentation (obligatory for small feature):
* License and copyright ownership: as you require


mglisse:
Ah, I thought this would be a kNN graph with an input parameter k, but this is just the case k=1. It looks like there is an overload of `CGAL::nearest_neighbors` that takes a parameter `k`, if you decide to generalize the ipelet.

CGAL_ipelets/demo/CGAL_ipelets/nng.cpp
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  Triangulation t(pt_list.begin(), pt_list.end());
+
+  for(auto v = t.finite_vertices_begin();
+       v != t.finite_vertices_end();
+       ++v){
+
+    auto nn = CGAL::nearest_neighbor(t, v);
+    draw_in_ipe(Kernel::Segment_2(v->point(), nn->point()));
```
mglisse, line 55: It looks like if A is the nearest neighbor of B and vice versa, we can add the segment twice. Is that a feature (oriented graph), a "who cares", or a bug?

CGAL_ipelets/demo/CGAL_ipelets/nng.cpp
```diff
@@ -0,0 +1,60 @@
...
+  std::list<Point_2> pt_list;
+
+  read_active_objects(
+    CGAL::dispatch_or_drop_output<Point_2>(
+      std::back_inserter(pt_list)
+    )
+  );
+
+  if (pt_list.empty()) {
+    print_error_message("No mark selected");
```
mglisse, line 44: What happens if `pt_list` has a single point?

dfunke: With the move to the kNNG the return vector is simply empty

CGAL_ipelets/demo/CGAL_ipelets/nng.cpp
dfunke: This is a feature: if the user adds forward arrows to the output, then it the edges correctly point at A and B

dfunke:
Thank you for the excellent suggestion. I generalized the Ipelet to compute the kNNG.

> Ah, I thought this would be a kNN graph with an input parameter k, but this is just the case k=1. It looks like there is an overload of `CGAL::nearest_neighbors` that takes a parameter `k`, if you decide to generalize the ipelet.

afabri:
Can you please change the name to something with "nearest_neighbors" in it.

sloriot:
Successfully tested in CGAL-6.0-Ic-116

PR #7786, opened 2023-10-11
Remove double defgroup definition

albert-github:
Remove double defgroup definition


sloriot:
Successfully tested in CGAL-6.0-Ic-86

PR #7795, opened 2023-10-12
CGAL Lab: fix reading of VTK files

lrineau:
## Summary of Changes

CGAL Lab: fix reading of VTK files

  - fix a memory leak
  - allow to read a polygon soup, in addition to surface meshes
  - fix reading of vtp files

## Release Management

* Affected package(s): CGAL Lab
* License and copyright ownership: unchanged, maintenance by GF

### DIFF
--- a/Polyhedron/demo/Polyhedron/TextRenderer.cpp
+++ b/Polyhedron/demo/Polyhedron/TextRenderer.cpp
@@ -1,7 +1,6 @@
 #include <CGAL/Three/TextRenderer.h>
 #include <CGAL/Three/Scene_item.h>
 #include <CGAL/Three/Scene_print_item_interface.h>
-#include "Scene_polyhedron_selection_item.h"
 void TextRenderer::draw(CGAL::Three::Viewer_interface *viewer, const QVector3D& scaler)
 {
     QPainter *painter = viewer->getPainter();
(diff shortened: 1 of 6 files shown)


Polyhedron/demo/Polyhedron/TextRenderer.cpp
```diff
@@ -1,7 +1,6 @@
 #include <CGAL/Three/TextRenderer.h>
 #include <CGAL/Three/Scene_item.h>
 #include <CGAL/Three/Scene_print_item_interface.h>
-#include "Scene_polyhedron_selection_item.h"
```
afabri, line 4: Without this `#include`  the target  `scene_points_with_normal_items`  fails with a [linker error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-93/Polyhedron_Demo/TestReport_Christo_MSVC-2022-Community-Release.gz).

lrineau: > I do not understand why the CI does not catch this. It compiles the Polyhedron demo.

The CI compiles the demo under Linux. The linker has a different behavior.

I have fixed the issue with 7dc4b624aaae6c0270deb67a5efbf993a65a97c5. Now the destructor code has its own `.cpp` file, compiled into the library `demo_framework`.

sloriot:

PR #7809, opened 2023-10-19
fix out-of-bound warnings

sloriot:
The sanity value was not used so let's set it to 3.


sloriot:

PR #7810, opened 2023-10-19
Filtered_kernel: Remove dead code

### LINKED ISSUE #7718: Dead code in static filter Power_side_of_oriented_power_sphere_3
https://github.com/CGAL/cgal/blame/187409888bd8590ad71f2a4912f53d9014a6cc27/Filtered_kernel/include/CGAL/Filtered_kernel/internal/Static_filters/Power_side_of_oriented_power_sphere_3.h#L374
As far as I can see, `int_tmp_result` can only be ±1 (line 348 or 354), and `cmp` is a copy, so the comparison `cmp!=0` should always be true. Since the `if` branch ends with a `return`, this suggests that the next 170 lines are unreachable. Did I misread the code?
(if I was more serious, I would add an assertion and run some tests to confirm, or check if there is a coverage report of CGAL's testsuite somewhere online)
See also #2067 since this is the version with 4 points.

afabri:
## Summary of Changes

Remove dead code spotted by @mglisse that was generated by the filter generator.

## Release Management

* Affected package(s):  Filtered_kernel
* Issue(s) solved (if any): fix #7718


sloriot:

PR #7820, opened 2023-10-23
Prevent the removal of fixed boundary edges

sloriot:
Add a check before calling `remove_degerenate_faces()` to avoid removing patch border edges


sloriot:

PR #7822, opened 2023-10-23
Kernel_23: Exploit Uncertain in collinear_3()

afabri:
## Summary of Changes

Make `collinear_3()` exploit uncertainty to make it faster. See discussion in PR #5207  

## Release Management

### DIFF
--- a/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
+++ b/Cartesian_kernel/include/CGAL/predicates/kernel_ftC3.h
@@ -102,12 +102,14 @@ collinearC3(const FT &px, const FT &py, const FT &pz,
   FT dqx = qx-rx;
   FT dpy = py-ry;
   FT dqy = qy-ry;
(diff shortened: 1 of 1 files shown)


sloriot:
I think the new code and the old code are doing exactly the same thing when it comes to filter failure. Indeed the only way you would save something is if the first determinant is -1 or +1, which is not possible as 0 would also be a positive value in that case.

mglisse:
> the only way you would save something is if the first determinant is -1 or +1

Why? If the first sign of determinant is an uncertain containing 0 and at least one of ±1, the old code will throw trying to convert to bool in `if`, while the new code will continue (and if the second determinant is clearly non-zero, it will be able to conclude without going to exact computation).

sloriot:
Indeed I missed the fact that `is_zero` is uncertain. Looking again at `CGAL_AND`, I agree that exact computation will be triggered only if the interval/static det signs are all indeterminate.

sloriot:

PR #7824, opened 2023-10-24
Fix example mesh_3D_gray_vtk_image

ange-clement:
## Summary of Changes

Fixes example mesh_3D_gray_vtk_image.
Changed : Meshing all value **under** isovalue --> Meshing all values **over** isovalue.

## Release Management

* Affected package(s): Mesh_3
* Issue(s) solved (if any): 

### DIFF
--- a/Mesh_3/examples/Mesh_3/mesh_3D_gray_vtk_image.cpp
+++ b/Mesh_3/examples/Mesh_3/mesh_3D_gray_vtk_image.cpp
@@ -36,14 +36,15 @@ typedef CGAL::Mesh_complex_3_in_triangulation_3<Tr> C3t3;
 // Criteria
 typedef CGAL::Mesh_criteria_3<Tr> Mesh_criteria;
 
-class Less {
+template <bool less_than_iso>
+class Is_interior {
   double iso;
 public:
-  Less(double iso): iso(iso) {}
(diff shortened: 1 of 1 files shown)


janetournois:
Using `Less` or `More` actually depends on the input data.
`More` is the default behavior when using `isovalue` only, and using `Less` gives another example. This should be better documented

janetournois (commented):
Using `Less` or `More` actually depends on the input data.
Using `More` is the default behavior when using `isovalue` only, and using `Less` gives another example. This should be better documented

lrineau:
@janetournois @ange-clement As far as I understand, that pull-request is debated. I will not merge it, even if it was tested successfully.

sloriot:
@janetournois what is the status of this PR?

janetournois:
This modifies the example of Mesh_3 to make it work with different data sets, and it does not bring any new info or feature, so imho it can be closed.

sloriot:
What about adding an option in the parameter to pick less or more?

janetournois:
Good idea @sloriot !
However meshing this data will fail with `less=false`

sloriot:
> Good idea @sloriot ! However meshing this data will fail with `less=false`

which is why the default is true

sloriot:
Successfully tested in CGAL 6.1-Ic-14

PR #7825, opened 2023-10-24
Mesh_3 - avoid a timeout in the testsuite

janetournois:
## Summary of Changes

Tests of Mesh_3 in the daily testsuite (too) often end with timeouts, because of the test `test_meshing_without_features_determinism`. This happens on many different platforms.

This PR reduces the target dihedral angle from 10 degrees to 9 degrees for `CGAL::perturb_mesh_3()` to fit in the testsuite timeout.

My experiments showed that the perturber always succeeds to get to 10 degrees, but it may be long and go beyond the timeout, in particular in Debug mode. This test is not about the perturber itself, but about determinism, so it must not be considered as a regression.

## Release Management

* Affected package(s): Mesh_3 tests

### DIFF
--- a/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
+++ b/Mesh_3/test/Mesh_3/test_meshing_without_features_determinism.cpp
@@ -24,6 +24,7 @@
 // To avoid verbose function and named parameters call
 using namespace CGAL::parameters;
 
+
 template <typename Concurrency_tag>
 void test()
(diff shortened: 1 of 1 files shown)


sloriot:
7 timeouts in 6.0-I-90 and 4 in CGAL-6.0-Ic-93

sloriot:
Successfully tested in CGAL-6.0-Ic-96 (no time out!)

Issue #7121, opened 2022-12-13
CGAL overrides BUILD_TESTING for parent project when consumed as a third-party library


puneetmatharu:
## Issue Details

I'm working on a project that consumes CGAL as a third-party library via FetchContent. After installing my project as well as CGAL, my `<project-name>Config.cmake` file locates CGAL using `find_dependency(...)`. When CGAL is found this way, it inevitably reads the file `<INSTALL_DIR>/lib/cmake/CGAL/CGAL_add_test.cmake` which hardcodes the value of `BUILD_TESTING` as a cache variable. When I build my tests as a separate project, this disables testing for them and I have to explicitly pass `-DBUILD_TESTING=ON` at CMake configure time to reenable them. 

According to Scott Craig's book ["Professional CMake: A Practical Guide"](https://crascit.com/professional-cmake/), it is best to avoid explicitly setting `BUILD_TESTING` in case the project is being consumed as a subproject:

```txt
31.2. Don’t Assume A Top Level Build

...

Another common problem is modifying variables that have the potential to affect the whole build,
not just the project. A classic example is forcing the value of cache variables like BUILD_SHARED_LIBS,
BUILD_TESTING, CMAKE_BUILD_TYPE, CMAKE_MODULE_PATH or CMAKE_RUNTIME_OUTPUT_DIRECTORY. Using the FORCE
or INTERNAL keywords with the set() command on these variables prevents a parent project from
overriding that choice. Quite often, these variables shouldn’t be defined by the project as cache
variables at all, and maybe shouldn’t even be set by the project to begin with. For variables like
CMAKE_MODULE_PATH, append to existing values rather than replace them.
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): not relevant
* Compiler: not relevant
* Release or debug mode: not relevant
* Specific flags used (if any): not relevant
* CGAL version: 5.5.1
* Boost version: 1.80.0
* Other libraries versions if used (Eigen, TBB, etc.): none

## Solution

I think you can get around this by renaming `BUILD_TESTING`. For example, to make it project-specific you may wish to use `CGAL_BUILD_TESTING`. I have attached a patch file (see below) that contains the required changes applied to CGAL release version 5.5.1.

[patch.zip](https://github.com/CGAL/cgal/files/10222532/patch.zip)

lrineau:
Hi, I am glad you are using CGAL for your project. :tada:

puneetmatharu:
[...]

PR #7845, opened 2023-11-02
Patch usage of BUILD_TESTING to handle the case that CGAL is a subproject

puneetmatharu:
## Summary of Changes

Solves issue [#7121](https://github.com/CGAL/cgal/issues/7121).

Currently, CGAL overrides the `BUILD_TESTING` CMake variable. As is, if CGAL is pulled in as a subproject (without setting `BUILD_TESTING`), it will cause tests in parent projects to also be disabled. I have changed the name of this variable to `CGAL_ENABLE_TESTING` to be project-specific. 

In Scott Craig's book ["Professional CMake: A Practical Guide"](https://crascit.com/professional-cmake/), he states that it is best to avoid explicitly setting `BUILD_TESTING` in case the project is being consumed as a subproject:

```text
31.2. Don’t Assume A Top Level Build

...

Another common problem is modifying variables that have the potential to affect the whole build,
not just the project. A classic example is forcing the value of cache variables like BUILD_SHARED_LIBS,
BUILD_TESTING, CMAKE_BUILD_TYPE, CMAKE_MODULE_PATH or CMAKE_RUNTIME_OUTPUT_DIRECTORY. Using the FORCE
or INTERNAL keywords with the set() command on these variables prevents a parent project from
overriding that choice. Quite often, these variables shouldn’t be defined by the project as cache
variables at all, and maybe shouldn’t even be set by the project to begin with. For variables like
CMAKE_MODULE_PATH, append to existing values rather than replace them.
```

## Release Management

* Issue(s) solved (if any): fix [#7121](https://github.com/CGAL/cgal/issues/7121)
* Link to compiled documentation (obligatory for small feature): *Not run.*
* License and copyright ownership: [REUSE Compliance Check](https://github.com/puneetmatharu/cgal/actions/workflows/reuse.yml)
* [Link to triggered GitHub workflows](https://github.com/puneetmatharu/cgal/actions)


puneetmatharu:
@lrineau here is the PR, as per your request in https://github.com/CGAL/cgal/issues/7121.

lrineau (approved):

lrineau:
Thanks @puneetmatharu 

@sloriot I think this PR is ready to be tested. The only reserve I have is about the needed changes to `Scripts/developer_scripts/run_testsuite_with_ctest`:

https://github.com/CGAL/cgal/pull/7845/files#diff-c4688d05e1fdb810fb3c3410acfdedc7a188a0b424aec7d8ce638921a4e97234

But I think that script is called from `autotest_cgal_with_ctest` and found in the internal release itself. That should work.

afabri:
No need to backport to the bug fix branch?

lrineau:
> No need to backport to the bug fix branch?

I do not think so. It is a bug-fix because it does fix a bug report. But the scenario that is fixed is not mainstream (though it seems to be used more and more in the CMake ecosystem). And at the same time that is a small breaking change, because the name of the CMake option (in `CMakeCache.txt` and CMake GUI) changes. It seems right to make the change in `master` only (targeting CGAL-6.0).

sloriot:
Testsuite showed that for docker `BUILD_TESTING=ON` should imply `CGAL_BUILD_TESTING=ON` (and it also seems reasonable).

lrineau:
@puneetmatharu We tested this PR last night, and the result were not great:
https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-104.shtml

All those blue columns are because `-DBUILD_TESTING=ON` was passed to the CMake command line, but not `-DCGAL_BUILD_TESTING=ON`.

We could "fix" that, in our test scripts, but I think there might have other users of CGAL using `BUILD_TESTING` (in particular I think about the various CGAL packages in package managers). I think that one should add a compatibility: the default value of `CGAL_BUILD_TESTING` should be the value of `BUILD_TESTING`.

According to the [documentation of the CMake `option` command][doc], the default value can be empty. That is why I suggest to define `CGAL_BUILD_TESTING` that way:

```cmake
option(CGAL_ENABLE_TESTING "Build the testing tree." ${BUILD_TESTING})
```

If `BUILD_TESTING` is undefined, then it will evaluate to an empty value, and the default value of `CGAL_BUILD_TESTING`, according to [the documentation][doc], will be `OFF`.

[doc]: https://cmake.org/cmake/help/latest/command/option.html

puneetmatharu:
> @puneetmatharu We tested this PR last night, and the result were not great: https://cgal.geometryfactory.com/CGAL/testsuite/results-6.0-Ic-104.shtml
> 
> All those blue columns are because `-DBUILD_TESTING=ON` was passed to the CMake command line, but not `-DCGAL_BUILD_TESTING=ON`.
> 
> We could "fix" that, in our test scripts, but I think there might have other users of CGAL using `BUILD_TESTING` (in particular I think about the various CGAL packages in package managers). I think that one should add a compatibility: the default value of `CGAL_BUILD_TESTING` should be the value of `BUILD_TESTING`.
> 
> According to the [documentation of the CMake `option` command](https://cmake.org/cmake/help/latest/command/option.html), the default value can be empty. That is why I suggest to define `CGAL_BUILD_TESTING` that way:
> 
> ```cmake
> option(CGAL_ENABLE_TESTING "Build the testing tree." ${BUILD_TESTING})
> ```
> 
> If `BUILD_TESTING` is undefined, then it will evaluate to an empty value, and the default value of `CGAL_BUILD_TESTING`, according to [the documentation](https://cmake.org/cmake/help/latest/command/option.html), will be `OFF`.

I absolutely agree -- this is an oversight on my side. Thank you for spotting and raising this. I have changed my code accordingly.

puneetmatharu:
Brilliant, thank you.

sloriot:
Successfully tested in CGAL-6.0-Ic-116

PR #7848, opened 2023-11-06
CGAL::IO::Internal::read_OBJ(): allow arbitrary whitespace in "f" lines

sxntxn:
The "OBJ" file format does not specify which whitespace to use, so be more tolerant about it.
This allows correct reading of OBJ files where only ```\t``` (tabulator) is used as a separator in the ```f``` lines.


sloriot:
@MaelRL what TODO do you have in mind?

sloriot:
@MaelRL ping!!!!

MaelRL (approved):

sloriot:
Successfully tested in CGAL-6.0-Ic-224

PR #7860, opened 2023-11-13
Add missing #include to Surface_sweep_2_impl.h

vslashg:
This header uses std::set, but doesn't include <set>.  This can break builds depending on how this header is included.


sloriot:
Thanks for taking the time to submit a patch. Note that the CI does check for such things. This header is internal is not meant to be included by a user. To me the patch is not needed.

lrineau:
Spam (the GSoC 2024 has been announced, right?)

PR #7873, opened 2023-11-20
Isotropic remeshing: add example with custom dummy sizing

sloriot:
Add a sizing interpolating the size along the y-axis and also improve the concept.

Screenshot from 2023-11-20 09-43-26

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
@@ -60,7 +60,7 @@ Point_3 split_placement(const halfedge_descriptor h,
                         const PolygonMesh& pmesh) const;
 
 /// function called after the addition of the split vertex `v` in `pmesh`.
-/// This function can be used for exemple to update a pre-computed sizing field.
+/// This function can be used for example to update a pre-computed sizing field.
 void register_split_vertex(const vertex_descriptor v,
                            const PolygonMesh& pmesh);
 
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
@@ -16,6 +16,7 @@ class PMPSizingField{
 public:
 
 /// @name Types
+///  These types are used for the documentation of the functions of the concept and not needed implementation wise.
 /// @{
 
 /// Vertex descriptor type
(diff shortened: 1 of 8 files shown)


Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
```diff
@@ -38,27 +39,29 @@ typedef unspecified_type FT;
...
 
-/// a function returning the location of the split point of the edge of `h`.
+/// returns the position of the new vertex created when splitting the edge of `h`.
 Point_3 split_placement(const halfedge_descriptor h,
                         const PolygonMesh& pmesh) const;
 
-/// a function that updates the sizing field value at the vertex `v`.
-void update(const vertex_descriptor v,
-            const PolygonMesh& pmesh);
+/// function called after the addition of the split vertex `v` in `pmesh`.
```
janetournois, line 62: shouldn't we mention that the sizing field at `v` may have to be updated?

sloriot: The sizing field class does register it and does what it needs to do. Uniform_sizing is not using it and neither does the custom example. Something like, "can be used to updated pre-computed sizing" can be added if you want to insist on that point.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/Adaptive_sizing_field.h
```diff
@@ -38,6 +38,8 @@ namespace Polygon_mesh_processing
 * Edges too long with respect to the local target edge length are split in two, while
 * edges that are too short are collapsed.
 *
+* This class depends on the Eigen library.
```
albert-github, line 41: `Eigen`  -> `\eigen` ?

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPSizingField.h
janetournois: Yes I think that it is important to say it. If there is nothing to update, then it's easy. Otherwise it's good to know that it is here that the user should do it.

sloriot:

lrineau:
This PR was successfully tested, but there are two unresolved discussion:
- https://github.com/CGAL/cgal/pull/7873#pullrequestreview-1739595023 between @janetournois and @sloriot,
- https://github.com/CGAL/cgal/pull/7873#pullrequestreview-1739686797 between @albert-github  and @sloriot.

For the moment, I will not merge this PR. I let you go on with the discussions.

sloriot:
I fixed both requests that are doc only in doc only files. Local doc testsuite is fine with the changes. Restoring "Tested" flag.

PR #7892, opened 2023-11-28
fix dihedral angle computation

sloriot:


lrineau (approved):

Kernel_23/test/Kernel_23/test_approximate_dihedral_angle_3.cpp
```diff
@@ -10,10 +10,51 @@ struct query {
...
+  assert( CGAL::approximate_dihedral_angle(a, b, d, c) < 0);
+  assert( CGAL::approximate_dihedral_angle(c, a, d, b) < 0);
+  assert( CGAL::approximate_dihedral_angle(a, d, c, b) < 0);
+  assert( CGAL::approximate_dihedral_angle(c, b, a, d) < 0);
+  assert( CGAL::approximate_dihedral_angle(d, b, c, a) < 0);
+  assert( CGAL::approximate_dihedral_angle(d, c, a, b) < 0);
+}
+
+auto almost_equal_angle(double a, double b) {
+  return std::min(std::abs(a - b), std::abs(a + 360 - b)) < 0.1;
```
afabri, line 33:
```suggestion
  return (std::min)(std::abs(a - b), std::abs(a + 360 - b)) < 0.1;
```

afabri: Is 0.1  a reasoned choice?

afabri:
@sloriot can you check that[ `compare_dihedral_angle()`](https://doc.cgal.org/latest/Kernel_23/group__compare__dihedral__angle__grp.html#ga52404107d79baa56301bfc8199134228)  is correct.

Kernel_23/test/Kernel_23/test_approximate_dihedral_angle_3.cpp
lrineau: (Sorry I forgot about the protection against the `min` macro.)

About `0.1`, that is a small value (for an angle in degrees) and it worked for me. We could try another smaller value, as soon as it passes the tests. I let you modify that value if you want.

sloriot:

PR #7900, opened 2023-12-01
Polyhedron demo - fix `DoubleEdit`

janetournois:
## Summary of Changes

With Qt6, scientific notation was used by default. This PR changes it to normal notation for doubles.

~~Note that I had to remove the call to `QDoubleValidator::fixup(input);` to make it work, otherwise I could not add a "." in the number I wanted to enter. I did not find any satisfying explanation for that, but it works...~~

## Release Management

### DIFF
--- a/Polyhedron/demo/Polyhedron/CGAL_double_edit.cpp
+++ b/Polyhedron/demo/Polyhedron/CGAL_double_edit.cpp
@@ -9,18 +9,16 @@ public:
     : QDoubleValidator(parent)
   {
     setLocale(QLocale::C);
-    setNotation(QDoubleValidator::StandardNotation);
   }
 
   void fixup ( QString & input ) const
   {
     input.replace(".", locale().decimalPoint());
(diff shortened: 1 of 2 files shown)


Polyhedron/demo/Polyhedron/CGAL_double_edit.cpp
```diff
@@ -9,13 +9,14 @@ class DoubleValidator : public QDoubleValidator
...
     setLocale(QLocale::C);
+    setNotation(QDoubleValidator::StandardNotation);
   }
 
   void fixup ( QString & input ) const
   {
     input.replace(".", locale().decimalPoint());
     input.replace(",", locale().decimalPoint());
-    QDoubleValidator::fixup(input);
+//    QDoubleValidator::fixup(input);
```
lrineau, line 19: `fixup` should call its base `fixup` implementation. If that does not work, then it means that the setting of the validator is wrong. Maybe the range?

sloriot:

PR #7907, opened 2023-12-05
Prevent out-of-bound warnings

sloriot:
Fixes warnings for example [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-121/Mesh_3/TestReport_lrineau_Fedora-rawhide-Release.gz)


lrineau (changes requested):
As I said in a private discussion, I am opposed to that change, unless all other possible ways to fix that warning are tested before (and in particular `CGAL_assume(i != j)` in CGAL functions using that table).

lrineau (approved):

afabri:
Does the `assume` make the warning go away ????

sloriot:
Successfully tested in CGAL-6.0-Ic-122 (the warning is gone)

Issue #6267, opened 2022-01-24
Wrong result in CGAL::intersection() for planes


daniel-zint:
## Issue Details

For some specific values, the computation of plane intersections delivers significantly different results in comparison with solving the linear system of equations with Eigen. Note that when changing the input planes only slightly, the results are the same again.

## Source Code

```
typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
typedef Kernel::Point_3 Point;
typedef CGAL::Plane_3<Kernel> Plane_3;

Plane_3 p1( -1.0, 0.0, 0.0, 0.42 );
Plane_3 p2( -2.7782134062978145e-17, 1.7751377254852134e-17, 0.99999999999999978, -0.099999999999999964 );
Plane_3 p3( -0.84267313319221382, 0.53842547357644333, -3.2969051662693791e-17, 0.38199815134456822 );
auto intersection_point = CGAL::intersection( p1, p2, p3 );
auto cgal_point         = boost::get<Point>( intersection_point.get() );
std::cout << "CGAL point:  " << cgal_point << std::endl;

Eigen::Matrix3d A;
A << p1.a(), p1.b(), p1.c(), p2.a(), p2.b(), p2.c(), p3.a(), p3.b(), p3.c();
Eigen::Vector3d b( -p1.d(), -p2.d(), -p3.d() );
Eigen::Vector3d x = A.colPivHouseholderQr().solve( b );
Point eigen_point = { x[0], x[1], x[2] };
std::cout << "Eigen point: " << eigen_point << std::endl;
```

Output:
```
CGAL point:  0.42 0 0.1
Eigen point: 0.42 -0.0521436 0.1
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 64 bits
* Compiler: Visual Studio 2019 (v142)
* Release or debug mode: both
* Specific flags used (if any):
* CGAL version: Github Version from July 4th, commit: e7f2e6f162
* Boost version: 1.77.0
* Other libraries versions if used (Eigen, TBB, etc.): Eigen 3.3.9

sloriot:
Useful FAQ entry:
https://www.cgal.org/FAQ.html#inexact_NT

Did you try with `Exact_predicates_exact_constructions_kernel` and got the same issue?

lrineau:
There is a general issue in the CGAL code for constructions:
  - the code was written with exact arithmetic in mind,
[...]

PR #7913, opened 2023-12-06
Increase precision of the intersection of 3 planes

sloriot:
Addresses #6267


sloriot:
the patch was compiling with master but it does not in 5.5. I need to adjust it...

PR #7914, opened 2023-12-06
PMP remesh_planar_patches: handle const meshes

sloriot:


sloriot:
Successfully tested in CGAL-6.0-Ic-124

PR #7917, opened 2023-12-07
PMP: Write into different files to avoid concurrent writing

afabri:
## Summary of Changes

As test programs may be executed in parallel we have to write in files with different names,

## Release Management


sloriot:
Successfully tested in CGAL-6.0-Ic-124

Issue #7859, opened 2023-11-13
Interpolated Curvature remaining TODOs


sloriot:
Remaining issue to be handled before 6.0 in order not to block the integration of the small feature branch on which two other PRs depend on.

- [ ] Update Results & Performance section
- [ ] Add a comment in the user manual about the relationship to the package [Estimation of Local Differential Properties of Point-Sampled Surfaces](https://doc.cgal.org/latest/Manual/packages.html#PkgJetFitting3)
- [ ] Move `average_edge_length()` and maybe other helper function in `measure.h` or other files.
- [ ] Check that the display plugin works as expected
- [ ] No plugin for principal curvature directions?
- [ ] remove Surface_mesh properties created in the plugin
- [ ] restrict named parameters (for @sloriot it is easy to make a mistake with and without the_map suffix)
- [ ] Why is ball radius default -1 and not 0

PR #7919, opened 2023-12-08
Interpolated Curvature remaining TODOs

hoskillua:
## Summary of Changes
Remaining issue #7859 to be handled before 6.0 in order not to block the integration of the small feature branch on which two other PRs depend on.

Todos (taken from the issue)
* [x]  Update Results & Performance section
* [x]  Add a comment in the user manual about the relationship to the package [Estimation of Local Differential Properties of Point-Sampled Surfaces](https://doc.cgal.org/latest/Manual/packages.html#PkgJetFitting3)
* [x]  Move `average_edge_length()` and maybe other helper function in `measure.h` or other files.
* [x]  Check that the display plugin works as expected
* [x]  plugin for principal curvature directions? 
* [x]  plugin for principal curvature directions -> make sure the directions make sense in the concave (both negative) surface areas
* [x]  <s>remove Surface_mesh properties created in the plugin</s> not specific to this method and not a big deal.
* [ ]  restrict named parameters (for @sloriot it is easy to make a mistake with and without the_map suffix)
* [x]  Why is ball radius default -1 and not 0

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Polygon_mesh_processing.txt
@@ -1027,7 +1027,7 @@ These computations are performed using (on all vertices of the mesh) `CGAL::Poly
 where function named parameters are used to select the curvatures (and possibly directions) to be computed. An overload function with the same name
 but taking a given vertex is also available in case the computation should be done only for that vertex.
 
-\subsection ICCResults Results & Performance
+\subsection ICCResults Results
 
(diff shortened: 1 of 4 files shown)


sloriot:
We should add in the doc what happens if **val = 0** as it is not documented for now.

hoskillua:
> We should add in the doc what happens if **val = 0** as it is not documented for now.

This was added here: https://cgal.github.io/7919/v1/Polygon_mesh_processing/index.html#ICCBackground

sloriot:
Successfully tested in CGAL-6.0-Ic-156

PR #7935, opened 2023-12-18
CGAL Lab: Transformation of point normals

soesau:
## Summary of Changes

Affine transformation transforms points, but didn't transform their normals. This PR takes care of normal transformation of point sets.

## Release Management

* Affected package(s): Polyhedron Demo


sloriot:
Successfully tested in CGAL-6.0-Ic-134

PR #7941, opened 2023-12-19
Add a switch to get rid of QP_Solver dependency

sloriot:


sloriot:
Successfully tested in CGAL-6.0-Ic-134

PR #7953, opened 2024-01-02
Remove unused variable

sloriot:
The variable is not used as shadowed by the loop variable and the `maybe-unused` requires c++17

### DIFF
--- a/Kernel_23/test/Kernel_23/test_approximate_dihedral_angle_3.cpp
+++ b/Kernel_23/test/Kernel_23/test_approximate_dihedral_angle_3.cpp
@@ -53,7 +53,6 @@ int main() {
 
   Point_3 a = {0, 0, 0};
   Point_3 b = {0, -1, 0}; // ab is oriented so that it sees the plan xz positively.
-  // c can be any point in the half-plane xy, with x>0
 
   const query queries[] = {
     { {  1, 0,  0},    0.},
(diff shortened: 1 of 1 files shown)


Kernel_23/test/Kernel_23/test_approximate_dihedral_angle_3.cpp
```diff
@@ -53,7 +53,6 @@ int main() {
 
   Point_3 a = {0, 0, 0};
   Point_3 b = {0, -1, 0}; // ab is oriented so that it sees the plan xz positively.
-  [[maybe_unused]] Point_3 c = {1, 0, 0};
   // c can be any point in the half-plane xy, with x>0
```
albert-github, line 56: Shouldn't comment be removed as well (as `c` is a bit hanging in thin air now)?

sloriot: c is defined below which is why I left the comment.

albert-github: Sure it is below and I saw it as well, but for the quick reader this is not clear.
Maybe a reference to the variable `c` in the loop is in place here.
It is good that `c` has been removed as it was hiding the real one.

sloriot:
Successfully tested in CGAL-6.0-Ic-142

PR #7974, opened 2024-01-11
PMP/soup_autorefinement.cpp: Make sure we can read the file before continuing 

sloriot:


sloriot:
Succesfully tested in CGAL-6.0-Ic-154

Issue #7973, opened 2024-01-10
Segment_Delaunay_graph_2 draw_dual_edge linearizes Parabola_segment_2


Yvee1:
The `draw_dual` function of `Segment_Delaunay_graph_2` provides a way to draw the dual segment Voronoi diagram. It passes lines, segments, and rays to a stream provided by the caller such that they can provide their own drawing implementation. However, the parabolic segments present in such a Voronoi diagram are linearized with a coarse step size of 2, which cannot be customized. In line with the goal of providing the caller flexibility on how to draw the geometric primitives, it would be better if the parabolic segment was not linearized, but instead passed directly, such that the caller can, for example, use a quadratic Bézier curve to draw it. 

A related issue is that the start and end point of a parabolic segment are protected members of the `Parabola_segment_2` class, so cannot be accessed if it would be passed to a caller. Let me know if I should make a separate issue for this.

PR #7976, opened 2024-01-11
Ease drawing of SDG duals

afabri:
## Summary of Changes

Issue #7973 mentions improvements to do on the undocumented class `Parabola_segment_2`.  To get started we make it a struct. Mael will add an example with some better drawing implementation.

## Release Management

* Affected package(s):  Segment_Delaunay_graph_2

### DIFF
--- a/Apollonius_graph_2/include/CGAL/Parabola_segment_2.h
+++ b/Apollonius_graph_2/include/CGAL/Parabola_segment_2.h
@@ -27,7 +27,7 @@ namespace Qt {
 }
 
 template < class Gt >
-class Parabola_segment_2 : public Parabola_2< Gt >
+struct Parabola_segment_2 : public Parabola_2< Gt >
 {
   typedef CGAL::Parabola_2<Gt>            Base;
   typedef typename Base::Site_2           Site_2;
@@ -39,10 +39,10 @@ class Parabola_segment_2 : public Parabola_2< Gt >
   using Base::t;
   using Base::f;
 
-protected:
+
   Point_2 p1, p2;
 
-public:
+
   Parabola_segment_2() : Parabola_2< Gt >() {}
 
   template<class ApolloniusSite>
--- a/Segment_Delaunay_graph_2/examples/Segment_Delaunay_graph_2/CMakeLists.txt
+++ b/Segment_Delaunay_graph_2/examples/Segment_Delaunay_graph_2/CMakeLists.txt
@@ -4,13 +4,19 @@
 cmake_minimum_required(VERSION 3.1...3.23)
 project(Segment_Delaunay_graph_2_Examples)
 
-find_package(CGAL REQUIRED)
+find_package(CGAL REQUIRED OPTIONAL_COMPONENTS Core)
 
-# create a target per cppfile
-file(
-  GLOB cppfiles
-  RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
-  ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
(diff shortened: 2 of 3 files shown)


MaelRL:
@Yvee1 I have added in [958eabe](https://github.com/CGAL/cgal/pull/7976/commits/958eabea9ff7281cc4bd654dafd6c2c87f99bb1b) a comprehensive example on how to generate a cropped dual of a generic segment Delaunay graph, using both an inexact kernel during the computation (for speed) and an exact kernel with exact square roots in the construction of the dual (for precision of the positions).

Here is some example:
Screenshot from 2024-01-11 12-10-12

Let me know if you have any question.

gdamiand:
@MaelRL The pull request #6967 changes many things about drawing functions. It is maybe interesting to base this pr on the new version?

MaelRL:
It doesn't draw per se or use basic viewers; it's more like "print": it only creates a list of segments that can be then read by the polyhedron demo.

Yvee1:
Thanks for the example. It is not clear to me though why you would want to use an exact kernel for drawing, could you elaborate on that? Also you seem to still discretize the parabolic segments.

In case it is of any help to others, and to make my issue clearer: my temporary solution was to rewrite the `draw_dual_edge` function so that it passes the parabolic segment to the stream as below.
```c++
template <class Stream,
          class K,
          class Gt  = CGAL::Segment_Delaunay_graph_traits_2<K>,
          class SDG = CGAL::Segment_Delaunay_graph_2<Gt>,
          class ST = CGAL::Segment_Delaunay_graph_storage_traits_2<Gt>,
          class D_S = CGAL::Triangulation_data_structure_2<CGAL::Segment_Delaunay_graph_vertex_base_2<ST>, CGAL::Segment_Delaunay_graph_face_base_2<Gt>>>
Stream& draw_dual_edge(const SDG& dg, typename D_S::Edge e, Stream& str)
{
	typename Gt::Line_2    l;
	typename Gt::Segment_2 s;
	typename Gt::Ray_2     r;
	CGAL::Parabola_segment_2<Gt> ps;

	CGAL::Object o = dg.primal(e);

	if (CGAL::assign(l, o))   str << l;
	if (CGAL::assign(s, o))   str << s;
	if (CGAL::assign(r, o))   str << r;
	if (CGAL::assign(ps, o))  str << ps;

	return str;
}
```
Then I can handle it as follows.
```c++
VoronoiDrawer& VoronoiDrawer::operator<<(const CGAL::Parabola_segment_2<Gt>& p) {
	auto dir = p.line();
	auto focus = p.center();

	// Roundabout way to obtain start and end of parabolic segment because they are protected
	std::vector<Gt::Point_2> pts;
	p.generate_points(pts, Gt::FT(1000000));
	auto start = pts.front();
	auto end = pts.back();

	auto control = CGAL::circumcenter(focus, dir.projection(start), dir.projection(end));

	// Draw quadratic Bezier curve from start to end with control as control point

	return *this;
}
```

Yvee1:
Right, thank you for the explanation. I did not realize that the SDG_2 performs no constructions; though, it makes sense as a Delaunay graph of points does not. It's strange though that I get an assertion violation when construction a SDG_2 with inexact constructions but not with exact constructions (https://github.com/CGAL/cgal/issues/7978), because if it would perform no constructions there should be no difference right?

sloriot:
Successfully tested in CGAL-6.0-Ic-156

afabri:
@MaelRL can one use the kerrnel without sqrt for  the construction and the kernel with sqrt only where it is needed?

PR #7927, opened 2023-12-12
Introduce Arr_observer<Arrangement_2>

efifogel:
## Summary of Changes

1. Renamed the template Arr_observer to Aos_observer
2. Added the nested type 'Observer' in the Aos_2 class template, which is an alias to Arr_observer<Aos_2>.
3. Finally, introduced Arr_observer<Arrangement_2>, which is an alias to Arr_observer<Aos_2>, where Arrangement_2 derives from Aos_2 (for cases where Arrangement_2 is not Aos_2).

The first 2 items eliminate the error-prone c-type casting that was used to define observers.
The 3rd item is for backward compatibility.

## Release Management

* Affected package(s): Arrangement_on_surface_2, Minkowski_sum_2, Visibility_2, and Envelope_3
* Issue(s) solved (if any): fix #7901
* Feature/Small Feature (if any):
* Link to compiled documentation (obligatory for small feature) [*wrong link name to be changed*](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership:

PR #8000, opened 2024-01-24
Do not skip whitespace while reading quotient

sloriot:
Fixes issue mentioned in https://github.com/CGAL/cgal/pull/7927

`sentry` also eats end-of-line and arrangement formatter has a skip-until-eol after reading vertex coordinates

We decide to not allow whitespaces between `num` `/` `dum`

### DIFF
--- a/Number_types/include/CGAL/Quotient.h
+++ b/Number_types/include/CGAL/Quotient.h
@@ -421,7 +421,6 @@ operator>>(std::istream& in, Quotient<NT>& r)
   NT num,den=1;
   in >> num;
   if(!in) return in;
-  std::istream::sentry s(in); // skip whitespace
   if(in.peek()!='/'){
           if(!in.good()){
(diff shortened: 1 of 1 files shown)


mglisse:
> We decide to not allow whitespaces between `num` `/` `dum`

IIUC, after the patch, you accept whitespace before num, before den, but not between num and '/'.
This seems to match what Gmpq (more generally read_float_or_quotient) does, and is less strict than gmpxx (which also refuses a space after '/'), so it shouldn't break much.
Ideally there would be a bit of testing and/or documentation to go with it, but I can understand if you don't think it is worth the trouble.

sloriot:
Actually it's doc with no space at all around '/'. I can force no space after if you think it makes sense.

mglisse:
> I can force no space after if you think it makes sense.

Bah, as long as it isn't causing problems, it can wait...

sloriot:

PR #8002, opened 2024-01-25
Core: Use Expr::is_zero() of AST

afabri:
## Summary of Changes

Accelerate comparison with zero

## Release Management

* Affected package(s):  Number_types

### DIFF
--- a/Number_types/include/CGAL/CORE_Expr.h
+++ b/Number_types/include/CGAL/CORE_Expr.h
@@ -115,6 +115,14 @@ template <> class Algebraic_structure_traits< CORE::Expr >
         };  */
     };
 
+    class Is_zero
+      : public CGAL::cpp98::unary_function< Type, bool > {
+      public:
+        bool operator()( const Type& x ) const {
+          return x.isZero();
+        }
+    };
+
 };
 
 template <> class Real_embeddable_traits< CORE::Expr >
--- a/Number_types/include/CGAL/CORE_Expr.h
+++ b/Number_types/include/CGAL/CORE_Expr.h
@@ -132,7 +132,7 @@ template <> class Algebraic_structure_traits< CORE::Expr >
           if((inf > 1) || (sup < 1)){
             return false;
           }
-          if((inf == 1) && (sup == 1)){
+          if(inf == sup){
             return true;
           }
           return x.cmp(Type::getOne()) == 0;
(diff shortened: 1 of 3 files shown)


Number_types/include/CGAL/CORE_Expr.h
```diff
@@ -115,6 +115,30 @@ template <> class Algebraic_structure_traits< CORE::Expr >
...
+    class Is_one
+      : public CGAL::cpp98::unary_function< Type, bool > {
+      public:
+        bool operator()( const Type& x ) const {
+          double inf, sup;
+          x.doubleInterval(inf,sup);
+          if((inf > 1) || (sup < 1)){
+            return false;
+          }
+          if((inf == 1) && (sup == 1)){
```
mglisse, line 135:
```suggestion
          if(inf == sup){
```

mglisse:
Is this implementation of Is_one really faster? From what I can see, doubleInterval itself should already be more costly than `==1`.

afabri:
When comparing to 1 that constructs an `Expr(1)`, and performs a subtraction. I wanted to avoid that.  
I am also wondering if one should not use `Lazy_exact_nt<Expr>`  instead of doing a filtering, but I said myself that `Expr`  is somehow the same, but didn't find the interval hardcoded.
And I am wondering if I first have to call `Expr::approx()` as done [here](https://github.com/CGAL/cgal/blob/96f698ca09b61b6ca7587d43b022a0db43519699/Number_types/include/CGAL/CORE_Expr.h#L167)

mglisse:
> When comparing to 1 that constructs an `Expr(1)`, and performs a subtraction. I wanted to avoid that.

https://github.com/CGAL/cgal/blob/96f698ca09b61b6ca7587d43b022a0db43519699/CGAL_Core/include/CGAL/CORE/Expr_impl.h#L84-L89

`doubleInterval` also constructs an Expr from a double and performs a subtraction...

> And I am wondering if I first have to call `Expr::approx()`

Looks like `doubleValue` already does it? Although the different precision is suspicious.
https://github.com/CGAL/cgal/blob/96f698ca09b61b6ca7587d43b022a0db43519699/CGAL_Core/include/CGAL/CORE/Expr.h#L252-L253

> but didn't find the interval hardcoded.

IIUC (but I know little about CORE) it stores an approximate value as a `Real` and a precision, not directly an interval.

sloriot:

lrineau:
@afabri @mglisse Can you please go on with the discussion? I think Andreas has to answer to Marc's last remarks.

sloriot:
Successfully tested in CGAL-6.0-Ic-192

PR #8003, opened 2024-01-26
Add a hook CGAL_hook_fix_ctest_dependencies

lrineau:
## Summary of Changes

Add a hook `CGAL_hook_fix_ctest_dependencies`.

This hook function detects dependencies between targets tests, and add proper dependencies between compilation tests of those targets.

## Release Management

* Affected package(s): Installation, Polyhedron/demo


lrineau:
Actually, this could be a bug fix for release branches as well, but the patch is quite heavy.

@sloriot Do you want to fix the CMake issues in the testsuite for release branches as well?

sloriot:
> Actually, this could be a bug fix for release branches as well, but the patch is quite heavy.
> 
> @sloriot Do you want to fix the CMake issues in the testsuite for release branches as well?

It does not seem necessary to me.

sloriot:
Windows does not seem to like this change. See 6.0-Ic-168

lrineau:
> Windows does not seem to like this change. See 6.0-Ic-168

Should be fixed by 3664fc8fdb24b99918a256b74c458f0465453121. I have not tested on Windows. I have only test on Linux that there was no regression.

sloriot:

PR #8004, opened 2024-01-26
Point_set_3: property uchar refers to the range 0..255 not the char type

### LINKED ISSUE #7996: planes_point_set_3.ply in wrong format
## Issue Details

region_growing_planes_on_point_set_3 creates a file named planes_point_set_3.ply. In this file the colors are stored in a wrong format. If you open the file in MeshLab you will get an error message.

## Source Code

region_growing_planes_on_point_set_3.cpp

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Windows 64 bit
* Compiler: MS Visual C++ 2022
* Release or debug mode: both
* Specific flags used (if any): -
* CGAL version: 5.6
* Boost version: the latest
* Other libraries versions if used (Eigen, TBB, etc.): the latest

afabri:
## Summary of Changes

In the [ply file format](https://paulbourke.net/dataformats/ply/) one specifies in the header the type of properties. For a color one typically writes `property uchar red`, and then later writes a number between 0 and 255. We wrote a single char which is a mistake and was reported in issue #7996.   Note that even the CGAL ply reader cannot read it. And there is a [comment](https://github.com/CGAL/cgal/blob/96f698ca09b61b6ca7587d43b022a0db43519699/Stream_support/include/CGAL/IO/PLY/PLY_reader.h#L157) in the reader which explains that it reads an integer string and not a single char.

## Release Management

* Issue(s) solved (if any): fix #7996

### DIFF
--- a/Point_set_3/test/Point_set_3/CMakeLists.txt
+++ b/Point_set_3/test/Point_set_3/CMakeLists.txt
@@ -10,6 +10,7 @@ find_package(CGAL REQUIRED)
 create_single_source_cgal_program("point_set_test.cpp")
 create_single_source_cgal_program("point_set_test_join.cpp")
 create_single_source_cgal_program("test_deprecated_io_ps.cpp")
+create_single_source_cgal_program("issue7996.cpp")
 
 #Use LAS
 #disable if MSVC 2017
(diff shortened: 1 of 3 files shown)


sloriot:

Issue #8009, opened 2024-01-29
No matching function for call to `midpoint(const Point_3&, const Point_3&)` on `bounded_error_Hausdorff_distance`


OpenByteDev:
## Issue Details

I get the following build errors when trying to use `CGAL::Polygon_mesh_processing::bounded_error_Hausdorff_distance`:

```
[main] Building folder: terrain-builder 
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /mnt/[...]/build --config Debug --target unittests_terrainmerger -j 18 --
[build] [  0%] Built target CLI11
[build] [  1%] Built target fmt
[build] [  8%] Built target glm
[build] [ 10%] Built target spdlog
[build] [ 11%] Built target radix
[build] [ 93%] Built target Catch2
[build] [ 93%] Building CXX object src/CMakeFiles/terrainmergerlib.dir/terrainmerger/simplify.cpp.o
[build] In file included from /mnt/[...]/src/terrainmerger/simplify.cpp:4:
[...]

PR #8010, opened 2024-01-31
Change inclusion order to check dependencies

afabri:
## Summary of Changes

Issue #8009 revealed that some header files do not `#include` all what they need.  We do not see that in the testsuite, as we "usually" first include a kernel and then only header files of higher level functions.  This PR includes the  high level first and then only a kernel. We start with the _Polygon Mesh Processing_ package, but will extend this to the other packages.

This PR should fix the reported issue, but changes in the testsuite itself should reveal other similar issues. 

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any):  tbd #8009

### DIFF
--- a/Polygon_mesh_processing/examples/Polygon_mesh_processing/corefinement_parallel_union_meshes.cpp
+++ b/Polygon_mesh_processing/examples/Polygon_mesh_processing/corefinement_parallel_union_meshes.cpp
@@ -10,6 +10,7 @@
 
 typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
 typedef Kernel::Point_3 Point_3;
+typedef Kernel::Aff_transformation_3 Aff_transformation_3;
 typedef CGAL::Surface_mesh<Point_3> Mesh;
 
 namespace PMP = CGAL::Polygon_mesh_processing;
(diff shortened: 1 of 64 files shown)


Polygon_mesh_processing/examples/Polygon_mesh_processing/corefinement_parallel_union_meshes.cpp
```diff
@@ -10,6 +10,7 @@
 
 typedef CGAL::Exact_predicates_inexact_constructions_kernel Kernel;
 typedef Kernel::Point_3 Point_3;
+typedef Kernel::Aff_transformation_3 Aff_transformation_3;
```
sloriot, line 13: why changing it in an example?

afabri: It is not about inclusion order. I only  saw that we mix   `Class<Kernel>`  and `Kernel::Class`  which makes no sense to me even if they are equivalent,

sloriot: my point is why replacing in "user code" a free function call by a functor call.

PR #8011, opened 2024-01-31
CGAL_error_msg -> CGAL_NEF_TRACEN to allow convex decomposition on particular input

aothms:
## Summary of Changes

Apologies for the programming by accident here as I don't understand the majority of the code here. However, I propose here changing one particular occurrence of `CGAL_error_msg` to `CGAL_NEF_TRACEN `, which may be warranted giving the surrounding lines. `CGAL_error_msg` gives an assertion error and terminates program (at least on my configuration) however the code seems to be allowed to fall through since a null handle is returned on the next line which is actually handled at the call site.

### Data

[nef.txt](https://github.com/CGAL/cgal/files/14110073/nef.txt)

### Steps to reproduce

```C++
int main(int, char**) {
	std::ifstream ifs("nef.txt");
	CGAL::Nef_polyhedron_3<CGAL::Epeck> nef;
	ifs >> nef;
	CGAL::convex_decomposition_3(nef);
}
```

- Without proposed fix throws assertion violation
- With proposed fix runs (seemlingly) normal

### Output

afbeelding

(the apparent self intersections, I think are rendering artefacts from the z offset of edge rendering in blender).

## Release Management

* Affected package(s): Convex Decomposition of Polyhedra / 3D Boolean Operations on Nef Polyhedra
* Issue(s) solved (if any): - 
* Link to compiled documentation: -
* License and copyright ownership: -

### DIFF
--- a/Nef_3/include/CGAL/Nef_3/SNC_const_decorator.h
+++ b/Nef_3/include/CGAL/Nef_3/SNC_const_decorator.h
@@ -273,7 +273,7 @@ public:
 
     SFace_handle sf;
     if(!CGAL::assign(sf,o)) {
-      CGAL_error_msg( "it is not possible to decide which one is a visible facet (if any)");
+      CGAL_NEF_TRACEN( "it is not possible to decide which one is a visible facet (if any)");
       return Halffacet_handle();
     }
 


sloriot:
Looking a bit in the debugger it seems that you are most probably affected by the issue mentioned in the comment here,
which is somehow a good thing since we have a case to debug it.
```
    // The ray here has an special property since it is shooted from the lowest
    // vertex in a shell, so it would be expected that the ray goes along the
    // interior of a volume before it hits a 2-skeleton element.
    // Unfortunately, it seems to be possible that several shells are incident
    // to this lowest vertex, and in consequence, the ray could also go along
    // an edge or a facet belonging to a different shell.
    // This fact invalidates the precondition of the get_visible_facet method,
    // (the ray must pierce the local view of the hit object in a sface).
    // This situation has not been analyzed and has to be verified. Granados.
```

PR #8014, opened 2024-02-02
Warnings in documentation when running xmllint

albert-github:
When running xmllint over the documentation output we get a number of waning's like:

```
save_Periodic_2_triangulation_2/class_c_g_a_l_1_1_periodic__2__triangulation__2.html:1420: parser error : Opening and ending tag mismatch: dd line 1420 and div
"This is an advanced function.">is_triangulation_in_1_sheet()</a></code>  </div>
                                                                               ^
```
they are a consequence of the fact that a a command is not properly ended in this case we have a `\pre` command and its documentation states:
> The `\pre` command ends when a blank line or some other sectioning command is encountered.

and in this case the command is followed by the `\cgalAdvancedEnd` command that does not properly end the `\pre` command. The same is valid in the other cases.


sloriot:

PR #8015, opened 2024-02-03
Incorrect layout bisector precondition

albert-github:
On the page Kernel_23/group__bisector__grp.html we see for `bisector` with the `Point_3` type arguments in the precondition:
```
‘p != q’
```
instead of
```
p != q
```
due to the fact that there is not a backtick but a single quote as closing part


sloriot:

PR #8016, opened 2024-02-05
Space before equal sign in ALIASES

albert-github:
There is a space before the `=`-sign and in the newer version of doxygen this is not accepted anymore. (probably an unknown side effect).


sloriot:

PR #8022, opened 2024-02-07
Don't set the point again when reinserting the same point

mglisse:
## Summary of Changes

In Triangulation and Delaunay_triangulation (but not Regular_triangulation with equal weights), if we insert a point and locate notices that there is already a vertex for this point, we call `set_point` to set this point again. I don't understand why we do that. In most cases it should be useless but harmless. To notice a difference, we would need a `Point_d` type for which 2 objects may represent the same position but differ in some way (probably include extra data like a color). But even in that case, it isn't obvious to me that using the color of the last copy of this point that gets inserted is better than the color of the first copy. If someone was using this mechanism to store the index of points, having a vertex that changes what point it represents in the middle of the construction could break some uses.

It would be nice if someone else could check the logic, not just trust me.

(I don't have any particular use for this, I only noticed it while copy-pasting `insert` so I could customize it)

## Release Management

### DIFF
--- a/Installation/CHANGES.md
+++ b/Installation/CHANGES.md
@@ -76,6 +76,12 @@ Release date: October 2023
 -   Added three functions `vertices()` to the class `Triangulation_3`.
     Each of them returns an array containing the vertices of the given triangulation simplex.
 
+### [dD Triangulations](https://doc.cgal.org/6.0/Manual/packages.html#PkgTriangulations)
+-   **Breaking change**: Inserting multiple unweighted points in the same
+    position now keeps the first one, instead of switching to the latest. This
+    only affects custom point types where not all points in the same position
+    are equivalent.
+
 [Release 5.6](https://github.com/CGAL/cgal/releases/tag/v5.6)
 -----------
 
--- a/Triangulation/doc/Triangulation/CGAL/Delaunay_triangulation.h
+++ b/Triangulation/doc/Triangulation/CGAL/Delaunay_triangulation.h
@@ -125,8 +125,8 @@ value of `lt`:
 <DT><B>`OUTSIDE_AFFINE_HULL`</B><DD> Point
 `p` is inserted so as to increase the current dimension of the Delaunay
 triangulation.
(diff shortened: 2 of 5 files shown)


afabri:
What does the documentation say?  How is that handled in 2D and 3D?

lrineau (changes requested):
The previous behavior is the one that is documented:

- in [Triangulation::insert](https://doc.cgal.org/5.6/Triangulation/classCGAL_1_1Triangulation.html#ac98183b1c784328d4280ec9870398831)
  > `ON_VERTEX`
  > The point of the `pVertex` described by `f` is set to `p`.
- in [Delaunay_triangulation::insert](https://doc.cgal.org/latest/Triangulation/classCGAL_1_1Delaunay__triangulation.html#a79e606a88db53a80ee4f2c29eeb690c0)
  > `ON_VERTEX`
  > The position of the vertex `v` described by `f` is set to `p`. `v` is returned.

That behavior seems odd. If one want to change it, first check if that was mentioned during the review, and then eventually change the documentation and publish a breaking change.

mglisse:
> How is that handled in 2D and 3D?

In 3d and 2d:
https://doc.cgal.org/latest/Triangulation_3/classCGAL_1_1Triangulation__3.html#ad8d7a1aa3b310ba8d86ede726746fcb3
https://doc.cgal.org/latest/Triangulation_2/classCGAL_1_1Triangulation__2.html#a1025cd7e7226ccb44d82f0fb1d63ad4e
> If point p coincides with an already existing vertex, this vertex is returned and the triangulation remains unchanged.

I am trying to change dD so it behaves likes 2D/3D.

mglisse:
> The previous behavior is the one that is documented:

Oops, I lost the reflex to look for the documentation in a completely different place than the code...

> That behavior seems odd. If one want to change it, first check if that was mentioned during the review,

I searched for "insert" in the 3 rounds of reviews of Triangulation_d and didn't find anything related.

> and then eventually change the documentation and publish a breaking change.

I set the PR to draft in the meantime.

lrineau (approved):
Looks good.

PR #8023, opened 2024-02-08
Autorefinement: insert points in edge to avoid filter failures

sloriot:
Collect intersection points on edges and in faces to use CDT dedicated functions.
Also add structural filtering with projection traits and restore "classical" projection traits for testing.

### DIFF
--- a/Intersections_3/include/CGAL/Intersections_3/internal/Triangle_3_Triangle_3_intersection.h
+++ b/Intersections_3/include/CGAL/Intersections_3/internal/Triangle_3_Triangle_3_intersection.h
@@ -39,11 +39,15 @@ struct Point_on_triangle
 {
   // triangle points are not stored in this class but are expected
   // to always be passed in the same order. For a triangle pqr,
-  // edge 0 is pq, edge 1 qr and edge 2 rp. Point 0 is p, 1 is q and 2 is r.
+  // edge 1 is pq, edge 2 qr and edge 3 rp. Point -1 is p, -2 is q and -3 is r.
   //
-  // (id, -1) point on t1
-  // (-1, id) point on t2
-  // (id1, id2) intersection of edges
+  //  (0,-) vertex of t2 inside t1
+  //  (-,-) common point of t1 and t2
+  //  (-,0) vertex of t1 inside t2
+  //  (-,+) vertex of t1 on an edge of t2
+  //  (+,+) edge of t1 intersecting edge of t2
+  //  (+,-) vertex of t2 on an edge of t1
+  //
   std::pair<int, int> t1_t2_ids;
   boost::container::flat_set<int> extra_t1; // store other ids of edges containing the point
   typename Kernel::FT alpha; //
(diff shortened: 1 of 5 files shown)


Intersections_3/include/CGAL/Intersections_3/internal/Triangle_3_Triangle_3_intersection.h
```diff
@@ -39,11 +39,15 @@ struct Point_on_triangle
...
   // to always be passed in the same order. For a triangle pqr,
-  // edge 0 is pq, edge 1 qr and edge 2 rp. Point 0 is p, 1 is q and 2 is r.
+  // edge 1 is pq, edge 2 qr and edge 3 rp. Point -1 is p, -2 is q and -3 is r.
+  //
+  //  (0,-) vertex of t2 inside t1
+  //  (-,-) common point of t1 and t2
+  //  (-,0) vertex of t1 inside t2
+  //  (-,+) vertex of t1 on an edge of t2
+  //  (+,+) edge of t1 intersecting edge of t2
+  //  (+,-) vertex of t2 on an edge of t1
```
albert-github, line 49: Wouldn't it be  better to have these "magical" numbers as as elements of and `enum` or `enum class` ?

sloriot:
Successfully tested in CGA-6.0-Ic-173 (but TODOs remaining)

sloriot:
There are some runtime errors in CGAL-6.0-Ic-176

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/autorefinement.h
```diff
@@ -602,9 +680,9 @@ void generate_subtriangles(std::size_t ti,
               {
                 CGAL_AUTOREF_COUNTER_INSTRUCTION(counter.timer6.start();)
                 typename EK::Point_3 pt = typename EK::Construct_planes_intersection_point_3()(
-                  triangles[in_triangle_ids[i]][0], triangles[in_triangle_ids[i]][1],triangles[in_triangle_ids[i]][2],
-                  triangles[in_triangle_ids[j]][0], triangles[in_triangle_ids[j]][1],triangles[in_triangle_ids[j]][2],
-                  triangles[ti][0], triangles[ti][1],triangles[ti][2]);
+                  all_triangle_data[in_triangle_ids[i]].points[0], all_triangle_data[in_triangle_ids[i]].points[1],all_triangle_data[in_triangle_ids[i]].points[2],
```
afabri, line 683: With VC2022 and the call `test_autorefinement.exe data-autoref/four_cubes.off 12 94 184 1 78 0 106 352`   I get a segfault in the parallel execution.   My guess (only based on gut feeling) is  that while a thread is in this functor another thread executes the exact construction of one of the input points

sloriot: I can produce an error locally on the same case (need several runs). Thanks for the pointer. Investigating...

sloriot:
Successfully tested in CGAL-6.0-Ic-181

PR #8026, opened 2024-02-10
Call 3d static filters from NewKernel_d

mglisse:
## Summary of Changes

NewKernel_d already had wrappers so it could call some 2d static filters. I blindly copy-pasted so it could also handle 3d.
IIRC, those static filters were just a proof of concept and could do with more work (I don't think they are used by Epeck_d for instance), but I don't have a lot of time for it, and this simple patch (minimal effort) immediately gave a nice speed-up when building a Delaunay triangulation with `Epick_d<Dimension_tag<3>>` (enough that it isn't worth having a special case for 3d using Delaunay_triangulation_3 in my application). I might add the weighted versions in another PR later...

The main issue with all those filters is that it isn't clear if the testsuite really tests all the layers, but for now it should be ok since Epeck_d does not use static filters.

## Release Management


sloriot:

PR #8035, opened 2024-02-15
Using std::optional for Property_container::get<T>

soesau:
## Summary of Changes
Switching from `std::pair<Property_map<T>, bool>` to `std::optional` in `Property_container::get<T>`

Introducing `Pair_optional_adaptor` for backward compatibility which extends `std::optional<T>` to interface of `std::pair`

using `Pair_optional_adaptor` for `Surface_mesh` and `Point_set_3`

## Release Management

* Affected package(s): Point_set_3, Surface_mesh, STL_Extension

### DIFF
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -243,9 +243,9 @@ public:
   Point_set_3& operator= (const Point_set_3& ps)
   {
     m_base = ps.m_base;
-    m_indices = this->property_map<Index> ("index").first;
-    m_points = this->property_map<Point> ("point").first;
-    m_normals = this->property_map<Vector> ("normal").first;
+    m_indices = this->property_map<Index> ("index").value();
+    m_points = this->property_map<Point> ("point").value();
+    m_normals = this->property_map<Vector> ("normal").value();
     m_nb_removed = ps.m_nb_removed;
     return *this;
   }
@@ -255,9 +255,9 @@ public:
   Point_set_3 (const Point_set_3& ps)
   {
     m_base = ps.m_base;
-    m_indices = this->property_map<Index> ("index").first;
-    m_points = this->property_map<Point> ("point").first;
-    m_normals = this->property_map<Vector> ("normal").first;
+    m_indices = this->property_map<Index> ("index").value();
+    m_points = this->property_map<Point> ("point").value();
+    m_normals = this->property_map<Vector> ("normal").value();
     m_nb_removed = ps.m_nb_removed;
   }
   /// \endcond
@@ -816,9 +816,7 @@ public:
 
     \param name Name of the property.
 
-    \return Returns a pair containing: the specified property map and a
-    Boolean set to `true` or an empty property map and a Boolean set
-    to `false` (if the property was not found).
+    \return Returns an optional property map (empty if the property was not found).
   */
   template <class T>
   std::optional<Property_map<T>>
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -359,8 +359,8 @@ public:
     other.resize(m_base.size());
     other.transfer(m_base);
     m_base.swap(other);
-    m_indices = *this->property_map<Index>("index");
-    m_points = *this->property_map<Point>("point");
+    m_indices = this->property_map<Index>("index").value();
+    m_points = this->property_map<Point>("point").value();
   }
 
   /*!
--- a/Point_set_3/include/CGAL/Point_set_3.h
+++ b/Point_set_3/include/CGAL/Point_set_3.h
@@ -359,8 +359,8 @@ public:
     other.resize(m_base.size());
     other.transfer(m_base);
     m_base.swap(other);
-    m_indices = this->property_map<Index>("index").first;
-    m_points = this->property_map<Point>("point").first;
+    m_indices = *this->property_map<Index>("index");
+    m_points = *this->property_map<Point>("point");
   }
 
   /*!
@@ -784,9 +784,9 @@ public:
   template <typename T>
   bool has_property_map (const std::string& name) const
   {
-    std::pair<Property_map<T>, bool>
+    std::optional<Property_map<T>>
       pm = m_base.template get<T> (name);
-    return pm.second;
+    return pm.has_value();
   }
 
   /*!
@@ -806,10 +806,7 @@ public:
   std::pair<Property_map<T>, bool>
   add_property_map (const std::string& name, const T t=T())
   {
-    Property_map<T> pm;
-    bool added = false;
-    std::tie (pm, added) = m_base.template add<T> (name, t);
-    return std::make_pair (pm, added);
+    return m_base.template add<T>(name, t);
   }
 
   /*!
@@ -824,13 +821,10 @@ public:
     to `false` (if the property was not found).
   */
   template <class T>
-  std::pair<Property_map<T>,bool>
+  std::optional<Property_map<T>>
   property_map (const std::string& name) const
   {
-    Property_map<T> pm;
-    bool okay = false;
-    std::tie (pm, okay) = m_base.template get<T>(name);
-    return std::make_pair (pm, okay);
+    return m_base.template get<T>(name);
   }
 
   /*!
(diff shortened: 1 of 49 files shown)


STL_Extension/include/CGAL/Pair_optional_adaptor.h
```diff
@@ -0,0 +1,92 @@
...
+
+  Pair_optional_adaptor(std::pair<T, bool>& p) : std::optional<T>(p.second ? p.first : std::optional<T>()), first(t_storage.t), second(p.second), t_storage(p.first) {}
+
+#ifndef CGAL_NO_DEPRECATED_CODE
+  CGAL_DEPRECATED_MSG("you are using the deprecated API, please update your code")
+  operator std::pair<T, bool>() {
+    return std::pair<T, bool>(first, second);
+  }
+
+  CGAL_DEPRECATED_MSG("you are using the deprecated API, please update your code")
```
sloriot, line 37: why tuples?

Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
```diff
@@ -2078,7 +2079,7 @@ class Surface_mesh
     /// In case it does not exist the Boolean is `false` and the behavior of
     /// the property map is undefined.
     template <class I, class T>
-    std::pair<Property_map<I, T>,bool> property_map(const std::string& name) const
+    Pair_optional_adaptor<Property_map<I, T>> property_map(const std::string& name) const
```
sloriot, line 2082: the doc should indicate that the return type is an optional

Point_set_3/include/CGAL/Point_set_3.h
```diff
@@ -824,13 +824,10 @@ class Point_set_3
     to `false` (if the property was not found).
   */
   template <class T>
-  std::pair<Property_map<T>,bool>
+  Pair_optional_adaptor<Property_map<T>>
```
sloriot, line 827: the doc should mention that the output is an optional

STL_Extension/include/CGAL/Pair_optional_adaptor.h
```diff
@@ -0,0 +1,92 @@
...
+    operator std::tuple<std::_Ignore const&, bool&>() {
+    return std::tuple<std::_Ignore const&, bool&>(std::ignore, second);
+  }
+
+  CGAL_DEPRECATED_MSG("you are using the deprecated API, please update your code")
+    operator std::tuple<std::_Ignore const&, std::_Ignore const&>() {
+    return std::tuple<std::_Ignore const&, std::_Ignore const&>(std::ignore, std::ignore);
```
lrineau, line 74: `std::_Ignore` does not exist! Use `decltype(std::ignore)` instead.

lrineau (changes requested):
This class `Pair_optional_adaptor<T>` has four data members:
  - its base class, of type `std::optional<T>`,
  - `T &first` and `bool second` for the compatibility with `std::pair<T, bool>`,
  - and `t_storage` (I am not sure why it is an union instead of a plain `T`.

That makes that class source-incompatible with:

```c++

auto [pmap, ok] = point.get<Foo>("bar");
```

See live on compiler explorer: https://godbolt.org/z/3xY6qzPTq

lrineau:
I am in favor of a real breaking change:

- change the API of `CGAL::Property_container` (undocumented anyway),
- In `Surface_mesh.h`, add:

  - `vertex_property_map`,
  - `halfedge_property_map`,
  - `edge_property_map`, and
  - `face_property_map`,

  with the new API, and keep the old `property_map`,
- a breaking change in `Point_set_3.h`.

STL_Extension/include/CGAL/Pair_optional_adaptor.h
soesau: Otherwise, it doesn't work with boost::tie or std::tie as can be found in some examples and tests.

soesau:
>     * and `t_storage` (I am not sure why it is an union instead of a plain `T`).

I used the union originally to avoid construction of T in case there is no default constructor. Sebastien mentioned already that T will be some lightweight object anyway that has to be default constructible for being put into a pair.

So, I'll remove the union and switch to a plain T.

sloriot:
We decided in a live meeting to break the compatibility and use std::optional directly.

sloriot:
@soesau  Note that this PR is a release blocker and it should be updated following the discussion we had with @lrineau  and @afabri

lrineau:
@sloriot, in a message dated Feb 2024:
> We decided during meeting to break the compatibility and use std::optional directly.

@sloriot, yesterday:
> @soesau Note that this PR is a release blocker and it should be updated following the discussion we had with @lrineau and @afabri

@sloriot, why is it release blocker?

@soesau If this PR is really blocking the next release, then we should hurry. Do you need help on it?

sloriot:
The API change is possible thanks to std::optional that is introduced in CGAL APIs with CGAL 6.0, and Orthtree is already using it for its dynamic property system.

Surface_mesh/include/CGAL/Surface_mesh/IO/OFF.h
```diff
@@ -390,12 +390,10 @@ bool write_OFF_with_or_without_fcolors(std::ostream& os,
 
   const bool has_fcolors = !(is_default_parameter<CGAL_NP_CLASS, internal_np::face_color_map_t>::value);
 
-  typename Mesh::template Property_map<Face_index, Color> fcolors;
-  bool has_internal_fcolors;
-  std::tie(fcolors, has_internal_fcolors) = sm.template property_map<Face_index, CGAL::IO::Color>("f:color");
+  auto fcolors  = sm.template property_map<Face_index, CGAL::IO::Color>("f:color");
```
sloriot, line 393: not using `Color` produces a unused typedef for lines 387 and 431

Surface_mesh/include/CGAL/Surface_mesh/IO/PLY.h
```diff
@@ -463,9 +463,8 @@ bool fill_simplex_specific_header(std::ostream& os,
   bool okay = false;
   if(prop == "v:normal")
   {
-    Vector_map pmap;
-    std::tie(pmap, okay) = sm.template property_map<VIndex, Vector>(prop);
-    if(okay)
+    auto pmap = sm.template property_map<VIndex, Vector>(prop);
```
sloriot, line 466: `okay` is no longer used

Surface_mesh/include/CGAL/Surface_mesh/IO/PLY.h
```diff
@@ -644,25 +641,23 @@ void fill_header_impl(std::tuple<T,TN...>,
   bool okay = false;
   {
     typedef typename Surface_mesh<Point>::template Property_map<Simplex, T>   Pmap;
-    Pmap pmap;
-    std::tie(pmap, okay) = sm.template property_map<Simplex,T>(pname);
-    if(okay)
+    auto pmap  = sm.template property_map<Simplex,T>(pname);
```
sloriot, line 644: `okay` no longer used

Surface_mesh/include/CGAL/Surface_mesh/IO/PLY.h
```diff
@@ -520,16 +518,15 @@ bool fill_simplex_specific_header(std::ostream& os,
   bool okay = false;
   if(prop == "f:color")
   {
-    Fcolor_map pmap;
-    std::tie(pmap, okay) = sm.template property_map<FIndex, Color>(prop);
-    if(okay)
+    auto pmap = sm.template property_map<FIndex, Color>(prop);
```
sloriot, line 521: `okay` no longer used

Surface_mesh/include/CGAL/draw_surface_mesh.h
```diff
@@ -112,33 +112,35 @@ struct Graphics_scene_options_surface_mesh
...
     else
     { this->colored_edge=[](const SM &, edge_descriptor)->bool { return false; }; }
 
-    std::tie(fcolors, found)=
+    auto _fcolors =
         amesh.template property_map<face_descriptor, CGAL::IO::Color>("f:color");
-    if(found)
+    if(_fcolors.has_value())
     {
+      fcolors = _fcolors.value();
```
lrineau, line 143: There should be a consistent use of `auto` versus the full type. I prefer auto, by the way, because it will always avoid type casting.

lrineau (approved):

Surface_mesh/include/CGAL/draw_surface_mesh.h
lrineau: But that is not a reason to delay the merge of the PR.

sloriot:

lrineau (approved):

PR #8049, opened 2024-02-27
be compatible with non-C++17 compilers

sloriot:


sloriot:
Successfully tested in 6.0-Ic-184

PR #8060, opened 2024-03-06
Qt6: fix a bug with devicePixelRatio and picking

lrineau:
## Summary of Changes

The picking was no longer working reliably on my machine (Linux with Wayland). That was due to the handling of pixel ratio (Retina/HiDPI). On my machine I have a screen at 200% and another at 125%.

## Release Management

* Affected package(s): Polyhedron/demo


sloriot:
Successfully tested in CGAL-6.0-Ic-192

Issue #8075, opened 2024-03-16
Documentation in Triangulation package


albert-github:
In the file Triangulation/doc/Triangulation/Triangulation.txt around line 408 we have the construct:
```
hole is ``star shaped'' around 
```

and doxygen will transform this into (see Triangulation/index.html):
```
hole is `&lsquo;star shaped&rsquo;' around
```
resulting in:
![image](https://github.com/CGAL/cgal/assets/5223533/edf76952-67b5-4d04-9f38-5f00860fdfd1)

The question here is which type of quotes are intended here (as in my opinion the current output is not so nice) ?

mglisse:
The intent was probably "the same behavior as latex".

albert-github:
@mglisse thanks for the pointer so something like:

![image](https://github.com/CGAL/cgal/assets/5223533/4305f7e1-fafe-4eab-98eb-ae146203da50)

albert-github:
I've just pushed a proposed patch, pull request #8076

PR #8076, opened 2024-03-16
issue #8075 Documentation in Triangulation package

albert-github:
replace the, for doxygen and thus for the output, strange constructs with better constructs.


sloriot:
CGAL-6.0-Ic-198 output is correct.

PR #8087, opened 2024-03-19
Fix implicit gmp to double conversion failures

pentacular:

## Summary of Changes

Trivial fix for errors in compiling isotropic remeshing using gmp which fails to convert implicitly to double.

## Release Management

* Affected package(s):

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/internal/Isotropic_remeshing/remesh_impl.h
@@ -400,9 +400,9 @@ namespace internal {
       for(edge_descriptor e : edge_range)
       {
         const halfedge_descriptor he = halfedge(e, mesh_);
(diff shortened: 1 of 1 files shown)


janetournois (approved):

janetournois:
@sloriot would it make sense to add a test that checks that isotropic remeshing compiles with GMP?

sloriot:
do you think it would work to have double -> FT? From what I remember we said it would not be a good idea. Note also that with a specific vertex point map (on-the-fly conversion), this patch would be useless.

janetournois:
Another option for this PR would be to replace the definition of `Boost_bimap`.
For now we have
```
      typedef boost::bimap<
        boost::bimaps::set_of<halfedge_descriptor>,
        boost::bimaps::multiset_of<double, std::less<double> > >  Boost_bimap;
 ```
 hence the use of `CGAL::to_double()` in this PR.
 
 Maybe it would be better to have
 ```
      typedef boost::bimap<
        boost::bimaps::set_of<halfedge_descriptor>,
        boost::bimaps::multiset_of<FT, std::less<FT> > >  Boost_bimap;
 ```
 ?

pentacular:
I think this is probably obsolete -- shall I close it?

PR #8096, opened 2024-03-26
fix dependencies

lrineau:
Fix dependencies so that Triangulation_2 is not a dependency of most of CGAL.


sloriot:

PR #8103, opened 2024-03-27
Improve information on exact nt backend

sloriot:
Add a test checking what we expect is the reality and a function to print the name of the exact backend


sloriot:

lrineau (approved):

PR #8108, opened 2024-03-28
CMake: simplify, now that C++17 is required

lrineau:


sloriot:
Successfully tested in CGAL-6.0-Ic-21[56]

PR #8129, opened 2024-04-08
Make triangulate_faces deterministic

sloriot:
Triangulate quads the same way

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
@@ -549,6 +549,7 @@ public:
                               NamedParameters,
                               Def_Kernel>::type;
     using FT = typename Traits::FT;
+    using PID = typename std::iterator_traits<typename Polygon::const_iterator>::value_type;
 
     // Visitor
     using Visitor = typename internal_np::Lookup_named_param_def<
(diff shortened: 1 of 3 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/triangulate_faces.h
```diff
@@ -219,8 +219,26 @@ class Triangulate_polygon_mesh_modifier
 
       const FT p1p3 = cross_product(p2-p1, p3-p2) * cross_product(p0-p3, p1-p0);
       const FT p0p2 = cross_product(p1-p0, p1-p2) * cross_product(p3-p2, p3-p0);
-      halfedge_descriptor res = (p0p2>p1p3) ?  CGAL::Euler::split_face(v0, v2, pmesh)
-                                            :  CGAL::Euler::split_face(v1, v3, pmesh);
+
+      halfedge_descriptor res = boost::graph_traits<PolygonMesh>::null_halfedge();
+
+      if (p1p3!=p0p2)
```
lrineau, line 225: Maybe we should have a predicate. Because here we compare computed values, without an exact kernel.

afabri: How do you want to call it?    `Comparison_result  fct(const Point_& p0,..... p3)`

afabri: Also the code might profit from computing some differences only once.

sloriot: I don't think there is any need to use an exact predicate here. I don't think we could select an invalid diagonal.

PR #8131, opened 2024-04-10
Boolean Set Operations: Fix brief descriptions

afabri:
## Summary of Changes

Add a "\\" after "resp." so that doxygen is not mislead by the "." as end of a brief description.

## Release Management

* Affected package(s): Boolean_set_operations_2

### DIFF
--- a/Boolean_set_operations_2/doc/Boolean_set_operations_2/Boolean_set_operations_2.txt
+++ b/Boolean_set_operations_2/doc/Boolean_set_operations_2/Boolean_set_operations_2.txt
@@ -407,7 +407,7 @@ aforementioned free functions is actually realized by a set overriding
 member function.
 
 member functions of the `Polygon_set_2`
-(resp. `General_polygon_set_2`) that perform Boolean set-operations
+(respectively `General_polygon_set_2`) that perform Boolean set-operations
 come in two flavors: for example, `S.join(P, Q)` computes the union of
 \f$ P\f$ and \f$ Q\f$ and assigned the result to \f$ S\f$, while
 `S.join(P)` performs the operation \f$ S \longleftarrow S \cup P\f$.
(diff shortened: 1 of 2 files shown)


afabri:
It is fixed in the built documentation, e.g., for[ `join()`](https://cgal.github.io/8131/v0/Boolean_set_operations_2/group__boolean__join.html#gabb13a0ed958cda84ea0e4d2857c9d67a) to check that it works in principle.  In case someone (@sloriot , @efifogel )wants to check  it in detail I think it is easier to just read the diff in the "Files Changed" tab instead of going through all html pages.

sloriot:
I'd prefer using "respectively"

afabri:
@JacksonCampolattaro do you have an opinion on the usage of  _respectively_ or _resp._  [here](https://cgal.github.io/8131/v0/Boolean_set_operations_2/group__boolean__join.html#gabb13a0ed958cda84ea0e4d2857c9d67a)?    I am even wondering if it should only be used in the second parenthesis.

JacksonCampolattaro:
> @JacksonCampolattaro do you have an opinion on the usage of _respectively_ or _resp._ [here](https://cgal.github.io/8131/v0/Boolean_set_operations_2/group__boolean__join.html#gabb13a0ed958cda84ea0e4d2857c9d67a)? I am even wondering if it should only be used in the second parenthesis.

I prefer using the full word; the sentence is already pretty dense, so abbreviation could make it less clear.

I'm not used to seeing respecively used in the middle of a sentence; I don't think it's wrong, but reversing the phrasing would be more intuitive to me:

> If substituted with [CGAL::Tag_true](https://cgal.github.io/8131/v0/STL_Extension/group__PkgSTLExtensionUtilities.html#gab3e2296107b5d26c32c8183028a217f1) (default), pgn1 and pgn2 are bounded by polylines before the operation is performed, converting them into a general polygon and a general polygon with holes, respectively.

sloriot:
Successfully tested in CGAL-6.0-Ic-228

Issue #8134, opened 2024-04-10
Isotropic remeshing creates self intersections


ning-zelin:
Isotropic remeshing creates self intersections

## Issue Details 
My tests using the following data found that self-intersections were produced after remesh, even if I set `do_collapse(false)`.
## Source Code

```c++ 
if(PMP::does_self_intersect(mesh)){
  std::cout<<"input mesh self-intersect"<<std::endl;
}
PMP::isotropic_remeshing(faces(mesh), 0.3, mesh,
                         CGAL::parameters::number_of_iterations(10)
                                          .protect_constraints(true).do_collapse(false)); //i.e. protect border, here
if(PMP::does_self_intersect(mesh)){
  std::cout<<"out-put mesh self-intersect"<<std::endl;
}
 ```

* Operating system: Ubuntu20.04 
* Input mesh: [merged_3.zip](https://github.com/CGAL/cgal/files/14933487/merged_3.zip)
* CGAL 5.6.1

**Comment by sloriot** [comment]
@janetournois It seems to come from the projection step. The vertex 2459 is very close to a face in the input:
![image](https://github.com/CGAL/cgal/assets/1224632/87d58807-0b69-4c06-acc7-02fd7e33c603)
In the output, I guess it gets projected onto the face rather that the original vertex (probably because of the smoothing that moves it). I'm not sure we can do much in such a case. For sure adding that vertex as constrained vertex would fix the issue, but it's a "manual" fix.

As it seems the mesh was not combinatorially modified near that vertex, maybe the smoothing should only apply to modified regions?

**Comment by ning-zelin** [comment]
But it still has self-intersections after setting do_projection to false.
```c++
  PMP::isotropic_remeshing(faces(mesh), 0.3, mesh,
                           CGAL::parameters::number_of_iterations(10)
                                            .protect_constraints(true).do_project(false)); 
```

sloriot:
As I tried to say, I guess that the smoothing move the vertex across the face and the projection put it on the face.

sloriot:
You can try disabling the smoothing

ning-zelin:
> As I tried to say, I guess that the smoothing moves the vertex across the face and the projection put it on the face.

Thanks. I set `number_of_relaxation_steps(0)`, and the problem was solved.

ning-zelin:
> > As I tried to say, I guess that the smoothing moves the vertex across the face and the projection put it on the face.
> 
[...]

PR #8150, opened 2024-04-18
PMP::isotropic_remeshing() - add NP `allow move functor`

janetournois:
## Summary of Changes

The function `tangential_relaxation()` has a named parameter called `allow_move_functor` that enables the user to prevent and cancel some vertex moves.

With this PR, it is now possible to use the same parameter for the relaxation steps of `CGAL::Polygon_mesh_processing::isotropic_remeshing()`

## Release Management

* Affected package(s): PMP
* Issue(s) solved (if any): partially solves issue #8134

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
@@ -171,6 +171,16 @@ namespace Polygon_mesh_processing {
 *     \cgalParamDefault{`false`}
 *   \cgalParamNEnd
 *
+*   \cgalParamNBegin{allow_move_functor}
+*     \cgalParamDescription{A function object used to determinate if a vertex move should
+*                           be allowed or not during the relaxation step.}
+*     \cgalParamType{Unary functor that provides `bool operator()(vertex_descriptor v, Point_3 src, Point_3 tgt)`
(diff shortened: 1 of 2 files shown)


lrineau:
> To solve #8134, we can think of a smart functor that would prevent self-intersections from being introduced by the relaxation step

Should we close the issue #8134, how that this PR is merged, or leave it open? Because it seems the actual solution to #8134 is
  - this PR #8150,
  - plus something else (see the quote above).

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/remesh.h
```diff
@@ -170,6 +171,16 @@ namespace Polygon_mesh_processing {
 *     \cgalParamDefault{`false`}
 *   \cgalParamNEnd
 *
+*   \cgalParamNBegin{allow_move_functor}
+*     \cgalParamDescription{A function object used to determinate if a vertex move should
+*                           be allowed or not during the relaxation step.}
+*     \cgalParamType{Unary functor that provides `bool operator()(vertex_descriptor v, Point_3 src, Point_3 tgt)`
```
lrineau, line 177: Actually, the sentence itself contradicts the word "Unary" (that would take _one_ parameter). That is not a unary functor.

Oops

PR #8169, opened 2024-04-25
Add a way to collect input coplanar faces after corefinement

sloriot:
Add a method in the corefinement visitor to collect the list of output faces corresponding to coplanar input faces

### DIFF
--- a/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
+++ b/Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
@@ -119,6 +119,9 @@ public:
   /// The point has already been put in the vertex point map.
   void after_vertex_copy(vertex_descriptor v_src, const Triangle_mesh& tm_src,
                          vertex_descriptor v_tgt, const Triangle_mesh& tm_tgt);
+  /// called for each face `f` in the output mesh `tm` that belongs to the intersection of two coplanar faces of the input meshes.
+  void input_coplanar_face(face_descriptor f, TriangleMesh& tm);
+
 /// @}
 
 /// @name Functions used by corefine() for progress tracking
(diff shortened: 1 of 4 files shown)


sloriot:
Successfully tested in CGAL-6.0-Ic-245 (but doc still missing)

Polygon_mesh_processing/doc/Polygon_mesh_processing/Concepts/PMPCorefinementVisitor.h
```diff
@@ -119,6 +119,9 @@ class PMPCorefinementVisitor{
   /// The point has already been put in the vertex point map.
   void after_vertex_copy(vertex_descriptor v_src, const Triangle_mesh& tm_src,
                          vertex_descriptor v_tgt, const Triangle_mesh& tm_tgt);
+  /// called for each face `f` in the output mesh `tm` that belongs to the intersection of two coplanar faces of the input meshes.
+  void input_coplanar_face(face_descriptor f, TriangleMesh& tm);
```
MaelRL, line 123: The (singular) "face" + "input" is a bit misleading because at first sight it looks like it's a face of the input, whereas it's an output face.

Is it related with these other functions:
```cpp
  /// called before processing intersections between the `n` pairs of coplanar faces.
  void start_handling_intersection_of_coplanar_faces(std::size_t n);
  /// called each time a pair of coplanar faces is processed.
  void intersection_of_coplanar_faces_step() const;
  /// called after processing all intersections between coplanar faces.
  void end_handling_intersection_of_coplanar_faces() const;
```
?

If so, maybe:

```cpp
/// Called for each face `f` in the output mesh `tm` that *results from* the intersection of two coplanar faces in the input meshes.
void intersection_of_coplanar_faces(face_descriptor f, TriangleMesh& tm);
```

and move it to be with the other functions

sloriot: not really, as it is only in the output. I proposed something that seems better after taking your input into account.

MaelRL (approved):

sloriot:
Successfully tested in CGAL-6.1-Ic-119

PR #8173, opened 2024-04-29
[PSP3] Adapting to new LASTools version

### LINKED ISSUE #8111: LAStools support
LAStools has evolved since CGAL forked it in https://github.com/CGAL/LAStools. I think we can now try to use it directly.

- LAStools now supports Cmake natively
- Among other new features, LAStools now support the COPC extension, which enables embedding an octree section in the las/laz (a VLR) for efficient pruning of the point range given a query 3D bbox. 
- Version 240319 introduces a breaking change in the constructor of all `LASreader`s. They now expect a `LASreadOpener*` argument : https://github.com/LAStools/LAStools/commit/c3012b6e94ba7abc90236faa4899bf34d38b4dc8    See also this reported [issue](https://github.com/LAStools/LAStools/issues/193)
- Missing functions in the dll. See this [pull request](https://github.com/LAStools/LAStools/pull/194)

soesau:
## Summary of Changes

Switching to the official LAStools repository

## Release Management

* Affected package(s): PSP3, Point_set_3, Stream_support
* Issue(s) solved (if any): fix #8111 

-----

Edit, @lrineau, 2024/05/24: depends on https://github.com/LAStools/LAStools/pull/194, that has been merged.

### DIFF
--- a/Documentation/doc/Documentation/Third_party.txt
+++ b/Documentation/doc/Documentation/Third_party.txt
@@ -248,9 +248,8 @@ imported target `CGAL::LASLIB_support` provided in
 \laslib information can be obtained from
 <a href="https://lastools.github.io/">https://lastools.github.io/</a> and
 <a href="https://rapidlasso.de/product-overview/">https://rapidlasso.de/product-overview/</a>.
-\laslib is usually distributed along with LAStools: for simplicity, \cgal
-provides <a href="https://github.com/CGAL/LAStools">a fork with a
-CMake based install procedure</a>.
+\laslib is usually distributed along with LAStools. Current versions of \laslib provide CMake support.
+BUILD_SHARED_LIBS needs to be set to true on windows to create a dynamic linked library.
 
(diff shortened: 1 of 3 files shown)


lrineau:
TODO: document that the "last" version of LASTools is required, and precise since which version it will work.

sloriot:
This PR makes the `laslib` distributed by CGAL incompatible with CGAL-6.0 (see [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.0-Ic-252/Lab_Demo/TestReport_gimeno_ArchLinux-clang-CXX17-Release.gz) for example). Any chance to fix that ? The reason why it would be nice to have is that we would install a single version on test platforms and it would work both with master and the release branches.

sloriot:
Actually, if this PR target `5.5.x-branch` then we would be good with test plateforms being updated.

sloriot:
Successfully tested in CGAL-6.0-Ic-324

PR #8179, opened 2024-05-02
CORE: Accelerate writing to a stream if precision <= 17

afabri:
## Summary of Changes

Writing an `Expr`to as stream is slow. We make it faster by writing `Expr::doubleValue()` if the precision is smaller or equal to 17.

## Release Management

* Affected package(s): CORE


sloriot:
@afabri CGAL-6.0-Ic-238 includes your latest commit. You can have a look at the results and remove the last commit.

afabri:
@sloriot please retest so that we can merge

sloriot:

PR #8204, opened 2024-05-15
Generator: Add example

afabri:
## Summary of Changes

I several time wrote a 2D or 3D point generator. So let's just add one in the Generator package.
To get started it takes the number of points and the dimension and generates in a square/cube.

I did not add it to the User Manual. Shall I?

## Release Management

* Affected package(s): Generator


sloriot:

PR #8215, opened 2024-05-22
Surface_mesh:  Add exact_num_faces(const SM&)

### LINKED ISSUE #5720: edges(sm).size() takes linear time
## Issue Details

Would it make sense to enrich the iterator range class with a data member for storing the size in case we know it?  This would avoid to traverse the range in order to count the elements.  This is for example necessary for a ` Surface_mesh` as soon as  it has garbage (simplices only marked as deleted).

### LINKED ISSUE #8158: Extremely slow Face_count_stop_predicate
Hi, I just wanted to make you aware of a potential performance bug in `Surface_mesh_simplification`:

https://github.com/CGAL/cgal/blob/c4165fe5f9f97099bb91dc4e11708df014e9952f/Surface_mesh_simplification/include/CGAL/Surface_mesh_simplification/Policies/Edge_collapse/Face_count_stop_predicate.h#L41

I have the suspicion that `exact_num_faces()` is somehow O(faces) instead of constant-time reading an integer.

https://github.com/CGAL/cgal/blob/c4165fe5f9f97099bb91dc4e11708df014e9952f/BGL/include/CGAL/boost/graph/internal/helpers.h#L162-L167

I noticed that using `Face_count_stop_predicate` instead of `Edge_count_stop_predicate` makes simplification 10000x slower, and changing the line to

```c++
    const std::size_t current_face_count = profile.surface_mesh().number_of_faces();
```

Fixes it.

(There's the caveat that I'm using `Face_count_stop_predicate.h` from CGAL master (5.6.1+) with my local CGAL 5.5.2 where `Face_count_stop_predicate.h` didn't exist yet.)

Just in case this is a free 10000x speedup :)

afabri:
## Summary of Changes

Add a partial specialization for `CGAL::internal::exact_num_faces(const Surface_mesh&)`  to have an `O(1)` time.  This was pointed out as important for the `Face_count_stop_predicate`  in Issue #8158 

## Release Management

* Issue(s) solved (if any): fix #5720

### DIFF
--- a/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
+++ b/Surface_mesh/include/CGAL/Surface_mesh/Surface_mesh.h
@@ -26,6 +26,7 @@
 #include <CGAL/boost/graph/Euler_operations.h>
 #include <CGAL/boost/graph/iterator.h>
 #include <CGAL/boost/graph/named_params_helper.h>
+#include <CGAL/boost/graph/internal/helpers.h>
 #include <CGAL/Named_function_parameters.h>
 #include <CGAL/circulator.h>
 #include <CGAL/Handle_hash_function.h>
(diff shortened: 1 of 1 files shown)


sloriot:
why only one and not the four?

lrineau:
TODO: that new function overload needs a test.

sloriot:

Issue #8192, opened 2024-05-07
CMake find_package with a minimal CGAL version number does not detect properly 6.X CGAL version


VincentRouvreau:
## Issue Details

CMake `find_package(CGAL 5.1.0)` (with a minimal cgal version number) does not seem to behave as expected with master version.

## Source Code

Here is a CMake file to reproduce the behaviour:
```cmake
cmake_minimum_required(VERSION 3.8)
project(minimal_cgal_version)

find_package(CGAL 5.1.0)

if (TARGET CGAL::CGAL)
  message("++ CGAL version: ${CGAL_VERSION}. Includes found in ${CGAL_INCLUDE_DIRS}")
endif ()
```

With CGAL 5.1.5, it works fine:
```txt
++ CGAL version: 5.1.5. Includes found in /home/gailuron/workspace/contrib/CGAL-5.1.5/include
```
(tested with 5.5.3 also).

But, with 6.0-I-207 (master from some days ago):
```txt
CMake Warning at src/cmake/modules/GUDHI_third_party_libraries.cmake:30 (find_package):
  Could not find a configuration file for package "CGAL" that is compatible
  with requested version "5.1.0".

  The following configuration files were considered but not accepted:

    /home/gailuron/workspace/contrib/CGAL-6.0-I-207/build/CGALConfig.cmake, version: 6.0

Call Stack (most recent call first):
  CMakeLists.txt:20 (include)
```

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits): Linux 64 bits
* Compiler: g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
* Release or debug mode: Release mode
* Specific flags used (if any):
* CGAL version: 5.1.5, 5.5.3 and 6.0-I-207 (master from some days ago)
* Boost version: 1.84.0
* Other libraries versions if used (Eigen, TBB, etc.):
cmake                   3.29.0               hcfe8598_0    conda-forge
eigen                     3.4.0                h00ab1b0_0    conda-forge
gmp                       6.3.0                h59595ed_1    conda-forge
mpfr                      4.2.1                h9458935_0    conda-forge

VincentRouvreau:
Will the major changes for CGAL 6.0 invalidate the CMake code above?
[...]

PR #8221, opened 2024-05-23
Fix issue #8192

lrineau:
Fix issue #8192: CMake `find_package` with a minimal CGAL version number does not detect properly 6.X CGAL version


sloriot:

VincentRouvreau:
Works fine with CGAL-6.0-I-254, thanks for your reactivity !

PR #8222, opened 2024-05-24
CGAL Lab: Add a target CGALlab_compile_all_plugins

lrineau:
## Summary of Changes

Fix a usability issue with Visual Studio Code. There is now a new target `CGALlab_compile_all_plugins` that compiles the executable `CGALlab` and all the plugins.

That target, as seen in the vscode Make Project Outline, is placed just below the `CGALlab` target.

Screenshot_20240524_165017

If the executable `CGALlab_compile_all_plugins` is launched, it just displays:

```
This executable is not intended to be run.
Please run the CGALlab executable instead.
```

## Release Management

* Affected package(s): Lab

### DIFF
--- a/Lab/demo/Lab/CGALlab_compile_all_plugins.cpp
+++ b/Lab/demo/Lab/CGALlab_compile_all_plugins.cpp
@@ -1,0 +1,6 @@
+#include <iostream>
+int main()
+{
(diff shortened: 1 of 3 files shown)


sloriot:

PR #8225, opened 2024-05-24
Fix property map calls in the demo

sloriot:
Fix bad optional access errors.


sloriot:

PR #8228, opened 2024-05-28
Fix using the fcolor pmap when there is no face color

MaelRL:
## Summary of Changes

This is a quick band aid such that the demo can be used. The proper fix would be to not have a Boolean on whether the pmap is defined as item member, while also passing the pmap as parameter of functions...

## Release Management

* Affected package(s): `Lab`
* Issue(s) solved (if any): -


sloriot (approved):

sloriot:
Successfully tested in CGAL-6.0-Ic-256

PR #8238, opened 2024-05-29
Fix CC selection in selection item

MaelRL:
## Summary

And remove CC selection from the deformation item, where it had limited use.

## Release Management

* Affected package(s): `Lab`
* Issue(s) solved (if any): -

### DIFF
--- a/Lab/demo/Lab/Plugins/PMP/Selection_plugin.cpp
+++ b/Lab/demo/Lab/Plugins/PMP/Selection_plugin.cpp
@@ -211,8 +211,13 @@ public:
     connect(ui_widget.selectionOrEuler, SIGNAL(currentChanged(int)), this, SLOT(on_SelectionOrEuler_changed(int)));
     connect(ui_widget.editionBox, SIGNAL(currentIndexChanged(int)), this, SLOT(on_editionBox_changed(int)));
 
+    ui_widget.Sharp_edges_label->hide();
+    ui_widget.Sharp_angle_spinbox->hide();
+    ui_widget.Select_sharp_edges_button->hide();
+    ui_widget.Select_boundaryButton->hide();
     ui_widget.Add_to_selection_button->hide();
     ui_widget.Select_all_NTButton->hide();
+
     QObject* scene = dynamic_cast<QObject*>(scene_interface);
     if(scene) {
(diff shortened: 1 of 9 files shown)


sloriot:
Works as expected, thanks! One remark: Components is now the default tab, it should be Simplices IMO

afabri:
> Works as expected, thanks! One remark: Components is now the default tab, it should be Simplices IMO

And in Simplices it should be faces    (statistically for me).

MaelRL:
> > Works as expected, thanks! One remark: Components is now the default tab, it should be Simplices IMO
> 
> And in Simplices it should be faces (statistically for me).

There are a lot of instances of "facet(s)", so I leave it out of scope of this PR.

sloriot:
> > > Works as expected, thanks! One remark: Components is now the default tab, it should be Simplices IMO
> > 
> > 
> > And in Simplices it should be faces (statistically for me).
> 
> There are a lot of instances of "facet(s)", so I leave it out of scope of this PR.

I guess Andreas meant putting face(t)s as default selected simplex instead of vertex

MaelRL:
The plugin uses everywhere the combo box ID (aka the position of "Facet" or "Vertex" in the list of options) so it's a lot more work than just changing the order in the UI. Not worth it right now.

sloriot:
Successfully tested in CGAL-6.0-Ic-257

afabri:
> Successfully tested in CGAL-6.0-Ic-257

Do we have to test it in the GUI on a Windows machine?

MaelRL:
> > Successfully tested in CGAL-6.0-Ic-257
> 
> Do we have to test it in the GUI on a Windows machine?

Can't hurt, both for the buttons that i have reorganized and so you can check the behavior of the CC buttons since you had the issue.

Note that I left a behavior that a bit is surprising at first but makes sense : when you click "select CCs with simplices count under threshold", it updates the threshold to a new value. That new value is the number of simplices for the next smallest, unselected CC. Thus, you keep clicking and it adds one CC at the time.

lrineau:
> > Successfully tested in CGAL-6.0-Ic-257
> 
> Do we have to test it in the GUI on a Windows machine?

I have actually merged the pull-request. Now the GUI can be tester from a build of CGAL Lab from `master`.

PR #8258, opened 2024-06-05
Add a Bbox class with dimension as parameter

afabri:
## Summary of Changes

For the upcoming Frechet Distance package we need a dD bounding box.  We do not have that in `Kernel_d`, and there is such a class in ign's Distributed Delaunay Triangulation package to come ([link](https://github.com/mbredif/ddt_cgal/blob/master/DDT/include/CGAL/DDT/kernel/Bbox.h)).   This PR copies it to the Kernel package (with the ok of @mbredif), but in the `CGAL` namespace.  

This draft pull request is to get a discussion started.

- In the initial version even the number type is a template parameter. Do we want to keep that?   
- Do we want to make `Bbox_2/3`  a `typedef` of this new class?
- We have to go through the API and probably enrich it with what `Bbox_2` offers additionally.
- Do we have to add functions in `Kernel_d`?

### TODO
- [ ] create a small feature and get it approved

## Release Management

* Feature/Small Feature (if any): tbd
* License and copyright ownership:  LGPL , owned by IGN

### DIFF
--- a/Kernel_23/include/CGAL/Bbox.h
+++ b/Kernel_23/include/CGAL/Bbox.h
@@ -20,6 +20,7 @@
 #include <iterator>
 #include <CGAL/assertions.h>
 #include <CGAL/Dimension.h>
+#include <CGAL/use.h>
 
 namespace CGAL {
 namespace Impl {
@@ -112,6 +113,7 @@ protected:
 
     template <typename I>
     void init(int d, I b, I e) {
+        CGAL_USE(e);
         CGAL_assertion(d == std::distance(b,e));
         for(int i=0; i<d; ++i,++b)
         {


afabri:
@mglisse I am also wondering why there is no bounding box class in `Epick_d` etc. Apparently there is no need for it so far, but would you like to see it integrated?  With functors etc.

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,159 @@
...
+    Container max_values;
+
+public:
+    Bbox& operator+=(const Bbox& bbox)
+    {
+        CGAL_assertion(min_values.size() == 0 || min_values.size() == bbox.min_values.size());
+        int dim = bbox.min_values.size();
+        for(int i=0; i<dim; ++i)
+        {
+            if(min_values[i] > bbox.min_values[i]) min_values[i] = bbox.min_values[i];
```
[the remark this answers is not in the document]
mbredif: That's wrong indeed.
min_values.size()==0 happens for a default constructed Bbox with a dynamic dimension, which models an empty bbox. In this case, values should just be copied from bbox to this.

afabri:
@mbredif do I have to make you collaborator, or is this already the case?

mglisse:
> @mglisse I am also wondering why there is no bounding box class in `Epick_d` etc.

I remember thinking about it, maybe discussing it, but I don't remember the content...

> Apparently there is no need for it so far,

That may have played a big role :wink: (it doesn't seem to be part of the Kernel_d concept)

> but would you like to see it integrated? With functors etc.

If you have a use for it, we probably should.
To be sure I understand: the current bbox_23 always use `double`, whatever the kernel, making them not really kernel objects. Is the plan the same for the new bbox, i.e. ignore the generality offered by the template parameter and only use it with `double` (otherwise it becomes Iso_rectangle_2/Iso_cuboid_3)?
It looks like the kernel integration can be done later in a separate PR (this one adds the class), which gives me a bit of time to think about how that fits with the design of NewKernel_d.

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,159 @@
...
+{
+public:
+    inline int dimension() const { return this->min_values.size(); }
+    Bbox(int d = 0          ) { init_values(d); this->init(d       ); }
+    Bbox(int d, double range) { init_values(d); this->init(d, range); }
+
+protected:
+    void init_values(int d) {
+        this->min_values.resize(d);
+        this->max_values.resize(d);
```
mglisse, line 133: It seems a bit wasteful to use 2 containers of size d instead of one of size 2*d, but that's an implementation detail so it doesn't matter now.

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,161 @@
...
+    }
+};
+
+template<typename Container, typename Derived>
+std::ostream& operator<<(std::ostream& out, const Impl::Bbox<Container, Derived>& bbox)
+{
+    int d = bbox.dimension();
+    for(int i=0; i<d; ++i)
+        out  << (bbox.min)(i) << "  " << (bbox.max)(i) << " ";
```
afabri, line 145: `Bbox_2/3` first print lower left and then upper right.

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,161 @@
...
+
+// A D-dimensional axis aligned box
+template<unsigned int N, typename T>
+class Bbox : public Impl::Bbox<std::array<T, N>, Bbox<N,T>>
+{
+    enum { D = N };
+public:
+    inline constexpr int dimension() const { return D; }
+    Bbox(int d = 0          ) { assert(d==N || d==0); this->init(d       ); }
+    Bbox(int d, double range) { assert(d==N || d==0); this->init(d, range); }
```
afabri, line 121: double should be T, right?

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// A D-dimensional axis aligned box
+template<unsigned int N, typename T>
+class Bbox : public Impl::Bbox<std::array<T, N>, Bbox<N,T>>
+{
+    enum { D = N };
+public:
+    inline constexpr int dimension() const { return D; }
+    Bbox(int d = 0          ) { assert(d==N || d==0); this->init(d       ); }
```
afabri, line 120: I find it surprising that for a Bbox with non-dynamic dimension I have to pass parameter `d`.

mbredif: that was to offer a common API whatever the dimension dynamics

afabri:
We also need a constructor taking an iterator range of coordinates. And maybe also for a range of pairs of coordinates.
The former is needed when constructing a bbox from a point with floating point coordinates. The latter when constructing
it from exact coordinates transformed to a `std::pair<double,double>` using `CGAL::to_double()`.

afabri:
As `Epick_d` may have a dynamic and a static dimension, depending on a template parameter, the functor `Epick_d::Construct_bbox_d()`  must return either a bbox type with dynamic and static dimension, respectively.

sloriot:
Kernel_23 failing in CGAL-6.0-Ic-284

sloriot:
Kernel_23 tests are failing in

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,174 @@
...
+    void init(int d, T range = -std::numeric_limits<T>::infinity()) {
+        for(int i=0; i<d; ++i)
+        {
+            min_values[i] = -range;
+            max_values[i] =  range;
+        }
+    }
+
+    template <typename I>
+    void init(int d, I b, I e) {
```
sloriot, line 111: `e` is not used

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,174 @@
...
+        for(int i=0; i<dimension(); ++i) {
+            result *= (std::min)((max)(i), (bbox.max)(i)) -
+                      (std::max)((min)(i), (bbox.min)(i));
+            if (result <= 0) return 0;
+        }
+        return result;
+    }
+
+    bool operator==(const Bbox& bbox) const {
+        for(int i=0; i<dimension(); ++i)
```
afabri, line 93: @mbredif don't we have to check that the dimensions are equal?

afabri:
@mglisse how can I obtain for `Epick_d`   if it has a dynamic dimension or not?   I want to add a `Compute_bbox`  in `function_objects_cartesian.h`  and when I want to specify the `result_type`  I have to provide the appropriate `Bbbox` type which has a specialization for the dynamic bbox.

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
```diff
@@ -1006,6 +1007,29 @@ template<class R_> struct Squared_length : private Store_kernel<R_> {
...
+        result_type operator()(Point const&a)const{
+                CI ci(this->kernel());
+                typename Real_embeddable_traits<RT>::To_interval f;
+                typename Get_functor<R, Point_dimension_tag>::type pd(this->kernel());
+                return result_type(pd(a), make_transforming_iterator(ci(a,Begin_tag()),f), make_transforming_iterator(ci(a,End_tag())), f);
+        }
+};
+}
+
+CGAL_KD_DEFAULT_FUNCTOR(Construct_bbox_tag,(CartesianDKernelFunctors::Construct_bbox<K>),(Point_tag),(Construct_ttag<Point_cartesian_const_iterator_tag>));
```
sloriot, line 1030: @mglisse do you have any doc for the macro `CGAL_KD_DEFAULT_FUNCTOR`. I'm pretty sure this line is incorrect but I don't know how to fix it.

afabri:
I am wondering if it is not just something missing in `Lazy_cartesian.h`   as I now got it working for `Epick_d`.

NewKernel_d/include/CGAL/NewKernel_d/functor_tags.h
```diff
@@ -201,6 +201,7 @@ namespace CGAL {
         CGAL_DECL_COMPUTE(Squared_distance);
         CGAL_DECL_COMPUTE(Squared_distance_to_origin);
         CGAL_DECL_COMPUTE(Squared_length);
+        CGAL_DECL_COMPUTE(Construct_bbox);
```
mglisse, line 204: I think it makes more sense as CGAL_DECL_MISC. "Compute" is for things that return a Lazy_exact_nt in Epeck_d.
And then Lazy_cartesian needs something similar to Point_dimension_tag, assuming that the Bbox is supposed to be built from the approximate object (maybe merge the cases Point_dimension_tag, Vector_dimension_tag, and this new one in a partial specialization `struct Functor<T,D,Misc_tag>`?)

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
mglisse: The line looks ok to me. You have the tag, the class, the required objects, the required other functors, that's it. Ah, maybe you are missing Point_dimension_tag in the last list (I am not sure this list is really used currently).

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
```diff
@@ -1006,6 +1007,29 @@ template<class R_> struct Squared_length : private Store_kernel<R_> {
...
+        CGAL_FUNCTOR_INIT_STORE(Construct_bbox)
+        typedef R_ R;
+        typedef typename R::Dimension Dimension;
+        typedef typename Get_type<R, RT_tag>::type RT;
+        typedef typename Get_type<R, Point_tag>::type Point;
+        typedef typename Get_functor<R, Construct_ttag<Point_cartesian_const_iterator_tag> >::type CI;
+
+        typedef Bbox<Dimension,double> result_type;
+        typedef Point argument_type;
+        result_type operator()(Point const&a)const{
```
mglisse, line 1021: Out of curiosity, is it likely to stick to just this function, or do you expect to add overloads for Segment, Sphere, Iso_box, etc? I am asking because ideally (I stopped before making things consistent), overload resolution should happen in Kernel_d_interface, and dispatch to different non-overloaded functors like Bbox_from_point, Bbox_from_segment.

mglisse:
> In fact there is the nested type `Dimension`

Yes, that's also what we use in Gudhi to access the compile-time dimension of a CGAL Kernel. It is even documented in the Kernel_d concept :four_leaf_clover: .

afabri:
@sloriot and I, we are  blocked with getting `Epick_d.cpp` compiled.  There is probably a specialization we have to add, but we have no idea what exactly,

NewKernel_d/include/CGAL/NewKernel_d/functor_tags.h
```diff
@@ -260,6 +260,7 @@ namespace CGAL {
         CGAL_DECL_CONSTRUCT(Construct_circumcenter,Point);
         CGAL_DECL_CONSTRUCT(Point_drop_weight,Point);
         CGAL_DECL_CONSTRUCT(Power_center,Weighted_point);
+        CGAL_DECL_CONSTRUCT(Construct_bbox,Bbox);
```
mglisse, line 263: I suggested MISC though, since this doesn't return a Lazy object in Epeck_d.

mglisse:
> @sloriot and I, we are blocked with getting `Epick_d.cpp` compiled. There is probably a specialization we have to add, but we have no idea what exactly,

I mentioned looking at Point_dimension_tag in Lazy_cartesian.h. Something like

```diff
--- a/NewKernel_d/include/CGAL/NewKernel_d/Lazy_cartesian.h
+++ b/NewKernel_d/include/CGAL/NewKernel_d/Lazy_cartesian.h
@@ -309,24 +309,15 @@ struct Lazy_cartesian :
     template<class T,class D> struct Functor<T,D,Construct_tag> {
             typedef Lazy_construction2<T,Kernel> type;
     };
-    template<class D> struct Functor<Point_dimension_tag,D,Misc_tag> {
-            typedef typename Get_functor<Approximate_kernel, Point_dimension_tag>::type FA;
-            struct type {
-              FA fa;
-              type(){}
-              type(Kernel const&k):fa(k.approximate_kernel()){}
-              template<class P>
-              int operator()(P const&p)const{return fa(CGAL::approx(p));}
-            };
-    };
-    template<class D> struct Functor<Vector_dimension_tag,D,Misc_tag> {
-            typedef typename Get_functor<Approximate_kernel, Vector_dimension_tag>::type FA;
+    template<class T,class D> struct Functor<T,D,Misc_tag> {
+            // By default (Point_dimension, Construct_bbox), apply to the approximate object
+            typedef typename Get_functor<Approximate_kernel, T>::type FA;
             struct type {
               FA fa;
               type(){}
               type(Kernel const&k):fa(k.approximate_kernel()){}
-              template<class V>
-              int operator()(V const&v)const{return fa(CGAL::approx(v));}
+              template<class P...>
+              int operator()(P&&...p)const{return fa(CGAL::approx(std::forward<P>(p))...);}
             };
     };
     template<class D> struct Functor<Linear_base_tag,D,Misc_tag> {
```
(completely untested)
If that still doesn't work, I'll checkout the branch to experiment, but not right now.

sloriot:
@mglisse in the meantime I achieve to make it compile and run. Not sure if that's the right way to do it.

mglisse:
> @mglisse in the meantime I achieved to make it compile and run. Not sure if that's the right way to do it.

See https://github.com/mglisse/cgal/tree/Kernel_23-Bbox_d-ign, where I added commit https://github.com/mglisse/cgal/commit/316ab90c981f46542dc64a1fa95e70232db53147 .

sloriot:
> > @mglisse in the meantime I achieved to make it compile and run. Not sure if that's the right way to do it.
> 
> See https://github.com/mglisse/cgal/tree/Kernel_23-Bbox_d-ign, where I added commit [mglisse@316ab90](https://github.com/mglisse/cgal/commit/316ab90c981f46542dc64a1fa95e70232db53147) .

That's indeed much simpler. I did not see that MISC macro. Thanks!

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
afabri: It would be natural to generalize.  We started this only for the Frechet Distance submission, where we only deal with points.

sloriot:

sloriot:
Still a draft and small feature missing.

sloriot:
> Still a draft and small feature missing.

Ping @afabri

sloriot:
@afabri what's the plan for this PR?

afabri:
We can close this PR as this was introduced while working on Frechet Distance.

PR #6849, opened 2022-09-15
Isosurfacing_3 (new package)

JulyCode:
## Summary of Changes

Extracting isosurfaces from scalar functions is a common method for generating 3D meshes. An isosurface is a subspace with equal values of the function.

### ToDo:
- [ ] Check branch size
- [x] TMC_functor::compute_edge_index() uses operator[] on cell_descriptor (violates concept)
- [x] Check Orthtree as partition for Marching Cubes after fixing compute_edge_index()
- [x] MC vs TMC picture

### Extra:
- [ ] Use Isosurfacing_3 in the demo (Isosurface_3_plugin)
- [ ] Optimize parallelism

## Release Management

* Affected package(s): `Isosurfacing_3`
* Issue(s) solved (if any): - 
* Feature/Small Feature (if any): [Isosurfacing_3](https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Isosurfacing_3)
* License and copyright ownership: see feature

PR #8259, opened 2024-06-05
Hyperbolic surface triangulation 2 (new package)

loic-dubois:
## Summary of Changes

This package enables building and handling triangulations of closed orientable hyperbolic surfaces: a basic tool for any future implementation of algorithms operating on such surfaces. Functionalities are offered such as the Delaunay flip algorithm, and the construction of a portion of the lift of the triangulation in the Poincaré disk model of the hyperbolic plane. Triangulations can be generated by triangulating a convex fundamental domain in the Poincaré disk. A method is offered that generates such domains in genus two. 

## Release Management

* Affected package(s): `Hyperbolic_surface_triangulation_2`
* Issue(s) solved (if any): -
* Feature/Small Feature (if any): https://cgal.geometryfactory.com/CGAL/Members/wiki/Features/Hyperbolic_Surface_Triangulation
* Link to compiled documentation: [link](https://cgal.github.io/8259/v0/Manual/packages.html#PkgHyperbolicSurfaceTriangulation2)
* License and copyright ownership: GPL INRIA


Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s)     : Vincent Despré, Loïc Dubois, Monique Teillaud
+
+// This file contains the declaration and the implementation of the class Hyperbolic_fundamental_domain_2
+
+#ifndef CGAL_HYPERBOLIC_FUNDAMENTAL_DOMAIN_2
+#define CGAL_HYPERBOLIC_FUNDAMENTAL_DOMAIN_2
+
+#include "Complex_without_sqrt.h"
```
afabri, line 18:
```suggestion
#include <CGAL/Complex_without_sqrt.h>
```

afabri: and the same everywhere

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+template<class Traits>
+Hyperbolic_fundamental_domain_2<Traits>::Hyperbolic_fundamental_domain_2() {}
+
+template<class Traits>
+Hyperbolic_fundamental_domain_2<Traits>::Hyperbolic_fundamental_domain_2(const std::vector<Point>& vertices, const std::vector<int>& pairings){
+  set(vertices, pairings);
```
afabri, line 72: do not use the `set()` function but data member initialization

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+template<class Traits>
+int Hyperbolic_fundamental_domain_2<Traits>::size() const{
+  return _vertices.size();
+}
+
+template<class Traits>
+typename Hyperbolic_fundamental_domain_2<Traits>::Point Hyperbolic_fundamental_domain_2<Traits>::vertex(int index) const{
```
afabri, line 91: Use `const&` so that points do not get copied.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+
+template<class Traits>
+void Hyperbolic_fundamental_domain_2<Traits>::from_stream(std::istream& s){
+  _vertices.clear();
+  _pairings.clear();
+
+  std::string line;
+  s >> line;
+  int size = std::stoi(line);
+  for (int k=0; k<size; k++){
```
afabri, line 140: Use `reserve(size)` for the two vectors.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_factory_2.h
```diff
@@ -0,0 +1,262 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+template<class Traits>
+typename Traits::FT Hyperbolic_fundamental_domain_factory_2<Traits>::exact_number_from_float(float x){
+  if (x< 0){
+    return _FT(0)-exact_number_from_float(-x);
+  }
+  return _FT(int(x*_DENOMINATOR_FOR_GENERATION)%_DENOMINATOR_FOR_GENERATION)/_FT( _DENOMINATOR_FOR_GENERATION);
```
afabri, line 154: macros must start with `CGAL_`

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_isometry_2.h
```diff
@@ -0,0 +1,183 @@
...
+/*
+Represents a hyperbolic isometry in the Poincare disk model the hyperbolic plane. The isometry f is stored as list (c0, c1, c2, c3) of 4 complex numbers,
+so that f(z) = (c0 z + c1) / (c2 z + c3) holds on every complex z in the open unit disk.
+*/
+template<class Traits>
+class Hyperbolic_isometry_2{
+public:
+  typedef Hyperbolic_isometry_2<Traits>                    Self;
+  typedef typename Traits::FT                              FT;
+  typedef typename Traits::Complex                         ComplexNumber;
```
afabri, line 31: No CamelCase in CGAL.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
```diff
@@ -0,0 +1,721 @@
...
+  }
+
+  std::tuple<Dart_const_handle,Point,Point,Point> value = std::make_tuple(_anchor.dart, positions[_anchor.dart], positions[const_ccw(_anchor.dart)], positions[const_cw(_anchor.dart)]);
+  realizations.push_back(value);
+
+  ComplexNumber anchor_z0 (_anchor.vertices[0].x(), _anchor.vertices[0].y());
+  ComplexNumber anchor_z1 (_anchor.vertices[1].x(), _anchor.vertices[1].y());
+  ComplexNumber anchor_z2 (_anchor.vertices[2].x(), _anchor.vertices[2].y());
+
+  double weight_of_anchor_dart = CGAL::to_double(anchor_z0.squared_modulus() + anchor_z1.squared_modulus());
```
afabri, line 396: Spotting a `double` without understanding what the function is about immediately leads to the question if this could lead to wrong results.

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_complex.cpp
```diff
@@ -0,0 +1,96 @@
...
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s)     : Vincent Despré, Loïc Dubois, Monique Teillaud
+
+#include <CGAL/Complex_without_sqrt.h>
+#include <CGAL/Gmpq.h>
```
afabri, line 14: Use `CGAL::Exact_rational` instead.

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_circular_kernel.cpp
```diff
@@ -0,0 +1,78 @@
...
+//
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s)     : Vincent Despré, Loïc Dubois, Monique Teillaud
+
+#include <iostream>
```
afabri, line 13: Recently we found errors where CGAL header files had did miss to include what they depend on. 
In order to test that we should include in test files the new headers before including number types, std files etc.
So below lines 22-25 should be moved before line 13.

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2_Demo/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+# Find includes in corresponding build directories
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+include_directories(${CMAKE_BINARY_DIR})
+
+# Instruct CMake to run moc automatically when needed.
+set(CMAKE_AUTOMOC ON)
+
+
+# CGAL and its components
+find_package(CGAL REQUIRED COMPONENTS Core Qt5)
```
afabri, line 14: We are at Qt6.  The transition is rather smooth

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2_Demo/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+cmake_minimum_required(VERSION 3.1...3.15)
+
+project( demo )
```
afabri, line 3: make the name more specific. Also the one of the cpp

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
```diff
@@ -0,0 +1,721 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+template<class Traits>
+Hyperbolic_surface_triangulation_2<Traits>::Hyperbolic_surface_triangulation_2(const Domain& domain){
+  // (Triangulates by adding an internal edge between domain.vertex(size-1) and the other vertices)
+  _combinatorial_map.clear();
+  int size = domain.size();
+
+  // Make the triangles
+  Dart_handle dart_of_triangle[size-2];
```
afabri: You must use a `std::vector`.   A built-in array VC++ does not allow a size known at runtime. I just fixed that.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+  s << std::to_string(n) << std::endl;
+
+  for (int k=0; k<n; k++){
+    s << paired_side(k) << std::endl;
+  }
+
+  for (int k=0; k<n; k++){
+    s << vertex(k) << std::endl;
+  }
+}
```
lrineau, line 130: Instead of `void`, return the stream:
```suggestion
std::ostream& Hyperbolic_fundamental_domain_2<Traits>::to_stream(std::ostream& s) const{
  int n = size();

  s << std::to_string(n) << std::endl;

  for (int k=0; k<n; k++){
    s << paired_side(k) << std::endl;
  }

  for (int k=0; k<n; k++){
    s << vertex(k) << std::endl;
  }
  return s;
}
```

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+
+template<class Traits>
+void Hyperbolic_fundamental_domain_2<Traits>::from_stream(std::istream& s){
```
lrineau, line 133: Same here: return the reference to `s`.

That allows nice code like:

```c++
if(!hyperbolic_domain.from_stream(s).fail()) { // ...
```

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_circular_kernel.cpp
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,210 @@
...
+  return _pairings[index];
+}
+
+template<class Traits>
+Hyperbolic_isometry_2<Traits> Hyperbolic_fundamental_domain_2<Traits>::side_pairing(int index) const{
+  int n = size();
+  int paired_index = paired_side(index);
+
+  Point p1,p2,q1,q2;
+  q1 = vertex(index);
```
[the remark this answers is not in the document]
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_isometry_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_complex.cpp
pougetma: done

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_circular_kernel.cpp
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_factory_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
pougetma: The weights instantiated via a to_double are just used here for a heuristic algorithm to deal with a priority queue during the lift, the result of the lift is exact but the order in which the lift is processed does not need more than a heuristic.
@loic-dubois : anything to add?

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2_Demo/CMakeLists.txt
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2_Demo/CMakeLists.txt
pougetma: done

sloriot:
check [this doc page](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-your-membership-in-organizations/publicizing-or-hiding-organization-membership)

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_without_sqrt.h
```diff
@@ -0,0 +1,173 @@
...
+
+template<class FT>std::ostream& operator<<(std::ostream& s, const Complex_without_sqrt<FT>& z);
+template<class FT>void operator>>(std::istream& s, Complex_without_sqrt<FT>& z);
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+template<class FT>
+Complex_without_sqrt<FT>::Complex_without_sqrt(){
```
afabri, line 63: Change to a member initializer list.

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_without_sqrt.h
```diff
@@ -0,0 +1,173 @@
...
+
+template<class FT>
+void Complex_without_sqrt<FT>::set_imaginary_part(const FT& imaginary_part){
+  _imag = imaginary_part;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+template<class FT>
+FT Complex_without_sqrt<FT>::real_part() const{
```
afabri, line 95:
```suggestion
const FT& Complex_without_sqrt<FT>::real_part() const{
```

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_isometry_2.h
```diff
@@ -0,0 +1,184 @@
...
+  Hyperbolic_isometry_2();
+
+  void set_to_identity();
+
+  // Can be used to set the coefficients manually. Be careful when doing so : the implementation does not check that the resulting moebius transform fixes the unit circle.
+  void set_coefficients(const Complex_number& c0, const Complex_number& c1, const Complex_number& c2, const Complex_number& c3);
+  void set_coefficient(int index, const Complex_number& coefficient);
+
+  // Returns the index-th coefficient
+  Complex_number get_coefficient(int index) const;
```
afabri, line 43: return `const&`

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
```diff
@@ -0,0 +1,747 @@
...
+    queue.pop();
+
+    Dart_const_handle invaded = const_opposite(invader);
+
+    if (!_combinatorial_map.is_marked(invaded, visited_darts_mark)){
+      _combinatorial_map.mark(invaded, visited_darts_mark);
+      _combinatorial_map.mark(const_ccw(invaded), visited_darts_mark);
+      _combinatorial_map.mark(const_cw(invaded), visited_darts_mark);
+
+      Point a = positions[const_ccw(invader)];
```
afabri, line 430: make points `const&`

afabri:
@loic-dubois @pougetma What is not offered by [`std::complex`](https://en.cppreference.com/w/cpp/numeric/complex) for that you come up with your own class ?

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/CGAL/Hyperbolic_fundamental_domain_factory_2.h
```diff
@@ -3,7 +3,10 @@ namespace CGAL{
 /*!
 \ingroup PkgHyperbolicSurfaceTriangulation2MainClasses
 
-Factory class, whose purpose is to generate some convex domains of surfaces of genus two.
+Factory class, whose only purpose is to construct random fundamental domains of
+closed orientable hyperbolic surfaces. The method
+make_hyperbolic_fundamental_domain_g2 constructs such a domain for a surface of
```
afabri, line 8: backtick the function.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_isometry_2.h
pougetma: done

afabri:
@pougetma can you please configure your text editor that it does remove trailing white space. And remove the current WS.

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_without_sqrt.h
pougetma: done

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
pougetma: done

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/CGAL/Hyperbolic_fundamental_domain_factory_2.h
pougetma: done

pougetma:
> > @loic-dubois @pougetma What is not offered by [`std::complex`](https://en.cppreference.com/w/cpp/numeric/complex) for that you come up with your own class ?
> 
> We represent each complex number by the pair of its real and imaginary parts. We require that the number type of the real and imaginary parts supports the basic arithmetic operations +, -, *, /, and computes them exactly (without any rounding). But we dot not require anything else. In particular, we do not want to use square roots or things like that. This is a constraint that we chose, hoping that it would make computations faster.
> 
> So for example, if we were to use CGAL:Gmpq for the real and imaginary parts, then we could represent the complex number 1 + i, but we should not be able to compute its norm (for it is not a rational number).
> 
> But then std::complex has methods like "norm" or "arg". So I guess that std::complex, templated by CGAL:Gmpq, should not even compile?

It seems to me that std::complex is fine as a model for the concept
ComplexWithoutSqrt, (note that the "norm" is in fact the squared modulus).
We "only" have to modify the member names of our concept so that it fits the
one we need from std::complex. But the fact that sdt::complex provides more is
not a pb.

afabri:
> We "only" have to modify the member names of our concept so that it fits the
> one we need from std::complex. But the fact that sdt::complex provides more is
> not a pb.

I would suggest to switch to `std::complex` then.  The less we have to maintain  because it is offered by `std::` types, the better.
With the functions to change, I guess you mean `std::complex::imag()`.   I agree.

afabri:
Do I get it right that [`Exact_complex` ](https://github.com/CGAL/cgal/blob/master/Hyperbolic_triangulation_2/include/CGAL/Hyperbolic_triangulation_2/internal/Exact_complex.h) is completely independent from `Complex_without_sqrt`  and not used at all in your package?

afabri:
Replacling with `std::complex`   and modifying some functions compiles but I get the following warning from VC++. 
````

1>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\include\complex(2054,24): warning C4996: 'std::complex<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational,boost::multiprecision::et_on>>::complex': warning STL4037: The effect of instantiating the template std::complex for any type other than float, double, or long double is unspecified. You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress this warning.
````

**Comment by pougetma** [comment]
> Replacling with `std::complex` and modifying some functions compiles but I get the following warning from VC++.
> 
> ```
> 
> 1>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\include\complex(2054,24): warning C4996: 'std::complex<boost::multiprecision::number<boost::multiprecision::backends::gmp_rational,boost::multiprecision::et_on>>::complex': warning STL4037: The effect of instantiating the template std::complex for any type other than float, double, or long double is unspecified. You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress this warning.
> ```
> 
> The example also runs, but no idea if it does what you would expect. Shall I commit and push it somewhere else?

I'm working on the move to std::complex, so please let me the token.

**Comment by afabri** [comment]
> I'm working on the move to std::complex, so please let me the token.

The question is if we want to switch to `std::complex`  or not.  The warning gives me doubts. Any feedback is welcome ( @mglisse @lrineau @sloriot  ?)

**Comment by mglisse** [comment]
> The question is if we want to switch to `std::complex` or not. The warning gives me doubts.

The warning is quite clear. Even if it happens to work today (you would need to check each standard library with each CGAL number type), some implementation might add constraints tomorrow (or tweak their implementation in a way that breaks expression templates) and `std::complex<Exact_rational>` would stop compiling. I don't think it is worth the risk. Besides, even if it works with non-fundamental types, std::complex was not optimized for them and may perform many unnecessary copies.
I don't know if boost provides a more relaxed `complex` template. If necessary, having one in CGAL (preferably in the Number_types package?) is not a huge amount of code.

Comments about Complex_without_sqrt
* the preferred form for binary operators (like +, *, <<, etc) is "hidden friend"
* for a proper number type, you also want +=, -=, etc (those are members)
* At least the constructor with 2 arguments should handle rvalues more efficiently (easiest would be to make it a template and forward its arguments)

**Comment by afabri** [comment]
And what about aligning the names of functions to `std::complex`   instead of the CGAL naming style?   I would prefer to align.

**Comment by mglisse** [comment]
> And what about aligning the names of functions to `std::complex` instead of the CGAL naming style? I would prefer to align.

No opinion. It would indeed make it easier to replace it with some other implementation at some point in the future.

**Comment by pougetma** [comment]
So the idea would be to add CGAL::Complex<FT> in the number types package. I will first do it locally in this package and comply to both Field concept of cgal and the naming of std::complex.

**Comment by afabri** [comment]
I had made the replacement with `std::complex`.  I have pushed it to `afabri`:  https://github.com/afabri/cgal/pull/new/Hyperbolic_surface_triangulation_2-dubois . Just in case it might be useful for you, Marc.

**Comment by pougetma** [comment]
I managed to get a class that looks like a number type. 
@mglisse I didn't get your point: "At least the constructor with 2 arguments should handle rvalues more efficiently (easiest would be to make it a template and forward its arguments)"  ?

**Comment by mglisse** [comment]
> I didn't get your point: "At least the constructor with 2 arguments should handle rvalues more efficiently (easiest would be to make it a template and forward its arguments)" ?

```c++
template<class U,class V>
Complex_without_sqrt(U&& real_part, V&& imaginary_part): _real(std::forward<U>(real_part)), _imag(std::forward<V>(imaginary_part)) {}
```

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_without_sqrt.h
```diff
@@ -0,0 +1,230 @@
...
+  _Self conjugate() const;
+
+  _Self& operator+=(const _Self& other);
+  _Self& operator-=(const _Self& other);
+  _Self& operator*=(const _Self& other);
+  _Self& operator/=(const _Self& other);
+  _Self& operator=(const _Self& other);
+
+  template<class FT_>
+  friend Complex_without_sqrt<FT_> operator+(const Complex_without_sqrt<FT_>& z);
```
mglisse, line 50: Unary operators were good before, as members.

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_without_sqrt.h
```diff
@@ -0,0 +1,230 @@
...
+  template<class FT_>
+  friend Complex_without_sqrt<FT_> operator-(const Complex_without_sqrt<FT_>& z);
+  template<class FT_>
+  friend bool operator==(const Complex_without_sqrt<FT_>& z1, const Complex_without_sqrt<FT_>& z2);
+  template<class FT_>
+  friend bool operator!=(const Complex_without_sqrt<FT_>& z1, const Complex_without_sqrt<FT_>& z2);
+  template<class FT_>
+  friend Complex_without_sqrt<FT_> operator+(const Complex_without_sqrt<FT_>& z1, const Complex_without_sqrt<FT_>& z2);
```
mglisse, line 58: Note that I was advocating **hidden** friends, i.e. entirely inside the class
```c++
friend Complex_without_sqrt operator+(Complex_without_sqrt const&z1, Complex_without_sqrt const&z2) {
  return { z1.real()+z2.real(), z1.imag()+z2.imag() };
}
```

pougetma:
If we agree on the naming and members of the Complex_nb class, I will fix the doc and test.

afabri:
> If we agree on the naming and members of the Complex_nb class, I will fix the doc and test.

I discussed this with @MaelRL and @sloriot, and we suggest that the concept should use the API of `std::complex()`  for accessing and modifying the real and imaginary part.  Also the functions for norm etc should be free functions and not member functions.

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/Concepts/HyperbolicSurfaceTraits_2.h
```diff
@@ -13,7 +13,7 @@ class HyperbolicSurfaceTraits_2 {
   /// \name Types
   /// @{
   /*!
-          represents a complex number over a field: must be a model of ComplexWithoutSqrt.
+          represents a complex number over a field: must be a model of ComplexNumber.
```
afabri, line 16:
```suggestion
          represents a complex number over a field: must be a model of `ComplexNumber`.
```

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/Concepts/ComplexNumber.h
```diff
@@ -0,0 +1,156 @@
...
+    /// @}
+
+  /// \name Operations
+  /// @{
+  /*!
+    returns the square of the modulus.
+  */
+  FT squared_modulus() const;
+
+  /*!
```
afabri, line 78: There is no `std::complex<>::conjugate()`,  but `std::conj()`.   
I went to adapting your branch. Have a look [here](https://github.com/afabri/cgal/commit/5e3ab0a60a92f6c5d0f6eef22073d228cbfaa846).

afabri: And similar for `squared_modulus()`

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_complex.cpp
```diff
@@ -1,16 +1,15 @@
...
 #include <iostream>
 #include <sstream>
 #include <complex>
 
 using namespace CGAL;
 
-typedef Complex_without_sqrt<Gmpq> Complex_gmpq;
-typedef Complex_without_sqrt<Exact_integer> Complex_integer;
-typedef Complex_without_sqrt<Interval_nt<>> Complex_interval;
+typedef Complex_number<Gmpq> Complex_gmpq;
```
afabri, line 10: No need to test with `Gmpq`.  There will be test platforms where `Exact_rational`  is this type.  And there will be plarforms where `GMP`  is not even installed.

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(Hyperbolic_surface_triangulation_2_Tests)
 
 find_package(CGAL REQUIRED)
 
-set(CMAKE_BUILD_TYPE "Release")
+set(CMAKE_BUILD_TYPE "Debug")
 
 include_directories(../../include/)
```
afabri, line 11: No need for this

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2/Hyperbolic_surface_triangulation_2_demo.cpp
```diff
@@ -0,0 +1,54 @@
...
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
+//
+// Author(s)     : Vincent Despré, Loïc Dubois, Monique Teillaud
+
+#include "window.h"
+
+#include <CGAL/Gmpq.h>
```
afabri, line 15: Use `Exact_rational`

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_traits_2.h
```diff
@@ -0,0 +1,80 @@
...
+template<class ComplexType>
+class Hyperbolic_point_2 {
+private:
+  ComplexType _z;
+
+public:
+  typedef ComplexType                           Complex;
+
+  Hyperbolic_point_2() { _z = Complex(); }
+  Hyperbolic_point_2(const Complex& z) { _z = z; }
```
afabri, line 34:
```suggestion
  Hyperbolic_point_2(const Complex& z) : _z(z) {}
```
The same may apply elsewhere, so please check.

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_traits_2.h
```diff
@@ -0,0 +1,80 @@
...
+
+  Complex get_z() const { return _z; }
```
afabri, line 36:
```suggestion
  const Complex& get_z() const { return _z; }
```

Hyperbolic_surface_triangulation_2/include/CGAL/Hyperbolic_surface_triangulation_2.h
```diff
@@ -0,0 +1,748 @@
...
+  _combinatorial_map.unmark_all(visited_darts_mark);
+
+  struct Compare {
+    bool operator()(std::pair<Dart_const_handle,double> const & x, std::pair<Dart_const_handle,double> const & y) {
+        return x.second > y.second;
+    }
+  };
+  std::priority_queue<std::pair<Dart_const_handle,double>, std::vector<std::pair<Dart_const_handle,double>>, Compare> queue;
+
+  std::map<Dart_const_handle, Point> positions;
```
afabri, line 383: Maybe not critical but you might use `std::unordered_map` instead which hashes.

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_isometry.cpp
```diff
@@ -0,0 +1,86 @@
+#include <CGAL/Hyperbolic_surface_traits_2.h>
+#include <CGAL/Complex_number.h>
+#include <CGAL/Hyperbolic_isometry_2.h>
+
+#include <iostream>
+#include <CGAL/Exact_rational.h>
+#include <CGAL/Cartesian.h>
+#include <CGAL/Hyperbolic_Delaunay_triangulation_traits_2.h>
+
+using namespace CGAL;
```
afabri, line 10: Please do not use `using namespace CGAL;`

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_isometry.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+typedef Cartesian<Gmpq>                                                 Kernel;
```
afabri, line 12: This `Gmpq`  will compile in case `Exact_rational` happens to be `Gmpq`.

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/hs_test_isometry.cpp
```diff
@@ -0,0 +1,86 @@
+#include <CGAL/Hyperbolic_surface_traits_2.h>
+#include <CGAL/Complex_number.h>
```
afabri, line 2: Only `#include` what is required, so that the testsuite can check that the required headers include everything they need.

Hyperbolic_surface_triangulation_2/README.md
```diff
@@ -0,0 +1,3 @@
+This folder contains the package Hyperbolic_surface_triangulation_2.
```
afabri, line 1: Remove this file.  When we generate a tarball for a CGAL distribution this file would be right at the root.

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_number.h
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) 2024
+// INRIA Nancy (France), and Université Gustave Eiffel Marne-la-Vallée (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
```
afabri, line 9: As this will sooner or later go into the Number_types package it should have the LGPL header.

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_number.h
```diff
@@ -0,0 +1,202 @@
...
+//
+// Author(s)     : Vincent Despré, Loïc Dubois, Monique Teillaud
+
+// This file contains the declaration and the implementation of the class Complex_number
+
+#ifndef CGAL_COMPLEX_NUMBER
+#define CGAL_COMPLEX_NUMBER
```
afabri, line 16:
```suggestion
#ifndef CGAL_COMPLEX_NUMBER_H
#define CGAL_COMPLEX_NUMBER_H
```

Hyperbolic_surface_triangulation_2/include/CGAL/Complex_number.h
```diff
@@ -0,0 +1,202 @@
...
+ template<class FT>
+Complex_number<FT>& Complex_number<FT>::operator=(const Complex_number<FT>& other) {
+  _real = other.real();
+  _imag = other.imag();
+  return *this;
+}
+
+} // namespace CGAL
+
+#endif // CGAL_COMPLEX_WITHOUT_SQRT
```
afabri, line 202:
```suggestion
#endif // CGAL_COMPLEX_NUMBER_H
```

Hyperbolic_surface_triangulation_2/test/Hyperbolic_surface_triangulation_2/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+
+include_directories(../../include/)
+
+# create a target per cppfile
+file(
+  GLOB cppfiles
+  RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
+  ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
+foreach(cppfile ${cppfiles})
+  create_single_source_cgal_program("${cppfile}" ../../include/)
```
afabri, line 19: No idea what happens with the `../../include/`   It is not needed.

Hyperbolic_surface_triangulation_2/demo/Hyperbolic_surface_triangulation_2/window.cpp
```diff
@@ -0,0 +1,248 @@
+// Copyright (c) 2024
+// INRIA Nancy (France), and Université Gustave Eiffel Marne-la-Vallee (France).
+// All rights reserved.
+//
+// This file is part of CGAL (www.cgal.org)
+//
+// $URL$
+// $Id$
+// SPDX-License-Identifier: GPL-3.0-or-later OR LicenseRef-Commercial
```
afabri, line 9: For demo programs we usually do not have a license header.

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,105 @@
...
+
+  /*!
+      returns the \f$ i \f$-th vertex.
+
+      \pre <code> is_valid() </code>
+  */
+  const Point& vertex(int i) const;
+
+  /*!
+      returns the index of the side paired to the \f$ i \f$-th side.
```
afabri, line 57: Just write i-th    that is no backtick no dollar sign.  Mael went through many CGAL files to unify that in [this PR](https://github.com/CGAL/cgal/pull/6849/files)

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/Concepts/ComplexNumber.h
```diff
@@ -0,0 +1,156 @@
...
+    returns the imaginary part.
+  */
+  FT imag() const;
+    /// @}
+
+  /// \name Operations
+  /// @{
+  /*!
+    returns the square of the modulus.
```
afabri, line 74: These two functions  are not member functions. Move them right behind the class and use the  [`\relate` ](https://www.doxygen.nl/manual/commands.html#cmdrelates)command.

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/dependencies
```diff
@@ -0,0 +1,5 @@
+Manual
+Stream_support
+Number_types
+Combinatorial_map
+Hyperbolic_triangulation_2
```
afabri, line 5:
```suggestion
Hyperbolic_triangulation_2
Algebraic_foundations
```
Add this to get a link to `Field`

afabri:
In the traits classes I only see constructions being made when I want Voronoi.    Does that not mean that I could use the `Exact_predicates_inexact_constructions_kernel`?

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/PackageDescription.txt
```diff
@@ -0,0 +1,56 @@
...
+
+\cgalPkgDescriptionEnd
+
+
+\cgalClassifedRefPages
+
+\cgalCRPSection{Concepts}
+
+- `HyperbolicSurfaceTraits_2` is the concept for the template parameter of most classes of the package.
+- `ComplexWithoutSqrt` describes a complex number type that does not use square root.
```
afabri, line 42:
```suggestion
- `ComplexNumber` describes a complex number type that does not use square root.
```

pougetma:
> In the traits classes I only see constructions being made when I want Voronoi. Does that not mean that I could use the `Exact_predicates_inexact_constructions_kernel`?
> 
> In a model of a base concept of the traits class you write for exactness one needs the exact construction kernel with sqrt.

I tried EPIC in the test hs_test_triangulation_EPIC.cpp but I failed to construct a point from an exact  rational ...

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/CGAL/Hyperbolic_surface_traits_2.h
```diff
@@ -9,6 +9,16 @@ namespace CGAL{
 */
 template<class HyperbolicTraitsClass>
 class Hyperbolic_surface_traits_2 : public HyperbolicTraitsClass {
+public:
+  /// \name Types
+  /// @{
+  /*!
+          represents a complex number, model of
+          `ComplexNumber', over the field HyperbolicSurfaceTraits_2::FT for its real and
```
afabri, line 17:
```suggestion
          `ComplexNumber', over the field `HyperbolicSurfaceTraits_2::FT` for its real and
```

sloriot:
Red testsuite in CGAL-6.1-Ic-47

sloriot:
Please check CGAL-6.1-Ic-60 for errors and warnings.

Hyperbolic_surface_triangulation_2/doc/Hyperbolic_surface_triangulation_2/dependencies
```diff
@@ -0,0 +1,6 @@
...
+Algebraic_foundations
```
albert-github, line 6: Add `Periodic_4_hyperbolic_triangulation_2` to, probably, get the link in the documentation:
```
/home/cgal-testsuite/cgal_doc_build/CGAL-6.1-Ic-65/doc/Hyperbolic_surface_triangulation_2/Hyperbolic_surface_triangulation_2.txt:20: warning: unable to resolve reference to 'Chapter_2D_Periodic_Hyperbolic_Triangulations' for \ref command
```
correct

Periodic_4_hyperbolic_triangulation_2/doc/Periodic_4_hyperbolic_triangulation_2/Periodic_4_hyperbolic_triangulation_2.txt
```diff
@@ -17,9 +17,7 @@ namespace CGAL {
 
 
 
-This package enables the computation of Delaunay triangulations of the Bolza
-surface, which is the most symmetric surface of genus 2. The Bolza surface is
-a hyperbolic closed compact orientable surface.
+This package enables the computation of Delaunay triangulations of the Bolza surface, which is the most symmetric surface of genus 2. The Bolza surface is a hyperbolic closed compact orientable surface. For triangulations on general hyperbolic surfaces, we refer to the package \ref Chapter_Hyperbolic_Surface_Triangulations "Hyperbolic Surface Triangulations".
```
albert-github, line 20: Probably need to add `Hyperbolic_surface_triangulation_2` to the dependencies of `Periodic_4_hyperbolic_triangulation_2` to resolve the documentation warning:
```
/home/cgal-testsuite/cgal_doc_build/CGAL-6.1-Ic-65/doc/Periodic_4_hyperbolic_triangulation_2/Periodic_4_hyperbolic_triangulation_2.txt:20: warning: unable to resolve reference to 'Chapter_Hyperbolic_Surface_Triangulations' for \ref command
```

MaelRL:
> When compiling examples or tests, it seems a file is missing: #include <CGAL/license/Triangulation_on_hyperbolic_surface_2.h> @MaelRL : did you forget to commit it?

No, the license file is there: https://github.com/CGAL/cgal/pull/8259/files#diff-74d7fd69497c42e2489e25b8ea8b5796a3debbff0098cb3243b636938b720994

Triangulation_on_hyperbolic_surface_2/demo/Triangulation_on_hyperbolic_surface_2/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+cmake_minimum_required(VERSION 3.1...3.15)
+
+project( Triangulation_on_hyperbolic_surface_2_Demo )
+
+include_directories(/Users/pougetma/dev/cgal-loic-submission-2024/Installation/include/)
```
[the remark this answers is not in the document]
pougetma: ok, I removed links, but dont understand why my cmake cannot find the Installation/include.license dir ??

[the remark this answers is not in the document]
albert-github: No idea, probably something missing in respect to the cmake `include_directories` or in `${CGAL_INCLUDE_DIRS}` or maybe in the `packag.txt`.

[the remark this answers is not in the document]
MaelRL: > ok, I removed links, but dont understand why my cmake cannot find the Installation/include.license dir ??

I merged master now so everything is up to date on other places, can you please try again?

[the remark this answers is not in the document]
pougetma: Everything seems fine, thanks

Triangulation_on_hyperbolic_surface_2/demo/Triangulation_on_hyperbolic_surface_2/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+cmake_minimum_required(VERSION 3.1...3.15)
```
albert-github, line 1: Everywhere else:
```
cmake_minimum_required(VERSION 3.12...3.29)
```
is used, so I think this should used here as well (in all `CMakeLists.txt` of this project)

pougetma: Right, done. Thanks

Triangulation_on_hyperbolic_surface_2/include/CGAL/Hyperbolic_fundamental_domain_2.h
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  // Check that the _pairings vector encodes a perfect matching of the set {0,1,\dots,n-1}
+  std::vector<bool> already_paired(n);
+  for (int k=0; k<n; k++){
+    already_paired[k] = false;
+  }
+  for (int k=0; k<n; k++){
+    std::size_t paired_side = _pairings[k];
+    if ((paired_side<0) || (paired_side>=n)){
```
sloriot, line 173: paired_side is unsigned

sloriot: you store the values as int but seems to always use them as std::size_t

sloriot (changes requested):

afabri:
Runtime errors in [CGAL-6.1-Ic-111](https://cgal.geometryfactory.com/CGAL/testsuite/summary-6.1-Ic-111.html?package=Triangulation_on_hyperbolic_surface_2)

Triangulation_on_hyperbolic_surface_2/demo/Triangulation_on_hyperbolic_surface_2/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+if (NOT CGAL_Qt6_FOUND  OR NOT Qt6_FOUND)
+  message(STATUS "This project requires the Qt6 library, and will not be compiled.")
+  return()
+endif()
+
+# ui files, created with Qt Designer
+qt6_wrap_ui(UIS drawing_window_description.ui)
+
+add_executable( Triangulation_on_hyperbolic_surface_2_demo Triangulation_on_hyperbolic_surface_2_demo.cpp window.cpp ${UIS})
+target_link_libraries(Triangulation_on_hyperbolic_surface_2_demo PRIVATE CGAL::CGAL CGAL::CGAL_Qt6 Qt6::Widgets )
```
lrineau, line 25: The call to `cgal_add_compilation_test` is missing:

```suggestion
add_executable( Triangulation_on_hyperbolic_surface_2_demo Triangulation_on_hyperbolic_surface_2_demo.cpp window.cpp ${UIS})
target_link_libraries(Triangulation_on_hyperbolic_surface_2_demo PRIVATE CGAL::CGAL CGAL::CGAL_Qt6 Qt6::Widgets )
cgal_add_compilation_test(Triangulation_on_hyperbolic_surface_2_demo)
```

That is why the [CGAL testsuite did nothing with `Triangulation_on_hyperbolic_surface_2_Demo` with the last internal release](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.1-Ic-117.shtml#Triangulation_on_hyperbolic_surface_2_Demo).

PR #8265, opened 2024-06-07
CMake - Add one CMakeLists per package

janetournois:
## Summary of Changes

To be able to test a full package (tests, examples, and if they exist benchmark and demo) at once, for example using `ctest` (or simply compiling in visual studio), I suggest to add one CMakeLists per package, at the root of the package.

This PR introduces two such CMakeLists, which I think are very handy.

### DIFF
--- a/Mesh_3/CMakeLists.txt
+++ b/Mesh_3/CMakeLists.txt
@@ -1,0 +1,16 @@
+# This is the CMake script for compiling this folder.
+
+cmake_minimum_required(VERSION 3.1...3.23)
+project(Mesh_3)
+
+if(CGAL_ENABLE_TESTING)
+  enable_testing()
+endif()
+
+option(CGAL_MESH_3_BENCHMARK "Add benchmark to project" OFF)
+if(CGAL_MESH_3_BENCHMARK)
+  add_subdirectory("benchmark/Mesh_3")
+endif()
+
+add_subdirectory("examples/Mesh_3")
+add_subdirectory("test/Mesh_3")
--- a/Poisson_surface_reconstruction_3/CMakeLists.txt
+++ b/Poisson_surface_reconstruction_3/CMakeLists.txt
@@ -1,0 +1,11 @@
+# This is the CMake script for compiling this folder.
+
+cmake_minimum_required(VERSION 3.1...3.23)
+project(Poisson_surface_reconstruction_3)
+
+if(CGAL_ENABLE_TESTING)
+  enable_testing()
+endif()
+
+add_subdirectory("examples/Poisson_surface_reconstruction_3")
+add_subdirectory("test/Poisson_surface_reconstruction_3")


Mesh_3/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# This is the CMake script for compiling this folder.
+
```
lrineau, line 2: Useless comment.
```suggestion
```

Mesh_3/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+cmake_minimum_required(VERSION 3.1...3.23)
+project(Mesh_3)
+
+if(CGAL_ENABLE_TESTING)
+  enable_testing()
+endif()
+
```
lrineau, line 9: Useless, because it will be done in sub-directory, as soon as `CGALConfig.cmake` is found:

`CGALConfig.cmake` includes `CGAL_CreateSingleSourceCGALProgram.cmake`, that includes `CGAL_add_test.cmake`, and the later has:

https://github.com/CGAL/cgal/blob/2c13f1fe6bcb7496b5f5c68095790ae4f894f3dc/Installation/cmake/modules/CGAL_add_test.cmake#L10-L14

So:
```suggestion
```

Mesh_3/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+option(CGAL_MESH_3_BENCHMARK "Add benchmark to project" OFF)
+if(CGAL_MESH_3_BENCHMARK)
+  add_subdirectory("benchmark/Mesh_3")
+endif()
+
+add_subdirectory("examples/Mesh_3")
+add_subdirectory("test/Mesh_3")
```
lrineau, line 16: To make it more copy-pastable:

```suggestion
option(CGAL_MESH_3_BENCHMARK "Add benchmark to project" OFF)
if(CGAL_MESH_3_BENCHMARK)
  add_subdirectory("benchmark/${PROJECT_NAME}")
endif()

add_subdirectory("examples/${PROJECT_NAME}")
add_subdirectory("test/${PROJECT_NAME}")
```

I do not know what to do with benchmarks.

Poisson_surface_reconstruction_3/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+cmake_minimum_required(VERSION 3.1...3.23)
+project(Poisson_surface_reconstruction_3)
+
+if(CGAL_ENABLE_TESTING)
+  enable_testing()
+endif()
+
+add_subdirectory("examples/Poisson_surface_reconstruction_3")
+add_subdirectory("test/Poisson_surface_reconstruction_3")
```
lrineau, line 11:
```suggestion
cmake_minimum_required(VERSION 3.1...3.23)
project(Poisson_surface_reconstruction_3)

add_subdirectory("examples/${PROJECT_NAME}")
add_subdirectory("test/${PROJECT_NAME}")
```

lrineau (commented):
Nice addition.

Note that we will have to do something about those `CMakeFiles.txt` during the construction of the internal release. I add a https://github.com/CGAL/cgal/labels/TODO for that reason.

janetournois:
@lrineau for "benchmark" I suggest [808fa60](https://github.com/CGAL/cgal/pull/8265/commits/808fa60c6c4a841c051cb6cf3893728097c7482f)

sloriot:
Instead of manually creating cmake scripts, can't we create custom targets in case we have `CGAL_ENABLE_TESTING=ON`?

janetournois:
I don't like much that option, because it forces to use the top-level CMakeLists, while the solution I suggested is "package-centered"

PR #8272, opened 2024-06-10
Fix iteration on dynamic bit set

sloriot:
if `ID = unsigned int`, this can lead to an infinite loop


lrineau:
I suggest you add a test for the corresponding issue.

lrineau:
> If `ID == unsigned int`, we had:
> 
> ```
> warning: comparison is always true due to limited range of data type [-Wtype-limits]
> ```

That is what I mean. I suggest to add tests so that PMP is also tested with such inputs: when the ranges of indices is a range of `int` or `unsigned int`. That is quite common, for the graphics community as far as I know.

sloriot:
Successfully tested in CGAL-6.0-Ic-264

PR #8277, opened 2024-06-11
PMP: Change the string of the property

afabri:
## Summary of Changes

When calling[ `compute_vertex_normals()`](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__normal__grp.html#ga74e6b247d7e28beb677076aad5614d8c) we better change the string associated to the property from `"v:normals"`  to `" v:normal"` as otherwise it will not be written by[ `write_PLY()`](https://doc.cgal.org/latest/Surface_mesh/group__PkgSurfaceMeshIOFuncPLY.html#ga50f0e9f2b293855d2c7f1a62939cbe8d).

Note that is not a change in CGAL code, but just in an example (that not even saves the file).

## Release Management

* Affected package(s): PMP

### DIFF
--- a/Polygon_mesh_processing/examples/Polygon_mesh_processing/compute_normals_example.cpp
+++ b/Polygon_mesh_processing/examples/Polygon_mesh_processing/compute_normals_example.cpp
@@ -29,8 +29,8 @@ int main(int argc, char* argv[])
     return 1;
   }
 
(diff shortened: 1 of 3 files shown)


sloriot:
Successfully tested in CGAL-6.0-Ic-265

PR #8284, opened 2024-06-12
[WIP] New package: Frechet Distance

lrineau:
WIP 

------ 

## Summary of Changes

_Describe what your pull request changes to CGAL (this can be skipped if it solves an issue already in the tracker or if it is a Feature or Small Feature submitted to the CGAL Wiki)._

## Release Management

* Affected package(s):


NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
```diff
@@ -63,6 +63,12 @@ struct Cartesian_filter_K : public Base_
     typedef typename Store_kernel<EK_>::reference_type EK_rt;
     EK_rt exact_kernel()const{return sek.kernel();}
 
+
+    // Added to be similar to CGAL::Filtered_kernel
+    typedef EK Exact_kernel;
+    enum { Has_filtered_predicates = true };
+    typedef Boolean_tag<Has_filtered_predicates> Has_filtered_predicates_tag;
+
```
lrineau, line 71: @mglisse please review

NewKernel_d/test/NewKernel_d/Epick_d.cpp
```diff
@@ -745,6 +745,11 @@ static_assert(std::is_same<CGAL::Dimension_tag<3>,Ker3::Dimension>::value);
 static_assert(std::is_same<CGAL::Dynamic_dimension_tag,Kerd::Dimension>::value);
 static_assert(std::is_same<CGAL::Dimension_tag<2>,CGAL::Ambient_dimension<Ker2::Point_d>::type>::value);
 static_assert(std::is_same<CGAL::Dimension_tag<3>,CGAL::Ambient_dimension<Ker3::Point_d,Ker3>::type>::value);
+
+static_assert(Ker2::Has_filtered_predicates_tag::value);
+static_assert(Ker3::Has_filtered_predicates_tag::value);
+static_assert(Kerd::Has_filtered_predicates_tag::value);
+
```
lrineau, line 752: @mglisse please review

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
mglisse: As I said elsewhere, if that's really all that's needed (there is an example so I assume it compiles and runs, although I am rather surprised), ok.
(I would have mentioned in the comment the name of the package that needs it, but I guess this comment is good enough)

NewKernel_d/test/NewKernel_d/Epick_d.cpp
mglisse: ok, we can always modify this if we start testing non-filtered kernels.

afabri:
Working on the dD case I realize that `Difference_of_points_d`  and   `Construct_vector_3`  construct opposite vectors.  The functor `Construct_vector_d`  has no operator for two points, and Kernel_23 has no `Difference_of_points_2/3`.   In which direction shall we unify @mglisse @sloriot ?

mglisse:
> Working on the dD case I realize that `Difference_of_points_d` and `Construct_vector_3` construct opposite vectors. The functor `Construct_vector_d` has no operator for two points, and Kernel_23 has no `Difference_of_points_2/3`. In which direction shall we unify @mglisse @sloriot ?

I am ok with either option.

*Internally (Construct_LA_vector), I don't want to add more overloading.
But in Kernel_d_interface.h, I don't mind turning Construct_vector_d into a wrapper, like we did for Construct_point_d so it could take a weighted point.*

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
```diff
@@ -114,19 +115,25 @@ class KernelD_converter_<Final_,K1,K2,typeset<> > {
 
 
 // TODO: use the intersection of Kn::Object_list.
-template<class K1, class K2, class List_=
-typename typeset_intersection<typename K1::Object_list, typename K2::Object_list>::type
+template<class K1, class K2,
+         class List_= Default,
+         typename NTc = NT_converter<typename Get_type<K1, FT_tag>::type, typename Get_type<K2, FT_tag>::type>
```
mglisse, line 120: IIUC:
* You want to be able to convert from an approximate (Interval_nt) kernel to an exact kernel: can you explain in what conditions this is used? Are the intervals guaranteed to be points, or is any value in the interval ok for your purpose?
* NT_converter does not support conversion from Interval_nt to anything but double: if converting to something else makes sense, shouldn't we specialize NT_converter?
* Instead of passing a functor that converts to the rational type, you pass one that converts to double and rely on an implicit conversion happening afterwards. At first this feels like a hack, but problems seem unlikely (and it might even have slightly better performance than the "clean" solution for some kernels), so that's ok.

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
```diff
@@ -114,19 +115,25 @@ class KernelD_converter_<Final_,K1,K2,typeset<> > {
...
+                      typename typeset_intersection<typename K1::Object_list,
+                                                    typename K2::Object_list>::type>::type List;
         typedef KernelD_converter Self;
         typedef Self Final_;
-        typedef KernelD_converter_<Self,K1,K2,List_> Base;
+        typedef KernelD_converter_<Self,K1,K2,List> Base;
         typedef typename Get_type<K1, FT_tag>::type FT1;
         typedef typename Get_type<K2, FT_tag>::type FT2;
-        typedef NT_converter<FT1, FT2> NTc;
         NTc c; // TODO: compressed storage as this is likely empty and the converter gets passed around (and stored in iterators)
```
mglisse, line 137: If you are motivated, you can replace this old comment with CGAL_NO_UNIQUE_ADDRESS. (you don't have to, I just happened to notice it while looking at the adjacent lines)

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
mglisse: Ok, if the intervals are points, the conversion makes sense. To go back to my second point: what do you think of specializing `NT_converter<Interval_nt,X>` so that
1. it asserts that the interval is a point (CGAL_precondition?)
2. it chains `NT_converter<Interval_nt,double>` (or actually calls `.sup()` directly, no need to average 2 identical values) and `NT_converter<double,X>`

?
Adding a parameter to the kernel converter is not a bad idea, but I think it should only happen if we first reject the simpler solution (maybe you already did and I simply missed earlier discussions).

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_base.h
```diff
@@ -22,6 +22,7 @@ namespace CGAL {
 template < typename FT_, typename Dim_, typename Derived_=Default>
 struct Cartesian_base_d : public CGAL_BASE
 {
+    typedef FT_  FT;
```
mglisse, line 25: Do you still need this, or was it solved by using Kernel_d_interface?

afabri: right. just removed it again.

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
```diff
@@ -63,6 +64,12 @@ struct Cartesian_filter_K : public Base_
     typedef typename Store_kernel<EK_>::reference_type EK_rt;
     EK_rt exact_kernel()const{return sek.kernel();}
 
+
+    // Added to be similar to CGAL::Filtered_kernel
+    typedef Kernel_d_interface<EK> Exact_kernel;
```
mglisse, line 69: Note to @mglisse : think if there is a better place to put this, using Kernel_d_interface at this low level doesn't look right. Maybe in Kernel_d_interface, I can check if there is a type Exact_kernel, and in that case override it with a typedef using Kernel_d_interface (recursively)?

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
afabri: Personally  I find the NT converter as template parameter cleaner than the specialization.  Also I like that in dD it is like in 2D/3D.

mglisse: * Is there any real use of that third parameter in 2D/3D? The only uses I found (not an exhaustive search) were with To_double or To_interval and they look like they would probably work just as well with the default parameter.
* Having a parameter does make it easier to have different versions, for instance this one that expects a trivial interval vs what I did in (the abandoned) Epack_d where I accepted conversions depending on `has_smaller_relative_precision`. So :+1: for the extra parameter.
* I guess this just leaves the comment in https://github.com/CGAL/cgal/pull/8284#discussion_r1765255498 about replacing `NT_converter<distance_t,double>` with something more specific at some point in the future.

Frechet_distance/doc/Frechet_distance/Concepts/FrechetDistanceTraits.h
```diff
@@ -34,5 +38,50 @@ using Kernel = unspecified;
 */
 using Point = unspecified_type;
 
+/*! The number type of the filtered kernel. If
```
albert-github, line 41: `If` ? something missing?

afabri: Better look at it once I ask for review. Sorry for these WIP commits.

mglisse:
> If I get it correctly `Epick_d` has functors that are not in the concept `Kernel_d`. That is they are probably not in `Cartesian_d`.

Yes. The package Triangulation defines for instance a concept DelaunayTriangulationTraits that is not implied by the Kernel_d concept, it has Ep[ie]ck_d as models but not Cartesian_d.

Kernel_23/include/CGAL/Bbox.h
```diff
@@ -0,0 +1,234 @@
...
+        for(int i=0; i<d; ++i)
+        {
+            min_values[i] = -range;
+            max_values[i] =  range;
+        }
+    }
+
+    template <typename I>
+    void init(int d, I b, I e) {
+        CGAL_assertion(d == std::distance(b,e));
```
sloriot, line 146:
```suggestion
        CGAL_USE(e);
        CGAL_assertion(d == std::distance(b,e));
```

afabri:
@mglisse is it acceptable to add the nested type `using Dimension = DimensionTag` to  `Epick_d<DimensionTag` ?

mglisse:
> @mglisse is it acceptable to add the nested type `using Dimension = DimensionTag` to `Epick_d<DimensionTag` ?

I don't understand, with CGAL-6.0, `CGAL::Epick_d<CGAL::Dimension_tag<2>>::Dimension` already exists (same with `Dynamic_dimension_tag`). What do you want to add where?

afabri:
> > @mglisse is it acceptable to add the nested type `using Dimension = DimensionTag` to `Epick_d<DimensionTag` ?
> 
> I don't understand, with CGAL-6.0, `CGAL::Epick_d<CGAL::Dimension_tag<2>>::Dimension` already exists (same with `Dynamic_dimension_tag`). What do you want to add where?

`Dimension` is in the concept, but in the [model ](https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epick__d.html) it is missing in the documentation (not the code) that  `using Dimension = DimensionTag;`

mglisse:
> `Dimension` is in the concept, but in the [model ](https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epick__d.html) it is missing in the documentation (not the code) that `using Dimension = DimensionTag;`

The model documents that it satisfies the concept, and what it has in addition. It doesn't repeat what is in the concept, that would clutter the documentation and make the differences hard to see. At least that's the choice I made when I wrote this doc. Adding just `Dimension` would be strange, if you really want to add it I think you should copy everything else from the concept at the same time, although I would not recommend that.

afabri:
> > `Dimension` is in the concept, but in the [model ](https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epick__d.html) it is missing in the documentation (not the code) that `using Dimension = DimensionTag;`
> 
> The model documents that it satisfies the concept, and what it has in addition. It doesn't repeat what is in the concept, that would clutter the documentation and make the differences hard to see. At least that's the choice I made when I wrote this doc. Adding just `Dimension` would be strange, if you really want to add it I think you should copy everything else from the concept at the same time, although I would not recommend that.

But when writing `using Dimension = unspecified_type;`   don't we then need a concept for `Dimension`. And even in the model don't  we have to make explicit that `Dimension` will be what is passed in as `DimensionTag`.

Frechet_distance/include/CGAL/Frechet_distance.h
```diff
@@ -27,6 +28,61 @@
 namespace CGAL
 {
 
+#ifdef CGAL_EIGEN3_ENABLED
+template<class Dim> struct Epick_d;
+template<class Dim> struct Epeck_d;
+
+template < typename P, int D>
+struct Ambient_dimension<P, Epick_d<Dimension_tag<D>>>
```
afabri, line 36: Should that go into `Epick_d.h`    @sloriot  ?

sloriot: @mglisse ?

mglisse: I guess we could add a template Ambient_dimension inside the class Kernel_d_interface that says everything has Base::Dimension as ambient dim, since that's what CGAL::Ambient_dimension looks for by default.

Frechet_distance/include/CGAL/Frechet_distance.h
```diff
@@ -0,0 +1,207 @@
...
+ * \param distance_bound the distance to compare against
+ * \param np an optional sequence of \ref bgl_namedparameters "Named Parameters" containing the one listed below:
+ *
+ * \cgalNamedParamsBegin
+ *   \cgalParamNBegin{geom_traits}
+ *     \cgalParamDescription{an instance of a geometric traits class}
+ *     \cgalParamType{a model of `FrechetDistanceTraits`}
+ *     \cgalParamDefault{`Frechet_distance_traits_2`, `Frechet_distance_traits_3`, or `Frechet_distance_traits_d`, depending on the dimension of the point type
+ *                       deduced from the point type, using `Kernel_traits`.}
+ *     \cgalParamExtra{The input point type (the range's value type) must be equal to the traits' `Point_d` type.
```
MaelRL, line 104:
```suggestion
 *     \cgalParamExtra{The input point type (the range's value type) must be equal to the traits' `Point_d` type.}
```

afabri:
Why is[ this](https://github.com/CGAL/cgal/blob/249a6ac47edf7f40d53667c5d6c21cb7e942c9e5/NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Point_d.h#L52)  `typedef` in the class `Point_d`  a comment and not made available, @mglisse , @sloriot ?

mglisse:
> Why is[ this](https://github.com/CGAL/cgal/blob/249a6ac47edf7f40d53667c5d6c21cb7e942c9e5/NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Point_d.h#L52) `typedef` in the class `Point_d` a comment and not made available, @mglisse , @sloriot ?

I don't remember, but I would guess because it wasn't really needed. We could add it though, under the condition that you don't use it :wink: .

> In fact when I uncomment it, and if I choose as dimension 4 instead of 2 examples/Frechet_DS_d.cpp compiles.

(I assume it doesn't compile without the change)
That's suspicious. You should be getting this typedef from the traits class, not from the point type, no?

afabri:
> > Why is[ this](https://github.com/CGAL/cgal/blob/249a6ac47edf7f40d53667c5d6c21cb7e942c9e5/NewKernel_d/include/CGAL/NewKernel_d/Wrapper/Point_d.h#L52) `typedef` in the class `Point_d` a comment and not made available, @mglisse , @sloriot ?
> That's suspicious. You should be getting this typedef from the traits class, not from the point type, no?

Good point. Fixed.

afabri:
@sloriot can you help me with [this error](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-99/Frechet_distance/TestReport_Christo_MSVC2017-Release-64bits.gz) with VC2017

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+are uncertain.  This means for `Simple_cartesian<double>` that there are no guarantees, for
+`Simple_cartesian<Exact_rational>` that all computations are performed with exact rationals,
+and for kernels such as `Exact_predicates_inexact_constructions_kernel`,  `Exact_predicates_exact_constructions_kernel`,
+as well as `Epick_d` and `Epeck_d`, the computation is filtered and hence both fast and guaranteed to be correct.
+
+
+\section secFrechetDistancePerformance Performance
+
+The performance of the algorithm is quadratic in the number of vertices in the worst case.
+For all kernels with filtered predicates the performance is roughly the same, the overhead of using interval arithmetic is negligible.
```
MaelRL, line 65:
```suggestion
Kernels with filtered predicates have roughly the same performance as others: the overhead of using interval arithmetic is negligible.
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+The algorithm is not concerned by the <em>curse of dimensionality</em>.
+The running time increases when the curves are closer or when the bound for the approximation error becomes smaller.
+For a very detailed analysis in practice we refer to the journal paper entitled <em>Walking the Dog Fast in Practice: Algorithm Engineering of the Fréchet Distance</em>
```
MaelRL, line 68:
```suggestion
For a very detailed analysis in practice, we refer to the journal paper entitled <em>Walking the Dog Fast in Practice: Algorithm Engineering of the Fréchet Distance</em>
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+\cgalCite{cgal:bkn-wdfp-21}.
+
+\section secFrechetDistanceImplementation Implementation
+
+The algorithms in this package are an adaption of the implementation devised by Bringmann et al. (\cgalCite{cgal:bkn-wdfp-19}
+and \cgalCite{cgal:bkn-wdfp-21}). In particular, the implementation can decide non-difficult cases very
+fast while for difficult cases it still has the quadratic running time guarantee of the classical Fréchet
+distance algorithm by Alt and Godau \cgalCite{ag-cfdbt-95}. This is achieved by using fast bounding
```
MaelRL, line 76:
```suggestion
quickly, while it still has the quadratic running time guarantee of the classical Fréchet distance algorithm
by Alt and Godau \cgalCite{ag-cfdbt-95} for difficult cases. This is achieved by using fast bounding
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+box filtering methods and a divide and conquer algorithm with pruning rules on the free-space diagram.
```
MaelRL, line 77:
```suggestion
box filtering methods and a divide-and-conquer algorithm with pruning rules on the free-space diagram.
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+
+
+\section secFrechetDistanceExamples Examples
+
+In the examples we use different kernels to illustrate that the functions
```
MaelRL, line 82:
```suggestion
In the examples, we use different kernels to illustrate that the functions
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+can be used with either inexact or exact kernels.
+
+\subsection subsecFrechetDistanceFirstExample Decision for 2D Polylines
+
+The following example shows how we can use `is_Frechet_distance_larger()`
+to determines if the Frechet distance between two polylines is larger than a given distance bound.
```
MaelRL, line 88:
```suggestion
to determine if the Frechet distance between two polylines is larger than a given distance bound.
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+
+\cgalExample{Frechet_distance/Frechet_distance_2.cpp}
+
+\subsection subsecFrechetDistanceSecondExample Distance Computation for 3D Polylines
+
+The following example shows how we can compute the Fréchet distance up to a given error bound on two
+polylines in 3-dimensional Euclidean space using `bounded_error_Frechet_distance()`.
+As we use `Simple_cartesian<double>` the algorithm uses plain floating point arithmetic
```
MaelRL, line 96:
```suggestion
As we use `Simple_cartesian<double>`, the algorithm uses plain floating point arithmetic
```

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+\subsection subsecFrechetDistanceImageCredits  Image Credits
+
+The character image is a visualization of two data points from the <a href="https://archive.ics.uci.edu/dataset/175/character+trajectories">Character Trajectories</a> data set.
+
+\subsection subsecFrechetDistanceImplementation  Implementation History
+
+An initial version using floating point arithmetic was developed by the authors
+while working at the Max-Planck Institute for Informatics in Saarbrücken, Germany.
+André Nusser, together with Sebastien Loriot and Andreas Fabri, introduced
+the usage of interval arithmetic and square root extensions to alleviate issues stemming from rounding errors and hence ensuring correctness of the computation.
```
MaelRL, line 139:
```suggestion
the usage of interval arithmetic and square root extensions to alleviate issues stemming from rounding errors and hence ensure correctness of the computation.
```

Frechet_distance/examples/Frechet_distance/Frechet_DS_d.cpp
```diff
@@ -0,0 +1,57 @@
...
+using Kernel = CGAL::Epick_d<CGAL::Dimension_tag<4>>;
+using Traits = CGAL::Frechet_distance_traits_d<Kernel>;
+using Point = Traits::Point_d;
+using Curve = std::vector<Point>;
+using Curves = std::vector<Curve>;
+
+int main()
+{
+    Curves curves;
+#if 0
```
MaelRL, line 20: "if 0" in examples

Frechet_distance/doc/Frechet_distance/Frechet_distance.txt
```diff
@@ -0,0 +1,142 @@
...
+\section secFrechetDistanceAPI API
+
+The main entry points are the following functions and classes:
+- The function `bounded_error_Frechet_distance()` computes an approximation of the Fréchet distance between two polylines, up to a given approximation error.
+  It returns an interval that contains the true distance.
+- The function `is_Frechet_distance_larger()` decides if the Fréchet distance between two polylines is larger than a given bound.
+- The class `Frechet_distance::Neighbor_search` stores a set of polylines and offers a query functions to find polylines closer to a query curve than a given bound.
+
+Both functions and the class have as template parameter a traits class defining the dimension and the point type.
+The traits classes have as template parameter a kernel, model of the concept `Kernel`.
```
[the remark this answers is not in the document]
afabri: Why do you suggest to remove this?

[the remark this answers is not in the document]
MaelRL: it is not true for _d

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
afabri: Why are you surprised that it compiles?

mglisse: There have been a number of changes since then (I think other pieces were missing at the time), so the comment may not be relevant anymore.

Frechet_distance/include/CGAL/Frechet_distance/internal/Frechet_distance.h
```diff
@@ -0,0 +1,193 @@
...
+          using AT = Frechet_distance_traits_3<Simple_cartesian<Interval_nt_advanced>>;
+          using ET = Frechet_distance_traits_3<Simple_cartesian<Exact_rational>>;
+
+          using Filtered_traits = std::pair<AT,ET>;
+
+          return Curve<Filtered_traits, true>(point_range, traits);
+        }
+        else
+        {
+          //TODO: not implemented
```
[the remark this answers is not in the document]
anusser: I honestly don't know exactly what's going on there. We can have a call if you want to discuss this directly.

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
sloriot: Hi @mglisse, any chance you have some time to do the right fix? Also Epeck_d is currently not considered as filtered. What shall we do in Lazy_cartesian.h?

Frechet_distance/test/Frechet_distance/Frechet_distance_test.cpp
```diff
@@ -0,0 +1,344 @@
...
+    using Traits = CGAL::Frechet_distance_traits_d<Kernel>;
+    using Point = Kernel::Point_d;
+    std::cout <<"CGAL::Epick_d\n";
+    double t1=Test_struct<Kernel, Traits, Point>::testFrechetDistance();
+    std::cout << t1 << "\n";
+  }
+
+  if (test_set.empty() || test_set.count(8))
+  {
+    using Kernel = CGAL::Epeck_d<CGAL::Dimension_tag<100>>;
```
mglisse, line 335: If I replace 100 with 3, it fails to compile (interaction between `Bbox_3` and `Bbox_d<Dimension_tag<3>>`).
Curiously, the Epick_d version does compile with dim 3, although it gives an (expected) assertion failure at runtime because the file format differs.

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
mglisse: https://github.com/mglisse/cgal/commits/frechet/ (commit https://github.com/mglisse/cgal/commit/c5b2721c5fd616d18fe963c9caa760f504919930) would keep Cartesian_filter "purer" (not knowing anything about the Kernel_d interface). I hesitated about sticking to the names AK and EK internally, and use the longer Approximate_kernel/Exact_kernel only for those with a Kernel_d interface, that's still a possibility. It would also be possible to key it on something else than the existence of the 2 types (maybe the `Has_filtered_predicates_tag` you added). Using internals this way is bound to be fragile anyway...

For Epeck_d, if I add Has_filtered_predicates(_tag), it does compile and is recognized as filtered, but it does not help with the running time of Frechet_distance_test.cpp, it spends most of its time doing Lazy_exact_nt computations. I did not check what is done for Epeck, the first thing to try would be Epeck_d in dim 3 for a fair comparison.

sloriot:
Compilation error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-136/Frechet_distance/TestReport_Christo_MSVC2017-Release-64bits.gz)

afabri:
> Compilation error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-136/Frechet_distance/TestReport_Christo_MSVC2017-Release-64bits.gz)

With `/permissive-`  a `constexpr`  used to conditionally compile is ignored, that is the compiler sees a block that it should ignore.
We removed   `/permissive-`  on the VC 2017 test platform.

Frechet_distance/test/Frechet_distance/Frechet_distance_test.cpp
afabri: This is now fixed.

Frechet_distance/test/Frechet_distance/Frechet_distance_test.cpp
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  if (test_set.empty() || test_set.count(3))
+  {
+    using Kernel = CGAL::Exact_predicates_exact_constructions_kernel_with_sqrt;
+    using Traits = CGAL::Frechet_distance_traits_2<Kernel>;
+    using Point = Kernel::Point_2;
+    std::cout <<"Exact_predicates_exact_constructions_kernel_with_sqrt (force filtering)\n";
+    double t1=Test_struct<Kernel, Traits, Point>::testFrechetDistanceNearNeighborsDS();
+    double t2=Test_struct<Kernel, Traits, Point>::testFrechetDistance<true>();
```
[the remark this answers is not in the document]
afabri: `Exact_predicates_exact_constructions_kernel_with_sqrt`  is a `Simple_cartesian<CORE::Expr>`  for which we do not do any filtering.

afabri:
@sloriot compared the performance of the first commit at hash ad109d217a2b1bf2e785c131eeabc6f274661825
with `Simple_cartesian<double>` the latter with and without filtering. 
With filtering the same performance as the initial version, and without filtering even faster than the initial version.
So with filtering is satisfying as it is certified, whereas without filtering we have no guarantee.
`Epick` and `Epeck` with filtering have the same performance.   For `Epeck_d` it is not yet completely satisfying, but that should not be blocking.

NewKernel_d/include/CGAL/NewKernel_d/Cartesian_filter_K.h
afabri: I just pushed what you suggested. Thank's a lot.

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
afabri: @mglisse I am willing to do something here, but I have no idea what exactly.  Can you give me the recipe?

mglisse: Maybe just
```diff
-        NTc c; // TODO: compressed storage as this is likely empty and the converter gets passed around (and stored in iterators)
+        CGAL_NO_UNIQUE_ADDRESS NTc c;
```

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
```diff
@@ -1170,6 +1193,41 @@ template<class R_> struct Compare_lexicographically : private Store_kernel<R_> {
...
+namespace CartesianDKernelFunctors {
+template<class R_> struct Compare_squared_distance : private Store_kernel<R_> {
+        CGAL_FUNCTOR_INIT_STORE(Compare_squared_distance)
+        typedef R_ R;
+        typedef typename Get_type<R, RT_tag>::type RT;
+        typedef typename Get_type<R, Comparison_result_tag>::type result_type;
+        typedef typename Get_functor<R, Construct_ttag<Point_cartesian_const_iterator_tag> >::type CI;
+        // TODO: This is_exact thing should be reengineered.
+        // the goal is to have a way to tell: don't filter this
+        typedef typename CGAL::Uses_no_arithmetic<CI> Uses_no_arithmetic;
```
mglisse, line 1205: Please remove those 3 lines (there are arithmetic operations in this function).

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
```diff
@@ -1170,6 +1193,41 @@ template<class R_> struct Compare_lexicographically : private Store_kernel<R_> {
...
+                auto b_begin=ci(b,Begin_tag());
+                auto a_end=ci(a,End_tag());
+                RT sqdist(0);
+                result_type res;
+                // can't we do slightly better for Uncertain<*> ?
+                // after res=...; if(is_uncertain(res))return indeterminate<result_type>();
+                do{
+                  RT d = (*a_begin++ - *b_begin++);
+                  sqdist += d*d;
+                }while(a_begin!=a_end);
```
mglisse, line 1221: If you decided to compare sqdist to c after each iteration of the loop (just a random example, not a suggestion), it would make sense to have special code. But if you just compute the squared distance, why not to use `typedef typename Get_functor<R, Squared_distance_tag>::type CSD;` like `Compare_distance` does?

afabri: changed.

Frechet_distance/include/CGAL/Frechet_distance/internal/curve.h
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        this->init();
+    }
+
+    Rational_point rpoint(PointID const& i) const
+    {
+      if constexpr (Approximate_traits::Dimension::value<=3)
+      {
+        // TODO: this assumes that input interval are all tight --> need a PM!
```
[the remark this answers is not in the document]
afabri: We came to the conclusion that `Epeck` points as result of a construction cannot be a problem here.

NewKernel_d/include/CGAL/NewKernel_d/KernelD_converter.h
afabri: done as suggested.

sloriot:

NewKernel_d/include/CGAL/NewKernel_d/function_objects_cartesian.h
mglisse: > I removed `CGAL_NTS`

Meanwhile, #8921 is adding `::CGAL::` (aka `CGAL_NTS`) in front of every `compare` it can find.
(I don't know which one is right, just pointing out the inconsistency)

PR #8291, opened 2024-06-18
Improve generation of polylines

sloriot:
We now have several options:
- 2d
- triangulate (kept even if it's not really a polyline)
- polygon
- one per line (within the same item)

image


sloriot:
Successfully tested in CGAL-6.0-Ic-274

PR #8314, opened 2024-06-27
Shape_detection: using proper triangulated polygonal faces for linear_least_squares

### LINKED ISSUE #7992: Least_squares_plane_fit_region is not using PCA
`CGAL::Shape_detection::Polygon_mesh::Least_squares_plane_fit_region` says in the doc that it uses PCA but that's not the case.

From a quick look only centroids are used for a point on the plane and normals are only evaluated using 3 consecutive points of a face (which is not a good idea for non triangular faces).

soesau:
## Summary of Changes

Polygonal faces are now triangulated and face normals are calculated using PMP::compute_face_normal.

The calculated face normals and triangulations are buffered in Least_squares_plane_fit_region. However, Least_squares_plane_fit_sorting is independent and does not benefit from that buffered data.

## Release Management

* Issue(s) solved (if any): fix #7992

### DIFF
--- a/Shape_detection/include/CGAL/Shape_detection/Region_growing/Polygon_mesh/Least_squares_plane_fit_region.h
+++ b/Shape_detection/include/CGAL/Shape_detection/Region_growing/Polygon_mesh/Least_squares_plane_fit_region.h
@@ -93,9 +93,6 @@ namespace Polygon_mesh {
     using Scalar_product_3 = typename GeomTraits::Compute_scalar_product_3;
     using Cross_product_3 = typename GeomTraits::Construct_cross_product_vector_3;
 
-    std::unordered_map<Item, Vector_3, internal::hash_item<Item> > m_face_normals;
-    std::unordered_map<Item, std::vector<Triangle_3>, internal::hash_item<Item> > m_face_triangulations;
-
   public:
     /// \name Initialization
     /// @{
@@ -163,7 +160,10 @@ namespace Polygon_mesh {
     m_squared_length_3(m_traits.compute_squared_length_3_object()),
     m_squared_distance_3(m_traits.compute_squared_distance_3_object()),
     m_scalar_product_3(m_traits.compute_scalar_product_3_object()),
-    m_cross_product_3(m_traits.construct_cross_product_vector_3_object()) {
+    m_cross_product_3(m_traits.construct_cross_product_vector_3_object()),
+    m_face_normals( get(CGAL::dynamic_face_property_t<Vector_3>(), pmesh) ),
+    m_face_triangulations( get(CGAL::dynamic_face_property_t<std::vector<Triangle_3>>(), pmesh) )
+    {
 
 #ifdef CGAL_SD_RG_USE_PMP
     auto get_face_normal = [this](Item face, const PolygonMesh& pmesh)
@@ -194,7 +194,7 @@ namespace Polygon_mesh {
 #endif
 
       for (const Item &i : faces(pmesh)) {
-        m_face_normals[i] = get_face_normal(i, pmesh);
+        put(m_face_normals, i, get_face_normal(i, pmesh));
         std::vector<Point_3> pts;
         auto h = halfedge(i, pmesh);
         auto s = h;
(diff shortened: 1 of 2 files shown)


lrineau:
@afabri @sloriot @MaelRL: This bug-fix pull-request adds new dependencies to non-trivial parts of PMP. For the license checking, that means a user of CGAL-5.6.1 might have a new license issue after an update to CGAL-5.6.2. That does not seem right.

soesau:
Sebastien motivated yesterday having a basic triangulation for exactly cases like this.
Here a simple triangulation of simple polygons is required. Using PMP::triangulate_hole_polyline creates all those mentioned dependencies.

Shape_detection/include/CGAL/Shape_detection/Region_growing/internal/utils.h
```diff
@@ -395,6 +397,58 @@ namespace internal {
...
   }
 
+  template<
+    typename Traits,
+    typename FaceGraph,
+    typename Region,
+    typename FaceToTrianglesMap>
+  std::pair<typename Traits::Plane_3, typename Traits::FT>
+    create_plane_from_triangulated_faces(
+      const FaceGraph& face_graph,
```
sloriot, line 407: ` warning: unused parameter 'face_graph'`

Shape_detection/include/CGAL/Shape_detection/Region_growing/internal/utils.h
```diff
@@ -395,6 +397,58 @@ namespace internal {
...
+    std::vector<ITriangle_3> triangles;
+    CGAL_precondition(region.size() > 0);
+    triangles.reserve(region.size());
+    const IConverter iconverter = IConverter();
+
+    for (const typename Region::value_type face : region) {
+      const std::vector<Triangle_3>& tris = face_to_triangles_map.at(face);
+      CGAL_precondition(tris.size() > 0);
+
+      for (const auto tri : tris)
```
sloriot, line 432: `/mnt/testsuite/include/CGAL/Shape_detection/Region_growing/internal/utils.h:432:23: warning: loop variable 'tri' creates a copy from type 'const CGAL::Triangle_3<CGAL::Simple_cartesian<double> >' [-Wrange-loop-construct]`

sloriot:
runtime error of `test_region_growing_on_degenerated_mesh`for shape detection in  CGAL-6.0-Ic-279

sloriot:
Successfully tested in CGAL-6.0-Ic-282

afabri:
Can we not identify three points which are maximal un-collinear, by computing Eigen values/vectors ?

Shape_detection/include/CGAL/Shape_detection/Region_growing/Polygon_mesh/Least_squares_plane_fit_region.h
```diff
@@ -383,25 +415,6 @@ namespace Polygon_mesh {
       return Point_3(x, y, z);
     }
 
-    // Compute normal of the face.
-    template<typename Face>
-    Vector_3 get_face_normal(const Face& face) const {
```
sloriot, line 388: I think the issue encountered by @lrineau would be fixed by only fixing this function. We should look for 3 non-collinear vertices to get a normal. I would then skip the proper 2D CDT and keep the centroid triangulation for the plane fitting (only matters for non exactly coplanar points anyway). @afabri what do you think?

afabri: I also thought that for computing a face normal it is enough to find 3 sufficiently non-collinear points, but Sven told  me that for non-convex faces he has to compute some weight and for that he needs a triangulation.  We have to check with him.

sloriot:
Successfully tested in CGAL-6.0-Ic-307

Shape_detection/include/CGAL/Shape_detection/Region_growing/Polygon_mesh/Least_squares_plane_fit_region.h
soesau: These were two separate problems:
1. get_face_normal is used for fitting a plane to a single face and for checking whether a face fits to an adjacent region. This was likely the issue encountered by @lrineau.
2. PCA was not using triangulated faces.

lrineau:
I have removed the label https://github.com/CGAL/cgal/labels/Not%20yet%20approved.

I hope this merge PR was eventually approved. The issue I pointed out was the addition of dependencies: https://github.com/CGAL/cgal/pull/8314#issuecomment-2195684005.

PR #8341, opened 2024-07-05
Fix initialization for partial hedge range

sloriot:
Fix bug introduced #7712


sloriot:
Successfully tested in CGAL-6.0-Ic-286

PR #8365, opened 2024-07-22
Workaround missing links

sloriot:


albert-github:
Note the problem looks like to be present in BGL/doc/BGL/BGL.txt as well.

albert-github:
I was still a bit intrigued by the problem and it looks like doxygen gets a bit confused due to the, undocumented, definitions at the lines 37 and 122. The definition at line 188 is documented.
Would it be possible to either:
- move the function (plus its documentation) from line 188 before the functions on the lines 37 and 122
- put the functions from the lines 37 and 122 in a `#ifndef DOXYGEN_RUNING ... #endif` block
  (non related side note the `DOXYGEN_RUNNING` is defined in `BaseDoxyfile.in` but also in `Kinetic_space_partition/Doxyfile.in` and `Kinetic_surface_reconstruction/Doxyfile.in` I don't think the later 2 definitions shouldn't be there)

sloriot:
Successfully tested in CGAL-6.0-Ic-294

PR #8371, opened 2024-07-22
fix copy/paste error

### LINKED ISSUE #8363: draw_nef_3.cpp demo does not work with two passed filenames
## Issue Details

because of index error P1 and P2 get read from same file, and N1-N2 results in empty display always.

## Source Code

https://github.com/CGAL/cgal/blob/master/Nef_3/examples/Nef_3/draw_nef_3.cpp#L19
```
  std::ifstream ifs2((argc > 1) ? argv[1] : CGAL::data_file_path("meshes/beam.off"));
```
needs to be
```
  std::ifstream ifs2((argc > 2) ? argv[2] : CGAL::data_file_path("meshes/beam.off"));
```
  
## Environment

* Compiler: g++
* Specific flags used (if any): see CGAL cmake file
* CGAL version: 5.6.1
* Boost version:

sloriot:
Fixes #8363


sloriot:
Successfully tested in CGAL-6.0-Ic-294

PR #8378, opened 2024-07-25
Tet remeshing - remove assertion in Adaptive sizing field

janetournois:
## Summary of Changes

The assertion is invalid (at least) when the function takes a far vertex, inserted by parallel version of Mesh_3.

## Release Management

* Affected package(s): Mesh_3


sloriot:
Can't you use something like `v->is_far() || !edges.empty()`

janetournois:
The input triangulation can come from any user code. So there can be more vertices with more different dimensions in the input.

Also, we can think of more cases, depending on how `cell_selector` is defined. For example, take a vertex of dimension `2` that is not part of the selection, then it has no incident edges that will be pushed in `edges`, and again the assertion is wrong.

sloriot:
Successfully tested in CGAL-6.0-Ic-300

PR #8380, opened 2024-07-26
Speed up copy_face_graph

sloriot:
Mostly when tm is not empty and almost make it less dependent on tm size

<s>Should probably be backported</s>


sloriot:
Successfully tested in CGAL-6.0-Ic-300

afabri:
> Mostly when tm is not empty and almost make it less dependent on tm size
> 
> Should probably be backported

please backport, unless it is non-trivial

sloriot:
Successfully tested in CGAL-6.0-Ic-307

PR #8398, opened 2024-08-06
Fix for Line_3 Ray_3 do_intersect

MaelRL:
## Summary of Changes

Some non intersection is wrongfully detected as an intersection for the Line_3 Ray_3 combination. Here is an illustration:

image

The base code is :
```cpp
  if(!do_intersect(l, r.supporting_line()))
    return false;

  typename K::Coplanar_orientation_3 pred = k.coplanar_orientation_3_object();
  CGAL::Orientation p0p1s = pred(l.point(0), l.point(1), r.source());
  if(p0p1s == COLLINEAR)
    return true;

  CGAL::Orientation stp0 = pred(r.source(), r.second_point(), l.point(0));
  if(stp0 == COLLINEAR)
    return Ray_3_has_on_collinear_Point_3(r,l.point(0),k);

  return (p0p1s != stp0);
```

The supporting line of the ray intersects the line, and then nothing is collinear, so we check L0,L1,R1 vs R0,R1,L0, which are of opposite sign so there is an intersection... but there is none.

This check is very similar to what is done in Ray_3-Ray_3, where it is correct.

I did a naive fix for now: without loss of generality we can go to the configuration below

image

and so we have an intersection as long as the ray "points" in the other slab wrt to the line. And so, in coplanar orientation checks, if the orientation L0,L1,R0 differs from L0, L1, (L0 + (R1-R0)).

I feel like I could probably skip at least a translation by having something like Coplanar_orientation(P3, P3, P3, V3), and maybe even more...

## Release Management

* Affected package(s): `Intersections_3`
* Issue(s) solved (if any): -

### DIFF
--- a/Intersections_3/include/CGAL/Intersections_3/internal/Line_3_Ray_3_do_intersect.h
+++ b/Intersections_3/include/CGAL/Intersections_3/internal/Line_3_Ray_3_do_intersect.h
@@ -38,11 +38,8 @@ do_intersect(const typename K::Line_3& l,
   if(p0p1s == COLLINEAR)
     return true;
 
-  CGAL::Orientation stp0 = pred(r.source(), r.second_point(), l.point(0));
-  if(stp0 == COLLINEAR)
-    return Ray_3_has_on_collinear_Point_3(r,l.point(0),k);
-
-  return (p0p1s != stp0);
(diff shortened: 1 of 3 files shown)


afabri:
What about backporting this?

sloriot (approved):

Issue #781, opened 2016-02-14
CGAL::Assertion_exception when refining mesh


petersadro:
I am using the default CDT mesh refiner.  I sometimes hit an assertion when refining.

I am running tip of tree CGAL as of Feb 12 - SHA c624d3d835afe666885e32e234f9209b88f208f5

terminate called after throwing an instance of 'CGAL::Assertion_exception'
  what():  CGAL ERROR: assertion violation!
Expr: n == zone.fh
/usr/local/include/CGAL/Mesh_2/Refine_edges.h
Line: 434
Aborted (core dumped)

I've created a gist (https://gist.github.com/petersadro/55470eccde4ce8df3684) with the backtrace, example code, and data to reproduce the issue.

Sorry for the large amount of data.  I actually pared down the free points to the minimum to still reproduce the problem.  Removing the constraints is problematic.

Thanks,
Pete

lrineau:
It seems you create the mesher object and then use it without passing any criteria object. That is almost certainly the reason why the mesher loops endlessly until the local size of triangles is so small that the geometric constructions are buggy.

petersadro:
Hello,

Sorry for not getting back to this sooner.  I really only have time on the weekend.
The original program did have a criteria attached ( when the data was generated), just my test program left it out.  I did have the typedef, though...

I've verified the test program still crashes with the criteria set, and have added this new test source code to the above gist.  The test data is the same.

Thanks,
Pete

lrineau:
@afabri That may be related to the new constraints hierarchy. Is there an easy way to reuse the old implementation?

afabri:
I can reproduce the assertion.  Note however that `num_constraints` is much bigger than  the number of constraints that follows.  If I only insert the constraints that are in the file the mesher finishes.
I still should find out why there is an assertion, but maybe my remark is a quick fix for your problem.

petersadro:
Ah, when the constraints are written to the file, I have them in a vector.  I use size to write numconstraints.  When writing the constraints, I check for source==target, and drop if true.

afabri:
In new_tritest.cpp  it goes wrong for the segment with coordinates  ((0.5, 0.5), (0.5, 0.5))  where 0.5 is the last number in the data file.

petersadro:
It looks to me like my data file is truncated.
Let me try again.

afabri:
[...]

PR #8429, opened 2024-08-23
Add minimal program for Issue #781

afabri:
## Summary of Changes

Added the minimal test case that shows the error reported in Issue #781

Todo:  Fix the bug. @lrineau proposed something (see disussion in the other issue)

## Release Management

* Affected package(s): Mesh_2
* Issue(s) solved (if any): 

### DIFF
--- a/Mesh_2/test/Mesh_2/issue_781.cpp
+++ b/Mesh_2/test/Mesh_2/issue_781.cpp
@@ -31,8 +31,6 @@ int main(int argc, char* argv[])
 
     cdt.insert( pt );
 
-
-    meshTriPoint s(5.4688567999999904, 44.256729199999995);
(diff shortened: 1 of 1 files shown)


afabri:
The input is a segment and a point close to the segment. So the domain is just a triangle.   The first refinement point is the midpoint of an edge, and it is outside of the domain. 
The assertion is triggered in `Refine_edges_base::conflicts_zone_impl()`, namely in[ Line 429](https://github.com/CGAL/cgal/blob/6b366985e18fba5d3694d76e2c67a1a8f0c0d268/Mesh_2/include/CGAL/Mesh_2/Refine_edges.h#L429)

MaelRL:
Replaced by https://github.com/CGAL/cgal/pull/9290.

PR #8433, opened 2024-08-25
Fix exception double free in Convex decomposition

### LINKED ISSUE #422: convex_decomposition_3: Uncatchable exception causes crash after copying Nef_3
This looks like a bug related to taking a copy of a Nef_polyhedron_3.
To reproduce, I've edited one of the built-in examples (https://github.com/CGAL/cgal/blob/master/Convex_decomposition_3/examples/Convex_decomposition_3/list_of_convex_parts.cpp). See below for my code. I'm essentially taking a copy of the read Nef_polyhedron_3 before passing the copy to _convex_decomposition_3_.

**Issue A**: This causes an assertion failure. The assertion failure only happens if I make a copy of the Nef_polyhedron_3.

**Issue B:** The assertion isn't catchable, because the Nef_polyhedron_3 _destructor_ crashes. It could be that the object is left in an unrecoverable state (which isn't very nice as it forces me to let my app crash).

The test data: https://raw.githubusercontent.com/openscad/openscad/openscad-2019.01-RC1/cgal/data/issue1455.nef3

FYI: The nef3 example is generated based on user input to OpenSCAD. My goal is to shield users from having their process crash due to feeding corner cases into the software.

Edit: Tested on Mac OS X using CGAL-4.6.3.

list_of_convex_parts.cpp:

```
#include <CGAL/IO/Nef_polyhedron_iostream_3.h>
#include <CGAL/Nef_3/SNC_indexed_items.h>
#include <CGAL/convex_decomposition_3.h> 
#include <list>
#include <CGAL/assertions_behaviour.h>

typedef CGAL::Polyhedron_3<Kernel> Polyhedron_3;
typedef CGAL::Nef_polyhedron_3<Kernel, CGAL::SNC_indexed_items> Nef_polyhedron_3;
typedef Nef_polyhedron_3::Volume_const_iterator Volume_const_iterator;

int main() {

  CGAL::set_error_behaviour(CGAL::THROW_EXCEPTION);
#if 1
  Nef_polyhedron_3 tmp;
  std::cin >> tmp;
  Nef_polyhedron_3 N(tmp);
#else
  Nef_polyhedron_3 N;
  std::cin >> N;
#endif

  try {
    CGAL::convex_decomposition_3(N);
  }
  catch (...) {
    std::cerr << "Assertion failed\n";
    return 1;
  }
  std::list<Polyhedron_3> convex_parts;

  // the first volume is the outer volume, which is 
  // ignored in the decomposition
  Volume_const_iterator ci = ++N.volumes_begin();
  for( ; ci != N.volumes_end(); ++ci) {
    if(ci->mark()) {
      Polyhedron_3 P;
      N.convert_inner_shell_to_polyhedron(ci->shells_begin(), P);
      convex_parts.push_back(P);
    }
  }
  std::cout << "decomposition into " << convex_parts.size() << " convex parts " << std::endl;

}
```

GilesBathgate:
## Summary of Changes

Change the order of delete in External_structure_builder.h to avoid a double free in the exception case

## Release Management

* Issue(s) solved (if any): fix #422 

### DIFF
--- a/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/External_structure_builder.h
+++ b/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/External_structure_builder.h
@@ -81,10 +81,10 @@ class External_structure_builder : public Modifier_base<typename Nef_::SNC_and_P
     SNC_point_locator* old_pl = pl;
     pl = pl->clone();
     sncpl.pl = pl;
-    delete old_pl;
     SNC_external_structure C(*sncp,pl);
     C.clear_external_structure();
     C.build_external_structure();
+    delete old_pl;
   }
 };
 
--- a/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/External_structure_builder.h
+++ b/Convex_decomposition_3/include/CGAL/Convex_decomposition_3/External_structure_builder.h
@@ -77,51 +77,7 @@ class External_structure_builder : public Modifier_base<typename Nef_::SNC_and_P
     SNC_structure* sncp(sncpl.sncp);
     SNC_point_locator* pl(sncpl.pl);
 
-
-
     Unique_hash_map<SHalfedge_handle, SFace_handle> sedge2sface;
-    /*
-    SFace_iterator sfi;
-    CGAL_forall_sfaces(sfi, *sncp) {
-      SFace_cycle_iterator sfc;
-      for(sfc = sfi->sface_cycles_begin(); sfc != sfi->sface_cycles_end(); ++sfc) {
-        if(sfc.is_shalfedge()){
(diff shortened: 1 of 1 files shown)


sloriot:

sloriot:
Seems worth going into 5.5

sloriot:
AFAIS, it does fix the segfault of 422 but not the assertion failing. I'll try to have a look later to see if the fix is easy

GilesBathgate:
Yes I think issue #422 was about the segfault. Didn't @kintel raise another issue regarding the assertion fail, or do we even expect the nef input provided to be decompose-able?

sloriot:
I think you are right. There is an isolated vertex that seems to be responsible for not being decomposable. I'll merge the PR as it fixes an issue.

Issue #8327, opened 2024-07-01
Reachability of file ftp://ftp.mpi-sb.mpg.de/pub/outgoing/CGAL/Ridges_3_datafiles.tgz


albert-github:
In the file Ridges_3/doc/Ridges_3/Ridges_3.txt we have the code:
```
\cgalFootnote{Model data may be downloaded via ftp://ftp.mpi-sb.mpg.de/pub/outgoing/CGAL/Ridges_3_datafiles.tgz . The mechanical part model has been provided courtesy of Dassault System to produce \cgalFigureRef{figmechanical_crest_filteredintro}, due to copyright issues the available model is not the same, it is provided by the AIM\@SHAPE Shape Repository.}
```

I was not able to retrieve this linked ftp file.
Wouldn't it be better to have this file somewhere on a CGAL server or is this not possible due to copyright restrictions?

sloriot:
@pougetma do you have some files locally or in your archives?

afabri:
There are more things to fix: 

-  "cgal:p-gd-01"      in the sentence "the reader may consult [[3]](https://doc.cgal.org/latest/Ridges_3/citelist.html#CITEREF_cgal:hgygm-ttdpf-99),cgal:p-gd-01, as well as"
- The examples in the User Manual do as if a `data` directory existed

PR #8440, opened 2024-08-28
Ridges 3 bug fix doc pougetma

pougetma:
## Summary of Changes

Add a rep example/data with 2 off files

## Release Management

Solves issue 8327 
https://github.com/CGAL/cgal/issues/8327

### DIFF
--- a/Ridges_3/doc/Ridges_3/Ridges_3.txt
+++ b/Ridges_3/doc/Ridges_3/Ridges_3.txt
@@ -44,8 +44,8 @@ functions of the package are provided in Section \ref Ridges_3Examples.
 
 For a detailed introduction to ridges and related topics, the reader
 may consult
-\cgalCite{cgal:hgygm-ttdpf-99},cgal:p-gd-01, as well as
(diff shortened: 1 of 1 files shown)


sloriot:
@pougetma Thanks for the PR. Do you agree that shipping the data is actually not needed and only the text should be updated?

pougetma:
Do you mean I need to store the data outside of CGAL? I’m not aware of the practice of data management within cgal.
The 2 examples are light weight, so  it seemed reasonable to have them in cgal.
M.

sloriot:
Well, it does increase the history size of 0.5MB and I don't see any added value: None of them are used in tests, rocker-arm.off is a plain copy of aim@shape data and the ellipsoid could be regenerated using an implicit function with `Mesh_3`. Even worse for rocker-arm, it's not the data used to generate the figure. Do you see what I mean? (I can call you if I'm not clear :) )

pougetma:
Ok, agree, I removed the data files

sloriot:

PR #8445, opened 2024-08-29
Gsoc2024 hexahedral mesh generation tbenard

theo97490:
## Summary of Changes

Implementation of the method used in the paper "A template-based approach for parallel hexahedral two-refinement", Steven J. Owen, Ryan M. Shih, Corey D. Ernst (2015, 2017); 

Adds two functions to compute the hexahedral two-refinement of a mesh under the Linear Cell Complex package, respectively a single-threaded and multi-threaded version of the algorithm.

[Additionnal information on the project public gist here](https://gist.github.com/theo97490/961d6e67d6456d660067f5229368099c)

We asked ourselves with gdamiand if this this going to be a package or a part of the LCC package. I think the algorithm has no real purpose to have its own package. Currently it is written as an example under the LCC package, we might move it in the end into the include folder of the package.

## TODO

- [ ] Fix a bug where stored dart handles to planes (x,y,z) may be invalidated due to refinement.
- [ ] Finish the Multi Threaded version :
  - [ ] Finish the MT for levels of refinement above 1
  - [ ] Sew all LCCs into one after all threads finishes
  - [ ] A test to check if any decomposition of the domain doesn't change the result of the MT function, with the test being truthy. 
- [ ] Vertices displacement procedure ( see 2017 version of the paper )
- [ ] Cleanup the code
- [ ] Move the code from "examples" to "include" inside the LCC package


sloriot:
Replaced by #9045

PR #8453, opened 2024-09-02
simple fix to remove the segfault with Qt>=6.7

### LINKED ISSUE #8232: Simple drawing fails with Qt6 6.7.1
## Issue Details

Issue #8058 is very similar, it reported that simple drawing did not work with 6.8.
It also reported on some problems with find_package of CGAL with Qt6.
The latter was fixed, and it turned out that 6.8 hasn't been officially published, so it was closed.
However, it makes me think that the same reason that caused the segmentation with 6.8 is also the reason for the segmentation with 6.7. (This piece of info may be useless...)

All the example programs that use the simple drawing mechanism end with a segmentation fault, e.g., Polygon/draw_polygon, Arrangement_2/draw_arrangement, Surface_mesh/sm_draw_small_faces. Other programs that are based on Qt6 work fine. This is, of course, not a proof that the problem is in our code, but it's suspicious.

I've experienced the same behavior on two separate machines running Ubuntu 22.04.

Here is the trace of the stack:
[stack_qt_issue.txt](https://github.com/CGAL/cgal/files/15467708/stack_qt_issue.txt)

## Environment

* Operating system Linux, 64 bits):
* Compiler: g++ 11.4.0
* Release or debug mode: both
* CGAL version: latest
* Boost version: 1.82

lrineau:
## Summary of Changes

This should fix #8232. I have tested locally. I hope that does not 

## Release Management

* Affected package(s): GraphicsView, Qt6
* Issue(s) solved (if any): fix #8232
* License and copyright ownership: bug fix

### DIFF
--- a/GraphicsView/include/CGAL/Qt/qglviewer_impl.h
+++ b/GraphicsView/include/CGAL/Qt/qglviewer_impl.h
@@ -354,7 +354,6 @@ camera is manipulated) : main drawing method. Should be overloaded. \arg
 postDraw() : display of visual hints (world axis, FPS...) */
 CGAL_INLINE_FUNCTION
 void CGAL::QGLViewer::paintGL() {
-  makeCurrent();
   // Clears screen, set model view matrix...
   preDraw();
   // Used defined method. Default calls draw()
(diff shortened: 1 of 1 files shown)


sloriot:

sloriot:
Also locally tested (also runtime with Qt 6.4)

PR #8485, opened 2024-09-19
Only install headers when component is enabled.

mering:
## Summary of Changes

When Qt or demos are disabled, `make install` should not install the respective headers. This PR takes the existing flags into account to only install these headers when the respective feature is enabled.

## Release Management

* Affected package(s):
* Issue(s) solved (if any):
* Link to compiled documentation:


Installation/CMakeLists.txt
```diff
@@ -646,8 +646,10 @@ cache_get(CGAL_3RD_PARTY_INCLUDE_DIRS  )
 cache_get(CGAL_3RD_PARTY_LIBRARIES     )
 cache_get(CGAL_3RD_PARTY_LIBRARIES_DIRS)
 
-install(DIRECTORY "${CGAL_GRAPHICSVIEW_PACKAGE_DIR}/include/CGAL/Qt/" DESTINATION "${CGAL_INSTALL_INC_DIR}/CGAL/Qt" COMPONENT CGAL_Qt5)
-if(CGAL_BRANCH_BUILD)
+if(WITH_CGAL_Qt6)
```
sloriot, line 649:
```suggestion
if(WITH_CGAL_Qt5)
```

sloriot: The PR targets 5.6.x and qt6 is introduced in CGAL 6.0

sloriot (approved):

sloriot:
Thanks for the contribution!

sloriot:
Breaks `test_config_file_non_standard_install_lib_cmake_with_qt` in Installation tests. I don't understand why yet.

lrineau:
Since CGAL is header-only, the variable `WITH_CGAL_Qt5` is not something we want to support.

CMake install components are already well defined. If you want to install without the component `CGAL_Qt5`, then use the command `cmake --install ` with the option `--component Unspecified`.

```shell
cmake --install build/Debug --component Unspecified
```

lrineau:
@sloriot @SaillantNicolas

We should use [`CMAKE_INSTALL_DEFAULT_COMPONENT_NAME`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_DEFAULT_COMPONENT_NAME.html#cmake-install-default-component-name) to choose a better name instead of `Unspecified`.

Maybe:

```cmake
set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME CGAL)
```

sloriot:
if we cannot point to any documentation page, I suggest not to close this issue and add that documentation page (or give a link here if it does exist).

mering:
> Since CGAL is header-only, the variable `WITH_CGAL_Qt5` is not something we want to support.

@lrineau If you do not want to support this variable, you should remove it from your documentation like https://github.com/CGAL/cgal/blob/f1af52edeb741a05f86d959984ed943f25f5fcd4/Documentation/doc/Documentation/advanced/Configuration_variables.txt#L30 or https://github.com/CGAL/cgal/blob/f1af52edeb741a05f86d959984ed943f25f5fcd4/Documentation/doc/Documentation/advanced/Installation.txt#L53-L64

lrineau:
Hi @mering, you are right that our documentation pages are not updated. The switch to header-only was from [CGAL-5.3](https://github.com/CGAL/cgal/releases/tag/v5.3), released in 2021. But the cleaning of the documentation and CMake scripts is still pending. I filled the issue #4815, a few years ago, but it has never had sufficient priority for us to address it fully.

Is the command line with `cmake --install` and the option `--component` suitable for your need? If not, please tell us more about your need, so that we can find a solution.

mering:
> Hi @mering, you are right that our documentation pages are not updated. The switch to header-only was from [CGAL-5.3](https://github.com/CGAL/cgal/releases/tag/v5.3), released in 2021. But the cleaning of the documentation and CMake scripts is still pending. I filled the issue #4815, a few years ago, but it has never had sufficient priority for us to address it fully.

This option is also still used in tests and scripts: https://github.com/search?q=repo%3ACGAL%2Fcgal%20WITH_CGAL_Qt&type=code.

> Is the command line with `cmake --install` and the option `--component` suitable for your need? If not, please tell us more about your need, so that we can find a solution.

I am just testing this in our workflow but so far this seems to work flawlessly! Thanks a lot for the hint!

PR #8492, opened 2024-09-24
Arrangements:  Unify mix of pre and post condition macros

afabri:
## Summary of Changes

We had `postcondition_code()`  followed by `precondition()` .  I saw it when I only switched off postconditions with a macro.
We might do that for an entire testsuite. 

## Release Management

* Affected package(s): Arrangement on Surface


sloriot:

PR #8496, opened 2024-09-25
CGAL: Deal with CGAL_assertion(false)

### LINKED ISSUE #8295: CGAL_assertion(false)/CGAL_error() replace by CGAL_unreachable()?
There are more than 250 occurrences of `CGAL_assertion(false)` in CGAL. And other occurrences of `CGAL_error()`. Should not we replace them with `CGAL_unreachable()`?

afabri:
## Summary of Changes

Replace or remove some `CGAL_assertion(false)`.  

This is WIP as some need  further thought or discussion.

## Release Management

* Affected package(s):
* Issue(s) solved (if any): fix #8295

### DIFF
--- a/Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/internal/smooth_vertices.h
+++ b/Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/internal/smooth_vertices.h
@@ -875,7 +875,7 @@ std::size_t smooth_vertices_on_surfaces(C3t3& c3t3,
               return CGAL::midpoint(s->source(), s->target());
             else
             {
-              CGAL_assertion(false);
+              CGAL_assertion(false);   AF:  So are we not completely sure?
               return std::nullopt;
             }
           };
--- a/Algebraic_kernel_d/include/CGAL/Algebraic_kernel_d/Curve_analysis_2.h
+++ b/Algebraic_kernel_d/include/CGAL/Algebraic_kernel_d/Curve_analysis_2.h
@@ -890,7 +890,7 @@ private:
                 break;
             }
             default:{
-              CGAL_assertion(false); // !!! Never reached
+              CGAL_unreachable();
             }
             }
         }
--- a/Alpha_shapes_2/include/CGAL/Alpha_shapes_2/internal/Lazy_alpha_nt_2.h
+++ b/Alpha_shapes_2/include/CGAL/Alpha_shapes_2/internal/Lazy_alpha_nt_2.h
@@ -212,7 +212,7 @@ public:
         exact_ = Exact_squared_radius()( to_exact(*data().p0),to_exact(*data().p1),to_exact(*data().p2) );
         break;
       default:
-        CGAL_assertion(false);
+        CGAL_unreachable();
     }
   }
 
@@ -229,7 +229,7 @@ public:
         approx_ = Approx_squared_radius()( to_approx(*data().p0),to_approx(*data().p1),to_approx(*data().p2) );
         break;
       default:
-        CGAL_assertion(false);
+        CGAL_unreachable());
     }
   }
 
--- a/Alpha_shapes_3/include/CGAL/Alpha_shapes_3/internal/Lazy_alpha_nt_3.h
+++ b/Alpha_shapes_3/include/CGAL/Alpha_shapes_3/internal/Lazy_alpha_nt_3.h
@@ -224,7 +224,7 @@ public:
         approx_ = Approx_squared_radius()( to_approx(*data().p0),to_approx(*data().p1),to_approx(*data().p2),to_approx(*data().p3) );
       break;
       default:
-        CGAL_assertion(false);
+        CGAL_unreachable();
     }
   }
 
--- a/Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_X_trapezoid.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arr_point_location/Td_X_trapezoid.h
@@ -605,10 +605,9 @@ public:
     case CGAL_TD_VERTEX:
       return TD_VERTEX;
     default:
-       CGAL_assertion(false);
-       return TD_TRAPEZOID;
+       CGAL_unreachable();
     }
-    }
+  }
 
   /*! Access trapezoid type flag. */
   CGAL_TD_INLINE unsigned char type_flag() const
--- a/Cone_spanners_2/include/CGAL/Compute_cone_boundaries_2.h
+++ b/Cone_spanners_2/include/CGAL/Compute_cone_boundaries_2.h
@@ -91,10 +91,8 @@ public:
     DirectionOutputIterator operator()(const unsigned int cone_number,
                                        const Direction_2& initial_direction,
                                        DirectionOutputIterator result)  {
-        if (cone_number<2) {
-            std::cout << "The number of cones must be larger than 1!" << std::endl;
-            CGAL_assertion(false);
-        }
+
+        CGAL_assertion_msg(cone_number>1, "The number of cones must be larger than 1");
 
         *result++ = initial_direction;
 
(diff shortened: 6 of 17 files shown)


Triangulation_on_sphere_2/include/CGAL/Triangulation_on_sphere_2.h
```diff
@@ -964,18 +964,12 @@ march_locate_2D(Face_handle f,
...
         li = (o0 != COLLINEAR) ? 2 : (o1 != COLLINEAR) ? 0 : 1;
         break;
       }
-      default:
-      {
-        // impossible
-        CGAL_assertion(false);
-        return f;
-      }
     }
```
lrineau, line 973: For the compiler, those two versions of the code are different.

* With
  
  ```c++
  switch(sum)
  {
    case 0: // ...
    case 1: // ...
    case 2: // ...
    default: CGAL_unreachable();
  }
  ```
  
  then the compiler knows `sum` can only be 0, 1, or 2.
* With
  
  ```c++
  switch(sum)
  {
    case 0: // ...
    case 1: // ...
    case 2: // ... // do not forget the break!
    default: CGAL_unreachable();
  }
  ```
  
  then the compiler has no knowledge about values of `sum`, and might produce a slightly less efficient code (and maybe warnings).

In that case, where `sum` is initialized with:

```c++
    int sum =  ((o0 == COLLINEAR) ? 1 : 0)
             + ((o1 == COLLINEAR) ? 1 : 0)
+ ((o2 == COLLINEAR) ? 1 : 0);
```

maybe the compiler was already able to deduce the value range of `sum`. The version with `default: CGAL_unreachable();` is more explicit.

lrineau (commented):
Looks good, so far, except for one case.

Actually, I am not so sure with should replace `CGAL_assertion(false)` with `CGAL_unreachable()` everywhere. For users activating the CGAL assertions even in production, the code will behave differently.

mglisse:
> Actually, I am not so sure with should replace `CGAL_assertion(false)` with `CGAL_unreachable()` everywhere. For users activating the CGAL assertions even in production, the code will behave differently.

Oh, I always assumed that `CGAL_unreachable()` was the same as `CGAL_assume(0)`, i.e. that when assertions are enabled it would error out, but I now see this isn't the case. Was this a deliberate choice, or an accident?

lrineau:
> > Actually, I am not so sure with should replace `CGAL_assertion(false)` with `CGAL_unreachable()` everywhere. For users activating the CGAL assertions even in production, the code will behave differently.
> 
> Oh, I always assumed that `CGAL_unreachable()` was the same as `CGAL_assume(0)`, i.e. that when assertions are enabled it would error out, but I now see this isn't the case. Was this a deliberate choice, or an accident?

I am not sure it was intended. The commit that introduced that behavior is https://github.com/CGAL/cgal/pull/6052/commits/604d2fd1ffcd0072958ba53b606d80fe7b993883. Actually, I think it was intentional. The goal was to convince the compiler that the branch was dead, to remove warnings.

That means `CGAL_unreachable()` no longer triggers the `CGAL::Assertion_exception` exception, if CGAL assertion are enabled. And `CGAL_assertion(false)` is not equivalent to `CGAL_unreachable()`.

`CGAL_unreachable()` should be restricted to cases where it is impossible the branch can be reached. Or if invalid user-inputs are already checked by other assertions. Otherwise, in Debug mode, users might get segfaults instead of CGAL assertion failures.

mglisse:
Does that mean we should use `CGAL_assume(false)` by default, and only use `CGAL_unreachable()` in a few "safe" cases?

Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/internal/smooth_vertices.h
```diff
@@ -871,12 +871,9 @@ std::size_t smooth_vertices_on_surfaces(C3t3& c3t3,
           {
             const auto intersection = proj.value().first;
             using Segment = typename Tr::Geom_traits::Segment_3;
-            if (const Segment* s = std::get_if<Segment>(&intersection))
+            if (const Segment* s = std::get_if<Segment>(&intersection)){
+              CGAL_assertion(s != nullptr);
```
afabri, line 875: assertion comes to late.

afabri:
Concerning `CGAL::assume()`  it is unlucky that it has not the same semantics as `[[ assume ]]`.   
We have probably not many users of `CGAL::assume()` so do we want to break it or do we want to a add a CGAL::cpp23::assume()`? 
Or simply not use something like  `[[ assume ]]`?

Triangulation_on_sphere_2/include/CGAL/Triangulation_on_sphere_2.h
afabri: I reverted this change.

afabri:
@sloriot this is no longer WIP.

sloriot:

afabri:
@sloriot what has to be approved?

Segment_Delaunay_graph_Linf_2/include/CGAL/Segment_Delaunay_graph_Linf_2/Vertex_conflict_C2.h
```diff
@@ -780,7 +780,7 @@ class Vertex_conflict_C2
 
         // philaris: execution should never reach here
         // philaris: code added to avoid warnings on some compilers
-        CGAL_assertion( false );
+        CGAL_assertion(false);
```
sloriot, line 783: no change in this file?

Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/internal/smooth_vertices.h
```diff
@@ -871,13 +871,9 @@ std::size_t smooth_vertices_on_surfaces(C3t3& c3t3,
...
             const auto intersection = proj.value().first;
             using Segment = typename Tr::Geom_traits::Segment_3;
-            if (const Segment* s = std::get_if<Segment>(&intersection))
-              return CGAL::midpoint(s->source(), s->target());
-            else
-            {
-              CGAL_assertion(false);
-              return std::nullopt;
-            }
+            const Segment* s = std::get_if<Segment>(&intersection);
```
sloriot, line 874: it's not clear to me that here it can't happen. We are using a kernl with inexact constructions. @janetournois

sloriot (commented):
What I don't like with `CGAL_unreachable()` replacement is that if there is a bug in CGAL code and that we reach a place marked as unreachable no assertion will be thrown. Maybe we should modify the macro to also raise `CGAL_assertion(false)`.

Tetrahedral_remeshing/include/CGAL/Tetrahedral_remeshing/internal/smooth_vertices.h
sloriot: as the assertion still exist and nullptr output is not handled in the calling code, it's not a regression.

afabri:
> What I don't like with `CGAL_unreachable()` replacement is that if there is a bug in CGAL code and that we reach a place marked as unreachable no assertion will be thrown. Maybe we should modify the macro to also raise `CGAL_assertion(false)`.

The whole point is to tell the compiler that it can go ahead and optimize things. For example for the three orientation switch.

sloriot:
> > What I don't like with `CGAL_unreachable()` replacement is that if there is a bug in CGAL code and that we reach a place marked as unreachable no assertion will be thrown. Maybe we should modify the macro to also raise `CGAL_assertion(false)`.
> 
> The whole point is to tell the compiler that it can go ahead and optimize things. For example for the three orientation switch.

I know, what I'm saying is that the doc of `_builtin_unreachable` says that it's undefined behavior if the program "reaches" it. So without an assertion, we can't detect it.

Issue #8320, opened 2024-06-30
Missing build dependencies


albert-github:
During working on issue #8318 (resulting in proposed pull request #8319) is was discovered that in the file: Documentation/doc/CMakeLists.txt the rule:
```
execute_process(
  COMMAND ${Python3_EXECUTABLE} ${CGAL_DOC_SCRIPT_DIR}/generate_how_to_cite.py
          ${CGAL_ROOT} ${CMAKE_BINARY_DIR} "${CGAL_BRANCH_BUILD}"
  RESULT_VARIABLE GENERATE_HOW_TO_CITE_RESULT)
if(NOT GENERATE_HOW_TO_CITE_RESULT EQUAL "0")
  message(FATAL_ERROR "generate_how_to_cite did not exit correctly")
endif()
```
does not depend on the different PackageDescription.txt used

afabri:
Do you mean that if the bib entry of the package is changed this does not trigger the rebuild of the page with the bib entries for the chapters of a release?

albert-github:
I don't 100% remember anymore.
We see in #8318 that the `PackageDescription.txt` of 2 packages changed and that this, apparently, doesn't trigger a rebuild.
We see in `./doc/scripts/generate_how_to_cite.py` that in the main routine the used `PackageDescription.txt` are determined, but these dependencies are (of course) not known to CMake and hence no rebuild.

It will, very likely, not be easy to let cmake know the dependencies. Probably some multi-stage approach is necessary (but I have no good idea for this).

afabri:
The error will not show up during the CI or a build request in a PR.   But we should get an error in the nightly documentation testsuite.  We could consider that as good enough.

sloriot:
You won't have any issue if you start with a fresh build. So it is only a developer issue.

albert-github:
Well probably the major issue looks like to be that doxygen doesn't emit a warning for the double but different bib definition.
It looks like I found it when I ran a link checker on the project and got a warning like (on June 30, 2024): 
```
List of duplicate and empty anchors
        cgal:ol-kinetic-24a     Lines: 531, 536
```
on `Manual/how_to_cite_cgal.html` and `Manual/how_to_cite.html`.

So it might go unnoticed for a long time.

I made a note of it so I could see what can be done in doxygen ...

Issue #8318, opened 2024-06-30
Double biblio rererence


albert-github:
On the documentation output page: doc_output/Manual/how_to_cite.html

We see:
![image](https://github.com/CGAL/cgal/assets/5223533/82250e72-e898-49d2-8181-55f6feea4171)

so twice the generated label: `cgal:ol-kinetic-24a`

and on the page: doc_output/Manual/how_to_cite_cgal.html

![image](https://github.com/CGAL/cgal/assets/5223533/6274cf5a-6702-415e-9fe7-68aa7dc6f1cc)

Resulting from:
- Kinetic_space_partition/doc/Kinetic_space_partition/PackageDescription.txt
- Kinetic_surface_reconstruction/doc/Kinetic_surface_reconstruction/PackageDescription.txt

PR #8504, opened 2024-09-27
Double use of citation name

albert-github:
Based on #8320 and https://github.com/doxygen/doxygen/pull/11157 creating an assertion for builds from scratch when a citation name is used multiple times.

### DIFF
--- a/Documentation/doc/scripts/generate_how_to_cite.py
+++ b/Documentation/doc/scripts/generate_how_to_cite.py
@@ -278,7 +278,9 @@ def main():
         encoding="utf-8",
     )
     k = 2
+    citeDic = {}
     for line in f:
+        foundDouble=False
         match = pattern.match(line)
         if match:
             pkg = match.group(1)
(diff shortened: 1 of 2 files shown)


afabri:
It would be good to add a duplicate entry so that we can see that it gets catched in the manual testsuite.

albert-github:
Well that would probably be a bit problematic as it has to be in `PackageDescription.txt` files and we would see it every night in the overview https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/ 
Or are you referring to another place?

afabri:
You added  a duplicate detection feature to a script that is run to build the full documentation.
In order to test your feature we should add a duplicate, and merge this branch in integration.
If your feature works we will see it in the documentation testsuite page. 
Then we can remove the duplicate again, and merge this PR in master.

albert-github:
That would be one scenario. Another would be to do it locally, like I did for testing.
I changed the original problematic ones (#8318), i.e. the ones in 
-    Kinetic_space_partition/doc/Kinetic_space_partition/PackageDescription.txt
-    Kinetic_surface_reconstruction/doc/Kinetic_surface_reconstruction/PackageDescription.txt
to `cgal:ol-kinetic` and ran the documentation generation locally so that I got the error.

sloriot:
Locally tested, works great. Thanks @albert-github

PR #8509, opened 2024-10-01
Polygon: various clean-ups

sloriot:
Adds a `GraphicsItem`  for multipolygons with holes. Adds a `z` for items. Adds reading from `argc` in an example.


afabri:
@sloriot it broke I-c-337 and needs retesting. Fixed by last commit.

sloriot:

PR #8516, opened 2024-10-04
remove extra template keyword

sloriot:


sloriot:

PR #8527, opened 2024-10-08
Triangulation_2: Do not include kernel header

afabri:
## Summary of Changes

Do not include `Exact_predicates_inexact_constructions_kernel.h`  and `Simple_cartesian.h`.
Change order of `#include` statements so that the  T2 files come first.
Forward declare `Epick` in `Constrained_triangulation_plus_2` where it is used for defining a `Default` for the template parameter (which is not documented).

## Release Management


sloriot:
If I write an example that is not including EPICK.h and that I use `Default` as template parameter, it will not compile.

sloriot:
You should simply remove the default parameter

Triangulation_2/include/CGAL/Constrained_triangulation_plus_2.h
```diff
@@ -24,16 +24,17 @@
...
 #include <CGAL/Triangulation_2/insert_constraints.h>
+
 #include <boost/container/flat_set.hpp>
 
 #include <type_traits>
 
 namespace CGAL {
 
+ class Epick;
+
```
sloriot, line 37:
```suggestion
```

PR #8528, opened 2024-10-08
CGAL: Modernize CMakeLists.txt

afabri:
## Summary of Changes

- Changed `PUBLIC` to `PRIVATE`  in the `target_link_libraries()`
- ~~Link against `Boost::<COMPONENT>`~~   as we do more than just linking and must use the `CGAL::_.._support`
- Use `target_compile_definitions()` and `add_compile_definitions()`

More to be done.

## Release Management

* Affected package(s): all
* Issue(s) solved (if any): fix one item of  #4815


afabri:
Why do we have this cmake if-then-else [here](https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L22)?   Is it for backward compatibility for older versions of cmake?  @lrineau

lrineau:
> Why do we have this cmake if-then-else [here](https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L22)? Is it for backward compatibility for older versions of cmake? @lrineau

Yes, I think so. Maybe it can be removed.

afabri:
> > Why do we have this cmake if-then-else [here](https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L22)? Is it for backward compatibility for older versions of cmake? @lrineau
> 
> Yes, I think so. Maybe it can be removed.

I guess we have a test platform where we use the "minimal required" version of cmake?

afabri:
> Also can we then remove the line `add_definitions("-DCGAL_USE_BOOST_PROGRAM_OPTIONS")` ??

It is needed.   Does that need "modernization" @lrineau ?

sloriot:
> > > Why do we have this cmake if-then-else [here](https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L22)? Is it for backward compatibility for older versions of cmake? @lrineau
> > 
> > 
> > Yes, I think so. Maybe it can be removed.
> 
> I guess we have a test platform where we use the "minimal required" version of cmake?

My guess is that it will depend how boost is found (config mode or with FindBoost.cmake).

> 
> Also can we then remove the line `add_definitions("-DCGAL_USE_BOOST_PROGRAM_OPTIONS")` ??
IIRC, this is to disable something at runtime in the example. So if you remove it you need to update the example and the cmake script to not compile it.

Lab/demo/Lab/Plugins/PCA/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ cgal_lab_plugin(create_obb_mesh_plugin Create_obb_mesh_plugin)
 target_link_libraries(create_obb_mesh_plugin PUBLIC scene_surface_mesh_item
                                                     scene_polygon_soup_item
                                                     scene_selection_item
-                                                    scene_points_with_normal_item)
+                                                    scene_points_with_normal_item CGAL::Eigen3_support)
```
lrineau, line 34: Actually, it should be inherited from the items, unless the code of `scene_points_with_normal` itself uses Eigen.

lrineau:
> CGAL_USE_BOOST_PROGRAM_OPTIONS

No. That preprocessor macro is used in `Ridges_3` examples.

lrineau:
> > Also can we then remove the line `add_definitions("-DCGAL_USE_BOOST_PROGRAM_OPTIONS")` ??
> 
> It is needed. Does that need "modernization" @lrineau ?

https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L21-L30

It swaps three lines of `target_compile_definitions` against one line `add_definitions` that is for the three programs after. I think it is all right. But, as for all the other cases, that is a matter of taste. If the other way around looks better for you, change the code.

Lab/demo/Lab/Plugins/PCA/CMakeLists.txt
afabri: It only inherits if any of the dependency declares its Eigen dependency public I guess. Note that `create_obb_mesh_plugin.cpp`   used Eigen directly.

afabri:
TBB seems also to use keywords like `emit()`, as can be seen in the CI
```
/usr/include/oneapi/tbb/profiling.h:229:15: error: expected unqualified-id before ‘)’ token
  229 |     void emit() { }
  ```

**Comment by lrineau** [comment]
> TBB seems also to use keywords like `emit()`, as can be seen in the CI
> 
> ```
> /usr/include/oneapi/tbb/profiling.h:229:15: error: expected unqualified-id before ‘)’ token
>   229 |     void emit() { }
> ```

Bless the CI!

And I thanks myself for that initiative (started by #4234).

**Comment by afabri** [comment]
There is a disussion on[ github/TBB](https://github.com/oneapi-src/oneTBB/issues/547).  There I even saw @mglisse.
My conclusion is that we just keep `QT_NO_KEYWORDS`

**Comment by sloriot** [comment]
> > > > Why do we have this cmake if-then-else [here](https://github.com/CGAL/cgal/blob/4305f01384659ecc200db2c41f41aa48d43e92d2/Ridges_3/examples/Ridges_3/CMakeLists.txt#L22)? Is it for backward compatibility for older versions of cmake? @lrineau
> > > 
> > > 
> > > Yes, I think so. Maybe it can be removed.
> > 
> > 
> > I guess we have a test platform where we use the "minimal required" version of cmake?
> 
> My guess is that it will depend how boost is found (config mode or with FindBoost.cmake).
> 
> > Also can we then remove the line `add_definitions("-DCGAL_USE_BOOST_PROGRAM_OPTIONS")` ??
> 
> IIRC, this is to disable something at runtime in the example. So if you remove it you need to update the example and the cmake script to not compile it.

Starting cmake 3.5, `Boost::<COMPONENT>` are also defined by `FindBoost.cmake` so we can get rid of the else case.

**Comment by afabri** [comment]
@lrineau what is `CGAL::Data`   in `target_link_libraries()` about ?

**Comment by sloriot** [comment]
> @lrineau what is `CGAL::Data` in `target_link_libraries()` about ?

it is used to set the path to the data directory

**Comment by afabri** [comment]
Concering VTK it seems that we still have to link against [${VTK_LIBRARIES}](https://docs.vtk.org/en/latest/getting_started/index.html#using-c-and-cmake)

**Comment by lrineau** [comment]
> @lrineau what is `CGAL::Data` in `target_link_libraries()` about ?

That is the CMake support for [the function `CGAL::data_file_path`](https://doc.cgal.org/6.0/Manual/namespaceCGAL.html#acdae9a147ad2a3998cc21f88bc292dac). As for CGAL-6.0, it only adds a compile definition `-DCGAL_DATA_DIR=</path/to/Data/data>`.

File: Point_set_processing_3/examples/Point_set_processing_3/CMakeLists.txt [hunk 2]
```diff
@@ -25,13 +22,9 @@ endif()
...
 option(CGAL_ACTIVATE_CONCURRENT_PSP3 "Enable concurrency" OFF)
 
-set(CGAL_libs CGAL::CGAL)
 if(CGAL_ACTIVATE_CONCURRENT_PSP3 OR "$ENV{CGAL_ACTIVATE_CONCURRENT_PSP3}")
   find_package(TBB REQUIRED)
   include(CGAL_TBB_support)
-  if(TARGET CGAL::TBB_support)
-    set(CGAL_TBB_target ${CGAL_libs} CGAL::TBB_support)
-  endif()
 endif()
```
lrineau, line 29: With `CGAL_ACTIVATE_CONCURRENT_PSP3`, TBB is searched, but never used.

The problems date from PR #4636. See the culprit change at https://github.com/CGAL/cgal/pull/4636/files#diff-16a24da0dde4987fe4dfe346bf9547c0985d8e02341d32fbe191b6197d0ea1f8.

Shape_detection/test/Shape_detection/CMakeLists.txt
```diff
@@ -45,23 +45,13 @@ if(TARGET CGAL::Eigen3_support)
...
-    add_library(libproto STATIC  ${proto_src} ${proto_misc_src})
-    add_executable(test_validity_sampled_data "test_validity_sampled_data.cpp")
-    target_link_libraries(test_validity_sampled_data libproto CGAL::CGAL CGAL::Data CGAL::Eigen3_support)
-  else()
-    add_executable(test_validity_sampled_data "test_validity_sampled_data.cpp")
-    target_link_libraries(test_validity_sampled_data CGAL::CGAL CGAL::Data CGAL::Eigen3_support)
-  endif()
+
+  add_executable(test_validity_sampled_data "test_validity_sampled_data.cpp")
+  target_link_libraries(test_validity_sampled_data CGAL::CGAL CGAL::Data CGAL::Eigen3_support)
```
lrineau, line 54: The support for `RANSAC_PROTO_DIR` is removed. Is it intentional?

afabri: yes. After a discussion with @soesau

Point_set_processing_3/examples/Point_set_processing_3/CMakeLists.txt
afabri: Shall we fix this in this PR, or shall we handle that independently so that it also makes it into 5.6.1 ?

lrineau: I think it is better in an other PR. This one is already complex enough.

afabri: @soesau can you please take care of this.

Shape_detection/test/Shape_detection/CMakeLists.txt
lrineau: Right. That seems to be commit 9edf6eadc7a93ef451b184d0b2cc3355fcec9412. Thanks.

Point_set_processing_3/examples/Point_set_processing_3/CMakeLists.txt
lrineau: I have created https://github.com/CGAL/cgal/issues/8553 to track that issue.

lrineau (approved):

Lab/demo/Lab/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(Lab_Demo)
 include(FeatureSummary)
 
 
-add_definitions ( -DCGAL_NO_DEPRECATED_CODE )
+add_compile_definitions ( -DCGAL_NO_DEPRECATED_CODE )
```
sloriot, line 6: Looks like magritte does not like it: https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-354/Lab_Demo/TestReport_magritte_arm64-apple_bigsur_clang1200-release-64bits.gz

sloriot: *New in version 3.26: Any leading -D on an item will be removed.* so we have to remove the -D

sloriot:
Successfully tested in CGAL-6.1-Ic-8 (CGAL-6.1-Ic-7 for Rubens)

PR #8529, opened 2024-10-09
Spatial_searching: Do not quiet warning in CMakeLists.txt

afabri:
## Summary of Changes

It is not a good idea to quiet a warning in the `CMakeLists.txt` file. 

_Todo:_  Fix it in the code, or  disable warning in the header file itself.

There is another warning disabled in the test of [`Distance_3`](https://github.com/CGAL/cgal/blob/master/Distance_3/test/Distance_3/CMakeLists.txt#L9).  The question is if we have a test case where we force it into the problem, or if the same warning can show up for any user code.   The warning does not show up outside this testsuite which might be an indication.

## Release Management

### DIFF
--- a/Spatial_searching/examples/Spatial_searching/CMakeLists.txt
+++ b/Spatial_searching/examples/Spatial_searching/CMakeLists.txt
@@ -7,11 +7,6 @@ project(Spatial_searching_Examples)
 # CGAL and its components
 find_package(CGAL REQUIRED)
 
-if(MSVC)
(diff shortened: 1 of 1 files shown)


afabri:
I do not get a warning with `/W4`, so let's see what the testsuite says. @sloriot

lrineau:
Note that those lines were from a77205a33c0ee342f6e28cfc0c277e70ce944189, about ten years ago.

sloriot:

PR #8530, opened 2024-10-09
[KSR] Fix crash in collect_connected_border()

soesau:
## Summary of Changes

using LCC with `GenericMap::upper_bound_on_dart_ids()` instead of `GenericMap::number_of_darts()`.

`number_of_darts()` only provides the number, but not the largest id.

## Release Management

* Affected package(s): KSR


sloriot:

PR #8540, opened 2024-10-14
[Lab] Fix transparency

soesau:
## Summary of Changes

The viewer does not return the dimensions of the framebuffer, thus the viewport information is used to create the framebuffers for depth peeling. Also the correct width and height are communicated to the shader.

## Release Management

* Affected package(s): Lab


afabri:
I am wondering what the semantics of `viewer.width()`  is?

sloriot:

PR #8548, opened 2024-10-15
Bug fix in LCC method set_automatic_attributes_management

gdamiand:
## Summary of Changes

Correct the name in LCC of the method. The previous name is set_update_attributes which is wrong.
The correct name is set_automatic_attributes_management (which is documented).

Add a non-documented internal method allowing to change the update attribute mode without correcting the attributes.

## Release Management


sloriot:

PR #8550, opened 2024-10-15
add some override keywords

sloriot:
Try to fix some warnings with gcc master


sloriot:

PR #8562, opened 2024-10-20
Replace usage of `\htmlonly` with `<div>`

albert-github:
The HTML `<div>` construct with attributes is supported by doxygen (in HTML, on other places it is ignored) and can theerfore be removed.
- the advantage is that a mismatch of `<div>` / `</div>` is notice (e.g. when they are in different `ALIASES` and one of the 2 has been forgotten, see also #8560)
- the `ALIASES`: `cgalModifBegin` and `cgalModifEnd` are not used (anymore) and are therefore removed


afabri:
What do you mean with that `cgalModifBegin` are not used. They do not appear in the master branch but they are useful when proposing changes in the manual. So they are just needed for "development".

albert-github:
Reinstated the"modif" `ALIASES`.

- what to do with the `\latexonly` part ?
- interesting question is are these 2 `ALIASES` filtered out when a release is made so they won't accidentally slip into a release?

afabri:
I think we could give up `\latexonly`  as we have no ambition to create something like pdf.  @MaelRL any opinion?

sloriot:
Successfully tested in CGAL-6.1-Ic-50

albert-github:
@sloriot @lrineau @afabri 

Any chance that this will be integrated in CGAL or is there something further to be done here

afabri:
Is all what remains removing the `\latexonly` ?

albert-github:
@afabri 

Looks like it is, so what to do with the `\latexonly` ?
(looks like to be only in what is now called `Documentation/doc/resources/1.14.0/BaseDoxyfile.in`, won't change the older versions).

afabri:
just remove  it.

albert-github:
Merged and removed `\latexonly` as well

sloriot:
Successfully tested in CGAL-6.2-Ic-122

PR #6088, opened 2021-10-28
Add functions to compute discrete curvatures

sloriot:
## Summary of Changes

Add functions to compute discrete curvatures.

Tentative TODO:
- [ ] discrete curvature for boundary vertices
- [ ] not naive way of computing curvatures when computed at all the vertices of a mesh

## Release Management

* Affected package(s): `Polygon_mesh_processing`
* Issue(s) solved (if any): -
* Feature/Small Feature (if any): [Add discrete curvature](https://cgalwiki.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/PMP-Discrete_curvatures) -- Pre-approved 2025-02-06
* Link to compiled documentation (obligatory for small feature) [compiled doc](https://cgal.geometryfactory.com/~mrouxell/PMP-discrete_curvatures/Polygon_mesh_processing/group__PMP__measure__grp.html)
* License and copyright ownership: no change

PR #8564, opened 2024-10-21
PMP:  Add vertex_angle_sum()

afabri:
## Summary of Changes

The vertex angle sum can be used as an indication for local flatness.

Writing it I am wondering if we need a generic apply functors to vertices.

The name itself must be discussed. Should it be `compute_..` ?

I would also like to see integrated this function in the demo where we color code.   I am wondering what to do with border vertices where the sum is in general not 2 pi

## Release Management

* Affected package(s): PMP
* Feature/Small Feature (if any): tbd 
* Link to compiled documentation tbd 


MaelRL:
merged with https://github.com/CGAL/cgal/pull/6088.

PR #8597, opened 2024-11-08
use assume from c++23 when possible

sloriot:


Installation/include/CGAL/config.h
```diff
@@ -156,7 +156,11 @@
 // Same for C++20
 #if __cplusplus >= 202002L || _MSVC_LANG >= 202002L
 #  define CGAL_CXX20 1
-#endif
+#endif<
+// Same for C++23
+#if __cplusplus >= 202302L || _MSVC_LANG >= 202302L
+#  define CGAL_CXX23 1
+#endif<
```
lrineau, line 163:
```suggestion
#endif
// Same for C++23
#if __cplusplus >= 202302L || _MSVC_LANG >= 202302L
#  define CGAL_CXX23 1
#endif
```

Installation/include/CGAL/config.h
```diff
@@ -351,6 +355,9 @@ using std::max;
 #endif
 
 // Macro CGAL_ASSUME and CGAL_UNREACHABLE
+#ifdef CGAL_CXX23
+#  define CGAL_ASSUME(EX) [[ assume(EX) ]]
+#else
```
lrineau, line 360: Please use [`elif`](https://en.cppreference.com/w/cpp/preprocessor/conditional) to avoid the nesting of conditionals. I find it more readable that way.

- If C++23,
- else if `__builtin_unreachable` exists,
- else if `__assume` exists...

lrineau (changes requested):

sloriot:
Successfully tested in CGAL-6.1-Ic-18

PR #8598, opened 2024-11-08
PMP: Initialize to avoid warning

afabri:
## Summary of Changes

An uninitialized variable gets assigned a value inside a `for` loop,
The compiler cannot know that we always enter the loop and [warns]( https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-15/Polygon_mesh_processing_Examples/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz) when we later use the variable.
We could use `[[assume(x)]]` to indicate that we always will enter the loop but just initializing does the job.

## Release Management

* Affected package(s): PMP 

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
+++ b/Polygon_mesh_processing/include/CGAL/Polyhedral_envelope.h
@@ -714,7 +714,7 @@ private:
   {
     Vector_3 bmin, bmax;
 
-    Orientation ori;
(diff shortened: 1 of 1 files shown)


afabri:
Seems to be fixed as no longer [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-18/Polygon_mesh_processing_Examples/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz)

sloriot:
Successfully tested in CGAL-6.1-Ic-18

PR #8609, opened 2024-11-14
Flip it if you can't collapse it

### LINKED ISSUE #8605: Failed to test mesh self-intersection using CGAL::Polygon_mesh_processing::does_self_intersect
## Issue Details
Hi there!

I've been working with the Polygon Mesh Processing Package and attempted to test whether a triangulated surface mesh has any self-intersections. However, I ran into an issue where it seems to have failed to detect the intersections between two triangles within the mesh.

This particular triangulated surface mesh consists of 5 vertices and 2 faces. Strangely enough, when I used the `bool CGAL::Polygon_mesh_processing::does_self_intersect` function, it indicated that this mesh does not self-intersect, which is contrary to what I expected based on my visual inspection of the mesh.

Here are the coordinates of the vertices for reference:
- Vertices_1 coordinates: (97.1358, -291.711, 132.143)
- Vertices_2 coordinates: (97.14, -291.719, 132.147)
- Vertices_3 coordinates: (97.1361, -291.711, 132.144)
- Vertices_4 coordinates: (97.1298, -291.719, 132.145)
- Vertices_5 coordinates: (97.1289, -291.709, 132.148)

Mesh file was attached. Any help would be greatly appreciated!

[intersection.zip](https://github.com/user-attachments/files/17712240/intersection.zip)

## Source Code:

```c++
SurfaceMesh mesh1;
CGAL::IO::read_polygon_mesh(".../two_triangles.stl", mesh1);
if (PMP::does_self_intersect(mesh1))
	std::cout << "Input mesh self-intersects" << std::endl;
else
	std::cout << "Input mesh does not self-intersect" << std::endl;
```

## Environment
* CGAL version: CGAL 6.0.1

sloriot:
if a needle is also a cap, it will be handled as a needle only. But if we can't collapse the edge, a flip will not be tried. Try the flip too.

Note that this will slow down the method as the status of a triangle is tested twice.

Fix #8605

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -757,6 +757,24 @@ bool remove_almost_degenerate_faces(const FaceRange& face_range,
     int kk=0;
     std::ofstream(std::string("tmp/n-00000.off")) << tmesh;
 #endif
+
+    auto run_cap_check = [&](halfedge_descriptor h, bool consider_for_collapse=true)
+    {
+      halfedge_descriptor cap_h = internal::is_it_a_cap(face(h, tmesh), tmesh, vpm, vcm, ecm, gt,
+                                                        cap_threshold, flip_triangle_height_threshold_squared);
+      if(cap_h != null_h)
+      {
+#ifdef CGAL_PMP_DEBUG_REMOVE_DEGENERACIES_EXTRA
+        std::cout << "\t\t But the face is a cap" << std::endl;
+#endif
+        edges_to_flip.insert(cap_h);
+      }
+      else
+      {
+        if (consider_for_collapse) next_edges_to_collapse.insert(h);
+      }
+    };
+
     while(!edges_to_collapse.empty())
     {
       // note that on the first iteration, 'h' does not indicate a known needle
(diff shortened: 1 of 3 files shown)


Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -112,14 +151,17 @@ void collect_badly_shaped_triangles(const typename boost::graph_traits<TriangleM
...
                                                            collapse_length_threshold, flip_triangle_height_threshold_squared);
 
   if(res[0] != boost::graph_traits<TriangleMesh>::null_halfedge())
   {
 #ifdef CGAL_PMP_DEBUG_REMOVE_DEGENERACIES_EXTRA
-    std::cout << "add new needle: " << edge(res[0], tmesh) << std::endl;
+    if (res[1] == boost::graph_traits<TriangleMesh>::null_halfedge())
+      std::cout << "add new needle: " << edge(res[0], tmesh) << std::endl;
+    else
+      std::cout << "add new needle (also a cap): " << edge(res[0], tmesh) << std::endl;
```
MaelRL, line 164: Do you not fill both ranges on purpose? There's still:

```
  ...
  else // let's not make it possible to have a face be both a cap and a needle (for now)
 ...
```

a little below; so, if it's a needle & a cap, you fill the needle range, but not the cap range.

MaelRL:
Here are some changes:
- Avoid a lot of redundant calls of is_it_a_needle / is_it_a_cap
  - Always avoid testing for cap-ness if it's already a needle
  - Wait for a needle to no longer be a needle, or not be treatable to actually test cap-ness
  - No need to fill both ranges at the beginning: if we start from all faces, needle-ness will be tested, and the exit towards cap-ness will happen if its not a needle
- When the mesh is modified, do not test everything immediately, but put them in the queue of the next iteration (which will test for needle-ness through the "is it still a needle?" mechanism)

Result: Fewer calls to shape predicates; for example

- master -
needle calls = 828 668
cap calls = 803 330

- PR -
needle calls = 803 554
cap calls = 795 317

Should be fewer but the base there is a fixed base cost at 735k faces and two effects are balancing each other for the calls:
- much fewer calls to shape predicates (is_it_a_needle, is_it_a_cap) by not calling until absolutely necessary (but still checking at pop time).
- when we modify the mesh, I no longer fill the CURRENT cap and needle ranges, but the next ones. The point is to really prioritize all caps   BEFORE treating needles, whereas otherwise we would treat some needles at the current iteration while there are still caps to treat.

MaelRL:
Gives a slight improvement on the quality too (PR = before last batch of commits, TMP = after last batch of commits):

Screenshot From 2024-12-07 15-11-28

MaelRL (approved):

sloriot:

PR #8611, opened 2024-11-15
Remove extra rounding mode changes for sqrt for interval

sloriot:
The only place where the function is call is in `Interval_nt::sqrt()` and that functions start with the creation of a protector.
Even worse, if we are using interval with no protection, we get an incorrect rounding mode after a call to sqrt.

### DIFF
--- a/Number_types/include/CGAL/FPU.h
+++ b/Number_types/include/CGAL/FPU.h
@@ -618,6 +618,7 @@ inline double IA_sqrt_toward_zero(double d) {
 #ifdef CGAL_ALWAYS_ROUND_TO_NEAREST
   return (d > 0.0) ? nextafter(std::sqrt(d), 0.) : 0.0;
 #else
+  CGAL_assertion(FPU_get_cw()==CGAL_FE_UPWARD);
   FPU_set_cw(CGAL_FE_DOWNWARD);
   double i = (d > 0.0) ? CGAL_IA_FORCE_TO_DOUBLE(CGAL_BUG_SQRT(CGAL_IA_STOP_CPROP(d))) : 0.0;
   FPU_set_cw(CGAL_FE_UPWARD);


mglisse:
Uh? I agree that it isn't optimal (there is already a comment saying so), it may even be wrong in the no-protection case as you mention, but completely removing the only place that sets the rounding mode to "down" cannot be right. For the lower bound, we need to call sqrt with the rounding mode set towards -inf or 0 (all other interval operations use +inf).

sloriot:
> Uh? I agree that it isn't optimal (there is already a comment saying so), it may even be wrong in the no-protection case as you mention, but completely removing the only place that sets the rounding mode to "down" cannot be right. For the lower bound, we need to call sqrt with the rounding mode set towards -inf or 0 (all other interval operations use +inf).

As stated above, the only function calling it has a protector  [here](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/Interval_nt.h#L1131) so it seems correct to me.

mglisse:
> As stated above, the only function calling it has a protector [here](https://github.com/CGAL/cgal/blob/master/Number_types/include/CGAL/Interval_nt.h#L1131) so it seems correct to me.

That protector sets the rounding mode to UP, which is good for the upper bound, but not for the lower bound, which needs the rounding mode set to DOWN. For other operations, we use tricks (-(-a)+(-b) instead of a+b for instance) so we can emulate rounding DOWN using UP, but for sqrt I am not aware of any such trick, we really need to compute sup and inf using 2 different rounding modes.

sloriot:
I wrongly thought that the protector was setting the rounding mode downward and since I got a checking failing on the rounding mode while exiting a program, I spotted this but you are right the patch is completely wrong. Thanks for checking!

sloriot:
But the error is now back so there is something I need to understand ...

mglisse:
AFAIU, there are 2 modes
1. `Interval_nt<true>`: we don't assume anything about the rounding mode (typically to nearest), sqrt sets it to up, down, calls sqrt, up, calls sqrt, and resets it to its initial value.
2. `Interval_nt<false>`: we assume that the rounding mode is up when the function is called. sqrt sets it to down, calls sqrt, up, calls sqrt.

Both versions look right, even if the first one sets the rounding mode one more time that necessary. And I think there is an assertion in debug mode that checks the precondition of version 2.

sloriot:
I found the pb, I was missing a protector.

mglisse:
I thought this assertion was already covered:
* in `sqrt(Interval_nt)`, before calling `IA_sqrt_toward_zero`, there is `typename Interval_nt<Protected>::Internal_protector P;`
* `typedef Checked_protect_FPU_rounding<Protected>   Internal_protector;`
* `Checked_protect_FPU_rounding` has in its constructor `CGAL_expensive_assertion(FPU_get_cw() == CGAL_FE_PROTECTED);`

Ah! It was an **expensive** assertion, so not enabled by default in debug mode, that explains the difference. If you enable expensive assertions, does it detect your bug without need for this patch? If the test is considered too expensive in general, does it also apply to this patch?

sloriot:
> I thought this assertion was already covered:
> 
>     * in `sqrt(Interval_nt)`, before calling `IA_sqrt_toward_zero`, there is `typename Interval_nt<Protected>::Internal_protector P;`
> 
>     * `typedef Checked_protect_FPU_rounding<Protected>   Internal_protector;`
> 
>     * `Checked_protect_FPU_rounding` has in its constructor `CGAL_expensive_assertion(FPU_get_cw() == CGAL_FE_PROTECTED);`
> 
> 
> Ah! It was an **expensive** assertion, so not enabled by default in debug mode, that explains the difference. If you enable expensive assertions, does it detect your bug without need for this patch? If the test is considered too expensive in general, does it also apply to this patch?

Locally in the sqrt function vs all the time you do an operation with EPECK for example is a huge difference I think. + we are not using `sqrt` that much. I'd like to keep it that way.

sloriot:

PR #8622, opened 2024-11-22
Testsuite: Custom platform names display in test results

SaillantNicolas:
## Summary of Changes

Update Perl scripts to support custom platform display names through CGAL_SUMMARY_NAME variable.


sloriot:
@SaillantNicolas I don't see any issue in CGAL-6.1-Ic-33, please check by yourself if it means that it means that it's "Tested".

SaillantNicolas:
> @SaillantNicolas I don't see any issue in CGAL-6.1-Ic-33, please check by yourself if it means that it means that it's "Tested".

We tested it with @lrineau , but there were issues with retrieving the names.

lrineau:
Tested in production.

PR #8629, opened 2024-11-27
Fixing ICPC usage with modern cmake and wrappers

Marc-Pierre-Barbier:
## Summary of Changes

Related to conan's conan-io/conan-center-index#25843

replaced 'fp-model strict' by 'fp-model=strict' https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/fp-model-fp.html for more compatibility with nvcc. as 'fp-model=strict' will be directly transmited to the compiler while 'fp-model strict' will have 'strict' being treated as a file input.

Also added a CXX only filter for newer cmake 3.3+

## Release Management

* Affected package(s): at least conan


sloriot:

PR #8633, opened 2024-11-29
Speeding up the boolean when some part of the mesh are identical

sloriot:
Add dedicated code in corefinement based operations when identical faces are present in both input meshes. Identical faces matching are directly passed to the output builder that can directly decide what to do with them.

The identical faces are also not passed to the intersection filtering


PMP_Boolean_operations/include/CGAL/Polygon_mesh_processing/internal/Corefinement/intersection_impl.h
```diff
@@ -1738,14 +1743,202 @@ class Intersection_of_triangle_meshes
...
+    {
+      visitor.finalize(nodes,tm1,tm2,vpm1,vpm2,{});
+      return output;
+    }
+
+    Bbox_3 bb12((std::max)(tm1_bb.xmin(),tm2_bb.xmin()), (std::max)(tm1_bb.ymin(),tm2_bb.ymin()), (std::max)(tm1_bb.zmin(),tm2_bb.zmin()),
+                (std::min)(tm1_bb.xmax(),tm2_bb.xmax()), (std::min)(tm1_bb.ymax(),tm2_bb.ymax()), (std::min)(tm1_bb.zmax(),tm2_bb.zmax()));
+
     // used only if throw_on_self_intersection == true
+    // TODO: use a hashmap or dynamic_property?
```
sloriot, line 1756: @LeoValque did you notice these sets while tuning?

sloriot: Actually it is only used in the case we want to throw if there are some self-intersections in the input meshes

sloriot:
Successfully tested in CGAL-6.3-Ic-43

sloriot:
There is a know "regression" in `corefine()` that should have all the edges of the coplanar faces marked as constrained (which is not the case for `corefine_and_foobar()`). It will be fixed in an upcoming PR.

PR #8637, opened 2024-12-02
Errors in bibliography  and incorrect usage of `\f[` command

albert-github:
- The `&` in the bibliography has to be escaped
- the doxygen command `\f[` should not be used to directly change the environment, for this the doxygen command `\f{` exists.


afabri (approved):

sloriot:

PR #8639, opened 2024-12-03
Periodic_3_mesh_3:   Avoid Making Copies

afabri:
## Summary of Changes

Turn several copies into `const&`  to avoid allocation of data members of traits object with multiprecision coordinates.   In release mode the gain is around 10%..   

To address the timeouts in the testsuite in Debug mode, I also changed cell size and approximation error.

## Release Management

* Affected package(s): Periodic_triangulation_3_mesh_3

### DIFF
--- a/Mesh_3/include/CGAL/Mesh_3/min_dihedral_angle.h
+++ b/Mesh_3/include/CGAL/Mesh_3/min_dihedral_angle.h
@@ -139,7 +139,7 @@ minimum_dihedral_angle(
 
 template <typename K>
 typename K::FT
-minimum_dihedral_angle(const typename K::Tetrahedron_3& t, K k = K() )
+minimum_dihedral_angle(const typename K::Tetrahedron_3& t , const K&  k = K())
 {
   return minimum_dihedral_angle(t[0],t[1],t[2],t[3],k);
 }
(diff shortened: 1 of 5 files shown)


Mesh_3/include/CGAL/Mesh_3/min_dihedral_angle.h
```diff
@@ -84,7 +84,7 @@ minimum_dihedral_angle(
      const typename K::Point_3& p1,
      const typename K::Point_3& p2,
      const typename K::Point_3& p3,
-     K k = K())
+     const K& k)
```
MaelRL, line 87:
```suggestion
     const K& k = K())
```

Mesh_3/include/CGAL/Mesh_3/min_dihedral_angle.h
```diff
@@ -139,7 +139,7 @@ minimum_dihedral_angle(
 
 template <typename K>
 typename K::FT
-minimum_dihedral_angle(const typename K::Tetrahedron_3& t, K k = K() )
+minimum_dihedral_angle(const typename K::Tetrahedron_3& t , const K&  k = K())
```
MaelRL, line 142:
```suggestion
minimum_dihedral_angle(const typename K::Tetrahedron_3& t , const K& k = K())
```

sloriot:
Only 1 timeout in CGAL-6.1-Ic-37 vs 13 in CGAL-6.1-I-36

Successfully tested

sloriot:
> Note that this a timeout when in Release mode with CGAL_NDEBUG. Should it be investigated?

I would not bother, `Fedora-Release` has the timeout and not `Fedora-rawide-Release` while both have similar compilers and flags, but running on different computers.

PR #8649, opened 2024-12-06
Bug-fix speeding up coplanar case

sloriot:
The new node propagation status only needs to be done the first time the node is seen, the classification will always be the same and propagation is already done.


sloriot:
Successfully tested in CGAL-6.1-Ic-37

PR #8668, opened 2024-12-18
Fix Coref Boolean for identical nested patches

sloriot:


sloriot:

PR #8669, opened 2024-12-20
Stream_support: Document that WKT can also handle 3D points

afabri:
## Summary of Changes

Fix in the documentation, namely that also 3D points may be used.

## Release Management


afabri:
While fixing this I was also checking the **Attention**  _Only Cartesian Kernels with double or float as FT are supported._
It is true that ` Simple_cartesian<Exact_rational>`  does not work, while ` Epeck`   does.   How to express this correctly?

afabri:
Successfully tested in[ 6.-Ic-55-](https://cgal.geometryfactory.com/CGAL/testsuite/results-6.1-Ic-55.shtml)  @sloriot

sloriot:
Successfully tested in CGAL-6.1-Ic-59

PR #8681, opened 2025-01-08
Fix missing requirement in the testsuite

sloriot:


sloriot:
Does not work in CGAL-6.1-Ic-59

sloriot:
The logic has changed with the new scripts, a `r` is put only if no exe was run for a test/example category.

PR #8682, opened 2025-01-08
Handle copy of non manifold umbrella in a Face_filtered_graph

sloriot:


sloriot:
Successfully tested in CGAL-6.1-Ic-59

PR #8687, opened 2025-01-14
WIP: Changes for clang on Windows 

afabri:
## Summary of Changes

Currently choosing clang in vscode leads to compilation and link errors.
The first commit of this PR addresses a compilation error.

It works in Visual Studio, so I probably only do something wrong when I configure inside vscode. Let's have a look on Thursday @lrineau 

## Release Management

### DIFF
--- a/Number_types/include/CGAL/Mpzf.h
+++ b/Number_types/include/CGAL/Mpzf.h
@@ -66,7 +66,7 @@
 
 #include <boost/cstdint.hpp>
 
-#ifdef _MSC_VER
+#ifdef _MSC_VER && (! defined(__clang__))
 #include <intrin.h>
 #pragma intrinsic(_BitScanForward64)
 #pragma intrinsic(_BitScanReverse64)
@@ -190,7 +190,7 @@ template <class T, class = void> struct no_pool {
 
 // Only used with an argument known not to be 0.
 inline int ctz (std::uint64_t x) {
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) && (! defined(__clang__))
   unsigned long ret;
   _BitScanForward64(&ret, x);
   return (int)ret;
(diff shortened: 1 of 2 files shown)


mglisse:
Longer term, we could add a version of that code using C++20 functions like [`countr_zero`](https://en.cppreference.com/w/cpp/numeric/countr_zero) (need to check that special cases, for instance 0, work properly).

afabri:
In [` CGAL/enable_third_party_libraries.h`](https://github.com/CGAL/cgal/blob/299a89f2129d8390ee73e784990bbacc7a05310f/Installation/include/CGAL/Installation/internal/enable_third_party_libraries.h#L54)   I changed locally:
```
-    (!defined _WIN32 || defined _WIN64) && \
+    (!defined _WIN32 || defined _WIN64 || (defined _MSC_VER && defined(__clang__))) && \
```

sloriot:
@afabri why WIP? shall we postpone to 6.2 or shall we test/merge as is?

Number_types/include/CGAL/Mpzf.h
```diff
@@ -66,7 +66,7 @@
 
 #include <boost/cstdint.hpp>
 
-#ifdef _MSC_VER
+#ifdef _MSC_VER && (! defined(__clang__))
```
lrineau, line 69: We should use [`BOOST_MSVC` from `<boost/config.hpp>`](https://www.boost.org/doc/libs/master/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.boost_informational_macros):

> `BOOST_MSVC`
> Defined if the compiler is really Microsoft Visual C++, as opposed to one of the many other compilers that also define `_MSC_VER`. Has the same value as `_MSC_VER`.

afabri: Just found this:   `Installation/include/CGAL/auto_link/auto_link.h:#  define BOOST_MSVC _MSC_VER`

afabri:
We can close this issue.  I made the mistake to choose the most bottom line, which is 32 bit.
image

PR #8704, opened 2025-01-21
Fix bad detection of warnings + remove outdated scripts

sloriot:
for examples/tests a warning might be reported as `t`


sloriot:
New warnings are detected in CGAL-6.1-Ic-69 and there is a testsuite :)

sloriot:
Works in CGAL-6.1-Ic-70 and CGAL-6.1-Ic-69

PR #8710, opened 2025-01-22
Aos 2 dcel extension efif

efifogel:
## Summary of Changes

The changes fix a bug in the extensions of vertex and halfedge types of the DCEL when used to instantiate Arrangement_with_history_2 or similar arrangement classes that derive from Arrangement_2.
Typically, only the face type is extended or all the 3 types (vertex, halfedge, and face) are extended, and for such extensions we have dedicated class templates. The bug shows up when trying to extend any other subset of the 3 types, and apparently, this hasn't been used much with Arrangement_with_history_2.

## Release Management

* Issue(s) solved (if any): NA
* Link to compiled documentation: NA


sloriot:
Successfully tested in 6.1-Ic-109

PR #8712, opened 2025-01-23
Fix warnings

sloriot:
Revealed by #8704

### DIFF
--- a/SMDS_3/test/SMDS_3/test_c3t3_with_features.cpp
+++ b/SMDS_3/test/SMDS_3/test_c3t3_with_features.cpp
@@ -346,7 +346,7 @@ struct Tester
     // Test vertex iterators
     //-------------------------------------------------------
     std::cout << "Test vertex iterators\n";
-    const Vertex_handle& vertex_to_modify = c3t3.vertices_in_complex_begin();
+    Vertex_handle vertex_to_modify = c3t3.vertices_in_complex_begin();
     Vertex_handle vertex_to_modify_copy = vertex_to_modify;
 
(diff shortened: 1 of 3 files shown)


SMDS_3/test/SMDS_3/test_c3t3_with_features.cpp
```diff
@@ -346,7 +346,7 @@ struct Tester
     // Test vertex iterators
     //-------------------------------------------------------
     std::cout << "Test vertex iterators\n";
-    const Vertex_handle& vertex_to_modify = c3t3.vertices_in_complex_begin();
+    Vertex_handle vertex_to_modify = c3t3.vertices_in_complex_begin();
```
janetournois, line 349: can you please remove the comment below? `vertex_to_modify` is not a dangling reference anymore with this change, it's just a `Vertex_handle`

STL_Extension/test/STL_Extension/test_Concurrent_compact_container.cpp
```diff
@@ -45,7 +45,7 @@ struct Node_1
   void set_time_stamp(const std::size_t& ts) {
     time_stamp_ = ts;
   }
-  std::size_t time_stamp_;
+  std::size_t time_stamp_ = 0;
```
sloriot, line 48: runtime errors in CGAL-6.1-Ic-70

sloriot: which makes me think that the test might be buggy

sloriot:
Successfully tested in CGAL-6.1-Ic-78

PR #8714, opened 2025-01-24
Always init time_stamp to -2

sloriot:
It is not clear to me that this is the right thing to do:
- it is done for some classes
- but in some test, a comment mention that it should be uninitialized (the test works if init to -1 but not to 0).
- Base_time_stamper is doing an init to -2


sloriot:
- Runtime error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-76/STL_Extension/TestReport_lrineau_Ubuntu-latest-GCC6.gz) (and also on other plateforms)

lrineau (approved):

sloriot:
Still errors in STL_extension test in CGAL-6.1-Ic-79

lrineau:
> Still errors in STL_extension test in CGAL-6.1-Ic-79

That's right. There was an issue with `Compact_container::reserve` and timestamps. Fixed in lastest commit fa83d0bd459e819c0861543c8ba74c872e1ce8c6.

sloriot:

PR #8715, opened 2025-01-24
boost::tie -> std::tie

sloriot:


SharonIV0x86:
I noticed that CGAL is transitioning from Boost components to C++ standard components. If there's any way I can contribute or assist with this effort, please let me know.

lrineau:
Note that `std::tie` requires the inclusion of `<tuple>`.

Installation/cmake/modules/config/support/test_BOOST.cpp
```diff
@@ -23,7 +23,7 @@
 
 using boost::tuple;
 using boost::make_tuple;
-using boost::tie;
```
MaelRL, line 26: keep it boost?

sloriot: not even sure this is still used

sloriot:

PR #8724, opened 2025-01-31
BGL - add `CGAL::dijkstra_shortest_path(vs, vt, mesh)`

janetournois:
## Summary of Changes

computes the shortest path between two vertices, using Dijkstra algorithm

Todo
- [x] write doc and integrate to 
- [x] add np
- [x] add measure to define what "shortest" means
- [x] add to ref manual top page
- [x] use in demo code

## Release Management

* Affected package(s): BGL
* [Small feature](https://cgalwiki.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/Add_CGAL::shortest_path(vs,_vt))
* Link to compiled documentation [CGAL::shortest_path_between_two_vertices()](https://cgal.github.io/8724/v0/BGL/group__PkgBGLTraversal.html#ga5627dee8a6fe4d238cb5c91e123aa763)

### DIFF
--- a/BGL/examples/BGL_surface_mesh/shortest_path.cpp
+++ b/BGL/examples/BGL_surface_mesh/shortest_path.cpp
@@ -60,6 +60,9 @@ int main(int argc, char** argv)
   CGAL::dijkstra_shortest_path(vs, vt, sm,
     std::back_inserter(halfedge_sequence));
 
+  assert(source(halfedge_sequence.front(), sm)==vs);
+  assert(target(halfedge_sequence.back(), sm)==vt);
+
   // dump
   std::cout << "Shortest path between vertices " << i0 << " and " << i1
             << " is made of " << halfedge_sequence.size() << " halfedges." << std::endl;
@@ -68,7 +71,7 @@ int main(int argc, char** argv)
   auto vpmap = get(CGAL::vertex_point, sm);
 
   std::ofstream out("shortest_path.polylines.txt");
-  out << halfedge_sequence.size() << " " << get(vpmap, source(halfedge_sequence.front(),sm));
+  out << halfedge_sequence.size()+1 << " " << get(vpmap, source(halfedge_sequence.front(),sm));
   for (const halfedge_descriptor he : halfedge_sequence)
   {
     const vertex_descriptor v = target(he, sm);
--- a/BGL/examples/BGL_surface_mesh/shortest_path.cpp
+++ b/BGL/examples/BGL_surface_mesh/shortest_path.cpp
@@ -57,7 +57,8 @@ int main(int argc, char** argv)
   }
 
   std::vector<halfedge_descriptor> halfedge_sequence;
-  CGAL::shortest_path_between_two_vertices(vs, vt, sm, halfedge_sequence);
+  CGAL::shortest_path_between_two_vertices(vs, vt, sm,
+    std::back_inserter(halfedge_sequence));
 
   // dump
   std::cout << "Shortest path between vertices " << i0 << " and " << i1
--- a/BGL/examples/BGL_surface_mesh/shortest_path.cpp
+++ b/BGL/examples/BGL_surface_mesh/shortest_path.cpp
@@ -1,0 +1,79 @@
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Surface_mesh.h>
+#include <CGAL/boost/graph/graph_traits_Surface_mesh.h>
+
+#include <CGAL/boost/graph/shortest_path.h>
+#include <CGAL/IO/polygon_mesh_io.h>
+
+
+#include <string>
+#include <vector>
+#include <fstream>
+#include <exception>
+#include <algorithm>
+
+using K = CGAL::Exact_predicates_inexact_constructions_kernel;
+using Point = K::Point_3;
+using Mesh = CGAL::Surface_mesh<Point>;
+
+using vertex_descriptor = boost::graph_traits<Mesh>::vertex_descriptor;
+using edge_descriptor = boost::graph_traits<Mesh>::edge_descriptor;
+using halfedge_descriptor = boost::graph_traits<Mesh>::halfedge_descriptor;
+
+namespace PMP = CGAL::Polygon_mesh_processing;
+namespace params = CGAL::parameters;
+
+
+// Example main
+int main(int argc, char** argv)
+{
+  const std::string filename = (argc > 1) ? argv[1] : CGAL::data_file_path("meshes/elephant.off");
+
+  // Try building a surface_mesh
+  Mesh sm;
+  bool ok = CGAL::IO::read_polygon_mesh(filename, sm);
+  if (!ok || !sm.is_valid() || sm.is_empty())
+  {
+    std::cerr << "Error: Invalid facegraph" << std::endl;
+    std::cerr << "Filename = " << filename << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  const std::size_t i0 = 0;
+  const std::size_t i1 = num_vertices(sm) / 2;
+
+
+  // Get the vertex descriptors of the source and target vertices
+  const vertex_descriptor vs = *vertices(sm).first;
+  vertex_descriptor vt;
+  std::size_t vid = 0;
+  for (const vertex_descriptor v : vertices(sm))
+  {
+    if (vid++ == i1)
+    {
+      vt = v;
+      break;
+    }
+  }
+
+  std::vector<halfedge_descriptor> halfedge_sequence;
+  CGAL::shortest_path_between_two_vertices(vs, vt, sm, halfedge_sequence);
+
+  // dump
+  std::cout << "Shortest path between vertices " << i0 << " and " << i1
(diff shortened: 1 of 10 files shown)


BGL/include/CGAL/boost/graph/shortest_path.h
```diff
@@ -0,0 +1,160 @@
...
+  internal::Stop_at_target_Dijkstra_visitor<Graph> vis;
+  vis.destination_vd = vt;
+  try
+  {
+    boost::dijkstra_shortest_paths(g, vs,
+      boost::predecessor_map(pred_pmap)
+      .visitor(vis)
+      .weight_map(w_map));
+  }
+  catch (const std::exception& e){}
```
lrineau, line 114: That looks dangerous to catch and dismiss all `std::exception` exceptions. Imagine the supplied "edge weight" has a bug in its implementation, and an assertion throws a `CGAL::Failure_exception`.  It would be silently dismissed.

Let's catch only the exception the normal behavior of the code is supposed to throw. And remove the name `e` of the unused variable to avoid a warning.

```suggestion
  catch (const internal::Dijkstra_end_exception&){}
```

sloriot:
Compilation error [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-80/Isosurfacing_3_Examples/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz)

BGL/include/CGAL/boost/graph/shortest_path.h
```diff
@@ -0,0 +1,160 @@
...
+    {
+      if (vd == destination_vd)
+        throw Dijkstra_end_exception();
+    }
+  };
+} // namespace internal
+
+/*!
+* \ingroup PkgBGLTraversal
+* Computes the shortest path between two vertices in a graph `g`
```
afabri, line 67: We have to add that the vertices must be in the same connected component

BGL/include/CGAL/boost/graph/shortest_path.h
```diff
@@ -0,0 +1,160 @@
...
+  vertex_on_path vop;
+  vop.vertex = constrained_vertices.back();
+  vop.is_constrained = true;
+  path.push_back(vop);
+
+  // Display path
+  for (auto path_it = path.begin(); path_it != path.end() - 1; ++path_it)
+  {
+    const std::pair<halfedge_descriptor, bool>
+      h = halfedge((path_it + 1)->vertex, path_it->vertex, g);
```
afabri, line 150: Can't we obtain directly the halfedges when running Dijkstra, instead of  having a sequence of vertices?

sloriot:

BGL/include/CGAL/boost/graph/shortest_path.h
afabri: I think we can by implementing a `DijkstraVisitor::edge_relaxed(e)`  and store the edge in a property map of the target vertex. See the documentation [here](https://www.boost.org/doc/libs/1_42_0/libs/graph/doc/dijkstra_shortest_paths.html):  _`vis.edge_relaxed(e, g)` is invoked on edge `(u,v)` if `d[u] + w(u,v) < d[v]`. The edge `(u,v)` that participated in the last relaxation for vertex `v` is an edge in the shortest paths tree._

BGL/include/CGAL/boost/graph/shortest_path.h
```diff
@@ -0,0 +1,190 @@
...
+* @param halfedge_sequence_oit the output iterator holding the output sequence
+* of halfedges that form the shortest path from `vs` to `vt` on `g`
+* @param np an optional sequence of \ref bgl_namedparameters "Named Parameters" among the ones listed below
+*
+*  \cgalNamedParamsBegin
+*   \cgalParamNBegin{edge_weight_map}
+*    \cgalParamDescription{a property map associating to each edge in the graph its weight or ``length''.
+*                          The weights must all be non-negative.}
+*    \cgalParamType{a class model of `ReadablePropertyMap` with `boost::graph_traits<PolygonMesh>::%edge_descriptor`
+*                   as key type and `FT` as value type.}
```
afabri, line 95: No need to introduce `FT`.    It is probably specified in the BGL what has to be provided for "adding" and "comparing" weights.

afabri: I would like to suggest
```

*    \cgalParamType{a class model of `ReadablePropertyMap` with `boost::graph_traits<PolygonMesh>::%edge_descriptor`
*                   as key type and a value type which as specified in the names parameter `distance_map`of the function
                    <A href="https://www.boost.org/doc/libs/release/libs/graph/doc/dijkstra_shortest_paths.html">`boost::graph::dijkstra_shortest_paths()`</A>,
                    with any model of `RingNumberType` fulfilling the requirements. }
```

BGL/include/CGAL/boost/graph/shortest_path.h
```diff
@@ -0,0 +1,190 @@
...
+*     \cgalParamType{a class model of `ReadablePropertyMap` with `boost::graph_traits<PolygonMesh>::%vertex_descriptor`
+*                    as key type and `std::size_t` as value type}
+*     \cgalParamDefault{an automatically indexed internal map}
+*   \cgalParamNEnd
+*  \cgalNamedParamsEnd
+*/
+template<typename Graph,
+         typename OutputIterator,
+         typename NamedParameters = parameters::Default_named_parameters>
+OutputIterator shortest_path_between_two_vertices(
```
afabri, line 110:
```suggestion
OutputIterator shortest_path(
```

afabri: @sloriot any opinion on the function name?

sloriot: `dijkstra_shortest_path()` so that it's clear it's not geodesic

BGL/include/CGAL/boost/graph/dijkstra_shortest_path.h
```diff
@@ -0,0 +1,192 @@
...
+  try
+  {
+    boost::dijkstra_shortest_paths(g, vs,
+      boost::predecessor_map(pred_pmap)
+      .visitor(vis)
+      .weight_map(w_map)
+      .vertex_index_map(vim));
+  }
+  catch (const internal::Dijkstra_end_exception& ){}
+
```
afabri, line 148: Can all what follows not be replaced with this
```

  std::list<halfedge_descriptor> path;
  vertex_descriptor t = vt;
  do {
    path.push_front(halfedge(relaxed_edges_map[t]),g);
    t = get(pred_pmap, t);
  }while (t != vs);
  for(auto he : path){
    *halfedge_sequence_oit++ = he;
  }
  return halfedge_sequence_oit;
```

BGL/include/CGAL/boost/graph/dijkstra_shortest_path.h
```diff
@@ -0,0 +1,192 @@
...
+    {
+      return "Dijkstra shortest path: reached the target vertex.";
+    }
+  };
+
+  /// Visitor to stop Dijkstra's algorithm once the given target turns 'BLACK',
+  /// that is when the target has been examined through all its incident edges and
+  /// the shortest path is thus known.
+  template<typename Graph, typename VertexEdgeMap>
+  class Stop_at_target_Dijkstra_visitor : boost::default_dijkstra_visitor
```
sloriot, line 43: if you inherit from the default I don't think you need empty function

BGL/include/CGAL/boost/graph/dijkstra_shortest_path.h
```diff
@@ -0,0 +1,160 @@
...
+* @param vs source vertex
+* @param vt target vertex
+* @param g the graph
+* @param halfedge_sequence_oit the output iterator holding the output sequence
+* of halfedges that form the shortest path from `vs` to `vt` on `g`
+* @param np an optional sequence of \ref bgl_namedparameters "Named Parameters" among the ones listed below
+*
+*  \cgalNamedParamsBegin
+*   \cgalParamNBegin{edge_weight_map}
+*    \cgalParamDescription{a property map associating to each edge in the graph its weight or ``length''.
```
lrineau, line 88:
```suggestion
*    \cgalParamDescription{a property map associating to each edge in the graph its weight or "length".
```

BGL/examples/BGL_surface_mesh/shortest_path.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+using K = CGAL::Exact_predicates_inexact_constructions_kernel;
+using Point = K::Point_3;
+using Mesh = CGAL::Surface_mesh<Point>;
+
+using vertex_descriptor = boost::graph_traits<Mesh>::vertex_descriptor;
+using edge_descriptor = boost::graph_traits<Mesh>::edge_descriptor;
+using halfedge_descriptor = boost::graph_traits<Mesh>::halfedge_descriptor;
+
+namespace PMP = CGAL::Polygon_mesh_processing;
```
sloriot, line 23:
```suggestion
```

sloriot:

PR #8734, opened 2025-02-12
Fix for foldovers so that they are not considered as coplanar

sloriot:
It is still not clear to me if we want to support those configurations as they are self-intersections.


sloriot:
Successfully tested in CGAL-6.1-Ic-132

PR #8737, opened 2025-02-13
Cmap add test gdamiand

gdamiand:
## Summary of Changes

Solve a bug for CMap and GMap when we iterate on an empty map with an iterator that needs to mark cells, and when using index version.

## Release Management

* Affected package(s): CMap GMap

### DIFF
--- a/Combinatorial_map/include/CGAL/Cell_iterators.h
+++ b/Combinatorial_map/include/CGAL/Cell_iterators.h
@@ -86,7 +86,8 @@ namespace CGAL {
                               Tag_true>::value);
       CGAL_assertion(amap.is_whole_map_unmarked(mcell_mark_number));
 
-      mark_cell<Map,i,dim>(amap, adart, mcell_mark_number);
+      if(this->cont())
+      { mark_cell<Map,i,dim>(amap, adart, mcell_mark_number); }
     }
 
     /// Destructor.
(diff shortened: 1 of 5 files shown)


gdamiand:
Maybe it is better for release branch? (but the bug is very specific thus it is ok to use master)

Combinatorial_map/include/CGAL/Cell_iterators.h
```diff
@@ -86,7 +86,8 @@ namespace CGAL {
                               Tag_true>::value);
       CGAL_assertion(amap.is_whole_map_unmarked(mcell_mark_number));
 
-      mark_cell<Map,i,dim>(amap, adart, mcell_mark_number);
+      if(this->cont())
+      { mark_cell<Map,i,dim>(amap, adart, mcell_mark_number); }
```
albert-github, line 90: Wouldn't it be better to have the test inside the function (as this is anyway provided) and this will in one go protect all calls to the function.

gdamiand: the mark_cell function takes a descriptor as input; and has a requirement that this descriptor is "valid".
thus I think the current use is correct.

sloriot:
Successfully tested in CGAL-6.1-Ic-90

PR #8738, opened 2025-02-14
Remove check doc todos

sloriot:
We can consider them as done now


sloriot:
Successfully tested in CGAL-6.1-Ic-90

PR #8739, opened 2025-02-14
Tutorial: 2D Triangulations

afabri:
## Summary of Changes

Add a tutorial for getting started with 2D Triangulations in an easier way than the User Manual.
This could as well become a section in the User Manual.


sloriot:
@afabri what is the status?

Triangulation_2/examples/Triangulation_2/Tutorial_Triangulation_2.cpp
```diff
@@ -0,0 +1,100 @@
+//! [TutoT2-include]
+#include <Exact_predicates_inexact_constructions_kernel.h>
```
sloriot:
```suggestion
#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
```

Documentation/doc/Documentation/Tutorials/Tutorial_triangulation_2.txt
```diff
@@ -0,0 +1,177 @@
...
+So we start with some `#include` and some `using` statements to define types.
+
+\snippet Triangulation_2/Tutorial_Triangulation_2.cpp TutoT2-include
+
+When we insert a set of points in the triangulation, it generates a set of finite
+vertices corresponding to the points, and the decomposition of the convex hull
+of the points in finite triangular faces.
+Additionally, the triangulation generates infinite faces, which are incident to the edges of the convex hull
+and which have as third vertex a single infinite vertex. When we draw the triangulation
+we only draw finite vertices and faces.
```
sloriot, line 41: shouldn't you mention that the points are not inserted in the order of the range?

Documentation/doc/Documentation/Tutorials/Tutorial_triangulation_2.txt
```diff
@@ -0,0 +1,177 @@
...
+
+FIGURE
+
+We inserted a range of points from a `std::array` but the insert functions
+can take any range type, e.g., a `std::vector` or `std::list`.
+When we insert an individual point we obtained a <em>vertex handle</em> `vh`.
+A handle is a pointer to a vertex object, and no surprise when we write
+`v->point()` into `std::cout` we will see `1 1` on the console.
+
+Just like the standard containers, the triangulation provides iterators to enumerate
```
sloriot, line 55: simply mention ranges?

afabri:
@albert-github can you help us to understand what is necessary for doxygen that it  adds a section **Examples**  in a reference manual page.

sloriot:
> @albert-github can you help us to understand what is necessary for doxygen that it adds a section **Examples** in a reference manual page.

What do you mean?

albert-github:
@afabri 

I don't get the question either which example where should it be etc.
Maybe this will help a little: the examples section is in doxygen based on the `\examples` command, so when the command `\example` is given you will also see and example section, see e.g. the following quick example I made (I didn't compile it to see whether or not it is 100% valid code): [example.tar.gz](https://github.com/user-attachments/files/24569423/example.tar.gz)

Documentation/doc/Documentation/Tutorials/Tutorial_triangulation_2.txt
```diff
@@ -20,7 +20,7 @@ which operates on a set of points and segments.
 Note that this tutorial is not about computational geometry, so we assume you are familiar
 with Delaunay triangulation with and without constraints.
 
-\section  Tutorial_Delaunay_2   Defining the Type of %Triangulation
+<H1>  Tutorial_Delaunay_2   Defining the Type of Triangulation</h1>
```
albert-github: The section numbers are special to CGAL, the way it is re-implemented here looses the possibility to refer to a section and the possibility of the TOC .

MaelRL:
@albert-github @sloriot 

The question was "what does it take for the list of examples using this function to show up at the bottom of the documentation of a class or a function".

For example :

[CGAL::Polygon_mesh_processing::repair_polygon_soup()](`https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__combinatorial__repair__grp.html#ga3b35133783759402828325b91ab559cc`).

--> used in the example [repair_polygon_soup_example.cpp](https://doc.cgal.org/latest/Polygon_mesh_processing/Polygon_mesh_processing_2repair_polygon_soup_example_8cpp-example.html), which is in [PMP/examples.txt](https://github.com/CGAL/cgal/blob/main/Polygon_mesh_processing/doc/Polygon_mesh_processing/examples.txt#L42)

but there is nothing at the bottom of the function's documentation, whereas we'd expect references to examples at the bottom of the function documentation:         

<hr>
image
<hr>

But in fact it seems related to the fact that there is no link in the doc output:

<hr>
image
<hr>

and that seems to be related to the fact that we use a namespace alias:

```cpp
```

If I change the example and put "CGAL::Polygon_mesh_processing::repair_polygon_soup", then the link shows up:

<hr>
image
<hr>

and I get the example showing in the function doc:

<hr>
image
<hr>

So the questions are:
- What can we do so links show up when using namespace aliases ?
- Are there other conditions such that the example links do not appear at the bottom of functions, for example for class template member functions?
- How does doxygen select which example to show in the list when there are many .cpp using that function? (CGAL::IO::read_polygon_mesh for example).

Thanks!

albert-github:
> What can we do so links show up when using namespace aliases ?

An intriguing problem.
I started with a small project (all files can be found in: [example.tar.gz](https://github.com/user-attachments/files/24695589/example.tar.gz) ) with am include file (aa.h):
```
/// \file


/// Namespace one
namespace N1
{
  /// Namespace two
  namespace N2
  {
     /// the class
     class cls
     {
        public:
          /// The function
          fie();
     };
  }
}
```

and an example file:
```
#include "aa.h"

namespace NN = N1::N2;

int example_main()
{
  NN::cls::fie();
  N1::N2::cls::fie();
  return 1;
}
```

and here we we don't see the link for the `NN` part, so I added a file using the the function as well (in aa.cpp, in the archive called aa.cpp.org):
```
// \file
#include "aa.h"

/// The namespaces
namespace NN = N1::N2;

/// the main function
int main()
{
  NN::cls::fie();
  N1::N2::cls::fie();
  return 1;
}
```
and now the link was present, so this brought me to the idea to add (ommented out in the archive):
```
namespace NN = N1::N2;
```
and removing again the `.cpp` file and now the link appeared again.
to the `.h` file

The problem is that in this case the `example.cpp` is not scanned for function / classes etc. that are defined in the used source (as is quite common in CGAL as a lot of files are documented through a separate documentation file in the `doc` directory of the project (and these files only contain prototypes).

- So the solution is to add `namespace NN = N1::N2;` to a file that is used / scanned in the documentation.


**Edit**

It looks like that for PMP the documentation is not going the `doc` directory, so here it can, probably, be added straight at the end of `Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h` (maybe needs some protection like `@ifdef DOXYGEN_RUNNING`.

albert-github:
> Are there other conditions such that the example links do not appear at the bottom of functions, for example for class template member functions?
> How does doxygen select which example to show in the list when there are many .cpp using that function? (CGAL::IO::read_polygon_mesh for example).

To be honest I was a bit amazed that this link back was working, it looks like that the solution mentioned in https://github.com/CGAL/cgal/pull/8739#issuecomment-3765251072 (so defining the `namespace NN = N1::N2;` also solves this problem, unless you can find an example that shows the contrary.


**Edit**

The "Examples" part is only shown when there is a detailed description.

MaelRL:
Thanks a lot for the analysis!

However, I am not sure I like having to add this extra namespace alias in headers (even hidden behind doxygen macros), I'd rather not use "PMP" in examples, then... but it's not a great solution either because the complete namespace is quite long.

No chance that this could be fixed from within doxygen?

albert-github:
@MaelRL 
I can imagine your objections.
It is not possible to fix this in doxygen as the examples are nor run through the "scanner.l" just through the "code.l"
I got an idea, you could add the `namespace NN = N1::N2;` part to the end of `examples.txt` outside the comment part, this is read by doxygen (scanner.l) but is not used in the compilation of CGAL and does not pollute the header files.

MaelRL:
> @MaelRL I can imagine your objections. It is not possible to fix this in doxygen as the examples are nor run through the "scanner.l" just through the "code.l" I got an idea, you could add the `namespace NN = N1::N2;` part to the end of `examples.txt` outside the comment part, this is read by doxygen (scanner.l) but is not used in the compilation of CGAL and does not pollute the header files.

Seems to work well (tested in the other PR): the links now show up in examples

<hr>
image
<hr>

and in the function, we now have example refs:

<hr>
image
<hr>

I have not seen any obvious side effect nor addition when grepping in doc_output.

Thanks!

PR #7125, opened 2022-12-16
Modify flipping criteria in PMP::isotropic_remeshing and PMP::refine

MaelRL:
### Criterion in `PMP::isotropic_remeshing()`

The criterion that takes the scalar_product of the cross products  is maybe adapted to minimize the curvature when triangulating faces,  but should_flip() is used in `PMP::isotropic_remeshing()`, `PMP::refine()`, and `PMP::remove_almost_degenerate_faces()`. These algorithms aim to produce well-shaped elements. The criterion is not adapted to these algorithms: for example, on a flat mesh the scalar product is meaningless so it will pick the diagonal which maximizes the product of the lengths and product of sines, but this might create very anisotropic elements since the sine of obtuse angles is still positive. In `PMP::isotropic_remeshing()`, we actually do not notice how bad the flip is because other operations (smoothing, collapsing, splitting) counter balance it.
 
The "new" criterion is simply the criterion used in mesh smoothing and the typical Delaunay criterion for surfaces. The old criterion is still used in `PMP::triangulate_faces()` since we care for a nice surface and not nice elements.

Below is an example of `PMP::isotropic_remeshing()` without collapse (left - input, middle - master, right - new flip criterion)

image

### Criterion in `PMP::refine()`

The current criterion is some kind of Delaunay ball (it calls `side_of_bounded_sphere()`), which might work OK for flat regions, but can produce super thin wedges (see issue: https://github.com/CGAL/cgal/issues/6982) when the mesh is not flat.

The criterion used instead is the one used in `PMP::isotropic_remeshing()` and `PMP::remove_almost_degenerate_faces()`, which is the typical angle-based surface Delaunay criterion.

Below is input, hole filling with old refine criterion, old filling with new refine criterion.

image

## Release Management

* Affected package(s): `PMP`
* Issue(s) solved (if any): fix #6982
* Feature/Small Feature (if any): -
* License and copyright ownership: no change

PR #8743, opened 2025-02-17
Better flip-check criterion for almost degenerate face removal

MaelRL:
## Summary of Changes

While trying to remove caps, `PMP::remove_almost_degenerate_faces()` does a Delaunay-like check before allowing the edge flip. Unfortunately, the current criterion (see https://github.com/CGAL/cgal/pull/7125) can prevent some edge flips that would resolve a cap because sometimes this Delaunay-like criterion can prefer a cap if it is balanced by a very small angle on the other side. Then you have alpha + beta < 180 and you're stuck with the cap, but maybe you would have liked to flip still because the largest angles created by a flip would be e.g. 120 and 120, which is not that bad.

However in this function, we are maybe not so interested in having nicely shaped elements as much as we are trying to remove the worst elements. So this PR tries to improve the worst cap angle instead, by checking only (and greedily) if we do not create a worse cap angle when we flip.

This produces better results for the cases that were problematic with alpha + beta < 180

## 1
image

## 2
image

## 3
image

## Release Management

* Affected package(s): `PMP`
* Issue(s) solved (if any): 

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
@@ -27,7 +27,7 @@
 
 #ifdef CGAL_PMP_REMOVE_DEGENERATE_FACES_DEBUG
 #include <CGAL/Polygon_mesh_processing/polygon_soup_to_polygon_mesh.h>
-#include <CGAL/IO/OFF.h>
+#include <CGAL/IO/polygon_mesh_io.h>
 #endif
 
 #include <boost/algorithm/minmax_element.hpp>
@@ -343,7 +343,8 @@ template <typename TriangleMesh, typename VPM, typename Traits>
 bool should_flip(typename boost::graph_traits<TriangleMesh>::edge_descriptor e,
                  const TriangleMesh& tmesh,
                  const VPM& vpm,
-                 const Traits& gt)
+                 const Traits& gt,
+                 const double cap_threshold)
 {
   typedef typename boost::graph_traits<TriangleMesh>::halfedge_descriptor halfedge_descriptor;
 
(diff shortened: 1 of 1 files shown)


afabri:
WIP, so what is the todo?

sloriot:
Note that it makes `hole_filling_example_LCC` fail [fo example here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-93/Polygon_mesh_processing_Examples/TestReport_lrineau_Debian-stable-Release.gz)

MaelRL:
This shouldn't be tested as it is: the API changes in the internal function (I added the capness threshold) breaks isotropic remeshing and related functions.

I didn't want to fix it until we discuss this criterion change (and with @janetournois).

afabri:
@MaelRL can you assign this.

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -343,19 +343,58 @@ template <typename TriangleMesh, typename VPM, typename Traits>
...
+  CGAL_USE(gt);
+  const Point_ref p0 = get(vpm, target(h, tmesh));
+  const Point_ref p1 = get(vpm, target(next(h, tmesh), tmesh));
+  const Point_ref p2 = get(vpm, source(h, tmesh));
+  const Point_ref p3 = get(vpm, target(next(opposite(h, tmesh), tmesh), tmesh));
+
+  // reject if flipping the edge would create a worse cap
+  auto is_worse_cap = [&](const Point_ref p, const Point_ref q, const Point_ref r)
+  {
+    internal::Is_cap_angle_over_threshold<Traits> pred;
```
sloriot: useless

Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/repair_degeneracies.h
```diff
@@ -339,6 +339,52 @@ get_best_edge_orientation(typename boost::graph_traits<TriangleMesh>::edge_descr
...
+
+  const Point_ref p0 = get(vpm, target(h, tmesh));
+  const Point_ref p1 = get(vpm, target(next(h, tmesh), tmesh));
+  const Point_ref p2 = get(vpm, source(h, tmesh));
+  const Point_ref p3 = get(vpm, target(next(opposite(h, tmesh), tmesh), tmesh));
+
+  // reject if flipping the edge would create a larger cap angle (2 new faces, 3 angles each)
+  auto is_worse_cap = [&](const Point_ref p, const Point_ref q, const Point_ref r)
+  {
+    bool res = (angle_cmp(p, q, r, p0, p1, p2) == CGAL::LARGER ||
```
MaelRL, line 370:
```suggestion
    // Note that it is useful to test all angles, including the angle(s) at p1:
    // it checks the (unlikely since p0p1p2 is a cap at p1) case where
    // the diagonal post-flip is not in the cone p0p1p2 since then one
    // the angle would be larger than the angle p0p1p2.
    bool res = (angle_cmp(p, q, r, p0, p1, p2) == CGAL::LARGER ||
```

sloriot:
Successfully tested in CGAL-6.2-Ic-85

PR #8747, opened 2025-02-19
Spelling correction

albert-github:
Spelling correction

As this is a spelling correction is in code (and outside my normal realm of PRs) I created a separate PR for it. I cannot test it,  b I'm quite confident that it won't lead to problems. I think the old name contained a typo.


sloriot:
Successfully tested in CGAL-6.1-Ic-93

PR #8751, opened 2025-02-24
T2: avoid maybe uninitialized warning

afabri:
## Summary of Changes

We know that the variable `i` will be set by `is_edge()`, we try to indicate it with a `CGAL_assume()` but that does not help the compiler, so let's just initialize it.  We get the warning [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-I-95/Triangulation_2/TestReport_lrineau_Ubuntu-latest-GCC6-Release.gz).

I am wondering if the `CGAL_assertion()` I have put makes sense. 

## Release Management


afabri:
As expected the warning is gone. @MaelRL do you agree that I can remove the assertion?

sloriot:
Successfully tested in CGAL-6.1-Ic-104 (warning is gone)

MaelRL:
> As expected the warning is gone. @MaelRL do you agree that I can remove the assertion?

If you go up in the history, there is a bit of back and forth (removed -> assume -> removed...). Is it clear why the "assume" way failed?

PR #8754, opened 2025-02-25
Number types:  Add double ceil(NT)

afabri:
## Summary of Changes

Add function `double ceil(const NT&)`.

## Release Management

* Affected package(s): Algebraic_foundations, Number_types


afabri:
I commented out several test cases in `Number_types/test/utilities.cpp`, where I do not understand why they do not compile.
I did not address `CORE::Expr`. Also I have no idea what to put for `Sqrt_extension`.

mglisse:
The standard function `std::ceil` returns `long double` when passed a `long double` and essentially tries to return the same type as its input. It is only for integers that they decided to force the return type to `double`, probably for consistency with the historical C behavior.
If you want a function that always returns a `double`, I think using a different name might be clearer.

sloriot:
@LeoValque do you want to take over this PR or shall we close it? You mentioned that your `double_cell()` function would be needed in 2 packages.

afabri:
After a discussion with @LeoValque we decided that this function would not even come in for snap rounding.

PR #8763, opened 2025-03-03
Update CMakeLists.txt

### LINKED ISSUE #8699: Inconsistent use of `cmake_minimum_required`
In the file `Documentation/doc/CMakeLists.txt` we see (like in other `CMakeLists.txt` files):
```
cmake_minimum_required(VERSION 3.12...3.29)
```

but now also the line:
```
cmake_minimum_required(VERSION 3.18..3.29) # for list(SORT ... COMPARE NATURAL)
```
has been added.

Either:
- in the file `Documentation/doc/CMakeLists.txt` the first `cmake_minimum_required` line should be replaced by the second `cmake_minimum_required` line and the second `cmake_minimum_required`  line should be removed

or
- everywhere only the second `cmake_minimum_required`  line should be used.

or 
- use second `cmake_minimum_required` line only in the top level `CMakeLists.txt` and remove all other occurrences of `cmake_minimum_required` (over 300 files).

Dhruv-Sharma01:
## Summary of Changes

In the file `Documentation/doc/CMakeLists.txt`, the first `cmake_minimum_required` line is replaced by the second `cmake_minimum_required` line, and the second `cmake_minimum_required` line is removed.

## Release Management

* Issue(s) solved (if any): Fix #8699 

### DIFF
--- a/Documentation/doc/CMakeLists.txt
+++ b/Documentation/doc/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12...3.31)
+cmake_minimum_required(VERSION 3.18...3.31) 
 project(Documentation NONE)
 
 # Check whether this cmake script is the top level one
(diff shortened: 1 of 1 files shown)


lrineau:
If the option `BUILD_DOC` is `OFF`, there is a `return` line 22:

https://github.com/CGAL/cgal/blob/8330b77234f4dcc310936f85b2c8007816b36f8b/Documentation/doc/CMakeLists.txt#L20-L23

and in that case, CMake versions 3.12 to 3.17 are still usable with CGAL. That is why there are two distinct calls to `cmake_minimum_required`.

PR #8764, opened 2025-03-03
Update orient_polygon_soup.h

### LINKED ISSUE #8693: Infinite loop in orient_polygon_soup in a degenerate case
I'm filling an issue as I observe an Infinite loop in `orient_polygon_soup()` in the function `void duplicate_singular_vertices()` that is extracting links of vertices to detect if a vertex is non-manifold (singular). I cannot reproduce the error in an isolated program, only in a complex one.

The problem happens here:
```
        if (next==v_id){
          /// turn the otherway round
          next = neighbors[0];
          do{
            P_ID other_p_id;
            std::tie(next, other_p_id) = next_ccw_vertex_around_target(next, v_id, polygons, edges, marked_edges);
            if (next==v_id) break;
            visited_polygons.insert(other_p_id);
            if(nb_link_ccs != 1)
              vertices_to_duplicate.back().second.push_back(other_p_id);
          }
          while(true);
        }
```

the pb happens because of a face having a degenerate edge, with the duplicated vertex being a T junction:

```
B----------------A
 |                     |
 |                     |
C-------DD-----E
 |           |         |
 |           |         |
 |           |         |
F------- G------H
```
f0: A-B-C-D-D-E
f1: E-D-G-H
f2: D-C-F-G

I'm not even sure we want to support such cases.

Dhruv-Sharma01:
## Summary of Changes

This pull request fixes an infinite loop in the function `duplicate_singular_vertices()` in the file `Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orient_polygon_soup.h`. The issue was caused by consecutive duplicate vertices in a cycle (degenerate edge at a T-junction). The fix introduces a check to break out of the loop when a consecutive duplicate is detected, thereby preventing the infinite loop.

## Release Management

* Issue(s) solved (if any): fix #8693 
* Feature/Small Feature (if any): Bug fix for handling degenerate edges in polygon soup orientation.

### DIFF
--- a/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orient_polygon_soup.h
+++ b/Polygon_mesh_processing/include/CGAL/Polygon_mesh_processing/orient_polygon_soup.h
@@ -388,20 +388,24 @@ struct Polygon_soup_orienter
             vertices_to_duplicate.back().second.push_back(other_p_id);
         }
         while(next!=neighbors[0]);
-
         if (next==v_id){
-          /// turn the otherway round
+          /// turn the other way round
           next = neighbors[0];
+          V_ID last_next = next; // store the initial value
           do{
             P_ID other_p_id;
             std::tie(next, other_p_id) = next_ccw_vertex_around_target(next, v_id, polygons, edges, marked_edges);
             if (next==v_id) break;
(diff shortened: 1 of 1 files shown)


Dhruv-Sharma01:
@MaelRL issue #8693 is still unresolved right?

Issue #8748, opened 2025-02-21
The find_alpha_solid function lacks proper checks for handling collinear points


DM-LPW:
## Issue Details

./main
CGAL version: 6.0.0
terminate called because of function find_alpha_solid
cgal/Alpha_shapes_2/include/CGAL /Alpha_shape_2.h 
Line: 1453
reason: When the point set is collinear and contains more than 2 points, the function enters the for loop via the number_of_version() check. However, since no adjacent faces exist for collinear points, the --_interval_face_map.end() operation returns a NULL memory address, leading to invalid memory access. The input geometry could be a linestring (1 2, 3 4, 5 6, 7 8).
## Source Code
```
#include <iostream>
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/point_generators_2.h>
#include <CGAL/algorithm.h>
#include <CGAL/Alpha_shape_2.h>
#include <CGAL/Delaunay_triangulation_2.h>
#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
#include <CGAL/Triangulation.h>
typedef CGAL::Exact_predicates_exact_constructions_kernel Kernel;
typedef CGAL::Alpha_shape_vertex_base_2<Kernel> Vb;
typedef CGAL::Alpha_shape_face_base_2<Kernel> Fb;
typedef CGAL::Triangulation_data_structure_2<Vb, Fb> Tds;
typedef CGAL::Delaunay_triangulation_2<Kernel, Tds> Triangulation_2;
typedef CGAL::Point_2<Kernel> Point_2;
typedef CGAL::Segment_2<Kernel> Segment_2;
typedef CGAL::Alpha_shape_2<Triangulation_2> Alpha_shape_2;

template <class OutputIterator>

void
alpha_edges(const Alpha_shape_2& A, OutputIterator out)
{
    auto it = A.alpha_shape_edges_begin();
    auto end = A.alpha_shape_edges_end();
    for (; it != end; ++it)
    {
        if (A.classify(*it) == 2)
        {
            *out++ = A.segment(*it);
        }
    }
}

int main() {
    double alpha = 10000;
    size_t nb_components = 1;
    double result = -1;
    // create points
    std::vector<Point_2> points;
    points.push_back(Point_2(1, 2));
    points.push_back(Point_2(3, 4));
    points.push_back(Point_2(5, 6));
    points.push_back(Point_2(7, 8));
    //compute distance
[...]

PR #8767, opened 2025-03-03
Deal with dimension != 2

sloriot:
Fixes for #8748


sloriot:
Successfully tested in CGAL-6.1-Ic-104

PR #8770, opened 2025-03-05
Fix SFINAE usage in insert and do_intersect functions on Ubuntu 24.04

ViNN280801:
## Summary of Changes

I fixed an error in the code related to the incorrect use of std::is_same. In the source code, there was an attempt to use std::is_some<int, int>::type as a function parameter, which caused a compilation error. I removed this unnecessary parameter and added type checking using SFINAE (std::enable_if).
I have Ubuntu 24.04:

```bash
loveit0@loveit0-Nitro-AN515-58:~$ uname -a
Linux loveit0-Nitro-AN515-58 6.11.0-17-generic #17~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 20 22:48:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

I used the CGAL in my project for intersections with AABB and when I compiling my code, I faced with these kind of errors:

```
/usr/local/include/CGAL/Arrangement_2/Arrangement_on_surface_2_global.h:145:27: error: ‘std::is_same<__remove_cv(int), int>::type’ is not a type
  145 |             const PointLocation& pl, ZoneVisitor &visitor,
      |                           ^~~
/usr/local/include/CGAL/Arrangement_2/Arrangement_on_surface_2_global.h:471:27: error: ‘std::is_same<__remove_cv(int), int>::type’ is not a type
  471 |             InputIterator begin, InputIterator end,
      |                           ^~~
/usr/local/include/CGAL/Arrangement_2/Arrangement_on_surface_2_global.h:1534:5: error: ‘std::is_same<__remove_cv(int), int>::type’ is not a type
 1534 |              const PointLocation& pl, std::is_same<int, int>::type)
      |     ^  
```

## Release Management

- Affected package(s): Arrangement_on_surface_2
- Issue(s) solved (if any): -
- Feature/Small Feature (if any): N/A
- License and copyright ownership: This contribution is licensed under the terms used by CGAL, and the authorship remains with the contributors.

### DIFF
--- a/Arrangement_on_surface_2/include/CGAL/Arrangement_2/Arrangement_on_surface_2_global.h
+++ b/Arrangement_on_surface_2/include/CGAL/Arrangement_2/Arrangement_on_surface_2_global.h
@@ -139,11 +139,11 @@ void insert(Arrangement_on_surface_2<GeometryTraits_2, TopologyTraits>& arr,
 //
 //
 template <typename GeometryTraits_2, typename TopologyTraits,
-          typename PointLocation, typename ZoneVisitor>
+          typename PointLocation, typename ZoneVisitor,
+          typename = std::enable_if_t<std::is_same<PointLocation, int>::value>>
 void insert(Arrangement_on_surface_2<GeometryTraits_2, TopologyTraits>& arr,
             const typename GeometryTraits_2::X_monotone_curve_2& c,
-            const PointLocation& pl, ZoneVisitor &visitor,
-            std::is_same<int, int>::type)
+            const PointLocation& pl, ZoneVisitor &visitor)
 {
   typedef GeometryTraits_2                              Gt2;
   typedef TopologyTraits                                Tt;
@@ -466,10 +466,10 @@ void insert(Arrangement_on_surface_2<GeometryTraits_2, TopologyTraits>& arr,
 // std::bool_constant< true>)'
 //
 template <typename GeometryTraits_2, typename TopologyTraits,
-          typename InputIterator>
+          typename InputIterator,
+          typename = std::enable_if_t<std::is_same<InputIterator, int>::value>>
 void insert(Arrangement_on_surface_2<GeometryTraits_2, TopologyTraits>& arr,
-            InputIterator begin, InputIterator end,
-            std::is_same<int, int>::type)
+            InputIterator begin, InputIterator end)
 {
   typedef GeometryTraits_2                              Gt2;
   typedef TopologyTraits                                Tt;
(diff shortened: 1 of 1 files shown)


sloriot:
Thanks for the PR! Do you have an example that we could compile and that is showing the error?

ViNN280801:
btw, tests are successfull

image

efifogel:
I would also like to have a look at the fix. I'll do it next week.
BW, perhaps you mentioned it and I missed it: which version of compiler are
you using?

ViNN280801:
my project uses CUDA, so, I used `nvcc_wrapper` from [kokkos](https://github.com/kokkos/kokkos/tree/develop/bin)
but what about gcc/g++

image
image

efifogel:
The current solution is using TAG dispatching, but because of a bug in one
of the compilers 20 years ago, instead of using std::true_type, it uses the
cumbersome is_same<int, int>::type.
The suggested PR is using SFINAE. (As a matter of fact, the suggested code
is a mix, and it has some incorrect pieces, but all this I find irrelevant;
see below.)

For the specific scenario—dispatching to one of two internal
implementations based on a compile-time trait—*(Tag Dispatching) is
generally considered better* than the SFINAE-based option (which uses via
enable_if).

However, using C++17 or newer, there is a third option—if constexpr—that is
superior to both.

More specifically, we have the following interface function:

template <typename GeometryTraits_2, typename TopologyTraits, typename
Curve, typename PointLocation, typename ZoneVisitor>
void insert(Arrangement_on_surface_2<GeometryTraits_2, TopologyTraits>& arr,
            const Curve& c,
            const PointLocation& pl,
            ZoneVisitor &visitor);

We need to dispatch based on the type `Curve`.
The better way is:

{
  using Gt2 = GeometryTraits_2;
  using X_monotone_curve_2 = typename Gt2::X_monotone_curve_2;
  using Is_x_monotone = typename std::is_same<Curve, X_monotone_curve_2>::type;
                                                        ;
  if constexpr (Is_x_monotone) {
    // Handling x-monotone implementation
  }
  else {
    // Handling non x-monotone implementation
  }
}

So, no, either change to the better option, or delete this PR, and I'll
change to the better option, when I get a chance.

Efi

                         _/

sloriot:
With #9303 being merged into main, you should no longer have the issue. I'm closing the PR, do not hesitate to come back to us if you are still having issue when compiling with the main branch.

ViNN280801:
@sloriot, ok, nice, thank you!

ViNN280801:
Hi, @efifogel,

I sincerely apologize for the long silence - I only noticed your comment now, and I'm sorry for not responding sooner.

Thank you for the thorough and thoughtful review. You're absolutely right: `if constexpr` is the cleanest solution here, and I should have gone that route from the start. Since you mentioned you'd handle it when you get a chance, I'll close this PR to avoid any confusion - it makes much more sense for it to land with the proper fix.

I'd love to contribute to CGAL again in the future. Thanks again for taking the time to explain the reasoning so clearly - it was educational for me.

Best regards

PR #8779, opened 2025-03-10
fix an issue with the offset meshing plugin

lrineau:


sloriot:
Successfully tested in CGAL-6.1-Ic-109

PR #8784, opened 2025-03-13
Fixed TODO: find a better name in AABB_tree/include/CGAL/AABB_tree/internal/AABB_node.h 🚀

Qt-puppy:
- Changed current method name from `traversal_with_priority_and_group_traversal` to `grouped_priority_traversal`, which is more concise and meaningful.


Qt-puppy:
@sloriot Just a friendly follow-up—whenever you get a chance, I’d love your feedback on this PR.

PR #8785, opened 2025-03-14
Tetrahedral_remeshing - fix edges accidentally removed from complex

janetournois:
## Summary of Changes

Fixes for split step in tetrahedral remeshing :
+ Some complex edges could get removed from complex before checking that the splitting operation was totally allowed
+ it could happen that midpoint would invert sub-cells during split, for very small volumes. This PR proposes a trick to try other possible refinement points

## Release Management


sloriot:
Successfully tested in CGAL-6.1-Ic-114

PR #8793, opened 2025-03-20
SMDS: Allow whitespace in the line

afabri:
## Summary of Changes

Make the function `read_MEDIT()`  more tolerant.

## Release Management

* Affected package(s): SMDS_3

### DIFF
--- a/SMDS_3/include/CGAL/SMDS_3/tet_soup_to_c3t3.h
+++ b/SMDS_3/include/CGAL/SMDS_3/tet_soup_to_c3t3.h
@@ -611,9 +611,11 @@ bool build_triangulation_from_file(std::istream& is,
       continue;
     }
 
-    if(line == "Vertices")
+    if(line.find("Vertices") != std::string::npos)
     {
       is >> nv;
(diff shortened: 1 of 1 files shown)


afabri:
We have another issue to fix. When I uncomment the calls to  writing the mesh in the example [mesh_polyhedral_complex_sm.cpp](https://github.com/CGAL/cgal/blob/6.0.x-branch/Mesh_3/examples/Mesh_3/mesh_polyhedral_complex_sm.cpp), and feed `out.mesh`     into the example [c3t3_example.cpp](https://github.com/CGAL/cgal/blob/6.0.x-branch/SMDS_3/examples/SMDS_3/c3t3_example.cpp), and then load `after_remeshing.mesh`  into `CGALlab.exe`   many border facets are missing:

image

afabri:
Note that when loading `elephant.mesh`   which is the default everything is fine, so it must have to do with dealing with domains.

sloriot:
Successfully tested in CGAL-6.1-Ic-114

PR #8810, opened 2025-03-28
Fix Compare_xy_2 for the generic 3D projection traits

MaelRL:
## Summary of Changes

The previous implementation defined `Compare_xy_2` as simply `Compare_xyz_3`. This does define an order over the projected points, but it has a strong downside: it ignores the direction of the normal, so its order can be the opposite of the lexicographical order
would give. This new order is in conflict with the existing `Compare_x_2` and `Compare_y_2`, which take the normal's direction into account.

(Also with the `Less_x_2`, `Less_y_2`, and `Less_xy_2` family, which also exist for the projection traits.)

## Release Management

* Issue(s) solved (if any): -


sloriot:

Issue #8759, opened 2025-02-28
CGAL Bug or GCC6 Compiler Bug ?


afabri:
## Issue Details

Error message in a master branch [testsuite](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-I-95/Convex_hull_3/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz)

## Environment

* Operating system (Windows/Mac/Linux, 32/64 bits):
* Compiler:
* Release or debug mode:
* Specific flags used (if any):
* CGAL version:  CGAL-6.1-I-95, that is github master branch
* Boost version:
* Other libraries versions if used (Eigen, TBB, etc.):

afabri:
What happens is that [this overload](https://github.com/CGAL/cgal/blob/299a89f2129d8390ee73e784990bbacc7a05310f/Convex_hull_3/include/CGAL/convex_hull_3.h#L1083) of `convex_hull_3()` should be picked, but it isn't'

lrineau:
That is a CGAL bug. The [call is](https://github.com/CGAL/cgal/blob/ad7e4754aa55c275fbe6dfcd3b52d73abe3217e7/Convex_hull_3/test/Convex_hull_3/test_quickhull_indexed_triangle_set_3.cpp#L23):

```c++
  typedef  std::vector<Point_3> Vertices;
  typedef std::vector<std::array<int,3> > Faces;

  Vertices vertices;
  Faces faces;

  std::vector<Point_3> points;
  // [...]
  CGAL::convex_hull_3(points.begin(), points.end(), vertices, faces);
```

There are two possible matches:

- https://github.com/CGAL/cgal/blob/299a89f2129d8390ee73e784990bbacc7a05310f/Convex_hull_3/include/CGAL/convex_hull_3.h#L993-L996
- https://github.com/CGAL/cgal/blob/299a89f2129d8390ee73e784990bbacc7a05310f/Convex_hull_3/include/CGAL/convex_hull_3.h#L1082-L1088

And none of them is rejected by the `std::enable_if_t`.

afabri:
Well, all other compilers succeed.   Do you have an idea for a fix @lrineau ?

sloriot:
I tried to reproduce it with gcc master but it could not. I would suggest to remove the traits and introduce a named parameter (the PolygonMesh parameter is anyway missing the VPM parameter).

afabri:
[...]

PR #8818, opened 2025-04-01
missing include in boost + extra care for overload resolution

sloriot:
Backward compatible fix for #8759

Thanks for the idea @mglisse


sloriot:
The other PR is still bringing a better API

sloriot:

PR #8820, opened 2025-04-02
Workaround bug in gcc

sloriot:
Fixes [these errors](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-122/Alpha_wrap_3/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz)

cc @mglisse

### DIFF
--- a/STL_Extension/include/CGAL/Compact_container.h
+++ b/STL_Extension/include/CGAL/Compact_container.h
@@ -177,6 +177,8 @@ namespace internal {
     template <typename Element>
     static void set_erase_counter(Element &, unsigned int) {}
     template <typename Element>
+    static void restore_erase_counter(Element*, unsigned int) {}
+    template <typename Element>
     static void increment_erase_counter(Element &) {}
   };
 
@@ -193,9 +195,21 @@ namespace internal {
     }
 
     template <typename Element>
+    static unsigned int erase_counter(Element* e)
+    {
+      return e->erase_counter();
+    }
+
+    template <typename Element>
     static void set_erase_counter(Element &e, unsigned int c)
     {
       e.set_erase_counter(c);
+    }
+
+    template <typename Element>
+    static void restore_erase_counter(Element* e, unsigned int c)
+    {
+      e->set_erase_counter(c);
     }
 
     template <typename Element>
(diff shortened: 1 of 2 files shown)


lrineau:
1. Has the code (without the `volatile`) been checked with `[ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html)`?
2. I really would like that workaround to be conditioned on a check of the culprit (major, or major+minor) version of gcc.
3. Has a bug been filled on gcc?

sloriot:
AFAIU, we have the issue starting with an internal release of gcc 14 up to now. I'm working on the repro but it's not that simple.

I have an alternative fix:

`unsigned int m_erase_counter=0;` instead of `volatile unsigned int erase_cou

sloriot:
I have a repro using only the Compact_container, I guess it's still too large?

sloriot:
I thought that I could nail it down to:
https://gist.github.com/sloriot/a22d7576c9da6eda1daaa12de2194a60
but even the gcc of the system is not returning the expected value, so I guess I made a mistake simplifying.
@lrineau might be worth discussing this issue when you are available...

sloriot:
All errors [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-122/Alpha_wrap_3/TestReport_lrineau_Ubuntu-GCC_master_CXX20-Release.gz) are fixed in CGAL-6.1-Ic-123 (even if there is a warning about ++ on volatile).

afabri:
I think it would be helpful to put in the description what is the problem, and what the fix is.

sloriot:
Successfully tested in CGAL-6.1-Ic-128

lrineau:
@slorio This bug fix should be backported to our release branches, no?

sloriot:
if you are using gcc master you are must likely not using an old CGAL version.

lrineau:
> if you are using gcc master you are must likely not using an old CGAL version.

[gcc-15 is already released](https://gcc.gnu.org/gcc-15/). It is not a development version. And CGAL-6.1 is not released.

On all updated Linux distributions, CGAL is version 6.0.1 and gcc and version 15.1, and any user of CGAL triangulations, on those Linux distributions, is impacted by the runtime bugs.

lrineau:
This fix should be at least backported to CGAL-6.0.x.

lrineau:
> > if you are using gcc master you are must likely not using an old CGAL version.
> 
> [gcc-15 is already released](https://gcc.gnu.org/gcc-15/). It is not a development version. And CGAL-6.1 is not released.
> 
> On all updated Linux distributions, CGAL is version 6.0.1 and gcc and version 15.1, and any user of CGAL triangulations, on those Linux distributions, is impacted by the runtime bugs.

#8876 is actually such an example, with a user on Fedora 42.

PR #8830, opened 2025-04-06
// Fix polygon orientation function to handle near-degenerate cases v…

### LINKED ISSUE #8828: optimal_convex_partition_2 fails with very simple polygon
## Issue Details

CGAL error: precondition violation!
Expression : is_simple_2(first, last, traits)
File       : /opt/homebrew/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
Line       : 542
Explanation: 
Refer to the bug-reporting instructions at https://www.cgal.org/bug_report.html
libc++abi: terminating due to uncaught exception of type CGAL::Precondition_exception: CGAL ERROR: precondition violation!
Expr: is_simple_2(first, last, traits)
/opt/homebrew/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
Line: 542

## Source Code

void test(){
  
    using Kern = CGAL::Exact_predicates_inexact_constructions_kernel;
    using Tra = CGAL::Partition_traits_2<Kern>;
    Tra::Polygon_2  polygon;
    std::list<Tra::Polygon_2>   partition_polys;

    const std::vector<double> vec = {-640,-160,-640,-384,-448,-384,-448,-448,-320,-448,-320,-384
    -128,-384,-128,-152,-128,-128,-128,0,-128,24,-128,256,-320,256,-448,64,-320,64,-320,-192,-448,-192,-512,-192,-512,-160
    };
    
    for (unsigned i = 0; i < vec.size(); i += 2) {
        polygon.push_back(Tra::Point_2{vec[i], vec[i + 1]});
    }
       
    assert(polygon.orientation() == CGAL::Orientation::COUNTERCLOCKWISE);
    assert(polygon.is_simple());

    CGAL::optimal_convex_partition_2(polygon.vertices_begin(),
                                    polygon.vertices_end(),
                                    std::back_inserter(partition_polys));
 
    assert(CGAL::partition_is_valid_2(polygon.vertices_begin(),
                                     polygon.vertices_end(),
                                     partition_polys.begin(),
                                     partition_polys.end()));
 
}
## Environment

* OS: Mac 16.3
* Compiler: Clang (system default C++)
* CGAL version: 6.0.1
* Boost version: 1.87.0_1

Abhishek-Atole:
# Fix polygon orientation function to handle near-degenerate cases via precondition fallback

## Summary of Changes
Adds robust error handling to the `orientation_2` function to handle cases where the `is_simple_2` precondition check fails during polygon partitioning operations, even when the polygon passes `polygon.is_simple()` initially.

The issue occurs due to numerical precision problems with nearly collinear points. The solution implements a try-catch approach that first attempts the original precondition check, but gracefully falls back to the `orientation_2_no_precondition` version if the check fails.

This prevents the partitioning algorithms (particularly `optimal_convex_partition_2`) from failing with seemingly simple polygons, making CGAL more robust for real-world applications with imprecise geometry.

## Release Management
- **Affected package(s):** Polygon
- **Issue(s) solved:** fix #8828
- **Feature/Small Feature:** N/A
- **License and copyright ownership:** Same as the existing file

### DIFF
--- a/Polygon/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
+++ b/Polygon/include/CGAL/Polygon_2/Polygon_2_algorithms_impl.h
@@ -539,8 +539,21 @@ Orientation orientation_2(ForwardIterator first,
                           ForwardIterator last,
                           const Traits& traits)
 {
-  CGAL_precondition(is_simple_2(first, last, traits));
-  return Polygon::internal::orientation_2_no_precondition(first, last, traits);
+  // The simplicity check can fail during partitioning even when polygon.is_simple() passes
+  // due to numerical precision issues with nearly collinear points
+  
+  try {
+    // Try with the original precondition first
+    CGAL_precondition(is_simple_2(first, last, traits));
+    return Polygon::internal::orientation_2_no_precondition(first, last, traits);
+  }
+  catch(const CGAL::Precondition_exception&) {
+    // If the simplicity check fails, fall back to the no_precondition version
+#ifdef CGAL_POLYGON_DEBUG
+    std::cerr << "Warning: orientation_2 called with non-simple polygon" << std::endl;
+#endif
+    return Polygon::internal::orientation_2_no_precondition(first, last, traits);
+  }
 }
 
 } //namespace CGAL


sloriot:
I don't see any added value to the patch, precondition will already provide enough information without the duplicated test.

PR #8832, opened 2025-04-07
Lab - Fixed loading of OpenMesh files without selection

soesau:
## Summary of Changes

A `Scene_polyhedron_selection_item`  is created when loading an OpenMesh file. When there was no selection inside the file, this selection item was neither added to the scene nor deleted. Thus, it still received Qt signals and caused crashes when the actual mesh was removed.

## Release Management

* Affected package(s): CGALlab


afabri:
@sloriot tested locally. The bug is not detected by the testsuite, nor it's fix. Read the diff and maybe merge in master.

PR #8834, opened 2025-04-08
fix deprecation warning

sloriot:
Fix this deprecation [warning]


sloriot:
Successfully tested in CGAL-6.1-Ic-128

Issue #8344, opened 2024-07-08
A question about read_OBJ(): How to read an OBJ mesh with normals and UVs?


qwerty10086:
According to the document, it seems this funciton does not provide any parameters about normals or UVs. Is there another way to meet this need?

MaelRL:
Currently there is no way to read this information with the OBJ reader provided by CGAL.

PR #8835, opened 2025-04-08
Internal support for parsing texture and normal indices in OBJ loader and lil typo

mehull-26:
## Summary of Changes

This PR adds the ability to parse normals and texture coordinates from .obj files using the existing face parsing logic. This is currently internal-only, and not exposed in the documented API or parameter sets.

The goal is to fully integrate OBJ normal and texture support in a future PR, once the writer logic is updated, and a broader set of test cases are validated.

- Internal parsing logic implemented
- Unit test added for face parsing with v/vt/vn syntax
- Not exposed to public interface yet

Why not public yet?
- Writer does not yet support output of texture and normal attributes.
- Further edge case testing is necessary before public exposure.

Once these conditions are met, the feature will be promoted to the public API.

## Release Management

* Affected package(s): [*Stream Support*](https://github.com/CGAL/cgal/tree/master/Stream_support)
* Issue(s) solved (if any): partially worked on [#8344](https://github.com/CGAL/cgal/issues/8344)
* Feature/Small Feature (if any): added capability of reading normals and texture coordinates in OBJ file format (not yet exposed to public API though, only kept to internal::read_obj() )
* Link to compiled documentation (obligatory for small feature) [**Not yet exposed to the public API to write documentation for**](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership: Same as CGAL

### DIFF
--- a/Documentation/doc/Documentation/Developer_manual/Chapter_intro.txt
+++ b/Documentation/doc/Documentation/Developer_manual/Chapter_intro.txt
@@ -225,7 +225,7 @@ functionalities for circles, circular arcs living on a 3D sphere
 
 \cgal currently provides several models for the \cgal 2D and 3D kernel
 concept, and one model for the 2D circular and the 3D spherical kernel
-concepts.. Those are again parameterized and differ in their
+concepts. Those are again parameterized and differ in their
 representation of the geometric objects and the exchangeability of
 the types involved.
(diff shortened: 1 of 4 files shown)


mehull-26:
**updated it with a cleaner PR at [#8855](https://github.com/CGAL/cgal/pull/8855)**

PR #8841, opened 2025-04-11
Restrict shared vertices on the boundary

sloriot:
function internally used not already exposed


sloriot:
Successfully tested in CGAL-6.1-Ic-132

PR #8444, opened 2024-08-29
Gsoc2024 basic viewer glfw tgrillon

tgrillon:
##### Features

The main addition of this pull request is a new basic viewer based on GLFW. Its purpose is to provide the same features as the Qt viewer while using a lighter API. 

Here is a summary of the main features that both viewers offer:
###### Camera features
- Two type of camera: Orbiter & Free-fly (FPS).
- Two camera mode: Perspective view (3D) & Orthographic view (2D).
- FOV modification (40-90).
- Translate/rotate the visualized structure.
- Rotation/translation speed modification. 
- Zoom in/out.
- Align camera to world axis. 
- Align camera to clipping plane.
- Recenter the scene.
- Reset position/orientation.
- Rotation constraint axis (`up`, `right` and `forward` axis). 
###### Clipping plane features
- Four modes: 
  - `CLIPPING_PLANE_OFF`: Disables the clipping plane.
  - `CLIPPING_PLANE_SOLID_HALF_TRANSPARENT_HALF`: Renders half of the clipped object as solid and half as transparent.
  - `CLIPPING_PLANE_SOLID_HALF_WIRE_HALF`: Renders half of the clipped object as solid and half with wireframe.
  - `CLIPPING_PLANE_SOLID_HALF_ONLY`: Only renders the positive side of the clipped object.
- Align clipping plane to camera. 
- Rotate/Translate clipping plane. 
- Translate clipping plane along its normal. 
- Translate clipping plane along camera forward direction. 
- Rotation constraint axis (`up` and `right` axis). 
- Show/Hide clipping plane rendering. 
###### Scene features
- Show/Hide `edges`/`vertices`/`faces`/`lines`/`rays`/`mesh triangles`.
- Single color mode: 
  - Draw primitives using their default color defined in the graphics scene.
- Multiple color mode: 
  - Draw primitives using their defined color. 
- Draw edge as cylinder.
- Draw vertex as sphere. 
- Draw vertex or face normal.
- Inverse normal. 
- Use mono or direction based color for normal.
- Show/Hide XY grid.  
- Show/Hide world axis.
- Take a screenshot.   
###### Feature only in GLFW viewer
[...]

PR #8850, opened 2025-04-18
Basic viewer improvements

gdamiand:

## Summary of Changes

Some improvements to the basic viewer.

Following the GSOC 2024 project, see PR #8444, extract all the modifications that concern the current basic_viewer qt (and keep in the original PR the new basic viewer glfw that could be considered later).

## Changes to the documentation

  - in `Graphics_scene`,  functions to set/get default colors: [diff](https://github.com/CGAL/cgal/pull/8850/files#diff-74c5383fef398c7cb3dbfaa8cbc836dbc5a3ad0a7b450442948e7e32d54ced47), and [documentation](https://cgal.github.io/8850/doc/Basic_viewer/classCGAL_1_1Graphics__scene.html)
  - in `Basic_viewer`, same functions, plus functions for sizes: [diff](https://github.com/CGAL/cgal/pull/8850/files#diff-cc9bfe7d72d61000248f90b7648c07decb71fd8a400ca54e3ecc8c808b0f70fa), and [documentation](https://cgal.github.io/8850/doc/Basic_viewer/classCGAL_1_1Qt_1_1Basic__viewer.html).

## Release Management

* Affected package(s): Basic_viewer


sloriot:
Are you sure it is a good idea to remove the target `CGAL::CGAL_Basic_viewer`? I understand why you added the `_Qt` version but I think it would be better to still have the old target with a default choice (qt or glfw if both are available).
Side question why not `CGAL::Basic_viewer` for the target name?

sloriot:
> Are you sure it is a good idea to remove the target `CGAL::CGAL_Basic_viewer`? I understand why you added the `_Qt` version but I think it would be better to still have the old target with a default choice (qt or glfw if both are available). Side question why not `CGAL::Basic_viewer` for the target name?

you can even configure this behavior with a cmake config flag. Then no modif of the cmakelists are needed.

gdamiand:
Ok to keep the target CGAL::CGAL_Basic_viewer.
(this name was decided in the previous small feature)

sloriot:
Tests in 6.1-Ic-138 looks good. We are missing the implementation of the target.

lrineau:
> Ok to keep the target CGAL::CGAL_Basic_viewer. (this name was decided in the previous small feature)

That name is consistent with the current naming scheme:

  - `CGAL::CGAL`,
  - `CGAL::CGAL_Qt6`,
  - `CGAL::CGAL_Core`,
  - `CGAL::CGAL_ImageIO`, and thus
  - `CGAL::CGAL_Basic_viewer`.

I agree we would transition to another naming scheme without the `CGAL_` prefix:

  - `CGAL::CGAL`,
  - `CGAL::Qt6`,
  - `CGAL::Core`,
  - `CGAL::ImageIO`,
  - `CGAL::Basic_viewer`.

There is a way to keep the old names as well as CMake target aliases.

gdamiand:
> Tests in 6.1-Ic-138 looks good. We are missing the implementation of the target.

Done.

sloriot:
I also wouldn't modify the CMakeLists of all examples/tests. But I can make that change myself.

gdamiand:
> I also wouldn't modify the CMakeLists of all examples/tests. But I can make that change myself.

Not sure. I think it maybe better to use the "new" name in our examples, and keep the alias to keep backward compatibility ?

sloriot:
IMO, we should make the choice at the cmake level ("globally"), and update the alias accordingly. You are still able to force it by using the dedicated targets. Do you see what I mean?

sloriot:
was the plan to have it in 6.1?

gdamiand:
> was the plan to have it in 6.1?

Why not?

lrineau:
Currently, with the renaming with `*_Qt`, that is difficult to see what are the actual contributions of this pull-requests. Maybe that is only technical, like a bug-fix, without any addition to the documentation.

gdamiand:
There are many improvements of the basic viewer: better shaders; can view vertices/edges with spheres/cylinders; can view normals ...

No modification of the API, and thus no modification of users/refs manuals.

lrineau (approved):
I pre-approve theses changes.

sloriot:
Successfully tested in CGAL-6.1-Ic-151

PR #8851, opened 2025-04-18
SMDS_3 - add `c3t3.set_triangulation()`

janetournois:
When loading a .mesh file to a `Triangulation_3`, and then a `C3t3`, as done in [this example](https://doc.cgal.org/latest/SMDS_3/SMDS_3_2tetrahedron_soup_to_c3t3_example_8cpp-example.html) leads to a `C3t3` for which `number_of_facets()` and `number_of_cells()` return 0, because the internal counters have not been updated by `c3t3.triangulation() = tr`.

It is not clear to the user why these numbers are 0, and the function `rescan_after_load_of_triangulation()` (which fixes the counters) is not documented.

## Summary of Changes

~~Add the use of `rescan_after_load_of_triangulation()` in an example.~~

~~The other option would be to use it automatically in the functions that load a triangulation.~~

~~@lrineau @MaelRL do you have an opinion?~~

~~Todo : discuss the API~~

Add and document a function `c3t3.set_triangulation(const Tr& tr)` that internally calls `rescan_after_load_of_triangulation()`, and is easier to use

## Release Management

* Affected package(s): SMDS_3

### DIFF
--- a/SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
+++ b/SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
@@ -328,12 +328,18 @@ public:
   const Triangulation& triangulation() const { return tr_; }
 /// @}
 
+#ifndef DOXYGEN_RUNNING
 /// \name Non const access
 /// @{
     /// returns a reference to the triangulation
+    /// \cgalAdvancedBegin
+    /// This function should only be used by advanced users: it merely swaps the triangulation without
+    /// rebuilding critical C3T3 information such as the number of simplexes in complex.
+    /// On the other hand, this is performed by `set_triangulation()`
+    /// \cgalAdvancedEnd
   Triangulation& triangulation() { return tr_; }
 /// @}
-
+#endif
 
 /// \name Modifiers
 /// @{
--- a/SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
+++ b/SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
@@ -351,6 +351,14 @@ public:
     far_vertices_.clear();
   }
 
+  /** sets the internal triangulation to \p tr
+  */
+  void set_triangulation(const Triangulation& tr)
+  {
+    tr_ = tr;
+    rescan_after_load_of_triangulation();
+  }
+
   /** adds cell \p cell to the 3D complex, with subdomain index \p index
   */
   void add_to_complex(const Cell_handle& cell, const Subdomain_index& index)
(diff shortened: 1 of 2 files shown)


sloriot:
Successfully tested in CGAL-6.1-Ic-136

sloriot:
@janetournois  Mergeable?

janetournois:
The function is not documented, so I would say no.
We need to discuss it!

SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
```diff
@@ -351,6 +351,14 @@ class Mesh_complex_3_in_triangulation_3
...
   }
 
+  /** sets the internal triangulation to \p tr
+  */
+  void set_triangulation(const Triangulation& tr)
+  {
+    tr_ = tr;
+    rescan_after_load_of_triangulation();
+  }
+
```
lrineau, line 361: add a version doing a move:

```suggestion
  void set_triangulation(const Triangulation& tr)
  {
    tr_ = tr;
    rescan_after_load_of_triangulation();
  }
  /** sets the internal triangulation to \p tr
  */
  void set_triangulation(Triangulation&& tr)
  {
    tr_ = std::move(tr);
    rescan_after_load_of_triangulation();
  }

```

janetournois:
> should not we remove the `triangulation()` returning a reference to non-const?

It is used quite a lot in Tetrahedral_remeshing (maybe also in Mesh_3) so I would be against that change.

sloriot:
Successfully tested in CGAL-6.1-Ic-151

SMDS_3/include/CGAL/Mesh_complex_3_in_triangulation_3.h
```diff
@@ -328,12 +328,18 @@ class Mesh_complex_3_in_triangulation_3
...
+#ifndef DOXYGEN_RUNNING
 /// \name Non const access
 /// @{
     /// returns a reference to the triangulation
+    /// \cgalAdvancedBegin
+    /// This function should only be used by advanced users: it merely swaps the triangulation without
+    /// rebuilding critical C3T3 information such as the number of simplexes in complex.
+    /// On the other hand, this is performed by `set_triangulation()`
+    /// \cgalAdvancedEnd
   Triangulation& triangulation() { return tr_; }
```
lrineau, line 340: I do not understand:
> it merely swaps the triangulation without...

The function just returns the reference to the non-const triangulation. It does not swap anything. What was the intent?

MaelRL: I went too fast, it was meant in the context of using the reference to assign a different triangulation, i.e. something like, "If you use this reference to call the change the triangulation, [...]"

PR #8857, opened 2025-04-22
feat: read support for vn and vt added for OBJ file format

mehull-26:
## Summary of Changes

This PR adds the ability to parse normals and texture coordinates from .obj files using the existing face parsing logic. This is currently internal-only, and not exposed in the documented API or parameter sets.

The goal is to fully integrate OBJ normal and texture support in a future PR, once the writer logic is updated, and a broader set of test cases are validated.

- Internal parsing logic implemented
- test added for face parsing with v/vt/vn syntax
- Not exposed to public interface yet

Why not public yet?
- Writer does not yet support output of texture and normal attributes.

Once these conditions are met, the feature will be promoted to the public API.

## Release Management

* Affected package(s): [*Stream Support*](https://github.com/CGAL/cgal/tree/master/Stream_support)
* Issue(s) solved (if any): partially worked on [#8344](https://github.com/CGAL/cgal/issues/8344)
* Feature/Small Feature (if any): added capability of reading normals and texture coordinates in OBJ file format (not yet exposed to public API though, only kept to internal::read_obj() )
* Link to compiled documentation (obligatory for small feature) [**Not yet exposed to the public API to write documentation for**](httpssss://wrong_URL_to_be_changed/Manual/Pkg)
* License and copyright ownership: Same as CGAL

### DIFF
--- a/Stream_support/include/CGAL/IO/OBJ.h
+++ b/Stream_support/include/CGAL/IO/OBJ.h
@@ -24,6 +24,8 @@
 #include <boost/range/value_type.hpp>
 #include <CGAL/Named_function_parameters.h>
 
+#include <CGAL/Kernel_traits.h>
+
 #include <algorithm>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


sloriot:
@afabri is it something that we want to support and are you willing to review the PR or come up with a simpler version? A CLA would be needed to integrate this version.

afabri:
Hello @mehull-26, 
Does this really correspond to a need you have, or is it an effort to bring you in a good position for an upcoming Google Summer of Code ? 
We appreciate your effort, but it would be a partial contribution.  It would be "only"  for a polygon soup and not for a `FaceGraph`.
The public API should have named parameters for passing references to the vectors, etc.     And if added, there would be functionality for  the `OBJ` but not for the `PLY` format. For the latter the API for a `FaceGraph`  has the feature to read additional properties.
So I tend to close this pull request.

mehull-26:
@afabri, it was an effort for GSOC, though it was last year. Yes, I understand, and personally, I won't really appreciate my code there either. I was exploring the organisation and trying to understand the codebase, and I may not have been consistent with my codebase. IMO, it would be better to lose it.

afabri:
> @afabri, it was an effort for GSOC, though it was last year. Yes, I understand, and personally, I won't really appreciate my code there either. I was exploring the organisation and trying to understand the codebase, and I may not have been consistent with my codebase. IMO, it would be better to lose it.

Was it an effort for a CGAL GSoC  project?

mehull-26:
yes

PR #8859, opened 2025-04-22
workaround API change in QT 6.9

sloriot:
Working around `qstring::arg()` API change in Qt 6.9

should fix:
https://cgal.geometryfactory.com/CGAL/testsuite/summary-6.1-I-135.html?package=Lab_Demo


sloriot:
Successfully tested in CGAL-6.1-Ic-137 (error is gone on the 4 platforms)

PR #8862, opened 2025-04-23
Triangulation: Add static filter for 5D and 6D orientation predicate

afabri:
## Summary of Changes

This PR adds a static filter for  the orientation test of 4D, 5D and 6D points.  It is ~20% faster than just interval arithmetic.

- [x] Add for 4D
- [x] ~~Add in sphere predicate~~

## Release Management


mglisse:
> So far it is slower than just interval arithmetic

I tried quickly running the benchmark (on a file generated with the new generator).
With master, I get 11s.
Branch with `-DCGAL_NO_STATIC_FILTER_5` (so presumably just the clear/mark change): 10.3s
Branch (with static filter): 8.5s

That doesn't look slower to me :confused: 

* The TDS change is exactly what I had in mind
* orientationC5.h doesn't seem to be used
* we could write the static filter directly with the internal interface instead of defining an adapter (I used the adapter approach for 2d and 3d because I wanted to reuse the static filters from Kernel_23 without modifying them), but if you prefer with the intermediate adapter, that's fine with me.

Triangulation/doc/Triangulation/CGAL/TDS_full_cell_default_storage_policy.h
```diff
@@ -0,0 +1,10 @@
+
+namespace CGAL {
+
+  /*!
+  \ingroup PkgTriangulationsRef
+  A tag class to indicate that mirror indices are not stored in full cells.
+  */
+class TDS_full_cell_mirror_storage_policy
```
mglisse, line 8: copy-pasto?

(we may want to introduce a new name/alias for this one, especially if we end up using "mirror" by default)

afabri:
@sloriot can you please create the 4D version, plus the in-sphere tests with the filter generator.

afabri:
Hi @mglisse I just added the 4D version.  The user we know only constructs a convex hull,  and all points are on the hull, so the triangulation class does the job, and I have no need for a faster insphere test. 
It might make sense to add a convex hull function that does not insert if the locate type is not outside convex hull, but not in this PR.

Triangulation/doc/Triangulation/CGAL/TDS_full_cell_mirror_storage_policy.h
```diff
@@ -0,0 +1,11 @@
+
+namespace CGAL {
+
+  /*!
+  \ingroup PkgTriangulationsRef
+  A tag class to indicate that mirror indices are stored in full cells.
+  \precondition The dimension is not larger than 127.
```
albert-github: You probably meant here `\pre` (warning in overnight documentation build:CGAL-6.1-Ic-188)

sloriot:

afabri:
In 6D with 100K points the runtime is 207 vs 94 seconds.

sloriot:

PR #8865, opened 2025-04-27
Removing page outline panel

albert-github:
In the current master version of doxygen  (and thus upcoming version 1.14.0) a page outline panel is introduced, setting: `PAGE_OUTLINE_PANEL`, though it would be better to disable this in CGAL.

**With the setting `PAGE_OUTLINE_PANEL=YES` (the default)**:

image

**With the setting `PAGE_OUTLINE_PANEL=NO`**:

image

### DIFF
--- a/Documentation/doc/resources/1.10.0/BaseDoxyfile.in
+++ b/Documentation/doc/resources/1.10.0/BaseDoxyfile.in
@@ -523,6 +523,16 @@ DISABLE_INDEX          = YES
 
 GENERATE_TREEVIEW      = YES
 
+# When GENERATE_TREEVIEW is set to YES, the PAGE_OUTLINE_PANEL option determines
+# if an additional navigation panel is shown at the right hand side of the
+# screen, displaying an outline of the contents of the main page, similar to
+# e.g. https://developer.android.com/reference If GENERATE_TREEVIEW is set to
+# NO, this option has no effect.
+# The default value is: YES.
+# This tag requires that the tag GENERATE_HTML is set to YES.
+
+PAGE_OUTLINE_PANEL     = NO
+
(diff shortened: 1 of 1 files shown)


afabri:
I guess it cannot replace our navigation panel ?

albert-github:
@afabri Which "our navigation panel " do you mean ?

afabri:
> @afabri Which "our navigation panel " do you mean ?

The panel on the left.

afabri:
@sloriot do you think we have to test this or can you simply merge ?

albert-github:
No this has not been introduced in 1.10.0 but will be available in 1.14.0 release (and thus in the current master).
The directory is called 1.10.0 as it was originally based on 1.10.0 master.

sloriot:
So I need to create a new directory otherwise we'll get warnings with version between 1.10 and 1.14.

albert-github:
I don't think so.
The directory that is now called 1.10.0 is only used for the tests with the master version of doxygen.
The other directory and version currently used for generation the documentation is 1.9.6.
When I remember correctly the 1.8.13 directory and version is only used to compares some XML output from doxygen.

I think it would be better to wait for the release of the doxygen 1.14.0 version (which is expected in the next 2 weeks) and use this version for future new versions (6.2 and up) of CGAL. At that moment it would be best to rename the 1.10.0 directory to 1.14.0 and create a new directory which should in my opinion be called master.
Also the 1.8.13 directory should be dropped and when necessary replaced by, preferably, 1.14.0 as well for all versions (as it is only used internally by CGAL.

(I think there might be already an issue regarding discarding 1.8.13 completely and using a newer version.)

sloriot:

PR #8875, opened 2025-05-06
Kernel: Concepts are CamelCase

afabri:
## Summary of Changes

When `CGAL::Weighted_point_3` was moved into `Kernel::` we forgot to switch to CamelCase, for  example [here](https://cgal.geometryfactory.com/CGAL/doc/master/Kernel_23/classKernel_1_1ConstructPoint__3.html).
The same for `Iso_rectangle_2`  and `Iso_cuboid_3`

## Release Management

### DIFF
--- a/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
+++ b/Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
@@ -417,7 +417,7 @@ public:
     \ref CGAL::ON_BOUNDARY, depending on where point `p` is relative to
     rectangle `r`.
   */
-  Bounded_side operator()(const Kernel::IsoRectangle_2& r,
+  Bounded_side operator()(const Kernel::Iso_rectangle_2& r,
                           const Kernel::Point_2&p);
 
   /// @}
@@ -466,7 +466,7 @@ public:
     \ref CGAL::ON_BOUNDARY, depending on where point `p` is with
   respect to iso-cuboid `c`.
   */
-  Bounded_side operator()(const Kernel::IsoCuboid_3& c,
+  Bounded_side operator()(const Kernel::Iso_cuboid_3& c,
                           const Kernel::Point_3&p);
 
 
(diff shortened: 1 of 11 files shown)


MaelRL:
There are plenty of similar cases in other doc files, such as doc/Point_2.h, doc/squared_distance_2.h, ...

MaelRL:
Indeed, sorry.

There are only a few hits, looking more precisely, like:

```cpp
  /*!
  A type representing an
  axis-aligned rectangle. It must be a model of
  `Kernel::Iso_rectangle_2`.
  */
  typedef unspecified_type Iso_rectangle_2;
```

in `Periodic_2TriangulationTraits_2`

Kernel_23/doc/Kernel_23/Concepts/FunctionObjectConcepts.h
```diff
@@ -417,7 +417,7 @@ class BoundedSide_2 {
     \ref CGAL::ON_BOUNDARY, depending on where point `p` is relative to
     rectangle `r`.
   */
-  Bounded_side operator()(const Kernel::Iso_rectangle_2& r,
```
sloriot, line 420: Does not seem valid to me. Here we are not talking about the concept, but of the nested type `Iso_rectangle_2` from the `Kernel`

afabri: You are right. But that mean that the links  like `Kernel::Point_2` with a link to the concept are wrong. Would be quite some work to remove them with a `%`.

sloriot: Best solution would be to rename `Kernel` namespace to `Kernel_concepts`

MaelRL: What about `KernelObjects` / `KernelFunctors`?

afabri: Why is it nost just   `operator()( Point_3)`, because we  are in the scope of `Kernel` ?

sloriot: That's a good point but you are not really in a Kernel scope. You define the requirement for a functor that could live outside of the Kernel (says Projection_traits for example). Kernel does not make sense and there should be nested types in the functor.

MaelRL: `Kernel::Point_3` is the `Point_3` typedef defined in the `Kernel` concept, so it's well defined (?)

sloriot:

PR #8854, opened 2025-04-21
C++17 implementation of Dispatch_(or_drop)?_output_iterator

lrineau:
## Summary of Changes

This is a new implementation of `Dispatch_output_iterator` and `Dispatch_or_drop_output_iterator`, using:

  - derivation from template parameter pack (C++11),
  - with a using-declaration based on a pack (C++17).

And `*_or_drop*` is implemented using an enable-if trick with a boolean template parameter.

That new implementation loops around the issue with gcc-15, and fixes issue #8827.

See:
  - the CGAL https://github.com/CGAL/cgal/issues/8827,
  - and the upstream gcc bug entry https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119859

## Release Management

* Affected package(s): STL_Extension
* Issue(s) solved (if any): fix #8827
* License and copyright ownership: maintenance by GeometryFactory

PR #8879, opened 2025-05-07
Avoid ambiguous overloads of operator= with optional and variant

### LINKED ISSUE #8827: Error in gcc master detected in Boolean_set_operations_2_examples
Errors [here](https://cgal.geometryfactory.com/CGAL/testsuite/summary-6.1-Ic-124.html?package=Boolean_set_operations_2_Examples) are due to what looks like a bug in gcc master.

Bug reported to gcc: submitted a gcc bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119859

The following does not compile (ambiguous)

```c++
struct Curve{};

struct A
{
  template<typename T>
  A& operator=(const T& t) {
    return *this;
  }
};

struct B : public A
{
  using A::operator=;
  template<typename T>
  B& operator=(const T& t) {
    return *this;
  }
};

int main()
{
  Curve c;

  B b;
  b=c;
}
```

Observed with:

```
commit 03ac8886e5c1fa16da90276fd721a57fa9435f4f (HEAD -> master, origin/master, origin/HEAD)
Author: Jonathan Wakely [jwakely@redhat.com](mailto:jwakely@redhat.com)
Date: Wed Mar 26 11:47:05 2025 +0000
```

sloriot:
Does not replace #8854 that is a better implementation but currently not fully compatible with older compilers.

Also loops around the issue with gcc-15, and fixes issue #8827.

See:
  - the CGAL https://github.com/CGAL/cgal/issues/8827,
  - and the upstream gcc bug entry https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119859

### DIFF
--- a/STL_Extension/include/CGAL/iterator.h
+++ b/STL_Extension/include/CGAL/iterator.h
@@ -1286,10 +1286,10 @@ struct Output_visitor {
 
 namespace internal {
 
-template < typename D, typename V = std::tuple<>, typename O = std::tuple<> >
+template < typename D, bool with_drop, typename V = std::tuple<>, typename O = std::tuple<> >
 struct Derivator
 {
-  typedef Derivator<D, V, O> Self;
+  typedef Derivator<D, with_drop, V, O> Self;
   Derivator() = default;
   Derivator(const Self&) = default;
   Self& operator=(const Self&) = delete;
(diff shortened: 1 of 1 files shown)


sloriot:
Successfully tested in CGAL-6.1-Ic-151 (errors are gone)

PR #8883, opened 2025-05-12
Developer Manual: trivial fixes

afabri:
## Summary of Changes

Fix that functions shall be written with  `()`.     
Removed an example with the no longer existing `Window_stream`

This part of the manual needs a real review, as a lot is now directly on github.


sloriot:

PR #8888, opened 2025-05-14
Fix range issue in weights

MaelRL:
## Summary of Changes

Outward construction does an inward construction for each hole, but the hole weights start at [1] in the weight range, not [0].

## Release Management

* Affected package(s): `Straight_skeleton_extrusion_2`
* Issue(s) solved (if any): -


sloriot:

PR #8894, opened 2025-05-15
Fix how to set BOOST_ROOT

afabri:
## Summary of Changes

We require at least 1.74, but we gave an example with 1.70
We did not give the path to the boost cmake config file.   

## Release Management

* Affected package(s): documentation

### DIFF
--- a/Documentation/doc/Documentation/Third_party.txt
+++ b/Documentation/doc/Documentation/Third_party.txt
@@ -70,7 +70,7 @@ from <A HREF="https://sourceforge.net/projects/boost/files/boost-binaries/">`htt
 
 As there is no canonical directory for where to find \boost on Windows,
 we recommend that you define the environment variable
-`BOOST_ROOT` and set it to where you have installed \boost, e.g., `C:/local/boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80.0`.
+`Boost_ROOT` (or `BOOST_ROOT`) and set it to where you have installed \boost, e.g., `C:/local`, or set `Boost_DIR` to the path of the directory containing the file `BoostConfig.cmake`, e.g.`C:/local/boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80.0` (be careful of the case of `Boost_DIR`).
 
 \subsection thirdpartyMP Multi Precision Number Type Library
 
--- a/Documentation/doc/Documentation/Third_party.txt
+++ b/Documentation/doc/Documentation/Third_party.txt
@@ -70,7 +70,7 @@ from <A HREF="https://sourceforge.net/projects/boost/files/boost-binaries/">`htt
 
 As there is no canonical directory for where to find \boost on Windows,
 we recommend that you define the environment variable
-`BOOST_ROOT` and set it to where you have installed \boost, e.g., `C:\boost\boost_1_70_0`.
+`BOOST_ROOT` and set it to where you have installed \boost, e.g., `C:/local/boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80.0`.
 
 \subsection thirdpartyMP Multi Precision Number Type Library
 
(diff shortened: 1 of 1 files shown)


afabri:
Is it ok that I changed "\"  to "/" ?

sloriot:
@lrineau I'm not sure about the fix. `Boost_DIR` shall point to where `BoostConfig.cmake` is located but `BOOST_ROOT` should point to the root of the installation, isn't it?

janetournois:
[The doc of FindBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html) says "BOOSTROOT : preferred installation prefix"
On my installation, it is set to where `BoostConfig.cmake` is, and it works

afabri:
> [The doc of FindBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html) says "BOOSTROOT : preferred installation prefix" On my installation, it is set to where `BoostConfig.cmake` is, and it works

But also with the underscore on the page you gave.

sloriot:
That's my point: BOOT_ROOT installation prefix while config mode should be using `Boost_DIR`.

sloriot:
ping @lrineau

Documentation/doc/Documentation/Third_party.txt
```diff
@@ -70,7 +70,7 @@ from <A HREF="https://sourceforge.net/projects/boost/files/boost-binaries/">`htt
 
 As there is no canonical directory for where to find \boost on Windows,
 we recommend that you define the environment variable
-`BOOST_ROOT` and set it to where you have installed \boost, e.g., `C:\boost\boost_1_70_0`.
+`BOOST_ROOT` and set it to where you have installed \boost, e.g., `C:/local/boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80.0`.
```
lrineau:
```suggestion
`Boost_ROOT` (or `BOOST_ROOT`) and set it to where you have installed \boost, e.g., `C:/local`, or set `Boost_DIR` to the path of the directory containing the file `BoostConfig.cmake`, e.g.`C:/local/boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80.0` (be careful of the case of `Boost_DIR`).
```

afabri: I do not get how, if you only set `BOOST_ROOT` the cmake directory will ever be found automatically.

lrineau:
Actually, if Boost is installed in a standard directory, like `C:/ProgramFiles(x86)`, it can be found automatically. See the [documentation of `CMAKE_SYSTEM_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PREFIX_PATH.html).

Documentation/doc/Documentation/Third_party.txt
sloriot: We can also stick `Boost_DIR` and that's it.

afabri: But your "or"  in "set .. or set .."  is not correct, that is when I set `BOOST_ROOT` to `C:/local` the job is not done at all.

lrineau: > I do not get how, if you only set `BOOST_ROOT` the cmake directory will ever be found automatically.

Given a prefix, [`find_package` searches in several sub-directories](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure), including:
> `<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/`

given that:
> In all cases the `<name>` is treated as case-insensitive and corresponds to any of the names specified (`<PackageName>` or names given by `NAMES`).

Then the sub-path `boost_1_80_0/lib64-msvc-14.3/cmake/Boost-1.80` matches `<name>*/lib*/cmake/<name>*`.

lrineau: > But your "or" in "set .. or set .." is not correct, that is when I set `BOOST_ROOT` to `C:/local` the job is not done at all.

That is strange: I tested that successfully yesterday on `christo`.

sloriot:

PR #8898, opened 2025-05-21
Restore erase counter after destroy/creation (6.0.x)

sloriot:
6.0.x version of #8820


sloriot:
@lrineau please review that I did not miss something (will only be tested in the bug-fix release branch)

sloriot:
Merging so that we can test it tonight

PR #8899, opened 2025-05-21
Triangulation: filter_iterator

afabri:
## Summary of Changes

The testsuite of [Triangulation](https://cgal.geometryfactory.com/CGAL/testsuite/summary-6.1-Ic-159.html?package=Triangulation) has N's for some platforms. The test platforms have in common to use boost_1_88_0, but with VC++ or clang on Windows I can't reproduce the error with this version of boost.

In this PR I use the filter_iterator adaptor of CGAL, to see if the boost filter_iterator  is the problem.

## Release Management

### DIFF
--- a/STL_Extension/test/STL_Extension/test_Compact_container.cpp
+++ b/STL_Extension/test/STL_Extension/test_Compact_container.cpp
@@ -5,6 +5,9 @@
 #include <list>
 #include <vector>
 #include <type_traits>
+
+#include <boost/iterator/filter_iterator.hpp>
+
 #include <CGAL/Compact_container.h>
 #include <CGAL/Random.h>
 
@@ -57,9 +60,10 @@ class Node_2
     Node_2 * p;
     void * p_cc;
   };
-  int      rnd;
 
 public:
+
+  int      rnd;
 
   Node_2()
   : p(nullptr), rnd(CGAL::get_default_random().get_int(0, 100)) {}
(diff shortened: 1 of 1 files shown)


Triangulation/include/CGAL/Triangulation.h
```diff
@@ -437,15 +448,28 @@ class Triangulation
...
     { return Finite_full_cell_iterator(Finiteness_predicate(*this), full_cells_begin(), full_cells_end()); }
     Finite_full_cell_iterator finite_full_cells_end()
     { return Finite_full_cell_iterator(Finiteness_predicate(*this), full_cells_end(), full_cells_end()); }
+
     Finite_full_cell_const_iterator finite_full_cells_begin() const
     { return Finite_full_cell_const_iterator(Finiteness_predicate(*this), full_cells_begin(), full_cells_end()); }
     Finite_full_cell_const_iterator finite_full_cells_end() const
     { return Finite_full_cell_const_iterator(Finiteness_predicate(*this), full_cells_end(), full_cells_end()); }
+#else
+/*
```
afabri, line 462: @sloriot When I uncomment the non-const version I get a compilation error. Can you explain why?

sloriot: Fixed. I don't get why I need to define extra operators. I would have expect that there would be an implicit conversion then.

mglisse:
See https://github.com/boostorg/iterator/issues/92 for the source of the issue.

afabri:
@mglisse    I only changed the filter for cells and not for vertices, which is enough for the testsuite. What is different for vertices?  Is it maybe only not tested?

sloriot:
Successfully tested in CGAL-6.1-Ic-170

PR #8904, opened 2025-05-22
Rename home page of reference manuals to "Reference Manual"

sloriot:
Before: "Pkg Reference"
After: "Reference Manual"

No need to repeat the package name as it is on the line above.

Example output
Screenshot from 2025-05-22 16-18-54


sloriot:
Successfully tested in CGAL-6.1-Ic-161

PR #8909, opened 2025-05-26
TDS_3: use indices, and keep backward compatibility with handles

afabri:
## Summary of Changes

Instead of changing the API of TDS, Triangulation, and everything using it, this PR takes another approach:
We store indices in the data structures instead of handles, and generate the handles on the fly, which hold a pointer to the TDS and an index. 

We also split `Triangulation_data_structure_3`  in two:  A base class which knows about storage and index/vs pointer and provides an API with handles.  And a derived class that uses this API with handles, which works for both base classes as in the derived class we do not use pointers or indices directly. 

We checked that the Triangulation with indexed storage can be used in Alpha Wrapping without as only change of code the `typedef`  of the `TDS`. 

### Todo

See tracking issue #8969.

## Release Management

* Affected package(s): TDS_3


lrineau:
With last push, on Linux, the testsuite of TDS_3, Triangulation_3, and STL_Extension, pass in Debug (with ASAN and UBSAN):

```
STL_Extension_Examples      =   2.27 sec*proc (14 tests)
STL_Extension_Tests         =  19.46 sec*proc (72 tests)
TDS_3_Examples              =   1.29 sec*proc (8 tests)
TDS_3_Tests                 =   1.11 sec*proc (6 tests)
Triangulation_3_Demo        =   0.18 sec*proc (1 test)
Triangulation_3_Examples    =  29.26 sec*proc (42 tests)
Triangulation_3_Tests       = 608.20 sec*proc (46 tests)
```

TDS_3/include/CGAL/TDS_3/Indexed_storage.h
```diff
@@ -0,0 +1,1802 @@
...
+    free_list_type freelist_{Index_type::invalid_index};
+    size_type anonymous_property_nb = 0;
+    static constexpr bool recycle_ = true;
+    bool garbage_ = false;
+    Property_map<Index_type, Storage_type, Concurrency_tag> storage_ =
+        add_property_map<Storage_type>(prefix + std::string(":storage"), Storage_type()).first;
+    Property_map<Index_type, bool, Concurrency_tag> removed_ =
+        add_property_map<bool>(prefix + std::string(":removed"), false).first;
+
+    Property_map<Index_type, Point, Concurrency_tag> points_;
```
afabri: Doing it here we also get a property map for cells as soon it has a nested type `Point`. This is already the case for cells with cached circumcenter?

Triangulation_3/include/CGAL/Triangulation_3.h
```diff
@@ -1355,46 +1448,88 @@ class Triangulation_3
...
 
-    cell_stack.push(d);
+    cell_stack.push(tds().to_cell_descriptor(d));
     d->tds_data().mark_in_conflict();
 
-    *it.second++ = d;
+    *it.second++ = d; // = tds().to_cell_descriptor(d);
 
+    auto check_this_facet_must_be_in_the_cz = [&](cell_descriptor cd, int i, bool on_boundary = false) {
+      if(!this_facet_must_be_in_the_cz || !the_facet_is_in_its_cz) return;
```
afabri: In vtune I see both lambdas.  Before we performed the test of line above to decide if we call the lambda or not. I guess what we see in vtune is the parameters getting copied.

Triangulation_3/include/CGAL/Delaunay_triangulation_3.h
```diff
@@ -879,6 +894,13 @@ side_of_sphere(VertexDescriptor v0, VertexDescriptor v1,
       return t->side_of_sphere(c, p, true) == ON_BOUNDED_SIDE;
     }
 
+    template <typename CellDescriptor,
+              std::enable_if_t<false == std::is_same_v<Cell_handle, CellDescriptor>, bool> = true>
+    bool operator()(const CellDescriptor& c) const
+    {
+      return t->side_of_sphere(c, p, true) == ON_BOUNDED_SIDE;
+    }
+
```
lrineau, line 910: In both cases the code is the same, and forwards to `*t`. You can simplify to just one generic overload:

```c++
    template <typename CellDescriptor>
    bool operator()(const CellDescriptor& c) const
    {
      return t->side_of_sphere(c, p, true) == ON_BOUNDED_SIDE;
    }
```

Triangulation_3/include/CGAL/Delaunay_triangulation_3.h
```diff
@@ -904,6 +926,14 @@ side_of_sphere(VertexDescriptor v0, VertexDescriptor v1,
       return t->side_of_circle(c, 3, p, true) == ON_BOUNDED_SIDE;
     }
 
+
+    template <typename CellDescriptor,
+              std::enable_if_t<false == std::is_same_v<Cell_handle, CellDescriptor>, bool> = true>
+    bool operator()(const CellDescriptor& c) const
+    {
+      return t->side_of_circle(c, 3, p, true) == ON_BOUNDED_SIDE;
+    }
```
lrineau, line 942: Same here: it can be factorize in just one generic version.

afabri: Not the same code because we call the function `handle(c)`  in line 828.   But you are nevertheless right as we can `handle(c)` even if `c`  is a `Cell_handle`as that is the identity. Thank you.

afabri:
`DT3_benchmark.cpp`   is 10% slower in this branch than in [afabri/Triangulation_3-benchmark-GF](https://github.com/afabri/cgal/tree/Triangulation_3-benchmark-GF)  which is main with `DT3_benchmark.cpp` added.
The branch [afabri/Triangulation_3_Indices-GF]((https://github.com/afabri/cgal/tree/Triangulation_3-Indices-GF) goes further and completely removes handles, without being faster.

The benchmark  of main:
image


The benchmark of [afabri:Triangulation_3-Handles-GF](https://github.com/afabri/cgal/tree/Triangulation_3-Handles-GF)
image

The benchmark of [afabri:Triangulation_3-Indices-GF](https://github.com/afabri/cgal/tree/Triangulation_3-Indices-GF)
image

PR #8918, opened 2025-06-02
Thirdparty: bugfix for libpointmatcher

soesau:
## Summary of Changes

Added support for latest version of libpointmatcher.
Updated doc as libpointmatcher uses the deprecated boost/timer.hpp which can cause a compilation error when building the library.

There is a PR at libpointmatcher to fix the deprecation bug since months, but it still needs to be reviewed.

## Release Management


sloriot:

PR #8919, opened 2025-06-02
Fix warning: avoid copy

sloriot:


sloriot:

PR #8923, opened 2025-06-04
Remove double clear and take the link of vh and not the infinite vertex

sloriot:


Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -40,11 +40,10 @@ link_to_face_graph(const Triangulation_3& t,
   clear(fg);
   vertex_descriptor inf;
   vertex_descriptor nullvertex = boost::graph_traits<FG>::null_vertex();
-  fg.clear();
   typedef boost::unordered_map<Vertex_handle, vertex_descriptor> Vertex_map;
   Vertex_map vertex_map;
   std::vector<Cell_handle>  cells;
-  t.incident_cells(t.infinite_vertex(),std::back_inserter(cells));
+  t.incident_cells(vh, std::back_inserter(cells));
```
afabri, line 46: Note that we also have a function `visit_just_incident_cells()`  in case this makes a difference in performance.

afabri:
Is there an issue?  Taking the infinite vertex instead of `v` looks like a serious bug.

sloriot:
> Is there an issue? Taking the infinite vertex instead of `v` looks like a serious bug.

It is a serious bug :)

sloriot:
Successfully tested in CGAL-6.1-Ic-170

PR #8924, opened 2025-06-04
Improve link_to_face_graph API

sloriot:
also use it in Tetra remeshing to illustrate how to detect the sharp features of the domain


sloriot:
@janetournois The output (even before my change) does not look correct. Would you mind having a look and let me know if it is as expected. My issue is that `target_edge_length` does not seem to be taken into account on the domain boundary.

sloriot:

janetournois:
Indeed there is a problem with the surface of the cube, but it is also in master.
So it's not blocking this PR, and I'm taking care of it in another branch.
Thanks @sloriot for the report

janetournois:
> Indeed there is a problem with the surface of the cube, but it is also in master. So it's not blocking this PR, and I'm taking care of it in another branch. Thanks @sloriot for the report

There is actually no problem here.
The issue with the example is that the density of the input is very different between the surface and the inside of the cube. One iteration is not enough to uniformize edge lengths.
Also note that the target edge length (0.02) is very small compared to input edge lengths (even inside) so you need quite a few iterations to reach it.

I will push an improvement to this PR

Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -22,33 +22,87 @@
...
+
+\param t the 3D triangulation
+\param vh the vertex handle of the vertex
+\param tm the triangle mesh
+
+\param np an optional sequence of \ref bgl_namedparameters "Named Parameters" among the ones listed below
+
+\cgalNamedParamsBegin
+  \cgalParamNBegin{ignore_infinite_faces}
+    \cgalParamDescription{if `true` and if `vh` is on the convex hull of `t`, infinite facets are ignored and a triangle mesh with a border is generated}
```
afabri, line 47: Should the word _facets_ be replaced with _faces_?

sloriot:

Triangulation_3/include/CGAL/link_to_face_graph.h
sloriot: we are in the Triangulation_3 context so I'd say no.

Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -22,33 +22,87 @@
...
+#include <CGAL/boost/graph/named_params_helper.h>
 
 namespace CGAL {
 
+/*!
+\ingroup PkgTriangulation3Ref
 
-template<class Triangulation_3,class FG>
-typename boost::graph_traits<FG>::vertex_descriptor
+fills the face graph `tm` with the <A HREF="https://en.wikipedia.org/wiki/Simplicial_complex#Closure.2C_star.2C_and_link">link</a> of triangulation vertex `vh`.
```
MaelRL:
```suggestion
clears and fills the face graph `tm` with the <a href="https://en.wikipedia.org/wiki/Simplicial_complex#Closure.2C_star.2C_and_link">link</a> of triangulation vertex `vh`.
```

Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -22,33 +22,87 @@
...
+  \cgalParamNBegin{vertex_point_map}
+    \cgalParamDescription{a property map associating points to the vertices of `tm`}
+    \cgalParamType{a class model of `WritablePropertyMap` with `boost::graph_traits<TriangleMesh>::%vertex_descriptor`
+                   as key type and `Triangulation_3::Point_3` as value type}
+    \cgalParamDefault{`boost::get(CGAL::vertex_point, tm)`}
+  \cgalParamNEnd
+\cgalNamedParamsEnd
+
+\returns the vertex descriptor of the triangle mesh `tm` corresponding to the infinite vertex of `t`,
+         if `vh` is on the convex hull of the triangulation, and if `no_infinite_faces == false`.
```
MaelRL:
```suggestion
         if `vh` is on the convex hull of the triangulation, and if `ignore_infinite_faces` is `false`.
```

Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -22,33 +22,87 @@
...
+\param np an optional sequence of \ref bgl_namedparameters "Named Parameters" among the ones listed below
+
+\cgalNamedParamsBegin
+  \cgalParamNBegin{ignore_infinite_faces}
+    \cgalParamDescription{if `true` and if `vh` is on the convex hull of `t`, infinite facets are ignored and a triangle mesh with a border is generated}
+    \cgalParamType{Boolean}
+    \cgalParamDefault{`true`}
+  \cgalParamNEnd
+  \cgalParamNBegin{vertex_to_vertex_output_iterator}
+    \cgalParamDescription{an `OutputIterator` where pairs of input and output vertices are put upon creation in `tm`.}
```
MaelRL:
```suggestion
    \cgalParamDescription{an `OutputIterator` where pairs of input vertices and output descriptors are put upon creation in `tm`.}
```

sloriot:
For some reason, it seems to be responsible for time out in Tetremesh testsuite.

afabri:
@sloriot can you please improve the very first paragraph in this PR.  It cannot start with "also use in ..."
The title speaks about the improvement of an API. How can that lead to a timeout?

afabri:
> For some reason, it seems to be responsible for time out in Tetremesh testsuite.

@sloriot can you add it again in a testsuite, to see if the timeout is still there,  In Ic-208 were tons of integrated pull requests.

sloriot:
Successfully tested in CGAL-6.2-Ic-135

lrineau:
TODO: this PR is a breaking change, no?

Triangulation_3/include/CGAL/link_to_face_graph.h
```diff
@@ -61,22 +117,38 @@ link_to_face_graph(const Triangulation_3& t,
...
+typename boost::graph_traits<TriangleMesh>::vertex_descriptor
+link_to_face_graph(const Triangulation_3& t,
+                   typename Triangulation_3::Vertex_handle vh,
+                   TriangleMesh& tm,
+                   bool ignore_infinite_faces)
+{
+  link_to_face_graph(t, vh, tm, CGAL::parameters::ignore_infinite_faces(ignore_infinite_faces));
+}
+#endif
+#endif
```
lrineau, line 150: I see this part of the code ensures the backward compatibility. But:

  - it should have been marked as deprecated,
  - and it should be tested.

PR #8940, opened 2025-06-17
Lab - fix "save as" for a `c3t3` that was loaded

janetournois:
## Summary of Changes

Load a `.mesh` file, remesh it or do nothing, and "save as..." fails because the `c3t3` has its `is_valid` field set to `false`.

This PR fixes that issue : now only `.binary.cgal` requires being certainly valid, i.e. generated by Mesh_3

thanks @IasonManolas for the report

## Release Management

* Affected package(s): Lab


sloriot:

PR #8944, opened 2025-06-18
Reset depth when updating exact

sloriot:
Only meaningful when CGAL_PROFILE is defined.

### DIFF
--- a/Filtered_kernel/include/CGAL/Lazy.h
+++ b/Filtered_kernel/include/CGAL/Lazy.h
@@ -233,14 +233,14 @@ print_dag(const Return_base_tag&, std::ostream& os, int level)
 
 struct Depth_base {
 #ifdef CGAL_PROFILE
-  int depth_;
+  mutable int depth_;
 
   Depth_base()
     : depth_(0)
   {}
 
   int depth() const { return depth_; }
-  void set_depth(int i)
+  void set_depth(int i) const
   {
     depth_ = i;
     CGAL_HISTOGRAM_PROFILER(std::string("[Lazy_kernel DAG depths]"), i);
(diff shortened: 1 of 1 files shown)


Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -660,6 +660,7 @@ class Lazy_rep_n final :
     auto* p = new typename Base::Indirect(ec()( CGAL::exact( std::get<I>(l) ) ... ) );
     this->set_at(p);
     this->set_ptr(p);
+    this->set_depth(0);
```
mglisse, line 663: Is that also needed in Lazy_rep_optional_n?

Filtered_kernel/include/CGAL/Lazy.h
```diff
@@ -660,6 +660,7 @@ class Lazy_rep_n final :
...
     if(!noprune || is_currently_single_threaded())
       lazy_reset_member(l);
```
mglisse, line 665: Is the goal of the depth to measure how much work a future `exact` call will take, or to measure the memory used to store the DAG? In the first case, I think you put set_depth in the right place. In the second case, it might make sense to move it next to `lazy_reset_member`.

sloriot: It's to get the current depth of an object in the DAG so thanks for confirming that's the right place.

Filtered_kernel/include/CGAL/Lazy.h
sloriot: I also asked myself the question, also about adding the conditional reset. As it is specific code maybe @afabri knows more.

sloriot:

PR #8945, opened 2025-06-19
Tetrahedral_remeshing - make it determistic

janetournois:
## Summary of Changes

Tetrahedral_remeshing is not deterministic (though outputs are very close in 2 consecutive runs).
Making it deterministic...

## Release Management


Tetrahedral_remeshing/test/Tetrahedral_remeshing/test_tetrahedral_remeshing_determinism.cpp
```diff
@@ -0,0 +1,104 @@
+#define CGAL_TETRAHEDRAL_REMESHING_VERBOSE
```
sloriot: already present in the cmake script. Shouldn't it be removed here and the cmake script?

janetournois: I prefer keeping it because it provides some info in the testsuite results before actually starting to debug

sloriot:

PR #8959, opened 2025-06-27
Orthtree: Node insertion during orthtree build has linear complexity

soesau:
## Summary of Changes

Added inactive list for deleted nodes (nodes cannot be deleted for now)
avoiding linear search time for node insertion during Orthtree build

Reusing indices in properties from deleted nodes for insertion of a group of nodes (i.e., during Orthtree refinement) is deactivated. This has no impact as nodes cannot be deleted anyway.

## Release Management

* Affected package(s): Orthtree, Property_map

### DIFF
--- a/Property_map/include/CGAL/Property_container.h
+++ b/Property_map/include/CGAL/Property_container.h
@@ -50,7 +50,7 @@ public:
 
   virtual void append(const Property_array_base<Index>& other) = 0;
 
-  virtual void reserve(std::size_t n) = 0;
+  virtual void resize(std::size_t n) = 0;
 
   virtual void shrink_to_fit() = 0;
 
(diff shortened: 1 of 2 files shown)


sloriot:
Runtime error [here](https://cgal.geometryfactory.com/CGAL/testsuite/summary-6.1-Ic-188.html?package=Property_map)

sloriot:
Successfully tested in CGAL-6.1-Ic-196

Property_map/include/CGAL/Property_container.h
```diff
@@ -441,16 +445,15 @@ class Property_container {
   }*/
 
 public:
+  void resize(std::size_t n) {
```
sloriot, line 447: I don't really like this change. resize and reserve does not do the same thing. Here you don't set the properties as only reserve is called for each property. I also don't understand why `m_active_indices` is resized (even in the original code).

soesau: It is actually just bad naming.
The reserve method of the property arrays also resize the underlying container. I'll rename it to be consistent.

sloriot:

PR #8993, opened 2025-07-23
Shape_detection: Plane and line region angle check bugfix

soesau:
## Summary of Changes

Checking sign before using squared cos angle
The squared cos angle is used to avoid sqrt for vector normalization.

This changes the behavior to be more strict regarding unoriented normals. Maybe, a parameter to decide whether to use the absolute value of the cos(angle) is better?

## Release Management


sloriot:
@sven why work in progress?

sloriot:
Fully red in  CGAL-6.2-Ic-6

sloriot:
@soesau what is the status of this PR? Shall you commit the problematic test case?

PR #8994, opened 2025-07-23
Reference Mpzf/Gmpzf/MP_Float from each other's doc

### LINKED ISSUE #8991: Mention Mpzf in the docs of Gmpzf and MP_Float
The 3 types `MP_Float`, `Gmpzf` and `Mpzf` provide very similar functionality. It would be useful if the documentation of each one mentioned the other 2, possibly with a hint of when to use which. (I just had the case of someone who saw Gmpzf, found it slow, and reimplemented their own version without noticing Mpzf)
There is also `cpp_float`, but since it isn't documented...

mglisse:
## Release Management

* Issue(s) solved (if any): fix #8991

### DIFF
--- a/Number_types/doc/Number_types/CGAL/MP_Float.h
+++ b/Number_types/doc/Number_types/CGAL/MP_Float.h
@@ -24,8 +24,8 @@ plan to also have a multiprecision exponent to fix this issue.
 
 The implementation of `MP_Float` is simple but provides a quadratic
 complexity for multiplications. This can be a problem for large operands.
-For faster implementations of the same functionality with large integral
-values, you may want to consider using `GMP` or `LEDA` instead.
+For faster implementations of the same functionality, if `GMP` is available,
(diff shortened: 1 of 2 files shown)


Number_types/doc/Number_types/CGAL/MP_Float.h
```diff
@@ -24,8 +24,8 @@ plan to also have a multiprecision exponent to fix this issue.
 
 The implementation of `MP_Float` is simple but provides a quadratic
 complexity for multiplications. This can be a problem for large operands.
-For faster implementations of the same functionality with large integral
-values, you may want to consider using `GMP` or `LEDA` instead.
+For faster implementations of the same functionality, if `GMP` is available,
```
afabri:
```suggestion
For faster implementations of the same functionality, if \gmp is available,
```

sloriot:

PR #9000, opened 2025-07-28
Make sure boost system is found

sloriot:


sloriot:
boost system is now header only, changed done directly in the master branch. See ref https://github.com/boostorg/system/issues/132

Backward compatibility shouldn't be an issue as the lib was "empty" for last releases (testsuite will tell us).

sloriot:

PR #9001, opened 2025-07-28
Fix precondition and lazy fill of maps

sloriot:
Follow up of #8609


sloriot:

PR #9003, opened 2025-07-29
Fix unused var warning

sloriot:
See [here](https://cgal.geometryfactory.com/CGAL/testsuite/CGAL-6.1-Ic-206/Boolean_set_operations_2/TestReport_lrineau_Ubuntu-latest-GCC6-CXX1z.gz) for example


sloriot:

PR #9004, opened 2025-07-29
Homological Discrete Vector Fields

umenohana13:
## Summary of Changes

The pull request contains the first version of:
* user manual (still incomplete - persistant homology and Alexander duality are not documented)
* reference manual (slightly incomplete as well - the IO classes stores in tools_io are "lightly" documented as they will be revised to integrate IOs inside CGAL)
* code (same: the tools_io file will be revised)

## Release Management

* Affected package(s): `HDVF` / `Surface_mesh_topology`
* Issue(s) solved (if any): 
* Feature/Small Feature (if any): https://cgalwiki.geometryfactory.com/CGAL/Members/wiki/Features/Homological_Discrete_Vector_Fields
* License and copyright ownership: Not yet validated by my lab but most probably LGPL / LIS laboratory


afabri:
@umenohana13 I started fixing errors showing up in the CI, but there are too many to make. It is now headers compiled in isolation which do not `#include` all they use.  And missing `std::` before `cout` etc.

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -0,0 +1,132 @@
...
+/*!
+\brief Returns all boundary matrices.
+
+The function returns constant reference to a vector of column-major sparse matrices. The `q`-th element of this vector is the matrix of \f$\partial_q\f$, which gives the boundary of cells of dimension `q`(as a linear combination of `q`-1 cells).
+
+ */
+const vector<Col_matrix> & boundary_matrices() const;
+
+/*!
+ \brief Returns the boundary matrix of dimension `q` (ie.\ the matrix of \f$\partial_q}\f$).
```
afabri:
```suggestion
 \brief Returns the boundary matrix of dimension `q` (ie.\ the matrix of \f$\partial_q\f$).
```

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -0,0 +1,132 @@
...
+
+/// @}
+
+/// \name Access functions
+/// @{
+
+/*!
+Returns the dimension of the complex, that is, the largest dimension of cells.
+ */
+int dim();
```
afabri: Should be `dimension()`.  The same for `number_of_cells()`

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -0,0 +1,132 @@
...
+class AbstractChainComplex
+{
+public:
+/// \name Types
+/// @{
+
+/*!
+ Type of column-major chains (returned by the boundary operator)
+ */
+typedef CGAL::OSM::Sparse_chain<CoefficientType, CGAL::OSM::COLUMN> Col_chain;
```
afabri: Here I would also go for `Column_chain`

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -0,0 +1,132 @@
...
+*/
+Col_chain d(size_t id_cell, int q);
+
+
+/*!
+ \brief Returns the co-boundary of the cell of index `id_cell` in dimension `q`.
+
+This boundary is a finite linear combination of cells of dimension `q`+1. It is encoded as a row-major chain (which maps each cell with a non-zero coefficient to this coefficient). This co-boundary  is thus the `id_cell`-th row of the boundary matrix in dimension `q`+1.
+*/
+Row_chain cod(size_t id_cell, int q);
```
afabri, line 120: What does `cod` stand for?

HDVF/doc/HDVF/Concepts/Filtration.h
```diff
@@ -0,0 +1,165 @@
+/*!
+\ingroup PkgHDVFConcepts
+\cgalConcept
+
+The concept `Filtration` describes the requirements for persistent filtrations associated to persistent homology computation.
+
+A filtration is associated to an `AbstractSimplicialComplex`. Each cells is equiped with a scalar value (called its degree) and the filtration is an enumeration of cells in any dimension by increasing degrees.
```
afabri: `AbstractSimplicialComplex`   has no link.

HDVF/doc/HDVF/Concepts/Filtration.h
```diff
@@ -0,0 +1,165 @@
...
+
+A filtration class provides:
+
+ - an iterator to visit all cells by increasing degrees.
+ - getters to get the degree of a cell, and the cell of a given index along the filtration.
+ - an overload of `<<` to output filtrations.
```
afabri:
```suggestion
 - an overload of `operator<<()` to output filtrations.
```

HDVF/doc/HDVF/Concepts/Filtration.h
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * As stated in `AbstractChainComplex`, cells are identified by their dimension, together with their index along this dimension. The type `Cell_index_dimension` stores this pair:
+     *
+     * - First element of the pair: index of the cell.
+     * - Second element of the pair: dimension of the cell.
+     */
+    typedef std::pair<size_t, int> Cell_index_dimension ;
+
+    /*! \brief Value type of the filtration iterator.
+     * Contains a cell (identified by its index and dimenion in a `Cell_index_dimension`) and its associated degree.
```
afabri:
```suggestion
     * Contains a cell (identified by its index and dimension in a `Cell_index_dimension`) and its associated degree.
```

HDVF/doc/HDVF/Concepts/Filtration.h
```diff
@@ -0,0 +1,165 @@
...
+    /*! \brief Gets the cell (that is cell index and dimension) at the index `i` of the filtration.
+     */
+    Cell_index_dimension cell_index_dimension (size_t i) const;
+
+    /*! \brief Gets the degree of the `i`th element of the filtration.
+     */
+    Degree_type degree (size_t i) const;
+
+    // Filtration verification
+    /*! \brief Checks that a filtration is valid
```
afabri:
```suggestion
    /*! \brief Checks that a filtration is valid.
```

HDVF/doc/HDVF/Concepts/GeometricChainComplex.h
```diff
@@ -0,0 +1,45 @@
+/*!
+\ingroup PkgHDVFConcepts
+\cgalConcept
+
+The concept `GeometricChainComplex` refines the concept `AbstractChainComplex` and describes the requirements for (topological) chain complexes associated to geometric complexes used in the concept `HDVF`.
+ It adds to `AbstractChainComplex` methods to get vertices coordinates.
```
afabri:
```suggestion
 It adds to `AbstractChainComplex` methods to get vertex coordinates.
```

HDVF/doc/HDVF/Concepts/GeometricChainComplex.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+typedef std::vector<double> Point ;
+/// @}
+
+
+/// \name Access functions
+/// @{
+
+/*!
+Returns the vector of vertices coordinates.
```
afabri:
```suggestion
Returns the vector of vertex coordinates.
```

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+
+ <img src="HDVF_ex1_perfect_cohom1.png" align="center" width=30%/>
+ <img src="HDVF_ex1_perfect_cohom2.png" align="center" width=30%/>
+ <img src="HDVF_ex1_perfect_cohom3.png" align="center" width=30%/>
+
+
+\cgalHasModelsBegin
+\cgalHasModelsBare{`CGAL::HDVF::Hdvf_core<Ring, AbstractChainComplex, SparseChain, Sparse_matrix>`}
+\cgalHasModelsBare{`CGAL::HDVF:Hdvf<Ring, AbstractChainComplex>`}
+\cgalHasModelsBare{`CGAL::HDVF:Hdvf_persistence<Ring, AbstractChainComplex>`}
```
afabri: In the documentation of the class you do have `CoefficientType`  for the first template parameter.

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+ * The function returns in each dimension the vector of cells with a given `flag`.
+ */
+std::vector<std::vector<size_t> > flag (FlagType flag) const;
+
+/*!
+ * \brief Gets cells with a given `flag` in dimension `q`.
+ *
+ * The function returns the vector of cells of dimension `q` with a given `flag`.
+ */
+std::vector<size_t> flag_dim (FlagType flag, int q) const;
```
afabri: I would just call it `flag()` .

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+
+/*!
+ * \brief Gets the flag of the cell `tau` in dimension `q`.
+ */
+FlagType cell_flag (int q, size_t tau) const;
+
+/*!
+ * \brief Gets HDVF option.
+ */
+int hdvf_opts ();
```
afabri, line 228: `option()`  or `options()`

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+ * \brief Gets the column-major matrix of \f$\partial'\f$, reduced boundary operator (from the reduction associated to the HDVF).
+ */
+const Col_matrix& matrix_dd (int q) const;
+
+/*!
+ * \brief Tests if a HDVF is perfect.
+ *
+ * The function returns `true` if the reduced boundary matrix is null and `false` otherwise.
+ */
+    bool is_perfect_hdvf();
```
afabri, line 255: `is_perfect()`

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+
+
+/// @}
+
+
+/// \name HDVF functions and operations
+/// @{
+
+/*!
+ * \brief A operation: pairs two critical cells.
```
afabri: I think you should systematically write `A`-operation to avoid confusion with the word "A".

afabri: In fact at some places you write `A()`  operation which is also ok.

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+ *
+ * As long as valid pairs for A exist, the function selects the first available pair (returned by `find_pair_A()`) and applies the corresponding `A()` operation.
+ * If the `Ring` of coefficients is a field, this operation always produces a perfect HDVF (ie.\ the reduced boundary is null and the reduction provides homology and cohomology information).
+ *
+ * If the HDVF is initially not trivial (some cells have already been paired), the function completes it into a perfect HDVF.
+ *
+ * \param verbose If `true`, all intermediate reductions are printed out.
+ *
+ */
+std::vector<Pair_cells> compute_perfect_hdvf(bool verbose = false);
```
afabri: Maybe replace `Pair_cells` with `Cell_pair`

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+ * \brief Exports primary/secondary/critical labels (in particular for vtk export).
+ *
+ * The method exports the labels of every cell in each dimension.
+ *
+ * \return A vector containing, for each dimension, the vector of labels by cell index.
+ */
+virtual std::vector<std::vector<int> > psc_labels () const;
+
+/*!
+ * \brief Returns a chain containing the homology generator associated to `cell` (critical cell) of dimension  `q` (in particular for vtk export).
```
afabri:
```suggestion
 * \brief Returns a chain containing the homology generator associated to `cell` (critical cell) of dimension  `q`.
```

afabri (changes requested):

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
umenohana13: OK done

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
umenohana13: Done

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
umenohana13: It goes with d : usually we use d for the boundary operator, and d^* for the co-boundary, I wrote cod for co-d.
In our (finite dimension) context, it's just the operator which matrix is the transpose of d (but in a more general context, the definition is more complex.
Is it unclear?

HDVF/doc/HDVF/Concepts/Filtration.h
umenohana13: Oups, it's "AbstractChainComplex". 
DONE

HDVF/doc/HDVF/Concepts/Filtration.h
umenohana13: DONE

HDVF/doc/HDVF/Concepts/Filtration.h
umenohana13: OK

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,413 @@
...
+
+/*! \brief Structure to represent data for HDVF operations (pairs of cells).
+ *
+ * Cells are always sorted so that the dimension of `sigma` is lesser than the dimension of `tau`.
+ */
+struct Pair_cells {
+    size_t sigma;  /// Index of the first cell
+    size_t tau;    /// Index of the second cell
+    int dim;    /// Dimension of cells: `dim`/`dim`+1 for A and R, `dim`/`dim` for other operations
+};
```
afabri:
```suggestion
struct Pair_cells {
    /// Index of the first cell
    size_t sigma; 
    /// Index of the second cell
    size_t tau;   
    /// Dimension of cells: `dim`/`dim`+1 for `A()` and `R()`, `dim`/`dim` for other operations
    int dim;    
};
```

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1172 @@
...
+        del_row(_F_row[q + 1], tau2);
+    }
+
+    // ---- Update _G_col
+
+    if (_hdvf_opt & (OPT_FULL | OPT_G))
+    {
+        // Update _G_col[q + 1]
+        // Subtract the product of (G11 * D11_inv) and D12 from _G_col[q+1]
+        _G_col[q + 1] -= (G11 * D11_inv) * D12;
```
afabri, line 1446: Visual C++ finds this `operator*()` ambiguous:
```
example_hdvf_simplicial.cpp
C:\GitHub\cgal\HDVF\include\CGAL\HDVF\Hdvf_core.h(783,23): error C2593: 'operator *' is ambiguous [C:\GitHub\cgal\HDVF\examples\HDVF\build\example_hdvf_simplicial.vcxproj]
  (compiling source file '../example_hdvf_simplicial.cpp')
      C:\GitHub\cgal\HDVF\include\CGAL\OSM\Sparse_chain.h(285,25):
      could be 'CGAL::OSM::Sparse_chain<CoefficientType,1> CGAL::OSM::operator *<1>(const CGAL::OSM::Sparse_chain<CoefficientType,1> &,const CoefficientType &)'
          with
          [
              CoefficientType=CoefficientType
          ]
      C:\GitHub\cgal\HDVF\include\CGAL\OSM\Sparse_chain.h(285,25):
      or       'CGAL::OSM::Sparse_chain<CoefficientType,2> CGAL::OSM::operator *<1>(const CGAL::OSM::Sparse_chain<CoefficientType,1> &,const CoefficientType &)'
          with
          [
              CoefficientType=CoefficientType
          ]
      C:\GitHub\cgal\HDVF\include\CGAL\HDVF\Hdvf_core.h(783,23):
      while trying to match the argument list '(CGAL::OSM::Sparse_chain<CoefficientType,1>, CoefficientType)'
          with
          [
              CoefficientType=CoefficientType
          ]
      C:\GitHub\cgal\HDVF\include\CGAL\HDVF\Hdvf_core.h(783,23):
      the template instantiation context (the oldest one first) is
          C:\GitHub\cgal\HDVF\examples\HDVF\example_hdvf_simplicial.cpp(38,18):
```

afabri: @sloriot now that you master VC++, can you please have a look.

HDVF/include/CGAL/OSM/Sparse_chain.h
```diff
@@ -0,0 +1,843 @@
...
+    /**
+     * \brief Applies multiplication on each coefficient.
+     *
+     * \param[in] chain The  chain.
+     * \param[in] lambda The factor to apply.
+     *
+     * \return A new chain representing the result.
+     */
+    template <int _CTF>
+    friend Sparse_chain operator*(const Sparse_chain<CoefficientType, _CTF> &chain, const CoefficientType& lambda) {
```
afabri: The return type has as second template parameter `ChainTypeFlag`.   I find it surprising that chain can have a `__CTF` different from  `ChainTypeFlag`.

HDVF/include/CGAL/HDVF/Hdvf_core.h
afabri: got fixed.

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,839 @@
...
+            p.sigma = i ;
+            p.tau = sigma ;
+            p.dim = q-1 ;
+        }
+    }
+    return p;
+}
+
+template<typename CoefficientType, typename ComplexType, typename DegreeType, typename FiltrationType>
+Cell_pair Hdvf_persistence<CoefficientType, ComplexType, DegreeType, FiltrationType>::step_persist(bool& found, bool verbose)
```
afabri: The function must return something.

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,839 @@
...
+        _masks.at(q) = OSM::Bitboard(this->_K.number_of_cells(q)) ;
+    }
+
+    // Init boundary matrices
+    std::vector<Column_matrix> _DD_per(this->_K.dimension()+1) ;
+
+    // Copy _DD_col with filtration order (for dimensions q>0)
+    for (int q = 0 ; q <= this->_K.dimension(); ++q)
+    {
+        const std::pair<int, int> s(this->_DD_col.at(q).dimensions()) ;
```
afabri, line 796: You only  pass one argument to the constructor.

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,474 @@
...
+
+// =========== Simplicial chain complex tools
+// Build K sub-chain complex of L (homeomorphic to B^n)
+
+/*!
+ \ingroup PkgHDVFAlgorithmClasses
+
+The class `Duality_simplicial_complex_tools` is dedicated to Alexander duality for 3D surface meshes. Starting from a simplicial chain complex (encoding a 3D surface mesh), it provides methods to embed the complex into a larger icosphere and generate a 3D constrained Delaunay triangulation.
+
+Technically, starting from a Simplicial_chain_complex `_K`, the method `simp_complex_bb` builds a Simplicial_chain_complex L and a Sub_chain_complex_mask K.
```
afabri:
```suggestion
Technically, starting from a Simplicial_chain_complex `_K`, the method `simplicial_complex_bb()` builds a Simplicial_chain_complex L and a Sub_chain_complex_mask K.
```

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,474 @@
...
+// =========== Cubical chain complex tools
+// Build K sub-chain complex of L (homeomorphic to B^n)
+// Adjust L size
+
+/*!
+ \ingroup PkgHDVFAlgorithmClasses
+
+The class `Duality_cubical_complex_tools` is dedicated to Alexander duality for 3D binary volumes.
+
+Starting from a Cubical_chain_complex `_K`, the method `cubical_chain_complex_BB` builds a Cubical_chain_complex L and Sub_chain_complex_mask K.
```
afabri:
```suggestion
Starting from a Cubical_chain_complex `_K`, the method `cubical_chain_complex_bb()` builds a Cubical_chain_complex L and Sub_chain_complex_mask K.
```

afabri: I think it would read nicer if you replaced Cubical_chain_complex with cubical chain complex. In case you want to keep it you must backtick, as it is a type.  Why the letter "K"?  Can you avoid the starting underscore?

afabri: Also the different classes should better be in their own header files.

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+     * The method returns the image of a cycle by the morphism \f$f\f$, that is, a linear combination of critical cells (corresponding to the decomposition of the cycle in the homology basis).
+     * If the annotation has a single non zero coefficient for a given critical cell \f$\sigma\f$, then the cycle belongs to the class of the homology generator \f$g(\sigma)\f$.
+     *
+     * \warning Will raise an error is the chain provided is not a cycle.
+     * \warning The HDVF must be perfect.
+     *
+     * \param[in] chain The cycle to annotate in the homology basis.
+     * \param[in] dim Dimension of the cycle.
+     */
+    typename HDVF_core_type::Column_chain get_annotation(typename HDVF_core_type::Column_chain chain, int dim) const
```
afabri: You better add `typedef HDVF_core_type::Column_chain Column_chain;`  earlier and document it,. and then use just `Column_chain` here.  Should it be a `const&`

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+     */
+    void MW(size_t pi, size_t sigma, int q);
+
+    /**
+     * \brief Gets the annotation of a cycle in the homology basis.
+     *
+     * The method returns the image of a cycle by the morphism \f$f\f$, that is, a linear combination of critical cells (corresponding to the decomposition of the cycle in the homology basis).
+     * If the annotation has a single non zero coefficient for a given critical cell \f$\sigma\f$, then the cycle belongs to the class of the homology generator \f$g(\sigma)\f$.
+     *
+     * \warning Will raise an error is the chain provided is not a cycle.
```
afabri, line 623: In which form the error is raised?  An exception thrown?   A return value. Please document it.

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+    
+    /*!
+     Type of parent Hdvf_core class.
+     */
+    typedef Hdvf_core<ComplexType, OSM::Sparse_chain, OSM::Sparse_matrix> HDVF_core_type ;
+
+    /**
+     * \brief Default constructor.
+     *
+     * Builds a "empty" HDVF associated to K (with all cells critical). By default, the HDVF option is set to OPT_FULL (full reduction computed).
```
afabri:
```suggestion
     * Builds an" empty" HDVF associated to K (with all cells critical). By default, the HDVF option is set to OPT_FULL (full reduction computed).
```

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,841 @@
...
+#include <CGAL/HDVF/Filtration_lower_star.h>
+
+namespace CGAL {
+namespace HDVF {
+
+// Types for persistent homology / cohomology
+
+/*! \brief Type to store persistent intervals filtration indices (birth/death indices)
+ *
+ * For "infinite" intervals borne at index i, the interval is set to (i,i-1)
```
afabri:
```suggestion
 * For "infinite" intervals born at index i, the interval is set to (i,i-1)
```

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,841 @@
...
+ * - First element of the pair: cell entailing the birth of the hole.
+ * - Second element of the pair: cell entailing the death of the hole.
+ *
+ * For infinite intervals, the "infinite cells" is defined as (-1,q+1) where q is the dimension of the chain complex.
+ */
+typedef std::pair<Cell, Cell> CellsPerInterval ;
+
+/*! \brief Template for persistent intervals degrees (birth/death degrees)
+ *
+ * For "infinite" intervals borne at degree d, the interval is set to (d,d-1)
```
afabri:
```suggestion
 * For "infinite" intervals born at degree d, the interval is set to (d,d-1)
```

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
umenohana13: Thanks, corrected

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
umenohana13: OK, corrected

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+
+/** \brief HDVF Enum for the label of cells. */
+static enum FlagType {
+    PRIMARY,
+    SECONDARY,
+    CRITICAL,
+    NONE // For duality and persistence
+};
+
+/** \brief HDVF option (compute only reduced boundary). */
```
afabri, line 39: The constants should go into an `enum`

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+#include <iostream>
+#include <random>
+#include <CGAL/OSM/OSM.h>
+#include <CGAL/OSM/Bitboard.h>
+
+namespace CGAL {
+namespace HDVF {
+
+/** \brief HDVF Enum for the label of cells. */
+static enum FlagType {
```
afabri: @SLoriot does it makes sense that it is `static`?

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+
+/** \brief Structure to represent data for HDVF operations (pairs of cells).
+ *
+ * Cells are always sorted so that the dimension of `sigma` is lesser than the dimension of `tau`.
+ */
+struct Cell_pair {
+    size_t sigma;  /// Index of the first cell
+    size_t tau;    /// Index of the second cell
+    int dim;    /// Dimension of cells: `dim`/`dim`+1 for A and R, `dim`/`dim` for other operations
+};
```
afabri, line 60:
```suggestion
struct Cell_pair {
    /// Index of the first cell
    size_t sigma;
    /// Index of the second cell
    size_t tau;
    /// Dimension of cells: `dim`/`dim`+1 for A and R, `dim`/`dim` for other operations
    int dim;
};
```

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
```diff
@@ -0,0 +1,1081 @@
...
+
+namespace CGAL {
+namespace HDVF {
+
+// Forward declaration of Duality_cubical_complex_tools
+template<typename T> class Duality_cubical_complex_tools ;
+
+// TOOLS
+
+std::ostream & operator<<(std::ostream & out, std::vector<size_t> c)
```
afabri: I am not sure if this will lead to ambiguities as soon as a user also defines this, even when in another namespace.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
```diff
@@ -0,0 +1,1081 @@
...
+    
+    /** \brief Type of vertices coordinates */
+    typedef std::vector<double> Point ;
+private:
+    /* \brief Vector of VTK types associated to cells in each dimension
+     e.g. {1, 3, 8, 11} */
+    static const std::vector<int> VTK_cubtypes ;
+
+public:
+    /** \brief Type used to encode primal or dual construction. */
```
afabri, line 109: Should this `enum` be more global?

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
```diff
@@ -0,0 +1,1081 @@
...
+     * - if the `type` is DUAL and the Cub_object_io contains only cells of maximal dimension (ie.\ binary object), the constructor build the dual associated complex (see below right), which comes to encode \f$2q\f$ connectivity (with \f$q\f$ the dimension of the complex)
+     *
+     *<img src="primal_dual.png" align="center" width=20%/>
+     *
+     * \param[in] cub A Cub_object_io containing a set of "cubical" cells.
+     * \param[in] type Type of construction used (PRIMAL or DUAL).
+     */
+    Cubical_chain_complex(const Cub_object_io& cub,typeComplexCube type);
+
+    /** \brief Friend class `Duality_cubical_complex_tools` provides tools for Alexander duality. */
```
afabri: A comment about "friends".   Usually you declare something friend if it accesses private functions or data members.  Is this the case here, and globally, because I see `friend` declarations all over the place.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
```diff
@@ -0,0 +1,1081 @@
...
+    /**
+     * \brief Assignment operator for cubical chain complexes.
+     *
+     * Stores a copy of an cubical chain complex in *this.
+     *
+     * \param[in] complex The cubical chain complex which will be copied.
+     */
+    Cubical_chain_complex& operator=(const Cubical_chain_complex& complex)
+    {
+        _dim = complex._dim;
```
afabri:
```suggestion
    Cubical_chain_complex& operator=(const Cubical_chain_complex& complex)
    : _dim(complex._dim)
    {
```

afabri: And the same for the initialization of other data members in a constructor.

afabri: And probably the same in constructors of other classes.

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,474 @@
...
+ * <img src="HDVF_dtorus_cohom1_co.png" align="center" width=25%/>
+ * <img src="HDVF_dtorus_cohom2_co.png" align="center" width=25%/>
+ *
+ * All homology / cohomology generators:
+ *
+ *<img src="HDVF_dtorus_all.png" align="center" width=30%/>
+ */
+
+template <typename ComplexType, template <typename, int> typename _ChainType = OSM::Sparse_chain, template <typename, int> typename _SparseMatrixType = OSM::Sparse_matrix, typename VertexIdType = size_t>
+void hdvf_geometric_chain_complex_output_vtk (Hdvf_core<ComplexType, _ChainType, _SparseMatrixType> &hdvf, ComplexType &complex, std::string filename = "test", bool co_faces = false)
```
afabri: For IO functions we try to unify.  This function should be `CGAL::IO::write_VTK()`.  The first parameter makes its unambiguous.

afabri (changes requested):

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: Yes I needed these classes to be friend as the tools for duality really manipulate complexes (and private parts).
I'll check but I'm pretty sure I needed this friendship

But the one I needed more was in Sparse_chain and Sparse_matrix to overload operators "symmetrically" ... I hope it's OK for you...

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: I thought I had been careful about this but now I have a doubt. I'll double check...

umenohana13: I think it's OK (data member initialised by a method called on an argument cannot be initialised in the : ?

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: I confirm that Duality classes are very technical and it was more convenient to make them friend (as they produce new chain complexes + chain complex masks starting from the parent chain complex)

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: This enum is also very technical (it contains the codes used to encode respectively vertices, edges, squares and cubes in VTK). So normally users should not need it

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: You're right, I was lazy... I'll check where I use this and correct...

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
afabri: Sorry about that. Member initialization is for constructors and not for the assignment operator.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
afabri: I had not checked whether friendship is needed or not.  I commented as I rarely saw so many friend declarations in a package, and this might be an indication to better add public functions.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: Done

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
umenohana13: OK so I should move these functions to the IO namespace? 
I'm currently trying to revise all these classes. So I'll modify as you suggest

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: OK I'll check again to see if I can improve public functions and remove friends.
A question: for Sparse_chain and Sparse_matrix, there are a lot of friends because I was told that operators are better overloaded using friends (in order to define "symmetrical" operators). Are friends OK in this setting?

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
umenohana13: Done
-> also renamed "chain_complex_chain_to_vtk" into "chain_to_vtk", but I did not remove both VTK output methods (chain_complex_to_vtk and chain_to_vtk) from GeometricChainComplex models : should I ? As these methods are quite technical, I thought they could remain as such...

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
afabri: If you implement a binary `operator+()`  by calling a public member `operator+=()`  I don't see a need for the binary operator becoming a friend.  In the doxygen comment you should `\relates`  so that it appears on the page of the class.
Do you agree @sloriot ?

albert-github:
@afabri Looks like the documentation didn't succeed as https://cgal.geometryfactory.com could not be reached, looks like the machine / connection is down.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: OK - I trust you. So if you think operators should not be friends I will modify Sparse_chain and Sparse_matrix accordingly. 
I add it to my TODO list

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
umenohana13: I corrected tparam names in the technical functions

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: I removed all friends in Cubical_chain_complex and Simplicial_chain_complex. You were right: I lacked some public methods. Now there's no friendship anymore between chain complex methods and Geometric_chain_complex_tools

HDVF/doc/HDVF/Concepts/Ring.h
```diff
@@ -0,0 +1,50 @@
+/*!
+\ingroup PkgHDVFConcepts
+\cgalConcept
+
+The concept `Ring` describes the requirements for the ring of coefficients used to compute homology in the `HDVF` concept. Besides ring operators, it also specifies the functions needed to test invertibility in the ring.
```
afabri: I am wondering if the `Ring` concept is identical to [`IntegralDomainWithoutDivision`](https://doc.cgal.org/latest/Algebraic_foundations/classIntegralDomainWithoutDivision.html).
If it is different this concept gives an idea what you have to add to the new concept.

HDVF/doc/HDVF/Concepts/SparseChain.h
```diff
@@ -0,0 +1,305 @@
...
+    /// \name Creation, filling
+    /// @{
+
+    /*!
+     * \brief Creates new empty sparse chain.
+     */
+    SparseChain();
+
+    /*!
+     * \brief Creates new empty sparse chain of given size.
```
afabri: It is not empty.  Do you mean that all coefficients have a particular value?

HDVF/doc/HDVF/Concepts/SparseMatrix.h
```diff
@@ -0,0 +1,402 @@
...
+
+    /*!
+     * \brief Type of coefficients stored in the matrix (a model of `Ring`).
+     */
+    typedef Ring CoefficientRing;
+
+    /*!
+     * \brief Matrix and chain type (either ROW or COLUMN).
+     */
+    typedef int ChainTypeFlag;
```
afabri: I am not sure about the choice of the name.  The Eigen library refers to a [page](https://netlib.org/linalg/html_templates/node90.html) (20 years old) where they speak about "Storage Formats"

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+
+ The class `Hdvf_core` is the core implementation of homological discrete vector fields (HDVF for short). The ring of coefficients for homology computation must be a model of `Ring`.
+
+ An enumeration `Flag_type` is defined in the `HDVF` namespace and the `Hdvf_core` class maps each cell to one of the flags (namely `PRIMARY`, `SECONDARY`, `CRITICAL`). The NONE flag is used in child classes (such as `Hdvf_duality` or `Hdvf_reduced`) when computing relative homology on a sub-complex.
+ The flag of each cell is stored in an appropriate structure and getters are provided to access to this information.
+
+ The `Hdvf_core` class stores the associated reduction in sparse matrices: row-major for \f$f\f$, and column-major for \f$g\f$, \f$h\f$ and \f$\partial'\f$. Getters are provided to access this information. However, according to the chosen HDVF computation option (`OPT_BND`, `OPT_F`, `OPT_G`, `OPT_FULL`) the reduction can be computed only partially (and thus faster).
+
+ The class provides perfect HDVF constuction operations: `compute_perfect_hdvf()` and `compute_rand_perfect_hdvf()`, which build perfect HDVFs by pairing iteratively critical cells through the `A()` operation.
```
afabri:
```suggestion
 The class provides perfect HDVF construction operations: `compute_perfect_hdvf()` and `compute_rand_perfect_hdvf()`, which build perfect HDVFs by pairing iteratively critical cells through the `A()` operation.
```

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+#include <cassert>
+#include <iostream>
+#include <random>
+#include <CGAL/OSM/OSM.h>
+#include <CGAL/OSM/Bitboard.h>
+
+namespace CGAL {
+namespace HDVF {
+
+/** \brief HDVF Enum for the label of cells. */
```
afabri, line 31: Let's find a better name than `FlagType` in a zoom session.

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+    }
+    return out ;
+}
+
+/*!
+ \ingroup PkgHDVFAlgorithmClasses
+
+ The class `Hdvf_core` is the core implementation of homological discrete vector fields (HDVF for short). The ring of coefficients for homology computation must be a model of `Ring`.
+
+ An enumeration `Flag_type` is defined in the `HDVF` namespace and the `Hdvf_core` class maps each cell to one of the flags (namely `PRIMARY`, `SECONDARY`, `CRITICAL`). The NONE flag is used in child classes (such as `Hdvf_duality` or `Hdvf_reduced`) when computing relative homology on a sub-complex.
```
afabri: `Hdvf_reduced`   does not exist.

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1213 @@
...
+ The `Hdvf_core` class stores the associated reduction in sparse matrices: row-major for \f$f\f$, and column-major for \f$g\f$, \f$h\f$ and \f$\partial'\f$. Getters are provided to access this information. However, according to the chosen HDVF computation option (`OPT_BND`, `OPT_F`, `OPT_G`, `OPT_FULL`) the reduction can be computed only partially (and thus faster).
+
+ The class provides perfect HDVF construction operations: `compute_perfect_hdvf()` and `compute_rand_perfect_hdvf()`, which build perfect HDVFs by pairing iteratively critical cells through the `A()` operation.
+
+ If the user wishes to build an HDVF using other criteria, several `find_pair_A()` functions are provided (searching for valid pairs of cells for `A`respecting various constraints). The `A` operation can be applied to any pair returned by these functions.
+
+ Homology/cohomology generators are actually algebraic objects, namely chains. Methods `homology_chain()` and `cohomology_chain()` return the homology and cohomology generator chain associated to a given critical cell. VTK export functions output all the cells of such chains with non zero coefficients.
+
+
+ \cgalModels{HDVF}
```
afabri, line 91: This links to namespace and not to the concept. Maybe @sloriot can help.

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * The method exports the labels of every cells in each dimension.
+     * Encoding used:
+     * - `PRIMARY`: -1
+     * - `SECONDARY`: +1
+     * - `CRITICAL`: 0
+     *
+     * \returns A vector containing, for each dimension, the vector of labels by cell index.
+     */
+    virtual std::vector<std::vector<int> > psc_labels () const
```
afabri: I am not sure about the need for `virtual` functions.  In general you need that for polymorphic types, that is for a `vector<Shape>` storing `Circle` and `Rectangle` both derived from `Shape` with a virtual `draw()` function.   Are we in this kind of situation?

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
```diff
@@ -0,0 +1,821 @@
...
+    struct Persistent_hole {
+        size_t time_birth, time_death;
+        Degree degree_birth, degree_death;
+        Cell cell_birth, cell_death;
+        
+        /** \brief Computes the (degree) duration of a persistent interval (ie.\ persistent hole)
+         *
+         * By definitions of "default" values, infinite intervals have a duration of -1.
+         */
+        inline Degree duration () const
```
afabri, line 114: We have to discuss in the next zoom call  the terms duration and degree.

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
umenohana13: I limited friends as much as possible. All the remaining friends use private or protected members and are "friends" because they provide optimised overloads depending on chain/matrix types (so that the choice of the right method is guided by the ChainTypeFlag).

HDVF/doc/HDVF/Concepts/Ring.h
umenohana13: Yes, you're right. Actually in homology theory, it is usually not specified if the ring must be free of "divisors of zero" (ie an integral domain) or not. But in practice, we always use integral domains. So I think it's nice to refer to this concept (I normally corrected all concepts and models).

HDVF/doc/HDVF/Concepts/SparseChain.h
umenohana13: Yes, the chain is empty and thus encodes the 0 chain (as empty <=> 0). 
I clarifier (I think) as follows:
```
/*!
     * \brief Creates new empty sparse chain.
     *
     * Creates a sparse chain encoding an empty linear combination of cells.
     */
    SparseChain();

    /*!
     * \brief Creates new empty sparse chain (ie. zero-chain) of given size.
     *
     * Constructor with size, initializes an empty sparse chain encoding a linear combination of cells with all coefficients null.
     */
    SparseChain(size_t chainSize);
```

HDVF/doc/HDVF/Concepts/SparseMatrix.h
umenohana13: We can modify without problem. Yes "storage format" is maybe clearer (either COLUMN or ROW). 
So I replaced ChainTypeFlag by StorageFormat "all over the place".
I also slightly updated the comment (in all concepts and classes):
`* \brief Matrix and chain storage format (either ROW or COLUMN).`

HDVF/include/CGAL/HDVF/Hdvf_core.h
umenohana13: I fully agree. 
Before the zoom, some tries  (usually in the papers we say that P, S or C is a label or that we define a partition of cells into primary, secondary, critical):
- HDVFFlag
- HDVFLabel
- HDVFPartitionLabel
I think YOU should tell me (I'm too much used to HDVFs to be clear on this)
-

HDVF/include/CGAL/HDVF/Hdvf_core.h
umenohana13: You're right ... actually "not yet" exist (I was short of time). But it's a straightforward adaptation of Hdvf_duality that can be interesting for users (reduced homology is pretty much used in practice while no software provide its computation...). 
For now I removed it ... later :)

HDVF/include/CGAL/HDVF/Hdvf_core.h
umenohana13: I'm lost Andreas - can we discuss this during our zoom?

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
umenohana13: I thought I did not use polymorphism, BUT it seems I use some (as I noticed that if I remove the virtual of "flag" method in Hdvf_core, I meet problems for Hdvf_duality - where only flags of the current sub_chain_complex_mask are exported).
I suggest that I write tests for HDVFs quickly and that I use these tests then to remove all unnecessary virtual (anyway all virtual in Hdvf_duality and Hdvf_persistance are useless - and just come from rough "copy" ...)

HDVF/include/CGAL/HDVF/Hdvf_persistence.h
umenohana13: OK

HDVF/doc/HDVF/Concepts/GeometricChainComplex.h
```diff
@@ -0,0 +1,50 @@
...
+class GeometricChainComplex
+{
+public:
+/// \name Types
+/// @{
+
+/*!
+Type of coefficients used to compute homology (model of `IntegralDomainWithoutDivision`)
+ */
+typedef IntegralDomainWithoutDivision CoefficientRing;
```
afabri:
```suggestion
typedef unspecified_type Coefficient_ring;
```

afabri: Additionally, is this typedef not already in the concept  `AbstractChainComplex`?

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -0,0 +1,145 @@
...
+
+/// \name Output functions
+/// @{
+
+/*!
+Outputs the chain complex in text mode.
+
+By default, outputs the complex to `std::cout`.
+*/
+friend std::ostream& operator<<(std::ostream& out = std::cout, const AbstractChainComplex& complex) const;
```
afabri: A first argument cannot have a default, if there is a second one without a default.

afabri: Also it makes no sense to declare a function `friend`  in a concept.

HDVF/include/CGAL/HDVF/Simplex.h
```diff
@@ -0,0 +1,147 @@
...
+    // Indices of the simplex vertices - sorted vector
+    Simplex_data_type _vertices;
+
+    public:
+    /** \brief Constructor from a vector of vertices indices.
+     *
+     * \param[in] vertices Vector of the simplex vertices indices (a simplex of dimension \f$q\f$ has \f$q+1\f$ vertices and the vector must be **sorted**).
+     * \param[in] sort_data If `true` vectors of vertex indices are sorted, if `false` indices are assumed to be sorted (faster).
+     */
+    Simplex(const Simplex_data_type& vertices, bool sort_data = false) : _vertices(vertices) {
```
afabri: What is `Simplex_data_type` ?

HDVF/doc/HDVF/Concepts/SparseChain.h
```diff
@@ -0,0 +1,307 @@
...
+class SparseChain {
+
+public:
+    /// \name Types
+    /// @{
+
+    /*!
+     * \brief Type of coefficients stored in the matrix (a model of `IntegralDomainWithoutDivision`).
+     */
+    typedef IntegralDomainWithoutDivision CoefficientRing;
```
afabri:
```suggestion
    typedef unspecified_type Coefficient_ring;
```

afabri: And the same for the matrix class.

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,494 @@
...
+    /** \brief Type returned by `simplicial_chain_complex_bb()`.
+     *
+     * The structure contains a triple:
+     * - A simplicial chain complex `L` (homeomorphic to \f$\mathbb B^3\f$).
+     * - A sub chain complex mask `K` (encoding the initial mesh)
+     * - The vector of vertices coordinates of `L`
+     */
+    typedef struct {
+        ChainComplex& L ;
+        Sub_chain_complex& K ;
```
afabri: @sloriot any idea why we do not see the second data member [here](https://cgal.github.io/9004/v1/HDVF/structCGAL_1_1Homological__discrete__vector__field_1_1Duality__simplicial__complex__tools_1_1Complex__duality__data.html)?

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,494 @@
...
+- K (Sub_chain_complex_mask) : Sub_chain_complex_mask identifying _K inside L
+
+\tparam CoefficientRing a model of the `IntegralDomainWithoutDivision` concept providing the ring used to compute homology.
+*/
+
+template<typename CoefficientRing>
+class Duality_simplicial_complex_tools {
+public:
+    /** \brief Type of simplicial chain complex encoding `L`. */
+    typedef Simplicial_chain_complex<CoefficientRing> ChainComplex ;
```
afabri: Should not be CamelCase

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
```diff
@@ -3,6 +3,7 @@
 \cgalConcept
 
 The concept `AbstractChainComplex` describes the requirements for (topological) chain complexes associated to abstract complexes used in the concept `HDVF`.
+The concept `AbstractChainComplex` describes the requirements for (topological) chain complexes associated to abstract complexes used in the concept `CGAL::HDVF`.
```
afabri: NO, concepts are not in the namespace.

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
umenohana13: Done :)

HDVF/doc/HDVF/Concepts/AbstractChainComplex.h
umenohana13: OK so just `HDVF` (corrected, sorry...)

HDVF/include/CGAL/OSM/Sparse_matrix.h
```diff
@@ -0,0 +1,2072 @@
...
+     *
+     * Removes column of index `index` whatever the `StorageFormat` of the matrix. For column matrices, it just comes to the `\=` operator and for row matrices, it entails a traversal of the matrix.
+     *
+     * \param[in] matrix Reference on the matrix to modify.
+     * \param[in] index The index to remove.
+     *
+     * \return The modified matrix representing the result.
+     */
+    template <typename _CT, int _CTF>
+    friend Sparse_matrix<_CT, _CTF>& del_column(Sparse_matrix<_CT, _CTF>& matrix, size_t index);
```
afabri: Can you please replace `del_`  with `remove_`

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,414 @@
...
+ */
+typedef unspecified_type Complex;
+    
+/*! \brief Type of coefficients used to compute homology. */
+typedef Complex::Coefficient_ring Coefficient_ring;
+
+/*!
+ * \brief Type of sparse chains (a model of `SparseChain`).
+ */
+    typedef unspecified_type SparseChainType;
```
afabri:
```suggestion
    typedef unspecified_type Sparse_chain;
```

HDVF/doc/HDVF/Concepts/HDVF.h
```diff
@@ -0,0 +1,414 @@
...
+
+/*!
+ * \brief Type of sparse matrices (a model of `SparseMatrix`).
+ */
+    typedef unspecified_type SparseMatrixType;
```
afabri:
```suggestion
    typedef unspecified_type Sparse_matrix;
```

afabri: and then you have to fix the chain types below.

HDVF/include/CGAL/HDVF/Hdvf_core.h
```diff
@@ -0,0 +1,1219 @@
...
+    }
+    return out ;
+}
+
+/*!
+ \ingroup PkgHDVFAlgorithmClasses
+
+ The class `Hdvf_core` is the core implementation of homological discrete vector fields (HDVF for short). The ring of coefficients for homology computation must be a model of `IntegralDomainWithoutDivision`.
+
+ An enumeration `PSC_flag` is defined in the `HDVF` namespace and the `Hdvf_core` class maps each cell to one of the flags (namely `PRIMARY`, `SECONDARY`, `CRITICAL`). The NONE `PSC_flag` is used in child classes (such as `Hdvf_duality`) when computing relative homology on a sub-complex.
```
afabri:
```suggestion
 An enumeration `PSC_flag` is defined in the `Homological_discrete_vector_field` namespace and the `Hdvf_core` class maps each cell to one of the flags (namely `PRIMARY`, `SECONDARY`, `CRITICAL`). The NONE `PSC_flag` is used in child classes (such as `Hdvf_duality`) when computing relative homology on a sub-complex.
```

HDVF/include/CGAL/HDVF/Filtration_lower_star.h
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /*! \brief Constructor from a function mapping vertices to degrees.
+     *
+     * The constructor computes all cells degrees as the minimum of the degrees of their vertices (obtained through `deg_fun`) and sorts all the cells of the complex to fulfill the filtration ordering constraints.
+     *
+     * \param[in] K Constant reference to the underlying complex.
+     * \param[in] deg_fun Function mapping vertices of `K` to their degree.
+     */
+    Filtration_lower_star(const ChainComplex& K, std::function<Degree(size_t)>& deg_fun) : Filtration_parent(K)
```
afabri: The parameter `deg_fun`  must be renamed as otherwise there is a doxygen link. See [here](https://cgal.github.io/9004/v1/HDVF/classCGAL_1_1Homological__discrete__vector__field_1_1Filtration__lower__star.html#abd5d7eac3c5509887aa5bcbf32ab2abd)

HDVF/include/CGAL/HDVF/Cubical_chain_complex.h
```diff
@@ -0,0 +1,1120 @@
...
+        std::vector<double> res ;
+        for (size_t c : coords)
+            res.push_back(c/2. + .5) ;
+        for (size_t i = coords.size(); i<3; ++i) // points must be 3D
+            res.push_back(0) ;
+        return res ;
+    }
+
+    /** \brief Gets the vector of vertex coordinates  */
+    const std::vector<Point>& get_vertices_coords() const
```
afabri: I am not sure that the return type can be a ` const&`   as you return a temporary variable. @sloriot ??

umenohana13: As soon as we are done with a "proper" integration of points, I will build the coordinates vector as a private data. So this will become OK.
But you're right, I'm playing with C++ "permissions" for cubical complexes. It works but might be unsafe...

HDVF/include/CGAL/HDVF/Filtration_lower_star.h
umenohana13: Done

HDVF/doc/HDVF/Concepts/HDVF.h
umenohana13: done

HDVF/doc/HDVF/Concepts/HDVF.h
umenohana13: done

HDVF/include/CGAL/OSM/Sparse_matrix.h
umenohana13: Done

HDVF/include/CGAL/HDVF/Mesh_object_io.h
```diff
@@ -0,0 +1,658 @@
...
+    double at(size_t i) const { return _coords.at(i) ;}
+    double& operator[](size_t i) { return _coords.at(i) ;}
+    Io_node_type& operator=(const Io_node_type& v) { _coords = v._coords ; return *this ; }
+    void push_back(double x) { _coords.push_back(x) ; }
+    std::vector<double> get_coords() const { return _coords; }
+
+    friend Io_node_type & operator+(const Io_node_type &v1, const Io_node_type &v2)
+    {
+        assert(v1.size() == v2.size()) ;
+        Io_node_type &tmp = *new(Io_node_type)(v1.size(),0) ;
```
afabri: You have a memory leak here.

umenohana13:
The Hdvf_traits_3 works great, but we would need a "generic" Hdvf_traits_d class. Especially for cubical complexes, dimension of points needs to be "generic" (and not only 2 or 3).

HDVF/include/CGAL/HDVF/Mesh_object_io.h
umenohana13: Fixed (by adding a _points vector in Cubical_chain_complex.cpp)
But I'm bothered: how to generate points properly according to the dimension: Point_2 and Point_3 are created by given x, y ... but Point_d are created using an iterator...
I've tried to create overloaded "to_point" functions, taking the vector of doubles of coordinates and returning Point in the right dimension. This function is called by "compute_point" in the class. I thought that overloading would solve the problem. But as soon as I try to add the Point_2 version, I get compilation errors...

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+ Therefore, `are_same_cycles` will return `true` for \f$\alpha\f$  and \f$g(\sigma)\f$,  but `false` for \f$\beta\f$  and \f$g(\sigma)\f$.
+ \cgalFigureEnd
+
+\cgalModels{HDVF}
+
+\tparam ChainComplex a model of the `AbstractChainComplex` concept, providing the type of abstract chain complex used.
+ */
+
+template<typename ChainComplex>
+class Hdvf : public Hdvf_core<ChainComplex, OSM::Sparse_chain, OSM::Sparse_matrix> {
```
afabri: How can that compile without tparams for `OSM::Sparse_chain`

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+public:
+    /*! \brief Type of coefficients used to compute homology. */
+    typedef typename ChainComplex::Coefficient_ring Coefficient_ring;
+
+    /*!
+     Type of parent Hdvf_core class.
+     */
+    typedef Hdvf_core<ChainComplex, OSM::Sparse_chain, OSM::Sparse_matrix> HDVF_parent ;
```
afabri: Can we rename that to `Base`

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1474 @@
...
+Hdvf<ChainComplex>::Hdvf(const ChainComplex& K, int hdvf_opt) : Hdvf_core<ChainComplex, OSM::Sparse_chain, OSM::Sparse_matrix>(K, hdvf_opt) { }
+
+
+
+// Methods to find a pair of cells for M in dimension q
+//  -> <f(cell1),cell2> = +- 1 with cell1 primary, cell2 critical
+// First version: returns a pair of dimensions q
+// Second version: returns all the pairs containing sigma
+
+// \brief find a valid Cell_pair for M in dimension q
```
afabri, line 738: This is NOT a doxygen comment.

HDVF/include/CGAL/HDVF/Hdvf.h
```diff
@@ -0,0 +1,1477 @@
...
+     * The method returns the image of a cycle by the morphism \f$f\f$, that is, a linear combination of critical cells (corresponding to the decomposition of the cycle in the homology basis).
+     * If the annotation has a single non zero coefficient for a given critical cell \f$\sigma\f$, then the cycle belongs to the class of the homology generator \f$g(\sigma)\f$.
+     *
+     * \warning Will raise an error is the chain provided is not a cycle.
+     * \warning The HDVF must be perfect.
+     *
+     * \param[in] chain The cycle to annotate in the homology basis.
+     * \param[in] dim Dimension of the cycle.
+     */
+    typename Base::Column_chain get_annotation(typename Base::Column_chain chain, int dim) const
```
afabri: There is no need to write `Base::Column_chain`   as we inherit `Column_chain`  from the base class.

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -0,0 +1,582 @@
...
+
+    /** \brief Type returned by `simplicial_chain_complex_bb()`.
+     *
+     * The structure contains a triple:
+     * - A simplicial chain complex `L` (homeomorphic to \f$\mathbb B^3\f$).
+     * - A sub chain complex mask `K` (encoding the initial mesh)
+     * - The vector of vertex coordinates of `L`
+     */
+    typedef struct {
+        Chain_complex& L ;
```
afabri: This should not be a reference, and there should not be a ` new` , but something with a ` std::move()`, to be explained by @sloriot

sloriot: This is not problematic by itself but code like `Chain_complex& L = *new Chain_complex(tetL) ;` is as the ownership is then taken by the struct. It cannot be a public class IMO. Note that if you build the class locally in a function and return it, the return type optimization will make the class to be built in the calling scope. Then there should not be any copy.

HDVF/include/CGAL/HDVF/Icosphere_object_io.h
```diff
@@ -0,0 +1,133 @@
...
+namespace CGAL {
+namespace Homological_discrete_vector_field {
+
+/* Class used to build an icosphere embedding a triangular mesh (for Alexander Duality)
+
+ \tparam Traits a geometric traits class model of the `HDVFTraits` concept.
+*/
+
+template <typename Traits>
+class Icosphere_object_io : public Mesh_object_io<Traits>
```
afabri, line 38: Note that CGAL has the function [`make_icosahedron()`](https://cgal.github.io/9115/v1/BGL/group__PkgBGLGeneratorFct.html#ga03d4d6b1b7cb4ef6c460487240e6bb10)

HDVF/include/CGAL/HDVF/Hdvf.h
umenohana13: Yes: methods are documented in the class (I just wanted to keep some documentation here, out of doxygen, to help and understand the code)

umenohana13:
The word "**cell**" is used in Abstract_simplicial_chain_complex and Simplicial_chain_complex to speak about simplices: this is actually the only way to have a coherent description of complexes whatever their type (simplicial, cubical and later on, linear - such as combinatorial maps).

HDVF/include/CGAL/HDVF/Hdvf.h
umenohana13: Done
Commit: 22931c62ba6830eecd585afe2a812f84dfa2fa63

HDVF/include/CGAL/HDVF/Icosphere_object_io.h
umenohana13: Yes, in the CGAL version, I use this function to generate the icosahedrons Surface_mesh. But here, I need to create a Mesh_object_io ... and anyway as soon as possible, this tetgen version will disappear:)

HDVF/include/CGAL/HDVF/Geometric_chain_complex_tools.h
```diff
@@ -298,24 +298,38 @@ namespace Homological_discrete_vector_field {
...
+ * \tparam ChainComplex Type of the underlying chain complex.
  *
  * The structure contains a pair:
- * - A simplicial chain complex `L_complex` (homeomorphic to \f$\mathbb B^3\f$).
- * - A sub chain complex mask `K_complex` (encoding the initial mesh inside `L_complex`)
+ * - `L_complex`: a pointer over simplicial chain complex (homeomorphic to \f$\mathbb B^3\f$).
+ * - `K_complex`: a sub chain complex mask  (encoding the initial mesh inside `*L_complex`)
  */
 template <typename ChainComplex>
 class Complex_duality_data_t {
```
afabri, line 329: As I wrote in my review I do not see a need for this class.   The function `dualize()` should just return a `std::pair`  of `std::shared_ptr` for the two complexes.

afabri:
@albert-github I have a doxygen question.
Why does a link to a member function gets generated for a parameter [here](https://cgal.github.io/9004/v1/HDVF/classCGAL_1_1Homological__discrete__vector__field_1_1Filtration__lower__star.html#adc27cd5dc5b7d50ae32c3582bd8ddffc)  ??

albert-github:
@afabri 

The link you provided leads me to:

image

in the "Protected Member Functions" section.
I don't see a link here or do you mean why is this function here at all?

albert-github:
After my initial remark / question 
I did dive a bit into the matter and saw that the function is protected but is also documented and doxygen has no mechanism like `EXTRACT_PRIVATE` to exclude protected members (as far as I know).
See also the `create_shape` function in the `Shape_detection` package e.g. at https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/CGAL-6.2-Ic-48/master/Shape_detection/classCGAL_1_1Shape__detection_1_1Shape__base.html#a4f4e29f440b04dccebd9597f98bb3b2e

**Edit**
Looks like the request for `EXTRACT_PROTECTED` is already an older request https://github.com/doxygen/doxygen/issues/898

afabri:
I mean this one:

image

albert-github:
I was looking at the wrong function (sorry).

the word `degree` here is seen as a function and doxygen's automatic linking mapping kicks in here a function with this name exists as well.
Best is to rename the argument.

Note the same happens for other functions as well (e.g. `Filtration_lower_star` in the public member function section)

umenohana13:
Questions for May 20th:
- Names / tparams related to Sparse_matrix (Sparse_matrix_core and Sparse_matrix curryfication providing Sparse_matrix_type and corresponding underlying Sparse_chain_type)
- How to describe this in the concepts?
- is_invertible() and inverse() for integers?
- Hdvf_persistence and Hdvf_duality must take Complexes built over Sub_sparse_matrix... is there a way to guarantee / enforce it?

HDVF/doc/HDVF/dependencies
```diff
@@ -0,0 +1,9 @@
+Manual
+Kernel_23
+Kernel_d
+STL_Extension
+Algebraic_foundations
+Number_types
+Circulator
+Stream_support
+Surface_mesh
```
afabri, line 9:
```suggestion
Surface_mesh
Triangulation_3

```

afabri: We then get a link [here](https://cgal.github.io/9004/v0/HDVF/classCGAL_1_1Homological__discrete__vector__field_1_1Triangulation__3__io.html)

HDVF/include/CGAL/OSM/Sparse_matrix.h
```diff
@@ -0,0 +1,2232 @@
...
+     * @param other The second matrix.
+     * @return A boolean.
+     * @{
+     */
+
+    /** \relates Sparse_matrix_core
+     *
+     * \brief Comparison of two COLUMN matrices.
+     */
+    template <typename _CT, template <typename, int> typename SCT>
```
afabri, line 330: You have to write what `SCT`   means

HDVF/include/CGAL/OSM/Sparse_matrix.h
```diff
@@ -0,0 +1,2232 @@
...
+    friend bool operator==(const Sparse_matrix_core<_CT, OSM::COLUMN, SCT>& matrix, const Sparse_matrix_core<_CT, OSM::COLUMN, SCT> &other);
+
+    /** \relates Sparse_matrix_core
+     *
+     * \brief Comparison of a COLUMN  and a ROW matrix.
```
afabri, line 335:
```suggestion
     * \brief Comparison of a `COLUMN`  and a `ROW` matrix.
```

PR #9005, opened 2025-07-30
Mesh_3 - fix "facet is not in its conflict zone" crash

janetournois:
## Summary of Changes

In theory in Mesh_3, a Steiner point cannot be hidden.
In practice with inexact constructions (even with the robust circumcenter), it can happen that the Steiner point is hidden.
This case leads to an early exit of `RT3::find_conflicts()` which was inconsistent with `facet_is_in_its_cz`

This PR fixes it

## Release Management

* Affected package(s): Triangulation_3, Mesh_3


sloriot:

PR #9006, opened 2025-07-30
Add to way to indicate that features were detected outside of the class

sloriot:
This is an issue for the default initialization that will redo the feature detection and overwrite the face patch ids


sloriot:

PR #9007, opened 2025-07-31
Allow Side_of_triangle to take ownership on an external AABB-tree

sloriot:
Convenient for using only a subset of faces for example, or to force the building of the tree


sloriot:

Issue #8962, opened 2025-06-30
CDT_3 - output surface has a hole though input surface does not


janetournois:
## Issue Details

The surface of the CDT of this input PLC has holes, though the input does not have holes.

[final_remeshed_no_tri.zip](https://github.com/user-attachments/files/20981945/final_remeshed_no_tri.zip)

The input faces are not perfectly coplanar, but we still don't expect holes.

The experiments were made in CGAL lab

## Environment

* CGAL version: master branch 2504f0ccac31faa380284d6cc3dffacace1f0a3d

hihry:
I tried to solve it, could you review it?

hihry:
hey @lrineau I worked on the issue could you have a look, this is like my first contribution to a a big codebase and i am unwantedly excited

PR #9015, opened 2025-08-01
Fix: Merge duplicate points and orient polygons before triangulation to prevent invalid meshes

hihry:
Fix: #8962 Clean polygon soup before triangulation

- Merge duplicate points to avoid degenerate vertices
- Orient polygon soup for consistent face winding
- Prevents invalid meshes and triangulation failures in
  make_conforming_constrained_Delaunay_triangulation_3()


lrineau:
I do not think we can merge this pull-request. The preprocessing has to be done in a preprocessing step, and not in that function. If the user already as done some sort of preprocessing, that would be a waste of computation time to do it again.

Closing now.

PR #9027, opened 2025-08-07
Avoid cascade with EPECK centroid 

sloriot:


sloriot:
Successfully tested CGAL-6.1-Ic-217

PR #9031, opened 2025-08-12
3D Constrained Delaunay:  Add example for vertex_with_info()

### LINKED ISSUE #8955: CDT_3: point cell with infos

afabri:
## Summary of Changes

This example exploits the fact that the first vertices correspond to the input points, assuming that they are unique.

## Release Management

* Affected package(s): Constrained_triangulation_3
* Issue(s) solved (if any): fix #8955


ImperatorS79:
I just tested it, and the infos end up being initialized to 0, not 1,2,3,4... What version of cgal did you use (I used the master one) ?

afabri:
I didn't add any code to transfer it, but if you loop over the first N vertices you can set the info yourself.

ImperatorS79:
Do I have any guarantee that the order of the vertices will be kept when I iterate over them using CGAL's iterators ? I have specific infos to add to each vertices.

afabri:
In the current implementation you have the guarantee.

sloriot:

sloriot:
@lrineau Waiting for your review before merging.

afabri:
The goal was to unblock the user. Up to you @lrineau to do the real thing, I guess we cannot do spatial sort as then the vertices would  be out of order.

sloriot:
@lrineau do we merge or do we close?

PR #9038, opened 2025-08-20
Stream Support: Return false if nothing was read

afabri:
## Summary of Changes

The `read_..._WKT()`  functions have `bool` as return type, but currently only check if the stream is bad, which is not the case if no input was read.

This PR was triggered by the fact that CGALlab could not WTK files with polygons, but only polylines.  It now can read polygons, but does only draw the contours and not the interior.

## Release Management


sloriot:
Successfully tested in CGAL-6.1-Ic-240

PR #9044, opened 2025-08-27
Fix the handling of non-simply connected faces in the output of clip with plane

sloriot:
In case we triangulate, we have only one CDT2
In case we don't triangulate, we arrange intersection polygons and only triangulate nested polygons (it's a bit more than non-simply connected faces as the last level of nesting could be handled similarly to simply connected faces).


sloriot:

PR #9050, opened 2025-09-01
Optimize the `do_intersect()` functions of the 2D Regularized Boolean Set Operation" package (made it tolerant to inexact kernels.)

efifogel:
## Summary of Changes

Optimized `do_intersect(polygon, polygon)`, `do_intersect(begin, end)`, and `do_intersect(begin1, end1, begin2, end2)`:
(i) Terminated the execution once an intersection is detected. (In the past, the intersection was computed in one phase and examined in a subsequent phase.)
(ii) Made the variants of the free functions `do_intersect()` that apply to linear polygons, robust even with an inexact-construction kernel. The variants that apply to generalized polygons endure inexact constructions much more than before; however, there are rare degenerate cases that are still require an exact construction kernel.

In general, the changes described here do not affect the default interface, so a small feature is not required. However, it is a major impact, and it does affect the interface as described bellow, and even somehow break backward compatibility.

Recently, the code of the package "2D Regularized Boolean Set Operations" was optimized. In particular, a 3rd optional parameter was introduced in the free functions. It determined whether the boundaries of the input polygons are treated as cyclic sequences of single (`x`-monotone) segments or as a cyclic sequences of (`x`-monotone) polylines. The change described here eliminates this 3rd parameter, and brings the interface of the `do_intersect() function back to the original design with two input polygons.

## Release Management

* Affected package(s): Boolean_set_operations_2, Surface_sweep, Arrangement_on_surface_2
* Feature/Small Feature (if any): [here](https://cgalwiki.geometryfactory.com/CGAL/Members/wiki/Features/Small_Features/do_intersect_polygon_2_predicates_only)


sloriot:
Is there anything from https://github.com/CGAL/cgal/pull/5284 that we want to keep/import?

efifogel:
The two conclusions in #5284 <https://github.com/CGAL/cgal/pull/5284> still
hold:

   1. Rename the package to Regularized_boolean_set_operations_2
   2. Introduce the namespace Regularized_boolean_set_operations_2

BW, it might be easier to re-implement the above rather than to integrate
the old branch.
Perhaps we can do it as well before the next version is released.

PS, I've realized that my implementation is deficient.
It works well for linear curves, but if the input consists of general
polygons bounded by non-linear curves, it may fail.
In the general case 2 curves may intersect (in their interiors) at a point
where they only touch (their second derivatives have opposite signs).
In such cases two polygons bounded by such curves may not necessarily
intersect in their interiors.

Anyhow, a simple solution is to add a tag to the relevant traits that
indicates that the curves are linear.

If the curves are linear, continue as planned, that is, abort the procedure
and return true (to indicate that an intersection has been detected).
Otherwise, check the multiplicity of the intersection.
  If the curves change relative places, abort the procedure and return true.
  Otherwise, continue

The consequence of the above is that the procedure will not be stable when
using EPIC for non linear curves (cause we generate a new point---the
intersection point, and we use it further down the road).
It will still abort once a definitive intersection is detected.

I wanted to add a tag that indicates that curves are linear for a while now.
I believe that we can use it to expedite other procedures as well.

The bottom line is the PR is not ready yet...
Naturally, I'll add the necessary tests.

Efi
                         _/

efifogel:
This PR is now ready for integration
                         _/

sloriot:
```
 The following files have tabs:
Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
```

efifogel:
Fixed
                         _/

sloriot:
Tests/examples are failing in CGAL-6.1-Ic-234 (not sure if it's a bad interaction with another PR).

efifogel:
Some failures were because of this PR.
I fixed them.
Let's try again...
                         _/

sloriot:
Only warnings left now! There might be other warnings coming from the other branch.

efifogel:
I fixed one thing.
There might be more, but I'm abroad and the Internet connection is really
bad, so I'll check again when I'm back (Tue.)
Feel free to run it again or not till then...
                         _/

efifogel:
I fixed the remaining issues.
Please re-test; thanks.
                         _/

Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_on_surface_base_2.h
```diff
@@ -1268,160 +1165,214 @@ class Gps_on_surface_base_2
...
+  bool do_intersect_divide_and_conquer2(InputIterator begin, InputIterator end, std::size_t k, Merge merge) {
+    std::vector<Arr_entry> arr_entries;
+    arr_entries.reserve(k);
+    arr_entries.resize(1);
+    arr_entries[0].first = m_arr;
+    arr_entries[0].second = m_vertices;
+    std::size_t size = std::distance(begin, end);
+    auto it = begin;
+    while (it != end) {
+      std::size_t num = std::min(size+1, k);
```
afabri:
```suggestion
      std::size_t num = (std::min)(size+1, k);
```

efifogel:
Fixed
                         _/

Boolean_set_operations_2/include/CGAL/Boolean_set_operations_2/Gps_agg_op_surface_sweep_2.h
```diff
@@ -70,21 +71,17 @@ class Gps_agg_op_surface_sweep_2 : public Ss2::Surface_sweep_2<Visitor_> {
...
-    typedef Unique_hash_map<Vertex_handle, Event*>      Vertices_map;
-    typedef typename Gt2::Compare_xy_2                  Compare_xy_2;
+    using Vertices_map = Unique_hash_map<Vertex_handle, Event*>;
+    using Compare_xy_2 = typename Gt2::Compare_xy_2;
 
-    this->m_visitor->before_sweep();
     // Allocate all of the Subcurve objects as one block.
-    this->m_num_of_subCurves =
-      static_cast<unsigned int>(std::distance(curves_begin, curves_end));
+    this->m_num_of_subCurves = std::distance(curves_begin, curves_end);
```
sloriot:
```suggestion
    this->m_num_of_subCurves = static_cast<unsigned int>(std::distance(curves_begin, curves_end));
```

sloriot:
Successfully tested in CGAL-6.2-Ic-5

sloriot:
Successfully tested in CGAL-6.2-Ic-39

TODO: after the merge create an issue to mention to create a non-regularized version of `do_intersect`, following the discussion (that I cannot find for now) where we agreed to add a new namespace for the regularized version so that the version in the CGAL namespace has the same behavior as the Kernel do-intersect

PR #9053, opened 2025-09-02
Tolerance for FT without sqrt

sloriot:
Used at a non critical part used for discretization


sloriot:

PR #9054, opened 2025-09-03
Hide warning messages that are too verbose

sloriot:
Note that prior to 6.0 all those messages were written into a debug file.


sloriot:
Successfully tested in CGAL-6.1-Ic-232

